PR #101, opened 2022-12-09
Select profiles

edlerd:
Selector for one or multiple profiles, including order to define a hierarchy between chosen profiles.

### DIFF
--- a/static/js/components/ProfileSelect.tsx
+++ b/static/js/components/ProfileSelect.tsx
@@ -1,0 +1,166 @@
+import React, { FC, Fragment } from "react";
+import {
+  Button,
+  Col,
+  Icon,
+  Label,
+  List,
+  Row,
+} from "@canonical/react-components";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "../util/queryKeys";
+import { NotificationHelper } from "../types/notification";
+import { fetchProfiles } from "../api/profiles";
+import { LxdProfile } from "../types/profile";
+
+interface Props {
+  notify: NotificationHelper;
+  selected: string[];
+  setSelected: (profiles: string[]) => void;
+}
+
+const ProfileSelect: FC<Props> = ({ notify, selected, setSelected }) => {
+  const toggleProfile = (item: LxdProfile) => {
+    const newSelection = selected.includes(item.name)
+      ? selected.filter((i) => i !== item.name)
+      : [...selected, item.name];
+    setSelected(newSelection);
+  };
+
+  const moveUp = (item: LxdProfile, index: number) => {
+    if (index === 0) {
+      return;
+    }
+    const newSelection = [...selected];
+    newSelection.splice(index, 1);
+    newSelection.splice(index - 1, 0, item.name);
+    setSelected(newSelection);
+  };
+
+  const moveDown = (item: LxdProfile, index: number) => {
+    if (index === selected.length) {
+      return;
+    }
+    const newSelection = [...selected];
+    newSelection.splice(index, 1);
+    newSelection.splice(index + 1, 0, item.name);
+    setSelected(newSelection);
+  };
+
+  const { data: profiles = [], error } = useQuery({
+    queryKey: [queryKeys.profiles],
+    queryFn: fetchProfiles,
+  });
+
+  if (error) {
+    notify.failure("Could not load profiles.", error);
+  }
+
+  const unselected = profiles.filter(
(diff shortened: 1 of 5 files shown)


static/js/components/ProfileSelect.tsx
```diff
@@ -0,0 +1,168 @@
...
+                      </Button>
+                      <Button
+                        appearance="base"
+                        className="p-profile-select-button"
+                        dense
+                        hasIcon
+                        small
+                        type="button"
+                        disabled={i === selected.length - 1}
+                        onClick={() => moveDown(profile)}
```
lorumic, line 142: You could pass the `profile` (or just the profile's name) and the index `i` as a parameter to `moveUp` and `moveDown`, and then avoid an unnecessary search in the array inside the function.

static/js/components/ProfileSelect.tsx
```diff
@@ -0,0 +1,168 @@
...
+    if (index === selected.length) {
+      return;
+    }
+    const newSelection = [...selected];
+    newSelection.splice(index, 1);
+    newSelection.splice(index + 1, 0, item.name);
+    setSelected(newSelection);
+  };
+
+  const { data: projectProfiles = [], error } = useQuery({
```
lorumic, line 52: Why `projectProfiles` instead of `profiles`?

static/js/components/ProfileSelect.tsx
```diff
@@ -0,0 +1,168 @@
...
+          <hr />
+          <div className="p-profile-select-list">
+            <List
+              items={selected.map((name, i) => {
+                const profile = projectProfiles.find(
+                  (profile) => profile.name === name
+                );
+                if (!profile) {
+                  return null;
+                }
```
lorumic, line 112: Having `selected` as a string array is causing many unnecessary checks and type inconsistency between the `selected` and the `unselected` array - wouldn't it be better to have both as `LxdProfile` arrays instead? And then you map the array in the parent (`InstanceFormGuided`).

static/js/components/ProfileSelect.tsx
edlerd: There was some clash during dev with the profiles name, as it was used for other stuff. I'll simplify it now -- thats much better!

static/js/components/ProfileSelect.tsx
edlerd: I had the full profiles in the state before, but the problem is the initial state. It should be only the profile with name "default", and when I use strings I can just hardcode the name as init state. If it would be the object, I'd need to fetch it in the parent and select it.

static/js/components/ProfileSelect.tsx
edlerd: good idea

static/js/components/ProfileSelect.tsx
lorumic: Good point - hadn't noticed that. But is it safe to assume that a profile called `default` always exists? Because currently what you are doing is basically adding it blindly to the `selected` list with no checks about its existence. So, what I propose to do instead, is to move the assignment of the `default` profile to the `ProfileSelect` component, where you would need to fetch all the profiles anyway, and there, after checking that a profile called `default` exists, you add it to the `selected` list by default.

edlerd: Yeah it exists always and is always called default. There is no way to ask the default profile for a project. Because it is alwas a profile called default. You can't remove or rename it or set a new default profile.

edlerd: See the discussion at https://chat.canonical.com/canonical/pl/kkyq6d1zq3nadbae6nzgpiz1yr

lorumic: Then it's something that needs to be addressed, because currently you could delete the default profile. I just submitted https://github.com/canonical/lxd-ui/pull/117 for that.

Back on topic, I would propose one of the following:
- Moving the default assignment logic to the child (`ProfileSelect`)
- Keep the logic in the parent but create a mock `LxdProfile` object with only the name set to `default`

And then have `selected` as a `LxdProfile` array.

edlerd: Thanks for the #117 PR!

I think keeping the selected profiles as a string array with the names is reducing complexity.

lorumic: Ok, LGTM then!

lorumic (approved):

PR #102, opened 2022-12-09
Enhance API response models

lorumic:
Continues #95


static/js/types/common.d.ts
```diff
@@ -0,0 +1,27 @@
+export interface LxdConfigPair {
```
edlerd, line 1: Common is a place that tends to grow because it is not specific. What do you think about splitting this into `config` and `device` files?

edlerd (approved):

static/js/types/common.d.ts
lorumic: Sure, done @ [bebfcc3](https://github.com/canonical/lxd-ui/commit/bebfcc3ad5861579eb99377d4e5e112ec510c0fd)

PR #104, opened 2022-12-09
Allow reconnect terminal with custom parameters

edlerd:
this allows connection to alpine containers, that only have `sh` and no `bash`


static/js/modals/TerminalPayloadForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+        <Row>
+          <Col size={4}>
+            <Label>Environment variables</Label>
+          </Col>
+          <Col size={4}>Key</Col>
+          <Col size={4}>Value</Col>
+        </Row>
+        {formik.values.environment.map((variable, index) => (
+          <Row key={index}>
+            <Col size={4}> </Col>
```
lorumic, line 134: Since it's empty, this can be a self-closing tag: `<Col size={4} />`

static/js/modals/TerminalPayloadForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+                onBlur={formik.handleBlur}
+                onChange={formik.handleChange}
+                value={formik.values.environment[index].value}
+                stacked
+              />
+            </Col>
+          </Row>
+        ))}
+        <Row>
+          <Col size={4}></Col>
```
lorumic, line 158: Same here

static/js/modals/TerminalPayloadForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+            <Button
+              onClick={() =>
+                removeEnvironmentRow(formik.values.environment.length - 1)
+              }
+              type="button"
+              hasIcon
+            >
+              <Icon name="minus" />
+              <span>Remove variable</span>
+            </Button>
```
lorumic, line 173: This only allows the user to remove the last added variable, which is not ideal UX. What if I add a bunch of N variables and at some point, I realise that I don't want the variable at position N-10 anymore? I would need to copy the last added variable to replace the variable at N-10, and then remove the last entry: not very user-friendly.
I suggest to put an X/bin icon on the RH side of each entry to allow the user to delete that specific one.

lorumic (approved):

PR #108, opened 2022-12-13
add(storages): create storages list page, storage APIs and storage types

xlmnxp:
Signed-off-by: Salem Yaslem <s@sy.sa>

I cannot find Icon list to use for side-bar (Navigator)


static/js/components/Navigation.tsx
```diff
@@ -101,6 +101,16 @@ const Navigation: FC = () => {
                         Projects
                       </NavLink>
                     </li>
+                    <li className="p-side-navigation__item--title">
+                      <NavLink
+                        className="p-side-navigation__link"
+                        to="/storages"
+                        onClick={toggleMenu}
+                      >
+                        <i className="p-icon--switcher-environments is-light p-side-navigation__icon"></i>{" "}
```
edlerd, line 110: We can use the `p-icon--pods` icon for now. For new icons, ensure to include it in `styles.scss`.

I will request a dedicated icon for vanilla, but it may take some time to become available. Until then, let's use the pods icon. Thank you.

static/js/types/storage.d.ts
```diff
@@ -0,0 +1,12 @@
...
+    "volume.block.filesystem": string;
+    "volume.size": string;
+  };
+  description: string;
+  driver: string;
+  locations: string[];
+  name: string;
+  status: string;
+  used_by?: string[];
+}
```
edlerd, line 12: Sorry for the nitpick, there should be an empty line here.

I wonder why the linter doesn't catch this atm.

static/js/components/Navigation.tsx
edlerd: PS: We use vanilla as CSS framework. The available icons can be found here: https://vanillaframework.io/docs/patterns/icons

static/js/types/storage.d.ts
xlmnxp: @edlerd resolved

static/js/components/Navigation.tsx
xlmnxp: @edlerd resolved

xlmnxp:
@edlerd you can start workflow :+1:

xlmnxp:
@edlerd how to-do that? I don't think there issue with multiple commits because it will happen more often in future

edlerd:
> @edlerd how to do that? I don't think there issue with multiple commits because it will happen more often in future

It is explained here: https://www.internalpointers.com/post/squash-commits-into-one-git 
Ideally we have one commit here: https://github.com/canonical/lxd-ui/pull/108/commits and it can have the commit message of the first commit `add(storages): create storages list page, storage APIs and storage types`.

The idea is to have a clean history with not too many smallish commits.

xlmnxp:
@edlerd done :+1:

edlerd (approved):

PR #110, opened 2022-12-13
Move action buttons to controls area of instance overview

lorumic:


static/js/InstanceOverview.tsx
```diff
@@ -28,6 +34,47 @@ const InstanceOverview: FC<Props> = ({ instanceName }) => {
     queryFn: () => fetchInstance(instanceName),
   });
 
+  useEffect(() => {
+    if (instance) {
+      setControls(
+        <div style={{ display: "inline", marginRight: "1rem" }}>
```
edlerd, line 40: We should avoid adding CSS in JS. Better fix it with a vanilla class.

I think the problem is the tooltip in the delete button. So you could give the tooltip the right utility class from vanilla, to get correct spacing.

edlerd: PS: Not sure vanilla has utility classes for horizontal spacing. Maybe ask the vanilla pros for advice.

anthonydillon: No, this is generally handled by the grid.

static/js/InstanceOverview.tsx
```diff
@@ -28,6 +34,47 @@ const InstanceOverview: FC<Props> = ({ instanceName }) => {
     queryFn: () => fetchInstance(instanceName),
   });
 
+  useEffect(() => {
+    if (instance) {
+      setControls(
+        <div className="u-instance-overview-controls">
```
edlerd, line 40: This is inline by default. So we can use an empty fragment here as a wrapper. And clean-up the `u-instance-overview-controls` class.

edlerd (approved):

static/js/InstanceOverview.tsx
lorumic: Very strange. I tried that approach before first, and it wasn't working. It was causing buttons to be stacked vertically. That's why I had applied that style. Anyway, now that I tried again, it works as you said. Done @ [1ea1594](https://github.com/canonical/lxd-ui/commit/1ea159429bc0928d054872d87ec3c930dbe025d7)

PR #111, opened 2022-12-13
Hide XTerm when terminal session could not be open

lorumic:


xlmnxp:
@lorumic Is there placeholder for it when empty?

PR #112, opened 2022-12-13
Extended profile creation form with additional fields [WD-874]

lorumic:


static/js/components/DiskDeviceSelector.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { fetchStorages } from "../api/storages";
+
+interface Props {
+  notify: NotificationHelper;
+  setDiskDevice: Dispatch<SetStateAction<LxdDiskDevice | null>>;
+}
+
+const DiskDeviceSelector: FC<Props> = ({ notify, setDiskDevice }) => {
+  const [pool, setPool] = useState<string>("");
+  const [path, setPath] = useState<string>("");
```
edlerd, line 22: Just like the LimitSelector, this component can be stateless, when the parent passes in the diskDevice as a prop.

static/js/components/NicDeviceSelector.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { LxdNicDevice } from "../types/device";
+
+interface Props {
+  notify: NotificationHelper;
+  setNicDevice: Dispatch<SetStateAction<LxdNicDevice | null>>;
+}
+
+const NicDeviceSelector: FC<Props> = ({ notify, setNicDevice }) => {
+  const [nic, setNic] = useState<string>("");
+  const [nameInInstance, setNameInInstance] = useState<string>("");
```
edlerd, line 22: Same idea with state here: parent can pass in nicDevice and this component could become stateless and much simpler. I.e. the updateNic can become a one liner with spreading: setNicDevice({...nicDevice, network: e.target.value})

static/js/panels/ProfileForm.tsx
```diff
@@ -75,6 +132,64 @@ const ProfileForm: FC = () => {
...
+                          setMeasure={setMeasure}
+                        />
+                      </>
+                    ),
+                    title: "Limits",
+                  },
+                  {
+                    content: (
+                      <>
+                        <CloudInitConfig
```
edlerd, line 172: Couldn't we have just one input for the cloud init yaml with the three blocks prefilled but empty?

static/js/components/NicDeviceSelector.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <Select
+        name="nic"
+        label="Select a network device (NIC)"
+        options={getNetworkOptions()}
+        onChange={updateNic}
+        value={nic}
+        stacked
+      />
+      <Input
+        id="nameInInstance"
```
edlerd, line 79: Shall we add a placeholder with a hint what possible values look like? I.e. eth0 or what is commonly used

static/js/components/DiskDeviceSelector.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <h4>Root disk device</h4>
+      <Select
+        name="pool"
+        label="Select a storage pool"
+        options={getStoragePoolOptions()}
+        onChange={updatePool}
+        value={pool}
+        stacked
+      />
+      <Input
```
edlerd, line 81: We might want to hardcode this to / and then we won't need an input for the path.

static/js/panels/ProfileForm.tsx
```diff
@@ -75,6 +132,64 @@ const ProfileForm: FC = () => {
...
+                          setNicDevice={setNicDevice}
+                        />
+                        <hr />
+                        <DiskDeviceSelector
+                          notify={notify}
+                          setDiskDevice={setDiskDevice}
+                        />
+                      </>
+                    ),
+                    title: "Devices",
```
edlerd, line 152: Wouldn't it make more sense to split them into two sections storage and network?

static/js/panels/ProfileForm.tsx
```diff
@@ -28,7 +43,49 @@ const ProfileForm: FC = () => {
     },
     validationSchema: ProfileSchema,
     onSubmit: (values) => {
-      createProfile(values.name, values.description)
+      // create LxdDevices object
+      const devicesObj: LxdDevices = {};
+      if (nicDevice?.network) {
+        devicesObj[nicDevice.name === "" ? nicDevice.network : nicDevice.name] =
```
edlerd, line 49: Why fall back to the network name as device name? This is a little wired, because the front-end doesn't explain this behaviour.

static/js/components/LimitSelector.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <>
+      <Input
+        id="cpuLimit"
+        name="cpuLimit"
+        type="number"
+        label="CPU limit"
+        onChange={updateCpuLimit}
+        value={cpuLimit}
+        stacked
```
edlerd, line 41: Can we add a help text for this ... Is it a value between 1 and 100 or 0 and 1?

static/js/components/LimitSelector.tsx
```diff
@@ -0,0 +1,77 @@
...
+      />
+      <Input
+        id="memoryLimit"
+        name="memoryLimit"
+        type="number"
+        label="Memory limit"
+        onChange={updateMemoryLimit}
```
edlerd, line 48: I think we should simplify this so the parent always receives the memory limit in MiB, and passes in the value as MiB. Only in here we have the additional unit selector. On change and when receiving the limit from the parent we convert to MiB. Hence the parent has not to deal with the unit complexity, we rather keep it local.

static/js/panels/ProfileForm.tsx
lorumic: We could, but I don't think we should. They have separate config keys, so grouping them in one single `YamlEditor` means that we should then build a parser to split them when we build the `config` object. Not optimal and very error prone. And user-wise it's also not desirable, because it confuses the user by making them think that they are a single block of config, while they aren't. They are separate configs that can be set or not independently, and I think this should be apparent to the user.

static/js/panels/ProfileForm.tsx
lorumic: It would make sense to do so if we allowed the user to add multiple networks and/or multiple storage volumes, but since now they are the only 2 options, I think grouping them into one "devices" section is better than having 2 sections with only 1 element inside each.

static/js/panels/ProfileForm.tsx
lorumic: Seemed like a reasonable default behaviour, considering that it is going to be applied only if the user has selected a network from the list. Of course any alternative proposal is always welcome. As for the lack of explanation... Well, also when you create an instance and leave the name field blank you find yourself with an instance named `freaky-panda` with no explanation anywhere... ;)

static/js/components/DiskDeviceSelector.tsx
lorumic: Sure, just one question: should this still be shown to the user (e.g. as a disabled `Input`, or in any other way), so that they are aware of the `path` value that is being applied, or should it be completely transparent since they cannot modify it anyway? Which approach is better?

edlerd: I think we should simplify the form and remove the field.

static/js/panels/ProfileForm.tsx
edlerd: True, let's add explanatory text in both places. I think the help prop on the Input component is a good place for this. Though it is only useful to the user once. Maybe let's discuss it with UX.

static/js/panels/ProfileForm.tsx
```diff
@@ -75,6 +132,64 @@ const ProfileForm: FC = () => {
                 }
                 stacked
               />
+              <Accordion
+                titleElement="h3"
```
edlerd, line 136: I think h3 is a little too big here. Maybe try h4 or h5? Or one of the small text utility classes.

static/js/panels/ProfileForm.tsx
lorumic: I was thinking that maybe it could go in the `placeholder` as well, together with the example (e.g. eth0), so that it doesn't take further space. What do you think?

static/js/components/LimitSelector.tsx
lorumic: Looking into this, I just found out that this field should be able to accommodate either:

- a number, or
- a range of numbers, or
- a set of comma-separated integers

So a simple numeric input is not good, see: https://linuxcontainers.org/lxd/docs/master/reference/instance_options/#cpu-limits

I will use a `text` input and validate its content with regular expressions.

edlerd: Maybe three radio buttons to select the type of limit and a changing input based on which option was selected is better here?

edlerd: > a set of comma-separated integers
This could be checkboxes for each core that you can enable/disable

static/js/components/RootDiskDeviceSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { LxdDiskDevice } from "../types/device";
+import { fetchStorages } from "../api/storages";
+
+interface Props {
+  notify: NotificationHelper;
+  rootDiskDevice: LxdDiskDevice;
+  setRootDiskDevice: Dispatch<SetStateAction<LxdDiskDevice>>;
+}
+
+const RootDiskDeviceSelector: FC<Props> = ({
```
edlerd, line 15: Can we shorten this to `RootDiskSelector`. Actually the component has no notion of root, maybe `StorageSelector`?

static/js/components/NicDeviceSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { queryKeys } from "../util/queryKeys";
+import { LxdNicDevice } from "../types/device";
+
+interface Props {
+  notify: NotificationHelper;
+  nicDevice: LxdNicDevice;
+  setNicDevice: Dispatch<SetStateAction<LxdNicDevice>>;
+}
+
+const NicDeviceSelector: FC<Props> = ({ notify, nicDevice, setNicDevice }) => {
```
edlerd, line 15: How about `NetworkSelector` or `NicSelector`?

static/js/components/NicDeviceSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+        value={nicDevice.network}
+        stacked
+      />
+      <Input
+        id="nameInInstance"
+        name="nameInInstance"
+        type="text"
+        label="Name in instance"
+        placeholder="e.g. eth0"
+        help="The network name will be used if left blank."
```
edlerd, line 59: I'd shorten it: "Network name will be used if left blank"

static/js/components/LimitSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+const LimitSelector: FC<Props> = ({
+  cpuLimit,
+  setCpuLimit,
+  memoryLimit,
+  setMemoryLimit,
+}) => {
+  const cpuLimitHelp = (
+    <>
+      See{" "}
+      <a href="https://linuxcontainers.org/lxd/docs/latest/reference/instance_options/#cpu-limits">
```
edlerd, line 21: Should add `target="_blank" rel="noreferrer"` to the link.

We should change the link text, see https://heyoka.medium.com/dont-use-click-here-f32f445d1021
Maybe "See docs for the allowed formats." and link on the "allowed formats"?

static/js/components/LimitSelector.tsx
lorumic: Ok, the limits should be fixed too @ [888ddae](https://github.com/canonical/lxd-ui/commit/888ddaeaca03ce4bf85a1d7ea782c0b4281e4840)

static/js/panels/ProfileForm.tsx
```diff
@@ -11,11 +11,35 @@ import { createProfile } from "../api/profiles";
...
+  DEFAULT_DISK_DEVICE,
+  DEFAULT_MEM_LIMIT,
+  DEFAULT_NIC_DEVICE,
+} from "../util/defaults";
 
 const ProfileForm: FC = () => {
   const navigate = useNavigate();
   const notify = useNotification();
   const queryClient = useQueryClient();
+  const [nicDevice, setNicDevice] = useState<LxdNicDevice>(DEFAULT_NIC_DEVICE);
```
edlerd, line 34: We can use formik to store the state. It is intended for it and we don't need to define state setter/getter next to it.

Below an example how to use formik for the state of the nicDevice.

```
Subject: [PATCH] fix(instance) simplify instance actions
 - use portal to set actions on instance detail page
 - close instance edit stays on current page
 - extract create snapshot btn
---
Index: static/js/components/NetworkSelector.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/static/js/components/NetworkSelector.tsx b/static/js/components/NetworkSelector.tsx
--- a/static/js/components/NetworkSelector.tsx	(revision 706eb18f5a4b51c0b994db09c7217d8886974d55)
+++ b/static/js/components/NetworkSelector.tsx	(date 1671110410115)
@@ -9,7 +9,7 @@
 interface Props {
   notify: NotificationHelper;
   nicDevice: LxdNicDevice;
-  setNicDevice: Dispatch<SetStateAction<LxdNicDevice>>;
+  setNicDevice: (nic: LxdNicDevice) => void;
 }
 
 const NetworkSelector: FC<Props> = ({ notify, nicDevice, setNicDevice }) => {
Index: static/js/panels/ProfileForm.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/static/js/panels/ProfileForm.tsx b/static/js/panels/ProfileForm.tsx
--- a/static/js/panels/ProfileForm.tsx	(revision 706eb18f5a4b51c0b994db09c7217d8886974d55)
+++ b/static/js/panels/ProfileForm.tsx	(date 1671110547987)
@@ -12,7 +12,7 @@
 import { queryKeys } from "../util/queryKeys";
 import SubmitButton from "../buttons/SubmitButton";
 import NetworkSelector from "../components/NetworkSelector";
-import { LxdDevices, LxdDiskDevice, LxdNicDevice } from "../types/device";
+import { LxdDevices, LxdDiskDevice } from "../types/device";
 import StorageSelector from "../components/StorageSelector";
 import CpuLimitSelector from "../components/CpuLimitSelector";
 import CloudInitConfig from "../components/CloudInitConfig";
@@ -31,7 +31,6 @@
   const notify = useNotification();
-  const [nicDevice, setNicDevice] = useState<LxdNicDevice>(DEFAULT_NIC_DEVICE);
   const [rootDiskDevice, setRootDiskDevice] =
     useState<LxdDiskDevice>(DEFAULT_DISK_DEVICE);
   const [cpuLimit, setCpuLimit] = useState<CpuLimit>(DEFAULT_CPU_LIMIT);
@@ -49,14 +48,16 @@
     initialValues: {
       name: "",
       description: "",
+      cpuLimit: DEFAULT_CPU_LIMIT,
+      nicDevice: DEFAULT_NIC_DEVICE,
     },
     validationSchema: ProfileSchema,
-    onSubmit: (values) => {
+    onSubmit: ({ name, description, cpuLimit, nicDevice }) => {
       // create LxdDevices object
       const devicesObj: LxdDevices = {};
       if (nicDevice.network) {
-        devicesObj[nicDevice.name === "" ? nicDevice.network : nicDevice.name] =
-          nicDevice;
+        const name = nicDevice.name === "" ? nicDevice.network : nicDevice.name;
+        devicesObj[name] = nicDevice;
       }
       if (rootDiskDevice.pool) {
         devicesObj.root = rootDiskDevice;
@@ -111,9 +112,9 @@
       // build final LxdProfile object with all the data
       const profile: LxdProfile = {
         config: configObj,
-        description: values.description,
+        description: description,
         devices: devicesObj,
-        name: values.name,
+        name: name,
       };
 
       createProfile(profile)
@@ -170,8 +171,10 @@
                     content: (
                       <NetworkSelector
                         notify={notify}
-                        nicDevice={nicDevice}
-                        setNicDevice={setNicDevice}
+                        nicDevice={formik.values.nicDevice}
+                        setNicDevice={(v) =>
+                          void formik.setFieldValue("nicDevice", v)
+                        }
                       />
                     ),
                     title: "Network",
```

static/js/components/MemoryLimitSelector.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return (
+    <>
+      <Label>Memory limit</Label>
+      <RadioInput
+        label="fixed (bytes)"
+        checked={memoryLimit.selectedType === MEM_LIMIT_TYPE.FIXED}
+        onChange={() =>
+          setMemoryLimit({
+            ...DEFAULT_MEM_LIMIT,
```
edlerd, line 54: I think the fields in the `MemoryLimit` apart from type should be optional. Then this call is simplified to `setMemoryLimit({ selectedType: MEM_LIMIT_TYPE.FIXED })`. So we don't need to spread the default with null values into it.

This also applies to the CpuLimit radio options.

static/js/types/limits.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export enum CPU_LIMIT_TYPE {
+  DYNAMIC = 0,
+  FIXED_RANGE = 1,
+  FIXED_SET = 2,
+}
+
+export interface CpuLimit {
+  dynamicValue: number | null;
```
edlerd, line 35: I think those (and MemoryLimit above) should be optional, as in `dynamicValue?: number`.

static/js/components/MemoryLimitSelector.tsx
lorumic: If we don't spread the default null values, the old value will be kept. This will result in a dirty object and possibly wrong values being applied, at least for the `MemoryLimit`. Unless we differentiate explicitly the value field for the 2 cases, setting a `percentageValue` field and a `bytesValue` field.

static/js/panels/ProfileForm.tsx
```diff
@@ -28,7 +52,71 @@ const ProfileForm: FC = () => {
...
+        devicesObj[nicDevice.name === "" ? nicDevice.network : nicDevice.name] =
+          nicDevice;
+      }
+      if (rootDiskDevice.pool) {
+        devicesObj.root = rootDiskDevice;
+      }
+
+      // create LxdConfigPair object for limits
+      const limits: LxdConfigPair = {};
+      switch (cpuLimit.selectedType) {
```
edlerd, line 67: I'd push this complexity into a function to  make it more readable:

```
Index: static/js/panels/ProfileForm.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/static/js/panels/ProfileForm.tsx b/static/js/panels/ProfileForm.tsx
--- a/static/js/panels/ProfileForm.tsx	(revision 706eb18f5a4b51c0b994db09c7217d8886974d55)
+++ b/static/js/panels/ProfileForm.tsx	(date 1671111390214)
@@ -27,6 +27,27 @@
   DEFAULT_NIC_DEVICE,
 } from "../util/defaults";
 
+const getCpuLimit = (cpuLimit: CpuLimit) => {
+  switch (cpuLimit.selectedType) {
+    case CPU_LIMIT_TYPE.DYNAMIC:
+      if (cpuLimit.dynamicValue) {
+        return `${cpuLimit.dynamicValue}`;
+      }
+      return null;
+    case CPU_LIMIT_TYPE.FIXED_RANGE:
+      if (cpuLimit.rangeValue?.from && cpuLimit.rangeValue.to) {
+        return `${cpuLimit.rangeValue.from}-${cpuLimit.rangeValue.to}`;
+      }
+      return null;
+    case CPU_LIMIT_TYPE.FIXED_SET:
+      if (cpuLimit.setValue) {
+        return cpuLimit.setValue;
+      }
+      return null;
+  }
+  return null;
+};
+
 const ProfileForm: FC = () => {
   const notify = useNotification();
@@ -64,31 +85,11 @@
 
       // create LxdConfigPair object for limits
       const limits: LxdConfigPair = {};
-      switch (cpuLimit.selectedType) {
-        case CPU_LIMIT_TYPE.DYNAMIC:
-          if (cpuLimit.dynamicValue) {
-            limits["limits.cpu"] = `${cpuLimit.dynamicValue}`;
-          }
-          break;
-        case CPU_LIMIT_TYPE.FIXED_RANGE:
-          if (!cpuLimit.rangeValue) break;
-          if (
-            cpuLimit.rangeValue.from !== null &&
-            cpuLimit.rangeValue.to !== null
-          ) {
-            limits[
-              "limits.cpu"
-            ] = `${cpuLimit.rangeValue.from}-${cpuLimit.rangeValue.to}`;
-          }
-          break;
-        case CPU_LIMIT_TYPE.FIXED_SET:
-          if (cpuLimit.setValue) {
-            limits["limits.cpu"] = cpuLimit.setValue;
-          }
-          break;
-        default:
-          break;
+      const limitsCpu = getCpuLimit(cpuLimit);
+      if (limitsCpu) {
+        limits["limits.cpu"] = limitsCpu;
       }
+
       if (memoryLimit.value) {
         limits["limits.memory"] = `${memoryLimit.value}${memoryLimit.unit}`;
       }
```

edlerd: Little bug in there (also in the patch I proposed): The setValue can be a single number and needs to be turned into something like 1-1 in that case.

static/js/components/MemoryLimitSelector.tsx
edlerd: After a `setMemoryLimit({ selectedType: MEM_LIMIT_TYPE.FIXED })` the `memoryLimit` object from upstream will not have a `value` property. I don't see how the old value will be kept?

lorumic: Right, I missed the fact that you were passing a whole object there.

edlerd: But there is a problem with the memory type, because we'd fall back to bits and want GiB to be selected, which comes from the default spread. I think we should make it explicit in this case, so when fixed is selected, we also set the unit to GiB.

static/js/panels/ProfileForm.tsx
lorumic: Good idea to move the `switch` to a function, but your patch actually introduces a bug: when the value in the `from`/`to` field is set to 0 (which is a possible and allowed value for these fields), the expression `if (cpuLimit.rangeValue?.from && cpuLimit.rangeValue.to)` will equal to `false` and will return `null`, which is wrong.
In other words: I wasn't putting the explicit `null` checks without a reason. ;)

edlerd: Right.

lorumic:
All solved @ [4b00f28](https://github.com/canonical/lxd-ui/commit/4b00f2885898777640a0d49c57bd3047f9f74e94)

static/js/panels/ProfileForm.tsx
```diff
@@ -11,6 +11,49 @@ import { createProfile } from "../api/profiles";
...
+      if (
+        cpuLimit.rangeValue.from !== null &&
+        cpuLimit.rangeValue.to !== null
+      ) {
+        return `${cpuLimit.rangeValue.from}-${cpuLimit.rangeValue.to}`;
+      }
+      return null;
+    case CPU_LIMIT_TYPE.FIXED_SET:
+      if (cpuLimit.setValue) {
+        if (cpuLimit.setValue.length > 1) {
```
edlerd, line 48: I think we need to look for commas, because the value could be a multiple digit one.

lorumic: Ok, is it ok to keep it simple, without crazy validation through regex? (which I'm trying to avoid, at least at this stage)

lorumic: Done @ [86a81f1](https://github.com/canonical/lxd-ui/commit/86a81f106d7d3fceca7bd45be2f5972edebc365a)

static/js/panels/ProfileForm.tsx
```diff
@@ -25,10 +68,70 @@ const ProfileForm: FC = () => {
...
+      description,
+      nicDevice,
+      rootDiskDevice,
+      cpuLimit,
+      memoryLimit,
+      userDataConfig,
+      vendorDataConfig,
+      networkConfig,
+    }) => {
+      // create LxdDevices object
```
edlerd, line 91: I'd propose to simplify this a little bit:

```
      const config: LxdConfigPair = {};
      const limitsCpu = getCpuLimit(cpuLimit);
      if (limitsCpu) {
        config["limits.cpu"] = limitsCpu;
      }
      if (memoryLimit.value) {
        config["limits.memory"] = `${memoryLimit.value}${memoryLimit.unit}`;
      }
      if (userDataConfig.trim()) {
        config["cloud-init.user-data"] = `|\n${userDataConfig.trim()}`;
      }
      if (vendorDataConfig.trim()) {
        config["cloud-init.vendor-data"] = `|\n${vendorDataConfig.trim()}`;
      }
      if (networkConfig.trim()) {
        config["cloud-init.network-config"] = `|\n${networkConfig.trim()}`;
      }

      const devices: LxdDevices = {};
      if (nicDevice.network) {
        const name = nicDevice.name === "" ? nicDevice.network : nicDevice.name;
        devices[name] = nicDevice;
      }
      if (rootDiskDevice.pool) {
        devices.root = rootDiskDevice;
      }

      const profile: LxdProfile = {
        config,
        description,
        devices,
        name,
      };
```

static/js/panels/ProfileForm.tsx
edlerd: Yeah I think this is fine.

edlerd (approved):

PR #113, opened 2022-12-14
Simplify instance actions

edlerd:
- use portal to set actions on instance detail page
 - close instance edit stays on current page
 - extract create snapshot btn

### DIFF
--- a/static/js/InstanceOverview.tsx
+++ b/static/js/InstanceOverview.tsx
@@ -1,10 +1,4 @@
-import React, {
-  Dispatch,
-  FC,
-  ReactNode,
-  SetStateAction,
-  useEffect,
-} from "react";
+import React, { FC } from "react";
 import { fetchInstance } from "./api/instances";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "./util/queryKeys";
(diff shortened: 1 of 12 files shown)


static/js/buttons/instances/CreateSnapshotBtn.tsx
```diff
@@ -0,0 +1,36 @@
...
+  instance?: LxdInstance;
+  notify: NotificationHelper;
+}
+
+const CreateSnapshotBtn: FC<Props> = ({ instance, notify }) => {
+  const [isModal, setModal] = useState(false);
+
+  return (
+    <>
+      {instance && isModal && (
```
lorumic, line 17: I think checking that `instance` is not `undefined` should not be done here, as we shouldn't even have this rendered if `instance` is not defined. We also do the same in other places I think, check in the parent and then pass a clean, "defined" prop to the children.

static/js/InstanceOverview.tsx
```diff
@@ -93,10 +48,31 @@ const InstanceOverview: FC<Props> = ({ instanceName, setControls }) => {
...
+      {controlTarget &&
+        createPortal(
+          <>
+            <StartInstanceBtn label="Start" {...btnProps} />
+            <StopInstanceBtn label="Stop" {...btnProps} />
+            <EditInstanceBtn label="Edit" {...btnProps} />
+            <DeleteInstanceBtn
+              onFinish={() => navigate("/instances")}
+              {...btnProps}
+            />
```
lorumic, line 71: This is missing the `label="Delete"` prop.

static/js/modals/CreateSnapshotForm.tsx
```diff
@@ -79,28 +79,25 @@ const CreateSnapshotForm: FC<Props> = ({ instance, close, notify }) => {
...
+      buttonRow={
+        <>
+          <Button className="u-no-margin--bottom" type="button" onClick={close}>
+            Cancel
+          </Button>
+          <SubmitButton
+            isSubmitting={formik.isSubmitting}
+            isDisabled={!formik.isValid}
+            buttonLabel="Create snapshot"
+            onClick={() => void formik.submitForm}
```
lorumic, line 94: Why the `Modal` outside the `Form`? We could have avoided the redirection of the `click` event.

edlerd: It creates a visual bug, the CreateSnapshot Button goes into its own row when the modal is wrapped in a form.

PR #119, opened 2022-12-16
Fix profile limits

lorumic:
- [adjust max memory value (fixed) based on the selected unit](https://github.com/canonical/lxd-ui/pull/119/commits/30f1672265c1f6276221b8990af57145b9bf54ee)
- [add info about total values for cpu and memory limits](https://github.com/canonical/lxd-ui/pull/119/commits/3e110befce119c5567cf8303a16cac170dad4b41)

### DIFF
--- a/static/js/components/MemoryLimitSelector.tsx
+++ b/static/js/components/MemoryLimitSelector.tsx
@@ -11,7 +11,7 @@ import { useQuery } from "@tanstack/react-query";
 import { fetchResources } from "../api/server";
 import { queryKeys } from "../util/queryKeys";
 import { NotificationHelper } from "../types/notification";
-import { MEM_UNITS, MemoryLimit, MEM_LIMIT_TYPE } from "../types/limits";
+import { BYTES_UNITS, MemoryLimit, MEM_LIMIT_TYPE } from "../types/limits";
 import { DEFAULT_MEM_LIMIT } from "../util/defaults";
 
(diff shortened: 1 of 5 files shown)


static/js/components/MemoryLimitSelector.tsx
```diff
@@ -97,14 +130,21 @@ const MemoryLimitSelector: FC<Props> = ({
...
               }
               value={memoryLimit.unit}
             />
           </Col>
         </Row>
       )}
+      {maxMemory && (
+        <p>
+          <span>
+            Total memory: <b>{humanFileSize(maxMemory, true)}</b>
```
lorumic, line 144: I don't know if this value should also be formatted according to the currently selected unit (only for the `fixed (bytes)` case), what do you think? @edlerd

edlerd (approved):
LGTM, up to you if you want to add the total memory formatting according to current selected unit.

static/js/components/MemoryLimitSelector.tsx
lorumic: Ok, done, but please check once again because I'm not sure I'm doing it the best way @ [0c1ec36](https://github.com/canonical/lxd-ui/commit/0c1ec3671413bf9b746bfc20e1931c167231b1f8) 
If it looks good, you can merge. Otherwise, I'm happy to improve it. :)

PR #121, opened 2022-12-16
Add profile details page [WD-873]

lorumic:


static/js/buttons/profiles/DeleteProfileBtn.tsx
```diff
@@ -8,9 +8,20 @@ import { deleteProfile } from "../../api/profiles";
...
+  isDense?: boolean;
+  label?: string;
 }
 
-const DeleteProfileBtn: FC<Props> = ({ name, notify }) => {
+const DeleteProfileBtn: FC<Props> = ({
+  name,
+  notify,
+  appearance = "base",
+  className = "p-contextual-menu__link",
```
edlerd, line 21: I think the button props can be unified and simplified. In the current situation handling the possible customisations and settings is getting a bit too noisy.

How about a single prop on all *Btn classes with name btnStyle. It is an object with all props from Button component in react-conponents and typed accordingly. In the *Btn component we just spread that object into the Button. Hence all our *Btn can be fully customized in the use case.

Also we can have some presets for the btnStyle that can be imported from any use case. So we don't repeat the btnProps object in our components, but just import it.

Does it make sense?

static/js/ProfileDetail.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const { data: profile, error } = useQuery({
+    queryKey: [queryKeys.profiles, name],
+    queryFn: () => fetchProfile(name),
+  });
+
+  if (error) {
+    notify.failure("Could not load profile details.", error);
+  }
+
+  if (!profile) {
```
edlerd, line 38: I think we can also end in this branch, when the query is loading. Generally we miss the loading state in our components.

static/js/ProfileDetail.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  const btnProps = {
+    profile: profile,
+    notify: notify,
+    appearance: "",
+    className: "u-no-margin--bottom",
+    isDense: false,
+  };
+
+  const containsDevices = (type: "nic" | "disk") => {
```
edlerd, line 50: I think we could benefit from using the same approach as in this PR with isFooDevice helpers: https://github.com/canonical/lxd-ui/pull/120/files#diff-37a7120764c7cadda0eec0b5bf2b8783704700be479fb051f60a8d751382c69a

But that is not in Main yet....

static/js/buttons/profiles/DeleteProfileBtn.tsx
lorumic: Sure, nice idea. I can submit a dedicated PR to implement this over all the *Btn components.
But yeah, I also agree that it is out of scope for this PR. So I will mark it as solved even though I have not applied any change to the code in this regard.

static/js/ProfileDetail.tsx
lorumic: Right. I remembered that `useQuery` gives us the `isLoading` flag, so I used it to differentiate the loading case from the case where the profile details could not be loaded.
It is a temporary solution, and we should discuss how to visually present the loading state to the user in a nice way, but I think that for the scope of this PR it should be enough - and will also serve as a reminder for a future PR to generalise and improve this behaviour.

static/js/ProfileDetail.tsx
lorumic: I added those helper functions and replaced `containsDevices` and `getDevices` with them @ [67c3f03](https://github.com/canonical/lxd-ui/commit/67c3f037beb960c5ca2c14b7ee095e108bda37de)

static/js/ProfileDetail.tsx
```diff
@@ -0,0 +1,232 @@
...
+              <th>Instances using this profile</th>
+              <td>{getUsedByNames()?.join(", ")}</td>
+            </tr>
+            {Object.values(profile.devices).some((device) =>
+              isNicDevice(device)
+            ) && (
+              <tr>
+                <th>Network devices</th>
+                <td>
+                  {Object.values(profile.devices)
```
edlerd, line 108: In my opinion, it is better to always show the rows with headings and not filter them out if there is no content. The content in that case can be "none" or just "-". The reason I think this is better, is users remembering the place where to find an information if the row with heading is always there. If the row is sometimes there and sometimes not, users tend to be confused or at least they don't get the mental model of finding this information in that place as fast as in the other approach.

static/js/ProfileDetail.tsx
```diff
@@ -0,0 +1,232 @@
...
+            <tr>
+              <th>Name</th>
+              <td>{profile.name}</td>
+            </tr>
+            <tr>
+              <th>Description</th>
+              <td>{profile.description}</td>
+            </tr>
+            <tr>
+              <th>Instances using this profile</th>
```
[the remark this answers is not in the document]
lorumic: I also recalled this information from our discussions, but then I could not find any evidence of this in the docs. It is my understanding that profiles can be applied to instances only, so representing something different in the UI would contradict what the user can find in the docs.

Or can you link me any reference in the docs where it says that a profile can be applied to anything that is not an instance?

For the link: sure, just keep in mind that the same shouldr be done in the instance overview tab, with profiles. I can take care of that in a future PR.

static/js/ProfileDetail.tsx
lorumic: Sure, good point. I will fix that. I am also already using this approach for the `used_by` information after all.

static/js/ProfileDetail.tsx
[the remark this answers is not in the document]
edlerd: True, would be good to link both ways.
Regarding entities, it might be undocumented. I remember it was mentioned that it actually not a common use case to link profiles to images, but to support windows it is used for attaching the TPM device. In that case the UI would be wired, because the heading says "Instances" but the entity is not an instance.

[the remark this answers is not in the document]
lorumic: Ok, but I think the point that we need to understand here is: does the back-end ever return anything that is not an instance in the `used_by` field here? Only in that case it would make sense to adapt the UI accordingly. It might be the case that, even if you can somehow link a profile to an image, that would not be returned in this list here.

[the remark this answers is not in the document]
edlerd: You might be right, this seems like an instance only used by at https://github.com/lxc/lxd/blob/master/lxd/profiles.go#L197

Let's keep it as is and remember to double check on the next core sync.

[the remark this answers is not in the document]
lorumic: Added links to instance details page @ [3b362fd](https://github.com/canonical/lxd-ui/commit/3b362fdf2a1568565b63d47c47b20775395fa671)

static/js/ProfileDetail.tsx
```diff
@@ -0,0 +1,232 @@
...
+              <td>{getUsedByNames()?.join(", ")}</td>
+            </tr>
+            {Object.values(profile.devices).some((device) =>
```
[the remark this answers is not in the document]
lorumic: Fixed @ [3b362fd](https://github.com/canonical/lxd-ui/commit/3b362fdf2a1568565b63d47c47b20775395fa671)

static/js/ProfileDetail.tsx
lorumic: Fixed @ [3b362fd](https://github.com/canonical/lxd-ui/commit/3b362fdf2a1568565b63d47c47b20775395fa671)

edlerd (approved):
LGTM, let's remember to
 1. ask about the used by with core team in sync
 2. make a nice and reusable loading / error handling. Maybe vanilla has something at hand? If not Yurii has something in Anbox we might want to use.
 3. Think about improving Btn components. I started a local pr, but it's far from nice or complete.

PR #131, opened 2023-01-05
Add e2e tests

edlerd:


tests/instances.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  await deleteInstance(page, instance);
+});
+
+test("snapshot create restore and remove ", async ({ page }) => {
+  const instance = randomInstanceName();
+  const snapshot = randomSnapshotName();
+  await createInstance(page, instance);
+  await openInstanceDetail(page, instance);
```
lorumic, line 14: See the other comments: with this, we are opening the instance detail twice: first by click (in `createInstance`) and then by navigation (in `openInstanceDetail`).

tests/instances.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    .getByRole("link", { name: instanceName })
+    .click();
+};
+
+const deleteInstance = async (page: Page, instanceName: string) => {
+  await openInstanceDetail(page, instanceName);
+  await page.waitForTimeout(10000).then(async () => {
+    await page.keyboard.down("Shift");
+    await page.getByRole("button", { name: "Delete Delete" }).click();
+    await page.keyboard.up("Shift");
```
lorumic, line 72: Unlike all the other confirmation cases, here we are confirming the action with the `Shift` shortcut. Should the function be called `deleteInstanceNoConfirm` to make it clear?

tests/instances.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    await page.waitForSelector("text=Add instance", TEXT_TIMEOUT);
```
lorumic, line 73: I would make this behaviour consistent with the profile delete test. Either always check the selector, or always check the page URL.

edlerd:
Addressed all the issues and change it, so we have a shared instance for the snapshot tests.
@lorumic can you try running with `workers: 3` in the `playwright.config.ts` if it succeeds for you as well?

lorumic:
> @lorumic can you try running with `workers: 3` in the `playwright.config.ts` if it succeeds for you as well?

It's much better than before. With 3 workers each single test takes longer but in total it takes around 10 seconds less. I would still keep the single worker because with that every test completes in less than 30 seconds, whereas with the 3 workers the "instance create and remove" took 40-55 seconds.

lorumic (approved):
Good, I like the idea of having an instance created in the initial setup that can be used for the snapshot tests. This allowed to complete them in few seconds.

PR #132, opened 2023-01-09
Remove actions from profile list view, simplify profile actions

edlerd:


static/js/pages/profiles/actions/EditProfileBtn.tsx
```diff
@@ -11,25 +11,17 @@ interface Props {
   label?: string;
```
lorumic, line 11: We are left with unused props in the interface up here.

edlerd: missed that one, removed now.

lorumic (approved):

PR #134, opened 2023-01-09
Switch to start/stop instance [WD-1035]

lorumic:


static/js/pages/instances/InstanceDetail.tsx
```diff
@@ -21,6 +29,27 @@ const InstanceDetail: FC = () => {
...
+    queryKey: [queryKeys.instances, name],
+    queryFn: () => fetchInstance(name),
+  });
+
+  if (error) {
+    notify.failure("Could not load instance details.", error);
+  }
+
+  if (isLoading) {
+    return <Loader text="Loading instance details..." />;
```
edlerd, line 46: I think the loading and error state when no instance is present should be displayed within the BaseLayout. It could both be in a row instead of the tabs. Otherwise we render a just white page with the loader/error.

static/js/pages/instances/InstanceOverview.tsx
```diff
@@ -103,10 +77,6 @@ const InstanceOverview: FC<Props> = ({ controlTarget, instanceName }) => {
             <th>Description</th>
```
edlerd, line 77: While you're at this file, can you remove the Label/Value heading? It seems a bit odd.

static/sass/styles.scss
```diff
@@ -163,3 +163,12 @@
 .u-row-selected {
   background-color: $color-mid-x-light;
 }
+
+.u-instance-status {
```
edlerd, line 167: better use the vanilla standard `u-text--muted` `p-text--small`. We should avoid custom CSS.

static/js/pages/instances/InstanceDetail.tsx
```diff
@@ -31,11 +60,27 @@ const InstanceDetail: FC = () => {
...
-      title={`Instance details for ${name}`}
+      title={
+        <List
+          inline
+          items={[
+            <span key="name">{name}</span>,
+            <span key="status" className="u-instance-status">
+              {instance.status}
+            </span>,
+            <StartStopInstanceBtn
```
edlerd, line 71: I think a dense button might be better fitting the text height.

static/sass/styles.scss
```diff
@@ -163,3 +163,12 @@
...
+  font-size: 80%;
+  font-style: italic;
+}
+
+.u-button-icon {
```
edlerd, line 172: This shouldn't be needed.

The misplacement comes from the button being inside the h4, which shouldn't be the case. The h4 should only wrap the instance name.

Maybe don't use the `BaseLayout` here?

static/js/pages/instances/actions/StartStopInstanceBtn.tsx
```diff
@@ -0,0 +1,103 @@
...
+const StartStopInstanceBtn: FC<Props> = ({
+  instance,
+  notify,
+  hasCaption = true,
+}) => {
+  const [isLoading, setLoading] = useState(false);
+  const queryClient = useQueryClient();
+
+  const isRunning = instance.status === "Running";
+  const isStopped = instance.status === "Stopped";
```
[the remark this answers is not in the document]
lorumic: What do you mean by "fails"? I don't think we should consider every other status (error, unknown) as "stopped", otherwise this means that you would be able to start an errored/unknown instance. Is this the correct behaviour? If yes, then I will be happy to change as you suggest. Otherwise, as it is now, we have 3 cases:

1) Instance running: show the "Stop" button (which stops the instance if clicked)
2) Instance stopped: show the "Start" button (which starts the instance if clicked)
3) Other status: _component is not rendered as we would not be able to stop nor start it_

Or did I misunderstand anything about the errored/unknown statuses?

static/js/pages/instances/InstanceDetail.tsx
lorumic: You mean in `InstanceDetailPanel` or here? I think here is fine - the other controls are also not dense.

static/js/pages/instances/InstanceDetail.tsx
lorumic: Done @ [4f62c9e](https://github.com/canonical/lxd-ui/commit/4f62c9e2df86d2b692c4298deac8212f25ffcc02)

static/js/pages/instances/InstanceOverview.tsx
lorumic: Done @ [4f62c9e](https://github.com/canonical/lxd-ui/commit/4f62c9e2df86d2b692c4298deac8212f25ffcc02)

static/sass/styles.scss
lorumic: Done @ [4f62c9e](https://github.com/canonical/lxd-ui/commit/4f62c9e2df86d2b692c4298deac8212f25ffcc02)

(`u-text--muted` not needed here)

static/sass/styles.scss
lorumic: Done @ [4f62c9e](https://github.com/canonical/lxd-ui/commit/4f62c9e2df86d2b692c4298deac8212f25ffcc02)

static/js/pages/instances/actions/StartStopInstanceBtn.tsx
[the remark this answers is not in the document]
edlerd: Hmm, good question. I was assuming you would still want to try to start an instance from an error state.

static/js/pages/instances/InstanceDetail.tsx
edlerd: I meant the detail page, but ymmw - Just thought it should be part of the Title and match its height. But consistency with the other buttons on the right is also a valid aspect.

static/js/pages/instances/InstanceStatusIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface Props {
+  instance: LxdInstance;
+}
+
+const InstanceStatusIcon: FC<Props> = ({ instance }) => {
+  // todo: which states are used - can error/unknown/init be removed?
+  const getIconClassForStatus = (status: string) => {
+    return {
+      error: "p-icon--oval-red",
```
edlerd, line 12: Let's make this gren on running, grey on stopped and yellow in all other cases.

static/js/pages/instances/actions/StartStopInstanceBtn.tsx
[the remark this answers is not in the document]
edlerd: Can you ask the core team in the LXD UI channel if we can start from an error state? Or generally which other states than running/stopped are actually in use and how the start/stop should behave in them.

static/js/pages/instances/InstanceStatusIcon.tsx
lorumic: Resolved @ [15828c6](https://github.com/canonical/lxd-ui/commit/15828c618a635e47275a8b2e59458573733c7bc0)

edlerd (approved):

PR #139, opened 2023-01-11
Certificate generation in the browser

edlerd:


static/js/pages/certificates/BrowserImport.tsx
```diff
@@ -0,0 +1,76 @@
...
+              (tab === "firefox" && !activeTab),
+            onClick: () => handleTabChange(tab.toLowerCase()),
+          }))}
+        />
+
+        {activeTab === "firefox" && (
+          <div tabIndex={0} role="tabpanel" aria-labelledby="firefox">
+            <ul className="p-list--divided">
+              <li className="p-list__item">
+                Paste this link into the address bar{" "}
```
lorumic, line 26: I would put a colon at the end:

> Paste this link into the address bar:

static/js/pages/certificates/BrowserImport.tsx
```diff
@@ -0,0 +1,76 @@
...
+                <a
+                  href="about:preferences#privacy"
```
lorumic, line 28: This link (and the one for Chrome as well) doesn't seem to work, so maybe it should be removed?

static/js/util/certificate.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const keys = forge.pki.rsa.generateKeyPair(2048);
+  const cert = forge.pki.createCertificate();
+  cert.publicKey = keys.publicKey;
+
+  // @ts-expect-error toString can have an argument!
+  cert.serialNumber = "01" + getRandomBytes(19).toString(16);
+  cert.validity.notBefore = new Date();
+  cert.validity.notAfter = new Date(
+    new Date().getTime() + 1000 * 60 * 60 * 24 * validityDays
```
lorumic, line 39: You can use `Date.now()` instead of `new Date().getTime()`.

static/js/pages/certificates/CertificateGenerate.tsx
```diff
@@ -0,0 +1,124 @@
...
+interface Certs {
+  crt: string;
+  pfx: string;
+}
+
+const CertificateGenerate: FC = () => {
+  const [certs, setCerts] = useState<Certs | null>(null);
+
+  const createCert = () => {
+    const certs = generateCert();
```
lorumic, line 16: Currently the UI gets frozen for a few seconds. I know that there is a warning to the user that this might happen, but usually it is better to guard this scenario with a loader.

static/js/pages/certificates/BrowserImport.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const [activeTab, handleTabChange] = useState("firefox");
+
+  return (
+    <Row>
+      <Col size={8}>
+        <Tabs
+          links={TABS.map((tab) => ({
+            label: tab,
+            active:
+              tab.toLowerCase() === activeTab ||
```
lorumic, line 16: Do we need to have it all lowercase for the `activeTab` state? Having it consistent with the tab title would save us a `toLowerCase` conversion.

static/js/pages/certificates/BrowserImport.tsx
```diff
@@ -0,0 +1,76 @@
...
+                  href="chrome://settings/certificates"
+                  target="_blank"
+                  rel="noreferrer"
+                >
+                  chrome://settings/certificates
+                </a>
+              </li>
+
+              <li className="p-list__item">
+                Click the blue Import button and select the{" "}
```
lorumic, line 65: Should `Import` be in a `<code>` element?

edlerd:
Todo: Add chrome windows and rename chrome linux

static/js/pages/certificates/CertificateGenerate.tsx
```diff
@@ -0,0 +1,136 @@
...
+                <Col size={8}>
+                  Download <code>lxd-ui.crt</code> and add it to the LXD trust
+                  store
+                  <div className="p-code-snippet">
+                    <pre className="p-code-snippet__block--icon">
+                      <code>lxc config trust add Downloads/lxd-ui.crt</code>{" "}
+                      <CopyButton
+                        text="lxc config trust add Downloads/lxd-ui.crt"
+                        tooltipPosition="right"
+                      />
```
lorumic, line 86: As mentioned, this would probably better right-aligned.

static/js/pages/certificates/BrowserImport.tsx
```diff
@@ -0,0 +1,97 @@
...
+        {activeTab === FIREFOX && (
+          <div tabIndex={0} role="tabpanel" aria-label="firefox">
+            <ul className="p-list--divided">
+              <li className="p-list__item">
+                Paste this link into the address bar:{" "}
+                <code>about:preferences#privacy</code>{" "}
+                <CopyButton
+                  text="about:preferences#privacy"
+                  tooltipPosition="right"
+                />
```
lorumic, line 33: This and the other `CopyButton`s could be right-aligned; in addition they could be wrapped inside the `pre` element, just like in the previous step (Trust).

static/js/pages/certificates/CertificateGenerate.tsx
edlerd: It is just the buttons to download and copy would be too close to each other.

lorumic (approved):
LGTM

PR #145, opened 2023-01-12
Alternate display of IPv4/IPv6 information [WD-1204]

lorumic:


static/sass/styles.scss
```diff
@@ -122,6 +122,13 @@
   }
 }
 
+.p-address-row {
```
edlerd, line 125: You can use `u-hide--large` from https://vanillaframework.io/docs/utilities/hide

static/js/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -86,6 +86,14 @@ const InstanceDetailPanel: FC<Props> = ({ instance, notify, onClose }) => {
               <th>Last used</th>
               <td>{isoTimeToString(instance.last_used_at)}</td>
             </tr>
+            <tr className="p-address-row">
+              <th>Address</th>
```
edlerd, line 90: I think we should split into ipv6 and ipv4, as we do in the table. We should probably adopt the overview tab on the instance detail page as well.

edlerd: Also the ipv6 can be fairly long (i.e. fd42:34d1:2187:887a:216:3eff:fe9e:f091), overflowing the width of the panel. Maybe wrap every ipv6 with a `<div className="u-truncate" title={ipv6}>{ipv6}</div>`. The title makes it possible to hover with the mouse and see the full address, while the truncate adds a nice ellipsis when it is too long.

lorumic: > Also the ipv6 can be fairly long (i.e. fd42:34d1:2187:887a:216:3eff:fe9e:f091), overflowing the width of the panel. Maybe wrap every ipv6 with a `<div className="u-truncate" title={ipv6}>{ipv6}</div>`. The title makes it possible to hover with the mouse and see the full address, while the truncate adds a nice ellipsis when it is too long.

Honestly, I don't think this is a good idea for IP addresses. The reason is: the most common use case, at least in my opinion, is **copy-pasting** the address. If you put it into a `title`, you cannot copy it anymore.
My proposal is: we keep it as it is now, which is not so bad since it is already wrapping on the ":" (so every 4 chars at most), and we add the title. So you still get a way to see the full address (including the few characters overflowing the width) and a way to copy-paste the address. What do you think?

edlerd: In chrome, it doesn't wrap, it looks a bit odd for me:

Screenshot from 2023-01-13 09-42-33

edlerd: It even could wrap in a very unfortunate way, so you see only part of an address and think it is complete.

lorumic: It's wrapping on the colon, as I said. You are missing only 3 characters. It looks odd, but be sure that as soon as you ellipsise it you will be bombarded by requests of angry users that want it to be copy-pastable. So another solution needs to be found in my opinion.

edlerd: Maybe truncate and add a copy button next to it?

edlerd: We have `p-icon--copy`

static/sass/styles.scss
lorumic: I applied that class on the `tr` and I cannot get it to be shown in any way. Also, it is inconsistent with the style just above:

```
.p-instance-table {
  @media screen and (max-width: 1400px) {
    th,
    td {
      &:nth-child(4),
      &:nth-child(5) {
        display: none;
      }
    }
  }
}
```

edlerd: You are right, I forgot we can't use large, because it is `1036px` and we don't want to change it either because it has other side effects. So I guess we have to keep it as is.

static/js/pages/instances/InstanceDetailPanel.tsx
lorumic: Good idea, I'll try that!

static/js/components/CopyButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+}) => {
+  const [copied, setCopied] = useState(false);
+
+  const handleCopy = async () => {
+    try {
+      await navigator.clipboard.writeText(text);
+      setCopied(true);
+
+      setTimeout(() => {
+        setCopied(false);
```
edlerd, line 24: This is a bit odd. It changes the tooltip message and the icon back, but why not keep it as it is without the timeout / revert function?

static/js/pages/instances/InstanceDetailPanel.tsx
edlerd: Yeah, it is very neat. I like it.

static/js/components/CopyButton.tsx
lorumic: It's taken from anbox dashboard with no modifications. Since it was in main I assumed it was reviewed and good already?

edlerd: I might have missed it there.

lorumic: I think it's good as it is now. The user would get confused with no revert as they might want to copy it again and would not know that the button is back to the "can-copy" state. Also, it is a common way to handle this kind of functionality, I've seen it everywhere done like this.

edlerd: ok

edlerd (approved):

PR #149, opened 2023-01-12
Add freeze/unfreeze button for instances [WD-1205]

lorumic:


static/js/pages/instances/actions/FreezeInstanceBtn.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const isDisabled =
+    isLoading || instance.status === "Stopped" || instance.status === "Error";
+
+  return (
+    <Tooltip
+      message={isDisabled ? "Instance must be running" : undefined}
+      position="btm-center"
+    >
+      {instance.status !== "Frozen" && (
```
edlerd, line 66: Isn't `instance.status === "Frozen"  ? (...) : (...)` simpler?

static/js/pages/instances/actions/FreezeInstanceBtn.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <ConfirmationButton
+          className="u-no-margin--bottom u-no-margin--right"
+          isLoading={isLoading}
+          iconClass={
+            isLoading
+              ? "p-icon--spinner u-animation--spin"
+              : "p-icon--disconnect"
```
edlerd, line 73: This looks a bit too much like networking, no?

static/js/pages/instances/actions/FreezeInstanceBtn.tsx
```diff
@@ -0,0 +1,108 @@
...
+        notify.failure("Error on instance unfreeze.", e);
+      });
+  };
+
+  const isDisabled =
+    isLoading || instance.status === "Stopped" || instance.status === "Error";
+
+  return (
+    <Tooltip
+      message={isDisabled ? "Instance must be running" : undefined}
```
edlerd, line 63: I am a bit confused, this would show up when loading as well.

static/js/pages/instances/actions/FreezeInstanceBtn.tsx
lorumic: This is the closest I could find within vanilla icon set. The only alternative that comes to my mind is to remove the icon, which would make it look inconsistent with the Start/Stop button (as well as all the other buttons), unless I remove it from there as well.
Any other solution?

edlerd: Let's keep it as is, and I'll make a request for a new icon.

static/js/pages/instances/actions/FreezeInstanceBtn.tsx
lorumic: Right, I had a different structure before, and I overlooked this part. Now fixed @ [12afd4e](https://github.com/canonical/lxd-ui/commit/12afd4ee74f4b5eec68e0e678f3e45d68e280243)

static/js/pages/instances/actions/FreezeInstanceBtn.tsx
lorumic: Fixed @ [12afd4e](https://github.com/canonical/lxd-ui/commit/12afd4ee74f4b5eec68e0e678f3e45d68e280243)

edlerd (approved):

PR #152, opened 2023-01-16
Add empty state for instance list [WD-1273]

lorumic:


static/js/pages/instances/InstanceList.tsx
```diff
@@ -263,7 +263,35 @@ const InstanceList: FC = () => {
                 isLoading ? (
                   <Loader text="Loading instances..." />
                 ) : (
-                  "No data to display"
+                  <Row className="empty-state-message">
```
edlerd, line 266: In the snapshot table, the column names are not shown with the empty state. Can make it the same here? Not sure why this differs though.
Or at least at some spacing to the above element, the empty states is a bit cramped below the table headers atm.

lorumic: Instance list:

image

Snapshot list:

image

I don't see any difference and the column headers are showing in both cases. I will add some spacing to both of them, so that hopefully they look less cramped.

edlerd: I didn't get the table headers in empty snapshots yesterday, confused why they appear today.

I still think it would look better with 3rem margin above the empty states.

anthonydillon:
We shouldn't use center align. Here is the empty table pattern: https://vanillaframework.io/docs/base/tables#empty

static/js/pages/instances/InstanceList.tsx
lorumic: The old version of the snapshot empty state, which was not using the `emptyStateMsg` prop, used to be without the column headers, because it was conditionally rendered instead of the table itself. But that was the first change that I applied - even before the one for instances - as reflected by the commit order, so probably you were running a cached version of the app or wrong branch.

I have changed the margin to 3rem, but I will also check Ant's link now and commit again soon.

edlerd: From the example in vanilla I guess we should use a wrapper with the `p-strip` class to get the 3rem spacing.

edlerd (approved):

PR #154, opened 2023-01-16
Fix broken format-js script in package.json

lorumic:


package.json
```diff
@@ -9,7 +9,7 @@
     "build-js": "webpack",
     "build-css": "sass static/sass/styles.scss build/ui/styles.css --load-path=node_modules --style=compressed && postcss --use autoprefixer --replace 'build/**/*.css' --no-map",
     "build": "yarn run build-js && yarn run build-css",
-    "format-js": "prettier 'static/js/**/*.{js,json,jsx}' --write",
+    "format-js": "eslint 'static/js/**/*.{js,json,jsx,tsx}' --fix",
```
edlerd, line 12: Why change it? It peeked at ubuntu.com package json and they use prettier as well.

lorumic: Right, I should have mentioned the reasons since the beginning:

1) We already use eslint for the `lint-js` script, so I don't see why we need another package for `format-js` since we can do exactly the same thing with the package that we already have - and indeed, I realised that I should also remove prettier from the `devDependencies` in package.json.

2) The `format-js` script with prettier has never really worked for me:

image

So what I'm doing since 3 months by now is replacing it with eslint locally, which means that I need to stash the changes to package.json at every rebase, and I am kind of fed up of doing this.

Of course I am open to alternative solutions, but considering what I just said I don't see a valid reason to keep using prettier instead of eslint.

PR #155, opened 2023-01-16
Display freeze button when not frozen (not the opposite)

lorumic:


static/js/pages/instances/actions/FreezeInstanceBtn.tsx
```diff
@@ -65,7 +65,7 @@ const FreezeInstanceBtn: FC<Props> = ({
       }
       position="btm-center"
     >
-      {instance.status === "Frozen" ? (
+      {instance.status !== "Frozen" ? (
```
edlerd, line 68: It is usually better (i.e. easier to read) when using positive comparisons. could you flip this around?

lorumic: Usually, yes, but in this case flipping it around would mean that the "Unfreeze" button appears before the "Freeze" one in the code - which is not tidy, in my opinion. It's better to have the buttons in the order they are now - and hence, we need the negative comparison.

edlerd (approved):

static/js/pages/instances/actions/FreezeInstanceBtn.tsx
edlerd: I disagree, but lets keep it as is for now.

lorumic: For future disagreements, feel free to add a further reviewer to get advice: it's more constructive than merging and then stating here that you disagree. :)

PR #156, opened 2023-01-17
Instance creation improvements [WD-1275]

lorumic:


static/js/pages/instances/InstanceFormGuided.tsx
```diff
@@ -91,111 +124,169 @@ const InstanceFormGuided: FC = () => {
     }
   };
 
+  const getInstanceCreationObj = (): LxdInstanceCreation => {
+    return (
+      isYamlConfig
```
edlerd, line 129: This is confusing, because there are two callers to the getInstanceCreationObj  function and the call from switchToYaml  doesn't actually care about the isYamlConfig. Or it is always false in that case. I think we should split the functionionality to have one function turning the formik values into an object.

The isYamlConfig should actually live in the onSubmit function directly.

Also I'd pass in the formik values to that function. Because when calling from onSubmit, the values passed to onSubmit can differ from formik.values. I.e. the onSubmit should use the values from the argument it receives, not the formik.values.

static/js/pages/instances/InstanceFormGuided.tsx
```diff
@@ -91,111 +124,169 @@ const InstanceFormGuided: FC = () => {
...
+          }
+    ) as LxdInstanceCreation;
+  };
+
+  const setYaml = (text: string) => {
+    void formik.setFieldValue("instanceYaml", text);
+  };
+
+  const switchToYaml = () => {
+    setYaml(dumpYaml(getInstanceCreationObj()));
```
edlerd, line 151: Here I'd split this to multiple lines as well to un-nest.

static/js/pages/instances/InstanceFormGuided.tsx
```diff
@@ -43,17 +50,29 @@ const InstanceFormGuided: FC = () => {
...
   });
 
+  const create = () => {
+    setWithStart(false);
+    formik.handleSubmit();
+  };
+
+  const createAndStart = () => {
+    setWithStart(true);
+    formik.handleSubmit();
```
edlerd, line 60: This could fail, when the state update doesn't land in time before it is queried. It it usually better to pass the boolean directly to the function handling it. But it might be hard in this case, because handlesubmit doesn't allow argument passing easily.
I think we should split the actual logic to create and start into a function that receives the values and the boolean. Then we can call this function from here, from line 54 above and from formiks onSubmit as well. Does it make sense?

edlerd (commented):
I really like the new behavior of the creation flow. It is so much nicer than before. Just a bunch of comments, sorry it was so many things.

static/js/pages/instances/InstanceList.tsx
```diff
@@ -186,23 +185,13 @@ const InstanceList: FC = () => {
...
-          <ContextualMenu
-            hasToggleIcon
-            links={[
-              {
-                children: "Quick create instance",
-                onClick: () => panelParams.openInstanceFormGuided(),
-              },
-              {
-                children: "Custom create instance (YAML)",
-                onClick: () => panelParams.openInstanceFormYaml(),
```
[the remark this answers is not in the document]
lorumic: Yes, I was already taking that into account, but since the other YAML-based instance creation form is also used to edit instances, I think it would be better to update it so that it handles the edit case only, instead of deleting it completely. So I created a dedicated ticket for this ([WD-1358](https://warthogs.atlassian.net/browse/WD-1358)), feel free to add details there about this point in case I forgot anything. I plan to work on that right after this PR gets merged, but for what I have just mentioned I would prefer to consider it a separate issue and out of the scope of this PR.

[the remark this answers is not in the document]
edlerd: sounds good, thanks.

lorumic:
All the points in the comments (plus the ones listed during the call) should have been addressed @ [f7a8bd5](https://github.com/canonical/lxd-ui/commit/f7a8bd5f191385bd12d4fc62ded080037019f993)

static/js/pages/instances/InstanceList.tsx
```diff
@@ -36,6 +36,13 @@ const InstanceList: FC = () => {
   const [status, setStatus] = useState<string>("any");
   const [type, setType] = useState<string>("any");
 
+  const { setSharedNotify } = useSharedNotify();
+  useEffect(() => {
+    if (setSharedNotify) {
+      setSharedNotify(notify);
+    }
+  }, [setSharedNotify]);
```
edlerd, line 44: I forgot, this useEffect also depends on notify, we should add that to the array of dependencies.

static/js/pages/instances/InstanceFormGuided.tsx
```diff
@@ -43,17 +51,80 @@ const InstanceFormGuided: FC = () => {
     instanceType: Yup.string().required("Instance type is required"),
   });
 
-  const formik = useFormik<{
+  interface InstanceCreationForm {
```
edlerd, line 54: Maybe `FormValues`?

edlerd (approved):
Very good stuff. I like how this turned out much nicer UX and also code wise. (Just some minor ideas for naming stuff and one dependency).

PR #157, opened 2023-01-18
fix(instance) add padding to terminal

edlerd:
* added 1 rem padding to terminal view

### DIFF
--- a/static/js/pages/instances/InstanceTerminal.tsx
+++ b/static/js/pages/instances/InstanceTerminal.tsx
@@ -135,6 +135,8 @@ const InstanceTerminal: FC<Props> = ({ controlTarget, notify }) => {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


static/js/pages/instances/InstanceTerminal.tsx
```diff
@@ -135,6 +135,8 @@ const InstanceTerminal: FC<Props> = ({ controlTarget, notify }) => {
       return;
     }
 
+    xtermRef.current?.terminal.element?.style.setProperty("padding", "1rem");
```
lorumic, line 138: Isn't it better to add a class with this style instead of setting it inline?

edlerd: It messes up the fit addon when done in another way. so the cols/rows won't match actual screen sice.

lorumic (approved):

PR #167, opened 2023-01-23
Show action buttons on row hover in instance list [WD-1451]

lorumic:


static/sass/_instance_list.scss
```diff
@@ -0,0 +1,42 @@
...
+
+.u-row:hover {
+  .u-instance-actions {
+    visibility: visible;
+  }
+}
+
+.u-row-selected:hover {
+  .u-instance-actions {
+    display: none;
```
edlerd, line 40: This is a little wired. I would also show them on the selected row on hover.

static/sass/_instance_list.scss
```diff
@@ -0,0 +1,42 @@
...
+    &:nth-child(4),
+    &:nth-child(5) {
+      display: none;
+    }
+  }
+}
+
+.u-row:hover,
+.u-row-selected {
+  background-color: $colors--light-theme--background-hover;
```
edlerd, line 25: I think the hover and selected should slightly differ. Lets follow up with design on this question.

edlerd (commented):
There is no way on mobile to hover, do we have a solution for table/phone clients?

static/sass/_instance_list.scss
lorumic: I don't see why: we have the actions in the side panel, for the selected row.

edlerd: I think it would be better for consistency.

static/sass/_instance_list.scss
lorumic: I checked this and the design for now seems to use the same.

static/sass/_instance_list.scss
lorumic: Let's keep this point open and ask design input about it.

lorumic:
> There is no way on mobile to hover, do we have a solution for tablet/phone clients?

This also needs design input to be answered. What behaviour do we want to have on mobile clients? But above all, do we expect to have mobile clients at all for this UI?

edlerd:
> This also needs design input to be answered. What behaviour do we want to have on mobile clients? But above all, do we expect to have mobile clients at all for this UI?

As discussed in standup: We can always show the buttons on small screens. And yes it is not our focus, but the UI should work on mobile as well.

lorumic:
> > This also needs design input to be answered. What behaviour do we want to have on mobile clients? But above all, do we expect to have mobile clients at all for this UI?
> 
> As discussed in standup: We can always show the buttons on small screens. And yes it is not our focus, but the UI should work on mobile as well.

Ok, no problem, I will implement that.
As for the other 2 points above, I will mark them as resolved, since my understanding was confirmed to be correct and no action is needed, at least in the short term.

edlerd (approved):

PR #168, opened 2023-01-24
Create instance yaml x will go back instead of close panel

edlerd:


lorumic (approved):
Might be a good idea to generally safeguard the cancel action on the create new instance form.

PR #170, opened 2023-01-24
Create snapshot error should not close modal and display message inside

edlerd:


lorumic (approved):
Looks good, but if we could also prevent some failures it would be better - e.g., we should properly validate the snapshot name field so that any incorrect input is warned in advance of the submission. :)

PR #175, opened 2023-01-25
Simplify dev setup

edlerd:


README.md
```diff
@@ -5,64 +5,31 @@ Targets small and large scale private clouds.
...
-
-    sudo service haproxy restart
+    sudo lxc config trust add keys/lxd-ui.crt
 
 Install dotrun as described in https://github.com/canonical/dotrun#installation Launch it from the head of this repo
 
     dotrun
 
-Browse through https://localhost:9443/ and **avoid** querying port 3000 directly. Requests to the lxd core won't reach HAProxy on the port 3000.
+You might want to enable a firewall as the last step exposes an api to start or interact with unprivileged containers on your public ip.
```
lorumic, line 28: Which step? The last step is launching with `dotrun`, but I'm not sure it's the one you're referring to. Also, since we're here, it might be better to specify that if the selected port is open on your machine (allowed in `ufw` for example) but not open on your router, you expose the API only within the LAN, which is something that might be desired.
In summary, I would rephrase this paragraph so that it clarifies some information (what's the exact step that exposes you to a risk) and includes further useful information about using the API within the LAN and exclude it from the WAN.

Also, I don't understand why this completely replaces the information about how to reach the UI in a browser. `Browse through https://localhost:9443/`should be left there in my opinion.

edlerd: We hide the wrong notification about reaching the ui on port 3000. Also dotrun ends with a message `booting on https://localhost:9443`. So I assumed that is enough.
I just simplified a bit more, so dotrun also generates the keys. Now we end the readme with a not about how to reach the UI again. (see latest commit below)

lorumic (approved):
Thanks, LGTM.

PR #179, opened 2023-01-27
Keep start/stop button visible during operation progress

lorumic:


static/js/pages/instances/actions/StartStopInstanceBtn.tsx
```diff
@@ -67,7 +67,7 @@ const StartStopInstanceBtn: FC<Props> = ({
     <>
       {canBeStarted && (
         <ConfirmationButton
-          className={className}
+          className={isLoading ? "" : className}
```
edlerd, line 70: While I agree to the problem, this is a little wired to just not forward the `className` to the button if it is loading. It surely fits our current use cases, but if we use the Buttons in a different context, then removing the `className` on load might not be the right thing.
It looks funny when the single button appears and disappears while loading. I wonder if we should just hide the button in the hover section, or always display it? Not sure about a good solution here.

lorumic: Yeah I thought exactly the same, but I didn't want to add too much overhead to a PR when our current use case doesn't need it. So my idea fundamentally was: let's handle it this way for now, and then, if and when a new case arises, we'll figure out a cleaner way to handle it. Also, this button might be changed more heavily, in appearance and function, at a later design iteration, so I wouldn't want to overthink something that is still subject to changes. What do you think?

edlerd (approved):

static/js/pages/instances/actions/StartStopInstanceBtn.tsx
edlerd: Sounds OK, let's keep this in mind to be improved.

PR #184, opened 2023-01-31
Add start page for unauthenticated user and adapt nav menu [WD-1348]

lorumic:


static/js/App.tsx
```diff
@@ -18,18 +18,29 @@ import StorageList from "pages/storages/StorageList";
 import ProfileDetail from "pages/profiles/ProfileDetail";
 import CertificateGenerate from "pages/certificates/CertificateGenerate";
 import { SharedNotifyProvider } from "./context/sharedNotify";
+import NoAuth from "pages/certificates/NoAuth";
 
 const queryClient = new QueryClient();
 
 const App: FC = () => {
+  const [isAuthenticated, setAuthenticated] = useState(false);
```
yurii-vasyliev, line 26: @lorumic , wouldn't it be better to create a context + a custom hook for that?
Like 
```
const { isAuthenticated } = useAuth();
```
You wouldn't need to pass that state down to the components in that case.

static/js/pages/certificates/CertificateGenerate.tsx
```diff
@@ -121,6 +121,15 @@ const CertificateGenerate: FC = () => {
             <h3 className="p-stepped-list__title">
               All done <Icon name="success" />
             </h3>
+            <div className="p-stepped-list__content">
+              <Row>
+                <Col size={8}>
+                  <p>
+                    <a href="/">Go back to the main page</a>.
```
yurii-vasyliev, line 128: Does it have to be an `a`? With page reloading?

static/js/pages/certificates/NoAuth.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Row, Col } from "@canonical/react-components";
+
+const NoAuth: FC = () => {
+  return (
+    <BaseLayout title="Client authentication needed">
+      <Row>
+        <Col size={12}>
+          <h5>Uh-oh! Looks like you are not authenticated.</h5>
+          <p>
+            <a href="/ui/certificates">
```
yurii-vasyliev, line 12: The same question here about page reloading.

static/js/App.tsx
lorumic: Maybe, but at this point this information is really only needed by `App` and `Navigation`, not by any other component, so I thought a simple state that is passed to `Navigation` was enough, without introducing too much overhead. Of course, it can be always changed later, if the information needs to be passed elsewhere or if it's considered a bad practice to do it like this. This is just, let's say, a first iteration - the design of the `NoAuth` page is also likely to change a lot.

static/js/pages/certificates/CertificateGenerate.tsx
lorumic: You mean with `useNavigate`? Or, can you give me a quick code example of what you mean?

static/js/pages/certificates/NoAuth.tsx
lorumic: See the other comment - the page is not the same.

static/js/pages/certificates/CertificateGenerate.tsx
yurii-vasyliev: I mean using `<Link to="/">Go back to the main page</Link>` from `react-router-dom` instead of `a`. If you use a HTML anchor tag it will always reload the page and lose your state.

lorumic: Ah, yes, now I understand your question. Yes, I need to lose my state here because I need to check again if the user is now (= after following the steps for certificate generation and whitelisting) authenticated or not - if I don't, the user will see the `NoAuth` page forever until they reload the page. So yes, in short, a page reload is what we want here.

static/js/pages/certificates/NoAuth.tsx
lorumic: I deleted my other comment since I misunderstood your original question. It is safe to use `Link` here, I will update this part.

static/js/App.tsx
```diff
@@ -26,34 +29,76 @@ const App: FC = () => {
...
-              <Route path="/ui/settings" element={<Settings />} />
-              <Route
-                path="/ui/certificates"
-                element={<CertificateGenerate />}
-              />
-              <Route path="*" element={<NoMatch />} />
-            </Routes>
-            <Panels />
-          </div>
+          <AuthProvider>
```
yurii-vasyliev, line 32: Not a big deal, but you could create an `index.tsx` and move all providers there, which would keep this file cleaner. And this is a more usual way to handle the file structure.
https://github.com/canonical/anbox-cloud-dashboard/blob/main/static/js/index.tsx

static/js/context/auth.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { createContext, FC, ReactNode, useContext } from "react";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+
+const checkAuth = async () =>
+  await fetch("/1.0/certificates").then((response) => response.status !== 403);
```
yurii-vasyliev, line 6: What is the point of waiting the response here? You could just return the promise as it is :)

static/js/context/auth.tsx
```diff
@@ -0,0 +1,44 @@
...
+  isAuthLoading: true,
+};
+
+export const AuthContext = createContext<ContextProps>(initialState);
+
+interface ProviderProps {
+  children: ReactNode;
+}
+
+export const AuthProvider: FC<ProviderProps> = ({ children }) => {
```
yurii-vasyliev, line 24: Could you catch errors and show them in the console at least?

lorumic: Currently we don't have a practice of logging errors to the console in this project.

static/js/App.tsx
lorumic: Done

yurii-vasyliev (approved):

static/js/context/auth.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { createContext, FC, ReactNode, useContext } from "react";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+
+const checkAuth = () =>
+  fetch("/1.0/certificates").then((response) => response.status !== 403);
```
edlerd, line 6: I think this should be in a dedicated file in API folder. Also this check fails on the demo server. Maybe it can be extended to also work there?

static/js/context/auth.tsx
lorumic: Ah, sorry, I didn't understand you were talking about the `async` / `await` thing. You're right. That can be removed and I did it with the last commit. Actually, it was a "leftover" because initially I wasn't using `useQuery`, I was fetching directly instead.

static/js/pages/certificates/CertificateGenerate.tsx
```diff
@@ -12,6 +15,15 @@ interface Certs {
 const CertificateGenerate: FC = () => {
   const [isGenerating, setGenerating] = useState(false);
   const [certs, setCerts] = useState<Certs | null>(null);
+  const { isAuthenticated, isAuthLoading } = useAuth();
```
edlerd, line 18: Not sure if the process here works. Would something trigger the new check if it is authenticated after following the steps? Maybe we need a button to verify it and notify in case of success instead of redirecting?

lorumic: It works, because I'm invalidating the query. I tested it many, many times. :)

static/js/context/auth.tsx
lorumic: Well, I don't know because I cannot do the add/remove thing from the lxc trusted list on the demo server, so I don't really have any way to test it...

static/js/pages/certificates/CertificateGenerate.tsx
edlerd: +1

static/js/context/auth.tsx
edlerd: Right, maybe the better approach is to make the demo proxy return 403 for any calls in https://github.com/canonical/lxd-ui/blob/main/haproxy-demo.cfg#L12
If is_lxd_core is true and is_backend_allowed is false. If you feel like writing some HAProxy config feel free the add it, or I can look into it later on.

lorumic: I moved the solitary function to its own file.

About the demo server, I found a way to return 403 for any call and already tested it locally (= on the haproxy-dev config first), before pushing anything to the remote. However, I don't see how this can be so important since the demo server is already configured correctly and hence working without the auth check, so we don't really need to test this functionality on it. I don't even understand how you realised that it was failing there without manipulating the trust list.

However, I'm happy to test the method I found on the haproxy-demo config anyway, but out of the scope of this PR. I would either open another one later myself, expressly for this purpose, or you can feel free to check this issue later in a follow-up.

If there are no further remarks, I'd be happy to proceed with the merge. I can keep the PR unmerged until next week anyway, no problems with that.

static/js/pages/certificates/CertificateGenerate.tsx
lorumic: Sorry, I had a false memory about query invalidation. So I'll add further explanation of my understanding of what happens here: the query gets re-fetched on window focus. So everything is automatic: once you switch back to the browser window after running the `lxc config trust add` command in the terminal, you'll get a fresh fetch that will give you a 200 instead of 403, hence updating the UI "automatically".

edlerd (approved):
LGTM

al-bakalova:
Please align the Gerenate, Download a crt and Download fx buttons with the labels as discussed on MM.

The rest looks good

The following design issue has been added to the backlog - https://warthogs.atlassian.net/browse/WD-1875

PR #185, opened 2023-02-04
Add project showcase section

JacobCoffee:
It would be nice to include a little showcase of how nice the project interface actual is.

Compared to other LXD UI solutions this one is by far my favorite and best.. but I found myself wary to even try it because of how others performed and looked.

If I were looking for something like this I would immediately want to know how clunky or not the UI is. Having this would help do that without going through the whole process of standing the app up.

Perhaps including these sample images would give some people to push to try it out; additionally, it can be used for easy screen captures to share on social media platforms to generate buzz around the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,12 @@ The tests expect the environment on localhost to be accessible. Execute `dotrun`
 Generate new tests with helper
 
     npx playwright codegen --ignore-https-errors https://localhost:9443/
+    
+# Examples
+
+### Home Page
+![image](https://user-images.githubusercontent.com/45884264/216782634-855c4d85-4f9a-409c-8443-b069dbba258c.png)
+
+
+### Container / VM Creation - Image Selection Menu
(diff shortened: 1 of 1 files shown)


edlerd:
I think we can use better screenshots with content in them. See the ones below, they might be more informative.

### Instance list
instance-list

### Instance detail
instance-detail

### Snapshots
snapshot

### Terminal
terminal

### Profile detail
profile-detail

JacobCoffee:
Good idea! I added your images but in the end it was super bulky. I am trying this table layout. You can still click into each image and expand it. What do you think?

image

edlerd (approved):
Looks good, thanks for adding this :)

JacobCoffee:
@edlerd do i need to do anything else to have this merged? just curious, its still in my GH notifications and wondering if I can mark it done or not😄

edlerd:
Yes, I'll merge it right away. Thanks again!

PR #187, opened 2023-02-07
Fix cert redirect

edlerd:
* we should always rediriect to /ui, because the / part is the api without the ui.


lorumic:
Not really needed - we'd still go to `InstanceList` by virtue of:

`<Route path="/" element={<ProtectedRoute outlet={<InstanceList />} />} />`

edlerd:
> Not really needed - we'd still go to `InstanceList` by virtue of:
> 
> `<Route path="/" element={<ProtectedRoute outlet={<InstanceList />} />} />`

That doesn't work when the ui is shipped with lxd, because lxd would not serve the ui under the / route.

lorumic (approved):

PR #189, opened 2023-02-07
Represent projects in the url of relevant components WD-1287

edlerd:


static/js/pages/profiles/ProfileDetail.tsx
```diff
@@ -20,11 +20,13 @@ import Loader from "components/Loader";
...
-
   const { name } = useParams<{ name: string }>();
+  const { project } = useParams<{
+    project: string;
+  }>();
 
-  if (!name) {
-    return <>Missing name</>;
+  if (!name || !project) {
+    return <>Missing name or project</>;
```
lorumic, line 29: See the other comment.

static/js/App.tsx
```diff
@@ -25,39 +25,39 @@ const App: FC = () => {
...
       />
       <Route
-        path="/ui/profiles/:name"
+        path="/ui/:project/profiles/:name"
         element={<ProtectedRoute outlet={<ProfileDetail />} />}
       />
       <Route
-        path="/ui/networks"
+        path="/ui/:project/networks"
         element={<ProtectedRoute outlet={<NetworkList />} />}
```
lorumic, line 53: The networks api file and all the points where `fetchNetworks` gets referenced (`NetworkList` and `NetworkSelector`) need to be updated too.

lorumic (approved):
LGTM

static/js/App.tsx
```diff
@@ -19,45 +19,42 @@ import ProtectedRoute from "components/ProtectedRoute";
 const App: FC = () => {
   return (
     <Routes>
-      <Route path="/" element={<ProtectedRoute outlet={<InstanceList />} />} />
+      <Route path="/" element={<Navigate to="/ui/default/instances" />} />
+      <Route path="/ui" element={<Navigate to="/ui/default/instances" />} />
```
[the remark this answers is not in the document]
lorumic: That doesn't change the fact that if you don't persist the user-selected project, they'd have to select it every time they use the UI, which is probably not ideal.

PR #190, opened 2023-02-07
Overhaul instance list header [WD-1892]

lorumic:


static/js/util/instanceOptions.tsx
```diff
@@ -1,10 +1,10 @@
 export const instanceTypeOptions = [
   {
-    label: "Container",
+    label: "Containers",
```
edlerd, line 3: This label is used for the selector on instance creation. There we need the singular form. So maybe pluralize it in place or have two of those?

static/js/pages/instances/InstanceList.tsx
```diff
@@ -218,127 +217,133 @@ const InstanceList: FC = () => {
...
+        <div className="p-panel__header">
+          <Row className="p-panel__title p-instance-header">
+            <Col size={2}>
+              <h4>
+                {visibleInstances.length}{" "}
+                {visibleInstances.length === 1 ? "Instance" : "Instances"}
+              </h4>
+            </Col>
+            <Col size={8}>
+              <Row>
```
edlerd, line 231: I think we don't need the nesting here and can remove the two layers for 
```
            <Col size={8}>
              <Row>
```

static/js/pages/instances/InstanceList.tsx
```diff
@@ -218,127 +217,133 @@ const InstanceList: FC = () => {
...
+                    options={[
+                      {
+                        label: "Containers and VMs",
+                        value: "any",
+                      },
+                      ...instanceTypeOptions,
+                    ]}
+                    value={type}
+                  />
+                </Col>
```
edlerd, line 274: The filters and the button look a little funny with screen width between 1000 and 1200 px.

I think we should give the button a col size of 3 at least. The filter and search can be together in one col with size 7. Then the two filters can have a fixed width and the search bar can adjust its width to fill the col-7. Maybe consult with design, as the responsive behaviour is not specified. But from the screens, I assume the filters are fixed width and the search input adjusts.

lorumic: Ok, let's first hear from @al-bakalova if your assumptions are correct, and then, once the responsive behaviour becomes clearer, I'll adapt the sizes accordingly.

edlerd (approved):
LGTM

al-bakalova:
Looks good to me as well. Thank you!

[The following issue ](https://warthogs.atlassian.net/browse/WD-1915)has been created to further improve the "create new" button's behaviour.

PR #191, opened 2023-02-07
Configurable columns on instance table

edlerd:


static/js/pages/instances/InstanceList.tsx
```diff
@@ -28,6 +28,19 @@ import StartStopInstanceBtn from "./actions/StartStopInstanceBtn";
...
+const loadHidden = () => {
+  const saved = localStorage.getItem("InstanceListHidden");
+  if (!saved) {
+    return [];
+  }
+  return JSON.parse(saved) as string[];
+};
+
+const saveHidden = (columns: string[]) => {
+  localStorage.setItem("InstanceListHidden", JSON.stringify(columns));
```
lorumic, line 42: Strange naming convention here - I would expect the key to be either camelCased or snake_cased. Not very important though, so feel free to resolve and keep it as is.

static/js/pages/instances/InstanceList.tsx
```diff
@@ -92,8 +111,19 @@ const InstanceList: FC = () => {
       sortKey: "snapshots",
       className: "u-align--center",
     },
-    { content: "" },
-  ];
+    {
+      content: (
+        <TableColumnsSelect
+          columns={["Status", "Name", "Type", "IPv4", "IPv6", "Snapshots"]}
```
lorumic, line 117: I'd extract these to a global, enum-like constant that we can use to reference the headers/aria-labels wherever they are used, so that if by any chance in the future we change them, they get automatically updated everywhere.

lorumic (commented):
Some high-level/design comments:

- I am not sure the `ContextualMenu` works well here, because I would expect the menu to stay open after I check/uncheck an item, so that I can apply multiple selections without having to open the menu every single time. Can this be done with the `ContextualMenu`?
- The right border of the menu should be aligned with the table's row separators:

image

static/js/pages/instances/InstanceList.tsx
edlerd: Good point, this was very unstable before, the constants make the dependency clear now.

edlerd:
All things above are resolved.

On small screens < 1400px width, we hide ipv4, ipv6 and snapshots by default. Users can configure all columns, but then the table will look odd, I guess this behaviour is the best we can do. Users will realize and pick the right ones for their size.

lorumic:
> I had to pull the TableColumnsSelect out of the table, so it won't get closed on selection. React would remount the whole table on adding/removing columns. Now we pull the gear into the table header with CSS.

The problem with this is that it loses the alignment with the `>` to open the instance detail panel. I'm afraid this won't make it pass the design review...

image

edlerd:
> The problem with this is that it loses the alignment with the `>` to open the instance detail panel. I'm afraid this won't make it pass the design review...

That happens because we want the alignment of the context menu with the right side of the table (your [previous comment](https://github.com/canonical/lxd-ui/pull/191#pullrequestreview-1289045256)). I will remove the left padding of the arrow icon to make the gear and arrow aligned.

al-bakalova:
Looks perfect to me! Thank you!

static/js/components/TableColumnsSelect.tsx
```diff
@@ -0,0 +1,62 @@
...
+      toggleClassName="has-icon"
+      toggleLabel={<Icon name="settings" />}
+      toggleAppearance="base"
+    >
+      <div className="table-column-select-list">
+        <CheckboxInput
+          checked={hidden.length === 0}
+          indeterminate={selectedCount > 0 && selectedCount < columns.length}
+          label={`${selectedCount} out of ${columns.length} selected`}
+          onChange={() => hidden.length > 0 && setHidden([])}
```
lorumic, line 45: I find this a bit confusing. What I would expect here is: if I click this checkbox when some of the checkboxes below are deselected, they turn to "all selected" (and this works), and if I click this checkbox when all checkboxes below are selected, they get all deselected - but this doesn't happen.

static/js/pages/instances/InstanceList.tsx
```diff
@@ -42,6 +73,34 @@ const InstanceList: FC = () => {
   const [query, setQuery] = useState<string>("");
   const [status, setStatus] = useState<string>("any");
   const [type, setType] = useState<string>("any");
+  const [userHidden, setUserHidden] = useState<string[]>(loadHidden());
+  const [sideHidden, setSideHidden] = useState<string[]>([]);
+
+  const hidden = userHidden.concat(sideHidden);
```
lorumic, line 79: I would expect this to go after L98 since that call to `resizeColsForSidepanel` might update the `sideHidden` state.

static/js/components/TableColumnsSelect.tsx
edlerd: Adjusted it so it matches expected behaviour.

static/js/pages/instances/InstanceList.tsx
edlerd: It shouldn't matter, because there will be another render cycle after the resize function updates the state. so even if it is listed after the function, it will not receive the updated values on the 1st render.

I put it there, so it is clear for the code reader how the two states are just combined.

lorumic (approved):

PR #193, opened 2023-02-09
Add text console

edlerd:


static/js/api/instances.tsx
```diff
@@ -160,3 +160,35 @@ export const fetchInstanceVga = (
...
+    fetch(`/1.0/instances/${name}/console?project=${project}`, {
+      method: "GET",
+    })
+      .then(handleTextResponse)
+      .then((data: string) => resolve(data))
+      .catch(reject);
+  });
+};
+
+export const fetchInstanceConsole = (
```
lorumic, line 178: It's a bit weird to have this called "fetch"; shouldn't it use the word "connect" instead? (Like [in the API specs](https://linuxcontainers.org/lxd/docs/latest/api/#/instances/instance_console_post))

static/js/pages/instances/InstanceTerminal.tsx
```diff
@@ -62,7 +62,7 @@ const InstanceTerminal: FC<Props> = ({ controlTarget, notify }) => {
     }
 
     setTerminalLoading(true);
-    const result = await fetchInstanceExec(name, project, payload).catch(
+    const result = await connectInstanceExec(name, project, payload).catch(
       (e) => {
         setTerminalLoading(false);
```
lorumic, line 67: While comparing side-by-side the `InstanceTerminal` and the `InstanceTextConsole`, I just happened to notice that the name of this state `[isTerminalLoading, setTerminalLoading]` sounds a bit "too customised", considering that there is no `isLoading` state in this component and we use `isLoading` for the state of the `TextConsole`.
I suggest renaming it to `[isLoading, setLoading]` for consistency, but it's a very tiny and unimportant thing, so I'll approve and feel free to keep it as it is.

lorumic (approved):

PR #196, opened 2023-02-13
Instance logs

edlerd:


static/js/pages/instances/DownloadButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    if (maybeExistingLink) {
+      maybeExistingLink.click();
+      return;
+    }
+
+    const data =
+      content ||
+      (await fetchInstanceLogFile(instance.name, instance.project, fileName));
+    const blob = new Blob([data], { type: "text/plan" });
```
lorumic, line 42: There is a typo in the MIME type, it should be `text/plain`.

static/js/pages/instances/DownloadButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+    link.setAttribute("download", fileName);
+    link.setAttribute("id", linkId);
+    link.style.display = "none";
+
+    return link;
+  };
+
+  const handleDownload = async () => {
+    const linkId = `link-${fileName.toLowerCase().replace(" ", "-")}`;
+    const maybeExistingLink = document.getElementById(linkId);
```
lorumic, line 32: I find this name confusing; I'd rather use something like `alreadyCreatedLink`.

static/js/pages/instances/FileRow.tsx
```diff
@@ -0,0 +1,81 @@
...
+          aria-label={`File ${fileName} URL`}
+          className="p-button--base has-icon is-dense u-no-margin--bottom"
+        >
+          <Icon name="get-link" />
+        </a>
+        <DownloadButton
+          fileName={fileName}
+          instance={instance}
+          content={logContent ?? ""}
+        />
```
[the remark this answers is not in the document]
edlerd: I didn't want to use the onClick, because then we can't copy the link or open it in a new tab easily.
But removing the two helper classes from the link adjust the layout to be the same.

static/js/pages/instances/DownloadButton.tsx
edlerd: good catch!

PR #197, opened 2023-02-13
Adjust empty state of instance list page [WD-1957]

lorumic:


static/js/components/EmptyState.tsx
```diff
@@ -0,0 +1,56 @@
...
+  title,
+  message,
+  linkMessage,
+  linkURL,
+  buttonLabel,
+  buttonAction,
+}) => {
+  return (
+    <Row>
+      <Col size={3} />
```
edlerd, line 27: We shouldn't use empty cols, better use `col-start-large-3 col-start-medium-2` or something like that from https://vanillaframework.io/docs/patterns/grid

static/sass/styles.scss
```diff
@@ -48,11 +48,19 @@
...
 
-.empty-state-message {
-  .icon {
-    height: 100%;
-    width: 35%;
-  }
+.u-empty-state-icon {
+  height: 2.5rem;
+  margin-bottom: 20px;
+  margin-top: 25px;
```
edlerd, line 54: We shouldn't use `px` values, but always prefer the `rem` unit or even better the `$sp-large` or alike variables from vanilla

edlerd: Gives this a read, then you might not need css for it at all: https://vanillaframework.io/docs/settings/spacing-settings

static/sass/styles.scss
```diff
@@ -48,11 +48,19 @@
   max-width: none;
 }
 
-.empty-state-message {
-  .icon {
-    height: 100%;
-    width: 35%;
-  }
+.u-empty-state-icon {
```
edlerd, line 51: I just realized now, the `u-` prefix is meant for utility classes, not sure if it's needed here.
I started to group the CSS in files and import them, maybe add one for the empty state?
Then in the file, wrap the whole content with a class and give that to the outer wrapper of the `EmptyState` component. This way the content of the file can't leak into other use cases and the browser can ignore (in all other cases) all the content in that file easily by just looking at the wrapper class.

edlerd (approved):
Just some possible optimization for the css.

PR #198, opened 2023-02-13
Adjust start stop buttons in instance list, add force stop option WD-1961

edlerd:


static/js/pages/instances/InstanceList.tsx
```diff
@@ -237,20 +258,17 @@ const InstanceList: FC = () => {
               items={[
                 <StartStopInstanceBtn
                   key="startstop"
-                  className="u-instance-actions"
+                  className={classnames(
+                    "u-instance-actions",
+                    starting.concat(stopping).includes(instance.name) &&
+                      "u-hide"
```
lorumic, line 264: I suggest: `"u-hide": [...starting, ...stopping].includes(instance.name)`
Shorter and easier to read and understand.

static/js/pages/instances/actions/StartStopInstanceBtn.tsx
```diff
@@ -2,105 +2,145 @@ import React, { FC, useState } from "react";
...
+          confirmationExtra={
+            <span className="u-float-left">
+              <CheckboxInput
+                inline
+                label="Force stop"
+                defaultChecked={isForce}
+                onClick={() => setForce((prev) => !prev)}
+              />
+            </span>
+          }
```
lorumic, line 139: I don't really like the fact that this checkbox gets the focus automatically once the modal gets opened. Is this an intended/desired behaviour?

edlerd: Right, that is a little wired. Good catch.

static/js/pages/instances/InstanceList.tsx
edlerd: Yeah, that is better indeed.

edlerd: I kept the `concat` version, because prettier allows it on fitting on one line. But the `:` version I adopted, because that is easier to read.

lorumic (approved):

PR #200, opened 2023-02-14
Add create and delete for storage pools WD-2011

edlerd:


static/js/pages/storages/StorageForm.tsx
```diff
@@ -0,0 +1,160 @@
...
+          size: size ? `${size}GiB` : undefined,
+        },
+      };
+
+      createStoragePool(storagePool, panelParams.project)
+        .then(() => {
+          void queryClient.invalidateQueries({
+            queryKey: [queryKeys.storage],
+          });
+          panelParams.clear();
```
lorumic, line 65: Here we should pass a success notification to the storage list page.

static/js/pages/storages/StorageForm.tsx
```diff
@@ -0,0 +1,160 @@
...
+                onChange={formik.handleChange}
+                value={formik.values.driver}
+                required
+                stacked
+              />
+              <Input
+                id="size"
+                name="size"
+                type="number"
+                help="Optional, defaults to 20% of free disk space, >= 5 GiB and <= 30 GiB"
```
lorumic, line 121: I'm confused, the value should be between 5 and 30? Are decimals allowed? (I already know the answer from what the BE returns: no)
The point is that I tried with values lower than 5 and it succeeds. With 31 also succeeds. Where do those extremes come from?
Maybe some constraints should be added by using the `min`, `max` and `step` properties.

static/js/pages/storages/StorageSize.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const StorageSize: FC<Props> = ({ storage }) => {
+  const { data: resources } = useQuery({
+    queryKey: [queryKeys.storage, storage.name, queryKeys.resources],
+    queryFn: () => fetchStorageResources(storage.name),
+  });
+
+  if (!resources) {
+    return null;
```
lorumic, line 20: Maybe we should return something more meaningful for the user in this case?

static/js/pages/storages/StorageForm.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { createStoragePool } from "api/storages";
+import { getSourceHelpForDriver, storageDrivers } from "util/storageOptions";
+
+const StorageForm: FC = () => {
+  const panelParams = usePanelParams();
+  const notify = useNotification();
+  const queryClient = useQueryClient();
+  const controllerState = useState<AbortController | null>(null);
+
+  const SnapshotSchema = Yup.object().shape({
```
lorumic, line 24: This should be called `StorageSchema` or similar, I guess.

static/js/pages/storages/StorageForm.tsx
edlerd: This is from the docs, the default is 20%, but in that range. I'll remove all the help text apart from optional.

lorumic: That "Optional" alone seems out of place; the "Description" is also optional, but we don't have a help text below it just to say that it is optional.

edlerd: Right, I circled it back to "When left blank, defaults to 20% of free disk space. Default will be between 5GiB and 30GiB". Hope it is less confusing?

lorumic: A bit less, but still confusing. I cannot think of a better wording though, so let's keep it as it is for the time being, and maybe iterate on it later.

lorumic (approved):

PR #202, opened 2023-02-15
Adjust navigation to match new design [WD-1955]

lorumic:


edlerd:
It works better on very small screens / mobile. A couple bugs that need to be fixed:
in mid size
 - on the instance list  (~1k pixels width) the menu doesn't close. Closing works on other pages though.
 - the arrow to open/close is missing
in small size
 - we should always take the full width. It looks wired, if the nav is overlaying half of the content
 - adding a close x button on the top of the menu -- the arrow to close the menu is on the bottom. On actual devices it might be hidden, due to the url bar being there and making some part of screen.height not visible (it is shifted down and up, based on the scrolling behavior on the page)

edlerd (approved):

PR #204, opened 2023-02-17
Overhaul the instance detail panel [WD-1962]

lorumic:


static/js/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -30,107 +50,183 @@ const InstanceDetailPanel: FC<Props> = ({ instance, notify, onClose }) => {
...
-            className="u-no-margin--bottom"
-            hasIcon
-            onClick={onClose}
-          >
-            <i className="p-icon--close"></i>
-          </Button>
+        <div className="p-panel__title-wrapper">
+          <div className="p-panel__title">
+            <span className="u-no-margin--bottom">Instance summary</span>
+            <span className="u-no-margin--bottom u-truncate">
```
edlerd, line 62: Maybe add a title with content "view more", so when truncated you can see the text on hover.

static/js/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -30,107 +50,183 @@ const InstanceDetailPanel: FC<Props> = ({ instance, notify, onClose }) => {
         );
       });
```
edlerd, line 51: When we add `?? []` to the end of  `getIpAddresses`, it always returns an array. Then we can save a bunch of checks down below.

static/js/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -30,107 +50,183 @@ const InstanceDetailPanel: FC<Props> = ({ instance, notify, onClose }) => {
...
-                <Row>
-                  <Col
-                    size={4}
-                    className="u-truncate"
-                    title={instance.config["image.description"]}
-                  >
-                    {instance.config["image.description"]}
-                  </Col>
-                </Row>
+                {ip4Addresses && ip4Addresses.length > 0 ? (
```
edlerd, line 151: Here we can remove the check with the above function change.

static/js/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -30,107 +50,183 @@ const InstanceDetailPanel: FC<Props> = ({ instance, notify, onClose }) => {
...
+                  "-"
+                )}
+              </td>
+            </tr>
+            <tr>
+              <th className="u-text--muted">Architecture</th>
+              <td>{instance.architecture}</td>
+            </tr>
+            <tr>
+              <th className="u-text--muted">PID</th>
```
edlerd, line 197: Did you find out why the PID is important to display it here?

static/js/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -139,46 +235,61 @@ const InstanceDetailPanel: FC<Props> = ({ instance, notify, onClose }) => {
...
+            <Col
+              size={1}
+              className="p-snapshot-link p-text--small u-align--right u-no-margin--bottom u-truncate"
             >
-              Manage snapshots
-            </Link>
-          </Col>
+              <strong>
+                <Link
+                  to={`/ui/${instance.project}/instances/${instance.name}/snapshots`}
```
edlerd, line 267: This link should always be visible, also with no snapshots available. Otherwise, it is hard to create your first snapshot.

static/js/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -139,46 +235,61 @@ const InstanceDetailPanel: FC<Props> = ({ instance, notify, onClose }) => {
...
+            instance.snapshots
+              ?.slice(0, RECENT_SNAPSHOT_LIMIT)
+              .map((snapshot) => (
+                <Row key={snapshot.name}>
+                  <Col size={2} className="u-truncate" title={snapshot.name}>
+                    {snapshot.name}
+                  </Col>
+                  <Col
+                    size={2}
+                    className="p-snapshot-creation p-text--small u-align--right u-text--muted u-no-margin--bottom"
```
edlerd, line 287: This looks a little odd, because the text is smaller - I think in the design it is normal size but italics.

static/js/pages/instances/actions/OpenVgaBtn.tsx
```diff
@@ -25,11 +25,11 @@ const OpenVgaBtn: FC<Props> = ({
     if (instance.status !== "Running") {
       return "Instance must be running";
     }
-    return undefined;
+    return "VGA Console";
   };
 
   return (
-    <Tooltip message={getTooltipMessage()} position="btm-center">
+    <Tooltip message={getTooltipMessage()} position="top-right">
```
edlerd, line 32: This should be the `OpenConsoleBtn` now, that links to either the text console (for containers) or the graphical console (for virtual machines).

static/sass/_pattern_icon.scss
```diff
@@ -11,30 +11,30 @@
   .p-icon--oval-green {
     @extend %icon;
 
-    background-image: url("static/assets/img/status/status-succeeded.svg");
+    background-image: url("static/assets/img/status/status-running.svg");
```
edlerd, line 14: Do we need to include the custom icons? Can't we directly use the vanilla `p-icon--status-foo` icons? I think then we can remove the SVG's and this extra SCSS bit here.

static/sass/_instance_detail_panel.scss
```diff
@@ -0,0 +1,86 @@
+.p-instance-detail-panel {
```
edlerd, line 1: Good idea to extract this in a dedicated file!

static/js/pages/instances/InstanceDetailPanel.tsx
lorumic: No, as I said, I assume it's there for a reason, so I don't really want to question that. I might quickly ask it out of curiosity during the next UX/UI sync, and that's all.

static/js/pages/instances/InstanceDetailPanel.tsx
lorumic: This needs design input because showing a link that says "View all" when you have "None" written just below seems counterintuitive to me.
Please discuss this with design and once clarified I'll be happy to update.

static/js/pages/instances/actions/OpenVgaBtn.tsx
lorumic: There is no `OpenConsoleBtn` now and it would not be the right timing to create it since the VD of the instance detail page are still pending validation. Once this get clarified, it can be updated here as well, but out of the scope of this PR I'd say since it depends on a design that is still pending approval. Same applies to the instance actions probably, if we want to switch to the "icons only" view, I'd keep them as they are (namely, only the start/stop button) until the VD gets updated, and take care of the changes in a later PR.

static/js/pages/instances/InstanceDetailPanel.tsx
edlerd: I discussed with Alex, and we agreed this will likely change as soon as we start working on snapshots in a couple of weeks.
For now, we can change the label to from "View all" to "Manage", when all items in this view are visible (i.e. no items are hidden).

lorumic: Sounds good, I'm not sure I understood correctly the part about "all items in this view are visible" - can they even be hidden? By the user? Apart from that, I changed the label and removed the check that was hiding it when no snapshot was present.

edlerd: What I meant is:
We limit the number of displayed snapshots in the side panel. If there are more available, the label should say "view all", otherwise "manage"

static/js/pages/instances/InstanceList.tsx
```diff
@@ -215,19 +215,23 @@ const InstanceList: FC = () => {
         },
         {
           content: instance.state?.network?.eth0?.addresses
-            .filter((item) => item.family === "inet")
+            .filter(
+              (item) =>
+                item.family === "inet" && !item.address.startsWith("127")
+            )
             .map((item) => item.address)
-            .filter((address) => !address.startsWith("127"))
```
edlerd, line 220: This is the same as before. The old version seems easier to read to me. Same for the `inet6` below. Why change both?

lorumic: We are filtering the same array twice. Not very good from a performance point of view. What do we care about more, performance or readability? (Personally, I find it perfectly readable this way too...)

edlerd: ok, seems right.

al-bakalova:
Looks good to me! Nice work integrating the split layout!
@lyubomir-popov could you please check the styles one last time? Thank you

edlerd (approved):

PR #206, opened 2023-02-20
Add storage details WD-2289

edlerd:
* Improved the storage list view
* Added a storage detail page
* Added simple (directory based) storage creation (done last week, not in the PR, but still worth to give it a quick review)

### DIFF
--- a/static/js/pages/images/ImageName.tsx
+++ b/static/js/pages/images/ImageName.tsx
@@ -1,0 +1,28 @@
+import React, { FC } from "react";
+import { queryKeys } from "util/queryKeys";
+import { fetchImage } from "api/images";
+import { useQuery } from "@tanstack/react-query";
+
+interface Props {
+  id: string;
+  project: string;
+}
+
+const ImageName: FC<Props> = ({ id, project }) => {
+  const { data: image } = useQuery({
+    queryKey: [queryKeys.images, id],
+    queryFn: () => fetchImage(id, project),
+  });
+
+  const label = image?.properties.description
+    ? image.properties.description
+    : id;
+
+  return (
+    <div className="u-truncate" title={label}>
+      {label}
+    </div>
+  );
+};
+
+export default ImageName;
--- a/static/js/pages/storages/StorageDetail.tsx
+++ b/static/js/pages/storages/StorageDetail.tsx
@@ -1,20 +1,15 @@
 import React, { FC } from "react";
-import { Link, useParams } from "react-router-dom";
+import { useParams } from "react-router-dom";
 import { useQuery } from "@tanstack/react-query";
 import BaseLayout from "components/BaseLayout";
 import NotificationRow from "components/NotificationRow";
 import { queryKeys } from "util/queryKeys";
 import useNotification from "util/useNotification";
-import { List, Row } from "@canonical/react-components";
+import { Row } from "@canonical/react-components";
 import Loader from "components/Loader";
 import { fetchStorage } from "api/storages";
 import StorageSize from "pages/storages/StorageSize";
-import ImageName from "pages/images/ImageName";
-
-interface UsedByType {
-  name: string;
-  project: string;
-}
+import StorageUsedBy from "pages/storages/StorageUsedBy";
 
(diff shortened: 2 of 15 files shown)


static/js/pages/storages/StorageDetail.tsx
```diff
@@ -0,0 +1,141 @@
...
+              </td>
+            </tr>
+            <tr>
+              <th className="u-text--muted">Images using this storage</th>
+              <td>
+                {usedByImages?.length ? (
+                  <List
+                    className="u-no-margin--bottom"
+                    items={usedByImages.map((name) => ({ name }))}
+                  />
```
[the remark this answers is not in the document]
edlerd: It is kind of important, when you want to delete a storage pool and the images are still on it.
I think long run we'll have instance management in the ui, for now we just display the fingerprints.

static/js/pages/images/ImageName.tsx
```diff
@@ -0,0 +1,19 @@
...
+  id: string;
+}
+
+const ImageName: FC<Props> = ({ id }) => {
+  const { data: image } = useQuery({
+    queryKey: [queryKeys.images, id],
+    queryFn: () => fetchImage(id),
+  });
+
+  return <>{image ? image.properties.description : id}</>;
```
lorumic, line 16: Shouldn't we use the `project` parameter for the query here?
Also, since we're falling back on the `id` (the fingerprint), which is a very long chunk of text, I'd still apply the `truncate` and `title` to be safe.

edlerd: True about the truncate / fallback.

The images are independent of projects, so it should be fine with the network call as it is.

lorumic (approved):

al-bakalova:
Screenshot 2023-02-28 at 13 47 01

When deleting a storage, please use the following message:

Title: Error
Body: The storage pool is currently used.

- Please align the "used by" column to the left.

- Actions - please remove the button stroke and use only the icon. This should be closer to the instance list (future improvements)

- Create storage pool: once a driver, different than ZFS is selected, please hide the hint. It makes sense to initially have it and to keep it when ZFS is selected. When a different driver type is selected, the message is a bit misleading.

al-bakalova:
Thanks for applying the requested changes!

static/js/pages/storages/StorageUsedBy.tsx
```diff
@@ -0,0 +1,152 @@
...
+        return path.startsWith(`/1.0/${type}`);
+      })
+      .map((path) => {
+        const name: string = path.split("/").slice(-1)[0] ?? "";
+        return {
+          name: name.split("?")[0],
+          project: name.includes("?project=")
+            ? name.split("=").slice(-1)[0]
+            : project,
+          instance: type === "snapshots" ? path.split("/")[3] : "",
```
lorumic, line 47: I propose adding one or more comments above/inside this chunk with one or more example paths to give some context, so that when one reads it they can figure out what's happening more easily and without recurring to external sources (the API specs).

static/js/pages/storages/StorageUsedBy.tsx
```diff
@@ -0,0 +1,152 @@
...
+  data[INSTANCES].sort((a, b) => {
+    return a.project < b.project
+      ? -1
+      : a.project > b.project
+      ? 1
+      : a.name < b.name
+      ? -1
+      : a.name > b.name
+      ? 1
+      : 0;
```
lorumic, line 67: Why only instances are sorted this way?

static/js/pages/storages/StorageUsedBy.tsx
edlerd: Agree, this was quite hacky, I changed it to be more straight forward. Also extracted this into a util file, so we can reuse it for the other pages.

static/js/pages/storages/StorageUsedBy.tsx
edlerd: true dat.

lorumic (approved):
LGTM

PR #209, opened 2023-02-23
Make the instance list responsive, hide columns based on the actual content

edlerd:
* instance list columns get hidden on resize based on the current content


edlerd:
Will be part of #204

PR #210, opened 2023-02-24
Cleanup instance summary WD-2288

edlerd:
* make the header fully responsive (hide filters when too few space is available -- to keep it simple)
* make the aside full height to fix horizontal alignment
* align text in the status column with header
* fix spacing around status icon
* make table rows clickable and remove the icon for opening the aside
* add wrapper for all instance table styles to avoid leaking into other parts
* use titles instead of tooltip where appropriate
* logo uses same text font and size as other headings
* reposition the close for the navigation on mobile


al-bakalova:
Good job applying these changes! A few small comments/requests:

- at 1095 screen width the instance label and the search still overlap Screenshot 2023-02-28 at 13 37 20
- Please make the page header sticky on top of the page
Screenshot 2023-02-28 at 13 42 31
- Please make sure the aside panel's header is also sticky
- Please make the column headers sticky as well. Only the table content is to be scrollable.
- When a notification is displayed on top of the page, please keep it sticky on top of the table. Again, only the table content is to remain scrollable, the rest is to always be visible to the user.

Thank you

static/js/pages/instances/DownloadButton.tsx
```diff
@@ -48,19 +48,18 @@ const DownloadButton: FC<DownloadButtonProps> = ({
...
-      </Button>
-    </Tooltip>
+    <Button
+      appearance="base"
+      className="u-no-margin--bottom"
+      onClick={() => void handleDownload()}
+      title={`Download ${fileName}`}
+      aria-label={`Download ${fileName} file`}
+      has-icon
+      is-dense
```
yurii-vasyliev, line 58: `dense`

static/js/pages/instances/DownloadButton.tsx
```diff
@@ -48,19 +48,18 @@ const DownloadButton: FC<DownloadButtonProps> = ({
...
+      ref={buttonRef}
+    >
+      <Icon name="begin-downloading" alt="download" />
+    </Button>
```
yurii-vasyliev, line 62: Aria-label is missing

static/js/pages/instances/actions/OpenTerminalBtn.tsx
```diff
@@ -21,20 +21,16 @@ const OpenTerminalBtn: FC<Props> = ({
...
-        dense
-        hasIcon
-        onClick={handleOpen}
-        disabled={isDisabled}
-      >
-        <i className="p-icon--code" />
-      </Button>
-    </Tooltip>
+      <i className="p-icon--code" />
+    </Button>
```
yurii-vasyliev, line 33: Aria-label is missing

static/js/pages/instances/actions/OpenVgaBtn.tsx
```diff
@@ -29,19 +29,18 @@ const OpenVgaBtn: FC<Props> = ({
...
+      dense
+      hasIcon
+      onClick={handleOpen}
+      disabled={
+        instance.status !== "Running" || instance.type !== "virtual-machine"
+      }
+      title={getTitle()}
+    >
+      <i className="p-icon--canvas" />
+    </Button>
```
yurii-vasyliev, line 43: Aria-label is missing

yurii-vasyliev (changes requested):

static/sass/_pattern_navigation.scss
```diff
@@ -1,39 +1,42 @@
...
+    .p-panel__logo-image {
+      height: 54px;
     }
-  }
 
-  .align-with-logo {
-    margin-right: -$sph--small;
-    padding-top: $spv--small;
+    .logo-text {
+      left: 61px;
```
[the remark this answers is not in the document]
edlerd: No, we discussed this with @lyubomir-popov on friday, the Logo text should be the exact same size, font, etc as the other headings on that line of the screen. Making it actual text ensures that.

al-bakalova:
Good job implementing all my requests! Thank you

al-bakalova:
Two small changes though:
- Could you please make sure the instance count in the header "# instances" is aligned with the "Name" column header?
- Could you please align the "Create new" button with the end of the table? Looks like there is a double spacing.

Everything else looks good!

al-bakalova:
On my screen resolution, which is 1470x956, when the side panel is open, the search and filters are not visible - this is something that we will resolve as a future improvement.
Screenshot 2023-03-01 at 9 16 14

However, once I close the side navigation, there is plenty of space for the search and filters to reappear, but they do not.
Screenshot 2023-03-01 at 9 17 04

Is there a way to take the state of the side navigation into account when hiding/ showing the search and the filters?

Also, the table pagination seems to be missing and it looks like we are currently using an endless list. 
Let's discuss the best way to solve this. In my opinion it should be sticky at the bottom and only the column rows should be scrollable.

lorumic:
I was reviewing this and I noticed one design thing that I wanted to confirm. Are we missing a bottom border/hr element below the sticky header in the "Instance summary" panel? @al-bakalova @edlerd

al-bakalova:
@lorumic Yes, we have removed the HR. The side panel is going to be slightly redesigned as part of a future improvement so please do not pay too much attention right now :)

static/js/pages/instances/InstanceList.tsx
```diff
@@ -283,62 +302,65 @@ const InstanceList: FC = () => {
...
-        <div className="p-panel">
-          <div className="p-panel__content p-instance-content">
+          <div className="p-panel__header instance-list-header">
+            <div className="instance-header-left">
+              <h4>
+                {visibleInstances.length}
+                &nbsp;
+                {visibleInstances.length === 1 ? "Instance" : "Instances"}
+              </h4>
+              <SearchBox
```
yurii-vasyliev, line 318: Sorry, I missed that yesterday. Is there a way to add labels for all the filters here with the `u-off-screen` class?

static/js/pages/instances/actions/StartStopInstanceBtn.tsx
```diff
@@ -56,7 +56,7 @@ const StartStopInstanceBtn: FC<Props> = ({
   const canBeStarted =
     instance.status === "Stopped" || instance.status === "Frozen";
   const canBeStopped =
-    instance.status === "Running" ||
+    instance.status.endsWith("ing") ||
```
yurii-vasyliev, line 59: This approach might be working, but it's not straightforward for other developers. I would created a variable instead, like:
```
const startStatuses = ["Stopped", "Frozen"];
const stopStatuses = ["Running", ...];

const isDisabled = [...startStatuses, ...stopStatuses].includes(instance.status); // Or just pass that directly to `isDisabled` prop
```

static/js/pages/instances/InstanceList.tsx
edlerd: Good catch! I just added the aria-labels.

static/js/pages/instances/actions/StartStopInstanceBtn.tsx
edlerd: Good idea, adjusted to this.

al-bakalova:
Looks good to me, thank you!

yurii-vasyliev (approved):
LGTM

PR #211, opened 2023-02-28
Instance create notify WD-1964

edlerd:
* close sidepanel right after submitting instance creation form
* adjust notifications for instance creation

### DIFF
--- a/static/js/components/NotificationRow.tsx
+++ b/static/js/components/NotificationRow.tsx
@@ -24,11 +24,31 @@ const NotificationRow: FC<Props> = ({ notify }) => {
   if (!notify.notification) {
     return null;
   }
-  const { type, message } = notify.notification;
+  const { actions, title, type, message } = notify.notification;
+  const figureTitle = () => {
+    if (title) {
+      return title;
+    }
+    switch (type) {
+      case "negative":
+        return "Error";
+      case "positive":
+        return "Success";
+      case "information":
+        return "Info";
+      default:
+        return "";
+    }
+  };
   return (
     <div ref={ref}>
       <Row>
-        <NotificationComponent severity={type} onDismiss={notify.clear}>
+        <NotificationComponent
+          title={figureTitle() || undefined}
+          actions={actions}
+          severity={type}
+          onDismiss={notify.clear}
+        >
           {message}
         </NotificationComponent>
       </Row>
--- a/static/js/pages/instances/CreateInstanceForm.tsx
+++ b/static/js/pages/instances/CreateInstanceForm.tsx
@@ -1,4 +1,4 @@
-import React, { FC, MouseEvent, useState } from "react";
+import React, { FC, MouseEvent, ReactNode, useState } from "react";
 import {
   Button,
   Col,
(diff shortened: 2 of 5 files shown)


al-bakalova:
Looks good to me.
For the info message please use the following copy:

Title: Instance creation
Body: A new instance has successfully been requested.

al-bakalova:
Looks good to me, thank you!

static/js/components/NotificationRow.tsx
```diff
@@ -24,11 +24,31 @@ const NotificationRow: FC<Props> = ({ notify }) => {
...
+    }
+    switch (type) {
+      case "negative":
+        return "Error";
+      case "positive":
+        return "Success";
+      case "information":
+        return "Info";
+      default:
+        return null;
```
yurii-vasyliev, line 40: I'd return an empty string to stick to the same type. And below, when you call it, you could do:
```
title={figureTitle() ?? undefined}
```

static/js/pages/instances/CreateInstanceForm.tsx
```diff
@@ -59,85 +75,103 @@ const CreateInstanceForm: FC = () => {
...
-  interface FormValues {
-    name: string;
-    image: RemoteImage | undefined;
-    instanceType: string;
-    profiles: string[];
-    yaml?: string;
-    unchangedYaml?: string;
+  function notifyInProgress() {
+    instanceListNotify?.info(
+      <>A new instance has successfully been requested.</>,
```
yurii-vasyliev, line 80: I think it should be `A new instance has been successfully requested.` or just `A new instance has been requested.`

static/js/pages/instances/CreateInstanceForm.tsx
```diff
@@ -59,85 +75,103 @@ const CreateInstanceForm: FC = () => {
...
+  function notifyLaunchedAndStarted(instanceLink: ReactNode) {
     instanceListNotify?.success(
-      <>
-        Instance{" "}
-        <Link to={`/ui/${panelParams.project}/instances/${instanceName}`}>
-          {instanceName}
-        </Link>{" "}
-        {action}.
-      </>
+      <>Instance {instanceLink} successfully launched and started.</>
```
yurii-vasyliev, line 87: Here and everywhere else - maybe it's better to extend the type for the first argument to `string | ReactNode`, so you don't have to wrap it with a fragment every time?

yurii-vasyliev (changes requested):

static/js/pages/instances/CreateInstanceForm.tsx
edlerd: The `instanceLink` used inside is a `ReactNode`, so we'd still need the fragment wrapper.

static/js/components/NotificationRow.tsx
edlerd: That is not valid typescript, because empty string can't be used as a left-hand side of the `??` chain.

static/js/pages/instances/CreateInstanceForm.tsx
edlerd: Simplified the message.

static/js/components/NotificationRow.tsx
yurii-vasyliev: Sure, you're right. `title={figureTitle() || undefined}` then.

edlerd: right, that works and is better.

yurii-vasyliev (approved):

PR #213, opened 2023-03-02
Overhaul instance creation experience WD-2200 WD-2199

edlerd:
* moved instance create form to a dedicated page
* created reusable form components to create/customize an instance/profile
* created a reusable navigation component for the instance form
* introduced global context for notifications
* image selector: filter images by architectures supported by the current environment configuration
* added instance configuration tab to instance detail page with a form (view mode is pending design input)
* added profile creation overhaul
* added profile editing overhaul


static/js/App.tsx
```diff
@@ -33,6 +34,10 @@ const App: FC = () => {
         path="/ui/:project/instances"
         element={<ProtectedRoute outlet={<InstanceList />} />}
       />
+      <Route
+        path="/ui/:project/instances/1/create"
```
yurii-vasyliev, line 38: That `1` added on purpose, right?

static/js/pages/instances/actions/snapshots/CreateSnapshotBtn.tsx
```diff
@@ -2,30 +2,27 @@ import React, { FC, useState } from "react";
...
 
 interface Props {
   instance: LxdInstance;
-  notify: NotificationHelper;
 }
 
-const CreateSnapshotBtn: FC<Props> = ({ instance, notify }) => {
+const CreateSnapshotBtn: FC<Props> = ({ instance }) => {
+  const notify = useNotify();
   const [isModal, setModal] = useState(false);
```
yurii-vasyliev, line 13: You could probably that by naming it `[open, setOpen]` or `[modalOpen, setModalOpen]`

static/js/pages/instances/forms/InstanceDetailsForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { isContainerOnlyImage, isVmOnlyImage } from "util/images";
+import { instanceCreationTypes } from "util/instanceOptions";
+import { FormikProps } from "formik/dist/types";
+import { FormValues } from "pages/instances/CreateInstanceForm";
+import { RemoteImage } from "types/image";
+
+export interface InstanceDetailsFormValues {
+  name: string;
+  description?: string;
+  image: RemoteImage | undefined;
```
yurii-vasyliev, line 14: Could be `image?: RemoteImage`

static/js/pages/profiles/ProfileSelector.tsx
```diff
@@ -77,10 +72,14 @@ const ProfileSelector: FC<Props> = ({
 
   return (
     <>
-      <Label className="u-sv-1">Apply profiles</Label>
-      <div className="u-text--muted u-sv1">
-        Add profiles and create a hierarchy
-      </div>
+      <Row>
+        <Col size={12}>
+          <Label className="u-sv-1">Apply profiles</Label>
```
yurii-vasyliev, line 77: Do we need `label` here? If we need only the styles, maybe we could only apply needed styles or use existing Vanilla class?

static/js/util/useNotificationHelper.tsx
```diff
@@ -0,0 +1,64 @@
...
+import isEqual from "lodash/isEqual";
+import { useLocation } from "react-router-dom";
+
+const useNotificationHelper = (): NotificationHelper => {
+  const location = useLocation() as QueuedNotification;
+  const [notification, setNotification] = useState<Notification | null>(
+    location.state?.queuedNotification ?? null
+  );
+
+  window.history.replaceState({}, "");
```
yurii-vasyliev, line 16: Do we need this on every re-render?

yurii-vasyliev (approved):
Please also check my comments.

static/js/App.tsx
edlerd: Yes, to reuse the /instances path prefix but ensure we don't block a detail page for an instance. Because instance names can't start with a number. It's a bit wired maybe, but I didn't find a better solution.

static/js/pages/profiles/ProfileSelector.tsx
edlerd: This is a WIP component, that will fundamentally change when the design is ready. Currently we need the label as the rest of the component is custom.

static/js/util/useNotificationHelper.tsx
edlerd: This is a leftover from the previous implementation wit sharednotufy context. I'll clean this up.

static/js/App.tsx
lorumic: What about `/instances?create` or `/instances?new`?

static/js/pages/instances/actions/snapshots/CreateSnapshotBtn.tsx
edlerd: yeah that is better.

static/js/pages/instances/forms/InstanceDetailsForm.tsx
edlerd: yes!

static/js/App.tsx
edlerd: With `?` the URL is also not so nice either.

I'd like to use `/instance-create`, but I don't know how to convince the `NavLink` from react-router for `/instances`, that it should also be active for a URL that does not match its own target.

static/js/pages/instances/forms/DevicesForm.tsx
```diff
@@ -0,0 +1,215 @@
...
+      <Button onClick={addDeviceRow} type="button" hasIcon>
+        <Icon name="plus" />
+        <span>Add device</span>
+      </Button>
+      <Link
+        href="https://linuxcontainers.org/lxd/docs/latest/reference/devices/"
+        target="_blank"
+        rel="noreferrer"
+      >
+        What is a device <Icon name="external-link" />
```
lorumic, line 209: The style of the icon is inconsistent with that of the link. What I did for the empty state was to apply the link style to it through a class: https://github.com/canonical/lxd-ui/blob/main/static/sass/_empty_state.scss#L10
However, there is a problem with this: when a link is visited, its text colour changes, but that of the icon does not.

static/js/App.tsx
yurii-vasyliev: @edlerd , what if you make it `/instances/new` and put that route before this one https://github.com/canonical/lxd-ui/blob/main/static/js/App.tsx#L36 ? Will the router catch it up earlier?

edlerd: Yeah, that will work, but I'd like to avoid hogging the detail page for an instance with the name `new`, hence the workaround with the `/1`, because instances names can't start with a number.

lorumic: It's very hacky, and confusing for the user, to have that `/1/` in the path. I honestly find it more natural to have the `?` in the URL - what do you think @yurii-vasyliev ?

static/js/pages/instances/forms/DevicesForm.tsx
edlerd: I think it is quite common to not color the icon. Especially as the color will differ with a visited link, I lean towards keeping it as it is.

lorumic: @al-bakalova how should we proceed here? I ask this so that we can also adapt the style for empty states, if needed.

static/js/App.tsx
al-bakalova: Using the /1/ might mislead the user into thinking there is some sort of additional grouping. I'd try to avoid this approach. Maybe have something like "create-new" or "create-new-instance" can prevent us from having an issue?

static/js/pages/instances/forms/DevicesForm.tsx
al-bakalova: I'd apply the link style to the icon. After all, the icon itself is also part of the link.

static/js/App.tsx
edlerd: true, lets keep it simple. if someone has an instance with the name create-new it would be wired anyways.

al-bakalova:
Looks good, thanks for applying all the small changes!

static/js/pages/instances/forms/SnapshotsForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+            options={booleanFields(isInstance)}
+            value={formik.values.snapshots_schedule_stopped}
+          />
+          <hr />
+          <div className="snapshot-schedule">
+            <RadioInput
+              labelClassName="right-margin"
+              label="Cron syntax"
+              checked={cronSyntax}
+              onChange={() => setCronSyntax(true)}
```
lorumic, line 91: Why not setting back the value to the empty string? I find this behaviour confusing:

Peek 2023-03-07 12-43

1) "Cron syntax" selected by default, with no default value (empty string).
2) I switch to "Manual configuration", every minute is selected by default.
3) I switch back to "Cron syntax" and I find a value filled in that is equivalent to every minute.

Is this a desired behaviour? Or is this just confusing the user? (also for @al-bakalova @yurii-vasyliev )

edlerd: Right, clearing the value is a better approach here. Also, the default for manual would better be `@daily`.

PR #218, opened 2023-03-07
Use empty state component for empty snapshot list WD-2127

edlerd:
* changed the empty state for snapshots (added a create button and aligned the visual appearance)

### DIFF
--- a/static/js/pages/instances/InstanceSnapshots.tsx
+++ b/static/js/pages/instances/InstanceSnapshots.tsx
@@ -1,5 +1,5 @@
-import React, { FC } from "react";
-import { Col, Icon, MainTable, Row } from "@canonical/react-components";
+import React, { FC, useState } from "react";
+import { MainTable, Row } from "@canonical/react-components";
 import { isoTimeToString } from "util/helpers";
 import DeleteSnapshotBtn from "./actions/snapshots/DeleteSnapshotBtn";
 import RestoreSnapshotBtn from "./actions/snapshots/RestoreSnapshotBtn";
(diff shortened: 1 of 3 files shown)


static/js/pages/instances/InstanceSnapshots.tsx
```diff
@@ -83,45 +87,40 @@ const InstanceSnapshots: FC<Props> = ({ controlTarget, instance, notify }) => {
...
+            rows={rows}
+            paginate={30}
+            responsive
+            sortable
+            className="u-table-layout--auto"
+          />
+        ) : (
+          <EmptyState
+            iconName="settings"
+            iconClass="p-empty-instances"
```
lorumic, line 114: In order to apply the correct colour, this should be changed to some new class called `p-empty-snapshots` which includes the mixin of the correct (temporary) icon:
```
.p-empty-snapshots {
  @include vf-icon-settings($color-mid-light);
}
```

edlerd: I changed it [here](https://github.com/canonical/lxd-ui/pull/213/files#diff-f5a5b0429dba7d86dd96396c39539ab3a091d90536bccb9db008a3dbc487f83dR102). Will pick that change to this branch, to allow reusing of the icon class.

lorumic: This is a different thing: not the icon for a link, but the icon representing the snapshots - as it was `vf-icon-containers` for instances, it is `vf-icon-settings` (temporarily) for snapshots.
It is to this icon that the correct colour (`$color-mid-light`) should be applied.

edlerd: Right, I mixed up the external link icon and the snapshot icon. Now it should be fixed.

lorumic (approved):
LGTM

PR #223, opened 2023-03-09
Adjust profile and image selector after design review WD-2561

edlerd:
* adjusted profile selector to be simple dropdowns with add/remove logic
* adjust image selector to have more space between filters and content, also hide filters in mobile
* adjust creation-navigation to have disabled style
* ensure all form elements in instance/profile creation/editing have a placeholder


al-bakalova:
Good work! One final request:
- When browsing the "Select base image" modal on smaller screens, please hide the Alias column to make sure the actions are always visible.

al-bakalova:
When resizing the browser window, the space below the table seems to double. Can we keep it consistent?
Some screenshots to illustrate it -
Screenshot 2023-03-14 at 10 31 24

Screenshot 2023-03-14 at 10 31 47

edlerd:
> When resizing the browser window, the space below the table seems to double. Can we keep it consistent?

@al-bakalova The spacing was adjusted, please check again.

al-bakalova:
Looks good to me. Thanks

src/pages/profiles/ProfileSelector.tsx
```diff
@@ -70,109 +36,66 @@ const ProfileSelector: FC<Props> = ({ project, selected, setSelected }) => {
...
-                );
-                if (!profile) {
-                  return null;
+      {selected.map((value, i) => (
+        <Row key={value}>
+          <Col size={8}>
+            <Select
+              help={
+                i > 0 &&
+                "Additional profile, overrides settings specified through profiles above"
```
lorumic, line 54: Do we really need to show this help text below every added profile? Wouldn't it be enough to show it only on the last one?

src/pages/instances/forms/InstanceFormMenu.tsx
```diff
@@ -54,7 +62,19 @@ const InstanceFormMenu: FC<Props> = ({
...
-          <MenuItem label={YAML_CONFIGURATION} {...menuItemProps} />
+          {isConfigDisabled ? (
+            <li className="p-side-navigation__item">
+              <button
+                className="p-side-navigation__link p-button--base"
+                disabled={true}
+                title="Please select an image before adding custom configuration"
+              >
+                {YAML_CONFIGURATION}
+              </button>
```
lorumic, line 73: Why aren't we using react-component's `Button` here? (and also @ L39 above and in `ProfileFormMenu` @ L37)

Also, please note that `p-button--base` has `text-align: center;`. You might want to override that:

image
(disabled, no image selected)

vs

image
(enabled, after image selection)

lorumic (changes requested):

src/pages/instances/forms/InstanceFormMenu.tsx
edlerd: Good catch, adjusted the usages to `Button` and fixed the alignment / width.

src/pages/profiles/ProfileSelector.tsx
edlerd: You are right, having it only on the last element is a good idea.

edlerd:
@lorumic resolved the issues, can you give it a 2nd pass?

src/sass/_instance_forms.scss
```diff
@@ -36,6 +41,16 @@
       margin-bottom: 6px;
     }
 
+    .profile-action-btn {
+      margin-right: 1rem;
+      margin-top: 2.5rem;
+    }
+
+    .profile-delete-btn {
+      margin-left: 1rem;
```
lorumic, line 50: Shouldn't we use vanilla spacing variables when possible? The two horizontal margins here should have `$sph--large` instead of `1rem`.

src/pages/profiles/ProfileSelector.tsx
```diff
@@ -70,109 +36,99 @@ const ProfileSelector: FC<Props> = ({ project, selected, setSelected }) => {
...
+                  newSelection.splice(index + 1, 0, value);
+                  setSelected(newSelection);
+                }}
+                type="button"
+                aria-label="move profile down"
+                title="move profile down"
+                disabled={index === selected.length - 1}
+              >
+                <Icon name="chevron-down" />
+              </Button>
```
lorumic, line 110: Is there a particular reason to disallow the reordering of the first profile? It looks odd and unbalanced without the arrows on the first profile as well. I can understand that it cannot be deleted, but it should still be possible to move it down, right? Maybe you can hide the arrows only if there is just 1 profile.

edlerd: True, that makes sense to also show them on the 1st profile in case there are multiple.
Just pushed all the changes @lorumic can you give it another pass?

lorumic (approved):

PR #226, opened 2023-03-14
Prevent TypeError when image.description is undefined

lorumic:
The field `config["image.description"]` of an instance can be undefined. In such case, this can cause a `TypeError` when searching the instance list:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'toLowerCase')
```


edlerd (approved):
Good catch!

lorumic:
> Good catch!
> 
> There might be other cases where we access this field. I guess you checked those as well?

Yes, I checked and this was the only case where a function was called (`toLowerCase`) on `image.description`. In the other cases we are just assigning the value, so even if the value is undefined, that doesn't make the app crash.

PR #228, opened 2023-03-14
Overhaul Instance detail tab: Snapshots [WD-2552]

lorumic:
- Added search box.
- Moved "Create new" button inside the tab.
- Updated column labels (table headers).
- Moved notifications inside the tab.
- Fixed actions style.
- Displayed error for "name already in use" inside the creation form.
- Fixed alignment of "Stateful" column.


src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -1,30 +1,47 @@
...
-  controlTarget?: HTMLSpanElement | null;
   instance: LxdInstance;
 }
 
-const InstanceSnapshots: FC<Props> = ({ controlTarget, instance }) => {
+const InstanceSnapshots: FC<Props> = ({ instance }) => {
+  const [query, setQuery] = useState<string>("");
   const [isModalOpen, setModalOpen] = useState(false);
+
+  const visibleSnapshots =
```
edlerd, line 24: Better use `filteredSnapshots`. I also renamed it in the InstanceList on the pagination PR.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -73,28 +90,45 @@ const InstanceSnapshots: FC<Props> = ({ controlTarget, instance }) => {
...
+            <SearchBox
+              className="search-box margin-right"
+              type="text"
+              onChange={(value) => {
+                setQuery(value);
+              }}
+              placeholder="Search for snapshots"
+              value={query}
+              aria-label="Search for snapshots"
+            />
```
edlerd, line 111: I think the input is a little too wide on wide screens, also the responsive behaviour seems a bit off.

al-bakalova:
A few pending improvements:
- The error should appear in the tab and not above the tabs
Screenshot 2023-03-14 at 13 10 57
- Please remove the icons from the Restore and delete buttons. The correct button order (from left to right) should be Delete, Restore.
- The Restore and Delete buttons should only appear on hover and be always visible only on smaller screens
- The list of snapshots should be paginated and always display the pagination, item count and items per page (sticky footer)
- The name already in use error should go below the name feed and should be displayed as soon as the user enters the name, not after clicking on "Create snapshot" (see mockups)
- The "Expires at" field should be separated into two different fields "Expiration date" and "Expiration time" - both fields should use the browser options that we have combined in the previous version (see mockups)

src/pages/instances/InstanceSnapshots.tsx
lorumic: @al-bakalova what is the expected responsive behaviour here?

al-bakalova:
The additional improvements have been added to the following JIRA task. https://warthogs.atlassian.net/browse/WD-2629

src/pages/instances/InstanceSnapshots.tsx
al-bakalova: Please make sure to always fit the search and the Create new button on the same line.
As discussed earlier, please reuse the max width from the Instance header.

lorumic:
@al-bakalova All the comments in the previous review, plus the other things that we have discussed, have been addressed.

al-bakalova:
If possible, please make the Stateful info tooltip text to be on two lines.
Screenshot 2023-03-16 at 15 49 56

Confirmation messages/ " This action cannot be undone, and can result in data loss." should be on a new line
Screenshot 2023-03-16 at 15 53 12

Screenshot 2023-03-16 at 15 54 11

Please make the header and the tabs sticky. The search bar and the create new button should also be sticky on the top. The only content to be scrollable is the table content.

The page should be paginated. The footer with the count, pagination and items per page should be sticky at the bottom.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -71,46 +113,107 @@ const InstanceSnapshots: FC<Props> = ({ controlTarget, instance }) => {
...
+  const {
+    pageData: pageSnapshots,
+    currentPage,
+    paginate: setCurrentPage,
+  } = usePagination(rows, {
+    itemsPerPage: pageSize,
+    autoResetPage: true,
+  });
+
+  const updateTBodyHeight = () => {
```
edlerd, line 125: I wonder if we can generalize this, there is already [this shared](https://github.com/canonical/lxd-ui/blob/main/src/util/updateMaxHeight.tsx) in util. We could move this function there and use it from here and from the `InstanceList`?

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -71,46 +113,107 @@ const InstanceSnapshots: FC<Props> = ({ controlTarget, instance }) => {
     };
   });
 
+  const {
+    pageData: pageSnapshots,
+    currentPage,
+    paginate: setCurrentPage,
+  } = usePagination(rows, {
```
edlerd, line 120: We could create an adapter for `usePagination`, that also holds the `pageSize`. We currently need to define the latter separately every time (also in `InstanceList`). I think we will always want a `pageSize` when we use pagination, so it might be good to combine them. I think we should add this as a small story to the backlog to not extend the scope of this further.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -71,46 +113,107 @@ const InstanceSnapshots: FC<Props> = ({ controlTarget, instance }) => {
...
+          onSuccess={onSuccess}
         />
       )}
-      <Row>
-        {instance.snapshots && instance.snapshots.length > 0 ? (
+      {hasSnapshots && (
+        <div className="upper-controls-bar">
+          <div className="search-box-wrapper">
+            <SearchBox
+              className="search-box margin-right"
```
edlerd, line 155: This is missing an `id` field. If it is there, a hidden label will be linked to the input. Otherwise, we get a hidden label, that has no connected form element.

src/pages/instances/actions/snapshots/ActionButtons.tsx
edlerd, line 1: Maybe a better name like `SnapshotActions` or similar?

src/pages/instances/actions/snapshots/ActionButtons.tsx
```diff
@@ -0,0 +1,90 @@
...
+  onFailure,
+}) => {
+  const [isLoading, setLoading] = useState(false);
+  const queryClient = useQueryClient();
+
+  const handleDelete = () => {
+    setLoading(true);
+    deleteSnapshot(instance, snapshot)
+      .then(() => {
+        setLoading(false);
```
edlerd, line 30: I started to put the `setLoading` and `invalidateQueries` into a [finally block](https://github.com/canonical/lxd-ui/pull/231/files#diff-e7ef45c9e95848b69d5608e22eaae94b539591811aa5b705ded87d29291f9a92R41). I think it makes the code a bit more structured. And we might want to refresh state even in case an error happened.

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
```diff
@@ -47,6 +69,14 @@ const CreateSnapshotForm: FC<Props> = ({ instance, close }) => {
   const SnapshotSchema = Yup.object().shape({
     name: Yup.string()
       .required("This field is required")
+      .test("deduplicate", "Snap name is already in use", (value) =>
```
edlerd, line 72: "Snapshot name" or "Name"? why "Snap name" ...

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
```diff
@@ -126,29 +161,54 @@ const CreateSnapshotForm: FC<Props> = ({ instance, close }) => {
...
-          label="Stateful"
-          wrapperClassName="row"
-          help={getStatefulHelp()}
-          disabled={!isStateful || !isRunning}
-          onChange={formik.handleChange}
-          onBlur={formik.handleBlur}
+        <Row className="expiration-wrapper">
+          <Col size={6}>
+            <Input
+              id="expirationDate"
```
edlerd, line 169: I wonder if we can make more clear the dates are optional. In our user tests, no one realized that, and everyone started to enter a date. Might be a future task for design, though.

edlerd (commented):
Code looks good, just some small things and a couple of future ideas.

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
lorumic: This is the label we have in the HFM currently, should it be changed @al-bakalova?

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
lorumic: Sure, feel free to add a note to discuss it in the next UX-UI sync with @al-bakalova

src/pages/instances/InstanceSnapshots.tsx
lorumic: [Done](https://warthogs.atlassian.net/browse/WD-2713)

src/pages/instances/InstanceSnapshots.tsx
edlerd: I was wrong. What I wrote is true for input, select and other form elements. For searchbox we must have a name, but not an id. The component wants to id itself as search, so it shouldn't be set. Something we could fix upstream at one point, so the provided I'd won't break it.

src/pages/instances/InstanceSnapshots.tsx
lorumic: Sure, good idea! I've moved the function to a shared util.

src/pages/instances/InstanceSnapshots.tsx
lorumic: Added a `name` to the `SearchBox`.

src/pages/instances/actions/snapshots/ActionButtons.tsx
lorumic: Renamed to `SnapshotActions`.

src/pages/instances/actions/snapshots/ActionButtons.tsx
lorumic: Very good idea! I applied it to both action handlers.

edlerd (approved):
LGTM, just the question about the snap/snapshot thing in deduplicate. And a follow-up thing about making it obvious, the expiration in snapshot modal is optional. But the later one shouldn't block this.

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
lorumic: Ok, done.

PR #231, opened 2023-03-14
Side panel spacing and instance action overhaul WD-2592

edlerd:
* adjusted spacing in the instance side panel
* linking heading in side panel, remove custom links and styles
* add generic `InstanceStateAction` component with new start/restart/pause/stop icons on the side panel, instance list and instance detail header


edlerd:
@al-bakalova all issues should be resolved, please check again.

al-bakalova:
Everything looks great and pixel perfect! Thank you!
I have one final adjustment to ask for -
Please remove the "Force pause" action from the Confirm pause modal. We can only either force stop or force reboot.

lorumic:
Alignment of the spinner icon seems a bit wrong:

image

image

src/pages/instances/actions/OpenConsoleBtn.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <Button
+      appearance="base"
+      dense
+      hasIcon
+      onClick={handleOpen}
+      title="Console"
+    >
+      <i className="p-icon--canvas">Open VGA Terminal</i>
```
lorumic, line 33: We should probably use react-component's `Icon`, here but also in all the other 22 places where we still use the `i` tag. It's out of the scope of this PR: I'll file a JIRA ticket about this.

src/pages/instances/InstanceList.tsx
```diff
@@ -263,38 +243,24 @@ const InstanceList: FC = () => {
...
-                    })}
-                    instance={instance}
-                    hasCaption={true}
-                    onStarting={addStarting}
-                    onStopping={addStopping}
-                    onFinish={removeLoading}
-                  />,
-                ]}
+              <InstanceStateActions
+                key="startstop"
```
lorumic, line 255: We probably don't need this `key` anymore?

src/pages/instances/actions/StopInstanceBtn.tsx
```diff
@@ -0,0 +1,78 @@
...
+      confirmationExtra={
+        <span className="u-float-left">
+          <CheckboxInput
+            inline
+            label="Force stop"
+            tabIndex={-1}
+            defaultChecked={isForce}
+            onClick={() => setForce((prev) => !prev)}
+          />
+        </span>
```
lorumic, line 67: This is used in 2 places, maybe it could be good to extract it to a dedicated component? So that we only need to pass 3 props: the `label`, the `defaultChecked` and the click handler - possibly 2 props only if we pass the whole `force` state...

src/pages/instances/actions/RestartInstanceBtn.tsx
```diff
@@ -0,0 +1,81 @@
...
+      .finally(() => {
+        instanceLoading.setFinish(instance);
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.instances],
+        });
+      });
+  };
+
+  const disabledStatuses = ["Stopped", "Frozen"];
+  const isDisabled = isLoading || disabledStatuses.includes(instance.status);
```
lorumic, line 52: This would allow the user to restart an instance that has the "error" status. Is this actually admissible?

lorumic (changes requested):

src/pages/instances/actions/RestartInstanceBtn.tsx
edlerd: The reason is that sometimes a normal stop gets stuck. And I wanted to keep a way to then trigger a force stop.

lorumic: I see, should we then add the "Error" status to the `disabledStatuses` maybe? Just to be sure that we don't allow something that should not be allowed...

src/index.tsx
```diff
@@ -17,15 +18,17 @@ const root = createRoot(rootElement);
...
-      <QueryClientProvider client={queryClient}>
-        <AuthProvider>
-          <div className="l-application" role="presentation">
-            <Navigation />
-            <App />
-            <Panels />
-          </div>
-        </AuthProvider>
-      </QueryClientProvider>
+      <InstanceLoadingProvider>
```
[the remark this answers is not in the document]
edlerd: I don't have an opinion on the hierarchy, reordering seems fine though.

src/sass/styles.scss
```diff
@@ -39,6 +39,8 @@
 @include vf-p-icon-video-play;
 @include vf-p-icon-warning-grey;
 
+$border-thin: 1px solid $color-mid-light !default;
```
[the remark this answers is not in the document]
edlerd: That was also my first try, but we don't have access to the `$color-mid-light` variable in settings, it only exists after we import vanilla. The settings contain variables for vanilla itself.

src/pages/instances/actions/OpenConsoleBtn.tsx
edlerd: I can fix it here and keep it in mind for the future.

src/pages/instances/actions/RestartInstanceBtn.tsx
edlerd: I misunderstood your first comment. Yes - that makes sense.

edlerd:
@lorumic please give this a 2nd pass, all issues should be resolved. The action buttons misalignments will be addressed later as soon as we get the final images for them -- the current ones are just a wip to unblock development.

src/pages/instances/actions/OpenConsoleBtn.tsx
lorumic: Sorry, again on this point - should the "Open VGA Terminal" be set as `aria-label` of the `Button` here? And have the `<Icon />` self-closing then?

edlerd: Yes, this is the way.

lorumic (approved):

PR #233, opened 2023-03-14
Image field should be optional in edit details form

lorumic:


src/pages/instances/forms/InstanceEditDetailsForm.tsx
```diff
@@ -7,7 +7,7 @@ import { EditInstanceFormValues } from "pages/instances/EditInstanceForm";
 export interface InstanceEditDetailsFormValues {
   name: string;
   description?: string;
-  image: string;
+  image?: string;
```
edlerd, line 10: This won't fix the current error, as we reference the original image config and not the form image.

I propose to remove image and instanceType completely as both are immutable, see also https://github.com/canonical/lxd-ui/pull/234

lorumic: Good point, I approved #234. I'll close this one then.

PR #236, opened 2023-03-15
Fix heading structure and form label connections

edlerd:
* fixed heading structure on all pages (except instance detail)
* fixed all form elements to have corresponding labels


src/pages/instances/forms/MemoryLimitSelector.tsx
```diff
@@ -160,6 +162,7 @@ const MemoryLimitSelector: FC<Props> = ({ memoryLimit, setMemoryLimit }) => {
           </Col>
           <Col size={4}>
             <Select
+              id="memUnitSelect"
               name="unitSelect"
```
lorumic, line 166: Should we have consistent `id` and `name`?

src/pages/instances/forms/NetworkForm.tsx
```diff
@@ -75,6 +76,7 @@ const NetworkForm: FC<Props> = ({ formik, project, children }) => {
                   <Input
                     label="Network name"
                     name={`devices.${index}.name`}
+                    id={`networkName-${index}`}
```
lorumic, line 79: With or without the `-` before the index?
(I'd say without since it's camelCased, anyway it's better to stick with one convention and keep it everywhere)

lorumic (changes requested):
Should `SearchBox`es have `id` set? Or `name`? Or neither of them? You said in [this comment](https://github.com/canonical/lxd-ui/pull/228#discussion_r1140671898) that they should have the `id` set, but what I see in this PR is:

- `name` set, `id` not set (in `ImageSelector`)
- `id` set, `name` not set (in `Settings`)
- neither set (in `InstanceList`)

lorumic:
QA: this is how the page of the instance list looks like when I open the [demo](https://lxd-ui-236.demos.haus/ui/default/instances) of this PR:

Peek 2023-03-17 22-18

src/components/TableColumnsSelect.tsx
```diff
@@ -34,7 +34,13 @@ const TableColumnsSelect = ({
       dropdownProps={{ "aria-label": "columns menu" }}
       position="right"
       toggleClassName="has-icon"
-      toggleLabel={<Icon name="settings" />}
+      toggleLabel={
+        <>
+          <Icon name="settings">
+            <span className="u-off-screen">Columns selection toggle</span>
```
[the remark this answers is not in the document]
edlerd: True, we can elevate this as an aria-label of the button.

edlerd:
All issues should be resolved @lorumic 
* The SearchBox should have a name, but not an id.
* The bug in the instance list is resolved upstream through the new pagination component. That one needed some more adjustments to be a11y conform.

PR #242, opened 2023-03-20
replace yaml editor to Monaco editor

xlmnxp:


edlerd:
Thank you for the PR, Monaco Editor is way better than what we currently use.

I think we should set `options={{ fontSize: 18 }}` to make it more consistent with the rest of the app.

Seeing the fixed height of 32rem makes me think if we should switch it to something more responsive. Ideally, the editor takes the full available height and avoids additional scrolling. Also, the width would be good to be full space/responsive.

Lastly, we should probably also use the editor in the `CloudInitConfig` component. So we can drop the `@focus-reactive/react-yaml` dependency in `package.json` and `yarn.lock`.

xlmnxp:
> Thank you for the PR, Monaco Editor is way better than what we currently use.
> 
> I think we should set `options={{ fontSize: 18 }}` to make it more consistent with the rest of the app.
> 
> Seeing the fixed height of 32rem makes me think if we should switch it to something more responsive. Ideally, the editor takes the full available height and avoids additional scrolling. Also, the width would be good to be full space/responsive.
> 
> Lastly, we should probably also use the editor in the `CloudInitConfig` component. So we can drop the `@focus-reactive/react-yaml` dependency in `package.json` and `yarn.lock`.

there issue with Monaco editor height when changed to 100% (default value) which make the editor expand for ever ( I tried to solve it but without luck) so I provide height property to change it when needed ( also i think it responsive because it based on rem)

xlmnxp:
@edlerd fixed and implement auto resize for container, also I modify CloudInitConfig to use same editor

edlerd:
Thank you, this is great!

I'll merge it right away and change a little bit the height adjustment and adjust it to our formatting -- you can use `dotrun format-js` to make it autoformat.

PR #243, opened 2023-03-21
Adjusted design of the terminal tab [WD-2553]

lorumic:
- Moved "Reconnect" button inside the tab
- Moved notifications inside the tab
- Made header, notification rows and control row sticky - only terminal is to be scrollable
- Adjusted terminal's background colour

### DIFF
--- a/src/pages/instances/InstanceDetail.tsx
+++ b/src/pages/instances/InstanceDetail.tsx
@@ -132,7 +132,7 @@ const InstanceDetail: FC = () => {
 
               {activeTab === "terminal" && (
                 <div role="tabpanel" aria-labelledby="terminal">
-                  <InstanceTerminal controlTarget={controlTarget} />
+                  <InstanceTerminal />
                 </div>
               )}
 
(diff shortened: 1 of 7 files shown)


edlerd (commented):
LGTM, there is one visual bug with the scrollbar of the terminal, to reproduce:
* make the browser window small
* hit enter ~40 times
* maximize browser window
It sometimes happens, but not always. I assume this is some race condition in play with with resize / re-render.

al-bakalova:
Good job, @lorumic! Looks good to me.

lorumic:
> LGTM, there is one visual bug with the scrollbar of the terminal, to reproduce:
> 
> * make the browser window small
> * hint enter ~40 times
> * maximize browser window
>   It sometimes happens, but not always. I assume this is some race condition in play with with resize / re-render.

What's happening exactly? I cannot see anything strange on the scrollbar trying the steps that you listed. Also, on which browser(s) did you experience this?

edlerd:
> What's happening exactly? I cannot see anything strange on the scrollbar trying the steps that you listed. Also, on which browser(s) did you experience this?

It gets a little distorted, I can post a picture tomorrow or share it with you. I could reproduce it when making the window quite small and then maximizing it. Interestingly, when changing the size slowly it didn't happen, and it also fixed itself when resizing slowly..

lorumic:
> It gets a little distorted, I can post a picture tomorrow or share it with you. I could reproduce it when making the window quite small and then maximizing it. Interestingly, when changing the size slowly it didn't happen, and it also fixed itself when resizing slowly.

Ok, anyway it doesn't sound like something we should really worry about, given that it's so hard to reproduce (and notice), or does it?

lorumic:
@edlerd Added the `setTimeout` fix as discussed, you can proceed with the review when you have time.

edlerd (approved):

PR #244, opened 2023-03-21
Set dark theme for YAML editor [WD-2549]

lorumic:
Thanks to @xlmnxp we now have a much better YAML Editor, which also has a nice dark theme. Could you please have a look at it @al-bakalova?

image

### DIFF
--- a/src/pages/instances/forms/YamlForm.tsx
+++ b/src/pages/instances/forms/YamlForm.tsx
@@ -48,6 +48,7 @@ const YamlForm: FC<Props> = ({
         <Editor
           defaultValue={yaml}
           language="yaml"
+          theme="vs-dark"
           onChange={(value) => value && setYaml(value)}
           options={{
             fontSize: 18,


al-bakalova:
@lorumic Every time I click on the demo link, I get a "Building" message. 
Could this page be stuck in the pipeline?

lorumic:
> @lorumic Every time I click on the demo link, I get a "Building" message. Could this page be stuck in the pipeline?

There has been a general outage of the demos. I just triggered a manual rebuild and now the demo is live again.

edlerd (approved):

al-bakalova:
@lorumic This looks good to me!
Before we decide to adopt it, I would only like to make sure we can set a max height for the yaml editor. What I have in mind is the cloud init configuration section, as well as the yaml config one.

lorumic:
> @lorumic This looks good to me! Before we decide to adopt it, I would only like to make sure we can set a max height for the yaml editor. What I have in mind is the cloud init configuration section, as well as the yaml config one. Of course, these modifications will be part of a different PR. I just want us to explore the possibilities in advance.

Yes, we can set a max height for it. Here's a live demo:

Current situation (no max height):

no-max-height

Example with a max height of 200px:

with-max-height

edlerd:
My two cents regarding the height of the editor:
* In the YAML Configuration tab the editor should take the full height and start scrolling at one point, the page itself will never scroll, because the editor will not grow more than the page height
* in the cloud init, we have 3 editors, so we can't let them take the full height. also we don't want the editors to scroll as well as the outer window to scroll, hence we grow them with the content.

I think this is the best we can get for both cases.

al-bakalova:
@lorumic Thanks for checking and providing examples! Looks like the editor meets our needs!

@edlerd I agree with your points. Should we create a separate task to implement these?

edlerd:
> @edlerd I agree with your points. Should we create a separate task to implement these?

That is how it works atm :)

PR #247, opened 2023-03-23
Overhaul Instance detail tab: Console [WD-2554]

lorumic:


al-bakalova:
The tabs look very good. A few improvements to still apply:
- Both the header and the tabs should always be sticky 
- When the instance is stopped, please do not treat it as an error. Let's introduce an empty state screen and invite the user to start the instance. Here's an example:
Screenshot 2023-03-24 at 9 54 30

- Graphic console tab: please do not introduce a scrollable page. The graphic console itself should fit within the existing page. The user can always click on Fullscreen to see more. If you need to deal with specific ratio, please always align left the content.
- Text console - please make sure the space around the text console is always 1,5 rems. Currently the top and bottom space looks funny.

al-bakalova:
Looks good to me! Thank you!

src/pages/instances/InstanceVga.tsx
```diff
@@ -17,36 +17,43 @@ declare global {
...
 interface Props {
-  controlTarget?: HTMLSpanElement | null;
+  onMount: (handler: () => void) => void;
+  onFailure: (message: string, e: unknown) => void;
+  inTabNotification: Notification | null;
 }
 
-const InstanceVga: FC<Props> = ({ controlTarget }) => {
-  const notify = useNotify();
+const InstanceVga: FC<Props> = ({ onMount, onFailure, inTabNotification }) => {
```
edlerd, line 25: Now that we call it "Graphic" in the frontend under the console tab, shall we rename this to `InstanceGraphicConsole`?

edlerd (approved):
LGTM, just one idea to rename the vga maybe?

PR #248, opened 2023-03-23
Overhaul Instance detail tab: Overview [WD-2551]

lorumic:


src/sass/_instance_detail_overview.scss
```diff
@@ -0,0 +1,86 @@
...
+    td {
+      &:nth-child(2) {
+        width: 80px;
+      }
+
+      &:nth-child(3) {
+        width: 94px;
+
+        @include extra-large {
+          width: 105px;
```
edlerd, line 62: Better use `rem` than px as a unit.
The type and managed columns are too narrow, when sorting by them the heading gets truncated. A bit more space would be good.

src/pages/instances/InstanceOverview.tsx
```diff
@@ -31,162 +38,282 @@ const InstanceOverview: FC<Props> = ({ controlTarget, instance }) => {
...
+  if (profileError) {
+    setInTabNotification(failure("Could not load profiles.", profileError));
   }
 
+  const profileHeaders = [
+    { content: "Name", sortKey: "name" },
+    { content: "Description", sortKey: "description" },
+  ];
+
+  const profileRows = profiles
```
edlerd, line 139: The default display should be in order from the `instance.profiles, not in order of the profile list response.

Maybe extract the profile table with the extra query into a dedicated component, same for the networks table.

edlerd: I forgot to mention: the network table doesn't list the up addresses, at least on the demo server the cells were always empty.

src/sass/_instance_detail_overview.scss
lorumic: We've always been using `px` for column widths - see [these](https://github.com/canonical/lxd-ui/blame/f1dbf916f78f56640231d943527d17de9cf23ff8/src/sass/_instance_list.scss#L107) that you defined for the instance table. For consistency, it doesn't make much sense to stop now all of a sudden. Rather, we could have a separate ticket to convert all column widths from `px` to `rem` in the future, if you like.

src/pages/instances/InstanceOverview.tsx
lorumic: > I forgot to mention: the network table doesn't list the IP addresses, at least on the demo server the cells were always empty.

It does, it's just that almost all the instances on the demo server have been assigned a disconnected network device. Instances with `lxdbr0` have the IP addresses populated: [one example](https://lxd-ui-248.demos.haus/ui/default/instances/detail/testing123).

al-bakalova:
This tab also looks very good. A few small modifications to make it perfect:
- As we discussed with the core team, maybe the Overview tab is the place to introduce the PID - this is why it was included in the mockups
- Please use the column title look and feel for both the keys (base image, description, type, architecture, PID, date created, last used, memory, disk) and the table columns (name, type, managed, ipv4, ipv6, name, description). These should all match. You can reuse the the look and feel from the instance list, if needed. Currently, these look too big.
- The section titles (General, Usage, Networks, Profiles) also seem to be too big. These should be H5.
- And one final request, could you please align the first column title with the rest of the labels. Ex. "Name" in the table should be left aligned with disk, memory, last used, ...

src/pages/instances/InstanceOverview.tsx
edlerd: The IP's are sometimes missing. I.e. This [overview](https://lxd-ui-248.demos.haus/ui/default/instances/detail/alex-is-testing) has no IP on the fan network listed. If we check the `instance.state.networks.eth0` (which responds to the `lxdfan0` network name of `eth0`, then we see there is an ipv4 and also an ipv6 listed.
So the logic for picking the addresses to show in the overview tab should rely on the `instance.state.networks`. Though not sure if this is exhaustive, or if further addresses could be under the `instance.expanded_devices`, might be worth to ask it to the core team.

lorumic: I don't understand. The IPs that are being shown here are supposed to match the ones shown in the [Networks](https://lxd-ui-248.demos.haus/ui/default/networks) page, right? This at least is what I can reckon by looking at the HFM. So it seems just right as it is to me.

edlerd: The network page is also broken in that regard. We should focus to get the instance overview tab fixed in this PR.

src/sass/_instance_detail_page.scss
```diff
@@ -1,4 +1,8 @@
 .instance-detail-page {
+  #control-target {
+    margin-right: $sph--large;
```
edlerd, line 3: We don't need this, the header has its own spacing.

src/pages/instances/InstanceOverviewMetrics.tsx
edlerd, line 1: It is much nicer with the split components. With it, the loading behaviour improved as well.

al-bakalova:

src/pages/instances/InstanceOverview.tsx
```diff
@@ -1,192 +1,168 @@
...
+                    instanceCreationTypes.filter(
+                      (item) => item.value === instance.type
+                    )[0].label
+                  }
+                </td>
+              </tr>
+              <tr>
+                <th className="p-muted-heading">IPv4</th>
+                <td>
+                  {ip4Addresses.length ? (
```
edlerd, line 76: This can be a reused component to limit / show all items from an array. We pass it the items and the initial limit. It contains the show all button and holds the state of how many items are currently visible.
We can reuse it here for ipv4 and ipv6, and also use it for both in the detail panel.

We should probably also use it in the Instance list columns for ipv4 and v6 -- but maybe not in scope of this PR.

lorumic: Nice idea! I just implemented that.

src/pages/instances/ExpandableList.tsx
edlerd, line 1: This could live under src/components, but probably we can move it there once we have a 2nd use outside of instances.

edlerd (approved):

PR #249, opened 2023-03-24
Set a min- and max-height for cloud-init config fields

lorumic:


lorumic:
Now that #240 is merged, this can also be reviewed. @al-bakalova @edlerd

edlerd:
The read mode is missing the height constraints. It comes in with the [code renderer](https://github.com/canonical/lxd-ui/blob/main/src/pages/instances/forms/CloudInitForm.tsx#L31)

lorumic:
> The read mode is missing the height constraints. It comes in with the [code renderer](https://github.com/canonical/lxd-ui/blob/main/src/pages/instances/forms/CloudInitForm.tsx#L31). An example on the demo server is [here](https://lxd-ui-249.demos.haus/ui/default/profiles/detail/new-profile-name/configuration).

Right - well spotted. Should be good now.

edlerd (approved):
Thanks for changes, lgtm

al-bakalova:
Looks good to me. I'd like to clarify one point though:
- Why do we get line numbers only for the first three lines? It looks kind of broken to me. Is this expected?

Screenshot 2023-03-30 at 10 40 39

al-bakalova:
@lorumic It all makes sense! Thanks for clarifying this, Michele!

PR #252, opened 2023-03-27
Update logo and favicon, decrease top padding on all pages WD-2762

edlerd:
## Done

- update logo
- update favicon
- reduce top spacing on all pages, to make heading in line with the new logo

Fixes WD-2762

## QA


al-bakalova:
Looks good to me, thank you!

src/sass/_pattern_navigation.scss
```diff
@@ -1,35 +1,34 @@
...
 
   @include extra-large {
     .p-panel__logo-image {
-      height: 54px;
+      height: 40px;
     }
 
     .logo-text {
-      left: 61px;
+      left: 52px;
```
lorumic, line 31: I see the logo text "jumping up" when passing the `x-large` breakpoint:

Peek 2023-03-28 12-23

Maybe it needs some `top` adjustment too? I think it's ugly to see it jump, even though the title of the instance list does exactly the same...

edlerd: Good catch! the top spacing of the logo text should be `.25rem` instead of `5px`.

lorumic: I don't see any change in the `x-large` view, it's still jumping just as shown in the screen capture. But maybe this is expected? I see that it happens with the title as well, it's just that it looks super-ugly to me...

edlerd: Before it was not in line with the other headings, now it is. As for the jumping, I think that is intended for larger screens, if understand your concern correctly, even the [vanilla example](https://vanillaframework.io/docs/examples/layouts/application/JAAS) has the same behaviour.

lorumic (approved):

PR #253, opened 2023-03-27
Add renaming of profiles and instances WD-2780

edlerd:
## Done

- allow renaming of profiles and instances via the edit forms
- restrict the default profile to not be renamed.

Fixes WD-2780

## QA

- try renaming an instance/profile
- try updating an instance/profile at the same time as renaming it

### DIFF
--- a/src/components/Pagination.tsx
+++ b/src/components/Pagination.tsx
@@ -6,7 +6,7 @@ interface Props {
   pageSize: number;
   setPageSize: (val: number) => void;
   currentPage: number;
-  setCurrentPage: (val: number) => void;
+  paginate: (val: number) => void;
   visibleCount: number;
   totalCount: number;
   totalPages: number;
@@ -17,7 +17,7 @@ const Pagination: FC<Props> = ({
   pageSize,
   setPageSize,
   currentPage,
-  setCurrentPage,
+  paginate,
   visibleCount,
   totalCount,
   totalPages,
@@ -37,7 +37,7 @@ const Pagination: FC<Props> = ({
         disabled={currentPage === 1}
         onClick={() => {
           if (currentPage > 1) {
-            setCurrentPage(currentPage - 1);
+            paginate(currentPage - 1);
           }
         }}
       >
@@ -54,7 +54,7 @@ const Pagination: FC<Props> = ({
             Math.max(1, parseInt(e.target.value))
           );
 
-          setCurrentPage(newPage);
+          paginate(newPage);
         }}
         value={currentPage}
         type="number"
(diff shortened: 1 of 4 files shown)


al-bakalova:
This is a great start. A few improvements to make it perfect:

- Once an item's name has been successfully updated, please add the confirmation notification below the tab.

- For both flows, please review the copy of the confirmation messages and change it to "**Success**. Configuration updated"

al-bakalova:
Updating only the instance name and clicking on save - the page freezes with an error but the name still gets updated - 
Screenshot 2023-03-28 at 8 59 35

Updating the name and the storage - we get a very long message with two different colons - 
Screenshot 2023-03-28 at 9 20 12

Maybe let's try to shorten these to 
**Error**. Configuration update for "device-name" failed. Instance DNS name already in use.

Screenshot 2023-03-28 at 9 23 48


When updating more than one configurations that usually have errors, you only get an error for the latter.

edlerd:
Fixed the errors and the notifications. @al-bakalova please have another look at this one if it is ready to go forward.

edlerd:
> @edlerd Looks like the page is stuck in the pipeline. Could you please redeploy?

@al-bakalova it works for me. Might be that @lorumic fixed it in the meantime.

lorumic:
> @al-bakalova it works for me. Might be that @lorumic fixed it in the meantime.

Yeah, I saw the message and triggered a rebuild, it has just finished.

al-bakalova:
When performing an action within a tab, the notification should appear below the tabs. This is a general observation and a rule we should always follow.

- Seems to me that renaming instances is no longer possible. I got the same message trying to rename a few different instances. A couple of screenshots to share -

- Renaming the instance + updating the description + updating a network, results in an error. The description was successfully updated though:
Screenshot 2023-03-30 at 10 31 55

- Another attempt to update an instance name:
Screenshot 2023-03-30 at 10 33 55

- Updating the profile names works every time. The notification should appear below the tabs:
Screenshot 2023-03-30 at 10 35 48

edlerd:
How shall we proceed here, should we find a copy for the case when an instance is updated, but the rename failed, or do we separate the rename action and add the pencil to the header for renaming?
Personally I think the pencil in the header might be a bit too visible, but maybe there is a way to make that not too distractive?

al-bakalova:
@edlerd 
For the time being, I think it is best to separate these two actions. Let me spend some time on the header and I will come back with suggestions.

al-bakalova:
Design ticket for the issue https://warthogs.atlassian.net/browse/WD-2986

PR #256, opened 2023-03-28
Redesign the Reconnect terminal modal [WD-2772]

lorumic:
## Done

- Redesigned the "Reconnect terminal" modal.

Fixes [WD-2772](https://warthogs.atlassian.net/browse/WD-2772)

## QA

1. Browse an instance detail page, [terminal tab](https://lxd-ui-256.demos.haus/ui/default/instances/detail/alex-is-testing/terminal), and click on the "Reconnect" button.
2. Check the new design of the modal.

[WD-2772]: https://warthogs.atlassian.net/browse/WD-2772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/TerminalPayloadForm.tsx
+++ b/src/pages/instances/TerminalPayloadForm.tsx
@@ -1,17 +1,10 @@
-import React, { FC, KeyboardEvent } from "react";
-import {
-  Button,
-  Col,
-  Form,
-  Icon,
-  Input,
-  Label,
-  Modal,
-  Row,
-} from "@canonical/react-components";
+import React, { FC, KeyboardEvent, useEffect, useRef } from "react";
+import { Button, Form, Icon, Input, Modal } from "@canonical/react-components";
 import * as Yup from "yup";
 import { useFormik } from "formik";
 import { LxdTerminalPayload } from "types/terminal";
+import { updateMaxHeight } from "util/updateMaxHeight";
+import useEventListener from "@use-it/event-listener";
 
 interface Props {
   close: () => void;
@@ -19,6 +12,8 @@ interface Props {
 }
 
 const TerminalPayloadForm: FC<Props> = ({ close, onFinish }) => {
+  const ref = useRef<HTMLDivElement>(null);
+
   const TerminalSchema = Yup.object().shape({
     command: Yup.string().required("This field is required"),
     environment: Yup.array().of(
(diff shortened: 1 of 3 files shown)


src/pages/instances/TerminalPayloadForm.tsx
```diff
@@ -100,89 +112,72 @@ const TerminalPayloadForm: FC<Props> = ({ close, onFinish }) => {
...
-          <Col size={8}>
-            <Button onClick={addEnvironmentRow} type="button" hasIcon>
-              <Icon name="plus" />
-              <span>Add variable</span>
+        <Input type="submit" hidden />
+        <Label className="u-no-margin--bottom">Environment variables</Label>
+        {formik.values.environment.map((_variable, index) => (
+          <div key={index} className="env-variables">
+            <Input
+              type="text"
```
edlerd, line 149: This and the other input is missing a unique id. You can use the index to generate one.
We should also add a `label`, and the `labelClassName="u-off-screen"`.

src/pages/instances/TerminalPayloadForm.tsx
```diff
@@ -100,89 +112,72 @@ const TerminalPayloadForm: FC<Props> = ({ close, onFinish }) => {
...
-          </Row>
-        ))}
-        <Row>
-          <Col size={4} />
-          <Col size={8}>
-            <Button onClick={addEnvironmentRow} type="button" hasIcon>
-              <Icon name="plus" />
-              <span>Add variable</span>
+        <Input type="submit" hidden />
+        <Label className="u-no-margin--bottom">Environment variables</Label>
```
edlerd, line 145: This shouldn't be a label, because it doesn't refer to any form element.

edlerd (approved):

al-bakalova:
When all the environment variables are removed, the modal resizes. Could we please prevent it from doing so?
Screenshot 2023-03-28 at 14 59 33

The second key/value pair should only be visible upon request. 
Screenshot 2023-03-28 at 15 00 53

When too many key/value pairs are added, the modal becomes too long and the footer even hides.
Screenshot 2023-03-28 at 15 01 53

Can you please add a content scroll? Please make sure to make the actions (Cancel, Reconnect) sticky and always visible.

al-bakalova:
Looks very good to me! 
One final suggestion -  could you please make sure to always scroll down the content so that the "Add variable" button is visible even after adding a couple of variable key-value pairs?

lorumic:
> Looks very good to me! One final suggestion - could you please make sure to always scroll down the content so that the "Add variable" button is visible even after adding a couple of variable key-value pairs?

Done - now the content should smoothly scroll so that the "Add variable" button stays visible after a new pair is added.

al-bakalova:
@lorumic Looks good to me! Thanks for applying all the changes!

lorumic:
@edlerd Quite a few things have changed in terms of code, so it would be probably better to redo the review before merging?

src/pages/instances/TerminalPayloadForm.tsx
```diff
@@ -79,110 +74,128 @@ const TerminalPayloadForm: FC<Props> = ({ close, onFinish }) => {
     }
   };
 
+  const updateContentHeight = () => {
+    updateMaxHeight("content-wrapper", "p-modal__footer", 64, "max-height");
```
edlerd, line 78: Where does the 64 come from?
Is it for the modal padding? The modal has a padding of `1rem` top and bottom, so shouldn't we use `2rem` or read the modal padding interactively?

edlerd (approved):
LGTM, just curious about the magic 64

src/pages/instances/TerminalPayloadForm.tsx
lorumic: Yes, currently this "magic" value was obtained empirically from manual tests at different resolutions, trying to find a `px` amount that allowed to keep the footer always visible without taking too much vertical space around the modal.
But the idea for a future improvement would be to make the `updateMaxHeight` function to receive an `additionalOffset` parameter that is not necessarily in the `px` unit, but rather a full "value+unit" string that you then pass to the `calc` function:
```
  const style = `${targetProperty}: calc(100vh - ${offset}px - ${additionalOffset})`;
```
So if you pass "2rem", and you calculate an "above+below" offset of 20px, you'd have: `calc(100vh - 20px - 2rem)`.

PR #258, opened 2023-03-29
Limit number of ips displayed in the instance list to 5

edlerd:
## Done

- limit the number of ip addresses in the instance list to 5 with a show more link in case there are more

## QA

1. Run the LXD-UI:
    - check the instance list with an instance that has multiple ip addresses


edlerd:
Here are some screenshot showing the different stages (with & without side panel
Screenshot from 2023-03-29 15-27-58
Screenshot from 2023-03-29 15-27-53
Screenshot from 2023-03-29 15-27-48
Screenshot from 2023-03-29 15-27-37

al-bakalova:
Thanks for providing the screenshots.
I believe, when having more than one IP address to display in the table, our initial idea was to show a count ([see mockup](https://app.zeplin.io/project/63e128139e0eff100dad0cbd/screen/640f5bda4a254e220bc5a264) - second line) and then display the exhaustive list in the side panel ([see mockup](https://app.zeplin.io/project/63e128139e0eff100dad0cbd/screen/63e202305e0ea60e95e98327)). 

This is what the side panel is for. Let's try not to overload the table.

edlerd:
Right, that makes sense, I updated the PR, now it looks like in the screen below (same for the ipv4)
Screenshot from 2023-03-29 22-05-21

al-bakalova:
@edlerd Looks good to me! Can I get a quick screenshot of how it looks like in the side panel?
I do realise this is out of scope, we can add it as a separate task.

edlerd:
@al-bakalova in the side panel we'd have the full list, but only up to 5 with a "view all" link that shows all.

Screenshot from 2023-03-30 13-39-24

al-bakalova:
@edlerd Yes! This is the expected behaviour! Thanks for sharing. I think this is good to go. Danke!

lorumic (approved):

PR #260, opened 2023-03-29
Overhaul instance detail header [WD-2550]

lorumic:
## Done

- Adjusted the header of the instance detail page.

Fixes [WD-2550](https://warthogs.atlassian.net/browse/WD-2550)

## QA

1. Run the LXD-UI:
    - Visit any instance detail page ([example](https://lxd-ui-260.demos.haus/ui/default/instances/detail/alex-is-testing)), and check the appearance of the header.

[WD-2550]: https://warthogs.atlassian.net/browse/WD-2550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/InstanceDetail.tsx
+++ b/src/pages/instances/InstanceDetail.tsx
@@ -1,8 +1,8 @@
 import React, { FC } from "react";
-import { List, Row, Tabs } from "@canonical/react-components";
+import { Row, Tabs } from "@canonical/react-components";
 import InstanceOverview from "./InstanceOverview";
 import InstanceTerminal from "./InstanceTerminal";
-import { useNavigate, useParams } from "react-router-dom";
+import { Link, useNavigate, useParams } from "react-router-dom";
 import InstanceSnapshots from "./InstanceSnapshots";
 import NotificationRow from "components/NotificationRow";
 import { useNotify } from "context/notify";
@@ -71,18 +71,31 @@ const InstanceDetail: FC = () => {
     <main className="l-main">
       <div className="p-panel instance-detail-page">
         <div className="p-panel__header">
+          <h1 className="u-off-screen">{name}</h1>
           {instance ? (
-            <List
-              className="p-panel__title"
-              inline
-              items={[
-                <span key="name">{name}</span>,
-                <i key="status" className="p-text--small">
-                  {instance.status}
-                </i>,
-                <InstanceStateActions key="state" instance={instance} />,
-              ]}
-            />
+            <div className="p-panel__title">
+              <nav
+                key="breadcrumbs"
+                className="p-breadcrumbs"
+                aria-label="Breadcrumbs"
+              >
+                <ol className="p-breadcrumbs__items">
+                  <li className="p-breadcrumbs__item">
+                    <Link to={`/ui/${project}/instances`}>Instances</Link>
+                  </li>
+                  <li
+                    className="p-breadcrumbs__item instance-name u-truncate"
+                    title={name}
+                  >
+                    {name}
+                  </li>
+                </ol>
+              </nav>
+              <div>
+                <i className="u-text--muted">({instance.status})</i>
+                <InstanceStateActions key="state" instance={instance} />
+              </div>
+            </div>
           ) : (
             <h4 className="p-panel__title">{name}</h4>
           )}
--- a/src/pages/instances/actions/DeleteInstanceBtn.tsx
+++ b/src/pages/instances/actions/DeleteInstanceBtn.tsx
@@ -2,7 +2,6 @@ import React, { FC, useState } from "react";
 import { deleteInstance } from "api/instances";
 import { LxdInstance } from "types/instance";
 import ConfirmationButton from "components/ConfirmationButton";
-import { Tooltip } from "@canonical/react-components";
 import { useNavigate } from "react-router-dom";
 import { useNotify } from "context/notify";
 
@@ -34,24 +33,18 @@ const DeleteInstanceBtn: FC<Props> = ({ instance }) => {
   const isDisabled = isLoading || instance.status !== "Stopped";
 
   return (
-    <Tooltip
-      message={isDisabled ? "Instance must be stopped" : undefined}
-      position="btm-center"
-    >
-      <ConfirmationButton
-        className="u-no-margin--bottom"
-        isLoading={isLoading}
-        iconClass="p-icon--delete"
-        iconDescription="Delete"
-        title="Confirm delete"
-        toggleCaption="Delete"
-        confirmationMessage={`Are you sure you want to delete instance "${instance.name}"? This action cannot be undone, and can result in data loss.`}
-        posButtonLabel="Delete"
-        onConfirm={handleDelete}
-        isDense={false}
-        isDisabled={isDisabled}
-      />
-    </Tooltip>
+    <ConfirmationButton
+      toggleAppearance="base"
+      className="delete-instance-btn"
+      isLoading={isLoading}
+      icon="delete"
+      title="Confirm delete"
+      confirmationMessage={`Are you sure you want to delete instance "${instance.name}"? This action cannot be undone, and can result in data loss.`}
+      posButtonLabel="Delete"
+      onConfirm={handleDelete}
+      isDense={true}
+      isDisabled={isDisabled}
+    />
   );
 };
 
--- a/src/pages/instances/actions/StartInstanceBtn.tsx
+++ b/src/pages/instances/actions/StartInstanceBtn.tsx
@@ -3,10 +3,11 @@ import { LxdInstance } from "types/instance";
 import { useQueryClient } from "@tanstack/react-query";
 import { startInstance, unfreezeInstance } from "api/instances";
 import { queryKeys } from "util/queryKeys";
-import { Button } from "@canonical/react-components";
+import { Button, Icon } from "@canonical/react-components";
 import { useNotify } from "context/notify";
 import { useInstanceLoading } from "context/instanceLoading";
 import InstanceLink from "pages/instances/InstanceLink";
+import classnames from "classnames";
 
 interface Props {
   instance: LxdInstance;
@@ -62,12 +63,9 @@ const StartInstanceBtn: FC<Props> = ({ instance }) => {
       aria-label={isLoading ? "Starting" : "Start"}
       title="Start"
     >
-      <i
-        className={
-          isLoading
-            ? "p-icon--spinner u-animation--spin"
-            : "p-icon--start-instance"
-        }
+      <Icon
+        className={classnames({ "u-animation--spin": isLoading })}
+        name={isLoading ? "spinner" : "play"}
       />
     </Button>
   );
--- a/src/sass/_instance_detail_page.scss
+++ b/src/sass/_instance_detail_page.scss
@@ -1,10 +1,62 @@
 .instance-detail-page {
+  .p-panel__title {
+    display: flex;
+    flex-direction: column;
+    max-width: none;
+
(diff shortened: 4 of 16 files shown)


src/sass/_pattern_icon.scss
```diff
@@ -9,32 +9,7 @@
...
-  background-image: url("https://assets.ubuntu.com/v1/24df595c-state=pause.svg");
-  background-position: center;
-  display: inline-block;
-  min-height: 1rem;
-  min-width: 1rem;
-}
-
-.p-icon--restart-instance {
-  background-image: url("https://assets.ubuntu.com/v1/d205cd5a-state=state4.svg");
+.instance-state-icon {
```
edlerd, line 12: I think this class can be removed, vanilla default styling looks correct.

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -34,24 +33,20 @@ const DeleteInstanceBtn: FC<Props> = ({ instance }) => {
...
   return (
-    <Tooltip
-      message={isDisabled ? "Instance must be stopped" : undefined}
-      position="btm-center"
-    >
+    <div className="delete-instance-btn-wrapper">
       <ConfirmationButton
-        className="u-no-margin--bottom"
+        toggleAppearance="base"
+        className="delete-instance-btn u-no-margin--bottom"
```
edlerd, line 39: We don't need the margin-bottom class here, instead I'd move the margins from the `delete-instance-btn-wrapper` to the `delete-instance-btn`. It should simplify the whole bit in here.

src/pages/instances/InstanceDetail.tsx
```diff
@@ -76,10 +76,23 @@ const InstanceDetail: FC = () => {
               className="p-panel__title"
               inline
               items={[
-                <span key="name">{name}</span>,
```
edlerd, line 79: I think we don't need this list as a wrapper. Removing it should also simplify the styling. I.e. the magic 0.4 in `p-breadcrumbs__item margin-bottom: 0.4rem;` should be a 0. Because the flex container with `align-items: center` should do the centring. The problem with the list is, that the items are hidden away under the list and don't get centred, when they differ in height. I hope you understand what I meant.

lorumic: I quickly tried to remove the list and it was causing only additional problems. If you have something specific in mind that you tested and it's working, feel free to submit a suggestion with the suggestion button:

image

edlerd: This is what I came up with. Quite a few styles still needed, but at least the list is gone:
```
            <div className="p-panel__title">
              <nav
                key="breadcrumbs"
                className="p-breadcrumbs"
                aria-label="Breadcrumbs"
              >
                <ol className="p-breadcrumbs__items">
                  <li className="p-breadcrumbs__item">
                    <Link to={`/ui/${project}/instances`}>Instances</Link>
                  </li>
                  <li className="p-breadcrumbs__item">
                    <span>
                      {name}{" "}
                      <i className="u-text--muted">({instance.status})</i>
                    </span>
                  </li>
                </ol>
              </nav>
              <InstanceStateActions key="state" instance={instance} />
            </div>
```
and the styles will become

```

  .p-panel__title {
    align-items: center;
    display: flex;
    padding-bottom: $spv--x-small;

    .p-breadcrumbs {
      margin-bottom: 0.2rem;
      width: inherit;

      .p-breadcrumbs__items {
        margin-bottom: 0;
      }

      .p-breadcrumbs__item {
        margin-top: 0;
      }

      .p-breadcrumbs__item:first-of-type {
        margin-right: $sp-x-small;
      }
    }

    .actions-list {
      margin-bottom: 0;
      margin-left: $sph--large;
    }
  }
```

src/pages/instances/InstanceDetail.tsx
```diff
@@ -76,10 +76,23 @@ const InstanceDetail: FC = () => {
...
+                  className="p-breadcrumbs"
+                  aria-label="Breadcrumbs"
+                >
+                  <ol className="p-breadcrumbs__items">
+                    <li className="p-breadcrumbs__item">
+                      <Link to={`/ui/${project}/instances`}>Instances</Link>
+                    </li>
+                    <li className="p-breadcrumbs__item">
+                      <span>
+                        {name}{" "}
```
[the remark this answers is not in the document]
lorumic: Yes, I agree, having an `h1` with the off-screen class is probably the best solution here.

src/pages/instances/InstanceDetail.tsx
lorumic: This is super helpful, thanks! I also came up with a similar solution eventually, but I was struggling with the alignment on smaller screens. This has a bit more styles than my solution, but it also works nicer at the different resolutions.

edlerd: Maybe the actions need a `min-width` in general, to avoid them from breaking in two lines. We'd have to check if that works well in the instance list and here in the header. Might be good for both cases.

lorumic: Ok, I added also the `min-width`, but they're still breaking if you zoom enough - I don't know if there's much we can do about it.

al-bakalova:
@lorumic, that's a great start. A few points to make it look better.

-  Could you please align the logo text with the breadcrumb and the icons?
Screenshot 2023-03-30 at 10 18 06

- For the breadcrumb, please use the same font size as the one we are using for the title on the Instance list.

- Could you please align the text in the tabs with the "Project" label in the side navigation? - I do realise this is not per design, I would like us to explore it as an option.

al-bakalova:
When performing an action within a tab, the notification should appear below the tabs. This is a general observation and a rule we should always follow.

- Seems to me that renaming instances is no longer possible. I got the same message trying to rename a few different instances. A couple of screenshots to share -

- Renaming the instance + updating the description + updating a network, results in an error. The description was successfully updated though:
Screenshot 2023-03-30 at 10 31 55

- Another attempt to update an instance name:
Screenshot 2023-03-30 at 10 33 55

- Updating the profile names works every time. The notification should appear below the tabs:
Screenshot 2023-03-30 at 10 35 48

lorumic:
> When performing an action within a tab, the notification should appear below the tabs. This is a general observation and a rule we should always follow.
> 
> * Seems to me that renaming instances is no longer possible. I got the same message trying to rename a few different instances.

Wrong PR? :sweat_smile:

al-bakalova:
@lorumic Yes! Please ignore the comments.

lorumic:
@al-bakalova I have applied the changes, could you please have a look?

al-bakalova:

src/sass/_instance_list.scss
```diff
@@ -140,7 +140,7 @@
     }
 
     .actions {
-      width: 160px;
+      width: 210px;
```
edlerd, line 143: There is an issue now with the side panel open and ~1100 px screen width. I think we should not show the action column, when the side panel is open - what do you think?

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -62,12 +62,13 @@ const StartInstanceBtn: FC<Props> = ({ instance }) => {
       aria-label={isLoading ? "Starting" : "Start"}
       title="Start"
     >
-      <i
+      <Icon
         className={
           isLoading
-            ? "p-icon--spinner u-animation--spin"
-            : "p-icon--start-instance"
+            ? "instance-state-icon u-animation--spin"
```
edlerd, line 68: Still a leftover of the `instance-state-icon`, which can be removed. We could use the `classnames` helper here maybe?

src/pages/instances/InstanceDetail.tsx
```diff
@@ -71,18 +71,31 @@ const InstanceDetail: FC = () => {
...
+                className="p-breadcrumbs"
+                aria-label="Breadcrumbs"
+              >
+                <ol className="p-breadcrumbs__items">
+                  <li className="p-breadcrumbs__item">
+                    <Link to={`/ui/${project}/instances`}>Instances</Link>
+                  </li>
+                  <li
+                    className="p-breadcrumbs__item instance-name u-truncate"
+                    title={name}
```
edlerd, line 88: The name can be up to 63 characters long. Considering M as the widest character, [this](https://lxd-ui-260.demos.haus/ui/default/instances/detail/MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM) is the worst it can get. We should find a way for long instance names not to break the layout. I think removing the `max-width` of the `p-panel__title` should fix in most cases, except there is a small range ~1100px width where it is still funny.

src/sass/_instance_detail_page.scss
```diff
@@ -1,10 +1,62 @@
...
+    }
+
+    .instance-name {
+      margin-right: $sph--x-small;
+      max-width: 35vw;
+      overflow: clip !important;
+    }
+
+    .actions-list {
+      line-height: 2rem;
```
edlerd, line 39: Why the line-height?

edlerd (changes requested):

src/sass/_instance_list.scss
lorumic: I had a false memory that we were already hiding the actions when the side panel was open, but maybe we only discussed it and never implemented it. Anyway, yes, I agree, it's a good idea.

src/pages/instances/actions/StartInstanceBtn.tsx
lorumic: Whoops, yes, it can be removed, and good idea about the `classnames` - this actually made me realise that we need exactly the same thing in the `ConfirmationButton`, because when I removed the `baseIcon` I forgot to apply the `u-animation--spin` class when loading, leaving it with a "still" spinner.

src/pages/instances/InstanceDetail.tsx
lorumic: Ok, I set the `max-width` of the `p-panel__title` here to `none`.

src/sass/_instance_detail_page.scss
lorumic: This is a leftover of a previous change, before we introduced the new line on small and medium sizes. Because the `line-height` of the `p-panel__title` gets reduced from `2rem` to `1.5rem` below the large breakpoint, and this was causing a mis-alignment with the breadcrumbs. Now, instead, when we go below the large breakpoint, the breadcrumbs will always be on the first line and the status + actions on a new line, so we don't need to "force" the `line-height` of the latter to `2rem`.

src/pages/instances/InstanceDetail.tsx
edlerd: Let's lower the max-width for the instance-name from `35vw` to `33vw`. Otherwise, it breaks into two lines at ~1100px screen width.

edlerd (approved):
Just one issue with layouting of the instance-name (see comment above). Apart from that ready to go. Good stuff :+1:

src/pages/instances/InstanceDetail.tsx
al-bakalova: Screenshot 2023-03-31 at 9 24 45
e name at some point and maybe display the full one on a hover. Mobile users will only see it

lorumic: Done - thanks for the suggestion. Setting it to `33vw` fixes the problem indeed.

PR #261, opened 2023-03-30
Change defaults for the instance terminal, raise state so it is persisted when reconnecting multiple times

edlerd:
## Done

- change defaults when opening a terminal to use uid/gid 0, with /root as home
- raise the state for the terminal reconnect button, so it is persisted when reconnecting multiple times
- reconnect terminal and graphic/text console on starting of an instance, close error notifications before reconnecting

## QA

1. Run the LXD-UI:
    - open terminal
    - reconnect terminal multiple times
    - start/stop instance when displaying terminal, and graphic/text console. They should reconnect on start and close the error notification.

### DIFF
--- a/src/pages/instances/actions/ReconnectTerminalBtn.tsx
+++ b/src/pages/instances/actions/ReconnectTerminalBtn.tsx
@@ -2,6 +2,8 @@ import React, { FC, useState } from "react";
 import { Button, Icon } from "@canonical/react-components";
 import { LxdTerminalPayload } from "types/terminal";
 import TerminalPayloadForm from "../TerminalPayloadForm";
+import * as Yup from "yup";
+import { useFormik } from "formik";
 
 interface Props {
   onFinish: (data: LxdTerminalPayload) => void;
@@ -10,8 +12,7 @@ interface Props {
 const ReconnectTerminalBtn: FC<Props> = ({ onFinish }) => {
   const [isModal, setModal] = useState(false);
 
-  const confirmModal = (result: LxdTerminalPayload) => {
-    onFinish(result);
+  const closeModal = () => {
     setModal(false);
   };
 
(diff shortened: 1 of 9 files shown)


al-bakalova:
@edlerd, looks good to me. Thank you!

On a different note, when starting or rebooting an instance, while browsing the Terminal tab, please reload the page so that the Terminal itself can reappear.

edlerd:
@al-bakalova good idea! I added the auto-reconnect for the terminal and also for the graphic and text console.

src/pages/instances/actions/ReconnectTerminalBtn.tsx
```diff
@@ -10,23 +12,64 @@ interface Props {
...
+          value: "xterm-256color",
+        },
+        {
+          key: "HOME",
+          value: "/root",
+        },
+      ],
+      user: 0,
+      group: 0,
+    },
```
lorumic, line 50: I see these are defined in multiple places ([here](https://github.com/canonical/lxd-ui/blob/b25dc0b887787e3d491c53cf95f84b7a2e232d73/src/pages/instances/InstanceTerminal.tsx#L24)) and I was wondering, can we extract these shared props so that we can easily reuse them? For the `defaultPayload`, we can probably spread the additional properties that it has.
Also, giving it a type, or at least to the objects inside the `environment` array, would probably not be a bad idea.

edlerd: The formats are different. Here we have a format for the form, in the other case we have the request payload form. We have some logic 5 lines below to transform the form-type into the request-type.

Thinking about it -- We can probably move that logic into the API connector and pass an object with the form-type to it. Then we can also share the default object.

lorumic (approved):

PR #262, opened 2023-03-30
Add interface name to the network table

edlerd:
## Done

- added the interface name to the network table on the instance overview

## QA

1. Run the LXD-UI:
    - check the network table on the instance overview of various instances


al-bakalova:
@edlerd Looks good to me. I wonder if we should rearrange the columns and bring it to a second position. I will leave the choice to you as you have a better understanding of the topic! :)

edlerd:
> @edlerd Looks good to me. I wonder if we should rearrange the columns and bring it to a second position. I will leave the choice to you as you have a better understanding of the topic! :)

Yeah, that is better, I switched it around.

lorumic (approved):

PR #264, opened 2023-03-31
Instance and profile edit fixes

edlerd:
## Done

- ensure all fields from the instance and profile are kept, even if they are not supported by the UI
- simplify edit code
- added rename API functions for instances and profiles (unused atm, but we'll need it shortly)
- this is taken from https://github.com/canonical/lxd-ui/pull/253 but avoiding any visible UI changes.
- adjusted error and success notifications on edit

## QA

1. Run the LXD-UI:
    - edit instances and a profiles

### DIFF
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -1,5 +1,9 @@
 import { watchOperation } from "./operations";
-import { handleResponse, handleTextResponse } from "util/helpers";
+import {
+  handleEtagResponse,
+  handleResponse,
+  handleTextResponse,
+} from "util/helpers";
 import { LxdInstance } from "types/instance";
 import { LxdTerminal, TerminalConnectPayload } from "types/terminal";
 import { LxdApiResponse } from "types/apiResponse";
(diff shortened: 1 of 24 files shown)


src/pages/profiles/EditProfileForm.tsx
```diff
@@ -123,6 +123,23 @@ const EditProfileForm: FC<Props> = ({ profile }) => {
...
+
+    const supportedMainKeys = new Set([
+      "name",
+      "description",
+      "devices",
+      "config",
+    ]);
+    const additionalMainKeys = Object.fromEntries(
+      Object.entries(profile).filter(([key]) => !supportedMainKeys.has(key))
+    );
```
lorumic, line 141: This looks very similar to the one in `EditInstanceForm`. I was wondering if we could somehow reuse this code. Maybe one idea could be to have a `util` function that takes as input the `supportedMainKeys` (which are different between instances and profiles) and the `LxdInstance | LxdProfile` object, from which it can extract the `config` object. The function would then return the `additionalConfigKeys` and `additionalMainKeys` as a tuple, for example. What do you think?

edlerd: yeah, that is much better to put this logic into a reusable function.

src/api/instances.tsx
```diff
@@ -74,6 +65,26 @@ export const updateInstance = (
...
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/instances/${oldName}?project=${project}`, {
+      method: "POST",
+      body: JSON.stringify({
+        name: newName,
+      }),
+    })
+      .then(handleResponse)
+      .then((data: LxdOperation) => {
+        watchOperation(data.operation, 120).then(resolve).catch(reject);
```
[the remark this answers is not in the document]
edlerd: I don't see a benefit here. Calling the constants medium/long/short is just hiding the details for not much gain.

lorumic (approved):

PR #265, opened 2023-03-31
Fix navigation toggle and logo are sticky WD-2987

edlerd:
## Done

- fix navigation on small height screens
- avoid overlaying the logo with the project selector
- make the navigation toggle at the bottom sticky

Fixes WD-2987

## QA

1. Run the LXD-UI:
    - test navigation on all possible screen dimensions

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -218,19 +218,21 @@ const Navigation: FC = () => {
                     </a>
                   </li>
                 </ul>
-                <Button
-                  appearance="base"
-                  aria-label={`${
-                    menuCollapsed ? "expand" : "collapse"
-                  } main navigation`}
-                  hasIcon
-                  dense
-                  className="sidenav-toggle is-dark u-no-margin l-navigation-collapse-toggle u-hide--small"
-                  onClick={hardToggleMenu}
-                >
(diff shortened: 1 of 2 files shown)


lorumic:
The toggle "jumps up" the external links on large screens:

Peek 2023-03-31 11-38

edlerd:
> The toggle "jumps up" the external links on large screens:

Right, I didn't test the large bit, now it should work smoothly.

al-bakalova:
The problem seems to still be there @edlerd 
Screenshot 2023-03-31 at 15 06 07

edlerd:
Will have another look. Maybe will have to buy a bigger monitor in between :D

edlerd:
The problem was with the non-x-large screens. Now it works on all dimensions.

al-bakalova:
Looks good to me! Danke!

lorumic:
I see that it gets sometime clipped by the scrollbar:

image

I wonder if we should increase its distance from the right edge? :thinking:

al-bakalova:
Maybe we can try to align the collapse/expand button with the edge of the project switch field?

edlerd:
> Maybe we can try to align the collapse/expand button with the edge of the project switch field?

did that.

lorumic (approved):

PR #266, opened 2023-04-03
Add bulk select and delete for snapshots WD-3025

edlerd:
## Done

- added bulk selection for snapshots
- allow deletion of multiple snapshots
- added a reusable component `SelectableMainTable` that adds checkboxes to all rows.

Fixes [WD-3025](https://warthogs.atlassian.net/browse/WD-3025)

## QA

1. Run the LXD-UI:
    - Select instance snapshots (all on page, all on instance)
    - Browse snapshot pages and search while/before selecting
    - Delete multiple selected snapshots

[WD-3025]: https://warthogs.atlassian.net/browse/WD-3025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/SelectableMainTable.tsx
+++ b/src/components/SelectableMainTable.tsx
@@ -163,7 +163,7 @@ const SelectableMainTable: FC<Props> = ({
               ) : (
                 <>
                   <b>{selectedNames.length}</b> {itemName}
-                  {selectedNames.length > 1 && "s"} selected{" "}
+                  {selectedNames.length > 1 && "s"} selected.{" "}
                   <Button
                     appearance="link"
                     className="u-no-margin--bottom u-no-padding--top"
--- a/src/components/SelectableMainTable.tsx
+++ b/src/components/SelectableMainTable.tsx
@@ -1,0 +1,192 @@
+import React, { FC } from "react";
+import {
+  MainTableRow,
+  Props as MainTableProps,
+} from "@canonical/react-components/dist/components/MainTable/MainTable";
+import {
+  Button,
+  CheckboxInput,
+  ContextualMenu,
+  Icon,
+  MainTable,
+  Notification,
+} from "@canonical/react-components";
+import classnames from "classnames";
+
+interface SelectableMainTableProps {
+  allNames: string[];
+  itemName: string;
+  parentName: string;
+  selectedNames: string[];
+  setSelectedNames: (val: string[]) => void;
+  processingNames: string[];
+  rows: MainTableRow[];
+}
+
+type Props = SelectableMainTableProps & MainTableProps;
+
+const SelectableMainTable: FC<Props> = ({
+  allNames,
+  itemName,
+  parentName,
+  selectedNames,
+  setSelectedNames,
+  processingNames,
+  rows,
+  headers,
+  ...props
+}: Props) => {
+  const isAllSelected = selectedNames.length === allNames.length;
+  const isSomeSelected = selectedNames.length > 0;
+
+  const selectAll = () => {
+    setSelectedNames(allNames);
+  };
+
+  const selectPage = () => {
+    setSelectedNames(rows.map((row) => row.name ?? ""));
+  };
+
+  const selectNone = () => {
+    setSelectedNames([]);
+  };
+
+  const headersWithCheckbox = [
+    {
+      content: (
+        <>
+          <CheckboxInput
+            label=""
+            labelClassName="multiselect-checkbox"
+            checked={isAllSelected}
+            indeterminate={isSomeSelected && !isAllSelected}
+            onChange={isSomeSelected ? selectNone : selectPage}
+            aria-label="Select all"
+          />
+          <ContextualMenu
+            position="left"
+            title="Multiselect"
+            toggleAppearance="base"
+            toggleClassName="has-icon u-no-margin--bottom"
+            toggleLabel={<Icon name="chevron-down" />}
+            toggleProps={{
+              "aria-label": "multiselect rows",
+            }}
+            links={[
+              {
+                children: `Select all ${itemName}s on this page`,
+                onClick: selectPage,
+              },
+              {
+                children: `Select all ${parentName} ${itemName}s`,
+                onClick: selectAll,
+              },
+            ]}
+          />
+        </>
+      ),
+      className: "select select-header",
+    },
+    ...(headers ?? []),
+  ];
+
+  const rowsWithCheckbox = rows.map((row) => {
+    const isRowSelected = selectedNames.includes(row.name ?? "");
+    const isRowProcessing = processingNames.includes(row.name ?? "");
+
+    const toggleRow = () => {
+      const newSelection = isRowSelected
+        ? selectedNames.filter((candidate) => candidate !== row.name)
+        : [...selectedNames, row.name ?? ""];
+      setSelectedNames(newSelection);
+    };
+
+    row.columns = [
+      {
+        content: (
+          <CheckboxInput
+            label=""
+            labelClassName="u-no-margin--bottom"
+            checked={isRowSelected}
+            onChange={toggleRow}
+            disabled={isRowProcessing}
+            aria-label={`Select ${row.name ?? ""}`}
+          />
+        ),
+        role: "rowheader",
+        className: "select",
+      },
+      ...(row.columns ?? []),
+    ];
+
+    row.className = classnames(row.className, {
+      "selected-row": isRowSelected,
+      "processing-row": isRowProcessing,
+    });
+    row.key = row.name;
+
+    return row;
+  });
+
+  const selectionState = isSomeSelected
+    ? [
+        {
+          className: "select-notification",
+          key: "select-info",
+          expanded: true,
+          expandedContent: (
+            <Notification
+              borderless
+              className="u-no-margin--bottom"
+              title="Selection"
+            >
+              {isAllSelected ? (
+                <>
+                  {allNames.length === 1 ? (
+                    <>
+                      <b>1</b> {itemName} of this {parentName} is selected.{" "}
+                    </>
+                  ) : (
+                    <>
+                      All <b>{allNames.length}</b> {itemName}s of this{" "}
+                      {parentName} are selected.{" "}
+                    </>
+                  )}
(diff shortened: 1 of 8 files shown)


al-bakalova:
@edlerd  Looks good and works as expected. A few design comments:
- We are on board with your suggestion to introduce [the selected background ](https://discourse.ubuntu.com/t/row-selection/19847)
- The "Selection" message is a reused notification component. We have decided to keep the close button in its top right corner.
- Could you please try to bold the numbers in the selection message? 
As an example -  **14** snapshots selected on this page. Select all **354** instance snapshots.
- The "Clear selection" button has an additional close icon to the right (as initially introduced in MAAS)

- And finally, we are currently discussing the Search + select all experience with the MAAS and MAAS Site Manager teams. Let me get back to you with our final decision.

edlerd:
@al-bakalova all points should be resolved. Apart from the last one, that is blocked due to external input.

al-bakalova:
@edlerd This looks great!
One small issue - could you please align the "Delete" and the "Clear selection" button labels?

Screenshot 2023-04-04 at 13 42 32

I have also had a discussion with the design team. The reason MAAS does this is lack of space. If we have enough space to keep both the action buttons (delete, clear selection) and the search bar, we can do so. Maybe on smaller screens, we can keep hiding the search bar.

al-bakalova:
Looks good to me! I think this is good to go.
One think to mention here - bulk deleting snaps takes quite some time (even when I am testing with only 2 or 3 items). 
Is this due to us using a test environment? If not, maybe we should consider a loader? wdyt?

edlerd:
Brought back the loading state/spinner for the bulk delete and also for the single item actions restore/delete.
Also made the header sticky and adjusted the column widths.
@al-bakalova  please have another look at it.

al-bakalova:
@edlerd The sticky selection message looks perfect, and resizes in a way that makes sense.

When it comes to the loader, I believe we had a hard time to explain our ideas.

To clarify the problem - deleting items takes time, once the user confirms the deletion, they go back to the table where the table rows are still selected, the restore and the delete (row) buttons still being active. This experience is rather confusing. We should find a better way to provide feedback to the user.

Your recent improvements make sense but the item rows are still selected, the buttons in the table still being active.

What I had in mind is to somehow "lock" the table while performing the actions. I have a couple ideas that I am happy to discuss further (the illustrations are from other design systems):

- Disable the rows that we are about to delete and make these disappear as soon as the deletion is done
Screenshot 2023-04-05 at 9 39 52
- Disable the whole table and add a spinner on top of it
Screenshot 2023-04-05 at 9 39 20

edlerd:
> To clarify the problem - deleting items takes time, once the user confirms the deletion, they go back to the table where the table rows are still selected, the restore and the delete (row) buttons still being active. This experience is rather confusing. We should find a better way to provide feedback to the user.

I added some code so we show something similar to the 1st screenshot with the disabled rows - setting opaque to 0.5, disabling the select and hiding the actions on the rows that are bulk-processing.

al-bakalova:
@edlerd Looks good, thank you!

al-bakalova:
- Case 1: That's expected. Now that you mention the two cases side by side, it probably makes sense to keep it and continue to show the selected items count. It will be useful.
- Case 2: The page contains 50 items but the selection was "All in the project" before the user decided to remove one. So the count still makes sense. The checkbox on the other hand - should change to indeterminate state, that's a good catch.
@lorumic @edlerd

lorumic:
Yeah the case 1 was really just a base case to highlight that there might be something wrong in case 2. It was there for the sake of comparison.
At this point, I wonder if it could help to remove the "on this page" part from the message, because it's rather confusing. Also, if we change something in the pagination (the current page or the page size), that message doesn't change, so it's not really reflecting the current selection on the current page. I propose to just say "N snapshots selected".

edlerd:
Changed the selection notification, it's text and also the select all checkbox state as discussed @lorumic  @al-bakalova

lorumic:
> Changed the selection notification, it's text and also the select all checkbox state as discussed @lorumic @al-bakalova

The state of the general checkbox is indeterminate even when all snapshots are selected:

image

edlerd:
> The state of the general checkbox is indeterminate even when all snapshots are selected:

Whoops, that didn't happen to me before. Change the condition for indeterminate as suggested.

src/components/SelectableMainTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+          expandedContent: (
+            <Notification
+              borderless
+              className="u-no-margin--bottom"
+              title="Selection"
+            >
+              {isAllSelected ? (
+                <>
+                  All <b>{allNames.length}</b> {itemName}s of this {parentName}{" "}
+                  are selected.{" "}
```
lorumic, line 143: Missing a singular/plural check here:

image

src/components/SelectableMainTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+                    appearance="link"
+                    className="u-no-margin--bottom u-no-padding--top"
+                    onClick={selectNone}
+                  >
+                    Clear selection
+                  </Button>
+                </>
+              ) : (
+                <>
+                  <b>{selectedNames.length}</b> {itemName}s selected{" "}
```
lorumic, line 154: And here as well, the singular vs plural:

image

src/components/SelectableMainTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+                  <Button
+                    appearance="link"
+                    className="u-no-margin--bottom u-no-padding--top"
+                    onClick={selectAll}
+                  >
+                    Select all <b>{allNames.length}</b> {parentName} {itemName}s
```
lorumic, line 160: If we make the `parentName` optional, we should also make sure that there is no repeated space when `parentName` is `undefined`.

src/pages/instances/actions/snapshots/SnapshotBulkDelete.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const queryClient = useQueryClient();
+
+  const handleDelete = () => {
+    setLoading(true);
+    onStart();
+    deleteSnapshotBulk(instance, snapshotNames)
+      .then(() =>
+        onSuccess(
+          <>
+            <b>{snapshotNames.length}</b> Snapshots deleted.
```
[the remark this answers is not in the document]
lorumic: Oh, and also, I realised now that this, too, needs a singular/plural check.

lorumic:
Immediately after we select a snapshot (except the first of the page), the previous one gets the focus.
Similarly, after we deselect a snapshot (except the last of the page), the next one gets the focus.
See this illustrated by the following screen capture:

Peek 2023-04-05 18-24

Can we fix this? It might be annoying in terms of accessibility.

src/components/SelectableMainTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+              {
+                children: `Select all ${itemName}s on this page`,
+                onClick: selectPage,
+              },
+              {
+                children: `Select all ${parentName} ${itemName}s`,
+                onClick: selectAll,
+              },
+            ]}
+          />
```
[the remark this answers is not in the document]
edlerd: I think we do always have a `parentName`. For instances, the parent is a project.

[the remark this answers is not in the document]
lorumic: I didn't think that you wanted to specify the project as the parent for instances as I considered that implicit information, but I guess in that case you could keep it as it.

edlerd:
> Can we fix this? It might be annoying in terms of accessibility.

Adding a key based on the unique snapshot names fixes it. Also addressed the various things mentioned above.

src/components/SelectableMainTable.tsx
edlerd: As mentioned above, I think we'll always have a parentName.

al-bakalova:
Once I select a single snapshot the system is already inviting me to select the exhaustive list of snapshots. Shouldn't we keep the two step suggestions here? -

1 instance selected. Select all instances on the page.
50 snapshots selected. Select all 333 instance snapshots.

@edlerd @lorumic

edlerd:
> Once I select a single snapshot the system is already inviting me to select the exhaustive list of snapshots. Shouldn't we keep the two step suggestions here? -
> 
> 1 snapshot selected. Select all instances on the page. 50 snapshots selected. Select all 333 instance snapshots.

This gets complicated quickly, because the user might select 2 items on different pages and we'd need to figure out if that is the case to refer to the right page - or to not have that 1st step in that case. I'd keep it simple as it is.

al-bakalova:
@edlerd I hear your points and I agree. Let's not overcomplicate it.

Side question, I believe the delete message should say "Are you sure you want to delete **these**?" and not "them" but my English grammar is sometimes questionable. 

Screenshot 2023-04-06 at 10 50 27

lorumic:
> Side question, I believe the delete message should say "Are you sure you want to delete **these**?" and not "them" but my English grammar is sometimes questionable.

I have suggested changing **these** with **them**. My reasoning was: we use "it" for the singular, and ["them" is the (object) plural of "it"](https://forum.wordreference.com/threads/plural-of-it.2557149/).

src/components/SelectableMainTable.tsx
```diff
@@ -0,0 +1,192 @@
...
+                    className="u-no-margin--bottom u-no-padding--top"
+                    onClick={selectNone}
+                  >
+                    Clear selection
+                  </Button>
+                </>
+              ) : (
+                <>
+                  <b>{selectedNames.length}</b> {itemName}
+                  {selectedNames.length > 1 && "s"} selected{" "}
```
lorumic, line 166:
```suggestion
                  {selectedNames.length > 1 && "s"} selected.{" "}
```
Sorry, I just happened to notice this now - in the `isAllSelected` case above, we have a "." at the end of the message, but here we don't. I suggest adding it here too for consistency - and also because it looks better, given that we have another sentence right next to it.

lorumic (approved):

PR #268, opened 2023-04-03
Add snapshot configuration WD-3026

edlerd:
## Done

- added configuration button to instance snapshot tab
- added modal to configure automatic snapshots for the above button

Fixes WD-3026
## QA

1. Run the LXD-UI:
    - go to the snapshots tab for an instance
    - see configuration, change it and save it

### DIFF
--- a/src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
+++ b/src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
@@ -1,0 +1,107 @@
+import React, { FC, KeyboardEvent } from "react";
+import { Button, Modal } from "@canonical/react-components";
+import { LxdInstance } from "types/instance";
+import { useFormik } from "formik";
+import { updateInstance } from "api/instances";
+import { queryKeys } from "util/queryKeys";
+import { EditInstanceFormValues } from "pages/instances/EditInstanceForm";
+import { useQueryClient } from "@tanstack/react-query";
+import { useNotify } from "context/notify";
+import SnapshotsForm from "pages/instances/forms/SnapshotsForm";
+import { useParams } from "react-router-dom";
+import {
+  getInstanceEditValues,
+  getInstancePayload,
+  InstanceEditSchema,
+} from "util/instanceEdit";
+
+interface Props {
+  instance: LxdInstance;
+  close: () => void;
+}
+
+const ConfigureSnapshotModal: FC<Props> = ({ instance, close }) => {
+  const notify = useNotify();
+  const { project } = useParams<{ project: string }>();
+  const queryClient = useQueryClient();
+
+  const formik = useFormik<EditInstanceFormValues>({
+    initialValues: getInstanceEditValues(instance),
+    validationSchema: InstanceEditSchema,
+    onSubmit: (values) => {
+      const instancePayload = getInstancePayload(
+        instance,
+        values
+      ) as LxdInstance;
+
+      updateInstance(instancePayload, project ?? "")
+        .then(() => {
+          notify.success("Snapshot configuration updated.");
+          close();
+        })
+        .catch((e: Error) => {
+          notify.failure("Snapshot configuration update failed", e);
+        })
+        .finally(() => {
+          void queryClient.invalidateQueries({
+            queryKey: [queryKeys.instances],
+          });
+        });
+    },
+  });
+
+  const handleEscKey = (e: KeyboardEvent<HTMLElement>) => {
+    if (e.key === "Escape") {
+      close();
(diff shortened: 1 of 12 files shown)


al-bakalova:
@david, the snapshot configuration modal looks good and works as expected. However, there are a couple of things that I would like to clarify:
- Please move the "Edit configuration" button to the left and use the default style. The reasons why - it is not the primary action in the modal. The primary goal here is for the user to "consult" the configuration. What is more, the current placement and styling might be misleading and instead of closing the modal (usually there is a green OK button that just closes the modal), our users will always enter the edit mode. This is why the button is intentionally placed on the left side of the modal.

edlerd:
> * Please move the "Edit configuration" button to the left

Changed the button style and placement.

al-bakalova:
@edlerd Looks good to me! Thank you!
Works as expected. Once the config is updated, it also updates in the Configuration tab.
I think this is good to go.

src/pages/instances/forms/ConfigurationTable.tsx
lorumic, line 1: You should set the `zIndex` prop for this [`Tooltip`](https://github.com/canonical/lxd-ui/blob/70da6f227dc00160ccaaa6db083ef800b591966d/src/pages/instances/forms/ConfigurationTable.tsx#L91) if you want it to be displayed over a modal.

src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
lorumic, line 1: When the modal gets opened, the element that gets the focus is the "Edit configuration" button. For this reason, if I hit the Enter key after opening the modal, before any other interaction, this would trigger the switch to edit mode. Is this expected/desired? If not, maybe we should try to move the focus to the "Close" button?

al-bakalova: I agree with Michele, good catch! It would be better to focus on the "Close" button as our primary goal here is to "consult" rather than edit the configs. 

Also, the height of the modal is now different. There is no need to introduce all the white space below the table.
Screenshot 2023-04-04 at 13 38 47

src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+const ConfigureSnapshotModal: FC<Props> = ({ instance, close }) => {
+  const notify = useNotify();
+  const { project } = useParams<{ project: string }>();
+  const queryClient = useQueryClient();
+
+  const InstanceSchema = Yup.object().shape({
+    name: Yup.string().required("Instance name is required"),
+    instanceType: Yup.string().required("Instance type is required"),
+  });
```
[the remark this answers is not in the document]
edlerd: For creation it is different, but we can reuse for the two edits.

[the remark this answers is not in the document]
lorumic: The only difference that I see is that the name is optional in creation and required in edit. But I would try to find a workaround for this, and still reuse the same schema for both creation and edit. Especially because we aren't doing the "deduplicate" test on edit, which should be done instead.
You can make it a function `getInstanceSchema` and pass the mutation (`optional` / `required`) to it maybe?

edlerd:
All things mentioned above should be resolved @al-bakalova and @lorumic

src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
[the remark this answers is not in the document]
edlerd: I think it is better not to mix creation/editing schemas.

src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <Modal
+      close={close}
+      className="edit-snapshot-config"
+      title="Snapshot configuration"
+      buttonRow={
+        formik.values.readOnly ? (
+          <div className="u-space-between u-flex-row-reverse">
+            <Button className="u-no-margin--bottom" onClick={close}>
+              Close
+            </Button>
```
lorumic, line 69: You might want to add the `u-no-margin--right` here so that the button keeps the correct right alignment.

lorumic:
> All things mentioned above should be resolved @al-bakalova and @lorumic

LGTM, is everything good for you as well @al-bakalova?

al-bakalova:
Looks good to me! Thank you!

PR #272, opened 2023-04-04
fix(errors) use titled notifications for all error notifications

edlerd:
## Done

- changed the failure function so it expects a title and the error, optional message and actions
- adjusted all failures, so we have concise and consistent titles
- this should make it easier for the user to read the actual error message (in the notification body). Before we'd mix our message with the API error message in the body, making it hard to gasp the notification.

## QA

1. Run the LXD-UI:
    - check the errors that are outputted


edlerd:
@al-bakalova @lorumic probably not possible to test this on the demo server. Basically, I changed all the failures where we notify the user, so that the bold title of the notification comes from us and is very concise (i.e. "Image deletion failed"). We still show the API error, which might hint at a solution or a root cause. But that one will not be bold. Instead the API error is part of the notification in standard text-weight.

I was aiming to not hide the API error behind our message. Before, we'd have a bold title of "Error" in any case and prepend our message to the API error in the notification body. This led to the situation where the API error is hard to find and the root cause easy to miss.

al-bakalova:
@edlerd Good idea, your solution makes perfect sense. Should we merge and then start paying more attention to our error messages just to QA these later? :)

al-bakalova:
@edlerd The messages look good. However, the general structures seem to be:

**- [action] failed**
For example:
- Instance deletion failed
- Instance creation failed
- Instance update failed
- Connection failed (to replace "Failed to connect")
- Terminal connection failed

Following this logic, should we change all the "Could not load [item]" to
- **Loading [item] failed**

For example:
- Loading instance details failed
- Loading metrics failed
- Loading networks failed
- Loading profiles failed

Maybe it's good to confirm these with @ru-fu. I am not sure my English is on point :D

edlerd:
Changed the text accordingly to your suggestions @al-bakalova

al-bakalova:
@edlerd These look good, thank you!

src/pages/profiles/ProfileDetail.tsx
```diff
@@ -67,7 +67,7 @@ const ProfileDetail: FC = () => {
         <div className="p-panel__content">
           <NotificationRow />
           {isLoading && <Loader text="Loading profile details..." />}
-          {!isLoading && !profile && <>Could not load profile details.</>}
+          {!isLoading && !profile && <>Loading profile details.</>}
```
lorumic, line 70:
```suggestion
          {!isLoading && !profile && <>Loading profile failed</>}
```

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -28,7 +28,7 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
         );
       })
       .catch((e) => {
-        notify.failure(`Error deleting project ${project.name}.`, e);
+        notify.failure(`Project deletion failed`, e);
```
lorumic, line 31:
```suggestion
        notify.failure("Project deletion failed", e);
```

src/pages/storages/StorageDetail.tsx
```diff
@@ -35,13 +35,13 @@ const StorageDetail: FC = () => {
   });
 
   if (error) {
-    notify.failure("Could not load storage details.", error);
+    notify.failure("Loading storage  failed", error);
```
lorumic, line 38:
```suggestion
    notify.failure("Loading storage details failed", error);
```

lorumic (changes requested):

lorumic (approved):

PR #273, opened 2023-04-04
Improve snapshots section in instance side panel [WD-2710]

lorumic:
## Done

- Changed heading of the snapshots section to "Snapshots"
- Added local sort logic to sort the snapshots from the most recent, so that we do not rely on the BE implementation for the order
- Added a "View all (N)" link below the most 5 recent snapshots (if N > 5)

Fixes [WD-2710](https://warthogs.atlassian.net/browse/WD-2710)

## QA

1. Run the LXD-UI:
    - Open an instance side panel and check the snapshots section

[WD-2710]: https://warthogs.atlassian.net/browse/WD-2710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/InstanceDetailPanel.tsx
+++ b/src/pages/instances/InstanceDetailPanel.tsx
@@ -201,32 +201,53 @@ const InstanceDetailPanel: FC = () => {
                         <Link
                           to={`/ui/${instance.project}/instances/detail/${instance.name}/snapshots`}
                         >
-                          Recent snapshots
+                          Snapshots
                         </Link>
                       </h3>
                       {instance.snapshots?.length ? (
-                        <List
-                          items={instance.snapshots
-                            .reverse()
-                            .slice(0, RECENT_SNAPSHOT_LIMIT)
-                            .map((snapshot) => (
-                              <Row key={snapshot.name} className="no-grid-gap">
-                                <Col
-                                  size={4}
-                                  className="u-truncate"
-                                  title={snapshot.name}
+                        <>
+                          <List
+                            className="u-no-margin--bottom"
+                            items={instance.snapshots
+                              .slice()
+                              .sort((snap1, snap2) => {
+                                const a = snap1.created_at;
+                                const b = snap2.created_at;
+                                return a > b ? -1 : a < b ? 1 : 0;
+                              })
+                              .slice(0, RECENT_SNAPSHOT_LIMIT)
+                              .map((snapshot) => (
+                                <Row
+                                  key={snapshot.name}
+                                  className="no-grid-gap"
                                 >
-                                  <ItemName item={snapshot} />
-                                </Col>
-                                <Col
-                                  size={8}
-                                  className="p-snapshot-creation u-align--right u-text--muted u-no-margin--bottom"
-                                >
-                                  <i>{isoTimeToString(snapshot.created_at)}</i>
-                                </Col>
-                              </Row>
-                            ))}
-                        />
+                                  <Col
+                                    size={4}
+                                    className="u-truncate"
+                                    title={snapshot.name}
+                                  >
+                                    <ItemName item={snapshot} />
+                                  </Col>
+                                  <Col
+                                    size={8}
+                                    className="p-snapshot-creation u-align--right u-text--muted u-no-margin--bottom"
+                                  >
+                                    <i>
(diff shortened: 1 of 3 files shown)


al-bakalova:
@lorumic Good job! the snapshots section looks good and works as expected. A few small improvements to suggest:
- Could you please align left the creation date and time? These should be left aligned with the values from the sections above.

src/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -201,32 +201,53 @@ const InstanceDetailPanel: FC = () => {
...
-                            .reverse()
-                            .slice(0, RECENT_SNAPSHOT_LIMIT)
-                            .map((snapshot) => (
-                              <Row key={snapshot.name} className="no-grid-gap">
-                                <Col
-                                  size={4}
-                                  className="u-truncate"
-                                  title={snapshot.name}
+                        <>
+                          <List
```
edlerd, line 209: Also to align the width with elements above, I think we shouldn't use a list here (or the row / col below). Maybe use a tr for each snapshot to match the structure above.

src/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -195,40 +195,52 @@ const InstanceDetailPanel: FC = () => {
...
+                            >
+                              <ItemName item={snapshot} />
+                            </th>
+                            <td className="u-text--muted u-no-margin--bottom">
+                              <i>{isoTimeToString(snapshot.created_at)}</i>
+                            </td>
+                          </tr>
+                        ))}
+                      {instance.snapshots.length > RECENT_SNAPSHOT_LIMIT && (
+                        <Link
```
edlerd, line 233: I think this is not valid here to output a link in the table body.

src/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -195,40 +195,52 @@ const InstanceDetailPanel: FC = () => {
...
+                        .slice(0, RECENT_SNAPSHOT_LIMIT)
+                        .map((snapshot) => (
+                          <tr key={snapshot.name} className="u-no-border">
+                            <th
+                              title={snapshot.name}
+                              className="snapshot-name u-truncate"
+                            >
+                              <ItemName item={snapshot} />
+                            </th>
+                            <td className="u-text--muted u-no-margin--bottom">
```
edlerd, line 227: `u-no-margin--bottom` can be removed

edlerd (commented):
Some small notes.

I think the empty state in the side panel looks a bit odd, because of the two links with the caption "snapshot" next to each other.

al-bakalova:
@edlerd @lorumic Looks good to me!
Personally, I don't mind the the links issue. If you both agree that's odd, we can remove the second line of the empty state message.

al-bakalova:
When I hover on long snapshot names, I get the text pointer, and not the arrow one. Do we have control over this?

edlerd:
> @edlerd @lorumic Looks good to me! Personally, I don't mind the the links issue. If you both agree that's odd, we can remove the second line of the empty state message.

We can keep it as is, maybe it was just me at that time wondering about it.

edlerd:
> When I hover on long snapshot names, I get the text pointer, and not the arrow one. Do we have control over this?
> It seems to be a more general issue, since I also have the same pointer when hovering the general details

This is standard in browsers, when hovering text it becomes the text-cursor, when hovering links the hand-cursor and if none of the two match, then the arrow cursor.
Though we do have control over the cursor for each element - I think we shouldn't override it. Unless we have a good reason for it.

lorumic:
> > @edlerd @lorumic Looks good to me! Personally, I don't mind the the links issue. If you both agree that's odd, we can remove the second line of the empty state message.
> 
> We can keep it as is, maybe it was just me at that time wondering about it.

Yes, I also think it should be kept, because it's not immediate that by clicking on the link in the header you can create a snapshot. Ideally, as an idea for future improvement (= out of scope for this PR), we could make the two links behave differently. The one in the header, just as it does now, will open the list of snapshots, and the one in the empty state that says "Create" will open the snapshots tab but with the creation modal already open.

edlerd (approved):

PR #274, opened 2023-04-04
Extend usePagination hook, move pageSize state into it [WD-2713]

lorumic:
## Done

- Created an adapter for react-components' `usePagination` hook that holds the `pageSize` and exposes it together with a function to update it.
- Reset the current page to 1 when the page size is updated.

Fixes [WD-2713](https://warthogs.atlassian.net/browse/WD-2713)

## QA

1. Run the LXD-UI:
    - Check that the pagination is working as expected in the instance list as well as the snapshot list.
    - Check that when changing the page size on a page that is not 1, the current page gets reset to 1.

[WD-2713]: https://warthogs.atlassian.net/browse/WD-2713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/pagination.tsx
```diff
@@ -0,0 +1,42 @@
...
+    itemsPerPage: _pageSize,
+    autoResetPage: true,
+  });
+
+  const setPageSize = (val: number) => {
+    usePagination.paginate(1);
+    _setPageSize(val);
+  };
+
+  return { ...usePagination, pageSize: _pageSize, setPageSize };
```
edlerd, line 41: Now that we have the pageSize in here, can we also move the totalPages calculation into this hook?

src/pages/instances/InstanceList.tsx
```diff
@@ -279,10 +277,10 @@ const InstanceList: FC = () => {
     pageData: pageInstances,
```
edlerd, line 277: I wonder if it would be better to not unpack here: `const page = usePagination...`. Then we could simplify below to `<Pagination {...page}` and adjust the 3-4 usages of details from it to calling `page.pageData`. But we'd need to rename `setCurrentPage` to `paginate` in the `Pagination` component to make the props match.

lorumic: I would avoid this change, because I don't see any particular benefit and it complicates referencing `pageInstances` - e.g. as the `rows` prop of the `MainTable`, or for the check in `visibleCount`.
For example, you'd need to do something like this:
```
<MainTable
  headers={getHeaders(userHidden.concat(sizeHidden))}
  rows={page.pageData}
```
And similar for the `visibleCount`. Seems to me that we're complicating things instead of making them simpler and cleaner.

edlerd (approved):

PR #275, opened 2023-04-05
Add instance and profile renaming

edlerd:
## Done

- added instance renaming
- added profile renaming

Fixes [WD-2780](https://warthogs.atlassian.net/browse/WD-2780)

## QA

1. Run the LXD-UI:
    - rename some instances
    - rename some profiles

[WD-2780]: https://warthogs.atlassian.net/browse/WD-2780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/InstanceDetailHeader.tsx
+++ b/src/pages/instances/InstanceDetailHeader.tsx
@@ -1,0 +1,74 @@
+import React, { FC, useState } from "react";
+import { Link } from "react-router-dom";
+import InstanceStateActions from "pages/instances/actions/InstanceStateActions";
+import DeleteInstanceBtn from "./actions/DeleteInstanceBtn";
+import { LxdInstance } from "types/instance";
+import InstanceRename from "pages/instances/forms/InstanceRename";
+import { Tooltip } from "@canonical/react-components";
+
+interface Props {
+  name: string;
+  instance?: LxdInstance;
+  project: string;
+}
+
+const InstanceDetail: FC<Props> = ({ name, instance, project }) => {
+  const [isRename, setRename] = useState(false);
+  const canRename = instance?.status === "Stopped";
+
+  return (
+    <div className="p-panel__header">
+      <h1 className="u-off-screen">{name}</h1>
+      {instance ? (
+        <div className="p-panel__title">
+          <nav
+            key="breadcrumbs"
+            className="p-breadcrumbs"
+            aria-label="Breadcrumbs"
+          >
+            <ol className="p-breadcrumbs__items">
+              <li className="p-breadcrumbs__item">
+                <Link to={`/ui/${project}/instances`}>Instances</Link>
+              </li>
+              {isRename ? (
+                <li className="p-breadcrumbs__item instance-rename">
+                  <InstanceRename
+                    instance={instance}
+                    project={project}
+                    closeForm={() => setRename(false)}
+                  />
+                </li>
+              ) : (
+                <li
+                  className="p-breadcrumbs__item instance-name u-truncate"
+                  onClick={() => canRename && setRename(true)}
+                  title={name}
+                >
+                  <Tooltip
+                    message={!canRename && "Stop the instance to rename"}
+                    position="btm-left"
+                  >
+                    {name}
+                  </Tooltip>
+                </li>
+              )}
+            </ol>
+          </nav>
+          {!isRename && (
+            <div>
+              <i className="status u-text--muted">{instance.status}</i>
+              <InstanceStateActions key="state" instance={instance} />
+            </div>
+          )}
+        </div>
+      ) : (
+        <h4 className="p-panel__title">{name}</h4>
+      )}
+      <div className="p-panel__controls">
+        {instance && !isRename && <DeleteInstanceBtn instance={instance} />}
+      </div>
+    </div>
+  );
+};
+
+export default InstanceDetail;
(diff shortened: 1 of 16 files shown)


al-bakalova:
@edlerd, the experience looks good and behaves as expected. A few improvements:

-Could you please capitalise the "Save" button label
Screenshot 2023-04-05 at 17 56 03

- Please shorten the copy a little. Suggested copy:
**Renaming failed.** Renaming running instances not allowed.
Screenshot 2023-04-05 at 17 59 59

- Success message
**Success**. Instance renamed.

al-bakalova:
@edlerd These changes look good to me. I have just realised that we cannot rename frozen/ paused instances either. 
Maybe we need a separate error message for this state?

al-bakalova:
This looks good. A few comments on my side:
- on hover, the background includes both the caret and the item's name. Is it possible to add the hover only to the items name, where the field will eventually appear in edit mode?
- hovering on the instance name looks good, however, hovering on the profile seems to cover a larger zone. I guess it's because the new header for profiles is not yet introduced, but I wanted to make sure this will later be fixed.
- Could you please change the tooltips to highlight the problem first "Stop the instance to rename". Using the Vanilla tooltip makes sense, good catch that the other one is already in use.

al-bakalova:
@edlerd Looks good to me, thank for applying all the changes!

lorumic:
The text in the button goes new line when processing:

image

This happens only for [an instance](https://lxd-ui-275.demos.haus/ui/default/instances/detail/a-new-instance) rename. Is this expected/desired?

src/pages/instances/forms/InstanceRename.tsx
```diff
@@ -0,0 +1,63 @@
...
+    setSubmitting(true);
+    renameInstance(instance.name, name, project)
+      .then(() => {
+        navigate(
+          `/ui/${project}/instances/detail/${name}`,
+          notify.queue(notify.success("Instance renamed."))
+        );
+        closeForm();
+      })
+      .catch((e) => notify.failure("", e, undefined, "Renaming failed"))
```
lorumic, line 35: This should be updated according to the new order of parameters for the `failure` function, as set in #272.

src/pages/profiles/forms/ProfileRename.tsx
```diff
@@ -0,0 +1,63 @@
...
+    setSubmitting(true);
+    renameProfile(profile.name, name, project)
+      .then(() => {
+        navigate(
+          `/ui/${project}/profiles/detail/${name}`,
+          notify.queue(notify.success("Profile renamed."))
+        );
+        closeForm();
+      })
+      .catch((e) => notify.failure("", e, undefined, "Renaming failed"))
```
lorumic, line 35: This should be updated according to the new order of parameters for the `failure` function, as set in #272.

src/pages/instances/forms/InstanceRename.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <>
+      <Input
+        autoFocus
+        className="u-no-margin--bottom name-input"
+        onChange={(e) => setName(e.target.value)}
+        onKeyUp={(e) => e.key === "Enter" && handleSubmit()}
+        type="text"
+        value={name}
+      />{" "}
```
lorumic, line 48: What is the purpose of the `{" "}` here?

src/pages/profiles/forms/ProfileDetailsForm.tsx
```diff
@@ -19,10 +19,13 @@ export const profileDetailPayload = (values: CreateProfileFormValues) => {
 
 interface Props {
   formik: FormikProps<CreateProfileFormValues>;
+  isEdit?: boolean;
 }
 
-const ProfileDetailsForm: FC<Props> = ({ formik }) => {
+const ProfileDetailsForm: FC<Props> = ({ formik, isEdit = true }) => {
```
lorumic, line 25: This should be `false` by default, no? Given that you're not passing it to the `CreateProfileForm`.

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+                  </Tooltip>
+                </li>
+              )}
+            </ol>
+          </nav>
+          {!isRename && (
+            <div>
+              <i className="status u-text--muted">{instance.status}</i>
+              <InstanceStateActions key="state" instance={instance} />
+            </div>
```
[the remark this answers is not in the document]
edlerd: I think it can't, it would move into the nav and into the list, where it should be out of it.

src/pages/instances/forms/InstanceRename.tsx
edlerd: right, now that it is merged, we can adjust it as well.

src/pages/instances/forms/InstanceRename.tsx
edlerd: Wired, who put the space in there? oO

src/pages/profiles/forms/ProfileDetailsForm.tsx
edlerd: true! I'll better make it non-optional to avoid confusion on it.

src/pages/instances/InstanceDetailHeader.tsx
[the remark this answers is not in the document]
lorumic: Ah, right, I missed the 2 tags closed right above it. Then it's ok as it is now.

src/pages/instances/forms/InstanceRename.tsx
lorumic: :eyes: 
Thanks for removing it from the `ProfileRename`. It's still there for `InstanceRename` though.

lorumic (approved):

PR #279, opened 2023-04-05
Disable delete project and edit project features, when the project is not empty or the default project

edlerd:
## Done

- Fixed the edit project experience to disable delete and change project features, if the project is not empty

## QA

1. Run the LXD-UI:
    - try to delete or edit a project that is empty or not empty


al-bakalova:
The suggested experience seems okay. However I struggle with the following case - I wanted to update the customised features. Everything but the Network zones was disabled, as expected. When I decided to add the network zones, I got an error about the networks. Is this right? 

Screenshot 2023-04-05 at 20 27 18

edlerd:
> The suggested experience seems okay. However I struggle with the following case - I wanted to update the customised features. Everything but the Network zones was disabled, as expected. When I decided to add the network zones, I got an error about the networks being disabled ... but they already were and the checkbox is disabled for me. Is this expected?

I think you ended up there by switching from custom to default features and back. Then the network switch is disabled. We should disable the feature switch if networks are enabled and the project is non-empty. I adjusted the code accordingly.

al-bakalova:
Reviewing the project features is a very hard task. Should we also request a review from someone on the core team?
I have a few observations but I am not sure these make sense: 

- If the user cannot change the features of the default project, let's just disable the Features dropdown for them.
Currently, I am able to spend time updating these just to get an error that I can not to so.
- Following the copy guidelines, the message should say "You cannot change the features of the default project". The tech writers advise us against using the short forms as "can't".

Screenshot 2023-04-06 at 11 10 07

- The MMMM project has a custom features config. I am still able to enter edit mode but removing networks and zones is still not possible. If it is not possible for me to select or unselect a feature, can we just disable the checkbox instead?

Screenshot 2023-04-06 at 11 13 12

edlerd:
> * If the user cannot change the features of the default project, let's just disable the Features dropdown for them.
>   Currently, I am able to spend time updating these just to get an error that I can not to so.
> * Following the copy guidelines, the message should say "You cannot change the features of the default project". The tech writers advise us against using the short forms as "can't".
> * The MMMM project has a custom features config. I am still able to enter edit mode but removing networks and zones is still not possible. If it is not possible for me to select or unselect a feature, can we just disable the checkbox instead?

@al-bakalova all this should be exactly as described (disabled fields on default project, and on the MMM project as well). Were you testing this on another demo link than the one from this PR? When I test it on this PRs link, it works exactly as you describe it should behave.

al-bakalova:
@edlerd My bad. I was probably browsing another PR. This one works as expected. Thank you

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -226,7 +226,7 @@ const ProjectDetailsForm: FC<Props> = ({ formik }) => {
                   )
                 }
                 checked={formik.values.features_networks_zones}
-                disabled={formik.values.readOnly}
+                disabled={isReadOnly || isDefaultProject}
```
lorumic, line 229: Actually, I tried disabling the network zones on a non-empty project and this is what I get:

image

src/util/projects.tsx
```diff
@@ -21,3 +23,14 @@ export const getSubpageFromUrl = (url: string) => {
...
   return undefined;
 };
+
+export const isProjectEmpty = (project: LxdProject) => {
+  if (!project.used_by) {
+    return true;
+  }
+
+  return (
+    project.used_by.filter((item) => !item.includes("/1.0/profiles/"))
```
lorumic, line 33: [A project with profiles](https://lxd-ui-279.demos.haus/ui/baz/configuration) is non-empty too, and as such, can't be deleted:

image

edlerd: So we should only exclude the default profile here, right.

src/pages/projects/forms/ProjectDetailsForm.tsx
edlerd: I amended the commit, so we disable the network zones when it is active and the project is non-empty.

lorumic (approved):

PR #280, opened 2023-04-05
Use shorter copy in the pagination on smaller screens

edlerd:
## Done

- avoid the pagination text to break into multiple lines on small screens

## QA

1. Run the LXD-UI:
    - test the pagination footer on smallish screens


src/components/Pagination.tsx
```diff
@@ -1,6 +1,9 @@
 import { Button, Icon, Input, Select } from "@canonical/react-components";
-import React, { FC } from "react";
+import React, { FC, useState } from "react";
 import { paginationOptions } from "util/pagination";
+import useEventListener from "@use-it/event-listener";
+
+const figureSmallScreen = () => window.innerWidth < 850;
```
lorumic, line 6: As you already noticed, I don't really like seeing these values hardcoded here and there. It's not very important, but if we can assign this to a `const` in a central place and reference it from there, that would make me happier.

Also, why is it 850? I still experience the 2 lines between 850 and 863 width:

image

al-bakalova: I am no longer able to reproduce this issue.

al-bakalova:
Looks good to me @edlerd 

Please keep in mind this might slightly change once we finalise our Vanilla recommendations.

src/components/Pagination.tsx
edlerd: > Also, why is it 850? I still experience the 2 lines between 850 and 863 width:

This was wired, it is also dependent on the side panel being open or not. I figured it is way safer to measure the available space for the text directly and this should resolve it.

lorumic: Ok, so to summarise, we're happy with it taking 2 lines with the side panel open? Because this is what I currently get at 1240 width as soon as I open a side panel:

image

If this is expected/intended, just let me know.

al-bakalova: I believe this is not expected. I have just checked, it does not behave that way for me. It goes straight to "50 out of 59"

al-bakalova: Screenshot 2023-04-06 at 12 28 26

lorumic: This is what I experience with Chrome version 112.0.5615.49 on Ubuntu:

Peek 2023-04-06 11-44

edlerd: Oh right, the pagination does not get updated on sidepanel open. I think we should throw a resize event when we open or close the aside. I'll add that in.

src/util/usePanelParams.tsx
```diff
@@ -22,13 +22,23 @@ type ParamMap = Record<string, string>;
 const usePanelParams = (): PanelHelper => {
   const [params, setParams] = useSearchParams();
 
+  const craftEvent = () => {
```
lorumic, line 25: Maybe call it `craftResizeEvent`?

PR #282, opened 2023-04-11
New design and features for snapshot empty state [WD-3030]

lorumic:
## Done

- Updated design of snapshot empty state
- Added "See configuration" button to the empty state
- Disabled "Create snapshot" button when snapshots are disabled in the current project

Fixes [WD-3030](https://warthogs.atlassian.net/browse/WD-3030)

## QA

1. Run the LXD-UI:
    - Open the snapshots tab of an instance with no snapshots
    - Check that the empty state looks and behaves as expected
    - Check also that the "Create snapshot" button is disabled when the snapshot feature is disabled in the current project

[WD-3030]: https://warthogs.atlassian.net/browse/WD-3030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/EmptyState.tsx
+++ b/src/components/EmptyState.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from "react";
+import React, { FC, ReactNode } from "react";
 import { Button, Col, Icon, Row } from "@canonical/react-components";
 
 interface Props {
@@ -10,6 +10,8 @@ interface Props {
   linkURL: string;
   buttonLabel: string;
   buttonAction: () => void;
+  isDisabled?: boolean;
+  extraButton?: ReactNode;
 }
 
 const EmptyState: FC<Props> = ({
@@ -21,6 +23,8 @@ const EmptyState: FC<Props> = ({
   linkURL,
   buttonLabel,
   buttonAction,
+  isDisabled = false,
+  extraButton,
 }) => {
   return (
     <Row className="empty-state">
@@ -39,10 +43,12 @@ const EmptyState: FC<Props> = ({
             <Icon className="external-link-icon" name="external-link" />
           </a>
         </p>
+        {extraButton}
         <Button
           className="empty-state-button"
           appearance="positive"
           onClick={buttonAction}
+          disabled={isDisabled}
         >
           {buttonLabel}
         </Button>
--- a/src/components/EmptyState.tsx
+++ b/src/components/EmptyState.tsx
@@ -1,17 +1,12 @@
 import React, { FC, ReactNode } from "react";
-import { Button, Col, Icon, Row } from "@canonical/react-components";
+import { Col, Icon, Row } from "@canonical/react-components";
 
 interface Props {
   iconName: string;
   iconClass: string;
   title: string;
   message: string | ReactNode;
-  linkMessage?: string;
-  linkURL?: string;
-  buttonLabel?: string;
-  buttonAction?: () => void;
-  isDisabled?: boolean;
-  customButton?: ReactNode;
+  children: ReactNode;
 }
 
 const EmptyState: FC<Props> = ({
@@ -19,12 +14,7 @@ const EmptyState: FC<Props> = ({
   iconClass,
   title,
   message,
-  linkMessage,
-  linkURL,
-  buttonLabel,
-  buttonAction,
-  isDisabled = false,
-  customButton,
+  children,
 }) => {
   return (
     <Row className="empty-state">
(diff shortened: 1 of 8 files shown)


lorumic:
> * Please not that I am not able to reproduce this case since I am currently not able to create new instances in the empty projects. Could you please provide a screenshot?

I created a [project with disabled snapshots](https://lxd-ui-282.demos.haus/ui/no-snapshots/configuration) and then an [instance](https://lxd-ui-282.demos.haus/ui/no-snapshots/instances/detail/noted-terrapin/snapshots) in it. Let me know if this allows you to reproduce the case or if I misunderstood anything.

I will implement the other changes that you asked. So, both buttons should be disabled (not hidden, right?) when the snapshots are disallowed on a project level. And then the text needs to be updated adding a line with "Snapshots are disabled for Project A."
Is my understanding correct?

src/components/EmptyState.tsx
```diff
@@ -1,15 +1,17 @@
...
-  linkURL: string;
-  buttonLabel: string;
-  buttonAction: () => void;
+  message: string | ReactNode;
+  linkMessage?: string;
+  linkURL?: string;
+  buttonLabel?: string;
+  buttonAction?: () => void;
+  isDisabled?: boolean;
+  customButton?: ReactNode;
```
edlerd, line 14: The interface of EmptyState is getting a bit chaotic. How about we just output children in it? Children can be one or multiple buttons. Then we don't need the button* isDisabled, customButton, etc props here.

@yurii-vasyliev build a nice and clean component for [anbox](https://github.com/canonical/anbox-cloud-dashboard/blob/main/static/js/components/layout/EmptyState.tsx) that might serve as inspiration how to keep the interface tidy.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -29,6 +33,29 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
...
+  const onSuccess = (message: ReactNode) => {
+    setInTabNotification(success(message));
+  };
+
+  const onFailure = (title: string, e: unknown) => {
+    setInTabNotification(failure(title, e));
+  };
+
+  const {
+    data: project,
```
edlerd, line 45: We should feed the project as a prop from the parent, that already loaded it.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -172,7 +193,10 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
                   aria-label="Search for snapshots"
                 />
               </div>
-              <ConfigureSnapshotsBtn instance={instance} />
+              <ConfigureSnapshotsBtn
+                instance={instance}
+                className="u-no-margin--right"
```
edlerd, line 198: I think we can keep this className inside the component, it won't do any harm in the empty state usage.

edlerd (changes requested):

src/pages/instances/InstanceSnapshots.tsx
lorumic: It was needed to move it out, and it still will be even after the change suggested above, in order to avoid the two buttons to be "glued together" with no margin between them. And in general, a reusable component like this one should not "force" a style (like the no-margin--right) that applies to one use case only.

src/pages/instances/InstanceSnapshots.tsx
lorumic: It doesn't seem to me that the parent (`InstanceDetail`) has loaded the project config.

edlerd (changes requested):
@al-bakalova we seem to be using the container icon in the empty state for snapshots. Should we use no icon here or a different one?

al-bakalova:
@edlerd @lorumic I thought the icon makes sense as it is a snapshot of the instance. I don't we have a better icon available. Now that you mention it, if it is misleading to both of you, we can also remove it.

lorumic:
> Now that you mention it, if it is misleading to both of you, we can also remove it.

No, not for me. I thought that changing the icon from settings to container was actually a good idea, which made totally sense for the same reason that you mentioned. Also, I think we should always keep an icon for consistency with the other empty states.

edlerd:
OK, I was just wondering. We can of course reuse the icon for the snapshots.

PR #285, opened 2023-04-11
Empty state in graphic console for a non-running VM instance [WD-2834]

lorumic:
## Done

- Added an empty state for the console tab (graphic radio selected), to be shown when the VM instance is not running

Fixes [WD-2834](https://warthogs.atlassian.net/browse/WD-2834)

## QA

1. Run the LXD-UI:
    - Open the detail page of a non-running VM instance
    - Go to the "Console" tab
    - Check that the new empty state looks and behaves as expected

[WD-2834]: https://warthogs.atlassian.net/browse/WD-2834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/InstanceTerminal.tsx
+++ b/src/pages/instances/InstanceTerminal.tsx
@@ -14,6 +14,9 @@ import NotificationRowLegacy from "components/NotificationRowLegacy";
 import { failure } from "context/notify";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import { LxdInstance } from "types/instance";
+import EmptyState from "components/EmptyState";
+import SubmitButton from "components/SubmitButton";
+import { useInstanceStart } from "util/instanceStart";
 
 const XTERM_OPTIONS = {
   theme: {
@@ -55,6 +58,8 @@ const InstanceTerminal: FC<Props> = ({ instance }) => {
   const [controlWs, setControlWs] = useState<WebSocket | null>(null);
   const [payload, setPayload] = useState(defaultPayload);
   const [fitAddon] = useState<FitAddon>(new FitAddon());
+
+  const isRunning = instance.status === "Running";
 
   const openWebsockets = async (payload: TerminalConnectPayload) => {
     if (!name) {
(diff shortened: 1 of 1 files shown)


al-bakalova:
@lorumic I seem to browsing an older version. We still treat it as an error. I have checked multiple instances, stopped and started these but I was not able to achieve the desired result. 
I did check and I seem to be browsing demo environment. Could you please check on your side?

Screenshot 2023-04-13 at 9 31 43

lorumic:
> I did check and I seem to be browsing demo environment. Could you please check on your side?

You're browsing a container. This change is about virtual machines: https://lxd-ui-285.demos.haus/ui/default/instances/detail/alex-loves-testing-instances/console

al-bakalova:
@lorumic Sorry, my bad. This looks good.

This should be a different PR, but doesn't it make sense to have a similar empty page for the text console as well?

lorumic:
> This should be a different PR, but doesn't it make sense to have a similar empty page for the text console as well?

The text console is not necessarily empty when an instance is not running. The shutdown is retained and can be displayed there for some time after an instance is stopped, see [here](https://lxd-ui-285.demos.haus/ui/default/instances/detail/A-very-very-very-very-very-very-very-very-very-very-long-demo/console) for an example.

al-bakalova:
@lorumic yes, but we are still treating this with a big error message on top where it is not really an error. I will create a separate issue for this one. Let's discuss it during the daily standup

lorumic:
> @lorumic yes, but we are still treating this with a big error message on top where it is not really an error. I will create a separate issue for this one. Let's discuss it during the daily standup

Sure, we can hide the error - even in the scope of this PR as it is a very small change.

My point there was another: that we should not hide the console itself by displaying an empty state, because we would hide some actual content that might be useful for the user.

al-bakalova:
@lorumic Yes, yes, I understand you and I agree with you.
If possible, in this case, can we turn the error notification into an info one that says:

**Instance stopped.**
Start the instance to (enable/use/write in/) the text console.

I am not a dev so I count on you to chose the best verb here :)

src/pages/instances/InstanceConsole.tsx
```diff
@@ -29,6 +45,40 @@ const InstanceConsole: FC<Props> = ({ instance }) => {
     handleFullScreen = childHandleFullScreen;
   };
 
+  const switchConsole = (isGraphic: boolean) => {
```
edlerd, line 48: The name is quite ambiguous, like what does `switchConsole(true)` mean?

src/pages/instances/InstanceConsole.tsx
```diff
@@ -29,6 +45,40 @@ const InstanceConsole: FC<Props> = ({ instance }) => {
...
+    setInTabNotification(null);
+    setGraphic(isGraphic);
+  };
+
+  const handleStart = () => {
```
edlerd, line 53: I wonder if we can share this code with the StartInstanceButton. Especially with all the isLoading/isRunning calculations and the useNotify, useQueryClient and useInstanceLoading bits - would be good to have them in one place only.

lorumic:
> (Maybe improve the error message experience though, but that seems not straightforward -- when is it an error to display?).

The logic I'm using for the graphic console is: if the instance is not running, that's the source of the error. I don't display the notification in that case, but the empty state.

src/pages/instances/InstanceConsole.tsx
lorumic: I already thought about this, and could not find a straightforward solution. Any proposal?

edlerd: Seems right in combination with the empty state, but getting a non-interactive text console with the logs -- not sure if you'd need an error or not to clarify why it is not interactive.

edlerd:
> > (Maybe improve the error message experience though, but that seems not straightforward -- when is it an error to display?).
> 
> The logic I'm using for the graphic console is: if the instance is not running, that's the source of the error. I don't display the notification in that case, but the empty state.

Seems right in combination with the empty state, but getting a non-interactive text console with the logs -- not sure if you'd need an error or not to clarify why it is not interactive.

lorumic:
> Seems right in combination with the empty state, but getting a non-interactive text console with the logs -- not sure if you'd need an error or not to clarify why it is not interactive.

You need a **notification**, not an error. That's why @al-bakalova was proposing to display an **info** notification in that case instead. I think it makes sense. I would just use a slightly different text:

**Instance not running** **
Start the instance to interact with the text console.

** to cover also the frozen/paused state

src/pages/instances/InstanceConsole.tsx
edlerd: You can directly use the `StartInstanceButton` in here and give it a single prop `btnProps: ButtonProps` that is spread into the Button. It will allow styling it to match this use cases layout.

edlerd (approved):
LGTM

PR #287, opened 2023-04-12
Add loading state and show error on configure snapshots

edlerd:
## Done

- added loading state to configure snapshots
- make error visible when snapshot config update failed

## QA

1. Run the LXD-UI:
    - update snapshot configuration from snapshot tab of instance detail

### DIFF
--- a/src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
+++ b/src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
@@ -14,6 +14,7 @@ import {
   getInstancePayload,
   InstanceEditSchema,
 } from "util/instanceEdit";
+import SubmitButton from "components/SubmitButton";
 
 interface Props {
   instance: LxdInstance;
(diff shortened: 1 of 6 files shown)


src/pages/instances/actions/snapshots/ConfigureSnapshotModal.tsx
```diff
@@ -37,12 +38,12 @@ const ConfigureSnapshotModal: FC<Props> = ({ instance, close }) => {
       updateInstance(instancePayload, project ?? "")
         .then(() => {
           notify.success("Configuration updated.");
-          close();
         })
         .catch((e: Error) => {
           notify.failure("Configuration update failed", e);
```
lorumic, line 43: I thought only instance status updates should be notified in the area between the header and the tabs. What would be the expected behaviour here @al-bakalova?

edlerd: Pushed a change, so the notification is below the tabs.

PR #288, opened 2023-04-12
Responsive snapshot list in instance detail page

edlerd:
## Done

- collapsing columns in the snapshot table for smaller screens
- always show buttons of snapshot table on smaller screens
- align content vertically in snapshot table

Fixes [WD-3031](https://warthogs.atlassian.net/browse/WD-3031)

## QA

1. Run the LXD-UI:
    - check instance snapshots on various screen sizes

[WD-3031]: https://warthogs.atlassian.net/browse/WD-3031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/InstanceSnapshots.tsx
+++ b/src/pages/instances/InstanceSnapshots.tsx
@@ -17,6 +17,10 @@ import SelectableMainTable from "components/SelectableMainTable";
 import SnapshotBulkDelete from "pages/instances/actions/snapshots/SnapshotBulkDelete";
 import ConfigureSnapshotsBtn from "pages/instances/actions/snapshots/ConfigureSnapshotsBtn";
 
+const collapsedViewMaxWidth = 1250;
+export const figureCollapsedScreen = (): boolean =>
+  window.innerWidth <= collapsedViewMaxWidth;
+
 interface Props {
   instance: LxdInstance;
 }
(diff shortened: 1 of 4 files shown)


al-bakalova:
@edlerd This looks very good. A few suggestions to make it perfect:

- The checkbox still seems to be slightly misaligned:
Screenshot 2023-04-13 at 9 04 35

- Is there a way for the first column to benefit from more space? It currently feels like there is a lot of empty space between stateful and the actions.
Screenshot 2023-04-13 at 9 06 42

- Is there a way for us to understand whether the usually is browsing on a mobile device or on a resized desktop browser?
If yes, maybe we should always display all the buttons only on mobile devices and still allow the desktop user to hover on the different rows? This might be out-of-scope, we can definitely list it as a future improvement, if possible.

edlerd:
> * The checkbox still seems to be slightly misaligned:

fixed that

> * Is there a way for the first column to benefit from more space? It currently feels like there is a lot of empty space between stateful and the actions.

Made the stateful column slightly smaller. We need to reserve space for the arrow, when sorting by it. Also, the action buttons need some space, because when they are loading, they become wider due to the spinner.

> * Is there a way for us to understand whether the usually is browsing on a mobile device or on a resized desktop browser?

Actually, there is a media query for it. It detects if the main input is a pointing device with fine resolution. I changed it, so we don't rely on screen size any more to hide buttons, but tot use that in the snapshots and instance table.

al-bakalova:
@edlerd Awesome, thanks for applying these changes. The media query for mobile devices makes me happy!

About the stateful/actions, the buttons will be replaces with icons as part of a different PR. Let's review the spacing once we have the new buttons. This is already good.

lorumic (approved):

PR #289, opened 2023-04-12
Table sorting across multiple pages, sort snapshots by recency

edlerd:
## Done

- fix instance and snapshot table sorting, when more than one page is available
- order snapshot table by recency

Fixes WD-3052

## QA

1. Run the LXD-UI:
    - sort instances and snapshots when more than one page of items exists

### DIFF
--- a/src/pages/instances/InstanceSnapshots.tsx
+++ b/src/pages/instances/InstanceSnapshots.tsx
@@ -91,12 +91,12 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
         <>
           Name
           <br />
-          Date created
+          <div className="created-header--collapsed">Date created</div>
         </>
       ) : (
         "Name"
       ),
-      sortKey: "name",
+      sortKey: isSmallScreen ? "created_at" : undefined,
       className: "name",
     },
     ...(isSmallScreen
@@ -178,7 +178,6 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
         },
       ],
       sortData: {
-        name: snapshot.name,
         created_at: snapshot.created_at,
         expires_at: snapshot.expires_at,
         stateful: snapshot.stateful,
@@ -186,7 +185,7 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
     };
   });
 
-  const pagination = usePagination(rows);
+  const pagination = usePagination(rows, "created_at", "descending");
 
   const resize = () => {
     updateTBodyHeight("snapshots-table-wrapper");
@@ -287,6 +286,9 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
             allNames={
               instance.snapshots?.map((snapshot) => snapshot.name) ?? []
             }
+            onUpdateSort={pagination.updateSort}
+            defaultSort="created_at"
+            defaultSortDirection="descending"
           />
           <Pagination
             {...pagination}
--- a/src/util/pagination.tsx
+++ b/src/util/pagination.tsx
@@ -1,6 +1,7 @@
 import { useState } from "react";
 import { usePagination as _usePagination } from "@canonical/react-components";
 import { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import { SortDirection } from "@canonical/react-components/dist/types";
 
 export const paginationOptions = [
   {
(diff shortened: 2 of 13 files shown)


al-bakalova:
@edlerd A few observations:

- Sorting the instance table by name works as expected
- Sorting by description seems unnecessary and is not working right now

- When sorting the different columns, the table "jumps" a couple of pixels. The issue is solved for Snapshot but is persistent in Instances.

- When sorting the snapshot table by name, we still sort only the list in the current page rather than the exhaustive list

edlerd:
> * Sorting by description seems unnecessary and is not working right now

Fixed sorting by description, there was missing a sort criterion for it.

> * When sorting the different columns, the table "jumps" a couple of pixels. The issue is solved for Snapshot but is persistent in Instances.

Fixed the height of the headings in the instance list.

> * When sorting the snapshot table by name, we still sort only the list in the current page rather than the exhaustive list

I think this is working as expected. The problem you might see is, we are sorting a mix of letters and numbers. So snap-100 comes before snap-50.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -287,6 +287,9 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
             allNames={
               instance.snapshots?.map((snapshot) => snapshot.name) ?? []
             }
+            onUpdateSort={pagination.updateSort}
+            defaultSort="created_at"
+            defaultSortDirection="descending"
```
lorumic, line 292: For small screens, this is currently working only when you open the Snapshots tab. Then, as soon as you alter the sorting of any column, there's no way to "restore" this default way of sorting the data. It will be either sorted by name - even though the "arrow" is placed next to "Date created", which is confusing - or, if you "reset" the sort status of this column (= no arrow visible anymore), it will be sorted by date created but ascending instead of descending:

Peek 2023-04-14 10-50

edlerd: This is a little tricky to solve. Some options that come to mind:
* Sort the collapsed column by date created instead of name
* Change which columns are collapsed, and not allow sorting by the collapsed column
* Disallow sorting the collapsed column as it is

edlerd: I applied the 1. Idea and think it is actually pretty neat.

edlerd:
> * The clickable space between the snapshots and the status (instance table)

I just found a way to avoid this. Should now be limited to the text "Status".

> * Ignoring the capitalisation (name column)

This was already working. I realize in snapshots it doesn't work nicely, because we have a mixture of text and numbers. And I think that is quite a frequent use case with automated snapshots. So I disabled sort by name, the default sort by creation date should be more useful anyways.

al-bakalova:
@edlerd Looks good to me, thanks for applying the changes. The instance table performs as expected.

However, there seem to be an issue with the Name column of the Snapshots table. I am somehow still able to click on it to remove the sorting from the other columns but then I never get an ascending or a descending arrow. The text never turns blue either. The mouse pointer stays at "type pointer" and I believe the column is not sorted either. Could you please check on your side?

edlerd:
> However, there seem to be an issue with the Name column of the Snapshots table. I am somehow still able to click on it to remove the sorting from the other columns but then I never get an ascending or a descending arrow. The text never turns blue either. The mouse pointer stays at "type pointer" and I believe the column is not sorted either. Could you please check on your side?

This is a bug in the `MainTable` in react component. I created a [PR](https://github.com/canonical/react-components/pull/913) to fix it there. We have to keep it as is in our project until that is merged.

al-bakalova:
@edlerd Got it. It makes sense! From my point of view, we can proceed with the code review. I will give it a Design +1. Please let me know when this problem is solved for us so that I can check it again :)

lorumic:
> * When sorting the different columns, the table "jumps" a couple of pixels. The issue is solved for Snapshot but is persistent in Instances.

I'm still experiencing this "jump" issue on the collapsed snapshot table:

Peek 2023-04-17 13-03

edlerd:
> I'm still experiencing this "jump" issue on the collapsed snapshot table:

Good catch on this collapsed state! I added a fix for it.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -178,15 +178,15 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
         },
       ],
       sortData: {
-        name: snapshot.name,
+        name: snapshot.name.toLowerCase(),
```
lorumic, line 181: We don't need this anymore, right? The table cannot be sorted by name.

src/util/pagination.tsx
```diff
@@ -17,18 +18,51 @@ export const paginationOptions = [
...
-): {
+export interface Pagination {
   pageData: MainTableRow[];
   currentPage: number;
   paginate: (pageNumber: number) => void;
   pageSize: number;
   setPageSize: (newSize: number) => void;
   totalPages: number;
-} => {
+  updateSort: (sort: string | null | undefined) => void;
```
lorumic, line 28: I think this can be shortened to:

```suggestion
  updateSort: (sort?: string) => void;
```

src/util/pagination.tsx
```diff
@@ -17,18 +18,51 @@ export const paginationOptions = [
...
+  data: MainTableRow[],
+  defaultSort?: string,
+  defaultSortDirection?: SortDirection
+): Pagination => {
   const [_pageSize, _setPageSize] = useState(paginationOptions[0].value);
 
+  const [sort, setSort] = useState<string | null | undefined>(defaultSort);
+  const [sortDirection, setSortDirection] = useState(
+    defaultSortDirection ?? "ascending"
+  );
```
lorumic, line 41: We can have an initializer, and get rid of the `??` operator in the state:

```suggestion
  defaultSortDirection: SortDirection = "ascending"
): Pagination => {
  const [_pageSize, _setPageSize] = useState(paginationOptions[0].value);

  const [sort, setSort] = useState<string | null | undefined>(defaultSort);
  const [sortDirection, setSortDirection] = useState(defaultSortDirection);
```

src/util/pagination.tsx
```diff
@@ -17,18 +18,51 @@ export const paginationOptions = [
...
+        return sortDirection === "ascending" ? 1 : -1;
+      }
+      if (aVal < bVal) {
+        return sortDirection === "ascending" ? -1 : 1;
+      }
+      return 0;
+    });
+  }
+
+  const updateSort = (newSort: string | null | undefined) => {
```
lorumic, line 57: See comment above, about the interface for this function:
```suggestion
  const updateSort = (newSort?: string) => {
```

src/util/pagination.tsx
edlerd: ah right, good idea.

src/util/pagination.tsx
lorumic: Wouldn't it still be compatible? I tested it locally and it seemed to work.

src/pages/instances/InstanceSnapshots.tsx
edlerd: true, good catch.

src/util/pagination.tsx
lorumic: True, with `sort?: string | null` it's already better anyway.

lorumic (approved):

PR #293, opened 2023-04-13
Add instance bulk select WD-2609

edlerd:
## Done

- add bulk select of instances

Fixes [WD-2609](https://warthogs.atlassian.net/browse/WD-2609)

## QA

1. Run the LXD-UI:
    - bulk select instances
    - start/stop/pause instances


al-bakalova:
@edlerd This looks really good. A few suggestions on how to make it perfect:
- The action buttons seem to be too far from the checkboxes. Could you please left align these? The idea is for the buttons to replace the search bar so maybe we can left align with where the search bar used to be
- To stick to the recent copy changes suggested by other UXers reviewing the functionality, when all the instances of the project are selected, please shorten the copy to "All ## instances selected. Clear selection".
- Same as in the Snapshots task - when I have filtered the list by a keyword and start selecting instances afterwards, the suggested total number should match the number of prefiltered instances. In this case, we should suggest "2 instances selected. Select all 6 instances." (these that match my search "alex")
Screenshot 2023-04-14 at 18 37 37
- Bulk actions/ confirmation - can we please add "already" to the instances that are already running with the desired status. I would like to make it more obvious why we would ignore the action. For example "30 already stopped/started/paused instances will be ignored." [See mockups](https://app.zeplin.io/project/63e128139e0eff100dad0cbd/screen/643814b20854750a5b7a7b1a), if needed
- When a single instance is selected, can we please reuse the confirmation modals from the table actions? There is no need to repeat "1 instance selected. 1 running instance will be restarted".

edlerd:
> * The action buttons seem to be too far from the checkboxes. Could you please left align these?

Aligned them in place of the search

> * To stick to the recent copy changes suggested by other UXers reviewing the functionality, when all the instances of the project are selected, please shorten the copy to "All ## instances selected. Clear selection".

Changed the copy

> * Same as in the Snapshots task - when I have filtered the list by a keyword and start selecting instances afterwards, the suggested total number should match the number of prefiltered instances. 

Changed the behaviour, when a filter is active.

> * Bulk actions/ confirmation - can we please add "already" to the instances that are already running with the desired status. I would like to make it more obvious why we would ignore the action.

Right, I added the "already" keyword for stop/pause/start, but not for the restart cases that are ignored.

> * When a single instance is selected, can we please reuse the confirmation modals from the table actions? There is no need to repeat "1 instance selected. 1 running instance will be restarted".

Having the exact same modal is a bit complicated, I removed the 1st line in case all instances have the same state (which includes obviously the 1 item selected case). Maybe that is enough to make those cases work?

al-bakalova:
@edlerd All the applied changes look good to me, good job!

A few examples of a there our copy does not seem to work as we tell the user that these instances are already having the desired state but we still ask the user if they want to proceed with the action:
Screenshot 2023-04-17 at 13 29 38

Screenshot 2023-04-17 at 13 30 03

Screenshot 2023-04-17 at 13 32 43

I almost feel like we should complicate the logic behind. In case all the selected instances are already running (stopped) and we still want them to start (stop), we should just tell the user - "The 3 selected instances are already stopped(running/paused)" so we only need a "Close/Cancel" button in the modal.

What do you think?

edlerd:
> I almost feel like we should complicate the logic behind. In case all the selected instances are already running (stopped) and we still want them to start (stop), we should just tell the user - "The 3 selected instances are already stopped(running/paused)" so we only need a "Close/Cancel" button in the modal.

Right, this is a bit wired. I changed it to adjust the modal. Maybe we should instead disable the button - or is the modal necessary to explain the context?

al-bakalova:
@edlerd I have just checked the environment.
I feel like some work can be done with the copy. How about:

-All the 3 selected instances are already running (paused/stopped). My goal here is to provide a more positive message rather than telling the user the action will be ignored. I hope this makes sense.

edlerd:
> -All the 3 selected instances are already running (paused/stopped). My goal here is to provide a more positive message rather than telling the user the action will be ignored. I hope this makes sense.

This is a bit of a rabbit hole, because it makes non-generic copy, see the two examples below, that won't work With the "All x selected instances are already y" copy.
There might be more exceptions.

Screenshot from 2023-04-17 17-16-17
Screenshot from 2023-04-17 17-15-57

al-bakalova:
@edlerd Looks like another case we need to cover. Can we use a non generic copy instead and adjust it to each case?
For paused and stopped instances we need a warning message telling the user " Already paused or stopped instances cannot be paused".

edlerd:
> @edlerd Looks like another case we need to cover. Can we use a non generic copy instead and adjust it to each case? For paused and stopped instances we need a warning message telling the user " Already paused or stopped instances cannot be paused".

How about keeping it simple and just disabling the buttons?
Then we don't even need to find copy for the modals because they won't open and we can avoid this problem.

al-bakalova:
@edlerd The suggested experience makes a lot of sense and works as expected. I am happy with this solution, good idea!

If possible, it would be nice to tell the user why buttons are disabled in the action bar. Would it be possible to introduce a message when hovering the disabled buttons ? Something like "The selected instances are already in the desired state". (Maybe there is a better copy here).

lorumic:
@al-bakalova what would be the expected border behaviour for disabled buttons? It is currently getting the .33 opacity from the disabled button, but I would expect it not to, because it is weird to see the lines split like this:

image

al-bakalova:
> @al-bakalova what would be the expected border behaviour for disabled buttons? It is currently getting the .33 opacity from the disabled button, but I would expect it not to, because it is weird to see the lines split like this:
> 
> image

I doesn't look like we have guidelines for this one, this is why I didn't comment on it. If we have control over it, I think it might look a little better if we don't add opacity to the button strokes.

al-bakalova:
@edlerd Here's the copy suggested by the content team: 

**5** instances selected:
- **1** paused instance will be started.
- **2** stopped instances will be started.
- No action for **2** instances that are already running.

Are you sure you want to start these instances?

If possible and not too complicated for you to implement, we would like to reorder the lines so that the "ignored" instances always come at the end.

al-bakalova:
@edlerd Thanks for applying these changes so quickly! 
The disabled buttons also look great!

lorumic:
> Adjusted the borders to be not affected by the disabled buttons

Thanks, but we still have the .33 opacity on the vertical borders (the separators between the buttons). Is that ok?

edlerd:
> Thanks, but we still have the .33 opacity on the vertical borders (the separators between the buttons). Is that ok?

For me the difference seems barely visible, probably not a big deal.

edlerd:
> Thanks, but we still have the .33 opacity on the vertical borders (the separators between the buttons). Is that ok?

Seems ok for me.

al-bakalova:
> > Thanks, but we still have the .33 opacity on the vertical borders (the separators between the buttons). Is that ok?
> 
> Seems ok for me.

Barely visible on a retina laptop. I think we can ignore it.

lorumic:
Quick question: should we clear selection after a bulk operation is finished? What is the expected/desired behaviour here? I see that currently the selection is kept, but how likely is that the user will perform another bulk operation on the selected instances? Or is that done only to highlight the affected instances?

image

edlerd:
> Quick question: should we clear selection after a bulk operation is finished? What is the expected/desired behaviour here? I see that currently the selection is kept, but how likely is that the user will perform another bulk operation on the selected instances? Or is that done only to highlight the affected instances?

I think we should keep the selection - we do remove items if they don't match the current filter any more though. So if you filter by "running", then select some and stop them, we will unselect them as they become stopped.

al-bakalova:
> > Quick question: should we clear selection after a bulk operation is finished? What is the expected/desired behaviour here? I see that currently the selection is kept, but how likely is that the user will perform another bulk operation on the selected instances? Or is that done only to highlight the affected instances?
> 
> I think we should keep the selection - we do remove items if they don't match the current filter any more though. So if you filter by "running", then select some and stop them, we will unselect them as they become stopped.

We currently keep the selection as there is no "notification centre" where the user can track what has just been updated. We don't show it in the notification either. I'd keep it as it is. We are likely to adjust it at a later stage

src/util/instanceBulkActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+    Frozen: "unfreeze",
+    Stopped: "start",
+  },
+  restart: {
+    Running: "restart",
+  },
+  freeze: {
+    Running: "freeze",
+  },
+  stop: {
```
lorumic, line 26: I think we should enable actions for the "freezing" state. Currently, we can restart and stop a freezing instance from the detail page, but not from here. This can be useful when instances get stuck in the freezing state, like it often happens on the demo server. The freezing instance is even escaping the "ignored" counter currently:

Screenshot from 2023-04-19 18:02:21

```suggestion
    Freezing: "restart",
    Running: "restart",
  },
  freeze: {
    Running: "freeze",
  },
  stop: {
    Freezing: "stop",
```

src/pages/instances/actions/InstanceBulkActions.tsx
```diff
@@ -0,0 +1,111 @@
...
+    updateInstanceBulkAction(actions, isForce)
+      .then(() => {
+        const summary = actionSummary(actions);
+
+        notify.success(
+          summary.map(({ count, action }) => (
+            <Fragment key={action}>
+              <b>{count}</b> {pluralizeInstance(count)} {action}.<br />
+            </Fragment>
+          ))
```
lorumic, line 43: The "actions" here will always be only one kind of action, right? Or will there be cases when we can have different actions (with different outcomes) from the same bulk operation? I could not imagine any, but here it seems like we're treating this as an array of possibly different actions. Which edge cases am I failing to imagine here?

src/pages/instances/actions/InstanceBulkActions.tsx
```diff
@@ -0,0 +1,111 @@
...
+        );
+        setForce(false);
```
lorumic, line 45: Shouldn't this go in the `finally` block?

src/pages/instances/actions/InstanceBulkAction.tsx
```diff
@@ -0,0 +1,142 @@
...
+    currentState: LxdInstanceStatus,
+    desiredAction: LxdInstanceAction
+  ) => {
+    const count = instances.filter(
+      (instance) => instance.status === currentState
+    ).length;
+
+    const status = statusLabel(currentState) ?? "";
+    const actionRaw = instanceAction(desiredAction, currentState);
+    const action = instanceActionLabel(actionRaw);
```
lorumic, line 50:
```suggestion
    const action = actionRaw ? instanceActionLabel(actionRaw) : "ignored";
```
Which also has the benefit of making the code clearer in this context. So that when a few lines below we have `if (action === "ignored")`, we clearly know where that "ignored" is coming from, without needing to switch to the context of the util function.

src/pages/instances/actions/InstanceBulkAction.tsx
```diff
@@ -0,0 +1,142 @@
...
+      confirmationExtra={confirmationExtra}
+      confirmationMessage={
+        <>
+          {selectedSummary}
+          {getLineOrder().map((state) => statusLine(state, action))}
+          <br />
+          Are you sure you want to {confirmationLabel.toLowerCase()}{" "}
+          {instanceActions(instances, action).filter(
+            (item) => item.action !== undefined
+          ).length === 1
```
lorumic, line 124:
```suggestion
          {instanceActions(instances, action).length === 1
```
In case you update the `instanceActions` function as suggested, `filter`ing here becomes not needed anymore.

src/pages/instances/InstanceList.tsx
```diff
@@ -143,6 +145,18 @@ const InstanceList: FC = () => {
...
+    const validNames = new Set(
+      filteredInstances.map((instance) => instance.name)
+    );
+    const validSelections = selectedNames.filter((name) =>
+      validNames.has(name)
+    );
+    if (validSelections.length !== selectedNames.length) {
+      setSelectedNames(validSelections);
+    }
+  }, [filteredInstances]);
```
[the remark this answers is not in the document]
edlerd: There is the state filter, so you can filter running instances, then stop them and they vanish.

src/util/instanceBulkActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+  instances: LxdInstance[],
+  desiredAction: LxdInstanceAction
+): InstanceBulkAction[] => {
+  return instances.map((instance) => {
+    return {
+      name: instance.name,
+      project: instance.project,
+      action: instanceAction(desiredAction, instance.status),
+    };
+  });
```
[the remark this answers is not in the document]
edlerd: This is pretty neat, I'll use it directly as it is.

src/util/instanceBulkActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+    paused: 0,
+    stopped: 0,
+  };
+
+  actions.map(({ action }) => {
+    const label = instanceActionLabel(action);
+    if (label && label in summary) {
+      summary[label] += 1;
+    }
+  });
```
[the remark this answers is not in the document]
edlerd: With the suggestions above it gets even simpler, because label is guaranteed to be in summary.

src/util/instanceBulkActions.tsx
edlerd: We can't restart a frozen instance. The api responds with `Instance is not running`.

As for the screenshot, I can't reproduce that any more.

src/util/instanceBulkActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  return action && action in actionToLabel ? actionToLabel[action] : "ignored";
+};
+
+export const pluralizeInstance = (count: number) => {
+  return count === 1 ? "instance" : "instances";
+};
+
+export const statusLabel = (currentState: LxdInstanceStatus) => {
+  const stateToLabel: Partial<Record<LxdInstanceStatus, string>> = {
```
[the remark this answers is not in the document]
edlerd: status should be the thing, right!

src/util/instanceBulkActions.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const actionToLabel = {
+    unfreeze: "started",
+    start: "started",
+    restart: "restarted",
+    freeze: "paused",
+    stop: "stopped",
+  };
+
+  return action && action in actionToLabel ? actionToLabel[action] : "ignored";
+};
```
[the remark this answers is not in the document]
edlerd: I simplified even more, by not allowing a null as action to come in, we don't need the ?? part.

src/pages/instances/actions/InstanceBulkActions.tsx
edlerd: Paused and stopped instances can be started
Running and paused instances can be stopped
Basically the `actionsToPerform` array with duplicate entries.

src/pages/instances/actions/InstanceBulkActions.tsx
edlerd: This is intentional, if it failed, you want to redo and then it should stay as it was.

src/pages/instances/actions/InstanceBulkAction.tsx
edlerd: Good idea, though with the not allowing undefined actions this becomes even simpler!

src/util/instanceBulkActions.tsx
lorumic: I wasn't talking about _frozen_ instances, I was talking about _freezing_ instances. They can be restarted and stopped, no?

image

As for reproducing, it's very simple. The instance stuck in the freezing state is still there: https://lxd-ui-293.demos.haus/ui/default/instances

Just select it and another one (I used the frozen one right below it for the example above) and try clicking on the allowed actions (start or stop), and you'll reproduce the issue easily. It's something that should be addressed.

src/pages/instances/actions/InstanceBulkActions.tsx
lorumic: Well, it should stay as it was if you had a dialog left open to let the user retry. But since the dialog is gone, and you reset everything in the `finally` block, you should definitely reset this prop as well. Otherwise, given that it's a shared prop between actions, you'd potentially leak a previous state to subsequent, unrelated, bulk operations.

src/pages/instances/actions/InstanceBulkActions.tsx
lorumic: We're talking about the summary here, right? Namely, the sentence that, at the of an operation, says for example:

> 2 instances stopped

So, my question is: is this an array? Can we have multiple lines out of it? Because that is what the code suggests. And I'm failing to imagine a case where you'd have multiple lines as the outcome of a single type of action.

src/util/instanceBulkActions.tsx
edlerd: Ah right, we have also the `Starting` and `Freezing` states. I added the later one to restart and stop and the starting to stop as well.

src/pages/instances/actions/InstanceBulkActions.tsx
edlerd: Ah true, user might do something else next. I'll change it.

src/pages/instances/actions/InstanceBulkActions.tsx
edlerd: Right we changed this along the way and here we can only have one action in user's peek. Because unfreeze and start are the same thing for the user here.

edlerd:
This changed quite a bit, we should test it thoroughly with all cases and variations before merging.

lorumic (approved):
I checked the code and QAed the bulk operations again on the demo server. Everything looks good and in a working condition.

PR #295, opened 2023-04-17
Snapshot actions [WD-3032]

lorumic:
## Done

- Use icons for snapshot actions - edit, restore, delete
- Allow user to opt out restoring the instance state when restoring a stateful snapshot
- Add edit feature to update the name and expiry date of a snapshot

Fixes [WD-3032](https://warthogs.atlassian.net/browse/WD-3032)

## QA

1. Run the LXD-UI:
    - Open the instance detail page, snapshots tab
    - Check the look and feel of snapshot actions (with icon buttons instead of text buttons)
    - Restore a stateful snapshot - with or without the instance state
    - Edit a snapshot (rename it)
    - Edit a snapshot (update its expiry date)
    - Edit a snapshot (rename it and update its expiry date)

[WD-3032]: https://warthogs.atlassian.net/browse/WD-3032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/snapshotEdit.tsx
```diff
@@ -0,0 +1,14 @@
+export interface CreateEditSnapshotFormValues {
```
edlerd, line 1: Can this be shorter, as `SnapshotFormValues` ?

edlerd (commented):
I think the change-version icon for "restore snapshot" is unclear. @al-bakalova 
Screenshot from 2023-04-17 18-08-27

src/pages/instances/actions/snapshots/EditSnapshotForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+const EditSnapshotForm: FC<Props> = ({
+  instance,
+  snapshot,
+  close,
+  onSuccess,
+}) => {
+  const notify = useNotify();
+  const queryClient = useQueryClient();
+  const controllerState = useState<AbortController | null>(null);
+  console.log(snapshot);
```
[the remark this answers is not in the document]
lorumic: The PR was not meant to be reviewed yet.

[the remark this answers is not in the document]
edlerd: whoops, sorry I didn't realize earlier today

al-bakalova:
@lorumic I cannot access the environment. I have been getting the Building message.

al-bakalova:
> I think the change-version icon for "restore snapshot" is unclear. @al-bakalova Screenshot from 2023-04-17 18-08-27

This is what we have in Vanilla so we should stick to it.

lorumic:
> I cannot access the environment. I have been getting the Building message.

I triggered a rebuild and it is now live.

al-bakalova:
@lorumic This looks good to me!

Checking the edit modal, I realise that the stateful checkbox is not needed. Once you take a snapshot that does not keep the state, you should not be able to request it at a later stage.

Do you want me to log this as a bug or should we fix it with this issue?

lorumic:
> Checking the edit modal, I realise that the stateful checkbox is not needed. Once you take a snapshot that does not keep the state, you should not be able to request it at a later stage.

True, we cannot edit the `stateful` property - if I understood it correctly, it seems that the only field we can change with an update is the expiry date. But I think the current UI is good - we still show the checkbox (so that the user can see if the snapshot is stateful or not), but it is always disabled. I'm not sure this is the best approach though, so if you think it shouldn't be shown (e.g. let's not show anything that we cannot edit in the edit modal), I will always hide it in edit mode. Just let me know.

Also, about [this ticket](https://warthogs.atlassian.net/browse/WD-3269) - my bad, I have inopportunely modified some code. I will push a fix soon.

al-bakalova:
@lorumic Thanks for providing further details. Yes, I believe not showing the stateful checkbox in the Edit mode makes more sense. I have updated the mockup accordingly.

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
```diff
@@ -34,68 +28,38 @@ const CreateSnapshotForm: FC<Props> = ({ instance, close, onSuccess }) => {
...
-  };
-
-  const handleEscKey = (e: KeyboardEvent<HTMLElement>) => {
-    if (e.key === "Escape") {
-      close();
-    }
-  };
-
-  const SnapshotSchema = Yup.object().shape({
+  const SnapshotSchema: unknown = Yup.object().shape({
```
edlerd, line 31: I see you put some of the validations in the snapshot util. Can we put the whole SnapshotSchema into a shared util and reuse it for create/edit?

src/util/snapshots.tsx
```diff
@@ -0,0 +1,78 @@
...
+export interface SnapshotFormValues {
+  name: string;
+  stateful: boolean;
+  expirationDate: string | null;
+  expirationTime: string | null;
+}
+
+export const getExpiresAt = (
+  expirationDate: string,
+  expirationTime: string | null
```
edlerd, line 15: All callers of this function make the expiration time mandatory. We should either remove the null here, or allow the callers to pass in a data string and a null time.

src/util/snapshots.tsx
```diff
@@ -0,0 +1,78 @@
...
+          instance.project,
+          controllerState,
+          `instances/${instance.name}/snapshots`
+        )
+      );
+    },
+  ];
+};
+
+export const testForbiddenChars = (): [
```
edlerd, line 46: A whitelist might be better, I think we only allow a-z, 0-9 and -.

lorumic: A place where I could check this would help. The `.` is also allowed, so I need to know exactly what characters to allow in order to make a whitelist.

src/api/snapshots.tsx
```diff
@@ -76,13 +76,15 @@ export const deleteSnapshotBulk = (
 
 export const restoreSnapshot = (
   instance: LxdInstance,
-  snapshot: LxdSnapshot
+  snapshot: LxdSnapshot,
+  stateful: boolean
```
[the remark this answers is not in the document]
lorumic: Good catch! This was causing the restore of stateless snapshots to fail.

src/components/ConfirmationButton.tsx
```diff
@@ -15,6 +15,7 @@ interface Props {
   confirmationExtra?: ReactNode;
   confirmationMessage: string | ReactNode;
   confirmButtonLabel: string;
+  confirmButtonAppearance?: string;
```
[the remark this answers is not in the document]
lorumic: Yes, I suggested the same, at least for the `-Button` props, in [this comment](https://github.com/canonical/lxd-ui/pull/293#discussion_r1171297171). I would keep the `-Message`, `-Extra`, etc. as they are now, and refactor all of them together in a future PR.

src/util/snapshots.tsx
lorumic: Right, I forgot to update this after making the `expirationTime` mandatory.

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
lorumic: I also thought that, but couldn't find a way to do it easily. Currently we're referencing the `formik` object inside the schema itself, and we're allowed to do that only because we're in the same scope where it gets defined. If you move the `SnapshotSchema` to a shared util, you'd need to pass the `formik` object to it somehow, right? And that is not allowed because it's not defined yet. As an extreme countermeasure I even tried to store the `formik` object in the state, but that triggers a chain of unpleasant effects that make the code super messy, and eventually it wasn't working anyway because it was creating the "render loop" error. So I gave up eventually.

edlerd: We shouldn't use formik in the schema. I think it is only used to access other fields. We can do that [without the formik object](https://stackoverflow.com/questions/63058945/get-the-value-of-another-field-for-validation-in-yup-schema).

src/util/snapshots.tsx
edlerd: All I found was [this list](https://github.com/lxc/lxd/blob/086f33b9b0f198eb91e551a75d00d1fc9bfe5650/lxd/instance/instance_utils.go#L705), which is exactly like yours. When testing also other special chars were rejected, so there must be other places validating the name.

lorumic: Yeah, I also tested with different characters yesterday. What I noticed is that the error messages are often different, meaning that they're probably validated in different places. I propose to note this down for a future improvement, and check with the core team if they can provide us with an exhaustive list of allowed characters.

src/pages/instances/actions/snapshots/CreateSnapshotForm.tsx
lorumic: This is exactly what I needed!

PR #297, opened 2023-04-18
Add tests for project, profile and instance forms and rename WD-2926

edlerd:
## Done

- added tests for the profile forms
- added tests for the instance forms
- added tests for the project forms
- added tests for the profile rename
- added tests for the instance rename

Fixes [WD-2926](https://warthogs.atlassian.net/browse/WD-2926)

## QA

1. Run the LXD-UI:
    - validate test suite is passing

[WD-2926]: https://warthogs.atlassian.net/browse/WD-2926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lorumic (commented):
I wasn't able to QA yet because the tests are mysteriously failing on my machine and I couldn't find a reason. I will keep on trying/investigating. In the meanwhile, I have one important, high-level comment/request: let's exclude operations that are not relevant for the single test that one tries to run from the global setup. For example, at the moment, we always create an instance, even for tests where we don't actually need it (e.g. profile or project operations).

tests/profiles.spec.ts
```diff
@@ -1,23 +1,147 @@
...
+  await page.getByText("Cloud init").click();
+
+  await setCodeInput(page, "Network config", "foo:\n" + " - abc");
+  await setCodeInput(page, "User data", "bar:\n" + " - def");
+  await setCodeInput(page, "Vendor data", "baz:\n" + " - ghi");
+  await saveProfile(page);
+
+  await assertCode(page, "foo:");
+  await assertCode(page, "bar:");
+  await assertCode(page, "baz:");
```
lorumic, line 143: This test is failing because it's trying to write that `foo` text in the network config field before it's loaded - see the test video, around 0:18:

[3ce50750618b5f9e510f0234fe8efc2bac6262dc.webm](https://user-images.githubusercontent.com/56583786/233929638-560bdb05-4b3f-489a-9cbe-69bde49bcda8.webm)

edlerd: Yes there was some logic to figure out when the editor is finished loading based on timeouts. I'll better wait for the class of the editor to appear.

tests/projects.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+async function createProject(page: Page, project: string) {
+  await page.goto("/ui/");
+  await page.getByRole("button", { name: "default" }).click();
+  await page.getByRole("button", { name: "Create project" }).click();
+  await page.getByPlaceholder("Enter name").click();
+  await page.getByPlaceholder("Enter name").fill(project);
+  await page.getByRole("button", { name: "Create" }).click();
+  await page.waitForSelector(`text=Project ${project} created.`, TIMEOUT);
+}
```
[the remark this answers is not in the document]
edlerd: Right, i threw some code to ensure the navigation is loaded.

playwright.config.ts
```diff
@@ -54,13 +54,6 @@ const config: PlaywrightTestConfig = {
         ...devices["Desktop Firefox"],
       },
     },
-
-    {
-      name: "webkit",
-      use: {
-        ...devices["Desktop Safari"],
-      },
-    },
```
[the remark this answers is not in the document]
edlerd: It was always the slowest of all browsers. I removed it in the development phase. Though now we can add it back.

edlerd:
> I wasn't able to QA yet because the tests are mysteriously failing on my machine and I couldn't find a reason. I will keep on trying/investigating. In the meanwhile, I have one important, high-level comment/request: let's exclude operations that are not relevant for the single test that one tries to run from the global setup. For example, at the moment, we always create an instance, even for tests where we don't actually need it (e.g. profile or project operations).

I couldn't figure out how to do that. Because the setup of the global test instance is in the gloablSetup and there we have no context over which tests are run. Also we don't have a signal which test needs the gloabal instance and which doesn't.

playwright.config.ts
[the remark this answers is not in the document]
lorumic: Yes, I think it's reasonable to do so. We can always reintroduce it later if it's really needed, but for now it's probably better to do without it.

lorumic:
> I couldn't figure out how to do that. Because the setup of the global test instance is in the globalSetup and there we have no context over which tests are run. And we don't have a signal which tests need the global instance and which don't. We always run the full test suite, so it would just add complexity, and not help the suite to run more stable or faster.

It would help the single test (or batch of tests in a single file) to run more stable and faster though. Or do you expect to run the full test suite even when the number of tests will have increased so much that they're in the order of hundreds/thousands?
The solution that I proposed is fairly simple and does not add complexity, it removes it. No instance creation should happen in the `globalSetup`, because the `globalSetup` is for things affecting all tests and not all tests really need an instance creation. Rather, once the features in the app increase, an always smaller percentage of the tests will require an instance creation. And since an instance creation requires around 10-30 seconds to complete, you're adding a 30 seconds overhead to many tests for something that is useless for that single test when you run it alone or in its batch of tests that are unrelated to instances.
We can skip this for now, since the number of tests is still limited, but keeping in mind that we're doing things wrong there, and that it should be fixed appropriately sooner or later.

edlerd:
> > I couldn't figure out how to do that. Because the setup of the global test instance is in the globalSetup and there we have no context over which tests are run. And we don't have a signal which tests need the global instance and which don't. We always run the full test suite, so it would just add complexity, and not help the suite to run more stable or faster.
> 
> It would help the single test (or batch of tests in a single file) to run more stable and faster though. Or do you expect to run the full test suite even when the number of tests will have increased so much that they're in the order of hundreds/thousands? The solution that I proposed is fairly simple and does not add complexity, it removes it. No instance creation should happen in the `globalSetup`, because the `globalSetup` is for things affecting all tests and not all tests really need an instance creation. Rather, once the features in the app increase, an always smaller percentage of the tests will require an instance creation. And since an instance creation requires around 10-30 seconds to complete, you're adding a 30 seconds overhead to many tests for something that is useless for that single test when you run it alone or in its batch of tests that are unrelated to instances. We can skip this for now, since the number of tests is still limited, but keeping in mind that we're doing things wrong there, and that it should be fixed appropriately sooner or later.

I removed the globally shared instance. We were only using it for snapshots tests, and those create their own random instance now.

lorumic:
What's happening here? (especially at the end)

[4f1b5c791ba8b9c4b9a8d205c5336de3d4acfa46.webm](https://user-images.githubusercontent.com/56583786/234242186-a4e8d172-5f9d-4f5e-92ab-be8881a5d1ae.webm)

lorumic:
> What's happening here? (especially at the end)

And also here, for profile edit:

[f45e08cf2e3d6f38262af92e27d4334163589750.webm](https://user-images.githubusercontent.com/56583786/234243379-dea3b6b6-7952-4482-8001-e13a241d534b.webm)

edlerd:
There was a delay before deleting instances which I removed, also added a visit profile/instance function. The suite runs now a bit quicker.

tests/profiles.spec.ts
```diff
@@ -37,9 +156,7 @@ async function createProfile(page: Page, profile: string) {
 }
 
 async function deleteProfile(page: Page, profile: string) {
```
lorumic, line 158: Shouldn't this and all the functions below be moved to a `profile-helpers` file, just like for the instances?

tests/profiles.spec.ts
```diff
@@ -48,38 +165,30 @@ async function deleteProfile(page: Page, profile: string) {
...
+  await page.goto("/ui/");
+  await page.getByRole("link", { name: "Profiles" }).click();
+  await page.getByPlaceholder("Search").click();
+  await page.getByPlaceholder("Search").fill(profile);
+  await page.getByRole("link", { name: profile }).first().click();
+}
+
+async function renameProfile(page: Page, old: string, newName: string) {
+  await visitProfile(page, old);
+  await page.getByRole("listitem", { name: old }).getByText(old).click();
```
lorumic, line 178: Maybe `oldName` instead of just `old`?

tests/instances.spec.ts
```diff
@@ -3,10 +3,117 @@ import {
...
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+
+  const newName = instance + "-rename";
+  await renameInstance(page, instance, newName);
+
+  await deleteInstance(page, newName);
+});
+
+test("instance configuration edit", async ({ page }) => {
```
lorumic, line 33: This is quite big, doing all the configs in just one test, I wonder if we should split it like for profiles?

tests/instances.spec.ts
```diff
@@ -3,10 +3,117 @@ import {
...
+  await page.getByPlaceholder("Search").click();
+  await page.getByPlaceholder("Search").fill(instance);
+  await page.getByRole("link", { name: instance }).first().click();
+  await page.getByTestId("tab-link-Configuration").click();
+  await page.getByRole("button", { name: "Edit instance" }).click();
+
+  await page.getByText("Resource limits").click();
+  await setOption(page, "Memory swap", "true");
+  await setOption(page, "Disk priority", "1");
+  await setInput(page, "Max number of processes", "Enter number", "2");
```
lorumic, line 46: This is just a subset, cpu and memory limits are not tested?

tests/instances.spec.ts
```diff
@@ -3,10 +3,117 @@ import {
...
+  await setOption(
+    page,
+    "Prevent instance file system from being UID/GID shifted",
+    "true"
+  );
+  await setInput(page, "Base host id", "Enter ID", "11");
+  await setInput(page, "Idmap size", "Enter number", "22");
+  await setOption(page, "Unique idmap usage", "true");
+  await setOption(page, "Allow /dev/lxd in the instance", "true");
+  await setOption(page, "Make /1.0/images API available", "true");
```
lorumic, line 60: Maybe we should add another test where a VM is created, so that it allows to test also the missing config here (which is for VMs only)?

tests/instances.spec.ts
```diff
@@ -3,10 +3,117 @@ import {
...
+
+  await page
+    .getByRole("tabpanel", { name: "Configuration" })
+    .getByText("Snapshots")
+    .click();
+  await setInput(page, "Snapshot name", "Enter name pattern", "snap123");
+  await setInput(page, "Expire after", "Enter expiry expression", "3m");
+  await setOption(page, "Snapshot stopped instances", "true");
+
```
lorumic, line 69: Missing the "Schedule" field here as well.

tests/configuration-helpers.ts
```diff
@@ -0,0 +1,113 @@
...
+    type === "fixed"
+      ? "Comma-separated core numbers"
+      : "Number of exposed cores";
+  await page.getByPlaceholder(placeholder).click();
+  await page.getByPlaceholder(placeholder).press("Control+a");
+  await page.getByPlaceholder(placeholder).fill(limit);
+  await page.getByRole("button", { name: "Save changes" }).click();
+  await page.getByTestId("tab-link-Configuration").click();
+  await page.getByText("Resource limits").click();
+  await page.getByRole("gridcell", { name: `Exposed CPUs ${limit}` }).click();
```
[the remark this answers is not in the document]
edlerd: this code is older than the assertReadMode function, but you are right, we should use it here!

tests/instances.spec.ts
edlerd: VMs are much slower to create, so I avoided adding tests for them.

lorumic: A VM for alpine should be reasonably fast to create, no? At least so it seems on my (extremely slow) local environment...

tests/profiles.spec.ts
```diff
@@ -1,23 +1,142 @@
...
+test("profile snapshots", async ({ page }) => {
+  const profile = randomProfileName();
+  await createProfile(page, profile);
+  await editProfile(page, profile);
+  await page.getByText("Snapshots").click();
+
+  await setInput(page, "Snapshot name", "Enter name pattern", "snap123");
+  await setInput(page, "Expire after", "Enter expiry expression", "3m");
+  await setOption(page, "Snapshot stopped instances", "true");
+  await saveProfile(page);
```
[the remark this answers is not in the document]
edlerd: It is a bit complicated because the Schedule keyword is already used in the help text for the Snapshot name pattern. I'll add a custom function for the schedule.

edlerd:
> * Basic details (description and, for instances, applied profiles) are not tested
> * Storage config is not tested
> * Networks config is not tested
> * YAML config is not tested

Covered the basic case. I'd postpone network and storage, because we can't reliably ensure a given storage or network is present without having created them. And both features are probably changing a lot shortly. So after they are stable we should add tests for using them in instances and profiles as well.

lorumic (commented):
I think a test for the YAML configuration could be added, no? Even if just to update something really simple - the description, for example - and prove that the save works for that as well.

edlerd:
> I think a test for the YAML configuration could be added, no? Even if just to update something really simple - the description, for example - and prove that the save works for that as well.

Good idea. Actually that is broken atm, because we don't refresh the form state after submitting the yaml changes.

tests/projects.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+  await createProject(page, project);
+
+  await page.getByRole("link", { name: "Configuration" }).click();
+  await page.getByRole("button", { name: "Edit configuration" }).click();
+  await page.getByPlaceholder("Enter description").fill("desc");
+  await page
+    .getByRole("combobox", { name: "Features" })
+    .selectOption("customised");
+  await page.locator("span").filter({ hasText: "Networks" }).click();
+  await page.locator("label").filter({ hasText: "Network zones" }).click();
```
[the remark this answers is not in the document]
lorumic: Also the value in the description field is not checked.

lorumic:
> Good idea. Actually that is broken atm, because we don't refresh the form state after submitting the yaml changes.

I see, we can then skip it for now and then add a dedicate test for the YAML configuration once the issue gets fixed.

lorumic (approved):

PR #298, opened 2023-04-19
Profile list overhaul [WD-2673]

lorumic:
## Done

- Updated profile list page to match the look and feel of the instance list page
- Handled the peculiar UI differences for the default project and for projects with inherited profiles

Fixes [WD-2673](https://warthogs.atlassian.net/browse/WD-2673)

## QA

1. Run the LXD-UI:
    - Check the profile list page for the default project
    - Check the profile list page for a project with profiles enabled
    - Check the profile list page for a project with profiles disabled

[WD-2673]: https://warthogs.atlassian.net/browse/WD-2673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/profiles/ProfileList.tsx
+++ b/src/pages/profiles/ProfileList.tsx
@@ -209,7 +209,7 @@ const ProfileList: FC = () => {
           <NotificationRow />
           <Row className="no-grid-gap">
             <Col size={12}>
-              {!featuresProfiles && (
+              {!isLoading && !featuresProfiles && (
                 <Notification severity="caution" title="Profiles disabled">
                   The feature has been disabled on a project level. All the
                   available profiles are inherited from the default project.
--- a/src/pages/profiles/ProfileList.tsx
+++ b/src/pages/profiles/ProfileList.tsx
@@ -1,8 +1,13 @@
-import React, { FC } from "react";
-import { MainTable, Row, Button } from "@canonical/react-components";
-import NotificationRow from "components/NotificationRow";
+import React, { FC, useEffect, useState } from "react";
+import {
+  MainTable,
+  Row,
+  Button,
+  SearchBox,
+  Col,
+} from "@canonical/react-components";
 import { fetchProfiles } from "api/profiles";
-import BaseLayout from "components/BaseLayout";
+import classnames from "classnames";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { useNotify } from "context/notify";
@@ -10,15 +15,24 @@ import Loader from "components/Loader";
 import { Link, useNavigate, useParams } from "react-router-dom";
 import ItemName from "components/ItemName";
 import { getProfileInstances } from "util/usedBy";
+import usePanelParams from "util/usePanelParams";
+import { usePagination } from "util/pagination";
+import useEventListener from "@use-it/event-listener";
+import { updateTBodyHeight } from "util/updateTBodyHeight";
+import Pagination from "components/Pagination";
+import NotificationRow from "components/NotificationRow";
 
 const ProfileList: FC = () => {
   const navigate = useNavigate();
   const notify = useNotify();
+  const panelParams = usePanelParams();
   const { project } = useParams<{ project: string }>();
+  const [query, setQuery] = useState<string>("");
 
   if (!project) {
     return <>Missing project</>;
   }
+  const isDefaultProject = project === "default";
 
   const {
     data: profiles = [],
@@ -33,15 +47,26 @@ const ProfileList: FC = () => {
     notify.failure("Loading profiles failed", error);
   }
 
+  const filteredProfiles = profiles.filter((item) => {
+    if (query) {
+      const q = query.toLowerCase();
+      if (
+        !item.name.toLowerCase().includes(q) &&
+        !item.description.toLowerCase().includes(q)
+      ) {
+        return false;
+      }
+    }
+    return true;
+  });
+
   const headers = [
     { content: "Name", sortKey: "name" },
     { content: "Description", sortKey: "description" },
     { content: "Used by", sortKey: "used_by", className: "u-align--right" },
   ];
 
-  const rows = profiles.map((profile) => {
-    const isDefaultProject = project === "default";
-
+  const rows = filteredProfiles.map((profile) => {
     return {
       columns: [
         {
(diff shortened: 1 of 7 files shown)


src/pages/profiles/ProfileList.tsx
```diff
@@ -1,46 +1,119 @@
...
   const headers = [
     { content: "Name", sortKey: "name" },
     { content: "Description", sortKey: "description" },
-    { content: "Used by", sortKey: "used_by", className: "u-align--right" },
+    {
+      content: isDefaultProject ? (
+        <>
+          Total instances
+          <br />
+          <div className="header-second-row">Used by</div>
```
edlerd, line 103: The split heading is confusing, can we change it to just say "Used by" and clarify the two numbers in the cells?

src/pages/profiles/ProfileList.tsx
```diff
@@ -56,62 +129,122 @@ const ProfileList: FC = () => {
...
-          ).length,
+          content: (
+            <>
+              {isDefaultProject && (
+                <>
+                  {total} in all projects
+                  <br />
+                </>
+              )}
+              {usedBy} instances
```
edlerd, line 149: this need pluralization for 1 instance case.

src/pages/profiles/ProfileList.tsx
```diff
@@ -56,62 +129,122 @@ const ProfileList: FC = () => {
...
-          content: getProfileInstances(
-            project,
-            isDefaultProject,
-            profile.used_by
-          ).length,
+          content: (
+            <>
+              {isDefaultProject && (
+                <>
+                  {total} in all projects
```
edlerd, line 145: Can we hide this if the usedBy matches the total?

al-bakalova:
@lorumic This looks great! I have a couple of suggestions on how to make it perfect:
- The cell content of the combined lines - the first line is supposed to be dark (#111) like the rest of the content. Only the additional line (used by total) should be light-dark (#757575)

- The project label (side nav) and the table headers are not aligned.
Screenshot 2023-04-19 at 12 12 45

- Here's how it looks like when resized:
Screenshot 2023-04-19 at 12 14 13

- When inherited, the notification is available. The columns should be aligned with the Instances (side nav)
Screenshot 2023-04-19 at 12 18 17

src/pages/profiles/ProfileList.tsx
```diff
@@ -56,62 +129,122 @@ const ProfileList: FC = () => {
...
-        controls={
+    <main className="l-main profile-list">
+      <div
+        className={classnames("p-panel", {
+          "has-side-panel": !!panelParams.profile,
+        })}
+      >
+        <div className="p-panel__header profile-list-header">
+          <h1 className="p-heading--4 u-no-margin--bottom">Profiles</h1>
+          <SearchBox
```
edlerd, line 187: Should the search be left aligned like in the instance list?
The small input looks a little lost on large widths atm.

src/sass/_instance_list.scss
```diff
@@ -213,3 +205,11 @@
     }
   }
 }
+
+.table-column-select-list {
```
edlerd, line 209: Why move it out of the instance list selector?

src/sass/_profile_list.scss
```diff
@@ -0,0 +1,104 @@
...
+      max-width: 20rem;
+      width: 100%;
+    }
+
+    .margin-right {
+      margin-right: $sph--large;
+    }
+
+    .profiles-disabled {
+      visibility: hidden;
```
edlerd, line 17: Can we not render those items, instead of hiding it with css?

src/pages/profiles/ProfileList.tsx
edlerd: Maybe it should instead say "x instances in other projects" on a 2nd line?

src/pages/profiles/ProfileList.tsx
lorumic: This is a question for @al-bakalova

src/pages/profiles/ProfileList.tsx
lorumic: This is a question for @al-bakalova

src/pages/profiles/ProfileList.tsx
lorumic: This is a question for @al-bakalova - currently in the design it doesn't seem to be left-aligned like in the instance list.

src/sass/_instance_list.scss
lorumic: Try moving it back in and you'll get this:

image

src/sass/_profile_list.scss
lorumic: Not with the current implementation. Because with `space-between`, the search box would stick to the right edge.

src/pages/profiles/ProfileList.tsx
al-bakalova: > The split heading is confusing, can we change it to just say "Used by" and clarify the two numbers in the cells?

Once we switch these, it is getting confused, I agree. What if we keep only "Used by" and the have:
10 instances
100 in all projects 

But then we sort by the number of instances in the current project?

lorumic: Sure, it works for me.

src/pages/profiles/ProfileList.tsx
al-bakalova: Good catch!
We should probably left align it to match with the Instance table and that way it won't jump when we switch the pages.

lorumic: Ok good, I will update it accordingly. And this will also solve the issue mentioned in [this other comment](https://github.com/canonical/lxd-ui/pull/298#discussion_r1171084327).

src/pages/profiles/ProfileList.tsx
al-bakalova: I am not sure I get the question. I'd prefer to always keep both lines and never hide (for consistency reasons).

10 instances
10 in all projects

It kind of currently repeats the total but gives the user an idea that the profiles of the default project are actually used across different projects.

lorumic:
All the points mentioned above should have been addressed @al-bakalova @edlerd

src/pages/profiles/ProfileList.tsx
```diff
@@ -1,46 +1,111 @@
...
     error,
-    isLoading,
+    isLoading: isProfilesLoading,
   } = useQuery({
     queryKey: [queryKeys.profiles, project],
     queryFn: () => fetchProfiles(project),
   });
 
+  const {
+    data: projectObj,
```
edlerd, line 49: I'm not a fan of the `*Obj` suffix. Maybe going forward, we can start aliasing the `project` from useParams to `projectName`, and have the project variable name free for the actual object.

Anyhow, not a big issue, feel free to keep it as is for now.

edlerd (approved):
It looks very good now, I like it!

Left some code ideas, one tiny fix and one idea to improve the notification.

al-bakalova:

src/sass/_profile_list.scss
```diff
@@ -0,0 +1,105 @@
...
+    tbody {
+      height: calc(100vh - 323px);
+    }
+
+    thead tr,
+    tbody tr {
+      display: table;
+      table-layout: fixed;
+      width: 100%;
+    }
```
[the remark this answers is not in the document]
lorumic: As discussed during the sync, let's keep this noted down for a future improvement :)

PR #299, opened 2023-04-19
Rename header component and rename project

edlerd:
## Done

- factored out a common RenameHeader component
- added option to rename the project

Fixes [WD-3352](https://warthogs.atlassian.net/browse/WD-3352)

## QA

1. Run the LXD-UI:
    - Check the headers for instances detail, profile detail and project configuration while or before and after renaming.

[WD-3352]: https://warthogs.atlassian.net/browse/WD-3352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


al-bakalova:
@edlerd I have just reviewed and tested the functionality. It all looks good to me. Thank you!

src/components/RenameHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    onChange={(e) => setNewName(e.target.value)}
+                    onKeyUp={(e) => e.key === "Enter" && handleRename()}
+                    type="text"
+                    value={newName}
+                  />
+                  <Button
+                    appearance="base"
+                    className="cancel"
+                    dense
+                    onClick={handleRename}
```
lorumic, line 62:
```suggestion
                    onClick={() => setRename(false)}
```

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -13,61 +14,56 @@ interface Props {
...
-      )}
-      <div className="p-panel__controls">
-        {instance && !isRename && <DeleteInstanceBtn instance={instance} />}
-      </div>
-    </div>
+    <RenameHeader
+      name={name}
+      parentItem={<Link to={`/ui/${project}/instances`}>Instances</Link>}
+      renameDisabledReason={
+        !instance || instance.status !== "Stopped"
```
lorumic, line 48: Shouldn't this be:
```suggestion
        instance?.status !== "Stopped"
```
?

src/components/RenameHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+              {isRename ? (
+                <li className="p-breadcrumbs__item rename">
+                  <Input
+                    autoFocus
+                    className="u-no-margin--bottom name-input"
+                    onChange={(e) => setNewName(e.target.value)}
+                    onKeyUp={(e) => e.key === "Enter" && handleRename()}
+                    type="text"
+                    value={newName}
+                  />
```
lorumic, line 57: I wonder if we can somehow plug in the `checkDuplicate` logic here. Because we're currently missing it for the rename, so we fall back to the default "postponed feedback".
One idea that I had is to wrap the `checkDuplicate` logic inside a function in the parent, and then pass it to this FC as a prop. Then you can call it inside `onChange` or `onKeyDown`.

lorumic (changes requested):

src/components/RenameHeader.tsx
edlerd: I think for validation it is better to use formik, so I rewrote the bit to have its own form with schema etc.
Also deduplicating now.

src/pages/instances/InstanceDetailHeader.tsx
lorumic: Why was this first committed and then reverted? :thinking:

edlerd: The change should be represented, I force pushed it with the other changes.

src/sass/_rename_header.scss
lorumic, line 1: We have again the problem of the submit button going new line while processing:

image

Window width: 1243px

src/pages/profiles/ProfileDetailHeader.tsx
```diff
@@ -18,61 +22,71 @@ const ProfileDetailHeader: FC<Props> = ({
...
-  const canRename = profile && profile.name !== "default";
+  const navigate = useNavigate();
+  const notify = useNotify();
+  const controllerState = useState<AbortController | null>(null);
+
+  const RenameSchema = Yup.object().shape({
+    name: Yup.string()
+      .test(
+        "deduplicate",
+        "An profile with this name already exists",
```
lorumic, line 33:
```suggestion
        "A profile with this name already exists",
```

src/components/RenameHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    onKeyUp={(e) =>
+                      e.key === "Enter" && void formik.submitForm()
+                    }
+                    type="text"
+                  />
+                  <div>
+                    <Button
+                      appearance="base"
+                      className="cancel"
+                      onClick={() => formik.setFieldValue("isRenaming", false)}
```
lorumic, line 64: Can we also reset the value in the input field when we cancel a rename?

Peek 2023-04-20 17-58

lorumic (changes requested):

src/pages/instances/InstanceDetailHeader.tsx
edlerd: Right, now it is included.

src/sass/_rename_header.scss
edlerd: Adjusted the rename to full width with no line breaks to avoid this.

PR #301, opened 2023-04-20
Ensure creation form sections are active, simplify section…

edlerd:
## Done

- fixed profile and instance creation forms: active sections should be active
- simplify state handling in the forms
- fixed memory/cpu limits in creation (should not be overridded initially)
- fixed read only output of memory limits

## QA

1. Run the LXD-UI:
    - Check profile/instance create and edit

### DIFF
--- a/src/pages/instances/CreateInstanceForm.tsx
+++ b/src/pages/instances/CreateInstanceForm.tsx
@@ -229,7 +229,7 @@ const CreateInstanceForm: FC = () => {
   };
 
   const updateSection = (newItem: string) => {
-    if (section === YAML_CONFIGURATION && newItem !== YAML_CONFIGURATION) {
+    if (Boolean(formik.values.yaml) && newItem !== YAML_CONFIGURATION) {
       void formik.setFieldValue("yaml", undefined);
     }
     setSection(newItem);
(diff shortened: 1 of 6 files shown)


lorumic (commented):
It's looking good and working as expected. I think it should have been specified that the "active section" problem was in the project configuration page only. I struggled a bit to understand why you wrote "fixed profile and instance creation forms: active sections should be active", since the active was working correctly in those pages.

At this point, I have a request. Unlike instances and profiles, the project configuration page still doesn't have the linkable sections. Could you please introduce them in the context of this PR? I think we should have a consistent behaviour, it feels weird that config sections are linkable in the profiles and instances pages only.

edlerd:
> It's looking good and working as expected. I think it should have been specified that the "active section" problem was in the project configuration page only. I struggled a bit to understand why you wrote "fixed profile and instance creation forms: active sections should be active", since the active was working correctly in those pages.
> 
> At this point, I have a request. Unlike instances and profiles, the project configuration page still doesn't have the linkable sections. Could you please introduce them in the context of this PR? I think we should have a consistent behaviour, it feels weird that config sections are linkable in the profiles and instances pages only.

It was broken in instance and profile creation, as well as project edit and project creation. I don't think we need the urls in the creation forms. WDYT?

lorumic:
> It was broken in instance and profile creation, as well as project edit and project creation. I don't think we need the urls in the creation forms. WDYT?

Right, I was mixing up creation and edit. I agree, we should not introduce the urls wherever we can have disabled sections, in order to avoid leaking links to sections that the user shouldn't be allowed to access (because they're disabled). This means that we should not have them in project edit either, I guess.

lorumic (approved):

PR #302, opened 2023-04-20
Update images

edlerd:
## Done

- changed image order to be ubuntu images from newest to oldes with lts on top
- added image update script
- updated images (in draft, because of the pending release of lunar lobster)

## QA

1. Run the LXD-UI:
    - check image list for creating instances


lorumic (commented):
```
~/dev/lxd-ui$ scripts/update-images.sh 
scripts/update-images.sh: line 3: jq: command not found
```
This is what I get if I try to run the script in my environment. Maybe we should refrain from including in the project something that is unlikely to work without additional effort?
Or is the script there for reference only? Ideally, what I would expect here would be:

- the JSON files are empty (or non-existent) in the project by default
- the script is run automatically with the `dotrun` command

Currently what we have is a script "for reference", that cannot work unless you have `jq` installed, and pre-filled JSON files.

edlerd:
> Currently what we have is a script "for reference", that cannot work unless you have `jq` installed, and pre-filled JSON files.

The script is there to have it shared to make it easier to update the images in the future. I can also remove it and just keep a local copy. Wdyt?

lorumic:
> The script is there to have it shared to make it easier to update the images in the future. I can also remove it and just keep a local copy. Wdyt?

I think having the script is good, but ideally I would like to see it run automatically at every project run, so that basically what we achieve is an automatic check of the images, and always have the most up-to-date version available in the app. Would that be a feasible and good idea?

edlerd:
> I think having the script is good, but ideally I would like to see it run automatically at every project run, so that basically what we achieve is an automatic check of the images, and always have the most up-to-date version available in the app. Would that be a feasible and good idea?

I am afraid to make the build more complex. Because we'll add the dependency to the image servers for a successful build. I think it is more reliable if we update it manually and keep track in git.
Though I'll add a check to the script to warn in case jq is not installed.

lorumic (approved):

PR #304, opened 2023-04-20
Fixed crash happening on detail page/panel for instances without networks

lorumic:
## Done

- An instance can have zero network devices. The app wasn't currently able to handle this case correctly, and was crashing when opening the side panel or the detail page of instances with no networks. This PR fixes this bug.

## QA

1. Run the LXD-UI:
    - Open the [side panel](https://lxd-ui-304.demos.haus/ui/default/instances?panel=instance-summary&instance=test-no-network&project=default) or [detail page](https://lxd-ui-304.demos.haus/ui/default/instances/detail/test-no-network) of an instance without networks.
    - Confirm that the "no networks" case is correctly handled.

### DIFF
--- a/src/pages/instances/InstanceDetailPanel.tsx
+++ b/src/pages/instances/InstanceDetailPanel.tsx
@@ -212,9 +212,8 @@ const InstanceDetailPanel: FC = () => {
                           <Link
                             to={`/ui/${instance.project}/instances/detail/${instance.name}/configuration/networks`}
                           >
-                            Configure
-                          </Link>{" "}
-                          instance networks.
+                            Configure instance networks
+                          </Link>
                         </p>
                       </td>
                     </tr>
@@ -274,9 +273,8 @@ const InstanceDetailPanel: FC = () => {
                           <Link
                             to={`/ui/${instance.project}/instances/detail/${instance.name}/snapshots`}
                           >
-                            Manage
-                          </Link>{" "}
-                          instance snapshots.
+                            Manage instance snapshots
+                          </Link>
                         </p>
                       </td>
                     </tr>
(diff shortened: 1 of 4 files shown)


edlerd (approved):
LGTM

al-bakalova:
@lorumic @edlerd This is a good catch!
If possible, I'd like to address the Networks section in the side panel the same way we do for snapshots.
A good first step would be to add "No networks found". I tried to test but I am still unsure if configuring a network is possible ... I have been getting a "_Invalid expanded devices: Failed detecting root disk device: No root device could be found_" error.

Anyway, if configuring a network for an instance that doesn't already have one is possible, I think it makes sense to add a link to it. Something like:

NETWORKS
No networks found
Configure instance networks (link to Networks in configuration).

al-bakalova:
@lorumic  
The current look and feel makes it impossible to distinguish the Networks section.
Additional spacing needs to be added before the section separator (Let's use the Profiles section as a reference)

Screenshot 2023-04-20 at 20 22 18

lorumic:
> I tried to test but I am still unsure if configuring a network is possible

Yes, sorry, it's definitely possible, but without thinking I forgot to add the root storage device to the profile assigned to that instance. And that's why you're getting that error. Now I added it and you should be able to manipulate the networks as well.

Actually, this fact made me realise that the crash happens only when nor the profile, nor the instance that is using it, has specified any device - no storage, no networks. Only in that case the `expanded_devices` config prop will be `undefined`, causing the app to crash. It's probably a very edge case then, but it can still happen I guess.

Anyway, I just pushed a fix for the design issues that you mentioned. Please have a look and let me know if it's ok now.

al-bakalova:
@lorumic Thanks for clarifying.

We have a pending issue for the Snapshots section. We are supposed to update the copy and make the whole sentence "Manage instance snapshots" clickable, as well as to remove the dot at the end.

The goal is to follow the same copy for networks. Could you please make "Configure instance networks" a link and remove the dot?

Otherwise, it looks good to me.

Thank you

lorumic:
> The goal is to follow the same copy for networks. Could you please make "Configure instance networks" a link and remove the dot?

Updated for both networks and snapshots.

al-bakalova:
@lorumic Looks good to me, thank you!

edlerd (approved):

PR #305, opened 2023-04-21
Add profile summary panel (side panel) [WD-3020]

lorumic:
## Done

- Added profile summary (side panel)

Fixes [WD-3020](https://warthogs.atlassian.net/browse/WD-3020)

## QA

1. Run the LXD-UI:
    - Open profile list page
    - Click on the description column or the used by column to open the side panel
    - Check that the panel looks and behaves as expected

[WD-3020]: https://warthogs.atlassian.net/browse/WD-3020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/profiles/ProfileDetailPanel.tsx
+++ b/src/pages/profiles/ProfileDetailPanel.tsx
@@ -1,0 +1,177 @@
+import React, { FC } from "react";
+import { Button, Icon } from "@canonical/react-components";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import Loader from "components/Loader";
+import Aside from "components/Aside";
+import usePanelParams from "util/usePanelParams";
+import { useNotify } from "context/notify";
+import { fetchProfile } from "api/profiles";
+import ProfileLink from "./ProfileLink";
+import { fetchProject } from "api/projects";
+import { isProjectWithProfiles } from "util/projects";
+import ExpandableList from "components/ExpandableList";
+import { isDiskDevice, isNicDevice } from "util/devices";
+import { getProfileInstances } from "util/usedBy";
+import ProfileUsedByProject from "./ProfileUsedByProject";
+
+const ProfileDetailPanel: FC = () => {
+  const notify = useNotify();
+  const panelParams = usePanelParams();
+
+  const profileName = panelParams.profile;
+  const projectName = panelParams.project;
+
+  const {
+    data: profile,
+    error,
+    isLoading: isProfileLoading,
+  } = useQuery({
+    queryKey: [queryKeys.profiles, profileName, projectName],
+    queryFn: () => fetchProfile(profileName ?? "", projectName),
+    enabled: profileName !== null,
+  });
+
+  const {
+    data: project,
+    error: projectError,
+    isLoading: isProjectLoading,
+  } = useQuery({
+    queryKey: [queryKeys.projects, projectName],
+    queryFn: () => fetchProject(projectName),
+  });
+
+  if (error) {
+    notify.failure("Loading profile failed", error);
+  }
+
+  if (projectError) {
+    notify.failure("Loading project failed", error);
+  }
+  const isLoading = isProfileLoading || isProjectLoading;
+
+  const featuresProfiles = isProjectWithProfiles(project);
+
+  const isDefaultProject = projectName === "default";
+  const usageCount = getProfileInstances(
+    projectName,
+    isDefaultProject,
+    profile?.used_by
+  ).length;
+
+  return (
+    <Aside width="narrow" pinned className="u-hide--medium u-hide--small">
+      {isLoading && <Loader />}
+      {!isLoading && !profile && <>Loading profile failed</>}
+      {profile && (
+        <div className="p-panel profile-detail-panel">
+          <div className="p-panel__header">
+            <h2 className="p-panel__title">Profile summary</h2>
+            <div className="p-panel__controls">
+              <Button
+                appearance="base"
+                className="u-no-margin--bottom"
+                hasIcon
+                onClick={panelParams.clear}
+                aria-label="Close"
+              >
+                <Icon name="close" />
+              </Button>
+            </div>
+          </div>
+          <div className="p-panel__content panel-content">
+            <div className="content-scroll">
+              <table className="u-table-layout--auto u-no-margin--bottom">
+                <tbody>
+                  <tr>
+                    <th className="u-text--muted">Name</th>
+                    <td>
+                      <ProfileLink
+                        profile={{ name: profile.name, project: projectName }}
+                      />
+                    </td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Description</th>
+                    <td>{profile.description ? profile.description : "-"}</td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted last-of-section">
+                      Defined in
+                    </th>
+                    <td>{featuresProfiles ? "Current" : "Default"} project</td>
+                  </tr>
+                  <tr className="u-no-border">
+                    <th colSpan={2}>
+                      <h3 className="p-muted-heading p-heading--5">Devices</h3>
+                    </th>
+                  </tr>
+                  <tr className="u-no-border list-wrapper">
+                    <th className="u-text--muted">Networks</th>
+                    <td>
+                      {Object.values(profile.devices).some(isNicDevice) ? (
+                        <ExpandableList
+                          items={Object.values(profile.devices)
+                            .filter(isNicDevice)
+                            .map((device) => (
+                              <div
+                                key={device.network}
+                                className="u-truncate list-item"
+                                title={device.network}
+                              >
+                                {device.network}
+                              </div>
+                            ))}
+                        />
+                      ) : (
+                        <>-</>
+                      )}
+                    </td>
+                  </tr>
+                  <tr className="u-no-border list-wrapper">
+                    <th className="u-text--muted last-of-section">Storage</th>
+                    <td>
+                      {Object.values(profile.devices).some(isDiskDevice) ? (
+                        <ExpandableList
+                          items={Object.values(profile.devices)
+                            .filter(isDiskDevice)
+                            .map((device) => (
+                              <div
+                                key={device.path}
+                                className="u-truncate list-item"
+                                title={device.pool}
+                              >
+                                {device.pool}
+                              </div>
+                            ))}
+                        />
+                      ) : (
+                        <>-</>
+                      )}
+                    </td>
+                  </tr>
+                  <tr>
+                    <th colSpan={2}>
+                      <h3 className="p-muted-heading p-heading--5">
+                        Used by ({usageCount})
+                      </h3>
+                    </th>
+                  </tr>
+                  {!isDefaultProject && <th />}
+                  <ProfileUsedByProject
+                    profile={profile}
+                    project={projectName}
+                    headingClassName="u-text--muted"
+                    hasTableParent
+                  />
+                </tbody>
+              </table>
+            </div>
+          </div>
+        </div>
+      )}
+    </Aside>
+  );
+};
+
+export default ProfileDetailPanel;
--- a/src/pages/profiles/ProfileDetailPanel.tsx
+++ b/src/pages/profiles/ProfileDetailPanel.tsx
@@ -1,9 +1,6 @@
 import React, { FC } from "react";
-import { Button, Icon } from "@canonical/react-components";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
-import Loader from "components/Loader";
-import Aside from "components/Aside";
 import usePanelParams from "util/usePanelParams";
 import { useNotify } from "context/notify";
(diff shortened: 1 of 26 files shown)


al-bakalova:
@lorumic This looks very good. One suggestion, would it be possible to make sure the instance count never goes on a second line?
Screenshot 2023-04-21 at 13 24 54

edlerd:
> @lorumic This looks very good. One suggestion, would it be possible to make sure the instance count never goes on a second line?

This issue is also on the detail page, maybe if we find a fix, we can apply it there at the same time?

lorumic:
> > @lorumic This looks very good. One suggestion, would it be possible to make sure the instance count never goes on a second line?
> 
> This issue is also on the detail page, maybe if we find a fix, we can apply it there at the same time?

I was already aware of this. The problem is that the count is an element of variable width. We don't know how much width it will take because it depends on the value - 5 will have one width, 50 another, 500 another again.
And we have a project name that can possibly very long. So my open question is: how can we put inline a (possibly truncated) project name and its instance count which has a variable width? Any idea to achieve this easily @edlerd ?

edlerd:
> And we have a project name that can possibly be very long. So my open question is: how can we put inline a (possibly truncated) project name and its instance count which has a variable width? Any idea to achieve this easily @edlerd ?

Maybe [this](https://stackoverflow.com/questions/11905509/truncated-text-in-a-variable-width-container) helps?

lorumic:
> Maybe [this](https://stackoverflow.com/questions/11905509/truncated-text-in-a-variable-width-container) helps? The basic idea is to float right the number container. That number container can grow with its content.

What fixed min-width would you give to the number container, given that it could be 5, 50, 500, 5000...?

lorumic:
> No min-width for the number, just let it grow naturally, maybe some margin or padding though. the float right should make it take the space it needs and the name will adjust with truncation to the remaining space.

Then it's unlikely to work in the way you suggested, because the link you posted assumes a fixed width for the container on the right.

edlerd:
> > No min-width for the number, just let it grow naturally, maybe some margin or padding though. the float right should make it take the space it needs and the name will adjust with truncation to the remaining space.
> 
> Then it's unlikely to work in the way you suggested, because the link you posted assumes a fixed width for the container on the right.

It's just in the example, but I changed it in the linked jsfiddle, and it worked as well without a given width, being flexible with the content.

lorumic:
> It's just in the example, but I changed it in the linked jsfiddle, and it worked as well without a given width, being flexible with the content.

Thanks for the help - it should be looking as expected now @al-bakalova

al-bakalova:
@lorumic This looks good. A few points that I have just seen:

- When switching between a profile with no network or storage and another one that has both, the storage key pair value jumps a little. To reproduce, switch between the default profile and a-new-name-for-this-profile.
Screenshot 2023-04-24 at 18 13 51
Screenshot 2023-04-24 at 18 14 08

- For some reason, there is no section separator between devices and used by sections for alex-is-testing-profiles
Screenshot 2023-04-24 at 18 13 06

- When no instances are available, we need to apply the same spacing between the "No instances" label and the HR, as the one below (between a single instance and the next HR)
Screenshot 2023-04-24 at 18 14 57

- I feel like we need more space between the instance count and the first instance name. Would it be possible to add a little more? Like half a rem?
Screenshot 2023-04-24 at 18 17 22

- The problem with the space between the profile name and the description not being clickable (to open the side panel) seems to still persist.

al-bakalova:
> > * The problem with the space between the profile name and the description not being clickable (to open the side panel) seems to still persist.
> 
> This is no different from the instance list page. I don't think we can do much about it, you'd lose the link to the detail page otherwise.

Hm, we can than ignore this one. Thanks for clarifying

lorumic:
> Hm, we can than ignore this one. Thanks for clarifying

Forget what I've just said - I fixed it for both the instance table and the profile table :upside_down_face:

al-bakalova:
@lorumic This looks very very good. Thanks for applying all these changes!

Would it be a quick fix for you to add an empty state for the "Used by" section? 
An additional label, just like we did for snapshots - "No items found". 
Let's use "items" instead of "instances" as it might sometimes refer to projects too.

Thank you!

al-bakalova:
@lorumic The update is not available on my end. Could you please check on your end?

lorumic:
> @lorumic The update is not available on my end. Could you please check on your end?

Demos are failing again. I'll let you know once this is fixed.

al-bakalova:
@lorumic 
Could you please align left also for projects other than the default one?

Screenshot 2023-04-25 at 18 39 44

al-bakalova:
@lorumic Looks good to me! Thanks for applying all these changes!

src/pages/profiles/ProfileDetailPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const projectName = panelParams.project;
+
+  const {
+    data: profile,
+    error,
+    isLoading: isProfileLoading,
+  } = useQuery({
+    queryKey: [queryKeys.profiles, profileName, projectName],
+    queryFn: () => fetchProfile(profileName ?? "", projectName),
+    enabled: profileName !== null,
```
edlerd, line 32: Not sure if we need the enabled flag here. And if we do, why don't we need it below in the project query?

src/pages/profiles/ProfileDetailPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  const isDefaultProject = projectName === "default";
+  const usageCount = getProfileInstances(
+    projectName,
+    isDefaultProject,
+    profile?.used_by
+  ).length;
+
+  return (
+    <Aside width="narrow" pinned className="u-hide--medium u-hide--small">
```
edlerd, line 63: The skeleton and content for this panel is pretty much the same as the instance detail panel. We should extract a common component and reuse the CSS and structure to avoid duplication of code.

src/pages/profiles/ProfileUsedByProject.tsx
```diff
@@ -6,13 +6,21 @@ import { LxdProfile } from "types/profile";
...
+  headingClassName?: string;
+  hasTableParent?: boolean;
 }
 
-const ProfileUsedByProject: FC<Props> = ({ profile, project }) => {
+const ProfileUsedByProject: FC<Props> = ({
+  profile,
+  project,
+  headingClassName,
+  hasTableParent = false,
```
edlerd, line 22: Better avoid this kind of props, that let a component behave in two different ways. The two usages look so similar it should be possible to make this component generic to fit both without a need to distinguish.

src/sass/_profile_detail_panel.scss
```diff
@@ -0,0 +1,89 @@
+.profile-detail-panel {
```
edlerd, line 1: I guess most of this CSS is identical to the instance detail panel. With a shared structure component, this would probably vanish.

src/pages/profiles/ProfileUsedByProject.tsx
edlerd: I think the component is doing too many things. It renders fundamentally different based on default or non-default project. I think this logic might be better pushed into child components and contained there.
As for the table logic - I think we should not even use tables here. Something like a grid or flex layout might be better, as the data is not tabular by nature.

lorumic: The parent in the summary panel is a table, so unless you want to completely change the structure of the summary panel itself (which sounds pretty much out of scope, because it involves changing the instance's one as well for consistency), the child component needs to be wrapped inside a table row.

Apart from that, I tried to follow your suggestions, and simplified the code by creating 2 different children components for the default/non-default cases. Nonetheless, the flag is still needed, because in the summary panel we have the single project list right aligned, whereas in the overview we don't need such thing.

src/pages/profiles/ProfileDetailPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+            </tr>
+            <tr className="u-no-border">
+              <th colSpan={2}>
+                <h3 className="p-muted-heading p-heading--5">Devices</h3>
+              </th>
+            </tr>
+            <tr className="u-no-border list-wrapper">
+              <th className="u-text--muted">Networks</th>
+              <td>
+                {Object.values(profile.devices).some(isNicDevice) ? (
```
edlerd, line 93: Let put this into a dedicated component like ProfileNetworkList and reuse it here and in the overview.

src/pages/profiles/ProfileDetailPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+                  />
+                ) : (
+                  <div className="list-item">-</div>
+                )}
+              </td>
+            </tr>
+            <tr className="u-no-border list-wrapper">
+              <th className="u-text--muted last-of-section">Storage</th>
+              <td>
+                {Object.values(profile.devices).some(isDiskDevice) ? (
```
edlerd, line 115: ProfileStorageList?

src/sass/_profile_detail_panel.scss
```diff
@@ -0,0 +1,48 @@
...
+    &:nth-child(8) {
+      border-top: none !important;
+    }
+  }
+
+  tr:last-child {
+    border-bottom: none;
+  }
+
+  .content-scroll {
```
edlerd, line 20: This can probably be removed (see above)

src/pages/profiles/ProfileUsedByProject.tsx
```diff
@@ -52,54 +59,17 @@ const ProfileUsedByProject: FC<Props> = ({ profile, project }) => {
 
   return (
```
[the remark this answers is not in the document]
lorumic: Not always, but I fixed it by moving the zero usage check to the only child that needed it.

lorumic:
@edlerd I have addressed all the changes that you requested. I also tried to check that QA-wise and design-wise nothing was broken, but the cases are well branched out and I'm not 100% sure I covered all of them in my tests. So please make sure to do a careful pass of QA and also design review (especially the things that Alex mentioned in the initial comments) because so many changes are likely to have broken something.

edlerd:
> @edlerd I have addressed all the changes that you requested. I also tried to check that QA-wise and design-wise nothing was broken, but the cases are well branched out and I'm not 100% sure I covered all of them in my tests. So please make sure to do a careful pass of QA and also design review (especially the things that Alex mentioned in the initial comments) because so many changes are likely to have broken something.

I was thinking about the many branches when reviewing as well. But then I wasn't sure how to simplify. We could generalize a bit more. One idea is this:

Create a structure similar to the one for the default profile. A list of entries. Each entry is a `{projectName?: string, instances: LxdInstance[]}`. For a non-default profile, there would be just 1 entry and the `projectName` would be null. I think with this, we can use the same renderer for default and non-default projects. For the default project, we'd render a `<th>` with the project name, and for non-default project we'd render an empty `<th>`. In case the list has no entries at all, we'd render `No instances found`. Instead of the list. I think this would work for overview and side panel.

Does it make sense? What do you think - is it reducing and simplifying the options?

lorumic:
> Does it make sense? What do you think - is it reducing and simplifying the options?

Sounds like a step back to me. Following [your suggestion](https://github.com/canonical/lxd-ui/pull/305#discussion_r1176912982), I decided to split the logic for default and non-default projects into 2 children components, instead of having them together in the same parent, which was the original approach, and was messy and unclear. I still think that the current is the cleanest approach, and that is better to keep the logic of the 2 cases separated into different children components.

As for the inconsistency of the empty state between the overview and the side panel, it's an easy fix. I'll push it soon.

edlerd:
> Sounds like a step back to me. Following [your suggestion](https://github.com/canonical/lxd-ui/pull/305#discussion_r1176912982), I decided to split the logic for default and non-default projects into 2 children components, instead of having them together in the same parent, which was the original approach, and was messy and unclear. I still think that the current is the cleanest approach, and that is better to keep the logic of the 2 cases separated into different children components.

Sometimes it takes a step or two until you understand the problem and can reduce complexity. Generalizing the cases - like trying to reduce the amount of `if` usages - works quite often. Because the amount of cases you have to think of is increasing exponentially with every nested case. That's where I was coming from with the idea - trying to push the branching into the leaves of the render tree, so the amount of cases won't multiply. But I guess the current state also works quite well.

lorumic:
> Sometimes it takes a step or two until you understand the problem and can reduce complexity. Generalizing the cases - like trying to reduce the amount of `if` usages - works quite often. Because the amount of cases you have to think of is increasing exponentially with every nested case. That's where I was coming from with the idea - trying to push the branching into the leaves of the render tree, so the amount of cases won't multiply. But I guess the current state also works quite well.

I think in this particular case the 2 approaches are more or less equivalent, because the logic inside the branched cases is still fairly straightforward, so it would work well also in the way you mentioned.
But as soon as this logic gets complex enough, having a parent with the logic of these 2 cases in it makes the code much harder to maintain, and eventually you'd still have to split the logic or you wouldn't understand anything anymore. So I'm taking a prudent, future-proof stance here, just in case things change and get more complex later on.

edlerd (approved):

PR #308, opened 2023-04-24
Add version in main navigation WD-3409

edlerd:
## Done

- added server version to main navigation
- added warning next to main nav in case version is too old

Fixes WD-3409

## QA

1. Run the LXD-UI:
    - check version is displayed correctly in the navigation

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -6,6 +6,7 @@ import classnames from "classnames";
 import Logo from "./Logo";
 import ProjectSelector from "pages/projects/ProjectSelector";
 import { getProjectFromUrl } from "util/projects";
+import ServerVersion from "components/ServerVersion";
 
 const isSmallScreen = () => {
   // using the max from both, because there is a bug in chrome, causing a 0 outerWidth for
@@ -254,6 +255,7 @@ const Navigation: FC = () => {
                       Report a bug
                     </a>
                   </li>
+                  <ServerVersion />
                 </ul>
               </div>
             </div>
--- a/src/components/ServerVersion.tsx
+++ b/src/components/ServerVersion.tsx
@@ -1,0 +1,42 @@
+import React, { FC } from "react";
+import { Icon, Tooltip } from "@canonical/react-components";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { fetchSettings } from "api/server";
+
+const ServerVersion: FC = () => {
+  const { data: settings } = useQuery({
+    queryKey: [queryKeys.settings],
+    queryFn: fetchSettings,
+  });
+
+  const version = settings?.environment?.server_version;
+  if (!version) {
+    return null;
+  }
+
+  const major = version.includes(".") ? version.split(".")[0] : undefined;
+  const recentMajor = 5;
+  const isOutdated = major ? parseInt(major) < recentMajor : false;
+
+  return (
+    <>
+      <hr className="p-side-navigation__list is-dark navigation-hr" />
+      <li className="p-side-navigation__link server-version">
+        {isOutdated && (
+          <Tooltip
+            message="You are using an outdated version.
+Update your LXD server to benefit from the latest features."
+            tooltipClassName="version-warning"
+            zIndex={1000}
+          >
+            <Icon name="warning" className="p-side-navigation__icon" />
+          </Tooltip>
+        )}
+        Server {version}
+      </li>
+    </>
+  );
+};
+
+export default ServerVersion;
--- a/src/sass/_pattern_navigation.scss
+++ b/src/sass/_pattern_navigation.scss
@@ -68,6 +68,26 @@
   z-index: 1001;
 }
 
+.l-navigation .server-version {
+  color: $color-mid-light;
+}
+
+.l-navigation .server-version:hover {
+  background: $colors--dark-theme--background-default !important;
+  color: $color-mid-light !important;
+}
+
+.version-warning .p-tooltip__message {
+  bottom: 1rem;
+  left: -3rem;
+}
+
+@include desktop {
+  .version-warning .p-tooltip__message {
+    left: -3.5rem;
+  }
+}
+
 .sidenav-top-ul::after {
   display: none;
 }
--- a/src/types/server.d.ts
+++ b/src/types/server.d.ts
@@ -5,5 +5,6 @@ export interface LxdSettings {
   config: LxdConfigPair;
   environment?: {
     architectures: string[];
+    server_version: ?string;
   };
 }


edlerd:
@al-bakalova see attached screens for how the version and the too old icon/tooltip look like.
Screenshot from 2023-04-24 16-18-19
Screenshot from 2023-04-24 16-17-55

src/components/Navigation.tsx
```diff
@@ -18,6 +21,17 @@ const Navigation: FC = () => {
...
+  const { data: settings } = useQuery({
+    queryKey: [queryKeys.settings],
+    queryFn: fetchSettings,
+  });
+
+  const serverVersion = settings?.environment?.server_version;
+  const mayorVersion = serverVersion?.includes(".")
+    ? serverVersion.split(".")[0]
+    : undefined;
+  const isOldVersion = mayorVersion ? parseInt(mayorVersion) < 5 : false;
```
lorumic, line 33:
```suggestion
  const majorVersion = serverVersion?.includes(".")
    ? serverVersion.split(".")[0]
    : undefined;
  const isOldVersion = majorVersion ? parseInt(majorVersion) < 5 : false;
```

edlerd:
> * When no alert is available, the label seems misaligned. "Server" should always be left aligned with "Report a bug".
> * I think the copy can benefit from a little tuning. What about something like - "Update version to benefit from the latest features"?

That makes sense. @al-bakalova please see the screens below:

Screenshot from 2023-04-25 08-33-45
Screenshot from 2023-04-25 08-33-40
Screenshot from 2023-04-25 08-33-15

al-bakalova:
@edlerd This looks very very good! 

Here's the copy suggested by the tech writers:
`You are using an outdated version. Update your LXD server to benefit from the latest features.
`
If we can advise on the number of the most recent version, we can also mention it. Like this - 
`You are using an outdated version. Update your LXD server to version X.Y to benefit from the latest features.`

Also, the team has also advised us to list he version as "Server v5.13", not "v. 5.13". Or even just "Server 5.13".

edlerd:
> Here's the copy suggested by the tech writers: `You are using an outdated version. Update your LXD server to benefit from the latest features. `
> 
> If we can advise on the number of the most recent version, we can also mention it. Like this - `You are using an outdated version. Update your LXD server to version X.Y to benefit from the latest features.`
> 
> Also, the team has also advised us to list the version as "Server v5.13", not "v. 5.13". Or even just "Server 5.13".

Adjusted copy and tooltip position. The demo server is also working again.

Screenshot from 2023-04-25 11-02-40

al-bakalova:
@edlerd This does look good. I have one final request after seeing it live.
Would it be possible to remove the hovering effect from the label. My goal here is not to confuse the users that the whole label is clickable (and redirecting the user to a page). The user should be able to hover only on the warning icon.
Thank you.

I will go ahead and give it a +1 since I might not be able to see the changes live.

edlerd:
> @edlerd This does look good. I have one final request after seeing it live. Would it be possible to remove the hovering effect from the label? My goal here is not to confuse the user that the whole label is clickable (and redirecting the user to a page). The user should be able to hover only on the warning icon. Thank you.

Right, the hover was odd, because this is not a link, I removed it with an override.

src/components/ServerVersion.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { data: settings } = useQuery({
+    queryKey: [queryKeys.settings],
+    queryFn: fetchSettings,
+  });
+
+  const serverVersion = settings?.environment?.server_version;
+  const majorVersion = serverVersion?.includes(".")
+    ? serverVersion.split(".")[0]
+    : undefined;
+  const isOldVersion = majorVersion ? parseInt(majorVersion) < 5 : false;
```
lorumic, line 17: Maybe this hard-coded value (5) here should be moved to a const, so that we can find it and update it more easily in the future? This is likely to be changed with time, no?

src/components/ServerVersion.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <hr className="p-side-navigation__list is-dark navigation-hr" />
+      <li className="p-side-navigation__link server-version">
+        {isOldVersion ? (
+          <Tooltip
+            message="You are using an outdated version.
+Update your LXD server to benefit from the latest features."
+            tooltipClassName="version-warning"
+            zIndex={1000}
+          >
+            <Icon name="warning" className="p-side-navigation__icon" />
```
lorumic, line 31: I'm not sure I understood correctly the previous comment from @al-bakalova, but I think only the Icon should be wrapped inside the Tooltip.
Or, to say that in a less technical way, you'd get to see the tooltip only by hovering over the icon, instead that both the icon and the label, no? What is the expected behaviour here @al-bakalova?

src/components/ServerVersion.tsx
```diff
@@ -0,0 +1,42 @@
...
+    ? serverVersion.split(".")[0]
+    : undefined;
+  const isOldVersion = majorVersion ? parseInt(majorVersion) < 5 : false;
+  const versionLabel = serverVersion ? `Server ${serverVersion}` : undefined;
+
+  return serverVersion ? (
+    <>
+      <hr className="p-side-navigation__list is-dark navigation-hr" />
+      <li className="p-side-navigation__link server-version">
+        {isOldVersion ? (
```
lorumic, line 24: These nested `?` operators are making the code harder to read, can we replace them with the `&&` syntax, e.g.:
```suggestion
  return serverVersion && (
    <>
      <hr className="p-side-navigation__list is-dark navigation-hr" />
      <li className="p-side-navigation__link server-version">
        {isOldVersion && (
```

src/components/ServerVersion.tsx
edlerd: I think it is better to extend the tooltip to the text and icon hover. I don't see a reason why we would want to not show the tooltip on the "Server 5.13" text hover.

src/components/ServerVersion.tsx
edlerd: I added an early exit above and restructured the rest a bit to make it easier to digest.

src/components/ServerVersion.tsx
edlerd: right, good idea!

src/components/ServerVersion.tsx
al-bakalova: Honestly, I don't think people will naturally hover on the label. I'd expect them to hover on the alarming icon though.
I think just an icon hover is fine, but the whole label might be useful too. As long as there is no tooltip when the user is running an okay version.

lorumic: Just looking at the other tooltips that we have in the app, almost all of them are wrapping only an icon. So that seemed to me the right behaviour for this case as well.

al-bakalova: Yeah, let's keep it consistent, if that's the logic behind. That's a good point

edlerd: It is actually broken, when applying the tooltip only to the icon the position gets calculated wrong. Lets keep it simple where it is - that works :)

Screenshot from 2023-04-25 15-03-07

lorumic: > It is actually broken, when applying the tooltip only to the icon the position gets calculated wrong.

Fix suggested [here](https://github.com/canonical/lxd-ui/pull/308#discussion_r1176502399)

lorumic (approved):

PR #314, opened 2023-04-25
Redirect /ui/:project to the instance list for that project [WD-3071]

lorumic:
## Done

- Made the `/ui/:project` path redirect to the instance list.

Fixes [WD-3071](https://warthogs.atlassian.net/browse/WD-3071)

## QA

1. Run the LXD-UI:
    - Visit any `/ui/:project` path of an existing project
    - Check that it correctly redirects to the instance list for that project

[WD-3071]: https://warthogs.atlassian.net/browse/WD-3071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -13,7 +13,7 @@ import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { useNotify } from "context/notify";
 import usePanelParams from "util/usePanelParams";
-import { useNavigate, useParams } from "react-router-dom";
+import { useLocation, useNavigate, useParams } from "react-router-dom";
 import Loader from "components/Loader";
 import {
   instanceStatuses,
@@ -53,6 +53,7 @@ const saveHidden = (columns: string[]) => {
 
 const InstanceList: FC = () => {
   const instanceLoading = useInstanceLoading();
+  const location = useLocation();
   const navigate = useNavigate();
   const notify = useNotify();
   const panelParams = usePanelParams();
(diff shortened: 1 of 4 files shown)


src/pages/instances/InstanceList.tsx
```diff
@@ -69,6 +70,10 @@ const InstanceList: FC = () => {
     return <>Missing project</>;
   }
 
+  if (!location.pathname.endsWith("/instances")) {
```
edlerd, line 73: I think we should not use the InstanceList as a redirector. It is a different functionality.

lorumic: Any alternative idea then?

src/util/projects.tsx
```diff
@@ -8,13 +8,7 @@ export const projectSubpages = [
...
 ];
 
-export const getProjectFromUrl = (url: string) => {
-  const parts = url.split("/");
-  if (projectSubpages.includes(parts[3])) {
-    return parts[2];
-  }
-  return undefined;
-};
+export const getProjectFromUrl = (url: string) => url.split("/")[2];
```
edlerd, line 11: This breaks, because we also have pages like https://lxd-ui-314.demos.haus/ui/settings and currently this function returns "settings" as project. We should keep this logic as is.

Better add a redirect via custom component to the new route.

src/pages/projects/ProjectRedirect.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const location = useLocation();
+
+  if (!project) {
+    return <>Missing project</>;
+  }
+
+  if (!location.pathname.endsWith("/instances")) {
+    return <Navigate to={`/ui/${project}/instances`} replace={true} />;
+  }
+  return <InstanceList />;
```
edlerd, line 16: We don't need the InstanceList here, can we always return Navigate?

edlerd (approved):

PR #318, opened 2023-04-26
Add volume list and iso imports

edlerd:
## Done

- Add storage volume list
- Add import local file to custom volume
- Add list of images (cached and ISO)

## QA Steps
* launch a new instance and upload a prepared ISO (this needs LXD 5.16 or later, upgrade with `snap refresh lxd --channel=latest/candidate`)
* use the iso image for a new instance
* ensure the instance is launched with a special message (linking to the console tab)
* ensure instance can't be started on creation directly
* check the storage detail page
* check the storage detail page volume tab
* ensure ISO can be deleted on the volume tab if it is not in use.
* check list of images
* try deleting / using a cached image and an iso image


lorumic:
Can you please add the expected QA steps?

edlerd:
Added QA steps.

src/pages/storage/StorageVolumes.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const filteredVolumes = volumes.filter((volume) => {
+    return volume.name.toLowerCase().includes(query.toLowerCase());
+  });
+
+  const rows = filteredVolumes.map((volume) => {
+    return {
+      columns: [
+        {
+          content: volume.name,
+          role: "rowheader",
```
lorumic, line 69: This (and all the following) should be:
```suggestion
          role: "cell",
```

src/pages/storage/StorageVolumes.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <div key={key}>
+              {key}: {val}
+            </div>
+          )),
+          role: "rowheader",
+          "aria-label": "Config",
+        },
+        {
+          content: volume.used_by?.map((entry) => (
+            <div key={entry}>{entry}</div>
```
lorumic, line 103: This doesn't look very well on medium/small screens. We should allow it to either wrap `anywhere` or add the `u-truncate` class and set the `title`.

image

src/util/images.tsx
```diff
@@ -1,10 +1,14 @@
 import { RemoteImage } from "types/image";
+import { LxdStorageVolume } from "types/storage";
 
 export const isVmOnlyImage = (image: RemoteImage) => {
-  return image.variant?.includes("desktop");
+  return image.variant?.includes("desktop") ?? image.server === "local-iso";
```
lorumic, line 5: I think this line should be:
```suggestion
  return image.variant?.includes("desktop") || image.server === "local-iso";
```
Otherwise, the local ISO would be detected as a "container only" image. Which would render a Win11 ISO useless.

image

src/pages/storage/StorageDetail.tsx
```diff
@@ -54,40 +68,28 @@ const StorageDetail: FC = () => {
...
-                </tr>
-                <tr>
-                  <th className="u-text--muted">Driver</th>
-                  <td>{storagePool.driver}</td>
-                </tr>
-              </tbody>
-            </table>
-            <h2 className="p-heading--5">Used by</h2>
-            <StorageUsedBy storage={storagePool} project={project} />
+            <Tabs
```
lorumic, line 71: Please make sure that only the tabs content is scrollable, and everything above is fixed at the top all the time, just like in the instance and profile detail pages. This is particularly important for the notification, because it's not being scrolled into view (intentionally? Or is it unexpected?), leaving the user confused with no apparent feedback from the operation. See the following screen capture to understand the problem more concretely:

Peek 2023-08-07 18-48

src/pages/instances/CreateInstanceForm.tsx
```diff
@@ -334,15 +372,17 @@ const CreateInstanceForm: FC = () => {
...
-                  onClick={() => submit(formik.values)}
-                />
+                {!isLocalIsoImage && (
+                  <SubmitButton
+                    isSubmitting={formik.isSubmitting}
+                    isDisabled={!formik.isValid || !formik.values.image}
+                    buttonLabel="Create and start"
+                    onClick={() => submit(formik.values)}
+                  />
+                )}
```
lorumic, line 385: I'm not sure I understand why we need to force the users to start the instance manually in this case. They would still need to start it in order to do... anything. So why removing the possibility to "create and start"?

lorumic (commented):
This is a good start. I'm adding a few suggestions and ideas for improvements.

src/util/images.tsx
lorumic: This is actually diriment - one cannot proceed further in any way until it's fixed.

Screenshot from 2023-08-07 18:39:06

src/pages/instances/CreateInstanceForm.tsx
edlerd: The user needs to be on the graphical console and finish the installation for the VM to work. Starting the VM here would imply the instance is usable after, which it is not. This is especially critical for windows, where you need to "press any key" to boot from the mounted ISO within a 5 second interval, and not from the empty disk. Starting the VM will bring the VM into a failed boot state.

So we decided to disable the "Create and start" option and have the changed notification that links to the console. So the users will likely go to the console, start it there and naturally understand to follow the installation process via the GUI.

src/pages/storage/UploadIsoImage.tsx
```diff
@@ -0,0 +1,142 @@
...
+          label: pool.name,
+          value: pool.name,
+        }))}
+        onChange={(e) => {
+          setPool(e.target.value);
+        }}
+        value={pool}
+        disabled={file === null}
+      />
+      <Notification severity="caution" title="Custom ISO images">
```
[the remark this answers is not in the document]
edlerd: moved it up.

src/pages/storage/UploadIsoImage.tsx
```diff
@@ -0,0 +1,142 @@
...
+        Image must be prepared with distrobuilder.{" "}
+        <a
+          href="https://discourse.ubuntu.com/t/how-to-install-a-windows-11-vm-using-lxd/28940"
+          target="_blank"
+          rel="noreferrer"
+        >
+          Learn more
+        </a>
```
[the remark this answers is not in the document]
edlerd: I think we will iterate on this note quite a bit, I tried to give more detailed description now and pick windows as an example.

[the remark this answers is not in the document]
lorumic: That doesn't really give a detailed description. It just provides a link with a tutorial that works for Windows only.

src/pages/storage/UploadIsoImage.tsx
```diff
@@ -0,0 +1,142 @@
...
+  if (storagePools.length > 0 && !pool) {
+    setPool(storagePools[0].name);
+  }
+
+  const importIsoFile = () => {
+    if (!file) {
+      return;
+    }
+    setLoading(true);
+    createIsoStorageVolume(pool, file, name, project?.name ?? "")
```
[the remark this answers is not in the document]
edlerd: Yes, we need a progress indicator here. I will add it shortly.

src/pages/instances/CreateInstanceForm.tsx
edlerd: We considered changing the redirect target to the console, but went against it, because it might be annoying to have different redirect targets for experienced users, that have an expectation of where they land on launch.

src/pages/storage/UploadIsoImage.tsx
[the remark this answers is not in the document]
edlerd: I commented before pushing, the note is now a bit different. I'll also discuss it with Piper from UX later today.

src/pages/images/ImageSelectorModal.tsx
```diff
@@ -0,0 +1,41 @@
...
+    >
+      {content === REMOTE_IMAGES && (
+        <RemoteImageSelector
+          onSelect={onSelect}
+          onUpload={() => setContent(UPLOAD)}
+        />
+      )}
+      {content === UPLOAD && (
+        <UploadIsoImage
+          onCancel={() => setContent(REMOTE_IMAGES)}
```
lorumic, line 33: Is there a way to actually cancel the current upload operation? So that if a user happens to change their mind while uploading, they can stop the upload for good instead of having it continue in the background?

src/pages/storage/UploadIsoImage.tsx
```diff
@@ -0,0 +1,155 @@
...
+        <>
+          <ProgressBar percentage={Math.floor(uploadState.percentage)} />
+          <p>
+            {humanFileSize(uploadState.loaded)} loaded of{" "}
+            {humanFileSize(uploadState.total ?? 0)}
+          </p>
+          {uploadState.loaded === uploadState.total && (
+            <p>
+              <i className="p-icon--spinner u-animation--spin" /> Validating ISO
+            </p>
```
lorumic, line 138: Maybe we can make the code a bit clearer here by using our `Loader` component?
```suggestion
            <Loader text="Validating ISO" />
```
If needed, you can customise the `u-loader` class to remove the margins in this context.

src/pages/storage/UploadIsoImageHint.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { FC } from "react";
+
+const UploadIsoImageHint: FC = () => {
+  return (
+    <div className={`p-notification--caution`}>
+      <div className="p-notification__content">
+        <h5 className="p-notification__title">
```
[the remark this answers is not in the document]
edlerd: I adjusted it a bit more and iterated on the dialogue layout to improve the experience.

src/pages/images/RemoteImageSelector.tsx
```diff
@@ -212,7 +229,7 @@ const ImageSelector: FC<Props> = ({ onClose, onSelect }) => {
         ],
         sortData: {
           os: item.os.toLowerCase(),
-          release: item.release.toLowerCase(),
+          release: item.release_title.toLowerCase(),
```
lorumic, line 232: Incidentally, this unapproved change from the other PR is making the app crash:
```
react-dom.development.js:12056 Uncaught TypeError: Cannot read properties of undefined (reading 'toLowerCase')
    at RemoteImageSelector.tsx:232:39
    at Array.map (<anonymous>)
    at RemoteImageSelector (RemoteImageSelector.tsx:166:6)
    at renderWithHooks (react-dom.development.js:16305:18)
    at updateFunctionComponent (react-dom.development.js:19588:20)
    at beginWork (react-dom.development.js:21601:16)
    at beginWork$1 (react-dom.development.js:27426:14)
    at performUnitOfWork (react-dom.development.js:26557:12)
    at workLoopSync (react-dom.development.js:26466:5)
    at renderRootSync (react-dom.development.js:26434:7)
```

src/types/image.d.ts
```diff
@@ -27,12 +27,14 @@ export interface ImportImage {
...
   arch: string;
+  created_at: number;
   lxd_requirements?: {
     secureboot: boolean;
   };
   os: string;
+  pool?: string;
   release: string;
-  release_title?: string;
+  release_title: string;
```
lorumic, line 37: It's not always there, see the other comment. This needs to be treated as optional.

src/pages/images/RemoteImageSelector.tsx
edlerd: Huh, this shouldn't have ended up in this branch, I'll reset the head to be without the commit about release_title.

src/types/image.d.ts
edlerd: Huh, this shouldn't have ended up in this branch, I'll reset the head to be without the commit about release_title.

lorumic: Might still be worth noting it down and fix it in the other branch.

edlerd: Yes, I changed it there to have it fixed.

src/pages/storage/UploadIsoImageHint.tsx
```diff
@@ -0,0 +1,40 @@
...
+const UploadIsoImageHint: FC = () => {
+  return (
+    <>
+      <div className={`p-notification--information`}>
+        <div className="p-notification__content">
+          <h5 className="p-notification__title">
+            Image must be prepared with distrobuilder
+          </h5>
+          <p>
+            For a windows image with name <code>WindowsIsoImage.iso</code> use
```
lorumic, line 12:
```suggestion
            For a Windows image with name <code>WindowsIsoImage.iso</code>, use
```

lorumic (commented):
Thanks for applying the changes. It looks and works much better now in my opinion. I'd still suggest to ask some UX/design expert to double-check the whole experience, just in case (especially the upload modal), but I will definitely remove the QA -1 label now.
Also adding a small copy suggestion.

edlerd:
Thank you for the design review @piperdeck 
I realized all the suggestions you mentioned above, and also the additional ones we discussed today, including
- redesign the "select base image" button / "change image" buttons on instance launch
- after custom image upload the users goes back to the image list with the new images select button positive

piperdeck (approved):
Everything seems good to me. The instance creation flow feels great!

PR #321, opened 2023-04-28
Use search and filter component for the instance list [WD-3189]

lorumic:
## Done

- Combines the previous search box + dropdown filters into a single "Search and filter" component.
- Allows filtering by instance status, type, and applied profile.

Fixes [WD-3189](https://warthogs.atlassian.net/browse/WD-3189)

## QA

1. Run the LXD-UI:
    - Visit any instance list page
    - Check the behaviour and the look-and-feel of the new search and filter component in the header

[WD-3189]: https://warthogs.atlassian.net/browse/WD-3189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -1,12 +1,5 @@
-import {
-  Button,
-  Col,
-  Icon,
-  Row,
-  SearchBox,
-  Select,
-} from "@canonical/react-components";
 import React, { FC, useEffect, useState } from "react";
+import { Button, Col, Icon, Row } from "@canonical/react-components";
 import { fetchInstances } from "api/instances";
 import NotificationRow from "components/NotificationRow";
 import { useQuery } from "@tanstack/react-query";
@@ -15,11 +8,7 @@ import { useNotify } from "context/notify";
 import usePanelParams from "util/usePanelParams";
 import { useNavigate, useParams } from "react-router-dom";
 import Loader from "components/Loader";
-import {
-  instanceStatuses,
-  instanceListTypes,
-  instanceCreationTypes,
-} from "util/instanceOptions";
+import { instanceCreationTypes } from "util/instanceOptions";
 import InstanceStatusIcon from "./InstanceStatusIcon";
 import TableColumnsSelect from "components/TableColumnsSelect";
 import useEventListener from "@use-it/event-listener";
@@ -35,6 +24,9 @@ import SelectableMainTable from "components/SelectableMainTable";
 import InstanceBulkActions from "pages/instances/actions/InstanceBulkActions";
 import { getIpAddresses } from "util/networks";
 import InstanceBulkDelete from "pages/instances/actions/InstanceBulkDelete";
+import InstanceSearchFilter from "./InstanceSearchFilter";
+import { InstanceFilters } from "util/instanceFilter";
+import { isWidthBelow } from "util/helpers";
 
 const STATUS = "Status";
 const NAME = "Name";
@@ -53,15 +45,22 @@ const saveHidden = (columns: string[]) => {
   localStorage.setItem("instanceListHiddenColumns", JSON.stringify(columns));
 };
 
+const isMediumScreen = () => isWidthBelow(820);
+
 const InstanceList: FC = () => {
   const instanceLoading = useInstanceLoading();
   const navigate = useNavigate();
   const notify = useNotify();
   const panelParams = usePanelParams();
   const { project } = useParams<{ project: string }>();
-  const [query, setQuery] = useState<string>("");
-  const [status, setStatus] = useState<string>("any");
-  const [type, setType] = useState<string>("any");
+  const [createButtonLabel, _setCreateButtonLabel] =
+    useState<string>("Create instance");
+  const [filters, setFilters] = useState<InstanceFilters>({
+    queries: [],
+    statuses: [],
+    types: [],
+    profileQueries: [],
+  });
   const [userHidden, setUserHidden] = useState<string[]>(loadHidden());
   const [sizeHidden, setSizeHidden] = useState<string[]>([]);
   const [selectedNames, setSelectedNames] = useState<string[]>([]);
@@ -83,6 +82,11 @@ const InstanceList: FC = () => {
   if (error) {
     notify.failure("Loading instances failed", error);
   }
+
+  const setCreateButtonLabel = () => {
+    _setCreateButtonLabel(isMediumScreen() ? "Create" : "Create instance");
+  };
+  useEventListener("resize", setCreateButtonLabel);
 
   const figureSizeHidden = () => {
     const wrapper = document.getElementById("instance-table-measure");
@@ -126,22 +130,31 @@ const InstanceList: FC = () => {
   };
 
   const filteredInstances = instances.filter((item) => {
-    if (query) {
-      const q = query.toLowerCase();
-      if (
-        !item.name.toLowerCase().includes(q) &&
-        !item.status.toLowerCase().includes(q) &&
-        !item.type.toLowerCase().includes(q) &&
-        !item.description.toLowerCase().includes(q) &&
-        !item.config["image.description"]?.toLowerCase().includes(q)
-      ) {
-        return false;
-      }
-    }
-    if (type !== "any" && item.type !== type) {
+    if (
+      !filters.queries.every(
+        (q) =>
+          item.name.toLowerCase().includes(q) ||
+          item.description.toLowerCase().includes(q) ||
+          item.config["image.description"]?.toLowerCase().includes(q)
+      )
+    ) {
       return false;
     }
-    if (status !== "any" && item.status !== status) {
+    if (
+      filters.statuses.length > 0 &&
+      !filters.statuses.includes(item.status)
+    ) {
+      return false;
+    }
+    if (filters.types.length > 0 && !filters.types.includes(item.type)) {
+      return false;
+    }
+    if (
+      filters.profileQueries.length > 0 &&
+      !filters.profileQueries.every((profile) =>
+        item.profiles.includes(profile)
+      )
+    ) {
       return false;
     }
     return true;
@@ -303,15 +316,16 @@ const InstanceList: FC = () => {
 
   const pagination = usePagination(getRows(userHidden.concat(sizeHidden)));
 
-  useEventListener("resize", () => updateTBodyHeight("instance-table-wrapper"));
-  useEffect(() => {
-    updateTBodyHeight("instance-table-wrapper");
-  }, [
+  const updateTableHeight = () => updateTBodyHeight("instance-table-wrapper");
+
+  useEventListener("resize", updateTableHeight);
+  useEffect(updateTableHeight, [
     instances,
     notify.notification,
-    query,
-    status,
-    type,
+    filters.queries,
+    filters.statuses,
+    filters.types,
+    filters.profileQueries,
     pagination.pageSize,
     pagination.currentPage,
   ]);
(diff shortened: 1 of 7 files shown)


src/pages/instances/InstanceList.tsx
```diff
@@ -124,22 +151,32 @@ const InstanceList: FC = () => {
...
+          item.description.toLowerCase().includes(keyword)
+        )
       ) {
         return false;
       }
     }
-    if (type !== "any" && item.type !== type) {
+    if (
+      statuses.length > 0 &&
+      !statuses.some((status) => item.status === status)
```
edlerd, line 169: `!statuses.includes(item.status)`

src/pages/instances/InstanceList.tsx
```diff
@@ -124,22 +151,32 @@ const InstanceList: FC = () => {
...
+    ) {
       return false;
     }
-    if (status !== "any" && item.status !== status) {
+    if (types.length > 0 && !types.some((type) => item.type === type.name)) {
```
edlerd, line 173: `!types.includes(item.type)`

src/pages/instances/InstanceList.tsx
```diff
@@ -307,6 +344,65 @@ const InstanceList: FC = () => {
...
+      .map((chip) => {
+        return {
+          name: chip.value === "VM" ? "virtual-machine" : "container",
+          uiName: chip.value,
+        } as InstanceListType;
+      });
+    const newProfileQueries = searchData
+      .filter((chip) => chip.lead === "profile")
+      .map((chip) => chip.value);
+    if (!isEqual(queries, newQueries)) setQueries(newQueries);
```
edlerd, line 399: I think it is easier now to combine all states into a single state variable with a complex type:

```
interface InstanceFilter = {
queries: string[];
status: LxdInstanceStatus[];
...

const [filter, setFilter] = useState<InstanceFilter>()
```
Then we don't need the multple isEqual checks here, but just replace the whole object.

src/pages/instances/InstanceList.tsx
```diff
@@ -354,61 +451,24 @@ const InstanceList: FC = () => {
...
-                  name="search-instance"
-                  type="text"
-                  onChange={(value) => {
-                    setQuery(value);
-                  }}
-                  placeholder="Search"
-                  value={query}
-                  aria-label="Search"
+              <div className="search-wrapper margin-right u-no-margin--bottom">
+                <SearchAndFilter
```
edlerd, line 455: Maybe contain this and all the code to build it in a dedicated component? When combining the filters into a single state, the interface (as amount of props) should be fairly small.

src/pages/instances/InstanceList.tsx
lorumic: `item.type` is a string. `types` is an array of objects.

src/pages/instances/InstanceList.tsx
lorumic: This could have worked, but eventually I decided to implement the "merged" logic for temporary statuses.

src/pages/instances/InstanceList.tsx
lorumic: We have a mismatch between the UI name and the BE name.

src/pages/instances/InstanceList.tsx
lorumic: Right, this way we can reduce the `isEqual` check to one instead of four, but you cannot replace the whole object blindly. You'd exceed the maximum update depth otherwise.

src/pages/instances/InstanceList.tsx
lorumic: Good idea, I was already planning to do this, because the `InstanceList` is already super-polluted - this is to say that, as a future improvement, we should extract many more things out of it - but eventually I got distracted by the countless layout issues created by the introduction of the `SearchAndFilter` component...

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const newProfileQueries = searchData
+      .filter((chip) => chip.lead === "profile")
+      .map((chip) => chip.value);
+    const newFilters = {
+      queries: newQueries,
+      statuses: newStatuses,
+      types: newTypes,
+      profileQueries: newProfileQueries,
+    };
+    if (!isEqual(filters, newFilters)) setFilters(newFilters);
```
edlerd, line 79: `onSearchDataChange` only gets called when some filter changed, no? I think we can drop the `isEqual` check here.

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -0,0 +1,92 @@
...
+      .filter((chip) => chip.quoteValue)
+      .map((chip) => chip.value);
+    const newStatuses = searchData
+      .filter((chip) => chip.lead === "status")
+      .map((chip) => chip.value as LxdInstanceStatus);
+    const newTypes = searchData
+      .filter((chip) => chip.lead === "type")
+      .map((chip) => {
+        return {
+          name: chip.value === "VM" ? "virtual-machine" : "container",
```
edlerd, line 66: Why can't we simplify this to an array of just the name? I think we don't need the uiName in the filter state, do we?

src/pages/instances/InstanceSearchFilter.tsx
lorumic: Try it yourself, no? :)

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -0,0 +1,92 @@
...
+      chips: profileSet.map((profile) => {
+        return { lead: "profile", value: profile };
+      }),
+    },
+  ];
+
+  const onSearchDataChange = (searchData: SearchAndFilterChip[]) => {
+    const newQueries = searchData
+      .filter((chip) => chip.quoteValue)
+      .map((chip) => chip.value);
```
edlerd, line 58: We can do the `chip.value.toLowerCase()` here, so we don't do it for every instance that is filtered.

src/pages/instances/InstanceSearchFilter.tsx
edlerd: Well the prop comment says `A function that is called when the search data changes.` so I am pretty sure. :)

lorumic: I'm also sure since I tried it and got the "maximum update depth exceeded" error. That's why I said to try it.

edlerd: Right, because we feed in `filter` and `setFilters` the whole component re-renders. We don't need filters and can just remove it.
Then `setFilter` still changes every time we call it and causes a re-render. But we can avoid that with `export default React.memo(InstanceSearchFilter);`. Then we can simplify the interface to this component (by removing the filters prop) and get rid of the `isEqual` check.

src/pages/instances/InstanceList.tsx
```diff
@@ -354,61 +360,23 @@ const InstanceList: FC = () => {
...
-                    ...instanceListTypes,
-                  ]}
-                  value={type}
-                  aria-label="Filter type"
-                />
-              </>
+              <InstanceSearchFilter
+                instances={instances}
+                filters={filters}
+                setFilters={setFilters}
```
edlerd, line 366: The component keeps its state when we switch project, we can avoid that by using it as a key: `key={project}`, hence forcing a clean render of it.

lorumic:
> * When the user shrinks the screen, can we force-collapse the side navigation?

Done, but the user can still click on the arrows to expand the side navigation, hence reproducing the problem.

> * When resized, would it be possible to keep the dropdown larger and not align it to the width of the field?

It's currently using 100% width of its parent, so I don't know an easy way to achieve this cleanly. Any ideas @edlerd?

> There is also a glitch when in mobile view, with the navigation open, and you see some part of the search + filter input - when clicking into the input, the dropdown overlays the navigation in a quite wired way.

What would be the solution to this? Since opening/closing the filters view doesn't trigger any event, I cannot think of any easily applicable countermeasure. Any proposal?

edlerd:
I think we should add a callback to the search and filter component to be notified when the dropdown is opened. So we can shrink the navigation on medium-sized screens in case it is expanded on that event. But lets do that as a follow up to unblock this PR.

One thing is still broken: The "Create instance" button grows to two rows when in medium size and manually opening the navigation - can you avoid it?

src/pages/instances/InstanceList.tsx
```diff
@@ -126,22 +120,33 @@ const InstanceList: FC = () => {
...
-        !item.status.toLowerCase().includes(q) &&
-        !item.type.toLowerCase().includes(q) &&
-        !item.description.toLowerCase().includes(q) &&
-        !item.config["image.description"]?.toLowerCase().includes(q)
-      ) {
-        return false;
-      }
+    if (
+      filters.queries.length > 0 &&
+      !filters.queries.every((q) => item.name.toLowerCase().includes(q)) &&
```
edlerd, line 125: The logic is a bit off. For every query we want it to match at least one of name, descripiton or image I think like this:
```
!filters.queries.every((q) => 
  item.name.toLowerCase().includes(q) || 
  item.description.toLowerCase().includes(q) ||
  item.config["image.description"]?.toLowerCase().includes(q)
)
```

edlerd: I think we can remove the `filters.queries.length > 0` because if is an empty array the `every` is implicitly true.

al-bakalova:
> I think we should add a callback to the search and filter component to be notified when the dropdown is opened. So we can shrink the navigation on medium-sized screens in case it is expanded on that event. But lets do that as a follow up to unblock this PR.
> 
> One thing is still broken: The "Create instance" button grows to two rows when in medium size and manually opening the navigation - can you avoid it?

I'd say the "Create instance" label should be shortened to just "Create" on smaller screens

lorumic:
> I'd say the "Create instance" label should be shortened to just "Create" on smaller screens

Good idea, it seems to me that this solved the problem well.

edlerd (approved):
LGTM, follow up ticket to consume the [upstream events](https://github.com/canonical/react-components/pull/919) and react to close navigation and adjust height is also created

al-bakalova:
@lorumic This looks very good. Two quick questions:
- Would it be possible to allow the search bar to get closer to the "Create" button and benefit from the remaining space?
- Would it be possible for the dropdown to be wider than the field above? Like the one we have for projects.
Screenshot 2023-05-16 at 18 24 20

al-bakalova:
@lorumic Looks good to me! Thank you for applying all these changes! It's exactly how I imagined it! Grazie!

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+} from "@canonical/react-components/dist/components/SearchAndFilter/types";
+
+interface Props {
+  instances: LxdInstance[];
+  setFilters: (newFilters: InstanceFilters) => void;
+}
+
+const InstanceSearchFilter: FC<Props> = ({ instances, setFilters }) => {
+  const profileSet = [
+    ...new Set(instances.flatMap((instance) => instance.profiles)),
```
edlerd, line 22: I think we should display all profiles in the project, not only the used ones. One application will be to link here (with that profile filter set) from a profile page. If the profile is not available (because no instances use it) that application might not work.

lorumic: The application that you mentioned will exist only for profiles that have at least 1 instance, not for the "empty" ones. There is an empty state for them, so they won't have any link, and hence there's no need to pollute the filter panel with all the empty profiles.

PR #336, opened 2023-05-10
Ask for certificate password on creation #331

### LINKED ISSUE #331: lxd-ui.pfx does not work on a Mac because of empty password.
When using any browser on a Mac, it uses the OS for certificates. Macos, at least Ventura doesn't allow blank passwords for PFX files. The easy fix would be for the generator on LXD US to prompt for a password when generating the certs, that way the user can type that in when they add it to the MacOS keychain. 

I did figure out a workaround... 
If I run this in the trerminal on my Mac, it will ass a password to a new file:
`openssl pkcs12 -in lxd-ui.pfx -password pass:"" -passin pass:"" -passout pass:"InsertPasswordHere" -out "lxd-ui-mac.pfx"`

Then all I had to do is double click on the lxd-ui-mac.pfx file, and enter the password I put in above. Now I have access!

edlerd:
## Done

- add prompt on certificate generation to allow setting a password.

Fixes #331 

## QA

1. Run the LXD-UI:
    - check cert generation

### DIFF
--- a/src/pages/certificates/CertificateGenerate.tsx
+++ b/src/pages/certificates/CertificateGenerate.tsx
@@ -5,6 +5,7 @@ import BrowserImport from "pages/certificates/BrowserImport";
 import { Navigate } from "react-router-dom";
 import { useAuth } from "context/auth";
 import Loader from "components/Loader";
+import PasswordModal from "pages/certificates/PasswordModal";
 
 interface Certs {
   crt: string;
@@ -13,6 +14,7 @@ interface Certs {
 
 const CertificateGenerate: FC = () => {
   const [isGenerating, setGenerating] = useState(false);
+  const [isModalOpen, setModalOpen] = useState(false);
   const [certs, setCerts] = useState<Certs | null>(null);
   const { isAuthenticated, isAuthLoading } = useAuth();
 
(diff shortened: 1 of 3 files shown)


al-bakalova:
Hey @edlerd 
This works well for me. However, I would like to clarify a couple of questions:
Screenshot 2023-05-10 at 13 27 46
- Would it be possible to keep the password hidden behind asterisks? I feel like that might not be the ideal representation and some users might be frustrated to type a password.
- We seem to be using a native browser prompt. Would it be possible to create a Vanilla one or are there any technical limitations that stop us from doing so?

edlerd:
> * Would it be possible to keep the password hidden behind asterisks? I feel like that might not be the ideal representation and some users might be frustrated to type a password.
> * We seem to be using a native browser prompt. Would it be possible to create a Vanilla one or are there any technical limitations that stop us from doing so?

You are right, this doesn't work with the native prompt. I'll change it to a vanilla modal.

edlerd:
@al-bakalova  please check again with the vanilla style modal.

al-bakalova:
@edlerd This is very good. I would apply a few changes to the copy:

Modal header: Add a password

Protect your certificate by adding a password (No need to explain they can skip it - we will add a button to tell them so).

Buttons: [Skip]  [Generate certificate]  (to match with the previous screen)

There is no error message when the two passwords do not match. I'd suggest something like:
Passwords do not match

Before entering a password, the primary button is enabled. Could you please disable it from the beginning?

edlerd:
Thank you for the copy updates @al-bakalova 
I evolved the password modal around them and added a 2nd input to avoid typos.

PR #344, opened 2023-05-15
Select 1st pool when configuring root storage

### LINKED ISSUE #342: Require selecting a storage pool, or leave the selected one
I used the UI to create an instance and wanted to increase the storage pool size, so I configured that before launching the instance:
Screenshot from 2023-05-12 14-00-13

The result was:
`Failed creating instance record: Failed initialising instance: Invalid devices: Device validation failed for \"root\": Root disk entry must have a \"pool\" property set`

Before I selected "Root storage pool" to configure the size, the `default` pool was selected there, so I didn't even look at the dropdown (the only thing I wanted to change was the size).
But as soon as I selected "Root storage pool", the dropdown selection changed to "Select option", and it let me launch the instance without selecting anything, which caused the error.

It makes sense in hindsight, but I think we could improve the usability here - either by requiring a pool to be selected, or by keeping the dropdown on `default` unless something else is selected.

edlerd:
## Done

- automatically select the 1st pool, when overriding root storage, disallow to unselect the pool completely

Fixes #342 

## QA

1. Run the LXD-UI:
    - Configure root storage on an instance/profil edit/creation

### DIFF
--- a/src/pages/instances/forms/RootStorageForm.tsx
+++ b/src/pages/instances/forms/RootStorageForm.tsx
@@ -67,6 +67,7 @@ const RootStorageForm: FC<Props> = ({ formik, project }) => {
       type: "disk",
       name: "root",
       path: "/",
+      pool: storagePools[0]?.name ?? undefined,
     });
     formik.setFieldValue("devices", copy);
     setTimeout(() => document.getElementById("rootStoragePool")?.focus(), 100);
@@ -83,6 +84,7 @@ const RootStorageForm: FC<Props> = ({ formik, project }) => {
       return {
         label: storagePool.name,
         value: storagePool.name,
+        disabled: false,
       };
     });
     options.unshift({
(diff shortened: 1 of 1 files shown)


al-bakalova:
@edlerd I am not sure what problem we are trying to solve here. Could we please review this one together tomorrow?

src/pages/instances/forms/RootStorageForm.tsx
```diff
@@ -83,6 +84,7 @@ const RootStorageForm: FC<Props> = ({ formik, project }) => {
       return {
         label: storagePool.name,
         value: storagePool.name,
+        disabled: false,
```
lorumic, line 87: Isn't this false by default? Do we need to specify it?

edlerd: It's a way to keep the types similar without specifying the type of the objects in the options array.

PR #346, opened 2023-05-17
Link to instance list with profile filter applied [WD-3874]

lorumic:
## Done

- Replace the previous "Show all" behaviour used in the profile detail overview and profile summary panel with a link to the instance list with the profile filter applied.

Fixes [WD-3874](https://warthogs.atlassian.net/browse/WD-3874)

## QA

1. Run the LXD-UI:
    - Go to profiles page
    - Open the profile detail overview or the profile summary panel of a profile with more than 5 instances
    - Click on the "Show all" link to see the new behaviour

[WD-3874]: https://warthogs.atlassian.net/browse/WD-3874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/profiles/ProfileUsedByDefaultProject.tsx
+++ b/src/pages/profiles/ProfileUsedByDefaultProject.tsx
@@ -1,9 +1,9 @@
 import React, { FC } from "react";
 import { LxdUsedBy } from "util/usedBy";
-import ExpandableList from "components/ExpandableList";
-import InstanceLink from "pages/instances/InstanceLink";
+import ViewProfileInstancesBtn from "./actions/ViewProfileInstancesBtn";
 
 interface Props {
+  profile: string;
   affectedProjects?: {
     name: string;
     instances: LxdUsedBy[];
@@ -12,6 +12,7 @@ interface Props {
 }
 
 const ProfileUsedByDefaultProject: FC<Props> = ({
+  profile,
   affectedProjects,
   headingClassName,
 }) => {
@@ -32,16 +33,9 @@ const ProfileUsedByDefaultProject: FC<Props> = ({
               <i className="u-text--muted no-instances">No instances</i>
             )}
             {project.instances.length > 0 && (
-              <ExpandableList
-                items={project.instances.map((instance) => (
-                  <div
-                    key={instance.name}
-                    className="u-truncate list-item"
-                    title={instance.name}
-                  >
-                    <InstanceLink instance={instance} />
-                  </div>
-                ))}
+              <ViewProfileInstancesBtn
+                profile={profile}
+                project={project.name}
               />
             )}
           </td>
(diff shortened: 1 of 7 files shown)


src/pages/profiles/ProfileUsedByDefaultProject.tsx
```diff
@@ -33,6 +40,13 @@ const ProfileUsedByDefaultProject: FC<Props> = ({
             )}
             {project.instances.length > 0 && (
               <ExpandableList
+                customShowAll={() => {
```
edlerd, line 43: We probably shouldn't use the Expandable list here, otherwise the experience of "show all" is inconsistent.

Also, I think the link to the filtered instance list is useful even with 2 instances to bulk-stop or restart them.

lorumic: Hmm, yes, could be, but what would be the best way to represent it visually then? @al-bakalova

src/pages/profiles/ProfileUsedByDefaultProject.tsx
```diff
@@ -32,17 +35,7 @@ const ProfileUsedByDefaultProject: FC<Props> = ({
...
-                  <div
-                    key={instance.name}
-                    className="u-truncate list-item"
-                    title={instance.name}
-                  >
-                    <InstanceLink instance={instance} />
-                  </div>
-                ))}
-              />
+              <ViewAllInstancesBtn profile={profile} project={projectName} />
```
edlerd, line 38: This is the wrong project when referencing another one. I think it should be `project.name`, and we don't need to pass in the `projectName` here.

src/pages/profiles/actions/ViewAllInstancesBtn.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { FC } from "react";
+import { useNavigate } from "react-router-dom";
+import { Button } from "@canonical/react-components";
+
+interface Props {
+  profile: string;
+  project: string;
+}
+
+const ViewAllInstancesBtn: FC<Props> = ({ profile, project }) => {
```
edlerd, line 10: ViewProfileInstancesBtn?

src/pages/profiles/actions/ViewAllInstancesBtn.tsx
```diff
@@ -0,0 +1,31 @@
...
+      small
+      onClick={() => {
+        navigate(`/ui/${project}/instances`, {
+          state: {
+            appliedProfile: profile,
+          },
+        });
+      }}
+    >
+      View all instances
```
edlerd, line 26: Maybe shorten to "View instances"?

al-bakalova:
@lorumic This looks good! A small comment on my side:
- Instead of "View all instances" can we please shorten it to just "Go to instances". I believe that way we will give the user a hint that we will redirect them to another page.

lorumic:
> @lorumic This looks good! A small comment on my side: Instead of "View all instances" can we please use "Go to instances". I believe that way we will give the user a hint that we will redirect them to another page.

Ok, I updated that

edlerd (approved):
LGTM

PR #355, opened 2023-05-24
Add scrollable table component WD-3408

edlerd:
## Done

- added reusable scrollable table component

Fixes WD-3408

## QA

1. Run the LXD-UI:
    - check scrollable tables  (instance list, snapshot list, profile list, profile/instance configuration)


lorumic:
A very high-level QA comment before diving into the code. The mechanism to auto-hide columns on medium screens in the instance table seems to be broken. See the following comparison with 1240px width:

BEFORE:
image

AFTER:
image

edlerd:
> A very high-level QA comment before diving into the code. The mechanism to auto-hide columns on medium screens in the instance table seems to be broken.

That is fixed just now. I had in the back of my head the idea to use a ref instead of giving a static `id` and using `getElementById`, which turns out to fix this issue as well :)

src/pages/instances/InstanceList.tsx
```diff
@@ -535,30 +521,42 @@ const InstanceList: FC = () => {
...
+                      instances,
+                      creationOperations,
+                      notify.notification,
+                      filters.queries,
+                      filters.statuses,
+                      filters.types,
+                      filters.profileQueries,
+                      pagination.pageSize,
+                      pagination.currentPage,
+                    ]}
```
lorumic, line 535: This looks quite chaotic. Can we create a `const` for this array up there and reference it here?

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -258,24 +250,32 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
...
-            defaultSortDirection="descending"
-          />
+          <ScrollableTable
+            dependencies={[
+              instance.snapshots,
+              inTabNotification,
+              query,
+              pagination.pageSize,
+              pagination.currentPage,
+            ]}
```
lorumic, line 260: Here too, and in all the other tables, I'd move this dependency array up in a `const` to make it easier to maintain.

src/components/ScrollableTable.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const ref = useRef<HTMLDivElement>(null);
+
+  const updateTBodyHeight = () => {
+    const table = ref.current?.children[0];
+    if (!table || table.children.length !== 2) {
+      return;
+    }
+    const tBody = table.children[1];
+    const above = tBody.getBoundingClientRect().top + 1;
+    const below = 85;
```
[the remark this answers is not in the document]
edlerd: True, I changed this by adding an optional `belowId` as a string of which we now read height + margintop + marginbottom to figure the right value.

src/pages/instances/InstanceList.tsx
edlerd: I think I found a simplification to pass the notification and the filtered object list. The rest is not needed.

src/pages/instances/InstanceSnapshots.tsx
edlerd: same as above, shortened the list.

src/components/ScrollableTable.tsx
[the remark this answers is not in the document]
lorumic: And you know what's the best about this? That incidentally, this fixes one of the 2 issues with the search and filter:

Peek 2023-05-25 23-27

:tada: :clap:

lorumic (approved):

PR #356, opened 2023-05-24
Expand collapsed navigation on hover WD-3186

edlerd:
## Done

- expand collapsed navigation on hvoer

Fixes WD-3186

## QA

1. Run the LXD-UI:
    - check expand / collapse hover / unhover navigation

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -17,6 +17,7 @@ const Navigation: FC = () => {
   const { isRestricted } = useAuth();
   const { menuCollapsed, setMenuCollapsed } = useMenuCollapsed();
   const { project, isLoading } = useProject();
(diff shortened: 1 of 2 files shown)


al-bakalova:
@edlerd A few notes here:

- Let's not merge it right now since there are ongoing "Application layout" efforts to redesign and come up with common side navigation guidelines. As soon as the team is ready, they will be in touch to share the guidance with you and the react Guild
- Looks like the correct behaviour is to slide over the navigation on hover and to push the content only when it is "pinned" as the collapse/expand button actually works as a pin (MAAS)
- The Design team is on it, so my suggestion would be to mark this issue as blocked and to wait for their final specification.

edlerd:
Rejected, we do not want an autocollapsing nav.

PR #360, opened 2023-05-25
Ensure instances are started after creation even with long running create processes WD-4050

edlerd:
## Done

- subscribe to events websocket to get live updates from LXD for any lifecycle changes
- use events to launch an instance that was created and is supposed to start.

Fixes WD-4050

## QA

1. Run the LXD-UI:
    - create or create+start some container and/or VMs and see the notifications work and the desired outcome is reached.

### DIFF
--- a/src/pages/instances/Events.tsx
+++ b/src/pages/instances/Events.tsx
@@ -1,0 +1,51 @@
+import React, { FC, useEffect, useState } from "react";
+import { LxdEvent } from "types/event";
+import { useEventQueue } from "context/eventQueue";
+
+const Events: FC = () => {
+  const eventQueue = useEventQueue();
+  const [eventWs, setEventWs] = useState<WebSocket | null>(null);
+
+  const handleEvent = (event: LxdEvent) => {
+    const eventCallback = eventQueue.get(event.metadata.id);
+    if (!eventCallback) {
+      return;
+    }
+    if (event.metadata.status === "Success") {
+      eventCallback.onSuccess();
+      eventQueue.remove(event.metadata.id);
+    }
+    if (event.metadata.status === "Failure") {
+      eventCallback.onFailure(event.metadata.err ?? "");
+      eventQueue.remove(event.metadata.id);
+    }
+  };
+
+  const connectEventWs = () => {
+    const wsUrl = `wss://${location.host}/1.0/events?type=operation,lifecycle`;
+    const ws = new WebSocket(wsUrl);
+    ws.onopen = () => {
+      setEventWs(ws);
+    };
+    ws.onclose = () => {
+      setEventWs(null);
+    };
+    ws.onmessage = (message: MessageEvent<Blob | string | null>) => {
+      if (typeof message.data !== "string") {
+        console.log("Invalid format on event api: ", message.data);
+        return;
+      }
+      const event = JSON.parse(message.data) as LxdEvent;
+      setTimeout(() => handleEvent(event), 1000); // handle with delay to allow operations to vanish
+    };
+  };
+
+  useEffect(() => {
+    if (!eventWs) {
+      connectEventWs();
+    }
+  }, [eventWs]);
+  return <></>;
+};
+
+export default Events;
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -41,9 +41,7 @@ export const createInstance = (
       body: body,
     })
       .then(handleResponse)
-      .then((data: LxdOperationResponse) => {
-        watchOperation(data.operation, TIMEOUT_120).then(resolve).catch(reject);
-      })
+      .then(resolve)
       .catch(reject);
   });
 };
--- a/src/context/eventQueue.tsx
+++ b/src/context/eventQueue.tsx
@@ -1,0 +1,48 @@
+import React, { createContext, FC, ReactNode, useContext } from "react";
+
+export interface EventQueue {
+  get: (operationId: string) => EventCallback | undefined;
+  set: (
+    operationId: string,
+    onSuccess: () => void,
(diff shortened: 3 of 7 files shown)


src/context/instanceLoading.tsx
```diff
@@ -1,53 +1,44 @@
...
+  type: LoadingType;
+  callback?: () => void;
+}
 
-  const setFinish = (instance: LxdInstance) => {
-    const newMap = new Map(instanceStates);
-    newMap.delete(instance.name);
-    setInstanceStates(newMap);
-  };
+const sharedMap = new Map<string, LoadingEntry>();
```
lorumic, line 31: This needs to go in the state just like it used to. Currently the loading logic is broken - you'll always get `false` for `isLoading`, and the spinner never appears:

Peek 2023-05-29 21-27

lorumic (changes requested):

src/pages/instances/Events.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const ws = new WebSocket(wsUrl);
+    ws.onopen = () => {
+      setEventWs(ws);
+    };
+    ws.onclose = () => {
+      setEventWs(null);
+    };
+    ws.onmessage = (message: MessageEvent<Blob | string | null>) => {
+      if (typeof message.data !== "string") {
+        console.log("Invalid format on event api: ", message.data);
```
lorumic, line 30:
```suggestion
        // TODO: remove this and other console.log calls
        console.log("Invalid format on event api: ", message.data);
```

src/pages/instances/Events.tsx
```diff
@@ -0,0 +1,46 @@
...
+        return;
+      }
+      const event = JSON.parse(message.data) as LxdEvent;
+      setTimeout(() => handleEvent(event), 1000); // handle with delay to allow operations to vanish
```
lorumic, line 34: Setting this timeout doesn't seem to be enough. Most of the times I tried I received the following error:

Screenshot from 2023-05-29 21:15:37

which makes me think that we get the "instance-created" event even earlier than 1 second before the "create" operation ceases to be "running". Ideally, we should check that the status of the "create" operation has transitioned from "running" to "success" - or even better, that the operation cannot be found anymore - before trying to start the instance.

src/pages/instances/Events.tsx
edlerd: This for discovery of errors in the future. I wouldn't want to remove this in the future. I see we added this todo comment to various places where we handle websockets and think it should be removed there as well. Or how would this be useful?

src/pages/instances/Events.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const handleEvent = (event: LxdEvent) => {
+    if (event.metadata.action === "instance-created") {
+      const instanceName = event.metadata.source.split("/").pop() ?? "";
+      const entry = instanceLoading.get(instanceName);
+      if (entry && entry.type === "Creating" && entry.callback) {
+        entry.callback();
+      }
+    }
+  };
```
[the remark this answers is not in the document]
edlerd: The idea is this will handle multiple things in the future. I just added the error case as well. We probably need to organize it a bit better when handling more types, but it should be generic for all types of events.

src/pages/instances/Events.tsx
edlerd: I saw this as well before, but with the latest implementation can't reproduce it any more even on our super slowish demo server. If it still happens, please reopen.

src/pages/instances/Events.tsx
lorumic: I don't understand what's special about websockets to justify these calls to `console.log`, which are discouraged for production code and a bad practice in general. They leak information about the app to the user.
Also, we don't apply the same error logging logic to, say, API calls, so I really don't see any good reason to make an exception here.
What we should really remove is these calls (and not introduce new ones), not the comments reminding us to clean up the code by removing them - sooner (better) or later:
```suggestion
```

edlerd: API calls that fail are prominently shown in the network tab in red. On failures, they are shown in the regular console.

Incoming messages on websockets are quite hidden under a WS tab in the Networks section and are prone to be overlooked. It is better to make the messages visible in the console if they are not handled than to drop them. The messages are not secret, because still discoverable with the debug interface of any decent browser - but a bit hard to find.

lorumic: Main point is: production code is not for debugging. You should not include any `console.log` there. Which is why I introduced those TODO comments in the first place. If you want to keep the debugging code in production code: fine, I'm just pointing out (for the last time) that doing so is not a good practice, and they should be removed if we want to have a clean, production-ready code.

src/pages/instances/Events.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { FC, useEffect, useState } from "react";
+import { LxdEvent } from "types/event";
+import { useEventQueue } from "context/eventQueue";
+
+const Events: FC = () => {
+  const eventQueue = useEventQueue();
+  const [eventWs, setEventWs] = useState<WebSocket | null>(null);
+
+  const handleEvent = (event: LxdEvent) => {
+    const entry = eventQueue.get(event.metadata.id);
```
lorumic, line 10: It took me some time (and a check to the relevant part of the code) to understand/remember what the `entry` was - and I had just checked the code in `eventQueue` shortly before. If one is not aware of the structure of an item in the `eventQueue`, they will have an even harder time. I propose renaming this to something that makes it immediately clear what the entry is about, maybe `eventCallbacks`?

lorumic (approved):
Just one small naming suggestion if you want, apart from that it's ready to go.

PR #363, opened 2023-05-31
Consume search and filter events [WD-3865]

lorumic:
## Done

- Adjust height of the page on search and filter expand
- Close navigation on search and filter panel open

Fixes [WD-3865](https://warthogs.atlassian.net/browse/WD-3865)

## QA

1. Run the LXD-UI:
    - Open instance list page
    - Set many filters so that they overflow the first line
    - Make sure that the list is filtered enough to require a height update to keep the pagination at the bottom (e.g. less than 5 filtered instances)
    - Check that the height update happens as expected
    - Open instance list page on a small screen (width below 620px)
    - Manually expand the side navigation by clicking on the "Menu" button (top-right)
    - With the side navigation expanded, click on the search and filter input
    - Check that when the search and filter panel opens, the side navigation gets automatically collapsed

[WD-3865]: https://warthogs.atlassian.net/browse/WD-3865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -31,8 +31,10 @@ const Navigation: FC = () => {
     e.stopPropagation();
   };
 
-  const collapseOnMediumScreen = () => {
-    setMenuCollapsed(isWidthBelow(820));
+  const collapseOnMediumScreen = (e: Event) => {
+    if (!("detail" in e) || e.detail !== "search-and-filter") {
+      setMenuCollapsed(isWidthBelow(820));
+    }
   };
 
   useEventListener("resize", collapseOnMediumScreen);
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -6,7 +6,6 @@ import classnames from "classnames";
 import Logo from "./Logo";
 import ProjectSelector from "pages/projects/ProjectSelector";
 import ServerVersion from "components/ServerVersion";
-import useEventListener from "@use-it/event-listener";
 import { isWidthBelow } from "util/helpers";
 import { useProject } from "context/project";
 import { useMenuCollapsed } from "context/menuCollapsed";
@@ -22,22 +21,14 @@ const Navigation: FC = () => {
   const { isAuthenticated } = useAuth();
   const softToggleMenu = () => {
     if (isSmallScreen()) {
-      setMenuCollapsed(!menuCollapsed);
+      setMenuCollapsed((prev) => !prev);
     }
   };
 
   const hardToggleMenu = (e: MouseEvent<HTMLElement>) => {
-    setMenuCollapsed(!menuCollapsed);
+    setMenuCollapsed((prev) => !prev);
     e.stopPropagation();
   };
-
-  const collapseOnMediumScreen = (e: Event) => {
-    if (!("detail" in e) || e.detail !== "search-and-filter") {
-      setMenuCollapsed(isWidthBelow(820));
-    }
-  };
-
-  useEventListener("resize", collapseOnMediumScreen);
 
   return (
     <>
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -1,4 +1,4 @@
-import React, { FC, MouseEvent, useState } from "react";
+import React, { FC, MouseEvent } from "react";
 import { NavLink } from "react-router-dom";
 import { Button, Icon } from "@canonical/react-components";
 import { useAuth } from "context/auth";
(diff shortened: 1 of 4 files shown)


al-bakalova:
@lorumic  Oh, this is great! Looks good to me! I wasn't able to identify any issues, good job!

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -91,6 +96,24 @@ const InstanceSearchFilter: FC<Props> = ({ instances, setFilters }) => {
...
+            ".p-search-and-filter__panel"
+          );
+          if (!searchAndFilterPanel) {
+            return;
+          }
+          const isShowing =
+            searchAndFilterPanel.getAttribute("aria-hidden") === "false";
+          if (!menuCollapsed && isShowing && isSmallScreen()) {
+            setMenuCollapsed(true);
+          }
```
edlerd, line 115: I think we can reduce the coupling between components here.

`onPanelToggle` will throw a custom event, similar to the one `onExpandChange` produces. In `menuCollapsed.tsx` we listen to that event and keep all the logic isSmallScreen, and the menuCollapsed / setMenuCollapsed local to it. We can then use that hook exclusively within Navigation and avoid the need for the `MenuCollapsedProvider`, because using the hook in only one place which has the state.

src/components/Navigation.tsx
```diff
@@ -9,30 +9,32 @@ import ServerVersion from "components/ServerVersion";
...
   const softToggleMenu = () => {
     if (isSmallScreen()) {
-      setMenuCollapsed((prev) => !prev);
+      setMenuCollapsed(!menuCollapsed);
     }
   };
 
   const hardToggleMenu = (e: MouseEvent<HTMLElement>) => {
-    setMenuCollapsed((prev) => !prev);
+    setMenuCollapsed(!menuCollapsed);
```
edlerd, line 30: The previous style with the callback is safer, because `menuCollapsed` comes from the wrapping component and is not guaranteed to be up to date, while the `prev` argument the callback provides comes with that guarantee.

src/components/Navigation.tsx
```diff
@@ -9,30 +9,32 @@ import ServerVersion from "components/ServerVersion";
...
 
-  const collapseOnMediumScreen = () => {
-    setMenuCollapsed(isWidthBelow(820));
+  const collapseOnMediumScreen = (e: Event) => {
+    if (!("detail" in e) || e.detail !== "search-and-filter") {
+      setMenuCollapsed(isWidthBelow(820));
+    }
   };
 
   useEventListener("resize", collapseOnMediumScreen);
```
edlerd, line 40: I think the event listener and handler should also move into the `menuCollapsed` hook.

src/util/helpers.tsx
```diff
@@ -159,8 +159,4 @@ export const getUrlParam = (paramName: string, url?: string): string | null => {
 export const defaultFirst = (p1: { name: string }, p2: { name: string }) =>
   p1.name === "default" ? -1 : p2.name === "default" ? 1 : 0;
 
-export const isWidthBelow = (width: number) => {
-  // using the max from both, because there is a bug in chrome, causing a 0 outerWidth for
-  // background tabs: https://bugs.chromium.org/p/chromium/issues/detail?id=719296
-  return Math.max(window.outerWidth, window.innerWidth) < width;
-};
+export const isWidthBelow = (width: number) => window.innerWidth < width;
```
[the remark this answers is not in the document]
lorumic: This was creating a much bigger problem than the small one it was solving. Taking the max of inner and outer width is not a good approach. We're not getting the breakpoint right with this approach, because what we need here is just the `innerWidth`. We don't really care about the outer one. When we're not opening a tab in the background with a small width (which is 99.9% of the use case, I'd say), we are taking the `outerWidth` with this approach, which is consistently bigger than the `innerWidth` (which is the correct one instead, and the one that we should always use). So when our breakpoint is 620, with the "max approach" we'd need to pass in 692 to the function in order to get that breakpoint right. But this will break other browsers (Firefox) because they don't have this bug. You'd need to pass just 620 for them, not 692. Hence my correction. If you have a better idea, that solves the small (negligible) problem that you mentioned without creating the much bigger one that I just described, feel free to propose it, but this logic of taking the max is broken and doesn't make much sense, as @aaryanporwal was also trying to tell us the other day.

src/components/Navigation.tsx
lorumic: Right, I was experimenting one thing and then I forgot to revert this change. It will be back as it was before.

src/util/helpers.tsx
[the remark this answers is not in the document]
edlerd: The `innerWidth` is not affected by the bug and actually the right thing to use, because it also encounters for zoom factor (which might explain the differences you encountered). I don't remember why we back in the day started with `outherWidth` and ran into the chrome bug.

So yeah, let's simplify this as proposed.

edlerd (approved):
LGTM, much cleaner now as before and still working well.

One more suggestion to simplify the `menuCollapsed` hook.

PR #367, opened 2023-06-06
Keep table headers sticky in the image selection modal [WD-4289]

lorumic:
## Done

- Keep table headers sticky in the image selection modal

Fixes [WD-4289](https://warthogs.atlassian.net/browse/WD-4289)

## QA

1. Run the LXD-UI:
    - Create an instance
    - Select image
    - Check that the table headers are sticky in the image selection modal

[WD-4289]: https://warthogs.atlassian.net/browse/WD-4289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/images/ImageSelector.tsx
+++ b/src/pages/images/ImageSelector.tsx
@@ -342,10 +342,7 @@ const ImageSelector: FC<Props> = ({ onClose, onSelect }) => {
             value={query}
           />
           <div className="image-list">
-            <ScrollableTable
-              belowId="modal-bottom-space"
-              dependencies={[images]}
-            >
+            <ScrollableTable dependencies={[images]}>
               <MainTable
                 className="table-image-select"
                 emptyStateMsg={
@@ -361,7 +358,6 @@ const ImageSelector: FC<Props> = ({ onClose, onSelect }) => {
                 sortable
               />
             </ScrollableTable>
-            <div id="modal-bottom-space" />
           </div>
         </Col>
       </Row>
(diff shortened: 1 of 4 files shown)


al-bakalova:
@lorumic This looks good. Thank you!

src/pages/images/ImageSelector.tsx
```diff
@@ -344,20 +345,25 @@ const ImageSelector: FC<Props> = ({ onClose, onSelect }) => {
...
-                  "No matching images found"
-                )
-              }
-              headers={headers}
-              rows={rows}
-              paginate={null}
-              sortable
-            />
+            <ScrollableTable
+              belowId="modal-bottom-space"
```
edlerd, line 349: we don't need the belowId here, do we?

src/sass/_forms.scss
```diff
@@ -140,6 +140,10 @@
 }
 
 .image-select-modal {
+  .p-modal__dialog {
+    overflow-y: hidden;
```
edlerd, line 144: Half of a filter option (i.e. its label) might be visible and with this there is no way to get to the dropdown to use the filter. Shall we remove this bit?

lorumic: This is the problem that style is trying to solve:

image

So, at around ~590px height, we have double scrollbars, on the table and on the modal. I applied that style to avoid that, because the one on the modal is unnecessary and can be hidden. I cannot reproduce/don't understand the issue that you mentioned.

PR #368, opened 2023-06-06
feat(cert) add certificate generation for macOS WD-4079 #331

edlerd:
## Done

- add browser section for macOS to the certificate generation page
- moved the download pfx button into the browser sections

Fixes WD-4079 #331 

## QA

1. Run the LXD-UI:
    - generate a cert with password for macos and test accessing the webui

### DIFF
--- a/src/pages/certificates/BrowserImport.tsx
+++ b/src/pages/certificates/BrowserImport.tsx
@@ -1,13 +1,24 @@
 import React, { FC, useState } from "react";
-import { Col, Row, Tabs } from "@canonical/react-components";
+import {
+  Button,
+  Col,
+  Notification,
+  Row,
+  Tabs,
+} from "@canonical/react-components";
 
 const FIREFOX = "Firefox";
 const CHROME_LINUX = "Chrome (Linux)";
 const CHROME_WINDOWS = "Chrome (Windows)";
 const EDGE = "Edge";
-const TABS: string[] = [FIREFOX, CHROME_LINUX, CHROME_WINDOWS, EDGE];
+const MACOS = "macOS";
+const TABS: string[] = [FIREFOX, CHROME_LINUX, CHROME_WINDOWS, EDGE, MACOS];
 
-const BrowserImport: FC = () => {
+interface Props {
+  sendPfx?: () => void;
+}
+
+const BrowserImport: FC<Props> = ({ sendPfx }) => {
   const [activeTab, handleTabChange] = useState(FIREFOX);
 
   const windowsDialogSteps = (
@@ -27,6 +38,17 @@ const BrowserImport: FC = () => {
     </>
   );
 
+  const downloadPfx = (
+    <li className="p-list__item u-clearfix">
+      Download <code>lxd-ui.pfx</code>
+      {sendPfx && (
+        <div className="u-float-right--large">
+          <Button onClick={sendPfx}>Download pfx</Button>
+        </div>
+      )}
+    </li>
+  );
+
   return (
     <Row>
       <Col size={8}>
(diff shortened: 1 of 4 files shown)


al-bakalova:
@edlerd A small comment on my side:
-  I feel like `The certificate must be protected by password. An empty password will fail to be imported on macOS.` belongs to the very first step as this is where the password is set. I would add it on a new line, after "Create a new certificate".

edlerd:
Right the current hint is too late. I added it also as a help text for the password modal.

Screenshot from 2023-06-07 11-39-47

al-bakalova:
@edlerd It does look good in the modal, that's a good idea. However, I'd still move it to the first point and remove it from the third one, if possible.

edlerd:
> @edlerd It does look good in the modal, that's a good idea. However, I'd still move it to the first point and remove it from the third one, if possible.

I was hesitant to add it to the 1st step, because it is just so prominent, and it will not be relevant for anyone on a different OS. We still need to show it at that time, so macOS users don't skip it accidentally, hence the note in the password modal.

al-bakalova:
@edlerd Makes sense. Thanks for applying the changes

src/pages/certificates/BrowserImport.tsx
```diff
@@ -132,6 +152,33 @@ const BrowserImport: FC = () => {
...
+                The certificate must be protected by password. An empty password
+                will fail to be imported on macOS.
+              </li>
+              {downloadPfx}
+              <li className="p-list__item">
+                Start the Keychain Access app on your Mac, select the login
+                keychain.
+              </li>
+              <li className="p-list__item">
+                Drag the certificate file onto the Keychain Access app.
```
lorumic, line 169: I find this a bit confusing. Maybe instead of "the certificate file" we could explicitly say "the lxd-ui.pfx file"? So that we disambiguate between the .crt and the .pfx.

src/pages/certificates/BrowserImport.tsx
```diff
@@ -132,6 +152,33 @@ const BrowserImport: FC = () => {
...
+              </li>
+              <li className="p-list__item">
+                If you’re asked to provide a name and password, type the name
```
lorumic, line 172: Let's consider this formal writing and do without contractions:
```suggestion
                If you are asked to provide a name and password, type the name
```

src/pages/certificates/BrowserImport.tsx
```diff
@@ -132,6 +152,33 @@ const BrowserImport: FC = () => {
             </ul>
           </div>
         )}
+
+        {activeTab === MACOS && (
+          <div role="tabpanel" aria-label="safari macos">
```
lorumic, line 157: Is this guide for Safari only, or for any macOS browser? It's currently quite unclear what browser(s) this tab is applicable to, and which operating systems the other tabs are applicable to - except the 2 ones that state it explicitly in parentheses.
For example, if I'm using Firefox on macOS, which tab should I select? The "Firefox" one or  the "macOS" one? This is not clear to me at the moment. I thought we decided to have a double selector: a first one for the OS and a second one for the browser. Did I misunderstand anything?

src/pages/certificates/BrowserImport.tsx
```diff
@@ -132,6 +158,36 @@ const BrowserImport: FC = () => {
...
+            <ul className="p-list--divided u-no-margin--bottom">
+              <li className="p-list__item">
+                <Notification severity="caution">
```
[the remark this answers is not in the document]
edlerd: good idea!

src/pages/certificates/BrowserImport.tsx
edlerd: We didn't find a good solution. With two dropdowns, you would need just so many clicks to get to the right browser/os combination. Mixing tabs and a dropdown is a very uncommon pattern.

But I see your point that we are not super clear here. Then again, on macOS there is no cert import in firefox, so I guess users will find the right section even with this ambiguity?

lorumic: I hope so. It's not too important to find a solution now. We can improve this in a future iteration, hopefully after receiving some other good feedback like the one that prompted these changes. :slightly_smiling_face:

PR #369, opened 2023-06-06
Target cluster group or member on instance creation WD-4263

edlerd:
## Done

- add location group/member selector to instance creation

Fixes WD-4263

## QA

1. Run the LXD-UI:
    - instance creation in clustered / unclustered configuration with a group and/or a member targetted

### DIFF
--- a/src/pages/instances/forms/InstanceLocationSelect.tsx
+++ b/src/pages/instances/forms/InstanceLocationSelect.tsx
@@ -1,0 +1,100 @@
+import React, { FC, useState } from "react";
+import { Select } from "@canonical/react-components";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import Loader from "components/Loader";
+import { useSettings } from "context/useSettings";
+import { fetchClusterGroups } from "api/cluster";
+import { FormikProps } from "formik/dist/types";
+import { CreateInstanceFormValues } from "pages/instances/CreateInstanceForm";
+
+interface Props {
+  formik: FormikProps<CreateInstanceFormValues>;
+}
+
+const figureDefaultGroup = (target?: string) => {
+  if (!target?.startsWith("@")) {
+    return "default";
+  }
+  return target.split("@")[1];
+};
+
+const figureDefaultMember = (target?: string) => {
+  if (!target || target.startsWith("@")) {
+    return "";
+  }
+  return target;
+};
+
+const InstanceLocationSelect: FC<Props> = ({ formik }) => {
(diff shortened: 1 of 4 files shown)


edlerd:
Some screens for the clustered backend. In case of no cluster, the two location selectors will not be visible.
The Location members change depending on the group.

Screenshot from 2023-06-06 17-12-09
Screenshot from 2023-06-06 17-12-00
Screenshot from 2023-06-06 17-11-55
Screenshot from 2023-06-06 17-11-51
Screenshot from 2023-06-06 17-11-45
Screenshot from 2023-06-06 17-11-12

al-bakalova:
@edlerd Thanks for sharing! I confirm that that was the logic that we were looking for. These screens look good to me.

src/pages/instances/forms/InstanceLocationSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+          return { label: group.name, value: group.name };
+        })}
+      />
+      <Select
+        id="locationMember"
+        label="Location member"
+        onChange={(e) => setMember(e.target.value)}
+        value={selectedMember}
+        options={[
+          { label: "any", value: "" },
```
lorumic, line 89: Maybe we should not show the "any" option if there are 0 members or 1 member. I did not remember that the group I was selecting had zero members and left "any" selected, and got the "No suitable cluster member could be found" error after trying to create the instance - a tedious postponed feedback. We could show a hint below explaining what the problem is and disable the create buttons.
Even better, we could also filter the `clusterGroup` list above so that we don't include groups that have zero members.

src/pages/instances/forms/InstanceLocationSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+    return "";
+  }
+  return target;
+};
+
+const InstanceLocationSelect: FC<Props> = ({ formik }) => {
+  const defaultGroup = figureDefaultGroup(formik.values.target);
+  const [selectedGroup, setSelectedGroup] = useState(defaultGroup);
+  const defaultMember = figureDefaultMember(formik.values.target);
+  const [selectedMember, setSelectedMember] = useState(defaultMember);
```
lorumic, line 33: These can be moved below the `isClustered` check, right?

src/pages/instances/forms/InstanceLocationSelect.tsx
edlerd: I think we should disable the empty groups, so they can't be selected.

lorumic (approved):

PR #370, opened 2023-06-06
Migration of instance to other cluster members WD-4264

edlerd:
## Done

- show location in configuration section of instance detail page when clustered environment
- add migrate button next to it
- handle instance migration

Fixes WD-4264

## QA

1. Run the LXD-UI:
    - test in clustered and unclustered environments
    - migrate instance to some cluster member


edlerd:
See screens below for design review

Screenshot from 2023-06-06 18-42-19
Screenshot from 2023-06-06 18-41-47
Screenshot from 2023-06-06 18-41-41
Screenshot from 2023-06-06 18-41-35
Screenshot from 2023-06-06 18-41-30

al-bakalova:
@edlerd That's the experience that we re looking for. However, the layout seems a bit off:

- The Migration notifications must be within the tab and not above the tabs as the action belongs to the tab and not to the header details
- I believe, when migrating an instance, we need to chose both the group and the member. Maybe confirm with the Core team
- I would prefer for us to have a slightly bigger migrate modal. It looks too condensed, let's let it breathe a little. I would double the width. That way, we can accommodate for longer instance names
- In the header, why are the status and the actions going to a next line. Looks like there is more than enough space for us to display all the details, the breadcrumb and the actions, on a single row.
- When not in "Edit mode", the Migrate button should be hidden.

edlerd:
@al-bakalova some updates addressing most of the points.

Re "group and the member", that is an open bug on lxd core, currently only member migration is supported.

Screenshot from 2023-06-06 22-18-19
Screenshot from 2023-06-06 22-17-30
Screenshot from 2023-06-06 22-17-17

al-bakalova:
@edlerd Thanks for applying all the changes. I think all the shared screens are good to go.

Were you able to slightly update the width of the Migrate modal?

edlerd:
Yes, gave it a min width - below a screenshot of the recent one.

Screenshot from 2023-06-07 09-36-58

al-bakalova:
@edlerd Perfect, thank you! You are getting a design +1 :)

src/pages/instances/actions/MigrateInstanceBtn.tsx
```diff
@@ -0,0 +1,90 @@
...
+    queryFn: fetchClusterMembers,
+  });
+
+  if (isLoading) {
+    return <Loader />;
+  }
+
+  const handleSuccess = () => {
+    setInTabNotification(
+      success(`Migration finished for instance ${instance}`)
```
lorumic, line 40: I think we should use the `ItemName` component, with the `bold` prop, here and in all the other notifications where we mention the instance name.

src/pages/instances/MigrateInstanceForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <Modal
+      close={close}
+      className="migrate-instance-modal"
+      title={`Migrate instance ${instance}`}
+      buttonRow={
+        <>
+          <Button
+            className="u-no-margin--bottom"
+            type="button"
+            aria-label="cancel reconnect"
```
lorumic, line 50:
```suggestion
            aria-label="cancel migrate"
```

src/pages/instances/actions/MigrateInstanceBtn.tsx
edlerd: In the failures, the name is part of the title which is bold itself, but here we can use it, sure.

src/pages/instances/actions/MigrateInstanceBtn.tsx
```diff
@@ -0,0 +1,90 @@
...
+    migrateInstance(instance, project, target)
+      .then((operation) => {
+        eventQueue.set(operation.metadata.id, handleSuccess, handleFailure);
+        setInTabNotification(info(`Migration started`));
+        closePortal();
+      })
+      .catch((e) => {
+        setInTabNotification(
+          failure(`Migration failed on instance ${instance}`, e)
+        );
```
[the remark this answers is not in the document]
edlerd: They seem, but are not directly compatible. The `handleFailure` uses a slightly different interface. It expects a message argument as a string and turns it into an error. While here, we pass the exception object directly to the failure method.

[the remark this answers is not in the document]
lorumic: But you can make the function compatible to both cases, right?

edlerd:
> Do you know why this is happening? Could it be that the corresponding formik value is not being updated after a query invalidation?

Yes, formik was still having a shadow copy of the instance with the old location. The way you asked here pointed at a easy fix to update that value after the migration finished.

lorumic (approved):

Issue #343, opened 2023-05-12
Support client tokens for authentication


stgraber:
The current TLS support is nice, but it has the downside of pushing users towards generating one client certificate per LXD server.
This then gets them into the annoying authentication flow of having to figure out which certificate to select in their browser.

Instead it may be interesting to support LXD's token based authentication where the user would basically:
 - Hit up /ui
 - UI detects that no client certificate is used (not sure if possible), then offers to generate a PFX for the user
 - UI instructs user to do `lxc config trust add --name some-name` on the LXD server
 - UI takes in the token string provided by the user and adds the browser's certificate to the target LXD's trust store

In this scenario, the user never needs to pass a .crt to the server and they can also rely on a single certificate per browser instead of per server.

edlerd:
Fixed with https://github.com/canonical/lxd-ui/pull/374

PR #374, opened 2023-06-07
Allow certificate adding with tokens WD-4251 #343

edlerd:
## Done

- added alternative to add existing certificate with token

Fixes WD-4251 #343 

## QA

1. Run the LXD-UI:
    - test adding a certificate already in the browser with token

### DIFF
--- a/src/api/certificates.tsx
+++ b/src/api/certificates.tsx
@@ -1,2 +1,19 @@
+import { handleResponse } from "util/helpers";
+
 export const checkAuth = () =>
   fetch("/1.0/certificates").then((response) => response.status !== 403);
+
+export const addCertificate = (token: string) => {
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/certificates`, {
+      method: "POST",
+      body: JSON.stringify({
+        type: "client",
+        password: token,
+      }),
+    })
+      .then(handleResponse)
+      .then(resolve)
+      .catch(reject);
+  });
+};
--- a/src/pages/certificates/CertificateAdd.tsx
+++ b/src/pages/certificates/CertificateAdd.tsx
@@ -1,0 +1,87 @@
+import React, { FC } from "react";
+import { Col, Row } from "@canonical/react-components";
+import { Navigate } from "react-router-dom";
+import { useAuth } from "context/auth";
+import Loader from "components/Loader";
+import CertificateAddForm from "pages/certificates/CertificateAddForm";
+import NotificationRow from "components/NotificationRow";
+
+const CertificateAdd: FC = () => {
+  const { isAuthenticated, isAuthLoading } = useAuth();
+
+  if (isAuthLoading) {
+    return <Loader />;
+  }
+
+  if (isAuthenticated) {
+    return <Navigate to="/ui" replace={true} />;
+  }
+
+  return (
+    <main className="l-main certificate-generate">
+      <div className="p-panel">
+        <div className="p-panel__header is-sticky">
+          <h1 className="p-panel__title">Add existing certificate</h1>
+        </div>
+        <div className="p-panel__content">
+          <NotificationRow />
+          <Row className="u-no-margin--left">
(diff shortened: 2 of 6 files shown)


src/pages/certificates/CertificateMain.tsx
```diff
@@ -32,6 +32,13 @@ const CertificateMain: FC = () => {
         >
           Setup
         </Button>
+        <p>
+          If you have imported a client certificate to your Browser previously,
```
lorumic, line 36:
```suggestion
          If you have imported a client certificate to your browser previously,
```

src/pages/certificates/CertificateMain.tsx
```diff
@@ -32,6 +32,13 @@ const CertificateMain: FC = () => {
...
+          add it to this LXD instance using tokens.
+        </p>
+        <Button appearance="" onClick={() => navigate("/ui/certificates/add")}>
```
lorumic, line 39:
```suggestion
        <Button onClick={() => navigate("/ui/certificates/add")}>
```

src/pages/certificates/CertificateAdd.tsx
```diff
@@ -0,0 +1,87 @@
...
+                        </div>
+                      </div>
+                    </Col>
+                  </Row>
+                </li>
+                <li className="p-stepped-list__item">
+                  <Row>
+                    <Col size={3}>
+                      <h2 className="p-stepped-list__title p-heading--3">
+                        Use token
```
lorumic, line 54: I'm not 100% convinced by this label. Maybe just "Import" as in the button label would be better?

src/api/certificates.tsx
```diff
@@ -1,2 +1,20 @@
...
   fetch("/1.0/certificates").then((response) => response.status !== 403);
+
+export const addCertificate = (token: string) => {
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/certificates`, {
+      method: "POST",
+      body: JSON.stringify({
+        type: "client",
+        password: token,
+        name: "some-name",
```
lorumic, line 13: Is the name mandatory? What's its purpose? (where would it be displayed/stored?)
I wonder if we should set some more appropriate value here... :sweat_smile:

src/pages/certificates/CertificateAddForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Button, Form, Textarea } from "@canonical/react-components";
+import { addCertificate } from "api/certificates";
+import { useNotify } from "context/notify";
+
+const CertificateAddForm: FC = () => {
+  const notify = useNotify();
+  const [token, setToken] = useState("");
+
+  const useToken = () => {
+    addCertificate(token)
```
lorumic, line 11: I would suggest to add some input sanitisation here, because I noticed that the CLI output includes a "Client lxd-ui certificate add token:" line before the token itself, and it's very easy to carelessly copy-paste the whole thing in this textarea and get a "not authorized" error that really doesn't help much to understand what went wrong. I know I'm being the naughty user here, but prevention is better than cure, right? :slightly_smiling_face: 
```suggestion
    const sanitisedToken = token.trim().split(/\r?\n|\r|\n/g).at(-1) ?? "";
    addCertificate(sanitisedToken)
```

src/pages/certificates/CertificateAdd.tsx
edlerd: yeah, import is better.

src/api/certificates.tsx
edlerd: I think we don't need it - let me test it. If I remember correctly, the name when creating the token via cli is used.

src/pages/certificates/CertificateAddForm.tsx
edlerd: yeah, good idea!

lorumic (approved):
Thanks for applying all the changes!

PR #380, opened 2023-06-08
Allow login with oidc WD-4314

edlerd:
## Done

- add oidc login support
- overhaul login page (two versions with and without oidc support by the backend)
- added logout when oidc login
- added oidc keys to the settings

Fixes WD-4314

## QA

1. Run the LXD-UI:
    - check oidc login, see [tutorial](https://github.com/canonical/lxd-ui/wiki/Setup-oidc-login) for the setup


lorumic:
If you could add to the QA steps above the detailed steps that you followed to enable the OIDC auth in the settings, that would be really helpful.

edlerd:
I wrote a [small tutorial](https://github.com/canonical/lxd-ui/wiki/Setup-oidc-login) from memory. Hope all steps are recorded. If not, I used the explanation @stgraber gave in the release video for LXD 5.13 on YouTube.

lorumic:
> I wrote a [small tutorial](https://github.com/canonical/lxd-ui/wiki/Setup-oidc-login) from memory. Hope all steps are recorded. If not, I used the explanation @stgraber gave in the release video for LXD 5.13 on YouTube.

Thanks, I can confirm that nothing is missing! I managed to set it up.

src/pages/login/CertificateAdd.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { useAuth } from "context/auth";
 import Loader from "components/Loader";
-import CertificateAddForm from "pages/certificates/CertificateAddForm";
+import CertificateAddForm from "pages/login/CertificateAddForm";
 import NotificationRow from "components/NotificationRow";
+import { useNotify } from "context/notify";
 
 const CertificateAdd: FC = () => {
   const { isAuthenticated, isAuthLoading } = useAuth();
+  const notify = useNotify();
```
lorumic, line 12: What is this for? It seems unused to me.

src/pages/login/Login.tsx
```diff
@@ -0,0 +1,71 @@
...
+              <p className="u-sv1">Choose your login method</p>
+              <a className="p-button--positive" href="/oidc/login">
+                Login with SSO
+              </a>
+              <h2 className="p-heading--5 u-sv-2">Other methods</h2>
+              <div>
+                Either{" "}
+                <a href="/ui/login/certificate-generate">
+                  create a new certificate
+                </a>
```
lorumic, line 40: Can this be a `Link`?

src/pages/login/Login.tsx
```diff
@@ -0,0 +1,71 @@
...
+              </div>
+              <div>
+                Or{" "}
+                <a href="/ui/login/certificate-add">use existing certificate</a>{" "}
+                that was added to your browser
```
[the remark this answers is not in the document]
lorumic: And also, the `Link` instead of the `a` element.

src/pages/login/Login.tsx
```diff
@@ -0,0 +1,71 @@
...
+              <Button
+                appearance={"positive"}
+                onClick={() => navigate("/ui/login/certificate-generate")}
+              >
+                Create a new certificate
+              </Button>
+              <p>
+                Or{" "}
+                <a href="/ui/login/certificate-add">use existing certificate</a>{" "}
+                that was added to your browser
```
lorumic, line 61: See the other comment, maybe reuse this bit of code so that you only need to update it once?

src/pages/login/Login.tsx
```diff
@@ -0,0 +1,71 @@
...
+  if (isAuthenticated) {
+    return <Navigate to="/ui" replace={true} />;
+  }
+
+  return (
+    <BaseLayout>
+      <Row className="empty-state">
+        <Col size={6} className="col-start-large-4">
+          <Icon name="containers" className="empty-state-icon lxd-icon" />
+          <h1 className="p-heading--4 u-sv-2">LXD Login</h1>
```
lorumic, line 28: Is the "LXD" prefix needed? Seems implicit to me.
```suggestion
          <h1 className="p-heading--4 u-sv-2">Login</h1>
```

lorumic (commented):
Just another couple of high-level comments in addition to the other ones:
1. The "security" icon seems much less relevant now that the side nav entry is called "Login". I think we should use a different one.
2. Please add a "Logout" button in the side nav, to be shown when the user is authenticated with OIDC. Unlike the client certificates, that can be easily managed in the CLI, there is currently only one intricate way to logout, which is to manually clear the cookies in the browser. Let's improve this.

src/pages/login/CertificateAdd.tsx
edlerd: We use it to either show the notification, or the warning, that a cert needs to be imported for this to work.

lorumic: It's not needed for the warning - I've just removed the `notify` and it still displays, since it's independent - and I don't see which other notification we could show in this context.

src/components/Navigation.tsx
```diff
@@ -206,21 +209,40 @@ const Navigation: FC = () => {
...
+                                window.location.reload()
+                              )
+                            }
+                          >
+                            <Icon
+                              className="is-light p-side-navigation__icon"
+                              name="power-off"
+                            />{" "}
+                            Log out
+                          </a>
```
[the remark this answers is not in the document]
edlerd: The button comes with some styling that doesn't match the navigation, hence using an anchor here.

edlerd:
> * The "security" icon seems much less relevant now that the side nav entry is called "Login". I think we should use a different one.

Replaced it with a profile one, that is commonly used for login

> * Please add a "Logout" button in the side nav, to be shown when the user is authenticated with OIDC. Unlike the client certificates, that can be easily managed in the CLI, there is currently only one intricate way to logout, which is to manually clear the cookies in the browser. Let's improve this.

There is a logout button, but it only works with latest/edge version of lxd. Because the option to see if the login method was oidc or tls was just added last week.

src/pages/login/CertificateAdd.tsx
edlerd: When either not having a cert in the browser (or select) or using an invalid token, there will be an error message.

lorumic: Right, I see the use case now. Thanks.

lorumic (approved):
Thanks for applying the changes. There is also this `pkce` cookie that seems to be stored in the context of an OIDC login. I wonder if that should be cleared as well? Leaving it there might be a security leak.

lorumic:
There is also the vertical spacing in the side nav to be updated maybe?

image

edlerd:
> Thanks for applying the changes. There is also this [`pkce`](https://oauth.net/2/pkce/) cookie that seems to be stored in the context of an OIDC login. I wonder if that should be cleared as well? Leaving it there might be a security leak.

I forwarded that question to the LXD-UI channel. If, it should be cleared by the backend route as it is http only cookie.

PR #383, opened 2023-06-09
Add shortcuts for common keyboard combinations #381

### LINKED ISSUE #381: UI should allow to pass "ctrl-alt-del" combination
When running Windows VM it is necessary to be able to pass "ctrl-alt-del" key combination in order to get to the login screen.

edlerd:
## Done

- add shortcuts to the console view of an instance

Fixes #381 

## QA

1. Run the LXD-UI:
    - test keyboard shortcuts in console view

### DIFF
--- a/src/lib/spice/src/inputs.js
+++ b/src/lib/spice/src/inputs.js
@@ -190,22 +190,45 @@ function handle_keyup(e)
     e.preventDefault();
 }
 
+function send_key(sc, keyCode)
+{
+    var msg = new Messages.SpiceMiniData();
+    var key = new Messages.SpiceMsgcKeyDown();
+
+    key.code = keyCode;
+    msg.build_msg(Constants.SPICE_MSGC_INPUTS_KEY_DOWN, key);
+    sc.inputs.send_msg(msg);
+
+    key.code = 0x80|keyCode;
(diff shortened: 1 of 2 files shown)


lorumic:
There's a fundamental UX problem here: once you go "Fullscreen" - and you're likely to do that immediately on a Windows VM since the mouse tracking is completely broken otherwise - you don't have access to the shortcuts anymore. Which means that each and every time you want to use one, you need to exit full screen, select the shortcut, and go full screen again.
What I have in mind is something similar to what TeamViewer does for example. In fullscreen mode, you move the pointer to the upper edge and an auto-hiding toolbar with controls appears. Would it be possible to do something like that here? Make some element overlay the spice wrapper - with CSS position - and making it visible only when the pointer is in some upper region of the full-screen window?
It can also be postponed if it takes too much time, and we want to release a quick fix.

lorumic (approved):
See above for a possible improvement. In general, the VGA console still has some major issues on low resolution screens - for Windows VMs at least.

edlerd:
The overlay is a good idea. Though hard to get right. Because it easily gets too often into the way if it shows to soon. Finding a good balance for it might take a bit bigger initiative. For now let's unblock the windows login.

lorumic:
Sure, sounds reasonable to me.

PR #386, opened 2023-06-14
Add management of networks

edlerd:
## Done

- add network creation (for bridges, ovn pending)
- add rename networks
- add remove networks
- add edit networks
- change style of instance and profile creation and editing forms 

## QA

1. Run the LXD-UI:
    - test network crud with clustered and non-clustered backend
    - test instance and profile forms 
    - test network overview


src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -0,0 +1,713 @@
...
+                        label: "Openvswitch",
+                        value: "openvswitch",
+                      },
+                    ]}
+                  />
+                  <Input
+                    {...getFormProps("bridge_external_interfaces")}
+                    type="text"
+                    label="Bridge external interfaces"
+                    help="Comma-separated list of unconfigured network interfaces to include in the bridge"
```
lorumic, line 332: I think all the "comma-separated list"s should be textareas. This is what we also do in the "Networks" section of a project configuration. We should keep a consistent design with that.

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -0,0 +1,713 @@
...
+                      options={optionYesNo}
+                      help="Whether to route traffic in and out of the bridge"
+                    />
+                  </>
+                )}
+            </>
+          )}
+          {[USER, ADVANCED_OVN].includes(section) && (
+            <>
+              <Label>User-provided free-form key/value pairs</Label>
```
lorumic, line 635: This "User" section should be removed now that we have added the YAML configuration section. Allowing the user to add free-form key-value pairs is not specific to networks. It's a feature common with instances, profiles, and whatnot. So I don't see why this should be here considering that we don't have it for instances and profiles, and also that with the YAML configuration section you can allow the user to achieve exactly the same result.

lorumic (commented):
See 2 high-level, UX-related comments below.
Also, in general, I think we should try to keep a consistent design and look-and-feel across pages and features. This is currently not the case. See the screen captures below:

Edit a project:

Peek 2023-06-23 16-45

Edit a network:

Peek 2023-06-23 16-46

I think the difference in terms of UX patterns is quite apparent.

edlerd:
> Also, in general, I think we should try to keep a consistent design and look-and-feel across pages and features. This is currently not the case. See the screen captures below:

I am unsure about this one. We started the table / override logic because of inheriting values from profiles to instances. We don't have inheritance on networks.

Fixed the other high level issues.

src/pages/networks/forms/NetworkFormMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+
+  const hasName = formik.values.name.length > 0;
+  const hasOvnTypeMissingUplink =
+    formik.values.type === "ovn" && (formik.values.network?.length ?? 0) < 1;
+  const disableReason = hasName
+    ? hasOvnTypeMissingUplink
+      ? "Please select an uplink network"
+      : undefined
+    : "Please enter a network name to enable this section";
```
lorumic, line 37: I like this feature, it makes a lot of sense. However, my proposal, to make it more reliable, would be to bind this not only to "any" name being entered in the name field, but to a valid name, so that we don't have an inconsistent state where we can input advanced config but cannot create the network because the current name is not valid. Also, shouldn't we apply the same logic to profile creation as well?

lorumic:
There are still quite a few design problems/inconsistencies (in my opinion at least):

1) You might not want to be forced to click the pencil button to see the inherited value. It's actually good to show it to the user first, so that they know what they're inheriting, and can decide if they actually want to override it or not. Of course this reasoning applies to instances and profiles too, if there's a defined inherited value that can be shown to the user.

image

2) All input fields of the network main conf have a width that doesn't match that of the (comparable) input fields in profile and instance main confs:

image

image

3) It's confusing that you're in edit mode by default when switching to the configuration tab of a network, but the same doesn't happen for profiles or instances, where you have to click an "edit" button instead.

4) We have "Update" as button label for network edit, and "Save changes" for instances and profiles.

5) I'm not sure the middle alignment of the "X" button works well in all cases. Maybe it's safer to top-align it instead?

image

src/pages/networks/forms/NetworkFormMenu.tsx
edlerd: I add it on the profile creation as well.
I'd avoid validating the name to keep it simple. If the name is invalid, the input for it should show the validation error.

edlerd:
Addressed all the points 1-5 mentioned by @lorumic above.

src/pages/networks/forms/ConfigurationRow.tsx
```diff
@@ -0,0 +1,150 @@
...
+export const getConfigurationRow = ({
+  formik,
+  name,
+  label,
+  children,
+  defaultValue,
+  disabled = false,
+}: Props): MainTableRow => {
+  const values = formik.values as unknown as Record<string, string | undefined>;
+  const isOverridden = values[name] !== undefined;
```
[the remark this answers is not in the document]
edlerd: Override should not take the LXD default into consideration. Because it is setting the value independent of the default. I think this is quite common, that users can set a config value even though it matches the default.

For the screen capture of networks: That is expected, an auto ip address is assigning an actual IP on creation. The defaults are written into the network config by the backend, I am not sure why that is done for `ipv4/6.nat`.

src/pages/networks/NetworkDetailOverview.tsx
```diff
@@ -0,0 +1,184 @@
...
+                <td>
+                  {humanFileSize(networkState?.counters.bytes_received ?? 0)} (
+                  {networkState?.counters.packets_received} packets)
+                </td>
+              </tr>
+              <tr className="list-wrapper">
+                <th className="p-muted-heading">TX</th>
+                <td>
+                  {humanFileSize(networkState?.counters.bytes_sent ?? 0)} (
+                  {networkState?.counters.packets_sent} packets)
```
lorumic, line 112: These should be guarded against `null`ish values, providing a suitable placeholder:

image

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -0,0 +1,732 @@
...
+      <Row className="form-contents" key={section}>
+        <Col size={12}>
+          {section === MAIN_CONFIGURATION && (
+            <React.Fragment>
+              <Row>
+                <Col size={8}>
+                  <Select
+                    {...getFormProps("type")}
+                    label="Type"
+                    help="Bridge (FAN) is only available on ubuntu, OVN needs to be configured to be available, Macvlan,  SR-IOV and Physical not available in the UI"
```
lorumic, line 191: As discussed, this help text can be confusing for inexperienced users who are not on a clustered server, tricking them into thinking that they just need to configure OVN in order to get the option enabled, which is currently not the case. The help text should reflect the actual restriction in place, which is the `network.ovn.northbound_connection` config being set.
Also, maybe a "Learn more" link which goes to this [page](https://documentation.ubuntu.com/lxd/en/latest/howto/network_ovn_setup/#set-up-a-lxd-cluster-on-ovn) can be helpful here.

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -0,0 +1,732 @@
...
+                  />
+                  {formik.values.type === "ovn" && (
+                    <Select
+                      {...getFormProps("network")}
+                      label="Uplink"
+                      help="Uplink network to use for external network access"
+                      options={getNetworkOptions()}
+                      required
+                      disabled={formik.values.readOnly}
+                    />
```
lorumic, line 261: It seems that with the CLI you can create an OVN network without explicitly passing the uplink parameter - see step 7 [here](https://documentation.ubuntu.com/lxd/en/latest/howto/network_ovn_setup/#set-up-a-lxd-cluster-on-ovn). It's probably picking some existing network by default.
I wonder if we should replicate the same approach in the UI, setting a default option for this dropdown, that the user can then change if they want, of course.

lorumic (commented):
This is getting in good shape. A few ideas for improvements follow.

lorumic:
The "Clear override" button can end up being not visible on medium/small screens:

image

(width is ~1240px in this case)

lorumic:
> The "Clear override" button can end up being not visible on medium/small screens:

And similarly, this is also problematic - but I think you're already aware of it:

image

src/pages/networks/NetworkDetailOverview.tsx
edlerd: oh good catch.

src/pages/networks/forms/NetworkForm.tsx
edlerd: Yes, the uplink is not a required parameter, I'll make it optional.

src/sass/_forms.scss
```diff
@@ -148,6 +159,17 @@
...
+  .override-form {
+    align-items: flex-end;
+    flex-direction: column-reverse;
+
+    > :first-child {
+      width: 100%;
+    }
+  }
+}
+
```
lorumic, line 172: Is this supposed to do anything about the width of the YAML editor being too small on medium screens? Because I still see it as small as it was before:

image

lorumic:
It seems that I can edit a non-managed network:

Peek 2023-08-08 14-44

Shouldn't the "Edit network" button be disabled in this case? Or even the whole "Configuration" tab inaccessible?

src/sass/_forms.scss
edlerd: On the demo server it looks all right, did you maybe need to pull?

lorumic: Right, apparently I pulled before the latest update and missed it.

lorumic (commented):
So far so good, I don't have other QA comments for now.

piperdeck (changes requested):
# 1

In the "Inherited column, I'd like to draw the eye toward the value itself a bit more, rather than the override source.

To do that, I suggest formatting the override source data in the same manner as the Configuration support text to its right:

Pasted image 20230828155446

# 2

A networks IPv4 and IPv6 addresses should probably be monospace-formatted in the Networks list view so that their octets line up and can be compared:

Pasted image 20230828155906

# 3

I found what might be an inconsistency unless I'm misunderstanding something here.

For some config values, if nothing at all is being inherited, the column is simply empty:

| Figure A |
| -------- |
|   Pasted image 20230828164843   |

For other config values, the column will show no value, but still say that it's inheriting from LXD:

| Figure B                             |
| ------------------------------------ |
|  Pasted image 20230828164951   |

And for yet other values, the column indicates a null value using a dash, and reports that it is inherited from LXD.

| Figure C                             |
| ------------------------------------ |
| Pasted image 20230828165032|

This prompts me to think about the different ways a field might end up being null, and how we should represent that in the UI.

Let's say we're creating an object O based on template T inside of project P:

```mermaid
flowchart TD

P --> T --> O
```

Now say there is a configuration `config` whose value is not specified in either P or T. When we create O, that config should not be considered to be inherited, because there does not exist a specification for it anywhere up the chain from O. Therefore it should show completely blank in the Inherited Column as in my first screenshot Figure A:

```mermaid
flowchart LR
subgraph I[Null, no override]
	direction TB
	P["P\nconfig="] --> T["T"] --> O["O"]
end

I --> L[Inherited column for config should be empty]
```

But suppose, on the other hand, that the config was set to `config=1` in P, and then overridden to null in T. In this case, the inherited column should have some visual indication of a null value (like the dash used), and report which parent it inherited that null value from as in Figure C:

```mermaid
flowchart LR
subgraph I[Null from override]
	direction TB
	P["P\nconfig=1"] --> T["T\nconfig="] --> O["O"]
end

I --> L[Inherited column for config should be:\n--\nFrom: T]
```

In either case, I think we should avoid Figure B, where e.g. "From: LXD" appears with no indication of a null value whatsoever.

However, I invite discussion on this because there are probably edge cases I haven't thought about.

edlerd:
Thank you for the review @piperdeck 
I agree with 1 and 3 - those changes were just implemented and are visible on the demo server now.

For 2 I am a bit hesitant. Should we always display IPs in a mono font? There is also the Instance list, instance detail page and some more places with IPs. Attaching a screen below with the network table with mono font - is that the intended result?

Screenshot from 2023-08-29 10-13-19

edlerd:
Hey @piperdeck as discussed, the defaults for memory/cpu are now adjusting to the chosen instance type of container or VM.

I also kept the font for the IPs as is.

Please have another pass on the changes, looking forward to your feedback :)

piperdeck (approved):
The changes look good to me!

PR #389, opened 2023-06-14
Make the "Delete project" button consistent with the other delete buttons in the app

lorumic:
## Done

- Adjusted position and appearance/responsiveness of the "Delete project" button.

## QA

1. Run the LXD-UI:
    - Visit the project configuration page.
    - Check the new position and appearance/responsiveness of the "Delete project" button.

### DIFF
--- a/src/pages/projects/actions/DeleteProjectBtn.tsx
+++ b/src/pages/projects/actions/DeleteProjectBtn.tsx
@@ -21,6 +21,16 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
   const [isLoading, setLoading] = useState(false);
   const navigate = useNavigate();
 
+  const getHoverText = () => {
+    if (project.name === "default") {
+      return "The default project cannot be deleted";
+    }
+    if (!isProjectEmpty(project)) {
+      return "Non-empty projects cannot be deleted";
+    }
+    return "Delete project";
+  };
+
   const handleDelete = () => {
     setLoading(true);
     deleteProject(project)
@@ -43,7 +53,7 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
 
   return (
     <ConfirmationButton
-      onHoverText="Delete project"
+      onHoverText={getHoverText()}
       toggleAppearance={isDeleteIcon ? "base" : "default"}
       className="u-no-margin--bottom"
       isLoading={isLoading}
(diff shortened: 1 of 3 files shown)


src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -41,26 +43,24 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
...
           <ItemName item={project} bold />?{"\n"}This action cannot be undone,
           and can result in data loss.
         </>
       }
-      confirmButtonLabel={
-        project.name === "default"
-          ? "The default project can't be deleted"
-          : "Delete"
-      }
+      confirmButtonLabel="Delete"
```
edlerd, line 59: I saw yesterday in another app they used "Delete permanently" which I think is better and can replace our extra sentence "This action cannot be undone, and can result in data loss". Wdyt, shall we adopt it here and in other delete buttons?

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -41,26 +43,24 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
 
   return (
     <ConfirmationButton
+      onHoverText="Delete project"
```
edlerd, line 46: In case the button is disabled, this should give the reason:
```
project.name === "default" ? "The default project can't be deleted" : (!isProjectEmpty(project) ? "Non empty projects can't be deleted" : "Delete project")
```

edlerd (commented):
Thanks for unifying this. Some ideas to improve it further below.

src/pages/projects/actions/DeleteProjectBtn.tsx
lorumic: I don't have a strong opinion for either, but I think the extra sentence is more noticeable than a "Delete permanently" button or title. Anyway, since it involves updating all the delete buttons, it's probably better suited for a separate effort.

src/pages/projects/actions/DeleteProjectBtn.tsx
lorumic: Done, but the same should be done also for profiles and instances. I'll create a ticket for that.

edlerd (approved):
Nice, thanks :)

PR #395, opened 2023-06-21
Use en-gb date language locale for snapshot expiry form

edlerd:
## Done

- fixed the date format on the expiry input for snapshot creation

## QA

1. Run the LXD-UI:
    - check snapshot creation/edit date picker

### DIFF
--- a/src/pages/instances/actions/snapshots/SnapshotForm.tsx
+++ b/src/pages/instances/actions/snapshots/SnapshotForm.tsx
@@ -98,6 +98,7 @@ const SnapshotForm: FC<Props> = ({
               id="expirationDate"
               name="expirationDate"
               type="date"
+              lang="en-GB"
               label="Expiry date"
               min={getTomorrow()}
               onChange={formik.handleChange}


lorumic:
What's the rationale behind forcing a locale to every user in the world?

edlerd:
> What's the rationale behind forcing a locale to every user in the world?

I didn't realize my setting is not UK English. I guess then it's better to keep this adopting to user prefs than setting the language for all.

lorumic:
> I didn't realize my setting is not UK English. I guess then it's better to keep this adopting to user prefs than setting the language for all.

Sorry, I don't get this. What's the problem of having one's own local date format displayed here?

edlerd:
> > I didn't realize my setting is not UK English. I guess then it's better to keep this adopting to user prefs than setting the language for all.
> 
> Sorry, I don't get this. What's the problem of having one's own local date format displayed here?

I meant this should be closed, because adopting to the locale setting is preferred. Just a previous confusion on my browsers locale (which was us, hence opening the PR, but setting it to uk is the better fix).

PR #403, opened 2023-07-05
Update link targets for docs and discuss

edlerd:
## Done

- updated link targets in main navigation

## QA

1. Run the LXD-UI:
    - check main navigation links that are changed

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -254,7 +254,7 @@ const Navigation: FC = () => {
                   <li className="p-side-navigation__item">
                     <a
                       className="p-side-navigation__link"
-                      href="https://linuxcontainers.org/lxd/docs/latest/"
+                      href="https://documentation.ubuntu.com/lxd/en/latest/"
                       target="_blank"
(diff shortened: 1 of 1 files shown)


src/components/Navigation.tsx
```diff
@@ -269,7 +269,7 @@ const Navigation: FC = () => {
                   <li className="p-side-navigation__item">
                     <a
                       className="p-side-navigation__link"
-                      href="https://discuss.linuxcontainers.org/"
+                      href="https://discourse.ubuntu.com/c/lxd/126"
```
lorumic, line 272: I guess this can be:
```suggestion
                      href="https://discourse.ubuntu.com/c/lxd"
```
And then that would redirect automatically?

edlerd: The canonical url is https://discourse.ubuntu.com/c/lxd/126 so I guess we should use it directly instead of relying on the redirict.

edlerd: That is not right, I just changed it.

lorumic (approved):

PR #408, opened 2023-07-07
fix(pagination) avoid passing pagination props to div

edlerd:
## Done

- fix errors in console on pagination usages, we shouldn't forward pagination props to the HTML div

## QA

1. Run the LXD-UI:
    - check pagination works as expected
    - check the console is clean of errors from the pagination component on the instance list

### DIFF
--- a/src/components/Pagination.tsx
+++ b/src/components/Pagination.tsx
@@ -2,6 +2,7 @@ import { Button, Icon, Input, Select } from "@canonical/react-components";
 import React, { FC, HTMLAttributes, useEffect, useState } from "react";
 import { paginationOptions } from "util/pagination";
 import useEventListener from "@use-it/event-listener";
+import { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
 
 const figureSmallScreen = () => {
   const descriptionElement = document.getElementById("pagination-description");
@@ -20,6 +21,10 @@ type Props = {
   totalCount: number;
   totalPages: number;
   keyword: string;
+  pageData: MainTableRow[];
+  itemsPerPage: number;
+  totalItems: number;
+  updateSort: (sort?: string | null) => void;
 } & HTMLAttributes<HTMLDivElement>;
 
 const Pagination: FC<Props> = ({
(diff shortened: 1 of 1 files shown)


src/components/Pagination.tsx
```diff
@@ -20,6 +21,10 @@ type Props = {
   totalCount: number;
   totalPages: number;
   keyword: string;
+  pageData: MainTableRow[];
+  itemsPerPage: number;
+  totalItems: number;
+  updateSort: (sort?: string | null) => void;
```
lorumic, line 27: I have a feeling these should be set as optional. Or at least, my IDE complains if they're not.
```suggestion
  pageData?: MainTableRow[];
  itemsPerPage?: number;
  totalItems?: number;
  updateSort?: (sort?: string | null) => void;
```

src/components/Pagination.tsx
```diff
@@ -31,6 +36,10 @@ const Pagination: FC<Props> = ({
   totalCount,
   totalPages,
   keyword,
+  pageData: _pageData,
+  itemsPerPage: _itemsPerPage,
+  totalItems: _totalItems,
+  updateSort: _updateSort,
```
lorumic, line 42: If we don't use them, maybe we can just avoid adding them here?
```suggestion
```

edlerd: It is nice to be able to spread them like `<Pagination {...pagination} ` as in `InstanceList`.

edlerd: They are not picked up by the divProps which pick up all the props not listed above. That's the reason I added them. Yes they are not used, hence renaming them to the _ prefix.

lorumic: So passing them or not to the props object doesn't change anything, because they're not being used inside the FC anyway. That's why I was proposing the change above.

edlerd: Those 4 props need to be avoided to fall into the divProps, not sure how that would be done otherwise.

lorumic: Ok, I understand the problem we're trying to solve here, but I think this solution is not very robust - ideally, we should use the return type from react-components' `usePagination` to prevent its fields from being passed to the `divProps`. This way, if that ever changes, we don't need to manually add/exclude fields here. However that object doesn't have its dedicated type, so I can't think of a better way to achieve that at the moment - except maybe declaring the type upstream and then update this later.

lorumic (approved):

Issue #422, opened 2023-07-28
Cannot edit any configuration option with unsupported network config


lathiat:
Hi Team,

I have LXD VMs created using the MAAS KVM Pod functionality. It has 3 network interfaces, bridged.

When I try edit the instance configuration in LXD-UI and attempt to change the CPU core count, it fails to save with the error:

> Instance update failed
> Invalid devices: Device validation failed for "eth0": Failed loading device "eth0": Unsupported device type

Despite having not edited that panel. It seems it tries to persist all config options even if not edited.

If I switch to the Networks panel, the network names (eth0/eth1/eth2) show but each "network device" dropdown has a list of all the interfaces on my system (including vsw0/vsw1, which are just linux bridges), but nothing is selected (hence the error, the text shows "Select option). Despite vsw0 being in the list.

Here is the network config from "lxc config show". In case it matters reproduction wise, this VM is in a non-default 'maas' project. 
I guess this bug is in two parts
- Firstly ideally it would not try to persist unchanged configuration particularly if it may have some unsupported feature (there are many ways to configure networks that don't seem currently supported in the UI). 
- Secondly I guess this basic bridged network configuration should work

Versions:
```
LXD snap:
tracking:     latest/stable/ubuntu-22.04
installed:          5.15-002fa0f             (25112) 181MB -

MAAS: 3.3.4
```

```
devices:
  eth0:
    boot.priority: "1"
    name: eth0
    nictype: bridged
    parent: vsw0
    type: nic
  eth1:
    name: eth1
    nictype: bridged
    parent: vsw1
    type: nic
  eth2:
    name: eth2
    nictype: bridged
    parent: vsw1
    type: nic
```

edlerd:
Thank you for reporting. This is a bug, the UI currently does not handle custom network configurations for instances correctly. We'll have it fixed shortly.

In the meantime, you can use the YAML editor to do any changes to those VMs, which will not alter the custom networks unless you edit the YAML code for them.

edlerd:
Fix landed in main next LXD 5.17 should contain it.

PR #424, opened 2023-07-28
fix instance and profile editing with custom networks applied

edlerd:
## Done

- When instances or profiles have custom networks applied, the guided forms cannot display the information sufficiently. Furthermore the payload for custom networks must not be altered through the guided forms. We treat such networks now as custom and do not allow editing them, but display "custom networks" instead.

Fixes WD-5384 #422 

## QA

1. Run the LXD-UI:
    - edit instance and profile networks
    - check with a custom network (can be set via the YAML input) like the one below. The editing of other parts of a profile or instance with such a custom network should succeed and not alter the custom network.
```
  eth0:
    boot.priority: '2'
    name: eth0
    nictype: bridged
    parent: lxcbr
    type: nic
```

## Screenshots

Screenshot from 2023-07-28 13-22-44
Screenshot from 2023-07-28 13-22-39
Screenshot from 2023-07-28 13-22-34

### DIFF
--- a/src/pages/instances/forms/NetworkForm.tsx
+++ b/src/pages/instances/forms/NetworkForm.tsx
@@ -19,6 +19,7 @@ import { EditInstanceFormValues } from "pages/instances/EditInstanceForm";
 import { getConfigurationRowBase } from "pages/instances/forms/ConfigurationRow";
 import Loader from "components/Loader";
 import { figureInheritedNetworks } from "util/instanceConfigInheritance";
+import { CustomNetworkDevice } from "util/formDevices";
 
 interface Props {
   formik: SharedFormikTypes;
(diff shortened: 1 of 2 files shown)


src/util/formDevices.tsx
```diff
@@ -45,8 +52,23 @@ export const formDeviceToPayload = (devices: FormDevice[]) => {
 export const parseDevices = (devices: LxdDevices): FormDevice[] => {
   return Object.keys(devices).map((key) => {
     const item = devices[key];
+
+    const isCustomNetwork =
+      item.type === "nic" &&
+      Object.keys(item).filter(
+        (key) => !["type", "name", "network"].includes(key)
+      ).length > 0;
```
lorumic, line 60: I think this can be simplified to:
```suggestion
    const isCustomNetwork =
      item.type === "nic" &&
      Object.keys(item).some(
        (key) => !["type", "name", "network"].includes(key)
      );
```

src/pages/instances/forms/NetworkForm.tsx
```diff
@@ -111,61 +112,66 @@ or remove the originating item"
...
-                    type="button"
-                    appearance="base"
-                    aria-label="delete network"
-                    hasIcon
-                  >
-                    <Icon name="delete" />
-                  </Button>
+            value:
+              device.type === "custom-nic" ? (
+                "custom network"
```
lorumic, line 131: Maybe we can have something more explicit than this? Even just adding an info icon next to this text, which shows a tooltip with something like "A custom network can be edited only from the YAML configuration" would make things clearer for the user?

src/pages/instances/forms/NetworkForm.tsx
```diff
@@ -111,61 +112,66 @@ or remove the originating item"
...
+                      name={`devices.${index}.name`}
+                      id={`networkName${index}`}
+                      onBlur={formik.handleBlur}
+                      onChange={formik.handleChange}
+                      value={
+                        (formik.values.devices[index] as LxdNicDevice).name
+                      }
+                      type="text"
+                      placeholder="Enter name"
+                    />
```
lorumic, line 159: I would still show this input field - as disabled - for custom networks. It's odd that currently we show the network name of a custom network in read mode but we hide it in edit mode.

edlerd: True - for custom networks, we can keep the network name visible.

lorumic (approved):
Thanks for applying the changes!

PR #430, opened 2023-08-09
Use release title for displayal #429

### LINKED ISSUE #429: When selecting base image, Ubuntu images should show 22.04 version number, not just jammy
When creating a VM and you click "Select base image", the offered Ubuntu releases simply have the shortname (e.g. focal, jammy, lunar). Ideally it would also display the release version (e.g. 22.04, for jammy). It would be even better if it indicated which as an LTS release.

In various products or places, we commonly interchange whether 22.04 or Jammy is used, however we have recently used the version number in some new places. For example, juju 3.x uses Ubuntu 'bases' of ubuntu@22.04, deprecated the older method of specifying 'series: jammy'. Ideally we'd have at a minimum consistency but realistically just display both in as many scenarios as possible.

While I work on Ubuntu every day and I mostly have the version to release name memorised, I struggle with this in other projects, for example OpenStack. I am sure the same happens for people using Ubuntu.

Screenshot 2023-08-09 at 1 29 03 pm

edlerd:
## Done

- changed field to display and filter release in image selection modal

Fixes #429 

## QA

1. Run the LXD-UI:
    - open the launch image / select image modal
    - check the release column is correctly showing the number and "LTS" for Ubuntu, i.e. "22.04 LTS" and its content is still reasonable for other distros.


src/pages/images/ImageSelector.tsx
```diff
@@ -264,7 +264,7 @@ const ImageSelector: FC<Props> = ({ onClose, onSelect }) => {
                 setRelease(v.target.value);
               }}
               options={getOptionList(
-                (item) => item.release,
+                (item) => item.release_title,
```
lorumic, line 267: By and large this should work, because we already have the shortname in the alias column, but that's not always the case, e.g.:

image

I think the original request from @lathiat was to show both. In this case the shortname "jammy" is not shown, could that be a problem?

src/pages/images/ImageSelector.tsx
```diff
@@ -170,7 +170,7 @@ const ImageSelector: FC<Props> = ({ onClose, onSelect }) => {
             onClick: selectImage,
           },
           {
-            content: item.release,
+            content: item.release_title,
```
lorumic, line 173: By and large this should work, because we already have the shortname in the alias column, but that's not always the case, e.g.:

image

I think the original request from @lathiat was to show both. In this case the shortname "jammy" is not shown, could that be a problem?

edlerd: He mentioned juju moved away from the short name to the Version plus LTS. I think the current way is good, showing both in the column would break into two lines and looks not right on smaller screens.
I think all users will be able to pick the right release from the version and LTS string.

lorumic: I read:

> Ideally it would **also** display the release version

and:

> Ideally we'd have at a minimum consistency but realistically **just display both in as many scenarios as possible**.

Which is not what we're doing here, so if that's not fixing the original issue, it would be better to know it in advance. Let's wait a confirmation from the OP.

edlerd: Iterating on this a bit more - we can use the empty variant column for displaying the code name. I think it looks best this way, though we are bending the meaning of columns a bit. But seems like the best compromise to 
1. Show both the version and the short name
2. Show them in correct order like Ubuntu 22.04 LTS jammy
3. Avoid spacing issues on smaller screens

lorumic (approved):
This is a nifty workaround. I like it!

PR #431, opened 2023-08-09
Update local EmptyStates to react-components [WD-5070]

aaryanporwal:
## Done

- [List of work items including drive-bys]

## QA

1. Run the LXD-UI:

## Screenshots


src/pages/storage/StorageList.tsx
```diff
@@ -130,29 +137,38 @@ const StorageList: FC = () => {
...
-                />
-              </>
-            </EmptyState>
+            <Row className="empty-state">
+              <Col size={6} className="col-start-large-4">
+                <EmptyState
+                  image={
+                    <Icon
+                      name="pods"
+                      className="empty-state-icon empty-storage-icon"
```
lorumic, line 146: We want to give up this approach because it creates a lot of CSS classes, and just for the purpose of slightly changing the colour of the icon. You can achieve the same with [opacity](https://github.com/canonical/anbox-cloud-dashboard/blob/main/templates/sass/styles.scss#L282) and that allows you to have a single shared CSS class for all the icons. Please check how this is done in Anbox dashboard.

src/pages/storage/StorageList.tsx
```diff
@@ -130,29 +137,38 @@ const StorageList: FC = () => {
...
-                  </a>
-                </p>
-                <AddStorageBtn
-                  project={project}
-                  className="empty-state-button"
-                />
-              </>
-            </EmptyState>
+            <Row className="empty-state">
+              <Col size={6} className="col-start-large-4">
```
lorumic, line 141: I have given up this `Row` and `Col` approach in Anbox because, below the large breakpoint, it was causing the `EmptyState` to "collapse" towards the side nav instead of staying centred in the page:

Peek 2023-08-09 15-07

Check the `.empty-state` CSS class in Anbox for an alternative way to do this.

aaryanporwal: The Row and Col approach is used to make the EmptyState centred, do we want it to be left un-centred?

src/pages/storage/StorageList.tsx
lorumic: No, never use inline styles. In the storybook we use them because we are forced to, and also to make the code examples self-contained. We want to have a single CSS class that can be used for all the empty state icons throughout the app. Just like we do in Anbox.

src/pages/networks/NetworkList.tsx
```diff
@@ -148,25 +149,23 @@ const NetworkList: FC = () => {
               }
             />
           )}
-          {!isLoading && !hasNetworks && (
+          {!isLoading && (
```
edlerd, line 152: Why remove the `!hasNetworks` condition?

src/pages/networks/NetworkList.tsx
```diff
@@ -12,7 +13,7 @@ import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import Loader from "components/Loader";
 import { useNavigate, useParams } from "react-router-dom";
-import EmptyState from "components/EmptyState";
```
edlerd, line 15: Please also remove the local component in `components/EmptyState.tsx`

src/pages/networks/NetworkList.tsx
aaryanporwal: I'm still working on this PR. I'll convert it to Draft. (I'm removing the conditions to force EmptyState render, of course for testing purpose only)

src/pages/networks/NetworkList.tsx
edlerd: I see, makes sense. Be careful with changes that should not be committed. One way is to add a `//todo: do not commit` comment above the change, so it won't accidentally make it into a commit.

aaryanporwal: That's a neat trick thanks :)

lorumic: Also, no need to push to remote any little and temporary work-in-progress. You can keep the changes locally, and push them only once at the end, unless you need to test them on the demo server, but it doesn't seem the best way to proceed since each build takes more than 5 minutes to complete.

aaryanporwal: You're right, I'll squash the ones here and the ones I have locally and push.

src/pages/instances/InstanceList.tsx
```diff
@@ -585,35 +586,38 @@ const InstanceList: FC = () => {
               )}
               {!hasInstances && (
                 <EmptyState
-                  iconName="containers"
-                  iconClass="empty-instances-icon"
```
edlerd, line 589: I think the `empty-instances-icon` class is now unused. Can you remove it from `styles.scss` as well?
Also check the other icon classes that were removed in other files.

src/pages/instances/InstanceList.tsx
```diff
@@ -17,7 +18,7 @@ import { instanceCreationTypes } from "util/instanceOptions";
 import InstanceStatusIcon from "./InstanceStatusIcon";
 import TableColumnsSelect from "components/TableColumnsSelect";
 import useEventListener from "@use-it/event-listener";
-import EmptyState from "components/EmptyState";
+
```
edlerd, line 21: Why leave an empty line here, I think this should be a single block - also in the other files.

edlerd (commented):
LGTM, some ideas for further clean-up of unused CSS and some style cosmetics below.

edlerd (approved):
LGTM, thank you :)

PR #432, opened 2023-08-09
Fix handling of restricted accounts #426

### LINKED ISSUE #426: No metrics displayed if user is restricted to see just one project
If I set with lxc config trust that user is restricted to a certain restricted project (project was auto-created by using multi-user setup), ui doesn't show Usage and is just displaying indefinitely "Loading metrics ...".

If I again set user to see all projects (with lxc config trust edit xxxxyyyyzzzz, set restricted to false), GUI shows everything, including metrics for an instance in user-1000 project.

edlerd:
## Done

- in useAuth hook figure if a current user is restricted
- use a different default project if user has no access to project with name "default" 
- on restricted users
  - avoid 403 requests to the metrics api
  - hide edit buttons for projects
  - hide warnings section in navigation
  - hide edit buttons for settings

Fixes #426 

## QA

1. Run the LXD-UI:
    - create a restricted account (lxc config trust edit $hash, change restricted to true and projects to [$project] where $hash is the hash of your certificate and $project is a non-default project
    - check the default redirect brings the user to a project that is acessible
    - check the instance detail page is displayed without error
    - check the server settings are not editable (no edit button)
    - check the project is not editable (no edit button)
    - check the warnings link is not in the main navigation
    - ensure all of the above still work correctly for a non-restricted user.
    - ensure the login process for a new user still works (identifying non-authenticated users is slightly changed)

### DIFF
--- a/src/pages/instances/Events.tsx
+++ b/src/pages/instances/Events.tsx
@@ -1,8 +1,10 @@
 import React, { FC, useEffect, useState } from "react";
 import { LxdEvent } from "types/event";
 import { useEventQueue } from "context/eventQueue";
+import { useAuth } from "context/auth";
 
 const Events: FC = () => {
+  const { isAuthenticated } = useAuth();
   const eventQueue = useEventQueue();
   const [eventWs, setEventWs] = useState<WebSocket | null>(null);
 
@@ -41,10 +43,10 @@ const Events: FC = () => {
   };
 
   useEffect(() => {
-    if (!eventWs) {
+    if (!eventWs && isAuthenticated) {
       connectEventWs();
     }
-  }, [eventWs]);
+  }, [eventWs, isAuthenticated]);
   return <></>;
 };
 
--- a/src/pages/instances/InstanceOverviewMetrics.tsx
+++ b/src/pages/instances/InstanceOverviewMetrics.tsx
@@ -37,7 +37,7 @@ const InstanceOverviewMetrics: FC<Props> = ({ instance, onFailure }) => {
   if (isRestricted) {
     return (
       <div className="u-text--muted">
-        Usage details unavailable for restricted users
+        Details are not available for restricted users
       </div>
     );
   }
--- a/src/pages/instances/InstanceOverviewMetrics.tsx
+++ b/src/pages/instances/InstanceOverviewMetrics.tsx
@@ -7,6 +7,7 @@ import { getInstanceMetrics } from "util/metricSelectors";
 import Meter from "components/Meter";
 import Loader from "components/Loader";
 import { LxdInstance } from "types/instance";
+import { useAuth } from "context/auth";
 
 interface Props {
   instance: LxdInstance;
@@ -14,6 +15,8 @@ interface Props {
 }
 
 const InstanceOverviewMetrics: FC<Props> = ({ instance, onFailure }) => {
+  const { isRestricted } = useAuth();
+
   const {
     data: metrics = [],
     error,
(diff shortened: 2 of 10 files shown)


src/pages/instances/Events.tsx
```diff
@@ -1,8 +1,14 @@
 import React, { FC, useEffect, useState } from "react";
 import { LxdEvent } from "types/event";
 import { useEventQueue } from "context/eventQueue";
+import { useAuth } from "context/auth";
 
 const Events: FC = () => {
+  const { isAuthenticated } = useAuth();
+  if (!isAuthenticated) {
+    return null;
+  }
```
lorumic, line 10: This is breaking the rules of hooks:
```
Warning: React has detected a change in the order of Hooks called by Events. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://reactjs.org/link/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useContext                 useContext
2. undefined                  useContext
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    at Events (https://localhost:8407/src/pages/instances/Events.tsx:24:7)
```

src/pages/instances/InstanceOverviewMetrics.tsx
```diff
@@ -32,7 +36,7 @@ const InstanceOverviewMetrics: FC<Props> = ({ instance, onFailure }) => {
 
   return (
     <>
-      {isLoading ? (
+      {isLoading && !isRestricted ? (
```
lorumic, line 39: I think this is wrong, UX-wise. If I understand correctly what's going on here, the restricted user would get shown the empty "-" placeholder. So, in principle, one cannot really distinguish whether they're seeing a "-" placeholder because they are restricted, or because the metrics are not available (e.g. the instance is not running).
A better UX is to clearly show the user that this data (the metrics) is not accessible due to an existing restriction, when this is the case.

src/pages/settings/SettingForm.tsx
```diff
@@ -113,6 +115,8 @@ const SettingForm: FC<Props> = ({ option, value }) => {
             Save
           </Button>
         </Form>
+      ) : isRestricted ? (
+        <span>{value ? value : "-"}</span>
```
lorumic, line 119: Doesn't this work properly here?
```suggestion
        <span>{value ?? "-"}</span>
```

src/pages/instances/Events.tsx
edlerd: True this changes when a user logs in, I'll change this slightly, so it is conforming.

src/pages/instances/InstanceOverviewMetrics.tsx
edlerd: Yes, we can show a dedicated message for restricted users, changed it accordingly.

src/pages/instances/InstanceOverviewMetrics.tsx
```diff
@@ -30,6 +34,14 @@ const InstanceOverviewMetrics: FC<Props> = ({ instance, onFailure }) => {
 
   const instanceMetrics = getInstanceMetrics(metrics, instance);
 
+  if (isRestricted) {
+    return (
+      <div className="u-text--muted">
+        Usage details unavailable for restricted users
```
lorumic, line 40: "Usage" is already the title of the section, so maybe it's redundant here:
```suggestion
        Details unavailable for restricted users
```

PR #447, opened 2023-09-01
Allow instance edit with a custom iso attached

edlerd:
## Done

- fixed instance edit with attached custom iso

## QA

1. Run the LXD-UI:
    - create an instance with an attached iso volume
    - edit the instance and hit save, it should succeed

### DIFF
--- a/src/util/formDevices.tsx
+++ b/src/util/formDevices.tsx
@@ -13,7 +13,7 @@ interface EmptyDevice {
 interface UnknownDevice {
   type: "unknown";
   name: string;
-  bare: string;
+  bare: unknown;
 }
 
 export interface IsoVolumeDevice {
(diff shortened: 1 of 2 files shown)


src/util/formDevices.tsx
```diff
@@ -75,16 +75,24 @@ export const parseDevices = (devices: LxdDevices): FormDevice[] => {
...
+        name: key,
+        bare: item,
+        type: "custom-nic",
+      };
+    }
+
+    if (key === "iso-volume") {
+      return {
+        name: key,
+        bare: item as never,
```
lorumic, line 89: `item` seems to have a value here, shouldn't this be treated as `unknown`?
```suggestion
        bare: item as unknown,
```

lorumic (approved):
LGTM, just one small type question.

src/util/formDevices.tsx
edlerd: Right, `unknown` is the better choice here. I also had to change the `UnknownDevice` type for this to work though.

aaryanporwal (approved):
lgtm

PR #452, opened 2023-09-05
Set the correct container element to avoid unspaced content in confirmation modals

lorumic:
## Done

- Set the correct container element to get an appropriate spacing between the main text and the shift+click hint in confirmation modals

## QA

1. Run the LXD-UI:
    - Open any of the confirmation modals.
    - Check that there is a correct spacing between the main text and the shift+click hint below it.


src/util/images.tsx
```diff
@@ -2,7 +2,7 @@ import { LxdImage, RemoteImage } from "types/image";
 import { LxdStorageVolume } from "types/storage";
 
 export const isVmOnlyImage = (image: RemoteImage) => {
-  return image.variant?.includes("desktop") || image.server === LOCAL_ISO;
+  return image.variant?.includes("desktop") ?? image.server === LOCAL_ISO;
```
edlerd, line 5: This condition was changed to `??` beforehand, and that introduced a bug. I don't remember the exact circumstances -- but do remember changing it back and forth :)

edlerd (commented):
Good catch, with the spacing it looks way better. Thank you for spotting, fixing and submitting a PR for it!

Just a tiny note on the `images.tsx`. I think we shouldn't change it.

src/util/images.tsx
lorumic: Right, I remember the circumstances. Weird that this triggers a warning and I didn't notice it back in the days. I will revert this change.

lorumic: I reverted the change, but please do investigate a way to remove the warning. It usually suggests that the thing could (and should) be written in a better way.

edlerd (approved):

PR #455, opened 2023-09-05
Add button to eject iso images from instance console

edlerd:
## Done

- added button to eject iso images from instance console

## QA

1. Run the LXD-UI:
    - launch instance with an iso, go to console start instance, check eject button is there and works.

### DIFF
--- a/src/pages/instances/actions/AttachIsoBtn.tsx
+++ b/src/pages/instances/actions/AttachIsoBtn.tsx
@@ -1,0 +1,110 @@
+import React, { FC, useState } from "react";
+import { updateInstance } from "api/instances";
+import { LxdInstance } from "types/instance";
+import { useParams } from "react-router-dom";
+import { ActionButton, useNotify } from "@canonical/react-components";
+import { getInstanceEditValues, getInstancePayload } from "util/instanceEdit";
+import { LxdIsoDevice } from "types/device";
+import { useQueryClient } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import usePortal from "react-useportal";
+import { RemoteImage } from "types/image";
+import CustomIsoModal from "pages/images/CustomIsoModal";
+import { remoteImageToIsoDevice } from "util/formDevices";
+
+interface Props {
+  instance: LxdInstance;
+}
+
+const AttachIsoBtn: FC<Props> = ({ instance }) => {
+  const { project } = useParams<{ project: string }>();
+  const notify = useNotify();
+  const queryClient = useQueryClient();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const [isLoading, setLoading] = useState(false);
+
+  const attachedIso = instance.devices["iso-volume"] as
+    | LxdIsoDevice
+    | undefined;
+
+  const detachIso = () => {
+    setLoading(true);
+    const values = getInstanceEditValues(instance);
+    values.devices = values.devices.filter((device) => {
+      return device.name !== "iso-volume";
+    });
+    const instanceMinusIso = getInstancePayload(
+      instance,
+      values
+    ) as LxdInstance;
+    updateInstance(instanceMinusIso, project ?? "")
+      .then(() =>
+        notify.success(
+          <>
+            ISO <b>{attachedIso?.source ?? ""}</b> detached
+          </>
+        )
+      )
+      .catch((e: Error) => notify.failure("Detach ISO failed", e))
+      .finally(() => {
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.instances, instance.name, project],
+        });
+        setLoading(false);
+      });
+  };
+
+  const handleSelect = (image: RemoteImage) => {
+    setLoading(true);
+    closePortal();
+    const values = getInstanceEditValues(instance);
+    const isoDevice = remoteImageToIsoDevice(image);
+    values.devices.push(isoDevice);
+    const instancePlusIso = getInstancePayload(instance, values) as LxdInstance;
+    updateInstance(instancePlusIso, project ?? "")
+      .then(() =>
+        notify.success(
+          <>
+            ISO <b>{image.aliases}</b> attached
+          </>
+        )
+      )
+      .catch((e: Error) => notify.failure("Attaching ISO failed", e))
+      .finally(() => {
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.instances, instance.name, project],
+        });
+        setLoading(false);
+      });
+  };
+
+  return attachedIso ? (
+    <>
+      <span className="u-text--muted margin-right">{attachedIso.source}</span>
+      <ActionButton
+        loading={isLoading}
+        onClick={detachIso}
+        className="u-no-margin--bottom"
+      >
+        Detach ISO
+      </ActionButton>
+    </>
+  ) : (
+    <>
+      <ActionButton
+        loading={isLoading}
+        onClick={openPortal}
+        className="u-no-margin--bottom"
+      >
+        Attach ISO
+      </ActionButton>
+      {isOpen && (
+        <Portal>
+          <CustomIsoModal onClose={closePortal} onSelect={handleSelect} />
+        </Portal>
+      )}
+    </>
+  );
+};
+
+export default AttachIsoBtn;
--- a/src/components/SubmitButton.tsx
+++ b/src/components/SubmitButton.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from "react";
+import React, { FC, MouseEventHandler } from "react";
 import { Button, Icon } from "@canonical/react-components";
 import classnames from "classnames";
 
@@ -8,7 +8,7 @@ interface Props {
   buttonLabel: string;
   appearance?: string;
   processingText?: string;
-  onClick?: () => void;
+  onClick?: MouseEventHandler<HTMLButtonElement>;
   dense?: boolean;
   className?: string;
 }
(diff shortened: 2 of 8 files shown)


src/pages/instances/actions/EjectIsoBtn.tsx
```diff
@@ -0,0 +1,51 @@
...
+      values
+    ) as LxdInstance;
+    updateInstance(instanceMinusIso, project ?? "")
+      .then(() => notify.success("ISO ejected"))
+      .catch((e: Error) => notify.failure("Eject ISO failed", e));
+  };
+
+  if (!attachedIso) {
+    return null;
+  }
```
lorumic, line 36: I think this can be moved before the `ejectIso` function.

lorumic (commented):
What I don't like here is that the action is not reversible. If I eject an ISO and want to "put it back" where it was, I need to manually edit the YAML, which is not something accessible for the inexperienced user.

In other words, we're adding a "user-friendly" button to eject the ISO, but then we leave the inexperienced user with no way to "roll back" to the previous state (instance with an ISO attached).

Can we improve the feature in this sense? For example, we could use the same "Eject ISO" button also as an "Attach ISO" button, hence showing it all the time for VM instances. Then we branch on the `attachedIso` variable: if it's undefined, we show the "Attach ISO" version, allowing the user to add an ISO volume (under the hood, this modifies the `devices` object of the instance), otherwise we use the logic that is already implemented to eject the ISO.

By doing this, we're essentially creating a virtual CD/DVD reader for Windows VMs - isn't that cool?

edlerd:
> What I don't like here is that the action is not reversible. If I eject an ISO and want to "put it back" where it was, I need to manually edit the YAML, which is not something accessible for the inexperienced user.

A valid concern and an interesting idea to have an "Attach ISO" button in case there is no ISO on the VM. Though, I am not sure if that is confusing for the most part of our users that are "just running" VMs. I'll discuss this later today with @piperdeck

lorumic:
> > What I don't like here is that the action is not reversible. If I eject an ISO and want to "put it back" where it was, I need to manually edit the YAML, which is not something accessible for the inexperienced user.
> 
> A valid concern and an interesting idea to have an "Attach ISO" button in case there is no ISO on the VM. Though, I am not sure if that is confusing for the most part of our users that are "just running" VMs. I'll discuss this later today with @piperdeck

I think that if they can grasp the concept of "ejecting" an ISO, they can easily grasp also the other way round. So I don't see it more confusing than the feature that is being added here.

edlerd:
> I think that if they can grasp the concept of "ejecting" an ISO, they can easily grasp also the other way round. So I don't see it more confusing than the feature that is being added here.

The eject is visible to users that want to install from an ISO. Most users will run a VM from an image server and will never see the eject button. My concern was to avoid polluting the interface with an action that is not relevant in the current context. However, writing this now we can distinguish both cases, because the ISO VMs will have no base image. So we can limit the attach new ISO button to the VMs that were launched with an ISO.

lorumic:
> The eject is visible to users that want to install from an ISO. Most users will run a VM from an image server and will never see the eject button. My concern was to avoid polluting the interface with an action that is not relevant in the current context. However, writing this now we can distinguish both cases, because the ISO VMs will have no base image. So we can limit the attach new ISO button to the VMs that were launched with an ISO.

Personally I would add it to all VMs, because I see it as the missing "virtual CD/DVD reader" feature that competitor products have (one example [here](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vm_admin.doc/GUID-C58B93A7-52CF-456D-95C1-8B5A906C9619.html)).

Now, if we're talking about how to avoid "polluting" the interface, that's a completely different topic because I think the interface is polluted already for a different reason: button labels. Which brings me to an "old" idea that I mentioned in the past: a "control bar" (see the comment [here](https://github.com/canonical/lxd-ui/pull/383#issuecomment-1585124378)) could be a good place to contain this kind of feature buttons, which should eventually be converted to _icon_ buttons.

edlerd:
Putting back in draft mode, because this is based on #445 hence appearing to have more changes than just the single commit it adds.

piperdeck (changes requested):
Overall, I would like more icons to make these buttons on the console screen more identifiable at a glance, but I realize we don't actually have the right assets to make that happen. For now, I would suggest putting the ISO file name next to the eject button, instead of in a tooltip. The first reason is that it makes it easier to quickly identify the ISO, and the second reason is that it provides a visual contrast between the two states of having or not having an ISO attached.

Frame 4 (2)

edlerd:
Thanks @piperdeck that looks much better indeed. I updated this PR to use the muted text instead of the tooltip next to the eject button. 

@lorumic I also updated from the upstream changes, so it is now ready for code review as well.

piperdeck (approved):
Looks great!

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -0,0 +1,120 @@
...
+    ) as LxdInstance;
+    updateInstance(instanceMinusIso, project ?? "")
+      .then(() =>
+        notify.success(
+          <>
+            ISO <b>{attachedIso?.source ?? ""}</b> ejected
+          </>
+        )
+      )
+      .catch((e: Error) => notify.failure("Detach ISO failed", e))
```
lorumic, line 48: Is there a reason why we use "eject" as the verb for the success message but "detach" here? I would use a bit of consistency and change this to:
```suggestion
      .catch((e: Error) => notify.failure("ISO ejection failed", e))
```

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -0,0 +1,120 @@
...
+          queryKey: [queryKeys.instances, instance.name, project],
+        });
+        setLoading(false);
+      });
+  };
+
+  return attachedIso ? (
+    <>
+      <span className="u-text--muted margin-right">{attachedIso.source}</span>
+      <SubmitButton
```
lorumic, line 92: If it's about having a button that shows a loader while an operation is running, the most semantically correct choice here would probably be react-component's [`ActionButton`](https://canonical.github.io/react-components/?path=/docs/actionbutton--default-story).

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -0,0 +1,120 @@
...
+        appearance=""
+        isSubmitting={isLoading}
+        isDisabled={false}
+        buttonLabel="Detach ISO"
```
lorumic, line 96: Here too, I suggest using the verb "eject", as it is probably more familiar in this kind of context, but if there's a particular reason you chose "detach", feel free to disregard this suggestion.
```suggestion
        buttonLabel="Eject ISO"
```

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -0,0 +1,120 @@
...
+    values.devices.push({
+      type: "iso-volume",
+      name: "iso-volume",
+      bare: {
+        "boot.priority": "10",
+        pool: image.pool ?? "",
+        source: image.aliases,
+        type: "disk",
+      },
+    });
```
lorumic, line 70: This chunk of code is currently used in 2 different places - here and in `CreateInstanceForm`. Maybe it would be good to create a reusable helper function for it?

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -0,0 +1,120 @@
...
+        isSubmitting={isLoading}
+        isDisabled={false}
+        buttonLabel="Detach ISO"
+        onClick={ejectIso}
+        className="u-no-margin--bottom"
+      />
+    </>
+  ) : (
+    <>
+      <SubmitButton
```
lorumic, line 103: Same as above - I suggest using `ActionButton` here.

src/pages/instances/actions/AttachIsoBtn.tsx
edlerd: We had a discussion about it and said it would be consistent to use detach/attach in more places of the UI.

lorumic: Ok, then maybe it would be worth removing all the "eject"s? We have a mixed usage at the moment.

src/pages/instances/actions/AttachIsoBtn.tsx
edlerd: Good point!

src/pages/instances/actions/AttachIsoBtn.tsx
edlerd: Very good point! I wanted to extract this but forgot -- will do that now!

src/pages/instances/actions/AttachIsoBtn.tsx
edlerd: I think the latest state of the branch removed them all.

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -0,0 +1,114 @@
...
+      });
+  };
+
+  return attachedIso ? (
+    <>
+      <span className="u-text--muted margin-right">{attachedIso.source}</span>
+      <ActionButton
+        appearance=""
+        loading={isLoading}
+        disabled={false}
```
lorumic, line 87: I think 2 props here are unneeded, since they're specifying the value that is already the default.
```suggestion
        loading={isLoading}
```

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -0,0 +1,114 @@
...
+      >
+        Detach ISO
+      </ActionButton>
+    </>
+  ) : (
+    <>
+      <ActionButton
+        appearance=""
+        loading={isLoading}
+        disabled={false}
```
lorumic, line 99: Same as above:
```suggestion
        loading={isLoading}
```

src/pages/instances/actions/AttachIsoBtn.tsx
edlerd: Ah right, I missed than when switching the component. Thanks :)

lorumic (approved):

PR #456, opened 2023-09-07
Extend iso tests to launch an instance with it

edlerd:
## Done

- add test to launch instance with custom iso

## QA

1. Run the LXD-UI:
    - ensure tests are passing i.e. `npx playwright test iso-volumes`

### DIFF
--- a/tests/iso-volumes.spec.ts
+++ b/tests/iso-volumes.spec.ts
@@ -1,5 +1,7 @@
 import { test } from "@playwright/test";
 import { randomNameSuffix } from "./helpers/name";
+import { deleteInstance, randomInstanceName } from "./helpers/instances";
+import { TIMEOUT } from "./helpers/constants";
 
 const ISO_FILE = "./tests/fixtures/foo.iso";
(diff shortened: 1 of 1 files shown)


tests/iso-volumes.spec.ts
```diff
@@ -37,3 +39,36 @@ test("upload and delete custom iso", async ({ page }) => {
   await page.getByText("Delete", { exact: true }).click();
   await page.getByText(`Storage volume ${isoName} deleted.`).click();
 });
+
+test("use custom iso for instance launch", async ({ page }) => {
+  test.skip(Boolean(process.env.CI), "github runners lack vm support");
```
lorumic, line 44: Do we need to skip this? We're just creating the instance, not starting it. Or would it not be able to even create it?

edlerd: Yes exactly, it fails to create the instance with type VM, see also this run: https://github.com/canonical/lxd-ui/actions/runs/6108398358

PR #467, opened 2023-09-13
feat(operations): Move operations list to global scope [WD-4462]

aaryanporwal:
## Done

- Operations page now lists operations across all projects!
- Moved operations from project scope to under the cluster tab
- Uses https://github.com/canonical/lxd/pull/12172
- Fixes https://github.com/canonical/lxd/issues/11808

## QA

1. Run the LXD-UI:
    - Run an operation (eg: create instance) > check the operations page to see if the operation is being shown in the list

## Screenshots

image

### DIFF
--- a/src/pages/operations/OperationList.tsx
+++ b/src/pages/operations/OperationList.tsx
@@ -80,7 +80,7 @@ const OperationList: FC = () => {
                 <OperationInstanceName operation={operation} />
               </div>
               <div className="u-text--muted">
-                Project name: {getProjectName(operation)}
+                Project: {getProjectName(operation)}
               </div>
             </>
           ),
--- a/src/pages/operations/OperationList.tsx
+++ b/src/pages/operations/OperationList.tsx
@@ -57,6 +57,7 @@ const OperationList: FC = () => {
   };
 
   const rows = operations.map((operation) => {
+    const projectName = getProjectName(operation);
     return {
       columns: [
         {
@@ -79,8 +80,8 @@ const OperationList: FC = () => {
               <div className="u-truncate u-text--muted">
                 <OperationInstanceName operation={operation} />
               </div>
-              <div className="u-text--muted u-truncate" title={getProjectName(operation)}>
-                Project: {getProjectName(operation)}
+              <div className="u-text--muted u-truncate" title={projectName}>
+                Project: {projectName}
               </div>
             </>
           ),
--- a/src/pages/operations/OperationList.tsx
+++ b/src/pages/operations/OperationList.tsx
@@ -79,7 +79,7 @@ const OperationList: FC = () => {
               <div className="u-truncate u-text--muted">
                 <OperationInstanceName operation={operation} />
               </div>
-              <div className="u-text--muted">
+              <div className="u-text--muted u-truncate" title={getProjectName(operation)}>
                 Project: {getProjectName(operation)}
               </div>
             </>
--- a/src/pages/operations/OperationList.tsx
+++ b/src/pages/operations/OperationList.tsx
@@ -10,9 +10,8 @@ import BaseLayout from "components/BaseLayout";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import Loader from "components/Loader";
-import { fetchOperations } from "api/operations";
+import { fetchAllOperations } from "api/operations";
 import CancelOperationBtn from "pages/operations/actions/CancelOperationBtn";
-import { useParams } from "react-router-dom";
 import { isoTimeToString } from "util/helpers";
 import { LxdOperationStatus } from "types/operation";
 import OperationInstanceName from "pages/operations/OperationInstanceName";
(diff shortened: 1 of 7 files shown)


src/pages/operations/OperationList.tsx
lorumic, line 1: This is more of a UX question, but shouldn't we allow filtering by project? I think we have that information in the response, right?

src/pages/operations/OperationList.tsx
```diff
@@ -10,29 +10,23 @@ import BaseLayout from "components/BaseLayout";
...
 
   const {
     data: operationList,
     error,
     isLoading,
   } = useQuery({
-    queryKey: [queryKeys.operations, project],
-    queryFn: () => fetchOperations(project),
+    queryKey: [queryKeys.operations],
+    queryFn: () => fetchAllOperations(),
```
lorumic, line 29:
```suggestion
    queryFn: fetchAllOperations,
```

src/pages/operations/OperationList.tsx
```diff
@@ -166,7 +158,7 @@ const OperationList: FC = () => {
               image={<Icon name="status" className="empty-state-icon" />}
               title="No operations found"
             >
-              <p>There are no ongoing operations in this project.</p>
+              <p>There are no ongoing operations right now.</p>
```
lorumic, line 161: Maybe "right now" is already implicit in "ongoing"?
```suggestion
              <p>There are no ongoing operations.</p>
```

aaryanporwal: Yeah you're right "ongoing" already implies that.

src/pages/operations/OperationList.tsx
aaryanporwal: @edlerd mentioned that as well. We can discuss the need for filtering by project. But I think adding filtering is out of the scope of this PR.

edlerd: Yeah, we discussed it briefly and concluded we expect the operation list to be not filled so a search and filter is not needed. We'll double check in the core team sync with their opinion on this and eventually add a story to add search + filter if needed.

edlerd, line 1: We forgot to update `OperationInstanceName`, it currently picks the `projectName` from the url with `useParams`. But we don't have the project in the url, hence the component does not link any more because the project is always missing.

We should parse the project from the operation. See how the `getInstanceName` function is doing it. We should add a similar function to `getProjectName`, that gives the value from the `project=<project_name>` parameter of the URL in the `operation.resources?.instances` field. Remember that there is a special case, when no project parameter is present, the project will be "default".

edlerd (changes requested):

src/pages/operations/OperationList.tsx
aaryanporwal: Thanks! just fixed it :)

src/util/operations.tsx
```diff
@@ -13,3 +13,17 @@ export const getInstanceName = (operation: LxdOperation): string => {
...
+
+export const getProjectName = (operation: LxdOperation): string => {
+  // the url can be
+  // /1.0/instances/<instance_name>?project=<project_name>
+  // when no project parameter is present, the project will be "default"
+
+  return (
+    operation.resources?.instances
+      ?.filter((item) => item.startsWith("/1.0/instances/"))
+      .map((item) => item.split("/")[4])
```
edlerd, line 25: Linking to non-default instances was broken -- I think because of this.
```suggestion
      .map((item) => item.split("/")[3])
```

edlerd (commented):
Thank you for the changes, some comment to fix the project in the link.
Also we discovered in the UX/UI sync that there should be an additional column for the project in the `OperationList.tsx` now, that we took it out of the project scope. Maybe as a 2nd column between time and action?

src/util/operations.tsx
aaryanporwal: Thanks, totally missed this.

edlerd:
> Thank you for the changes, some comment to fix the project in the link. Also we discovered in the UX/UI sync that there should be an additional column for the project in the `OperationList.tsx` now, that we took it out of the project scope. Maybe as a 2nd column between time and action?

@aaryanporwal this seems to be unchanged?

aaryanporwal:
So we decided to put the "Project: Project name" in muted font in the action column below instance name.

src/pages/operations/OperationList.tsx
```diff
@@ -84,6 +79,9 @@ const OperationList: FC = () => {
               <div className="u-truncate u-text--muted">
                 <OperationInstanceName operation={operation} />
               </div>
+              <div className="u-text--muted">
+                Project name: {getProjectName(operation)}
```
edlerd, line 83:
```suggestion
                Project: {getProjectName(operation)}
```
I think we can be a bit more concise here.

src/pages/operations/OperationList.tsx
```diff
@@ -84,6 +79,9 @@ const OperationList: FC = () => {
               <div className="u-truncate u-text--muted">
                 <OperationInstanceName operation={operation} />
               </div>
+              <div className="u-text--muted">
```
edlerd, line 82:
```suggestion
              <div className="u-text--muted u-truncate" title={getProjectName(operation)}>
```
Better be prepared for users with super long project names not overflowing into the next line. While still making the full name visible on hover with the title.

edlerd (approved):
Looking good, two tiny ideas to improve the operation list page.

src/pages/operations/OperationList.tsx
aaryanporwal: Yeah concise looks better

PR #468, opened 2023-09-14
Add ui version to navigation

edlerd:
## Done

- added ui version const
- change navigation version string to include backend and ui

## QA

1. Run the LXD-UI:
    - check version in nav

### DIFF
--- a/src/components/Version.tsx
+++ b/src/components/Version.tsx
@@ -1,26 +1,30 @@
 import React, { FC } from "react";
 import { Icon, Tooltip } from "@canonical/react-components";
 import { useSettings } from "context/useSettings";
+import { RECENT_MAJOR_SERVER_VERSION, UI_VERSION } from "util/version";
 
-const ServerVersion: FC = () => {
+const Version: FC = () => {
   const { data: settings } = useSettings();
 
-  const version = settings?.environment?.server_version;
-  if (!version) {
+  const serverVersion = settings?.environment?.server_version;
+  if (!serverVersion) {
     return null;
   }
 
-  const major = version.includes(".") ? version.split(".")[0] : undefined;
-  const recentMajor = 5;
-  const isOutdated = major ? parseInt(major) < recentMajor : false;
+  const serverMajor = serverVersion.includes(".")
+    ? serverVersion.split(".")[0]
+    : undefined;
+  const isOutdated = serverMajor
+    ? parseInt(serverMajor) < RECENT_MAJOR_SERVER_VERSION
+    : false;
 
   return (
     <>
       <hr className="p-side-navigation__list is-dark navigation-hr" />
-      <li className="p-side-navigation__link server-version">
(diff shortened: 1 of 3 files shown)


src/components/Version.tsx
```diff
@@ -1,37 +1,41 @@
...
+            message="You are using an outdated server version.
 Update your LXD server to benefit from the latest features."
             tooltipClassName="version-warning"
             zIndex={1000}
           >
             <Icon name="warning" className="p-side-navigation__icon" />
           </Tooltip>
         )}
-        Server {version}
+        Version {serverVersion}-ui-{UI_VERSION}
```
lorumic, line 35: I thought we had agreed on the shorthand "Ver"?
```suggestion
        Ver {serverVersion}-ui-{UI_VERSION}
```

src/util/version.tsx
```diff
@@ -0,0 +1,2 @@
+export const RECENT_MAJOR_SERVER_VERSION = 5;
+export const UI_VERSION = "5-1";
```
lorumic, line 2: The hyphen here is a bit confusing, shouldn't it be a dot instead?
```suggestion
export const UI_VERSION = "5.1";
```

src/components/Version.tsx
edlerd: I looked at the "Ver" before, and it seemed a bit unclear what the string even presents. The version gibberish after that is already quite cryptic. We have enough space to write out "Version", so maybe keep it as is?

src/util/version.tsx
edlerd: I was thinking about it -- I assume it might be confusing to have two version string with "5." concatenated together. Hence my suggestion with the minus. See also the [proposed process](https://docs.google.com/document/d/15xqgRFLMHtLC-SehjW3m0OX-dXjAGn9ylvs0ZHFdrdM/edit) for the ui release.

src/components/Version.tsx
lorumic: Might be. It's just that it looks a bit tight in space on my small-width screen:

image

Maybe ask @piperdeck if this is good? It might just be me, but I'd like to have a bit more margin on the right. It's also relevant to know if this can become any longer, e.g. 5.17-ui-5-1 might fit, but what if we have 8.88-ui-8-88? (picking 8 here since it's the widest number)

src/util/version.tsx
lorumic: Ok, if everybody agrees with that I'm fine too, it's just that it's a bit different from what we discussed (and noted down) in the sync with the core team yesterday. Hence my questions (this and the one above).

edlerd: We decided to make this . based in the doc after all :)

src/components/Version.tsx
edlerd: as discussed, using a much smaller font to give it more margin.

lorumic (approved):

PR #469, opened 2023-09-18
Validate instance name does not start with a number

edlerd:
## Done

- validate instance name does not start with a number

## QA

1. Run the LXD-UI:
    - create or rename instance to a name that starts with a number / not with a number

### DIFF
--- a/src/pages/instances/CreateInstanceForm.tsx
+++ b/src/pages/instances/CreateInstanceForm.tsx
@@ -106,6 +106,9 @@ const CreateInstanceForm: FC = () => {
       .matches(/^[A-Za-z0-9-]+$/, {
         message: "Only alphanumeric and hyphen characters are allowed",
       })
+      .matches(/^[A-Za-z].*$/, {
+        message: "Instance name must start with a letter",
+      })
       .optional(),
     instanceType: Yup.string().required("Instance type is required"),
   });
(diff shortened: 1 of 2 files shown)


edlerd:
Will also add validation to not start with a -

src/pages/instances/CreateInstanceForm.tsx
```diff
@@ -106,6 +106,9 @@ const CreateInstanceForm: FC = () => {
       .matches(/^[A-Za-z0-9-]+$/, {
         message: "Only alphanumeric and hyphen characters are allowed",
       })
+      .matches(/^[^0-9].*$/, {
+        message: "Name cannot start with a number",
```
aaryanporwal, line 110: Can this be renamed to instance name? feels a bit more explicit to me.
```suggestion
        message: "Instance name cannot start with a number",
```

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -33,6 +33,9 @@ const InstanceDetailHeader: FC<Props> = ({ name, instance, project }) => {
       .matches(/^[A-Za-z0-9-]+$/, {
         message: "Only alphanumeric and hyphen characters are allowed",
       })
+      .matches(/^[^0-9].*$/, {
+        message: "Name cannot start with a number",
```
aaryanporwal, line 37: same as above?

aaryanporwal (approved):
LGTM, there's a small nit^

src/pages/instances/CreateInstanceForm.tsx
edlerd: Good idea. I also changed the condition below, because it must start with a letter, - is also an invalid start.

PR #470, opened 2023-09-18
Add hostname from url to the page title

### LINKED ISSUE #464: Clearly distinguish LXD instances in page title
If you have multiple LXD UIs open in, e.g. Firefox tabs, then it's very hard to tell them apart.  The tabs are all named "LXD UI", and the pages are all visually the same with the name "Canonical LXD" at the top left.

image

Compare this with e.g. Syncthing, where each instance has a name, that name is the in name of the tab, and is clearly displayed at the top of the page.

image

Please consider adding a something to the LXD UI, ideally settable from the Settings page.

It would be nice if the defaults for this setting at least included the hostname.

Thanks!

edlerd:
## Done

- Add hostname from url to the page title

- Fixes #464

## QA

1. Run the LXD-UI:
    - check page title
    - customize page title from settings page with the user.ui_title key

### DIFF
--- a/public/assets/data/config-options.json
+++ b/public/assets/data/config-options.json
@@ -460,5 +460,12 @@
     "scope": "local",
     "default": "-",
     "description": "Volume to use to store the image tarballs (syntax is POOL/VOLUME)"
+  },
+  {
+    "key": "user.ui_title",
+    "type": "string",
+    "scope": "local",
+    "default": "-",
+    "description": "Custom identifier for the title of the UI when used on this server"
   }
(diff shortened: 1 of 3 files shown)


src/context/title.tsx
lorumic, line 1: I'm not sure about placing this in the context directory. It's not using React Context after all, so it might be misleading.

edlerd: right, moving it to `util/`

lorumic (approved):

PR #471, opened 2023-09-18
Add js test execution and first two test cases WD-6235

edlerd:
## Done

- add two js test cases
- add test runner to package.json
- schedule tests on github PR

Fixes WD-6235

## QA

1. Run the LXD-UI:
    - check PR tests include js-tests and are passing

### DIFF
--- a/src/util/operations.spec.tsx
+++ b/src/util/operations.spec.tsx
@@ -1,0 +1,46 @@
+import { getInstanceName } from "./operations";
+import { LxdOperation } from "types/operation";
+
+const craftOperation = (url: string) => {
+  return {
+    resources: {
+      instances: [url],
+    },
+  } as LxdOperation;
+};
+
+describe("getInstanceName", () => {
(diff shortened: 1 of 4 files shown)


src/util/operations.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it("snapshot operation", () => {
+    const operation = craftOperation(
+      "/1.0/instances/testInstance3/snapshots/testSnap"
+    );
+    const name = getInstanceName(operation);
+
+    expect(name).toBe("testInstance3");
+  });
```
lorumic, line 36: Maybe we can branch here as well, and check the snapshot operation both for the default and for a custom project?

src/util/operations.spec.tsx
lorumic, line 1: I would suggest to put test names that better match the "it" format - e.g. `it("correctly identifies an instance name from an operation in the default project")`, and so on.

PR #477, opened 2023-09-29
Overhaul storage section WD-6471

edlerd:
## Done

- overhaul the storage section following [this design](https://www.figma.com/file/akl8XNqp2K5egFKI3RIfoG/23.10-LXD---ISOs-and-Storage-Volumes?type=design&node-id=359-454&mode=design&t=zkJg5ePsCmTtdPYS-0)
- added pool selector to custom volume creation
- added pool editing

Fixes WD-6471

## QA

1. Run the LXD-UI:
    - check storage section with all subpages and detail pages
    - create pool / update / rename / delete it
    - create volume / update / rename / delete it
    - upload an iso / delete it

### DIFF
--- a/src/components/Pagination.tsx
+++ b/src/components/Pagination.tsx
@@ -61,16 +61,28 @@ const Pagination: FC<Props> = ({
   useEventListener("resize", resize);
   useEffect(resize, []);
 
+  const getVisibleCount = () => {
+    if (selectedNotification) {
+      return selectedNotification;
+    }
+
+    if (isSmallScreen) {
+      return `${visibleCount} out of ${totalCount}`;
+    }
+
+    if (visibleCount === totalCount && visibleCount > 1) {
+      return `Showing all ${totalCount} ${keyword}s`;
+    }
+
+    return `Showing ${visibleCount} out of ${totalCount} ${keyword}${
+      totalCount !== 1 ? "s" : ""
+    }`;
+  };
+
   return (
     <div className={classnames("pagination", className)} {...divProps}>
       <div className="description" id="pagination-description">
-        {selectedNotification
-          ? selectedNotification
-          : isSmallScreen
-          ? `${visibleCount}\xa0out\xa0of\xa0${totalCount}`
-          : `Showing ${visibleCount} out of ${totalCount} ${keyword}${
-              totalCount !== 1 ? "s" : ""
-            }`}
+        {getVisibleCount()}
       </div>
       <Button
         aria-label="Previous page"
--- a/src/pages/images/CachedImageList.tsx
+++ b/src/pages/images/CachedImageList.tsx
@@ -1,9 +1,10 @@
-import React, { FC } from "react";
+import React, { FC, useState } from "react";
 import {
   EmptyState,
   Icon,
   List,
   MainTable,
+  SearchBox,
   useNotify,
 } from "@canonical/react-components";
 import { humanFileSize, isoTimeToString } from "util/helpers";
@@ -15,10 +16,14 @@ import Loader from "components/Loader";
 import { useParams } from "react-router-dom";
 import CreateInstanceFromImageBtn from "pages/images/actions/CreateInstanceFromImageBtn";
 import { cachedLxdToRemoteImage } from "util/images";
+import ScrollableTable from "components/ScrollableTable";
+import { usePagination } from "util/pagination";
+import Pagination from "components/Pagination";
 
 const ImageList: FC = () => {
   const notify = useNotify();
   const { project } = useParams<{ project: string }>();
+  const [query, setQuery] = useState<string>("");
 
   if (!project) {
     return <>Missing project</>;
@@ -39,14 +44,28 @@ const ImageList: FC = () => {
 
   const headers = [
     { content: "Name", sortKey: "name" },
-    { content: "Architecture", sortKey: "architecture" },
-    { content: "Type", sortKey: "type" },
-    { content: "Upload date", sortKey: "uploaded_at" },
-    { content: "Size", sortKey: "size" },
+    {
+      content: "Architecture",
+      sortKey: "architecture",
+      className: "architecture",
+    },
+    { content: "Type", sortKey: "type", className: "type" },
+    {
+      content: "Upload date",
+      sortKey: "uploaded_at",
+      className: "uploaded_at",
+    },
+    { content: "Size", sortKey: "size", className: "u-align--right size" },
     { "aria-label": "Actions", className: "actions" },
   ];
 
-  const rows = images.map((image) => {
+  const filteredImages = images.filter(
+    (item) =>
+      !query ||
+      item.properties.description.toLowerCase().includes(query.toLowerCase()),
+  );
+
+  const rows = filteredImages.map((image) => {
     const actions = (
       <List
         inline
@@ -73,40 +92,50 @@ const ImageList: FC = () => {
           content: image.architecture,
           role: "cell",
           "aria-label": "Architecture",
+          className: "architecture",
         },
         {
           content: image.type,
           role: "cell",
           "aria-label": "Type",
+          className: "type",
         },
         {
           content: isoTimeToString(image.uploaded_at),
           role: "cell",
           "aria-label": "Upload date",
+          className: "uploaded_at",
         },
         {
           content: humanFileSize(image.size),
           role: "cell",
           "aria-label": "Size",
+          className: "u-align--right size",
         },
         {
           content: actions,
           role: "cell",
           "aria-label": "Actions",
-          className: "u-align--right",
+          className: "u-align--right actions",
         },
       ],
       sortData: {
         name: image.properties.description.toLowerCase(),
         architecture: image.architecture,
         type: image.type,
-        size: image.size,
+        size: +image.size,
         uploaded_at: image.uploaded_at,
       },
     };
   });
 
-  return !isLoading && images.length === 0 ? (
+  const pagination = usePagination(rows);
+
+  if (isLoading) {
+    return <Loader text="Loading cached images..." />;
+  }
+
+  return images.length === 0 ? (
     <EmptyState
       className="empty-state"
       image={<Icon name="mount" className="empty-state-icon" />}
@@ -115,15 +144,45 @@ const ImageList: FC = () => {
       <p>Images will appear here, when launching an instance from a remote.</p>
     </EmptyState>
   ) : (
-    <MainTable
-      headers={headers}
-      rows={rows}
-      paginate={30}
-      responsive
-      sortable
-      className="u-table-layout--auto"
-      emptyStateMsg={<Loader text="Loading images..." />}
-    />
+    <div className="cached-image-list">
+      <div className="upper-controls-bar">
+        <div className="search-box-wrapper">
+          <SearchBox
+            name="search-cached-images"
+            className="search-box margin-right"
+            type="text"
+            onChange={(value) => {
+              setQuery(value);
+            }}
+            placeholder="Search for cached images"
+            value={query}
+            aria-label="Search for cached images"
+          />
+        </div>
+      </div>
+      <Pagination
+        {...pagination}
+        id="pagination"
+        className="u-no-margin--top"
+        totalCount={images.length}
+        visibleCount={
+          filteredImages.length === images.length
+            ? pagination.pageData.length
+            : filteredImages.length
+        }
+        keyword="cached image"
+      />
+      <ScrollableTable dependencies={[images]}>
+        <MainTable
+          headers={headers}
+          rows={pagination.pageData}
+          sortable
+          className="cached-image-table"
+          emptyStateMsg="No images found matching this search"
+          onUpdateSort={pagination.updateSort}
+        />
+      </ScrollableTable>
+    </div>
   );
 };
 
--- a/src/pages/storage/EditStoragePool.tsx
+++ b/src/pages/storage/EditStoragePool.tsx
@@ -1,0 +1,124 @@
+import React, { FC, useState } from "react";
+import { useNotify, Row, Col, Button } from "@canonical/react-components";
+import { useQueryClient } from "@tanstack/react-query";
+import { updateStoragePool } from "api/storage-pools";
+import SubmitButton from "components/SubmitButton";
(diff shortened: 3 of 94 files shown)


piperdeck:
# 1

Is there a reason why volumes can't be sorted by "used by"?

# 2

For the pagination bar, when all pools/volumes/whatever are being shown, could we try the phrasing "Showing all 3 cached images" instead of "Showing 3 of 3 instances"? I feel that it reads a little more smoothly.

# 3

The "used by" column in the Pools table is not super intuitive to me. Could we please try a version like this:

Pools (1)

Note:

I'm not sure what the behaviour of this table is if there is only one project in the LXD server, but in that case I would use a different table which doesn't mention anything about projects, since this could introduce confusion to a user who isn't using projects and/or doesn't know what they are yet.

# 4

Is there a way for us to display the size of volumes in the volumes table? One of the reasons a user might be here could be to free up space by deleting volumes they don't need anymore.

I'm imagining that this might be complicated, for example because zfs deduplication means that the size of a volume is not necessarily the same as the amount of space that would be freed by deleting it. But it's worth thinking about.

# 5

The sort by type option doesn't seem to be working as expected. I would expect it to sort containers and container snapshots differently, but it seems to be treating them as one type.

image

# 6

So, this is a complicated issue. I would expect that when I select the filters "type:container" and "type:snapshot", that the list would be filtered to show only container snapshots. In other words, that the filters are applied with an AND operator between them. But this is not the case, as I get virtual machine snapshots in my filter results here:

image

Which leads me to conclude that the filter uses an OR operator. And I can see why it does this, since in the case that a user applied the "type:container" and "type:vm" filters, it's safe to assume they don't just want an empty list of all the volumes which are containers AND VMs.

But then again, when I search for another impossible combination "type:conatiner" and "content type:block", I do get an empty list. So it seems like the search terms are implicitly embedded into AND and OR phrases depending on which group of filters they're from.

Nevertheless, I think that it's legitimately confusing to have the behaviour in the screenshot. I'm not sure what a good solution to this is. I can think of a couple ideas:

- Allow the user to structure more complex filter rules, with ANDs, ORs, and parentheses where they want them.
    - But this is probably a pretty advanced feature and I'm not sure it would be worth the time and effort to build
- Apply new filter syntax so that the "volume type" filters are invisibly separated into two subgroups
    - But this may lead to confusion since the search bar doesn't actually say what operators it's using
- Find some way to display the filter syntax more explicitly so that users understand what filter they're applying.

I'm not really sure what the answer is. Would you be open to experimenting with some different filter behaviour?

piperdeck:
> We use the same way in profiles. Maybe it is an opportunity to learn about profiles? We could have a help icon linking to the project documentation.
> We could also have the simpler table if there is just 1 profile. I am afraid this introduces a bit of complexity to the code though, and we will have to check and think of all different ways the page might be rendered on future changes.

If it's going to introduce undue complexity, I'm happy leaving it as-is until I see testers being confused.

> We can get the size with an extra api call per volume. I added it to the detail page and the list. However we can't sort by it, because the list might be very long and we shouldn't have that many api calls.
> There are limitations. Snapshots do not give their size, as well as directory backed volumes.

This looks great, even with the limitations you mentioned. I think seeing sizes makes a big difference. I suspect that snapshots don't show their size because of the COW nature of ZFS, which makes sense.

> Anyhow let's take this as a dedicated issue outside this PR. It also affects the instance filter.

I agree completely, let's put a pin in this for future research

One more thing I'm noticing: The vertical alignment on some of these rows looks kind of crooked...

Frame 4 (3)

Would it be possible to make the row elements aligned with each other along the upper edge of the row? Ideally, the trashcan icon would be baseline aligned with text, while the storage usage graphic would be top-aligned with the text.

Otherwise, this is looking good!

edlerd:
> Would it be possible to make the row elements aligned with each other along the upper edge of the row? Ideally, the trashcan icon would be baseline aligned with text, while the storage usage graphic would be top-aligned with the text.

Thanks again for the review @piperdeck 
I adjusted the alignments and column widths for all the storage tables. Also made sure the numbers are all right aligned.

src/pages/images/CachedImageList.tsx
```diff
@@ -39,14 +44,24 @@ const ImageList: FC = () => {
...
+      className: "uploaded_at",
+    },
+    { content: "Size", sortKey: "size", className: "u-align--right size" },
     { "aria-label": "Actions", className: "actions" },
   ];
 
-  const rows = images.map((image) => {
+  const filteredImages = images;
+
+  const rows = filteredImages.map((image) => {
```
lorumic, line 64: Should filter based on the `query` here. Currently the query is set but not used to filter.

lorumic (changes requested):
Didn't to a full review yet, found the issue below on first pass.

src/pages/storage/StoragePools.tsx
```diff
@@ -108,24 +151,25 @@ const StoragePools: FC = () => {
...
+                className="u-table-layout--auto storage-pool-table"
+                emptyStateMsg={
+                  isLoading ? (
+                    <Loader text="Loading storage pools..." />
+                  ) : (
+                    "No data to display"
+                  )
+                }
+              />
+            </ScrollableTable>
```
lorumic, line 172: Adding the `ScrollableTable` is a step forward, but it still feels like something is missing. Switching tabs in the "Storage" page alone, I notice these discrepancies:

1. The absence of the search/search and filter.
2. The absence of pagination.

Would it be good to add these? The "Pools" tab looks "outdated" and inconsistent with the other pages/tabs otherwise.

edlerd: We don't expect more than a handful of pools, hence the tab is not paginated or filterable to make it lighter.

src/pages/images/CachedImageList.tsx
edlerd: Whoops, that is fixed now.

src/pages/storage/StoragePools.tsx
lorumic: Still doesn't seem right to have it everywhere else except here. @piperdeck what do you think about this?

lorumic: We can also treat it as a separate improvement. I see that other pages are inconsistent too: Profiles still have the pagination at the bottom, Networks don't have anything at all - not even the `ScrollableTable`... There's a lot of work to do to reach consistency :)

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -32,17 +32,31 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
-        <Input {...getFormProps("name")} type="text" label="Name" required />
+        <Input
+          {...getFormProps("name")}
+          type="text"
+          label="Name"
+          required
+          disabled={!formik.values.isCreating || formik.values.isReadOnly}
+          help={
+            !formik.values.isCreating
+              ? "Click the name in the header to rename the pool"
```
lorumic, line 43: Huh? The name in the header always says "Cannot rename storage pools".

src/pages/storage/StorageVolumesFilter.tsx
```diff
@@ -0,0 +1,115 @@
...
+const POOL = "Pool";
+const VOLUME_TYPE = "Volume type";
+const CONTENT_TYPE = "Content type";
+
+const StorageVolumesFilter: FC<Props> = ({ volumes, setFilters }) => {
+  const { state } = useLocation() as StorageFilterState;
+
+  useEffect(() => window.history.replaceState({}, ""), [state]);
+
+  const pools = [...new Set(volumes.flatMap((volume) => volume.pool))];
```
lorumic, line 47: `volume.pool` is a string, not an array of strings, so you don't need `flatMap` here:
```suggestion
  const pools = [...new Set(volumes.map((volume) => volume.pool))];
```

src/pages/storage/StorageVolumesFilter.tsx
```diff
@@ -0,0 +1,115 @@
...
+  "snapshot",
+  "image",
+  "custom",
+];
+
+const contentTypes: string[] = ["block", "filesystem", "iso"];
+
+const POOL = "Pool";
+const VOLUME_TYPE = "Volume type";
+const CONTENT_TYPE = "Content type";
```
lorumic, line 40: Very well done! :+1:

src/pages/storage/StorageVolumesFilter.tsx
```diff
@@ -0,0 +1,115 @@
...
+      heading: POOL,
+      chips: pools.map((pool) => {
+        return { lead: POOL, value: pool };
+      }),
+    },
+    {
+      id: 2,
+      heading: VOLUME_TYPE,
+      chips: volumeTypes.map((status) => {
+        return { lead: VOLUME_TYPE, value: status };
```
lorumic, line 61: Status?
```suggestion
      chips: volumeTypes.map((volumeType) => {
        return { lead: VOLUME_TYPE, value: volumeType };
```

src/pages/storage/StorageVolumesFilter.tsx
```diff
@@ -0,0 +1,115 @@
...
+      }),
+    },
+    {
+      id: 3,
+      heading: CONTENT_TYPE,
+      chips: contentTypes.map((type) => {
+        return { lead: CONTENT_TYPE, value: type };
```
lorumic, line 68: Maybe `contentType` is a bit more informative?
```suggestion
      chips: contentTypes.map((contentType) => {
        return { lead: CONTENT_TYPE, value: contentType };
```

src/components/Pagination.tsx
```diff
@@ -61,16 +61,28 @@ const Pagination: FC<Props> = ({
   useEventListener("resize", resize);
   useEffect(resize, []);
 
+  const getVisibleCount = () => {
+    if (selectedNotification) {
+      return selectedNotification;
+    }
+
+    if (isSmallScreen) {
+      return `${visibleCount}\xa0out\xa0of\xa0${totalCount}`;
```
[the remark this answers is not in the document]
edlerd: I remember testing on very small screens and this fixed an issue there.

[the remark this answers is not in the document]
lorumic: Can you show the specific case? I also tested on any screen size and didn't find any problem without the nbsp.

src/util/usedBy.tsx
```diff
@@ -31,8 +31,9 @@ export const filterUsedByType = (
       })
       .map((path) => {
         const url = new URL(`http://localhost/${path}`);
+        const encodedName = url.pathname.split("/").slice(-1)[0] ?? "";
         return {
-          name: url.pathname.split("/").slice(-1)[0] ?? "",
+          name: decodeURIComponent(decodeURIComponent(encodedName)),
```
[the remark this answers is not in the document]
edlerd: There was a bug in the used by routine, which is resolved, but not yet released: https://github.com/canonical/lxd/issues/12398

src/components/Pagination.tsx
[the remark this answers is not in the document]
edlerd: The instance list on mobile i.e. 400px width. Though there are more problems on such small screens and we don't usually optimize for those.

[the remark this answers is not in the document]
lorumic: Ok, it's not a big deal anyway, it's just that is quite hard to read as it is now, so maybe it's not worth it.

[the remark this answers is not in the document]
edlerd: yes, let's simplify it. For mobile we should better hide some other elements instead.

src/util/usedBy.tsx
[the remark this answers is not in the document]
edlerd: adding a comment referencing the bug.

edlerd:
Thanks for the great review with many simplifications and good suggestions @lorumic 
All things mentioned should be fixed, please give it a 2nd pass.

src/pages/storage/StorageVolumes.tsx
```diff
@@ -130,40 +175,78 @@ const StorageVolumes: FC = () => {
...
+    return <Loader text="Loading storage volumes..." />;
+  }
+
+  return volumes.length === 0 ? (
+    <EmptyState
+      className="empty-state"
+      image={<Icon name="mount" className="empty-state-icon" />}
+      title="No volumes found"
+    >
+      <p>Storage volumes will appear here</p>
```
lorumic, line 199: I wonder why in some empty states we have the formula "X will appear here", and in some other we have "There are no X (in/for/of Y)". :thinking:

src/util/usedBy.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+    expect(results[0].name).toBe("my_profile");
+    expect(results[0].project).toBe("foo");
+  });
+
+  it("decodes url encoded volume names", () => {
+    const paths = [
+      "/1.0/storage-pools/dir/volumes/custom/t%25C3%25BCdeld%25C3%25BC",
+    ];
+    const results = filterUsedByType("storage-pools", paths);
```
lorumic, line 33: It's a little bit confusing that the `type` is called "storage-pools", even though what we're really filtering by here is volumes.

lorumic (commented):
We're almost there, I found just a few more. :)

edlerd:
> We're almost there, I found just a few more. :)

It's getting simpler, I like it! Should all be resolved now.

src/pages/storage/EditStoragePool.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const notify = useNotify();
+  const queryClient = useQueryClient();
+  const { project } = useParams<{ project: string }>();
+  const controllerState = useState<AbortController | null>(null);
+
+  if (!project) {
+    return <>Missing project</>;
+  }
+
+  const StorageSchema = Yup.object().shape({
```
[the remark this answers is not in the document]
lorumic: It's still called `StorageSchema` in `CreateStoragePool`, would it be good to update it there as well?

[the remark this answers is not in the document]
edlerd: Changed it there as well.

lorumic (approved):
Thanks for applying all the changes! This became a monumental PR with over 90 files changed, but it's ready to go for me now. :)

PR #479, opened 2023-10-02
Use BaseLayout as main container wherever possible [WD-6364]

lorumic:
## Done

- Replaced the `main` element with the `BaseLayout` component throughout the app
- Minor code refactors related to the above change

Fixes [WD-6364](https://warthogs.atlassian.net/browse/WD-6364)

## QA

1. Run the LXD-UI:
    - Check that all the modified pages look as expected in terms of header and content appearance

[WD-6364]: https://warthogs.atlassian.net/browse/WD-6364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/BaseLayout.tsx
```diff
@@ -3,28 +3,33 @@ import classnames from "classnames";
 
 interface Props {
   title?: string | ReactNode;
-  controls?: ReactNode | null;
+  controls?: ReactNode;
+  customHeader?: ReactNode;
```
edlerd, line 7: This is creating an opinionated interface, namely when passing a `title`, the `customHeader` will be ignored. Maybe a better approach than making this `BaseLayout` cover all cases is to introduce a `CustomLayout` which receives the `header` component instead of `title` and `controls`?

src/components/BaseLayout.tsx
edlerd, line 1: I think there are a couple of more pages that can use a generic layout component. When searching for `<main ` the ProfileList, ClusterList and some Certificate components show up. They all look like they can also use the generic component.

edlerd (commented):
Looking good so far, two suggestions below.

src/components/BaseLayout.tsx
edlerd: Also we can make the `title` prop mandatory and the single exception of `Login` can also use the `CustomLayout`?

src/components/BaseLayout.tsx
lorumic: They all have the custom class applied to the `main` element, not to the `p-panel__content` div. So we cannot use the `contentClassName`.

src/sass/_detail_page.scss
```diff
@@ -0,0 +1,14 @@
+.detail-page {
+  margin-top: -$spv--medium;
+  padding-bottom: 0;
+}
+
+.instance-detail-title {
```
edlerd, line 6: It's good idea to have the shared content in a common scss file. This seems instance specific, maybe keep it in a dedicated file?

src/components/BaseLayout.tsx
lorumic: That would still leave out `InstanceList` which has a particular structure, with custom classes applied to... basically everything:
```
    <main className="l-main instance-list">
      <div
        className={classnames("p-panel", {
          "has-side-panel": !!panelParams.instance,
        })}
      >
        <div className="p-panel__header instance-list-header">
          <div className="instance-header-left">
...
```

lorumic: Also in the Certificate pages, we have the `is-sticky` class that is not applied by `BaseLayout`:
`<div className="p-panel__header is-sticky">`

edlerd: The `InstanceList` is quite custom. I think we could refactor it a bit to make it less custom. I.e. the `has-side-panel` can probably be applied to the main element as well. And the custom instance-list-header can probably be replaced by the `p-panel__header`. Though also fine to keep it as is and have a dedicated PR for making just this one generic.

The `Certificate*` components seem quite custom as well - the sticky also needs custom css for a white background and the z-index -- maybe keep it as is?

src/pages/instances/InstanceList.tsx
```diff
@@ -465,12 +466,12 @@ const InstanceList: FC = () => {
...
   return (
-    <main className="l-main instance-list">
-      <div
-        className={classnames("p-panel", {
-          "has-side-panel": !!panelParams.instance,
-        })}
-      >
+    <CustomLayout
+      mainClassName={classnames("instance-list", {
+        "has-side-panel": Boolean(panelParams.instance),
```
edlerd, line 471: I should have mentioned the css matching `has-side-panel` assumes it to be a child of `instance-list`, which changed now. Please fix that by changing the `.has-side-panel` to `&.has-side-panel`

lorumic: Right, good catch.

edlerd (approved):
LGTM, thanks :)

PR #480, opened 2023-10-04
Ensure certificate serial number is short enough for firefox

edlerd:
## Done

- Make generated certificates compatible to firefox, too long serial numbers cause problems in it. See [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1856925).
- Add hostname to the downloaded pfx and crt files
- Add hostname to certificate name

## QA

1. Run the LXD-UI:
    - create new certificates
    - import to firefox and test usage
    - ensure name contains hostname

### DIFF
--- a/src/pages/login/CertificateGenerate.tsx
+++ b/src/pages/login/CertificateGenerate.tsx
@@ -138,7 +138,12 @@ const CertificateGenerate: FC = () => {
                 {certs && (
                   <Col size={3}>
                     <Button
-                      onClick={() => downloadText("lxd-ui.crt", certs.crt)}
+                      onClick={() =>
+                        downloadText(
+                          `lxd-ui-${location.hostname}.crt`,
+                          certs.crt
+                        )
+                      }
                     >
                       Download crt
                     </Button>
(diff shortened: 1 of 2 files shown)


lorumic (commented):
Some QA notes:

1) "Scroll all the way down" it says in the tutorial for Firefox, but it's not at the very bottom. Maybe we could change this line to "Scroll down to the Certificates section, and then click the View Certificates button"?

2) For some reason, the spinner is not visible on Firefox while generating:

image

This works as expected on Chrome.

I cannot seem to be able to complete the process. When I restart the browser, I don't get asked to select the certificate by the browser. Neither on Firefox, nor on Chrome. Any idea why? Do I need to configure some auth setting in LXD?

edlerd:
> I cannot seem to be able to complete the process. When I restart the browser, I don't get asked to select the certificate by the browser. Neither on Firefox, nor on Chrome. Any idea why? Do I need to configure some auth setting in LXD?

Added the qa steps above. This only works when talking to LXD directly, the dev environment does not use the browsers client certificates.

edlerd:
> Some QA notes:

Addressed those just now.

lorumic:
> Test it against LXD directly under https://localhost:8443 with a locally UI enabled LXD

Can you elaborate on this a bit more? It seems to me that this tiny addition to the QA steps is still implying a lot of prior knowledge.

edlerd:
> > Test it against LXD directly under https://localhost:8443 with a locally UI enabled LXD
> 
> Can you elaborate on this a bit more? It seems to me that this tiny addition to the QA steps is still implying a lot of prior knowledge.

Follow the [steps from the readme](https://github.com/canonical/lxd-ui#install) to install LXD normally. Then use the new certificate that you generated from this branches dev environment.

lorumic (approved):

PR #482, opened 2023-10-09
Use websocket to track long running operations

edlerd:
## Done

- removed long-running requests to track instance operations
- use websocket to update on operation finish/fail

Fixes WD-6611

## QA

1. Run the LXD-UI:
    - check instance create, create and start, delete, start, stop, pause, unpause, also bulk actions, rename, update config, update snapshot config, attach/unattach iso, delete, bulk delete
    - the notifications and loading state should work as before.

### DIFF
--- a/src/sass/_instance_detail_snapshots.scss
+++ b/src/sass/_instance_detail_snapshots.scss
@@ -26,8 +26,7 @@
     }
   }
 
-  th:last-child,
-  td:last-child {
+  .p-table__expanding-panel {
     display: none !important;
   }
 
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -1,4 +1,3 @@
-import { TIMEOUT_120, TIMEOUT_60, watchOperation } from "./operations";
 import {
   handleEtagResponse,
   handleResponse,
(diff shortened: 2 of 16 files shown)


src/sass/_instance_detail_snapshots.scss
```diff
@@ -26,8 +26,7 @@
     }
   }
 
-  th:last-child,
-  td:last-child {
+  .p-table__expanding-panel {
```
lorumic, line 29: I cannot find this referenced anywhere.

lorumic (approved):
All looks good and works as expected - just one minor comment about a CSS class that seems unused.

I have found some bugs along the way, for which I've opened issues, so that they are tracked and can be addressed separately, since I think they are out of scope for this PR.

src/sass/_instance_detail_snapshots.scss
edlerd: When selecting a snapshot this hides the "x snapshots selected" line that causes the table to jump. The old css was causing a bug with the snapshot configuration, where it was hiding an important column. 
So I changed the CSS selector to fix it.
Not sure, it is ideal now, because on instances we have the multiselect just changed to have the pagination on top. I'll create a ticket to address the multiselect of snapshots as well.

lorumic: This `p-table__expanding-panel` CSS class does not seem to be used anywhere.

edlerd: It is coming in from the `MainTable` component in react components.

lorumic: Ok, that's the answer that I needed. :+1:

PR #483, opened 2023-10-09
Add documentation links to all main page headings WD-6525

edlerd:
## Done

- added an info icon linking to documentation for most main headings

Fixes WD-6525

## QA

1. Run the LXD-UI:
    - check the headings of all pages from the main navigation. Most should have an info icon linking to the docs of that page.

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -52,6 +52,7 @@ import ScrollableTable from "components/ScrollableTable";
 import NotificationRow from "components/NotificationRow";
 import SelectedTableNotification from "components/SelectedTableNotification";
 import CustomLayout from "components/CustomLayout";
+import HelpLink from "components/HelpLink";
 
 const loadHidden = () => {
   const saved = localStorage.getItem("instanceListHiddenColumns");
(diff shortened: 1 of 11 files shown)


piperdeck (changes requested):
I think this is a pretty safe change, but I'd make 2 minor changes:

1. I would change the icon to the "i" icon so that it matches the documentation icon in the side navigation
2. I would align the lower edge of the icon with the baseline of the header

Other than that, LGTM!

edlerd:
> I think this is a pretty safe change, but I'd make 2 minor changes:
> 
> 1. I would change the icon to the "i" icon so that it matches the documentation icon in the side navigation
> 2. I would align the lower edge of the icon with the baseline of the header

For 1. I changed the icon.
For 2. The icon was baseline aligned. I think because of the circle outline, it seemed to sit a bit too high from the baseline. I added a bit of top margin (which changes on very large screen widths, so better test it responsively with various widths). Wdyt?

piperdeck (approved):
Thanks for the additional info,

I think it probably also has to do with the fact that the Ubuntu Font is designed in such a way the the round bottoms of letters hang slightly below the baseline. I wouldn't worry about it too much, I think this is good to go.

src/pages/instances/InstanceList.tsx
```diff
@@ -478,7 +479,14 @@ const InstanceList: FC = () => {
       header={
         <div className="p-panel__header instance-list-header">
           <div className="instance-header-left">
-            <h1 className="p-heading--4 u-no-margin--bottom">Instances</h1>
+            <h1 className="p-heading--4 u-no-margin--bottom">
+              <HelpLink
+                href="https://documentation.ubuntu.com/lxd/en/latest/explanation/instances/#expl-instances"
+                title="Learn more about Instances"
```
lorumic, line 485:
```suggestion
                title="Learn more about instances"
```

lorumic (approved):
It's good, just one tiny suggestion.

PR #488, opened 2023-10-11
Ensure multiselect snapshots behaves the same as instances WD-6704

edlerd:
## Done

- Ensure multiselect snapshots behaves the same as instances

Fixes WD-6704

## QA

1. Run the LXD-UI:
    - open an instance detail page > snapshots
    - create some snapshots
    - ensure multiselect works as expected
    - multi delete some snapshots


piperdeck (dismissed):
# 1

For some reason, I can't seem to sort by "name" with this table

Other than that, this looks good to me :)

edlerd:
> For some reason, I can't seem to sort by "name" with this table

I fixed that.

edlerd:
> Actually, sorting by name on this table was disabled intentionally some time ago. See discussion [here](https://github.com/canonical/lxd-ui/pull/289#issuecomment-1509220027).

Thanks for finding this reference. I now remember the discussion back then. I'd support sorting by name today. Assuming, it is more frustrating to not be able to sort by name, than it is to get an unexpected sorting order. A user might expect alphanumeric order or not, we are using standard alphabetical sort, because we don't know what the user expects.

> Also, currently it doesn't seem to work as it should, see the screen capture:

This had an easy fix with lowercasing the strings.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -227,25 +229,14 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
...
-              </Button>
-            </>
+            <SnapshotBulkDelete
+              instance={instance}
+              snapshotNames={selectedNames}
+              onStart={() => setProcessingNames(selectedNames)}
+              onFinish={() => setProcessingNames([])}
+              onSuccess={onSuccess}
+              onFailure={onFailure}
+            />
```
lorumic, line 239: Shouldn't this be right-aligned, as the other CTAs are? It's a bit confusing to find this button on the left.

Peek 2023-10-12 11-57

What do you think @piperdeck ?

edlerd: I think it makes a lot of sense to keep the bulk actions on the left side together with the bulk select. This is also consistent with the instance list bulk actions.

lorumic: In the instance case it's different. There are a bunch of bulk actions so it's hard to miss them. In this case, I had to look around for a while, confused, to find it. This probably needs user research to confirm, but my opinion as in "the one who hadn't visited this page for a while", so more or less "fresh eyes", is that this is not the right place to put a button, especially given that in the base case you have the buttons on the opposite side. Let's see what @piperdeck thinks about it - maybe a third way is also possible.

piperdeck:
# 1

> In the instance case it's different. There are a bunch of bulk actions so it's hard to miss them. In this case, I had to look around for a while, confused, to find it. This probably needs user research to confirm, but my opinion as in "the one who hadn't visited this page for a while", so more or less "fresh eyes", is that this is not the right place to put a button, especially given that in the base case you have the buttons on the opposite side. Let's see what @piperdeck thinks about it - maybe a third way is also possible.

I think at some point we should revisit how the UI signals contextual list actions like this, perhaps by highlighting the action with a colour that matches the list selection colour. But since that's out of scope of this PR, I vote to right-align the CTA as @lorumic suggested, as I think that's the least confusing option in this case. I agree that it's inconsistent with the way it's done in the instance list, but there are other differences between this and the instance list already, so I think we may as well put the CTA in the spot where a user may have already registered the existence of CTAs.

# 2

When selecting snapshots from the list, the CTA to select all says "Select all 3 project snapshots". Is that a mistake? I'm not sure why it's calling them project snapshots, and would recommend changing the CTA to say "Select all 3 snapshots". (making sure that if the list is currently filtered, it still says "select all filtered snapshots")

# 3

I'm finding the padding around the pagination div to be a little too spacious to my eye.

After playing around, I think it looks better with `margin-top: 0` for the `pagination` div, and `margin-bottom: 0` for the `itemsPerPage` div:


image

edlerd:
All three points should be resolved, please have another pass for them.

piperdeck:
Great, looks good to me now!

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -255,9 +248,33 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
...
+            selectedNotification={
+              selectedNames.length > 0 && (
+                <SelectedTableNotification
+                  totalCount={instance.snapshots?.length ?? 0}
+                  itemName="snapshot"
+                  parentName="instance"
+                  selectedNames={selectedNames}
+                  setSelectedNames={setSelectedNames}
+                  filteredNames={filteredSnapshots.map(
+                    (instance) => instance.name
```
lorumic, line 269:
```suggestion
                    (snapshot) => snapshot.name
```

lorumic (approved):
Found just one naming issue, apart from that no problems.

PR #492, opened 2023-10-13
feat(operations): add test for getProjectName [WD-6612]

aaryanporwal:
## Done

- Add unit test for testing getProjectName, used in OperationList.
- Checks if project name is as expected from an instance operation when project parameter is present or not.

Fixes: [WD-6612](https://warthogs.atlassian.net/browse/WD-6612)

## QA

1. Run the LXD-UI:
    - Run `npm run test`

## Screenshots

[WD-6612]: https://warthogs.atlassian.net/browse/WD-6612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/util/operations.spec.tsx
+++ b/src/util/operations.spec.tsx
@@ -1,4 +1,4 @@
-import { getInstanceName } from "./operations";
+import { getInstanceName, getProjectName } from "./operations";
 import { LxdOperation } from "types/operation";
 
 const craftOperation = (url: string) => {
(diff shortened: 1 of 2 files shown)


src/util/operations.spec.tsx
```diff
@@ -44,3 +44,21 @@ describe("getInstanceName", () => {
...
+  it("identifies project name from an instance operation when no project parameter is present", () => {
+    const operation = craftOperation("/1.0/instances/testInstance1");
+    const name = getProjectName(operation);
+
+    expect(name).toBe("default");
+  });
+
+  it("identifies project name from an instance operation in a custom project", () => {
+    const operation = craftOperation(
+      "/1.0/instances/testInstance2?project=project"
```
edlerd, line 58: Better use a different value, to ensure we don't accidentally return the key and accept it in the test, something like `project=fooProject`.

src/util/operations.spec.tsx
```diff
@@ -44,3 +44,21 @@ describe("getInstanceName", () => {
...
+    );
+    const name = getProjectName(operation);
+
+    expect(name).toBe("project");
+  });
```
edlerd, line 63: Please add a third test for when project is not the first parameter, something like `/1.0/instances/testInstance1?foo=bar&project=barProject`. Looking at the implementation, this will likely fail, please fix the implementation for this case.

edlerd (commented):
Tests look good, some ideas to extend them below.

src/util/operations.spec.tsx
aaryanporwal: Updated the implementation

src/util/operations.spec.tsx
```diff
@@ -44,3 +44,30 @@ describe("getInstanceName", () => {
...
+
+  it("identifies project name from an instance operation in a custom project with other parameters", () => {
+    const operation = craftOperation(
+      "/1.0/instances/testInstance2?foo=bar&project=barProject"
+    );
+    const name = getProjectName(operation);
+
+    expect(name).toBe("barProject");
+  });
+});
```
edlerd, line 73: Thanks for extending the cases and fixing the implementation. I think there is still an uncovered edge case with this url: `/1.0/instances/testInstance1?foo=bar&project=barProject&baz=qux`

aaryanporwal: Fixed it. Thanks :)

edlerd (approved):
Looks good, thanks

PR #493, opened 2023-10-13
Move storage pool creation form from side panel to dedicated page [WD-6352]

lorumic:
## Done

- Moved storage pool creation form from side panel to dedicated page.

Fixes [WD-6352](https://warthogs.atlassian.net/browse/WD-6352)

## QA

1. Run the LXD-UI:
    - Go to storage page
    - Click on the "Create storage pool" button
    - Check the new form in a dedicated page: it should look and behave as expected

[WD-6352]: https://warthogs.atlassian.net/browse/WD-6352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edlerd (approved):
LGTM

lorumic:
@piperdeck Could you please do a design review when you have time? Thanks

edlerd (changes requested):
The demo build fails:
```
#14 8.763 [vite]: Rollup failed to resolve import "pages/instances/forms/FormMenuItem" from "/srv/src/pages/storage/forms/StoragePoolFormMenu.tsx".
#14 8.763 This is most likely unintended because it can break your application at runtime.
#14 8.763 If you do want to externalize this module explicitly add it to
#14 8.763 `build.rollupOptions.external`
#14 8.770 error during build:
#14 8.770 Error: [vite]: Rollup failed to resolve import "pages/instances/forms/FormMenuItem" from "/srv/src/pages/storage/forms/StoragePoolFormMenu.tsx".
#14 8.770 This is most likely unintended because it can break your application at runtime.
#14 8.770 If you do want to externalize this module explicitly add it to
#14 8.770 `build.rollupOptions.external`
#14 8.770     at viteWarn (file:///srv/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:48216:27)
#14 8.770     at onwarn (/srv/node_modules/@vitejs/plugin-react/dist/index.cjs:255:9)
#14 8.770     at onRollupWarning (file:///srv/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:48245:9)
#14 8.770     at onwarn (file:///srv/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:47976:13)
#14 8.770     at file:///srv/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:24276:13
#14 8.770     at Object.logger [as onLog] (file:///srv/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:25950:9)
#14 8.770     at ModuleLoader.handleInvalidResolvedId (file:///srv/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:24862:26)
#14 8.770     at file:///srv/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:24822:26
```
~~I guess something went bad on the merge resolve?~~
Edit: The import needs to pick from `components/forms`.

lorumic:
Demo is up and running with all the updates.

piperdeck:
The only issue I can find is that the description field should be a textarea, rather than a single-line field

lorumic:
> The only issue I can find is that the description field should be a textarea, rather than a single-line field

Good catch! I fixed that, and I also aligned all the description fields in other pages, which were still using the text input, to use textarea instead. Then I extracted the function to calculate the textarea's `rows`, so that it can be reused in the many places where it's needed, please have a look when you have time @edlerd

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -0,0 +1,86 @@
...
+            }
+            if (val === btrfsDriver) {
+              void formik.setFieldValue("source", "");
+            }
+            void formik.setFieldValue("driver", val);
+          }}
+          value={formik.values.driver}
+          required
+        />
+        <Input
```
edlerd, line 63: Can we use the more advanced `DiskSizeSelector` here?

src/util/formFields.tsx
```diff
@@ -28,3 +28,6 @@ export const optionRenderer = (
 
   return "";
 };
+
+export const getTextareaRows = (textLength?: number) =>
```
edlerd, line 32: This is very opinionated about the width of the textarea, which might be variable -- especially due to responsiveness.

Maybe introduce a new component `DynamicTextarea` (or find a better name), that wraps the Textarea. In it, we can use a ref to measure the actual width of the textarea and move `getTextareaRows`  into this component, to have the dynamic row height adjusting to the textarea width.

If it works well, we can later upstream it to react-components original `Textarea` with a boolean prop to opt into the row dynamics.

lorumic: Probably out of scope, let's consider this as a future improvement.

src/pages/storage/forms/StoragePoolFormMain.tsx
lorumic: We could but it's not a big gain. We'd have to add the label and help text manually. I don't know if it's worth it.

lorumic: Done, the `DiskSizeSelector` is now more flexible also for its use case in the storage volume creation form.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -0,0 +1,89 @@
...
+            }
+            if (val === btrfsDriver) {
+              void formik.setFieldValue("source", "");
+            }
+            void formik.setFieldValue("driver", val);
+          }}
+          value={formik.values.driver}
+          required
+        />
+        <DiskSizeSelector
```
edlerd, line 64: The rendering works fine, but we also need to handle the input correctly. See `src/pages/storage/forms/StorageVolumeForm.tsx:55` how it is done there, you can probably reuse the pattern.

edlerd (approved):
LGTM, thanks :)

PR #503, opened 2023-10-24
Boolean config options should have boolean default value

lorumic:
## Done

- Changed a string default value for 2 config options with type `bool`, so that no broken "Reset to default" button appears due to this type mismatch.

## QA

1. Run the LXD-UI:
    - Go to settings page
    - Check that `core.https_allowed_credentials` and `core.trust_ca_certificates` are showing a boolean value in read mode, and that the "Reset to default" functionality works as expected.

### DIFF
--- a/public/assets/data/config-options.json
+++ b/public/assets/data/config-options.json
@@ -150,7 +150,7 @@
     "key": "core.https_allowed_credentials",
     "type": "bool",
     "scope": "global",
-    "default": "",
+    "default": false,
     "description": "Whether to set Access-Control-Allow-Credentials HTTP header value to true"
   },
   {
@@ -241,7 +241,7 @@
     "key": "core.trust_ca_certificates",
     "type": "bool",
     "scope": "global",
-    "default": "",
+    "default": false,
     "description": "Whether to automatically trust clients signed by the CA"
   },
   {


edlerd (approved):

edlerd (commented):
When setting back to default it should null the value. This seems to not be possible any more. As the field description continues to be bold.

lorumic:
Seems that the back-end expects the empty string to reset these fields after all.

lorumic:
Opened an issue about this: https://github.com/canonical/lxd/issues/12459.
I can push a temporary fix, but maybe it's not so urgent and we can just wait that it is fixed upstream? What do you think? @edlerd

edlerd:
> Opened an issue about this: [canonical/lxd#12459](https://github.com/canonical/lxd/issues/12459). I can push a temporary fix, but maybe it's not so urgent and we can just wait that it is fixed upstream? What do you think? @edlerd

Thank you for finding this and filing the issue.

I wonder what to do with this PR -- we will remove the json file as soon as we start using the config API. So maybe we should close it?

lorumic:
I don't think we should close it. The fix for this on core side will probably be released much earlier than when we start using the config API. And I would discourage having a broken UI until then.
The 2 options that I see here are:
- The broken behaviour should be fixed now. Add a specific exception for the 2 fields, and remove it once it's fixed on core side.
- We can wait for the fix on core side, and then there's nothing to do here - hence ready to merge.

edlerd:
> We can wait for the fix on core side, and then there's nothing to do here - hence ready to merge

The fix is very simple and likely to be merged very soon, so let's keep this open and merge once the nightly build with the fix is available and works.

edlerd (approved):
All good now, thanks for picking up the ball and chasing the upstream fix as well :+1:

PR #504, opened 2023-10-24
Add storage pool and volume tests

edlerd:
## Done

- added CRUD tests for storage pools and volumes
- fix storage pool creation and editing in clustered environments

## QA

1. Run the LXD-UI:
    - ensure test suite passes

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -1,4 +1,8 @@
-import { handleEtagResponse, handleResponse } from "util/helpers";
+import {
+  handleEtagResponse,
+  handleResponse,
+  handleSettledResult,
+} from "util/helpers";
 import {
   LxdStoragePool,
   LxdStoragePoolResources,
@@ -10,13 +14,18 @@ import { LxdApiResponse } from "types/apiResponse";
 import { LxdOperationResponse } from "types/operation";
 import { TIMEOUT_300, watchOperation } from "api/operations";
 import axios, { AxiosResponse } from "axios";
+import { LxdClusterMember } from "types/cluster";
 
 export const fetchStoragePool = (
   pool: string,
   project: string,
+  target?: string,
 ): Promise<LxdStoragePool> => {
   return new Promise((resolve, reject) => {
-    fetch(`/1.0/storage-pools/${pool}?project=${project}&recursion=1`)
+    const targetParam = target ? `&target=${target}` : "";
+    fetch(
+      `/1.0/storage-pools/${pool}?project=${project}&recursion=1${targetParam}`,
+    )
       .then(handleResponse)
       .then((data: LxdApiResponse<LxdStoragePool>) => resolve(data.metadata))
       .catch(reject);
@@ -47,9 +56,14 @@ export const fetchStoragePoolResources = (
   });
 };
 
-export const createStoragePool = (pool: LxdStoragePool, project: string) => {
+export const createPool = (
+  pool: LxdStoragePool,
+  project: string,
+  target?: string,
+) => {
   return new Promise((resolve, reject) => {
-    fetch(`/1.0/storage-pools?project=${project}`, {
+    const targetParam = target ? `&target=${target}` : "";
+    fetch(`/1.0/storage-pools?project=${project}${targetParam}`, {
       method: "POST",
       body: JSON.stringify(pool),
     })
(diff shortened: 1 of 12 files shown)


src/api/storage-pools.tsx
```diff
@@ -71,6 +107,23 @@ export const updateStoragePool = (pool: LxdStoragePool, project: string) => {
   });
 };
 
+export const updateClusteredPool = (
+  pool: Partial<LxdStoragePool>,
+  project: string,
+  clusterMembers: LxdClusterMember[],
+): Promise<void> => {
```
lorumic, line 114: Is there a particular reason why the return type is specified here but not for the `createClusteredPool` above?

src/context/useClusterMembers.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const useClusterMembers = (): UseQueryResult<LxdClusterMember[]> => {
+  const { data: settings } = useSettings();
+  const isClustered = isClusteredServer(settings);
+
+  return useQuery({
+    queryKey: [queryKeys.cluster, queryKeys.members],
+    queryFn: fetchClusterMembers,
+    enabled: isClustered,
+  });
```
lorumic, line 17: Since we have this new hook, shouldn't we use it in all the other places where `fetchClusterMembers` is queried? This means, in addition to the components already changed so far, also these:
- ClusterGroupForm
- ClusterList
- MigrateInstanceBtn
- CreateNetwork

tests/helpers/storagePool.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export const visitPool = async (page: Page, pool: string) => {
+  await page.goto("/ui/");
+  await page.getByRole("link", { name: "Storage" }).click();
+  await page.getByRole("link", { name: pool }).first().click();
+};
+
+export const editPool = async (page: Page, profile: string) => {
+  await visitPool(page, profile);
```
lorumic, line 35:
```suggestion
export const editPool = async (page: Page, pool: string) => {
  await visitPool(page, pool);
```

tests/helpers/storageVolume.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export const deleteVolume = async (page: Page, pool: string) => {
+  await visitVolume(page, pool);
+  await page.getByRole("button", { name: "Delete volume" }).click();
+  await page
+    .getByRole("dialog", { name: "Confirm delete" })
+    .getByRole("button", { name: "Delete" })
+    .click();
+  await page.waitForSelector(`text=Storage volume ${pool} deleted.`, TIMEOUT);
```
lorumic, line 26:
```suggestion
export const deleteVolume = async (page: Page, volume: string) => {
  await visitVolume(page, volume);
  await page.getByRole("button", { name: "Delete volume" }).click();
  await page
    .getByRole("dialog", { name: "Confirm delete" })
    .getByRole("button", { name: "Delete" })
    .click();
  await page.waitForSelector(`text=Storage volume ${volume} deleted.`, TIMEOUT);
```

tests/helpers/storageVolume.ts
```diff
@@ -0,0 +1,48 @@
...
+  await page.getByRole("link", { name: "Storage" }).click();
+  await page.getByTestId("tab-link-Volumes").click();
+  await page.getByPlaceholder("Search and filter").fill(volume);
+  await page.getByPlaceholder("Search and filter").press("Enter");
+  await page.getByPlaceholder("Add filter").press("Escape");
+  await page.getByRole("link", { name: volume }).first().click();
+};
+
+export const editVolume = async (page: Page, profile: string) => {
+  await visitVolume(page, profile);
```
lorumic, line 40:
```suggestion
export const editVolume = async (page: Page, volume: string) => {
  await visitVolume(page, volume);
```

tests/storage.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  await deletePool(page, pool);
+});
+
+test("storage volume create, edit and remove", async ({ page }) => {
+  const volume = randomVolumeName();
+  await createVolume(page, volume);
+
+  await editVolume(page, volume);
+  await page.getByPlaceholder("Enter value").fill("50");
```
lorumic, line 36: Setting a 50 GiB size just for a test seems a bit risky. Maybe I'm paranoid, but I would set a smaller value, just in case. 5 is probably enough.

src/api/storage-pools.tsx
edlerd: A leftover from development

tests/storage.spec.ts
edlerd: The default is depending on the total size (for me, it is 30GB) and 50 ensures we do only extend it, because shrinking is not allowed.
I changed the creation to say 1GB, and here we put 2GB.

lorumic (approved):

PR #505, opened 2023-10-24
whitelist snapshot name characters

### LINKED ISSUE #334: UI Should Prohibit Certain Symbols from Snapshot Names
The UI should only allow valid characters for names when creating or editing Snapshots.

Creating:
lxd ui screenshot 3
Editing:
lxd ui screenshot 4

edlerd:
## Done

- add whitelist for snapshot names

Fixes #334

## QA

1. Run the LXD-UI:
    - check instance list > instance > snapshots > create snapshot form
    - ensure test suite is passing

### DIFF
--- a/src/util/snapshots.tsx
+++ b/src/util/snapshots.tsx
@@ -43,23 +43,6 @@ export const testDuplicateSnapshotName = (
   ];
 };
 
-export const testForbiddenChars = (): [
-  string,
-  string,
-  TestFunction<string | undefined, AnyObject>,
-] => {
-  return [
-    "forbiddenChars",
-    `The snapshot name cannot contain spaces or "/" characters`,
-    (value?: string) => {
-      if (!value) {
-        return true;
-      }
-      return !(value.includes(" ") || value.includes("/"));
-    },
-  ];
-};
-
 export const testValidDate = (): [
   string,
   string,
@@ -126,7 +109,10 @@ export const getSnapshotSchema = (
       .test(
         ...testDuplicateSnapshotName(instance, controllerState, snapshotName),
       )
-      .test(...testForbiddenChars()),
+      .matches(/^[A-Za-z0-9-_.:]+$/, {
+        message:
+          "Please enter only alphanumeric characters, underscores (_), periods (.), hyphens (-), and colons (:) in this field",
+      }),
     expirationDate: Yup.string()
       .nullable()
       .optional()


src/util/snapshots.tsx
```diff
@@ -126,7 +109,9 @@ export const getSnapshotSchema = (
       .test(
         ...testDuplicateSnapshotName(instance, controllerState, snapshotName),
       )
-      .test(...testForbiddenChars()),
+      .matches(/^[A-Za-z0-9-_]+$/, {
+        message: "Only alphanumeric, hyphen and underscore characters allowed",
+      }),
```
lorumic, line 114: `.` and `:` are allowed too.

edlerd: added them

aaryanporwal (approved):
lgtm

src/util/snapshots.tsx
```diff
@@ -126,7 +109,9 @@ export const getSnapshotSchema = (
       .test(
         ...testDuplicateSnapshotName(instance, controllerState, snapshotName),
       )
-      .test(...testForbiddenChars()),
+      .matches(/^[A-Za-z0-9-_.:]+$/, {
+        message: "Only alphanumeric and _ . - : characters allowed",
```
aaryanporwal, line 113: Can we make the message more readable and specific for the reader:

```suggestion
        message: "Only alphanumeric and _ . - : characters allowed",
```

src/util/snapshots.tsx
```diff
@@ -126,7 +109,9 @@ export const getSnapshotSchema = (
       .test(
         ...testDuplicateSnapshotName(instance, controllerState, snapshotName),
       )
-      .test(...testForbiddenChars()),
+      .matches(/^[A-Za-z0-9-_.:]+$/, {
```
lorumic, line 112: This is validating only the snapshot schema, hence for manual snapshot creation/edit only. Ideally, we should validate also the "Snapshot name pattern" config field. Otherwise we're solving only half of the problem - or even less, if we consider that scheduled snapshots are much more likely to be used than manual ones.

edlerd: The automatic pattern is not easily verifiable. The pogo2 template needs to be parsed and the resulting string put into this validator. But we don't have a parser at hand. So I'd keep that part untouched and fix the manual snapshot naming validation as reported in #334

lorumic: I guess that makes sense... at least until someone else opens an issue about the snapshot name pattern not being validated :)

PR #507, opened 2023-10-24
Add validation on instance create

### LINKED ISSUE #485: When adding a new network device, device name should be a required field
The addition of a new network device to an instance would fail on back-end side if no device name is specified. However, this field is currently treated as optional in the UI. It should be treated as required instead.

Screen capture:

Peek 2023-10-10 09-42

edlerd:
## Done

- added instance creation validation:
- for missing root storage
- missing network device name and
-  missing storage volume path 

Fixes #485

## QA

1. Run the LXD-UI:
    - Create an instance without root storage
    - Create an instance with a network device that has no name
    - Create an instance with a custom storage volume that has no mount point

### DIFF
--- a/.github/workflows/commits.yaml
+++ b/.github/workflows/commits.yaml
@@ -10,7 +10,7 @@ jobs:
     permissions:
       pull-requests: read   # to get list of commits from the PR
     name: Canonical CLA signed and Signed-off-by (DCO)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
     - name: Check if CLA signed
       uses: canonical/has-signed-canonical-cla@v1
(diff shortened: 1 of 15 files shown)


edlerd:
Hey @piperdeck 
I played a bit with your ideas for this and have another iteration ready.

See updated demo link above or the screencast link below for how it is now:

[Screencast from 17.11.2023 13:39:19.webm](https://github.com/canonical/lxd-ui/assets/1155472/8bf688be-1d85-4466-9037-db7bfc6d28cf)

piperdeck:
Discovered another area where we should catch missing information in instance config.

When you create a network device for an instance, it enforces giving the network a name but doesn't actually enforce choosing a network for that network device. It lets you click "save changes" but then throws an error. We should make the "Network" field mandatory as well.

edlerd:
> When you create a network device for an instance, it enforces giving the network a name but doesn't actually enforce choosing a network for that network device. It lets you click "save changes" but then throws an error. We should make the "Network" field mandatory as well.

Good point! I think the solution is to auto-select the 1st network when a network is added. That way, the Network selector can't be in an invalid state. I just pushed a change to do that. Please give the PR another pass @piperdeck

mas-who:
QA and code both looks good to me

PR #509, opened 2023-10-25
Preserve full sub-path when switching projects

lorumic:
## Done

- Improved the function to keep the context when switching project. Now the full sub-path is kept, allowing to preserve also the tab context of the "Storage" page.

## QA

1. Run the LXD-UI:
    - Visit "Storage" page, change tab, e.g. to "Volumes"
    - Switch project using the project selector in the side nav, and check that the tab context is preserved
    - Ensure that switching project from any page works as expected

### DIFF
--- a/src/util/projects.tsx
+++ b/src/util/projects.tsx
@@ -1,3 +1,4 @@
+import { STORAGE_TAB_PATHS } from "pages/storage/Storage";
 import { LxdProject } from "types/project";
 
 export const projectSubpages = [
@@ -12,9 +13,18 @@ export const projectSubpages = [
 
 export const getSubpageFromUrl = (url: string) => {
   const parts = url.split("/");
-  if (projectSubpages.includes(parts[4])) {
-    return parts[4];
+
+  const mainSubpage = parts[4];
+  const tabSubpage = parts[5];
+
+  if (mainSubpage === "storage" && STORAGE_TAB_PATHS.includes(tabSubpage)) {
+    return `${mainSubpage}/${tabSubpage}`;
   }
+
+  if (projectSubpages.includes(mainSubpage)) {
+    return mainSubpage;
+  }
+
   return undefined;
 };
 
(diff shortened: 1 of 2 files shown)


src/util/projects.tsx
```diff
@@ -13,7 +13,7 @@ export const projectSubpages = [
 export const getSubpageFromUrl = (url: string) => {
   const parts = url.split("/");
   if (projectSubpages.includes(parts[4])) {
-    return parts[4];
+    return parts.slice(4).join("/");
```
edlerd, line 16: This is too generic, i.e. when on a detail page instance, profile, network, etc. This asumes the detail page will exist for any project, which is not true, only for some subpages under the storage tab.

edlerd (changes requested):

src/util/projects.tsx
lorumic: Good point, I made it more specific. It should work as expected now.

src/util/projects.tsx
```diff
@@ -12,9 +14,19 @@ export const projectSubpages = [
 
 export const getSubpageFromUrl = (url: string) => {
   const parts = url.split("/");
-  if (projectSubpages.includes(parts[4])) {
-    return parts[4];
+
+  const mainSubpage = parts[4];
+  const tabSubpage = parts[5];
+  const slugifiedStorageTabs = STORAGE_TABS.map((tab) => slugify(tab));
```
edlerd, line 20: I don't like that we leak the slugify into this component, Can we define a `export const STORAGE_TAB_URLS = TABS.map((tab) => slugify(tab));` in `Storage.tsx` and then use it here directly?
Using the `STORAGE_TAB_URLS` could also simplify the `Storage.tsx` code a bit.

edlerd (commented):
It works now as expected. Some ideas to simplify below.

src/util/projects.tsx
lorumic: Sure, good idea.

edlerd (approved):
very good

PR #513, opened 2023-10-27
Ensure clustered backend pools are created and not left in pending state

edlerd:
## Done

- adding tests to ensure creating a storage pool on a clustered backend in created state. This would previously be a pending state for the pool, leaving it not fully usable.
- fixing pool creation for the cluster use case

## QA

1. Run the LXD-UI:
    - create a pool in a clustered backend (for the ZFS driver to work, the clustered test backend must not use containers, but VMs)
    - edit a pool in a clustered backend

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -56,7 +56,7 @@ export const fetchStoragePoolResources = (
 };
 
 export const createPool = (
-  pool: LxdStoragePool,
+  pool: Partial<LxdStoragePool>,
   project: string,
   target?: string,
 ) => {
@@ -72,6 +72,13 @@ export const createPool = (
   });
 };
 
+const getPoolForCluster = (pool: Partial<LxdStoragePool>) => {
+  const poolForCluster = { ...pool };
+  delete poolForCluster.config;
+
+  return poolForCluster;
+};
+
 export const createClusteredPool = (
   pool: LxdStoragePool,
   project: string,
@@ -84,6 +91,7 @@ export const createClusteredPool = (
       ),
     )
       .then(handleSettledResult)
+      .then(() => createPool(getPoolForCluster(pool), project))
       .then(resolve)
       .catch(reject);
   });
@@ -118,6 +126,7 @@ export const updateClusteredPool = (
       ),
     )
       .then(handleSettledResult)
+      .then(() => updatePool(getPoolForCluster(pool), project))
       .then(resolve)
       .catch(reject);
   });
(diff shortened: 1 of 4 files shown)


edlerd:
> When I try to create with a size value, I get this error:
Did you run the cluster on a VM or on containers?

> * Disable the size field, if it's not possible to set a fixed size because it's cluster member specific.
It should be possible. Did you review the code yet? Under the hood we are triggering requests for each cluster member to create a local pool with all fields. Afterwards one request to create the cluster wide pool. It could be the last request must not include a size though.

> * Disable ZFS driver option if the cluster is not VM-based - I don't know if this is feasible, but it's better than getting a postponed feedback, so it would be good to investigate a way to do it.
That is a good idea. Will try to find a way to know if the cluster nodes are running as container. If not, I am not sure about the relevance, it might be very uncommon to run a cluster on containers and its just us using this as a testing backend.

lorumic:
> Did you run the cluster nodes as containers or as VMs when testing?

VMs.

> Did you review the code yet? Under the hood we are triggering requests for each cluster member to create a local pool with all fields. Afterwards one request to create the cluster wide pool. It could be the last request must not include a size though.

Yes, I had already seen and understood the 3 changes in the code. And still, I'm always getting that error for a local cluster made of 3 VMs created following the guide in the [wiki](https://github.com/canonical/lxd-ui/wiki/Setup-local-LXD-cluster).

edlerd:
> Yes, I had already seen and understood the 3 changes in the code. And still, I'm always getting that error for a local cluster made of 3 VMs created following the guide in the [wiki](https://github.com/canonical/lxd-ui/wiki/Setup-local-LXD-cluster).

I see, thanks for providing the details. I think we have to strip the size (and likely other node specific properties) from the last request triggering the cluster-wide storage pool creation.

edlerd:
Adjusted the storage creation for clustered backends. It omits node-specific config for the cluster storage. Also updated the edit storage, which needs to also update the cluster wide pool and needs the same adjustment.

src/api/storage-pools.tsx
```diff
@@ -72,6 +72,13 @@ export const createPool = (
   });
 };
 
+const clusterPool = (pool: Partial<LxdStoragePool>) => {
```
lorumic, line 75: This name doesn't make it very clear what the function does. Maybe something more specific like `getPoolForCluster`?

lorumic (approved):
All good, just one tiny question about a function name.

src/api/storage-pools.tsx
edlerd: I was struggling to find a good name. Will go with your suggestion, that is def. better.

PR #514, opened 2023-10-27
feat(tab-links): Refactor into new TabLinks component [WD-6842]

aaryanporwal:
## Done

- Refactored all tab usages into a new TabLinks component to reduce code duplication across components.
- Made tabs links, that can be opened into a new tab etc.

## QA

1. Run the LXD-UI:
    - Go to storage page, instance, network, profile, storagepool, storagevolume detail pages and check if the tabs work well.
    - Check that the tabs can be right clicked and opened into a new tab etc. (because the tabs now have a href)

## Screenshots

Screen Recording showing tab links

### DIFF
--- a/src/components/TabLinks.tsx
+++ b/src/components/TabLinks.tsx
@@ -1,0 +1,40 @@
+import React, { FC } from "react";
+import { Tabs } from "@canonical/react-components";
+import { useNotify } from "@canonical/react-components";
+import { useNavigate } from "react-router-dom";
+import { slugify } from "util/slugify";
+
+interface Props {
+  tabs: string[];
+  activeTab?: string;
+  tabUrl: string;
+}
+
+const TabLinks: FC<Props> = ({ tabs, activeTab, tabUrl }) => {
+  const notify = useNotify();
+  const navigate = useNavigate();
+  const tabsPath = tabs.map((tab) => slugify(tab));
+
+  return (
+    <Tabs
+      links={tabs.map((tab, index) => {
+        const tabPath = tabsPath[index];
+        const href = tab === tabs[0] ? tabUrl : `${tabUrl}/${tabPath}`;
+
+        return {
+          label: tab,
+          id: tabPath,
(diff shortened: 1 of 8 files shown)


src/components/TabLinks.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useNavigate } from "react-router-dom";
+import { slugify } from "util/slugify";
+
+interface Props {
+  TABS: string[];
+  activeTab?: string;
+  tabUrl: string;
+}
+
+const TabLinks: FC<Props> = ({ TABS, activeTab, tabUrl }) => {
```
edlerd, line 13: The prop naming can be improved, how about `tabs, activeTab, baseUrl`? Prop names should always be `camlCase`.

src/components/TabLinks.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+        return {
+          label: tab,
+          id: tabPath,
+          active: tabPath === activeTab || (tab === TABS[0] && !activeTab),
+          onClick: (e) => {
+            e.preventDefault();
+            handleTabChange(tabUrl, tabPath);
+          },
+          href: tabPath === TABS[0] ? tabUrl : `${tabUrl}/${tabPath}`,
```
edlerd, line 41: Can we put this calculation into line 32, something like:
```
const href=tabPath === TABS[0] ? tabUrl : `${tabUrl}/${tabPath}`
```

Then we can use `href` directly.
And we can remove the `handleTabChange` function. Move the two lines `navigate(href)` and `notify.clear()` into the `onClick` handler.
This avoids the duplicate calculation of the href and should simplify the component a little bit.

edlerd (commented):
Looks excellent, thanks :)
Some ideas to simplify below.

src/components/TabLinks.tsx
aaryanporwal: Good idea 👍

edlerd (approved):
Very good, thank you for the changes :+1:

PR #527, opened 2023-11-09
Change instance filtering to url instead of state [WD-7026]

rubinaga:
## Done

- Changed instance filtering to use url search query parameters instead of state
- Changed eslintrc to ignore variables that start with underscore as they remain unused ( happens in the case of useSearchParams where the variable is not used in the file InstanceSearchFilter.tsx)

## QA

1. Run the LXD-UI:
    - Go to instances page
    - Apply any combination of filters
    - Check if they are applied correctly
    - Check if filters are applied when reloading the page or pasting an url with existing filters on it
    - Check if additional filters can be applied/removed in the case mentioned above

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -12,7 +12,7 @@ import { fetchInstances } from "api/instances";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import usePanelParams from "util/usePanelParams";
-import { useNavigate, useParams } from "react-router-dom";
+import { useNavigate, useParams, useSearchParams } from "react-router-dom";
 import Loader from "components/Loader";
 import { instanceCreationTypes } from "util/instanceOptions";
 import InstanceStatusIcon from "./InstanceStatusIcon";
@@ -29,7 +29,7 @@ import InstanceBulkActions from "pages/instances/actions/InstanceBulkActions";
 import { getIpAddresses } from "util/networks";
 import InstanceBulkDelete from "pages/instances/actions/InstanceBulkDelete";
 import InstanceSearchFilter from "./InstanceSearchFilter";
-import { InstanceFilters } from "util/instanceFilter";
+import { enrichStatuses } from "util/instanceFilter";
 import { isWidthBelow } from "util/helpers";
 import { fetchOperations } from "api/operations";
 import CancelOperationBtn from "pages/operations/actions/CancelOperationBtn";
@@ -53,6 +53,7 @@ import NotificationRow from "components/NotificationRow";
 import SelectedTableNotification from "components/SelectedTableNotification";
 import CustomLayout from "components/CustomLayout";
 import HelpLink from "components/HelpLink";
+import { LxdInstanceStatus } from "types/instance";
 
 const loadHidden = () => {
   const saved = localStorage.getItem("instanceListHiddenColumns");
@@ -73,12 +74,17 @@ const InstanceList: FC = () => {
   const { project } = useParams<{ project: string }>();
   const [createButtonLabel, _setCreateButtonLabel] =
     useState<string>("Create instance");
-  const [filters, setFilters] = useState<InstanceFilters>({
-    queries: [],
-    statuses: [],
-    types: [],
-    profileQueries: [],
-  });
+  // const location = useLocation();
+  const [searchParams] = useSearchParams();
+
+  const filters = {
+    queries: searchParams.getAll("queries"),
+    statuses: enrichStatuses(
+      searchParams.getAll("statuses") as LxdInstanceStatus[],
+    ),
+    types: searchParams.getAll("types"),
+    profileQueries: searchParams.getAll("profileQueries"),
+  };
   const [userHidden, setUserHidden] = useState<string[]>(loadHidden());
   const [sizeHidden, setSizeHidden] = useState<string[]>([]);
   const [selectedNames, setSelectedNames] = useState<string[]>([]);
@@ -501,11 +507,7 @@ const InstanceList: FC = () => {
               </>
             )}
             {hasInstances && selectedNames.length === 0 && (
-              <InstanceSearchFilter
-                key={project}
-                instances={instances}
-                setFilters={setFilters}
-              />
+              <InstanceSearchFilter key={project} instances={instances} />
             )}
           </div>
           {hasInstances && selectedNames.length === 0 && (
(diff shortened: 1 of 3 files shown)


src/pages/instances/InstanceList.tsx
```diff
@@ -73,12 +74,17 @@ const InstanceList: FC = () => {
...
+  const [searchParams] = useSearchParams();
+
+  const filters = {
+    queries: searchParams.getAll("queries"),
+    statuses: enrichStatuses(
+      searchParams.getAll("statuses") as LxdInstanceStatus[],
+    ),
+    types: searchParams.getAll("types"),
+    profileQueries: searchParams.getAll("profileQueries"),
+  };
```
lorumic, line 87: Now that these have become query parameters, they can probably get a more concise name: `queries` can become `q` or `query`, `statuses` can become `status` (we have one query parameter for each status that we want to filter, so it doesn't need the plural), `types` can be `type` and `profileQueries` can be `profile`. Also, we might want to explicitly type this object. So, in short, this can become:
```suggestion
  const filters: InstanceFilters = {
    queries: searchParams.getAll("q"),
    statuses: enrichStatuses(
      searchParams.getAll("status") as LxdInstanceStatus[],
    ),
    types: searchParams.getAll("type"),
    profileQueries: searchParams.getAll("profile"),
  };
```

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -56,24 +53,74 @@ const InstanceSearchFilter: FC<Props> = ({ instances, setFilters }) => {
...
+
+    filteredStatuses.forEach((status) => {
+      url.searchParams.append("statuses", status);
     });
+
+    filteredTypes.forEach((type) => {
+      url.searchParams.append("types", type);
+    });
+    profileQueries.forEach((profile) => {
+      url.searchParams.append("profileQueries", profile);
```
lorumic, line 87: See previous comment - if you decide to simplify the query params as suggested, these lines and the ones a bit below should also be updated.

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -56,24 +53,74 @@ const InstanceSearchFilter: FC<Props> = ({ instances, setFilters }) => {
...
+      .map((chip) => chip.value as LxdInstanceStatus);
+    const filteredTypes = searchData
+      .filter((chip) => chip.lead === "type")
+      .map((chip) => (chip.value === "VM" ? "virtual-machine" : "container"));
+
+    const profileQueries = searchData
+      .filter((chip) => chip.lead === "profile")
+      .map((chip) => chip.value);
+
+    const url = new URL(window.location.origin + window.location.pathname);
```
lorumic, line 73: Can't we use `window.location.href` directly here?

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -56,24 +53,74 @@ const InstanceSearchFilter: FC<Props> = ({ instances, setFilters }) => {
...
+        .filter(
+          (statusType) =>
+            statusType === "Running" ||
+            statusType === "Stopped" ||
+            statusType === "Frozen",
+        )
+        .map((status) => ({
+          lead: "status",
+          value: status,
+        })),
```
lorumic, line 107: This is filtering out the `Error` status. I don't think we need this filter at all, do we? Nor we need to enrich the statuses here. So this can become just:
```suggestion
      ...(searchParams.getAll("status") as LxdInstanceStatus[])
        .map((status) => ({
          lead: "status",
          value: status,
        })),
```

lorumic (commented):
Very well done :) I've left a few comments.

src/pages/instances/InstanceSearchFilter.tsx
edlerd: I think we don't need to construct the full url here, we could directly build the searchparams object.
Below we'd compare to the searchparams of the current url.

edlerd (commented):
This works really well, good work!
Some ideas to simplify the code in addition to @lorumic below.

rubinaga:
- Changed back the eslint file.
- Shortened the url parameters
- Refactored comparison from complete url to only search parameters
- Removed comment
- Removed unnecessary enrich of status

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -56,24 +49,71 @@ const InstanceSearchFilter: FC<Props> = ({ instances, setFilters }) => {
...
+    const filteredQueries = searchData
+      .filter((chip) => chip.quoteValue === true || chip.lead === "query")
+      .map((chip) => chip.value.toLowerCase());
+
+    const filteredStatuses = searchData
+      .filter((chip) => chip.lead === "status")
+      .map((chip) => chip.value as LxdInstanceStatus);
+    const filteredTypes = searchData
+      .filter((chip) => chip.lead === "type")
+      .map((chip) => (chip.value === "VM" ? "virtual-machine" : "container"));
```
edlerd, line 63: Rewriting it here has a wired side effect: When selecting VM type and reloading the page, we get containers instead of VMs. I think we can fix it by:
1. keep it here as `.map((chip) => chip.value)`.
2. move the rewrite to InstanceList: `types: searchParams.getAll("type").map((value) => value === "VM" ? "virtual-machine" : "container")`

edlerd (changes requested):

rubinaga:
- Fixed wrong type being applied on reload ( David's solution )

src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -64,59 +60,56 @@ const InstanceSearchFilter: FC<Props> = ({ instances }) => {
...
-    if (url.href !== window.location.href) {
-      setSearchParams(url.searchParams);
+    if (newSearchParams.toString() !== currentWindowSearchParams.toString()) {
+      setSearchParams(newSearchParams);
     }
   };
 
   const getChips = (): SearchAndFilterChip[] => {
     const searchParams = new URLSearchParams(window.location.search);
     const statusChipUrl = [
```
lorumic, line 94: I just realised that this now looks pretty similar to the items that we have below, inlined with the `newSearchData` array. I wonder if this should be inlined there too, instead of having, it alone, its own `const` declaration?

lorumic (approved):
It all looks good and works as expected. I have left a comment, but it's a minor thing. Feel free to implement it or not.

edlerd (approved):
LGTM, thanks for applying the changes :)

PR #529, opened 2023-11-13
Improve networking

edlerd:
## Done

- address various issues around network creation/editing discovered during in person review. Notes are in [this doc](https://docs.google.com/document/d/1BehzsvBhTVXrSrlSjFUoo7eMppzwmNLNfYQ_aDhfgi4/edit#heading=h.o2hu4ulgsezd)
- remove fan network type (simplifies bridge network handling)
- streamline copy on network forms
- fix ovn uplink network list to respect restricted.networks.uplinks setting for project, when that is unset, the non-ovn networks from the default project should be displayed in the list.

## QA

1. Run the LXD-UI:
    - created a bridge network
    - edit a bridge network
    - if ovn available, create / edit an ovn network
    - if ovn available, setup a project with restricted.networks.uplinks set. Enter some network names from the default project into  project configuration > Allow custom restrictions on a project level > Restrictions > Networks > Network uplinks. Only networks explicitly listed should be available as uplink, when creating an ovn network in that project.


lorumic (changes requested):
Please make the "Uplink" field a required one for OVN:

Peek 2023-11-14 13-54

edlerd:
> Please make the "Uplink" field a required one for OVN:

It can be empty, when no networks are available to the current user, the default network from the default profile will then be used. So changed it to be required, when at least 1 network to select from is available.

lorumic:
> It can be empty, when no networks are available to the current user, the default network from the default profile will then be used. So changed it to be required, when at least 1 network to select from is available.

Can we make it disable the "Create" button, just like the "Name" field does when you input a name that is already assigned to another network?

Peek 2023-11-14 14-36

lorumic:
I still have problems creating the OVN network. I get the "Option "network" value "UPLINK" is not one of the allowed uplink networks in project" error, even though I added it (and it is indeed correctly shown in the dropdown). See the screen capture below:

Peek 2023-11-15 09-53

edlerd:
> I still have problems creating the OVN network. I get the "Option "network" value "UPLINK" is not one of the allowed uplink networks in project" error, even though I added it (and it is indeed correctly shown in the dropdown). See the screen capture below:

I don't really know why this is happening, I suspect a wrong error message from the backend. With the [OVN setup described in the wiki](https://github.com/canonical/lxd-ui/wiki/Setup-OVN-cluster) I can create OVN networks through the UI.

lorumic:
Thanks, with microcloud it works. Probably I misconfigured something when I enabled OVN in the other cluster.

src/pages/networks/forms/NetworkSelector.tsx
```diff
@@ -4,31 +4,47 @@ import Loader from "components/Loader";
...
-  const options = networks.map((network) => {
+  const { data: project, isLoading: isProjectLoading } = useQuery({
+    queryKey: [queryKeys.projects, projectName],
+    queryFn: () => fetchProject(projectName),
+  });
+
+  const availableNetworks =
+    project?.config?.["restricted.networks.uplinks"]?.split(",") ||
+    networks
+      .filter((network) => ["bridge", "physical"].includes(network.type))
```
lorumic, line 33: I don't know if this is expected to be stable, but just in case I would suggest declaring a `const` for it above:
```
const UI_NETWORK_TYPES = ["bridge", "physical"];
```
And then this line becomes:
```suggestion
      .filter((network) => UI_NETWORK_TYPES.includes(network.type))
```
I think it's cleaner this way.

lorumic (approved):
Looks good overall, just a couple of small suggestions.

src/pages/networks/forms/NetworkSelector.tsx
edlerd: This component is heavily opinionated to be an OVN uplink selector. The types here are the ones supported by LXD as OVN uplinks. I don't think they will change much. I'll rename the component.

lorumic: Regardless of that, it's still tidier to have them declared in a `const` rather than hardcoded in the middle of the logic.

lorumic (approved):

PR #533, opened 2023-11-14
Link to local docs

edlerd:
## Done

- docs are now shipped with lxd, link to the local docs,  when server version is bigger than 5.18

## QA

1. Run the LXD-UI:
    - check documentation links for a lxd backend with a recent version (bigger than 5.18) -- spot checking should be enough and code review that they are unchanged apart from the base url.


lorumic (approved):
All good :+1:

PR #536, opened 2023-11-14
Use regular link from profile detail to instances with that profile

edlerd:
## Done

- Use a regular link from profile detail page to the instance list
- simplify the instance filters

## QA

1. Run the LXD-UI:
    - check link from profile detail page to instances in used by section
    - check instance filters above instance list (also with page reload)


src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -49,86 +42,43 @@ const InstanceSearchFilter: FC<Props> = ({ instances }) => {
...
 
-    const currentWindowSearchParams = new URLSearchParams(
-      window.location.search,
-    );
-    const newSearchParams = new URLSearchParams();
-    filteredQueries.forEach((query) => {
-      newSearchParams.append("q", query);
+    const newParams = new URLSearchParams();
+    searchTokensByLead("quote").forEach((query) => {
+      newParams.append("q", query);
```
lorumic, line 52: You renamed the lead to `quote`, but kept the value of the `forEach` argument as `query`. This is quite confusing. I advise to keep a consistent naming.

src/util/instanceFilter.tsx
```diff
@@ -2,7 +2,7 @@ import { LxdInstanceStatus } from "types/instance";
 
 export interface InstanceFilters {
   queries: string[];
-  statuses: LxdInstanceStatus[];
+  statuses: string[];
```
lorumic, line 5: I don't understand this change. We have a limited (and known) range of possible values for this, expressed by the `LxdInstanceStatus`. Why not using it?

src/pages/instances/InstanceSearchFilter.tsx
edlerd: `searchTokensByLead` gives the tokens for a given lead. In this exampl ethe lead is quote, so it returns all requies. In the forEach we iterate over all queries and each item is a query.

src/util/instanceFilter.tsx
edlerd: The filters are stored in the url, which is string only. This change avoids casting to and from string, when translating the statuses to and from the searchparameters. Because the url parameters can't be typed as status.

src/pages/instances/InstanceSearchFilter.tsx
lorumic: The lead is quote because you renamed it to quote. It used to be named query, and in that case the name was consistent. Now it is not.

src/util/instanceFilter.tsx
lorumic: I can let this go, but I still see it as a step backwards.

src/pages/instances/InstanceSearchFilter.tsx
edlerd: Ah, right. I bamboozled the two strings. Will fix it shortly.

src/util/instanceFilter.tsx
edlerd: Actually we can keep the status type as well. Through the simplifications, there is just a single cast left.

src/pages/instances/InstanceSearchFilter.tsx
lorumic: It's really a nitpick because the lead is invisible for this group of filters, but it left my brain confused for some seconds when reading the code because I was expecting a consistency between the lead name and the argument name of the `forEach` function.

edlerd: Yes and there was a bug, when you refreshed, the queries would get lost, they need to be without lead and have the `quoteValue:true` parameter instead.

lorumic (approved):
Thanks for applying the changes! :+1:

PR #538, opened 2023-11-15
Add local docs for images page

lorumic:
## Done

- Added the local docs logic to the Images page

## QA

1. Run the LXD-UI:
    - Navigate to the Images page
    - Check that, if LXD version is above 5.18, a local docs link is present for the info icon in the header instead of the remote one


edlerd (approved):
good catch, this was missed with the subsequent PRs.

PR #539, opened 2023-11-16
Show "Never" when instance has never been used

lorumic:
## Done

- Show "Never" when instance has never been used instead of "Jan 1, 1970 ..."

## QA

1. Run the LXD-UI:
    - Go to the detail page of an instance that was never used, e.g. [this](https://lxd-ui-539.demos.haus/ui/project/default/instances/detail/credible-chamois)
    - Check that "Last used" says "Never" instead of "Jan 1, 1970"

### DIFF
--- a/src/util/helpers.tsx
+++ b/src/util/helpers.tsx
@@ -14,7 +14,12 @@ export const isoTimeToString = (isoTime: string) => {
     return "";
   }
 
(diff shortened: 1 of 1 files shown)


edlerd (approved):
Nice, thanks for fixing this.

PR #540, opened 2023-11-16
Avoid conflict between filters and side panel params

lorumic:
## Done

- Fixed a conflict between the filters query parameters and the side panel query parameters, that essentially broke the "Instance summary" side panel, making it impossible to open - see screen capture below:

Peek 2023-11-16 13-56

## QA

1. Run the LXD-UI:
    - Go to instance list
    - Click on a row in the table - check that it opens the side panel correctly
    - Try adding some filters - with the side panel still open
    - Try reloading the page
    - Check that both types of query parameters (those related to filters and those related to the side panel) are correctly preserved

### DIFF
--- a/src/pages/instances/InstanceSearchFilter.tsx
+++ b/src/pages/instances/InstanceSearchFilter.tsx
@@ -8,6 +8,8 @@ import {
 } from "@canonical/react-components/dist/components/SearchAndFilter/types";
 import { useSearchParams } from "react-router-dom";
 
+const QUERY_PARAMS = ["query", "status", "type", "profile"];
+
 interface Props {
   instances: LxdInstance[];
 }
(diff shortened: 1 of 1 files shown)


src/pages/instances/InstanceSearchFilter.tsx
```diff
@@ -44,50 +46,52 @@ const InstanceSearchFilter: FC<Props> = ({ instances }) => {
...
+
+    QUERY_PARAMS.forEach((param) => {
+      const values = searchValuesByLead(param);
+      if (values.length === 0) {
+        newParams.delete(param);
+      }
+      values.forEach((value) => {
+        if (!newParams.has(param, value)) {
+          newParams.append(param, value);
+        }
```
edlerd, line 66: When having multiple chips for the same category selected and removing some (but not all), this doesn't remove the value from the params.

I.e. filter "running" or "stopped" and remove "stopped", this last change will not be reflected in the URL params.

edlerd (changes requested):
Thanks for picking this up, I realized this earlier, good to see already a fix for it :)

Though one edge case is not working well yet, see comment below.

src/pages/instances/InstanceSearchFilter.tsx
lorumic: Good catch! Should be fixed now.

edlerd (approved):
LGTM

PR #543, opened 2023-11-17
Use config api for server settings WD-7389 instance defaults WD-7390 and project defaults

edlerd:
## Done

- rely on the config api for:
    - server settings (used to be hard coded in a json file)
    - instance form defaults and help texts (used to be hard coded in a js object)
    - profile form defaults and help texts (used to be hard coded in a js object)
    - project form defaults and help texts (used to be hard coded in a js object)

Fixes WD-7389 WD-7390

## QA

1. Run the LXD-UI:
    - check settings page
        - edit settings
        - reset settings
    - check create instance form: all default values (From: LXD) and help texts (the gray explanation)
      - create instance > advanced > all sections and all possible overrides
    - check create profile form: all default values (From: LXD) and help texts (the gray explanation)
      - profiles > create profile > advanced > all sections and all possible overrides
    - check edit project form: all default values (From: LXD) and help texts (the gray explanation)
      - configuration > edit > all sections and all possible overrides


src/util/config.tsx
```diff
@@ -0,0 +1,101 @@
+import { LxcConfigOptionCategories, LxdConfigField } from "types/config";
+
+export const toConfigFields = (
+  categories: LxcConfigOptionCategories,
+): LxdConfigField[] => {
+  return Object.values(categories).flatMap((value) =>
```
mas-who, line 6: just a personal preference so not critical to change, I think the function might be easier to read if we use normal for loops. Having the tests helps with understanding though

tests/helpers/configuration.ts
```diff
@@ -84,9 +90,14 @@ export const setMemLimit = async (
 };
 
 export const setSchedule = async (page: Page, value: string) => {
-  await activateOverride(page, "Schedule - From: LXD");
+  await activateOverride(
+    page,
+    "Schedule Schedule for automatic instance snapshots - From: LXD",
```
mas-who, line 95: This seems a bit weird, "Schedule" is duplicated for the name attribute?

mas-who (commented):
Just some minor comments, I'd feel more comfortable if someone else also have a review since I may not be the most familiar with the code yet

src/util/config.tsx
edlerd: True, this used to be hard to read before. I'll change it.

tests/helpers/configuration.ts
mas-who: Yeah that makes sense :+1:

mas-who:
Checked during QA:

- Created a project `test-config-api-project` and checked every setting for edit saves and reset to default value.
- Checked create instance form, create profile form and edit project form for available settings as well as overrides. All settings and override options seem to display correct explanations.
- Checked all links within setting explanations that they navigate to the correct documentation.

mas-who (approved):

src/pages/projects/forms/ProjectResourceLimitsForm.tsx
```diff
@@ -42,37 +41,33 @@ const ProjectResourceLimitsForm: FC<Props> = ({ formik }) => {
...
           name: "limits_virtual_machines",
           label: "Max number of VMs",
-          help: "Maximum number of VMs that can be created in the project",
           defaultValue: "",
           children: <Input placeholder="Enter number" min={0} type="number" />,
         }),
 
         getInstanceConfigurationRow({
-          formik: formik as unknown as SharedFormikTypes,
+          formik: formik,
```
lorumic, line 68:
```suggestion
          formik,
```

src/pages/projects/forms/ProjectResourceLimitsForm.tsx
```diff
@@ -84,37 +79,33 @@ const ProjectResourceLimitsForm: FC<Props> = ({ formik }) => {
...
-          label: "Max sum of individual memory limits",
-          help: "Maximum value for the sum of individual limits.memory configurations set on the instances of the project",
+          label: "Max sum of memory limits",
           defaultValue: "",
           children: <Input placeholder="Enter number" min={0} type="number" />,
         }),
 
         getInstanceConfigurationRow({
-          formik: formik as unknown as SharedFormikTypes,
+          formik: formik,
```
lorumic, line 106:
```suggestion
          formik,
```

src/pages/settings/Settings.tsx
```diff
@@ -78,14 +86,15 @@ const Settings: FC = () => {
...
-      lastGroup = group;
+      const isNewCategory = lastCategory !== configField.category;
+      lastCategory = configField.category;
 
       return {
         columns: [
           {
-            content: isNewGroup && <h2 className="p-heading--5">{group}</h2>,
+            content: isNewCategory && (
+              <h2 className="p-heading--5">{configField.category}</h2>
```
lorumic, line 96: This has already been assigned to the `lastCategory` variable at this point, right? So we should be allowed to do:
```suggestion
              <h2 className="p-heading--5">{lastCategory}</h2>
```

src/pages/settings/Settings.tsx
```diff
@@ -131,39 +141,34 @@ const Settings: FC = () => {
             href={`${docBaseLink}/server/`}
             title="Learn more about server configuration"
           >
-            Settings
+            Server settings
```
lorumic, line 144: I wonder if this is just more confusing for the user? I preferred the old version. It also matches better the entry in the side navigation.

src/types/config.d.ts
```diff
@@ -1,9 +1,32 @@
...
+    }[];
+  };
+}
+
+export interface LxdConfigOptions {
+  configs: {
+    cluster: LxcConfigOptionCategories;
+    instance: LxcConfigOptionCategories;
+    project: LxcConfigOptionCategories;
+    server: LxcConfigOptionCategories;
```
lorumic, line 30: Related to Mason's question above - why are these 4 hard-coded here? Also, only `cluster` among them seems to be a "real" category that appears in the settings page. The other 3 are not there, and I don't see a bunch of others that are there instead (acme, core, images, etc).

src/util/config.tsx
```diff
@@ -0,0 +1,108 @@
...
+      ?.replaceAll("-", " ");
+    result = result.replaceAll(
+      key + "`",
+      `<a href="${href}" target="_blank" rel="noreferrer">${linkText}</a>`,
+    );
+  });
+
+  // code blocks
+  let count = 0;
+  while (result.includes("`") && count++ < 100) {
```
lorumic, line 103: Is this `100` an arbitrary limit, or where does it come from?

src/util/instanceConfigInheritance.tsx
```diff
@@ -107,11 +24,26 @@ export const figureInheritedValue = (
...
+    } else {
+      return ["1", "LXD (VM)"];
+    }
+  }
+
+  if (configKey === "limits.memory" && values.type === "instance") {
+    if (values.instanceType === "container") {
+      return ["-", "LXD (container)"];
+    } else {
+      return ["1GB", "LXD (VM)"];
```
lorumic, line 39: Where do these "magic values" for VMs come from?

src/pages/projects/forms/ClusterRestrictionForm.tsx
```diff
@@ -32,19 +31,17 @@ const ClusterRestrictionForm: FC<Props> = ({ formik }) => {
     <InstanceConfigurationTable
       rows={[
         getInstanceConfigurationRow({
-          formik: formik as unknown as SharedFormikTypes,
+          formik: formik,
```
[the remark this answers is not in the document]
edlerd: Thank you for flagging all of those, it seems simpler to search and replace `formik: formik,` with `formik,` locally to catch all the 86 occurrances.

src/components/forms/InstanceConfigurationRow.tsx
```diff
@@ -12,16 +12,20 @@ import { LxdProfile } from "types/profile";
...
   name: string;
   label: string | ReactNode;
   children: ReactElement;
   defaultValue?: string | CpuLimit | MemoryLimit;
   disabled?: boolean;
   readOnlyRenderer?: (value: unknown) => string | ReactNode;
-  help?: string;
 }
 
 export const getInstanceConfigurationRow = ({
```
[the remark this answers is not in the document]
edlerd: I am not happy with the whole config row and table architecture, either. Was thinking about this a while ago and it should be possible to simplify the components. I'll add a ticket to make this nicer. Currently, I'd keep it as is though -- for renaming I have trouble finding a good name. I guess when refactoring, the right components with proper names would pop out of that process.

src/pages/settings/Settings.tsx
```diff
@@ -7,50 +7,48 @@ import {
...
+  const getValue = (configField: ConfigField): string | undefined => {
     for (const [key, value] of Object.entries(settings?.config ?? {})) {
       if (key === configField.key) {
         return value;
       }
     }
-    if (typeof configField.default === "boolean") {
-      return configField.default ? "true" : "false";
+    if (configField.type === "bool") {
+      return configField.default === "true" ? "true" : "false";
```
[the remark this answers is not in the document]
edlerd: Nope, the default field might be empty string as well and we assume a default of "false" in that case.

src/pages/settings/Settings.tsx
```diff
@@ -64,8 +62,18 @@ const Settings: FC = () => {
...
 
-  let lastGroup = "";
-  const rows = configOptions
+  const configFields = toConfigFields(configOptions?.configs.server ?? {});
+
+  configFields.push({
+    key: "user.ui_title",
+    category: "user",
+    default: "",
+    shortdesc: "Title for LXD-UI in the browser. Shows hostname when unset.",
```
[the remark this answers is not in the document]
edlerd: Yeah, that is better!

src/pages/settings/Settings.tsx
edlerd: I am hesitant for this change, from reading I'd stumble -- why print the lastCategory here and not the current one. Probably better to keep that variable just to figure out if we reached a category boundry.

src/types/config.d.ts
edlerd: The response is structured in multiple hierarchies. Here on the top level we use only the part under server for the server settings. The part under instance holds configuration options for instances (like help texts and defaults).

The categories for the server settings are all under the server key here.

src/util/config.tsx
edlerd: Yes, I put it in to avoid infinite loops. It's a tactic of defensive programming.

src/util/instanceConfigInheritance.tsx
edlerd: Cpu and memory limit are special. The defaultDesc from the config api has a value of ``1Gib` (VMs)`. We can't parse it, so we have those two still hard coded in the UI with some logic around. I think it's reasonable to have the two exceptions.

src/pages/settings/Settings.tsx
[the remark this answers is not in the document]
lorumic: It should never be empty. We had [a case](https://github.com/canonical/lxd/issues/12459) in the past and it was considered a bug on core side.

src/components/forms/InstanceConfigurationRow.tsx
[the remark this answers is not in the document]
lorumic: As you prefer.

src/pages/settings/Settings.tsx
edlerd: True, especially with the UI setting in it, this is better with the shorter version.

src/types/config.d.ts
lorumic: It's very confusing, because they're at the top level when parsed in the settings form.

src/util/config.tsx
lorumic: Maybe put it as a `const` at the top with a comment explaining the choice, so that we don't have magic values scattered in the code?

src/util/instanceConfigInheritance.tsx
lorumic: Fair enough.

lorumic: The problem is that if they change on core side in the future, and we don't know/don't realise, we'll have an inconsistent value here.

src/types/config.d.ts
edlerd: Yeah it is a bit confusing from the server type here, the change is due to [this transition](https://github.com/canonical/lxd-ui/pull/543/files#diff-3ce5e2542470af50d3ea4bf0edb3ce1883fe5bb0b4300a5e90d808557cca83ecR65). I think the transition is good, because handle the transformed ConfigFields is way easier to handle in the UI, than dealing with the raw structure as returned from the API.

src/components/forms/InstanceConfigurationRow.tsx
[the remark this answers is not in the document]
lorumic: Ah, wait - nevermind. I see that we have another `getConfigurationRow` with different semantics, and even another `ConfigurationTable`. It's more tangled than I thought. Feel free to keep it as it is - possibly with a short term plan of improving it so that we don't have `Instance` things used by project and so on.

src/pages/settings/Settings.tsx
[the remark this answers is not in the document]
edlerd: Ticket reported as https://github.com/canonical/lxd/issues/12583

PR #561, opened 2023-11-28
chore(tests) created tests for instance and profile panels WD-7370

mas-who:
## Done

- Added tests for instance and profile side panels


edlerd (approved):
Code looks good, well done :)
Let's wait for the test suite to finish -- when it passes here, we can merge it.

PR #563, opened 2023-11-29
feat(storage) crud for custom volume snapshots [WD-7576]

mas-who:
## Done

- removed snapshot volume type from volumes list filter

## QA

1. Run the LXD-UI:
    - TBC


src/util/storageVolumeTable.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const POOL_COL = "Pool";
+export const TYPE_COL = "Type";
+export const CONTENT_TYPE_COL = "Content Type";
+export const CREATED_AT_COL = "Create At";
+export const SIZE_COL = "Size";
+export const USED_BY_COL = "Used By";
+export const SNAPSHOTS_COL = "Snapshots";
+export const ACTIONS_COL = "Actions";
+
+export const COLUMN_WIDTHS: Record<string, number> = {
```
edlerd, line 11: Not sure if this approach works with all different screen widths, we should be responsive for all sizes (though we don't focus on very narrow sizes below 800px).

src/pages/storage/StorageVolumes.tsx
```diff
@@ -65,42 +77,65 @@ const StorageVolumes: FC = () => {
...
       className: "actions",
       "aria-label": "Actions",
+      style: { width: `${COLUMN_WIDTHS[ACTIONS_COL]}px` },
     },
   ];
 
   const filteredVolumes = volumes.filter((item) => {
-    if (!filters.queries.every((q) => item.name.toLowerCase().includes(q))) {
+    // Do not show snapshot volumes at all
+    if (item.name.includes("/")) {
```
edlerd, line 126: Probably better to put this in a tiny helper function `isSnapshot(volume: LxdVolume)`, so the code gets speaking and doesn't need a comment here. Also good to reuse in other places.

src/util/storageVolumeTable.tsx
mas-who: I can change this to a css + media query approach? each column can get its own class name

mas-who: actually a percentage based approach should work too

src/pages/instances/actions/snapshots/useInstanceSnapshot.tsx
```diff
@@ -0,0 +1,253 @@
...
+          () => {
+            void queryClient.invalidateQueries({
+              predicate: (query) => query.queryKey[0] === queryKeys.instances,
+            });
+            onSuccess(
+              <>
+                Snapshot <ItemName item={values} bold /> created.
+              </>,
+            );
+            resetForm();
```
edlerd, line 88: The error handling does not surface the error message. I.e. when on a project that doesn't allow snapshot creation, the api responds with a message `Project \"default\" doesn't allow for snapshot creation`. We should show this message to the user. This is also relevant for the storage volume snapshot creation and edit snapshot on an instance or volume.

src/components/forms/VolumeSnapshotsForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useDocs } from "context/useDocs";
+import { FormikProps } from "formik";
+import { StorageVolumeFormValues } from "pages/storage/forms/StorageVolumeForm";
+
+interface Props {
+  formik: FormikProps<StorageVolumeFormValues>;
+  children?: ReactNode;
+}
+
+const VolumeSnapshotsForm: FC<Props> = ({ formik }) => {
```
edlerd, line 15: This seems like a non-generic component, better keep it in the storage directory?

src/components/forms/VolumeSnapshotsForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const docBaseLink = useDocs();
+  return (
+    <InstanceConfigurationTable
+      rows={[
+        getStorageConfigurationRow({
+          formik,
+          label: "Snapshot name pattern",
+          name: "snapshots_pattern",
```
edlerd, line 23: A custom pattern is not respected when creating new snapshots. This might be a backend bug. Can you investigate and communicate with the core team on this?

To reproduce: change the snapshot pattern to "snappy%d" on a volume. Create a snapshot, leave all fields default (especially empty name). The snapshot is created as "snap1".
Doing the same thing on an instance, results in a snapshot with the name "snappy1".

edlerd (commented):
This already works nicely on all the happy paths. I did a bit of QA and found some rough edges, see below comments for details.

src/pages/instances/actions/snapshots/useInstanceSnapshot.tsx
mas-who: I see, I didn't catch the edge case that if project is restricted after instance or volume snapshots were created, then the user can still create snapshots but this will fail. Will check why the error handling isn't showing the message here.

src/api/volume-snapshots.tsx
```diff
@@ -0,0 +1,159 @@
...
+  handleResponse,
+  pushFailure,
+  pushSuccess,
+} from "util/helpers";
+import { LxdOperationResponse } from "types/operation";
+import { LxdStorageVolume, LxdVolumeSnapshot } from "types/storage";
+import { LxdApiResponse, LxdSyncResponse } from "types/apiResponse";
+import { EventQueue } from "context/eventQueue";
+
+export const createSnapshot = (args: {
```
edlerd, line 12: I wonder if the function names should be unique. We are now introducing two functions with the same name `createSnapshot` (and similar for the others in this file below). Might be too easy to mix them up. Better keep them specific, like `createVolumeSnapshot`?

src/components/PortalModalBtn.tsx
```diff
@@ -0,0 +1,25 @@
...
+import usePortal from "react-useportal";
+
+interface Props {
+  renderModal: (closeModal: () => void) => ReactElement;
+  renderButton: (
+    openModal: (e: BaseSyntheticEvent<object, unknown, unknown>) => void,
+  ) => ReactElement;
+}
+
+const PortalModalBtn: FC<Props> = ({ renderButton, renderModal }) => {
```
edlerd, line 11: I see this is to reduce boilerplate code. I think it is trading readability for it. Because this component is coupling the renderModal and renderButton with its own local functions by passing them back to the caller, that is passing them on too. So the flow of dependencies is going in, then going out and then in again. I hope you know what I mean? Maybe better to keep the usePortal call in each specific button?

src/components/forms/SnapshotForm.tsx
```diff
@@ -1,59 +1,34 @@
...
-import { TOOLTIP_OVER_MODAL_ZINDEX } from "util/zIndex";
 import { SnapshotFormValues } from "util/snapshots";
 import { FormikProps } from "formik/dist/types";
-import NotificationRow from "components/NotificationRow";
 
 interface Props {
   isEdit: boolean;
-  formik: FormikProps<SnapshotFormValues>;
+  formik: FormikProps<
+    SnapshotFormValues<{ stateful?: boolean; description?: string }>
```
edlerd, line 18: I think we don't need the description here. Probably better to remove it.

src/pages/storage/StorageVolumeDetail.tsx
```diff
@@ -48,7 +50,9 @@ const StorageVolumeDetail: FC = () => {
     isLoading,
   } = useQuery({
     queryKey: [queryKeys.storage, pool, project, type, volumeName],
-    queryFn: () => fetchStorageVolume(pool, project, type, volumeName),
+    queryFn: () =>
+      loadCustomVolumeAndSnapshots({ pool, project, type, volumeName }),
```
edlerd, line 54: Why load the snapshots here as well? I think we only need the volume here and can load snapshots inside the `StorageVolumeSnapshots` component. I think they are only needed there.

src/context/loadCustomVolumeAndSnapshots.tsx
```diff
@@ -0,0 +1,28 @@
+import { fetchStorageVolume } from "api/storage-pools";
+import { fetchStorageVolumeSnapshots } from "api/volume-snapshots";
+import { LxdStorageVolume, LxdVolumeSnapshot } from "types/storage";
+import { splitVolumeSnapshotName } from "util/storageVolume";
+
+export const loadCustomVolumeAndSnapshots = async (args: {
```
edlerd, line 6: Not sure if we need this function, see below my comment on its usage.

src/pages/instances/InstanceDetail.tsx
```diff
@@ -47,6 +47,7 @@ const InstanceDetail: FC = () => {
   } = useQuery({
     queryKey: [queryKeys.instances, name, project],
     queryFn: () => fetchInstance(name, project),
+    refetchOnMount: (query) => query.state.isInvalidated,
```
edlerd, line 50: I think we shouldn't add this to all the queries. Why change it here?

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -10,34 +10,63 @@ import {
...
 export const figureCollapsedScreen = (): boolean =>
   window.innerWidth <= collapsedViewMaxWidth;
 
+interface SnapshotFormProps {
+  close: () => void;
+  instance: LxdInstance;
+  onSuccess: (message: ReactNode) => void;
+}
+
+const CreateInstanceSnapshotForm: FC<SnapshotFormProps> = (props) => {
```
edlerd, line 42: We should only have one component per file, where the filename matches the component name.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -54,7 +83,9 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
 
   const { project, isLoading } = useProject();
 
-  const snapshotsDisabled = project?.config["restricted.snapshots"] === "block";
+  const snapshotsDisabled =
+    project?.config["restricted.snapshots"] === "block" ||
+    project?.config["restricted"] === "true"; // Saw this on the demo server restricting snapshot creation;
```
edlerd, line 88: If the project is restricted, we have to check if restricted.snapshots is set to block or empty, because empty equals to block. We should not rely on the restricted setting alone, it doesn't say if snapshots are allowed or not.

src/util/snapshots.tsx
edlerd, line 1: Better split this file into a generic one, one about instances and another about volumes?

src/types/apiResponse.d.ts
```diff
@@ -1,3 +1,14 @@
...
   metadata: T;
 }
+
+export interface LxdSyncResponse {
+  type: "sync";
+  status: string;
+  status_code: number;
+  error_code: number;
+  error: string;
+  metadata: AnyObject;
```
edlerd, line 13: Surprised to see AnyObject from the yup package here. Maybe better use a generic type `LxdSyncResponse<T>` and `metadata: T`?

src/components/forms/VolumeSnapshotsForm.tsx
mas-who: An issue will be created for the core team to resolve this.

src/components/forms/SnapshotForm.tsx
mas-who: description is available for custom storage volume snapshots but not instance snapshot. Since this component is used for both creating instance and custom volume snapshots, the form may contain description field

src/pages/storage/StorageVolumeDetail.tsx
mas-who: This is my bad, I referenced the `InstanceDetail` logic and didn't think it through. The request to get snapshots is now moved to the `StorageVolumeSnapshots` component.

src/context/loadCustomVolumeAndSnapshots.tsx
mas-who: Yes, since the request for volume snapshots has been moved to its own component, this is no longer needed

src/components/forms/SnapshotForm.tsx
edlerd: I am not sure the description on the snapshot is useful. Maybe we should streamline and leave it out. Wdyt?

mas-who: Yes I agree, description is even less meaningful for scheduled snapshots. Maybe we can leave it out for now and add it if it turns out to be a super popular thing

src/pages/instances/InstanceSnapshots.tsx
mas-who: created a function for it in utils/snapshots.tsx

src/api/volume-snapshots.tsx
```diff
@@ -0,0 +1,159 @@
...
+            (msg) => pushFailure(results, msg),
+            () => continueOrFinish(results, snapshotNames.length, resolve),
+          );
+        });
+      }),
+    );
+  });
+};
+
+// NOTE: this api endpoint results in a synchronous operation
```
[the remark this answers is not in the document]
mas-who: as per discussion with the core team, will create issues for these

piperdeck:
Everything looks good to me, great work!

src/pages/instances/forms/CreateInstanceSnapshotForm.tsx
```diff
@@ -0,0 +1,33 @@
...
+import SnapshotForm from "components/forms/SnapshotForm";
+
+interface Props {
+  close: () => void;
+  instance: LxdInstance;
+  onSuccess: (message: ReactNode) => void;
+}
+
+const CreateInstanceSnapshotForm: FC<Props> = (props) => {
+  const { close, instance, onSuccess } = props;
```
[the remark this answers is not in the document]
mas-who: done as per previous comment

src/pages/instances/forms/EditInstanceSnapshotForm.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+interface Props {
+  instance: LxdInstance;
+  snapshot: LxdInstanceSnapshot;
+  close: () => void;
+  onSuccess: (message: ReactNode) => void;
+}
+
+const EditInstanceSnapshotForm: FC<Props> = (props) => {
+  const { instance, snapshot, close, onSuccess } = props;
```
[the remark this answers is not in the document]
mas-who: done as per previous comment

src/pages/instances/actions/snapshots/useInstanceSnapshot.tsx
```diff
@@ -0,0 +1,258 @@
...
+    </>
+  );
+  if (isStateful) {
+    statefulInfoMessage = `To create a stateful snapshot,\nthe instance must be running`;
+  }
+  if (isStateful && isRunning) {
+    statefulInfoMessage = "";
+  }
+
+  const statefulCheckbox = (
```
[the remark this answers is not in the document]
mas-who: I'm going to move the hook logic into the CreateInstanceSnapshotForm. Since the useCreateInstanceSnapshot hook is used exclusively in this component now. The downside is that the component contains more logic, but the code style would be more consistent with the rest of the codebase. Same thing will be done for the other relevant hooks and components. wdyt?

src/pages/instances/actions/snapshots/useInstanceSnapshot.tsx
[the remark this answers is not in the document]
mas-who: related to the previous comment. should probably remove these hooks

src/pages/instances/actions/snapshots/useInstanceSnapshot.tsx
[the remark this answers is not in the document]
edlerd: sounds good :+1:

src/pages/storage/StorageVolumeDetail.tsx
```diff
@@ -62,33 +64,37 @@ const StorageVolumeDetail: FC = () => {
...
-            <TabLinks
-              tabs={tabs}
-              activeTab={activeTab}
-              tabUrl={`/ui/project/${project}/storage/detail/${pool}/${type}/${volume.name}`}
-            />
+    <CustomLayout
+      header={<StorageVolumeHeader volume={volume} project={project} />}
+      contentClassName="sotrage-volume-form"
+    >
+      {activeTab !== "snapshots" && <NotificationRow />}
```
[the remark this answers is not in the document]
mas-who: The current behaviour with instance detail page is that snapshot related notifications appear below the tab links but everything else appear above. This is done in `InstanceSnapshots.tsx` with the `inTabNotification` state and the `NotificationRowLegacy` component. If we change this behaviour here then we would have to change it in all the other detail pages e.g. instance detail, network detail etc. Should we do that?

src/pages/storage/StorageVolumeNameLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface Props {
+  volume: LxdStorageVolume;
+  project: string;
+  isExternalLink?: boolean;
+  overrideName?: string;
+  className?: string;
+}
+
+const StorageVolumeNameLink: FC<Props> = (props) => {
```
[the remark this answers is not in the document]
mas-who: sorry not fully understanding what needs to be done here?

src/pages/storage/StorageVolumeDetail.tsx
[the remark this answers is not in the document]
edlerd: The idea for notification location was: Buttons above the tab list should produce notifications above them as well. In the instance detail page we have the state actions (start, stop, etc).

For buttons below the tab list, the notification should appear below that list. I think there are no actions for volumes above the tab list that trigger a notification. So it seemed, we can move the NotificationRow down and rely on it from all tab-conten-components. I think we don't have to change the other detail pages for this.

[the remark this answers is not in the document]
edlerd: Actually, I am not too happy about the in-tab notification implementation in instance detail. It couples the components very much for message passing. I hope it is something we'll improve very soon with the notification centre initiative.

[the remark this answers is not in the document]
mas-who: > The idea for notification location was: Buttons above the tab list should produce notifications above them as well. In the instance detail page we have the state actions (start, stop, etc).
> 
> For buttons below the tab list, the notification should appear below that list. I think there are no actions for volumes above the tab list that trigger a notification. So it seemed, we can move the NotificationRow down and rely on it from all tab-conten-components. I think we don't have to change the other detail pages for this.

I think one example is that if you go to the configurations tab, edit something for an instance or volume, the update notification appears above the tab list. This is because the `NotificationRow` component sits above `TabLinks` in `InstanceDetail`. This goes against the logic just described since the action originates from below the tab list. The problem now is that if we move `NotificationRow` below `TabLinks` in `InstanceDetail`, then notifications for instance actions above the tab list will also move below the tabs. Now looking at it, I think I understand why the snapshot notifications was done internal to the snapshots tab component.

src/api/volume-snapshots.tsx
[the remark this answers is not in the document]
mas-who: [issue link](https://github.com/canonical/lxd/issues/12637)

src/pages/storage/StorageVolumes.tsx
edlerd, line 1: This page has trouble with medium size, i.e. with 1000px width the content is hard to read.
I see two options:
1. Hide columns as we go below the medium breakpoint
2. Collapse columns -- i.e. like we do for the snapshots table, the created col wraps below the name with a muted text color. This gains some space for the other cols

I lean towards 2. Candidates to collapse are type & content type as well as name & created at. Though I am not 100% sure about it.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -216,23 +209,22 @@ const InstanceSnapshots: FC<Props> = ({ instance }) => {
...
+              <InstanceAddSnapshotBtn
+                instance={instance}
+                onSuccess={onSuccess}
+                onFailure={onFailure}
                 className="u-float-right"
-                onClick={() => setModalOpen(true)}
-              >
-                Create snapshot
-              </Button>
+              />
```
edlerd, line 223: I think those two buttons should also be disabled, when the project has restricted snapshot creation. Might be worth adding a tooltip or title text explaining why they are disabled. Or find a spot on the page to display the sentence, like in the empty state with disabled snapshots.
(same applies to the volume snapshots page)

src/pages/storage/StorageVolumeSnapshots.tsx
```diff
@@ -0,0 +1,340 @@
...
+                  aria-label="Search for snapshots"
+                />
+              </div>
+              <VolumeConfigureSnapshotBtn
+                volume={volume}
+                onSuccess={onSuccess}
+                onFailure={onFailure}
+                className="u-no-margin--right"
+              />
+              <VolumeAddSnapshotBtn volume={volume} className="u-float-right" />
```
edlerd, line 237: Those two buttons also need to be disabled when the project is restricting snapshot creation. Best with an explainer like on the instance snapshots (see other comment)

src/sass/_storage.scss
```diff
@@ -22,6 +40,34 @@
   button {
     margin-top: -6px;
   }
+
+  .volume-name-link {
+    max-width: 90%;
+  }
+
+  .u-row:hover {
+    background-color: $colors--light-theme--background-hover;
```
edlerd, line 49: I think we should only use hover effects, when the element is clickable.

src/sass/_storage.scss
```diff
@@ -22,6 +40,34 @@
...
+  }
+
+  .actions-list {
+    margin-top: -7px;
```
edlerd, line 53: I think this is not needed here, it looks better aligned without, no?

src/sass/styles.scss
```diff
@@ -259,3 +265,32 @@ $border-thin: 1px solid $color-mid-light !default;
 .no-border-top {
   border-top: none !important;
 }
+
+.snapshot-creation-modal {
```
edlerd, line 269: This seems not be generic enough for the main file, maybe into _snapshots.scss or a new file?

src/util/instanceSnapshots.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { AbortControllerState, checkDuplicateName } from "./helpers";
+import * as Yup from "yup";
+import { testFutureDate, testValidDate, testValidTime } from "./snapshots";
+
+/*** Instance snapshot utils ***/
+export const isInstanceStateful = (instance: LxdInstance) => {
+  return Boolean(instance.config["migration.stateful"]);
+};
+
+export const testDuplicateInsatnceSnapshotName = (
```
edlerd, line 11: Insatnce > Instance

src/util/storageVolume.tsx
```diff
@@ -86,16 +86,64 @@ export const getLxdDefault = (
...
+  return volumeDisplayName;
 };
 
 export const contentTypeForDisplay = (volume: LxdStorageVolume) => {
   return volume.content_type === "iso"
     ? "ISO"
     : capitalizeFirstLetter(volume.content_type);
 };
+
+export const generateLinkForVolumeDetail = (args: {
```
edlerd, line 102: shouldn't this be located in the StorageVolumeNameLink component? Though not sure what's best here.

edlerd (commented):
Thanks for all the changes, this is looking better and better.
Now, I finished the review with some new discoveries, when those are settled this should be ready to merge.

src/pages/storage/StorageVolumes.tsx
mas-who: Agreed, I think your suggestions of groupings makes sense, I will play around with grouping used by && snapshots as well

mas-who: I think it works nicely with used by && snapshots grouped for smaller screens 
Screenshot from 2023-12-08 14-37-49
wdyt?

src/pages/instances/InstanceSnapshots.tsx
mas-who: Yeah that makes sense, will add the help text and disable those buttons

src/pages/storage/StorageVolumes.tsx
edlerd: yep, looks good :+1: 

I remember one problem from the past: when sorting the combined columns, the arrow tends to be next to the lower line in the heading. You should probably make it consistent, so the expected sorting matches the actual one.

mas-who: Yeah I did notice that, made sure it's consistent across columns at small screen size

src/pages/instances/InstanceSnapshots.tsx
mas-who: I think a warning for configuring snapshot settings when snapshot is restricted makes sense. Perhaps this can be a separate task as it may require some design input. Cool if I create a Jira task for this?

edlerd: Yep, thank you :+1:

mas-who: [task link](https://warthogs.atlassian.net/browse/WD-7885)

src/sass/_storage.scss
mas-who: this may have been added in when I was battling with the styling and then forgotten... it does look better without it

src/sass/styles.scss
mas-who: moved

src/util/storageVolume.tsx
mas-who: there is a bit of logic in this function and it's just a preference to keep logic away from components whenever I can. I'm not too strongly opinionated about this though, if you feel it's more consistent with rest of the code to move logic into the component then that's also fine?

mas-who: just moved the function into the `StorageVolumeNameLink` component, still a function, that makes sense.

edlerd: We don't really have a tight rule around this. Some functions are in module specific files under util -- you probably have seen it already. When a function is component-specific and not likely to be reused, I tend to keep functions local with the component.

src/pages/storage/actions/snapshots/VolumeAddSnapshotBtn.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <Button
+          appearance="base"
+          hasIcon
+          dense={true}
+          onClick={openPortal}
+          type="button"
+          aria-label="Add Snapshot"
+          title={
+            isDisabled
+              ? `Snapshot is disabled for project ${volume.project}`
```
edlerd, line 39:
```suggestion
              ? `Snapshot creation is blocked for project ${volume.project}`
```

src/pages/storage/actions/snapshots/VolumeAddSnapshotBtn.tsx
```diff
@@ -0,0 +1,69 @@
...
+      ) : (
+        <Button
+          appearance="positive"
+          className={className}
+          onClick={openPortal}
+          disabled={isDisabled}
+        >
+          {isDisabled ? (
+            <Tooltip
+              message={`Snapshot is disabled for project ${volume.project}`}
```
edlerd, line 56:
```suggestion
              message={`Snapshot creation is blocked for project ${volume.project}`}
```

src/pages/instances/actions/snapshots/InstanceAddSnapshotBtn.tsx
```diff
@@ -0,0 +1,54 @@
...
+      )}
+      <Button
+        appearance="positive"
+        className={className}
+        onClick={openPortal}
+        disabled={isDisabled}
+      >
+        {isDisabled ? (
+          <Tooltip
+            message={`Snapshot is disabled for project ${instance.project}`}
```
edlerd, line 42:
```suggestion
            message={`Snapshot creation is blocked for project ${instance.project}`}
```

PR #571, opened 2023-12-07
Do not pass dynamicHeight as DOM attribute to textarea

lorumic:
## Done

- A tiny change to remove a React warning, complaining that an unrecognised "dynamicHeight" prop was being set on a DOM element:

Screenshot from 2023-12-07 20:38:06

## QA

1. Run the LXD-UI:
    - ~On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @lorumic or @edlerd for access.~ The warning seems to appear only when running locally (dev).
    - Open the browser console from the dev tools
    - Go to a page where there is an auto-expanding textarea, e.g. the instance detail page, configuration tab
    - Check that no warning like the one in the screenshot above appears in the console

### DIFF
--- a/src/components/AutoExpandingTextArea.tsx
+++ b/src/components/AutoExpandingTextArea.tsx
@@ -7,7 +7,8 @@ type Props = {
 } & TextareaProps;
 
 const AutoExpandingTextArea: FC<Props> = (props) => {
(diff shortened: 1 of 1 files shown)


mas-who:
Thank you for the fix!

PR #572, opened 2023-12-07
Add network forwards

edlerd:
## Done

- add forwards tab under network detail page
- added create network forward in the new tab
- added edit network forward in the new tab
- added Network name and subnet details to the create/edit forward screens
- removed html5 tooltips from required fields/edit forward screens
- changed help text for ports
- added description for forward

Fixes WD-7876

## QA

1. Run the LXD-UI:
    - create a network forward
    - edit a network forward with ports

### DIFF
--- a/src/pages/networks/NetworkForwardCount.tsx
+++ b/src/pages/networks/NetworkForwardCount.tsx
@@ -1,0 +1,32 @@
+import React, { FC } from "react";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { LxdNetwork } from "types/network";
+import { fetchNetworkForwards } from "api/network-forwards";
+
+interface Props {
+  network: LxdNetwork;
+  project: string;
+}
+
+const NetworkForwardCount: FC<Props> = ({ network, project }) => {
+  if (network.managed === false) {
+    return <>-</>;
+  }
+
+  const { data: forwards = [], isLoading } = useQuery({
+    queryKey: [
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network,
+      queryKeys.forwards,
+      project,
+    ],
+    queryFn: () => fetchNetworkForwards(network.name, project),
+  });
+
+  return <>{isLoading ? "" : forwards.length}</>;
+};
+
+export default NetworkForwardCount;
--- a/src/pages/networks/NetworkForwards.tsx
+++ b/src/pages/networks/NetworkForwards.tsx
@@ -1,0 +1,174 @@
+import React, { FC } from "react";
+import {
+  EmptyState,
+  Icon,
+  MainTable,
+  Row,
+  useNotify,
+} from "@canonical/react-components";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { LxdNetwork } from "types/network";
+import Loader from "components/Loader";
+import { fetchNetworkForwards } from "api/network-forwards";
+import { useDocs } from "context/useDocs";
+import DeleteNetworkForwardBtn from "pages/networks/actions/DeleteNetworkForwardBtn";
+import { Link } from "react-router-dom";
+import ExpandableList from "components/ExpandableList";
+import NetworkForwardPort from "pages/networks/NetworkForwardPort";
+
+interface Props {
+  network: LxdNetwork;
+  project: string;
+}
+
+const NetworkForwards: FC<Props> = ({ network, project }) => {
+  const docBaseLink = useDocs();
+  const notify = useNotify();
+
+  const {
+    data: forwards = [],
+    error,
+    isLoading,
+  } = useQuery({
+    queryKey: [
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network.name,
+      queryKeys.forwards,
+    ],
+    queryFn: () => fetchNetworkForwards(network.name, project),
+  });
+
+  if (error) {
+    notify.failure("Loading network forwards failed", error);
+  }
+
+  const hasNetworkForwards = forwards.length > 0;
+
+  const headers = [
+    { content: "Listen address", sortKey: "listenAddress" },
+    { content: "Description", sortKey: "description" },
+    { content: "Default target address", sortKey: "defaultTarget" },
+    { content: "Ports" },
+    { "aria-label": "Actions", className: "u-align--right actions" },
+  ];
+
+  const rows = forwards.map((forward) => {
+    return {
+      columns: [
+        {
+          content: forward.listen_address,
+          role: "cell",
+          "aria-label": "Listen address",
+        },
+        {
+          content: forward.description,
+          role: "cell",
+          "aria-label": "Description",
+        },
+        {
+          content: forward.config.target_address,
+          role: "cell",
+          "aria-label": "Default target address",
+        },
+        {
+          content: (
+            <ExpandableList
+              items={forward.ports.map((port) => (
+                <NetworkForwardPort key={port.listen_port} port={port} />
+              ))}
+            />
+          ),
+          role: "cell",
+          "aria-label": "Forwarded ports",
+        },
+        {
+          content: (
+            <>
+              <Link
+                className="p-button--base u-no-margin--bottom has-icon"
+                to={`/ui/project/${project}/networks/detail/${network.name}/forwards/${forward.listen_address}/edit`}
+              >
+                <Icon name="edit" />
+              </Link>
+              <DeleteNetworkForwardBtn
+                key={forward.listen_address}
+                network={network}
+                forward={forward}
+                project={project}
+              />
+            </>
+          ),
+          role: "rowheader",
+          className: "u-align--right actions",
+          "aria-label": "Actions",
+        },
+      ],
+      sortData: {
+        listenAddress: forward.listen_address,
+        description: forward.description,
+        defaultTarget: forward.config.target_address ?? "",
+      },
+    };
+  });
+
+  if (isLoading) {
+    return <Loader />;
+  }
+
+  return (
+    <>
+      <Link
+        className="p-button--positive u-no-margin--bottom u-float-right"
+        to={`/ui/project/${project}/networks/detail/${network.name}/forwards/create`}
+      >
+        Create forward
+      </Link>
+      <Row>
+        {hasNetworkForwards && (
+          <MainTable
+            headers={headers}
+            expanding
+            rows={rows}
+            paginate={30}
+            responsive
+            sortable
+            defaultSort="listenAddress"
+            defaultSortDirection="ascending"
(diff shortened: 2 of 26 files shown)


piperdeck:
# 1. 

My first thought is that on that page for the new network forward, there's nothing on this screen reminding you what of the subnet for the network that you're making this forward for. So if you didn't think to copy the network's IP address, you'll need to close this page and go back, then start over.

Here are my networks, with their subnet addresses showing.
image

Here's the details page for the network I want to forward to. It's IP address is still showing.
image

Now I go to the forwards tab....
image

And create a new forward, but it's not giving me any help by reminding me of the subnet address, which is most likely what I'll want here.
image

# 2.

Is this tooltip necessary? It appears as soon as I start editing the network forward
image

# 3.

I completely thought that this placeholder text was specifying the range of legal ports, and that we just have a weird restriction on which ports you can use. I would use support text like "eg. '80,90-100'" below the field rather than placeholder text here.

Incidentally, I noticed that the field won't validate that I've entered an illegal port number here. I suspect it won't break anything, but it's probably at least worth a yellow warning validation informing me that this won't work because ports only go up to 65353
image

# 4.

Is there a reason we don't have names or descriptions for forwards and ports? Both are supported on the back end

edlerd:

I addressed all mentioned points, please have a look at the updated result.

One point is still unresolved: The description for ports, I can't see how to fit them into the design. Maybe you have an idea, or maybe we don't need them in the UI?

edlerd:
> # Questions
> ## 1. Network forward with listen address that overlaps with another network's subnet?
> I created a network forward, setting the listen address to an IP that coincides with the subnet range of another network. I was expecting an error from the backend but was able to create the network forward successfully. According to the [docs](https://documentation.ubuntu.com/lxd/en/latest/howto/network_forwards/#network-forwards-listen-addresses) this is not allowed for a bridge network. Maybe my understanding is off?

That is okay, the forwards are firewall rules for the host, so any incoming traffic to a "listen address" is taken over by this rule, independent to which subnet or interface it belongs.

> ## 2. When creating ports for network forwards, duplicated listen ports are not allowed?
> I'm not a networking expert, thought maybe you could listen on one `ip:port` then forward to many internal `ip:port` addresses. But that's not true, I assume that's the correct behaviour?

Yes, you can only have one target for a given listen ip-port combination. What is pretty neat is, that for the same ip you can have different targets based on the port. So port 80 and 443 might go to one instance, while port 21 goes to another instance.

> ## 3. Invalid port numbers.
> This is related to the query from @piperdeck. When entering invalid port numbers we get a massive error output from the server. Maybe it would be a better user experience if we have some client side validation? Also noted that you can forward from different external ports to the same internal address ports.

True, when entering something like 500000 as a port, LXD responds with the raw error from the iptables config it tries to create. Though I would avoid validating high ports, as that becomes rather tricky with the supported formats for port ranges like `80-90` and lists like `80,443`.

piperdeck:
# 1

Can we please replace these with actual arrow symbols? Would look a little tidier

E.g.  :8000 ➝ 10.40.196.50:3000 (tcp)

image

# 2

It seems like this error notification has messed up the div sizes, leading to some weird scrolling behaviour. Also, the error is pretty unreadable. If we can't validate the input before the user presses the submit button, would it be possible to replace the error text with something more informative, along with highlighting the problematic field?

[Screencast from 2024-01-05 12-52-47.webm](https://github.com/canonical/lxd-ui/assets/45367207/776a206f-fe13-4bc9-92ff-905bd9d608d2)

edlerd:
> It seems like this error notification has messed up the div sizes, leading to some weird scrolling behaviour. Also, the error is pretty unreadable. If we can't validate the input before the user presses the submit button, would it be possible to replace the error text with something more informative, along with highlighting the problematic field?

We can reposition the error notification, so it is rendered in the correct container and doesn't mess up the page layout.

We can also validate the port numbers, so they are positive integers below 65535. It is a bit tricky to still allow ranges, and comma separated values, and a combination of the two. But that should be possible with regexp.

edlerd:
@piperdeck please have another pass at this.

Added the arrows, fixed the error notification placement and added port (-range and ,list) validation.

src/pages/networks/NetworkForwardCount.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { LxdNetwork } from "types/network";
+import { fetchNetworkForwards } from "api/network-forwards";
+
+interface Props {
+  network: LxdNetwork;
+  project: string;
+}
+
+const NetworkForwardCount: FC<Props> = ({ network, project }) => {
+  const { data: forwards = [] } = useQuery({
```
mas-who, line 13: maybe worthwhile to put a loader in this component to improve visuals instead of having numbers go from 0 -> count after some time?

src/pages/networks/NetworkForwards.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const {
+    data: forwards = [],
+    error,
+    isLoading,
+  } = useQuery({
+    queryKey: [queryKeys.networks, network, queryKeys.forwards, project],
+    queryFn: () => fetchNetworkForwards(network.name, project),
+  });
+
+  if (error) {
```
mas-who, line 38: maybe add Loader here?

src/util/queryKeys.tsx
```diff
@@ -2,6 +2,7 @@ export const queryKeys = {
   certificates: "certificates",
   cluster: "cluster",
   configOptions: "configOptions",
+  forwards: "forwards",
```
mas-who, line 5: I think `networkForwards` is a bit clearer?

src/pages/networks/actions/DeleteNetworkForwardBtn.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const notify = useNotify();
+  const queryClient = useQueryClient();
+  const [isLoading, setLoading] = useState(false);
+
+  const handleDelete = () => {
+    setLoading(true);
+    deleteNetworkForward(network, forward, project)
+      .then(() => {
+        notify.success(`Network forward for ${forward.listen_address} deleted`);
+        void queryClient.invalidateQueries({
```
mas-who, line 28: do we need to invalidate network queries? Maybe just invalidate network forward query for this specific network i.e. `query.queryKey[1] === network.name && query.queryKey[2] === queryKeys.forwards`. I think this might be more efficient when we get back to the `NetworkList` component so that we don't have to query the count for every network all over again.

src/pages/networks/NetworkForwards.tsx
```diff
@@ -0,0 +1,165 @@
...
+          role: "rowheader",
+          className: "u-align--right actions",
+          "aria-label": "Actions",
+        },
+      ],
+      sortData: {
+        listenAddress: forward.listen_address,
+        description: forward.description,
+        defaultTarget: forward.config.target_address ?? "",
+        ports: forward.ports.length,
```
mas-who, line 107: The ports column displays all ports as text, but we are sorting the column by count. Is this not slightly confusing for the users? WDYT?

src/pages/networks/NetworkForwardCount.tsx
edlerd: I think a loader would be too distracting, but we can keep it empty until the loading is finished.

src/util/queryKeys.tsx
edlerd: I was mimicking the api path, so this key is always used in combination with networks, i.e. `queryKey: [queryKeys.networks, network, queryKeys.forwards, project],`.

src/pages/networks/forms/NetworkForwardForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+const NetworkForwardForm: FC<Props> = ({
+  formik,
+  isEdit,
+  networkName,
+  project,
+}) => {
+  const notify = useNotify();
+
+  const { data: network } = useQuery({
```
mas-who, line 92: just a nit, so feel free to disregard. Instead of making a request, we could pass network data using React router. In `NetworkForwards` component we can use <Link to={{pathname: [path-to-create-or-edit-forms], state: {network}}}>. Then in `CreateNetworkForward` and `EditNetworkForward` components we can get the state via the `location` prop. Again, feel free to skip this one

src/pages/networks/forms/NetworkForwardForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+          />
+          <Input
+            {...formik.getFieldProps("description")}
+            id="description"
+            type="text"
+            label="Description"
+            placeholder="Enter description"
+            stacked
+          />
+          {formik.values.ports.length > 0 && (
```
mas-who, line 178: The add ports table makes this form component large and quite difficult to read from this point onwards. Can we create a `NetworkForwardPorts` table for this?

src/pages/networks/actions/DeleteNetworkForwardBtn.tsx
mas-who: or just use the exact query key for the network forward

mas-who:
QA looks good! Just a few minor comments with the code review.

src/pages/networks/NetworkForwardCount.tsx
mas-who: sounds good :+1:

src/util/queryKeys.tsx
mas-who: I see, thanks for clarifying!

src/pages/networks/forms/NetworkForwardForm.tsx
edlerd: Good point! This also reminds me we need to be consistent with queryKeys for networks, I'll refactor all the network related ones to start with `queryKeys.projects, project, queryKeys.networks`. We had some conflicts to that structure and also were introducing new conflicts in this PR. After this change, they should all align with the general structure (which always starts with `queryKeys.projects, project`. Which makes sense, because the entities live in a project as a root.

mas-who: Great, thanks for making that change! :pray:

src/pages/networks/actions/DeleteNetworkForwardBtn.tsx
edlerd: I'll limit this to the network.

edlerd: Limiting this to the exact forward will not invalidate the forwards list on the network.

src/pages/networks/NetworkForwards.tsx
edlerd: True, and there is no right solution sorting for multivalued fields. I'll remove the sort by port option.

mas-who:
LGTM. Thanks for making the changes

mas-who (approved):

piperdeck:
Awesome, looks good.

PR #573, opened 2023-12-08
feat(storage) default storage pool for volume creation [WD-7675]

mas-who:
## Done

- When a user creates a new custom storage volume from the volumes list, and if they’ve filtered this list to only show volumes from one storage pool, the volume creation form will now set the storage pool to match the filter by default.

## QA

1. Run the LXD-UI:
    - Go to Storage -> Volumes tab. In the search and filter input box, filter volumes by a **single** storage pool. Then try create a volume, the storage pool selection should be defaulted to the filtered storage pool.
    - Now try filter volumes by **multiple** storage pools. Then try create a volume, the storage pool selection should be defaulted to the default storage pool. Same behaviour should be observed if there are no storage pool filters for volumes.

### DIFF
--- a/src/pages/storage/StorageVolumes.tsx
+++ b/src/pages/storage/StorageVolumes.tsx
@@ -366,6 +366,9 @@ const StorageVolumes: FC = () => {
     return <Loader text="Loading storage volumes..." />;
   }
 
+  const defaultPoolForVolumeCreate =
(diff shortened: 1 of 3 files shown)


edlerd (approved):
QA and code looks good to me.

edlerd:
Might want to get a design review from @piperdeck  as he suggested the change originally.

piperdeck:
Everything looks good!

PR #574, opened 2023-12-08
fix(component) dynamic text area will now contain content exactly

mas-who:
## Done

Previously I was setting the text area height to be its scroll height + 1px. This was done so that the scrollbar doesn't show for the text area (if height of container === scroll height of container then the scrollbar shows), but I see this is causing the text area to grow faster than its content. Instead of doing that, we can simply just hide the scrollbar using css.

## QA

1. Run the LXD-UI:
    - Go anywhere that you can find a text area element, copy a body of text and paste into the text area. The text area should grow as tall as the content within.

## Screenshots
[Screencast from 08-12-2023 16:45:42.webm](https://github.com/canonical/lxd-ui/assets/151511568/217b8a35-031a-4bdc-8c00-fb237ba8a74a)


src/sass/styles.scss
```diff
@@ -259,3 +259,7 @@ $border-thin: 1px solid $color-mid-light !default;
 .no-border-top {
   border-top: none !important;
 }
+
+.text-area {
+  overflow: hidden;
```
lorumic, line 264: If we do this, it probably makes sense to also disable the `resize` - otherwise the user can still resize the textarea with the bottom-right handle, and that would result in a broken UI since the content is not scrollable.
```suggestion
  overflow: hidden;
  resize: none;
```

src/sass/styles.scss
```diff
@@ -259,3 +259,7 @@ $border-thin: 1px solid $color-mid-light !default;
 .no-border-top {
   border-top: none !important;
 }
+
+.text-area {
```
lorumic, line 263: Also, maybe this can be a bit more specific, to track down the context where it was added:

```suggestion
.auto-expanding-textarea {
```

lorumic: Sure, you could do that, or also keep it where it is - to avoid migrating all the other custom styles for the components that we might have in `styles.scss` at the moment.

By and large I think it's good to have it in `styles.scss` as it's just a custom style for a tiny custom component, and it would be an overkill to create a dedicated file for it - I think this is the reason why this file is now full of styles for different custom components. However, exactly due to the fact that we have a lot of classes in this file, we should try to be as specific as possible in terms of naming, so that just by the name one can grasp what is being done and why. :)

mas-who: okay perfect, will keep it as is with specific naming then, I can get behind that logic :+1: thanks!

src/components/AutoExpandingTextArea.tsx
```diff
@@ -26,6 +26,7 @@ const AutoExpandingTextArea: FC<Props> = (props) => {
       <Textarea
         {...(textAreaProps as TextareaProps)}
         rows={dynamicHeight ? undefined : rows}
+        className="text-area"
```
lorumic, line 29: I think you forgot to update it here :)
```suggestion
        className="auto-expanding-textarea"
```

mas-who: oops, I think I'm definitely a little tired :disappointed:

PR #576, opened 2023-12-11
chore(tests) add tests for server configs [WD-7703]

mas-who:
## Done

Added e2e tests for the server configuration / settings.

### DIFF
--- a/tests/helpers/server.ts
+++ b/tests/helpers/server.ts
@@ -1,0 +1,80 @@
+import { Locator, Page, expect } from "@playwright/test";
+import { TIMEOUT } from "./constants";
+
+export type ServerSettingType = "checkbox" | "text" | "number" | "password";
+
+export const visitServerSettings = async (page: Page) => {
+  await page.goto("/ui/");
+  await page.getByRole("link", { name: "Settings", exact: true }).click();
+};
+
+export const updateCheckbox = async (settingRow: Locator) => {
+  const checkbox = settingRow.locator(".p-checkbox__label");
+  await checkbox.click();
+};
+
+export const removePassword = async (page: Page, settingRow: Locator) => {
+  const removeButton = settingRow.getByRole("button", {
+    name: "Remove",
+    exact: true,
+  });
+  await removeButton.click();
+  await page.waitForSelector("text=Setting updated.", TIMEOUT);
+  await page.getByRole("button", { name: "Close notification" }).click();
+  await validateSettingValue(settingRow, "not set");
+};
+
+export const updateSetting = async (
+  page: Page,
+  settingName: string,
+  settingType: ServerSettingType,
+  content: string,
+) => {
+  const settingRow = page.locator("css=tr", { hasText: settingName });
+  await settingRow.getByRole("button").click();
+
+  if (settingType === "checkbox") {
+    await updateCheckbox(settingRow);
+  } else {
+    const settingInput = settingRow.locator(`css=input[type=${settingType}]`);
+    await settingInput.fill(content);
+  }
+  await settingRow.getByRole("button", { name: "Save", exact: true }).click();
+  await page.waitForSelector("text=Setting updated.", TIMEOUT);
+  await page.getByRole("button", { name: "Close notification" }).click();
+  await validateSettingValue(
+    settingRow,
+    settingType === "password" ? "set" : content,
+  );
+};
+
+export const validateSettingValue = async (
+  settingRow: Locator,
+  value: string,
+) => {
+  const readModeValue = settingRow.locator(".readmode-value");
+  await expect(readModeValue).toHaveText(value);
+};
+
+export const resetSetting = async (
+  page: Page,
+  settingName: string,
+  settingType: ServerSettingType,
+  defaultValue: string,
+) => {
+  const settingRow = page.locator("css=tr", { hasText: settingName });
+  await settingRow.getByRole("button").click();
+
+  if (settingType === "password") {
+    await removePassword(page, settingRow);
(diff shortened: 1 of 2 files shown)


tests/server.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    settingType: "text",
+    content: "Test User",
+  },
+];
+
+test("test all non-critical server settings", async ({ page }) => {
+  await visitServerSettings(page);
+  for (const setting of settings) {
+    // Update setting
+    await activateSetting(page, setting.settingName);
```
edlerd, line 47: The `activateSetting` call can be part of the `updateSetting`, making the test simpler to read. Then we don't need the comments any more, as the test code is straight forward.

tests/helpers/server.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (settingType === "checkbox") {
+    await updateCheckbox(settingRow);
+  } else {
+    const settingInput = settingRow.locator(`css=input[type=${settingType}]`);
+    if (content) {
+      await settingInput.fill(content);
+    }
+  }
+  await settingRow.getByRole("button", { name: "Save", exact: true }).click();
+  await page.waitForSelector("text=Setting updated.", TIMEOUT);
```
edlerd, line 47: I think we should close the notification, so the next check doesn't get marked as success by the notification from the previous setting that ran before. 

Maybe also validate the value if possible? I.e. true/false for checkbox, `value` for string and `set` for passwords?

tests/helpers/server.ts
```diff
@@ -0,0 +1,67 @@
...
+    await removeButton.click();
+    await page.waitForSelector("text=Setting updated.", TIMEOUT);
+    return;
+  }
+
+  await settingRow
+    .getByRole("button", { name: "Reset to default", exact: true })
+    .click();
+  await settingRow.getByRole("button", { name: "Save", exact: true }).click();
+  await page.waitForSelector("text=Setting updated.", TIMEOUT);
```
edlerd, line 66: This notification should also be closed.

Shall we add more validation logic that the setting, to ensure the value is actually a default now? We could ensure the value is printed without bold font (as that distinguishes the non-default values from default values). Or validate against a given value provided to the reset function?

tests/server.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  {
+    settingName: "loki.auth.password",
+    settingType: "password",
+    content: "abc123",
+  },
+  {
+    settingName: "user.ui_title",
+    settingType: "text",
+    content: "Test User",
+  },
```
edlerd, line 40: Thinking about this again -- maybe enough to keep just 1 setting per type: checkbox, text and password?

tests/helpers/server.ts
```diff
@@ -0,0 +1,67 @@
...
+      await settingInput.fill(content);
+    }
+  }
+  await settingRow.getByRole("button", { name: "Save", exact: true }).click();
+  await page.waitForSelector("text=Setting updated.", TIMEOUT);
+};
+
+export const resetSetting = async (page: Page, settingName: string) => {
+  const settingRow = page.locator("css=tr", { hasText: settingName });
+  if (await checkIfChangeButtonExist(settingRow)) {
```
edlerd, line 52: Maybe this branch for the password type should be its own function `removePassword`?

tests/server.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  updateSetting,
+  visitServerSettings,
+} from "./helpers/server";
+
+const settings: {
+  settingName: string;
+  settingType: ServerSettingType;
+  content?: string;
+}[] = [
+  { settingName: "acme.agree_tos", settingType: "checkbox" },
```
edlerd, line 15: Shall we drop the `setting` prefix from the properties? Or maybe with 3 settings left and much shorter test function, the array is not needed and simpler to write it out directly in the test?

tests/helpers/server.ts
mas-who: agreed, closing the notification for both update and reset. Also adding validations for both updates and resets.

tests/server.spec.ts
mas-who: I think let's keep the array, it doesn't hurt to keep this structure in case if we want to test more settings at a later stage then we can extend easily?

tests/helpers/server.ts
mas-who: yeah that's better, will do that

tests/server.spec.ts
mas-who: sure, we can do that :+1:

tests/helpers/server.ts
mas-who: this relates to the previous comment, I think validating against values is better since bold font is a design aspect that's more likely to change over time. I have added validations for resets.

edlerd (approved):
LGTM, thanks for making it nice.

When the test suite passes, this can be merged.

.github/workflows/canonical-web.code-workspace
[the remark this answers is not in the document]
mas-who: that's been removed, I accidentally saved my vs-code workspace settings in the repo earlier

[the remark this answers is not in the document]
edlerd: Maybe worth adding to `.gitignore`?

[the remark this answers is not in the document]
mas-who: no vs code workspace settings should not sit in any repo since it's a way for managing multiple repos. So I moved it elsewhere

PR #578, opened 2023-12-11
fix(app) disable configurations when entity creation is blocked [WD-7489]

mas-who:
## Done

- Fixed the issue with configuration side navigation items remaining accessible even when entity creation (i.e. instance, network, profile or storage volume) is blocked due to invalid form entries.

## QA

1. Run the LXD-UI:
    - While creating an instance, select a base image so that configurations becomes available. Click on the Advanced drop down, then deselect the base image. The drop down should collapse and become disabled.
    - While creating a profile, enter a name for the profile so that configurations becomes available. Click on the Advanced drop down, then clear the profile name input field. The drop down should collapse and become disabled.
    - While creating a network, enter a name for the network so that configurations becomes available. Click on the Advanced drop down, then clear the network name input field. The drop down should collapse and become disabled.
    - While creating a storage volume, enter a name for the volume so that configurations becomes available. Click on the Advanced drop down, then clear the volume name input field. The drop down should collapse and become disabled.

### DIFF
--- a/src/pages/instances/CreateInstance.tsx
+++ b/src/pages/instances/CreateInstance.tsx
@@ -347,7 +347,7 @@ const CreateInstance: FC = () => {
           active={section}
           setActive={updateSection}
           isConfigDisabled={!formik.values.image}
-          isConfigOpen={isConfigOpen}
+          isConfigOpen={!!formik.values.image && isConfigOpen}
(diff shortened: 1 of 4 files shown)


edlerd (approved):
LGTM, thanks for fixing this.

PR #579, opened 2023-12-12
chore: publish test reports to github pages [WD-7907]

mas-who:
## Done

- Created an orphan branch gh-pages from which github pages for this repository will deploy
- Added a new job in pr.yaml to use the artifact produced by the e2e test job and push it to the gh-pages branch. This takes into consideration that multiple pr actions may be running at the same time. Lastly, the job will run if the e2e tests failed or succeeded and output a url for the test reports.
- Added a new action triggered on PR merge so that test reprorts gets cleaned up on the gh-pages branch.

### DIFF
--- a/.github/workflows/delete_test_reports.yml
+++ b/.github/workflows/delete_test_reports.yml
@@ -1,0 +1,52 @@
+name: Delete playwright test reports
+on:
+    pull_request:
+        types:
+            - closed
+
+jobs:
+    delete_reports:
+        name: Delete Playwright Test Reports for PR ${{ github.event.number }}
+        runs-on: ubuntu-latest
+        env:
+            # Folder on gh-pages branch that contains all test reports for a PR
+            PR_REPORTS_DIR: reports/pr-${{ github.event.number }}
+        steps:
+            - uses: actions/checkout@main
+              with:
+                ref: gh-pages
+                token: ${{ secrets.GITHUB_TOKEN }}
+            - name: Set Git User
+              run: |
+                git config --global user.email "github-action@example.com"
+                git config --global user.name "GitHub Action"
+            - name: Check for workflow reports
+              id: has-reports
+              run: |
+                if [ -z "$(ls -A $PR_REPORTS_DIR)" ]; then
+                    echo "PR_REPORTS_EXISTS=false" >> $GITHUB_OUTPUT
+                else
+                    echo "PR_REPORTS_EXISTS=true" >> $GITHUB_OUTPUT
+                fi
+            - name: Delete reports from repo for PR ${{ github.event.number }}
+              if: ${{ steps.has-reports.outputs.PR_REPORTS_EXISTS == 'true' }}
+              timeout-minutes: 3
+              run: |
+                rm -rf reports/pr-${{ github.event.number }}
+                git add .
+                git commit -m "workflow: remove all reports for PR #${{ github.event.number }}"
+                
+                # In case if another action job pushed to gh-pages while we are rebasing for the current job
+                while true; do
+                    git pull --rebase
+                    if [ $? -ne 0 ]; then
+                        echo "Failed to rebase. Please review manually."
+                        exit 1
+                    fi
+        
+                    git push
+                    if [ $? -eq 0 ]; then
+                        echo "Successfully pushed HTML reports to repo."
+                        exit 0
+                    fi
+                done
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -2,7 +2,7 @@ name: PR checks
 on: pull_request
 
 permissions:
-  contents: read
+  contents: write
 
 jobs:
   commits:
@@ -66,6 +66,8 @@ jobs:
   e2e:
     name: e2e-tests
     runs-on: ubuntu-latest
+    outputs:
+      job_status: ${{job.status}}
     steps:
       - uses: actions/checkout@main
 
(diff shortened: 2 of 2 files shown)


.github/workflows/delete_test_reports.yml
```diff
@@ -0,0 +1,51 @@
...
+              if: ${{ steps.has-reports.outputs.BRANCH_REPORTS_EXIST == 'true' }}
+              timeout-minutes: 3
+              run: |
+                cd $BRANCH_REPORTS_DIR/..
+        
+                rm -rf ${{ github.event.ref }}
+                git add .
+                git commit -m "workflow: remove all reports for branch ${{ github.event.ref }}"
+        
+                while true; do
```
edlerd, line 39: Why the while loop?

.github/workflows/delete_test_reports.yml
```diff
@@ -0,0 +1,51 @@
...
+                else
+                    echo "BRANCH_REPORTS_EXIST=true" >> $GITHUB_OUTPUT
+                fi
+            - name: Delete reports from repo for branch
+              if: ${{ steps.has-reports.outputs.BRANCH_REPORTS_EXIST == 'true' }}
+              timeout-minutes: 3
+              run: |
+                cd $BRANCH_REPORTS_DIR/..
+        
+                rm -rf ${{ github.event.ref }}
```
edlerd, line 35:
```suggestion
                rm -rf reports/${{ github.event.ref }}
```
Simpler?

.github/useful_act_commands.md
edlerd, line 1: Why is this file needed?

.github/workflows/pr.yaml
```diff
@@ -155,3 +157,53 @@ jobs:
...
+          name: playwright-report
+          path: ${{ env.HTML_REPORT_URL_PATH }}
+      - name: Push HTML Report
+        timeout-minutes: 3
+        # In case if another action job pushed to gh-pages while we are rebasing for the current job
+        run: |
+          git add .
+          git commit -m "workflow: add HTML report for branch ${{ github.head_ref || github.ref_name }} (run_id: ${{ github.run_id }}, attempt:  ${{ github.run_attempt }})"
+
+          while true; do
```
edlerd, line 193: Why the loop?

.github/workflows/pr.yaml
```diff
@@ -155,3 +157,53 @@ jobs:
...
+
+  publish_report:
+    name: publish-e2e-report
+    # if job is cancelled we should skip this
+    if: success() || needs.e2e.outputs.job_status == 'failure'
+    needs: [e2e]
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    env:
+      HTML_REPORT_URL_PATH: reports/${{ github.head_ref || github.ref_name }}/${{ github.run_id }}/${{ github.run_attempt }}
```
edlerd, line 169: So we keep multiple runs on the same run available, nice :+1:

.github/workflows/delete_test_reports.yml
```diff
@@ -0,0 +1,51 @@
...
+                        echo "Failed to rebase. Please review manually."
+                        exit 1
+                    fi
+        
+                    git push
+                    if [ $? -eq 0 ]; then
+                        echo "Successfully pushed HTML reports to repo."
+                        exit 0
+                    fi
+                done
```
edlerd, line 51: Let's add a newline 
```suggestion
                done

```

edlerd (commented):
This is quite nice. Some questions and ideas below. Happy to discuss them over a call.

.github/workflows/pr.yaml
edlerd: Maybe use the PR number instead of the branch name? Then we can trigger the deletion on PR merge instead of branch removal. Just an idea, maybe simpler to keep as it is.

.github/useful_act_commands.md
mas-who: I wanted to keep a record of nektos commands for future reference, I will move this out of the repo.

.github/workflows/delete_test_reports.yml
mas-who: This is to cater for the scenario during a rebase, another pr action job pushed to the gh-pages branch. This means the push on the current job run will fail. If that's the case and the exit code is non-zero, then we will rebase again and push. Does that make sense or do you see some flaws in the logic?

.github/workflows/pr.yaml
mas-who: I don't see too much of a functional difference between tracking by PR number or branch name since this job is triggered with a pull request so it has to be an PR for a specific branch. The report delete action first checks if a folder with the branch name exists before deleting anything, so I don't think there is a risk there. Changing the implementation shouldn't be too bad but I do need to go through another round of testing which could take some time. If you think there is a definite benefit in doing this change then I can make it happen?

.github/workflows/pr.yaml
mas-who: same comment as before

.github/workflows/pr.yaml
edlerd: I just realized: The source branch can be in a fork, and then the current workflow would miss the "delete branch" event. So I think we have to track PRs and clean up when they are closed.

.github/workflows/delete_test_reports.yml
edlerd: Makes sense, maybe add the comment above the loop?

.github/workflows/pr.yaml
edlerd: Makes sense, maybe also add the comment in the line above the loop?

mas-who: sure done

.github/workflows/delete_test_reports.yml
mas-who: I don't think you can comment in a run command, maybe I just echo an explanation?

.github/workflows/pr.yaml
edlerd: On the cleanup job, you can use 
```
on:
  pull_request:
    types: [ closed ]
```

mas-who: the actions are now tracking by PR number.

Screenshot from 2023-12-13 17-09-53

.github/workflows/delete_test_reports.yml
edlerd: It should work if you start the line with a `#`, as that is a comment in a shell as well.

edlerd (approved):
LGTM, just one nit about the commenting of the while loops. Feel free to merge as is though.

mas-who:
> LGTM, just one nit about the commenting of the while loops. Feel free to merge as is though.

Aaah right, it's a shell script actually so it should be fine. I've moved the comment to within the run command on both actions now.

PR #580, opened 2023-12-12
Generalize config table helpers WD-7725

edlerd:
## Done

- generalized the config table helpers, and improved naming for all forms with inheritance behaviour

Fixes WD-7725

## QA

1. Run the LXD-UI:
    - check create and edit forms for
      - instances
      - profiles
      - projects
      - storage volumes
      - networks
    - ensure the test suite passes
    - no functional changes should be visible


src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+    case "network":
+      return getNetworkRowMetadata(name);
+  }
+};
+
+const getInstanceRowMetadata = (
+  values: InstanceAndProfileFormValues | ProjectFormValues,
+  name: string,
+): ConfigRowMetadata => {
+  const { data: configOptions } = useQuery({
```
mas-who, line 50: Although the queries are cached, still feels a bit weird that this request code is here. Wouldn't it make more sense to move the logic higher up before individual config rows are constructed and pass a configOptions lookup down to the child?

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+  const { project } = useParams<{ project: string }>();
+  const { data: profiles = [] } = useQuery({
+    queryKey: [queryKeys.profiles],
+    queryFn: () => fetchProfiles(project ?? ""),
+    enabled: Boolean(project),
+  });
+
+  // inherited values from applied profiles
+  if (Object.prototype.hasOwnProperty.call(values, "profiles")) {
+    const appliedProfiles = [
```
mas-who, line 68: suggestion: instead of reversing profiles, why not iterate profiles backwards to reduce number of array passes?

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+      ...(values as CreateInstanceFormValues | EditInstanceFormValues).profiles,
+    ].reverse();
+    for (const profileName of appliedProfiles) {
+      const profile = profiles.find((profile) => profile.name === profileName);
```
mas-who, line 72: nit: This is potentially an O^2 operation. We could iterate profiles before, construct a lookup of profile names? Although there might not be that many profiles so probably a non-issue

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+  return { value: lxdDefault, source: "LXD", configField };
+};
+
+export const getInheritedRootStorage = (
+  values: InstanceAndProfileFormValues,
+  profiles: LxdProfile[],
+): [LxdDiskDevice | null, string] => {
+  if (Object.prototype.hasOwnProperty.call(values, "profiles")) {
+    const appliedProfiles = [
```
mas-who, line 158: same comment as before

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+      ...(values as CreateInstanceFormValues | EditInstanceFormValues).profiles,
+    ].reverse();
+    for (const profileName of appliedProfiles) {
+      const profile = profiles.find((profile) => profile.name === profileName);
```
mas-who, line 162: same comment as before

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+export const getInheritedVolumes = (
+  values: InstanceAndProfileFormValues,
+  profiles: LxdProfile[],
+): InheritedVolume[] => {
+  const inheritedVolumes: InheritedVolume[] = [];
+  if (Object.prototype.hasOwnProperty.call(values, "profiles")) {
+    const appliedProfiles = [
+      ...(values as CreateInstanceFormValues | EditInstanceFormValues).profiles,
+    ].reverse();
```
mas-who, line 192: same comment as before

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+    for (const profileName of appliedProfiles) {
+      const profile = profiles.find((profile) => profile.name === profileName);
```
mas-who, line 194: same comment as before

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+export const getInheritedNetworks = (
+  values: InstanceAndProfileFormValues,
+  profiles: LxdProfile[],
+): InheritedNetwork[] => {
+  const inheritedNetworks: InheritedNetwork[] = [];
+  if (Object.prototype.hasOwnProperty.call(values, "profiles")) {
+    const appliedProfiles = [
+      ...(values as CreateInstanceFormValues | EditInstanceFormValues).profiles,
+    ].reverse();
```
mas-who, line 227: same comment as before

src/util/configInheritance.tsx
```diff
@@ -0,0 +1,246 @@
...
+    for (const profileName of appliedProfiles) {
+      const profile = profiles.find((profile) => profile.name === profileName);
```
mas-who, line 229: same comment as before

src/util/configInheritance.tsx
mas-who: I see quite a few functions needs to search through profiles, maybe having a function that generates a profileName lookup makes sense?

src/components/ConfigurationRow.tsx
```diff
@@ -34,13 +49,14 @@ export const getConfigurationRow = ({
...
-  isReadOnly,
-  value,
-  overrideValue,
+  readOnlyRenderer = (value) => <>{value}</>,
 }: Props): MainTableRow => {
+  const values = formik.values as unknown as Record<string, string | undefined>;
+  const value = values[name];
+  const isOverridden = value !== undefined;
+  const overrideValue = readOnlyRenderer(value === "" ? "-" : value);
+  const metadata = getConfigRowMetadata(formik.values, name);
```
mas-who, line 58: nice! If theese values were always fixed before then no need to pass them down as props

src/components/ConfigurationRow.tsx
```diff
@@ -123,16 +145,16 @@ export const getConfigurationRow = ({
...
+        {metadata && (
           <div className="p-text--small u-text--muted">
-            From: {inheritSource}
+            From: {metadata.source}
           </div>
         )}
       </div>
     ),
-    override: isReadOnly
+    override: formik.values.readOnly
```
mas-who, line 157: Unrelated to this PR, but the ternary below has the same indentation level which I find confusing since the last two conditions depends on the second condition (isOverriden). wdyt? Is that part of the general style guide?

src/pages/instances/forms/EditInstanceDetails.tsx
```diff
@@ -33,7 +33,7 @@ interface Props {
 }
 
 const EditInstanceDetails: FC<Props> = ({ formik, project }) => {
-  const isReadOnly = formik.values.readOnly;
+  const readOnly = formik.values.readOnly;
```
mas-who, line 36: how did this work before? InstanceEditDetailsFormValues interface only had readOnly, is it considering property from another unioned interface?

src/pages/instances/EditInstance.tsx
```diff
@@ -151,7 +151,7 @@ const EditInstance: FC<Props> = ({ instance }) => {
     return dumpYaml(bareInstance);
   };
 
-  const isReadOnly = formik.values.readOnly;
+  const readOnly = formik.values.readOnly;
```
mas-who, line 154: it's interesting, this file imports InstanceEditDetailsFormValues from EditInstanceDetails.tsx that file imports EditInstanceFormValues from this file, looks like a circular dependency? Maybe the typescript transpiler doesn't care since interfaces aren't a real construct in JS?

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -63,13 +63,14 @@ export interface NetworkFormValues {
   ipv6_ovn_ranges?: string;
   network?: string;
   yaml?: string;
+  type: "network";
```
mas-who, line 66: nit: the type attribute naming seems to be vague, maybe something like lxdEntityType might be better?

src/pages/networks/CreateNetwork.tsx
```diff
@@ -55,14 +55,22 @@ const CreateNetwork: FC = () => {
...
   });
 
   const formik = useFormik<NetworkFormValues>({
     initialValues: {
       readOnly: false,
       isCreating: true,
       name: "",
-      type: hasOvn ? "ovn" : "bridge",
+      networkType: hasOvn ? "ovn" : "bridge",
+      type: "network",
```
mas-who, line 73: refer to comment about field naming

src/pages/storage/forms/StorageVolumeForm.tsx
```diff
@@ -45,8 +45,9 @@ export interface StorageVolumeFormValues {
   zfs_remove_snapshots?: string;
   zfs_use_refquota?: string;
   zfs_reserve_space?: string;
-  isReadOnly: boolean;
+  readOnly: boolean;
   isCreating: boolean;
+  type: "storageVolume";
```
mas-who, line 50: nit: same comment about type being non-specific

src/pages/storage/forms/StorageVolumeCreate.tsx
```diff
@@ -43,13 +43,14 @@ const StorageVolumeCreate: FC = () => {
...
-      type: "custom",
+      volumeType: "custom",
       name: "",
       project: project,
       pool: searchParams.get(POOL) || "",
       size: "GiB",
-      isReadOnly: false,
+      readOnly: false,
       isCreating: true,
+      type: "storageVolume",
```
mas-who, line 53: refer to type comment from StorageVolumeForm.tsx

mas-who:
really nice refactor, makes so much more sense now that every config form is using the same config row component. I made some comments in `configInheritance` which I realise should have been part of the previous PR. They are just suggestions though so feel free to not implement them.

src/pages/networks/forms/NetworkForm.tsx
edlerd: entityType makes a lot of sense, I'll chagne it.

src/components/ConfigurationRow.tsx
edlerd: I'll change it to a function with early exits, which should be much easier to read.

src/util/configInheritance.tsx
edlerd: An instance won't have more than 1-5 profiles max. But still this code is quite redundant and can be optimized, as well as reused in all places to make the functions simpler.

src/pages/instances/EditInstance.tsx
edlerd: We can fix this to be one directional. `EditInstanceDetails.tsx` should only know about its own type `InstanceEditDetailsFormValues`, not about the union type from `EditInstance.tsx`.

src/pages/instances/forms/EditInstanceDetails.tsx
edlerd: Yes, the readOnly is in the `InstanceEditDetailsFormValues` type.

src/util/configInheritance.tsx
edlerd: I think it is fine to reverse it, as there will not be many applied profiles for an instance.

src/util/configInheritance.tsx
edlerd: Yes, I agree. Though currently, only instances, profiles and projects use the dynamic config lookup. Networks and storages don't need them. But both are planned to be added to it soon in the near future of this cycle. As soon as both are changed towards it, I am happy to move this query up the chain, as we'll need it in all cases anyway.

So for now let's keep the lookup here, in the future, let's keep in mind to move it up to the ConfigRow component.

src/util/configInheritance.tsx
edlerd: Settled on not altering the array, in favour of filtering and sorting the profiles array befor iterating over it. This is now in a dedicated function, making the overall code easier to gasp and quicker.

src/components/ConfigurationRow.tsx
edlerd: They needed adjustments to generalize. Before the values were different. But with some changes the generic nature became evident and indeed now the interface could be simplified and we can calculate the values here ... Further down.

src/pages/instances/EditInstance.tsx
mas-who: Maybe we can just move `InstanceEditDetailsFormValues` interface into `EditInstance.tsx`? The more specific interface `InstanceEditDetailsFormValues` is only ever used in creating the union type `EditInstanceFormValues` which lives in `EditInstance.tsx` where the `formik` is defined anyway.

edlerd: Yeah, that makes the types a bit better structured.

mas-who (approved):

PR #582, opened 2023-12-13
Rely on actual project list for selection of default project

### LINKED ISSUE #575: UI fails to login when the user isn't allowed access to the `default` project
In an environment where the user isn't allowed access to the `default` project, the UI just fails and goes back to the login page.

That's despite the UI having successfully retrieved `/1.0/projects` and therefore knowing the list of available projects and being able to detect that the `default` project isn't one of those.

I suspect you probably want to just pick the first project in the list if `default` isn't part of it.

image

edlerd:
## Done

- Rely on actual project list for selection of default project
- fix invalid object access for restricted users

Fixes #575

## QA

1. Run the LXD-UI:
    - try access to the UI with a restricted user (`lxc config trust list` then `lxc config trust edit $id` then change `restricted: true` and add a project which is or is not the default project)
    - The restricted user should be redirected to a project they have access to (i.e. non-default) when opening the `/ui` url in a browser.

### DIFF
--- a/src/context/auth.tsx
+++ b/src/context/auth.tsx
@@ -3,6 +3,7 @@ import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { fetchCertificates } from "api/certificates";
 import { useSettings } from "context/useSettings";
+import { fetchProjects } from "api/projects";
 
 interface ContextProps {
   isAuthenticated: boolean;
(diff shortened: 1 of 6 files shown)


src/context/auth.tsx
```diff
@@ -27,6 +28,17 @@ interface ProviderProps {
 export const AuthProvider: FC<ProviderProps> = ({ children }) => {
   const { data: settings, isLoading } = useSettings();
 
+  const { data: projects = [] } = useQuery({
+    queryKey: [queryKeys.projects, 1],
```
mas-who, line 32: nit: maybe assign 1 to a variable (e.g. recursion) to make it clear what the number means?

mas-who:
QA'd locally with the changes since I can't ssh into the demo server node yet. LGTM with just one small nit comment.

src/context/auth.tsx
edlerd: We always fetch with recursion set to 1 to hit the same cached result. So we can as well remove the parameter from the function. I did that together with a little refactoring of the other usages of the function to simplify.

mas-who: perfect, that makes sense :+1:

PR #591, opened 2024-01-03
fix: switch back to pull_request event trigger for gh actions

mas-who:
## Done

- Switch to use `pull_request` event trigger instead of `pull_request_target` for the pr check action. The `pull_request_target` event trigger was put in place so that we can get write permissions with the `GITHUB_TOKEN` for foreign (forked) PRs as actions would be run in the context of the base repository. However, it seems actions are actually running in context of the main branch as well which will disregard incoming changes.
- Will look for an alternative way to publish playwright reports.
- Also fixed some formatting issues that cropped up due to linting dependency updates.


edlerd (approved):
LGTM

PR #592, opened 2024-01-03
chore: pulish playwright reports using github action (v2)

mas-who:
## Done

- Publish playwright reports using github actions using `workflow_run` event trigger

### DIFF
--- a/.github/workflows/delete_test_reports.yml
+++ b/.github/workflows/delete_test_reports.yml
@@ -1,0 +1,53 @@
+name: Delete playwright test reports
+on:
+    pull_request_target:
+        types:
+            - closed
+
+jobs:
+    delete_reports:
+        name: Delete Playwright Test Reports for PR ${{ github.event.number }}
+        runs-on: ubuntu-latest
+        env:
+            # Folder on gh-pages branch that contains all test reports for a PR
+            PR_REPORTS_DIR: reports/pr-${{ github.event.number }}
+        steps:
+            - uses: actions/checkout@main
+              with:
+                ref: gh-pages
+                token: ${{ secrets.GITHUB_TOKEN }}
+            - name: Set Git User
+              run: |
+                git config --global user.email "github-action@example.com"
+                git config --global user.name "GitHub Action"
+            - name: Check for workflow reports
+              id: has-reports
+              run: |
+                if [ -z "$(ls -A $PR_REPORTS_DIR)" ]; then
+                    echo "PR_REPORTS_EXISTS=false" >> $GITHUB_OUTPUT
+                else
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_test_reports.yml
```diff
@@ -0,0 +1,82 @@
...
+    publish_report:
+        name: publish-e2e-report
+        runs-on: ubuntu-latest
+        continue-on-error: true
+        steps:
+          - uses: actions/checkout@v4
+            with:
+              ref: gh-pages
+              token: ${{ secrets.GITHUB_TOKEN }}
+          - name: Dlownload test report artifact
```
edlerd, line 17:
```suggestion
          - name: Download test report artifact
```

edlerd (approved):
LGTM

We could simplify further -- maybe no need to parse the pr number, run id and run attempt for the commit message. The commit message could be the same each time, or just contain the full path as it is. But also fine to keep as it is from my side.

mas-who:
> LGTM
> 
> We could simplify further -- maybe no need to parse the pr number, run id and run attempt for the commit message. The commit message could be the same each time, or just contain the full path as it is. But also fine to keep as it is from my side.

Oh my bad, I misunderstood you yesterday, thought you wanted to not store the additional information in the info.txt but still display them. Will just display the path in the commit message then.

.github/workflows/publish_test_reports.yml
```diff
@@ -0,0 +1,79 @@
...
+          # actual authentication is done using secrets.GITHUB_TOKEN with write permission
+          - name: Set Git User
+            run: |
+              git config --global user.email "github-action@example.com"
+              git config --global user.name "GitHub Action"
+          - name: Push HTML Report
+            timeout-minutes: 3
+            run: |
+              git add .
+              git commit -m "workflow: add HTML report publised at $HTML_REPORT_URL_PATH"
```
edlerd, line 60:
```suggestion
              git commit -m "workflow: add HTML report published at $HTML_REPORT_URL_PATH"
```

.github/workflows/delete_test_reports.yml
```diff
@@ -0,0 +1,52 @@
...
+                        echo "Failed to rebase. Please review manually."
+                        exit 1
+                    fi
+        
+                    git push
+                    if [ $? -eq 0 ]; then
+                        echo "Successfully pushed HTML reports to repo."
+                        exit 0
+                    fi
+                done
```
edlerd, line 52:
```suggestion
                done

```

edlerd (approved):
LGTM, thanks for the changes. Two minor typos below, then ready for merge :+1:

PR #594, opened 2024-01-05
feat: right align bottom buttons with forms across app [WD-7985]

mas-who:
## Done

- adjusted all creation and edit forms to have max width of 54rem
- right aligned bottom control buttons to all creation and edit form contents

## QA

1. Run the LXD-UI:
    - Check all forms with bottom control buttons. Make sure that buttons are right aligned to the form content for all screen sizes and also check if the forms visually look okay.

## Screenshots

Example
Screenshot from 2024-01-05 11-25-49

### DIFF
--- a/src/components/forms/FormFooterLayout.tsx
+++ b/src/components/forms/FormFooterLayout.tsx
@@ -1,0 +1,14 @@
+import React, { FC, PropsWithChildren } from "react";
+
+interface Props {}
+
+const FormFooterLayout: FC<Props & PropsWithChildren> = ({ children }) => {
+  return (
+    <div className="p-bottom-controls" id="form-footer">
+      <hr />
+      <footer className="u-align--right bottom-btns">{children}</footer>
+    </div>
+  );
+};
+
+export default FormFooterLayout;
--- a/src/components/forms/YamlForm.tsx
+++ b/src/components/forms/YamlForm.tsx
@@ -46,10 +46,7 @@ const YamlForm: FC<Props> = ({
   return (
     <>
       {children}
-      <div
-        ref={containerRef}
-        className="code-editor-wrapper restrict-editor-width"
-      >
+      <div ref={containerRef} className="code-editor-wrapper">
         <Editor
           defaultValue={yaml}
           language="yaml"
--- a/src/components/forms/YamlForm.tsx
+++ b/src/components/forms/YamlForm.tsx
@@ -46,7 +46,10 @@ const YamlForm: FC<Props> = ({
   return (
     <>
       {children}
-      <div ref={containerRef} className="code-editor-wrapper">
+      <div
+        ref={containerRef}
+        className="code-editor-wrapper restrict-editor-width"
+      >
         <Editor
           defaultValue={yaml}
           language="yaml"
--- a/src/pages/instances/CreateInstance.tsx
+++ b/src/pages/instances/CreateInstance.tsx
@@ -72,6 +72,7 @@ import {
   hasNetworkError,
   hasNoRootDisk,
 } from "util/instanceValidation";
+import FormFooterLayout from "components/forms/FormFooterLayout";
 
 export type CreateInstanceFormValues = InstanceDetailsFormValues &
   FormDeviceValues &
@@ -342,7 +343,7 @@ const CreateInstance: FC = () => {
 
   return (
     <BaseLayout title="Create an instance" contentClassName="create-instance">
-      <Form onSubmit={formik.handleSubmit} stacked className="form">
+      <Form onSubmit={formik.handleSubmit} className="form">
         <InstanceFormMenu
           active={section}
           setActive={updateSection}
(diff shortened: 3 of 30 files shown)


src/components/forms/YamlForm.tsx
```diff
@@ -46,7 +46,10 @@ const YamlForm: FC<Props> = ({
   return (
     <>
       {children}
-      <div ref={containerRef} className="code-editor-wrapper">
+      <div
+        ref={containerRef}
+        className="code-editor-wrapper restrict-editor-width"
```
edlerd, line 51: Why introduce a new class? We can use `code-editor-wrapper` to limit the editor width, no?

src/pages/storage/forms/StorageVolumeFormMain.tsx
```diff
@@ -37,6 +37,7 @@ const StorageVolumeFormMain: FC<Props> = ({ formik, project }) => {
             {...getFormProps(formik, "name")}
             type="text"
             label="Name"
+            stacked={false}
```
edlerd, line 40: I think stacked false is the default, no?
```suggestion
```

src/components/forms/YamlForm.tsx
mas-who: the `code-editor-wrapper` class is nested in the `cloud-init-config` class which isn't applicable for all forms. I will create a new selector with the same class but not nested instead.

src/pages/instances/CreateInstance.tsx
```diff
@@ -402,36 +402,34 @@ const CreateInstance: FC = () => {
...
-              onClick={() =>
-                navigate(
-                  location.state?.cancelLocation ??
-                    `/ui/project/${project}/instances`,
-                )
-              }
-            >
-              Cancel
-            </Button>
+        <footer className="u-align--right bottom-btns">
```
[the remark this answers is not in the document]
mas-who: Yeah that sounds like a great idea! Will check those additional classes

src/components/forms/FormFooterLayout.tsx
edlerd, line 1: Maybe shorten the name to `FormFooter`?

edlerd (commented):
Found two problems with narrow screens. And a third one, but that might be nasty and can be done as a follow up as well.

The profile selector for instances should not be full width. The up/down/remove buttons should be next to the profiles, see below comparison:

now:
Screenshot from 2024-01-05 12-32-22

before:
Screenshot from 2024-01-05 12-32-11


The secondary form navigation should be scrollable, when on a small screen. To make all the form screens accessible on small screens.

now:
Screenshot from 2024-01-05 12-39-11

before:
Screenshot from 2024-01-05 12-39-19

One last bit, but feel free to ignore it for a future ticket if out of scope. As this was inconsistent before. The scrollbar for the form content is sometimes on the very right of the page and sometimes in the middle at the content end. I think having the scroller in the middle is preferred. It is inconsistent mostly on the "main" form pages, and on the disk selector for instances.

scrollbar at content end:
Screenshot from 2024-01-05 12-45-14

scrollbar at the right
Screenshot from 2024-01-05 12-45-05

src/components/forms/FormFooterLayout.tsx
mas-who: - For the profile selector, I can revert to spacing from before. The buttons will align to the outer edge of the form so it still looks good. I also noticed that on really narrow screen size, even from before the arrows show above and below the selector. This is due to the arrows having fixed width regardless of the screen size. Maybe we should consider a different design for small screen sizes?
 - For the secondary form navigation, the current behaviour is caused by me removing the overflow css to keep the side navigation static. It's pretty tricky to get buttons aligned perfectly when the side navigation dynamically grows due to it being a scrollable container. I have adjusted css so that the buttons align mostly with the right edge of the form, but it's not perfect. Have a look on the demo server and let me know if it's okay? Alternatively, it might be possible to introduce a empty div right next to the footer container to mimic the behaviour of the side navigation so that the actual footer can be fixed width, but it would introduce non-semantic elements and might be tricky to get the div behave just like the side navigation.
 - For the inconsistent scrollbar issue, I think we should create a `ScrollableForm` component similar to the `ScrollableTable` component where it dynamically updates height attributes of the document body. I will create a ticket and circle back on this one later.
 - I think naming it `FormFooterLayout` indicates clearly that it is a wrapper instead of a self-sufficient component?

edlerd: > * For the profile selector, I can revert to spacing from before. The buttons will align to the outer edge of the form so it still looks good. I also noticed that on really narrow screen size, even from before the arrows show above and below the selector. This is due to the arrows having fixed width regardless of the screen size. Maybe we should consider a different design for small screen sizes?

I think this is fine now.
I can look into it in a follow-up to improve it for smaller screens.

> * For the secondary form navigation, the current behaviour is caused by me removing the overflow css to keep the side navigation static. It's pretty tricky to get buttons aligned perfectly when the side navigation dynamically grows due to it being a scrollable container. I have adjusted css so that the buttons align mostly with the right edge of the form, but it's not perfect. Have a look on the demo server and let me know if it's okay? Alternatively, it might be possible to introduce a empty div right next to the footer container to mimic the behaviour of the side navigation so that the actual footer can be fixed width, but it would introduce non-semantic elements and might be tricky to get the div behave just like the side navigation.

I don't fully understand. It seems to be working fine now, doesn't it? From your description it sounds like [scrollbar-gutter: stable](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) might solve the issue as it reserves space for the scrollbar even if it is not rendered.

> * For the inconsistent scrollbar issue, I think we should create a `ScrollableForm` component similar to the `ScrollableTable` component where it dynamically updates height attributes of the document body. I will create a ticket and circle back on this one later.

Yeah, that is a good idea, let's keep it as is for now then.

> * I think naming it `FormFooterLayout` indicates clearly that it is a wrapper instead of a self-sufficient component?

Sounds right.

edlerd (approved):
Code and QA looks good, thanks for improving this :+1:

PR #597, opened 2024-01-08
feature(storage): support ceph storage pool [WD-7968]

mas-who:
### Things required code changes
- CRUD storage pool with ceph (RBD) driver
- CRUD storage pool with ceph (RBD) driver in LXD cluster
- Prevent ceph storage volumes from being attached to more than one instance

### Things that didn't require code changes but tested locally
- Add, edit and delete custom storage volumes in ceph storage pool
- Add, edit, delete and restore snapshots for ceph custom storage volumes

## QA

1. Run the LXD-UI:
    - Setup a Ceph cluster using LXD VMs and microceph
    - Create, edit and delete storage pool with ceph driver
    - Create, edit and delete custom storage volume in ceph storage pool
    - Create, edit, delete and restore snapshots for ceph custom storage volumes
    - Create and run instances in ceph storage pool
    - Create and run instances with ceph custom storage volumes attached

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -14,6 +14,7 @@ import { LxdApiResponse } from "types/apiResponse";
 import { LxdOperationResponse } from "types/operation";
 import axios, { AxiosResponse } from "axios";
 import { LxdClusterMember } from "types/cluster";
+import { cephDriver } from "util/storageOptions";
 
 export const fetchStoragePool = (
   pool: string,
@@ -72,11 +73,23 @@ export const createPool = (
   });
 };
 
-const getPoolForCluster = (pool: Partial<LxdStoragePool>) => {
+const getClusterPool = (pool: Partial<LxdStoragePool>) => {
   const poolForCluster = { ...pool };
-  delete poolForCluster.config;
+  if (pool.driver !== cephDriver) {
+    delete poolForCluster.config;
+  }
 
   return poolForCluster;
+};
+
+const getMemberPool = (pool: Partial<LxdStoragePool>) => {
+  const poolForMember = { ...pool };
+  // config keys for ceph storage pool cannot be member specific
+  if (pool.driver === cephDriver) {
+    delete poolForMember.config;
+  }
+
+  return poolForMember;
 };
 
 export const createClusteredPool = (
@@ -84,14 +97,17 @@ export const createClusteredPool = (
   project: string,
   clusterMembers: LxdClusterMember[],
 ) => {
+  const memberPool = getMemberPool(pool);
   return new Promise((resolve, reject) => {
     void Promise.allSettled(
-      clusterMembers.map(async (item) =>
-        createPool(pool, project, item.server_name),
+      clusterMembers.map((item) =>
+        createPool(memberPool, project, item.server_name),
       ),
     )
       .then(handleSettledResult)
-      .then(() => createPool(getPoolForCluster(pool), project))
+      .then(() => {
+        return createPool(getClusterPool(pool), project);
+      })
       .then(resolve)
       .catch(reject);
   });
@@ -119,14 +135,15 @@ export const updateClusteredPool = (
   project: string,
   clusterMembers: LxdClusterMember[],
 ) => {
+  const memberPool = getMemberPool(pool);
   return new Promise((resolve, reject) => {
     void Promise.allSettled(
       clusterMembers.map(async (item) =>
-        updatePool(pool, project, item.server_name),
+        updatePool(memberPool, project, item.server_name),
       ),
     )
       .then(handleSettledResult)
-      .then(() => updatePool(getPoolForCluster(pool), project))
+      .then(() => updatePool(getClusterPool(pool), project))
       .then(resolve)
       .catch(reject);
   });
--- a/src/pages/storage/CreateStoragePool.tsx
+++ b/src/pages/storage/CreateStoragePool.tsx
@@ -8,12 +8,12 @@ import SubmitButton from "components/SubmitButton";
 import { useFormik } from "formik";
 import * as Yup from "yup";
 import { useNavigate, useParams } from "react-router-dom";
-import { LxdStoragePool } from "types/storage";
 import { queryKeys } from "util/queryKeys";
-import { zfsDriver, btrfsDriver } from "util/storageOptions";
+import { zfsDriver } from "util/storageOptions";
 import { testDuplicateStoragePoolName } from "util/storagePool";
 import StoragePoolForm, {
   StoragePoolFormValues,
+  storagePoolFormToPayload,
 } from "./forms/StoragePoolForm";
 import { useClusterMembers } from "context/useClusterMembers";
 import FormFooterLayout from "components/forms/FormFooterLayout";
@@ -45,20 +45,11 @@ const CreateStoragePool: FC = () => {
       driver: zfsDriver,
       source: "",
       size: "",
+      entityType: "storagePool",
     },
     validationSchema: CreateStoragePoolSchema,
-    onSubmit: ({ name, description, driver, source, size }) => {
-      const hasValidSize = size.match(/^\d/);
-      const storagePool: LxdStoragePool = {
-        name,
-        description,
-        driver,
-        source: driver !== btrfsDriver ? source : undefined,
-        config: {
-          size: hasValidSize ? size : undefined,
-        },
-      };
-
+    onSubmit: (values) => {
+      const storagePool = storagePoolFormToPayload(values);
       const mutation =
         clusterMembers.length > 0
           ? () => createClusteredPool(storagePool, project, clusterMembers)
--- a/src/pages/storage/EditStoragePool.tsx
+++ b/src/pages/storage/EditStoragePool.tsx
@@ -14,10 +14,12 @@ import { LxdStoragePool } from "types/storage";
 import { queryKeys } from "util/queryKeys";
 import StoragePoolForm, {
   StoragePoolFormValues,
+  storagePoolFormToPayload,
 } from "./forms/StoragePoolForm";
 import { checkDuplicateName } from "util/helpers";
 import { useClusterMembers } from "context/useClusterMembers";
 import FormFooterLayout from "components/forms/FormFooterLayout";
+import { getStoragePoolEditValues } from "util/storagePoolEdit";
 
 interface Props {
   pool: LxdStoragePool;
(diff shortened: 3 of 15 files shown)


src/pages/storage/forms/StorageVolumeFormMenu.tsx
```diff
@@ -76,14 +77,14 @@ const StorageVolumeFormMenu: FC<Props> = ({
                 disableReason={disableReason}
               />
               {contentType === "filesystem" &&
-                ["zfs", "lvm", "ceph"].includes(poolDriver) && (
+                [zfsDriver, lvmDriver, cephDriver].includes(poolDriver) && (
```
edlerd, line 80: Good idea to use the constants! Probably good to have this array in the `storageOptions.tsx` as well, some name like `export const driversWithFilesystemSupport = [...]`?

src/util/configInheritance.tsx
```diff
@@ -118,6 +121,11 @@ const getNetworkRowMetadata = (name: string): ConfigRowMetadata => {
   return getNetworkDefault(name);
 };
 
+// NOTE: this is only relevant for Ceph RBD storage pools at the moment
+const getStoragePoolRowMetadata = (name: string): ConfigRowMetadata => {
+  return getCephStoragePoolDefault(name);
```
edlerd, line 126: This is pretty neat for the follow-up, when the defaults come in via API, which is planned to be added soon.

src/util/storageOptions.tsx
```diff
@@ -20,12 +21,17 @@ export const storageDrivers = [
...
+    label: "Ceph",
+    value: cephDriver,
+  },
 ];
 
 const storageDriverToSourceHelp: Record<string, string> = {
   dir: "Optional, path to an existing directory",
   lvm: "Optional, path to an existing block device, loop file or LVM volume group",
   zfs: "Optional, path to an existing block device, loop file or ZFS dataset/pool",
+  ceph: "Optional, existing OSD storage pool to use",
```
edlerd, line 34: This is wired, because there is also the `OSD pool name` under config. Seems like both do the same thing.
Do we really need both, or can we hide the source input for the ceph driver?

edlerd (commented):
QA wise all works well. Some ideas to simplify and some questions below.

src/context/loadCustomVolumes.tsx
```diff
@@ -18,8 +19,11 @@ export const loadCustomVolumes = async (
       const contentTypes = ["filesystem", "block"];
       const isFilesystemOrBlock = contentTypes.includes(volume.content_type);
       const isCustom = volume.type === "custom";
+      // Ceph pool volumes cannot be shared by multiple instances
+      const isCephDriverAndUsed =
+        pool.driver === cephDriver && volume.used_by?.length;
```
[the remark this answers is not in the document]
mas-who: I put the logic here so that we don't have to iterate the returned result for filtering. do you think it's better to just do a filter in `CustomVolumeSelectModal`?

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -21,11 +21,11 @@ const StoragePoolSize: FC<Props> = ({ pool }) => {
   }
 
   const total = resources.space.total;
-  const used = resources.space.used;
+  const used = resources.space.used || 0;
```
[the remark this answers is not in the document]
mas-who: just did a double check, `used` can come back as undefined and total can come back as 0. This happens if the storage pool has an `unavailable` status, at least for the ceph storage pool. I have modified the `LxdStoragePoolResources` type to have and optional `used` attribute. The defaulting to 0 would be necessary still I think.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -21,11 +21,11 @@ const StoragePoolSize: FC<Props> = ({ pool }) => {
...
 
   return (
     <Meter
-      percentage={(100 / total) * used}
+      percentage={(100 / total) * used || 0}
```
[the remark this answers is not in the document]
mas-who: related to the previous comment, this would be needed because `total` can return as 0 resulting in a `NaN`.

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -14,16 +19,53 @@ export interface StoragePoolFormValues {
...
 }
 
 interface Props {
   formik: FormikProps<StoragePoolFormValues>;
+  section: string;
+  setSection: (section: string) => void;
 }
 
-const StoragePoolForm: FC<Props> = ({ formik }) => {
+export const storagePoolFormToPayload = (
```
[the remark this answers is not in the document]
mas-who: this is great! :+1:

src/pages/storage/forms/StoragePoolFormMenu.tsx
```diff
@@ -28,6 +38,11 @@ const StoragePoolFormMenu: FC<Props> = ({ active, setActive }) => {
       <nav aria-label="Storage pool form navigation">
         <ul className="p-side-navigation__list">
           <MenuItem label={MAIN_CONFIGURATION} {...menuItemProps} />
+          <MenuItem
```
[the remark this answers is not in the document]
mas-who: Hmmm this would go against existing behaviour on other forms where sections are disabled and never hidden. In any case, I think having the section visible but disabled indicates to the users that there are additional settings for ceph storage pools, which isn't bad?

src/pages/storage/forms/StoragePoolFormCeph.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <ConfigurationTable
+      rows={[
+        getConfigurationRow({
+          formik,
+          label: "Cluster name",
+          name: "ceph_cluster_name",
+          defaultValue: "",
+          help: "Name of the Ceph cluster in which to create new storage pools",
+          children: <Input type="text" />,
```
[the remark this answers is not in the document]
mas-who: done

src/pages/storage/forms/StorageVolumeFormMenu.tsx
mas-who: event better! :+1:

src/util/storageOptions.tsx
mas-who: now that you mentioned, it is weird since like you said it seems like they are doing the same thing. I recall Thomas mentioned that `source` should definitely be configurable, so let's rather remove the `osd_pool_name` config.

src/context/loadCustomVolumes.tsx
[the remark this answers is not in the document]
mas-who: on second thought, I think it's better to keep logic where it belongs, will move it to the component

src/api/storage-pools.tsx
```diff
@@ -74,24 +75,36 @@ export const createPool = (
...
   const poolForCluster = { ...pool };
-  delete poolForCluster.config;
+  if (pool.driver !== cephDriver) {
+    delete poolForCluster.config;
+  }
 
   return poolForCluster;
 };
 
+const getMemberPoolForCluster = (pool: Partial<LxdStoragePool>) => {
```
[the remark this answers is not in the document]
edlerd: Yeah, agreed. My point was solely about consistency among the two implementations.

src/context/loadCustomVolumes.tsx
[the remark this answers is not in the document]
edlerd: We could keep it here, but then the function name should announce what it does. From the current name, the filter is an unexpected side effect.

src/pages/storage/forms/StoragePoolFormMenu.tsx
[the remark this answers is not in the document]
edlerd: True, I'm fine keeping it.
We do indeed hide sections, i.e. on network creation/editing when you disable ipv4 or ipv6, the menu sections get removed. We sometimes disable them, because the users needs to first fill out something on the main form and then the other sections get unlocked.
I am not sure what is best option here, but seems like we should make things more consistent.

src/util/storageOptions.tsx
edlerd: This help text says "existing OSD storage pool to use", but the osd_pool_name is a custom name that will create the store. Maybe that is the difference and the reason both might be needed? Did you test the behaviour of the source field? will it also create a new storage pool if there is no pool with that name yet?

mas-who: Screenshot from 2024-01-10 18-43-10
Screenshot from 2024-01-10 18-44-13

src/pages/storage/forms/StoragePoolFormMenu.tsx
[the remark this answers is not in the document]
mas-who: Aaah I see, that makes sense then. Let's hide the Ceph section and disable it if storage name isn't specified

src/context/loadCustomVolumes.tsx
```diff
@@ -20,7 +20,7 @@ export const loadCustomVolumes = async (
       const isCustom = volume.type === "custom";
 
       if (isCustom && isFilesystemOrBlock) {
-        result.push({ ...volume, pool: pool.name });
+        result.push({ ...volume, pool: pool.name, driver: pool.driver });
```
edlerd, line 23: I think we don't need the driver here, do we?

src/pages/storage/CreateStoragePool.tsx
```diff
@@ -8,21 +8,23 @@ import SubmitButton from "components/SubmitButton";
...
 import { useClusterMembers } from "context/useClusterMembers";
 import FormFooterLayout from "components/forms/FormFooterLayout";
+import { MAIN_CONFIGURATION } from "./forms/StoragePoolFormMenu";
 
 const CreateStoragePool: FC = () => {
   const navigate = useNavigate();
   const notify = useNotify();
   const queryClient = useQueryClient();
   const { project } = useParams<{ project: string }>();
+  const [section, setSection] = useState(MAIN_CONFIGURATION);
```
edlerd, line 27: Probably better to move this state into `StoragePoolForm` as it is not needed here.

src/pages/storage/EditStoragePool.tsx
```diff
@@ -27,6 +30,7 @@ const EditStoragePool: FC<Props> = ({ pool }) => {
   const notify = useNotify();
   const queryClient = useQueryClient();
   const { project } = useParams<{ project: string }>();
+  const [section, setSection] = useState(MAIN_CONFIGURATION);
```
edlerd, line 33: Same as above:
Probably better to move this state into `StoragePoolForm` as it is not needed here.

src/context/loadCustomVolumes.tsx
mas-who: right this was put in place for filtering ceph volumes when attaching disk device to instances. It's no longer needed will remove.

src/util/storageOptions.tsx
mas-who: this is resolved, changed help text for `source` config and removed `ceph.osd.pool_name`

src/pages/storage/CreateStoragePool.tsx
mas-who: good point! Thanks

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -2,10 +2,14 @@ import React, { FC, useEffect, useState } from "react";
 import { Form, Input, Row, Col, useNotify } from "@canonical/react-components";
 import { FormikProps } from "formik";
 import StoragePoolFormMain from "./StoragePoolFormMain";
-import StoragePoolFormMenu from "./StoragePoolFormMenu";
+import StoragePoolFormMenu, { CEPH_CONFIGURATION } from "./StoragePoolFormMenu";
 import { MAIN_CONFIGURATION } from "./StorageVolumeFormMenu";
```
edlerd, line 6: This is the wrong import, it should be also taken from `StoragePoolFromMenu`. This is a previous mistake.

edlerd (approved):
LGTM, just one change of a wrong import from the previous code.

src/pages/storage/forms/StoragePoolForm.tsx
mas-who: good spot, changed

mas-who:
I don't think a design review is actually necessary for this one since everything we added was based on existing designs.

PR #598, opened 2024-01-10
feat: use TablePagination component from react-components for paginating tables [WD-8209]

mas-who:
## Done

- removed `Pagination.tsx`, `pagination.tsx` and `_pagination.scss`.
- The above component and styles have been upstreamed to react-components as `TablePagination`.
- Applied `TablePagination` component across the app where previously the `Pagination` component was used together with the `usePagination` hook.
- updated `ScrollableTable` to use the id of a child table element for updating its height in the DOM.

## QA

1. Run the LXD-UI:
    - check if pagination works as expected across the app. It's the same component, so checking in one place is sufficient. Suggest checking the instance list as it has the most complex pagination use case.


src/components/SelectableMainTable.tsx
```diff
@@ -120,13 +120,18 @@ const SelectableMainTable: FC<Props> = ({
...
-    row.key = row.name;
+    const key = row.name;
 
-    return row;
+    return {
+      ...row,
+      columns,
+      className,
+      key,
+    };
```
edlerd, line 134: Isn't this a performance issue, because we are copying all the data into a new object on each render cycle?

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    const visibleCount = pageData.length;
+    const filteredText = dataFiltered ? "filtered" : "";
+
+    if (isSmallScreen) {
+      return `${visibleCount} out of ${totalItems}`;
+    }
+
+    if (visibleCount === totalItems && visibleCount > 1) {
+      return `Showing all ${totalItems} ${filteredText} ${itemName}s`;
```
edlerd, line 112: There are too many spaces here when not filtered. One of the spaces should go in the `filteredtext` value.

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    return `Showing ${visibleCount} out of ${totalItems} ${filteredText} ${itemName}${
```
edlerd, line 115: same space issue

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+      totalItems !== 1 ? "s" : ""
+    }`;
+  };
+
+  const getPageLimitOptions = () => {
+    return generatePagingOptions(pageLimits || DEFAULT_PAGE_LIMITS);
```
edlerd, line 121: Why apply defaults here again, when already done above in the component where the props are received?

this seems better to be called directly instead of creating a callback, i.e.

```
const pageLimitOptions = generatePagingOptions(pageLimits);
```

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+  PropsWithChildren<Props> & HTMLAttributes<HTMLDivElement>
+> = ({
+  data,
+  containerClass,
+  itemName = "item",
+  displayTitle,
+  renderBelow,
+  dataPropOverride = "rows",
+  pageLimits = DEFAULT_PAGE_LIMITS,
+  dataFiltered = false,
```
edlerd, line 70: maybe better `isFiltered`?

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const getPageLimitOptions = () => {
+    return generatePagingOptions(pageLimits || DEFAULT_PAGE_LIMITS);
+  };
+
+  const totalPages = Math.ceil(data.length / pageSize);
+
+  return (
+    <>
+      {renderBelow && renderChildren(children, dataPropOverride, pageData)}
```
edlerd, line 128: Better have only one call to `renderChildren` than two and assign it to a `const` above the return block.

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+}
+
+const TablePagination: FC<
+  PropsWithChildren<Props> & HTMLAttributes<HTMLDivElement>
+> = ({
+  data,
+  containerClass,
+  itemName = "item",
+  displayTitle,
+  renderBelow,
```
edlerd, line 67: Maybe it should better be called `position` (like the `Tooltip` components prop in react-components) with two options "above" and "below" and a default of "above"?

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+  dataPropOverride = "rows",
```
edlerd, line 68: Maybe you can find a better name, though I don't have a good idea right away. Maybe keep as is.

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,193 @@
...
+  };
+
+  const totalPages = Math.ceil(data.length / pageSize);
+
+  return (
+    <>
+      {renderBelow && renderChildren(children, dataPropOverride, pageData)}
+      <div className={classnames("pagination", containerClass)} {...divProps}>
+        <div className="description" ref={descriptionRef}>
+          {getDisplayTitle()}
```
edlerd, line 131: description or displayTitle? Maybe settle for description?

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -243,47 +247,50 @@ const InstanceSnapshots = (props: Props) => {
...
+                    totalCount={instance.snapshots?.length ?? 0}
+                    itemName="snapshot"
+                    parentName="instance"
+                    selectedNames={selectedNames}
+                    setSelectedNames={setSelectedNames}
+                    filteredNames={filteredSnapshots.map((item) => item.name)}
+                  />
+                )
+              }
+              pageLimits={[1, 2, 3]}
```
edlerd, line 273: I guess this is here for testing only.

src/util/useSortTableData.tsx
```diff
@@ -0,0 +1,45 @@
+import { SortDirection } from "@canonical/react-components";
+import { useState } from "react";
+import { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+
+type Props = {
+  rows: MainTableRow[];
```
edlerd, line 6:
```suggestion
  rows: unknown[];
```

src/components/SelectableMainTable.tsx
mas-who: Yeah I agree it is less performant. I think then alternative approach here is to not append more columns to each row if it has the same lengh as the number of headers.

if (row.columns?.length !== headersWithCheckbox.length) {
    row.columns = [
    {
        content: (
        <CheckboxInput
            label={
            <div className="u-off-screen">Select {row.name ?? "row"}</div>
            }
            labelClassName="u-no-margin--bottom"
            checked={isRowSelected}
            onChange={toggleRow}
            disabled={isRowProcessing || !row.name}
        />
        ),
        role: "rowheader",
        className: "select",
    },
    ...(row.columns ?? []),
    ];
}

src/components/TablePagination.tsx
mas-who: done

src/components/TablePagination.tsx
mas-who: this will be removed based on ege's feedback

src/components/TablePagination.tsx
mas-who: we can think on this one a bit, can't think of a better name either

src/components/TablePagination.tsx
mas-who: filteredText no longer relevant, reverting to how it was before

src/components/TablePagination.tsx
mas-who: same as previous comment

src/components/TablePagination.tsx
mas-who: good spot, will remove

src/components/TablePagination.tsx
mas-who: yeah I agree that sounds better

src/pages/instances/InstanceSnapshots.tsx
mas-who: yes, this was the last component I tested, may have slipped through the craps when removing the prop

src/util/useSortTableData.tsx
mas-who: this hook does rely on the MainTableRow data format since we need the sortData field from rows objects

src/components/SelectableMainTable.tsx
mas-who: on second thought, I think destructuring each row wouldn't be that expensive. Most of the data sits within `row.colums`, when we destructure `row`, columns isn't destructured but passed by reference. The code would be safer that way, since mutating data by reference is a common source of bugs that is difficult to track down.

src/components/TablePagination.tsx
mas-who: maybe `propForDataOverride`?

edlerd: Seems too long and not concise enough. Maybe `dataForwardProp`?

mas-who: that's nice, thanks! :+1:

src/pages/storage/CustomIsoList.tsx
```diff
@@ -175,27 +175,22 @@ const CustomIsoList: FC<Props> = ({ project }) => {
...
-        />
+      <ScrollableTable dependencies={[images]} tableId="custom-iso-table">
+        <TablePagination
+          data={sortedRows}
+          id="pagination"
+          containerClass="u-no-margin--top"
+          itemName="custom ISO"
+        >
+          <MainTable
+            id="custom-id-table"
```
edlerd, line 186:
```suggestion
            id="custom-iso-table"
```

src/pages/storage/StorageVolumeSnapshots.tsx
```diff
@@ -254,45 +258,47 @@ const StorageVolumeSnapshots: FC<Props> = ({ volume }) => {
...
+                    totalCount={volumeSnapshots.length ?? 0}
+                    itemName="snapshot"
+                    parentName="volume"
+                    selectedNames={selectedNames}
+                    setSelectedNames={setSelectedNames}
+                    filteredNames={filteredSnapshots.map((item) => item.name)}
+                  />
+                )
+              }
+              pageLimits={[1, 2, 3]}
```
edlerd, line 281:
```suggestion
```

src/components/TablePagination.tsx
mas-who: yes that was forgotten, thanks for picking that up :+1:

mas-who:
All review changes done here in `TablePagination` component has been implemented in the relevant react-component PR.

edlerd (commented):
Thanks for the changes. One remaining question below.

src/components/TablePagination.tsx
```diff
@@ -0,0 +1,186 @@
...
+  position?: "below" | "above";
+  dataForwardProp?: string;
+  pageLimits?: number[];
+}
+
+const TablePagination: FC<
+  PropsWithChildren<Props> & HTMLAttributes<HTMLDivElement>
+> = ({
+  data,
+  containerClass,
```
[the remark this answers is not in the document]
mas-who: The thinking there was to make it clear that the styling would be applicable for the container element only, but this could be solved by writing good documentation, so will revert to `className` to avoid styling conflicts as you mentioned

edlerd (approved):
LGTM, good stuff :+1:

edlerd (approved):
LGTM, thanks for the iterations.

PR #602, opened 2024-01-15
fix: remove linuxcontainers images

mas-who:
## Done

- Removed linux containers images for selection when creating instances. This is due to the linuxcontainers image server no longer accepts lxd users.
- Fixed tests to use ubuntu jammy image instead of alpine

## QA
- Check only images from ubuntu or ubuntu minimal servers are available for selection when creating an instance


edlerd (approved):
LGTM, thanks for the fixes :+1:

mas-who:
> LGTM, thanks for the fixes 👍

Thank you for the quick review! :smiley:

PR #603, opened 2024-01-15
feat: implement scrollable form components [WD-8093]

mas-who:
## Done

- Implemented scrollable form component for all main configuration sections across the app (creation and edit)
- aligned scrollable behaviour for existing forms and tables across the application

## QA

1. Run the LXD-UI:
    - Check all main configuration sections, make sure that the form is scrollable but not the document body
    - Check instance creation and edit, network creation and edit, profile creation and edit, storage pool creation and edit, storage volume creation and edit, project creation and edit

## Screenshots

Example
Screenshot from 2024-01-15 18-31-27

### DIFF
--- a/src/components/ScrollableForm.tsx
+++ b/src/components/ScrollableForm.tsx
@@ -1,0 +1,21 @@
+import React, { FC, ReactNode } from "react";
+import ScrollableContainer from "./ScrollableContainer";
+import { useNotify } from "@canonical/react-components";
+
+interface Props {
+  children: ReactNode;
+}
+
+const ScrollableForm: FC<Props> = ({ children }) => {
+  const notify = useNotify();
+  return (
+    <ScrollableContainer
+      dependencies={[notify.notification]}
+      belowId="form-footer"
+    >
+      {children}
+    </ScrollableContainer>
+  );
+};
+
+export default ScrollableForm;
--- a/src/pages/instances/forms/EditInstanceDetails.tsx
+++ b/src/pages/instances/forms/EditInstanceDetails.tsx
@@ -7,6 +7,7 @@ import { useSettings } from "context/useSettings";
 import MigrateInstanceBtn from "pages/instances/actions/MigrateInstanceBtn";
 import { isClusteredServer } from "util/settings";
 import AutoExpandingTextArea from "components/AutoExpandingTextArea";
+import ScrollableForm from "components/ScrollableForm";
 
 export const instanceEditDetailPayload = (values: EditInstanceFormValues) => {
   return {
@@ -28,7 +29,7 @@ const EditInstanceDetails: FC<Props> = ({ formik, project }) => {
   const isClustered = isClusteredServer(settings);
 
   return (
-    <div className="details">
+    <ScrollableForm>
       <Row>
         <Col size={12}>
           <Input
@@ -91,7 +92,7 @@ const EditInstanceDetails: FC<Props> = ({ formik, project }) => {
         setSelected={(value) => void formik.setFieldValue("profiles", value)}
         readOnly={readOnly}
       />
-    </div>
+    </ScrollableForm>
   );
 };
 
--- a/src/sass/_forms.scss
+++ b/src/sass/_forms.scss
@@ -25,7 +25,7 @@
     display: flex;
 
     .form-navigation {
-      min-width: 13rem;
+      min-width: 14rem;
       overflow: auto;
 
       .p-side-navigation__link {
@@ -45,10 +45,10 @@
     }
 
     .form-contents {
-      overflow-y: auto;
+      overflow: auto hidden;
       padding-left: 2.5rem;
 
-      .details {
+      .content-details {
         max-width: 54rem;
       }
     }
@@ -330,7 +330,7 @@
 .storage-volume-form {
   .bottom-btns {
     margin-top: 1rem;
-    max-width: 71rem;
+    max-width: 72rem;
     padding-right: 1.5rem;
   }
 }
@@ -342,6 +342,6 @@
 .edit-storage-volume {
   .bottom-btns {
     margin-top: 1rem;
-    max-width: 69.5rem;
+    max-width: 70.5rem;
   }
 }
--- a/src/sass/_scrollable_container.scss
+++ b/src/sass/_scrollable_container.scss
@@ -1,0 +1,7 @@
+.scrollable-container {
+  .content-details {
+    display: block;
+    overflow: auto;
+    scrollbar-gutter: stable;
+  }
+}
--- a/src/sass/styles.scss
+++ b/src/sass/styles.scss
@@ -82,6 +82,7 @@ $border-thin: 1px solid $color-mid-light !default;
 @import "progress_bar";
 @import "project_select";
 @import "rename_header";
+@import "scrollable_container";
 @import "scrollable_table";
 @import "selectable_main_table";
 @import "settings_page";
--- a/src/components/ScrollableContainer.tsx
+++ b/src/components/ScrollableContainer.tsx
@@ -8,35 +8,34 @@ interface Props {
   belowId?: string;
 }
 
-const ScrollableTable: FC<Props> = ({
+const ScrollableContainer: FC<Props> = ({
   dependencies,
   children,
   belowId = "",
 }) => {
   const ref = useRef<HTMLDivElement>(null);
 
-  const updateTBodyHeight = () => {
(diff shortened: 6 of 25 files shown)


src/pages/instances/forms/EditInstanceDetails.tsx
```diff
@@ -26,72 +27,75 @@ const EditInstanceDetails: FC<Props> = ({ formik, project }) => {
...
-            onChange={formik.handleChange}
-            value={formik.values.description}
-            dynamicHeight
-            disabled={readOnly}
-          />
-        </Col>
-      </Row>
-      {isClustered && (
+    <ScrollableForm dependencies={[notify.notification]} belowId="form-footer">
+      <div className="details">
```
edlerd, line 34: I think the `<div className="details">` should move into `ScrollableForm`.

src/components/ScrollableForm.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const ScrollableForm: FC<Props> = ({
+  dependencies,
+  children,
+  belowId = "",
+}) => {
+  const ref = useRef<HTMLDivElement>(null);
+
+  const updateTBodyHeight = () => {
```
edlerd, line 18: updateFormHeight?

edlerd (commented):
The right scrolling behaviour is missing on
- instance/profile creation/edit > disk devices
- network>bridge, dns, ipv4 and ipv6
- storage volume > snapshots, filesystem
- storage pool > ceph

mas-who:
> The right scrolling behaviour is missing on
> 
> * instance/profile creation/edit > disk devices
> * network>bridge, dns, ipv4 and ipv6
> * storage volume > snapshots, filesystem
> * storage pool > ceph

Oh that is interesting. I assumed that all the config tables are already scrollable. I will do a scan through and make sure everything has consistent behaviour.

mas-who:
> > The right scrolling behaviour is missing on
> > 
> > * instance/profile creation/edit > disk devices
> > * network>bridge, dns, ipv4 and ipv6
> > * storage volume > snapshots, filesystem
> > * storage pool > ceph
> 
> Oh that is interesting. I assumed that all the config tables are already scrollable. I will do a scan through and make sure everything has consistent behaviour.

Picked up two more sections that was missing the same behaviour:
- storage volume > zfs
- network > network forward create/edit

src/components/ScrollableForm.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const style = `height: calc(100vh - ${offset}px); min-height: calc(100vh - ${offset}px)`;
+    form.setAttribute("style", style);
+  };
+
+  useEventListener("resize", updateFormHeight);
+  useEffect(updateFormHeight, [...dependencies, ref]);
+
+  return (
+    <div ref={ref} className="scrollable-form">
+      <div className="details">{children}</div>
```
edlerd, line 36: Sorry for the nitpick after asking to pull the details div into this component. The `details` is quite generic, better `form-details`? Or if we go with the generic `ScrollableComponent`, maybe find better suiting class names.

edlerd (commented):
Thanks for finding all the forms and aligning their layout. Some ideas to simplify and generalize below.

One thing I realized: the `min-width` of the `form-navigation` is too low, if we increase it to `14rem`, we avoid a vertical scrollbar if the screen is too small to fit the whole expanded menu on the screen. Maybe we can increase it here as well?

src/components/ScrollableForm.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const ref = useRef<HTMLDivElement>(null);
+
+  const updateFormHeight = () => {
+    const form = ref.current?.children[0];
+    if (!form) {
+      return;
+    }
+    const above = form.getBoundingClientRect().top + 1;
+    const below = getAbsoluteHeightBelow(belowId);
+    const parentsBottomSpacing = getParentsBottomSpacing(form as HTMLElement);
```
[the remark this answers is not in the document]
mas-who: Nice catch, thanks!

src/components/ScrollableForm.tsx
```diff
@@ -0,0 +1,41 @@
...
+import useEventListener from "@use-it/event-listener";
+import { getAbsoluteHeightBelow, getParentsBottomSpacing } from "util/helpers";
+
+interface Props {
+  children: ReactNode;
+  dependencies: DependencyList;
+  belowId?: string;
+}
+
+const ScrollableForm: FC<Props> = ({
```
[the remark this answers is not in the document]
mas-who: I prefer the generic component approach you mentioned, that's a really good idea and I have implemented that :+1:

src/components/ScrollableForm.tsx
mas-who: yeah, I renamed the class to `content-details`

src/sass/styles.scss
```diff
@@ -91,6 +90,8 @@ $border-thin: 1px solid $color-mid-light !default;
 @import "storage_pool_form";
 @import "storage_volume_form";
 @import "upper_controls_bar";
+@import "scrollable_container";
+@import "scrollable_table";
```
edlerd, line 94: Nit: those were sorted alphabetically

src/sass/_forms.scss
```diff
@@ -25,13 +25,13 @@
...
     .form-navigation {
-      min-width: 13rem;
+      min-width: 14rem;
       overflow: auto;
 
       .p-side-navigation__link {
         justify-content: flex-start;
         margin-bottom: 0;
-        min-width: 13rem;
+        min-width: 14rem;
```
edlerd, line 34: The link width shouldn't grow or we get again the vertical scroller:
image

src/sass/_scrollable_container.scss
```diff
@@ -0,0 +1,7 @@
+.scrollable-container {
+  .content-details {
+    display: block;
+    overflow: hidden auto;
```
edlerd, line 4:
```suggestion
    overflow: auto;
```
Otherwise on small screens part of a scrollable table (on the right) is impossible to reach. I think the other containers width need further adjustments so the vertical scroller is only visible on very small screens. With this change it is also displayed on large ones.

src/sass/_forms.scss
mas-who: Yeah I see, do you mean the horizontal scroller shouldn't show right? Adjusting the link min-width affects the horizontal spacing?

src/sass/_scrollable_container.scss
mas-who: Seems like the horizontal scroller starts to show when the screen width gets to ~400px. So users on mobile devices will most likely see this. This is an existing issue though, maybe we should consider vertically stack the form inputs with the side menu for smaller screen sizes? At the moment I think the horizontal scroller would show for either the side menu or the form

Screenshot from 2024-01-16 14-03-13

src/sass/_forms.scss
edlerd: We increased `form-navigation` to avoid the vertical scroller in the menu.

Here we increase its child `p-side-navigation__link`. This recreates the issue. Hence, we shouldn't change the `p-side-navigation__link` and everything should be fine.

src/sass/_scrollable_container.scss
edlerd: Check the scrollable table (i.e. resource limits tab in your screenshot) There the vertical scroller is much earlier, even on big screens where there is enough space with this change.
But we should have this change, so the config tables are usable on small screens.

edlerd: Edit: Just need some adjustments, to avoid the vertical scroller for large screens in the config tables.

src/sass/_forms.scss
mas-who: cool made the change

src/sass/_scrollable_container.scss
mas-who: Okay I finally see it, reminder to myself not to use the device view port in dev tools for testing these issues

mas-who: I have set the `overflow-y` behaviour to hidden on the outer container i.e. `forms-content` and only rely on the `scrollable-table` overflow behaviour for vertical scroll. For horizontal scroll, we should rely on the outer container `overflow-x` behaviour or else for smaller screen size we won't be able to scroll to the end of the table.

edlerd (approved):
LGTM, thanks for making all the forms consistent and fix them for smaller resolutions!

PR #607, opened 2024-01-17
feat: add floating notifications for applicable events [WD-8230]

mas-who:
## Done

- setup floating notification provider
- applied floating notifications to all relevant instance actions
- applied floating notifications to all relevant cluster actions
- applied floating notifications to all relevant profile actions
- applied floating notifications to all relevant network actions
- applied floating notifications to all relevant storage actions
- applied floating notifications to all relevant image actions
- applied floating notifications to all relevant operations actions
- applied floating notifications to all relevant project actions
- applied floating notifications to all relevant settings actions
- removed NotificationRowLegacy component
- removed NotificationRow from NetworkMap, not needed
- added filtering by severity for toast notifications
- added a status bar that shows the lxd and ui version, operations in progress and also a button that can expand the notifications list.

## QA

1. Run the LXD-UI:
    - QA relevant notifications using this [google sheet](https://docs.google.com/spreadsheets/d/18r_4jI2bZ12zFjJszrREkOy0b3wpvLyoGSTm_aup6e4/edit#gid=0)


edlerd (commented):
Functionality wise looks good already!

I think we need a way to remove toasts programmatically. I.e. when starting a new instance. Ideally we'd issue an info notification "creating your instance", which gets removed once the instance is created and about to be started. At that time issue a new info toast "Created instance {instanceLink}, now starting it". And finally, once the new instance is started, that 2nd info should be removed and a new notification be triggered with a success status "Created and started instance {instanceLink}.".
In the error case previous info should also be removed and a failure toast issued.
We didn't have this requirement previously, because the notifications were always replaced.

mas-who:
> Functionality wise looks good already!
> 
> I think we need a way to remove toasts programmatically. I.e. when starting a new instance. Ideally we'd issue an info notification "creating your instance", which gets removed once the instance is created and about to be started. At that time issue a new info toast "Created instance {instanceLink}, now starting it". And finally, once the new instance is started, that 2nd info should be removed and a new notification be triggered with a success status "Created and started instance {instanceLink}.". In the error case previous info should also be removed and a failure toast issued. We didn't have this requirement previously, because the notifications were always replaced.

sure, that can be done, 

> Functionality wise looks good already!
> 
> I think we need a way to remove toasts programmatically. I.e. when starting a new instance. Ideally we'd issue an info notification "creating your instance", which gets removed once the instance is created and about to be started. At that time issue a new info toast "Created instance {instanceLink}, now starting it". And finally, once the new instance is started, that 2nd info should be removed and a new notification be triggered with a success status "Created and started instance {instanceLink}.". In the error case previous info should also be removed and a failure toast issued. We didn't have this requirement previously, because the notifications were always replaced.

Yeah true, I've updated the implementation so that adding notifications will return the added notification with an id field, which can then be used to clear notifications programatically at a later stage

src/context/toastNotificationProvider.tsx
```diff
@@ -0,0 +1,150 @@
...
+  FC,
+  PropsWithChildren,
+  ReactNode,
+  createContext,
+  useContext,
+  useState,
+} from "react";
+
+export type ToastNotificationType = NotificationType & {
+  timestamp?: ReactNode;
```
edlerd, line 23: shouldn't this be string?
```suggestion
  timestamp?: string;
```

edlerd (commented):
This works and looks pretty nice already.

As discussed, the notification width can be a bit bigger on wider screens. Maybe use the medium, large and xlarge breakpoints from vanilla for proper spacing?

src/context/toastNotificationProvider.tsx
mas-who: yes, although the `Notification` component takes in a `ReactNode`, so I decided to just follow through with the same prop type

edlerd: I think it is fine to be more specific in a client component like this one. And string is more specific than ReactNode I guess. But then again, not too important anyhow.

mas-who: cool, will change it to string :+1:

piperdeck:
This is super cosmetic but I thought I would mention it now in case I forget later:

because objects with 1 pixel wide borders are appearing next to each other, there end up being 2 pixel wide borders separating the notifications. This doesn't need to change now while we're still in such early stages, but it would be nice if the finished product had only 1 pixel wide borders throughout.

image

Could we try putting a drop shadow on the notification and notification cluster components? I think it could help with visually distinguishing their borders with the borders of the page content they are floating in front of

Frame 613

Here's an example of a component in Vanilla that's using a similar drop shadow: https://vanillaframework.io/docs/layouts/application#example-structure

edlerd:
We should increase the z-index for the notifications, so they are not hidden behind the side navigation:

image

mas-who:
> I think the show all should open the full list, and scroll it down. Currently the most recent message is hidden below other things when expanding:

I tried scrolling but it doesn't feel very natural. So instead I reverse the array of notifications when expanding the container so that notifications are ordered top to bottom by time. Maybe try it out and see if you'd still prefer scrolling to the bottom?

src/util/instanceBulkActions.tsx
```diff
@@ -84,3 +88,22 @@ export const statusLabel = (status: LxdInstanceStatus) => {
   };
   return statusToLabel[status];
 };
+
+export const generateInstanceLink = (args: {
```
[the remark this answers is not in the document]
mas-who: I agree, couldn't find a decent place to put it so I dropped it in this file temporarily. Maybe I will just create a new file `instances.tsx` for it.

src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,190 @@
...
+            className={classnames(
+              "u-no-margin u-no-padding u-no-border filter-button",
+              { "filter-button-active": filters[severity] },
+            )}
+            onClick={() => handleFilterSelect(severity)}
+            disabled={!expanded}
+          >
+            <Icon name={iconLookup[severity]} />
+            <span>{groupedCount[severity]}</span>
+          </button>
```
edlerd, line 86: This is not a11y conform, probably should add a title to the button component, or an aria-label.

Also I am confused why the filters are shown, but in a disabled state, when the toast notifications are collapsed. Would it be better if they expand and apply filter in that case?

src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+    onDismissAll();
+  };
+
+  const getSeverityFilters = () => {
+    const filterButtons = severityOrder.map((severity) => {
+      if (groupedCount[severity]) {
+        return (
+          <button
+            key={`notification-filter-${severity}`}
```
edlerd, line 76:
```suggestion
            key={severity}
```
Can keep the keys simpler.

src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    onDismissAll();
+  };
+
+  const getSeverityFilters = () => {
+    const filterButtons = severityOrder.map((severity) => {
+      if (groupedCount[severity]) {
+        return (
+          <button
```
edlerd, line 75: The height of those buttons is too low to comply with a11y standards. Same for the "Dismiss all" and "Hide" clickable areas.

src/components/ToastNotifications.tsx
mas-who: made the button a toggle button with `aria-label` and `aria-pressed`. The disabled state issue will be resolved with the updated design.

src/components/ToastNotifications.tsx
mas-who: The updated design will resolve this issue.

src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,193 @@
...
+      const notification = array[lastNotificationIndex - index];
+      return (
+        <Notification
+          key={`notification-${index}`}
+          title={notification.title ?? DefaultTitles[notification.type]}
+          actions={notification.actions}
+          severity={notification.type}
+          onDismiss={() => onDismiss(notification)}
+          className={`u-no-margin--bottom toast-notification ${index === lastNotificationIndex ? "toast-notification--no-border-bottom" : ""}`}
+          timestamp={notification.timestamp}
```
edlerd, line 141: We probably should set `titleElement` to a non heading one like `div`, because the default of `h5` will not comply to the document heading structure.

src/pages/instances/InstanceDetail.tsx
```diff
@@ -64,9 +58,15 @@ const InstanceDetail: FC = () => {
       }
       contentClassName="detail-page"
     >
-      <NotificationRow />
       {isLoading && <Loader text="Loading instance details..." />}
-      {!isLoading && !instance && <>Loading instance failed</>}
+      {!isLoading && !instance && !error && <>Loading instance failed</>}
+      {error && (
+        <Strip>
+          <Notification severity="negative" title="Error">
```
edlerd, line 65: `titleElement` of the `Notification` should match the heading structure of the page. I think h3 is the right one here.

src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,193 @@
...
+    })
+  ) : (
+    <Notification
+      title={latestNotification.title ?? DefaultTitles[latestNotification.type]}
+      actions={latestNotification.actions}
+      severity={latestNotification.type}
+      onDismiss={() => onDismiss(latestNotification)}
+      className="u-no-margin--bottom toast-notification"
+      timestamp={latestNotification.timestamp}
+      role="alert"
```
edlerd, line 155: titleElement similar to above.

src/components/ToastNotifications.tsx
mas-who: yeah in this case I think `div` is appropriate since toast notifications exists outside of the app root.

src/pages/instances/InstanceDetail.tsx
mas-who: true, `h3` seems appropriate, although it is slightly strange that `h3` is appearing above the `h2` elements on this page. I think this can be included in the scope of the a11y audit issue rather since there are many places in the app where `Notification` is shown

mas-who:
Implementation based on the latest designs is done. Still need to do animations and tests.

src/util/layout.tsx
```diff
@@ -0,0 +1,65 @@
...
+  getHeightOffsetStyle,
+  getNegativeMargin,
+} from "./helpers";
+import useEventListener from "@use-it/event-listener";
+
+type adjustAppLayoutHeightArgs = {
+  dependencies?: DependencyList;
+};
+
+export const useAdjustAppLayoutHeight = ({
```
edlerd, line 13: This workaround shouldn't be needed. When I comment out the body of this it still aligns nicely on all pages. Only pages with forms have trouble, but that is because the `updateMaxHeight` function the forms use to calculate their height, doesn't consider the new status bar. I see you added some adjustments for the status bar to the `ScrollableTable` component, which might be needed for the `updateMaxHeight` as well.
After that we should be able to remove this workaround.

src/components/StatusBar.tsx
```diff
@@ -0,0 +1,53 @@
...
+  className?: string;
+}
+
+const StatusBar: FC<Props> = ({ className }) => {
+  const { toggleListView, notifications, countBySeverity, isListView } =
+    useToastNotification();
+
+  const notificationIcons = severityOrder.map((severity) => {
+    if (countBySeverity[severity]) {
+      return <Icon key={`status-${severity}`} name={iconLookup[severity]} />;
```
edlerd, line 19: This should have an aria label.

src/context/toastNotificationProvider.tsx
```diff
@@ -0,0 +1,252 @@
...
+      <ToastNotification
+        notification={latestNotification}
+        onDismiss={clear}
+        show={!!showNotification}
+      />
+      <ToastNotificationList
+        notifications={notifications}
+        onDismiss={clear}
+        onDismissAll={clearAll}
+        onDismissByFilter={clearByFilter}
```
edlerd, line 242: I think we should simplify the interface to a single callback that receives an array of notifications to dismiss. Then we can avoid the three distinct implementations of `clear`, `clearAll` and `clearAllByFilter`. I also think the filters should be a concept of the `ToastNotificationList`, that doesn't leak into the provider.

src/context/toastNotificationProvider.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const [notifications, setNotifications] = useState<ToastNotificationType[]>(
+    [],
+  );
+  const [countBySeverity, setCountBySeverity] =
+    useState<GroupedNotificationCount>({
+      caution: 0,
+      information: 0,
+      negative: 0,
+      positive: 0,
+    });
```
edlerd, line 82: We shouldn't keep the counts and the notifications in state. The counts should be calculated on the fly from the notifications. The problem with keeping both in state is, that the counts and the actual representation in the notifications mismatch. Also it complicates any state updates, because we need to update the notifications and the counts, instead of updating the notifications and relying on the render cycle to calculate the new counts.

src/util/layout.tsx
mas-who: thank you for suggesting this, I've made necessary changes in updateMaxHeight and it now works on form based pages :+1:

src/context/notificationProvider.tsx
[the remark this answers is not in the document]
mas-who: Yes I agree, will change that

src/pages/images/actions/DeleteImageBtn.tsx
```diff
@@ -33,9 +30,14 @@ const DeleteImageBtn: FC<Props> = ({ image, project }) => {
...
-          notify.success(`Image ${image.properties.description} deleted.`);
+          toastNotify.success(`Image ${image.properties.description} deleted.`);
         },
-        (msg) => notify.failure("Image deletion failed", new Error(msg)),
+        (msg) =>
+          toastNotify.failure(
+            "Error",
+            new Error(msg),
+            `Image ${image.properties.description} deletion failed`,
+          ),
```
edlerd, line 40: Why do we sometimes change the title to the message and user "Error" as title?
I think how it was previously was better. The title can already contain the details and we should avoid the generic "Error" title. Wdyt?

src/pages/instances/EditInstance.tsx
```diff
@@ -112,13 +113,20 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
+            toastNotify.success(<>Instance {instanceLink} updated.</>);
             void formik.setValues(getInstanceEditValues(instancePayload));
           },
-          (msg) => notify.failure("Instance update failed", new Error(msg)),
+          (msg) =>
+            toastNotify.failure(
+              "Error",
+              new Error(msg),
+              <>Instance {instanceLink} update failed.</>,
+            ),
```
edlerd, line 129: Error title?

src/pages/instances/Events.tsx
```diff
@@ -47,8 +51,9 @@ const Events: FC = () => {
         void queryClient.invalidateQueries({
           queryKey: [queryKeys.operations, event.project],
         });
+        void refetchOperations();
       }
-      setTimeout(() => handleEvent(event), 1000); // handle with delay to allow operations to vanish
+      setTimeout(() => handleEvent(event), 3000); // handle with delay to allow operations to vanish
```
edlerd, line 56: Why increase this to 3 seconds? This makes LXD appear a lot less responsive.

src/pages/instances/Events.tsx
```diff
@@ -26,6 +28,8 @@ const Events: FC = () => {
       eventCallback.onFinish?.();
       eventQueue.remove(event.metadata.id);
     }
+
+    setTimeout(() => void refetchOperations(), 3000);
```
edlerd, line 32: Why adding the refetch here with a 3s delay?

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -54,16 +56,31 @@ const InstanceDetailHeader: FC<Props> = ({ name, instance, project }) => {
...
-          (msg) => notify.failure("Renaming failed", new Error(msg)),
+          (msg) =>
+            toastNotify.failure(
+              "Error",
+              new Error(msg),
+              <>
+                Renaming instance {generateInstanceLink({ operation, project })}{" "}
+                failed.
+              </>,
+            ),
```
edlerd, line 83: Avoid the error title?

src/pages/instances/forms/CreateInstanceSnapshotForm.tsx
```diff
@@ -70,14 +72,19 @@ const CreateInstanceSnapshotForm: FC<Props> = ({
...
               );
               resetForm();
               close();
             },
             (msg) => {
-              notify.failure("Snapshot creation failed", new Error(msg));
+              toastNotify.failure(
+                "Error",
+                new Error(msg),
+                `Snapshot creation failed for instance ${instance.name}`,
```
edlerd, line 86: Error title?

tests/helpers/notification.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const toggleNotificationList = async (page: Page) => {
+  const listToggleButton = page.getByRole("button", {
+    name: "Expand notifications list",
+  });
+  await listToggleButton.click();
+};
+
+export const dismissNotificationFromList = async (page: Page) => {
```
edlerd, line 29: better `dismissFirstNotificationFormList`?

src/util/instances.tsx
```diff
@@ -0,0 +1,23 @@
+import { LxdOperationResponse } from "types/operation";
+import { getInstanceName } from "./operations";
+import InstanceLink from "pages/instances/InstanceLink";
+import React from "react";
+
+export const generateInstanceLink = (args: {
+  instanceName?: string;
+  operation?: LxdOperationResponse;
+  project?: string;
```
edlerd, line 9: I think this interface should be more concrete. Currently, you can give it either an instanceName or an operation. I think that should be two functions, one giving you a link from a string and one for an operation. Wdyt?

src/sass/styles.scss
```diff
@@ -87,12 +88,18 @@ $border-thin: 1px solid $color-mid-light !default;
...
 @import "storage_pool_form";
 @import "storage_volume_form";
 @import "storage";
+@import "toast";
 @import "upper_controls_bar";
 
+body {
+  overflow: hidden;
+}
+
```
edlerd, line 102: This is probably not needed.

src/pages/storage/forms/EditVolumeSnapshotForm.tsx
```diff
@@ -70,7 +72,11 @@ const EditVolumeSnapshotForm: FC<Props> = ({ volume, snapshot, close }) => {
           }
         },
         (msg) => {
-          notify.failure("Snapshot rename failed", new Error(msg));
+          toastNotify.failure(
+            "Error",
+            new Error(msg),
+            `Snapshot ${snapshot.name} rename failed`,
+          );
```
edlerd, line 79: Error title?

src/pages/storage/forms/CreateVolumeSnapshotForm.tsx
```diff
@@ -53,12 +55,16 @@ const CreateVolumeSnapshotForm: FC<Props> = ({ close, volume }) => {
...
               close();
               resetForm();
             },
             (msg) => {
-              notify.failure("Snapshot creation failed", new Error(msg));
+              toastNotify.failure(
+                "Error",
+                new Error(msg),
+                `Snapshot ${values.name} creation failed`,
+              );
```
edlerd, line 67: Error title?

src/pages/storage/actions/snapshots/VolumeSnapshotActions.tsx
```diff
@@ -34,8 +36,13 @@ const VolumeSnapshotActions: FC<Props> = ({ volume, snapshot }) => {
...
       eventQueue.set(
         operation.metadata.id,
-        () => notify.success(`Snapshot ${snapshot.name} deleted`),
-        (msg) => notify.failure("Snapshot deletion failed", new Error(msg)),
+        () => toastNotify.success(`Snapshot ${snapshot.name} deleted`),
+        (msg) =>
+          toastNotify.failure(
+            "Error",
+            new Error(msg),
+            `Snapshot ${snapshot.name} deletion failed`,
```
edlerd, line 44: Error title?

src/pages/storage/CreateStoragePool.tsx
```diff
@@ -63,12 +65,8 @@ const CreateStoragePool: FC = () => {
...
             queryKey: [queryKeys.storage],
           });
-          navigate(
-            `/ui/project/${project}/storage`,
-            notify.queue(
-              notify.success(`Storage pool ${storagePool.name} created.`),
-            ),
-          );
+          navigate(`/ui/project/${project}/storage`),
+            toastNotify.success(`Storage pool ${storagePool.name} created.`);
```
edlerd, line 69:
```suggestion
          navigate(`/ui/project/${project}/storage`),
            toastNotify.success(`Storage pool ${storagePool.name} created.`);
```
```suggestion
          navigate(`/ui/project/${project}/storage`),
            toastNotify.success(`Storage pool ${storagePool.name} created.`);
```
```suggestion
          navigate(`/ui/project/${project}/storage`);
          toastNotify.success(`Storage pool ${storagePool.name} created.`);
```

piperdeck:
Hi Mason, this is looking fantastic.

Some feedback:

# Status Bar

1. Can we please adjust the spacing of the notification count in the footer, so that the badges are closer to each other than they are to the total count? Something closer to this:

image

# Filters

2. I’ve noticed an inconsistency, where the filter buttons’ inactive and hover states are different dimensions from their active state. This causes the container to shift slightly when they are toggled. Could we please adjust them so that they are always the same size? specifically, I think the larger size of the active state is better.

| Hover state | Active state |
| --- | --- |
| image | image |

 # Notification List

3. If a notification comes in while the list is open, it pushes all the previous notifications down. I’m worried this could be error-prone, as the user might be about to dismiss a notification and dismiss the wrong one. Can we make it so that the scrollable area expands upwards when the list is open, so that visible notifications never move on their own if the list is open?

4. I’m finding that the show/hide animation for the notification list is a little bit on the slower side. Could we decrease the duration of that animation to make it a bit snappier?

5. The notification list interacts a little awkwardly with the status bar footer.
- During its animation, the borders of the notification list overlap with the footer in an awkward way.
- While expanded, the drop-shadow of the notification list fall on top of the status bar, which doesn’t quite look right to my eyes.
As a rule, could we make it so that the status bar is always on top of the notification list at all times, including its drop shadow and while it is being animated?

6. I’m finding the experience of manually dismissing notifications from the notification list to be a little bit jarring. Could we add an animation that slides notifications down or up to fill the gap left by a manually dismissed notification?
Dismissing filtered notifications is fine without an animation, since it totally resets the notification list.

7. A note on accessibility: I’ve noticed that while the notification list is open, it acts as a keyboard trap, since it prevents the user from tabbing to the button to close the list.
I propose 2 things:

- that the close button for the notification list be added to the tab order (last, of course)
- that the notification list be closable by pressing [Esc]

That's all I've found for now! Looking forward to getting some users and UX designers to have a look, which I'll do this week.

src/context/toastNotificationProvider.tsx
mas-who: yeah that makes sense, I have combined the 3 methods into `clear` and also moved the filter logic into `ToastNotificationList`

src/context/toastNotificationProvider.tsx
mas-who: This was done previously out of consideration for performance. I think I may have attempted to optimise too early in this case. The count is no longer a state and recalculated per render cycle.

src/util/notifications.tsx
```diff
@@ -0,0 +1,19 @@
+import { ICONS } from "@canonical/react-components";
+
+export type Filters = {
+  [filter: string]: boolean;
```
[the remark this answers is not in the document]
mas-who: I have adjusted the type for `Filters`. I think having filters as an object allows for faster filtering of notifications since we don't need to iterate the array of filters for each notification. Also when adding/removing filters, we wound't have to iterate the array of filters each time.

src/components/Version.tsx
```diff
@@ -20,20 +20,18 @@ const Version: FC = () => {
...
             tooltipClassName="version-warning"
             zIndex={1000}
           >
-            <Icon name="warning" className="p-side-navigation__icon" />
+            <Icon name="warning" className="version-warning-icon" />
           </Tooltip>
         )}
-        Version {serverVersion}-ui-{UI_VERSION}
-      </li>
+        LXD Version {serverVersion}-ui-{UI_VERSION}
```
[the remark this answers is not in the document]
mas-who: It was on the design :sweat_smile: I think piper wanted to add LXD to make it more clear the difference between lxd version and ui version

src/pages/images/actions/DeleteImageBtn.tsx
mas-who: I added a bit more detail to the error title and I was worried that the title might become too long because we don't really know the length of certain object properties. So I moved the title to the message body instead. But yeah I can change it back to how it was.

src/pages/instances/EditInstance.tsx
mas-who: In this case, the title can't be a react node and I wanted to have the `instanceLink` in the notification. I think I will have `Instance update failed.` in the title then just keep the link to the instance in the message body.

mas-who: I've done this for all generic `Error` titles

src/pages/instances/Events.tsx
mas-who: sorry I forgot to reverse this after experimenting. I found that the operations fetched from the server is not perfectly in sync with messages from the websocket so some delay is required or else we might have the number of operations stuck unless if we do a browser refresh.

src/pages/instances/Events.tsx
mas-who: This is related to the previous comment. When an event is handled, I found that if we refetch operations immediately then we might get back the same results from the server, it almost feels like operations on the backend isn't cleared immediately when events are emitted. The combination of delays on the event handling as well as refetching operations allows for both to keep in sync. The 3s delay is an empirical result from experimenting really.

src/pages/instances/Events.tsx
mas-who: so I put the delay on event handling to 1500ms which makes the ui feel a lot more fluid, but increased the `refetchOperations` delay to 3500ms to allow for operations to dissipate consistently, this seems to work consistently as well.

src/util/instances.tsx
mas-who: yeah that sounds good, I will put the operation link generator in the `util/operations` file

src/sass/styles.scss
mas-who: this is needed for the animation of translating along the y-axis. During the animation it overflows the body which causes a scrollbar to appear. If we don't want this styling maybe we can get rid of the sliding animation and just have fade in/out?

src/components/ScrollableTable.tsx
```diff
@@ -24,7 +24,8 @@ const ScrollableTable: FC<Props> = ({
     }
     const tBody = table.children[1];
     const above = tBody.getBoundingClientRect().top + 1;
-    const below = getAbsoluteHeightBelow(belowId);
+    const below =
+      getAbsoluteHeightBelow(belowId) + getAbsoluteHeightBelow("status-bar");
```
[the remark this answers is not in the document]
mas-who: yup, 100% better, also get rid of the assumption that status bar will have an ID of `status-bar`

src/components/Version.tsx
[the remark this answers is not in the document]
edlerd: Not sure if the LXD prefix clarifies UI vs. backend version. As this is LXD-UI, I guess a user won't make a difference between front- and backend. Also this is a combined version string from both parts.

@piperdeck what do you think? Shall we keep the footer string as "Version: 1.2.3" and strip the "LXD" prefix from it? I think we already say LXD-UI in the logo of the side navigation.

src/sass/styles.scss
edlerd: Ah right. I think sliding is the right animation type here. We could probably have a wrapper for the NotificationList and apply the overflow hidden to that wrapper only. But then again, I am not sure if this is even a better approach. Maybe keep as is -- but better add a comment to this CSS. Because currently, it is unclear why this overflow hidden on body is needed.

src/util/notifications.tsx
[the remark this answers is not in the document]
edlerd: Maybe a better approach is to use a `new Set([])` then? So we have an ordered array with constant time lookup and simple add/remove/isEmpty functions in the interface.

src/context/toastNotificationProvider.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+    const removeIdLookup: { [id: string]: boolean } = {};
+    notifications.forEach((notification) => {
+      removeIdLookup[notification.id] = true;
+    });
+
+    setNotifications((prev) => {
+      const newNotifications = prev.filter(
+        (existingNotification) => !removeIdLookup[existingNotification.id],
+      );
```
edlerd, line 145: Probably simpler to use a set?
```suggestion
    setNotifications((prev) => {
      const removeIdLookup = new Set(notifications);
      const newNotifications = prev.filter((item) => removeIdLookup.has(item));
```

src/context/toastNotificationProvider.tsx
edlerd: nice, thanks :)

src/components/StatusBar.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const StatusBar: FC<Props> = ({ className }) => {
+  const { toggleListView, notifications, countBySeverity, isListView } =
+    useToastNotification();
+
+  const notificationIcons = severityOrder.map((severity) => {
+    if (countBySeverity[severity]) {
+      return (
+        <Icon
+          key={`status-${severity}`}
```
edlerd, line 21:
```suggestion
          key={severity}
```
Keys can be as simple as possible

src/components/OperationStatus.tsx
```diff
@@ -0,0 +1,33 @@
...
+import React, { useState } from "react";
+import { isWidthBelow } from "util/helpers";
+import useEventListener from "@use-it/event-listener";
+import { Link } from "react-router-dom";
+import { pluralize } from "util/instanceBulkActions";
+
+const isSmallScreen = () => isWidthBelow(620);
+
+const OperationStatus = () => {
+  const [reduceText, setReduceText] = useState(false);
```
edlerd, line 12: Better call the state vars `isSmallScreen`?  I didn't get initially what `reduceText` is used for. We can directly call isWidthBelow in line 15 maybe to avoid duplicating the name?

src/context/toastNotificationProvider.tsx
```diff
@@ -0,0 +1,209 @@
...
+  FC,
+  PropsWithChildren,
+  ReactNode,
+  createContext,
+  useContext,
+  useEffect,
+  useState,
+} from "react";
+
+const HIDE_NOTIFICATION_DELAY = 5 * 1000;
```
edlerd, line 22:
```suggestion
const HIDE_NOTIFICATION_DELAY = 5_000;
```

src/pages/instances/Events.tsx
edlerd: I think a better solution is to only consider "running" operations in operationStatus. Then we don't need to change the timing here and can remove the refetch timer in this `Events` component.

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -241,15 +240,13 @@ const InstanceSnapshots = (props: Props) => {
...
-      <NotificationRowLegacy
-        notification={inTabNotification}
-        onDismiss={() => setInTabNotification(null)}
-      />
+      <NotificationRow />
       {hasSnapshots ? (
         <>
           <ScrollableTable
-            dependencies={[filteredSnapshots, inTabNotification]}
+            dependencies={[filteredSnapshots, notify.notification?.message]}
```
edlerd, line 247:
```suggestion
            dependencies={[filteredSnapshots, notify.notification]}
```

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -40,15 +42,25 @@ const AttachIsoBtn: FC<Props> = ({ instance }) => {
...
             <>
-              ISO <b>{attachedIso?.source ?? ""}</b> detached
+              ISO <b>{attachedIso?.source ?? ""}</b> detached from{" "}
+              {instanceLink}
             </>,
           ),
-        (msg) => notify.failure("Detach ISO failed", new Error(msg)),
+        (msg) =>
+          toastNotify.failure(
+            "Detaching ISO from failed.",
```
edlerd, line 60:
```suggestion
            "Detaching ISO failed.",
```

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -67,15 +79,24 @@ const AttachIsoBtn: FC<Props> = ({ instance }) => {
...
+          toastNotify.success(
             <>
-              ISO <b>{image.aliases}</b> attached
+              ISO <b>{image.aliases}</b> attached to {instanceLink}
             </>,
           ),
-        (msg) => notify.failure("Attaching ISO failed", new Error(msg)),
+        (msg) =>
+          toastNotify.failure(
+            "Attaching ISO to failed.",
```
edlerd, line 96:
```suggestion
            "Attaching ISO failed.",
```

src/pages/storage/CustomVolumeSelectModal.tsx
```diff
@@ -115,7 +115,7 @@ const CustomVolumeSelectModal: FC<Props> = ({
       <NotificationRow />
       <ScrollableTable
         dependencies={[volumes, rows, notify.notification]}
-        belowId="modal-footer"
+        belowIds={["modal-footer", "status-bar"]}
```
edlerd, line 118: Here we don't have a status-bar in the modal.
```suggestion
        belowIds={["modal-footer"]}
```

src/pages/storage/CustomVolumeCreateModal.tsx
```diff
@@ -62,7 +64,7 @@ const CustomVolumeCreateModal: FC<Props> = ({
           void queryClient.invalidateQueries({
             queryKey: [queryKeys.customVolumes],
           });
-          notify.success(`Storage volume ${values.name} created.`);
+          toastNotify.success(`Storage volume ${values.name} created.`);
```
edlerd, line 67: This doesn't work. The notification is displayed behind the modal, then the modal to select the newly created volume is shown and only after selection this modal closes and shows the notification. Maybe this is a candidate to keep an inline notification for the 2nd modal -- after the volume is created.

src/util/instances.tsx
```diff
@@ -0,0 +1,26 @@
+import { LxdOperationResponse } from "types/operation";
+import { getInstanceName } from "./operations";
+import InstanceLink from "pages/instances/InstanceLink";
+import React from "react";
+
+export const generateInstanceLinkFromName = (args: {
```
edlerd, line 6: Better `instanceLinkFromName` ?

src/util/instances.tsx
```diff
@@ -0,0 +1,26 @@
...
+  instanceName: string;
+  project?: string;
+}) => {
+  const { project, instanceName } = args;
+  return (
+    <InstanceLink instance={{ name: instanceName, project: project || "" }} />
+  );
+};
+
+export const generateInstanceLinkFromOperation = (args: {
```
edlerd, line 16: Better `instanceLinkFromOperation`?

src/util/notifications.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const severityOrder = [
+  "positive",
+  "caution",
+  "negative",
+  "information",
+] as const;
+
+export type Filters = Partial<{
+  [key in (typeof severityOrder)[number]]: boolean;
+}>;
```
edlerd, line 12: I think this can move into ToastNotificationList as it is a local concept.

src/util/updateMaxHeight.tsx
```diff
@@ -17,7 +19,8 @@ export const updateMaxHeight = (
   const below = belowElements
     ? belowElements[0].getBoundingClientRect().height + 1
     : 0;
-  const offset = Math.ceil(above + below + additionalOffset);
+  const statusBarHeight = getAbsoluteHeightBelow("status-bar");
+  const offset = Math.ceil(above + below + statusBarHeight + additionalOffset);
```
edlerd, line 23: Here we have the same situation as with the `ScrollableTable` and `ScrollableContainer` components. There are use cases like from `CustomVolumeCreateModal` where we don't need to take the status bar into account.

src/components/BaseLayout.tsx
```diff
@@ -7,6 +7,7 @@ interface Props {
   children: ReactNode;
   mainClassName?: string;
   contentClassName?: string;
+  dependencies?: DependencyList;
```
edlerd, line 10: We don't need the dependencies here, do we?

src/components/CustomLayout.tsx
```diff
@@ -1,11 +1,12 @@
-import React, { FC, ReactNode } from "react";
+import React, { DependencyList, FC, ReactNode } from "react";
 import classnames from "classnames";
 
 interface Props {
   header?: ReactNode;
   children: ReactNode;
   mainClassName?: string;
   contentClassName?: string;
+  dependencies?: DependencyList;
```
edlerd, line 9: We don't need the dependencies here, do we?

mas-who:
> Hi Mason, this is looking fantastic.
> 
> Some feedback:
> 
> # Status Bar
> 1. Can we please adjust the spacing of the notification count in the footer, so that the badges are closer to each other than they are to the total count? Something closer to this:
> 
> image

Done.

> # Filters
> 2. I’ve noticed an inconsistency, where the filter buttons’ inactive and hover states are different dimensions from their active state. This causes the container to shift slightly when they are toggled. Could we please adjust them so that they are always the same size? specifically, I think the larger size of the active state is better.
> 
> Hover state	 Active state
> image	image

Done.

> # Notification List
> 3. If a notification comes in while the list is open, it pushes all the previous notifications down. I’m worried this could be error-prone, as the user might be about to dismiss a notification and dismiss the wrong one. Can we make it so that the scrollable area expands upwards when the list is open, so that visible notifications never move on their own if the list is open?

Let's discuss this, I'm not sure if I am entirely following what you mean by the above.

> 4. I’m finding that the show/hide animation for the notification list is a little bit on the slower side. Could we decrease the duration of that animation to make it a bit snappier?

Done. I have halved the duration.

> 5. The notification list interacts a little awkwardly with the status bar footer.
> 
> * During its animation, the borders of the notification list overlap with the footer in an awkward way.
> * While expanded, the drop-shadow of the notification list fall on top of the status bar, which doesn’t quite look right to my eyes.
>   As a rule, could we make it so that the status bar is always on top of the notification list at all times, including its drop shadow and while it is being animated?

Done.

> 6. I’m finding the experience of manually dismissing notifications from the notification list to be a little bit jarring. Could we add an animation that slides notifications down or up to fill the gap left by a manually dismissed notification?
>    Dismissing filtered notifications is fine without an animation, since it totally resets the notification list.

Still looking at this as it it pretty tricky to implement.

> 7. A note on accessibility: I’ve noticed that while the notification list is open, it acts as a keyboard trap, since it prevents the user from tabbing to the button to close the list.
>    I propose 2 things:
> 
> * that the close button for the notification list be added to the tab order (last, of course)
> * that the notification list be closable by pressing [Esc]
> 
* I don't think altering the tab index for the close button makes sense as you have to click on the close button (or tab to the close button) first to open the notification list. You can navigate backwards by pressing SHIFT + Tab, but if you adjust the close button to be at the end of the tab order of the notification list then that will violate the document structure and cause a infinite loop.
* Implemented close notifications by pressing [Esc]

src/util/notifications.tsx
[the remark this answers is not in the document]
mas-who: good idea, I should use `Set` more!

src/context/toastNotificationProvider.tsx
mas-who: nice! :+1:

src/components/OperationStatus.tsx
mas-who: yeah that makes sense

src/context/toastNotificationProvider.tsx
mas-who: I forgot this syntax existed haha :+1:

src/pages/storage/CustomVolumeCreateModal.tsx
mas-who: Yeah I think inline is reasonable here since you can't really navigate anywhere while on this modal, unless if you close or cancel it

mas-who: great catch btw :+1:

src/util/updateMaxHeight.tsx
mas-who: this one is slightly more annoying to adjust, since the arguments to `updateMaxHeight` are all positional, if I add an extra argument `belowIds` then I would have to pass in a bunch of undefined inputs to `updateMaxHeight` everywhere that it is called. I'm thinking maybe I will define an input for considering the status bar or not, then default it to `true`, then only in specific locations I'd have to disable this behaviour for example in `CustomVolumeCreateModal`

src/components/BaseLayout.tsx
mas-who: no we don't, this should have been removed, will remove

src/components/CustomLayout.tsx
mas-who: no we don't, this should have been removed, will remove

src/util/updateMaxHeight.tsx
edlerd: Yeah, it has a lot of usages. Adding a property is probably the easiest way to get the few exceptions right.

mas-who: I can't seem to find any other relevant places than `CustoVolumeCreateModal`, not sure if I missed some?

src/context/toastNotificationProvider.tsx
```diff
@@ -0,0 +1,203 @@
...
+  failure: () => initialNotification,
+  clear: () => null,
+  toggleListView: () => null,
+  isListView: false,
+  countBySeverity: {},
+});
+
+interface Props {}
+
+const ToastNotificationProvider: FC<PropsWithChildren<Props>> = ({
```
edlerd, line 67:
```suggestion
const ToastNotificationProvider: FC<PropsWithChildren> = ({ children }) => {
```
No need for an empty interface.

src/context/toastNotificationProvider.tsx
```diff
@@ -0,0 +1,203 @@
...
+  children,
+}) => {
+  const [notifications, setNotifications] = useState<ToastNotificationType[]>(
+    [],
+  );
+  const [showList, setShowList] = useState(false);
+  const [notificationTimer, setNotificationTimer] =
+    useState<NodeJS.Timeout | null>(null);
+
+  // cleanup on timer if unmounted, not likely though since this component sits above App
```
edlerd, line 77:
```suggestion
  // cleanup on timer if unmounted
```

Would remove that, as the context this is used in might change, no?

src/pages/instances/Events.tsx
```diff
@@ -26,6 +28,8 @@ const Events: FC = () => {
       eventCallback.onFinish?.();
       eventQueue.remove(event.metadata.id);
     }
+
+    void refetchOperations();
```
edlerd, line 32:
```suggestion
```
We shouldn't need to refetch again, we already do that below. This results in a lot of additional requests.

src/pages/instances/Events.tsx
```diff
@@ -47,6 +51,7 @@ const Events: FC = () => {
         void queryClient.invalidateQueries({
           queryKey: [queryKeys.operations, event.project],
         });
+        void refetchOperations();
```
edlerd, line 54: I wonder if we should debounce the refetch. Especially with bulk actions there can be very many updates in short time.

src/components/forms/ScrollableConfigurationTable.tsx
```diff
@@ -20,7 +20,7 @@ const ScrollableConfigurationTable: FC<Props> = ({
   return (
     <ScrollableTable
       dependencies={[notify.notification]}
-      belowId="form-footer"
+      belowIds={["form-footer"]}
```
edlerd, line 23: We sometimes do need to encounter for the status bar here as well. I.e. on instance creation > network and others.

src/util/updateMaxHeight.tsx
edlerd: I think that is right, only that one I could spot as well!

src/components/Animate.tsx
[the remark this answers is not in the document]
mas-who: oooh interesting, I'm going to do some research how to support this in react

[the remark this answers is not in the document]
edlerd: I guess a [hook like this](https://github.com/Slashgear/use-reduced-motion/blob/master/src/index.ts) would do.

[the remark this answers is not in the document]
edlerd: We could also inline the hook in the Animate component.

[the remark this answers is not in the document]
mas-who: I think having a separate hook is better, in more complex situations we might need to use the web animation API outside of the `Animate` wrapper

src/pages/instances/Events.tsx
```diff
@@ -47,6 +49,7 @@ const Events: FC = () => {
...
       }
       setTimeout(() => handleEvent(event), 1000); // handle with delay to allow operations to vanish
```
edlerd, line 54: I think this timeout can be removed now. we can directly call handleEvent without a delay.

mas-who: what was the reason that this was needed in the first place?

edlerd: I think it was added because of some timing issues on the instance list for instances on creation. But that code is much more robust now and checking operation status as well as if the instance is in the get instances api response to avoid duplicates. Hence I suggest to make this instant without the timeout.

src/util/usePreferReducedMotion.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const listener = (event: MediaQueryListEvent) => {
+      setPrefersReducedMotion(event.matches);
+    };
+    mediaQuery.addEventListener("change", listener);
+    return () => {
+      mediaQuery.removeEventListener("change", listener);
+    };
+  }, []);
+
+  console.log(prefersReducedMotion);
```
edlerd, line 21:
```suggestion
```

src/components/ToastNotification.tsx
```diff
@@ -0,0 +1,57 @@
...
+            opacity: 0,
+          }}
+          to={{
+            opacity: 1,
+          }}
+          exitAnimation={[
+            { opacity: 1, transform: "translateY(0)" },
+            { opacity: 0, transform: "translateY(50px)" },
+          ]}
+          options={{ duration: 200 }}
```
edlerd, line 34: We need something like this class here as well.
```suggestion
      className="toast-notification-list-animate"

```

edlerd (approved):
This has come a long way. very good work.
Two open things remain:
1. the caption of the version in the status bar and 
2. adding the same fix for the notification list to a single notification.

I'll approve this ahead of time, seems very close to being ready to merge.

src/components/ToastNotification.tsx
mas-who: Thanks David, I've done the fix for the single notification, will chat to @piperdeck about the LXD version prefix

piperdeck:
> Maintain scroll position for new notifications when list is expanded.

This doesn't seem to be working, at least in my testing. In between these screenshots, a notification comes in:

| | | |
| --- | --- | --- |
| image | image | image |

> Animate dismissal of notifications when list is expanded

This looks awesome!

mas-who:
> > Maintain scroll position for new notifications when list is expanded.
> 
> This doesn't seem to be working, at least in my testing. In between these screenshots, a notification comes in:
> 
> image	image	image

Thanks for finding this issue! I see the main problem is that the scroll height should only be adjusted when the scroll bar is at the top most position. I was adjusting the scroll height after the initial offset causing it to shift exponentially. Please check again on this, it should be resolved now.

mas-who:
@piperdeck for pop up modals that appears above everything else on a page like the one shown below (when creating custom volumes from the instance configuration page), I think we should still have inline notifications since the user can't really navigate away from the popup modal unless they cancel what they are currently doing, so any messages shown would be completely scoped to the current context of the modal. wdyt?

Screenshot from 2024-02-08 09-45-13

src/components/ToastNotificationList.tsx
```diff
@@ -0,0 +1,239 @@
...
+  groupedCount = {},
+  show,
+}) => {
+  const [filters, setFilters] = useState<Set<FilterTypes>>(new Set());
+  const prevNotificationsSize = useRef<number>(notifications.length);
+  const containerRef = useRef<HTMLUListElement>(null);
+  const hasFilters = !!filters.size;
+  const preferReducedMotion = usePrefersReducedMotion();
+
+  useLayoutEffect(() => {
```
mas-who, line 37: @lorumic or @aaryanporwal , when you have some time could you review the content of this `useLayoutEffect`? The purpose for this here is to adjust the scroll height of the list container if a new item gets added when the scroll bar is at its initial top position. This makes the scroll visually stable, something requested by piper.

src/components/ToastNotificationList.tsx
```diff
@@ -0,0 +1,239 @@
...
+      const dismissText = (
+        <span className="dismiss-text">Dismiss {totalCount}</span>
+      );
+      return dismissText;
+    }
+
+    return <span>Dismiss all</span>;
+  };
+
+  const handleDismissNotification = (notification: ToastNotificationType) => {
```
mas-who, line 145: @lorumic or @aaryanporwal , this code here is used to animate dismissing a single notification from the notification list. The notification will fade out and have its height reduced to 0 before being unmounted from the DOM. In addition, if the user has the `prefers-reduced-motion` setting enabled on their OS, then no animation will be shown. You can emulate `prefers-reduced-motion` in chrome using this [guide](https://developer.chrome.com/docs/devtools/rendering/emulate-css#emulate_css_media_feature_prefers-reduced-motion).

src/components/Version.tsx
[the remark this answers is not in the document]
mas-who: @piperdeck suggested `lxd 5.20-ui-0.5` which I think looks good. Although this is a branding level decision which I'd like to get your final go ahead. For now, I will keep the version text the same as before, will adjust it in a separate PR later if we want to change it.

lorumic (approved):
Looks awesome and works really well! Well done, @mas-who!

piperdeck:
Having another look over it, and I just noticed that the filter button's hover state doesn't seem to be responsive to the size of the text inside it:

image

In this example, the button should be wider to accommodate the wider "12" inside it.

Other than that, I think this is ready to go.

piperdeck:
> for pop up modals that appears above everything else on a page like the one shown below (when creating custom volumes from the instance configuration page), I think we should still have inline notifications since the user can't really navigate away from the popup modal unless they cancel what they are currently doing, so any messages shown would be completely scoped to the current context of the modal. wdyt?

I agree, I think we can have inline notifications for errors that occur in an interaction which happens entirely within a modal

mas-who:
> Having another look over it, and I just noticed that the filter button's hover state doesn't seem to be responsive to the size of the text inside it:
> 
> image
> 
> In this example, the button should be wider to accommodate the wider "12" inside it.
> 
> Other than that, I think this is ready to go.

Fixed :slightly_smiling_face: 
Screenshot from 2024-02-08 18-59-22

piperdeck:
Awesome! This looks good to me

PR #608, opened 2024-01-17
Update dependencies to vite 5

edlerd:
## Done

- updated all dependencies
- adjust config
- adjust formatting

## QA

1. Run the LXD-UI:
    - check build and ci passes
    - spot check demo server


mas-who:
Reviewed the changes, looks good, just leaving this Vite v4 to v5 migration guide here for reference: https://vitejs.dev/guide/migration.html.

QA on the demo server looks mostly okay, clicked on spot checked most actions (links, buttons etc). Only saw one issue, when I tried to access the terminal, or console for an instance the below error pops up
Screenshot from 2024-01-18 09-47-48

I checked on demo 307 and terminal session works fine there. So I think it's got to do with some related dependency update?

edlerd:
> QA on the demo server looks mostly okay, clicked on all tabs and spot checked most actions (links, buttons etc). Only saw one issue, when I tried to access the terminal, or console for an instance the below error pops up

Well spotted, thanks! I reverted the xterm-addon-fit dependency. It expects xterm5, but xterm-react is limited to xterm 4. We should look into this independently. Also we shoudl add a test case for the terminal. I'll create some follow ups for both.

mas-who:
> > QA on the demo server looks mostly okay, clicked on all tabs and spot checked most actions (links, buttons etc). Only saw one issue, when I tried to access the terminal, or console for an instance the below error pops up
> 
> Well spotted, thanks! I reverted the xterm-addon-fit dependency. It expects xterm5, but xterm-react is limited to xterm 4. We should look into this independently. Also we shoudl add a test case for the terminal. I'll create some follow ups for both.

Thanks David for the update! Just checked again and all seems good now :+1:

mas-who (approved):

PR #614, opened 2024-01-19
fix: resolve notification provider issue with react-router-dom upgrade for good

mas-who:
## Done

- This is a sister PR for the react-components [fix](https://github.com/canonical/react-components/pull/1022) in `NotificationProvider`. Instead of calling `useLocation` in `NotificationProvider`, we will now pass the returned `state` and `pathname` as props to the provider component.
- Upgraded react-router-dom to 6.21.3

## QA

1. Run the LXD-UI:
    - Wait until the react component PR gets merged, then check that queued notifications still work.

### DIFF
--- a/src/context/notifications.tsx
+++ b/src/context/notifications.tsx
@@ -1,0 +1,15 @@
+import React, { FC, PropsWithChildren } from "react";
+import { NotificationProvider as RCNotificationProvider } from "@canonical/react-components";
+import { useLocation } from "react-router-dom";
+import { QueuedNotification } from "@canonical/react-components";
+
+const NotificationProvider: FC<PropsWithChildren> = ({ children }) => {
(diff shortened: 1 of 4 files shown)


src/context/notifications.tsx
edlerd, line 1: There's only a single use case of this and we don't expect more, so why not include the changes directly in the Root component?

mas-who: > There's only a single use case of this and we don't expect more, so why not include the changes directly in the Root component? Similar solution in #616

It was just a personal preference, since only notification provider needed state and path from useLocation I decided to keep it close to notification provider instead of Root. But your approach makes more sense :+1:

mas-who:
Closing this PR as the same issue was addressed by PR #616

PR #620, opened 2024-01-22
Allow bulk select/delete images in a project and unify pagination placement

### LINKED ISSUE #618: [FR] Allow deleting images by selection box/checkbox, similar to how deleting instances is today
While deploying and wiping LXD containers, VMs and images to test the caching of an upstream MiTM SSL proxy, it became clear that deleting the images from the UI is a lot more work than it could be, for large numbers of images. 

I'm proposing that the same methodology that is used in the instance list be used in the image list, so I can "select all" images and delete them in one shot, vs. clicking the little trash can to the right of each image directly. 

Yes, I can also do this via the CLI using `lxc image list -f json | jq -r '.[].fingerprint' | xargs rm`, but that's still clunky on its own. 

Ideally, something that allows the reuse of the same UX/components used elsewhere in the LXD UI would be ideal. 

Additionally, the ability to search for images/instances by a partial naming convention, and then "select all" of those matching for deletion, would be useful.

edlerd:
## Done

- allow bulk select and delete images
- extract `PageHeader` component
- refactor settings, instance list and project list to use the new header component
- move table pagination always below table

Fixes #618
## QA

1. Run the LXD-UI:
    - select one or many images and remove them from a project's image list
    - ensure layout of following pages (also responsive and with panel) is correct:
      - instance list
      - instance detail snapshots
      - profile list
      - image list
      - settings
      - storage volume list
      - custom storage volume snapshot list
      - custom iso list


mas-who:
I think we should align the page layout between images list and instances list? On the images page, when we bulk select, the delete button appears towards the right hand side of the page where on the instances page, the same action results in the delete button appear where the search bar is located (next to the other actions buttons). Maybe we can align layout and behaviour with the instances page?

**Images list**
Screenshot from 2024-01-23 09-43-39
Screenshot from 2024-01-23 09-43-50

**Instances list**
Screenshot from 2024-01-23 09-44-10
Screenshot from 2024-01-23 09-44-19

src/api/images.tsx
```diff
@@ -38,6 +44,29 @@ export const deleteImage = (
...
 
+export const deleteImageBulk = (
+  fingerprints: string[],
+  project: string,
+  eventQueue: EventQueue,
+): Promise<PromiseSettledResult<void>[]> => {
+  const results: PromiseSettledResult<void>[] = [];
+  return new Promise((resolve) => {
+    void Promise.allSettled(
+      fingerprints.map(async (name) => {
```
mas-who, line 55: there is no need to have `async/await` here

src/pages/images/ImageList.tsx
```diff
@@ -42,6 +46,16 @@ const ImageList: FC = () => {
     notify.failure("Loading images failed", error);
   }
 
+  useEffect(() => {
+    const validNames = new Set(images?.map((image) => image.fingerprint));
+    const validSelections = selectedNames.filter((name) =>
```
mas-who, line 51: this could be slightly more efficient, we could exit early if we found a mismatch?

```
for (const name of selectedNames) {
  if (!validNames.has(name)) {
    setSelectedNames(validSelections);
    break;
  }
}
```

edlerd: We have to calculate the `validSelections` as the overlap between selections and valid names. I don't see a way to do that without iterating and checking all selected items for validity.

mas-who: aah yes you are right, please disregard.

piperdeck:
This looks good to me, but I'm still not sure we should move the table pagination to the bottom of the page. For one thing, the reason we moved it to the top was because I observed users being confused and not finding the pagination controls when they were at the bottom, and for another thing, this might interact negatively with the new notifications which float at the bottom of the page.

Could we talk about this next week in a working together session?

edlerd:
> This looks good to me, but I'm still not sure we should move the table pagination to the bottom of the page. For one thing, the reason we moved it to the top was because I observed users being confused and not finding the pagination controls when they were at the bottom, and for another thing, this might interact negatively with the new notifications which float at the bottom of the page.
> 
> Could we talk about this next week in a working together session?

My main idea was to make the pagination placement consistent. In our discussion we concluded, due to users having trouble finding the select notifications if the pagination (with select info) was rendered below, to move all of them up.

Now, all `TablePagination`s are used with their default position of `above`.

Thank you for the fruitful discussion @piperdeck and @mas-who please give it another look as it is now.

src/components/PageHeader.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { FC, ReactNode } from "react";
```
mas-who, line 1: Based on our discussion from last week, an alternative to implement this component is to use a compound component pattern. This gets around having to use props for passing components, and have the additional benefit of allowing more control over children components. For example, we could expose className as a input prop for `HeaderLeftContainer` to allow styling customization for it. This is better than having a single `PageHeader` component, since if we want to customize any children components then we have to keep adding to the props of `PageHeader`. That being said, the current implementation acheives the same thing as your PageHeader, so feel free to keep things as is.

```
import React, { FC, PropsWithChildren, ReactNode } from "react";

interface HeaderLeftContainerProps {
  title: ReactNode;
}
const HeaderLeftContainer: FC<PropsWithChildren<HeaderLeftContainerProps>> = ({
  title,
  children,
}) => {
  return (
    <div className="page-header__left">
      <h1 className="p-heading--4 u-no-margin--bottom">{title}</h1>
      {children}
    </div>
  );
};

const SearchContainer: FC<PropsWithChildren> = ({ children }) => {
  return (
    <div className="page-header__search margin-right u-no-margin--bottom">
      {children}
    </div>
  );
};

const BaseActionsContainer: FC<PropsWithChildren> = ({ children }) => {
  return <div className="page-header__base-actions">{children}</div>;
};

interface PageHeader extends FC<PropsWithChildren> {
  HeaderLeftContainer: FC<PropsWithChildren<HeaderLeftContainerProps>>;
  SearchContainer: FC<PropsWithChildren>;
  BaseActionsContainer: FC<PropsWithChildren>;
}
const PageHeader: PageHeader = ({ children }) => {
  return <div className="p-panel__header page-header">{children}</div>;
};

PageHeader.HeaderLeftContainer = HeaderLeftContainer;
PageHeader.SearchContainer = SearchContainer;
PageHeader.BaseActionsContainer = BaseActionsContainer;
export default PageHeader;

/** Example usage */
const Example = () => {
  return (
    <PageHeader>
      <PageHeader.HeaderLeftContainer title="title">
        <BulkActions />
        <PageHeader.HeaderLeftContainer>
          <Search />
        </PageHeader.HeaderLeftContainer>
        <PageHeader.BaseActionsContainer>
          <BaseActions />
        </PageHeader.BaseActionsContainer>
      </PageHeader.HeaderLeftContainer>
    </PageHeader>
  );
};
```

src/pages/images/actions/BulkDeleteImageBtn.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const count = fingerprints.length;
+
+  const handleDelete = () => {
+    setLoading(true);
+    onStart();
+    void deleteImageBulk(fingerprints, project, eventQueue).then((results) => {
+      const { fulfilledCount, rejectedCount } =
+        getPromiseSettledCounts(results);
+      if (fulfilledCount === count) {
+        notify.success(
```
mas-who, line 37: making a mental note here to implement floating notificaiton once this is merged

src/pages/images/ImageList.tsx
```diff
@@ -162,49 +184,107 @@ const ImageList: FC = () => {
...
+          >
+            <p>
+              Images will appear here, when launching an instance from a remote.
+            </p>
+          </EmptyState>
+        )}
+        {images.length > 0 && (
+          <ScrollableTable
+            dependencies={[images]}
+            belowId="pagination"
```
mas-who, line 245: I don't think below ID is necessary if table pagination is rendered above

src/pages/instances/InstanceList.tsx
```diff
@@ -544,6 +549,7 @@ const InstanceList: FC = () => {
             <>
               <ScrollableTable
                 dependencies={[filteredInstances, notify.notification]}
+                belowId="pagination"
```
mas-who, line 552: belowId not necessary here

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -249,12 +249,14 @@ const InstanceSnapshots = (props: Props) => {
         <>
           <ScrollableTable
             dependencies={[filteredSnapshots, inTabNotification]}
+            belowId="pagination"
```
mas-who, line 252: belowId not necessary here

src/pages/profiles/ProfileList.tsx
```diff
@@ -218,10 +221,10 @@ const ProfileList: FC = () => {
             tableId="profile-table"
           >
             <TablePagination
-              data={sortedRows}
               id="pagination"
```
mas-who, line 224: can remove belowId for pagination

src/pages/storage/CustomIsoList.tsx
```diff
@@ -175,12 +175,16 @@ const CustomIsoList: FC<Props> = ({ project }) => {
         </div>
         <UploadCustomIsoBtn />
       </div>
-      <ScrollableTable dependencies={[images]} tableId="custom-iso-table">
+      <ScrollableTable
+        dependencies={[images]}
+        belowId="pagination"
```
mas-who, line 180: remove

src/pages/storage/StorageVolumeSnapshots.tsx
```diff
@@ -243,12 +243,14 @@ const StorageVolumeSnapshots: FC<Props> = ({ volume }) => {
         <>
           <ScrollableTable
             dependencies={[filteredSnapshots]}
+            belowId="pagination"
```
mas-who, line 246: remove

src/sass/_page_header.scss
```diff
@@ -0,0 +1,61 @@
+.page-header {
```
mas-who, line 1: nice consolidation of the page-header styling from all other components!

mas-who:
QA looks good, just some code review comments for consideration :slightly_smiling_face:

src/components/PageHeader.tsx
edlerd: good idea :+1:

mas-who:
LGTM.

mas-who (approved):

piperdeck:
When I select an item in the list, the delete button is appearing off to the side for some reason

image

This should be "created at"

image

Since this table's rows will always be at least 2 lines of text tall, maybe this description field should be able to occupy 2 lines before being truncated?

image

Other than that, this is looking good :)

edlerd:
Addressed the three issues you found @piperdeck -- please give it another pass.

piperdeck:
Everything looks good!

PR #621, opened 2024-01-22
Use vitest instead of react-scripts for tests

edlerd:
## Done

- remove react-scripts
- add vitest
- change script runner

## QA

1. Run the LXD-UI:
    - ensure ci is executing ts spec tests and succeeding


vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
```
mas-who, line 3: should we set the `globals` config so that we don't have to import test functions everywhere similar to jest? 

```
// vitest.config.ts
import { defineConfig } from 'vitest/config'

export default defineConfig({
  test: {
    globals: true,
  },
})
```

If we set this, then we just need to add this to `tsconfig.json` as well:

```
// tsconfig.json
{
  "compilerOptions": {
    "types": ["vitest/globals"]
  }
}
```

mas-who (approved):

mas-who:
LGTM. Just wondering if we should make vitest API imports global? Feel free to leave it as is.

vitest.config.ts
edlerd: Good idea, will change it :)

PR #623, opened 2024-01-24
Use config api for storage forms

edlerd:
## Done

- Use config api for storage forms for form fields help text and default values on storage pools and volumes 

Fixes WD-7393

## QA

1. Run the LXD-UI:
    - check create (and edit) storage pool (and storage volume) forms
    - pay attention to help texts and default values.


mas-who:
1. Snapshot name pattern config has weird description. I am aware there is a discussion with the core team about this, maybe we can manually remove this for now and revert the change when the backend sends more appropriate config descriptions?
Screenshot from 2024-01-26 09-23-05
2. Interesting that zfs.block_size config has a link for zfs.block_mode in its help text but the actual help text for zfs.block_mode config doesn't have a link. Not sure if we can do anything about this one though.
Screenshot from 2024-01-26 09-39-48

src/types/config.d.ts
```diff
@@ -28,5 +29,14 @@ export interface LxdConfigOptions {
     instance: LxcConfigOptionCategories;
     project: LxcConfigOptionCategories;
     server: LxcConfigOptionCategories;
+    "storage-btrfs": LxcConfigOptionCategories;
+    "storage-ceph": LxcConfigOptionCategories;
+    "storage-cephfs": LxcConfigOptionCategories;
```
mas-who, line 34: do we want to include `storage-cephfs` and `storage-cephobject` in this type now even if it's not supported in the ui yet?

src/util/storagePool.tsx
```diff
@@ -1,37 +1,45 @@
...
-  formField: string,
-): ConfigRowMetadata => {
-  if (formField in cephStoragePoolDefaults) {
-    return { value: cephStoragePoolDefaults[formField], source: "LXD" };
-  }
-  return { value: "", source: "LXD" };
-};
-
-const cephStoragePoolFormFieldToPayloadName: Record<string, string> = {
+export const storagePoolFormFieldToPayloadName: Record<string, string> = {
```
mas-who, line 5: Renaming this to a generic function name is a little weird for me since the lookup contains only ceph configs. That being said, only ceph storage pools have extra configs like these, so I guess generalising is fine since this lookup could contain configs for other driver types in the future. I'm just a little conflicted since I don't know if that future scenario is plausible looking at all the supported drivers from the lxd documentation.

src/util/config.tsx
```diff
@@ -39,7 +39,7 @@ export const configDescriptionToHtml = (
   // documentation links
   if (objectsInvTxt) {
     // tags like {ref}`instance-options-qemu` can be in the input string
-    const linkTags = input.match(/{(config|ref|config:option)}`[a-z-:.]+`/g);
+    const linkTags = input.match(/{(config|ref|config:option)}`[a-z-:._]+`/g);
```
mas-who, line 42: I wonder if there are other edge cases coming from sphinx

src/pages/storage/forms/StorageVolumeForm.tsx
```diff
@@ -130,6 +135,25 @@ const StorageVolumeForm: FC<Props> = ({ formik, section, setSection }) => {
   const poolDriver =
     pools.find((item) => item.name === formik.values.pool)?.driver ?? "";
 
+  if (!driversWithFilesystemSupport.includes(poolDriver)) {
```
mas-who, line 138:
```suggestion
  const getFormFieldsToReset = () => {
    if (!driversWithFilesystemSupport.includes(poolDriver)) {
      return getFilesystemVolumeFormFields();
    }

    if (poolDriver !== "zfs") {
      return getZfsVolumeFormFields();
    }

    return [];
  };

  const formFieldsToReset =
    getFormFieldsToReset() as (keyof StorageVolumeFormValues)[];

  for (const field of formFieldsToReset) {
    if (formik.values[field] !== undefined) {
      void formik.setFieldValue(field, undefined);
    }
  }
```

src/util/configInheritance.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { getConfigRowMetadata } from "util/configInheritance";
+import { StoragePoolFormValues } from "pages/storage/forms/StoragePoolForm";
+import { StorageVolumeFormValues } from "pages/storage/forms/StorageVolumeForm";
+
+beforeEach(() => {
+  vi.mock("@tanstack/react-query", () => ({
```
mas-who, line 6: where does `vi` come from? I assume vitest?

src/util/config.tsx
edlerd: This is what is being used as tag name in the LXD code. I think we can extend it if new ones arise.

src/util/storagePool.tsx
edlerd: There are plenty of options for [LVM](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_lvm/#storage-pool-configuration) and [ZFS](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_zfs/#storage-pool-configuration), which I'd like to include soonish. So this will extend to include keys for other drivers.

src/types/config.d.ts
edlerd: Types prefixed with `Lxd` usually relate to things exposed by the LXD api. So this represents what the api has to offer. I think it's okay to keep it here even though we don't use it yet.

src/pages/storage/forms/StorageVolumeForm.tsx
edlerd: Good idea!

src/util/configInheritance.spec.ts
edlerd: Yes, this is from vitest and is quite nice. One oddity is you can't reference any variables in the mock code, because of hoisting. Need to remember that.

edlerd:
> * Snapshot name pattern config has weird description. I am aware there is a discussion with the core team about this, maybe we can manually remove this for now and revert the change when the backend sends more appropriate config descriptions?

I added some lines to remove the placeholder.

> * Interesting that zfs.block_size config has a link for zfs.block_mode in its help text but the actual help text for zfs.block_mode config doesn't have a link. Not sure if we can do anything about this one though.

The doc has a section for all keys, but most of them are exactly as presented with the help texts we use now. I think cross-referencing makes sense here. The link is coming from the API, so we'd need to change it in LXD itself if we wanted to remove it.

src/util/storagePool.tsx
mas-who: aah okay, I missed those, that's perfect then :+1:

src/util/configInheritance.spec.ts
mas-who: oh that's a interesting one, good to know!

mas-who (approved):
LGTM thanks for the changes :+1: .

PR #625, opened 2024-01-30
feat: upgrade to xterm v5 and remove xterm-for-react dependency [WD-8311]

mas-who:
## Done

- Created a `Xterm` component which directly depends on `xterm.js`. This enabled the removal of `xterm-for-react` dependency and we can now upgrade `xterm` independently. The new Xterm component is a hook based implementation inspired by the [xterm-for-react](https://github.com/robert-harbison/xterm-for-react/blob/master/src/XTerm.tsx) implementation.
- Cleaned up `InstanceTerminal` and `InstanceTextConsole` components, removing unnecessary `useEffect` and `useLayoutEffect` calls.

## QA

1. Run the LXD-UI:
    - Visit the details page for a running instance, test the Terminal and Console tabs and make sure the terminal work as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,8 @@
     "react-useportal": "1.0.19",
     "serve": "14.2.1",
     "vanilla-framework": "4.6.0",
-    "xterm-addon-fit": "0.6.0",
-    "xterm-for-react": "1.0.4",
+    "xterm": "5.2.1",
+    "xterm-addon-fit": "0.8.0",
     "yup": "1.3.3"
   },
   "devDependencies": {
--- a/src/components/Xterm.tsx
+++ b/src/components/Xterm.tsx
@@ -1,0 +1,213 @@
+/**
+MIT License
+
+Copyright (c) 2020 Robert Harbison
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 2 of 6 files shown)


edlerd (commented):
The terminal is not responsive any more. When I open a session with a maximized browser window and resize it to a smaller size, the terminal should adjust automatically.

Important is, that it also sends the new size through the websocket control channel, so apps like vim or top know the user's terminal columns and rows. Though currently it doesn't adjust height/width of the container anymore -- which is on a higher level.

mas-who:
> The terminal is not responsive any more. When I open a session with a maximized browser window and resize it to a smaller size, the terminal should adjust automatically.
> 
> Important is, that it also sends the new size through the websocket control channel, so apps like vim or top know the user's terminal columns and rows. Though currently it doesn't adjust height/width of the container anymore -- which is on a higher level.

Aaah right, I mistakenly removed the event listeners for resize. Will add that back and it should resolve the issue.

src/sass/_xterm.scss
mas-who, line 1: This is a very strange one, when the terminal is rendered and opened, it causes an overflow on the `body` element but not the app root. Not an ideal fix but I can't seem to find the element that is overflowing

edlerd: We shouldn't do this. There is a [closed bug on Xterm.js](https://github.com/xtermjs/xterm.js/issues/4805) about this. But version 5.4 with the fix is not yet released. I think we should use version `5.2.1` to avoid this.

mas-who: thanks for finding this! :+1:

mas-who: tested on `xterm 5.2.1`, it works nicely without the additional styling

edlerd (approved):
LGTM

PR #626, opened 2024-01-30
chore: add test for using instance terminal [WD-8310]

mas-who:
## Done

- Added e2e test for instance terminal usage and warning before navigating away

## QA

1. Run the LXD-UI:
    - Ensure ci passes

### DIFF
--- a/tests/instances.spec.ts
+++ b/tests/instances.spec.ts
@@ -6,6 +6,7 @@ import {
   randomInstanceName,
   renameInstance,
   saveInstance,
+  visitAndStartInstance,
   visitInstance,
 } from "./helpers/instances";
 import {
@@ -23,6 +24,7 @@ import {
   deleteProfile,
   randomProfileName,
 } from "./helpers/profile";
+import { TIMEOUT } from "./helpers/constants";
 
 test("instance create and remove", async ({ page }) => {
   const instance = randomInstanceName();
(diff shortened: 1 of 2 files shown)


tests/instances.spec.ts
```diff
@@ -228,3 +230,21 @@ test("instance yaml edit", async ({ page }) => {
...
   await deleteInstance(page, instance);
 });
+
+test("instance terminal operations", async ({ page }) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  await visitAndStartInstance(page, instance);
+  await page.getByTestId("tab-link-Terminal").click();
+  await expect(page.locator(".xterm-screen")).toBeVisible(TIMEOUT);
+  const terminalInput = page.getByRole("textbox", { name: "Terminal input" });
```
edlerd, line 240: I think we can rely on autofocus here and directly use `page.keyboard.type`. Wdyt?

tests/instances.spec.ts
```diff
@@ -228,3 +230,21 @@ test("instance yaml edit", async ({ page }) => {
...
+  await terminalInput.fill("lsb_release -a");
+  await terminalInput.press("Enter");
+  await expect(page.locator(".xterm-rows")).toContainText("jammy");
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  page.on("dialog", async (dialog) => {
+    expect(dialog.type()).toEqual("beforeunload");
+    await dialog.accept();
+  });
```
edlerd, line 248: Can you extend this a bit? I.e. assert, the dialogue is displayed on navigation to another tab. This might also allow for the removal of the eslint  ignore.

tests/instances.spec.ts
```diff
@@ -228,3 +230,21 @@ test("instance yaml edit", async ({ page }) => {
...
+test("instance terminal operations", async ({ page }) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  await visitAndStartInstance(page, instance);
+  await page.getByTestId("tab-link-Terminal").click();
+  await expect(page.locator(".xterm-screen")).toBeVisible(TIMEOUT);
+  const terminalInput = page.getByRole("textbox", { name: "Terminal input" });
+  await terminalInput.fill("lsb_release -a");
+  await terminalInput.press("Enter");
+  await expect(page.locator(".xterm-rows")).toContainText("jammy");
```
edlerd, line 243: This is going to change for sure. Maybe check for the text "Distributor ID:", or something else that is not expected to change?

tests/instances.spec.ts
mas-who: that's a good point, we would test the focus logic as well then :+1:

tests/instances.spec.ts
mas-who: Changed the text check to `Ubuntu`. Although in `createInstance` we always select a `jammy` image, will we update that to the next lts version released?

edlerd (approved):
LGTM :+1:

tests/instances.spec.ts
mas-who: Yeah I think for this specific test using Ubuntu should be fine :+1:

PR #634, opened 2024-02-05
Add test coverage

edlerd:
## Done

- adding coverage for test runs

Fixes WD-8680

## QA

1. Perform the following QA steps:
    - ensure CI pipeline is passing
    - check the CI step emitted test coverage step i.e. [this text report](https://github.com/canonical/lxd-ui/actions/runs/7972445898/job/21764219744#step:13:15) or the [artifact on this summary](https://github.com/canonical/lxd-ui/actions/runs/7972445898)
    - push to gh-pages tested in [this fork action](https://github.com/edlerd/lxd-ui/actions/runs/7974396525) and verified it on [the forks gh-pages branch](https://github.com/edlerd/lxd-ui/tree/gh-pages/coverage)


.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,133 @@
...
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  e2e-coverage:
+    name: e2e-test-coverage
+    runs-on: ubuntu-latest
+    outputs:
```
mas-who, line 11: just realised this is actually not needed really, should remove from `pr.yml` as well, can we do it as part of this PR?

edlerd: I can remove it here, would keep the other workflow change dedicated not to mix too many things.

playwright.config.ts
```diff
@@ -8,22 +8,22 @@ import { TestOptions } from "./tests/fixtures/lxd-test";
...
   /* Run tests in files in parallel */
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : 3,
+  /* Opt out of parallel tests. */
+  workers: 1,
```
mas-who, line 26: we don't want to have a bit of a speed boost for local testing? or this parallel workers break the coverage test?

vitest.config.ts
```diff
@@ -4,9 +4,16 @@ import viteConfig from "./vite.config";
 export default mergeConfig(
   viteConfig,
   defineConfig({
+    build: {
+      sourcemap: "inline",
```
mas-who, line 8: is this set because separately generated sourcemap files interferes with coverage report generation?

package.json
```diff
@@ -21,7 +21,11 @@
     "serve": "./entrypoint",
     "test-js": "vitest --run",
     "test-e2e-edge": "npx playwright test --project chromium:lxd-latest-edge firefox:lxd-latest-edge",
-    "test-e2e-stable": "npx playwright test --project chromium:lxd-5.0-stable firefox:lxd-5.0-stable"
+    "test-e2e-stable": "npx playwright test --project chromium:lxd-5.0-stable firefox:lxd-5.0-stable",
+    "test-js-coverage": "vitest --run --coverage",
+    "test-e2e-coverage": "yarn test-e2e-coverage-playwright ; yarn test-e2e-coverage-report",
+    "test-e2e-coverage-playwright": "rm -rf coverage/playwright* ;  PW_TEST_HTML_REPORT_OPEN='never' npx playwright test --project coverage",
+    "test-e2e-coverage-report": "nyc report --reporter html --reporter cobertura --reporter text --reporter text-summary --temp-dir coverage/playwright --report-dir coverage/playwright-report"
```
mas-who, line 28: why so many reporters? Is it for html from playwright and the rest is for vitest?

edlerd: The reporters are quite cheap if all are in one nyc run. As for why so many:
- html for us to look at on github pages
- xml for an external dashboard to import
- text and text-summary to have the output available directly on the github action

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,131 @@
...
+          run: |
+            git config --global user.email "github-action@example.com"
+            git config --global user.name "GitHub Action"
+        - name: Push coverage Report
+          timeout-minutes: 3
+          run: |
+            git add .
+            git commit -m "workflow: update coverage report"
+            
+            # In case of another action job pushing to gh-pages while we are rebasing for the current job
```
mas-who, line 114: not sure if this check is necessary or is the right approach in this scenario. For normal e2e reports, they are stored in folders on `gh-pages` branch separated by their origin PR number as well as workflow run id. In that case, we don't want to completely stop a workflow run just because another workflow run is busy with the `gh-pages` branch. In the case of the coverage report, I think what we need to ensure is that a later workflow run does not finish earlier than prior workflow runs (e.g. two PRs closed in short succession). Not sure how this can be done though since the push event trigger completely separates the workflow run context...

tests/fixtures/coverage.ts
```diff
@@ -0,0 +1,75 @@
...
+      source,
+      sourceMap,
+    });
+    await converter.load();
+    converter.applyCoverage(entry.functions);
+    const istanbulCoverage = converter.toIstanbul() as CoverageMapData & {
+      [key: string]: { _coverageSchema: string };
+    };
+
+    const generateUUID = (): string => {
```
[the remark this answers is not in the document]
edlerd: Ah right, I put it under util/helpers to be reused.

.github/workflows/coverage.yaml
edlerd: This solves the problem of other actions pushing to gh-pages, that could be another PR test report.

There might be two merges in timely proximitry. If the 1st coverage is pushed after the 2nd coverage, we'd get a report that is slightly outdated. Though not a big issue I think as the next merge will trigger another new coverage report.

tests/instance-panel.spec.ts
```diff
@@ -1,45 +1,50 @@
...
+import { finishCoverage, startCoverage } from "./fixtures/coverage";
 
 let instance = randomInstanceName();
-let page: Page;
-test.beforeAll(async ({ browserName, browser }) => {
+
+test.beforeAll(async ({ browserName, browser, hasCoverage }) => {
   instance = `${browserName}-${instance}`;
-  page = await browser.newPage();
+  const page = await browser.newPage();
```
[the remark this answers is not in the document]
edlerd: We could create a page object in beforeAll and close it in afterAll, then startCoverage/finishCoverage on the edge of both functions.
Though I didn't like the inconsistency in some tests receiving the page object from a local reference in the file and other would receive it from the test function. So I opted for this approach instead. Wdyt?

.github/workflows/coverage.yaml
mas-who: Okay yeah I think let's leave it as is, slightly outdated coverage reports in this case wouldn't be a major issue in any case :+1:

tests/instance-panel.spec.ts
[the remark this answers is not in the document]
mas-who: yeah I think that's maybe the best approach, keeping it consistent and also not generating new pages :+1:

mas-who:
QA looks good https://mas-who.github.io/lxd-ui/coverage/

package.json
```diff
@@ -21,7 +21,11 @@
     "serve": "./entrypoint",
     "test-js": "vitest --run",
     "test-e2e-edge": "npx playwright test --project chromium:lxd-latest-edge firefox:lxd-latest-edge",
-    "test-e2e-stable": "npx playwright test --project chromium:lxd-5.0-stable firefox:lxd-5.0-stable"
+    "test-e2e-stable": "npx playwright test --project chromium:lxd-5.0-stable firefox:lxd-5.0-stable",
+    "test-js-coverage": "vitest --run --coverage",
```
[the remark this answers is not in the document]
edlerd: I added the merging of both reports just now.

mas-who (approved):

PR #639, opened 2024-02-08
Make code snippet in `ErrorPage` scrollable, keeping rest of the page fixed

lorumic:
## Done

- Improved responsiveness of the `ErrorPage` component by making the code snippet in it scrollable, and keeping everything else fixed.

## QA

1. Run the LXD-UI:
    - ~On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @lorumic or @edlerd for access.~ **Not applicable** - you need to craft an error in the code to see the changes in action.
    - Craft an error in the body of any main-content FC (e.g. `InstanceList`) or of the `Navigation` FC, and check that the error boundary works correctly. Example code to achieve this: `throw new Error("foo");`
    - Resize the window height and check that the code snippet containing the stack trace can be properly scrolled.

### DIFF
--- a/src/components/ErrorPage.tsx
+++ b/src/components/ErrorPage.tsx
@@ -1,10 +1,12 @@
-import React, { FC } from "react";
+import React, { FC, useEffect } from "react";
 import {
   CodeSnippet,
   CodeSnippetBlockAppearance,
   Notification,
   Strip,
 } from "@canonical/react-components";
+import { updateMaxHeight } from "util/updateMaxHeight";
+import useEventListener from "@use-it/event-listener";
 
 type Props = {
   error?: Error;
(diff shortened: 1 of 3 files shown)


src/components/ErrorPage.tsx
```diff
@@ -1,30 +1,39 @@
...
+import useEventListener from "@use-it/event-listener";
 
 type Props = {
   error?: Error;
 };
 
 const ErrorPage: FC<Props> = ({ error }) => {
   const body = encodeURIComponent(
-    `\`\`\`\n${error?.stack ?? "No stack track"}\n\`\`\``,
+    `\`\`\`\n${error?.stack ?? "No stack trace"}\n\`\`\``,
```
mas-who, line 17: good catch on the spelling error :)

PR #641, opened 2024-02-09
chore: speed up e2e tests [WD-8560]

mas-who:
## Done

- Adjusted test suites to reuse lxd resources across tests instead of creating new ones for every test.
- Use matrix strategy to run browser specific e2e tests. This splits up the test workload across two runners and essentially should halve the test run time.

## QA

1. Run the LXD-UI:
    - Ensure ci passes and test speed improves

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -62,11 +62,13 @@ jobs:
           reporter: github-pr-review
           fail-on-error: true
 
-  e2e:
+  browser-e2e-test:
     name: e2e-tests
     runs-on: ubuntu-latest
-    env:
-      HTML_REPORT_URL_PATH: reports/pr-${{ github.event.number }}/${{ github.run_id }}/${{ github.run_attempt }}
+    strategy:
+      fail-fast: false
+      matrix:
+        browser: ["chromium", "firefox"]
     outputs:
       job_status: ${{job.status}}
     steps:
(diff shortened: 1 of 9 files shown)


tests/instances.spec.ts
```diff
@@ -188,31 +168,45 @@ test("instance edit cloud init configuration", async ({ page }) => {
   await assertCode(page, "Network config", "foo:");
   await assertCode(page, "User data", "bar:");
   await assertCode(page, "Vendor data", "baz:");
+});
 
+test("instance terminal operations", async () => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
```
lorumic, line 175: Not reusing our browser-name-based `instance` here?

tests/instances.spec.ts
```diff
@@ -188,31 +168,45 @@ test("instance edit cloud init configuration", async ({ page }) => {
...
+    dialogPresent = true;
+    void dialog.accept();
+  });
+  await page.getByTestId("tab-link-Overview").click();
+  expect(dialogPresent).toEqual(true);
   await deleteInstance(page, instance);
 });
 
-test("instance create vm", async ({ page }) => {
+const vmInstance = randomInstanceName();
```
lorumic, line 193: Same as above - also not sure why this one is out of the test function scope. It seems to me that it doesn't get deleted also.

tests/networks.spec.ts
```diff
@@ -14,16 +14,20 @@ import {
   setOption,
 } from "./helpers/configuration";
 
-test("network create and remove", async ({ page }) => {
-  const network = randomNetworkName();
+let network = randomNetworkName();
+let page: Page;
+test.beforeAll(async ({ browser, browserName }) => {
+  network = `${browserName.substring(0, 2)}-${network}`;
```
lorumic, line 20: Why are we using only a substring of the browser name here, but the full string for instances, profiles and storage volumes? Is that because of some stricter limit on the length of the network name?

tests/notification.spec.ts
```diff
@@ -13,32 +15,36 @@ import {
...
+  await visitAndStopInstance(page, instance);
   await dismissNotification(page);
 });
 
-test("auto hide notification after a timeout", async ({ page }) => {
-  const instance = randomInstanceName();
-  await createInstance(page, instance);
+test("auto hide notification after a timeout", async () => {
+  await visitAndStartInstance(page, instance);
   await page.waitForTimeout(5000);
```
lorumic, line 43: If we export the `HIDE_NOTIFICATION_DELAY` const in `toastNotificationProvider.tsx`, we can directly reference that here, so that we track any changes on that value:
```suggestion
  await page.waitForTimeout(HIDE_NOTIFICATION_DELAY);
```

tests/profiles.spec.ts
```diff
@@ -19,39 +19,36 @@ import {
...
+  await deleteProfile(page, profile);
+  await page.close();
+});
 
+test("profile rename", async () => {
   const newName = profile + "-rename";
   await renameProfile(page, profile, newName);
-
-  await deleteProfile(page, newName);
+  profile = newName;
```
lorumic, line 38: Interesting approach, would it work for instances as well? So that we don't have to rename it twice - which is what is currently being done, see lines 48-49 of [instances.spec.ts](https://github.com/canonical/lxd-ui/pull/641/files#diff-53a4f34ae6b1ee87d7500934ca7b7e8daa3c3f826106db78fa47373904db7045)

tests/storage.spec.ts
```diff
@@ -18,7 +18,20 @@ import { activateOverride, setInput } from "./helpers/configuration";
...
+  await createVolume(page, volume);
+});
+
+test.afterAll(async () => {
+  await deleteVolume(page, volume);
+  await page.close();
+});
+
+test("storage pool create, edit and remove", async () => {
   const pool = randomPoolName();
```
lorumic, line 35: No browser-dependent name for the storage pool?

.github/workflows/pr.yaml
```diff
@@ -128,31 +130,69 @@ jobs:
...
+      - name: Upload ${{ matrix.browser }} blob reports to be merged
         if: always()
+        uses: actions/upload-artifact@v4
         with:
-          name: playwright-report
-          path: playwright-report/
-          retention-days: 30
+          name: blob-report-${{ matrix.browser }}
+          path: blob-report
+          retention-days: 1
```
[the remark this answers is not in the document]
mas-who: This is for retaining the blob data which is then used to merge into the final html report. We don't really want to keep this data for long since it serves no purpose after it is merged. The final reports are retained for 14 days (2 weeks) which I think is reasonable, previously the retention period of 30 days seems really long so I adjusted it.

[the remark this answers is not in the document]
lorumic: I see, thanks for the explanation. Makes sense :+1:

tests/instances.spec.ts
mas-who: running terminal operations on an existing instance created outside of this test consistently fails for some reason. I think it may be related to configurations set in another test that comes before this one but couldn't nail it down to what it was. So I left the instance creation inside of this test instead.

lorumic: > I think it may be related to configurations set in another test that comes before this one

If that's the case, it should work if this test is done before any other? Have you tried that?

mas-who: I have run the test by itself and it passed, but I didn't try to move it before other tests and run it like that, I will try that quick.

mas-who: Okay moving it to the beginning of instance tests allows it to pass, but in addition the instance needs to be stopped for the subsequent tests to pass so I've added that as well.

tests/instances.spec.ts
mas-who: Sorry my bad, I was experimenting with this one and forgot to move the `vmInstance` creation into the `beforeAll` block. Made the change now.

tests/networks.spec.ts
mas-who: yup, network names can only be 15 characters long. I chose to substring the browser name instead of the network name since `randomNetworkName` uses `randomNameSuffix` that's also used by many other tests e.g. storage, instance etc. I think best I leave a comment here for context, I see why this is confusing

tests/notification.spec.ts
mas-who: great idea, thanks!

tests/profiles.spec.ts
mas-who: Yeah it would, I think I forgot to apply the same approach to instances after doing it for the other tests.

tests/storage.spec.ts
mas-who: since the pool creation is scoped to a single test, we don't need browser specific name for it. The browser specific naming is really a safety net for parallel tests potentially editing the same resource causing a race condition, in this case that won't happen

tests/notification.spec.ts
mas-who: Had to revert this change unfortunately, during build time for tests we don't have scope of `react-components` module which is imported in the same file where the `HIDE_NOTIFICATION_DELAY` live. This causes a module not found error for any react components we import in `toastNotificationProvider`

lorumic: Wouldn't moving the const to a separate util file solve this? Then you would import it in both places (the test file and the `toastNotificationProvider`. Does it sound reasonable and likely to work?

mas-who: I thought about putting it in `util/notifications.tsx` but that file also imports from react-components :sweat: didn't think creating another new utils file was worth it.

src/context/toastNotificationProvider.tsx
```diff
@@ -19,7 +19,7 @@ import React, {
   useState,
 } from "react";
 
-const HIDE_NOTIFICATION_DELAY = 5_000;
+export const HIDE_NOTIFICATION_DELAY = 5_000;
```
lorumic, line 22: I think this can be reverted as well at this point.

lorumic (approved):
Thanks for applying the changes, one minor comment about one change that could also be reverted.

src/context/toastNotificationProvider.tsx
mas-who: aah yes, will revert that :+1:

PR #642, opened 2024-02-12
chore: fix app wide accessibility issues

mas-who:
## Done

- disable link in logo for small screen (reason: same adjacent link URL)
- changed title attribute so it's not identical to text content (InstanceList)
- TablePagination aria-label
- associate orphaned form label (InstanceCreateDetailsForm)
- add text to hidden input
- Resolve multiple label for the same form control (SnapshotScheduleInput)
- Remove uneccessary element titles or adjust for better meaning
- Adjusted heading elements to not skip heirarchical order
- Added aria-label for Select components that's missing labels

In addition, here is the [Jira ticket](https://warthogs.atlassian.net/browse/WD-8559) that documents all open a11y issues.

## QA

1. Run the LXD-UI:
    - Download Wave evaluation tool, check the application and see if there are any errors/alerts present on different pages and not documented as an open issue on this `[Jira ticket](https://warthogs.atlassian.net/browse/WD-8559)`.


src/components/RenameHeader.tsx
```diff
@@ -97,7 +97,6 @@ const RenameHeader: FC<Props> = ({
                 <li
                   className="p-breadcrumbs__item name u-truncate"
                   onClick={enableRename}
-                  title={name}
```
edlerd, line 100: We use the title attribute frequently, because if the content is very long (i.e. a very long instance name), the text will be truncated. Adding the title allows hovering to get a tooltip with the full, non-truncated name of the resource.

src/components/SnapshotScheduleInput.tsx
```diff
@@ -3,27 +3,27 @@ import { Input, RadioInput, Select } from "@canonical/react-components";
 
 const snapshotOptions = [
   {
-    label: "minute",
+    label: "Every minute",
```
edlerd, line 6: This looks a bit odd, having one option starting with uppercase and the others all lowercase. Shall we also write "Hourly", "Daily", etc?

src/pages/instances/InstanceList.tsx
```diff
@@ -330,7 +330,10 @@ const InstanceList: FC = () => {
         columns: [
           {
             content: (
-              <div className="u-truncate" title={instance.name}>
+              <div
+                className="u-truncate"
+                title={`Link for instance ${instance.name}`}
```
edlerd, line 335: Maybe just "Instance foo"?

src/pages/instances/InstanceIps.tsx
```diff
@@ -13,11 +13,7 @@ const InstanceIps: FC<Props> = ({ instance, family }) => {
   return addresses.length ? (
     <ExpandableList
       items={addresses.map((item) => (
-        <div
-          key={item.address}
-          className="ip u-truncate"
-          title={`${item.address} (${item.iface})`}
-        >
+        <div key={item.address} className="ip u-truncate">
```
edlerd, line 16: I think we should keep the title, i.e. when truncated. See right-hand side in the screen below

image

src/pages/instances/InstanceOverviewNetworks.tsx
```diff
@@ -64,7 +64,7 @@ const InstanceOverviewNetworks: FC<Props> = ({ instance, onFailure }) => {
             content: (
               <Link
                 to={`/ui/project/${instance.project}/networks/detail/${network.name}`}
-                title={network.name}
+                title={`link to ${network.name}`}
```
edlerd, line 67: I think we should avoid putting semantics like "this is a link" into captions. Could say "Network foo" though.

src/pages/instances/InstanceOverviewProfiles.tsx
```diff
@@ -49,7 +49,7 @@ const InstanceOverviewProfiles: FC<Props> = ({ instance, onFailure }) => {
           content: (
             <Link
               to={`/ui/project/${instance.project}/profiles/detail/${profile}`}
-              title={profile}
+              title={`link to ${profile}`}
```
edlerd, line 52: "Profile foo"?

src/pages/instances/InstanceSnapshots.tsx
```diff
@@ -128,7 +128,7 @@ const InstanceSnapshots = (props: Props) => {
         {
           content: (
             <>
-              <div className="u-truncate" title={snapshot.name}>
+              <div className="u-truncate">
```
edlerd, line 131: Should probably keep the title

src/pages/instances/forms/InstanceCreateDetailsForm.tsx
```diff
@@ -103,9 +103,7 @@ const InstanceCreateDetailsForm: FC<Props> = ({
       </Row>
       <Row>
         <Col size={12}>
-          <label className="p-form__label" htmlFor="base-image">
-            Base Image*
-          </label>
+          <p className="p-form__label">Base Image*</p>
```
edlerd, line 106: Now we can't click the label any more to trigger the modal. Shall we keep the reference and label type?

I think the problem is after the image is selected. In that special case we could fall back to a generic paragraph and avoid the label.

src/pages/networks/actions/DeleteNetworkBtn.tsx
```diff
@@ -49,7 +49,7 @@ const DeleteNetworkBtn: FC<Props> = ({ network, project }) => {
           ? "Can not delete, network is not managed"
           : isUsed
             ? "Can not delete, network is currently in use"
-            : "Delete network"
+            : "This network can be deleted"
```
edlerd, line 52: I think this can just be empty. No need to highlight, that we can delete it, when the button is active.

src/pages/networks/forms/NetworkForwardFormPorts.tsx
```diff
@@ -154,6 +150,7 @@ const NetworkForwardFormPorts: FC<Props> = ({ formik, network }) => {
                   hasIcon
                   className="u-no-margin--bottom"
                   type="button"
+                  aria-label="Delete port button"
```
edlerd, line 153: "Delete port $index"?

src/pages/profiles/ProfileList.tsx
```diff
@@ -114,7 +114,10 @@ const ProfileList: FC = () => {
       columns: [
         {
           content: (
-            <div className="u-truncate" title={profile.name}>
+            <div
+              className="u-truncate"
+              title={`link to profile ${profile.name}`}
```
edlerd, line 119: Here we have a list of profiles, so just the "profile.name" is probably the right thing to use?

src/pages/profiles/ProfileList.tsx
```diff
@@ -126,9 +129,7 @@ const ProfileList: FC = () => {
         },
         {
           content: (
-            <div className="profile-description" title={profile.description}>
-              {profile.description}
-            </div>
+            <div className="profile-description">{profile.description}</div>
```
edlerd, line 132: should keep the title for truncated case.

src/pages/storage/StorageVolumeNameLink.tsx
```diff
@@ -42,7 +42,7 @@ const StorageVolumeNameLink: FC<Props> = ({
     <div className={classnames("u-flex", className)}>
       <div
         className={classnames("u-truncate", "volume-name-link")}
-        title={volume.name}
+        title={`link to volume ${volume.name}`}
```
edlerd, line 45: just "volume.name"?

src/pages/storage/CustomVolumeSelectModal.tsx
```diff
@@ -63,7 +63,10 @@ const CustomVolumeSelectModal: FC<Props> = ({
             columns: [
               {
                 content: (
-                  <div className="u-truncate iso-name" title={volume.name}>
+                  <div
+                    className="u-truncate iso-name"
+                    title={`volume: ${volume.name}`}
```
edlerd, line 68: just "volume.name"?

src/pages/profiles/ProfileUsedByDefaultProject.tsx
```diff
@@ -22,9 +22,7 @@ const ProfileUsedByDefaultProject: FC<Props> = ({
         <tr key={project.name} className="instances-by-project list-wrapper">
           <th className={headingClassName}>
             <div className="flexible-container">
-              <div title={project.name} className="u-truncate">
-                {project.name}
-              </div>
+              <div className="u-truncate">{project.name}</div>
```
edlerd, line 25: keep the title?

src/pages/profiles/ProfileStorageList.tsx
```diff
@@ -15,11 +15,7 @@ const ProfileStorageList: FC<Props> = ({ profile }) => {
           items={Object.values(profile.devices)
             .filter(isDiskDevice)
             .map((device) => (
-              <div
-                key={device.path}
-                className="u-truncate list-item"
-                title={device.pool}
-              >
+              <div key={device.path} className="u-truncate list-item">
```
edlerd, line 18: keep the title?

src/pages/profiles/ProfileNetworkList.tsx
```diff
@@ -15,11 +15,7 @@ const ProfileNetworkList: FC<Props> = ({ profile }) => {
           items={Object.values(profile.devices)
             .filter(isNicDevice)
             .map((device) => (
-              <div
-                key={device.network}
-                className="u-truncate list-item"
-                title={device.network}
-              >
+              <div key={device.network} className="u-truncate list-item">
```
edlerd, line 18: keep the title?

mas-who:
Thanks for the review @edlerd . I see there are quite a few `title` attribute related comments. The related a11y issue is that `title` attribute should not have identical text as the element text content, and it should provide advisory information. Maybe to keep the hover effect, we can keep the `title` attribute through out but just add relevant prefixes e.g. `title="Profile profile-name"`?

src/components/RenameHeader.tsx
mas-who: As discussed, will keep all `title` attributes and adjust them to have slightly more descriptive text than the element text content.

src/components/Navigation.tsx
```diff
@@ -44,7 +44,7 @@ const Navigation: FC = () => {
       <header className="l-navigation-bar">
         <div className="p-panel is-dark">
           <div className="p-panel__header">
-            <Logo />
+            <Logo disableLink />
```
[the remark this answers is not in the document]
mas-who: This was a solution to the Wave alert for redundant links since we have two identical links for mobile and web view respectively. I did some more reading on this issue, I don't think it's a WCAG failure but just a Wave alert and it should be acceptable as this is part of the design. I think I will revert this change.

src/components/Logo.tsx
```diff
@@ -13,7 +17,7 @@ const Logo: FC = () => {
   };
 
   return (
-    <NavLink className="p-panel__logo" to={getLogoLink()}>
+    <NavLink className="p-panel__logo" to={disableLink ? "" : getLogoLink()}>
```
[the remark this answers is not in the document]
mas-who: Yes, it is related to the previous comment about the logo link approach, probably should just revert this change.

src/components/SnapshotScheduleInput.tsx
mas-who: This change resolves the issue of multiple labels pointing at the same form control since we have the configuration title that points at the `Input` or `Select` components but at the same time these components have their own labels. I'm sure you are already aware of this, just leaving the above comment here for future reference if we ever look back at the PR and try to figure out why this was done :slightly_smiling_face: 

I agree with the above comment, it looks a bit weird only one option starts with uppercase, will change the other options to look the same.

src/pages/instances/forms/InstanceCreateDetailsForm.tsx
mas-who: This is another Wave alert for Orphaned labels, the `for` attribute should point at `select-image` and not `base-image` but even with that change the alert remains because labelling buttons should be avoided according to [this MDN doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#buttons). I think the reason is that html buttons already have a11y built into it if text content is present within the button, having additional accessible labels will make it noisy for keyboard or screen reader users. For example, if you start your keyboard focus at the Instance name input, then you press tab to increment the form control focus, when you get to the Browse images button the focus will first go to the label then the button which seems to be an unnecessary step (all the other inputs will focus on the control only).

That being said, having the label clickable does make behaviour consistent and potentially benefits normal users, maybe  we don't need to go this far for a11y as it is a Wave alert and not WCAG failure, so I can revert this change if you think it's appropriate?

src/pages/networks/actions/DeleteNetworkBtn.tsx
mas-who: Nice! so empty string for the title attribute won't trigger hover text, thanks for the tip!

src/pages/instances/forms/InstanceCreateDetailsForm.tsx
edlerd: You are right, the button should not need a label for it. Then I guess the change is legit as it is?

mas-who: Yeah I am in favour of not having a label for buttons :slightly_smiling_face:

src/sass/_disk_device_form.scss
```diff
@@ -6,6 +6,11 @@
   .configuration {
     padding-left: 0;
     width: 8rem;
+
+    p {
```
edlerd, line 10: Probably better to target a custom class here, instead of all paragraphs?

edlerd (commented):
This is great, thanks for applying the a11y fixes!

QA is good. Code wise, just one comment about CSS below.

src/sass/_disk_device_form.scss
mas-who: true, I'm going to use vanilla instead actually

edlerd (approved):
All good, thanks.

PR #643, opened 2024-02-13
fix: resolve playwright report artifact upload conflict

mas-who:
## Done

- Removed workflow step that uploads the merged playwright reports for the second time

## QA

1. Run the LXD-UI:
    - Ensure CI passes and that the playwright report is visible at the URL produced by the `merge-e2e-report` job

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -13,7 +13,7 @@ const config: PlaywrightTestConfig = {
      * Maximum time expect() should wait for the condition to be met.
      * For example in `await expect(locator).toHaveText();`
      */
-    timeout: 5 * 1000,
+    timeout: 120 * 1000,
   },
   /* Run tests in files in parallel */
   fullyParallel: true,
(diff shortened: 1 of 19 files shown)


edlerd (approved):
LGTM, thanks for fixing

playwright.config.ts
```diff
@@ -24,7 +24,7 @@ const config: PlaywrightTestConfig = {
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : 3,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: process.env.CI ? "blob" : "html",
+  reporter: "blob",
```
edlerd, line 27: I think we should keep this, so local runs produce an html output

edlerd (changes requested):

edlerd (approved):
All good now, thanks for fixing it quickly!

PR #645, opened 2024-02-13
Simplify loading state and avoid races on start/stopping of instances

edlerd:
## Done

- Simplify loading state and avoid races on start/stopping of instances

## QA

1. Run the LXD-UI:
    - manually start/stop a bunch of instances quickly after one another (not necessarily bulk actions) to ensure no instance is left in starting/stopping state. This was the broken behaviour before.

### DIFF
--- a/src/context/instanceLoading.tsx
+++ b/src/context/instanceLoading.tsx
@@ -31,15 +31,19 @@ export const InstanceLoadingProvider: FC<Props> = ({ children }) => {
   );
 
   const setLoading = (instance: LxdInstance, loadingType: LoadingTypes) => {
-    const newMap = new Map(instanceStates);
-    newMap.set(instance.name, loadingType);
-    setInstanceStates(newMap);
+    setInstanceStates((oldMap) => {
+      const newMap = new Map(oldMap);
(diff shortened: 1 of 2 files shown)


src/context/instanceLoading.tsx
```diff
@@ -26,20 +20,14 @@ interface Props {
 }
 
 export const InstanceLoadingProvider: FC<Props> = ({ children }) => {
-  const [instanceStates, setInstanceStates] = useState(
-    new Map<string, LoadingTypes>(),
-  );
+  const instanceStates = new Map<string, LoadingTypes>();
```
mas-who, line 23: Does context trigger re-renders if we remove state here? i.e. does changes in the Map object trigger re-renders in children components subscribed to this context?

edlerd: Good point, this wasn't working properly when going to detail page / list page and back while a state change is in progress. Switched to a different approach just now.

mas-who: aah nice, by using the callback for `setState` calls you are guaranteed to use the correct previous state, nice! :+1:

mas-who (approved):

PR #646, opened 2024-02-13
fix: redirect back to specific form section on creation error [WD-8681]

mas-who:
## Done

- If a custom yaml error occurs on instance creation, the error notification will now redirect to the yaml config page with previously set values.

## QA

1. Run the LXD-UI:
    - Go to the instance creation form, then navigate to the YAML configuration section and enter some nonsense. Submit the form by clicking the create/create and start button. This should result in an error and a notification should appear at the top of the page. Click on `Check configuration` within the error notification, this should redirect you back to the YAML configuration section within the instance creation form. You should also see the erroneous values you have previously entered in the YAML editor.

## Screenshots

### DIFF
--- a/src/pages/instances/CreateInstance.tsx
+++ b/src/pages/instances/CreateInstance.tsx
@@ -87,6 +87,7 @@ interface PresetFormState {
   retryFormValues?: CreateInstanceFormValues;
   selectedImage?: RemoteImage;
   cancelLocation?: string;
+  retryFormSection?: string;
 }
 
 const CreateInstance: FC = () => {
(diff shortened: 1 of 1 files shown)


src/pages/instances/CreateInstance.tsx
```diff
@@ -259,14 +263,20 @@ const CreateInstance: FC = () => {
           operation.metadata.id,
           () => creationCompletedHandler(instanceName, shouldStart, isIsoImage),
           (msg) =>
-            notifyCreationFailed(new Error(msg), formUrl, values, "toast"),
+            notifyCreationFailed(
```
mas-who, line 266: I made this a bit more generic, if any error occurs the error notification would redirect to the relevant instance creation section. I think this is better than just redirecting back to the relevant section specifically for YAML errors

edlerd (approved):
Works well, thanks for fixing this!

PR #647, opened 2024-02-14
feat: provide UI backward compatibility for older lxd versions [WD-8671]

mas-who:
## Done

- Hide iso import functionality for storage volumes if the lxd server does not support the feature. This includes creating an instance based on custom ISO or attaching custom ISO to a running instance.
- Return empty server configs for lxd 5.0 instead of calling the `/1.0/metadata/configuration` endpoint
- Return empty doc string for lxd 5.0 instead of calling the `/documentation` api endpoint
- Remove `features.networks.zones` project config for lxd 5.0
- Remove `features.storage.buckets` project config for lxd 5.0
- Hide warnings page for lxd 5.0

## QA

1. Run the LXD-UI:
    - Downgrade lxd to 5.0/stable first
    - make sure you can't create an instance using a custom ISO, attach a custom ISO to a running instance via the Instance Console and the Custom ISOs tab is not visible on the storage page. 
    - make sure the server settings page does not error out even it's mostly empty and that there is a notification asking users to upgrade to lxd 5.19 or later.
    - check the network tab in the browser inspect tool that on any configuration page, no network requests are made to the `/documentation/objects.inv.txt` endpoint
    - make sure that you can create project successfully, then try edit a project and can save it successfully. On the project details page, main configuration section, you should not see the check boxes for Network zones and storage bucket features.
    - make sure the warnings page is hidden from the side navigation

### DIFF
--- a/src/api/server.tsx
+++ b/src/api/server.tsx
@@ -36,7 +36,13 @@ export const fetchResources = (): Promise<LxdResources> => {
   });
 };
 
-export const fetchConfigOptions = (): Promise<LxdConfigOptions> => {
+export const fetchConfigOptions = (
+  hasMetadataConfiguration: boolean,
+): Promise<LxdConfigOptions | null> => {
+  if (!hasMetadataConfiguration) {
+    return new Promise((resolve) => resolve(null));
+  }
+
   return new Promise((resolve, reject) => {
     fetch("/1.0/metadata/configuration")
       .then(handleResponse)
@@ -45,7 +51,13 @@ export const fetchConfigOptions = (): Promise<LxdConfigOptions> => {
   });
 };
 
-export const fetchDocObjects = (): Promise<string[]> => {
+export const fetchDocObjects = (
+  hasDocumentationObject: boolean,
+): Promise<string[]> => {
+  if (!hasDocumentationObject) {
+    return new Promise((resolve) => resolve([]));
+  }
+
   return new Promise((resolve, reject) => {
     fetch("/documentation/objects.inv.txt")
       .then(handleTextResponse)
--- a/src/context/useSupportedFeatures.tsx
+++ b/src/context/useSupportedFeatures.tsx
@@ -1,0 +1,24 @@
+import { useSettings } from "./useSettings";
+
+export const useSupportedFeatures = () => {
+  const { data: settings, isLoading, error } = useSettings();
+  const apiExtensions = new Set(settings?.api_extensions);
+
+  const serverVersion = settings?.environment?.server_version;
+  const serverMajor = parseInt(serverVersion?.split(".")[0] ?? "0");
+  const serverMinor = parseInt(serverVersion?.split(".")[1] ?? "0");
+
+  return {
+    settings,
+    isSettingsLoading: isLoading,
+    settingsError: error,
+    hasCustomVolumeIso: apiExtensions.has("custom_volume_iso"),
+    hasProjectsNetworksZones: apiExtensions.has("projects_networks_zones"),
+    hasStorageBuckets: apiExtensions.has("storage_buckets"),
+    hasMetadataConfiguration: apiExtensions.has("metadata_configuration"),
+    hasLocalDocumentation:
+      !!serverVersion && serverMajor >= 5 && serverMinor >= 19,
+    hasDocumentationObject:
+      !!serverVersion && serverMajor >= 5 && serverMinor >= 20,
+  };
+};
--- a/src/pages/projects/CreateProject.tsx
+++ b/src/pages/projects/CreateProject.tsx
@@ -41,6 +41,7 @@ import BaseLayout from "components/BaseLayout";
 import FormFooterLayout from "components/forms/FormFooterLayout";
 import { slugify } from "util/slugify";
 import { useToastNotification } from "context/toastNotificationProvider";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
 
 export type ProjectFormValues = ProjectDetailsFormValues &
   ProjectResourceLimitsFormValues &
@@ -56,6 +57,8 @@ const CreateProject: FC = () => {
   const queryClient = useQueryClient();
   const controllerState = useState<AbortController | null>(null);
   const [section, setSection] = useState(slugify(PROJECT_DETAILS));
+  const { hasProjectsNetworksZones, hasStorageBuckets } =
+    useSupportedFeatures();
 
   const ProjectSchema = Yup.object().shape({
     name: Yup.string()
@@ -88,6 +91,14 @@ const CreateProject: FC = () => {
             ...networkRestrictionPayload(values),
           }
         : {};
+
+      if (!hasProjectsNetworksZones) {
+        values.features_networks_zones = undefined;
+      }
+
+      if (!hasStorageBuckets) {
+        values.features_storage_buckets = undefined;
+      }
 
       createProject(
         JSON.stringify({
--- a/src/pages/projects/EditProject.tsx
+++ b/src/pages/projects/EditProject.tsx
@@ -32,6 +32,7 @@ import FormFooterLayout from "components/forms/FormFooterLayout";
 import { useNavigate, useParams } from "react-router-dom";
 import { slugify } from "util/slugify";
 import { useToastNotification } from "context/toastNotificationProvider";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
 
 interface Props {
   project: LxdProject;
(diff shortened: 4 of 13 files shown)


mas-who:
**Things to think about**

1. I wonder if there is a better and more generic approach to manage feature support? Going to think on this one for a bit.
2. Maybe we should start checking from lxd 4.0, then get a full list of features that doesn't work and that should inform which api extension to check for feature support. This is probably a good approach to establish a baseline assuming api extensions only have additions with lxd version increase from the perspective of the UI.

src/context/useSettings.tsx
```diff
@@ -4,9 +4,28 @@ import { fetchSettings } from "api/server";
...
+};
+
+export const useSettings = (): UseQueryResult<TransformedLxdSettings> => {
+  const fetchAndTransformSettings = async () => {
+    const settings = await fetchSettings();
+    const apiExtensions: TransformedLxdSettings["api_extensions"] = {};
+    settings.api_extensions?.forEach((extension) => {
+      apiExtensions[extension] = true;
+    });
+
```
edlerd, line 20: Probably better to keep this as it was and move the transformation to `useApiExtensions`, no?

edlerd:
> 2\. Maybe we should start checking from lxd 4.0, then get a full list of features that doesn't work and that should inform which api extension to check for feature support. This is probably a good approach to establish a baseline assuming api extensions only have additions with lxd version increase from the perspective of the UI.

I am not sure how much 4.0 is still in use. I took this on our sync agenda for tomorrow, we can get more input from the core team on which lowest version to support. Likewise, I was previously under the impression 5.0 is fine.

src/context/useSettings.tsx
mas-who: yeah 100%, also thinking changing `useApiExtensions` to `useSupportedFeatures` for better naming

mas-who:
> > 2. Maybe we should start checking from lxd 4.0, then get a full list of features that doesn't work and that should inform which api extension to check for feature support. This is probably a good approach to establish a baseline assuming api extensions only have additions with lxd version increase from the perspective of the UI.
> 
> I am not sure how much 4.0 is still in use. I took this on our sync agenda for tomorrow, we can get more input from the core team on which lowest version to support. Likewise, I was previously under the impression 5.0 is fine.

Aah great, thanks I was thinking about asking that tomorrow as well! I think if I know what the baseline lxd version is, then I can streamline the process to make this happen

src/api/server.tsx
```diff
@@ -36,7 +37,30 @@ export const fetchResources = (): Promise<LxdResources> => {
   });
 };
 
-export const fetchConfigOptions = (): Promise<LxdConfigOptions> => {
+export const fetchConfigOptions: QueryFunction<LxdConfigOptions> = ({
+  queryKey,
+}) => {
+  const [, metadataConfigurationSupported] = queryKey;
```
mas-who, line 43: having a `useQuery` call in here simply doesn't work, I think it's because this is the fetch function for another `useQuery` call. Instead of having `useSupportedFeatures` called here, I moved it one level above.

src/pages/settings/Settings.tsx
```diff
@@ -166,6 +171,15 @@ const Settings: FC = () => {
       >
         <NotificationRow />
         <Row>
+          {!metadataConfigurationSupport && (
```
mas-who, line 174: any suggestions for better copy?

mas-who:
Will update PR notes in the morning.

src/pages/settings/Settings.tsx
edlerd: Maybe `hasMetadataConfigApi`?

All of the neighbours could be `hasFooApi` instead of `fooSupport`, do you think that is better?

src/context/useSupportedFeatures.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return {
+    settings,
+    isSettingsLoading: isLoading,
+    settingsError: error,
+    customVolumeIsoSupport: apiExtensions.has("custom_volume_iso"),
+    projectsNetworksZonesSupport: apiExtensions.has("projects_networks_zones"),
+    storageBucketsSupport: apiExtensions.has("storage_buckets"),
+    metadataConfigurationSupport: apiExtensions.has("metadata_configuration"),
+    localDocumentationSupport:
+      serverVersion && serverMajor >= 5 && serverMinor >= 19,
```
edlerd, line 21: Good idea to move this over here.

src/pages/settings/Settings.tsx
mas-who: aah sorry if I was unclear, I meant for the notification text content and not the variable name. I tried to align these variable names with the actual api extension name, so maybe we can do `hasMetadataConfiguration` instead and similar pattern for the others? api extension can be related to sub features of an api endpoint so I think the `hasApi` pattern might be a bit ambiguous.

edlerd: Right I agree with variable naming either way.

As for the notification copy, the text seems OK, maybe add "or later" and "server" to the string:

"Update to the LXD v5.19.0 or later to access more server settings"

src/pages/settings/Settings.tsx
```diff
@@ -20,25 +20,30 @@ import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
...
+    hasMetadataConfiguration,
+    settings,
+    isSettingsLoading,
+    settingsError,
+  } = useSupportedFeatures();
 
   const { data: configOptions, isLoading: isConfigOptionsLoading } = useQuery({
-    queryKey: [queryKeys.configOptions],
+    queryKey: [queryKeys.configOptions, hasMetadataConfiguration],
     queryFn: fetchConfigOptions,
```
edlerd, line 38:
```suggestion
    queryKey: [queryKeys.configOptions],
    queryFn: () => fetchConfigOptions(hasMetadataConfiguration),
```
I doubt the `queryKey` is meant to pass information to  the `queryFn`. It hides the dependency quite well. Better pass the flag directly?

edlerd: Passing directly has also the advantage, that you can't forget to pass the flag.

src/pages/settings/Settings.tsx
mas-who: Yeah that sounds good, will add :+1:

src/pages/settings/Settings.tsx
mas-who: Aaah indeed this is a better approach, will change that :+1:

src/api/server.tsx
```diff
@@ -36,7 +36,29 @@ export const fetchResources = (): Promise<LxdResources> => {
...
+          server: {},
+          "storage-btrfs": {},
+          "storage-ceph": {},
+          "storage-cephfs": {},
+          "storage-cephobject": {},
+          "storage-dir": {},
+          "storage-lvm": {},
+          "storage-zfs": {},
+        },
+      }),
```
edlerd, line 58:
```suggestion
      resolve(null),
```
I think we can return null here and change the return type. All callers should be fine with this.

src/pages/projects/EditProject.tsx
```diff
@@ -61,6 +64,8 @@ const EditProject: FC<Props> = ({ project }) => {
     initialValues: initialValues,
     validationSchema: ProjectSchema,
     onSubmit: (values) => {
+      if (!hasProjectsNetworksZones) values.features_networks_zones = undefined;
+      if (!hasStorageBuckets) values.features_storage_buckets = undefined;
```
edlerd, line 68: similar to above comment regarding shorthand if style.

edlerd (commented):
QA looks good, tested with 5.0 and a recent backend.

I found one issue and some small optional comments on code simplifications below.

src/pages/projects/CreateProject.tsx
```diff
@@ -89,6 +92,8 @@ const CreateProject: FC = () => {
           }
         : {};
 
+      if (!hasProjectsNetworksZones) values.features_networks_zones = undefined;
+      if (!hasStorageBuckets) values.features_storage_buckets = undefined;
```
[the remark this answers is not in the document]
mas-who: yeah fair, I have always gone back and forth on this one, let's consistently use brackets rather

src/context/useSupportedFeatures.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return {
+    settings,
+    isSettingsLoading: isLoading,
+    settingsError: error,
+    hasCustomVolumeIso: apiExtensions.has("custom_volume_iso"),
+    hasProjectsNetworksZones: apiExtensions.has("projects_networks_zones"),
+    hasStorageBuckets: apiExtensions.has("storage_buckets"),
+    hasMetadataConfiguration: apiExtensions.has("metadata_configuration"),
+    hasLocalDocumentation:
+      !!serverVersion && serverMajor >= 5 && serverMinor >= 20,
```
edlerd, line 20: I think we misunderstood the local docs version: local docs are available since 5.19

We probably need another signal here, because the objects.inv.txt file is available since 5.20

edlerd (approved):
LGTM, thanks for applying the fixes :)

PR #653, opened 2024-02-16
chore(test): added tests for multiple lxd server versions [WD-8616]

mas-who:
## Done

- Adjusted CI to run e2e tests against multiple lxd server versions
- Added tests for lxd v5.0/stable
- Adjusted existing tests with features that does not exist in  lxd v5.0/stable to be skipped
- Added test fixture to annotate tests with lxd version

## QA

1. Run the LXD-UI:
    - ensure CI passes
    - check that the output playwright report looks fine

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -68,6 +68,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        lxd_channel: ["5.0/stable", "latest/edge"]
         browser: ["chromium", "firefox"]
     outputs:
       job_status: ${{job.status}}
@@ -114,7 +115,7 @@ jobs:
       - name: Install LXD
         uses: canonical/setup-lxd@v0.1.1
         with:
-          channel: latest/edge
+          channel: ${{ matrix.lxd_channel }}
 
       - name: Setup LXD
         shell: bash
@@ -132,17 +133,25 @@ jobs:
       - name: Install Playwright Browsers
         run: npx playwright install --with-deps ${{ matrix.browser }}
 
+      - name: Set lxd channel env variable
+        id: lxd-env
+        run: |
+          # need to change / to - in lxd channel string for report naming
+          LXD_CHANNEL=$(echo '${{ matrix.lxd_channel }}' | sed 's#/#-#g')
+          echo "LXD_CHANNEL=$LXD_CHANNEL" >> $GITHUB_OUTPUT
+
       - name: Run Playwright tests
-        run: npx playwright test --project ${{ matrix.browser }}
+        run: npx playwright test --project ${{ matrix.browser }}:lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}
 
       - name: Rename Playwright report
-        run: mv blob-report/report.zip blob-report/${{ matrix.browser }}-report.zip
+        if: always()
+        run: mv blob-report/report.zip blob-report/${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}-report.zip
 
-      - name: Upload ${{ matrix.browser }} blob reports to be merged
+      - name: Upload ${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }} blob reports to be merged
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: blob-report-${{ matrix.browser }}
+          name: blob-report-${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}
           path: blob-report
           retention-days: 1
 
--- a/package.json
+++ b/package.json
@@ -19,7 +19,9 @@
     "hooks-remove": "husky uninstall",
     "start": "concurrently --kill-others --raw 'vite --host | grep -v 3000' 'yarn serve'",
     "serve": "./entrypoint",
-    "test-js": "vitest --run"
+    "test-js": "vitest --run",
+    "test-e2e-edge": "npx playwright test --project chromium:lxd-latest-edge firefox:lxd-latest-edge",
+    "test-e2e-stable": "npx playwright test --project chromium:lxd-5.0-stable firefox:lxd-5.0-stable"
   },
   "dependencies": {
     "@canonical/react-components": "0.50.2",
--- a/tests/fixtures/lxd-test.ts
+++ b/tests/fixtures/lxd-test.ts
@@ -1,0 +1,10 @@
+import { test as base } from "@playwright/test";
+
+export type LxdVersions = "5.0-stable" | "latest-edge";
+export type TestOptions = {
+  lxdVersion: LxdVersions;
+};
+
+export const test = base.extend<TestOptions>({
+  lxdVersion: ["latest-edge", { option: true }],
+});
--- a/tests/iso-volumes.spec.ts
+++ b/tests/iso-volumes.spec.ts
@@ -1,4 +1,5 @@
-import { test } from "@playwright/test";
+import { expect } from "@playwright/test";
+import { test } from "./fixtures/lxd-test";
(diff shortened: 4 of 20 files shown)


.github/workflows/pr.yaml
```diff
@@ -68,6 +68,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        lxd_channel: ["5.0/stable", "latest/edge"]
```
mas-who, line 71: FYI, 256 runners can be used in a single workflow run according to github actions docs

tests/iso-volumes.spec.ts
```diff
@@ -71,3 +82,36 @@ test("use custom iso for instance launch", async ({ page }) => {
...
+  await page.goto("/ui/");
+  await page.getByRole("link", { name: "Storage", exact: true }).click();
+  await expect(page.getByTestId("tab-link-Custom ISOs")).toBeHidden();
+});
+
+test("not allowed to launch instance with custom iso for lxd v5.0/stable", async ({
+  page,
+  lxdVersion,
+}) => {
+  test.skip(Boolean(process.env.CI), "github runners lack vm support");
```
edlerd, line 103:
```suggestion
```

tests/projects.spec.ts
```diff
@@ -39,7 +40,9 @@ test("project edit configuration", async ({ page }) => {
     .getByRole("combobox", { name: "Features" })
     .selectOption("customised");
   await page.locator("span").filter({ hasText: "Networks" }).click();
-  await page.locator("label").filter({ hasText: "Network zones" }).click();
+  if (lxdVersion !== "5.0-stable") {
+    await page.locator("label").filter({ hasText: "Network zones" }).click();
+  }
```
edlerd, line 45: Shall we add two branches?
In 5.0 ensure the "Network zones" are hidden, in other click them.

tests/fixtures/lxd-test.ts
```diff
@@ -0,0 +1,23 @@
...
+export type LxdVersions = "5.0-stable" | "latest-edge";
+const LXD_CHANNEL = process.env.LXD_CHANNEL as LxdVersions;
+
+export const test = base.extend<{
+  lxdVersion: LxdVersions;
+  addLxdVersionAnnotation: null;
+}>({
+  lxdVersion: LXD_CHANNEL,
+  addLxdVersionAnnotation: [
+    // eslint-disable-next-line no-empty-pattern
```
edlerd, line 12: This is a bit ugly, maybe there is a better way?

tests/projects.spec.ts
mas-who: sure, will add an extra condition to check for 5.0-stable

tests/fixtures/lxd-test.ts
mas-who: have adjusted

mas-who: In playwright's [Typescript examples](https://playwright.dev/docs/test-fixtures#automatic-fixtures) they have empty objects in the first argument for those functions. Not sure how to overcome this one without disabling the linting rule.

tests/fixtures/lxd-test.ts
```diff
@@ -0,0 +1,26 @@
+import { TestInfo, test as base } from "@playwright/test";
+
+export type LxdVersions = "5.0-stable" | "latest-edge";
+const LXD_CHANNEL = process.env.LXD_CHANNEL as LxdVersions;
```
edlerd, line 4: Shall we have a fallback for running locally? I guess this would currently just fail, if `LXD_CHANNEL` is not set in the environment.

.github/workflows/pr.yaml
```diff
@@ -132,17 +133,25 @@ jobs:
...
+      - name: Set lxd channel env variable
+        id: lxd-env
+        run: |
+          # need to change / to - in lxd channel string for report naming
+          LXD_CHANNEL=$(echo '${{ matrix.lxd_channel }}' | sed 's#/#-#g')
+          echo "LXD_CHANNEL=$LXD_CHANNEL" >> $GITHUB_OUTPUT
+
       - name: Run Playwright tests
-        run: npx playwright test --project ${{ matrix.browser }}
+        run: LXD_CHANNEL=${{ steps.lxd-env.outputs.LXD_CHANNEL }} npx playwright test --project ${{ matrix.browser }}
```
edlerd, line 144: I don't really like, that we are using environment variables and the `--project` param to pass info to playwright. Though I see the project param is natively supported, which when used works easily. Where the different backend versions are not and need some custom extension. So not sure if there is a better way.

edlerd (commented):
As just discussed in person: Let's combine the browser and lxd-backend verison as project, that is communicated to playwright. This way we can avoid passing the LXD_VERSION as environment variable and get it on the playwright report on top level.

tests/fixtures/lxd-test.ts
edlerd: Edit: Not needed if we have the combined project name.

.github/workflows/pr.yaml
mas-who: Instead of doing it this way, we can parameterise projects with [this approach](https://playwright.dev/docs/test-parameterize#parameterized-projects). I think this also then gives the projects a good meaning instead of being placeholders.

package.json
```diff
@@ -19,7 +19,9 @@
     "hooks-remove": "husky uninstall",
     "start": "concurrently --kill-others --raw 'vite --host | grep -v 3000' 'yarn serve'",
     "serve": "./entrypoint",
-    "test-js": "vitest --run"
+    "test-js": "vitest --run",
+    "test-e2e-edge": "npx playwright test --project chromium:lxd-latest-edge firefox:lxd-latest-edge",
+    "test-e2e-stable": "npx playwright test --project chromium:lxd-5.0-stable firefox:lxd-5.0-stable"
```
edlerd, line 24: Nice additions here :+1:

tests/fixtures/lxd-test.ts
```diff
@@ -0,0 +1,10 @@
+import { test as base } from "@playwright/test";
+
+export type LxdVersions = "5.0-stable" | "latest-edge";
+export type TestOptions = {
+  lxdVersion: LxdVersions;
+};
+
+export const test = base.extend<TestOptions>({
+  lxdVersion: ["latest-edge", { option: true }],
```
edlerd, line 9: I am confused as the value passed here seems to be hardcoded. Is the actual value coming from the use block in playwright config? Maybe we don't need the value here, then?

Though, please keep this lxd-test file, because I am going to need it for the [coverage](https://github.com/canonical/lxd-ui/pull/634) very soon -- even if it turns out to be empty and just passes the base test at this point.

edlerd (commented):
QA from the report looks good. One question for a possible simplification below.

tests/fixtures/lxd-test.ts
mas-who: The value set here is the default for the project `use` parameter. I referenced the playwright docs for [parameterising projects](https://playwright.dev/docs/test-parameterize#parameterized-projects). I think { option: true } is necessary but not sure if we can pass in a empty value for the first parameter of the array. Although passing in an empty value will look strange as well I rate, wdyt?

edlerd: If we need this definition here, I think keeping `latest-edge` as a default is reasonable.

Maybe we should document the test setup, as it lately became a bit more complex. Maybe a future story to extend the [architecture file](https://github.com/canonical/lxd-ui/blob/main/ARCHITECTURE.MD)?

edlerd (approved):
LGTM, I like the solution :)

tests/fixtures/lxd-test.ts
mas-who: will do, I did add a bit of context in the HACKING.md, but I think adding to the architecture file is also necessary. Will create a task for it :+1:

PR #658, opened 2024-02-20
Update cookie name after upstream change #569

### LINKED ISSUE #569: Update OIDC cookie name
When https://github.com/canonical/lxd/pull/12628 is merged. The name of the `oidc_access` cookie will change to `oidc_identity`. It should be a simple search and replace :) 

I can see that this is referenced in two places: https://github.com/search?q=repo%3Acanonical%2Flxd-ui%20oidc_access&type=code

edlerd:
## Done

- updated OIDC cookie name as requested in #569 

Fixes #569 

## QA

1. Run the LXD-UI:
    - setup [OIDC locally](https://github.com/canonical/lxd-ui/wiki/Setup-oidc-login) and test the login with oidc against the LXD edge version

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -8,20 +8,18 @@ import ProjectSelector from "pages/projects/ProjectSelector";
 import { isWidthBelow, logout } from "util/helpers";
 import { useProject } from "context/project";
 import { useMenuCollapsed } from "context/menuCollapsed";
-import { getCookie } from "util/cookies";
 import { useDocs } from "context/useDocs";
 
 const isSmallScreen = () => isWidthBelow(620);
 
 const Navigation: FC = () => {
-  const { isRestricted } = useAuth();
+  const { isRestricted, isOidc } = useAuth();
   const docBaseLink = useDocs();
   const { menuCollapsed, setMenuCollapsed } = useMenuCollapsed();
   const { project, isLoading } = useProject();
   const [projectName, setProjectName] = useState(
     project && !isLoading ? project.name : "default",
   );
-  const hasOidcCookie = Boolean(getCookie("oidc_access"));
 
   useEffect(() => {
     project && project.name !== projectName && setProjectName(project.name);
@@ -225,7 +223,7 @@ const Navigation: FC = () => {
                           Settings
                         </NavLink>
                       </li>
-                      {hasOidcCookie && (
+                      {isOidc && (
                         <li className="p-side-navigation__item">
                           <a
                             className="p-side-navigation__link"
--- a/src/context/auth.tsx
+++ b/src/context/auth.tsx
@@ -8,6 +8,7 @@ import { fetchProjects } from "api/projects";
 interface ContextProps {
   isAuthenticated: boolean;
   isAuthLoading: boolean;
+  isOidc: boolean;
   isRestricted: boolean;
   defaultProject: string;
 }
@@ -15,6 +16,7 @@ interface ContextProps {
 const initialState: ContextProps = {
   isAuthenticated: false,
   isAuthLoading: true,
+  isOidc: false,
   isRestricted: false,
   defaultProject: "default",
 };
@@ -57,6 +59,7 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
     <AuthContext.Provider
       value={{
         isAuthenticated: (settings && settings.auth !== "untrusted") ?? false,
+        isOidc: settings?.auth_user_method === "oidc",
         isAuthLoading: isLoading,
         isRestricted,
         defaultProject,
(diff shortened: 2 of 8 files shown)


src/components/Navigation.tsx
```diff
@@ -21,7 +21,9 @@ const Navigation: FC = () => {
   const [projectName, setProjectName] = useState(
     project && !isLoading ? project.name : "default",
   );
-  const hasOidcCookie = Boolean(getCookie("oidc_access"));
+  const hasOidcCookie = Boolean(
```
mas-who, line 24: so I can login via OIDC now, but this never returns true for me which mean I can't logout via the UI. Not sure if I'm missing some oidc configs in lxd?

edlerd: Thank you for the QA. Upstream changed the cookie to be httponly, so we can't read it any more. I reworked the auth bits, so they become compatible to that change.

src/context/auth.tsx
```diff
@@ -57,6 +59,7 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
     <AuthContext.Provider
       value={{
         isAuthenticated: (settings && settings.auth !== "untrusted") ?? false,
+        isOidc: settings?.auth_user_method === "oidc",
```
mas-who, line 62: I see, so I assume if a valid certificate exists but the user authenticated using oidc, then oidc will be the preferred auth method? I guess maybe it's not typical that an lxd server would be set up with both tls and oidc authentication as it might relate to company policies etc

src/context/operationsProvider.tsx
```diff
@@ -40,7 +43,9 @@ const OperationsProvider: FC<Props> = ({ children }) => {
   } = useQuery({
     queryKey: [queryKeys.operations],
     queryFn: () => fetchAllOperations(),
+    enabled: isAuthenticated,
```
mas-who, line 46: thanks for implementing the auth check here as well as in `StatusBar`!

src/util/helpers.tsx
```diff
@@ -248,7 +246,11 @@ export const continueOrFinish = (
 };
 
 export const logout = (): void =>
-  void fetch("/oidc/logout").then(() => window.location.reload());
+  void fetch("/oidc/logout").then(() => {
+    if (!window.location.href.includes("/ui/login")) {
```
mas-who, line 250: do we actually need to do the redirect here? Since authentication check is already happening in App now maybe we let it handle the redirect there?

src/context/auth.tsx
mas-who: just tested the case by adding back my certificate while I'm logged in with oidc. When logging out, I get instantly logged back in haha, I think that's the expected behaviour though since both auth methods are valid

edlerd: It is possible both auth methods are available, I think it is even not possible to disable tls.

If LXD server prefers oidc or tls, when presented with both at the same time -- from testing it seems OIDC is preferred: But then again this is a very unlikely scenario. In the UI we wouldn't show a login to a tls user. You could fetch the `/oidc/login` route manually and be authenticated with both. The UI handles this well, and it seems the backend announces OIDC as auth in this case. The UI will show a logout option in the nav, when used the OIDC auth is removed, but the user stays login as the backend falls back to tls. All in all I think that is expected, though highly unusual to use both in parallel.

src/util/helpers.tsx
mas-who: actually nvm, the user would have to perform an action before they get logged out otherwise, please disregard my previous comment

src/context/operationsProvider.tsx
edlerd: We missed the login cases, when introducing those. Should keep those in mind next time we do something global.

mas-who (approved):

mas-who:
LGTM :+1:

src/util/helpers.tsx
edlerd: Doing it here catches three cases:
1. A user browses to a page that requires login, in this case the App triggers this function and it redirects to login.
2. A user has the UI open on a page that requires login. They loose access (i.e. the cert gets removed server side or the OIDC cookie expires) and come back. The UI refreshes its state with API calls and they hit a 403. App will trigger this function and it redirects to login.
3. A user clicks on logout in the Navigation. We want to immediately send the user to the login page after invalidating and removing their cookie.

edlerd: Your comment brought me to realized we don't need the "did we hit a 403" check in handling any api responses (in helper.tsx) to trigger logout. Becuase if we hit a 403, App will rerender and trigger this function. Which is better, because we could have multiple api calls triggering logout multiple times. App will only trigger it once.

mas-who: Yes, it's better to handle this check in one place :+1:

PR #662, opened 2024-02-21
fix: avoid empty label for a11y compliance

mas-who:
## Done

- Resolved empty label a11y issue for ip address selector 

## QA

1. Run the LXD-UI:
    - Check that the label is not visible for the ip address selector
    - Check wave evaluation tool that no a11y empty label warning is reported when the ip address selector is visible on the page

### DIFF
--- a/src/pages/networks/forms/IpAddressSelector.tsx
+++ b/src/pages/networks/forms/IpAddressSelector.tsx
@@ -26,7 +26,7 @@ const IpAddressSelector: FC<Props> = ({ id, address, setAddress }) => {
       </div>
       <div className="ip-address-selector ip-address-custom">
         <RadioInput
-          label=""
+          label="Custom"
           aria-label="custom"
           checked={isCustom}
           onChange={() => setAddress("")}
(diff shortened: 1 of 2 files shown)


edlerd:
This breaks the layout of the custom radio input:
image

mas-who:
> This breaks the layout of the custom radio input: image

Right, I see the reason is that the `labelClassname` prop is applied to the `label` element that wraps the `input`. I think some custom css will be necessary here, unless if we try making the change upstream again (something like a `labelTextClassname` prop)?

Screenshot from 2024-02-21 13-09-57

edlerd:
> Right, I see the reason is that the `labelClassname` prop is applied to the `label` element that wraps the `input`. I think some custom css will be necessary here, unless if we try making the change upstream again (something like a `labelTextClassname` prop)?

I think this is a very special use case and justifies introduction of a custom `labelClassName`, with custom styling in LXD-UI to hide the label here.

edlerd (approved):
Perfect, thanks for the fix!

PR #666, opened 2024-02-22
chore(doc): added new instructions for e2e test with different lxd versions [WD-9034]

mas-who:
## Done

- Added some context in ARCHITECTURE.md about multi lxd version support in the UI
- Updated instructions in HACKING.md for running e2e tests for stable and edge lxd versions

## QA

1. Run the LXD-UI:
    - Check if updates in docs is clear.

### DIFF
--- a/ARCHITECTURE.MD
+++ b/ARCHITECTURE.MD
@@ -47,3 +47,27 @@ A webhook triggers a [demo](https://github.com/canonical/demos.haus) build for a
 The [update_demo.yml](.github/workflows/update_demo.yml) workflow updates a branch called `demo` automatically on any merge to main. This keeps at least one demo instance running and always up to date.
 
 A secret is needed to enable HaProxy to forward requests to the LXD API backend. Ask one of the collaborators for the secret.
+
+## E2E test setup for multiple lxd versions
+
+The e2e test suite is run whenever a pull request is created and updated. This is done in the `browser-e2e-test` job from the `PR checks` github [workflow](https://github.com/canonical/lxd-ui/blob/6b574493501264dbac5722b1c2858f4d0020af75/.github/workflows/pr.yaml#L1). The job is run using the github actions matrix strategy so that the same tests can be carried out against different lxd versions and browser types in isolated github runner environments simultaneously. Currently e2e tests are run against the following parameter combinations:
+
+| lxd version  | browser  |
+| -----------  | -------- |
+| 5.0/stable   | chromium |
+| 5.0/stable   | firefox  |
+| latest/edge  | chromium |
+| latest/edge  | firefox  |
+
+For each parameter combination, a playwright project defined in [playwright.config.ts](https://github.com/canonical/lxd-ui/blob/6b574493501264dbac5722b1c2858f4d0020af75/playwright.config.ts#L1) is selected by name with reference syntax `browser:lxd-version` and executed. The execution flow is depicted in the diagram below:
+
+```mermaid
+    flowchart TD
+    A(pull request created) -->|other workflow jobs| B(browser-e2e-test)
+    B --> C{Github Matrix Runner}
+    C -->|5.0/stable chromium| D[Run chromium:lxd-5.0-stable]
+    C --> E[...]
+    C -->|latest/edge firefox| F[Run firefox:lxd-latest-edge]
+```
+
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -102,9 +102,19 @@ Install playwright and its browsers
 
     npx playwright install
 
-The tests expect the environment on localhost to be accessible. Execute `dotrun` and start tests with
+The e2e tests can be run against the LTS version (v5.0) or the latest release version for LXD. If you want to run the tests against the LTS version, first downgrade your lxd with
 
-    npx playwright test
+    snap refresh lxd --channel 5.0/stable
```
edlerd, line 107: Maybe suggest to use `latest/edge` here, so the next command below `test-e2e-edge` is connected?

ARCHITECTURE.MD
```diff
@@ -12,6 +12,8 @@ For the production release, [LXD](https://github.com/canonical/lxd) is pulling t
 
 In a clustered LXD environment the LXD-UI is available on all nodes, so you can query it from any cluster member. Authentication is shared between all nodes.
 
+LXD maintains different release branches in parallel. This includes long term support (LTS) releases which can have older versions of LXD bundled with LXD-UI. This means that LXD-UI should be compatible with older versions of LXD (v5.0 or later). Compatibility here means graceful degradation in the UI when certian features are missing in older LXD versions. 
+
```
edlerd, line 16: I was hoping we could describe how the test workflow is architected. Especially the matrix with channels and browsers. Also, how they are configured in `pr.yaml` and the connection to the projects in `playwright.config.ts`. Furthermore how the lxdVersion appears in the end to end tests and how it is used to enable / disable specific tests.

So instead of explaining why multiple version are there, explain how the different moving parts are working, and where they connect. Especially how they are configured, so we have a reference for future changes to the config.

mas-who: Aaaah sorry I misunderstood the assignment, will get that done :+1:

HACKING.md
```diff
@@ -102,9 +102,19 @@ Install playwright and its browsers
 
     npx playwright install
 
-The tests expect the environment on localhost to be accessible. Execute `dotrun` and start tests with
+The e2e tests can be run against the LTS version (v5.0) or the latest release version for LXD. If you want to run the tests against the edge version, first make sure your lxd is up to date with
```
edlerd, line 105:
```suggestion
The e2e tests can be run against the LXD 5.0, or the edge version of LXD. If you want to run the tests against the edge version, first make sure your lxd is up to date with
```
Maybe better to not assume 5.0 is "the LTS", as it will be a new LTS soon.

edlerd (approved):
This is great, thanks! some suggestions below, address them or keep as is -- up to you :)

ARCHITECTURE.MD
```diff
@@ -46,4 +46,30 @@ A webhook triggers a [demo](https://github.com/canonical/demos.haus) build for a
...
+    A(pull request created) -->|other workflow jobs| B(browser-e2e-test)
+    B --> C{Github Matrix Runner}
+    C -->|5.0/stable chromium| D[Run chromium:lxd-5.0-stable]
+    C --> E[...]
+    C -->|latest/edge firefox| F[Run firefox:lxd-latest-edge]
+```
+
+The parameter `lxdVersion` set in the configuration for each playwright project is available in every tests' context and may be used to disable a portion of individual tests or skip tests entirely. You can see an example implementation in this [test](https://github.com/canonical/lxd-ui/blob/6b574493501264dbac5722b1c2858f4d0020af75/tests/iso-volumes.spec.ts#L11) where the test is skipped for an older version of lxd.
+
+To add additional lxd versions for e2e testing, you would have to define two projects in `playwright.config.ts` for chromium and firefox browsers, then set the `lxdVersion` parameter accordingly. Then inside [pr.yaml](https://github.com/canonical/lxd-ui/blob/6b574493501264dbac5722b1c2858f4d0020af75/.github/workflows/pr.yaml#L1) include the new lxd versions for the `browser-e2e-test` job. Once these configurations are added, then you can write tests according to your specific use cases for the new lxd versions.
```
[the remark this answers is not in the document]
mas-who: also linking to playwirght.config.ts here

PR #671, opened 2024-02-24
fix: playwright coverage report showing incorrect results [WD-9070]

mas-who:
## Done

- Resolved issue with playwright coverage report not displaying all results.

## QA

1. Run the LXD-UI:
    - Run the test coverage script locally and inspect that the report produced is now correct

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -23,7 +23,7 @@ const config: PlaywrightTestConfig<TestOptions> = {
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : 3,
+  workers: 1,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: process.env.CI ? "blob" : "html",
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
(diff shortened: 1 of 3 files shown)


playwright.config.ts
```diff
@@ -23,7 +23,7 @@ const config: PlaywrightTestConfig<TestOptions> = {
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : 3,
+  workers: 1,
```
mas-who, line 26: Some of the tests reuse the same resources, running those tests in parallel will fail sometimes

tests/fixtures/lxd-test.ts
```diff
@@ -1,23 +1,24 @@
...
   hasCoverage: [false, { option: true }],
-});
-
-test.beforeEach(async ({ page, hasCoverage }) => {
-  await startCoverage(page, hasCoverage);
-});
-
-test.afterEach(async ({ page, hasCoverage }) => {
-  await finishCoverage(page, hasCoverage);
+  runCoverage: [
```
mas-who, line 14: so I found the issue wasn't with nyc, turns out global `beforeEach` and `afterEach` hooks in playwright should be done this way instead of using `test.beforeEach` and `test.afterEach`. This was discovered after debugging where `finishCoverage` didn't run for every test.

edlerd (approved):
Thanks for solving this!

PR #673, opened 2024-02-26
Make rows in iso and custom-volume modals clickable

edlerd:
## Done

- Make rows in iso and custom-volume modals clickable

Fixes WD-8689

## QA

1. Run the LXD-UI:
    - check "choose custom iso" modal on instance creation. Each iso row should have a hover effect, a pointer cursor and be clickable, same experience as the "select image" modal on instance creation
    - check the instance configuration > disk devices > attach custom device modal, row selection should be same as above iso/image selection

### DIFF
--- a/src/pages/images/CustomIsoSelector.tsx
+++ b/src/pages/images/CustomIsoSelector.tsx
@@ -39,7 +39,10 @@ const CustomIsoSelector: FC<Props> = ({
   ];
 
   const rows = images.map((image) => {
+    const selectIso = () => onSelect(image, "virtual-machine");
+
     return {
+      className: "u-row",
       columns: [
         {
           content: (
@@ -49,16 +52,19 @@ const CustomIsoSelector: FC<Props> = ({
           ),
           role: "cell",
           "aria-label": "Name",
+          onClick: selectIso,
         },
         {
           content: image.pool,
           role: "cell",
           "aria-label": "Storage pool",
+          onClick: selectIso,
         },
         {
           content: isoTimeToString(new Date(image.created_at).toISOString()),
           role: "cell",
           "aria-label": "Uploaded at",
+          onClick: selectIso,
         },
         {
           content:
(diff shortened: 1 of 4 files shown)


mas-who:
QA looks good! I did however pick up that the custom volume modal have some weird vertical scroll behaviour, I think maybe we should adjust the table height so that the action buttons are always in view?

Screenshot from 2024-02-26 20-17-06

Screenshot from 2024-02-26 20-19-59

edlerd:
> QA looks good! I did however pick up that the custom volume modal have some weird vertical scroll behaviour, I think maybe we should adjust the table height so that the action buttons are always in view? Feel free to skip this one though since it's not in the scope of this PR.

Yes, this was a display bug due to pagination -- which I think shouldn't have been enabled within the modals in the first place. I removed the pagination, and now it should be fixed.

mas-who:
> > QA looks good! I did however pick up that the custom volume modal have some weird vertical scroll behaviour, I think maybe we should adjust the table height so that the action buttons are always in view? Feel free to skip this one though since it's not in the scope of this PR.
> 
> Yes, this was a display bug due to pagination -- which I think shouldn't have been enabled within the modals in the first place. I removed the pagination, and now it should be fixed.

Great, thanks for the change, it looks good now! :+1:

piperdeck:
I like this. The only thing I notice that feels a bit wonky is something that probably doesn't have to do with this PR in particular, but it seems like the breakpoints for switching between the full and compact versions of this table are a little over-tuned. It switches to the compact version when the modal still has lots of room to remain full-sized:

https://github.com/canonical/lxd-ui/assets/45367207/d78df5f7-c69a-4ea6-a662-f88ce07f2fb3

Other than that, it looks good to me

edlerd:
> I like this. The only thing I notice that feels a bit wonky is something that probably doesn't have to do with this PR in particular, but it seems like the breakpoints for switching between the full and compact versions of this table are a little over-tuned. It switches to the compact version when the modal still has lots of room to remain full-sized:

This was due to the table configured as responsive, which is not really necessary. I removed that, and it should be fixed now.

piperdeck:
Awesome, this is looking good.

mas-who (approved):

PR #676, opened 2024-02-26
Add storage pool configuration as yaml

edlerd:
## Done

- Add storage pool configuration as yaml
- Hide cursor on yaml forms in read only mode
- Add prompt on switching between yaml and guided forms for networks, profiles, instances and storage pools if the yaml was edited
- Add notification with link to relevant docs page from the yaml config
- refresh yaml editor, when having changes in it and "cancel"
- fix profile edit, when doing changes and going to yaml, the yaml now represents the changed values from the form

Fixes WD-8614

## QA

1. Run the LXD-UI:
    - create a storage pool, switch to YAML configuration, ensure it works as intended
    - edit  a storage pool, switch to YAML configuration, ensure it works as intended

### DIFF
--- a/src/components/forms/YamlConfirmation.tsx
+++ b/src/components/forms/YamlConfirmation.tsx
@@ -1,0 +1,27 @@
+import { FC } from "react";
+import { ConfirmationModal } from "@canonical/react-components";
+
+interface Props {
+  onConfirm: () => void;
+  close: () => void;
+}
+
+const YamlConfirmation: FC<Props> = ({ onConfirm, close }) => {
+  return (
+    <ConfirmationModal
+      confirmButtonLabel="Discard changes"
+      onConfirm={onConfirm}
+      close={close}
+      title="Confirm"
+    >
+      <p>
+        Switching back to guided forms will discard all changes in the YAML
+        editor.
+        <br />
+        Are you sure you want to proceed?
+      </p>
+    </ConfirmationModal>
+  );
+};
+
+export default YamlConfirmation;
--- a/src/components/forms/YamlForm.tsx
+++ b/src/components/forms/YamlForm.tsx
@@ -4,6 +4,7 @@ import { updateMaxHeight } from "util/updateMaxHeight";
 import useEventListener from "@use-it/event-listener";
 import { editor } from "monaco-editor/esm/vs/editor/editor.api";
 import IStandaloneCodeEditor = editor.IStandaloneCodeEditor;
+import classnames from "classnames";
 
 export interface YamlFormValues {
   yaml?: string;
@@ -46,7 +47,10 @@ const YamlForm: FC<Props> = ({
   return (
     <>
       {children}
-      <div ref={containerRef} className="code-editor-wrapper">
+      <div
+        ref={containerRef}
+        className={classnames("code-editor-wrapper", { "read-only": readOnly })}
+      >
         <Editor
           defaultValue={yaml}
           language="yaml"
--- a/src/pages/instances/CreateInstance.tsx
+++ b/src/pages/instances/CreateInstance.tsx
@@ -1,4 +1,4 @@
-import { FC, ReactNode, useEffect, useState } from "react";
+import React, { FC, ReactNode, useEffect, useState } from "react";
 import {
   ActionButton,
   Button,
@@ -74,6 +74,7 @@ import {
 } from "util/instanceValidation";
 import FormFooterLayout from "components/forms/FormFooterLayout";
 import { useToastNotification } from "context/toastNotificationProvider";
+import { useDocs } from "context/useDocs";
 
 export type CreateInstanceFormValues = InstanceDetailsFormValues &
   FormDeviceValues &
@@ -91,6 +92,7 @@ interface PresetFormState {
 }
 
 const CreateInstance: FC = () => {
+  const docBaseLink = useDocs();
   const eventQueue = useEventQueue();
   const location = useLocation() as Location<PresetFormState | null>;
   const navigate = useNavigate();
@@ -381,6 +383,7 @@ const CreateInstance: FC = () => {
       <Form onSubmit={formik.handleSubmit} className="form">
         <InstanceFormMenu
           active={section}
+          formik={formik}
           setActive={updateSection}
           isConfigDisabled={!formik.values.image}
           isConfigOpen={!!formik.values.image && isConfigOpen}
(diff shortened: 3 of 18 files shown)


mas-who:
Testing on the demo server results in the below behaviour:

Screenshot from 2024-02-27 09-07-10

The above error actually occurs when creating a pool normally so it's not limited to the yaml editor. However, the pool actually gets created as shown below. Navigating to the pool configuration page works and updating the pool via the yaml editor also works perfectly.

Screenshot from 2024-02-27 09-08-00

I then tested locally and everything works, so I think something might be off on the demo server, maybe it's got to do with the OS update from last week?

edlerd:
> I then tested locally and everything works, so I think something might be off on the demo server, maybe it's got to do with the OS update from last week?

This error happens when the demo server is slow. We have a timeout in haproxy, and if that hits, you get the above error as the api request will be a haproxy error page that is not valid json.

mas-who:
> > I then tested locally and everything works, so I think something might be off on the demo server, maybe it's got to do with the OS update from last week?
> 
> This error happens when the demo server is slow. We have a timeout in haproxy, and if that hits, you get the above error as the api request will be a haproxy error page that is not valid json.

That's good to know, thanks! Then I think it's LGTM :)

mas-who (approved):

piperdeck:
I have a few thoughts about this, which I think could apply to YAML configuration in the UI generally.

# 1. 

The first is that there's a bit of a frustrating interaction where you click on the YAML, expecting to make some changes, and you get the error saying "cannot edit in read-only editor". I think we should make it more clear that the editor is read-only before the user has a chance to try to type in it.

- Desaturate the whole YAML box while it's in read-only mode, only setting it to full saturation when it's in edit mode
- Disable the text cursor from appearing when the editor is in read-only mode. If I click on a box and I see a text cursor appear, I expect to be able to input text at the location of the cursor
- Of course, the text inside should always be selectable, but perhaps the selection highlighting should be browser/OS native when the box is in read-only mode

# 2.

The inline message that appears above the text box is a little awkward. 

- Is it saying that if I ever go back to the inline forms and change anything, it will reset all the changes I made to the YAML? That's not consistent with a brief test of increasing the size through the YAML and then going back to the inline forms. When I do that, the change I made in the YAML seems to persist. 
- Is it saying that the YAML is always regenerated from the entity's existing configuration every time you open it? Then I would say that instead. I'd like to propose a new wording for this warning but I'm not sure exactly what the technical details are so I'm not sure.
- Is it saying that the YAML config will be lost, but only if I switch back to guided forms before saving? In this case, we could simply show the user a confirmation modal in the event that they click on the guided forms while they have unsaved changes in the YAML

# 3.

A user who switches to YAML configuration is choosing to lose some convenience, but maybe we can make it easier by showing them which YAML configurations are available through commented out config lines? example:

```
name: abc
description: ''
driver: zfs
source: /var/snap/lxd/common/lxd/disks/abc.img
config:
  size: 31GiB

# Uncomment to allow foo operations rather than regular operations
# foo: true

# Uncomment to choose the number of bars in pool
# bar: 0
```

edlerd:
> * Desaturate the whole YAML box while it's in read-only mode, only setting it to full saturation when it's in edit mode

I am not sure how to do this. I played locally with reduced opacity, but this doesn't transport the "this can't be edited" message. We could use a different theme for the editor, but that also seems a bit odd and doesn't transport the wanted message in my opinion.

> * Disable the text cursor from appearing when the editor is in read-only mode. If I click on a box and I see a text cursor appear, I expect to be able to input text at the location of the cursor

This should now work as proposed.

> The inline message that appears above the text box is a little awkward.

We can work around this, I changed how networks and storage pools work, so we translate the yaml form back to the guided forms. Also removed the note as it just works now to switch back and forth.

> A user who switches to YAML configuration is choosing to lose some convenience, but maybe we can make it easier by showing them which YAML configurations are available through commented out config lines? example:

The idea of the YAML forms is to allow adding options that we don't know about in the UI. It seems counterintuitive to maintain a list of possible keys and descriptions. We could use the metadata api, but it seems rather complex to build. This should be a dedicated ticket tough, as it is rather involved to build. A cheap alternative would be to link to the relevant page in the docs. Which might be the better thing to do anyway, as the docs can provide more info on the keys.

piperdeck:
> I am not sure how to do this. I played locally with reduced opacity, but this doesn't transport the "this can't be edited" message. We could use a different theme for the editor, but that also seems a bit odd and doesn't transport the wanted message in my opinion

Playing around with it, I think that removing the text cursor will suffice

> We can work around this, I changed how networks and storage pools work, so we translate the yaml form back to the guided forms. Also removed the note as it just works now to switch back and forth.

Wow, this works really well. I didn't know this solution was possible haha

However, it looks like this introduces a new issue, because if I change the name or driver of a pool in the YAML editor, those changes are adopted by the disabled fields in the guided form, and the "save changes" button is disabled. Maybe instead, if you're switching from the YAML editor to the forms, any changes that would alter the frozen fields should be thrown out.

Otherwise, it might just be safer to have a warning appear when you try to switch back to the forms, saying 

"Switching back to guided forms will discard all changes in the YAML editor. Are you sure you want to proceed?

Cancel
Proceed and discard changes"

> The idea of the YAML forms is to allow adding options that we don't know about in the UI. It seems counterintuitive to maintain a list of possible keys and descriptions. We could use the metadata api, but it seems rather complex to build. This should be a dedicated ticket tough, as it is rather involved to build. A cheap alternative would be to link to the relevant page in the docs. Which might be the better thing to do anyway, as the docs can provide more info on the keys.

A fair point. I think linking to the docs would probably be a suitable solution here.

edlerd:
> However, it looks like this introduces a new issue, because if I change the name or driver of a pool in the YAML editor, those changes are adopted by the disabled fields in the guided form, and the "save changes" button is disabled. Maybe instead, if you're switching from the YAML editor to the forms, any changes that would alter the frozen fields should be thrown out.
> 
> Otherwise, it might just be safer to have a warning appear when you try to switch back to the forms, saying
> 
> "Switching back to guided forms will discard all changes in the YAML editor. Are you sure you want to proceed?
> 
> Cancel Proceed and discard changes"

Yes, a prompt is better because of various problems the user could cause when editing yaml. I changed it for networks, profiles, instances and storage pools

Also fixed a bunch of other problems with yaml configuration (see description) and added the link to the docs from the yaml config pages.

edlerd:
Implemented the discussed changes, @piperdeck can you have another look at this?

PR #679, opened 2024-02-26
Add powerflex driver support

### LINKED ISSUE #677: Support for new Dell PowerFlex storage driver
Hello,

recently we have added support for the new Dell PowerFlex storage driver in LXD (https://github.com/canonical/lxd/pull/12304). The UI should also grow support for managing PowerFlex pools and volumes.

Whilst displaying the pool already works like expected, there seems to be an issue when trying to create custom storage volumes: 

Screenshot from 2024-02-26 15-25-19

It looks that also the pool creation requires some additional changes to accommodate the new driver.

The specification for the driver can be found in https://discourse.ubuntu.com/t/dell-powerflex-storage-driver/40611/6.
The end user documentation is in https://documentation.ubuntu.com/lxd/en/latest/reference/storage_powerflex/.

edlerd:
## Done

- Add powerflex driver support for storage pools

Fixes #677

## QA

1. Run the LXD-UI:
    - to test this we need a lab access, as it is not possible to run powerflex locally or in the demo environment. You can explore the options though on "Create new storage pool" after selecting the "Dell powerflex" driver.
    - see [docs on powerflex](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_powerflex/#storage-pool-configuration)
    - see [spec on powerflex](https://discourse.ubuntu.com/t/dell-powerflex-storage-driver/40611)


src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -28,6 +28,12 @@ export interface StoragePoolFormValues {
   ceph_rbd_clone_copy?: string;
   ceph_user_name?: string;
   ceph_rbd_features?: string;
+  powerflex_domain?: string;
+  powerflex_pool?: string;
+  powerflex_gateway?: string;
+  powerflex_user_password?: string;
```
roosterfish, line 34: Thanks for this! I think we should also have `powerflex.user.name` configurable as this might be different than the default `admin` right from the beginning. I just saw https://github.com/canonical/lxd-ui/pull/676 which would be perfectly fine to tune or modify other settings afterwards.

roosterfish (changes requested):

src/pages/storage/forms/StoragePoolForm.tsx
roosterfish: Nice, thanks!

mas-who:
A quick google search shows there is no easy way to setup a demo cluster with powerflex for testing, QA here is limited to the storage pool creation form for the powerflex driver. Unfortunately at this stage we can't test volume or snapshot creations which seems to be quite interesting based on the vTree structure of powerflex snapshot volumes. In any case, here are some observations on the configs available for creating a powerflex pool in LXD:

1. `powerflex.domain` is not a required field unless if `powerflex.pool` is set and it represents the pool name instead of id. By making it a required field in the UI we can prevent erroneous user entries but isn't that a deviation from API spec?
2. `powerflex.user.password` and `powerflex.user.name` are related settings, having them in separate forms seems a bit weird?

src/util/storagePool.tsx
```diff
@@ -59,3 +72,15 @@ export const testDuplicateStoragePoolName = (
     },
   ];
 };
+
+export const isPowerflexIncomplete = (
+  formik: FormikProps<StoragePoolFormValues>,
+): boolean => {
+  return (
+    formik.values.driver === powerFlex &&
+    (!formik.values.powerflex_domain ||
```
mas-who, line 81: since powerflex_domain is not required unless if powerflex_pool is set, do we try changing the logic here? Although it could be tricky since powerflex_domain is only reuiquired if powerflex_pool is the name and not id of the storage pool

src/util/storagePool.tsx
```diff
@@ -59,3 +72,15 @@ export const testDuplicateStoragePoolName = (
...
+      !formik.values.powerflex_pool ||
+      !formik.values.powerflex_gateway ||
+      !formik.values.powerflex_user_password)
```
mas-who, line 84: isn't `powerflex.user.name` also required based on the spec?

src/pages/storage/forms/StoragePoolFormPowerflex.tsx
```diff
@@ -0,0 +1,57 @@
...
+          name: "powerflex_gateway_verify",
+          defaultValue: "",
+          children: <Select options={optionTrueFalse} />,
+        }),
+        getConfigurationRow({
+          formik,
+          label: "Mode",
+          name: "powerflex_mode",
+          defaultValue: "",
+          children: <Input type="text" />,
```
mas-who, line 43: I think this should be a selector with options `nvme` and `tcp`. Although the lxd doc is saying only nvme is currently supported

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -111,21 +113,59 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
+                placeholder="Enter gateway"
+                help="Address of the PowerFlex Gateway"
+                required
+              />
+              <Input
+                {...formik.getFieldProps("powerflex_pool")}
+                type="text"
+                label="Pool"
+                placeholder="Enter pool"
+                help="ID of the PowerFlex storage pool"
```
mas-who, line 156: this can be the name of the pool as well

mas-who:
Another observation, for powerflex volumes, there is an additional config `block.type` which should be a selector with `thin` and `thick` as options. We should include that in `StorageVolumeFormBlock` if the driver type is powerflex.

src/util/storagePool.tsx
edlerd: Lxd has a default notion of the user being "admin". So it is optional as you only need to set it if it deviates from that default.

src/pages/storage/forms/StoragePoolFormPowerflex.tsx
edlerd: I'll remove this as it doesn't make sense with a single option. Especially as the docs state "The mode gets discovered automatically if the system provides the necessary kernel modules."

src/util/storagePool.tsx
edlerd: Well spotted! I'll relax the condition here to make the domain not required. If it is required, the backend will tell with an error anyways.

mas-who:
Been trying to test this one on demo server but I see the jenkins job keep failing, anyway I checked it locally and it looks good given the limited testing we can do for now.

piperdeck:
It seems to me like the "PowerFlex pool" field name is a little bit confusing, given that presumably the thing you're "creating" in this screen is a PowerFlex storage pool, and also that the very first field of the form is already called "Name", which one would presume is the name of this PowerFlex pool.

I'm not super sure about the details because I don't know a lot about PowerFlex, but assuming that this form is connecting to an existing PowerFlex Pool, this field might be better labelled as "Pool ID/Name", with the support text "Name or ID of remote PowerFlex Pool", and placeholder text "Enter name or ID"

Then, the support text under "Domain" would have to be adjusted to match the wording.

edlerd:
> It seems to me like the "PowerFlex pool" field name is a little bit confusing, given that presumably the thing you're "creating" in this screen is a PowerFlex storage pool, and also that the very first field of the form is already called "Name", which one would presume is the name of this PowerFlex pool.
> 
> I'm not super sure about the details because I don't know a lot about PowerFlex, but assuming that this form is connecting to an existing PowerFlex Pool, this field might be better labelled as "Pool ID/Name", with the support text "Name or ID of remote PowerFlex Pool", and placeholder text "Enter name or ID"

Yes, this is like with ceph -- the name of the pool in powerflex itself. And I think you are right it must be created before using it in lxd. We discussed this with @roosterfish and @mas-who last week, because we felt calling the "Pool name" as before is confusing. So we added the "Powerflex" prefix to it. I think calling it "Pool ID/Name" would still be confusing, as there is a "Name" input above. Maybe we should call use the label "Powerflex pool ID / Name"?

edlerd:
Updated the copy as discussed. @piperdeck can you have another look at this?

PR #685, opened 2024-02-29
Use shorter urls, without /detail/ in path

edlerd:
## Done

- shortened urls, so detail pages use singular forms, before: /ui/ui/project/default/instances/detail/holy-quagga now: ui/project/default/instance/holy-quagga. Same for all other pages with list and detail pages

## QA

1. Run the LXD-UI:
    - spot check links in the UI


src/components/NavLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface Props {
+  to: string;
+  title: string;
+  children: ReactNode;
+}
+
+const NavLink: FC<Props> = ({ to, title, children }) => {
+  // ignore last char to consider /instances and /instance as active
+  const matchPart = to.substring(0, to.length - 1);
+  const isActive = location.pathname.startsWith(matchPart);
```
mas-who, line 13: is it not better to use `useLocation` from `react-router` since it's better to subscribe to React's lifecycle via state variables?

src/components/NavLink.tsx
```diff
@@ -0,0 +1,27 @@
+import { FC, ReactNode } from "react";
+import { Link } from "react-router-dom";
+
+interface Props {
+  to: string;
+  title: string;
+  children: ReactNode;
+}
+
+const NavLink: FC<Props> = ({ to, title, children }) => {
```
mas-who, line 10: Is the reason to have this component instead of `NavLink` from `react-router` so that we don't have to pass `className` prop?

edlerd: I added the component, because we have the custom check if the NavLink is active with 1 character less from the `to` prop. Otherwise the menu entry for "Instances" does not get active, when browsing from /instances to /instance/foo

mas-who: I see, got it :+1:

src/components/NavLink.tsx
edlerd: Yes, that's better -- I changed it.

mas-who:
QA looks good, checked every detail page as well as links on the overview pages for resources. Thanks for make the changes.

mas-who (approved):

PR #687, opened 2024-03-04
feat: added zfs storage pool configs [WD-8615]

mas-who:
## Done

- Added storage pool configs for zfs driver type

## QA

1. Run the LXD-UI:
    - Create a storage pool for zfs driver type, check that zfs pool configs can be set.
    - Edit an existing zfs stroage pool and update its configs

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -14,7 +14,6 @@ import { LxdApiResponse } from "types/apiResponse";
 import { LxdOperationResponse } from "types/operation";
 import axios, { AxiosResponse } from "axios";
 import { LxdClusterMember } from "types/cluster";
-import { cephDriver } from "util/storageOptions";
 
 export const fetchStoragePool = (
   pool: string,
@@ -73,27 +72,32 @@ export const createPool = (
   });
 };
 
-const getClusterPool = (
-  pool: Partial<LxdStoragePool>,
-): Partial<LxdStoragePool> => {
-  const poolForCluster = { ...pool };
-  if (pool.driver !== cephDriver) {
-    delete poolForCluster.config;
+const getClusterAndMemberPools = (pool: Partial<LxdStoragePool>) => {
+  const memberSpecificConfigKeys = new Set([
+    "source",
+    "size",
+    "zfs.pool_name",
+    "lvm.thinpool_name",
+    "lvm.vg_name",
+  ]);
+  const configKeys = Object.keys(pool.config || {});
+  const memberConfig: LxdStoragePool["config"] = {};
+  const clusterConfig: LxdStoragePool["config"] = {};
+  for (const key of configKeys) {
+    if (memberSpecificConfigKeys.has(key)) {
+      memberConfig[key] = pool.config?.[key];
+    } else {
+      clusterConfig[key] = pool.config?.[key];
+    }
   }
 
-  return poolForCluster;
-};
+  const clusterPool = { ...pool, config: clusterConfig };
+  const memberPool = { ...pool, config: memberConfig };
 
-const getMemberPool = (
-  pool: Partial<LxdStoragePool>,
-): Partial<LxdStoragePool> => {
-  const poolForMember = { ...pool };
-  // config keys for ceph storage pool cannot be member specific
-  if (pool.driver === cephDriver) {
-    delete poolForMember.config;
-  }
-
-  return poolForMember;
+  return {
+    clusterPool,
+    memberPool,
+  };
 };
 
 export const createClusteredPool = (
@@ -101,7 +105,7 @@ export const createClusteredPool = (
   project: string,
   clusterMembers: LxdClusterMember[],
 ): Promise<void> => {
-  const memberPool = getMemberPool(pool);
+  const { memberPool, clusterPool } = getClusterAndMemberPools(pool);
   return new Promise((resolve, reject) => {
     void Promise.allSettled(
       clusterMembers.map((item) =>
@@ -110,7 +114,7 @@ export const createClusteredPool = (
     )
       .then(handleSettledResult)
       .then(() => {
-        return createPool(getClusterPool(pool), project);
+        return createPool(clusterPool, project);
       })
       .then(resolve)
       .catch(reject);
@@ -139,7 +143,7 @@ export const updateClusteredPool = (
   project: string,
   clusterMembers: LxdClusterMember[],
 ): Promise<void> => {
-  const memberPool = getMemberPool(pool);
+  const { memberPool, clusterPool } = getClusterAndMemberPools(pool);
   return new Promise((resolve, reject) => {
     void Promise.allSettled(
       clusterMembers.map(async (item) =>
@@ -147,7 +151,7 @@ export const updateClusteredPool = (
       ),
     )
       .then(handleSettledResult)
-      .then(() => updatePool(getClusterPool(pool), project))
+      .then(() => updatePool(clusterPool, project))
       .then(resolve)
       .catch(reject);
   });
--- a/src/components/ConfigurationRow.tsx
+++ b/src/components/ConfigurationRow.tsx
@@ -94,17 +94,19 @@ export const getConfigurationRow = ({
             })}
           </div>,
         )}
-        <div>
-          <Button
-            onClick={toggleDefault}
-            type="button"
-            appearance="base"
-            title="Clear override"
-            hasIcon
-          >
-            <Icon name="close" className="clear-configuration-icon" />
-          </Button>
-        </div>
+        {!disabled && (
+          <div>
+            <Button
+              onClick={toggleDefault}
+              type="button"
+              appearance="base"
+              title="Clear override"
+              hasIcon
+            >
+              <Icon name="close" className="clear-configuration-icon" />
+            </Button>
+          </div>
+        )}
       </div>
     );
   };
--- a/src/pages/storage/forms/StoragePoolForm.tsx
+++ b/src/pages/storage/forms/StoragePoolForm.tsx
@@ -5,14 +5,16 @@ import StoragePoolFormMain from "./StoragePoolFormMain";
 import StoragePoolFormMenu, {
   CEPH_CONFIGURATION,
   MAIN_CONFIGURATION,
+  ZFS_CONFIGURATION,
 } from "./StoragePoolFormMenu";
 import useEventListener from "@use-it/event-listener";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import { LxdStoragePool } from "types/storage";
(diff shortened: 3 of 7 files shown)


mas-who:
Instead of just adding `zfs.clone_copy`, I included the other storage pool level configs for the zfs driver as well. Reason is that the config form looks really empty with only one config, If that's too much I will remove the other configs.

edlerd (commented):
Some comments below. I didn't QA the clustered setup yet.

src/pages/storage/forms/StoragePoolFormZFS.tsx
```diff
@@ -0,0 +1,43 @@
...
+          formik,
+          label: "ZFS pool name",
+          name: "zfs_pool_name",
+          defaultValue: "",
+          children: <Input type="text" placeholder="Enter ZFS pool name" />,
+        }),
+        getConfigurationRow({
+          formik,
+          label: "Clone copy",
+          name: "zfs_clone_copy",
```
edlerd, line 27: In a clustered backend I got this error on creating a zfs pool with this option set to false:
`Config key "zfs.clone_copy" may not be used as member-specific key`

src/pages/storage/forms/StoragePoolFormZFS.tsx
```diff
@@ -0,0 +1,43 @@
...
+          defaultValue: "",
+          children: <Select options={optionTrueFalse} />,
+        }),
+        getConfigurationRow({
+          formik,
+          label: "Export",
+          name: "zfs_export",
```
edlerd, line 34: In a clustered backend I got this error on creating a zfs pool with this option set to false:
Config key `zfs.export" may not be used as member-specific key`

edlerd (commented):
Did QA on a cluster backend, some comments on it below.

src/pages/storage/forms/StoragePoolFormMenu.tsx
```diff
@@ -23,7 +28,9 @@ const StoragePoolFormMenu: FC<Props> = ({ formik, active, setActive }) => {
     setActive,
   };
 
-  const isCephDriver = formik.values.driver === cephDriver;
+  const isDriverWithConfig = [cephDriver, zfsDriver].includes(
+    formik.values.driver,
+  );
```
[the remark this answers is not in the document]
mas-who: Sure, I agree it is easier to read if we are explicit :+1:

src/pages/storage/forms/StoragePoolFormZFS.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const StoragePoolFormZFS: FC<Props> = ({ formik }) => {
+  return (
+    <ScrollableConfigurationTable
+      rows={[
+        getConfigurationRow({
+          formik,
+          label: "ZFS pool name",
+          name: "zfs_pool_name",
```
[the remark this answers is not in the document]
mas-who: Nice catch, the lxd docs has very minimal explanation on what this config does so I'm not sure what is the difference between `zfs.pool_name` and the normal storage pool name. Given that zfs driver pools are not remote pools, it is created and fully managed by lxd so is there a difference between the names? I did however find a [lxd forum thread ](https://discuss.linuxcontainers.org/t/how-to-recover-containers-from-an-imported-zfs-pool/16322)that used this setting though, so it's probably useful, not sure if we should hide it in the UI and let users edit it in the yaml?

src/pages/storage/forms/StoragePoolFormZFS.tsx
```diff
@@ -0,0 +1,43 @@
...
+          defaultValue: "",
+          children: <Input type="text" placeholder="Enter ZFS pool name" />,
```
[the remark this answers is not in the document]
mas-who: good catch, thanks!

src/pages/storage/forms/StoragePoolFormZFS.tsx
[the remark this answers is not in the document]
edlerd: Yes, I think this is the name within zfs. I wouldn't remove it from the UI, it seems useful. But probably we should flag the bug with the core team.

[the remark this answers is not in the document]
mas-who: Okay cool, I will flag it in the lxd ui channel :+1:

src/pages/storage/forms/StoragePoolFormZFS.tsx
mas-who: Cool will look into this :+1:

src/pages/storage/forms/StoragePoolFormZFS.tsx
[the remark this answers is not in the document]
mas-who: [issue](https://github.com/canonical/lxd/issues/13038) created.

src/pages/storage/forms/StoragePoolFormZFS.tsx
mas-who: so the issue is `zfs.clone_copy` and `zfs.export` are cluster level configs, we can't include those in the member specific POST requests for creating/updating pools. However, `zfs.pool_name` is a member specific config so we can't include it in the final POST request for creating/updating the cluster storage pool. Will need to think about this one for a bit, currently we have not catered for this scenario where a storage pool config object should be split into cluster configs and member specific configs.

src/components/ConfigurationRow.tsx
```diff
@@ -94,17 +94,19 @@ export const getConfigurationRow = ({
...
-            onClick={toggleDefault}
-            type="button"
-            appearance="base"
-            title="Clear override"
-            hasIcon
-          >
-            <Icon name="close" className="clear-configuration-icon" />
-          </Button>
-        </div>
+        {!disabled && (
```
mas-who, line 97: if the config is disabled we should be able to clear the override

src/api/storage-pools.tsx
```diff
@@ -14,7 +14,35 @@ import { LxdApiResponse } from "types/apiResponse";
...
+const getClusterAndMemberConfigs = (configs?: LxdStoragePool["config"]) => {
+  if (!configs) {
+    return {};
+  }
+
+  const memberSpecificConfigKeys = new Set([
+    "source",
+    "size",
+    "zfs.pool_name",
+    "lvm.thinpool_name",
```
mas-who, line 27: include lvm configs here as well since we will have yaml editing capabilities soon

mas-who:
tested ceph storage pools, no errors with config keys

src/api/storage-pools.tsx
```diff
@@ -77,9 +105,9 @@ const getClusterPool = (
   pool: Partial<LxdStoragePool>,
 ): Partial<LxdStoragePool> => {
   const poolForCluster = { ...pool };
-  if (pool.driver !== cephDriver) {
-    delete poolForCluster.config;
-  }
```
edlerd, line 82: The ceph case is removed here, but not handled anymore

src/api/storage-pools.tsx
```diff
@@ -88,10 +116,9 @@ const getMemberPool = (
   pool: Partial<LxdStoragePool>,
 ): Partial<LxdStoragePool> => {
   const poolForMember = { ...pool };
-  // config keys for ceph storage pool cannot be member specific
-  if (pool.driver === cephDriver) {
-    delete poolForMember.config;
-  }
```
edlerd, line 94: Ceph case is missing in the changed implementation

edlerd (commented):
It seems a bit complex to have the `getClusterPool`
and the `getMemberPool` functions, that both rely on a generic `getClusterAndMemberConfigs` function. I think all three functions should be defined closer to each other. And then maybe we should collapse them to a single function, because we need, the `memberPool` and the `clusterPool` always together. Currently we evaluate `getClusterAndMemberConfigs` twice and are only ever interested into one of the two return values.

src/api/storage-pools.tsx
mas-who: ah yes this was deliberate. The logic from before was keeping the config object for a ceph cluster pool creation because there are no member specific configs for ceph, with the current approach we are basically doing the same so this is no longer needed.

src/api/storage-pools.tsx
mas-who: this relates to the other comment for `getClusterPool`

src/api/storage-pools.tsx
edlerd: Right, everything not in memberSpecificConfigKeys is a cluster wide config option. Makes sense.

src/api/storage-pools.tsx
```diff
@@ -73,35 +72,40 @@ export const createPool = (
...
+  const memberSpecificConfigKeys = new Set([
+    "source",
+    "size",
+    "zfs.pool_name",
+    "lvm.thinpool_name",
+    "lvm.vg_name",
+  ]);
+  const configKeys = Object.keys(pool.config || {});
+  const memberConfigs: LxdStoragePool["config"] = {};
+  const clusterConfigs: LxdStoragePool["config"] = {};
```
edlerd, line 85: `memberConfig` and `clusterConfig`?

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -52,6 +58,13 @@ export const storagePoolFormToPayload = (
         [getPoolKey("ceph_rbd_features")]: values.ceph_rbd_features,
         source: values.source,
       };
+    } else if (isZFSDriver) {
+      return {
+        [getPoolKey("zfs_clone_copy")]: values.zfs_clone_copy,
+        [getPoolKey("zfs_export")]: values.zfs_export,
```
edlerd, line 64: Once I set an override for any of the two, on removing the override it will not remove the override on update, but the old value will stay.

mas-who: the issue is the `updatePool` api function makes a call to the `PATCH` endpoint, when we clear configs in `getConfigurationRow` we simply set the formik field to `undefined`. In this case, the data will not be updated since it's a PATCH request, we might need to use the PUT endpoint instead but also we need to watch out for fields that cannot be updated

edlerd: This should work, ensuring we have empty string in the payload, when the value is undefined:
```
        [getPoolKey("zfs_clone_copy")]: values.zfs_clone_copy ?? "",
```

mas-who: Does empty string become defaults in the backend?

mas-who: just tested, that works :astonished:

PR #689, opened 2024-03-05
Show error message, when API requests fail

edlerd:
## Done

- when updating an instance, the request might fail, we should show an error and stop the form from the submitting state
- the same is true for any other api requests, we should catch in case the api returns an error and display the message returned by the api if any is available.

## QA

1. Run the LXD-UI:
    - Edit an instance, provoke an error by going to Yaml configuration, add a 2nd profile, that doens't exist. Save changes and ensure an error is displayed.

### DIFF
--- a/src/pages/instances/EditInstance.tsx
+++ b/src/pages/instances/EditInstance.tsx
@@ -54,7 +54,7 @@ import { useEventQueue } from "context/eventQueue";
 import { hasDiskError, hasNetworkError } from "util/instanceValidation";
 import FormFooterLayout from "components/forms/FormFooterLayout";
 import { useToastNotification } from "context/toastNotificationProvider";
-import { instanceLinkFromName } from "util/instances";
+import InstanceLink from "pages/instances/InstanceLink";
 
 export interface InstanceEditDetailsFormValues {
   name: string;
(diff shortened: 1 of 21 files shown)


src/pages/instances/EditInstance.tsx
```diff
@@ -112,32 +112,37 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
+              formik.setSubmitting(false);
+              void queryClient.invalidateQueries({
+                queryKey: [queryKeys.instances],
+              });
+            },
+          );
+        })
+        .catch((e) => {
+          formik.setSubmitting(false);
+          toastNotify.failure("Instance update failed.", e);
```
mas-who, line 144: maybe we should include the `instanceLink` here? Since we are using `toastNotify`, user may visit the history of notifications at a later point and might want to navigate to the instance if they have navigated away

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -58,32 +58,38 @@ const InstanceDetailHeader: FC<Props> = ({ name, instance, project }) => {
...
+                "Renaming instance failed.",
+                new Error(msg),
+                instanceLinkFromOperation({ operation, project }),
+              ),
+            () => formik.setSubmitting(false),
+          );
+        })
+        .catch((e) => {
+          formik.setSubmitting(false);
+          toastNotify.failure("Renaming instance failed.", e);
```
mas-who, line 91: maybe we should add the name of the instance to the title?

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -26,27 +26,32 @@ const DeleteInstanceBtn: FC<Props> = ({ instance }) => {
...
+              new Error(msg),
+              <>
+                Instance <ItemName item={instance} bold />:
+              </>,
+            ),
+          () => setLoading(false),
+        );
+      })
+      .catch((e) => {
+        toastNotify.failure("Instance deletion failed", e);
```
mas-who, line 52: maybe add name of instance?

src/pages/instances/actions/FreezeInstanceBtn.tsx
```diff
@@ -25,31 +25,36 @@ const FreezeInstanceBtn: FC<Props> = ({ instance }) => {
...
+          () => {
+            instanceLoading.setFinish(instance);
+            void queryClient.invalidateQueries({
+              queryKey: [queryKeys.instances],
+            });
+          },
+        );
+      })
+      .catch((e) => {
+        toastNotify.failure("Instance freeze failed", e);
```
mas-who, line 55: maybe add name of instance?

edlerd:
Added links to the entities for most errors.

src/pages/instances/EditInstance.tsx
```diff
@@ -112,32 +113,41 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
 
-      void updateInstance(instancePayload, project).then((operation) => {
-        const instanceName = values.name;
-        const instanceLink = instanceLinkFromName({
-          instanceName,
-          project,
+      void updateInstance(instancePayload, project)
+        .then((operation) => {
+          const instanceName = values.name;
+          const instanceLink = instanceLinkFromName({
```
mas-who, line 119: we could move this outside of the then block and reuse the variable in the catch

edlerd: In the catch block we don't have an operation at hand. The op comes from the response to the request that failed in the catch case.

edlerd: Oh wait, this is not using the operation, but the name, yeah -- that makes sense what you said now :)

mas-who:
LGTM, thanks for the changes!

mas-who (approved):

PR #691, opened 2024-03-08
fix: prevent custom config keys being removed when updating resources

mas-who:
## Done

- Checked edit instances (using `getUnhandledKeyValues`)
- Checked edit profile (using `getUnhandledKeyValues`)
- Checked edit network (using yaml object)
- Checked edit network forward (no config keys)
- Checked edit storage pool (using PATCH request so it's not a problem)
- Checked edit storage volume (updated to use `getUnhandledKeyValues`)
- Checked edit snapshot (no config keys)
- Checked edit image (can't edit images so not a problem)
- Checked edit project (using `getUnhandledKeyValues`)

## QA

1. Run the LXD-UI:
    - Unfortunately not all resource types have yaml support yet e.g. storage pool, storage volume, project etc, QA for this PR will be a bit difficult on the demo server, recommend testing this locally.
    - For each resource type (e.g. instance, network, profile etc), create a new instance of that resource type. Using the lxd cli, set a custom config key for that resource with `lxc [resource-type] set [resource-name] user.foo=bar`, then check its configs using `lxc [resource-type] show [resource-name]` and confirm `user.foo` is present in the resource configs. From there, navigate to the edit page for that resource type, modify anything and then hit save. Run `lxc [resource-type] show [resource-name]` again and confirm that `user.foo` is still present in the configs. NOTE: the `lxc` command for setting and showing resource configs can be slightly different depending on the resource type, please check the MAN page relating to each resource type's cli commands.


src/util/storageVolume.tsx
```diff
@@ -61,12 +62,16 @@ export const getZfsVolumeFormFields = (): (keyof StorageVolumeFormValues)[] => {
 };
 
 export const getVolumeKey = (key: string): string => {
-  if (Object.keys(storageVolumeFormFieldToPayloadName).includes(key)) {
+  if (key in storageVolumeFormFieldToPayloadName) {
```
mas-who, line 65: I think this is a bit more efficient since we are only checking if the key exists in `storageVolumeFormFieldToPayloadName`

src/pages/storage/forms/StorageVolumeForm.tsx
```diff
@@ -79,6 +83,10 @@ export const volumeFormToPayload = (
       [getVolumeKey("zfs_remove_snapshots")]: values.zfs_remove_snapshots,
       [getVolumeKey("zfs_use_refquota")]: values.zfs_use_refquota,
       [getVolumeKey("zfs_reserve_space")]: values.zfs_reserve_space,
+      ...getUnhandledKeyValues(
+        (volume?.config as LxdConfigPair) || {},
+        getVolumeConfigKeys(),
+      ),
```
edlerd, line 89:
```suggestion
      ...getUnhandledKeyValues(volume?.config || {}, getVolumeConfigKeys()),
```
better add the `LxdConfigPair` to the `LxdStorageVolume` type and avoid the cast here:
```

  config: {
    "block.filesystem"?: string;
    "block.mount_options"?: string;
    "block.type"?: string;
    "volatile.rootfs.size"?: number;
    "security.shifted"?: string;
    "security.unmapped"?: string;
    "snapshots.expiry"?: string;
    "snapshots.pattern"?: string;
    "snapshots.schedule"?: string;
    "zfs.blocksize"?: string;
    "zfs.block_mode"?: string;
    "zfs.delegate"?: string;
    "zfs.remove_snapshots"?: string;
    "zfs.use_refquota"?: string;
    "zfs.reserve_space"?: string;
    size?: string;
  } & LxdConfigPair;
```

src/pages/storage/forms/StorageVolumeForm.tsx
```diff
@@ -79,6 +83,10 @@ export const volumeFormToPayload = (
...
     },
     project,
```
edlerd, line 91: We are handling unknown config keys, but there might also be main level keys missing. I.e. see in `src/pages/profiles/EditProfile.tsx:146` how we enrich config and main level keys from the original entity.

Also, might be easier to read, if we evaluate `getVolumeConfigKeys` above the return block and use a const in it.

edlerd (changes requested):
QA looks good, we should also handle unknown main keys (not only the nested ones under config). Added an idea for making the code a tiny bit simpler.

edlerd (approved):
LGTM, thanks for checking and fixing this!

PR #692, opened 2024-03-08
feat: multi-select rows with shift key [WD-8506]

mas-who:
## Done

- Add ability to multi select rows if the shift key is pressed.

## QA

1. Run the LXD-UI:
    - Check normal select rows and bulk actions on the instance list and images list tables still work as expected
    - Check that multi select rows with shift key pressed works as expected i.e. select one row, the hold down shift key and select another row, all rows in between should be selected.

### DIFF
--- a/src/components/SelectableMainTable.tsx
+++ b/src/components/SelectableMainTable.tsx
@@ -1,4 +1,4 @@
-import { FC } from "react";
+import { FC, PointerEvent, useState } from "react";
 import {
   MainTableRow,
   Props as MainTableProps,
@@ -10,6 +10,7 @@ import {
   MainTable,
 } from "@canonical/react-components";
 import classnames from "classnames";
+import useEventListener from "@use-it/event-listener";
 
 interface SelectableMainTableProps {
   filteredNames: string[];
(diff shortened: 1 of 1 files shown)


src/components/SelectableMainTable.tsx
```diff
@@ -34,20 +35,30 @@ const SelectableMainTable: FC<Props> = ({
   headers,
   ...props
 }: Props) => {
+  const [currentSelectedIndex, setCurrentSelectedIndex] = useState<number>();
   const isAllSelected =
     selectedNames.length === filteredNames.length && selectedNames.length > 0;
   const isSomeSelected = selectedNames.length > 0;
 
+  useEventListener<"mousedown">("mousedown", (event) => {
```
mas-who, line 43: this is needed because when holding down shift key and clicking in the browser, the default behaviour is to select all text content between two locations. This prevents the checkbox from being selected while the shift key is held down.

edlerd: I think disabling this so generally is not a good idea, because it prevents text selection. Better pin it down to `event.target` being a relevant checkbox?

src/components/SelectableMainTable.tsx
```diff
@@ -110,7 +150,9 @@ const SelectableMainTable: FC<Props> = ({
             }
             labelClassName="u-no-margin--bottom"
             checked={isRowSelected}
-            onChange={toggleRow}
+            onChange={(event: PointerEvent<HTMLInputElement>) =>
```
edlerd, line 153: I think we can keep this as it was before. In the `toggleRow` function, we can directly access `rowIndex` from the parent scope and the event will be automatically forwarded as first parameter.

src/components/SelectableMainTable.tsx
edlerd: Right, I see.

edlerd: Maybe add a comment about the text selection is prevented, as by the code alone it is not clear why this special listener is here.

edlerd (approved):
LGTM, thanks for adding this! Tiny nit below about clarifying the event listener for future us.

edlerd (approved):

PR #696, opened 2024-03-13
fix: create instance from custom iso

mas-who:
## Done

- Fixed issue when creating an instance from the custom ISO list, the redirect may result in invalid instance type locked for the instance creation page.

## QA

1. Run the LXD-UI:
    - upload foo.iso (from test fixtures), go to Storage -> Custom ISOs and try create an instance from the uploaded custom iso. No error should show (i.e. invalid instance type) when the instance gets created.


edlerd (approved):
Good catch, thanks for fixing it :)

PR #698, opened 2024-03-13
feat: refactor side panels and enabled overlay panels

mas-who:
## Done

- Removed the `DetailPanel` and `Aside` components and refactored them into a compound component `SidePanel`
- Introduced overlay styling variable for the side panel
- Updated `InstanceDetailPanel` and `ProfileDetailPanel` to use `SidePanel`

## QA

1. Run the LXD-UI:
    - Ensure that instance detail panel and profile detail panel still works as expected.


src/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -44,226 +44,239 @@ const InstanceDetailPanel: FC = () => {
...
-              inline
-              className="primary actions-list"
-              items={[
-                <OpenTerminalBtn key="terminal" instance={instance} />,
-                <OpenConsoleBtn key="console" instance={instance} />,
-              ]}
-            />
-            <div className="state">
-              <InstanceStateActions instance={instance} />
+  const content = instance && (
```
edlerd, line 47: Should this be a dedicated component?

src/pages/instances/InstanceDetailPanel.tsx
```diff
@@ -44,226 +44,239 @@ const InstanceDetailPanel: FC = () => {
...
+    </table>
+  );
+
+  return (
+    <SidePanel
+      loading={isLoading}
+      hasError={!instance}
+      className="detail-panel instance-detail-panel"
+    >
+      <SidePanel.HeaderStickyWrapper>
```
edlerd, line 247: This is a wired name. Maybe just `SidePanel.Sticky`, or maybe you have a better idea?

src/pages/profiles/ProfileDetailPanel.tsx
```diff
@@ -46,71 +46,88 @@ const ProfileDetailPanel: FC = () => {
     profile?.used_by,
   ).length;
 
+  const content = profile && (
```
edlerd, line 49: Might be worth a dedicated component?

src/sass/_overlay_panel.scss
```diff
@@ -0,0 +1,25 @@
...
+  width: 100vw;
+  z-index: 103 !important;
+
+  @include medium {
+    width: 500px;
+  }
+
+  @include large {
+    width: 650px;
+  }
```
edlerd, line 16: I think we should use the standard from vanilla for `is-wide` and `is-narrow` and avoid specifying our own sizes.

src/sass/_overlay_panel.scss
```diff
@@ -0,0 +1,25 @@
+.overlay-panel {
+  // mobile
+  height: 100vh;
+  padding-bottom: 0.75rem;
+  padding-top: 0.5rem;
+  position: absolute;
+  width: 100vw;
```
edlerd, line 7: Better rely on the vanilla width?

edlerd (commented):
The composable `SidePanel` is super nice. Some ideas on structure and naming things below.

src/components/SidePanel.tsx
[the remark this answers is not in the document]
mas-who: sure we could but I'm not seeing the immediate benefit of it. Isn't it a bit easier to look at all the components in a single file especially in this case these components are tightly related to each other?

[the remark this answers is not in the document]
edlerd: Yeah, especially as they are very short and not exported, we can probably keep it in one file.

src/pages/instances/InstanceDetailPanel.tsx
mas-who: Sticky sounds good :+1:

src/sass/_overlay_panel.scss
mas-who: this might require a bit of experiment as `is-narrow` is a little too narrow for the contents on some of the permissions side panels but `is-wide` is too wide :sweat_smile: In any case, I'd leave this for now as `_overlay_panel.scss` is currently just a place holder and will most likely be adjusted in the permissions PR.

src/sass/_overlay_panel.scss
mas-who: Sure, but let's maybe change this within the permissions PR since these styles might change still

edlerd: How about having just the default styling here in this pr and adding any custom styling in the other PR if we need it?
So this whole file becomes:
```
.overlay-panel {
  height: 100vh;
  position: absolute;
  z-index: 103 !important;
}
```

Because the other spacings are kind of custom tailored to the permissions already and we should def. use vanilla variables instead of the hardcoded rem values.

src/components/SidePanel.tsx
```diff
@@ -0,0 +1,138 @@
...
+  hasError,
+  className,
+}) => {
+  return (
+    <aside
+      className={classnames("l-aside", className, {
+        "is-narrow": !isOverlay,
+        "is-pinned": !isOverlay,
+        "is-split": isSplit,
+        "overlay-panel": isOverlay,
```
edlerd, line 100:
```suggestion
        "is-overlay": isOverlay,
```
Make it consistent?

mas-who: good idea, I think the scss file name we can still keep the same as `_overlay_panel.scss`?

edlerd: Yeah, sounds good.

edlerd: Or maybe it should match the component and be called side_panel.scss?

src/components/SidePanel.tsx
```diff
@@ -0,0 +1,138 @@
...
+  isSplit = false,
+  loading = false,
+  hasError,
+  className,
+}) => {
+  return (
+    <aside
+      className={classnames("l-aside", className, {
+        "is-narrow": !isOverlay,
+        "is-pinned": !isOverlay,
```
edlerd, line 98: This seems wrong, no?

mas-who: That is correct as is-pinned was always set for the old `DetailPanel`. Although I think it's probably better to pass those in as part of `className` wdyt?

mas-who: actually I will create some props for those classes like `width`, `pinned` etc similar to how it was done before. That's probably a more generalised approach

edlerd (approved):
Now it's perfect, thanks :)

PR #710, opened 2024-03-25
Include yaml editor code in build

edlerd:
## Done

- Include yaml editor code in build process
- Avoid to rely on external domain cdn.jsdelivr.net to load editor code

Fixes WD-9954

## QA

1. Run the LXD-UI:
    - ensure CI passes
    - after merge and LXD edge/latest is build, ensure the editor works in the nightly build as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   "scripts": {
     "clean": "rm -rf node_modules yarn-error.log *.log build/ .jekyll-metadata .bundle playwright-report test-results haproxy-local.cfg coverage",
     "build-html": "cp build/ui/index.html build/index.html",
-    "build": "npx vite build && yarn build-html",
+    "build-monaco": "mkdir build/ui/monaco-editor && cp -R node_modules/monaco-editor/min build/ui/monaco-editor",
+    "build": "npx vite build && yarn build-html && yarn build-monaco",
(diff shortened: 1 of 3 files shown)


mas-who (approved):

mas-who:
For context, this is related to the content-policy header from the backend correct?

edlerd:
> For context, this is related to the content-policy header from the backend correct?

Yes, the current CSP is sent as `Content-Security-Policy:
default-src 'self' data: https://*.ubuntu.com; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'`, disallowing to sideload js from the external cdn.jsdelivr.net domain. It is better not to allow it, and to include the yaml editor code in the actual build instead.

mas-who:
> > For context, this is related to the content-policy header from the backend correct?
> 
> Yes, the current CSP is sent as `Content-Security-Policy: default-src 'self' data: https://*.ubuntu.com; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'`, disallowing to sideload js from the external cdn.jsdelivr.net domain. It is better not to allow it, and to include the yaml editor code in the actual build instead.

Yes got it, thanks for the explanation :slightly_smiling_face:

PR #711, opened 2024-03-25
Ensure login button is not broken into two rows

edlerd:
## Done

- Changed login screen layout, so it won't break the "Create a new certificate" button is not split into more than one row on medium-sized screens

Fixes WD-9657

## QA

1. Run the LXD-UI:
    - test the login screen on various screen widths.

### DIFF
--- a/src/sass/_login.scss
+++ b/src/sass/_login.scss
@@ -1,0 +1,5 @@
+.login {
+  padding-left: $sph--large;
+  padding-right: $sph--large;
+  width: fit-content;
+}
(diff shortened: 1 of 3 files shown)


src/sass/_login.scss
```diff
@@ -0,0 +1,9 @@
+.login {
+  display: flex;
```
mas-who, line 2: I don't think this `display: flex` is necessary, `.login` is applied to a single `div`

mas-who:
QA looks good, just one small query regarding the new styles.

mas-who (approved):

PR #717, opened 2024-03-28
Avoid null pointer exception on custom images

### LINKED ISSUE #716: UI "Images" page fails to render when images exist that have empty properties.
In 5.20-f3dd836 (rev 27049), if there are images present that lack content in the 'properties' attribute:

$ lxc image show opnsense-23.7
auto_update: false
properties: {}
public: false
expires_at: 0001-01-01T00:00:00Z
profiles:
- default

then the "Images" menu item in the UI fails to render properly.  See attached screenshot.

image

edlerd:
## Done

- Avoid null pointer exception on custom images, that have no properties set.

Fixes #716

## QA

1. Run the LXD-UI:
    - Create a snapshot of an image, then publish that image, manipulate it to have empty properties and import it again:
```
lxc publish instance-name/snap0 --alias myImage
lxc image export myImage ./image
tar -xvf image.tar.gz 
vim metadata.yaml  # remove properties from the file
rm image.tar.gz 
lxc image import ./ --alias bar
```
    - then open the image list, where previously it would crash, now it renders fine

### DIFF
--- a/src/pages/images/ImageList.tsx
+++ b/src/pages/images/ImageList.tsx
@@ -89,7 +89,9 @@ const ImageList: FC = () => {
   const filteredImages = images.filter(
     (item) =>
       !query ||
-      item.properties.description.toLowerCase().includes(query.toLowerCase()) ||
+      (item.properties?.description ?? "")
+        .toLowerCase()
+        .includes(query.toLowerCase()) ||
       item.aliases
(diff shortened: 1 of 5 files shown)


mas-who (approved):

mas-who:
LGTM, thanks for the fix :+1:

PR #718, opened 2024-03-28
Use fetch all volumes endpoint

edlerd:
## Done

- use fetch all volumes endpoint
- we used to fetch all storage pools and then all volumes for each pool, with https://github.com/canonical/lxd/issues/12550 a new endpoint to directly fetch all storage volumes is available, this PR is starting to use it.
- This only works with the latest/edge version of lxd

## QA

1. Run the LXD-UI:
    - ensure storages -> volumes page is working as before

### DIFF
--- a/src/context/loadIsoVolumes.tsx
+++ b/src/context/loadIsoVolumes.tsx
@@ -1,13 +1,18 @@
-import { fetchStoragePools, fetchStorageVolumes } from "api/storage-pools";
+import {
+  fetchAllStorageVolumes,
+  fetchStoragePools,
+  fetchStorageVolumes,
+} from "api/storage-pools";
 import { isoToRemoteImage } from "util/images";
 import { RemoteImage } from "types/image";
 import { LxdStorageVolume } from "types/storage";
 
 export const loadIsoVolumes = async (
   project: string,
+  hasStorageVolumesAll: boolean,
 ): Promise<RemoteImage[]> => {
   const remoteImages: RemoteImage[] = [];
-  const allVolumes = await loadVolumes(project);
+  const allVolumes = await loadVolumes(project, hasStorageVolumesAll);
   allVolumes.forEach((volume) => {
     if (volume.content_type === "iso") {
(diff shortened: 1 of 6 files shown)


src/context/loadIsoVolumes.tsx
```diff
@@ -17,24 +16,3 @@ export const loadIsoVolumes = async (
 
   return remoteImages;
 };
-
-export const loadVolumes = async (
```
mas-who, line 21: I think we can't remove this function since we need to support `lxd 5.0/stable`.

mas-who: No adjustments needed for tests since the api call should always be using the right endpoints

src/context/loadIsoVolumes.tsx
```diff
@@ -20,6 +25,15 @@ export const loadIsoVolumes = async (
 
 export const loadVolumes = async (
   project: string,
+  hasStorageVolumesAll: boolean,
+): Promise<LxdStorageVolume[]> => {
+  return hasStorageVolumesAll
+    ? fetchAllStorageVolumes(project)
+    : await collectAllStorageVolumes(project);
```
mas-who, line 32: this `await` isn't needed since `loadVolumes` is awaited when its called.

mas-who:
Interesting observation:

old way of getting volumes:
Screenshot from 2024-04-04 11-16-35

new way of getting volumes:
Screenshot from 2024-04-04 11-18-02

The new way of fetching volumes actually resulted in longer response time from server, this is due to the old way of fetching volumes sends concurrent requests to the server. I think it's still better since we are not overloading the server.

mas-who:
QA looks good, just one small change for code :slightly_smiling_face:

src/context/loadIsoVolumes.tsx
edlerd: good catch, removed it.

mas-who (approved):

PR #732, opened 2024-04-09
Add custom image test WD-10108

edlerd:
## Done

- Add custom image test

Fixes WD-10108

## QA

1. Run the LXD-UI:
    - ensure test suite passes

### DIFF
--- a/tests/images.spec.ts
+++ b/tests/images.spec.ts
@@ -1,0 +1,20 @@
+import { test } from "./fixtures/lxd-test";
+import { deleteInstance, randomInstanceName } from "./helpers/instances";
+
+test("search for custom image and create an instance from it", async ({
+  page,
+}) => {
+  const customInstance = randomInstanceName();
+  const image = "my-custom-image"; // this is created in pr.yaml and coverage.yaml
+
+  await page.goto("/ui/");
+  await page.getByRole("link", { name: "Images", exact: true }).click();
+  await page.getByPlaceholder("Search").click();
+  await page.getByPlaceholder("Search").fill(image);
+  await page.getByRole("button", { name: "Create instance" }).first().click();
+  await page.getByLabel("Instance name").fill(customInstance);
+  await page.getByRole("button", { name: "Create", exact: true }).click();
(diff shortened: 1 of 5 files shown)


tests/instances.spec.ts
```diff
@@ -52,13 +52,15 @@ test.afterAll(async ({ browser }) => {
...
   await page.getByTestId("tab-link-Terminal").click();
+  await page.getByRole("button", { name: "Reconnect" }).click();
+  await page.getByLabel("Command").fill("sh");
+  await page.getByLabel("submit reconnect").click();
   await expect(page.locator(".xterm-screen")).toBeVisible();
-  await page.keyboard.type("lsb_release -a");
+  await page.keyboard.type("cat /etc/issue");
   await page.keyboard.press("Enter");
-  await expect(page.locator(".xterm-rows")).toContainText("Ubuntu");
+  await expect(page.locator(".xterm-rows")).toContainText("Alpine Linux");
```
mas-who, line 61: Are we changing to apline linux because it's less bulky than jammy?

tests/images.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { test } from "./fixtures/lxd-test";
+import { deleteInstance, randomInstanceName } from "./helpers/instances";
+
+test("search for custom image and create an instance from it", async ({
```
mas-who, line 4: this test would fail locally I think, maybe we should skip it for dev environment? Or we could setup a shell script to run before the e2e tests kicks off?

mas-who:
I see the test run time came down a fair bit, nice! Just one query about running the custom image test locally

tests/images.spec.ts
edlerd: Right, this hinges on the availability of the custom image.
I was planning to extend this test to create the custom image from the UI in the next cycle (as we have the item to add support for that on our roadmap). Maybe we can ignore this incompatibility for local test runs for now and manually create the custom image on our dev machines?

mas-who: Cool sounds good :+1:

mas-who (approved):

PR #743, opened 2024-04-15
Ensure instance type is derived from image type

edlerd:
## Done

- fix instance creation quicklinks:
- for custom images created from an instance snapshot. The instance creation flow should select the correct instance type, depending on the type of the custom image (vm or container)
- Keep supporting iso volume quick links from storage -> iso volumes

## QA

1. Run the LXD-UI:
    - publish an image from a vm and from a container snapshot
    - create an instance from the quick links in images 1. from the vm snapshot 2. from the container snapshot
    - upload an iso-volume, create an instance from the link in storage > iso volumes

### DIFF
--- a/src/pages/instances/CreateInstance.tsx
+++ b/src/pages/instances/CreateInstance.tsx
@@ -302,7 +302,7 @@ const CreateInstance: FC = () => {
 
   const isLocalIsoImage = formik.values.image?.server === LOCAL_ISO;
 
-  const handleSelectImage = (image: RemoteImage, type: LxdImageType | null) => {
+  const handleSelectImage = (image: RemoteImage, type?: LxdImageType) => {
     void formik.setFieldValue("image", image);
 
     const devices = formik.values.devices.filter(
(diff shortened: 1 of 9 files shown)


src/pages/instances/CreateInstance.tsx
```diff
@@ -314,19 +314,19 @@ const CreateInstance: FC = () => {
...
-    } else if (type) {
-      void formik.setFieldValue("instanceType", type);
     }
   };
 
   useEffect(() => {
     if (location.state?.selectedImage) {
       const type = location.state.selectedImage.volume
-        ? "iso-volume"
+        ? "virtual-machine"
```
mas-who, line 329: maybe here we can benefit from a comment stating why they type is set to `virtual-machine` if `volume` exists on `selectedImage`?

edlerd: I refactored this a bit so it becomes more straight forward, and we don't need the branch here.

mas-who (approved):

PR #747, opened 2024-04-18
fix(instance): add name length check for instance name (creation/edit)

lorumic:
## Done

- Added length validation check for instance name (creation and edit)

## QA

1. Run the LXD-UI:
    - Create a new instance, input some text that is longer than 63 characters in the "Instance name" field, and check that a proper error message appears
    - Edit an existing instance, change the name in the header by inputting some text that is longer than 63 characters and check that a proper error message appears


edlerd (approved):
Thanks for improving the validation and making it reusable.

PR #750, opened 2024-04-19
Run PR tests on `5.0/edge`, `5.21/edge` and `latest/edge`

simondeziel:
For CI jobs, it's best to catch issues soon after their introduction (`edge`) before they reach users (`stable`).


simondeziel:
Marking as ready for review despite the failing `5.0/edge` tests due to timeouts. I cannot retry those :/

playwright.config.ts
```diff
@@ -43,17 +43,31 @@ const config: PlaywrightTestConfig<TestOptions> = {
   /* Configure projects for major browsers */
   projects: [
     {
-      name: "chromium:lxd-5.0-stable",
+      name: "chromium:lxd-5.0-edge",
       use: {
         ...devices["Desktop Chrome"],
-        lxdVersion: "5.0-stable",
+        lxdVersion: "5.0-edge",
```
edlerd, line 49: Some tests depend on `5.0-stable`, need to change it there as well to `5.0-edge` get them to adjust behavior on the 5.0 backend: https://github.com/search?q=repo%3Acanonical%2Flxd-ui%205.0-stable&type=code
Or maybe this could be a constant that is used.

simondeziel: I've made the `s/5.0-stable/5.0-edge/` changes but I think ideally there would be API extension detection. Unfortunately, that is beyond my coding abilities to implement :/

mas-who: @simondeziel thanks for these changes! Yes we do have API extension detection within the UI components for graceful degradation (see [useSupportedFeatures.tsx](https://github.com/canonical/lxd-ui/blob/main/src/context/useSupportedFeatures.tsx)). Within the test environment, we are using lxd version to execute different tests for specific UI components so that we can ensure the graceful degradation behaves as expected.

ARCHITECTURE.MD
```diff
@@ -65,7 +67,7 @@ For each parameter combination, a playwright project defined in [playwright.conf
     flowchart TD
     A(pull request created) -->|other workflow jobs| B(browser-e2e-test)
     B --> C{Github Matrix Runner}
-    C -->|5.0/stable chromium| D[Run chromium:lxd-5.0-stable]
+    C -->|5.0/stable chromium| D[Run chromium:lxd-5.0-edge]
```
mas-who, line 70:
```suggestion
    C -->|5.0/edge chromium| D[Run chromium:lxd-5.0-edge]
```

edlerd (approved):
LGTM, just one comment above from mason to fix documentation.

Thanks for submitting the PR :)

playwright.config.ts
simondeziel: Ah, good news, thanks!

simondeziel:
@edlerd I've addressed the last comment and force pushed. Thanks!

PR #762, opened 2024-05-03
chore: add tests for permissions

mas-who:
## Done

- Added tests for permissions feature

## QA

1. Run the LXD-UI:
    - Ensure CI passes
    - Run permissions tests locally and make sure it passes.


.github/workflows/pr.yaml
```diff
@@ -148,9 +148,19 @@ jobs:
           LXD_CHANNEL=$(echo '${{ matrix.lxd_channel }}' | sed 's#/#-#g')
           echo "LXD_CHANNEL=$LXD_CHANNEL" >> $GITHUB_OUTPUT
 
+      - name: Create OIDC users
+        if: ${{ matrix.lxd_channel != '5.0/edge' }}
+        shell: bash
+        run: ./tests/scripts/create_oidc_identities
+
```
edlerd, line 155: We need this in the `coverage.yaml` workflow as well, but without the `if:`.

src/util/resourceDetails.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+
+describe("extractResourceDetailsFromUrl", () => {
+  it("extracts resource details from lxd entity urls", () => {
+    const serverPath = "/1.0";
+    const instancePath = "/1.0/instances/instance-1?project=default";
+    const storageVolumePath =
+      "/1.0/storage-pools/default/volumes/virtual-machine/volume-test?project=default&target=node-1";
+    const instanceSnapshotPath = "/1.0/instances/instance-1/snapshots/snap0";
+    const volumeSnapshotPath =
+      "/1.0/storage-pools/default/volumes/custom/volume-test/snapshots/snap0";
```
edlerd, line 11: shall we split this into one test case for each url?

tests/helpers/auth-identities.ts
```diff
@@ -0,0 +1,38 @@
+import { Page } from "@playwright/test";
+import { expect } from "../fixtures/lxd-test";
+
+export const identityBar = "bar@bar.com";
+export const identityFoo = "foo@foo.com";
```
edlerd, line 5: Add a comment referencing the create and delete scripts?

tests/helpers/auth-identities.ts
```diff
@@ -0,0 +1,38 @@
...
+      .locator("label")
+      .click();
+    const rowModifiedIcon = page
+      .getByRole("row", { name: `Select ${group}` })
+      .locator("i");
+    await expect(rowModifiedIcon).toBeVisible();
+  }
+};
+
+export const selectIdentities = async (page: Page, identities: string[]) => {
```
edlerd, line 31: There is only a slight difference in naming this vs the `toggleIdentities` in the other file. Can we find two names that are less ambiguous?

Same with the `toggleGroups` and `selectGroups`, which are named and used in the opposite way.

tests/permissions.spec.ts
```diff
@@ -0,0 +1,392 @@
...
+    lxdVersion !== "5.0-edge",
+    "Fine grained authorisation is supported in lxd 5.21 LTS and latest/edge",
+  );
+  await page.goto("/ui/");
+  await expect(page.getByRole("button", { name: "Permissions" })).toBeHidden();
+});
+
+test.describe("Permission groups", () => {
+  test("create and delete group", async ({ page, lxdVersion }) => {
+    skipPermissionTest(lxdVersion);
```
edlerd, line 59: This reads like we always skip, maybe we can include the conditional in the name while keeping the naming concrete and short?

tests/permissions.spec.ts
```diff
@@ -0,0 +1,392 @@
...
+    await deleteGroup(page, group);
+  });
+
+  test("edit and rename group", async ({ page, lxdVersion }) => {
+    skipPermissionTest(lxdVersion);
+    const group = randomGroupName();
+    const renameGroup = randomGroupName();
+    await createGroup(page, group, group);
+    await editGroup(page, group, renameGroup, renameGroup);
+    await editGroup(page, renameGroup, group, group);
```
edlerd, line 72: We seem to never read the description, shall we remove description parameters from the function arguments for create and edit group?
Can we call "edit group" as `renameGroup`, because I think that's what it does?

tests/permissions.spec.ts
```diff
@@ -0,0 +1,392 @@
...
+  test("add new permissions to group", async ({ page, lxdVersion }) => {
+    skipPermissionTest(lxdVersion);
+    const group = randomGroupName();
+    await createGroup(page, group, group);
+    await selectGroupAction(page, group, "Manage permissions");
+    await addPermission(page, "server", "server", "admin");
+    await removePermission(page, "server", "server", "admin");
+    await restorePermission(page, "server", "server", "admin");
+    await undoChange(page);
+    await redoChange(page);
```
edlerd, line 85: I think we should add some assertions between those, otherwise we don't really test them, they could all be no-ops.

edlerd (commented):
This is pretty good, some comments below.

src/util/resourceDetails.spec.ts
mas-who: sounds good :+1:

tests/helpers/auth-identities.ts
mas-who: agreed, in `auth-identities.ts`, `selectIdentities` should become `selectIdentitiesToModify` and toggleGroups should become `toggleGroupsForIdentities`. Same for the equivalent helpers in `auth-groups.ts`.

.github/workflows/pr.yaml
```diff
@@ -148,9 +148,19 @@ jobs:
...
       - name: Run Playwright tests
         run: npx playwright test --project ${{ matrix.browser }}:lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}
 
+      - name: Delete OIDC users
```
edlerd, line 159: Probably we don't need this clean-up in the workflow, we also don't call it in the coverage flow?

tests/helpers/auth-groups.ts
```diff
@@ -0,0 +1,160 @@
...
+    .getByRole("row", { name: `Select ${groupName} Name` })
+    .getByLabel("group actions menu")
+    .click();
+  await page.getByRole("button", { name: action }).first().click();
+};
+
+export const renameGroup = async (
+  page: Page,
+  groupName: string,
+  newGroupName?: string,
```
edlerd, line 39:
```suggestion
  newGroupName: string,
```
This seems more reasonable? we can also drop the condition with if below in the body.

tests/scripts/create_oidc_identities
```diff
@@ -1,6 +1,8 @@
 #! /usr/bin/env bash
 set -e
 
+sudo lxc config set user.show_permissions=true
```
edlerd, line 4:
```suggestion
lxc config set user.show_permissions=true
```
We probably don't need the sudo here, as the other commands below don't have it either?

Same for the delete script.

tests/permissions.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+    await selectGroupAction(page, group, "Manage permissions");
+    await addPermission(page, "server", "server", "admin");
+    await removePermission(page, "server", "server", "admin");
+    await restorePermission(page, "server", "server", "admin");
+    await undoChange(page);
+    await expect(page.getByText("1 permission will be modified")).toBeHidden();
+    await redoChange(page);
+    await assertModificationStatus(page, "1 permission will be modified");
+    await page.getByRole("button", { name: "Save changes" }).click();
+    await page.waitForSelector(`text=Permissions for group ${group} updated.`);
```
edlerd, line 89: Maybe add a check that the group has 1 permission assigned? Currently we don't ensure the permission is saved correctly.

tests/permissions.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+    await updateSetting(page, "oidc.groups.claim", "text", "test");
+    await visitIdpGroupsPage(page);
+    const configurationLink = page.getByRole("link", {
+      name: "configuration (oidc.groups.claim)",
+    });
+    await expect(configurationLink).toBeHidden();
+    await visitServerSettings(page);
+    await resetSetting(page, "oidc.groups.claim", "text", "-");
+    await visitIdpGroupsPage(page);
+    await expect(configurationLink).toBeVisible();
```
edlerd, line 391: The test should not change the server settings compared to before running the test. Can you restore the initial setting at the end of the test?

tests/permissions.spec.ts
edlerd, line 1: This file is pretty long. Shall we split this by the three features?

tests/scripts/delete_oidc_identities
```diff
@@ -6,3 +6,5 @@ lxd sql global "
     FROM identities
     WHERE name IN ('foo', 'bar');
 "
+
+sudo lxc config set user.show_permissions=false
```
edlerd, line 10: This is an unexpected side effect. When running the tests locally, this can leave the configuration in a different state as it was before. Maybe setting this config option shouldn't be part of the create and delete script? We can add it to the pr and coverage workflow instead, wdyt?

edlerd (commented):
Thanks for the adjustments, this is much better already. Some more comments, then this should be good to merge :)

.github/workflows/pr.yaml
mas-who: Yeah this is more needed for local testing, for cleanup purposes. Not needed in workflows as you mentioned :+1:

tests/scripts/delete_oidc_identities
mas-who: Sure, although to run these tests locally, we'd have to make sure `user.show_permissions` is set to `true`. This is the primary reason why I included the commands in the scripts. Should we then maybe just include this in the repository setup steps in HACKING.md?

mas-who: We can also have those commands included in the `package.json` scripts, although that would make the scripts very long and verbose.

edlerd: I think in package.json we'd run into the same issue: Running the tests would mess with the LXD configuration. I'd be fine adding them to the HACKING.md file.

mas-who: okay sounds good, will add that :+1:

tests/helpers/permissions.ts
edlerd, line 1: Why do we have `tests/helpers/permissions.ts` and `tests/helpers/auth.ts`? I think we should always stick to the `permission` prefix in the tests and in the UI in general. The exceptions being in the api folder, where we mimic the path of the endpoints in the file names. So they should be fine to be named auth-foo.tsx. But for tests, it might be better to stay consistent with the components under test, which all seem to use `PermissionFoo` as a prefix. That way we avoid the ambiguity like in this example here with the two files auth.ts and permissions.ts.

edlerd (commented):
Thanks for making this nice. :)

tests/helpers/server.ts
```diff
@@ -81,3 +81,9 @@ export const resetSetting = async (
   await page.getByRole("button", { name: "Close notification" }).click();
   await validateSettingValue(settingRow, defaultValue);
 };
+
+export const getInitialSettingValue = (page: Page, settingName: string) => {
```
edlerd, line 85:
```suggestion
export const getServerSettingValue = (page: Page, settingName: string) => {
```
This function doesn't know about its return value being initial or update or modified :)

tests/helpers/permissions.ts
mas-who: yeah agreed, will update the file names for tests

tests/helpers/permissions.ts
```diff
@@ -0,0 +1,51 @@
...
+    "Fine grained authorisation is not available for lxd 5.0",
+  );
+
+export const assertModificationStatus = async (
+  page: Page,
+  statusText: string,
+) => {
+  const status = page.getByText(statusText);
+  await expect(status).toBeVisible();
+};
```
edlerd, line 17:
```suggestion
export const assertTextVisible = async (
  page: Page,
  statusText: string,
) => {
  const status = page.getByText(statusText);
  await expect(status).toBeVisible();
};
```
This seems like a useful function. Shall we move it into a more generic helper file and adopt it in other tests as well? Could be a follow-up task. We can use this instead of the current approach: In many tests, we currently click on the text instead of using expect. Maybe create a ticket for this?

edlerd (approved):
Thanks for the adjustments, all good from my side.

tests/helpers/permissions.ts
mas-who: Good idea :+1: here is the [issue](https://warthogs.atlassian.net/browse/WD-11114)

mas-who:
@edlerd thanks for the review!

PR #765, opened 2024-05-08
Case insensitive search in image list on instance launch

edlerd:
## Done

- Case insensitive search in image list on instance launch

## QA

1. Run the LXD-UI:
    - create instance > browse image > search for "ubuntu" and "Ubuntu", it should bring up the same result. Before this would differ.


lorumic (approved):
LGTM

PR #774, opened 2024-05-22
Update react components

edlerd:
## Done

- use new empty state layout more in line with vanilla
- update react-components

## QA

1. Run the LXD-UI:
    - test empty state

## Screenshots

New instance empty state
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test-report-coverage": "cp coverage/unit/coverage-final.json coverage/playwright/ ; nyc report --reporter html --reporter cobertura --reporter text-summary --temp-dir coverage/playwright --report-dir coverage/playwright-report --exclude 'src/lib/**' --exclude 'src/types/**'"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


mas-who:
Thanks for this update. I think the empty state looks fine on normal pages but seems to be a bit far to the right in side panels. Should we collapse the left margin for empty state in side panels?

Screenshot from 2024-05-22 13-48-52

edlerd:
> Thanks for this update. I think the empty state looks fine on normal pages but seems to be a bit far to the right in side panels. Should we collapse the left margin for empty state in side panels?

Good catch, side panels are a special case. I updated them to be full width. Please check again.

mas-who:
> > Thanks for this update. I think the empty state looks fine on normal pages but seems to be a bit far to the right in side panels. Should we collapse the left margin for empty state in side panels?
> 
> Good catch, side panels are a special case. I updated them to be full width. Please check again.

Yeah I think it looks pretty good now :+1:

mas-who (approved):

PR #781, opened 2024-05-27
chore: add oidc login tests [WD-9461]

mas-who:
## Done

- Added OIDC secrets to the `ci-end-to-end` github environment to be access in the PR checks action.
- Added ui login test for OIDC identities. This test is run before other e2e tests since we need to remove tls certificates as part of the process.
- update CONTRIBUTING.md with relevant info for e2e tests

## NB!
Once this PR gets merged, we will need to review approve workflow runs as the browser e2e job is using the `ci-end-to-end` environment. 

## QA

1. Run the LXD-UI:
    - Ensure CI passes.


mas-who:
There is something weird when I tested this on my fork. It seems like the oidc user doesn't have the right permissions to view all the volumes, see [video](https://mas-who.github.io/lxd-ui/reports/pr-8/9243019129/1/#?testId=441c969b5ac029ba9e26-4df5f909a0c5d780fa281).  When I run the same test locally, it works, so it feels like there might be a lxd config difference between CI and local environment. Still trying to figure out what's going on so keeping this PR as draft.

.github/workflows/pr.yaml
```diff
@@ -149,10 +149,25 @@ jobs:
...
+          LXD_OIDC_PASSWORD: ${{ vars.LXD_OIDC_PASSWORD }}
+        run: |
+          echo "LXD_OIDC_CLIENT_ID=$LXD_OIDC_CLIENT_ID" >> .env.local
+          echo "LXD_OIDC_ISSUER=$LXD_OIDC_ISSUER" >> .env.local
+          echo "LXD_OIDC_AUDIENCE=$LXD_OIDC_AUDIENCE" >> .env.local
+          echo "LXD_OIDC_USER=$LXD_OIDC_USER" >> .env.local
+          echo "LXD_OIDC_PASSWORD=$LXD_OIDC_PASSWORD" >> .env.local
+          ./tests/scripts/setup_test
+
+      - name: show lxd configs
```
mas-who, line 169: this step will be removed after initial testing.

.gitignore
```diff
@@ -83,3 +83,4 @@ dist/
 tests/screenshots
 
 haproxy-local.cfg
+tests/.auth
```
edlerd, line 86:
```suggestion
tests/.auth

```

package.json
```diff
@@ -37,6 +37,7 @@
     "axios": "1.6.5",
     "cytoscape": "3.28.1",
     "cytoscape-popper": "2.0.0",
+    "dotenv": "^16.4.5",
```
edlerd, line 40: Can this be in dev dependencies?

tests/scripts/teardown_test
```diff
@@ -0,0 +1,12 @@
...
+
+lxd sql global "
+    DELETE
+    FROM identities
+    WHERE name IN ('foo', 'bar', 'admin');
+"
+lxc auth group delete login-admin
+
+# add tls cert back
+lxc config trust add keys/lxd-ui.crt
```
edlerd, line 12:
```suggestion
lxc config trust add keys/lxd-ui.crt

```

package.json
```diff
@@ -21,8 +21,8 @@
     "start": "concurrently --kill-others --raw 'vite --host | grep -v 3000' 'yarn serve'",
     "serve": "./entrypoint",
     "test-js": "vitest --run",
-    "test-e2e-edge": "tests/scripts/create_oidc_identities && npx playwright test --project chromium:lxd-latest-edge firefox:lxd-latest-edge && tests/scripts/delete_oidc_identities",
-    "test-e2e-5.21-edge": "tests/scripts/create_oidc_identities && npx playwright test --project chromium:lxd-5.21-edge firefox:lxd-5.21-edge && tests/scripts/delete_oidc_identities",
+    "test-e2e-edge": "tests/scripts/setup_test && npx playwright test --project chromium:lxd-latest-edge firefox:lxd-latest-edge && tests/scripts/teardown_test",
+    "test-e2e-5.21-edge": "tests/scripts/setup_test && npx playwright test --project chromium:lxd-5.21-edge firefox:lxd-5.21-edge && tests/scripts/teardown_test",
```
edlerd, line 25: Can we remove the *_oidc_identities scripts?

tests/scripts/setup_test
```diff
@@ -0,0 +1,63 @@
...
+
+# remove tls cert for oidc login
+FINGERPRINT=$(lxc config trust list | grep lxd-ui.crt | awk '{print $8}')
+lxc config trust remove ${FINGERPRINT}
+
+# setup oidc configs
+if [ -f .env.local ]
+then
+  set -o allexport; source .env.local; set +o allexport
+fi
```
edlerd, line 12: It is now mandatory to create the .env.local file to run the e2e tests locally. Can you document how to get the values and which config values are needed in the HACKING.md section about e2e tests?

playwright.config.ts
```diff
@@ -1,6 +1,12 @@
 import type { PlaywrightTestConfig } from "@playwright/test";
 import { devices } from "@playwright/test";
 import { TestOptions } from "./tests/fixtures/lxd-test";
+import dotenv from "dotenv";
+import path from "path";
+import { fileURLToPath } from "url";
+const __filename = fileURLToPath(import.meta.url); // get the resolved path to the file
+const __dirname = path.dirname(__filename); // get the name of the directory
+dotenv.config({ path: path.resolve(__dirname, ".env.local") });
```
edlerd, line 9: This seems to be unused?

playwright.config.ts
```diff
@@ -61,36 +78,46 @@ const config: PlaywrightTestConfig<TestOptions> = {
...
+        storageState: "tests/.auth/user.json",
       },
+      dependencies: ["setup-firefox"],
     },
     {
       name: "coverage",
       use: {
         ...devices["Desktop Chrome"],
         lxdVersion: "latest-edge",
+        storageState: "tests/.auth/user.json",
```
edlerd, line 117: This should be a reference to the same `authFile` constant we use in `auth.setup.ts`.

playwright.config.ts
```diff
@@ -61,36 +78,46 @@ const config: PlaywrightTestConfig<TestOptions> = {
       use: {
         ...devices["Desktop Chrome"],
         lxdVersion: "5.21-edge",
+        storageState: "tests/.auth/user.json",
       },
+      dependencies: ["setup-chromium"],
```
edlerd, line 83: Wouldn't this run the `auth.setup.ts` suite a second time?

tests/scripts/setup_test
```diff
@@ -0,0 +1,63 @@
...
+    VALUES 
+    (
+        (SELECT id from identities WHERE name='admin'),
+        2, 
+        5, 
+        'admin@admin.com', 
+        'admin', 
+        '{}'
+    );
+"
```
edlerd, line 30: Do we still need this? Can we log in with that user normally now and let LXD create the user?

We can setup the idp group mapping and a group with admin permissions to drive the permission allocation.

edlerd (changes requested):

package.json
mas-who: Yes! And also fixed

tests/scripts/setup_test
mas-who: Yes, I commented on the PR that this is a TODO. Just wanted you to have a look at the PR first in case if anything needed to change regarding the approach. Will add to HACKING.md.

playwright.config.ts
mas-who: The environment variables loaded here gets consumed in tests since [playwright.config.ts](https://github.com/canonical/lxd-ui/pull/781/files/af7ff8f7325d1bae1197af0656fc96189e5b714c#diff-f679bf1e58e8dddfc6cff0fa37c8e755c8d2cfc9e6b5dc5520a5800beba59a92) is the playwright entrypoint. Specifically, the LXD_OIDC_USER and LXD_OIDC_PASSWORD variables are required in `auth.setup.ts`. I will leave a comment here to make it clear.

playwright.config.ts
mas-who: Yeah agreed, decided to export `authFile` from `lxd-test.ts` since it's related to global setup.

playwright.config.ts
mas-who: In the local test scripts and in our workflows, we always run playwright tests with the `--project` flag for browser specific projects only. The setup projects only gets run once as a dependency to the browser specific projects.

tests/scripts/setup_test
mas-who: No, we don't need this but I just thought this is easier to setup and results in one less dependency to the auth0 account. But we can get it setup this way for sure.

tests/scripts/setup_test
mas-who: I've added instructions for setting up oidc to run the e2e tests with reference to the [oidc setup wiki page](https://github.com/canonical/lxd-ui/wiki/Setup-oidc-login). Let me know if this is sufficient?

mas-who:
@edlerd also updated the workflows to use repo secrets instead of repo variables.

playwright.config.ts
edlerd: Where is the comment?

tests/setup/auth.setup.ts
```diff
@@ -0,0 +1,25 @@
...
+
+setup("authenticate", async ({ page }) => {
+  await page.goto("/ui/");
+  await loginUser(page);
+  // Check logout functionality
+  await page.getByText("Log out").click();
+  await expect(
+    page.getByRole("link", { name: "Login with SSO" }),
+  ).toBeVisible();
+  await loginUser(page);
```
edlerd, line 22: Should this be a test case? A bit surprising to log in and log out during setup phase.

edlerd (commented):
The workflow seems to be failing, this might be because the secrets access is not available in the for PR, do you have a link to a passing workflow run in your fork?

tests/setup/auth.setup.ts
mas-who: Yeah I agree it's a little strange. Performing a logout in an underlying test would affect the global auth state since the session id would change. I can maybe try splitting the logout into a separate test but in the same file?

mas-who:
> The workflow seems to be failing, this might be because the secrets access is not available in this PR originating from a fork.
> 
> Do you have a link to a passing workflow run in your fork?

Sure here's the [action](https://github.com/mas-who/lxd-ui/actions/runs/9362847963/job/25772392689?pr=8) running on my fork. I had to rebase main quick to get around errors in the screenshot tests. Once it's done, you will likely see failing storage volume tests for the latest/edge and 5.21/edge tests due to the permission bug which should be fixed with this [PR](https://github.com/canonical/lxd/pull/13517). We should follow up on that PR though once merged though, since it would be immediately available in the latest/edge channel but not the 5.21/edge channel.

tests/setup/auth.setup.ts
edlerd: Why can't we have a "normal" test that performs log out and log in again?

mas-who: The idea is that this file should be run first before any other tests to setup the global auth file, which would contain session details for an authenticated identity. The dependency is specified in `playwright.config.ts` where the setup file is run in a separate project, prior to any other projects. What I am proposing is a "normal" test but still run inside the setup file, since a logout test would alter the authentication state and should be run before other tests.

edlerd: There is a way to [control test order](https://playwright.dev/docs/next/test-parallel#control-test-order) in CI, as playwright runs tests not in parallel and in alphabetical order. So the logout/login test could be the last one maybe?

edlerd: As discussed, this is a "normal" playwright test file, which is executed first before the other tests. I was under the wrong assumption this is a special setup file, which is not a test. To conclude this can stay as is, all good.

mas-who:
@edlerd this PR is now ready for review. Although the environment secrets are not accessble in the workflow due to the forked origin for the PR. Here's the action run on my [fork](https://github.com/mas-who/lxd-ui/actions/runs/10828519839) for validation.

CONTRIBUTING.md
```diff
@@ -137,6 +137,18 @@ The e2e tests can be run against LXD 5.0, or the edge version of LXD. If you wan
 
     snap refresh lxd --channel latest/edge
 
+The login e2e tests will require oidc setup for your lxd server. You may refer to the [Setup oidc login wiki page](https://github.com/canonical/lxd-ui/wiki/Setup-oidc-login) for setup instructions. Once you have completed the oidc setup, create a `.env.local` file at the root level of the project and ensure the environment variables shown below are set against the relevant lxd server oidc config values:
+    # Configs that enables OIDC authentication for the lxd server
+    LXD_OIDC_CLIENT_ID=[oidc.client.id]
+    LXD_OIDC_ISSUER=[oidc.issuer]
+    LXD_OIDC_AUDIENCE=[oidc.audience]
+    # Config required for provisioning the OIDC identity with admin permission
```
edlerd, line 145:
```suggestion
    # Config required for provisioning the OIDC identity with permissions through IDP group mappings
```

HACKING.md
```diff
@@ -1 +1 @@
-moved to [contribution guidelines](CONTRIBUTING.md)
\ No newline at end of file
+moved to [contribution guidelines](CONTRIBUTING.md)
```
edlerd, line 1: What changed here? oO

.github/workflows/pr.yaml
```diff
@@ -149,10 +150,24 @@ jobs:
...
+          LXD_OIDC_USER: ${{ secrets.LXD_OIDC_USER }}
+          LXD_OIDC_PASSWORD: ${{ secrets.LXD_OIDC_PASSWORD }}
+          LXD_OIDC_GROUPS_CLAIM: ${{ secrets.LXD_OIDC_GROUPS_CLAIM }}
+        run: |
+          echo "LXD_OIDC_CLIENT_ID=$LXD_OIDC_CLIENT_ID" >> .env.local
+          echo "LXD_OIDC_ISSUER=$LXD_OIDC_ISSUER" >> .env.local
+          echo "LXD_OIDC_AUDIENCE=$LXD_OIDC_AUDIENCE" >> .env.local
+          echo "LXD_OIDC_USER=$LXD_OIDC_USER" >> .env.local
+          echo "LXD_OIDC_PASSWORD=$LXD_OIDC_PASSWORD" >> .env.local
+          echo "LXD_OIDC_GROUPS_CLAIM=$LXD_OIDC_GROUPS_CLAIM" >> .env.local
```
edlerd, line 169:
```suggestion
```
Why write to a file?
I think we can read the environment variables in playwright directly, no need to put them into the env file first.

mas-who: Yeah I think you are right, although I may have had a reason to put it there like that previously. Just testing quick first

mas-who: okay it works, just had to move the env vars to the job level

HACKING.md
mas-who: apparently the existing HACKING.md does not have a new line and I added one here

PR #785, opened 2024-05-28
[WD-11261] Implementation of LXD PWA with a dynamic start_url

Kxiru:
## Done

- Implemented minimal PWA support for the LXD-UI client in index.html.
    - The manifest file/variable supports a dynamic start_url link.
- Added a Logo image for the minimum requirements of the PWA.
- Added a LXD-Screenshot image for the minimum requirements of the PWA.

Specification: [LX078 - PWA Support](https://docs.google.com/document/d/1OpYJVpJhUDPJebsY0Bb-0viLrHwOkcsHPO9yT6nXaDg/edit)

## QA

1. Run the LXD-UI:
    - Open the LXD-UI Client and note the PWA pop up, or, view the 
image
--icon in the url toolbar. One can click this to trigger the installation sequence.

## Screenshots
image
image
image


index.html
```diff
@@ -1,19 +1,68 @@
...
+        "src": window.location.origin + "/assets/img/canonical-lxd-512.png",
+        "type": "image/png",
+        "sizes": "512x512"
+      }
+    ],
+    "id": "LXDID",
+    "start_url": window.location.origin,
+    "background_color": "#E95420",
+    "display": "standalone",
+    "scope": "/",
```
edlerd, line 46:
```suggestion
    "scope": window.location.origin,
```
I think this should also be the main url.

index.html
edlerd, line 1: Please keep the previous formatting for the unchanged lines.

mas-who:
@Kxiru I think David has covered all the points. After making the changes for the screenshot reference path, it should work on the demo server as well.

In case if you are wondering why the path should be changed, you can go to the Source tab in your chrome inspector (see below for reference)

Screenshot from 2024-05-30 09-57-49

index.html
```diff
@@ -1,19 +1,80 @@
...
+  <script type="module" src="/src/index.tsx"></script>
+
+</body>
 
-  </body>
 </html>
+
+<script>
+
+  var screenshot = window.location.origin + "/assets/img/";
```
edlerd, line 29:
```suggestion
  var screenshot = window.location.origin + "/ui/assets/img/";
```
This needs to be under the ui route.

edlerd (changes requested):
The paths to the assets need to be prefixed with `/ui` and the scope field needs fixing.

I am not sure the orange menu bar is the best option, it seems a rather busy colour for it. Maybe use the grey/black from the left-hand navigation, or consult with design for this.

edlerd (changes requested):
It works on localhost, but not yet on the demo server -- and I fear it won't work in the real build either. See comments below.

Kxiru:
@edlerd , I've got it to work on the demo server and I've tidied up index.html a bit too :).

edlerd (commented):
This works great on the demo server with signed certificates!
I found some small improvements that should be done before merging it.

PR #789, opened 2024-06-05
[WD-11114] Feat: Improving tests by using text assertion

Kxiru:
## Done

- Renamed ```assertModificationStatus``` to ```assertTextVisible``` and replaced instances of the former.
- Reviewed all usages of "getByText()" in /lxd-ui/tests/ and replaced with assertTextVisible where appropriate.
- Rewrote the function to conditionally account for getByText instances where an Exact match was required.

```
export const assertTextVisible = async (
  page: Page,
  textValue: string,
  exact = false,
) => {
  const textLocator = page.getByText(textValue, { exact: exact });
  await expect(textLocator).toBeVisible();
};
```

## QA

1. Run the LXD-UI:
    - N/A

## Screenshots

N/A

### DIFF
--- a/tests/helpers/permissions.ts
+++ b/tests/helpers/permissions.ts
@@ -8,12 +8,13 @@ export const skipIfNotSupported = (lxdVersion: LxdVersions) =>
     "Fine grained authorisation is not available for lxd 5.0",
   );
 
-export const assertModificationStatus = async (
+export const assertTextVisible = async (
   page: Page,
-  statusText: string,
+  textValue: string,
+  exact = false,
 ) => {
-  const status = page.getByText(statusText);
-  await expect(status).toBeVisible();
+  const textLocator = page.getByText(textValue, { exact: exact });
+  await expect(textLocator).toBeVisible();
 };
 
 export const undoChange = async (page: Page) => {
--- a/tests/instances.spec.ts
+++ b/tests/instances.spec.ts
@@ -25,6 +25,7 @@ import {
   deleteProfile,
   randomProfileName,
 } from "./helpers/profile";
+import { assertTextVisible } from "./helpers/permissions";
 
 let instance = randomInstanceName();
 let vmInstance = randomInstanceName();
@@ -55,7 +56,7 @@ test("instance terminal operations", async ({ page }) => {
   await page.getByRole("button", { name: "Reconnect" }).click();
   await page.getByLabel("Command").fill("sh");
   await page.getByLabel("submit reconnect").click();
-  await expect(page.getByText("~ #")).toBeVisible();
+  await assertTextVisible(page, "~ #");
   await page.waitForTimeout(1000); // ensure the terminal is ready
   await page.keyboard.type("cat /etc/issue");
   await page.keyboard.press("Enter");
@@ -85,7 +86,7 @@ test("instance edit basic details", async ({ page }) => {
 
   await saveInstance(page, instance);
 
-  await page.getByText("DescriptionA-new-description").click();
+  await assertTextVisible(page, "DescriptionA-new-description");
   await expect(page.locator("#profile-0")).toHaveValue(profile);
   await expect(page.locator("#profile-1")).toHaveValue("default");
 });
@@ -213,11 +214,7 @@ test("instance yaml edit", async ({ page }) => {
   test.skip(Boolean(process.env.CI), "github runners lack vm support");
   await editInstance(page, vmInstance);
   await page.getByText("YAML configuration").click();
-
   await page.locator(".view-lines").click();
-  await page.keyboard.press("PageDown");
-  await page.keyboard.press("PageDown");
-  await page.keyboard.press("PageDown");
   await page.getByText("description: ''").click();
   await page.keyboard.press("End");
   await page.keyboard.press("ArrowLeft");
@@ -225,5 +222,5 @@ test("instance yaml edit", async ({ page }) => {
   await saveInstance(page, vmInstance);
 
   await page.getByText("Main configuration").click();
-  await page.getByText("DescriptionA-new-description").click();
+  await assertTextVisible(page, "DescriptionA-new-description");
 });
--- a/tests/iso-volumes.spec.ts
+++ b/tests/iso-volumes.spec.ts
@@ -1,6 +1,7 @@
 import { test, expect } from "./fixtures/lxd-test";
 import { randomNameSuffix } from "./helpers/name";
 import { deleteInstance, randomInstanceName } from "./helpers/instances";
+import { assertTextVisible } from "./helpers/permissions";
 
 const ISO_FILE = "./tests/fixtures/foo.iso";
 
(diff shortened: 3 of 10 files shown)


Kxiru:
_Observations from WD-11114:_

- In projects.spec.ts , test "project edit configuration" fails for both lxd-5.0-edge and lxd-latest-edge but for different reasons. For 5.0-edge, the checkbox option "Network Zones" is expected to be hidden, but it is visible.
- Continued from above, for lxd-latest-edge, the same test fails because "Clusters" does not exist on the page ("Cluster" does, but this does not hold any clusters and thus the test would still fail".
- In storage.spec.ts test "storage volume edit snapshot configuration" passes for Chromium:lxd-latest-edge but not for chromium:lxd-5.0-edge, as the test seems to have issues on the page.getByPlaceholder() line.

tests/helpers/permissions.ts
```diff
@@ -8,11 +8,12 @@ export const skipIfNotSupported = (lxdVersion: LxdVersions) =>
...
 
-export const assertModificationStatus = async (
+export const assertTextValue = async (
   page: Page,
   statusText: string,
+  exact = false,
 ) => {
-  const status = page.getByText(statusText);
+  const status = page.getByText(statusText, { exact: exact });
   await expect(status).toBeVisible();
```
edlerd, line 17: As this is not about status anymore, I'd rename the variables:
statusText -> textValue
status -> textNode

Maybe a better name for the function is `assertTextVisible`, wdyt?

edlerd:
> * In projects.spec.ts , test "project edit configuration" fails for both lxd-5.0-edge and lxd-latest-edge but for different reasons. For 5.0-edge, the checkbox option "Network Zones" is expected to be hidden, but it is visible.

Did you run the 5.0-edge tests against a LXD backend from the `5.0/edge` channel? The backend behaves differently depending on its version.

> * Continued from above, for lxd-latest-edge, the same test fails because "Clusters" does not exist on the page ("Cluster" does, but this does not hold any clusters and thus the test would still fail".

The cluster test relies on the clustering to be enabled. But be careful to enable it on your local host, it might block LXD on your machine in case the external ip changes -- which is not an expected behaviour for a cluster.

tests/iso-volumes.spec.ts
```diff
@@ -23,24 +24,24 @@ test("upload and delete custom iso", async ({ page, lxdVersion }) => {
...
+  await assertTextValue(page, "type: disk", true);
   await page.getByRole("button", { name: "Cancel", exact: true }).click();
 
   await page.getByPlaceholder("Search for custom ISOs").fill(isoName);
-  await page.getByText("3 B").click();
+  await assertTextValue(page, "3 B");
   await page.getByRole("button", { name: "Delete" }).click();
-  await page.getByText("Delete", { exact: true }).click();
-  await page.getByText(`Custom iso ${isoName} deleted.`).click();
+  await assertTextValue(page, "Delete", true);
```
edlerd, line 43: We need to click this one, currently this lets the tests fail.

edlerd (changes requested):

tests/helpers/permissions.ts
Kxiru: I suppose the function is ascertaining the visibility of the textValue, as opposed to checking it's value against something else, so a rename does make sense.

Same for the properties, although out of curiosity, why the name "textNode" for the element hosting the textValue?

edlerd: I was thinking getByText would return a [dom node](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction). Actually, it is returning a locator, so I suppose `textLocator` is a better name than `textNode`.

Kxiru:
> > * In projects.spec.ts , test "project edit configuration" fails for both lxd-5.0-edge and lxd-latest-edge but for different reasons. For 5.0-edge, the checkbox option "Network Zones" is expected to be hidden, but it is visible.
> 
> Did you run the 5.0-edge tests against a LXD backend from the `5.0/edge` channel? The backend behaves differently depending on its version.
> 
> > * Continued from above, for lxd-latest-edge, the same test fails because "Clusters" does not exist on the page ("Cluster" does, but this does not hold any clusters and thus the test would still fail".
> 
> The cluster test relies on the clustering to be enabled. But be careful to enable it on your local host, it might block LXD on your machine in case the external ip changes -- which is not an expected behaviour for a cluster.

I'm not sure what this means. Are you referring to the different Playwright projects / browsers that the tests can be performed on? If so, yes.

tests/helpers/permissions.ts
Kxiru:
```
export const assertTextVisible = async (
  page: Page,
  textValue: string,
  exact = false,
) => {
  const textNode = page.getByText(textValue, { exact: exact });
  await expect(textNode).toBeVisible();
};
```

tests/projects.spec.ts
```diff
@@ -46,8 +47,8 @@ test("project edit configuration", async ({ page, lxdVersion }) => {
   } else {
     await page.locator("label").filter({ hasText: "Network zones" }).click();
   }
-  await page.getByText("Allow custom restrictions on a project level").click();
 
+  await assertTextVisible(page, "Allow custom restrictions on a project level");
```
edlerd, line 51: This needs to be a click, otherwise we can't do the changes further down in the file.

edlerd:
> I'm not sure what this means. Are you referring to the different Playwright projects / browsers that the tests can be performed on? If so, yes.

The playwright tests for 5.0-edge are meant to run against a LXD backend with version 5.0 from the edge channel. If you run that test suite against a LXD version latest/stable or alike they will fail. You can check with `lxd version` in your back end.

I'd suggest to setup lxd inside a container or vm with various version (i.e. 1 container for 5.0 with `snap refresh lxd --channel=5.0/edge` or `snap install lxd --channel=5.0/edge`. Then another container with the latest/edge channel. And on the host you can keep the 5.21/stable version. That way you can switch out the `LXD_UI_BACKEND_IP` between host and the two containers, then restart dotrun and you have a backend with a different version.

tests/instances.spec.ts
```diff
@@ -213,17 +214,13 @@ test("instance yaml edit", async ({ page }) => {
   test.skip(Boolean(process.env.CI), "github runners lack vm support");
   await editInstance(page, vmInstance);
   await page.getByText("YAML configuration").click();
-
   await page.locator(".view-lines").click();
-  await page.keyboard.press("PageDown");
-  await page.keyboard.press("PageDown");
-  await page.keyboard.press("PageDown");
```
mas-who, line 220: why was this removed?

mas-who:
Looks good, thanks for doing this! :+1: Just one small clarification from my side.

tests/instances.spec.ts
Kxiru: The test wasn't running with them in. The item that needed to be viewed was not visible after 3 pagedown clicks.

edlerd (approved):
LGTM, thanks for improving the tests!

PR #790, opened 2024-06-05
Use doc link titles as link text in case they are available

edlerd:
## Done

- Use doc link titles as link text in case they are available from the object.inv.txt content

## QA

1. Run the LXD-UI:
    - open settings > core.https_address, override it. Under the text input should be written "see How to expose LXD to the network" with a link to the docs
    - ensure other links to docs are correct.

### DIFF
--- a/src/util/config.spec.ts
+++ b/src/util/config.spec.ts
@@ -81,7 +81,7 @@ describe("toConfigFields and replaceDocLinks", () => {
     );
 
     expect(result).toBe(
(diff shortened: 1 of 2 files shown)


mas-who (approved):

mas-who:
LGTM :+1: other links looks good and also links to the right doc sections.

As an idea, maybe we can add tests to check if specific sever setting links have the correct text? We can add those tests without actually overriding the settings. This could allow us catch issues ahead of time in case if documentation code changes in lxd on the latest/edge channel

Kxiru:
LGMT too! I verified all of the links and they point to the correct places.

edlerd:
> As an idea, maybe we can add tests to check if specific sever setting links have the correct text? We can add those tests without actually overriding the settings. This could allow us catch issues ahead of time in case if documentation code changes in lxd on the latest/edge channel

You mean an end-to-end test for the doc links? Sounds like a good idea.

PR #793, opened 2024-06-07
[WD-11660] LXD-UI e2e test for operations

Kxiru:
## Done

Created a test case in the new file ```operations.spec.ts``` to cover the following instance operations:

- create an instance
- validate create operation is in operation list
- start instance
- wait for the notification instance was started
- click on the “x operations” link in the status bar
- validate the operation to start the instance is in the operation list
- on instances page: stop instance
- validate stop operation is in operation list
- delete instance
- validate delete operation is in operation list

## QA

1. Run the LXD-UI:
    - N/A

## Screenshots

N/A

### DIFF
--- a/tests/operations.spec.ts
+++ b/tests/operations.spec.ts
@@ -1,0 +1,47 @@
+import { test } from "./fixtures/lxd-test";
+import {
+  createInstance,
+  deleteInstance,
+  randomInstanceName,
+  visitAndStartInstance,
+  visitAndStopInstance,
+} from "./helpers/instances";
+import { assertTextVisible } from "./helpers/permissions";
+import {
+  checkNotificationExists,
+  dismissNotification,
+} from "./helpers/notification";
+import { validateOperation } from "./helpers/operations";
+
+test("instance operations are recognised on the Operations page", async ({
+  page,
+}) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+
+  // validate create operation is in operation list
+  await page.goto("/ui/operations");
+  await validateOperation(page, "Creating Instance", instance);
+
+  // start instance and wait for the notification instance was started
+  await visitAndStartInstance(page, instance);
+  await checkNotificationExists(page);
+  await assertTextVisible(page, "Instance " + instance + " started");
+
+  // click on the “x operations” link in the status bar
+  await dismissNotification(page);
+
+  // validate the operation to start the instance is in the operation list
+  await page.goto("/ui/operations");
+  await validateOperation(page, "Starting Instance", instance);
+
+  // stop instance and validate stop operation is in operation list
+  await visitAndStopInstance(page, instance);
+  await page.goto("/ui/operations");
+  await validateOperation(page, "Stopping Instance", instance);
+
+  // delete instance and validate delete operation is in operation list
+  await deleteInstance(page, instance);
+  await page.goto("/ui/operations");
+  await validateOperation(page, "Deleting Instance", instance);
+});
--- a/tests/helpers/operations.ts
+++ b/tests/helpers/operations.ts
@@ -1,0 +1,12 @@
+import { Page } from "@playwright/test";
+import { expect } from "../fixtures/lxd-test";
+
+export const validateOperation = async (
+  page: Page,
+  title: string,
+  instance: string,
+) => {
+  await expect(
+    page.getByLabel("Action", { exact: true }).getByText(title + instance),
+  ).toBeVisible();
+};


tests/operations.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  await createProfile(page, profile);
+  await createInstance(page, instance);
+  await page.close();
+});
+
+test.afterAll(async ({ browser }) => {
+  const page = await browser.newPage();
+  await deleteProfile(page, profile);
+  await page.close();
+});
```
edlerd, line 36: We probably don't need the before and after for this test. Also, the profile is not needed. The only relevant line is with create instance. We can add that as a first thing to do in the test below.

tests/operations.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  page,
+}) => {
+  await page.goto("/ui/operations");
+
+  // validate create operation is in operation list
+  await page
+    .getByRole("row", { name: "Action" })
+    .filter({ hasText: "Starting Instance" })
+    .filter({ hasText: instance })
+    .getByText(instance);
```
edlerd, line 48:
```suggestion
  const operationLocator = await page
    .getByRole("row", { name: "Action" })
    .filter({ hasText: "Starting Instance" })
    .filter({ hasText: instance })
    getByText(instance);
   await expect(operationLocator).toBeVisible();
```
I think we need the extra expect. Might be worth to put this into a helper function like `validateOperation(title:string, instance: string)` and reuse it multiple time in this test. With that, the code might get very straight forward and reduce the need for some comments.

edlerd (commented):
This seems good already! Some comments to make this simpler below. The linter also has something.

tests/operations.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+import { assertTextVisible } from "./helpers/permissions";
+import {
+  checkNotificationExists,
+  dismissNotification,
+} from "./helpers/notification";
+
+let instance = randomInstanceName();
+let profile = randomProfileName();
+
+test.beforeAll(async ({ browser, browserName }) => {
```
mas-who, line 23: `beforeAll` and `afterAll` are not necessary in this test suite since there is only one test. Probably clearer to have these setup/tear-down steps inside the test, easier to read imo.

tests/operations.spec.ts
Kxiru: Playwright/code doesn't like this. I get an infinite loading screen when using the trace viewer, and when using the browser although the element is there, it isn't being picked up on. I also received this error:
imageAny thoughts on how to get around this?

Kxiru: @edlerd although I have commented out the usage in the latest commit, you can find the implementation of the helper function at tests/helpers/operations.ts.

mas-who: @Kxiru in vscode for the playwright extension, you can use the "Pick locator" functionality to identify the selector required for targeting a specific element in the test browser. See if that helps with the above error?

Screenshot from 2024-02-21 13-09-57

tests/operations.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  await visitAndStartInstance(page, instance);
+  await checkNotificationExists(page);
+  await assertTextVisible(page, "Instance " + instance + " started");
+
+  // click on the “x operations” link in the status bar
+  await dismissNotification(page);
+
+  // validate the operation to start the instance is in the operation list
+  await page.goto("/ui/operations");
+  await page
```
mas-who, line 49: Any playwright function that returns a `Locator` object shouldn't be awaited. e.g. `getByText`. Instead, actions or assertions on the `Locator` object should be awaited. e.g. `await page.getByText(instance).click()` or `await expect(page.getByText(instance)).toBeVisible()`

tests/operations.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+let instance = randomInstanceName();
+
+test("instance operations are recognised on the Operations page", async ({
+  browser,
+  browserName,
+}) => {
+  // Setup
+  instance = `${browserName}-${instance}`;
+  const page = await browser.newPage();
```
edlerd, line 24:
```suggestion
  page
}) => {
```
The framework is already giving us a new page for this test, no need to open a new page manually.

tests/operations.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  await page.goto("/ui/operations");
+  await validateOperation(page, "Stopping Instance", instance);
+
+  // delete instance and validate delete operation is in operation list
+  await deleteInstance(page, instance);
+  await page.goto("/ui/operations");
+  await validateOperation(page, "Deleting Instance", instance);
+
+  // Clean-up
+  await page.close();
```
edlerd, line 56:
```suggestion
```
We don't need to close the page at the end if we don't manually open it.

edlerd (changes requested):

tests/operations.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+test("instance operations are recognised on the Operations page", async ({
+  page,
+}) => {
+  await createInstance(page, instance);
+
+  await page.goto("/ui/operations");
+
+  // validate create operation is in operation list
+
+  await validateOperation(page, "Creating Instance", instance);
```
edlerd, line 27:
```suggestion
  // validate the operation to create the instance is in the operation list
  await page.goto("/ui/operations");
  await validateOperation(page, "Creating Instance", instance);
```
Tiny nit.

edlerd (approved):
LGTM, thanks for adding the tests.
Tiny nitpick below. Also, good to merge without it.

tests/operations.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  visitAndStopInstance,
+} from "./helpers/instances";
+import { assertTextVisible } from "./helpers/permissions";
+import {
+  checkNotificationExists,
+  dismissNotification,
+} from "./helpers/notification";
+import { validateOperation } from "./helpers/operations";
+
+let instance = randomInstanceName();
```
edlerd, line 16:
```suggestion
const instance = randomInstanceName();
```
The linter complained about this, it can be a const and it can be within the test, no need to have it on the root level.

edlerd (changes requested):

tests/operations.spec.ts
Kxiru: Resolved in the next push, thanks!

edlerd (approved):
LGTM, thanks for driving this forward quickly.

PR #795, opened 2024-06-10
[WD-11889] LXD add e2e testing for doc links

Kxiru:
## Done

- chore(tests): added e2e testing for doc links

## QA

1. Run the LXD-UI:
    - N/A

## Screenshots

N/A

### DIFF
--- a/tests/doc-links.spec.ts
+++ b/tests/doc-links.spec.ts
@@ -1,0 +1,232 @@
+import { test } from "./fixtures/lxd-test";
+import {
+  createInstance,
+  deleteInstance,
+  editInstance,
+  randomInstanceName,
+} from "./helpers/instances";
+import {
+  createProject,
+  deleteProject,
+  randomProjectName,
+} from "./helpers/projects";
+import { validateLink } from "./helpers/doc-links";
+import { openServerSetting, visitServerSettings } from "./helpers/server";
+import { activateOverride } from "./helpers/configuration";
+
+test("Ensure the documentation link text and link targets are present: Server Settings > Override", async ({
+  page,
+  lxdVersion,
+}) => {
+  test.skip(
+    lxdVersion === "5.0-edge",
+    "Generated documentation only started shipping with newer versions of LXD",
+  );
+  await visitServerSettings(page);
+
+  await openServerSetting(page, "cluster.https_address");
+  await validateLink(
+    page,
+    "Separate REST API and clustering networks",
+    "/documentation/howto/cluster_config_networks/#cluster-https-address",
+  );
+
+  await openServerSetting(page, "core.bgp_address");
+  await validateLink(
+    page,
+    "How to configure LXD as a BGP server",
+    "/documentation/howto/network_bgp/#network-bgp",
+  );
+
+  await openServerSetting(page, "core.dns_address");
+  await validateLink(
+    page,
+    "Enable the built-in DNS server",
+    "/documentation/howto/network_zones/#network-dns-server",
+  );
+
+  await openServerSetting(page, "core.https_address");
+  await validateLink(
+    page,
+    "How to expose LXD to the network",
+    "/documentation/howto/server_expose/#server-expose",
+  );
+
+  await openServerSetting(page, "core.metrics_address");
+  await validateLink(page, "How to monitor metrics", "/documentation/metrics/");
+
+  await openServerSetting(page, "core.storage_buckets_address");
+  await validateLink(
+    page,
+    "How to manage storage buckets and keys",
+    "/documentation/howto/storage_buckets/#howto-storage-buckets",
+  );
+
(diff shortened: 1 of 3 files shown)


tests/doc-links.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+}) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  await editInstance(page, instance);
+
+  await page.getByText("Security policies").click();
+  let buttons = page.getByRole("row").getByRole("button");
+  for (let i = 0; i < (await buttons.count()) - 1; i++) {
+    await buttons.nth(i).click();
+  }
```
edlerd, line 76: There is an `activateOverride` helper that might be useful instead of clicking all row buttons.

tests/helpers/doc-links.ts
```diff
@@ -0,0 +1,19 @@
...
+  linkText: string,
+  link: string,
+  setting = "",
+) => {
+  if (setting) {
+    await page
+      .locator("css=tr", { hasText: setting })
+      .getByRole("button")
+      .click();
+  }
```
edlerd, line 15: Probably better structured to split this into a dedicated `editServerSetting` function (or find a better suiting name). Then this function doesn't need an optional parameter and can focus on the doc link validation.

tests/doc-links.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+  deleteProject,
+  randomProjectName,
+} from "./helpers/projects";
+import { validateDocumentationLink } from "./helpers/doc-links";
+import { visitServerSettings } from "./helpers/server";
+
+test("Ensure the documentation link text and link targets are present: Server Settings > Override", async ({
+  page,
+}) => {
+  await visitServerSettings(page);
```
edlerd, line 19: I think for 5.0 we should skip this test, as the generated documentation only started shipping with the newer versions of LXD. You can find examples in other tests how to do this.

mas-who: I think we need to skip the instance config and project config tests for 5.0 as well since the required api extension `metadata_configuration` is not available for LXD v5.0-stable, see [doc](https://documentation.ubuntu.com/lxd/en/stable-5.0/api-extensions/).

fyi @Kxiru API extension is how we determine if certain features are available for a specific LXD installation. The available API extensions are different per LXD version. You can check the `useSupporteddFeatures` hook and see where it gets used in the code base to get a sense on how it works.

Kxiru: @mas-who , how can you tell what api extension is used and when/where?

I'll make the other tests also only available for LXD 5++

tests/doc-links.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+  await createProject(page, project);
+  await page.getByRole("link", { name: "Configuration" }).click();
+  await page.getByRole("button", { name: "Edit configuration" }).click();
+  await page.getByText("Allow custom restrictions on a project level").click();
+
+  await page.getByText("Resource limits").click();
+  let buttons = page.getByRole("row").getByRole("button");
+  for (let i = 0; i < (await buttons.count()); i++) {
+    await buttons.nth(i).click();
+  }
```
edlerd, line 143: Can we also use the `activateOverride` helper here? I think it works on all configuration pages.

tests/doc-links.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+    "Generated documentation only started shipping with newer versions of LXD",
+  );
+  await visitServerSettings(page);
+
+  await validateServerSettingLink(
+    page,
+    "Separate REST API and clustering networks",
+    "/documentation/howto/cluster_config_networks/#cluster-https-address",
+    "cluster.https_address",
+  );
```
edlerd, line 35: I think here was a misunderstanding. I'd suggest splitting the function to `editServerSetting` so the code becomes very descriptive: 
```suggestion
  await editServerSetting(page, "cluster.https_address");
  await validateDocumentationLink(
    page,
    "Separate REST API and clustering networks",
    "/documentation/howto/cluster_config_networks/#cluster-https-address"
  );
```

tests/helpers/doc-links.ts
```diff
@@ -0,0 +1,27 @@
+import { Page } from "@playwright/test";
+import { expect } from "../fixtures/lxd-test";
+
+export const validateDocumentationLink = async (
```
edlerd, line 4: Now the function is not bound to docs anymore, maybe just validateLink?

edlerd (approved):
LGTM, some more ideas to improve the code below.

tests/doc-links.spec.ts
Kxiru: Ah, I understand now, I'll action that.

edlerd (approved):
LGTM, thanks for the changes.

PR #797, opened 2024-06-14
[WD-11691] chore: minor UI change for vertical navigation resizing.

Kxiru:
## Done

- Minor scss changes to achieve the design stated in [this figma sheet](https://www.figma.com/design/t01jtaUMluSUP5fE5n4KzE/Navbar-Login-Status?node-id=0-1&t=xbuuiuOi0wtFyqty-0).

## QA

1. Run the LXD-UI:
    - N/A

## Screenshots

image

image

### DIFF
--- a/src/sass/_pattern_navigation.scss
+++ b/src/sass/_pattern_navigation.scss
@@ -55,7 +55,6 @@
   background: $colors--dark-theme--background-default;
   bottom: 0;
   padding: $spv--small $sph--x-large $spv--large $sph--large;
-  position: absolute;
   text-align: right;
   width: 100%;
 
@@ -91,20 +90,6 @@
   padding-bottom: 0;
 }
 
-.l-navigation.is-scroll {
-  .sidenav-bottom-ul.authenticated-nav {
-    bottom: initial;
-    margin-bottom: 0;
-    position: initial;
-  }
-
-  .sidenav-toggle-wrapper.authenticated-nav {
-    bottom: initial;
-    position: initial;
-    width: initial;
-  }
-}
-
 .l-navigation .p-panel__header {
   z-index: 1001;
 }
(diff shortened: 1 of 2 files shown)


src/sass/_pattern_navigation.scss
```diff
@@ -175,6 +175,24 @@
...
+  overflow: scroll;
+  -ms-overflow-style: none; /* Internet Explorer 10+ */
+  overflow-x: hidden;
+  scrollbar-width: none; /* Firefox */
+}
+
+.sidenav-top-container::-webkit-scrollbar {
+  display: none; /* Safari and Chrome */
+}
+
```
edlerd, line 189:
```suggestion
```
We should never hide the scrollbar, as users with those browsers expect it to be present.

edlerd (changes requested):
We shouldn't introduce a new concept for scrolling. We currently have [this function](https://github.com/canonical/lxd-ui/blob/main/src/components/Navigation.tsx#L89) to set the lower part of the nav to scroll. It works by measuring its height and comparing it to the total available height. This way we get a smooth transition on resize of the browser. It should also work by enabling scrolling on collapse/un-collapse of the accordions in the navigation (like for storage) in all possible height/width combinations.

So either remove that function and find a pure css solution or remove the css solution and rely on the existing one.

edlerd (changes requested):
See screen below:

1. We shouldn't hide the scrollbar in the navigation (see scrollbar is present on the right)
2. The top part of the nav is cut, even though there is enough space below. The height of the top part should be fully responsive, so it works in all possible screen widths/heights perfectly.

image

edlerd (changes requested):
On the demo server, the navigation doesn't scroll any more. I.e. when I shrink the browser windows height the lower part of the nav stays visible as required, but the middle part is cur with no way of reaching the hidden nav items. Is the demo not updated, or is it a problem with the changes?

Kxiru:
That being said, I have found that the top container is not scrollable. I added a fixing line for this in my last commit.

edlerd (changes requested):
QA looks good now, the linter has some errors, see below.

edlerd (approved):
This looks good, nice work :+1:

piperdeck (approved):
LGTM

PR #799, opened 2024-06-17
fix(network) show physical managed network state

edlerd:
## Done

- show physical managed network state in case it is available, show error message if it fails to load

## QA

1. Run the LXD-UI:
    - check detail page of a physical managed network

## Note

This is pending https://github.com/canonical/lxd/issues/13553 to improve the error message

### DIFF
--- a/src/pages/networks/NetworkDetailOverview.tsx
+++ b/src/pages/networks/NetworkDetailOverview.tsx
@@ -1,6 +1,6 @@
 import { FC, useEffect } from "react";
 import { Link, useParams } from "react-router-dom";
-import { Col, Row } from "@canonical/react-components";
+import { Col, Notification, Row } from "@canonical/react-components";
 import useEventListener from "@use-it/event-listener";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import ItemName from "components/ItemName";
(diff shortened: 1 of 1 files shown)


src/pages/networks/NetworkDetailOverview.tsx
```diff
@@ -96,12 +98,17 @@ const NetworkDetailOverview: FC<Props> = ({ network }) => {
...
-          <Col size={3}>
-            <h2 className="p-heading--5">Status</h2>
-          </Col>
-          <Col size={7}>
+      <Row className="section">
+        <Col size={3}>
+          <h2 className="p-heading--5">Status</h2>
+        </Col>
+        <Col size={7}>
+          {isNetworkStateError && (
```
mas-who, line 106: small nit: why not use a ternary here with `isNetworkStateError`?

mas-who:
QA and code LGTM, just one small nit comment, feel free to merge as is :)

src/pages/networks/NetworkDetailOverview.tsx
edlerd: I prefer the more verbose && and repeating the condition style here. Because the first conditional block has more than one or two lines. It makes the second conditional block easier to read and to understand. The condition of the ternary is so many lines up, and you need to keep it in mind to understand the condition for the else block. Repeating the condition tries to solve this.

mas-who: sounds good :+1:

mas-who (approved):

PR #805, opened 2024-06-25
fix(header) ensure instance name is unique and ot matched against previous name

edlerd:
## Done

- ensure instance name is unique and not matched against previous name on renaming

## QA

1. Run the LXD-UI:
    - rename an instance
    - change name to the same name as the current name. Previously it would cause a validation error, now it should be fine.

### DIFF
--- a/src/components/RenameHeader.tsx
+++ b/src/components/RenameHeader.tsx
@@ -35,14 +35,14 @@ const RenameHeader: FC<Props> = ({
   renameDisabledReason,
 }) => {
   const canRename = renameDisabledReason === undefined;
-  const enableRename = () => {
-    window.dispatchEvent(new Event("resize"));
+  const toggleRename = () => {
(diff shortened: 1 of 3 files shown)


mas-who:
code looks good and can confirm no validation error occurs once name is changed :slightly_smiling_face: 

I found a pre-existing behaviour that's a bit strange:
1. click on the rename header
2. change the instance name but don't save
3. click cancel
4. click on the rename header again and it shows the changed name in step 2.

Is this intended behaviour?

edlerd:
> 1. click on the rename header
> 2. change the instance name but don't save
> 3. click cancel
> 4. click on the rename header again and it shows the changed name in step 2.
> 
> Is this intended behaviour?

This is not ideal. Added a change to reset the value in the form on cancel.

mas-who:
LGTM, thanks for the fix :+1:

mas-who (approved):

Issue #814, opened 2024-07-12
Can't edit ipv4 dhcp


PVMChannel:
# Description

i just edited the ipv6 dhcp to false, i also cant figure out why my bridge network isnt working but thats probably unrelated

# Metadata

UI Version: 0.10
Path: /ui/project/default/network/lxdbr0/configuration/ipv6

# Stacktrace

```
Error: Minified React error #300; visit https://reactjs.org/docs/error-decoder.html?invariant=300 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Dy (https://url:8443/ui/assets/index-hQBqjpGf.js:44:19725)
    at pv (https://url:8443/ui/assets/index-hQBqjpGf.js:46:3139)
    at R1 (https://url:8443/ui/assets/index-hQBqjpGf.js:46:44804)
    at O1 (https://url:8443/ui/assets/index-hQBqjpGf.js:46:39766)
    at vT (https://url:8443/ui/assets/index-hQBqjpGf.js:46:39694)
    at xd (https://url:8443/ui/assets/index-hQBqjpGf.js:46:39547)
    at Ov (https://url:8443/ui/assets/index-hQBqjpGf.js:46:35914)
    at Cw (https://url:8443/ui/assets/index-hQBqjpGf.js:46:36717)
    at Pi (https://url:8443/ui/assets/index-hQBqjpGf.js:44:3274)
    at fo (https://url:8443/ui/assets/index-hQBqjpGf.js:46:37164)
```

edlerd:
Thanks for reporting this issue.

I was able to reproduce, and a PR with a fix is up above. This should be released within the next 1 to 2 weeks.

edlerd:
A fix was released with 5.21.2.

PR #815, opened 2024-07-12
fix(network) allow disabling of dhcp in ipv4/ipv6 network configuration

edlerd:
## Done

- fix(network) allow disabling of dhcp in ipv4/ipv6 network configuration
- i.e. on bridge networks. Avoid crash reported in #814

## QA

- test network configuration of a bridge, enable/disable dhcp for ipv4 and ipv6


mas-who (approved):

mas-who:
Thanks for the fix. LGTM :+1:

PR #816, opened 2024-07-12
fix(cert) ensure cert name is generated with only valid characters.

### LINKED ISSUE #809: Certificate generation results in invalid certificates when connecting using IPv6 without DNS
When generating a certificate, the servers hostname being passed as "organizationName": https://github.com/canonical/lxd-ui/blob/c67187fd6597e71a4003fb3b88d7375737849017/src/util/certificate.tsx#L29
When connecting to a host using its IPv6 address instead of IPv4 or DNS, the hostname is the IPv6-address resulting in a organizationName like `LXD UI [::1] (Browser Generated)`.

This parameter is marked as PRINTABLESTRING, which does not allow characters like `[` or `]`, as such the certificate is invalid and can not be imported to the trust store (`Error: x509: invalid RDNSequence: invalid attribute value: invalid PrintableString`). The problem does not occur when using IPv4 or a DNS hostname.

To avoid this, the ui should either use utf8String instead of printableString (maybe https://github.com/digitalbazaar/forge/issues/754#issuecomment-716656831) or filter out characters not allowed in PrintableString.

edlerd:
## Done

- ensure cert name is generated with only valid characters.

Fixes #809 WD-13031

### DIFF
--- a/src/util/certificate.spec.ts
+++ b/src/util/certificate.spec.ts
@@ -1,0 +1,23 @@
+import { sanitizeOrgName } from "./certificate";
+
+describe("sanitizeOrgName", () => {
+  it("replaces invalid chars from an ipv6 string", () => {
+    const result = sanitizeOrgName("LXD UI [::1] (Browser Generated)");
+
+    expect(result).toBe("LXD UI ::1 (Browser Generated)");
+  });
+
+  it("keeps a valid domain name", () => {
+    const result = sanitizeOrgName(
+      "LXD UI foo.example.com (Browser Generated)",
+    );
+
+    expect(result).toBe("LXD UI foo.example.com (Browser Generated)");
+  });
+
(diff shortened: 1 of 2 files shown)


Kxiru (commented):
LGTM!
Is it worth adding a test with invalid characters to make sure the string is valid, and so we know that if the logic changes in any way in the future or the test fails, we can trace it back to this?

mas-who (approved):

mas-who:
> LGTM! Is it worth adding a test with invalid characters to make sure the string is valid, and so we know that if the logic changes in any way in the future or the test fails, we can trace it back to this?

I think the unit tests added by David is sufficient for regression checks. Adding e2e may be tricky since we'd have to look at the downloaded certificate name somehow (potentially with fixtures). In any case, I don't think e2e will add a lot more value than the unit tests in this case

PR #822, opened 2024-07-23
[WD-13254] test: network forms

rubinaga:
## Done

- E2E tests for editing networks
- Fixed bug that didn't reset "edit network" form after clicking override and not making a change
- Removed unused parameter "placeholder" in "setTextArea" helper function

## QA

1. Run the LXD-UI:
    - To replicate the bug, edit a network, click override, do not make a change and submit the form. This is shown in the screenshots with the "dash" appearing after no changes being made. After refreshing the page the "dash" disappears
    - E2E tests need to be ran with playwright

## Screenshots
image
image

### DIFF
--- a/tests/helpers/network.ts
+++ b/tests/helpers/network.ts
@@ -1,6 +1,7 @@
 import { Page } from "@playwright/test";
 import { randomNameSuffix } from "./name";
 import { LxdNetworkType } from "types/network";
+import { activateAllTableOverrides } from "./configuration";
 
 export const randomNetworkName = (): string => {
   return `test-${randomNameSuffix()}`;
@@ -40,6 +41,7 @@ export const visitNetwork = async (page: Page, network: string) => {
   await page.goto("/ui/");
   await page.getByTitle("Networks (default)").click();
   await page.getByRole("link", { name: network }).first().click();
+  await page.getByTestId("tab-link-Configuration").click();
 };
 
 export const saveNetwork = async (page: Page, network: string) => {
@@ -48,17 +50,42 @@ export const saveNetwork = async (page: Page, network: string) => {
   await page.getByRole("button", { name: "Close notification" }).click();
 };
 
-export const editNetwork = async (page: Page, network: string) => {
-  await visitNetwork(page, network);
-  await page.getByTestId("tab-link-Configuration").click();
+export const editNetwork = async (page: Page) => {
   await page.getByRole("button", { name: "Edit network" }).click();
 };
 
-export const createNetworkForward = async (page: Page, network: string) => {
-  await visitNetwork(page, network);
+export const prepareNetworkTabEdit = async (
+  page: Page,
+  tabLocation: string,
+  networkName: string,
+) => {
+  await visitNetwork(page, networkName);
+  await page
+    .getByLabel("Network form navigation")
+    .getByText(tabLocation)
+    .click();
+  await editNetwork(page);
+  await activateAllTableOverrides(page);
+};
 
-  await page.getByTestId("tab-link-Configuration").click();
-  await page.getByRole("button", { name: "Edit network" }).click();
+export const visitNetworkConfiguration = async (page: Page, tab: string) => {
+  await page
+    .getByRole("button", {
+      name: "Advanced",
+    })
+    .click();
+  await page
+    .getByLabel("Network form navigation")
+    .getByText(tab, { exact: true })
+    .click();
+  await activateAllTableOverrides(page);
+};
+
+export const createNetworkForward = async (page: Page, network: string) => {
+  await createNetwork(page, network);
+  await visitNetwork(page, network);
+  await editNetwork(page);
+
   const networkSubnet = await page.inputValue("input#ipv4_address");
 
   const listenAddress = networkSubnet.replace("1/24", "1");
@@ -85,4 +112,6 @@ export const createNetworkForward = async (page: Page, network: string) => {
   await page
     .getByText(`:23,443-455 → ${targetAddress}:23,443-455 (tcp)`)
     .click();
+
+  await deleteNetwork(page, network);
 };
--- a/tests/networks.spec.ts
+++ b/tests/networks.spec.ts
@@ -1,151 +1,315 @@
 import { test } from "./fixtures/lxd-test";
 import {
+  activateAllTableOverrides,
+  checkAllOptions,
+  setTextarea,
+} from "./helpers/configuration";
+import {
   createNetwork,
   createNetworkForward,
   deleteNetwork,
   editNetwork,
+  prepareNetworkTabEdit,
   randomNetworkName,
-  saveNetwork,
   visitNetwork,
+  visitNetworkConfiguration,
 } from "./helpers/network";
 import {
-  activateOverride,
-  assertReadMode,
-  setOption,
-} from "./helpers/configuration";
+  getServerSettingValue,
+  updateSetting,
+  visitServerSettings,
+} from "./helpers/server";
 
 let network = randomNetworkName();
+let initialNorthboundNetworkValue: string = "";
 
-test.beforeAll(async ({ browser, browserName }) => {
-  // network names can only be 15 characters long
-  network = `${browserName.substring(0, 2)}-${network}`;
-  const page = await browser.newPage();
-  await createNetwork(page, network);
-  await page.close();
+test.describe("bridge type", () => {
+  test.beforeAll(async ({ browser, browserName }) => {
+    // network names can only be 15 characters long
+    network = `${browserName.substring(0, 2)}-${network}`;
+    const page = await browser.newPage();
+    await createNetwork(page, network);
+    await page.close();
+  });
+
+  test.afterAll(async ({ browser }) => {
+    const page = await browser.newPage();
+    await deleteNetwork(page, network);
+    await page.close();
+  });
+
+  test("configure main bridge network settings", async ({ page }) => {
+    await visitNetwork(page, network);
+    await editNetwork(page);
+
+    const DESCRIPTION = "A-new-description";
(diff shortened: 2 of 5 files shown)


tests/projects.spec.ts
```diff
@@ -92,14 +92,9 @@ test("project edit configuration", async ({ page, lxdVersion }) => {
...
-    page,
-    "Available networks",
-    "Enter network names",
-    "lxcbr0",
-  );
-  await setTextarea(page, "Network uplinks", "Enter network names", "lxcbr0");
-  await setTextarea(page, "Network zones", "Enter network zones", "foo,bar");
+  await setTextarea(page, "Available networks", "Enter network names");
+  await setTextarea(page, "Network uplinks", "Enter network names");
+  await setTextarea(page, "Network zones", "Enter network zones");
```
edlerd, line 97: We removed the placeholder in the function definition (3rd argument), but here you remove the 4th argument. This should be changed, so we remove the 3rd argument instead.

tests/networks.spec.ts
```diff
@@ -4,148 +4,370 @@ import {
...
+    const DHCP_RANGES = "192.168.1.100-192.168.1.150";
+    const OVN_RANGES = "10.141.1.50-10.141.1.99";
+    const DHCP_EXPIRY = "2h";
+
+    await prepareNetworkTabEdit(page, "IPv4", network);
+
+    await page
+      .getByRole("row")
+      .filter({
+        hasText: "Whether to allocate",
```
edlerd, line 101: This text is missing when testing against the LXD 5.0 backend. Either use a different text for it like "IPv4 DHCPtrue" or skip this bit of the test for the 5.0 backend. See other tests for how to identify the backend version from test code.

tests/networks.spec.ts
```diff
@@ -4,148 +4,378 @@ import {
...
+      .getByRole("textbox")
+      .inputValue();
+
+    await parent.getByRole("textbox").fill("foo");
+
+    await page
+      .getByRole("button", {
+        name: /save/i,
+      })
+      .click();
```
edlerd, line 229:
```suggestion
  await visitServerSettings(page);
  await updateSetting(page, "network.ovn.northbound_connection", "text", "foo");
```
We can use an existing abstraction for this. Also for the reset in after all below.

tests/networks.spec.ts
```diff
@@ -4,148 +4,378 @@ import {
...
+  test("set OVN bridge MTU and hardware address", async ({ page }) => {
+    await page
+      .getByRole("button", {
+        name: "Advanced",
+      })
+      .click();
+    await page
+      .getByLabel("Network form navigation")
+      .getByText("Bridge", { exact: true })
+      .click();
```
edlerd, line 306:
```suggestion
    await visitNetworkConfiguration(page, "Bridge");
```
This is duplicated quite a few times, maybe worth adding a helper function `visitNetworkConfiguration` Maybe we can even reuse the existing `prepareNetworkTabEdit`?

tests/networks.spec.ts
```diff
@@ -4,148 +4,378 @@ import {
...
+      .getByRole("button", {
+        name: "Advanced",
+      })
+      .click();
+    await page
+      .getByLabel("Network form navigation")
+      .getByText("IPv4", { exact: true })
+      .click();
+
+    for (const option of options) {
```
edlerd, line 344:
```suggestion

    const options = ["false", "true"];
    for (const option of options) {
```
Better define variables as close to their usage as possible to make this easier to read and maintain.

tests/networks.spec.ts
```diff
@@ -4,148 +4,378 @@ import {
...
+
+    await prepareNetworkTabEdit(page, "IPv4", network);
+
+    await setTextarea(page, "IPv4 OVN ranges", OVN_RANGES);
+    await setTextarea(page, "IPv4 gateway", NETWORK_CONSTANTS.IPV4_ADDRESS);
+    await activateOverride(page, "IPv4 routes Additional IPv4");
+    await page.getByLabel("IPv4 routes").fill(IPV4_ROUTES);
+    for (const option of options) {
+      await setOption(page, "IPv4 routes anycast", option);
+    }
```
edlerd, line 175: I am not a fan of repeating the for loop in every test case. Maybe add a helper `checkAllOptions(page, "name", options)` that we can use in all cases instead. The helper can have the for loop internally, so we have it exactly once and don't need to repeat it.

tests/networks.spec.ts
```diff
@@ -4,148 +4,378 @@ import {
...
+  test("set physical network IPv4 routes and gateway", async ({ page }) => {
+    const options = ["true", "false"];
+    const OVN_RANGES = "10.141.1.100-10.141.1.110";
+    const IPV4_ROUTES = "10.141.1.1";
+
+    await prepareNetworkTabEdit(page, "IPv4", network);
+
+    await setTextarea(page, "IPv4 OVN ranges", OVN_RANGES);
+    await setTextarea(page, "IPv4 gateway", NETWORK_CONSTANTS.IPV4_ADDRESS);
+    await activateOverride(page, "IPv4 routes Additional IPv4");
```
edlerd, line 171: Here we have the problem, that the help text is missing in LXD 5.0, similar to what was previously failing with the DHCP options.

edlerd (commented):
Thanks for adding the test, some ideas to simplify below.

tests/networks.spec.ts
edlerd: I forgot: We also need `getServerSettingValue` for the initial value.

tests/networks.spec.ts
edlerd: Thanks for adjusting this :+1:

edlerd (commented):
This is looking pretty good, thanks for the adjustments. A few smaller suggestions below.

tests/networks.spec.ts
```diff
@@ -1,151 +1,299 @@
...
+
+  test("configure OVN network IPv6 DHCP and L3 settings", async ({ page }) => {
+    await visitNetworkConfiguration(page, "IPv6");
+
+    const options = ["false", "true"];
+    for (const option of options) {
+      await page
+        .getByRole("combobox", { name: "IPv6 DHCP", exact: true })
+        .selectOption(option);
+    }
```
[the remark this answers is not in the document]
rubinaga: "IPv6 DHCP" also matches "IPv6 DHCP stateful". I haven't enforced "exact: true" at the function "checkAllOptions" since most helper functions don't have it.  That's the reason I made the exception and used a for loop.

tests/helpers/network.ts
```diff
@@ -48,17 +60,40 @@ export const saveNetwork = async (page: Page, network: string) => {
...
-  await page.getByTestId("tab-link-Configuration").click();
+export const editNetwork = async (page: Page) => {
   await page.getByRole("button", { name: "Edit network" }).click();
 };
 
+export const prepareNetworkTabEdit = async (
+  page: Page,
+  tabLocation: string,
+  networkName: string,
+  menu: string = "Network form navigation",
```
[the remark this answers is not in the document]
rubinaga: Good point, I thought I specified different menus when editing network / creating new network but it seems to work fine. I will remove this

tests/helpers/network.ts
```diff
@@ -1,6 +1,17 @@
...
+
+export const NETWORK_CONSTANTS = {
+  IPV4_ADDRESS: "10.141.1.1/24",
+  IPV6_ADDRESS: "2001:db8::1/64",
+  DESCRIPTION: "A-new-description",
+  MTU: "1300",
+  HARDWARE_ADDRESS: "00:11:22:33:44:55",
+  DNS_DOMAIN: "example.com",
+  DNS_SEARCH: "search.example.com",
+};
```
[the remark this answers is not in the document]
rubinaga: I agree, will change it up

tests/networks.spec.ts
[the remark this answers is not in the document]
edlerd: Right, makes sense. Thanks for clarifying.

tests/networks.spec.ts
```diff
@@ -1,151 +1,319 @@
...
-test("network forwards", async ({ page }) => {
-  await createNetworkForward(page, network);
+test.describe("OVN type", () => {
+  test.beforeAll(async ({ browser, browserName, lxdVersion }) => {
+    if (lxdVersion === "5.0-edge") {
+      return;
+    }
+
+    // network names can only be 15 characters long
+    network = `${browserName.substring(0, 2)}-${network}`;
```
edlerd, line 201:
```suggestion
```
We don't need the network name here.

tests/networks.spec.ts
```diff
@@ -1,151 +1,319 @@
...
-  await page.getByText("IPv4", { exact: true }).click();
-  await assertReadMode(page, "IPv4 DHCP expiry", "2h");
-
-  await page.getByText("IPv6", { exact: true }).click();
-  await assertReadMode(page, "IPv6 DHCP expiry", "3h");
-  await assertReadMode(page, "IPv6 DHCP stateful", "true");
+test.describe("physical type", () => {
+  test.beforeAll(async ({ browser, browserName }) => {
+    // network names can only be 15 characters long
+    network = `${browserName.substring(0, 2)}-${network}`;
```
edlerd, line 135:
```suggestion
```
Don't need the network name here.

edlerd (approved):
LGTM, seems also to be passing now. Thanks for adding and improving the tests :)

Two small nitpicks below, I think we don't need to redefine the network variable.

tests/networks.spec.ts
rubinaga: will remove it

tests/networks.spec.ts
rubinaga: will remove it here too

edlerd (approved):
All good, thanks for making this nice @rubinaga :+1:

PR #829, opened 2024-08-02
fix: invalidate query cache when creating custom storage volume [WD-13879]

mas-who:
## Done

- Previously we did not invalidate the query cache for custom volumes when creating a new one. This means when a user navigate back to an instance and try to attach the volume, they will not see the volume in the select modal unless if they refresh the page. This PR resolves this issue.

## QA

1. Run the LXD-UI:
    - Go to Instances -> select an instance to go to its detail page -> Configuration ->  Disk devices -> Attach disk divice (don't do anything)
    - Go to Volumes -> Create volume -> create a new volume -> then go back to the volume select modal in the previous step -> make sure that the new volume is in the list


lorumic (approved):
LGTM :+1:

PR #833, opened 2024-08-08
feat: [WD-13702] Add publish instance

Kxiru:
## Done

- Created PublishInstanceForm Component
- Created PublishInstanceBtn Component
- Added PIBtn to Instance detail page
- Added API call to `1.0/images/` to create a new Image.
- Added an e2e test for coverage.

## QA

1. Run the LXD-UI:
    - View an Instance detail page and publish an instance using the button.
    - Verify the instance has created an image by navigating to the image page.

## Screenshots
image
image
image

### DIFF
--- a/src/api/images.tsx
+++ b/src/api/images.tsx
@@ -114,3 +114,24 @@ export const createImageAlias = (
       .catch(reject);
   });
 };
+
+export const createImageFromInstance = (
+  instance: LxdInstance,
+  isPublic: boolean,
+): Promise<LxdOperationResponse> => {
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/images?project=${instance.project}`, {
+      method: "POST",
+      body: JSON.stringify({
+        public: isPublic,
+        source: {
+          name: instance.name,
+          type: "instance",
+        },
+      }),
+    })
+      .then(handleResponse)
+      .then(resolve)
+      .catch(reject);
+  });
+};
--- a/src/pages/instances/InstanceDetailHeader.tsx
+++ b/src/pages/instances/InstanceDetailHeader.tsx
@@ -19,6 +19,7 @@ import InstanceLink from "pages/instances/InstanceLink";
 import MigrateInstanceBtn from "./actions/MigrateInstanceBtn";
 import { useSettings } from "context/useSettings";
 import { isClusteredServer } from "util/settings";
+import CreateImageFromInstanceBtn from "./actions/CreateImageFromInstanceBtn";
 
 interface Props {
   name: string;
@@ -128,6 +129,7 @@ const InstanceDetailHeader: FC<Props> = ({
               {isClustered ? (
                 <MigrateInstanceBtn instance={instance} project={project} />
               ) : null}
+              <CreateImageFromInstanceBtn key="publish" instance={instance} />
               <DeleteInstanceBtn key="delete" instance={instance} />
             </>
           ) : null
--- a/src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
+++ b/src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
@@ -1,0 +1,46 @@
+import { FC } from "react";
+import { LxdInstance } from "types/instance";
+import { ActionButton } from "@canonical/react-components";
+import usePortal from "react-useportal";
+import CreateImageFromInstanceForm from "../forms/CreateImageFromInstanceForm";
+import { useInstanceLoading } from "context/instanceLoading";
+
+interface Props {
+  instance: LxdInstance;
+}
+
+const CreateImageFromInstanceBtn: FC<Props> = ({ instance }) => {
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const instanceLoading = useInstanceLoading();
+  const prohibitedStatuses = ["Error", "Frozen", "Running"];
+
+  const isDisabled =
+    prohibitedStatuses.includes(instance?.status) ||
+    Boolean(instanceLoading.getType(instance));
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <CreateImageFromInstanceForm
+            close={closePortal}
+            instance={instance}
+          />
+        </Portal>
+      )}
+      <ActionButton
+        appearance="default"
+        onClick={openPortal}
+        aria-label="Create image"
+        title={
+          isDisabled ? "Stop the instance to create an image" : "Create image"
+        }
+        disabled={isDisabled}
+      >
+        <span>Create Image</span>
+      </ActionButton>
+    </>
+  );
+};
+
+export default CreateImageFromInstanceBtn;
--- a/src/pages/instances/forms/CreateImageFromInstanceForm.tsx
+++ b/src/pages/instances/forms/CreateImageFromInstanceForm.tsx
@@ -1,0 +1,131 @@
+import { FC } from "react";
+import { LxdInstance } from "types/instance";
+import { useEventQueue } from "context/eventQueue";
+import { useFormik } from "formik";
+import { useToastNotification } from "context/toastNotificationProvider";
+import { createImageAlias, createImageFromInstance } from "api/images";
+import {
+  ActionButton,
+  Button,
+  Form,
+  Input,
+  Modal,
+} from "@canonical/react-components";
+import * as Yup from "yup";
+import { Link } from "react-router-dom";
+
+interface Props {
+  instance: LxdInstance;
+  close: () => void;
+}
+
+const CreateImageFromInstanceForm: FC<Props> = ({ instance, close }) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+
+  const notifySuccess = () => {
+    const created = (
+      <Link to={`/ui/project/${instance.project}/images`}>created</Link>
+    );
+    toastNotify.success(
+      <>
+        Image {created} from instance {instance.name}.
+      </>,
+    );
+  };
+
+  const formik = useFormik<{ alias: string; isPublic: boolean }>({
+    initialValues: {
+      alias: `from-instance-${instance.name}`,
+      isPublic: false,
+    },
+    validationSchema: Yup.object().shape({
+      alias: Yup.string(),
+    }),
+    onSubmit: (values) => {
+      const alias = values.alias;
+
+      createImageFromInstance(instance, values.isPublic)
+        .then((operation) => {
+          toastNotify.info(
+            <>
+              Creation of image from instance <b>{instance.name}</b> started.
+            </>,
+          );
+          close();
+          eventQueue.set(
+            operation.metadata.id,
+            (event) => {
+              if (alias) {
+                const fingerprint = event.metadata.metadata?.fingerprint ?? "";
+                void createImageAlias(fingerprint, alias).then(notifySuccess);
+              } else {
+                notifySuccess();
+              }
+            },
+            (msg) => {
+              toastNotify.failure(
+                `Image creation from instance "${instance.name}" failed.`,
+                new Error(msg),
+              );
+            },
+          );
+        })
+        .catch((e) => {
+          toastNotify.failure(
+            `Image creation from instance "${instance.name}" failed.`,
+            e,
+          );
+        });
+    },
+  });
+
+  return (
+    <Modal
+      close={close}
+      className="create-image-from-instance-modal"
+      title="Create image from instance"
+      buttonRow={
+        <>
+          <Button
+            appearance="base"
+            className="u-no-margin--bottom"
+            type="button"
+            onClick={close}
+          >
+            Cancel
+          </Button>
+          <ActionButton
+            appearance="positive"
+            className="u-no-margin--bottom"
+            loading={formik.isSubmitting}
+            disabled={!formik.isValid}
+            onClick={() => void formik.submitForm()}
+          >
+            Create image
+          </ActionButton>
+        </>
+      }
+    >
+      <Form onSubmit={formik.handleSubmit}>
+        <Input type="text" label="Instance" value={instance.name} disabled />
+        <Input
+          {...formik.getFieldProps("alias")}
+          type="text"
+          label="Alias"
+          error={formik.touched.alias ? formik.errors.alias : null}
+        />
+        <Input
+          {...formik.getFieldProps("isPublic")}
+          type="checkbox"
+          label="Make the image publicly available"
+          error={formik.touched.isPublic ? formik.errors.isPublic : null}
+        />
+        {/* hidden submit to enable enter key in inputs */}
+        <Input type="submit" hidden value="Hidden input" />
+      </Form>
+    </Modal>
+  );
+};
+
+export default CreateImageFromInstanceForm;
--- a/src/sass/_instance_detail_page.scss
+++ b/src/sass/_instance_detail_page.scss
@@ -7,3 +7,9 @@
     margin: 0 $sph--large;
   }
 }
+
+.create-image-from-instance-modal {
+  .p-modal__dialog {
+    width: 35rem;
+  }
+}
--- a/tests/publish-instance.spec.ts
+++ b/tests/publish-instance.spec.ts
@@ -1,0 +1,26 @@
+import { test } from "./fixtures/lxd-test";
+import { deleteAllImages } from "./helpers/images";
+import {
+  createInstance,
+  deleteInstance,
+  randomInstanceName,
+  visitInstance,
+} from "./helpers/instances";
+
+test("Publish an image from an instance", async ({ page }) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+
+  await page.goto("/ui/");
+  await visitInstance(page, instance);
+  await page.getByRole("button", { name: "Create image" }).click();
+  await page
+    .getByRole("dialog", { name: "Create image from instance" })
+    .getByRole("button", { name: "Create image" })
+    .click();
+
+  await page.waitForSelector(`text=Image created from instance ${instance}.`);
+
+  await deleteAllImages(page);
+  await deleteInstance(page, instance);
+});
--- a/src/components/SelectableMainTable.tsx
+++ b/src/components/SelectableMainTable.tsx
@@ -25,6 +25,7 @@ interface SelectableMainTableProps {
   disableSelect?: boolean;
   onToggleRow?: (rowName: string) => void;
   hideContextualMenu?: boolean;
+  defaultSortKey?: string;
 }
 
 type Props = SelectableMainTableProps & MainTableProps;
@@ -42,6 +43,7 @@ const SelectableMainTable: FC<Props> = ({
   disableSelect = false,
   onToggleRow,
   hideContextualMenu,
+  defaultSortKey,
   ...props
 }: Props) => {
   const [currentSelectedIndex, setCurrentSelectedIndex] = useState<number>();
@@ -211,6 +213,8 @@ const SelectableMainTable: FC<Props> = ({
       expanding={true}
       headers={headersWithCheckbox}
       rows={[...rowsWithCheckbox]}
+      defaultSort={defaultSortKey}
+      defaultSortDirection="descending"
       {...props}
     />
   );
--- a/src/pages/images/ImageList.tsx
+++ b/src/pages/images/ImageList.tsx
@@ -279,6 +279,7 @@ const ImageList: FC = () => {
                 headers={headers}
                 sortable
                 className="image-table"
+                defaultSortKey="uploaded_at"
                 emptyStateMsg="No images found matching this search"
                 onUpdateSort={updateSort}
                 selectedNames={selectedNames}


Kxiru:
Hello potential reviewers!
Within this PR, we are able to publish an instance in order to create an image. However, I have found that one is not able to publish more than one image from the same instance at a time (the previous one has to be deleted). I wonder if this is default behaviour, and welcome any suggestions to address this whilst I look into the matter further :).

mas-who:
> Hello potential reviewers! Within this PR, we are able to publish an instance in order to create an image. However, I have found that one is not able to publish more than one image from the same instance at a time (the previous one has to be deleted). I wonder if this is default behaviour, and welcome any suggestions to address this whilst I look into the matter further :).
> 
> image

I think it is fine to rely on the backend to inform us about an existing image, the error message is clear in this case.

mas-who:
It seems when publishing an image from a VM instance, the operation stops and does not progress. Not sure if this is an issue specifically on the demo server, please would you investigate this locally as well.

Screenshot from 2024-08-12 14-26-50

src/pages/instances/forms/PublishInstanceForm.tsx
```diff
@@ -0,0 +1,126 @@
...
+} from "@canonical/react-components";
+import * as Yup from "yup";
+import { Link } from "react-router-dom";
+
+interface Props {
+  instance: LxdInstance;
+  close: () => void;
+}
+
+const PublishInstanceForm: FC<Props> = ({ instance, close }) => {
```
mas-who, line 22:
```suggestion
const CreateImageFromInstanceForm: FC<Props> = ({ instance, close }) => {
```

src/pages/instances/forms/PublishInstanceForm.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+
+  const notifySuccess = () => {
+    const published = (
+      <Link to={`/ui/project/${instance.project}/images`}>published</Link>
+    );
+    toastNotify.success(<>Image {published} from instance.</>);
```
mas-who, line 30: let's include the instance name in the toast message here

src/pages/instances/forms/PublishInstanceForm.tsx
```diff
@@ -0,0 +1,126 @@
...
+            Cancel
+          </Button>
+          <ActionButton
+            appearance="positive"
+            className="u-no-margin--bottom"
+            loading={formik.isSubmitting}
+            disabled={!formik.isValid}
+            onClick={() => void formik.submitForm()}
+          >
+            Create Image
```
mas-who, line 100:
```suggestion
            Create image
```

mas-who (commented):
Thanks this looks pretty good already, left some feedback :+1:

src/pages/instances/actions/PublishInstanceBtn.tsx
```diff
@@ -0,0 +1,33 @@
...
+          <PublishInstanceForm close={closePortal} instance={instance} />
+        </Portal>
+      )}
+      <ActionButton
+        appearance={"default"}
+        onClick={openPortal}
+        aria-label="Publish instance"
+        title="Publish instance"
+      >
+        <span>Publish</span>
```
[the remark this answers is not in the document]
mas-who: actually, `span` is needed for button styling to apply correctly.

src/api/images.tsx
```diff
@@ -114,3 +114,25 @@ export const createImageAlias = (
       .catch(reject);
   });
 };
+
+export const createImageFromInstance = (
+  //Also known as create Image from an Instance.
```
mas-who, line 119: this comment can be removed now

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -27,6 +28,7 @@ const DeleteInstanceBtn: FC<Props> = ({ instance }) => {
 
   const handleDelete = () => {
     setLoading(true);
+    setDeleting(true);
```
mas-who, line 31: we can replace `isLoading` with `isDeleting` as a prop. no need to maintain state here and inside the parent anymore

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -119,7 +121,19 @@ const InstanceDetailHeader: FC<Props> = ({
         }
         controls={
           instance ? (
-            <DeleteInstanceBtn key="delete" instance={instance} />
+            <>
+              <PublishInstanceBtn
+                key="publish"
+                instance={instance}
+                isDeleting={isDeleting}
+                isStopping={instance?.status == "Stopped"}
```
mas-who, line 129: this can be done inside the component, we don't need this as a prop.

src/pages/instances/actions/PublishInstanceBtn.tsx
```diff
@@ -0,0 +1,47 @@
...
+          />
+        </Portal>
+      )}
+      <ActionButton
+        appearance="default"
+        onClick={openPortal}
+        aria-label="Publish instance"
+        title="Publish instance"
+        disabled={
+          instance?.status == ("Running" || "Restarting") ||
```
mas-who, line 36: did you test these statuses? From my testing, I can publish an instance while it's running, not sure about the other ones though. You should double check all statuses from `LxdInstanceStatus`, I think at least "Starting", "Stopping", "Restarting" and "Freezing" should result in the button being disabled, but please verify the other statuses as well.

You should also use `useInstanceLoading` to determine the status of an instance instead since it is managed via a global state provider.

src/pages/instances/actions/PublishInstanceBtn.tsx
```diff
@@ -0,0 +1,47 @@
...
+          !isStopping ||
```
mas-who, line 37: instead doing these `| |` conditions, rather define a list with applicable instance statuses and do something like `statuses.includes(instance.status)`, you should come up with a better variable naming for the list though. The logic `!isStopping` here makes `instance?.status == ("Running" || "Restarting")` redundant btw.

src/pages/instances/actions/PublishInstanceBtn.tsx
```diff
@@ -0,0 +1,47 @@
...
+import usePortal from "react-useportal";
+import CreateImageFromInstanceForm from "../forms/CreateImageFromInstanceForm";
+
+interface Props {
+  instance: LxdInstance;
+  isDeleting: boolean;
+  isStopping: boolean;
+}
+
+const PublishInstanceBtn: FC<Props> = ({
```
mas-who, line 13: let's also rename this component as well as file name i.e. `CreateInstanceFrom...`

mas-who (commented):
we can improve on how we determine instance statuses a bit here

mas-who:
@Kxiru let's also add a e2e test for this feature i.e. Create instance -> Navigate to instance detail page -> publish image -> check notification that it's successful

src/pages/instances/actions/PublishInstanceBtn.tsx
Kxiru: Yes, I did test them, and I cannot publish when running. I found all of the Instance statuses from another file as well but I will revisit this section and use useInstanceLoading.

src/pages/instances/actions/PublishInstanceBtn.tsx
Kxiru: If isStopping covers Running, Restarting and Frozen (as it seems to be in my case), then surely the only cases that need to be addressed are isStopping and isDeleting.

src/pages/instances/actions/PublishInstanceBtn.tsx
mas-who: I think it would be a combination between `instance.status` and status returned by using `useInstanceLoading` (which represents instance transient states)

Kxiru: image
In this file, RestartInstanceBtn, both useInstanceLoading and instance.status are used to determine whether an instance isLoading or not. Does that not mean that either application is valid?

src/pages/instances/actions/PublishInstanceBtn.tsx
mas-who: given and instance may be in the below states:
```
export type LxdInstanceStatus =
  | "Error"
  | "Freezing"
  | "Frozen"
  | "Restarting"
  | "Running"
  | "Starting"
  | "Stopped"
  | "Stopping";
  ```
  Your `!isStopping` condition would disable the button for every other status excepted for `Stopped`. This could be correct but we don't know if `Error`, `Frozen`, `Running` allows for image publish. I tried publishing an image while the instance is running and it worked, not sure about `Error` and `Frozen`. For `Freezing`, `Restarting`, `Starting` and `Stopping` we should disable the button since those are transient states for the instance

File: src/pages/instances/actions/PublishInstanceBtn.tsx (hunk 8 above)
Comment (mas-who) [reply src/pages/instances/actions/PublishInstanceBtn.tsx:36]: no because `instance.status` is fixed and its data will not change unless you make another API call. `useInstanceLoading` returns global states for the transient instance statuses i.e. `Starting`, `Freezing`, `Stopping` and `Restarting` which is dynamic. In fact I think the code is actually a bit wrong because those transient states does not actually exist for an instance in LXD, they exist only in the UI.

Comment (Kxiru) [reply src/pages/instances/actions/PublishInstanceBtn.tsx:36]: Given you said that you believe the solution will be a combo between instance.status and useInstanceLoading, are you insinuating I should replicate the above code?

Comment (mas-who) [reply src/pages/instances/actions/PublishInstanceBtn.tsx:36]: yeah pretty much, use `instanceLoading` for the transient statuses comparison, then use `instance.status` for the lxd instance statuses comparison. I just tested publishing an instance using the CLI while it is running, it actually failed so I'm not sure what happened yesterday when I tried it in the UI. So I guess `Running` should result in the button being disabled as well.

Comment (Kxiru) [reply src/pages/instances/actions/PublishInstanceBtn.tsx:36]: Nevermind, I think the new implementation I have addressed this.

File: tests/publish-instance.spec.ts [hunk 11]
```diff
@@ -0,0 +1,37 @@
...
+import {
+  createInstance,
+  deleteInstance,
+  randomInstanceName,
+  visitInstance,
+} from "./helpers/instances";
+
+let instance = randomInstanceName();
+
+test.beforeAll(async ({ browser, browserName }) => {
```
mas-who, line 12: there is no need to use `beforeAll` and `afterAll` hooks. Since there is only one test, you can just add the logic in the test directly

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,47 @@
...
+import CreateImageFromInstanceForm from "../forms/CreateImageFromInstanceForm";
+
+interface Props {
+  instance: LxdInstance;
+  isDeleting: boolean;
+}
+
+const CreateImageFromInstanceBtn: FC<Props> = ({ instance, isDeleting }) => {
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const prohibitedStatuses = [
```
mas-who, line 14: should use `useInstanceLoading` for `Starting`, `Stopping`, `Freezing` and `Restarting`

mas-who (commented):
almost there, some small fixes

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
Kxiru: I am not sure how you wish to implement this..

mas-who: Your implementation is good :)

tests/publish-instance.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  createInstance,
+  deleteInstance,
+  randomInstanceName,
+  visitInstance,
+} from "./helpers/instances";
+
+let instance = randomInstanceName();
+
+test("Publish an image from an instance", async ({ browserName, page }) => {
+  instance = `${browserName}-${instance}`;
```
mas-who, line 13:
```suggestion
  const instance = randomInstanceName();
```

tests/publish-instance.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+import {
+  createInstance,
+  deleteInstance,
+  randomInstanceName,
+  visitInstance,
+} from "./helpers/instances";
+
+let instance = randomInstanceName();
+
+test("Publish an image from an instance", async ({ browserName, page }) => {
```
mas-who, line 12:
```suggestion
test("Publish an image from an instance", async ({ page }) => {
```

tests/publish-instance.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { test } from "./fixtures/lxd-test";
+import { deleteAllImages } from "./helpers/images";
+import {
+  createInstance,
+  deleteInstance,
+  randomInstanceName,
+  visitInstance,
+} from "./helpers/instances";
+
+let instance = randomInstanceName();
```
mas-who, line 10: remove this

tests/publish-instance.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  await visitInstance(page, instance);
+  await page.getByRole("button", { name: "Publish" }).click();
+  await page
+    .getByRole("dialog", { name: "Create image from instance" })
+    .getByRole("button", { name: "Create image" })
+    .click();
+
+  await page.waitForSelector(`text=Image published from instance ${instance}.`);
+
+  deleteAllImages;
```
mas-who, line 24: this isn't being invoked, we need to call this to remove all the image created during this test

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+interface Props {
+  close: () => void;
+  project: string;
+}
+
+const ImportImageFromTarballForm: FC<Props> = ({ close, project }) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+
+  //   const notifySuccess = () => {
```
mas-who, line 29: something happened in this file, seems like a bunch of code got commented out?

Kxiru: This isn't supposed to be in this PR. I'm working on it. Youre so fast :joy:

tests/publish-instance.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  await page
+    .getByRole("dialog", { name: "Create image from instance" })
+    .getByRole("button", { name: "Create image" })
+    .click();
+
+  await page.waitForSelector(`text=Image published from instance ${instance}.`);
+
+  await deleteAllImages(page);
+  await deleteInstance(page, instance);
+  await page.close();
```
mas-who, line 26: last comment for the day, you don't need to call `page.close()` here since we are using standard `test` functionality here.

mas-who:
LGTM :+1: please make the last change for the tests and ask @piperdeck for a design review.

piperdeck:
Could we make this modal a little bit wider, just to give it more room to breathe?

Here I have it at 35rem wide:

Frame 2

Other than that, it looks great!

piperdeck (changes requested):
Left a comment about modal being a bit too narrow

Kxiru:
> Some observations
> 
> * when the instance is running, it is unclear why the "publish" button is disabled. See the delete button, which gives an explanation on hover for an improvement
> * the "publish" copy on the button might be unclear. I think "Create image" might be better, it matches the modal header and gives a better description of the functionality in my opinion. Though this is a bit longer, so ymmw.
> * when clicking on the link in the "Image published from instance ample-scam." success notification, the image appears on the bottom, hard to sport the new one. I think we should sort the list of images by newest upload date, so the new one always appears on the top.

Thanks for your comments, David. I've actioned those in the latest commit amend. Do you mind having a look and telling me what you think?

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const prohibitedStatuses = ["Error", "Frozen", "Running"];
+
+  const getHoverText = () => {
+    if (
+      prohibitedStatuses.includes(instance?.status) ||
+      Boolean(instanceLoading.getType(instance))
+    ) {
+      return "Stop the instance to publish it";
+    }
+    return "Delete instance";
```
edlerd, line 25:
```suggestion
    return "Create image";
```

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const instanceLoading = useInstanceLoading();
+  const prohibitedStatuses = ["Error", "Frozen", "Running"];
+
+  const getHoverText = () => {
+    if (
+      prohibitedStatuses.includes(instance?.status) ||
+      Boolean(instanceLoading.getType(instance))
+    ) {
+      return "Stop the instance to publish it";
```
edlerd, line 23:
```suggestion
      return "Stop the instance to create an image from it";
```

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,55 @@
...
+        onClick={openPortal}
+        aria-label="Publish instance"
+        title={getHoverText()}
+        disabled={
+          prohibitedStatuses.includes(instance?.status) ||
+          Boolean(instanceLoading.getType(instance)) ||
+          isDeleting
+        }
+      >
+        <span>Publish</span>
```
edlerd, line 49:
```suggestion
        <span>Create image</span>
```

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,55 @@
...
+          <CreateImageFromInstanceForm
+            close={closePortal}
+            instance={instance}
+          />
+        </Portal>
+      )}
+      <ActionButton
+        appearance="default"
+        onClick={openPortal}
+        aria-label="Publish instance"
```
edlerd, line 41:
```suggestion
        aria-label="Create image"
```

src/pages/instances/actions/DeleteInstanceBtn.tsx
edlerd, line 1: I would keep the deleting state local to keep things simple. Yes, there is the edge case when an instance is deleting then the image creation process could be started and will likely fail, but I think this is very obvious for a user to understand and we don't need to disable the publish button in case the instance is deleting.

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,55 @@
...
+import usePortal from "react-useportal";
+import CreateImageFromInstanceForm from "../forms/CreateImageFromInstanceForm";
+import { useInstanceLoading } from "context/instanceLoading";
+
+interface Props {
+  instance: LxdInstance;
+  isDeleting: boolean;
+}
+
+const CreateImageFromInstanceBtn: FC<Props> = ({ instance, isDeleting }) => {
```
edlerd, line 13:
```suggestion
const CreateImageFromInstanceBtn: FC<Props> = ({ instance }) => {
```
Probably can keep the interface simple.

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -34,6 +35,7 @@ const InstanceDetailHeader: FC<Props> = ({
   const navigate = useNavigate();
   const toastNotify = useToastNotification();
   const controllerState = useState<AbortController | null>(null);
+  const [isDeleting, setDeleting] = useState(false);
```
edlerd, line 38:
```suggestion
```
Probably better to keep this state local.

src/pages/instances/forms/CreateImageFromInstanceForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const CreateImageFromInstanceForm: FC<Props> = ({ instance, close }) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+
+  const notifySuccess = () => {
+    const published = (
+      <Link to={`/ui/project/${instance.project}/images`}>published</Link>
+    );
```
edlerd, line 29:
```suggestion
    const created = (
      <Link to={`/ui/project/${instance.project}/images`}>created</Link>
    );
```
Maybe be a bit more neutral here, as the published flag might be false on the created image?

src/pages/instances/forms/CreateImageFromInstanceForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+    toastNotify.success(
+      <>
+        Image {published} from instance {instance.name}.
+      </>,
+    );
+  };
+
+  const formik = useFormik<{ alias: string; isPublic: boolean }>({
+    initialValues: {
+      alias: "",
```
edlerd, line 39:
```suggestion
      alias: `from-instance-${instance.name}`,
```
I think we should use a default alias with the instance name here.

src/pages/instances/forms/CreateImageFromInstanceForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+            e,
+          );
+        });
+    },
+  });
+
+  return (
+    <Modal
+      close={close}
+      className="publish-instance-modal"
```
edlerd, line 86:
```suggestion
      className="create-image-from-instance-modal"
```

src/sass/_instance_detail_page.scss
```diff
@@ -7,3 +7,9 @@
     margin: 0 $sph--large;
   }
 }
+
+.publish-instance-modal {
```
edlerd, line 11:
```suggestion
.create-image-from-instance-modal {
```

edlerd (changes requested):
Thanks for the changes. Did a very detailed code review and found mostly copy suggestions. One request is a bit more involved: Can we keep the is deleting state local?

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const isDisabled = (): boolean => {
+    if (
+      prohibitedStatuses.includes(instance?.status) ||
+      Boolean(instanceLoading.getType(instance))
+    ) {
+      return true;
+    }
+    return false;
+  };
```
edlerd, line 25:
```suggestion
  const isDisabled = prohibitedStatuses.includes(instance?.status) ||
      Boolean(instanceLoading.getType(instance))
```
No need for a function here I think.

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,52 @@
...
+            instance={instance}
+          />
+        </Portal>
+      )}
+      <ActionButton
+        appearance="default"
+        onClick={openPortal}
+        aria-label="Create image"
+        title={
+          isDisabled() ? "Stop the instance to create an image" : "Create image"
```
edlerd, line 42:
```suggestion
          isDisabled ? "Stop the instance to create an image" : "Create image"
```

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -0,0 +1,52 @@
...
+        }
+        disabled={isDisabled()}
```
edlerd, line 44:
```suggestion
        disabled={isDisabled}
```

edlerd (dismissed):
LGTM, tiny simplification, I was a bit ambiguous before.

edlerd (dismissed):
LGTM, thanks for adding and improving the "create image from an instance" feature, and covering it with a test!

piperdeck (approved):
Looks good to me, great work!

PR #840, opened 2024-08-12
feat: bulk migrate instances [WD-13700]

mas-who:
## Done

- Moved the instance migration button from within the edit config page to instance details header
- Enabled bulk instance migration feature on instance list if lxd is clustered.

## QA

1. Run the LXD-UI:
In order to QA this PR, you will need to setup a LXD cluster with the following steps:
      2.1 create and start 3 VM instances (name them `node-1`, `node-2` and `node-3` for easier reference)
      2.2 for each instance, navigate to their detail page and access their terminals respectively
      2.3 in each instance's terminal, run `snap refresh lxd --channel=latest/edge`. Wait for lxd to update to the latest version i.e. 6.1.
      2.3 in the terminal for `node-1`, run `lxd init`. Then answer the question exactly like the example terminal output as shown below (use default values for empty inputs):
      ```
      Would you like to use LXD clustering? (yes/no) [default=no]: yes
      What IP address or DNS name should be used to reach this server? [default=10.147.196.215]: 
      Are you joining an existing cluster? (yes/no) [default=no]: no
      What member name should be used to identify this server in the cluster? [default=node-1]: 
      Do you want to configure a new local storage pool? (yes/no) [default=yes]: 
      Name of the storage backend to use (dir, lvm, zfs, btrfs) [default=zfs]: 
      Create a new ZFS pool? (yes/no) [default=yes]: 
      Would you like to use an existing empty block device (e.g. a disk or partition)? (yes/no) [default=no]: 
      Size in GiB of the new loop device (1GiB minimum) [default=5GiB]: 
      Do you want to configure a new remote storage pool? (yes/no) [default=no]: 
      Would you like to connect to a MAAS server? (yes/no) [default=no]: 
      Would you like to configure LXD to use an existing bridge or host interface? (yes/no) [default=no]: 
      Would you like to create a new Fan overlay network? (yes/no) [default=yes]: 
      What subnet should be used as the Fan underlay? [default=auto]: 
      Would you like stale cached images to be updated automatically? (yes/no) [default=yes]: 
      Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]: 
   ```
    2.4 Now join `node-2` and `node-3` to the cluster by:
    2.4.1 running `lxd init` in the terminals for `node-2` and `node-3` and answer the questions like the example shown below up until the "Please provide join token" question:
   ```
   Would you like to use LXD clustering? (yes/no) [default=no]: yes
   What IP address or DNS name should be used to reach this server? [default=10.147.196.240]: 
   Are you joining an existing cluster? (yes/no) [default=no]: yes
   Please provide join token: 
   ```
   2.4.2 open a terminal for `node-1` without closing the terminal for `node-2` and `node-3`. Inside the terminal generate join tokens for `node-2` and `node-3` with `lxc cluster add node-2` and `lxc cluster add node-3` respectively. Copy and paste the join tokens and paste it into the terminals for `node-2` and `node-3` and press enter. Then answer the rest of the questions for `node-2` and `node-3` like the example shown below:
   ```
   Please provide join token: eyJzZXJ2ZXJfbmFtZSI6Im5vZGUtNSIsImZpbmdlcnByaW50IjoiY2ZjZGRhZWU5MGVhYzg5NjY1Y2I2ODc3MGQwNTAxYzVhODVhOGUzMTlhYmZjNzZkODkzNjBiNDM3NjJmNzVkZiIsImFkZHJlc3NlcyI6WyIxMC4xNDcuMTk2LjIxNTo4NDQzIl0sInNlY3JldCI6IjBkMjI4NTcwZWMxMjM5ZTU4ZWQ5NTRkZWM1YjI5MjY4ODIzYzMzYmU4NzJmYjNiYzY4MTk3Y2Q0MWE4YjhhMTQiLCJleHBpcmVzX2F0IjoiMjAyNC0wOC0xM1QxMzozMzo0My40NzczMzgyOThaIn0=
   All existing data is lost when joining a cluster, continue? (yes/no) [default=no] yes
   Choose "zfs.pool_name" property for storage pool "local": 
   Choose "size" property for storage pool "local": 
   Choose "source" property for storage pool "local": 
   Would you like a YAML "lxd init" preseed to be printed? (yes/no) [default=no]: 
   ```
   2.5 Once the cluster is setup, set `LXD_UI_BACKEND_IP` to be the ipv4 address for `node-1`. Then start the UI with `dotrun`. Your UI should be displaying the lxd installation for `node-1` which is a clustered system.
   2.6 QA with the following steps:
   2.6.1 Create a few instances first
   2.6.2 Navigate to a instance's detail page, you should see the "Migrate" button at the top of the page. Click it and it should display a modal showing you available cluster members to which this instance can be migrated. Select a cluster member and ensure migration takes place successfully.
   2.6.3 On the instance list page, select multiple instances. You should see the migrate button at the top of the page. Click it and you will see the same modal displayed as the previous QA step. Select a cluster member and ensure migration takes place successfully for all selected instances. Note if an instance already exist on the target cluster member, it will result in a no-op for that instance.

#### Instance detail page
Screenshot from 2024-08-13 12-43-43
Screenshot from 2024-08-13 12-44-26
Screenshot from 2024-08-13 12-44-45
#### Instance list page
Screenshot from 2024-08-13 12-45-42
Screenshot from 2024-08-13 12-46-04
Screenshot from 2024-08-13 12-46-20


piperdeck:
After migrating instances, let's have it so that all the instances stay selected after the migration operation completes, regardless of whether the individual instance was already on the target node

piperdeck:
CleanShot 2024-08-14 at 10 32 32

Typo alert

mas-who:
@piperdeck I have made the following adjustments.

1. The migration form now uses checkboxes instead of the select buttons
Screenshot from 2024-08-15 11-47-40

2. After migration the instances remain selected. During migration the selected instances are disabled in the instance list table.

3. Fixed up typo that you spotted

4. In the instance detail page, if migration errors out, the loading state gets cleared correctly now. Also if we are in the process of Starting, Freezing, Restarting or Stopping an instance, the migrate button is disabled.

edlerd (changes requested):
I think we should simplify the interaction on the modal:

1. Change the title of the modal to "Choose the target for the 5 selected instances".
2. Remove the button row (cancel, migrate) from the modal
3. Add "migrate" buttons next to each row (same as in the "select" buttons on image selection on instance creation)

We need an additional step after the user selected the target, as some instances might be running and can't be migrated if live-migration is not possible. Also, some instances might be already on that host. So we should give a review step summarizing how many instances are going to be migrated and why and how many will be ignored.

mas-who:
> I think we should simplify the interaction on the modal:
> 
> 1. Change the title of the modal to "Choose the cluster member for the 5 selected instances".
> 2. Remove the button row (cancel, migrate) from the modal
> 3. Add "select" buttons next to each row (same as in the "select" buttons on image selection on instance creation)
> 
> We might need an additional step after the user selected the target, as some instances might be running and can't be migrated if live-migration is not possible. Also, some instances might be already on that host. So we should give a review step summarizing how many instances are going to be migrated and why and how many will be ignored.

I had a chat with @piperdeck about this, the button row was included so that the process isn't too sudden. I think an extra review step would achieve the same, and is nicer to show some summary. Currently we don't have any design for this though, should we just have a follow up modal with copy? I'm not sure if we should block live migration in the UI, is it possible to determine if live migration is possible from one of the API endpoints?

edlerd:
> Currently we don't have any design for this though, should we just have a follow up modal with copy?

I think we can make this up in close alignment to the bulk start/stop modal summaries.

> I'm not sure if we should block live migration in the UI, is it possible to determine if live migration is possible from one of the API endpoints?

Can you ask on the lxd channel if and how we can determine if an instance can be live migrated? I think we can find out from its properties. I agree we should not block it. If we can't find out if possible also fine to try and fail the instances that couldn't be migrated. I think that is better than blocking live migration completely.

mas-who:
> > Currently we don't have any design for this though, should we just have a follow up modal with copy?
> 
> I think we can make this up in close alignment to the bulk start/stop modal summaries.
> 
> > I'm not sure if we should block live migration in the UI, is it possible to determine if live migration is possible from one of the API endpoints?
> 
> Can you ask on the lxd channel if and how we can determine if an instance can be live migrated? I think we can find out from its properties. I agree we should not block it. If we can't find out if possible also fine to try and fail the instances that couldn't be migrated. I think that is better than blocking live migration completely.

I've made the changes to mimic the instance action confirmation modal. Please have a look and let me know what you think? Will see what the lxd team says regarding the question about figuring out if live migration is possible.

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  if (isLoading) {
+    return <Loader />;
+  }
+
+  const modalTitle = selectedMember
+    ? "Confirm migration"
+    : isBulkMigration
+      ? `Choose the cluster member for the ${instances.length} selected ${pluralize("instance", instances.length)}`
```
edlerd, line 137:
```suggestion
      ? `Choose target cluster member for ${instances.length} selected ${pluralize("instance", instances.length)}`
```
Simpler?

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      : `Choose the cluster member for instance ${instances[0].name}`;
```
edlerd, line 138:
```suggestion
      : `Choose target cluster member for instance ${instances[0].name}`;
```

src/pages/instances/InstanceList.tsx
```diff
@@ -94,6 +99,7 @@ const InstanceList: FC = () => {
   const [sizeHidden, setSizeHidden] = useState<string[]>([]);
   const [selectedNames, setSelectedNames] = useState<string[]>([]);
   const [processingNames, setProcessingNames] = useState<string[]>([]);
+  const [migratingInstances, setMigratingInstances] = useState<string[]>([]);
```
edlerd, line 102: I think we should not introduce additional local state for this. Can we rely on the global `instanceLoading` state instead, to check if an instance is currently migrating?

src/pages/instances/InstanceStatusIcon.tsx
```diff
@@ -21,6 +21,7 @@ const InstanceStatusIcon: FC<Props> = ({ instance }) => {
         Ready: "status-waiting-small",
         Running: "status-succeeded-small",
         Stopped: "status-queued-small",
+        Migrating: "status-in-progress-small",
```
edlerd, line 24: I don't think we ever get a Migrating status from the backend, do we?
```suggestion
```

edlerd (changes requested):
Thanks, from an interaction point of view I think this is good now.

Two issues with layout:
1. We should not limit or hard-code the height of the modal. I.e. here I have enough height available to display all cluster members, but the modal is restricted in height and I have to scroll
image

2. The center alignment of the text in the modal seems a bit odd. I think again we shouldn't have a min-height on the modal body?
image

mas-who:
> Thanks, from an interaction point of view I think this is good now.
> 
> Two issues with layout:
> 
> 1. We should not limit or hard-code the height of the modal. I.e. here I have enough height available to display all cluster members, but the modal is restricted in height and I have to scroll
>    image
> 2. The center alignment of the text in the modal seems a bit odd. I think again we shouldn't have a min-height on the modal body?
>    image

What I wanted to achieve was to have the modal's height stay fixed between the selection modal and the confirmation modal so that there is less of a jumpy transition. I think it should be fine to let the modal adjust its height based on content though, will take out the height restrictions

src/pages/instances/InstanceList.tsx
mas-who: to use `instanceLoading`, we would need to handle setting individual instance states at the API level i.e. in `migrateInstanceBulk`, and we would also have to update `instanceLoading.getType` to take in a string and not `LxdInstance`. 

I think better if we just use `processingNames` instead of creating a new state `migratingInstances`? I'm already setting `processingNames` on starting bulk migration. Note `processingNames` are cleared after migration is complete, regardless of successful completion of individual operations.

src/pages/instances/InstanceStatusIcon.tsx
mas-who: aah yes that's correct, will remove

src/pages/instances/InstanceList.tsx
edlerd: Okay, sounds good to me.

mas-who:
@edlerd I've been toying around with the modal layout a little and have some observations:

- When removing the height restrictions for the modal content container (in this case it's the table container), it will result in the modal footer getting pushed down and becoming inaccessible. I think for the current use case, maybe we can remove the buttons at the bottom when selecting a member since they are not needed at that stage?
Screenshot from 2024-08-19 17-57-12
- I see the `CustomIsoSelector` component has the same issue. This seems to be a more general problem with modals that have nondeterministic content length. The scrollable containers we have in LXD UI currently only adjusts content height with respect to the viewport height i.e. `ScrollableContainer` and `ScrollableTable`. Perhaps we should add a general hook/wrapper component that would adjust content height based on some parent restriction instead?

mas-who:
@edlerd I've made the change to the migrate instance modal to use `ScrollableTable`. Let me know what you think :)

Screenshot from 2024-08-20 13-29-38
Screenshot from 2024-08-20 13-29-25

edlerd:
This looks pretty good. Two small requests:
1. I think "instance" should not bold here, just the name of the instance:
image

2. When selecting the target cluster member, the whole row should be clickable and a hover effect active on the rows. Compare it image selection on instance creation or choosing a custom volume on adding a disk device to an instance.

mas-who:
> This looks pretty good. Two small requests:
> 
> 1. I think "instance" should not bold here, just the name of the instance:
>    image
> 2. When selecting the target cluster member, the whole row should be clickable and a hover effect active on the rows. Compare it image selection on instance creation or choosing a custom volume on adding a disk device to an instance.

Done :+1:

edlerd:
Thanks, nearly perfect, just one edge case that appeared now: 1 instance and selecting it's current location / row needs a copy fix:

image

mas-who:
> Thanks, nearly perfect, just one edge case that appeared now: 1 instance and selecting its current location / row needs a copy fix:
> 
> image

Aah good catch, I actually put in place logic that if only one instance is selected, the select button that corresponds to the current location of the instance would be disabled. Forgot to extend that to the entire row with the previous push. This scenario shouldn't be possible.

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+    const selectMember = () => {
+      if (disableReason) {
+        return;
+      }
+
+      if (selectedMember === member.server_name) {
+        setSelectedMember("");
+        return;
+      }
```
edlerd, line 75: Why is this needed, can this be removed?

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+          "aria-label": "Status",
+          onClick: selectMember,
+        },
+        {
+          content: (
+            <Button
+              onClick={selectMember}
+              dense
+              title={disableReason}
+              disabled={!!disableReason}
```
edlerd, line 120:
```suggestion
              disabled={Boolean(disableReason)}
```
I think this is easier to read.

src/sass/_migrate_instance.scss
```diff
@@ -0,0 +1,27 @@
+.migrate-instance-modal {
+  .migrate-instance-table {
+    min-height: 15rem;
```
edlerd, line 3:
```suggestion
```
Seems like a leftover?

edlerd (changes requested):
QA LGTM, three small comments on the code.

src/pages/instances/MigrateInstanceModal.tsx
mas-who: yes this is not needed indeed, will remove

src/sass/_migrate_instance.scss
mas-who: yes that's correct, now we dynamically adjust the table height in JS, will remove

edlerd (approved):
LGTM, thanks for adding the bulk migration :)

PR #843, opened 2024-08-14
feat: improve project deletion experience [WD-13431]

mas-who:
## Done

- Added tooltip for delete project button on the project configurations page
- The tooltip will show number of resources currently using the project, and will link to the page for each resource type

## QA

1. Run the LXD-UI:
    - Make sure you have resources in the project that you are testing with e.g. instances, profiles, images etc
    - Go to the project configuration page, hover over the delete project button
    - You should see a tooltip showing you the resources currently using the project
    - You should be able to click on the resource label in the tooltip and be directed to the relevant page

## Screenshots
image

### DIFF
--- a/src/pages/projects/actions/DeleteProjectBtn.tsx
+++ b/src/pages/projects/actions/DeleteProjectBtn.tsx
@@ -1,5 +1,5 @@
-import { FC, useState } from "react";
-import { useNavigate } from "react-router-dom";
+import { FC, ReactNode, useState } from "react";
+import { Link, useNavigate } from "react-router-dom";
 import { LxdProject } from "types/project";
 import { deleteProject } from "api/projects";
 import { queryKeys } from "util/queryKeys";
@@ -10,14 +10,69 @@ import { useSmallScreen } from "context/useSmallScreen";
 import {
   ConfirmationButton,
   Icon,
+  Tooltip,
   useNotify,
 } from "@canonical/react-components";
 import classnames from "classnames";
 import { useToastNotification } from "context/toastNotificationProvider";
+import { filterUsedByType } from "util/usedBy";
+import { ResourceType } from "util/resourceDetails";
 
 interface Props {
   project: LxdProject;
 }
+
+const generateProjectUsedByTooltip = (project: LxdProject) => {
+  const resourceLabelAndLink: Record<string, { label: string; link: string }> =
+    {
+      instance: {
+        label: "Instances",
+        link: `/ui/project/${project.name}/instances`,
+      },
+      profile: {
+        label: "Profiles",
+        link: `/ui/project/${project.name}/profiles`,
+      },
+      image: {
+        label: "Images",
+        link: `/ui/project/${project.name}/images`,
+      },
+      volume: {
+        label: "Custom volumes",
+        link: `/ui/project/${project.name}/storage/volumes`,
+      },
+    };
+
+  const resourceTypes = Object.keys(resourceLabelAndLink);
+  const usedByItems: ReactNode[] = [];
+  for (const resourceType of resourceTypes) {
+    const usedBy = filterUsedByType(
+      resourceType as ResourceType,
+      project.used_by,
+    );
+
+    if (usedBy.length > 0) {
+      const label = resourceLabelAndLink[resourceType].label;
+      const link = resourceLabelAndLink[resourceType].link;
+      usedByItems.push(
+        <li
+          key={resourceType}
+          className="p-list__item is-dark u-no-margin--bottom"
+        >
+          {<Link to={link}>{label}</Link>} ({usedBy.length})
+        </li>,
+      );
+    }
+  }
+
+  return (
+    <>
+      Non-empty project cannot be deleted.
+      <p className="u-no-margin--bottom">Project is used by:</p>
+      <ul className="p-list u-no-margin--bottom">{usedByItems}</ul>
+    </>
+  );
+};
 
 const DeleteProjectBtn: FC<Props> = ({ project }) => {
   const isDeleteIcon = useSmallScreen();
@@ -34,7 +89,7 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
       return "The default project cannot be deleted";
     }
     if (!isEmpty) {
-      return "Non-empty projects cannot be deleted";
+      return "";
     }
     return "Delete project";
   };
@@ -81,8 +136,16 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
       shiftClickEnabled
       showShiftClickHint
     >
-      {isDeleteIcon && <Icon name="delete" />}
-      {!isDeleteIcon && <span>Delete project</span>}
+      <Tooltip
+        message={
+          !isEmpty && !isDefaultProject
+            ? generateProjectUsedByTooltip(project)
+            : ""
+        }
+      >
+        {isDeleteIcon && <Icon name="delete" />}
+        {!isDeleteIcon && <span>Delete project</span>}
+      </Tooltip>
     </ConfirmationButton>
   );
 };
--- a/src/sass/styles.scss
+++ b/src/sass/styles.scss
@@ -288,7 +288,7 @@ body {
 
 .help-link {
   align-items: center;
-  display: flex;
+  display: inline-flex;
 
   .help-link-icon {
     height: 1rem;
--- a/src/util/resourceDetails.tsx
+++ b/src/util/resourceDetails.tsx
@@ -9,6 +9,13 @@ export type ResourceDetail = {
   type: string;
 };
 
+export type ResourceType =
+  | "instance"
+  | "profile"
+  | "snapshot"
+  | "image"
+  | "volume";
+
 // refer to api spec to see how the names can be extracted from resource url
 // https://documentation.ubuntu.com/lxd/en/latest/api/
 export const extractResourceDetailsFromUrl = (
--- a/src/util/usedBy.tsx
+++ b/src/util/usedBy.tsx
@@ -1,4 +1,4 @@
-import { extractResourceDetailsFromUrl } from "./resourceDetails";
+import { extractResourceDetailsFromUrl, ResourceType } from "./resourceDetails";
 export interface LxdUsedBy {
   name: string;
   project: string;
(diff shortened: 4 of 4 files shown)


src/util/projectDelete.tsx
```diff
@@ -0,0 +1,54 @@
+import { ReactNode } from "react";
+import { filterUsedByType } from "./usedBy";
+import { ResourceType } from "./resourceDetails";
+import { LxdProject } from "types/project";
+import { Link } from "react-router-dom";
+
+export const generateProjectUsedByTooltip = (project: LxdProject) => {
+  const resourceLableAndLink: Record<string, { label: string; link: string }> =
```
Kxiru, line 8: Spelling error
```suggestion
  const resourceLabelAndLink: Record<string, { label: string; link: string }> =
```

Kxiru:
Changes screenshot:
image

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -30,12 +32,12 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
   const isDefaultProject = project.name === "default";
   const isEmpty = isProjectEmpty(project);
   const getHoverText = () => {
+    if (!isEmpty) {
+      return "";
+    }
     if (isDefaultProject) {
       return "The default project cannot be deleted";
     }
```
Kxiru, line 40: The default project - which _is_ used by several things - displays a the message "Non-empty project cannot be deleted." rather than this message. Perhaps rather than a display either message X or message Y, this warning ("The default project cannot be deleted") can be conditionally appended later on.

Kxiru (changes requested):
Looks good to me! Just minor suggestions from me.

Additionally, the Project config title that is across 2 lines is not caused by this PR but could it be fixed?

src/pages/projects/actions/DeleteProjectBtn.tsx
mas-who: I think appending the default project message might make the tooltip popup too busy and crowded. Since a default project can never be deleted, I think we can just show the message "The default project cannot be delete"?

Kxiru: Can do, but the problem I am highlighting is the fact that according to the code logic, two different messages _should_ be showing up, but only one of them is. We could either merge them, which may cause crowding, as you mentioned, or just show one of them.

mas-who: The current logic will show the usedBy tooltip if the project is not empty, even if the project is the default project. The new approach will show "The default project cannot be delete" for the default project, and the usedBy tooltip for non-default projects that are not empty.

mas-who:
> Looks good to me! Just minor suggestions from me.
> 
> Additionally, the Project config title that is across 2 lines is not caused by this PR but could it be fixed?

On second thought I think you are right, it looks quite ugly that the breadcrumbs are split across multiple lines, made a fix for it :+1:

Kxiru:
> > Looks good to me! Just minor suggestions from me.
> > Additionally, the Project config title that is across 2 lines is not caused by this PR but could it be fixed?
> 
> On second thought I think you are right, it looks quite ugly that the breadcrumbs are split across multiple lines, made a fix for it 👍

Looks a lot better! Sorry to nitpick, but do you think the information icon should go at the end of the line rather than after the project name? I know the info pertains more towards the project config, but it may look a lot nicer.
image

mas-who:
> > > Looks good to me! Just minor suggestions from me.
> > > Additionally, the Project config title that is across 2 lines is not caused by this PR but could it be fixed?
> > 
> > 
> > On second thought I think you are right, it looks quite ugly that the breadcrumbs are split across multiple lines, made a fix for it 👍
> 
> Looks a lot better! Sorry to nitpick, but do you think the information icon should go at the end of the line rather than after the project name? I know the info pertains more towards the project config, but it may look a lot nicer. image

That's a fair point but in this case the header is actually a rename header. When you click on the project name, it becomes an input and if the help icon is after the project name, the link would jump around causing its position to shift which I don't think is a good behaviour.

Screenshot from 2024-08-15 17-14-10

piperdeck:
This looks good! Good solution for now until we find a way to force delete!

Kxiru (approved):
If Design is happy with the changes, let's move forwards. Code and QA LGTM! :)

edlerd:
I think the design is not right, the spacing looks huge between the different sections in the used by list.

The font colour of the visited/unvisited link is hardly readable on the black tooltip background. Probably should use the [inverted link class](https://vanillaframework.io/docs/patterns/links) or use a tooltip with a lighter background.

src/sass/styles.scss
```diff
@@ -288,7 +288,7 @@ body {
 
 .help-link {
   align-items: center;
-  display: flex;
+  display: inline-flex;
```
edlerd, line 291: Good catch!

mas-who:
> I think the design is not right, the spacing looks huge between the different sections in the used by list.
> 
> The font colour of the visited/unvisited link is hardly readable on the black tooltip background. Probably should use the [inverted link class](https://vanillaframework.io/docs/patterns/links) or use a tooltip with a lighter background.

@edlerd I've adjusted the spacing and link color (dark theme link) in the tooltip, let me know what you think. If possible I think we should avoid changing the tooltip background color since it won't be consistent with other apps that uses the tooltip component e.g. anbox.

src/util/projectDelete.tsx
```diff
@@ -0,0 +1,54 @@
...
+      usedByItems.push(
+        <p key={resourceType} className="is-dark u-no-margin--bottom">
+          {link ? <Link to={link}>{label}</Link> : label} ({usedBy.length})
+        </p>,
+      );
+    }
+  }
+
+  return (
+    <aside>
```
edlerd, line 48:
```suggestion
    <>
```
Not sure if aside is the right thing to use here. Haven't seen it used for the tooltip message before. Maybe just a plain fragment works as well?

src/util/projectDelete.tsx
```diff
@@ -0,0 +1,54 @@
+import { ReactNode } from "react";
+import { filterUsedByType } from "./usedBy";
+import { ResourceType } from "./resourceDetails";
+import { LxdProject } from "types/project";
+import { Link } from "react-router-dom";
+
+export const generateProjectUsedByTooltip = (project: LxdProject) => {
```
edlerd, line 7: This seems pretty tailored towards the `DeleteProjectBtn`, shouldn't the function live in that component file?

edlerd (commented):
Thanks for the design adjustments, those look all right to me now. Some smallish questions on code and structure below

src/util/projectDelete.tsx
mas-who: good point, I've moved the function into `DeleteProjectBtn` file now.

src/util/projectDelete.tsx
```diff
@@ -0,0 +1,54 @@
...
+      project.used_by,
+    );
+
+    if (usedBy.length > 0) {
+      const label = resourceLabelAndLink[resourceType].label;
+      const link = resourceLabelAndLink[resourceType].link;
+      usedByItems.push(
+        <p key={resourceType} className="is-dark u-no-margin--bottom">
+          {link ? <Link to={link}>{label}</Link> : label} ({usedBy.length})
+        </p>,
```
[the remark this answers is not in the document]
mas-who: Yes 100%, I've adjusted the html structure for this now.

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -10,15 +10,70 @@ import { useSmallScreen } from "context/useSmallScreen";
...
 import { useToastNotification } from "context/toastNotificationProvider";
+import { filterUsedByType } from "util/usedBy";
+import { ResourceType } from "util/resourceDetails";
 
 interface Props {
   project: LxdProject;
 }
 
+const generateProjectUsedByTooltip = (project: LxdProject) => {
+  const resourceLabelAndLink: Record<string, { label: string; link: string }> =
```
edlerd, line 26:
```suggestion
  const resourceLabelAndLink: Record<ResourceType, { label: string; link: string }> =
```

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -10,15 +10,70 @@ import { useSmallScreen } from "context/useSmallScreen";
...
+        label: "Custom volumes",
+        link: `/ui/project/${project.name}/storage/volumes`,
+      },
+    };
+
+  const resourceTypes = Object.keys(resourceLabelAndLink);
+  const usedByItems: ReactNode[] = [];
+  for (const resourceType of resourceTypes) {
+    const usedBy = filterUsedByType(
+      resourceType as ResourceType,
```
edlerd, line 50:
```suggestion
      resourceType,
```

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -10,15 +10,70 @@ import { useSmallScreen } from "context/useSmallScreen";
...
+
+    if (usedBy.length > 0) {
+      const label = resourceLabelAndLink[resourceType].label;
+      const link = resourceLabelAndLink[resourceType].link;
+      usedByItems.push(
+        <li
+          key={resourceType}
+          className="p-list__item is-dark u-no-margin--bottom"
+        >
+          {link ? <Link to={link}>{label}</Link> : label} ({usedBy.length})
```
edlerd, line 62:
```suggestion
          {<Link to={link}>{label}</Link>} ({usedBy.length})
```
I think link is never empty, when label is set, isn't it?

edlerd (commented):
LGTM, some very tiny nitpicks below.

src/pages/projects/actions/DeleteProjectBtn.tsx
mas-who: chose `string` here as I didn't want to specify a key for `snapshot`, also later when iterating the keys in `resourceLabelAndLink` we would still need to cast those keys from `string` to `ResourceType`?

edlerd: Ah right, thought we could streamline this.

PR #848, opened 2024-08-15
feat: [WD-13704] Create image from backup file

Kxiru:
## Done

- Added import tarball button as well as associated modal.
- Minor scss changes
- Additional API calls for importing the file and amending image properties.

## QA

1. Run the LXD-UI:
    - Upload a unified tarball image or a split tarball image to LXD-UI by selecting the "Import from Tarball" button on the Image list page.

## Screenshots

image
image
image

### DIFF
--- a/src/api/images.tsx
+++ b/src/api/images.tsx
@@ -9,6 +9,8 @@ import { LxdApiResponse } from "types/apiResponse";
 import { LxdOperationResponse } from "types/operation";
 import { EventQueue } from "context/eventQueue";
 import { LxdInstance } from "types/instance";
+import { UploadState } from "types/storage";
+import axios, { AxiosResponse } from "axios";
 
 export const fetchImage = (
   image: string,
@@ -103,6 +105,31 @@ export const createImage = (
       body: body,
     })
       .then(handleResponse)
+      .then(resolve)
+      .catch(reject);
+  });
+};
+
+export const uploadImage = (
+  body: File | FormData,
+  isPublic: boolean,
+  setUploadState: (value: UploadState) => void,
+): Promise<LxdOperationResponse> => {
+  return new Promise((resolve, reject) => {
+    axios
+      .post(`/1.0/images`, body, {
+        headers: {
+          "X-LXD-public": JSON.stringify(isPublic),
+        },
+        onUploadProgress: (event) => {
+          setUploadState({
+            percentage: event.progress ? Math.floor(event.progress * 100) : 0,
+            loaded: event.loaded,
+            total: event.total,
+          });
+        },
+      })
+      .then((response: AxiosResponse<LxdOperationResponse>) => response.data)
       .then(resolve)
       .catch(reject);
   });
--- a/src/pages/images/actions/forms/UploadImageForm.tsx
+++ b/src/pages/images/actions/forms/UploadImageForm.tsx
@@ -1,0 +1,198 @@
+import { ChangeEvent, FC, useState } from "react";
+import { useEventQueue } from "context/eventQueue";
+import { useFormik } from "formik";
+import { useToastNotification } from "context/toastNotificationProvider";
+import { createImageAlias, uploadImage } from "api/images";
+import {
+  ActionButton,
+  Button,
+  Form,
+  Input,
+  Modal,
+} from "@canonical/react-components";
+import * as Yup from "yup";
+import { Link } from "react-router-dom";
+import { humanFileSize } from "util/helpers";
+import ProgressBar from "components/ProgressBar";
+import { UploadState } from "types/storage";
+import Loader from "components/Loader";
+import { useQueryClient } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { LxdSyncResponse } from "types/apiResponse";
+import { AxiosError } from "axios";
+
+interface Props {
+  close: () => void;
+  project: string;
+}
+
+const UploadImageForm: FC<Props> = ({ close, project }) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+  const [uploadState, setUploadState] = useState<UploadState | null>(null);
+  const queryClient = useQueryClient();
+
+  const notifySuccess = () => {
+    const uploaded = <Link to={`/ui/project/${project}/images`}>uploaded</Link>;
+    toastNotify.success(<>Image {uploaded}.</>);
+  };
+
+  const changeFile = (e: ChangeEvent<HTMLInputElement>) => {
+    if (e.target.files) {
+      void formik.setFieldValue("fileList", e.target.files);
+    }
+  };
+
+  const getImageUploadBody = (fileList: FileList): File | FormData => {
+    if (fileList.length === 1) {
+      return fileList[0];
+    } else {
+      // Sorting by Size. The metadata file is very likely to be smaller than the image itself.
+      const formData = new FormData();
+      const sortedFiles = Array.from(fileList).sort((a, b) => a.size - b.size);
+
+      formData.append("metadata", sortedFiles[0]);
+      formData.append("rootfs.img", sortedFiles[1]);
+
+      return formData;
+    }
+  };
+
+  const formik = useFormik<{
+    alias: string;
+    isPublic: boolean;
+    fileList: FileList | null;
+  }>({
+    initialValues: {
+      alias: "",
+      isPublic: false,
+      fileList: null,
+    },
+    validationSchema: Yup.object().shape({
+      alias: Yup.string(),
+    }),
+    onSubmit: (values) => {
+      if (values.fileList) {
+        if (values.fileList.length > 2) {
+          close();
+          toastNotify.failure(
+            `Image upload failed.`,
+            new Error("Too many files selected"),
+          );
+          return;
+        }
+        uploadImage(
+          getImageUploadBody(values.fileList),
+          values.isPublic,
+          setUploadState,
+        )
+          .then((operation) => {
+            toastNotify.info(<>Creation of image from file started.</>);
+
+            eventQueue.set(
+              operation.metadata.id,
+              (event) => {
+                const fingerprint = event.metadata.metadata?.fingerprint ?? "";
+                if (values.alias) {
+                  void createImageAlias(fingerprint, values.alias);
+                }
+                void queryClient.invalidateQueries({
+                  queryKey: [queryKeys.images, project],
+                });
+                notifySuccess();
+              },
+              (msg) => {
+                toastNotify.failure(`Image upload failed.`, new Error(msg));
+              },
+            );
+          })
+          .catch((e: AxiosError<LxdSyncResponse<null>>) => {
+            const error = new Error(e.response?.data.error);
+            toastNotify.failure("Image upload failed", error);
+          })
+          .finally(() => {
+            close();
+          });
+      } else {
+        close();
+        toastNotify.failure(`Image upload failed`, new Error("Missing files"));
+      }
+    },
+  });
+
+  return (
+    <Modal
+      close={close}
+      title="Import image from file"
+      className="upload-image-modal"
+      buttonRow={
+        <>
+          {uploadState && (
+            <>
+              <ProgressBar percentage={Math.floor(uploadState.percentage)} />
+              <p>
+                {humanFileSize(uploadState.loaded)} loaded of{" "}
+                {humanFileSize(uploadState.total ?? 0)}
+              </p>
+              {uploadState.loaded === uploadState.total && (
+                <Loader text="Uploading file" />
+              )}
+            </>
+          )}
+          <Button
+            appearance="base"
+            className="u-no-margin--bottom"
+            type="button"
+            onClick={close}
+          >
+            Cancel
+          </Button>
+          <ActionButton
+            appearance="positive"
+            className="u-no-margin--bottom"
+            loading={formik.isSubmitting}
+            disabled={!formik.isValid || !formik.values.fileList}
+            onClick={() => void formik.submitForm()}
+          >
+            Upload image
+          </ActionButton>
+        </>
+      }
+    >
+      <Form
+        className={uploadState ? "u-hide" : ""}
+        onSubmit={formik.handleSubmit}
+      >
+        <Input
+          type="file"
+          name="fileList"
+          label="Image backup file"
+          onChange={changeFile}
+          multiple
+        />
+        <Input
+          {...formik.getFieldProps("alias")}
+          type="text"
+          label="Alias"
+          placeholder="Enter alias"
+          error={formik.touched.alias ? formik.errors.alias : null}
+        />
+        <Input
+          {...formik.getFieldProps("isPublic")}
+          type="checkbox"
+          label="Make the image publicly available"
+          error={formik.touched.isPublic ? formik.errors.isPublic : null}
+        />
+        {/* hidden submit to enable enter key in inputs */}
+        <Input
+          type="submit"
+          hidden
+          value="Hidden input"
+          disabled={!formik.isValid || !formik.values.fileList}
+        />
+      </Form>
+    </Modal>
+  );
+};
+
+export default UploadImageForm;
--- a/src/sass/_page_header.scss
+++ b/src/sass/_page_header.scss
@@ -21,11 +21,6 @@
       width: 45vw;
     }
   }
-
-  .page-header__base-actions {
-    align-self: flex-start;
-    flex: 1;
-  }
 }
 
 @media screen and (min-width: $breakpoint-large) and (width <= 1135px) {
--- a/src/pages/images/ImageList.tsx
+++ b/src/pages/images/ImageList.tsx
@@ -28,7 +28,8 @@ import { useDocs } from "context/useDocs";
 import CustomLayout from "components/CustomLayout";
 import PageHeader from "components/PageHeader";
 import CustomIsoBtn from "pages/storage/actions/CustomIsoBtn";
-import ExportImageTarballBtn from "./actions/ExportImageTarballBtn";
+import DownloadImageBtn from "./actions/DownloadImageBtn";
+import UploadImageBtn from "pages/images/actions/UploadImageBtn";
 
 const ImageList: FC = () => {
   const docBaseLink = useDocs();
@@ -112,7 +113,7 @@ const ImageList: FC = () => {
             project={project}
             image={localLxdToRemoteImage(image)}
           />,
-          <ExportImageTarballBtn key="export-image" image={image} />,
+          <DownloadImageBtn key="export-image" image={image} />,
           <DeleteImageBtn key="delete" image={image} project={project} />,
         ]}
       />
@@ -229,6 +230,7 @@ const ImageList: FC = () => {
             )}
           </PageHeader.Left>
           <PageHeader.BaseActions>
+            <UploadImageBtn project={project} />
             <CustomIsoBtn project={project} />
           </PageHeader.BaseActions>
         </PageHeader>
--- a/src/pages/images/actions/DownloadImageBtn.tsx
+++ b/src/pages/images/actions/DownloadImageBtn.tsx
@@ -7,7 +7,7 @@ interface Props {
   image: LxdImage;
 }
 
-const ExportImageTarballBtn: FC<Props> = ({ image }) => {
+const DownloadImageBtn: FC<Props> = ({ image }) => {
   const toastNotify = useToastNotification();
   const [isLoading, setLoading] = useState(false);
   const description = image.properties?.description ?? image.fingerprint;
@@ -51,4 +51,4 @@ const ExportImageTarballBtn: FC<Props> = ({ image }) => {
   );
 };
 
-export default ExportImageTarballBtn;
+export default DownloadImageBtn;
--- a/src/pages/images/actions/UploadImageBtn.tsx
+++ b/src/pages/images/actions/UploadImageBtn.tsx
@@ -1,0 +1,27 @@
+import { FC } from "react";
+import { ActionButton } from "@canonical/react-components";
+import usePortal from "react-useportal";
+import UploadImageForm from "./forms/UploadImageForm";
+
+interface Props {
+  project: string;
+}
+
+const UploadImageBtn: FC<Props> = ({ project }) => {
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <UploadImageForm close={closePortal} project={project} />
+        </Portal>
+      )}
+      <ActionButton className="u-no-margin--bottom" onClick={openPortal}>
+        <span>Upload image</span>
+      </ActionButton>
+    </>
+  );
+};
+
+export default UploadImageBtn;
--- a/src/sass/_images.scss
+++ b/src/sass/_images.scss
@@ -29,3 +29,11 @@
     margin-top: -7px;
   }
 }
+
+.upload-image-modal {
+  .p-modal__dialog {
+    @include large {
+      width: 35rem;
+    }
+  }
+}


src/api/images.tsx
```diff
@@ -114,3 +114,79 @@ export const createImageAlias = (
       .catch(reject);
   });
 };
+
+export const createImageFromInstance = (
```
mas-who, line 118: I think we should combine these functions into a single function `createImage`. This means we should try make adjustment to the existing `createImageFromInstanceSnapshot` function as well. The construction of the payload body should be determined outside of the API layer. Also be careful and don't forget about the `?project=` query parameter.

src/api/images.tsx
```diff
@@ -114,3 +114,79 @@ export const createImageAlias = (
...
+      .then(resolve)
+      .catch(reject);
+  });
+};
+
+export const createImageFromTarball = (
+  fileList: FileList,
+): Promise<LxdOperationResponse> => {
+  return new Promise((resolve, reject) => {
+    let obj: File | FormData | null = null;
```
mas-who, line 144: the logic here could be moved to a util file

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+            eventQueue.set(
+              operation.metadata.id,
+              (event) => {
+                const fingerprint = event.metadata.metadata?.fingerprint ?? "";
+                if (values.alias) {
+                  void createImageAlias(fingerprint, values.alias);
+                }
+                if (values.isPublic) {
+                  void changeImagePublicity(fingerprint, values.isPublic);
```
mas-who, line 72: we do not need to update the image to make it public here. You can set the public field for an image with the `X-LXD-public` header, refer to the [doc](https://documentation.ubuntu.com/lxd/en/latest/api/#/images/images_post). This is of course for a file upload request to `POST /1.0/images`.

src/api/images.tsx
```diff
@@ -114,3 +114,79 @@ export const createImageAlias = (
...
+      method: "POST",
+      body: obj,
+    })
+      .then(handleResponse)
+      .then(resolve)
+      .catch(reject);
+  });
+};
+
+export const changeImagePublicity = (
```
mas-who, line 177: this function is not needed. See later comment.

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+} from "@canonical/react-components";
+import * as Yup from "yup";
+import { Link, useNavigate } from "react-router-dom";
+
+interface Props {
+  close: () => void;
+  project: string;
+}
+
+const ImportImageFromTarballForm: FC<Props> = ({ close, project }) => {
```
mas-who, line 25: we should implement progress status for file upload similar to how it's done in `UploadCustomIso`. @edlerd I see we are using `axios` for this, although is it possible to track multiple file uploads at the same time (for the split tarball import)?

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+    },
+    validationSchema: Yup.object().shape({
+      alias: Yup.string(),
+    }),
+    onSubmit: (values) => {
+      if (fileList) {
+        createImageFromTarball(fileList)
+          .then((operation) => {
+            toastNotify.info(<>Creation of image from file started.</>);
+            close();
```
mas-who, line 62: If we are tracking the file upload progress, then we should not close the modal here

mas-who (commented):
Still need to do QA and go through the notification copy. Left some code feedback so long :)

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+interface Props {
+  close: () => void;
+  project: string;
+}
+
+const ImportImageFromTarballForm: FC<Props> = ({ close, project }) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+  const navigate = useNavigate();
+  const [fileList, setFileList] = useState<FileList | null>(null);
```
mas-who, line 29: I don't think we need this state, we can manage `fileList` using `formik`

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const published = (
+      <Link to={`/ui/project/${project}/images`}>published</Link>
+    );
+    toastNotify.success(<>Image {published} from file.</>);
+  };
+
+  const changeFile = (e: ChangeEvent<HTMLInputElement>) => {
+    if (e.target.files) {
+      const files = e.target.files;
+      setFileList(files);
```
mas-who, line 41: instead of managing file with additional state, you can use `formik.setFieldValue` here

mas-who:
@Kxiru QA for the importing functionality looks good, I was able to:
- import container and VM split tarballs
- import container and VM unified tarballs

src/api/images.tsx
Kxiru: Are you saying that createImageAlias should be integrated into both createImagefromInstanceSnapshot and createImageFromInstance?

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
Kxiru: @edlerd , should we implement an axios approach for this?

edlerd: I think axios is a nice helper to get the upload status. We can have a progress bar per file that is being uploaded, or restart the same progress bar for each new file.

src/api/images.tsx
mas-who: no I meant both `createImageFromInstanceSnapshot` and `createImageFromInstance` should become `createImage` with a body input determined from outside of the function.

edlerd: I think the `createImageFromInstance` function shouldn't be in this PR, it is introduced in https://github.com/canonical/lxd-ui/pull/833
though I agree, we could have a single `createImage` function, which receives a body. Then we can reuse it for the tarball upload and for the creation of an image from an instance.

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
Kxiru: I actually did try this before this implementation and it did not work. I also read somewhere online that for files, it is unsupported. I recreated it again just in case I might have done it wrong, but I got the same error: 
image

For reference, this was the new implementation:
```
  const changeFile = (e: ChangeEvent<HTMLInputElement>) => {
    if (e.target.files) {
      formik.setFieldValue("fileList", e.target.files);
    }
  };
```

mas-who: You can refer to #853 for how this can be done. The key thing here is to remove `formik.getFieldProps` from the file input and let `changeFile` handle selection changes

src/api/images.tsx
Kxiru: Created an issue to address this: https://warthogs.atlassian.net/browse/WD-14347?atlOrigin=eyJpIjoiMjcyYmI5Mjk3MjI1NGYxNWI4NWRlZGEwODViYjhlNDUiLCJwIjoiaiJ9

edlerd: Thanks

src/pages/images/actions/ImportFromTarballBtn.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <ImportImageFromTarballForm close={closePortal} project={project} />
+        </Portal>
+      )}
+      <ActionButton className="u-no-margin--bottom" onClick={openPortal}>
+        <span>Import from Tarball</span>
```
edlerd, line 21:
```suggestion
        <span>Upload image</span>
```
Maybe clearer?
Also this component could be UploadImageBtn and the form UploadImageForm, wdyt?

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+            eventQueue.set(
+              operation.metadata.id,
+              (event) => {
+                const fingerprint = event.metadata.metadata?.fingerprint ?? "";
+                if (values.alias) {
+                  void createImageAlias(fingerprint, values.alias);
+                }
+                close();
+                navigate(0);
```
edlerd, line 73: What is this here for?

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+            toastNotify.info(<>Creation of image from file started.</>);
+
+            eventQueue.set(
+              operation.metadata.id,
+              (event) => {
+                const fingerprint = event.metadata.metadata?.fingerprint ?? "";
+                if (values.alias) {
+                  void createImageAlias(fingerprint, values.alias);
+                }
+                close();
```
edlerd, line 72: The close should also be called in the failure case. It might be simplest to move the close call to the 4th argument of eventqueue.set, which is a callback that is called in both cases, on success and failure.

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const ImportImageFromTarballForm: FC<Props> = ({ close, project }) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+  const navigate = useNavigate();
+  const [uploadState, setUploadState] = useState<UploadState | null>(null);
+
+  const notifySuccess = () => {
+    const imported = <Link to={`/ui/project/${project}/images`}>imported</Link>;
+    toastNotify.success(<>Image {imported} from file.</>);
```
edlerd, line 33:
```suggestion
    toastNotify.success(<>Image {imported}.</>);
```
More concise and just as precise?

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+                if (values.alias) {
+                  void createImageAlias(fingerprint, values.alias);
+                }
+                close();
+                navigate(0);
+                notifySuccess();
+              },
+              (msg) => {
+                toastNotify.failure(
+                  `Image creation from file failed.`,
```
edlerd, line 78:
```suggestion
                  `Image upload failed.`,
```

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  new Error(msg),
+                );
+              },
+            );
+          })
+          .catch((e) => {
+            toastNotify.failure(`Image creation from file failed.`, e);
```
edlerd, line 85:
```suggestion
            toastNotify.failure(`Image upload failed.`, e);
```

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+}
+
+const ImportImageFromTarballForm: FC<Props> = ({ close, project }) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+  const navigate = useNavigate();
+  const [uploadState, setUploadState] = useState<UploadState | null>(null);
+
+  const notifySuccess = () => {
+    const imported = <Link to={`/ui/project/${project}/images`}>imported</Link>;
```
edlerd, line 32:
```suggestion
    const uploaded = <Link to={`/ui/project/${project}/images`}>uploaded</Link>;
```

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+            Cancel
+          </Button>
+          <ActionButton
+            appearance="positive"
+            className="u-no-margin--bottom"
+            loading={formik.isSubmitting}
+            disabled={!formik.isValid}
+            onClick={() => void formik.submitForm()}
+          >
+            Create image
```
edlerd, line 125:
```suggestion
            Upload image
```

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+          </ActionButton>
+        </>
+      }
+    >
+      <Form onSubmit={formik.handleSubmit}>
+        <Input
+          type="file"
+          name="fileList"
+          label="Local file"
```
edlerd, line 134:
```suggestion
          label="Image backup file"
```

src/sass/_page_header.scss
```diff
@@ -24,7 +24,6 @@
 
   .page-header__base-actions {
     align-self: flex-start;
-    flex: 1;
```
edlerd, line 27: If we remove the flex, we can also remove the align self and the whole block, as it won't have any effect.

src/util/helpers.tsx
```diff
@@ -290,3 +290,27 @@ export const getClientOS = (userAgent: string) => {
 
   return null;
 };
+
+export const getImageToTarballBody = (fileList: FileList): File | FormData => {
```
edlerd, line 294: This seems very specific to the ImportImageFromTarballForm, so I would keep it local to that file. This is very unlikely to be reused anywhere else.

src/util/helpers.tsx
```diff
@@ -290,3 +290,27 @@ export const getClientOS = (userAgent: string) => {
...
+  let obj: File | FormData | null = null;
+  const formData = new FormData();
+
+  if (fileList.length === 1) {
+    obj = fileList[0];
```
edlerd, line 299:
```suggestion
    return fileList[0];
```

This way we don't need the else, and we don't need to define obj above. We can directly return formData below.

src/util/helpers.tsx
```diff
@@ -290,3 +290,27 @@ export const getClientOS = (userAgent: string) => {
 
   return null;
 };
+
+export const getImageToTarballBody = (fileList: FileList): File | FormData => {
+  let obj: File | FormData | null = null;
+  const formData = new FormData();
```
edlerd, line 296: This should be defined closer to its use case.

src/api/images.tsx
```diff
@@ -116,6 +118,7 @@ export const createImageAlias = (
 };
 
 export const createImageFromInstance = (
+  //Also known as create Image from an Instance.
```
edlerd, line 121:
```suggestion
```
Seems unrelated?

src/pages/images/actions/ImportFromTarballBtn.tsx
Kxiru: Can do. I was following the naming convention of other buttons such as ExportImageTarballBtn, CreateInstanceFromImageBtn and UploadCustomIsoBtn. I can change this if you wish? I believe ImportFromTarballBtn is a bit more specific.

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
Kxiru: This refreshes the imageList page when the image is created as the created image does not display in the list unless it is rerendered/refreshed.

src/pages/images/actions/ImportFromTarballBtn.tsx
edlerd: CreateInstanceFromImageBtn is overly specific, because there is also another way of creating instances without selecting an image first.

Here I think we only have one way to upload/create an image, that's why I suggested the simpler UploadImage name. Maybe also better to call it ExportImage, as there is only one way of export, so we don't need the Tarball in the name. Or do we have avariations of export / upload image? Thinking of it -- maybe it should be DownloadImage instead of ExportImage as that more close to what it does?

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
edlerd: Should we better do a query cache invalidation for `[queryKeys.images, project]`? That should trigger a refresh of the cached images.

src/pages/images/actions/ImportFromTarballBtn.tsx
Kxiru: Alrighty, coming right up!
ImportFromTarballBtn -> UploadImageBtn (and Form)
ExportImageTarballBtn -> DownloadImageBtn (and Form)

src/pages/images/actions/forms/ImportImageFromTarballForm.tsx
Kxiru: Perfect! That works great, thanks David :)

Kxiru:
I believe this latest version is up to date, @edlerd , @mas-who . Please let me know your thoughts :)

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+        );
+      }
+
+      return formData;
+    }
+    return fileList[0];
+  };
+
+  const formik = useFormik<{
+    alias?: string;
```
edlerd, line 65:
```suggestion
    alias: string;
```
I think this is always string as it initializes with empty string.

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+              operation.metadata.id,
+              (event) => {
+                const fingerprint = event.metadata.metadata?.fingerprint ?? "";
+                if (values.alias) {
+                  void createImageAlias(fingerprint, values.alias);
+                }
+                void queryClient.invalidateQueries({
+                  queryKey: [queryKeys.images, project],
+                });
+                close();
```
edlerd, line 97:
```suggestion
```
I think we don't need this any more, the close call in the finally block below should trigger earlier.

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+            loading={formik.isSubmitting}
+            disabled={!formik.isValid}
+            onClick={() => void formik.submitForm()}
+          >
+            Upload image
+          </ActionButton>
+        </>
+      }
+    >
+      <Form onSubmit={formik.handleSubmit}>
```
edlerd, line 154: We should hide the form during upload. Condition would be `!uploadState` to hide it.

src/api/images.tsx
```diff
@@ -107,3 +109,28 @@ export const createImage = (
       .catch(reject);
   });
 };
+
+export const createImageFromTarball = (
```
edlerd, line 113:
```suggestion
export const uploadImage = (
```
More concise and just as precise? Also I think we won't always have a tarball here.

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+    toastNotify.success(<>Image {uploaded}.</>);
+  };
+
+  const changeFile = (e: ChangeEvent<HTMLInputElement>) => {
+    if (e.target.files) {
+      void formik.setFieldValue("fileList", e.target.files);
+    }
+  };
+
+  const getImageToTarballBody = (fileList: FileList): File | FormData => {
```
edlerd, line 44:
```suggestion
  const getImageUploadBody = (fileList: FileList): File | FormData => {
```
Tarball seems wrong, as this is also handling the case of root image plus metadata file, that can be gzip, without a tarball.

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+      if (!formData.has("metadata") || !formData.has("rootfs.img")) {
+        throw new Error(
+          "Both metadata (tar.xz / gzip) and rootfs.img files are required",
+        );
+      }
+
+      return formData;
+    }
+    return fileList[0];
```
edlerd, line 61: Can we turn the condition around `If filelist.length === 1 { ... }` to improve two things:
1. we will have a positive comparison, making it easier to read and
2. most of the function will not be double indented.

edlerd (commented):
Thanks for the adjustments, this is now much more straight forward to read.

src/api/images.tsx
```diff
@@ -107,3 +109,28 @@ export const createImage = (
...
+  body: File | FormData,
+  alias: string,
+  setUploadState: (value: UploadState) => void,
+): Promise<LxdOperationResponse> => {
+  return new Promise((resolve, reject) => {
+    axios
+      .post(`/1.0/images`, body, {
+        headers: {
+          "X-LXD-public": alias,
```
mas-who, line 122: Why is `alias` assigned to the `X-LXD-public` header?

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+  };
+
+  const changeFile = (e: ChangeEvent<HTMLInputElement>) => {
+    if (e.target.files) {
+      void formik.setFieldValue("fileList", e.target.files);
+    }
+  };
+
+  const getImageToTarballBody = (fileList: FileList): File | FormData => {
+    if (fileList.length != 1) {
```
mas-who, line 45: should we add an extra condition here for `fileList.length > 2` and throw an error? Not sure about what copy to provide though, the thinking here is that we could prevent a potentially expensive API call (if the files are large).

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+    initialValues: {
+      alias: "",
+      isPublic: false,
+      fileList: null,
+    },
+    validationSchema: Yup.object().shape({
+      alias: Yup.string(),
+    }),
+    onSubmit: (values) => {
+      if (values.fileList) {
```
mas-who, line 78: I don't think we need this check here if we disable form submission for when `fileList` is `null`

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const getImageToTarballBody = (fileList: FileList): File | FormData => {
+    if (fileList.length != 1) {
+      // Sorting by Size. The metadata file is very likely to be smaller than the image itself.
+      const formData = new FormData();
+      const sortedFiles = Array.from(fileList).sort((a, b) => a.size - b.size);
+
+      formData.append("metadata", sortedFiles[0]);
+      formData.append("rootfs.img", sortedFiles[1]);
+
+      if (!formData.has("metadata") || !formData.has("rootfs.img")) {
```
mas-who, line 53: Is this check necessary? If we have more than 1 file, then surely both form data fields will be populated.

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+      isPublic: false,
+      fileList: null,
+    },
+    validationSchema: Yup.object().shape({
+      alias: Yup.string(),
+    }),
+    onSubmit: (values) => {
+      if (values.fileList) {
+        createImageFromTarball(
+          getImageToTarballBody(values.fileList),
```
mas-who, line 80: have you checked that if an error was thrown from this function at this point, it does not crash the entire application? I don't think the `.catch` block will catch the error from this function since it's not part of `createImageFrromTarball`.

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+          values.alias || "",
```
mas-who, line 81: I think this should be `values.isPublic`?

src/pages/images/actions/forms/UploadImageForm.tsx
mas-who: Yes, maybe use the same approach as the other components that have this upload pattern? We can hide the form with `className={uploadState ? "u-hide" : ""}`

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+            type="button"
+            onClick={close}
+          >
+            Cancel
+          </Button>
+          <ActionButton
+            appearance="positive"
+            className="u-no-margin--bottom"
+            loading={formik.isSubmitting}
+            disabled={!formik.isValid}
```
mas-who, line 146: I think we should disable submission if no files are selected
```suggestion
            disabled={!formik.isValid || !formik.values.fileList}
```

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+          error={formik.touched.alias ? formik.errors.alias : null}
+        />
+        <Input
+          {...formik.getFieldProps("isPublic")}
+          type="checkbox"
+          label="Make the image publicly available"
+          error={formik.touched.isPublic ? formik.errors.isPublic : null}
+        />
+        {/* hidden submit to enable enter key in inputs */}
+        <Input type="submit" hidden value="Hidden input" />
```
mas-who, line 175: This should also be disabled if no files are selected.

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+          type="file"
+          name="fileList"
+          label="Image backup file"
+          onChange={changeFile}
+          multiple
+        />
+        <Input
+          {...formik.getFieldProps("alias")}
+          type="text"
+          label="Alias"
```
mas-who, line 165: Let's add a placeholder for this input "Enter alias"

src/pages/images/actions/forms/UploadImageForm.tsx
Kxiru: Wait, why would we hide the upload form when we have put the progress bar inside it?

src/pages/images/actions/forms/UploadImageForm.tsx
Kxiru: image

mas-who: aah I see, type checks

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+        formData.append("metadata", sortedFiles[0]);
+        formData.append("rootfs.img", sortedFiles[1]);
+
+        return formData;
+      }
+    } catch (e) {
+      close();
+      toastNotify.failure(`Image upload failed`, e);
+      throw e;
+    }
```
edlerd, line 68: I think we shouldn't throw/catch in here, the check if have more than two files should be in line85 with the `if (values.fileList)` check.

edlerd (dismissed):

src/pages/images/actions/forms/UploadImageForm.tsx
Kxiru: Thank you, please see my review :)

mas-who:
@Kxiru I just tested uploading the files again and it's failing on unified tarballs (see below). I suspect it's because in your axios API call, `Content-Type` header is not set to `application/octet-stream`. Please would you look into this thanks.

Screenshot from 2024-08-22 14-53-06

Kxiru:
> @Kxiru I just tested uploading the files again and it's failing on unified tarballs (see below). I suspect it's because in your axios API call, `Content-Type` header is not set to `application/octet-stream`. Please would you look into this thanks.
> 
> Screenshot from 2024-08-22 14-53-06
> 
> Edit: the same occurs for split image upload as well. Please check both cases.

Hmm... Looking into this as I am not having this issue locally, only on the demo server...

src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+      alias: "",
+      isPublic: false,
+      fileList: null,
+    },
+    validationSchema: Yup.object().shape({
+      alias: Yup.string(),
+    }),
+    onSubmit: (values) => {
+      if (values.fileList) {
+        if (values.fileList.length <= 2) {
```
mas-who, line 76: I think we had a misunderstanding, the `else` block for `if (values.fileList)` should be kept since it would inform the user if they didn't select any files. Then one level down, we can do an early return like:
```
if (values.fileList.length > 2) {
        toastNotify.failure(...)
        return;
   }
```
so that we don't need to wrap the entire api call in an if else block.

edlerd (approved):
LGTM, thanks for adding the upload.

PR #849, opened 2024-08-15
feat: [WD-14036] Duplicate instance (lxc copy)

Kxiru:
## Done

- Added DuplicateInstance Button and Form
- Added duplicate instance API call.

## QA

1. Run the LXD-UI:
    - Navigate to an instance
    - Select "Duplicate instance" in header.
    - Fill out form and select "Duplicate & Create"
    - Verify that another instance has been created with the same details.

## Screenshots

image


piperdeck:
So I'm getting an error where it says it can't find the instance I'm duplicating in the project I've selected. Is it possible that the project dropdown is being used to look for the source instance, rather than specifying the destination?

https://github.com/user-attachments/assets/1603d968-b6bd-4509-9ff7-e3d19e886bc0

edlerd:
Some observations
* the "Duplicate & start" copy on the new modal deviates from the "Create and start" copy on creation. We should better align those
* the copy "Created and started cloned instance abcd-1." of the success notification reads very odd to me.
* the copy "Creation for instance abcd-1 started." does not transport that this is a clone operation. I am not sure if this is the right thing to say in this event.
* I think we can simplify by duplicating and not starting the cloned instance? Currently, we have a complex series of multiple API calls going on, adding complexity on error handling.

src/types/instance.d.ts
```diff
@@ -100,3 +100,13 @@ export interface LxdInstance {
   type: string;
   etag?: string;
 }
+
+export interface LxdInstanceDuplicate {
```
edlerd, line 104: This is the type of form values. I think it should live in the form component.

src/api/instances.tsx
```diff
@@ -333,3 +337,30 @@ export const fetchInstanceLogFile = (
...
+        description: instance.description,
+        name: values.instanceName,
+        source: {
+          allow_inconsistent: values.allowInconsistent,
+          instance_only: values.instanceOnly,
+          project: values.project,
+          source: instance.name,
+          type: "copy",
+        },
+        stateful: values.stateful,
```
edlerd, line 359: This transformation should not happen in the API layer, but in the form component. Here we can receive the json directly. Actually I think we do not need a new endpoint here at all, but can reuse the `createInstance` one above.

edlerd (changes requested):

src/api/instances.tsx
```diff
@@ -333,3 +337,30 @@ export const fetchInstanceLogFile = (
...
+    fetch(`/1.0/instances`, {
+      method: "POST",
+      body: JSON.stringify({
+        //Does not yet consider targetClusterMember or storagePool
+        description: instance.description,
+        name: values.instanceName,
+        source: {
+          allow_inconsistent: values.allowInconsistent,
+          instance_only: values.instanceOnly,
+          project: values.project,
```
mas-who, line 355: This line here relates to the issue @piperdeck reported. Looking at the [API doc](https://documentation.ubuntu.com/lxd/en/latest/api/#/instances/instances_post), `source.project` is where the source instance is located. The target project should be specified by the `?project=` query param.

mas-who: also `source.project` won't be needed since in the UI, the instance detail page is already part of a project context

src/types/instance.d.ts
```diff
@@ -100,3 +100,13 @@ export interface LxdInstance {
...
+  instanceName: string;
+  targetClusterMember?: string;
+  storagePool?: string;
+  project?: string;
```
mas-who, line 108: let's call this `targetProject` to be precise

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { Link } from "react-router-dom";
+
+interface Props {
+  instance: LxdInstance;
+  close: () => void;
+}
+
+export interface DuplicateInstanceFormValues {
+  instanceName: string;
+  project: string;
```
mas-who, line 32: let's call this `targetProject` to be precise

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+    toastNotify.success(
+      <>
+        Created and started cloned instance <b>{newInstance}</b>.
+      </>,
+    );
+  };
+
+  const formik = useFormik<DuplicateInstanceFormValues>({
+    initialValues: {
+      instanceName: instance.name + "-1",
```
mas-who, line 76: maybe "-copy" or "-duplicate" is more appropriate as a suffix here?

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+      project: "default",
+      allowInconsistent: false,
+      instanceOnly: false,
+      stateful: false,
+    },
+    validationSchema: Yup.object().shape({
+      instanceName: Yup.string().required(),
```
mas-who, line 83: we should validate that instance name does not exist for the selected target project. See `InstanceSchema` in `CreateInstance.tsx` for reference.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }),
+    onSubmit: (values) => {
+      duplicateInstance(values, instance)
```
mas-who, line 86: noted by @edlerd already, we should just use `createInstance` here and do the form data -> payload transformation here

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+      }
+    >
+      <Form onSubmit={formik.handleSubmit}>
+        <Input
+          {...formik.getFieldProps("instanceName")}
+          type="text"
+          label="New instance name"
+        />
+        {isClustered ?? (
+          <Select
```
mas-who, line 175: in general, any selector that requires external data fetching should be handled with care. In the context of this form component, we should disable the selectors based on their relevant data loading states.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+                label: clusterMember.server_name,
+                value: clusterMember.server_name,
+              };
+            })}
+          />
+        )}
+        <Select
+          {...formik.getFieldProps("storagePool")}
+          id="storagePool"
+          label="Storage pool"
```
mas-who, line 190: let's call this "Target storage pool", also please adjust the variable names please

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+            type="button"
+            onClick={close}
+          >
+            Cancel
+          </Button>
+          <ActionButton
+            appearance="positive"
+            className="u-no-margin--bottom"
+            loading={formik.isSubmitting}
+            disabled={!formik.isValid}
```
mas-who, line 160: let's also disable the submit button if data is still loading for any one of the selectors.

src/pages/instances/forms/DuplicateInstanceForm.tsx
Kxiru: Good idea

src/pages/instances/forms/DuplicateInstanceForm.tsx
Kxiru: Please could you clarify this and can you give an example of your suggestion?
Something I saw elsewhere in the codebase was disabling the selectors when there was only one option available (because obviously there is nothing else to choose from) is that what you are referring to?

edlerd: We should disable the fetchClusterMembers query in case clusterMembers is true.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  const notifySuccess = (instanceName: string) => {
+    const newInstance = (
+      <Link to={`/ui/project/${instance.project}/instance/${instanceName}`}>
+        {instanceName}
+      </Link>
+    );
+    toastNotify.success(
+      <>
+        Created and started cloned instance <b>{newInstance}</b>.
```
edlerd, line 70:
```suggestion
        Created the instance <b>{newInstance}</b>.
```
We don't start it any more, and maybe this copy is more precise / in alignment with the button / modal text?

I found it a bit confusing to have "duplicate" on the button, and then "cloned" in the notification. Probably we can keep it as the default naming has the `-duplicate` suffix.

edlerd (changes requested):
Thanks for the adjustments, this looks much cleaner already. Two further suggestions below.

src/pages/instances/forms/DuplicateInstanceForm.tsx
mas-who: I would use the `isLoading` state from your various api calls

Kxiru:
Alrighty, that's all been covered, @edlerd , please let me know what you think :)

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+          {...formik.getFieldProps("targetProject")}
+          id="project"
+          label="Target project"
+          options={projects.map((project) => {
+            return {
+              label: project.name,
+              value: project.name,
+            };
+          })}
+        />
```
edlerd, line 193:
```suggestion
```
I think it is not possible to duplicate an instance into another project. So better remove this input?

Can also remove the initial value, and the projects query above.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+  const { data: storagePools = [], isLoading: storagePoolsLoading } = useQuery({
+    queryKey: [queryKeys.storage],
+    queryFn: () => fetchStoragePools(instance.project),
+  });
+
+  const { data: clusterMembers = [], isLoading: clusterMembersLoading } =
+    useQuery({
+      queryKey: [queryKeys.cluster],
+      queryFn: fetchClusterMembers,
+      enabled: !isClustered,
```
edlerd, line 60:
```suggestion
      enabled: isClustered,
```
The condition needs to be flipped around. We only have cluster members to fetch if the server is clustered. Otherwise, the api returns an error on fetching cluster members.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+          type="checkbox"
+          label="Copy instance without its snapshots"
+          error={
+            formik.touched.instanceOnly ? formik.errors.instanceOnly : null
+          }
+        />
+        <Input
+          {...formik.getFieldProps("stateful")}
+          type="checkbox"
+          label="Copy stateless"
```
edlerd, line 215:
```suggestion
          label="Copy stateful"
```
The copy is wrong here.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,225 @@
...
+          error={
+            formik.touched.allowInconsistent
+              ? formik.errors.allowInconsistent
+              : null
+          }
+        />
+        <Input
+          {...formik.getFieldProps("instanceOnly")}
+          type="checkbox"
+          label="Copy instance without its snapshots"
```
edlerd, line 207:
```suggestion
          label="Copy without instance snapshots"
```
Maybe easier to read?

edlerd (changes requested):

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,204 @@
...
+      </>,
+    );
+  };
+
+  const formik = useFormik<LxdInstanceDuplicate>({
+    initialValues: {
+      instanceName: instance.name + "-duplicate",
+      allowInconsistent: false,
+      instanceOnly: false,
+      stateful: false,
```
edlerd, line 75:
```suggestion
      stateful: false,
      targetClusterMember: clusterMembers[0]?.server_name,
```
We need the default or it will pick a random one, even though we display the first one in the list.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,204 @@
...
+          name: values.instanceName,
+          source: {
+            allow_inconsistent: values.allowInconsistent,
+            instance_only: values.instanceOnly,
+            source: instance.name,
+            type: "copy",
+          },
+          stateful: values.stateful,
+        }),
+        instance.project,
```
edlerd, line 98:
```suggestion
        instance.project,
        values.targetClusterMember,
```
We should tell the api layer which cluster member is selected.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,204 @@
...
+        </>
+      }
+    >
+      <Form onSubmit={formik.handleSubmit}>
+        <Input
+          {...formik.getFieldProps("instanceName")}
+          type="text"
+          label="New instance name"
+        />
+        {isClustered ?? (
```
edlerd, line 149:
```suggestion
        {isClustered && (
```
Otherwise this never shows.

edlerd (dismissed):

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,205 @@
...
+          error={formik.touched.stateful ? formik.errors.stateful : null}
+        />
+        {/* hidden submit to enable enter key in inputs */}
+        <Input type="submit" hidden value="Hidden input" />
+      </Form>
+    </Modal>
+  );
+};
+
+export default DuplicateInstanceForm;
```
edlerd, line 205:
```suggestion
import { FC, useState } from "react";
import { LxdInstance } from "types/instance";
import { useFormik } from "formik";
import { useToastNotification } from "context/toastNotificationProvider";
import {
  ActionButton,
  Button,
  Form,
  Input,
  Modal,
  Select,
} from "@canonical/react-components";
import * as Yup from "yup";
import { createInstance } from "api/instances";
import { isClusteredServer } from "util/settings";
import { useSettings } from "context/useSettings";
import { useQuery } from "@tanstack/react-query";
import { queryKeys } from "util/queryKeys";
import { fetchStoragePools } from "api/storage-pools";
import { fetchClusterMembers } from "api/cluster";
import { Link } from "react-router-dom";
import { instanceNameValidation } from "util/instances";
import { LxdDiskDevice } from "types/device";
import InstanceLink from "pages/instances/InstanceLink";
import { useEventQueue } from "context/eventQueue";

interface Props {
  instance: LxdInstance;
  close: () => void;
}

export interface LxdInstanceDuplicate {
  instanceName: string;
  targetClusterMember?: string;
  targetStoragePool?: string;
  allowInconsistent?: boolean;
  instanceOnly?: boolean;
  stateful?: boolean;
}

const DuplicateInstanceForm: FC<Props> = ({ instance, close }) => {
  const toastNotify = useToastNotification();
  const { data: settings } = useSettings();
  const isClustered = isClusteredServer(settings);
  const controllerState = useState<AbortController | null>(null);
  const eventQueue = useEventQueue();

  const { data: storagePools = [], isLoading: storagePoolsLoading } = useQuery({
    queryKey: [queryKeys.storage],
    queryFn: () => fetchStoragePools(instance.project),
  });

  const { data: clusterMembers = [], isLoading: clusterMembersLoading } =
    useQuery({
      queryKey: [queryKeys.cluster],
      queryFn: fetchClusterMembers,
      enabled: isClustered,
    });

  const notifySuccess = (instanceName: string) => {
    const newInstance = (
      <Link to={`/ui/project/${instance.project}/instance/${instanceName}`}>
        {instanceName}
      </Link>
    );
    toastNotify.success(
      <>
        Created the instance <b>{newInstance}</b>.
      </>,
    );
  };

  const formik = useFormik<LxdInstanceDuplicate>({
    initialValues: {
      instanceName: instance.name + "-duplicate",
      allowInconsistent: false,
      instanceOnly: false,
      stateful: false,
      targetClusterMember: clusterMembers[0]?.server_name,
      targetStoragePool:
        (instance.devices["root"] as LxdDiskDevice)?.pool ??
        storagePools[0]?.name,
    },
    validationSchema: Yup.object().shape({
      instanceName: instanceNameValidation(
        instance.project,
        controllerState,
        instance.name,
      ).required(),
    }),
    onSubmit: (values) => {
      createInstance(
        JSON.stringify({
          description: instance.description,
          name: values.instanceName,
          source: {
            allow_inconsistent: values.allowInconsistent,
            instance_only: values.instanceOnly,
            source: instance.name,
            type: "copy",
          },
          stateful: values.stateful,
          devices: {
            ...instance.devices,
            root: {
              path: "/",
              type: "disk",
              pool: values.targetStoragePool,
            },
          },
        }),
        instance.project,
        values.targetClusterMember,
      )
        .then((operation) => {
          eventQueue.set(
            operation.metadata.id,
            () => {
              notifySuccess(values.instanceName);
            },
            (msg) =>
              toastNotify.failure(
                "Instance duplication failed",
                new Error(msg),
                <InstanceLink instance={instance} />,
              ),
            () => close(),
          );
        })
        .catch((e) => {
          toastNotify.failure(
            `Instance creation for "${values.instanceName}" failed.`,
            e,
          );
          close();
        });
    },
  });

  return (
    <Modal
      close={close}
      className="duplicate-instances-modal"
      title="Duplicate Instance"
      buttonRow={
        <>
          <Button
            appearance="base"
            className="u-no-margin--bottom"
            type="button"
            onClick={close}
          >
            Cancel
          </Button>
          <ActionButton
            appearance="positive"
            className="u-no-margin--bottom"
            loading={formik.isSubmitting}
            disabled={
              !formik.isValid || storagePoolsLoading || clusterMembersLoading
            }
            onClick={() => void formik.submitForm()}
          >
            Duplicate
          </ActionButton>
        </>
      }
    >
      <Form onSubmit={formik.handleSubmit}>
        <Input
          {...formik.getFieldProps("instanceName")}
          type="text"
          label="New instance name"
          error={
            formik.touched.instanceName ? formik.errors.instanceName : null
          }
        />
        {isClustered && (
          <Select
            {...formik.getFieldProps("targetClusterMember")}
            id="targetClusterMember"
            label="Target cluster member"
            options={clusterMembers.map((clusterMember) => {
              return {
                label: clusterMember.server_name,
                value: clusterMember.server_name,
              };
            })}
          />
        )}
        <Select
          {...formik.getFieldProps("targetStoragePool")}
          id="storagePool"
          label="Storage pool"
          options={storagePools.map((storagePool) => {
            return {
              label: storagePool.name,
              value: storagePool.name,
            };
          })}
        />
        <Input
          {...formik.getFieldProps("allowInconsistent")}
          type="checkbox"
          label="Ignore copy errors for volatile files"
          error={
            formik.touched.allowInconsistent
              ? formik.errors.allowInconsistent
              : null
          }
        />
        <Input
          {...formik.getFieldProps("instanceOnly")}
          type="checkbox"
          label="Copy without instance snapshots"
          error={
            formik.touched.instanceOnly ? formik.errors.instanceOnly : null
          }
        />
        <Input
          {...formik.getFieldProps("stateful")}
          type="checkbox"
          label="Copy stateful"
          error={formik.touched.stateful ? formik.errors.stateful : null}
        />
        {/* hidden submit to enable enter key in inputs */}
        <Input type="submit" hidden value="Hidden input" />
      </Form>
    </Modal>
  );
};

export default DuplicateInstanceForm;
```
Adding error handling and support for the disk device

edlerd (changes requested):

src/pages/instances/forms/DuplicateInstanceForm.tsx
Kxiru: Oof, thank you for this.

Kxiru: I get this error when trying to duplicate an instance for the first time.
It is interesting because if I attempt to duplicate after this, I can. Or if I manually set the default, I can. This also resets every time the page is refreshed.
image

piperdeck (dismissed):
It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...

CleanShot 2024-08-19 at 15 44 20

src/pages/instances/forms/DuplicateInstanceForm.tsx
mas-who: Saw comment from @piperdeck regarding the project selector. @edlerd according to the [lxc copy](https://documentation.ubuntu.com/lxd/en/latest/reference/manpages/lxc/copy/) docs, it's possible to specify a target project for the copy operation (using the `--target-project` flag). I think we should include the project selector here?

mas-who:
> It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...

@piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).

src/pages/instances/forms/DuplicateInstanceForm.tsx
edlerd: I updated the suggestion above. There was missing the `enableReinitialize` setting on formik and the notification should be with an info and closing the modal right after submit and only once the background task to duplicate succeeds or fails we should show the final notification.

edlerd:
> > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> 
> @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).

I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.

mas-who:
> > > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> > 
> > 
> > @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).
> 
> I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.

@edlerd I think it is done using the `project` query parameter for the API call?

edlerd:
> > > > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> > > 
> > > 
> > > @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).
> > 
> > 
> > I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.
> 
> @edlerd I think it is done using the `project` query parameter for the API call?

Tested that, and it didn't work when duplicating from default into a different project.

mas-who:
> > > > > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> > > > 
> > > > 
> > > > @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).
> > > 
> > > 
> > > I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.
> > 
> > 
> > @edlerd I think it is done using the `project` query parameter for the API call?
> 
> Tested that, and it didn't work when duplicating from default into a different project.
> 
> Edit: It failed to find the source instance, as that would live in another project as the project param value.

Good catch! Sorry I haven't tested this myself but I think the trick is to set the `source.project` field in the payload. Looking at the [source code](https://github.com/canonical/lxd/blob/35c83ae168886d7aa5a4b0f5d0c5895df5b2476e/lxd/instances_post.go#L1101) it seems like if `source.project` is not set, then it gets set to the `targetProjectName` which is extracted from the request query params. This explains the behaviour you are experiencing.

Kxiru:
> > > > > > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> > > > > 
> > > > > 
> > > > > @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).
> > > > 
> > > > 
> > > > I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.
> > > 
> > > 
> > > @edlerd I think it is done using the `project` query parameter for the API call?
> > 
> > 
> > Tested that, and it didn't work when duplicating from default into a different project.
> > Edit: It failed to find the source instance, as that would live in another project as the project param value.
> 
> Good catch! Sorry I haven't tested this myself but I think the trick is to set the `source.project` field in the payload. Looking at the [source code](https://github.com/canonical/lxd/blob/35c83ae168886d7aa5a4b0f5d0c5895df5b2476e/lxd/instances_post.go#L1101) it seems like if `source.project` is not set, then it gets set to the `targetProjectName` which is extracted from the request query params. This explains the behaviour you are experiencing.
> 
> Edit: Tested this, it works if the `project` query param is set to the target project name, and in the payload `source.project` must be set to the current project for the instance.

I will reimplement this when I get around to this PR.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+      <>
+        Created instance <b>{newInstance}</b>.
+      </>,
+    );
+  };
+
+  const formik = useFormik<LxdInstanceDuplicate>({
+    initialValues: {
+      instanceName: instance.name + "-duplicate",
+      targetProject: "default",
```
edlerd, line 83:
```suggestion
      targetProject: instance.project,
```
Shall we better use the current project?

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+      instanceName: instanceNameValidation(
+        instance.project,
+        controllerState,
+        instance.name,
+      ).required(),
+    }),
+    onSubmit: (values) => {
+      createInstance(
+        JSON.stringify({
+          //Does not yet consider targetClusterMember or storagePool
```
edlerd, line 103:
```suggestion
```
Probably not needed anymore?

src/api/instances.tsx
```diff
@@ -36,14 +36,17 @@ export const fetchInstances = (project: string): Promise<LxdInstance[]> => {
 
 export const createInstance = (
   body: string,
-  project: string,
-  target?: string,
+  targetProject: string,
+  targetCluster?: string,
```
edlerd, line 40: Shall we better keep the old names, as they are exactly matching the URL parameter names of the API route that is being called?

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+}
+
+export interface LxdInstanceDuplicate {
+  instanceName: string;
+  targetProject: string;
+  targetClusterMember?: string;
+  targetStoragePool?: string;
+  allowInconsistent?: boolean;
+  instanceOnly?: boolean;
+  stateful?: boolean;
```
edlerd, line 40:
```suggestion
  targetStoragePool: string;
  allowInconsistent: boolean;
  instanceOnly: boolean;
  stateful: boolean;
```
I think those frou are not optional and always set.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+      targetStoragePool:
+        (instance.devices["root"] as LxdDiskDevice)?.pool ??
+        storagePools[0]?.name,
+    },
+    enableReinitialize: true,
+    validationSchema: Yup.object().shape({
+      instanceName: instanceNameValidation(
+        instance.project,
+        controllerState,
+        instance.name,
```
edlerd, line 97:
```suggestion
```
This is only meant for renaming.

edlerd (commented):
QA LGTM, some small nitpicks on the code.

src/pages/instances/InstanceDetailHeader.tsx
edlerd, line 1: Please rebase this on main, as another PR just landed with another button.

Kxiru:
Thanks, @edlerd , let me know your thoughts about this latest version :)

edlerd (dismissed):
LGTM, thanks for adding the duplicate feature!

piperdeck:
Got this strange cryptic error:

CleanShot 2024-08-20 at 14 26 40

It's both uninformative and also probably a bug, because I don't think there's any issue with the architecture. The instance is being duplicated to the same node as the original. This happened when I just clicked duplicate and then didn't change anything in the modal. 

> > > > > > > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> > > > > > 
> > > > > > 
> > > > > > @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).
> > > > > 
> > > > > 
> > > > > I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.
> > > > 
> > > > 
> > > > @edlerd I think it is done using the `project` query parameter for the API call?
> > > 
> > > 
> > > Tested that, and it didn't work when duplicating from default into a different project.
> > > Edit: It failed to find the source instance, as that would live in another project as the project param value.
> > 
> > 
> > Good catch! Sorry I haven't tested this myself but I think the trick is to set the `source.project` field in the payload. Looking at the [source code](https://github.com/canonical/lxd/blob/35c83ae168886d7aa5a4b0f5d0c5895df5b2476e/lxd/instances_post.go#L1101) it seems like if `source.project` is not set, then it gets set to the `targetProjectName` which is extracted from the request query params. This explains the behaviour you are experiencing.
> > Edit: Tested this, it works if the `project` query param is set to the target project name, and in the payload `source.project` must be set to the current project for the instance.
> 
> I will reimplement this when I get around to this PR.

It seems that this issue is still occurring for me. Is there any update on this @Kxiru ?

Kxiru:
> Got this strange cryptic error:
> 
> CleanShot 2024-08-20 at 14 26 40
> 
> It's both uninformative and also probably a bug, because I don't think there's any issue with the architecture. The instance is being duplicated to the same node as the original. This happened when I just clicked duplicate and then didn't change anything in the modal.
> 
> > > > > > > > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> > > > > > > 
> > > > > > > 
> > > > > > > @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).
> > > > > > 
> > > > > > 
> > > > > > I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.
> > > > > 
> > > > > 
> > > > > @edlerd I think it is done using the `project` query parameter for the API call?
> > > > 
> > > > 
> > > > Tested that, and it didn't work when duplicating from default into a different project.
> > > > Edit: It failed to find the source instance, as that would live in another project as the project param value.
> > > 
> > > 
> > > Good catch! Sorry I haven't tested this myself but I think the trick is to set the `source.project` field in the payload. Looking at the [source code](https://github.com/canonical/lxd/blob/35c83ae168886d7aa5a4b0f5d0c5895df5b2476e/lxd/instances_post.go#L1101) it seems like if `source.project` is not set, then it gets set to the `targetProjectName` which is extracted from the request query params. This explains the behaviour you are experiencing.
> > > Edit: Tested this, it works if the `project` query param is set to the target project name, and in the payload `source.project` must be set to the current project for the instance.
> > 
> > 
> > I will reimplement this when I get around to this PR.
> 
> It seems that this issue is still occurring for me. Is there any update on this @Kxiru ?



> Got this strange cryptic error:
> 
> CleanShot 2024-08-20 at 14 26 40
> 
> It's both uninformative and also probably a bug, because I don't think there's any issue with the architecture. The instance is being duplicated to the same node as the original. This happened when I just clicked duplicate and then didn't change anything in the modal.
> 
> > > > > > > > It looks like the "Target project" dropdown isn't showing for me anymore. Was this intentional? Can't find any discussion about this...
> > > > > > > 
> > > > > > > 
> > > > > > > @piperdeck this is related to this [comment](https://github.com/canonical/lxd-ui/pull/849#discussion_r1721863207).
> > > > > > 
> > > > > > 
> > > > > > I couldn't find a way to set the target project via api. Is there a way? If there is we should probably bring the selector back.
> > > > > 
> > > > > 
> > > > > @edlerd I think it is done using the `project` query parameter for the API call?
> > > > 
> > > > 
> > > > Tested that, and it didn't work when duplicating from default into a different project.
> > > > Edit: It failed to find the source instance, as that would live in another project as the project param value.
> > > 
> > > 
> > > Good catch! Sorry I haven't tested this myself but I think the trick is to set the `source.project` field in the payload. Looking at the [source code](https://github.com/canonical/lxd/blob/35c83ae168886d7aa5a4b0f5d0c5895df5b2476e/lxd/instances_post.go#L1101) it seems like if `source.project` is not set, then it gets set to the `targetProjectName` which is extracted from the request query params. This explains the behaviour you are experiencing.
> > > Edit: Tested this, it works if the `project` query param is set to the target project name, and in the payload `source.project` must be set to the current project for the instance.
> > 
> > 
> > I will reimplement this when I get around to this PR.
> 
> It seems that this issue is still occurring for me. Is there any update on this @Kxiru ?

Hey, Piper,
I have not been able to recreate your issue. I am doing some investigation.

edlerd (approved):
LGTM

Kxiru:
@Kxiru , Validate instance name(s) and increment tail if name already in use.

src/util/instanceDuplicate.tsx
```diff
@@ -0,0 +1,24 @@
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "./queryKeys";
+import { fetchInstances } from "api/instances";
+import { LxdInstance } from "types/instance";
+
+export const getDuplicatedInstanceName = (instance: LxdInstance): string => {
```
edlerd, line 6: This should better be located inside the DuplicateInstanceForm.

src/util/instanceDuplicate.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const { data: instances = [] } = useQuery({
+    queryKey: [queryKeys.instances],
+    queryFn: () => fetchInstances(instance.project),
```
edlerd, line 9:
```suggestion
    queryFn: () => fetchInstances(targetProject),
```
We should use the selected target project to pick a free name here, not the project the source instance lives in.

src/sass/_instance_detail_page.scss
```diff
@@ -8,8 +8,10 @@
   }
 }
 
+.duplicate-instances-modal,
 .create-image-from-instance-modal {
   .p-modal__dialog {
+    justify-self: center;
```
edlerd, line 14: Why is the justify-self needed?

Kxiru: The modal is not centered when someone opens something like the inspect-tab, vertically on the page, but it is now :) 
image

edlerd: It was always centred, wasn't it? Chrome dev tools report this as ignored:
image
I am confused.

src/util/instanceDuplicate.tsx
Kxiru: I don't think this information is accessible at this time as this function is being called to pass an initialValue to the form, and also from within the form I do not believe it can access it's own values.

edlerd: You are right, the initial value of the target project will match the instance project. And if the user later changes the target project, we won't regenerate a name -- so it should be fine to use the instance project here.

edlerd (commented):
One last comment on the hook for the deduplication, and there is the open discussion on the CSS change, then this should be good from my side.

src/sass/_instance_detail_page.scss
mas-who: The default behaviour for `.p-modal__dialog` from vanilla for smaller screen sizes should have the dialog stretch across 100vw with some padding.
Screenshot from 2024-08-22 08-54-10
The style override here sets a fixed width for `.p-modal__dialog` so for smaller screen sizes the modal would be off center.
Screenshot from 2024-08-22 08-53-30

I think either we override the modal width only for large screen size:
```
.p-modal__dialog {
    @include large{
      width: 35rem;
    }
  }
```
Or we can set margin to auto so that the modal will be centered for smaller screen sizes:
```
.p-modal__dialog {
    margin: auto;
    width: 35rem;
  }
```
The first approach might be preferable I rate?

It is however quite interesting that the browser shows `justify-self` not having an effect since removing the styling does revert the modal to what was shown in the first screenshot :thinking:

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+  };
+
+  const getDuplicatedInstanceName = (instance: LxdInstance): string => {
+    const { data: instances = [] } = useQuery({
+      queryKey: [queryKeys.instances],
+      queryFn: () => fetchInstances(instance.project),
+    });
+
+    const instanceNames = instances.map((instance) => instance.name);
+    const newInstanceName = instance.name + "-duplicate";
```
mas-who, line 87: there is one condition we need to check here, is to ensure `instance.name + "-duplicate"` does not exceed 63 characters thus violating lxd instance naming constraint. You can use `truncateInstanceName` from `util/instances` for this like `newInstanceName = truncateInstanceName(instance.name, "-duplicate")`. Then rest of the logic can remain the same.

src/sass/_instance_detail_page.scss
edlerd: +1 on limiting the size only for large screens

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+  });
+
+  const { data: clusterMembers = [], isLoading: clusterMembersLoading } =
+    useQuery({
+      queryKey: [queryKeys.cluster],
+      queryFn: fetchClusterMembers,
+      enabled: isClustered,
+    });
+
+  const notifySuccess = (instanceName: string, instanceProject: string) => {
```
mas-who, line 67: I think we can have a consistent approach here to how upload instance notification works. Instead of showing a link for the instance name, we can add a `Configure` action to the bottom of the notification. You can refer to `handleSuccess` in `UploadInstanceBtn` for how it is done. Make sure you have the latest main branch rebased though since it's a recently introduced change.

edlerd: Good idea :+1:

mas-who: We should also add invalidation for instance related queries here:
```
void queryClient.invalidateQueries({
      predicate: (query) => {
        return query.queryKey[0] === queryKeys.instances;
      },
    });
```

In general, a successful API call that would modify some resource in the backend, we should consider invalidating the query cache so that we don't get stale data in another part of the app where we are making `GET` api calls for the same resource.

mas-who:
@Kxiru just a few minor comments from my end and also good to go, great work! :)

src/pages/instances/forms/DuplicateInstanceForm.tsx
Kxiru: I've implemented it as you put it above, however why is this syntax different when the usual syntax is like the following:

```
        void queryClient.invalidateQueries({
          queryKey: [queryKeys.instances],
        });
```

mas-who: We need to invalidate the query because a new instance is created here. This means if a query is cached for `GET /1.0/instances` (or any other relevant endpoints), we would get stale data and therefore not show the newly created instance. The `predicate` syntax is for invalidating query caches based on generic key structure matching, instead of exact query key matching since at this point we don't know which other instance related queries may have been cached.

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,290 @@
...
+      </>
+    );
+
+    const actions = [
+      {
+        label: `${instanceName}`,
+        onClick: () =>
+          navigate(`/ui/project/${instanceProject}/instance/${instanceName}`),
+      },
+    ];
```
mas-who, line 87: Let's align the messages structure
```suggestion
    const actions = [
      {
        label: "Configure",
        onClick: () =>
          navigate(`/ui/project/${instanceProject}/instance/${instanceName}/configuration`),
      },
    ];
```

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -0,0 +1,290 @@
...
+              pool: values.targetStoragePool,
+            },
+          },
+        }),
+        values.targetProject,
+        values.targetClusterMember,
+      )
+        .then((operation) => {
+          toastNotify.info(`Duplication of instance ${instance.name} started.`);
+          close();
```
mas-who, line 159: there is a call to `close()` in both `then` and `catch` block. In this case, you can call `close()` once inside a `finally` block

src/pages/instances/forms/DuplicateInstanceForm.tsx
Kxiru: Thanks a bunch, that makes sense!

edlerd (approved):
LGTM, thanks for adding the duplication!

PR #852, opened 2024-08-20
fix(volume) custom volume cache key needs to include the current project

edlerd:
## Done

- added project to the cache key of the custom volume list
- when switching between projects, the custom volume list needs to be refreshed

## QA

1. Run the LXD-UI:
   - create instance > disk > select custom volume, see currently available volumes (ensure at least one is uploaded). Change project to a project with an isolated storage and ensure the list is different.


mas-who:
LGTM, thanks for the fix

mas-who (approved):

PR #858, opened 2024-08-26
feat: Add create instance from snapshot feature [WD-14411]

mas-who:
## Done

- Add duplicate snapshot feature

## Notes
- Currently using the `copy` vanilla icon for the duplicate snapshot button
- Removed the `allowInconsistent` and `instanceOnly` options. These are not relevant for snapshot duplication.
- Stateful copy is only relevant for stateful snapshots
- For a stateful copy of a snapshot, the instance name must remain the same. This means the instance cannot be created in the same project. Tested this in the CLI, it's the same behaviour (see code [ref](https://github.com/canonical/lxd/blob/68e1b1bd544a4973828009fcf4b3d59a9836f46c/lxd/instances_post.go#L580)). Therefore, I added some additional form validation logic. If the Copy stateful option is checked, then validation is performed on instance name (must be the same as the original instance name) and target project (must be a different project to the source instance project). 
Screenshot from 2024-08-26 13-28-43

## QA

1. Run the LXD-UI:
    - Create an instance
    - Create a stateless snapshot and a stateful snapshot for the instance (`migration.stateful` must be `true` and the instance must be running)
    - Try duplicate both snapshots. Pay close attention to the stateful snapshot form validation.

### DIFF
--- a/src/pages/instances/actions/snapshots/InstanceSnapshotActions.tsx
+++ b/src/pages/instances/actions/snapshots/InstanceSnapshotActions.tsx
@@ -13,6 +13,7 @@ import ConfirmationForce from "components/ConfirmationForce";
 import { useEventQueue } from "context/eventQueue";
 import InstanceEditSnapshotBtn from "./InstanceEditSnapshotBtn";
 import CreateImageFromInstanceSnapshotBtn from "pages/instances/actions/snapshots/CreateImageFromInstanceSnapshotBtn";
+import CreateInstanceFromSnapshotBtn from "./CreateInstanceFromSnapshotBtn";
 
 interface Props {
   instance: LxdInstance;
@@ -103,13 +104,6 @@ const InstanceSnapshotActions: FC<Props> = ({
             isDeleting={isDeleting}
             isRestoring={isRestoring}
           />,
-          <CreateImageFromInstanceSnapshotBtn
-            key="publish"
-            instance={instance}
-            snapshot={snapshot}
-            isRestoring={isRestoring}
-            isDeleting={isDeleting}
-          />,
           <ConfirmationButton
             key="restore"
             appearance="base"
@@ -142,6 +136,20 @@ const InstanceSnapshotActions: FC<Props> = ({
           >
             <Icon name="change-version" />
           </ConfirmationButton>,
+          <CreateImageFromInstanceSnapshotBtn
+            key="publish"
+            instance={instance}
+            snapshot={snapshot}
+            isRestoring={isRestoring}
+            isDeleting={isDeleting}
+          />,
+          <CreateInstanceFromSnapshotBtn
+            key="duplicate"
+            instance={instance}
+            snapshot={snapshot}
+            isDeleting={isDeleting}
+            isRestoring={isRestoring}
+          />,
           <ConfirmationButton
             key="delete"
             appearance="base"
--- a/src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
+++ b/src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
@@ -1,5 +1,5 @@
 import { FC, useState } from "react";
-import { LxdInstance } from "types/instance";
+import { LxdInstance, LxdInstanceSnapshot } from "types/instance";
 import { useFormik } from "formik";
 import { useToastNotification } from "context/toastNotificationProvider";
 import {
@@ -17,7 +17,7 @@ import { useSettings } from "context/useSettings";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { fetchStoragePools } from "api/storage-pools";
-import { useNavigate } from "react-router-dom";
+import { Link } from "react-router-dom";
 import { instanceNameValidation, truncateInstanceName } from "util/instances";
 import { fetchProjects } from "api/projects";
 import { LxdDiskDevice } from "types/device";
@@ -27,25 +27,66 @@ import ClusterMemberSelector from "pages/cluster/ClusterMemberSelector";
 
 interface Props {
   instance: LxdInstance;
+  snapshot: LxdInstanceSnapshot;
   close: () => void;
 }
 
-export interface LxdInstanceDuplicate {
+interface CreateInstanceFromSnapshotValues {
   instanceName: string;
   targetProject: string;
   targetClusterMember?: string;
   targetStoragePool: string;
-  allowInconsistent: boolean;
-  instanceOnly: boolean;
   stateful: boolean;
 }
 
-const DuplicateInstanceForm: FC<Props> = ({ instance, close }) => {
+const instanceFromSnapshotPayload = (
+  values: CreateInstanceFromSnapshotValues,
+  instance: LxdInstance,
+  snapshot: LxdInstanceSnapshot,
+) => {
+  const basePayload: Record<string, unknown> = {
+    name: values.instanceName,
+    description: instance.description,
+    architecture: instance.architecture,
+    ephemeral: snapshot.ephemeral,
+    mode: "pull",
+    devices: {
+      ...instance.devices,
+      root: {
+        path: "/",
+        type: "disk",
+        pool: values.targetStoragePool,
+      },
+    },
+  };
+
+  const source: Record<string, unknown> = {
+    project: instance.project,
+    type: "copy",
+    source: `${instance.name}/${snapshot.name}`,
+    base_image: snapshot.config["volatile.base_image"] ?? "",
+  };
+
+  if (snapshot.stateful && values.stateful) {
+    basePayload["stateful"] = true;
+    source["live"] = false;
+  }
+
+  return {
+    ...basePayload,
+    source,
+  };
+};
+
+const CreateInstanceFromSnapshotForm: FC<Props> = ({
+  instance,
+  snapshot,
+  close,
+}) => {
   const toastNotify = useToastNotification();
   const { data: settings } = useSettings();
   const isClustered = isClusteredServer(settings);
   const controllerState = useState<AbortController | null>(null);
-  const navigate = useNavigate();
   const eventQueue = useEventQueue();
 
   const { data: projects = [], isLoading: projectsLoading } = useQuery({
@@ -63,29 +104,21 @@ const DuplicateInstanceForm: FC<Props> = ({ instance, close }) => {
     queryFn: () => fetchInstances(instance.project),
   });
 
-  const notifySuccess = (instanceName: string, instanceProject: string) => {
-    const message = (
-      <>
-        Created instance <strong>{instanceName}</strong>.
-      </>
+  const notifySuccess = (name: string, project: string) => {
+    const instanceLink = (
+      <Link to={`/ui/project/${project}/instance/${name}`}>{name}</Link>
     );
 
-    const actions = [
-      {
-        label: "Configure",
-        onClick: () =>
-          navigate(
-            `/ui/project/${instanceProject}/instance/${instanceName}/configuration`,
-          ),
-      },
-    ];
-
-    toastNotify.success(message, actions);
+    const message = <>Created instance {instanceLink}.</>;
+    toastNotify.success(message);
   };
 
-  const getDuplicatedInstanceName = (instance: LxdInstance): string => {
+  const getNewInstanceName = (instance: LxdInstance): string => {
     const instanceNames = instances.map((instance) => instance.name);
-    const newInstanceName = truncateInstanceName(instance.name, "-duplicate");
+    const newInstanceName = truncateInstanceName(
+      instance.name,
+      `-${snapshot.name}-copy`,
+    );
 
     if (instanceNames.includes(newInstanceName)) {
       let count = 1;
@@ -98,12 +131,27 @@ const DuplicateInstanceForm: FC<Props> = ({ instance, close }) => {
     return newInstanceName;
   };
 
-  const formik = useFormik<LxdInstanceDuplicate>({
+  const validate = (values: CreateInstanceFromSnapshotValues) => {
+    const errors: Partial<CreateInstanceFromSnapshotValues> = {};
+    if (values.stateful) {
+      if (values.targetProject === instance.project) {
+        errors.targetProject =
+          "Stateful instance snapshot duplication must be in a different project";
+      }
+
+      if (values.instanceName !== instance.name) {
+        errors.instanceName =
+          "Instance name must be the same for stateful duplication";
+      }
+    }
+
+    return errors;
+  };
+
+  const formik = useFormik<CreateInstanceFromSnapshotValues>({
     initialValues: {
-      instanceName: getDuplicatedInstanceName(instance),
+      instanceName: getNewInstanceName(instance),
       targetProject: instance.project,
-      allowInconsistent: false,
-      instanceOnly: false,
       stateful: false,
       targetClusterMember: isClustered ? instance.location : "",
       targetStoragePool:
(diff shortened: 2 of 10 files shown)


src/pages/instances/actions/snapshots/DuplicateInstanceSnapshotBtn.tsx
```diff
@@ -0,0 +1,48 @@
...
+import DuplicateInstanceSnapshotForm from "../../forms/DuplicateInstanceSnapshotForm";
+
+interface Props {
+  instance: LxdInstance;
+  snapshot: LxdInstanceSnapshot;
+  isDeleting: boolean;
+  isRestoring: boolean;
+}
+
+const DuplicateInstanceSnapshotBtn: FC<Props> = ({
```
edlerd, line 14: I would call this `CreateInstanceFromSnapshot`, wdyt?

src/pages/instances/actions/snapshots/DuplicateInstanceSnapshotBtn.tsx
```diff
@@ -0,0 +1,48 @@
...
+        </Portal>
+      )}
+      <Button
+        appearance="base"
+        hasIcon
+        dense
+        aria-label="Duplicate instance snapshot"
+        disabled={isDeleting || isRestoring}
+        onClick={openPortal}
+        title="Duplicate instance snapshot"
```
edlerd, line 40:
```suggestion
        title="Create instance"
```

src/pages/instances/actions/snapshots/DuplicateInstanceSnapshotBtn.tsx
```diff
@@ -0,0 +1,48 @@
...
+      >
+        <Icon name="copy" />
```
edlerd, line 42:
```suggestion
        <Icon name="plus" />
```
I think the copy icon is more related to copy to clipboard, maybe the plus is better suited?

src/pages/instances/actions/snapshots/InstanceSnapshotActions.tsx
```diff
@@ -95,6 +96,13 @@ const InstanceSnapshotActions: FC<Props> = ({
           "u-snapshot-actions": !isDeleting && !isRestoring,
         })}
         items={[
+          <DuplicateInstanceSnapshotBtn
+            key="duplicate"
+            instance={instance}
+            snapshot={snapshot}
+            isDeleting={isDeleting}
+            isRestoring={isRestoring}
+          />,
```
edlerd, line 105: I sugges to change the order:
1. edit
2. restore
3. create image
4. create instance (this one)
5. delete snapshot

src/pages/instances/forms/DuplicateInstanceSnapshotForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+    ];
+
+    toastNotify.success(message, actions);
+  };
+
+  const getDuplicatedInstanceName = (instance: LxdInstance): string => {
+    const instanceNames = instances.map((instance) => instance.name);
+    const newInstanceName = truncateInstanceName(
+      instance.name,
+      "-snapshot-duplicate",
```
edlerd, line 139:
```suggestion
      "-snapshot-copy",
```
Maybe better -- or we could also include the snapshot name here. so it is "instance-snapshot-copy"

src/pages/instances/forms/DuplicateInstanceSnapshotForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+export interface LxdInstanceSnapshotDuplicate {
+  instanceName: string;
+  targetProject: string;
+  targetClusterMember?: string;
+  targetStoragePool: string;
+  stateful: boolean;
+}
+
+export const duplicateSnapshotPayload = (
```
edlerd, line 42:
```suggestion
export const instanceFromSnapshotPayload = (
```
We are not duplicating a snapshot.

src/pages/instances/forms/DuplicateInstanceSnapshotForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+          close();
+        });
+    },
+  });
+
+  return (
+    <Modal
+      close={close}
+      className="duplicate-instances-modal"
+      title="Duplicate instance snapshot"
```
edlerd, line 227:
```suggestion
      title="Create instance from snapshot"
```
maybe clearer.

src/pages/instances/forms/DuplicateInstanceSnapshotForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+            loading={formik.isSubmitting}
+            disabled={
+              !formik.isValid ||
+              storagePoolsLoading ||
+              clusterMembersLoading ||
+              projectsLoading
+            }
+            onClick={() => void formik.submitForm()}
+          >
+            Duplicate
```
edlerd, line 250:
```suggestion
            Create
```

src/pages/instances/forms/DuplicateInstanceSnapshotForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+    const actions = [
+      {
+        label: "Configure",
+        onClick: () =>
+          navigate(
+            `/ui/project/${instanceProject}/instance/${instanceName}/configuration`,
+          ),
+      },
+    ];
```
edlerd, line 130: We probably don't need the configure link, as we copy the instance with all values.
We should link the instance in the message directly.

edlerd (commented):
Oh nice, a new feature :)
Some suggestions on naming and copy. I think generally we should call it "Create instance from snapshot". And also update notifications to "Instance created" or "Instance creation failed". The context of coming from a snapshot is probably not important for the notifications.

src/pages/instances/actions/snapshots/DuplicateInstanceSnapshotBtn.tsx
mas-who: Yes agreed, especially since we already have the `CreateInstanceFromImageBtn` component, better to keep naming consistent.

src/pages/instances/actions/snapshots/InstanceSnapshotActions.tsx
mas-who: I've adjusted the ordering.

src/pages/instances/forms/DuplicateInstanceSnapshotForm.tsx
mas-who: having the snapshot name is a great idea :+1:

src/pages/instances/forms/DuplicateInstanceSnapshotForm.tsx
mas-who: Also updated the other toast notification copy to reflect "creation" instead of "duplication".

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
```diff
@@ -0,0 +1,308 @@
...
+  });
+
+  const notifySuccess = (instanceName: string, instanceProject: string) => {
+    const instanceLink = (
+      <Link to={`/ui/project/${instanceProject}/instance/${instanceName}`}>
+        {instanceName}
+      </Link>
+    );
+
+    const message = <>Created Instance {instanceLink}.</>;
```
edlerd, line 121:
```suggestion
    const message = <>Created instance {instanceLink}.</>;
```
I missed this in the other PRs. We should only use "Instance" if it is at the beginning of a notification. Could you please fix the other occasions in `UploadInstanceBtn` and `DuplicateInstanceForm` as well?

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
```diff
@@ -0,0 +1,308 @@
...
+    toastNotify.success(message);
+  };
+
+  const getDuplicatedInstanceName = (instance: LxdInstance): string => {
```
edlerd, line 125:
```suggestion
  const getNewInstanceName = (instance: LxdInstance): string => {
```
Maybe better, as this is not about duplication?

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
```diff
@@ -0,0 +1,308 @@
...
+        .finally(() => {
+          close();
+        });
+    },
+  });
+
+  return (
+    <Modal
+      close={close}
+      className="duplicate-instances-modal"
```
edlerd, line 216:
```suggestion
```
Seems unused.

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
```diff
@@ -0,0 +1,308 @@
...
+            {...formik.getFieldProps("targetClusterMember")}
+            id="targetClusterMember"
+            label="Target cluster member"
+            options={clusterMembers.map((clusterMember) => {
+              return {
+                label: clusterMember.server_name,
+                value: clusterMember.server_name,
+              };
+            })}
+          />
```
edlerd, line 269: Should we create a reusable `ClusterMemberSelector`? We could also use it in the `DuplicateInstanceForm`.

edlerd (commented):
QA looks mostly good. I found one issue where the instance name is wrongly reported as the instance snapshot name. I think this comes from reading the values of the instance creation operation, that contains both the snapshot and the instance as a resource. See the attached screen for more details on this. This is also visible on the instance list, when the instance is being created. I think both paths (operation list and instance list) use the same function to get the instance name from the operation.

Code also mostly good, some small comments below.

Screenshot from 2024-08-26 16-12-19

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
mas-who: fixed them all

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
mas-who: I reused the class for styling, I've added an extra class `create-instance-from-snapshot-modal` now to avoid any confusion.

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
mas-who: done and implemented in `DuplicateInstanceForm`

mas-who:
> QA looks mostly good. I found one issue where the instance name is wrongly reported as the instance snapshot name. I think this comes from reading the values of the instance creation operation, that contains both the snapshot and the instance as a resource. See the attached screen for more details on this. This is also visible on the instance list, when the instance is being created. I think both paths (operation list and instance list) use the same function `getInstanceName` to get the instance name from the operation.
> 
> Code also mostly good, some small comments below.
> 
> Screenshot from 2024-08-26 16-12-19

Thanks for the hint here @edlerd , I made some adjustments to the `getInstanceName` function to filter out non-instance resources which should fix the issue you observed.

src/pages/instances/ClusterMemberSelector.tsx
```diff
@@ -0,0 +1,38 @@
+import { Select, SelectProps } from "@canonical/react-components";
+import { useQuery } from "@tanstack/react-query";
+import { fetchClusterMembers } from "api/cluster";
+import { useSettings } from "context/useSettings";
+import React, { FC } from "react";
+import { queryKeys } from "util/queryKeys";
+import { isClusteredServer } from "util/settings";
+
+const ClusterMemberSelector: FC<SelectProps> = ({
```
edlerd, line 9: this should better be placed under `pages/cluster/forms`

src/util/operations.tsx
```diff
@@ -5,10 +5,14 @@ export const getInstanceName = (operation?: LxdOperation): string => {
...
+  // /1.0/instances/<instance_name>/<snapshot_name>
   return (
     operation?.resources?.instances
-      ?.filter((item) => item.startsWith("/1.0/instances/"))
-      .map((item) => item.split("/")[3])
+      ?.filter((item) => {
+        // for snapshots backend returns %2F instead of /, which is not decoded by the browser
+        return decodeURIComponent(item).startsWith("/1.0/instances/");
+      })
+      .map((item) => decodeURIComponent(item).split("/")[3])
```
edlerd, line 15: I think this is the wrong way to do it, now it reports to be creating an instance with the source instance name. I think we can filter out items with either more than 3 `/` or with `%2F` in it. Then the instance list and operation list should link to the correct/new instance name.

edlerd (changes requested):

src/util/operations.tsx
mas-who: aaah I was filtering out any resource url with more than 4 segments if split by `/`. But then decided to not do that and failed to see the problem with the source instance name. Will revert to filtering out resources.

edlerd: I think those tests are relevant. I.e. when you create a snapshot, you can see an operation in the operation list. And that operation should link to the correct instance. Maybe my suggestion to filter out the snapshot URLs wasn't the best idea here. There should be a solution that works for snapshot creation. Keeping the operation list link to the correct instance in that case. As well as for the instance and operation list linking to the new instance, when creating an instance from snapshot.

mas-who: The snapshot creation would be linked to the correct instance using the current approach. Since `resources.instances` contains the source instance for the snapshot?

edlerd: I see, then the tests seem to be wrong. I was wondering maybe an older version did report the resources differently. But even with 5.0 I did get the instances / instances_snapshots structure you mentioned above. So sounds good to me to remove those tests and change the logic.

Can you please report this as a bug to LXD? I think it is wrong, that the instances list includes a link to the snapshot when creating an instance from a snapshot.

mas-who: Sure will send a message to the UI channel :+1:

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+export interface CreateInstanceFromSnapshotValues {
+  instanceName: string;
+  targetProject: string;
+  targetClusterMember?: string;
+  targetStoragePool: string;
+  stateful: boolean;
+}
+
+export const instanceFromSnapshotPayload = (
```
edlerd, line 41:
```suggestion
const instanceFromSnapshotPayload = (
```

src/util/operations.tsx
```diff
@@ -5,9 +5,18 @@ export const getInstanceName = (operation?: LxdOperation): string => {
...
     operation?.resources?.instances
-      ?.filter((item) => item.startsWith("/1.0/instances/"))
+      ?.filter((item) => {
+        // for snapshots backend returns %2F instead of /, which is not decoded by the browser
+        const itemUrl = decodeURIComponent(item);
+        return (
+          itemUrl.startsWith("/1.0/instances/") &&
+          // exclude snapshots
+          itemUrl.split("/").length <= 4
+        );
```
edlerd, line 18: Can you link to the lxd issue reporting this bug so we can eventually clean this up when it is fixed there?

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
```diff
@@ -0,0 +1,287 @@
...
+    queryKey: [queryKeys.storage],
+    queryFn: () => fetchStoragePools(instance.project),
+  });
+
+  const { data: instances = [] } = useQuery({
+    queryKey: [queryKeys.instances],
+    queryFn: () => fetchInstances(instance.project),
+  });
+
+  const notifySuccess = (instanceName: string, instanceProject: string) => {
```
edlerd, line 106:
```suggestion
  const notifySuccess = (name: string, project: string) => {
```

edlerd (dismissed):
QA and code LGTM, some nitpicks below. Maybe change the commit message also to say "Create instance from snapshot" the current "duplicate snapshot" is misleading.

src/pages/instances/forms/CreateInstanceFromSnapshotForm.tsx
```diff
@@ -0,0 +1,287 @@
...
+import InstanceLink from "pages/instances/InstanceLink";
+import { useEventQueue } from "context/eventQueue";
+import ClusterMemberSelector from "pages/cluster/ClusterMemberSelector";
+interface Props {
+  instance: LxdInstance;
+  snapshot: LxdInstanceSnapshot;
+  close: () => void;
+}
+
+export interface CreateInstanceFromSnapshotValues {
```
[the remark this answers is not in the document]
mas-who: correct :+1:

src/util/operations.tsx
mas-who: Haven't created the issue yet, I will link it to a Jira issue later.

edlerd (approved):

PR #860, opened 2024-08-26
simpler one click edit mode for edit forms

edlerd:
## Done

- simpler one click edit mode for edit forms on instances, profiles networks, storage pools, storage volumes and projects

Fixes WD-14393

## QA

1. Run the LXD-UI:
    - check edit and creation of instances, profiles networks, storage pools, storage volumes and projects

### DIFF
--- a/src/components/forms/DiskDeviceFormCustom.tsx
+++ b/src/components/forms/DiskDeviceFormCustom.tsx
@@ -1,5 +1,5 @@
-import { FC } from "react";
-import { Icon, Input, Label } from "@canonical/react-components";
+import React, { FC } from "react";
+import { Button, Icon, Input, Label } from "@canonical/react-components";
 import { InstanceAndProfileFormikProps } from "./instanceAndProfileFormValues";
 import { EditInstanceFormValues } from "pages/instances/EditInstance";
 import { InheritedVolume } from "util/configInheritance";
@@ -13,6 +13,7 @@ import DetachDiskDeviceBtn from "pages/instances/actions/DetachDiskDeviceBtn";
 import classnames from "classnames";
 import { LxdStorageVolume } from "types/storage";
 import { isDiskDeviceMountPointMissing } from "util/instanceValidation";
+import { ensureEditMode } from "util/instanceEdit";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
@@ -30,6 +31,10 @@ const DiskDeviceFormCustom: FC<Props> = ({
     .filter((device) => device.name !== "root" && device.type === "disk")
     .map((device) => device as FormDiskDevice);
 
+  const focusField = (name: string) => {
+    setTimeout(() => document.getElementById(name)?.focus(), 100);
+  };
+
   const addVolume = (volume: LxdStorageVolume) => {
     const copy = [...formik.values.devices];
     copy.push({
@@ -42,7 +47,7 @@ const DiskDeviceFormCustom: FC<Props> = ({
     void formik.setFieldValue("devices", copy);
 
     const name = `devices.${copy.length - 1}.path`;
-    setTimeout(() => document.getElementById(name)?.focus(), 100);
+    focusField(name);
   };
 
   const changeVolume = (
@@ -71,6 +76,22 @@ const DiskDeviceFormCustom: FC<Props> = ({
     return candidate;
   };
 
+  const editButton = (fieldName: string) => (
+    <Button
+      appearance="base"
+      className="u-no-margin--bottom"
+      hasIcon
+      dense
+      title="Edit"
+      onClick={() => {
+        ensureEditMode(formik);
+        focusField(fieldName);
+      }}
+    >
+      <Icon name="edit" />
+    </Button>
+  );
+
   const rows: MainTableRow[] = [];
 
   customVolumes.map((formVolume) => {
@@ -83,15 +104,20 @@ const DiskDeviceFormCustom: FC<Props> = ({
           <RenameDiskDeviceInput
             name={formVolume.name}
             index={index}
-            readOnly={readOnly}
-            setName={(name) =>
-              void formik.setFieldValue(`devices.${index}.name`, name)
-            }
+            setName={(name) => {
+              ensureEditMode(formik);
+              void formik.setFieldValue(`devices.${index}.name`, name);
+            }}
           />
         ),
         inherited: "",
-        override: !readOnly && (
-          <DetachDiskDeviceBtn onDetach={() => removeDevice(index, formik)} />
+        override: (
+          <DetachDiskDeviceBtn
+            onDetach={() => {
+              ensureEditMode(formik);
+              removeDevice(index, formik);
+            }}
+          />
         ),
       }),
     );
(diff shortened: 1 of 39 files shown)


Kxiru:
To clarify, should the one click editing only work on dropdown and text fields? Not the overridable aspects of configs?
image

edlerd:
> To clarify, should the one click editing only work on dropdown and text fields? Not the overridable aspects of configs?

Previously you had to click "Edit instance", then those pencil icons would appear. When clicking a pencil to create an override, the input element appears.

Now the pencil icons are always visible, clicking them will go into edit-instanc-mode and show the input element.

mas-who:
Thanks for this change, the interaction is really nice now for editing configurations on resources :+1: 

Only two smallish comments from QA:

1. On the network device form, seems like the config row cell content alignment is a little off. In non-edit mode, the configuration column cell value seems to be off vertical center.
Screenshot from 2024-08-27 18-33-24

2. When editing network main configurations, clicking the edit icon for one of the config row puts all other config rows in edit mode. I think this seems to be a little inconsistent with other pages where config rows can go into edit mode individually?
[Screencast from 27-08-2024 18:41:31.webm](https://github.com/user-attachments/assets/5db8437a-c12f-43a6-9a3c-b93e271c0386)

edlerd:
> 1. On the network device form, seems like the config row cell content alignment is a little off. In non-edit mode, the configuration column cell value seems to be off vertical center.
>    Screenshot from 2024-08-27 18-33-24

This is a local override, hence the placement on the right. I agree this looks off, though. Maybe we can adopt a similar layout as in disk device section: with inherited and custom devices as headings.

> 2. When editing network main configurations, clicking the edit icon for one of the config row puts all other config rows in edit mode. I think this seems to be a little inconsistent with other pages where config rows can go into edit mode individually?
>    [Screencast from 27-08-2024 18:41:31.webm](https://github.com/user-attachments/assets/5db8437a-c12f-43a6-9a3c-b93e271c0386)

This is intended any fields with an override will go into edit mode once the user starts editing.

Both of those are implemented as intended. Maybe @piperdeck has an idea how to improve it?

piperdeck:
> > 1. On the network device form, seems like the config row cell content alignment is a little off. In non-edit mode, the configuration column cell value seems to be off vertical center.
> 
> This is a local override, hence the placement on the right. I agree this looks off, though. Maybe we can adopt a similar layout as in disk device section: with inherited and custom devices as headings.

I think Mason was talking about this alignment: 

CleanShot 2024-08-27 at 16 47 37

I think it would look better if we could give it this alignment:

CleanShot 2024-08-27 at 16 57 32


> > 2. When editing network main configurations, clicking the edit icon for one of the config row puts all other config rows in edit mode. I think this seems to be a little inconsistent with other pages where config rows can go into edit mode individually?
> >    [Screencast from 27-08-2024 18:41:31.webm](https://github.com/user-attachments/assets/5db8437a-c12f-43a6-9a3c-b93e271c0386)
> 
> This is intended. Any fields with an override will go into edit mode once the user starts editing. I have no idea if or how we could improve this.
> 
> Both of those are implemented as intended. Maybe @piperdeck has an idea how to improve it?

As for that point, I agree that this is implemented as intended, and personally I think it's working well

mas-who:
On the devices form, for custom devices, the "Mount point", "Read limit" and "Write limit" configs are not editable. When I make a change for anything else on the devices form, then those fields for a custom device becomes editable. Is it possible to have the edit button next to those fields too?

mas-who:
The alignment for the custom device "Pool / volume" field seems off 
Screenshot from 2024-08-28 09-45-26

src/pages/instances/EditInstance.tsx
```diff
@@ -253,19 +257,15 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
         {readOnly ? (
-          <Button
-            appearance="positive"
-            onClick={() => {
-              void formik.setFieldValue("readOnly", false);
-            }}
-          >
-            Edit instance
-          </Button>
+          <div style={{ height: "3.5rem" }} />
```
mas-who, line 260: what if we add `min-height: 3.5rem` to the `bottom-btns` class used by the `footer` element in `FormFooterLayout` component instead of adding this `div` here? Applicable for all the other `Edit...` components.

src/pages/networks/CreateNetwork.tsx
```diff
@@ -117,6 +117,7 @@ const CreateNetwork: FC = () => {
         project={project}
         section={section}
         setSection={updateSection}
+        version={0}
```
mas-who, line 120: `version` prop could be optional so that we don't need to pass it in for `Create...` components?

mas-who (commented):
This is a great change, the code also look cleaner than before on top of the UI improvements! Left some comments for feedback :)

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -13,6 +13,7 @@ import DetachDiskDeviceBtn from "pages/instances/actions/DetachDiskDeviceBtn";
 import classnames from "classnames";
 import { LxdStorageVolume } from "types/storage";
 import { isDiskDeviceMountPointMissing } from "util/instanceValidation";
+import { ensureEditMode } from "util/instanceEdit";
```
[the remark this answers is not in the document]
edlerd: I added the edit buttons to those fields.

src/pages/instances/forms/EditInstanceDetails.tsx
```diff
@@ -76,8 +78,10 @@ const EditInstanceDetails: FC<Props> = ({ formik, project }) => {
       <ProfileSelect
```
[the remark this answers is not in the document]
edlerd: Good idea about the naming.
As for the read only, we still need that during instance creation, because before the user selects and image, the read only is set to true there.

src/pages/instances/CreateInstance.tsx
```diff
@@ -450,7 +450,6 @@ const CreateInstance: FC = () => {
 
             {section === YAML_CONFIGURATION && (
               <YamlForm
-                key={`yaml-form-${formik.values.readOnly}`}
```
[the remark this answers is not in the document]
edlerd: This was a workaround to force the yaml form to rerender when the user switched between edit and read mode.

src/pages/networks/CreateNetwork.tsx
edlerd: good idea, we can default it to 0

src/pages/instances/EditInstance.tsx
edlerd: I was thinking about this, we might want to keep the "Edit ..." buttons here actually instead of removing them.

edlerd: Actually a great idea to use the min-height instead, applied that change now.

src/components/forms/DiskDeviceFormCustom.tsx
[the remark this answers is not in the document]
mas-who: I think if the tests pass then it's fine, was just thinking since the interaction for those fields changed tests may fail.

mas-who:
Code and QA LGTM, thanks for the updates :+1:

edlerd:
Addressed all the design requests @piperdeck please have another look at this.

piperdeck:
Looks good to me!

PR #861, opened 2024-08-27
feat: [WD-14389] Combine instance detail page actions

Kxiru:
## Done

- Moving buttons into InstanceDetailHeaderDropdown component.
- Minor scss changes
- Changed DeleteInstanceBtn to no longer use an icon as the situation in which this exists (when page is minimised) has changed.

## QA

1. Run the LXD-UI:
    - Navigate to instance detail page on a medium+ sized screen and view the instance buttons in a row.
    - Then, minimise screen (Can use inspect element displayed vertically to achieve this) and refresh the page/render the page for the first time to see the icons as a dropdown.

## Screenshots
Medium Screen:
image
image

Large screen:
image

### DIFF
--- a/src/components/RenameHeader.tsx
+++ b/src/components/RenameHeader.tsx
@@ -99,7 +99,7 @@ const RenameHeader: FC<Props> = ({
               </li>
             ) : (
               <li
-                className="p-heading--4 u-no-margin--bottom name continuous-breadcrumb"
+                className="p-heading--4 u-no-margin--bottom name continuous-breadcrumb u-truncate"
                 onClick={toggleRename}
                 title={`Rename ${name}`}
               >
--- a/src/pages/instances/InstanceDetailActions.tsx
+++ b/src/pages/instances/InstanceDetailActions.tsx
@@ -1,0 +1,73 @@
+import { FC, cloneElement, useState } from "react";
+import DeleteInstanceBtn from "./actions/DeleteInstanceBtn";
+import { LxdInstance } from "types/instance";
+import MigrateInstanceBtn from "./actions/MigrateInstanceBtn";
+import CreateImageFromInstanceBtn from "./actions/CreateImageFromInstanceBtn";
+import DuplicateInstanceBtn from "./actions/DuplicateInstanceBtn";
+import { ContextualMenu } from "@canonical/react-components";
+import { isWidthBelow } from "util/helpers";
+import useEventListener from "@use-it/event-listener";
+
+interface Props {
+  instance: LxdInstance;
+  project: string;
+  isLoading: boolean;
+}
+
+const InstanceDetailActions: FC<Props> = ({ instance, project, isLoading }) => {
+  const [isLargeScreen, setIsLargeScreen] = useState(isWidthBelow(1200));
+
+  useEventListener("resize", () => setIsLargeScreen(isWidthBelow(1200)));
+  const classname = isLargeScreen ? "p-contextual-menu__link" : "";
+
+  const menuElements = [
+    <MigrateInstanceBtn
+      key="migrate"
+      instance={instance}
+      project={project}
+      classname={classname}
+    />,
+    <CreateImageFromInstanceBtn
+      key="publish"
+      instance={instance}
+      classname={classname}
+    />,
+    <DuplicateInstanceBtn
+      key="duplicate"
+      instance={instance}
+      isLoading={isLoading}
+      classname={classname}
+    />,
+    <DeleteInstanceBtn
+      key="delete"
+      instance={instance}
+      classname={classname}
+    />,
+  ];
+
+  return (
+    <>
+      {isLargeScreen ? (
+        <ContextualMenu
+          closeOnOutsideClick={false}
+          toggleLabel="Actions"
+          position="left"
+          hasToggleIcon
+          title="actions"
+        >
+          {(close: () => void) => (
+            <span>
+              {[...menuElements].map((item) =>
+                cloneElement(item, { onClose: close }),
+              )}
+            </span>
+          )}
+        </ContextualMenu>
+      ) : (
+        <>{menuElements}</>
+      )}
+    </>
+  );
+};
+
+export default InstanceDetailActions;
--- a/src/pages/instances/InstanceDetailHeader.tsx
+++ b/src/pages/instances/InstanceDetailHeader.tsx
@@ -1,6 +1,5 @@
 import { FC, useState } from "react";
 import { Link, useNavigate } from "react-router-dom";
-import DeleteInstanceBtn from "./actions/DeleteInstanceBtn";
 import { LxdInstance } from "types/instance";
 import RenameHeader, { RenameHeaderValues } from "components/RenameHeader";
 import { renameInstance } from "api/instances";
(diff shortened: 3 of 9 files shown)


src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -41,6 +43,7 @@ const InstanceDetailHeader: FC<Props> = ({
   const controllerState = useState<AbortController | null>(null);
   const { data: settings } = useSettings();
   const isClustered = isClusteredServer(settings);
+  const isMediumScreen = () => isWidthBelow(820);
```
edlerd, line 46: This is not re-run when resizing the window. We might need a local state variable `isMidumScreen` here and add a `useEventListener("resize",...)`, that recalculates the state on the resize event to make it responsive.

src/pages/instances/InstanceDetailHeaderDropdown.tsx
```diff
@@ -0,0 +1,58 @@
...
+              instance={instance}
+              project={project}
+              dropdownItemClassname="dropdownItemClassname"
+            />
+          ) : null}
+          <CreateImageFromInstanceBtn
+            key="publish"
+            instance={instance}
+            dropdownItemClassname="dropdownItemClassname"
+          />
```
edlerd, line 40: This should look like in the [storybook example](https://canonical.github.io/react-components/?path=/story/components-contextualmenu--toggle). Without border and left aligned text. We might be able to pass the `p-contextual-menu__link` class to the buttons to get it get the layout correct. The custom `dropdownItemClassname` is probably not needed then.

src/pages/instances/InstanceDetailHeaderDropdown.tsx
```diff
@@ -0,0 +1,58 @@
...
+  project,
+  isLoading,
+}) => {
+  const { data: settings } = useSettings();
+  const isClustered = isClusteredServer(settings);
+
+  return (
+    <>
+      {instance ? (
+        <ContextualMenu toggleLabel="Options" hasToggleIcon visible>
```
edlerd, line 28:
```suggestion
        <ContextualMenu toggleLabel="Actions" hasToggleIcon visible>
```
I think we should advertise this as actions.

The contextual menu seems to be open when landing on the page. Initially it should be closed.

edlerd (changes requested):
The buttons should responsively collapse when resizing the browser

src/pages/instances/InstanceDetailHeader.tsx
Kxiru: Oop, I forgot to implement this in the recent push. Please hold.

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -125,20 +128,34 @@ const InstanceDetailHeader: FC<Props> = ({
...
+                    key="publish"
+                    instance={instance}
+                  />
+                  <DuplicateInstanceBtn
+                    key="duplicate"
+                    instance={instance}
+                    isLoading={isLoading}
+                  />
+                  <DeleteInstanceBtn key="delete" instance={instance} />
+                </>
```
edlerd, line 155: Probably best to move the buttons and all the logic to switch between dropdown and buttons into a dedicated `InstanceDetailActions` component. Maybe we can merge it with the `InstanceDetailHeaderDropdown`.

I think we can define an array of `actions = []` to which we push the buttons (depending on their conditions). Reuse that array for the context menu and for the button rendering. So we don't define the actions and their conditions twice. That would be hard to maintain in the long run and create an extra need to keep the two consistent. Best to rely on a single source of truth for it.

src/pages/instances/InstanceDetailHeader.tsx
Kxiru: All done!

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -39,8 +36,9 @@ const InstanceDetailHeader: FC<Props> = ({
   const navigate = useNavigate();
   const toastNotify = useToastNotification();
   const controllerState = useState<AbortController | null>(null);
-  const { data: settings } = useSettings();
-  const isClustered = isClusteredServer(settings);
+  const [isLargeScreen, setIsLargeScreen] = useState(isWidthBelow(1200));
+
+  useEventListener("resize", () => setIsLargeScreen(isWidthBelow(1200)));
```
edlerd, line 41: I think this can move into the `InstanceDetailActions`, we don't need it here, right?

src/pages/instances/InstanceDetailActions.tsx
```diff
@@ -0,0 +1,74 @@
...
+  isLargeScreen: boolean;
+}
+
+const InstanceDetailActions: FC<Props> = ({
+  instance,
+  project,
+  isLoading,
+  isLargeScreen,
+}) => {
+  const hasMediumScreenClass = isLargeScreen ? "p-contextual-menu__link" : "";
```
edlerd, line 22:
```suggestion
  const classname = isLargeScreen ? "p-contextual-menu__link" : "";
```
I think this better describes the variable contents.

src/pages/instances/actions/CreateImageFromInstanceBtn.tsx
```diff
@@ -18,19 +25,28 @@ const CreateImageFromInstanceBtn: FC<Props> = ({ instance }) => {
...
+            close={handleClose}
             instance={instance}
           />
         </Portal>
       )}
       <ActionButton
         appearance="default"
-        onClick={openPortal}
+        className={classNames("u-no-margin--bottom", classname)}
+        onClick={handleOpen}
```
edlerd, line 49:
```suggestion
        onClick={openPortal}
```
The `handleOpen` function seems redundant. Same for two other buttons below.

edlerd (commented):
QA LGTM, some small nitpicks on the code to simplify slightly.

I think three more changes are needed to make the header compatible with very large instance names. I.e. check with [this instance with a very long name on the demo server](https://lxd-ui-861.demos.haus/ui/project/default/instance/mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm). Below steps should make it truncate and avoid breaking the layout with any name:

1. Make the `.p-breadcrumbs--large .p-breadcrumbs__items .continuous-breadcrumb` as display `inline-block`
2. In the `RenameHeader`, where we use the `continuous-breadcrumb` class, we should also add `u-truncate`
3. Reduce the max width of `.rename-header .p-panel__title .name` to `26vw`.

image

Kxiru:
> QA LGTM, some small nitpicks on the code to simplify slightly.
> 
> I think three more changes are needed to make the header compatible with very large instance names. I.e. check with [this instance with a very long name on the demo server](https://lxd-ui-861.demos.haus/ui/project/default/instance/mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm). Below steps should make it truncate and avoid breaking the layout with any name:
> 
> 1. Make the `.p-breadcrumbs--large .p-breadcrumbs__items .continuous-breadcrumb` as display `inline-block`
> 2. In the `RenameHeader`, where we use the `continuous-breadcrumb` class, we should also add `u-truncate`
> 3. Reduce the max width of `.rename-header .p-panel__title .name` to `26vw`. (used to be `33vw`).
> 
> image

Thanks for the heads up, David. The header looks a lot better with long names, now.

src/components/RenameHeader.tsx
```diff
@@ -59,7 +59,7 @@ const RenameHeader: FC<Props> = ({
           <ol className="p-breadcrumbs__items">
             {parentItems.map((item, key) => (
               <li
-                className="p-heading--4 u-no-margin--bottom continuous-breadcrumb"
+                className="p-heading--4 u-no-margin--bottom continuous-breadcrumb u-truncate"
```
edlerd, line 62: The truncate class should be added to the 2nd item in the header further down in line 102 instead. This here is the parent -- the "Instances" link, which doesn't need truncation.

edlerd (commented):
Thanks for the changes, I really like how clean the header and the code is now. Just one small issue left, then this should be good to merge.

Kxiru:
> Thanks for the changes, I really like how clean the header and the code is now. Just one small issue left, then this should be good to merge.

Thanks for the clarification. Please let me know your thoughts now :).

edlerd (approved):
LGTM :+1:

PR #862, opened 2024-08-27
feat: [WD-14391] instance duplication tests

Kxiru:
## Done

- Created duplicate-instance.spec.ts file

## QA

1. Run the LXD-UI:
    - Within the codebase, run the playwright test.

## Screenshots

N/A

### DIFF
--- a/tests/instances.spec.ts
+++ b/tests/instances.spec.ts
@@ -224,3 +224,15 @@ test("instance yaml edit", async ({ page }) => {
   await page.getByText("Main configuration").click();
   await assertTextVisible(page, "DescriptionA-new-description");
 });
+
+test("Duplicate an instance", async ({ page }) => {
+  await visitInstance(page, instance);
+  await page.getByRole("button", { name: "Duplicate" }).click();
+  await page
+    .getByRole("dialog", { name: "Duplicate Instance" })
+    .getByRole("button", { name: "Duplicate" })
+    .click();
+
+  await page.waitForSelector(`text=Created instance ${instance}-duplicate.`);
+  await deleteInstance(page, `${instance}-duplicate`);
+});


tests/duplicate-instance.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { test } from "./fixtures/lxd-test";
+import { deleteAllImages } from "./helpers/images";
+import {
+  createInstance,
+  deleteInstance,
+  randomInstanceName,
+  visitInstance,
+} from "./helpers/instances";
+
+test("Duplicate an instance", async ({ page }) => {
```
edlerd, line 10: I think this can go into the instances.spec.ts suite

tests/duplicate-instance.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+
+  await page.goto("/ui/");
+  await visitInstance(page, instance);
```
edlerd, line 15:
```suggestion
  await visitInstance(page, instance);
```
We can use a shared instance from the instance suite.

tests/duplicate-instance.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  await page.getByRole("button", { name: "Duplicate" }).click();
+  await page
+    .getByRole("dialog", { name: "Duplicate Instance" })
+    .getByRole("button", { name: "Duplicate" })
+    .click();
+
+  await page.waitForSelector(`text=Created instance ${instance}-duplicate.`);
+
+  await deleteAllImages(page);
+  await deleteInstance(page, instance);
```
edlerd, line 25:
```suggestion
```
When using the shared instance, we don't need to clean up.
But we do need to remove the duplicate instance that gets created above.

edlerd (changes requested):

PR #863, opened 2024-08-27
feat: support vmdk import for vm instance creation

mas-who:
## Note
- This is currently a draft spike for importing VMDK support in the UI. The aim for now is to show case how to import the VMDK file by using the operations websocket from lxd.

## Done
- Add support for VMDK import to create VM instances.

## QA

1. Run the LXD-UI:
    - TBD


src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,328 @@
...
+    ];
+
+    toastNotify.success(message, actions);
+  };
+
+  const handleFailure = (msg: string) => {
+    toastNotify.failure("Instance creation failed.", new Error(msg));
+  };
+
+  const handleFinish = () => {
```
edlerd, line 134:
```suggestion
  const invalidateCache = () => {
```

src/util/helpers.tsx
```diff
@@ -290,3 +290,92 @@ export const getClientOS = (userAgent: string) => {
...
+      "riscv32",
+      "riscv64",
+      "armv6l",
+      "armv8l",
+      "loongarch64",
+    ];
+  }
+
+  return ["x86_64", "aarch64", "ppc64le", "s390x"];
+};
```
edlerd, line 318: As we only are interested in the vm options, I think we can keep the four options as a constant in the ConvertToInstanceBtn.

src/util/helpers.tsx
```diff
@@ -290,3 +290,92 @@ export const getClientOS = (userAgent: string) => {
...
+
+export const sendFileByWebSocket = (
```
edlerd, line 320: Better keep this in a dedicated file with name matching the function?

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,328 @@
...
+  const handleUploadProgress = (current: number, total: number) => {
+    setUploadState({
+      percentage: Math.floor((current / total) * 100),
+      loaded: current,
+      total: total,
+    });
+
+    if (current === total) {
+      handleCloseModal();
+      navigate(`/ui/project/${project?.name}/instances`);
```
edlerd, line 92: I think we should emit an info notification here to inform the upload completed and the import on the server is running in the background.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+        // establish websocket connection based on the instance creation operation
+        let wsUrl = `wss://${location.host}/1.0/operations/${operationId}/websocket`;
+        if (operationSecret) {
+          wsUrl += `?secret=${operationSecret}`;
+        }
+
+        const ws = new WebSocket(wsUrl);
+
+        ws.onopen = () => {
```
edlerd, line 161: lol, I was just commenting not to use use effect for this, but rely on ws.onopen and at that second you pushed this change.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,321 @@
...
+  }, [formik.resetForm, closePortal]);
+
+  return (
+    <>
+      <Button onClick={openPortal} type="button">
+        <span>Convert from image</span>
+      </Button>
+      {isOpen && (
+        <Portal>
+          <Modal
```
edlerd, line 220: Maybe split the modal and the upload logic into a dedicated component?

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,321 @@
...
+            close={handleCloseModal}
+            className="convert-image-modal"
+            title="Convert from image file"
```
edlerd, line 223: I wouldn't announce this feature as "Convert from image file". Imagine what a user would expect this feature under. Probably more a "Upload virtual machine image" or alike. Also, I think the confirm button should say "Upload" or "Upload and create" instead of "Convert".

On upload progress, I would change this to "Uploading file to LXD server" or something more brief and precise.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,321 @@
...
+                  disabled={!formik.values.imageFile}
+                  checked={formik.values.formatConversion}
+                />
+                <Input
+                  {...formik.getFieldProps("virtioConversion")}
+                  type="checkbox"
+                  label="Virtio"
+                  disabled={!formik.values.imageFile}
+                  checked={formik.values.virtioConversion}
+                />
```
edlerd, line 302: I think we need some help text here, as these are not self-explanatory to a user.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,321 @@
...
+            >
+              <Input
+                id="image-file"
+                name="imageFile"
+                type="file"
+                label="Image file"
+                onChange={handleFileChange}
+              />
+              <Input
+                {...formik.getFieldProps("name")}
```
edlerd, line 262: I think we should populate the name from the sanitized file name, wdyt?

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,321 @@
...
+    type: "virtual-machine",
+    name: values.name,
+    sourceDiskSize: fileSize,
+    architecture: values.architecture,
+  };
+
+  return JSON.stringify(instanceArgs);
+};
+
+const ConvertToInstanceBtn: FC = () => {
```
edlerd, line 73: We need to hide this depending on the api extension for conversion to be available.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
```diff
@@ -0,0 +1,321 @@
...
+    if (e.currentTarget.files) {
+      const file = e.currentTarget.files[0];
+      void formik.setFieldValue("imageFile", file);
+    }
+  };
+
+  const handleSuccess = (instanceName: string) => {
+    const message = (
+      <>
+        Created instance <strong>{instanceName}</strong>.
```
edlerd, line 113: I think we should link the instance name. Not sure if we need the configure link.

edlerd (commented):
QA for the happy path looks fine, some comments on the interaction and small ideas for the code.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
mas-who: haha yeah I was going through the code and realised use effect is not needed

src/pages/instances/actions/ConvertToInstanceBtn.tsx
mas-who: Sounds good, I've moved all util functions into `util/convertToInstance.tsx`, moved the form component into `ConvertToInstanceForm.tsx` and kept the upload logic in `ConvertToInstanceBtn.tsx`.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
mas-who: Yeah that sounds good, instead of "Upload virtual machine image", maybe "Upload virtual machine" could be better since the former makes me think an image would be created instead of an instance. wdyt?

src/pages/instances/actions/ConvertToInstanceBtn.tsx
mas-who: I've added some help text for both both options, let me know what you think.

src/pages/instances/actions/ConvertToInstanceBtn.tsx
mas-who: The thinking here is to keep the behaviour consistent with uploading an instance in `UploadInstanceBtn`. The lxd-migrate tool also allows the user to configure the instance during the migration setup steps, which I think would align with this behaviour?

src/pages/instances/actions/ConvertToInstanceBtn.tsx
mas-who: good catch!

mas-who:
@edlerd I've also added a function `isImageTypeRaw` to determine if a image is `raw` format, and based on that we would remove the `format` conversion option. The approach is based on heuristics so it's not foolproof, but I think this should be fine since if we fail to catch a raw image type then worst case it will go through conversion. Let me know what you think?

src/pages/instances/actions/ConvertToInstanceBtn.tsx
edlerd: Yes, agreed to the shorter copy

src/pages/instances/actions/ConvertToInstanceBtn.tsx
edlerd: I see, makes sense.

I think we should always link the instance name. I was conditioned to click on the name and got confused why in this occasion it is not clickable.

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+        <div className="conversion-options">
+          <Input
+            {...formik.getFieldProps("formatConversion")}
+            type="checkbox"
+            label={
+              <>
+                Format{" "}
+                <Icon
+                  name="information"
+                  title="Check this option to convert the image into raw format"
```
edlerd, line 144:
```suggestion
                  title="Convert the image into raw format. This can be skipped if the image is already in raw format to speed up the import."
```
Not sure if the 2nd part is actually true. I'm assuming it is?

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+          />
+          <Input
+            {...formik.getFieldProps("virtioConversion")}
+            type="checkbox"
+            label={
+              <>
+                Virtio{" "}
+                <Icon
+                  name="information"
+                  title="Check this option if the image requires Virtio drivers for conversion"
```
edlerd, line 159:
```suggestion
                  title="Add Virtio drivers to the image. This is needed if..."
```
Again, I am not sure if this is actually true. Can we explain why this should be enabled or disabled?

edlerd: Does this 2nd option depend on the first one to be checked? If so, we should disable it with a title tag stating the dependency in case the 1st option is disabled.

src/pages/instances/forms/ConvertToInstanceForm.tsx
mas-who: Yes that's correct, the backend will simply skip the conversion step for raw images

edlerd: Assuming the above is true, then the label of the checkbox should read "Convert the image into raw format" and the tooltip contain the 2nd sentence.

mas-who: Okay I'm going to vertically stack the checkboxes since the label is quite long :+1:

src/pages/instances/forms/ConvertToInstanceForm.tsx
mas-who: You can have both `format` and `virtio` conversion options. Agreed on the label :+1:

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+  project,
+  uploadState,
+}) => {
+  const handleFileChange = async (e: ChangeEvent<HTMLInputElement>) => {
+    const { onChange } = formik.getFieldProps("imageFile");
+    onChange(e);
+
+    if (e.currentTarget.files) {
+      const file = e.currentTarget.files[0];
+      const suffix = "-instance";
```
edlerd, line 45:
```suggestion
      const suffix = "-import";
```
Maybe better?

src/util/instances.tsx
```diff
@@ -63,3 +63,7 @@ export const truncateInstanceName = (
 
   return name + suffix;
 };
+
+export const sanitizeInstanceName = (name: string): string => {
+  return name.replace(/[^A-Za-z0-9-]/g, "");
```
edlerd, line 68:
```suggestion
  return name.replace(/[^A-Za-z0-9-]/g, "-");
```
Removing spaces can produce hard to read names, maybe better use a dash?

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+      const fileExtension = getFileExtension(file.name);
```
edlerd, line 46: This function doesn't really return the file extension, but an array. Maybe better to bundle this logic together with  sanitize instance name. Instead of splitting the work over two functions.

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+    formik.resetForm();
+    close();
+  }, [formik.resetForm, close]);
+
+  return (
+    <Modal
+      close={handleCloseModal}
+      className="convert-to-instance-modal"
+      title={
+        uploadState ? "Uploading image to LXD" : "Upload virtual machine image"
```
edlerd, line 73:
```suggestion
        uploadState ? "Uploading to LXD" : "Upload virtual machine"
```
As discussed, maybe better copy here without the "image" part.

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+            id: "pool",
+            label: "Root storage pool",
+            disabled: !project || !formik.values.imageFile,
+          }}
+        />
+        <Select
+          {...formik.getFieldProps("architecture")}
+          id="architecture"
+          label="Image architecture"
+          options={supportedVMArchOptions}
```
edlerd, line 131: Can this be an architecture that the host does not support? We have some logic in the base image selector to figure out which architectures are supported by the LXD server. We might want to reuse it here.

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+        />
+        <Input
+          {...formik.getFieldProps("virtioConversion")}
+          type="checkbox"
+          label={
+            <>
+              Add Virtio drivers to the image{" "}
+              <Icon
+                name="information"
+                title="This is needed if the image does not have Virtio drivers installed."
```
edlerd, line 158:
```suggestion
                title="Mandatory, if the image does not have Virtio drivers installed."
```

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+        <label htmlFor="">Conversion options</label>
+        <Input
+          {...formik.getFieldProps("formatConversion")}
+          type="checkbox"
+          label={
+            <>
+              Convert the image into raw format{" "}
+              <Icon
+                name="information"
+                title="This can be skipped if the image is already in raw format to speed up the import."
```
edlerd, line 143:
```suggestion
                title="Can be skipped if the image is already in raw format to speed up the import."
```

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+          options={supportedVMArchOptions}
+          disabled={!formik.values.imageFile}
+        />
+        <label htmlFor="">Conversion options</label>
+        <Input
+          {...formik.getFieldProps("formatConversion")}
+          type="checkbox"
+          label={
+            <>
+              Convert the image into raw format{" "}
```
edlerd, line 140:
```suggestion
              Convert to raw format{" "}
```

src/pages/instances/forms/ConvertToInstanceForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+          }
+          disabled={!formik.values.imageFile}
+          checked={formik.values.formatConversion}
+        />
+        <Input
+          {...formik.getFieldProps("virtioConversion")}
+          type="checkbox"
+          label={
+            <>
+              Add Virtio drivers to the image{" "}
```
edlerd, line 155:
```suggestion
              Add Virtio drivers{" "}
```

edlerd (commented):
This looks pretty neat already.

Some more suggestions on copy. Seeing it in action always sparks more ideas.
Also, some notes on the file name to instance name functionality.

src/pages/instances/forms/ConvertToInstanceForm.tsx
mas-who: `getFileExtension` returns a string? Not sure about lumping the logic with `sanitizeInstanceName` since not all instance names might need a split by file extension

edlerd: What I meant is: from a function called `getFileExtension(input: string)` I'd expect to get a string back, not an array.

mas-who: but `getFileExtension` returns a string, see below code for the function.

```
export const getFileExtension = (filename: string): string => {
  if (!filename.includes(".")) {
    return "";
  }

  return `.${filename.split(".").pop()}` || "";
};
```

mas-who: Aah I see, no problem :+1:

edlerd:
We discussed last week with piper to 
1. Merge the "Upload instance" and "Convert from image buttons". Think of a good description, so users can find both features from that caption
2. Use radio buttons to switch between the two form contents. Place them on the top of the modal. The two different forms are very similar, so switching would only add/remove a couple of fields. Implementation wise, we can keep the two forms in dedicated files, but include them dynamically based on the active radio.
3. We already had good ideas for the labels of the two radio buttons. @piperdeck should have a link to the design file with sketches and the copy. Though those sketches did not have the radio buttons yet. Though, the labels should translate easily into the radio btn idea. Important is that we list the file types -- maybe in a muted font as part of the label.

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -158,95 +136,96 @@ const UploadInstanceBtn: FC = () => {
...
+        })}
+        dependencies={[]}
+        belowIds={["modal-footer"]}
+      >
+        <Form onSubmit={formik.handleSubmit}>
+          <Input
+            id="instance-file"
+            name="instanceFile"
+            type="file"
+            label="Instance backup file"
```
edlerd, line 184: Should this match the radio description, including the example file extensions? Might be better.

src/pages/instances/forms/UploadInstanceFileModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+              inline
+              checked={fileType === "instance-backup"}
+              onChange={() => setFileType("instance-backup")}
+            />
+            <RadioInput
+              label="External format (.qcow2, .vmdk, etc...)"
+              inline
+              checked={fileType === "external-format"}
+              onChange={() => setFileType("external-format")}
+            />
```
edlerd, line 53: The styling looks a little busy, maybe placing them below each other will give this more order. Secondly, I think a bit more bottom spacing would be good for the two radios. See the `snapshot-schedule` CSS class for an example where we set some custom styling to make the radio placement a bit better.

mas-who:
@edlerd this PR is now ready for review. I've implemented the following as discussed:
1. Combine "upload instance" button with the new vm conversion button. In the upload modal, a file type selection is done via radio inputs. The content of the modal will change depending on the file type selected.
2. For external format file selection, if a raw image is selected, then the "format" conversion option will be removed and disabled.
3. For older LXD versions, file type selection will be hidden so users can only upload LXD backup archive files.
4. Added some logic to make the upload modal fixed height for larger screens (if there is enough viewport height). The aim here is to reduce the amount of layout shift in the modal when selecting different file formats. For screens with limited viewport height, the modal height will be dynamic and the form content will be scrollable. Let me know what you think.

edlerd (commented):
This is great, some comments on the layout below. Will give a more thorough review tomorrow.

src/pages/instances/forms/UploadInstanceFileModal.tsx
edlerd: We can also use the `u-sv1` or up to `u-sv3` helper classes, that might be better than introducing more custom css.

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
mas-who: Actually, what do you think about removing the label for this input? Since choosing the file type already relates to the file selection? For older LXD versions, we can show the label "LXD backup archive (.tar.gz)" since the file type selector will be hidden.

edlerd: Yes, I think that works well.

src/pages/instances/forms/UploadInstanceFileModal.tsx
mas-who: thanks I keep forgetting that we have these utility classes!

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+  const invalidateCache = () => {
+    void queryClient.invalidateQueries({
+      predicate: (query) => {
+        return query.queryKey[0] === queryKeys.instances;
+      },
+    });
+  };
+
+  const handleConfirmUpload = (values: UploadExternalFormatFileFormValues) => {
```
edlerd, line 126:
```suggestion
  const handleSubmit = (values: UploadExternalFormatFileFormValues) => {
```
There is no confirm here, is there?

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+        // set up event queue for the operation
+        eventQueue.set(
+          operationId,
+          () => handleSuccess(getInstanceName(operation.metadata)),
+          handleFailure,
+          invalidateCache,
+        );
+      })
+      .catch((e) => {
+        close();
```
edlerd, line 165:
```suggestion
```
I think we should keep this open, so the user could resubmit and won't lose the form status, but we should probably reset the upload state.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+          Upload completed. Now creating instance{" "}
+          <strong>{instanceName}</strong>.
+        </>,
+      );
+      navigate(`/ui/project/${project?.name}/instances`);
+    }
+  };
+
+  const handleUploadError = (error: Error) => {
+    toastNotify.failure("Image upload failed.", error);
```
edlerd, line 87: I think we have to reset the upload state here as well.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+          wsUrl += `?secret=${operationSecret}`;
+        }
+
+        const ws = new WebSocket(wsUrl);
+
+        ws.onopen = () => {
+          sendFileByWebSocket(
+            ws,
+            formik.values.imageFile,
+            FILE_CHUNK_SIZE,
```
edlerd, line 150: Shall we just hard code this in the function? Probably don't need to make this variable.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+        let wsUrl = `wss://${location.host}/1.0/operations/${operationId}/websocket`;
+        if (operationSecret) {
+          wsUrl += `?secret=${operationSecret}`;
+        }
+
+        const ws = new WebSocket(wsUrl);
+
+        ws.onopen = () => {
+          sendFileByWebSocket(
+            ws,
```
edlerd, line 148: Shall we move the creation of the ws into the function and pass in the wsUrl?

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+    onChange(e);
+
+    if (e.currentTarget.files) {
+      const file = e.currentTarget.files[0];
+      const suffix = "-import";
+      const fileExtension = getFileExtension(file.name);
+      const sanitisedFileName = sanitizeInstanceName(
+        file.name.split(fileExtension)[0],
+      );
+      const instanceName = truncateInstanceName(sanitisedFileName, suffix);
```
edlerd, line 202: When I select a file without a file extension, named "foo", I get "f-import", would expect "foo-import" for that case.

I think we don't need the file extension, maybe simplify this to be 1 function, that receives a filename and returns an instance name?

The other upload type uses "imported" suffix and uses a slightly different way to calculate the name. With a generic function, we could reuse it over there.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+        formik.setFieldError("name", undefined);
+      }
+
+      // If the image is already in raw format, remove the format conversion option
+      // this will optimise the conversion process since raw vm images do not need to be converted
+      const isRawImage = await isImageTypeRaw(file);
+      setIsRawImage(isRawImage);
+      if (isRawImage) {
+        await formik.setFieldValue("formatConversion", false);
+      }
```
edlerd, line 219: I don't fully understand why we set this field and then disable the selector for it.

Should we maybe only set the field to false and remove the `isRawImage` state and its side effect to disable the field? Disabling it seems rather drastic to me and might be experienced as not wanted by users.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+  const formik = useFormik<UploadExternalFormatFileFormValues>({
+    initialValues: {
+      name: "",
+      pool: "",
+      imageFile: null,
+      formatConversion: true,
+      virtioConversion: false,
+      architecture: supportedVMArchOptions(
+        settings?.environment?.architectures || [],
+      )[0].value,
```
edlerd, line 179: Can we use the `archOptions` variable, defined below? Probably can move that up a bit in the file and use it here.

src/util/uploadExternalFormatFile.tsx
```diff
@@ -0,0 +1,150 @@
...
+  name: string;
+  pool: string;
+  formatConversion: boolean;
+  virtioConversion: boolean;
+  architecture: string;
+}
+
+export const FILE_CHUNK_SIZE = 1024 * 1024;
+
+export const supportedVMArchOptions = (hostArchitectures: string[]) => {
```
edlerd, line 12:
```suggestion
export const externalUploadArchitectureOptions = (hostArchitectures: string[]) => {
```
Maybe clearer?

src/util/uploadExternalFormatFile.tsx
```diff
@@ -0,0 +1,150 @@
...
+          e.target?.error
+            ? e.target.error
+            : new Error("Failed to read read file"),
+        );
+      }
+    };
+
+    readNextFileChunk();
+  }
+};
```
edlerd, line 122: I found this hard to understand. Maybe splitting the responsibilities into a file reader, that just reports chunks and upload state, and another part that is hooking that up with the socket? Or maybe that is even worse, and the complexity is inherent.

edlerd (commented):
This works nicely and the design and layout now also looks very clean.

I think the error handling needs a bit of refinement. We just close the modal in both upload types, maybe we can keep it open not to lose the state? The user might want to retry, and preserving the state they created might be a better experience.

Apart from that a couple of small suggestions and ideas below.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+              label: "Root storage pool",
+              disabled: !project || !formik.values.imageFile,
+            }}
+          />
+          <Select
+            {...formik.getFieldProps("architecture")}
+            id="architecture"
+            label="Image architecture"
+            options={archOptions}
+            disabled={!formik.values.imageFile || archOptions.length < 2}
```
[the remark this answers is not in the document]
mas-who: also adding the message to all other disabled inputs

src/sass/_upload_instance_modal.scss
```diff
@@ -1,6 +1,19 @@
...
-    width: 35rem;
+  .conversion-options {
+    display: flex;
+    gap: 1rem;
+    margin-bottom: $spv--medium;
+  }
+
+  .fixed-height {
+    height: 33rem;
+  }
```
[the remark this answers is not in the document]
mas-who: `fixed-height` is used in `UploadExternalFormatFileForm` and `UploadInstanceBackupFileForm`.

src/pages/instances/forms/UploadInstanceFileModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+            <div className="u-sv1">
+              <RadioInput
+                label="LXD backup archive (.tar.gz)"
+                checked={fileType === "instance-backup"}
+                onChange={() => setFileType("instance-backup")}
+              />
+            </div>
+            <div className="u-sv3">
+              <RadioInput
+                label="External format (.qcow2, .vmdk, etc...)"
```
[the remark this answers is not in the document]
mas-who: Yup, those are QCOW, QCOW2, VDI, VHDX, or VMDK detailed in the [docs](https://documentation.ubuntu.com/lxd/en/latest/howto/import_machines_to_instances/). Will add to the label, thanks for the `<abbr>` tip!

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
mas-who: very good point! :+1:

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
mas-who: Yeah that makes sense, I guess the tooltip also indicates that the option can be skipped if it's a raw image, will remove the disabling logic.

src/util/uploadExternalFormatFile.tsx
mas-who: hmm not sure about this one, since these values are VM architectures supported by LXD in general and does not only relate to VM image uploads.

src/util/uploadExternalFormatFile.tsx
mas-who: I agree this is not the easiest code to understand but I am also not sure if we can split it like that since the socket and reader logic are tightly coupled. perhaps I can add some comments above the function describing how it works?

src/util/uploadExternalFormatFile.tsx
edlerd: I see, then the existing name seems correct.

src/util/uploadExternalFormatFile.tsx
edlerd: Yeah, a comment might help.

mas-who: I've added some comments explaining the function.

src/util/uploadExternalFormatFile.tsx
```diff
@@ -0,0 +1,156 @@
...
+        return;
+      }
+
+      reader.readAsArrayBuffer(chunk);
+    };
+
+    reader.onload = (event) => {
+      if (!file) {
+        return;
+      }
```
edlerd, line 95: Can this guard be at the top of the function, so we have it only once?

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,334 @@
...
+                    noFileSelectedMessage
+                      ? noFileSelectedMessage
+                      : "Can be skipped if the image is already in raw format to speed up the import."
+                  }
+                />
+              </>
+            }
+            disabled={!!noFileSelectedMessage}
+            checked={formik.values.formatConversion}
+          />
```
edlerd, line 289: There is some clipping when the checkboxes are active. The blue border is cut on the left-hand side.

Screenshot from 2024-09-11 18-06-07
Screenshot from 2024-09-11 18-06-02

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,334 @@
...
+        // establish websocket connection based on the instance creation operation
+        let wsUrl = `wss://${location.host}/1.0/operations/${operationId}/websocket`;
+        if (operationSecret) {
+          wsUrl += `?secret=${operationSecret}`;
+        }
+
+        sendFileByWebSocket(
+          wsUrl,
+          formik.values.imageFile,
+          1024 * 1024,
```
edlerd, line 143:
```suggestion
```
I think there was a misunderstanding. I meant keep the constant, but reference the constant directly in the function. In my opinion, we don't need it in the function parameters.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,334 @@
...
+  );
+
+  const formik = useFormik<UploadExternalFormatFileFormValues>({
+    initialValues: {
+      name: "",
+      pool: "",
+      imageFile: null,
+      formatConversion: true,
+      virtioConversion: false,
+      architecture: archOptions[0].value,
```
edlerd, line 174: I think If the settings are not loaded, the 0 index will cause an error, because we try to read `.value` from undefined.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,334 @@
...
+      if (!formik.errors.name) {
+        formik.setFieldError("name", undefined);
+      }
+
+      // If the image is already in raw format, remove the format conversion option
+      // this will optimise the conversion process since raw vm images do not need to be converted
+      const isRawImage = await isImageTypeRaw(file);
+      if (isRawImage) {
+        await formik.setFieldValue("formatConversion", false);
+      }
```
edlerd, line 208:
```suggestion

      await formik.setFieldValue("formatConversion", !isRawImage);
```
I think we should also enable it in case the image is not raw. Even though that is usually enabled as it is the default.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,334 @@
...
+  };
+
+  const handleCloseModal = useCallback(() => {
+    formik.resetForm();
+    close();
+    notify.clear();
+  }, [formik.resetForm, close, notify]);
+
+  const noFileSelectedMessage = !formik.values.imageFile
+    ? "A file must be selected first."
```
edlerd, line 219:
```suggestion
    ? "Please select a file before adding custom configuration."
```

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,334 @@
...
+                  title={
+                    noFileSelectedMessage
+                      ? noFileSelectedMessage
+                      : "Can be skipped if the image is already in raw format to speed up the import."
+                  }
+                />
+              </>
+            }
+            disabled={!!noFileSelectedMessage}
+            checked={formik.values.formatConversion}
```
edlerd, line 288:
```suggestion
            checked={formik.values.formatConversion}
            title={noFileSelectedMessage}
```
We should have the title on the checkboxes as well.

edlerd (commented):
Thanks for the adjustments!

If found one probably last visual flaw and a couple of very tiny suggestions to improve.

src/util/uploadExternalFormatFile.tsx
mas-who: unfortunately not, the external guard does not apply inside the context of the callbacks

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
mas-who: had to give the checkboxes a small `margin-left`

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
mas-who: aah I see, done

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
mas-who: for some reason when the input is a checkbox, `title` has no effect if specified on the `Input` component directly. Instead I added `title` attribute to the label.

edlerd: I think this is because the checkbox gets wrapped by the label.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
edlerd: This is a visual design crime, please don't misalign them with the rest of the modal. I think the root cause is the scrollable container in `UploadExternalFormatFileForm`. It also creates a half modal scrolling, see screen below. The same is true for the backup file. Maybe we should raise it to the `UploadInstanceFileModal` and not have scrollable container in the form components.

image

mas-who: I see, the issue with that is the modal buttons are inside the form components since they need access to the form submission handlers. Raising the scrollable container to the modal level would defeat the point since the aim is to always show the modal buttons and make form content scrollable. I can create a component for the file type radio selection and move that into the form components to deal with the half scrolling form problem. I think some padding on the form component will probably fix the check boxes without misalignment.

edlerd: Thanks for explaining, I now see the problem with that. Agree, your proposed solution sounds good.

mas-who: I've made the above changes, let me know what you think? :slightly_smiling_face:

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,346 @@
...
+      if (!formik.errors.name) {
+        formik.setFieldError("name", undefined);
+      }
+
+      // If the image is already in raw format, remove the format conversion option
+      // this will optimise the conversion process since raw vm images do not need to be converted
+      const isRawImage = await isImageTypeRaw(file);
+      if (isRawImage) {
+        await formik.setFieldValue("formatConversion", !isRawImage);
+      }
```
edlerd, line 214:
```suggestion
      await formik.setFieldValue("formatConversion", !isRawImage);
```
I think both cases are important, for non-raw images we should also enable conversion. This is the default, but a user might change between raw and non-raw files.

src/sass/_upload_instance_modal.scss
```diff
@@ -1,6 +1,17 @@
 .upload-instance-modal {
-  .p-modal__dialog {
-    margin: auto;
-    width: 35rem;
+  .fixed-height {
+    height: 35rem;
+  }
+
+  .content-details {
+    padding-left: $sph--x-small;
```
edlerd, line 7:
```suggestion
    padding-left: $sph--x-small;
    margin-left: -$sph--x-small;
```
Better keep the content aligned with the title in the header

src/pages/instances/forms/InstanceFileTypeSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface Props {
+  value: InstanceFileType;
+  onChange: (value: InstanceFileType) => void;
+  hide?: boolean;
+}
+
+const InstanceFileTypeSelector: FC<Props> = ({ value, onChange, hide }) => {
+  if (hide) {
+    return null;
+  }
```
edlerd, line 15: You can do it this way, personally prefer this decision in the parent to keep the interface clean. Like `{isConditional && <InstanceFileTypeSelector />}`

src/pages/instances/forms/UploadInstanceFileModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+      )}
+
+      {fileType === "instance-backup" && (
+        <UploadInstanceBackupFileForm
+          close={close}
+          uploadState={uploadState}
+          setUploadState={setUploadState}
+          fixedFormHeight={fixedFormHeight}
+          fileType={fileType}
+          onFileTypeChange={setFileType}
```
edlerd, line 53:
```suggestion
          setFileType={setFileType}
```
Maybe better?

src/pages/instances/forms/UploadInstanceFileModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+      )}
+
+      {fileType === "external-format" && (
+        <UploadExternalFormatFileForm
+          close={close}
+          uploadState={uploadState}
+          setUploadState={setUploadState}
+          fixedFormHeight={fixedFormHeight}
+          fileType={fileType}
+          onFileTypeChange={setFileType}
```
edlerd, line 64:
```suggestion
          setFileType={setFileType}
```
Maybe better?

edlerd (approved):
QA looks good, design as well, few very tiny suggestions below.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
mas-who: of course, that was my bad for not removing the condition check

src/sass/_upload_instance_modal.scss
mas-who: neat!

src/pages/instances/forms/InstanceFileTypeSelector.tsx
mas-who: agreed, that's better :+1:

edlerd (dismissed):
Good work, LGTM code and QA wise.

piperdeck:
This is a nitpick, but is there a reason we're using so much vertical space here? I think the modal might look a little bit better if its minimum vertical size was a bit smaller than this...

CleanShot 2024-09-15 at 21 10 21@2x

I'm noticing that we're not doing a lot to filter or validate the file extension of the file to be uploaded. Was this intentional? Seems like it might be a good idea to check and make sure the file that the user attaches at least has the correct extension, or maybe even set that as a parameter in the file upload dialog?

https://github.com/user-attachments/assets/36418b0d-18ca-4f19-aef6-2c8d719d9586

For example, when uploading pictures to Google Photos, the file picker is invoked in such a way as to disable anything that doesn't have an accepted file extension:

https://github.com/user-attachments/assets/aff5ac7a-b23b-40dc-8d75-9c861403fcfd

mas-who:
> This is a nitpick, but is there a reason we're using so much vertical space here? I think the modal might look a little bit better if its minimum vertical size was a bit smaller than this...
> 
> CleanShot 2024-09-15 at 21 10 21@2x
> 

This is done so that the modal does not change size depending on the file type selected. For the backup archive file type, there are less options to configure for the upload, so it appears to not utilise all the available space of the modal. Do you think we should rather have the modal adjust its size to the available content?

> I'm noticing that we're not doing a lot to filter or validate the file extension of the file to be uploaded. Was this intentional? Seems like it might be a good idea to check and make sure the file that the user attaches at least has the correct extension, or maybe even set that as a parameter in the file upload dialog?
> 
>  CleanShot.2024-09-15.at.21.21.33.mp4 
> For example, when uploading pictures to Google Photos, the file picker is invoked in such a way as to disable anything that doesn't have an accepted file extension:
> 
>  CleanShot.2024-09-15.at.21.25.04.mp4

For the file type extension, we can certainly restrict it during upload. However, the files does not necessarily need an extension for it to be valid. For example, a file named `lxd_vmdk_vm` or `lxd_vmdk_vm.vmdk` can both be a VMDK image. On the extreme side, a user can literally name their file `lxd_vmdk_vm.png` and it could still be a valid VMDK image. In this situation, we rely on the LXD backend to determine if a file uploaded is valid since it directly checks the binary content of the file. We would then show the error returned from the server if the file is invalid. Restricting file selection by extension may result in the UI blocking selection for valid files.

edlerd:
> For the file type extension, we can certainly restrict it during upload. However, the files does not necessarily need an extension for it to be valid. For example, a file named `lxd_vmdk_vm` or `lxd_vmdk_vm.vmdk` can both be a VMDK image. On the extreme side, a user can literally name their file `lxd_vmdk_vm.png` and it could still be a valid VMDK image. In this situation, we rely on the LXD backend to determine if a file uploaded is valid since it directly checks the binary content of the file. We would then show the error returned from the server if the file is invalid. Restricting file selection by extension may result in the UI blocking selection for valid files.

I think there is a way to have the file picker set to a list of allowed file extensions, but the user can still switch the filter to "any file type" if they really want. I think having the filter on the right file type might improve the experience.

edlerd:
I.e. choosing a file from a download directory with dozens or hundreds of files will be much easier with a filter on file extension applied.

mas-who:
> I.e. choosing a file from a download directory with dozens or hundreds of files will be much easier with a filter on file extension applied.

Yeah fair, I think then for instance backup files, we will filter file extensions to `.tar.bz2`, `.tar.gz`, `.tar.xz`, `.tar.lzma`, `.tar`, `.squashfs`, `.qcow2` and `.tar.zst` as per supported compression listed in LXD [source](https://github.com/canonical/lxd/blob/e6a6119184f3d98be0361abae7a5ff9f9389569b/shared/archive.go#L25). For external format file uploads, we will restrict to `.img`, `.qcow`, `.qcow2`, `.vdi`, `.vhdx` and `.vmdk` extensions as listed in the `lxd-migrate` [docs](https://documentation.ubuntu.com/lxd/en/latest/howto/import_machines_to_instances/).

piperdeck:
@mas-who I think I'm liking the fully responsive behaviour of the modal. What are your thoughts on it?

mas-who:
> @mas-who I think I'm liking the fully responsive behaviour of the modal. What are your thoughts on it?

@piperdeck I'm happy with this behaviour if you are happy! :)

mas-who:
@edlerd made the following changes based on feedback from @piperdeck :

1. adjusted the `accept` attribute for instance backup file picker. Turns out the support for this attribute on safari is not great, multi-segment file extensions such as `tar.gz`, `tar.bz2` have issues getting recognised resulting in the file picker to filter out those file extensions on safari. Instead, using MIME types should resolve this issue.

2. If an instance name is set on the create instance page, opening the file upload modal will automatically fill out the instance name input field. We do not generate an instance name based on the file name in this scenario

piperdeck:
Looks good to go!

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,335 @@
...
+      </Form>
+      <footer className="p-modal__footer" id="modal-footer">
+        <Button
+          appearance="base"
+          className="u-no-margin--bottom"
+          type="button"
+          onClick={handleCloseModal}
+        >
+          Cancel
+        </Button>
```
edlerd, line 320: We should also cancel the upload through websocket process on cancel. Currently after "cancel", the upload continues in the background and I think the success callback gets triggered.

edlerd (changes requested):

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
mas-who: @edlerd this is done, works nicely I think, thanks for the suggestion for removing the operation id, didn't even know that function existed

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+  const handleUploadError = (error: Error) => {
+    notify.failure("Image upload failed.", error);
+    setUploadState(null);
+  };
+
+  const handleCancelUpload = () => {
+    if (socket && socket.readyState === WebSocket.OPEN) {
+      socket.close(1000, CANCEL_UPLOAD_REASON);
+      eventQueue.remove(operationId);
```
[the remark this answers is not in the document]
mas-who: yes good catch :+1: no need to check if operationId is defined though since Set.delete will work anyway in eventQueue

edlerd (commented):
QA LGTM
Code also looks good, added a couple of very tiny suggestions to improve on it slightly, should be good to go.

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+    void queryClient.invalidateQueries({
+      predicate: (query) => {
+        return query.queryKey[0] === queryKeys.instances;
+      },
+    });
+  };
+
+  const handleSubmit = (values: UploadExternalFormatFileFormValues) => {
+    if (!values.imageFile) {
+      return;
```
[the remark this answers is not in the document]
mas-who: this is a unlikely condition since form submission is disabled if there is no file selected. I think I previously added this here to get around type checking, probably with some logic before the api call. I see that is no longer needed though, so I'd rather remove this condition than adding a notification.

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+  const handleUpload = () => {
+    if (!formik.values.instanceFile) {
+      return;
+    }
+
+    const uploadController = new AbortController();
+    setUploadAbort(uploadController);
+
+    const instanceName = formik.values.name;
```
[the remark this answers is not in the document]
mas-who: yeah not necessary

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+    void queryClient.invalidateQueries({
+      predicate: (query) => {
+        return query.queryKey[0] === queryKeys.instances;
+      },
+    });
+  };
+
+  const handleUpload = () => {
+    if (!formik.values.instanceFile) {
+      return;
```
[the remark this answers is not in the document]
edlerd: Perfect.

src/util/uploadExternalFormatFile.tsx
```diff
@@ -0,0 +1,165 @@
...
+    { value: "ppc64le", label: "ppc64le" },
+    { value: "s390x", label: "s390x" },
+  ];
+
+  return vmArchitectureOptions.filter((arch) =>
+    hostArchitectures.includes(arch.value),
+  );
+};
+
+export const CANCEL_UPLOAD_REASON = "cancel upload";
```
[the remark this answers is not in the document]
mas-who: no need :+1: I was trying to use this variable for something else previously and forgot to update it

src/util/uploadExternalFormatFile.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  return JSON.stringify(instanceArgs);
+};
+
+// This method facilitates the uploading of a file to a server via a websocket connection.
+// 1. When the websocket connection opens, a FileReader instance is created to read the file in chunks.
+// 2. The readNextFileChunk function slices the file into chunks of the specified chunk size and converts them into binary buffers.
+// 3. If the chunk size is zero, indicating the end of the file, the websocket connection is closed.
+// 4. The FileReader.onload event handler processes indiviual file chunks as they get read, which sends them
+// through the websocket connection.
```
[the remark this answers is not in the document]
mas-who: sure, it's also used to process data in chunks so we don't have to hold the entire file content in memory

edlerd (approved):
LGTM, good work :)

PR #868, opened 2024-08-29
feat: [WD-14383] Image export upload delete tests

Kxiru:
## Done

- Created new e2e tests for Image export and upload. (Implicitly expressed Image deletion as well)
- Created 3 new helper functions, createImage (from instance), delete image (singular) and getImageNameFromAlias.

Fixes
- Amended Publish Instance e2e test to address failing workflow e2e tests.

## QA

1. Run the LXD-UI:
    - Run new e2e tests locally

## Screenshots

N/A


tests/duplicate-instance.spec.ts
edlerd, line 1: I think this file is coming from a previous PR and should be removed?

tests/fixtures/e2e-test-image.tar.gz
edlerd, line 1: Do we need this? I think we can upload the image we exported with the test.

tests/helpers/images.ts
```diff
@@ -20,3 +21,46 @@ export const deleteAllImages = async (
   const notification = page.locator(".toast-notification");
   await expect(notification).toHaveText(/.*deleted.*/);
 };
+
+export const createImage = async (page: Page, instance: string) => {
+  //Creates an image from an instance
```
edlerd, line 26:
```suggestion
export const createImageFromInstance = async (page: Page, instance: string) => {
```
Naming it to what it does lets us remove the comment. I think this function should live in the instance helper file, as it is working on the instance detail page.

tests/helpers/images.ts
```diff
@@ -20,3 +21,46 @@ export const deleteAllImages = async (
...
+  const imageAlias = `${instance}-alias`;
+  await visitInstance(page, instance);
+  await page.getByRole("button", { name: "Create Image" }).click();
+  page.getByRole("dialog", { name: "Create image from instance" });
```
edlerd, line 30: Is something missing from this line - maybe `expect(...).toBeVisible()`?

tests/helpers/images.ts
```diff
@@ -20,3 +21,46 @@ export const deleteAllImages = async (
...
+  await page.waitForSelector(`text=Image ${imageName} deleted.`);
+};
+
+export const getImageNameFromAlias = async (page: Page, imageAlias: string) => {
+  await page.goto(`/ui`);
+  await page.getByRole("link", { name: "Images", exact: true }).first().click();
+
+  await page.getByPlaceholder("Search").click();
+  await page.getByPlaceholder("Search").fill(imageAlias);
+  await page.waitForTimeout(1000);
```
edlerd, line 61:
```suggestion
```
This shouldn't be necessary?

tests/images.spec.ts
```diff
@@ -18,3 +28,60 @@ test("search for custom image and create an instance from it", async ({
 
   await deleteInstance(page, customInstance);
 });
+
+test("Export an image", async ({ page }) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  const imageAlias = await createImage(page, instance);
+
+  await page.goto("/ui/");
```
edlerd, line 37:
```suggestion
```
I think we can click on the images without navigating first.

tests/images.spec.ts
```diff
@@ -18,3 +28,60 @@ test("search for custom image and create an instance from it", async ({
...
+  await page.getByRole("link", { name: "Images", exact: true }).click();
+  await page.getByPlaceholder("Search").click();
+  await page.getByPlaceholder("Search").fill(imageAlias);
+  page.getByText(imageAlias);
```
edlerd, line 41: Is this missing an expect ... to be visible?

tests/images.spec.ts
```diff
@@ -18,3 +28,60 @@ test("search for custom image and create an instance from it", async ({
...
+  const imageName = await getImageNameFromAlias(page, imageAlias);
+  await page.getByLabel("export image").click();
+  await page.waitForSelector(
+    `text=Image ${imageName} was successfully downloaded. Please check your downloads folder.`,
```
edlerd, line 45:
```suggestion
    `text=was successfully downloaded. Please check your downloads folder.`,
```
Maybe we can save the extra complexity to figure out the image name and rely on delete all images instead?

tests/images.spec.ts
```diff
@@ -18,3 +28,60 @@ test("search for custom image and create an instance from it", async ({
...
+
+test("Upload an image", async ({ page }) => {
+  const IMAGE_FILE = "./tests/fixtures/e2e-test-image.tar.gz";
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  await visitInstance(page, instance);
+  const instanceBaseImageName = await page
+    .locator("text=Base image")
+    .locator("xpath=following-sibling::*")
+    .innerText();
```
edlerd, line 60:
```suggestion
```
I think we don't need the instance or the image name for this test?

tests/images.spec.ts
```diff
@@ -18,3 +28,60 @@ test("search for custom image and create an instance from it", async ({
...
+});
+
+test("Delete an image", async ({ page }) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  const imageAlias = await createImage(page, instance);
+  const imageName = await getImageNameFromAlias(page, imageAlias);
+  await deleteImage(page, imageName);
+  await deleteInstance(page, instance);
+});
```
edlerd, line 87:
```suggestion
```
I think deletion is implicitly covered by the "Upload an image" case above.

edlerd (commented):
Good start, some ideas to simplify or streamline the test.

tests/images.spec.ts
edlerd: I see we need to create an instance. Because that will download an image. Because we deleted all images previously with another test. Maybe don't call delete all images in other tests that operate on the default project. Then we don't need an instance for this one.

tests/fixtures/e2e-test-image.tar.gz
Kxiru: I am unsure how to access the "filesystem" using the tests. That being said, this is how uploading the Custom ISO is achieved. Do you have any other suggestions? I can look into filesystem manipulation if that would be better. On my machine, the exported images are downloaded to my downloads folder and the download filename is inaccessible in advance through playwright.

tests/helpers/images.ts
Kxiru: I am not sure that would be much use within the createImageFromInstance function as a successful image is indicated by the toast notification, which we do check for. However it might be useful in getImageNameFromAlias.

tests/images.spec.ts
Kxiru: Won't deleteAllImages give us the error with the my-custom-image, again?

edlerd: Yes, absolutely right, lets keep this as is. When I wrote this comment I didn't know yet about that problem.

tests/images.spec.ts
Kxiru: I think you are right. I have recreated this function so it is much more simplified.

tests/images.spec.ts
Kxiru: Alrighty, I'll delete this function :')

tests/helpers/images.ts
edlerd: I was mainly confused why the line with `  page.getByRole("dialog", { name: "Create image from instance" });` is there after all. Maybe it can be removed?

Kxiru: Oooh, right I see. Yep, that is understandable, that can be removed.

tests/images.spec.ts
```diff
@@ -18,3 +24,43 @@ test("search for custom image and create an instance from it", async ({
 
   await deleteInstance(page, customInstance);
 });
+
+test("Export an image", async ({ page }) => {
+  const instance = randomInstanceName();
+  await createInstance(page, instance);
+  const imageAlias = await createImageFromInstance(page, instance);
```
edlerd, line 31:
```suggestion
  // creating an image from an instance, because we can only export unified images
  // and can't test the export with a standard image fetched from the image server
  // we can remove this step once export of split images is enabled by the backend
  const instance = randomInstanceName();
  await createInstance(page, instance);
  const imageAlias = await createImageFromInstance(page, instance);
```
Maybe explain why this step is needed, so we can remember to clean it up later, once we can export standard images fetched from the image server.

tests/images.spec.ts
```diff
@@ -18,3 +24,43 @@ test("search for custom image and create an instance from it", async ({
...
+    `text=Image ${imageName} was successfully downloaded. Please check your downloads folder.`,
+  );
+
+  await deleteImage(page, imageName);
+  await deleteInstance(page, instance);
+});
+
+test("Upload an image", async ({ page }) => {
+  const IMAGE_FILE = "./tests/fixtures/e2e-test-image.tar.gz";
+  const imageAlias = `${randomInstanceName()}-1`;
```
edlerd, line 50:
```suggestion
  const imageAlias = `${randomImageName()}-1`;
```
I think we should create a function creating us random image aliases in the image helper. We can also use it in `createImageFromInstance`

tests/images.spec.ts
```diff
@@ -18,3 +24,43 @@ test("search for custom image and create an instance from it", async ({
...
+  await page.getByLabel("Image backup file").setInputFiles(IMAGE_FILE);
+  await page.getByPlaceholder("Enter alias").fill(imageAlias);
+  await page
+    .getByLabel("Import image from file")
+    .getByRole("button", { name: "Upload image" })
+    .click();
+  await expect(page.getByText(imageAlias)).toBeVisible();
+  await page.waitForSelector(`text=Image uploaded.`);
+
+  await deleteImage(page, await getImageNameFromAlias(page, imageAlias));
```
edlerd, line 65:
```suggestion
  const imageName = await getImageNameFromAlias(page, imageAlias);
  await deleteImage(page, imageName);
```
Let's make this a bit easier to read.

tests/helpers/instances.ts
```diff
@@ -142,3 +142,15 @@ export const visitAndStopInstance = async (page: Page, instance: string) => {
     await page.waitForSelector(`text=Instance ${instance} stopped.`);
   }
 };
+
+export const createImageFromInstance = async (page: Page, instance: string) => {
+  const imageAlias = `${instance}-alias`;
+  await visitInstance(page, instance);
+  await page.getByRole("button", { name: "Create Image" }).click();
+  await page.getByLabel("Alias").click();
```
edlerd, line 150:
```suggestion
```
I think this is not needed.

tests/instances.spec.ts
```diff
@@ -238,3 +239,23 @@ test("Duplicate an instance", async ({ page }) => {
...
+  await page
+    .getByRole("dialog", { name: "Create image from instance" })
+    .getByRole("button", { name: "Create image" })
+    .click();
+
+  await page.waitForSelector(`text=Image created from instance ${instance}.`);
+
+  //Images created from an instance will have the same Image name and therefore deleteImage should be called twice.
+  await deleteImage(page, instanceBaseImageName);
+  await deleteImage(page, instanceBaseImageName);
```
edlerd, line 260: I think we shouldn't delete the base image here, only the image we created from the instance. We should give that image an alias on creation, so we can identify and delete only that one.

edlerd (changes requested):
Thanks for improving this, I think some more changes would make this good to merge.

tests/fixtures/e2e-test-image.tar.gz
edlerd: The playwright docs explain [how to access download files](https://playwright.dev/docs/downloads). I think we shouldn't save it as in the example in the docs. But once we have the download object from the on-download event, we can use its path to upload again.

tests/instances.spec.ts
Kxiru: To clarify, this section deletes the image created upon instance creation, as well as the image created by clicking through Create Image. The instanceBaseImageName gets the name of the first image (which I have called the base image) from the instances page.

EDIT: Giving the image an alias on creation is a great idea and it has remidned me that I have actually already created this function through createImageFromInstance. As such, I have refactored this test to include this function, as well as get an image alias back and delete both images via both alias and image name.

EDIT 2: So we shouldn't delete the base image, here? No cleanup?

Kxiru: EDIT 3: Perhaps we shouldn't delete the base image because due to the structure of the test suite, both an instance and vm instance are created before each test and for thorough cleanup, one would have to delete the image from both. Which is not impossible, but is out of scope.
I would be happy to add this cleanup functionality if you see it fit though, it would only be a few more lines of code in the .afterAll test.

tests/fixtures/e2e-test-image.tar.gz
Kxiru: A note with this, is that upon click of the button, the browser downloads a file to the default directory, AND playwright's download function effectively allows us to make a COPY of this file to another area, with a dedicated filename. So whilst we are achieving our objective of having a file in the place we want, (I have put it in tests/fixtures and it gets overwritten every time the test runs), the file is technically downloading twice. I wonder whether this will raise concerns for storage and random files. I will push this change for you to review.

tests/instances.spec.ts
edlerd: I think we should keep the base images and never remove them. This will speed up the tests. Cleaning them up means the next test needs to download the base image again. I think it is fine to keep the base image around.

If we create custom images from instances or by upload with alias, we should clean them up.

tests/instances.spec.ts
```diff
@@ -238,3 +240,11 @@ test("Duplicate an instance", async ({ page }) => {
   await page.waitForSelector(`text=Created instance ${instance}-duplicate.`);
   await deleteInstance(page, `${instance}-duplicate`);
 });
+
+test("Create an image from an instance", async ({ page }) => {
+  await page.goto("/ui/");
+  const imageAlias = await createImageFromInstance(page, instance);
+  await visitInstance(page, instance);
```
edlerd, line 247:
```suggestion
```
Don't think we need this navigation step?

tests/fixtures/image.tar.gz
edlerd, line 1: We should remove this file from the commit.

tests/images.spec.ts
```diff
@@ -18,3 +24,47 @@ test("search for custom image and create an instance from it", async ({
...
+
+  await expect(page.getByText(imageAlias)).toBeVisible();
+  let imageName = await getImageNameFromAlias(page, imageAlias);
+  await page.getByLabel("export image").click();
+  const download = await downloadPromise;
+  await page.waitForSelector(
+    `text=Image ${imageName} was successfully downloaded. Please check your downloads folder.`,
+  );
+  const IMAGE_FILE = "tests/fixtures/image.tar.gz";
+  await download.saveAs(IMAGE_FILE);
```
edlerd, line 49:
```suggestion
```
I think we can get the location of the download directly, without manually saving it here (see other comment)

tests/images.spec.ts
```diff
@@ -18,3 +24,47 @@ test("search for custom image and create an instance from it", async ({
...
+  await deleteImage(page, imageName);
+
+  //Upload an image
+  await page.goto("/ui/");
+  await page.getByRole("link", { name: "Images", exact: true }).click();
+
+  await page.getByRole("button", { name: "Upload image" }).click();
+  await page.getByLabel("Image backup file").setInputFiles(IMAGE_FILE);
```
edlerd, line 57:
```suggestion
  await page.getByLabel("Image backup file").setInputFiles(download.path());
```
I think we can directly use the download path, without saving it manually.

edlerd (commented):
Some small ideas to simplify.

tests/images.spec.ts
Kxiru: So apparently the download path cannot be used because it only works when playwright is used locally, not remotely, as it is being done when we run tests. I read this online somewhere but download.path is not an option for us and to circumvent it, we should use a static path.

edlerd (dismissed):
Code LGTM, just one issue found by the linter to use const instead of let.

Kxiru:
> Code LGTM, just one issue found by the linter to use const instead of let.

Yep, I just addressed that. Please could I get that approval again? :)

edlerd (approved):

PR #869, opened 2024-08-29
fix: remove bulk instance migration feature due to server loading concerns

mas-who:
## Done

- From the discussion with lxd team, there are concerns about potentially overloading the server if bulk migration is performed on a large number of instances. For now, we will remove the feature and wait for the team to add background task scheduling capability so we can do bulk actions in the UI without issues.

## QA

1. Run the LXD-UI:
    - For the UI connected to a lxd cluster, check that the `Migrate` button is no longer visible when selecting multiple instances on the instance list page.


Kxiru (commented):
Big oof. Looks good to me though. After all, it's not affecting anything else that is there. I'll allow David to give a final review.

edlerd (commented):
I am not a fan. If we remove it, we should also clean up the `isBulkMigration` from the `MigrateInstanceModal` and `migrateInstanceBulk` from the `api/instances.tsx` file.

mas-who:
> I am not a fan. If we remove it, we should also clean up the `isBulkMigration` from the `MigrateInstanceModal` and `migrateInstanceBulk` from the `api/instances.tsx` file.

Yeah I'm also not a huge fan of this change, but I think overloading the server is a valid point. If you don't want to merge this one, maybe when I get back we can look into processing batches of requests instead. In the interim, I've done some more cleaning up, removed all bulk migration related logic and also adjusted the modal component to process a single instance only.

PR #871, opened 2024-08-30
feat: [WD-14390] Add bulk instance action tests.

Kxiru:
## Done

- Added a test, "Bulk start, pause, unpause and stop instances".

## QA

1. Run the LXD-UI:

## Screenshots

N/A

### DIFF
--- a/tests/instances.spec.ts
+++ b/tests/instances.spec.ts
@@ -247,3 +247,64 @@ test("Create an image from an instance", async ({ page }) => {
   const imageName = await getImageNameFromAlias(page, imageAlias);
   await deleteImage(page, imageName);
 });
+
+test("Bulk start, pause, unpause and stop instances", async ({ page }) => {
+  await page.goto("/ui/");
+  await page.getByPlaceholder("Search").click();
+  await page.getByPlaceholder("Search and filter").fill(instance);
+  await page.getByPlaceholder("Search and filter").press("Enter");
+  await page.getByPlaceholder("Add filter").press("Escape");
+
+  //Bulk start instances
+  await page
+    .getByRole("row", { name: "select Name Type Description Status Actions" })
+    .getByLabel("multiselect rows")
+    .click();
+  await page
+    .getByRole("button", { name: "Select all instances on this" })
+    .click();
+  await page
+    .locator("button")
+    .filter({ hasText: /^Start$/ })
+    .click();
+  await page
+    .getByLabel("Confirm start")
+    .getByRole("button", { name: "Start" })
+    .click();
+  await page.waitForSelector(`text=instance started.`);
+
+  //Bulk restart instances
+  await page.locator("button").filter({ hasText: "Restart" }).click();
+  await page
+    .getByLabel("Confirm restart")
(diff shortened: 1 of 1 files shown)


tests/instances.spec.ts
```diff
@@ -247,3 +247,53 @@ test("Create an image from an instance", async ({ page }) => {
...
+  await page
+    .getByRole("row", { name: "select Name Type Description Status Actions" })
+    .getByLabel("Select all")
+    .click();
+  await page.getByRole("button", { name: "Start", exact: true }).click();
+  await page
+    .getByLabel("Confirm start")
+    .getByRole("button", { name: "Start" })
+    .click();
+  await page.waitForSelector(`text=instances started.`);
```
edlerd, line 266:
```suggestion
  await page.waitForSelector(`text=instance started.`);
```
Now it is a singular case, as we filter the other vm instance. Same for the other tests below.

edlerd: Maybe we should search for the container instance from the test suite instead of filtering by container type. So this also passes locally with a busy default project containing many containers.

Kxiru: I am not sure I understand. Are you saying we should put the two tests in its own suite and do the filter then, what would this achieve?
The tests pass locally for me when adjacent with other non related instances I have.

Kxiru: Or perhaps are you saying that the test should be conducted in a project of it's own so that it is created, empty?

Kxiru: Right, I believe I understand now. I'll get that done.

tests/instances.spec.ts
```diff
@@ -247,3 +247,54 @@ test("Create an image from an instance", async ({ page }) => {
   const imageName = await getImageNameFromAlias(page, imageAlias);
   await deleteImage(page, imageName);
 });
+
+test("Bulk start, pause, unpause and stop instances", async ({ page }) => {
+  await page.goto("/ui/");
+  await page.getByPlaceholder("Search").click();
+  await page.getByPlaceholder("Search and filter").fill(instance);
+  await page.getByText("Search for").click();
```
edlerd, line 255: This doesn't work for Firefox and lets the test suite fail.
I searched around our existing tests and found this bit confirming the search for volumes, I assume it would work on both chrome and firefox:
```
  await page.getByPlaceholder("Search and filter").fill(volume);
  await page.getByPlaceholder("Search and filter").press("Enter");
  await page.getByPlaceholder("Add filter").press("Escape");
```

edlerd (changes requested):

tests/instances.spec.ts
Kxiru: Thank you very much, David. I am not sure if this addressed the problem that I found when running Playwright against firefox, locally however I later found out that for some reason, Firefox and Chromium do not have the same locators for the checkbox. However, this was circumvented when I selected the dropdown option instead and bulk selected the instances on the page, that way. Please let me know your thoughts on the new implementation.

edlerd (approved):
LGTM

PR #872, opened 2024-08-30
Add export of instances WD-13701

edlerd:
## Done

- Add export of instances

Fixes WD-13701

## QA

1. Run the LXD-UI:
    - export an instance from the instance detail page

### DIFF
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -393,3 +393,26 @@ export const uploadInstance = (
       .catch(reject);
   });
 };
+
+export const createInstanceBackup = (
+  instanceName: string,
+  project: string,
+  expiresAt: string,
+  backupName: string,
+): Promise<LxdOperationResponse> => {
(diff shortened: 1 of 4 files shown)


Kxiru (commented):
Heya!
Thanks for this implementation. Exporting the image works well for me. Just one question though, for instances that are running, I find that the Export image button is not disabled. Should one be able to create a backup of an instance even when it is running?

edlerd:
> Heya! Thanks for this implementation. Exporting the image works well for me. Just one question though, for instances that are running, I find that the Export image button is not disabled. Should one be able to create a backup of an instance even when it is running?

Yes, we can export a running instance. That works in the`lxc export` cli, as well as in the api we are using here.

Kxiru (approved):
LGTM, thank you for the extra feature and for clarifying whether to export during a running instance :+1: .

PR #878, opened 2024-09-03
Enable VM tests and reduce flaky tests

edlerd:
## Done

- use DISABLE_VM_TESTS variable consistently in tests
- add test artefacts to `.gitignore` file
- show disabled override button in edit forms instead of hiding it, when the section is disabled, and we are in read mode
- fix instance tests to not set a config value that prevents the instance from being started.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,5 +81,7 @@ dist/
 /playwright/.cache/
 /coverage
 tests/screenshots
+tests/fixtures/image.tar.gz
+tests/fixtures/instance.tar.gz
 
 haproxy-local.cfg
--- a/src/components/ConfigurationRow.tsx
+++ b/src/components/ConfigurationRow.tsx
@@ -103,19 +103,20 @@ export const getConfigurationRow = ({
             })}
           </div>,
         )}
-        {!disabled && (
+        {
           <div>
             <Button
               onClick={toggleDefault}
               type="button"
               appearance="base"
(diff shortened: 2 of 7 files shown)


src/components/ConfigurationRow.tsx
```diff
@@ -144,24 +145,23 @@ export const getConfigurationRow = ({
...
+              if (!isOverridden) {
+                enableOverride();
+              }
+              focusOverride();
+            }}
+            className="u-no-margin--bottom"
+            type="button"
+            appearance="base"
+            title={isOverridden ? "Edit" : "Create override"}
+            disabled={disabled}
```
Kxiru, line 160: Since the disabled button is now shown, is it worth having a title or explanation on hover detailing why the button is disabled? Same for other instance of this.

.gitignore
```diff
@@ -81,5 +81,7 @@ dist/
 /playwright/.cache/
 /coverage
 tests/screenshots
+tests/fixtures/image.tar.gz
```
Kxiru, line 84: I think this is a great idea.

src/components/ConfigurationRow.tsx
edlerd: Ah yes, good idea. Will add the reason as a title.

Kxiru (approved):
LGTM!

PR #879, opened 2024-09-04
Show permissions in navigation also without oidc.audience being set.

edlerd:
## Done

- Show permissions in navigation also without oidc.audience being set. It is optional, after all


Kxiru (approved):
LGMT, fair enough.

PR #883, opened 2024-09-05
feat: [WD-14512] Proxy device configuration

Kxiru:
## Done

- Adding Proxy Configuration...

**Draft PR for Proxy device config**
Current issues / todos:
[] Correct implementation of the tiered Listen / Connect sections given the restrictions of the current components getInheritedDeviceRow and getConfigurationRowBase.
[] Correct / best case implementation of the UI design.

Upcoming:
[] Reviewing create/edit instance/profile pipeline with new api params.
[] Explanation tooltips for disabled buttons
[] NAT boolean logic (If false, connect and listen types must be the same)

## QA

1. Run the LXD-UI:
    - Navigate to the Create Instance page to see Proxy device config upon creating a new instance, or also access from the Create Profile / Edit Instance / Edit Profile pages.

## Screenshots

Pending...


src/components/forms/SelectProxyBtn.tsx
```diff
@@ -0,0 +1,19 @@
+import { FC } from "react";
+import { Button, Icon } from "@canonical/react-components";
+import { ProxyCard } from "types/resources";
+
+interface Props {
+  onSelect: (image: ProxyCard) => void;
+}
+
+const SelectProxyBtn: FC<Props> = ({ onSelect }) => {
```
edlerd, line 9:
```suggestion
const NewProxyBtn: FC<Props> = ({ onSelect }) => {
```
Closer to the functionality?

src/util/configInheritance.tsx
```diff
@@ -313,6 +315,12 @@ interface InheritedGPU {
   source: string;
 }
 
+interface InheritedProxy {
+  key: string;
+  proxy: LxdProxyDevice;
+  source: string;
+}
+
```
edlerd, line 323: Better put this interface directly next to the new function that is using it.

src/types/resources.d.ts
```diff
@@ -67,6 +67,22 @@ export interface GpuCard {
...
+    mode?: number;
+    address?: string;
+    port?: string;
+  };
+  connect?: {
+    type?: string;
+    address?: string;
+    port?: string;
+  };
+}
```
edlerd, line 85: I think we don't need this type. To translate the listen/connect strings into the nested object, there is a better strategy. As discussed, similar to the memory selection, we can have a component that receives the composed listen/connect string and breaks it down on render. On change, that component can rebuild the string and only communicate the recomposed string back up to the form state.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,303 @@
...
+    notify.failure("Loading profiles failed", profileError);
+  }
+
+  const inheritedProxies = getInheritedProxies(formik.values, profiles);
+
+  const existingDeviceNames: string[] = [];
+  existingDeviceNames.push(...inheritedProxies.map((item) => item.key));
+  existingDeviceNames.push(...formik.values.devices.map((item) => item.name));
+
+  const addProxyCard = (card: ProxyCard) => {
```
edlerd, line 62:
```suggestion
  const addProxy = () => {
```
I think we don't need an argument here, as the added proxy is static, we can hardcode the values below in this function

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,303 @@
...
+    const copy = [...formik.values.devices];
+    copy.push({
+      type: "proxy",
+      bind: card.bind,
+      nat: card.nat,
+      proxy_protocol: card.proxy_protocol,
+      mode: card.listen?.mode,
+      listen: `${card.listen?.type}:${card.listen?.address}:${card.listen?.port}`,
+      connect: `${card.listen?.type}:${card.listen?.address}:${card.listen?.port}`,
```
edlerd, line 71:
```suggestion
      type: "proxy",
```
I think we can keep the optional fields blank on creation

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+  if (profileError) {
+    notify.failure("Loading profiles failed", profileError);
+  }
+
+  const inheritedProxies = getInheritedProxies(formik.values, profiles);
+
+  const existingDeviceNames: string[] = [];
+  existingDeviceNames.push(...inheritedProxies.map((item) => item.key));
+  existingDeviceNames.push(...formik.values.devices.map((item) => item.name));
```
edlerd, line 60: I just realize now this approach is not 100% correct, because all device types share the same namespace. So we could make the `getExistingDeviceNames(formik)` function generic for all device pages. But we can add a follow-up task to do this. Probably best not to mix this change into this pr.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,303 @@
...
+    );
+
+    Object.keys(item.proxy).forEach((key) => {
+      if (key === "name" || key === "type") {
+        return null;
+      }
+
+      inheritedRows.push(
+        getInheritedDeviceRow({
+          label: deviceKeyToLabel(key),
```
edlerd, line 143: Here we might encounter any field in the proxy config (see docs for a [full list](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_proxy/#device-options)). So we should mention all possible keys in the `deviceKeyToLabel` function, also `uid` and alike, even though we don't support them otherwise in the ui.
For testing you could add them via the YAML editor to a proxy device.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,303 @@
...
+            title="Detach Proxy"
+          >
+            <Icon name="disconnect" />
+            <span>Detach</span>
+          </Button>
+        ),
+      }),
+    );
+
+    Object.keys(device).forEach((key) => {
```
edlerd, line 193: I think we should be more opinionated and write out the individual fields here. For several reasons, as 1. we want a specific ordering, and the Object.keys might give use a different one after saving and loading again. 2. We only want to represent some fields, not all supported ones. 3. To get the translation to dropdowns or adding help text for some fields, it will be easier or more straight forward, if the fields are listed explicitly.

edlerd (commented):
This looks pretty good, some comments below.

src/types/resources.d.ts
Kxiru: If we get rid of this type, this breaks things in NewProxyBtn (Previously SelectProxyBtn) and ProxyDeviceForm...

src/components/forms/ProxyDeviceForm.tsx
Kxiru: Sorry, are you saying one should create the "getExistingDeviceNames" function? How would this work?

edlerd: We can add that in another PR, I'll log a ticket for it. Just realized we should do that when reading this PR.

src/components/forms/ProxyDeviceForm.tsx
Kxiru: For some reason, if I do not put the fields here, then I cannot render the labels, but I can also no longer put the fields here anyway because upon implementing the suggestion to get rid of the card, I can no longer reference the card attributes.

This ProxyCard that is created in the AddProxy function -to my understanding - contains the attributes that will be shown on the proxydevice form. This blank slate if you will, is then pushed onto the devices stack, and later re-referenced within Object.keys at line 193 and getConfigurationRowBase around line 201 and iterated through to identify how to render each component. I don't think we can just get rid of the proxy card.

src/components/forms/ProxyDeviceForm.tsx
Kxiru: Is that not the current behaviour?

src/components/forms/NewProxyBtn.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+const NewProxyBtn: FC<Props> = ({ onSelect }) => {
+  return (
+    <>
+      <Button onClick={onSelect} type="button" hasIcon>
+        <Icon name="plus" />
+        <span>New Proxy Device</span>
+      </Button>
+    </>
```
edlerd, line 15:
```suggestion
      <Button onClick={onSelect} type="button" hasIcon>
        <Icon name="plus" />
        <span>New Proxy Device</span>
      </Button>
```

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+  const getProxyDeviceFormRows = (
+    label: string,
+    fieldName: string,
+    index: number,
+    options: {
+      label: string;
+      value: string;
+    }[],
+    help_text?: string,
```
edlerd, line 84:
```suggestion
    help?: string,
```
Let's keep it aligned with select component, also better to use camlCase than snake_case.

src/components/forms/ProxyDeviceForm.tsx
edlerd: You are right, uid and gid is already in the deviceKeyToLabel function, i think the security.uid and security.gid are missing and there might be others?

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+        { label: "Enabled", value: "true" },
+        { label: "Disabled", value: "false" },
+      ]),
+    );
+    customRows.push(
+      getProxyDeviceFormRows("Bind", "bind", index, [
+        { label: "Host", value: "host" },
+        { label: "Instance", value: "instance" },
+      ]),
+    );
```
edlerd, line 231: I think this one should be first and with a help text according to the latest design.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+              type="text"
+              className="u-no-margin--bottom"
+              placeholder="0644"
+              help={
+                <ConfigFieldDescription
+                  description={"Mode for the listening Unix socket"}
+                />
+              }
+              label="Mode"
+            />
```
edlerd, line 267:
```suggestion
```
I think we removed mode from the latest design.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+              type="text"
+              help={
+                <ConfigFieldDescription
+                  description={
+                    "Use the following format to specify the address and port: <type>:<addr>:<port>[-<port>][,<port>]"
+                  }
+                />
+              }
+              className="u-no-margin--bottom"
+              label="Value"
```
edlerd, line 290:
```suggestion
```
We probably don't need a label on this one?

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+              id={`devices.${index}.listen`}
+              key={`devices.${index}.listen`}
+              onBlur={formik.handleBlur}
+              onChange={(e) => {
+                void formik.setFieldValue(
+                  `devices.${index}.listen`,
+                  e.target.value,
+                );
+              }}
+              // value={formik.values.devices[index][fieldName]}
```
edlerd, line 279:
```suggestion
value={device.listen}
```
This should work and also satisfy the TS linter.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+        { label: "Host", value: "host" },
+        { label: "Instance", value: "instance" },
+      ]),
+    );
+    customRows.push(
+      getProxyDeviceFormRows("Use HAProxy Protocol", "proxy_protocol", index, [
+        { label: "Yes", value: "true" },
+        { label: "No", value: "false" },
+      ]),
+    );
```
edlerd, line 237:
```suggestion
      getProxyDeviceFormRows("Use HAProxy Protocol", "proxy_protocol", index, optionYesNo, device.proxy_protocol),
```
We have most of the yes/no, enabled/disabled, etc options defined as constants to reuse.

We can add another parameter to the `getProxyDeviceFormRows` to pass in the current value.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+      label: string;
+      value: string;
+    }[],
+    help_text?: string,
+  ) => {
+    const key = `devices.${index}.${fieldName}`;
+
+    return getConfigurationRowBase({
+      className: "no-border-top inherited-with-form",
+      configuration: <Label>{label}</Label>,
```
edlerd, line 90:
```suggestion
      configuration: <Label forId={key}>{label}</Label>,
```
We should connect this, so clicking on the label enables the input and for a11y to ensure all inputs have a label attached.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+  const existingDeviceNames: string[] = [];
+  existingDeviceNames.push(...inheritedProxies.map((item) => item.key));
+  existingDeviceNames.push(...formik.values.devices.map((item) => item.name));
+
+  const addProxy = () => {
+    const copy = [...formik.values.devices];
+    copy.push({
+      type: "proxy",
+      name: deduplicateName("proxy", 1, existingDeviceNames),
```
edlerd, line 66:
```suggestion
      name: deduplicateName("proxy", 1, existingDeviceNames),
      bind: "host"
```
Either we add a default for the bind field, or we need to add a 3rd option to the dropdown with a "Select value" label, disabled true and an empty value.

edlerd (commented):
Good progress, some more suggestions below.

src/components/forms/ProxyDeviceForm.tsx
Kxiru: I can add those. They should be in the ProxyDevice type/interface but I was not sure whether to add them.

src/components/forms/ProxyDeviceForm.tsx
Kxiru: Why do we need to pass through the current value?

src/components/forms/ProxyDeviceForm.tsx
Kxiru: I added the Select Value option to the dropdown, however the automatically selected option is still Host. I have also added the Select value to the other dropdowns. Please verify whether the behaviour is as you expect, because I would assume we want those Select Value options to show as the defaults and that is not what is currently happeing.

src/components/forms/ProxyDeviceForm.tsx
edlerd: We don't have to, but I think it is easiest to do it this way, because otherwise we have to pick the device at index again and cast it to the right typescript type. Passing the value seems a bit simpler.

src/components/forms/ProxyDeviceForm.tsx
edlerd: What we do in other places is give the "select an option" option a value of empty string, and make the value of the input empty string in case it is undefined.

src/components/forms/ProxyDeviceForm.tsx
Kxiru: Okay, but what attribute within the Select component should accept the value? I am not sure what is supposed to be done with it.

edlerd: On the select it can be `value={value ?? ""}`

Kxiru: Ah I see, thank you.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,364 @@
...
+    label: string,
+    fieldName: string,
+    index: number,
+    options: {
+      label: string;
+      value: string;
+      disabled?: boolean;
+    }[],
+    help?: string,
+    value?: string,
```
edlerd, line 88:
```suggestion
    value?: string,
    help?: string,
```
Best to switch them around, so we don't have to pass undefined below when no help text is defined.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,364 @@
...
+      ),
+    );
+
+    customRows.push(
+      getProxyDeviceFormRows(
+        "NAT mode",
+        "nat",
+        index,
+        optionEnabledDisabled,
+        undefined,
```
edlerd, line 247:
```suggestion
```
I think we can remove this once it is the last argument

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -0,0 +1,364 @@
...
+      ),
+    );
+
+    customRows.push(
+      getProxyDeviceFormRows(
+        "Use HAProxy Protocol",
+        "proxy_protocol",
+        index,
+        optionYesNo,
+        undefined,
```
edlerd, line 258:
```suggestion
```

piperdeck:
Hi Nkeiru! Are you planning on using this structure from the design file?

CleanShot 2024-09-10 at 09 38 43

edlerd (approved):
LGTM

PR #886, opened 2024-09-09
feat: [WD-14771]- Get existing device names

Kxiru:
## Done

- Implemented getExistingDeviceNames function in devices.tsx

## QA

1. Run the LXD-UI:
    - N/A

## Screenshots

N/A

### DIFF
--- a/src/components/forms/DiskDeviceFormCustom.tsx
+++ b/src/components/forms/DiskDeviceFormCustom.tsx
@@ -1,8 +1,7 @@
-import React, { FC } from "react";
+import { FC } from "react";
 import { Button, Icon, Input, Label } from "@canonical/react-components";
 import { InstanceAndProfileFormikProps } from "./instanceAndProfileFormValues";
 import { EditInstanceFormValues } from "pages/instances/EditInstance";
-import { InheritedVolume } from "util/configInheritance";
 import CustomVolumeSelectBtn from "pages/storage/CustomVolumeSelectBtn";
 import {
   deduplicateName,
@@ -18,18 +17,16 @@ import classnames from "classnames";
 import { LxdStorageVolume } from "types/storage";
 import { isDiskDeviceMountPointMissing } from "util/instanceValidation";
 import { ensureEditMode } from "util/instanceEdit";
+import { getExistingDeviceNames } from "util/devices";
+import { LxdProfile } from "types/profile";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
   project: string;
-  inheritedVolumes: InheritedVolume[];
+  profiles: LxdProfile[];
 }
 
-const DiskDeviceFormCustom: FC<Props> = ({
-  formik,
-  project,
-  inheritedVolumes,
-}) => {
+const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
   const readOnly = (formik.values as EditInstanceFormValues).readOnly;
   const customVolumes = formik.values.devices
     .filter((device) => device.name !== "root" && device.type === "disk")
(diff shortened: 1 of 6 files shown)


edlerd (commented):
The function seems all right, can you also adjust the three usages, where we currently manually craft the `const existingDeviceNames: string[] = [];` to rely on the new `getExistingDeviceNames` function instead of having the custom code?

I think in `DiskDeviceFormCustom` we have to change the interface a bit. We can remove the `inheritedVolumes` as a prop, but we need to add the `profiles` as a new one.

Kxiru:
> The function seems all right, can you also adjust the three usages, where we currently manually craft the `const existingDeviceNames: string[] = [];` to rely on the new `getExistingDeviceNames` function instead of having the custom code?
> 
> I think in `DiskDeviceFormCustom` we have to change the interface a bit. We can remove the `inheritedVolumes` as a prop, but we need to add the `profiles` as a new one.

Please let me know your thoughts on the implementation :).

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -38,10 +42,18 @@ const DiskDeviceFormCustom: FC<Props> = ({
...
+  const notify = useNotify();
+
+  const { data: profiles = [], error: profileError } = useQuery({
+    queryKey: [queryKeys.profiles],
+    queryFn: () => fetchProfiles(project),
+  });
+
+  if (profileError) {
+    notify.failure("Loading profiles failed", profileError);
+  }
```
edlerd, line 54: We can get the profiles from the parent, that seems the simpler approach than loading them again here.

src/util/devices.tsx
```diff
@@ -62,3 +65,26 @@ export const deviceKeyToLabel = (input: string): string => {
...
+  existingDeviceNames.push(...values.devices.map((item) => item.name));
+  // Returns devices within the profile
+  if (values.entityType === "instance") {
+    const appliedProfiles = getAppliedProfiles(values, profiles);
+    for (const profile of appliedProfiles) {
+      Object.values(profile.devices).forEach((device) => {
+        if ("name" in device) {
+          existingDeviceNames.push((device as { name: string }).name);
+        }
+      });
```
edlerd, line 85: The device.name is actually not the right thing to use here, we need the keys of the devices. This might make this block simpler.
You can test by applying a profile with any device i.e. a GPU in it. Attach the profile to an instance. Then add another GPU directly to that instance, the name chosen should be gpu-2, because gpu-1 is taken by a device inherited from the profile. This currently doesn't work, the inherited and the directly added GPU have the same name gpu-1

edlerd (changes requested):

src/util/devices.tsx
```diff
@@ -62,3 +65,24 @@ export const deviceKeyToLabel = (input: string): string => {
...
+
+export const getExistingDeviceNames = (
+  values: InstanceAndProfileFormValues,
+  profiles: LxdProfile[],
+): string[] => {
+  const existingDeviceNames: string[] = [];
+
+  // Returns devices within the instance
+  existingDeviceNames.push(...values.devices.map((item) => item.name));
+  // Returns devices within the profile
```
edlerd, line 77:
```suggestion
  // Returns devices from any of the profiles assigned to the instance
```

edlerd (dismissed):
QA and code LGTM, one tiny nitpick on the comment, but should be good to merge as is too.

src/util/devices.tsx
Kxiru: Thank you! I have addressed this now, I'll just need your approval again :).

edlerd (approved):

PR #888, opened 2024-09-11
feat: [WD-14770] Device inheritance simplification

Kxiru:
## Done

- Created new functions processIngeritedDevices and getLxdDefault

## QA

1. Run the LXD-UI:

## Screenshots

### DIFF
--- a/src/util/configInheritance.tsx
+++ b/src/util/configInheritance.tsx
@@ -10,6 +10,7 @@ import {
   isProxyDevice,
 } from "util/devices";
 import {
+  LxdDeviceValue,
   LxdDiskDevice,
   LxdGPUDevice,
   LxdNicDevice,
@@ -106,6 +107,12 @@ const getInstanceRowMetadata = (
   return getInstanceProfileProjectDefaults(values, configKey, configField);
 };
 
+const getLxdDefault = (configField: ConfigField | undefined): string => {
+  return configField?.default && configField?.default.length > 0
+    ? configField?.default
+    : "-";
+};
+
 const getProjectRowMetadata = (
   values: ProjectFormValues,
   name: string,
@@ -141,10 +148,7 @@ const getStorageVolumeRowMetadata = (
   const configKey = getVolumeKey(name);
   const configField = configFields.find((item) => item.key === configKey);
 
-  const lxdDefault =
-    configField?.default && configField?.default.length > 0
-      ? configField?.default
-      : "-";
+  const lxdDefault = getLxdDefault(configField);
 
   return { value: lxdDefault, source: "LXD", configField };
 };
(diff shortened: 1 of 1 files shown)


src/util/configInheritance.tsx
```diff
@@ -249,6 +244,31 @@ export const getInheritedRootStorage = (
...
+      .filter(([key, device]) => filterFn(device) && key !== "root")
+      .map(([key, dev]) => {
+        inheritedDevices.push({
+          key: key,
+          [device]: dev as T,
+          source: `${profile.name} profile`,
+        });
+      });
+  }
+};
```
edlerd, line 270: I think the problem is that in this generic function, we can't ensure the returned values are of a specific device type. I see you tried to solve that with a generic type and passing in a filter function. I think that could work, if the filter function is also using the same type `T`as the return value.

But maybe a simpler approach is to have the generic function collect all devices. In the specific functions, we can focus on filtering and mapping to the correct device type. With this, we don't need to pass the filter around, but apply it where it gets defined.

```suggestion

interface InheritedDevice {
  key: string;
  device: LxdDeviceValue;
  source: string;
}

const getInheritedDevices = (
  values: InstanceAndProfileFormValues,
  profiles: LxdProfile[],
): InheritedDevice[] => {
  const result: InheritedDevice[] = [];
  if (values.entityType === "instance") {
    const appliedProfiles = getAppliedProfiles(values, profiles);
    for (const profile of appliedProfiles) {
      Object.entries(profile.devices).map(([key, device]) => {
        result.push({ key, device, source: `${profile.name} profile` });
      });
    }
  }
  return result;
};

export interface InheritedVolume {
  key: string;
  disk: LxdDiskDevice;
  source: string;
}

export const getInheritedVolumes = (
  values: InstanceAndProfileFormValues,
  profiles: LxdProfile[],
): InheritedVolume[] => {
  return getInheritedDevices(values, profiles)
    .filter((item) => isDiskDevice(item.device) && item.key !== "root")
    .map((item) => ({
      ...item,
      disk: item.device as LxdDiskDevice,
    }));
};
```

Kxiru: This is a great solution. I love how you used the spread operator to transform the item object array and also add another property. That's a very cool technique that I will try to keep in mind. Thanks for the implementation :).

src/util/configInheritance.tsx
```diff
@@ -261,23 +258,12 @@ export const getInheritedVolumes = (
...
-            disk: disk as LxdDiskDevice,
-            source: `${profile.name} profile`,
-          });
-        });
-    }
-  }
-
-  return inheritedVolumes;
+  return getInheritedDevices(values, profiles)
+    .filter((item) => isDiskDevice(item.device))
```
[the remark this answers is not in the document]
Kxiru: Fair enough, thanks David! :)

edlerd (dismissed):
LGTM, just one linter error left, I think the type cast in line 280 is not needed, because isDiskDevice ensures the type already. Though below in line 284 we probably still need it.

edlerd (approved):

PR #890, opened 2024-09-12
feat: [WD-14882] Adjust Listen and Connect Inputs

Kxiru:
## Done

- Split Listen and Connect values into 3 different inputs.
- Disabled Connect Type is NAT is enabled.
- Tested for inherited devices.

## QA

1. Run the LXD-UI:
    - Review the config page for adding Proxy devices to an instance or profile.

## Screenshots

Updated:
image

### DIFF
--- a/src/components/forms/ProxyDeviceForm.tsx
+++ b/src/components/forms/ProxyDeviceForm.tsx
@@ -31,7 +31,8 @@ import { deviceKeyToLabel } from "util/devices";
 import { ensureEditMode } from "util/instanceEdit";
 import NewProxyBtn from "components/forms/NewProxyBtn";
 import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
-import { optionEnabledDisabled, optionYesNo } from "util/instanceOptions";
+import { optionEnabledDisabled } from "util/instanceOptions";
+import { getProxyAddress } from "util/proxyDevices";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
@@ -84,6 +85,8 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
     }[],
     value?: string,
     help?: string,
+    onChange?: (value: string) => void,
+    disabledText?: string,
   ) => {
     const key = `devices.${index}.${fieldName}`;
 
@@ -99,11 +102,14 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
           onChange={(e) => {
             ensureEditMode(formik);
             void formik.setFieldValue(key, e.target.value);
+            onChange?.(e.target.value);
           }}
           value={value ?? ""}
           options={options}
           help={<ConfigFieldDescription description={help} />}
           className="u-no-margin--bottom"
+          disabled={disabledText != undefined}
+          title={disabledText}
         />
       ),
       override: "",
@@ -187,6 +193,16 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
     }
     const device = formik.values.devices[index] as LxdProxyDevice;
 
+    const deviceListenParts = device.listen?.split(":") || [];
+    const listenType =
+      deviceListenParts.length > 0 ? deviceListenParts[0] : "tcp";
+
+    const deviceConnectParts = device.connect?.split(":") || [];
+    const connectAddress =
+      deviceConnectParts.length > 1 ? deviceConnectParts[1] : "";
+    const connectPort =
+      deviceConnectParts.length > 2 ? deviceConnectParts[2] : "";
+
     customRows.push(
       getConfigurationRowBase({
         className: "no-border-top custom-device-name",
@@ -233,6 +249,11 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
         ],
         device.bind,
         "Whether to bind the listen address to the instance or host",
+        (value) => {
+          if (value === "instance" && device.nat === "true") {
+            void formik.setFieldValue(`devices.${index}.nat`, "false");
+          }
+        },
       ),
     );
 
@@ -243,84 +264,24 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
         index,
         optionEnabledDisabled,
         device.nat,
+        undefined,
+        (value) => {
+          if (value === "true") {
+            void formik.setFieldValue(
+              `devices.${index}.connect`,
+              `${listenType}:${connectAddress}:${connectPort}`,
+            );
+          }
+        },
+        device.bind === "instance"
+          ? "Only host-bound proxies can use NAT"
+          : undefined,
       ),
     );
 
-    customRows.push(
-      getProxyDeviceFormRows(
-        "Use HAProxy Protocol",
-        "proxy_protocol",
-        index,
-        optionYesNo,
-        device.proxy_protocol,
-      ),
-    );
+    getProxyAddress(customRows, device, index, "listen", formik, "Listen");
 
-    customRows.push(
-      getConfigurationRowBase({
-        className: "no-border-top inherited-with-form",
-        configuration: <Label>Listen</Label>,
-        inherited: (
-          <Input
-            name={`devices.${index}.listen`}
-            id={`devices.${index}.listen`}
-            key={`devices.${index}.listen`}
-            onBlur={formik.handleBlur}
-            onChange={(e) => {
-              ensureEditMode(formik);
-              void formik.setFieldValue(
-                `devices.${index}.listen`,
-                e.target.value,
-              );
-            }}
-            value={device.listen}
-            type="text"
-            help={
-              <ConfigFieldDescription
-                description={
-                  "Use the following format to specify the address and port: <type>:<addr>:<port>[-<port>][,<port>]"
-                }
-              />
-            }
-            className="u-no-margin--bottom"
-          />
-        ),
-        override: "",
-      }),
-    );
-
-    customRows.push(
-      getConfigurationRowBase({
-        className: "no-border-top inherited-with-form",
-        configuration: <Label>Connect</Label>,
-        inherited: (
-          <Input
-            name={`devices.${index}.connect`}
-            id={`devices.${index}.connect`}
-            key={`devices.${index}.connect`}
-            onBlur={formik.handleBlur}
-            onChange={(e) => {
-              ensureEditMode(formik);
-              void formik.setFieldValue(
-                `devices.${index}.connect`,
-                e.target.value,
-              );
-            }}
-            value={device.connect}
-            type="text"
-            help={
-              <ConfigFieldDescription
-                description={
-                  "Use the following format to specify the address and port: <type>:<addr>:<port>[-<port>][,<port>]"
-                }
-              />
-            }
-            className="u-no-margin--bottom"
-          />
-        ),
-        override: "",
-      }),
-    );
+    getProxyAddress(customRows, device, index, "connect", formik, "Connect");
   });
 
   if (isProfileLoading) {
--- a/src/util/proxyDevices.tsx
+++ b/src/util/proxyDevices.tsx
@@ -1,0 +1,160 @@
+import { Input, Label, Select } from "@canonical/react-components";
+import { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import { getConfigurationRowBase } from "components/ConfigurationRow";
+import { LxdProxyDevice } from "types/device";
+import { ensureEditMode } from "./instanceEdit";
+import { proxyAddressTypeOptions } from "./instanceOptions";
+import { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+
+type ConnectionType = "listen" | "connect";
+
+export const getProxyAddress = (
+  customRows: MainTableRow[],
+  device: LxdProxyDevice,
+  index: number,
+  connectionType: ConnectionType,
+  formik: InstanceAndProfileFormikProps,
+  headingTitle: string,
+) => {
+  const deviceParts = device[connectionType]?.split(":") || [];
+  const deviceType = deviceParts.length > 0 ? deviceParts[0] : "tcp";
+  const deviceAddress = deviceParts.length > 1 ? deviceParts[1] : "";
+  const devicePort = deviceParts.length > 2 ? deviceParts[2] : "";
+
+  customRows.push(
+    getConfigurationRowBase({
+      className: "no-border-top inherited-with-form p-heading--6",
+      configuration: headingTitle,
+      inherited: "",
+      override: "",
+    }),
+  );
+
+  customRows.push(
+    getConfigurationRowBase({
+      className: "no-border-top inherited-with-form",
+      configuration: (
+        <Label forId={`devices.${index}.${connectionType}`}>Type</Label>
+      ),
+      inherited: (
+        <Select
+          id={`devices.${index}.${connectionType}`}
+          onChange={(e) => {
+            ensureEditMode(formik);
+            const newType = e.target.value;
+            void formik.setFieldValue(
+              `devices.${index}.${connectionType}`,
+              `${newType}:${deviceAddress}:${devicePort}`,
+            );
+
+            if (device.nat === "true") {
+              const connectParts = device.connect?.split(":") || [];
+              const connectAddress =
+                connectParts.length > 1 ? connectParts[1] : "";
+              const connectPort =
+                connectParts.length > 2 ? connectParts[2] : "";
+
+              void formik.setFieldValue(
+                `devices.${index}.connect`,
+                `${newType}:${connectAddress}:${connectPort}`,
+              );
+            }
+          }}
+          value={deviceType}
+          options={proxyAddressTypeOptions}
+          className="u-no-margin--bottom"
+          disabled={connectionType === "connect" && device.nat === "true"}
+          title={
+            device.nat
+              ? "This is determined by the listen type when nat mode is enabled"
+              : undefined
+          }
+        />
+      ),
+      override: "",
+    }),
+  );
+
+  customRows.push(
+    getConfigurationRowBase({
+      className: "no-border-top inherited-with-form",
+      configuration:
+        deviceType === "unix" ? (
+          <Label forId={`devices.${index}.${connectionType}.unixsocket`}>
+            *Socket path
+          </Label>
+        ) : (
+          <Label forId={`devices.${index}.${connectionType}.address`}>
+            *Address
+          </Label>
+        ),
+      inherited:
+        deviceType === "unix" ? (
+          <Input
+            id={`devices.${index}.${connectionType}.unixsocket`}
+            onChange={(e) => {
+              ensureEditMode(formik);
+              const socketPath = e.target.value;
+              void formik.setFieldValue(
+                `devices.${index}.${connectionType}`,
+                `unix:${socketPath}`,
+              );
+            }}
+            value={deviceAddress}
+            placeholder="/<socket_path>"
+            type="text"
+            className="u-no-margin--bottom"
+          />
+        ) : (
+          <Input
+            id={`devices.${index}.${connectionType}.address`}
+            onChange={(e) => {
+              ensureEditMode(formik);
+              const newAddress = e.target.value;
+              void formik.setFieldValue(
+                `devices.${index}.${connectionType}`,
+                `${deviceType}:${newAddress}:${devicePort}`,
+              );
+            }}
+            value={deviceAddress}
+            placeholder="127.0.0.1"
+            type="text"
+            className="u-no-margin--bottom"
+          />
+        ),
+      override: "",
+    }),
+  );
+
+  deviceType === "unix"
+    ? null
+    : customRows.push(
+        getConfigurationRowBase({
+          className: "no-border-top inherited-with-form",
+          configuration: (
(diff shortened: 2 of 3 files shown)


Kxiru:
It would be a nice to have to be able to have hover text on the disabled input for connect when NAT is enabled, however it does not take an onhover param. Any tips here, @edlerd ?

Also, @piperdeck , thanks in advance for the design review. The current code structure doesn't quite allow for the nicer look of having the input labels to the left of the input field. Even with the stacked attribute on the inputs, they end up looking indented and out of place. I tried exploring a few different ways to achieve your design but what not quite able. Perhaps some further light might be able to be shed on this by a reviewer...

src/components/forms/ProxyDeviceInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+  isNatEnabled,
+}) => {
+  const [inheritedType, inheritedAddress, inheritedPort] =
+    value?.split(":") || [];
+
+  const [type, setType] = useState(inheritedType || "");
+  const [address, setAddress] = useState(inheritedAddress || "");
+  const [port, setPort] = useState(inheritedPort || "");
+
+  const AddressTypeOptions = [
```
edlerd, line 24:
```suggestion
  const proxyAddressTypeOptions = [
```
I think this can move into the instance options file.

src/components/forms/ProxyDeviceInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+  value,
+  isConnect,
+  isNatEnabled,
+}) => {
+  const [inheritedType, inheritedAddress, inheritedPort] =
+    value?.split(":") || [];
+
+  const [type, setType] = useState(inheritedType || "");
+  const [address, setAddress] = useState(inheritedAddress || "");
+  const [port, setPort] = useState(inheritedPort || "");
```
edlerd, line 22:
```suggestion
  const parts = value?.split(":") || [];
  const type = parts.length > 0 ? parts[0] : "tcp";
  const address = parts.length > 1 ? parts[1] : "";
  const port = parts.length > 2 ? parts[2] : "";
```
I think we can do this without having local state in here. We can directly rely on the formik state as it comes in.

src/components/forms/ProxyDeviceInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+  ];
+
+  return (
+    <>
+      <Select
+        label="Type"
+        onChange={(e) => {
+          const newType = e.target.value;
+          setType(newType);
+          onChange?.(newType, address, port);
```
edlerd, line 51:
```suggestion
          onChange(newType, address, port);
```
We can directly update the parent state from here. Should work the same for the other fields below.

src/components/forms/ProxyDeviceInput.tsx
```diff
@@ -0,0 +1,85 @@
+import { FC, useState } from "react";
+import { Input, Select } from "@canonical/react-components";
+
+interface Props {
+  onChange?: (type?: string, address?: string, port?: string) => void;
+  value?: string;
+  isConnect?: boolean;
+  isNatEnabled?: string;
```
edlerd, line 8:
```suggestion
  isTypeDisabled,
```
nat is not a local concept to this component, I'd name this towards the local result of the variable.

src/components/forms/ProxyDeviceInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+        label="Type"
+        onChange={(e) => {
+          const newType = e.target.value;
+          setType(newType);
+          onChange?.(newType, address, port);
+        }}
+        value={type}
+        options={AddressTypeOptions}
+        className="u-no-margin--bottom"
+        disabled={isConnect && isNatEnabled === "true"}
```
edlerd, line 56:
```suggestion
        disabled={isTypeDisabled}
        title={isTypeDisabled ? "This is determined by the listen type when nat mode is enabled" : undefined}
```
Here we can set the hover text with a title.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -261,28 +262,15 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
...
-            key={`devices.${index}.listen`}
-            onBlur={formik.handleBlur}
-            onChange={(e) => {
+          <ProxyDeviceInput
+            onChange={(type, address, port) => {
               ensureEditMode(formik);
               void formik.setFieldValue(
                 `devices.${index}.listen`,
-                e.target.value,
+                `${type}:${address}:${port}`,
```
edlerd, line 270: If nat mode is enabled, we should also update the connect type from here. I think we could do as below 
```
const connectParts = formDevice.connect.split(":");
const connectAddress = connectParts.length > 1 ? connectParts[1] : "";
const connectPort = connectParts.length > 2 ? connectParts[2] : "";
void formik.setFieldValue(
                `devices.${index}.connect`,
                e.target.value,
                `${type}:connectAddress:connectPort`,
```

edlerd (commented):
Some ideas on the code simplification.
We need to be closer to the design. The labels should be in the left-hand column and listen/connect are not labels, and in italics.

piperdeck:
I'm getting some strange behaviour where my mouse pointer is showing the pointing finger when I hover over the field label. This is different from when I hover over any other field label that isn't on this page:

https://github.com/user-attachments/assets/4c6f2e5f-bf18-4ece-bb30-84fd18f0edf2

Also noticing that the address fields don't adapt to when Unix is selected as the connection type. Because unix sockets aren't like IP addresses, and they're of the form /path/to/socket.sock, the fields should change when I change the port type:

CleanShot 2024-09-15 at 20 58 33@2x

I'm not sure if this is what David is talking about in his previous comment, but when the user set NAT mode to true, the connection types for _listen_ and _connect_ should always be the same. Also, neither one can be Unix if NAT mode is enabled. But that behaviour isn't showing up at the moment:

CleanShot 2024-09-15 at 21 03 41@2x

These fields should probably be filled by their default values, rather than by "Select option":

CleanShot 2024-09-15 at 21 05 53@2x

I'm also noticing that the mode and UID/GID fields are missing when _Unix_ is selected for the listen address. Did we opt not to include those?

Kxiru:
@piperdeck , thanks for your review. A few comments :)

- I have since been able to replicate your design! Apologies for the delay and thanks for creating it.
- Enabling NAT now disables the type field in Connect and the Connect Type reflects the Listen type when it is changed.
- Currently working on automatically matching the Connect Type to the Listen Type if/when they are indifferent before NAT is enabled.

1. I was not aware of the Unix layout changing being a task or issue. I'll ask for more information on this.
2. Regarding the mouse pointer, I find that the mous pointer becomes a pointer finger when hovering over the labels of any config in the GPU, Proxy, Other and Network tabs. I believe it is thus inkeeping with the existing style. I can look into this further if you would like/need?
3. Regarding the default values, I was advised that Selecting an option would be a good way to move forward to avoid ... Missing some of the inputs? So users have to manually select each of their options.
4. Are Mode and GID/UID inputs specific to Unix types? @edlerd , please could you advise on how to proceed here, as I believe we were removing the mode section but I'm not sure about the GID/UID.

edlerd:
> * Regarding the mouse pointer, I find that the mous pointer becomes a pointer finger when hovering over the labels of any config in the GPU, Proxy, Other and Network tabs. I believe it is thus inkeeping with the existing style. I can look into this further if you would like/need?

I think this is default for HTML labels and should be all right from my point of view.

> * Regarding the default values, I was advised that Selecting an option would be a good way to move forward to avoid ... Missing some of the inputs? So users have to manually select each of their options.

I asked to remove the defaults, because otherwise there is no way for a user to create a proxy device without those values being set. Now I am not so sure. Maybe we should preselect the default values that LXD will enforce?

Another thought: Should we remove the HAProxy protocol selector? I am not so sure how relevant it is, have a suspicion it is kind of a special use case thing we might want to hide from the average UI user.

> * Are Mode and GID/UID inputs specific to Unix types? @edlerd , please could you advise on how to proceed here, as I believe we were removing the mode section but I'm not sure about the GID/UID.

I think Mode, GID and UID are for very special use cases, maybe we can ignore them in the UI to keep the form simple?

Kxiru:
> > * Regarding the mouse pointer, I find that the mous pointer becomes a pointer finger when hovering over the labels of any config in the GPU, Proxy, Other and Network tabs. I believe it is thus inkeeping with the existing style. I can look into this further if you would like/need?
> 
> I think this is default for HTML labels and should be all right from my point of view.
> 
> > * Regarding the default values, I was advised that Selecting an option would be a good way to move forward to avoid ... Missing some of the inputs? So users have to manually select each of their options.
> 
> I asked to remove the defaults earlier, because otherwise there is no way for a user to create a proxy device without those values being set. Now I am not so sure. Maybe we should preselect the default values that LXD will enforce?
> 
> Another thought: Should we remove the HAProxy protocol selector? I am not so sure how relevant it is, have a suspicion it is kind of a special use case thing we might want to hide from the average UI user.
> 
> > * Are Mode and GID/UID inputs specific to Unix types? @edlerd , please could you advise on how to proceed here, as I believe we were removing the mode section but I'm not sure about the GID/UID.
> 
> I think Mode, GID and UID are for very special use cases, maybe we can ignore them in the UI to keep the form simple?

Noted. @piperdeck , how far do you agree with this?
I am happy to move forward in either case.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -258,30 +277,58 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
...
+                `devices.${index}.listen`,
+                `${newType}:${listenAddress}:${listenPort}`,
+              );
+
+              if (device.nat === "true") {
+                void formik.setFieldValue(
+                  `devices.${index}.connect`,
+                  `${newType}:${connectAddress}:${connectPort}`,
+                );
+              }
```
edlerd, line 304: We also need to set the connect type to one matching in listen. Otherwise, we have this problem:

1. user selects connect type tcp
2. user select listen type udp
3. user enables nat mode

Now the list type is wrong, but the user can't change it to match.

edlerd (commented):
One observation from QA

src/components/forms/ProxyDeviceForm.tsx
Kxiru: Yep, that's what I'm working on now. :)

Kxiru: Alrighty, I have addressed (and pushed) this now. Moving onto the other issues.

piperdeck:
>> Regarding the default values, I was advised that Selecting an option would be a good way to move forward to avoid ... Missing some of the inputs? So users have to manually select each of their options.
> 
> I asked to remove the defaults earlier, because otherwise there is no way for a user to create a proxy device without those values being set. Now I am not so sure. Maybe we should preselect the default values that LXD will enforce?

My concern is that I want the user to actually know what the value of the default is. Like in the _Bind_ field, with it set to _Choose option_, there's no intuitive way to know whether this configuration is going to bind to the host or the instance (for example). This is a tricky question. Am I understanding correctly that on the back end, there's a difference between implicitly keeping the default value and explicitly setting the default value? 

In that case, is there any reason for use to ever leave that field implicit, rather than declaring it explicitly? I would lean toward that solution personally.

>> Regarding the mouse pointer, I find that the mous pointer becomes a pointer finger when hovering over the labels of any config in the GPU, Proxy, Other and Network tabs. I believe it is thus inkeeping with the existing style. I can look into this further if you would like/need?
>
> I think this is default for HTML labels and should be all right from my point of view.

I'm ok with the behaviour in theory, it's just that it's not consistent with field labels on our other pages. It's setting off alarm bells for me that there's something different in the implementation that might need to be brought back in line. But even if all the code is ok, I think that this kind of consistency is important for maintaining a general feeling of quality

Regarding HAproxy and GUI/ID options, if they're really advanced options, I'm ok with reserving them for the YAML editor since we have it. We might have actually discussed this before, apologies if it's just my bad for forgetting :P

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -292,35 +327,178 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
...
+              }}
+              value={connectPort}
+              placeholder="00[-00]"
+              type="text"
+              className="u-no-margin--bottom"
+            />
+          ),
+        override: "",
+      }),
+    );
```
edlerd, line 501: This is repeating very similar code as above.
Maybe put this into a helper file like `getProxyAddress`, which is parameterized for listen and connect, so we can call it twice and don't have to repeat all the fields and the logic around unix sockets, etc. twice.

Kxiru: Working on this... :)

src/util/proxyDevices.tsx
```diff
@@ -0,0 +1,131 @@
...
+            onChange={(e) => {
+              ensureEditMode(formik);
+              const newAddress = e.target.value;
+              void formik.setFieldValue(
+                `devices.${index}.${connectionType}`,
+                `${deviceType}:${newAddress}:${devicePort}`,
+              );
+            }}
+            value={deviceAddress}
+            placeholder="000.000.000.000"
```
edlerd, line 95:
```suggestion
            placeholder="127.0.0.1"
```
Maybe better, the 000.000.000.000 does not look like an ip to me.

src/util/proxyDevices.tsx
```diff
@@ -0,0 +1,131 @@
...
+        ),
+      inherited:
+        deviceType === "unix" ? (
+          <Input
+            onChange={(e) => {
+              ensureEditMode(formik);
+              const newAddress = e.target.value;
+              void formik.setFieldValue(
+                `devices.${index}.${connectionType}`,
+                `${newAddress}`,
```
edlerd, line 76: We are missing to prepend with type: here, no?

src/util/proxyDevices.tsx
```diff
@@ -0,0 +1,131 @@
...
+        ),
+      override: "",
+    }),
+  );
+
+  customRows.push(
+    getConfigurationRowBase({
+      className: "no-border-top inherited-with-form",
+      configuration:
+        deviceType === "unix" ? null : (
```
edlerd, line 108: I think we should rather wrap the whole customRows.push block with a conditional, instead of having the conditional twice inside this structure. Or we could return early above this block.

edlerd (commented):
This looks very close to finished now, thanks for that. Some comments below.

src/util/proxyDevices.tsx
Kxiru: I tried to do this but for some reason it does not register.

src/util/proxyDevices.tsx
Kxiru: Prepend?

edlerd: Yes. This is for the unix type, right? If so we could just hardcode it as `unix:${newAddress}`

edlerd: Maybe also rename newAddress to socketPath

Kxiru: Oh right. Yes that is true. I definitely missed needing to do that.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -99,6 +101,10 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
           onChange={(e) => {
             ensureEditMode(formik);
             void formik.setFieldValue(key, e.target.value);
+
+            if (e.target.value === "true") {
```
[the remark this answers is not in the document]
Kxiru: How should this be achieved? An event listener / use effect? I am wondering where the logic would actually go.

src/util/proxyDevices.tsx
```diff
@@ -0,0 +1,146 @@
...
+            onChange={(e) => {
+              ensureEditMode(formik);
+              const socketPath = e.target.value;
+              void formik.setFieldValue(
+                `devices.${index}.${connectionType}`,
+                `unix:${socketPath}`,
+              );
+            }}
+            value={deviceAddress}
+            placeholder="/"
```
edlerd, line 94:
```suggestion
            placeholder="/<socket_path_on_instance>"
```
Probably better?

src/components/forms/ProxyDeviceForm.tsx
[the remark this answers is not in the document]
edlerd: Here we call `onChange?.(e.target.value)` and in the caller we do
```
(value) => {
if (value) {
          void formik.setFieldValue(
            `devices.${index}.connect`,
            `${listenType}:${connectAddress}:${connectPort}`,
          );
}
        },
```

src/util/proxyDevices.tsx
```diff
@@ -0,0 +1,146 @@
...
+            const newType = e.target.value;
+            void formik.setFieldValue(
+              `devices.${index}.${connectionType}`,
+              `${newType}:${deviceAddress}:${devicePort}`,
+            );
+
+            if (device.nat === "true") {
+              void formik.setFieldValue(
+                `devices.${index}.connect`,
+                `${newType}:${deviceAddress}:${devicePort}`,
```
edlerd, line 53: We should not override the deviceAddress and devicePort of the connect. with this we are taking all the values from listen and copy them to connect, but we should only copy the type.

edlerd (commented):
Thanks for the adjustments, I found one probably last issue with changing the listen type when nat mode is enabled.

src/util/proxyDevices.tsx
edlerd: So we need to read the `device.index.connect` value and decompose it to get the address and port to use it here inside the conditional block.

src/components/forms/ProxyDeviceForm.tsx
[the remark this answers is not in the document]
Kxiru: Ahh, got it :)

src/util/proxyDevices.tsx
```diff
@@ -0,0 +1,152 @@
...
+            onChange={(e) => {
+              ensureEditMode(formik);
+              const socketPath = e.target.value;
+              void formik.setFieldValue(
+                `devices.${index}.${connectionType}`,
+                `unix:${socketPath}`,
+              );
+            }}
+            value={deviceAddress}
+            placeholder="/<socket_path_on_instance>"
```
edlerd, line 100:
```suggestion
            placeholder="/<socket_path>"
```
Sorry for going back and forth here. I just realized in listen the socket path on instance is wrong, and it should be socket path on host. The easiest solution is to make the copy generic.

edlerd (commented):
image

I think when switching to bind mode "instance", the NAT mode selector should be disabled with a title explaining "Only host-bound proxies can use NAT". If NAT mode is enabled, we should disable it. If NAT mode was not set, it shouldn't be overridden.

image
This is another restriction the form should represent. Though, it is much harder to represent. Maybe removing the HAProxy protocol selector is the easiest solution for it. As any other solution I can think of will be hard to understand for the user. And the field seems not so important to me.

src/util/proxyDevices.tsx
```diff
@@ -0,0 +1,152 @@
...
+  );
+
+  customRows.push(
+    getConfigurationRowBase({
+      className: "no-border-top inherited-with-form",
+      configuration:
+        deviceType === "unix" ? (
+          <Label forId={`devices.${index}.unixsocket`}>*Socket path</Label>
+        ) : (
+          <Label forId={`devices.${index}.address`}>*Address</Label>
```
edlerd, line 85: We need the connectionType in the id, otherwise the labels in connect relate to the fields in the listen section.

Kxiru:
@edlerd I'll remove the HAProxy input for now (and add this to the Spec), but isn't the solution just as simple as checking that NAT is disabled? Or is the problem moreso telling the User at the right time that certain configs aren't allowed?

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -40,6 +41,7 @@ interface Props {
 
 const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
   const notify = useNotify();
+  const [isInstanceBound, setInstanceBound] = useState(false);
```
edlerd, line 44: We already store this info in formik, we should avoid shadowing the state here. Rely on the respective formik value instead.

src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -233,6 +252,13 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
         ],
         device.bind,
         "Whether to bind the listen address to the instance or host",
+        (value) => {
```
edlerd, line 255: We need to ensure nat mode is not enabled. We can unset it here, in case it is enabled. If it is disabled, we should not update it.

edlerd (changes requested):

edlerd:
> isn't the solution just as simple as checking that NAT is disabled? Or is the problem moreso telling the User at the right time that certain configs aren't allowed?

There are many special cases here and handling them well is a can of worms. Probably best to avoid this complexity. Some examples:
1. user already selected listen mode unix and entered a socket. Then decides to switch on nat mode and then enables haproxy
2. user selected nat mode and haproxy protocol, now we need to only allow selecting type tcp, no udp or unix socket. This will be very confusing.
3. etc.

edlerd (dismissed):
This works so well now, good stuff :+1: 

QA and code LGTM, some very tiny nitpicks on the code, can also merge as is.

edlerd (approved):

PR #891, opened 2024-09-13
Update path-to-regexp

edlerd:
## Done

- Update path-to-regexp to avoid CVE-2024-45296

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,6 @@
     "src/**/*.scss": "stylelint"
   },
   "resolutions": {
-    "rollup": "4.20.0"
+    "path-to-regexp": "8.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -113,6 +113,6 @@
     "src/**/*.scss": "stylelint"
   },
   "resolutions": {
-    "rollup": "4.20.0"
```
mas-who, line 116: wouldn't this cause build issues on s390x again? Or was the issue resolved in rollup 4.13.x?

edlerd: I think it is not needed anymore, as the lockfile still resolves to 4.20.0

mas-who: miss that, thanks for clarifying :+1:

mas-who (approved):

mas-who:
LGTM

PR #893, opened 2024-09-13
Show change counts on save buttons for edit forms

edlerd:
## Done

- Show change counts on save buttons for edit forms
- added a function to count the pending changes for any edit form
- align naming of create/edit storage pool components. they have been renamed.
- change instance/profile configuration > cloud init forms. to improve the design in them. 
- fix gpu device input. on change of gpu that is specified by id, we should not unfocus the field when typing into it
- show type of inherited devices of type "other"

## QA

1. Run the LXD-UI:
    - edit instance, profile, project, network, storage pool and storage volume
    - ensure the save button shows the right number for all possible edits.

### DIFF
--- a/src/components/ConfigurationRow.tsx
+++ b/src/components/ConfigurationRow.tsx
@@ -103,20 +103,17 @@ export const getConfigurationRow = ({
             })}
           </div>,
         )}
-        {
-          <div>
-            <Button
-              onClick={toggleDefault}
-              type="button"
-              appearance="base"
-              title={disabled ? disabledReason : "Clear override"}
-              disabled={disabled}
-              hasIcon
-            >
-              <Icon name="close" className="clear-configuration-icon" />
-            </Button>
-          </div>
-        }
+        <Button
+          onClick={toggleDefault}
+          type="button"
+          appearance="base"
+          title={disabled ? disabledReason : "Clear override"}
+          disabled={disabled}
+          hasIcon
+          className="u-no-margin--bottom"
+        >
+          <Icon name="close" className="clear-configuration-icon" />
+        </Button>
       </div>
     );
   };
--- a/src/components/forms/CloudInitForm.tsx
+++ b/src/components/forms/CloudInitForm.tsx
@@ -1,13 +1,16 @@
 import { FC } from "react";
-import { Icon, Tooltip } from "@canonical/react-components";
+import { Button, Icon, Tooltip } from "@canonical/react-components";
 import CloudInitConfig from "components/forms/CloudInitConfig";
 import {
   InstanceAndProfileFormikProps,
   InstanceAndProfileFormValues,
 } from "./instanceAndProfileFormValues";
-import { getConfigurationRow } from "components/ConfigurationRow";
+import { getConfigurationRowBase } from "components/ConfigurationRow";
 import ScrollableConfigurationTable from "components/forms/ScrollableConfigurationTable";
 import { getInstanceKey } from "util/instanceConfigFields";
+import { ensureEditMode } from "util/instanceEdit";
+import classnames from "classnames";
+import { getConfigRowMetadata } from "util/configInheritance";
 
 export interface CloudInitFormValues {
   cloud_init_network_config?: string;
@@ -29,12 +32,74 @@ interface Props {
 }
 
 const CloudInitForm: FC<Props> = ({ formik }) => {
-  const codeRenderer = (value?: unknown) =>
-    value === "-" || value === undefined ? (
-      ""
-    ) : (
-      <CloudInitConfig config={value as string} />
-    );
+  const getCloudInitRow = (label: string, name: string, value?: string) => {
+    const metadata = getConfigRowMetadata(formik.values, name);
+    const isOverridden = value !== undefined;
+
+    return getConfigurationRowBase({
+      configuration: <strong>{label}</strong>,
+      inherited: (
+        <div
+          className={classnames({
+            "u-text--muted": isOverridden,
+            "u-text--line-through": isOverridden,
+          })}
+        >
+          <div className="mono-font">
+            <b>
+              <CloudInitConfig
+                key={`cloud-init-${name}`}
+                config={metadata.value as string}
+              />
+            </b>
+          </div>
+          {metadata && (
+            <div className="p-text--small u-text--muted">
+              From: {metadata.source}
+            </div>
+          )}
+        </div>
+      ),
+      override: isOverridden ? (
+        <>
+          <CloudInitConfig
+            config={value ?? ""}
+            setConfig={(config) => {
+              ensureEditMode(formik);
+              void formik.setFieldValue(name, config);
+            }}
+          />
+          <Button
+            onClick={() => {
+              ensureEditMode(formik);
+              void formik.setFieldValue(name, undefined);
+            }}
+            type="button"
+            appearance="base"
+            title={"Clear override"}
+            hasIcon
+            className="u-no-margin--bottom"
+          >
+            <Icon name="close" className="clear-configuration-icon" />
+          </Button>
+        </>
+      ) : (
+        <Button
+          onClick={() => {
+            ensureEditMode(formik);
+            void formik.setFieldValue(name, "\n\n");
+          }}
+          className="u-no-margin--bottom"
+          type="button"
+          appearance="base"
+          title="Create override"
+          hasIcon
+        >
+          <Icon name="edit" />
+        </Button>
+      ),
+    });
+  };
 
   return (
     <div className="cloud-init">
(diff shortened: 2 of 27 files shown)


mas-who:
Very cool that you are implementing this across the app! Functionality wise it works, although I think that the edit count should also decrease for non-effective changes. Below are some QA observations so far:

1. On instance main configuration page, if you have multiple profiles, editing individual profile does not add to the number of edits. Not sure if this is the right behaviour, I was expecting individual profiles to each contribute to the edit count. I guess perhaps it could be clearer if there was a dot next to the config title to show that the edit relates to the "Profiles" field for an instance?
2. Changing a config value, then changing it back to the original value does not reduce the edit count. I think this is misleading since technically that's not a change. This may not be feasible but is it possible to track all changes with respect to the initial form values?
3. On the yaml editor, shall we disable the edit count display? Since the number will always be 1 here.

src/util/instanceEdit.tsx
```diff
@@ -120,8 +120,23 @@ export const InstanceEditSchema: Yup.ObjectSchema<{
...
+export const trackChangedField = (
+  formik: ConfigurationRowFormikProps,
+  fieldName?: string,
+) => {
+  const oldValues = formik.values.changedFields ?? new Set<string>();
+  const changedFields = new Set([...oldValues, fieldName]);
+  if (changedFields.size > oldValues.size) {
+    void formik.setFieldValue("changedFields", changedFields);
+  }
 };
```
mas-who, line 142: Here is an rough idea (excuse the messy code), maybe we can do something like the below example to track field value changes against initial form values? There might be issues with the comparison since here I am assuming that field values have primitive types always e.g. if a field value is an object, then we would probably need a deep comparison.

Lastly, we probaly don't have to call this function in `ensureEditMode` and `enableOverride`.

```suggestion
export const trackChangedField = (
  formik: ConfigurationRowFormikProps,
  fieldName?: string,
) => {
  const oldValues = formik.values.changedFields ?? new Set<string>();
  const changedFields = new Set(oldValues);

  const values = formik.values as unknown as Record<string, string | undefined>;
  const initialValues = formik.initialValues as unknown as Record<
    string,
    string | undefined
  >;

  if (fieldName) {
    if (values[fieldName] === initialValues[fieldName]) {
      changedFields.delete(fieldName);
    } else {
      changedFields.add(fieldName);
    }
  }

  if (changedFields.size !== oldValues.size) {
    void formik.setFieldValue("changedFields", changedFields);
  }
};
```

edlerd: Using initial values is a good idea. That works well on primitive fields.
As you mention: unsolved are changes to devices and attached profiles, both are complex fields which are important to get right as well.

edlerd: Using initial values is a great idea. I completely changed the approach now, because we don't need to track the lifecycle of changes, we can calculate the changes on the fly based on the initial values and comparing it to current values. It is a bit involved to get that to a single number for complex objects, but much easier on containing the complexity.

mas-who:
This works so nice now! Just one QA observation, shall we disable the submit button if number of changes is zero?

src/pages/storage/forms/CreateStorageVolume.tsx
```diff
@@ -20,7 +20,7 @@ import { POOL } from "../StorageVolumesFilter";
 import FormFooterLayout from "components/forms/FormFooterLayout";
 import { useToastNotification } from "context/toastNotificationProvider";
 
-const StorageVolumeCreate: FC = () => {
+const CreateStorageVolume: FC = () => {
```
mas-who, line 23: amazing, thanks for changing these component names :)

src/components/forms/FormSubmitBtn.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const FormSubmitBtn: FC<Props> = ({ formik, disabled, isYaml = false }) => {
+  const changeCount = getFormChangeCount(formik);
+
+  return (
+    <ActionButton
+      appearance="positive"
+      loading={formik.isSubmitting}
+      disabled={!formik.isValid || disabled}
```
mas-who, line 20: maybe disabled if `changeCount === 0`?

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +32,76 @@ interface Props {
 }
 
 const CloudInitForm: FC<Props> = ({ formik }) => {
-  const codeRenderer = (value?: unknown) =>
-    value === "-" || value === undefined ? (
-      ""
-    ) : (
-      <CloudInitConfig config={value as string} />
-    );
+  const getCloudInitRow = (label: string, name: string, value?: string) => {
```
mas-who, line 35: nice refactor, the cloud init form looks much better now. Shall we move `getCloudInitRow` into it's own component?

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +32,76 @@ interface Props {
...
+      configuration: <b>{label}</b>,
+      inherited: (
+        <div
+          className={classnames({
+            "u-text--muted": isOverridden,
+            "u-text--line-through": isOverridden,
+          })}
+        >
+          <div className="mono-font">
+            <b>
```
mas-who, line 49: what's the `<b>` tag for?

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +32,76 @@ interface Props {
...
+      override: isOverridden ? (
+        <>
+          <CloudInitConfig
+            config={value ?? ""}
+            setConfig={(config) => {
+              ensureEditMode(formik);
+              void formik.setFieldValue(name, config);
+            }}
+          />
+          <div>
```
mas-who, line 72: is this `<div>` needed?

src/components/ConfigurationRow.tsx
```diff
@@ -103,20 +103,19 @@ export const getConfigurationRow = ({
...
+            type="button"
+            appearance="base"
+            title={disabled ? disabledReason : "Clear override"}
+            disabled={disabled}
+            hasIcon
+            className="u-no-margin--bottom"
+          >
+            <Icon name="close" className="clear-configuration-icon" />
+          </Button>
+        </div>
```
mas-who, line 118: unrelated to this change but is this `div` needed?

src/util/formChangeCount.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { FormDevice, FormDeviceValues } from "util/formDevices";
+import { ResourceLimitsFormValues } from "components/forms/ResourceLimitsForm";
+import { InstanceEditDetailsFormValues } from "pages/instances/EditInstance";
+
+const getPrimitiveFieldChanges = (
+  formik: ConfigurationRowFormikProps,
+): number => {
+  let changeCount = 0;
+
+  const ignoredFields = [
```
mas-who, line 14: small improvement, this could be a `Set` which is better for lookup and `Set.has()` is not any more complex than `Array.includes()` wdyt?

src/util/formChangeCount.tsx
```diff
@@ -0,0 +1,143 @@
...
+  ];
+
+  for (const key in formik.values) {
+    if (ignoredFields.includes(key)) {
+      continue;
+    }
+    const keyType = key as keyof ConfigurationRowFormikValues;
+    if (formik.values[keyType] !== formik.initialValues[keyType]) {
+      changeCount++;
+      console.log(key, formik.values[keyType], formik.initialValues[keyType]);
```
mas-who, line 30:
```suggestion
    
```

src/util/formChangeCount.tsx
```diff
@@ -0,0 +1,143 @@
+import {
```
mas-who, line 1: I think this file's functions can benefit from some unit tests. Perhaps a follow up task?

src/util/formChangeCount.tsx
```diff
@@ -0,0 +1,143 @@
...
+const getProfileChanges = (formik: ConfigurationRowFormikProps): number => {
+  if (!Object.hasOwn(formik.values, "profiles")) {
+    return 0;
+  }
+
+  const initProfiles = (formik.initialValues as InstanceEditDetailsFormValues)
+    .profiles;
+  const profiles = (formik.values as InstanceEditDetailsFormValues).profiles;
+
+  const hasOrderChange = () => {
```
mas-who, line 46: since profile names are unique, maybe it's easier to iterate the items in `initProfiles` and check if the corresponding item in `profiles` with the same index have the same name? Then do an early exit if any mismatch is found

mas-who (commented):
Thanks, the implementation is now much easier to understand! Left some comments for potential minor improvements.

I am a little concerned about the maintainability for this change though. Since adding any config related features in the future may require changes in `formChangeCount` and it would not be obvious. What are your thoughts about this?

src/components/forms/CloudInitForm.tsx
edlerd: As it is only relevant to this component, and we are accessing values from the component, moving it would increase the interface, probably better to keep it local.

src/components/forms/CloudInitForm.tsx
edlerd: We should use strong instead. b is to make things bold, but more legacy.

src/components/forms/CloudInitForm.tsx
edlerd: Yes, to force the button to a new line.

src/util/formChangeCount.tsx
edlerd: I added some tests to it.

src/util/formChangeCount.tsx
edlerd: yes, good idea -- changed this to be more simple and accurate.

edlerd:
> I am a little concerned about the maintainability for this change though. Since adding any config related features in the future may require changes in `formChangeCount` and it would not be obvious. What are your thoughts about this?

Not too concerned about this. Adding fields with primitive type should just work. Adding devices should just work. Adding new complex types should be rare. And it will not allow editing, which should surface latest in the QA step when checking the editing of the additional field.

mas-who:
> > I am a little concerned about the maintainability for this change though. Since adding any config related features in the future may require changes in `formChangeCount` and it would not be obvious. What are your thoughts about this?
> 
> Not too concerned about this. Adding fields with primitive type should just work. Adding devices should just work. Adding new complex types should be rare. And it will not allow editing, which should surface latest in the QA step when checking the editing of the additional field.

Good point, since we disable submit if the new field does not result in a change. LGTM! Thanks for this change :)

mas-who (approved):

PR #902, opened 2024-09-16
fix(idp-group) deselect idp group after deletion, deselct idp group on cancel of edit panel

edlerd:
## Done

- this fixes two issues with idp group selection
- 1. on edit idp group and cancel, we should unselect the group
- 2. on select idp group, then delete the single row, we should remove the deleted group from the selection.

## QA

1. Run the LXD-UI:
    - edit idp group, in the panel click cancel, ensure the group is unselected
    - select and idp group, delete it from the row (not bulk delete), ensure the group gets unselected
    - bulk delete idp groups, ensure they get unselected.


mas-who (approved):

mas-who:
Thanks for the fixes, LGTM :+1:

PR #903, opened 2024-09-16
Link identity and permission counts to the edit panel sections

edlerd:
## Done

- Link identity and permission counts to the edit panel sections

## QA

1. Run the LXD-UI:
    - Check permission > groups
    - click on the number of identities or permissions links to the edit panel section

### DIFF
--- a/src/pages/permissions/PermissionGroups.tsx
+++ b/src/pages/permissions/PermissionGroups.tsx
@@ -118,13 +118,31 @@ const PermissionGroups: FC = () => {
           title: group.description,
         },
         {
-          content: allIdentityIds.length,
+          content: (
+            <Button
+              appearance="link"
+              dense
+              onClick={() => panelParams.openEditGroup(group.name, "identity")}
+            >
+              {allIdentityIds.length}
+            </Button>
+          ),
           role: "cell",
           className: "u-align--right identities",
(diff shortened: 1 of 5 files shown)


edlerd:
@piperdeck please have a look at this one also for the changed group management that was merged in https://github.com/canonical/lxd-ui/pull/887

mas-who:
Design related question, maybe we show the identities and permissions count as links so that it's clear you can click on them?

edlerd:
I'm not sure we do want to link the numbers. But you are right @mas-who with the numbers as links in place this might be better.
 Still not sure if we should deviate from the table pattern of only linking in the first column to detail pages and having action buttons I. The last one.

mas-who:
> I'm not sure we do want to link the numbers. But you are right @mas-who with the numbers as links in place this might be better. Still not sure if we should deviate from the table pattern of only linking in the first column to detail pages and having action buttons in The last one.

Aah I see, yes it does deviate away from the common table pattern we have... but IMO I still think in this case having links serves a purpose and make the interaction clearer. Maybe @piperdeck will have a different perspective after design review :)

mas-who:
LGTM from my end, thanks for the changes :+1:

piperdeck:
Good to go!

PR #904, opened 2024-09-16
feat: show cached images for instance creation [WD-14414]

### LINKED ISSUE #839: Show cached images during selection for instance creation
It would be nice to show which images have been downloaded previously during image selection for creating instances. This could be indicated in the image selection modal as shown below. @piperdeck perhaps this could be an extra column in the modal table, wdyt?

Screenshot from 2024-08-12 09-16-33

mas-who:
## Done

- Show cached images when creating instances

Closes #839 

## QA

1. Run the LXD-UI:
    - Create an instance with a base image from any remote image server to download it for LXD
    - Then try create an instance again, make sure the previously downloaded image is available for selection in the image selection modal

## Screenshots
Screenshot from 2024-09-16 17-10-44

### DIFF
--- a/src/pages/images/ImageSelector.tsx
+++ b/src/pages/images/ImageSelector.tsx
@@ -1,6 +1,7 @@
 import { FC, OptionHTMLAttributes, useState } from "react";
 import {
   Button,
+  CheckboxInput,
   Col,
   MainTable,
   Modal,
@@ -19,6 +20,7 @@ import {
   isContainerOnlyImage,
   isVmOnlyImage,
   LOCAL_ISO,
+  LOCAL_IMAGE,
 } from "util/images";
 import Loader from "components/Loader";
 import { getArchitectureAliases } from "util/architectures";
@@ -55,6 +57,7 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
   const [arch, setArch] = useState<string>("amd64");
   const [type, setType] = useState<LxdImageType | undefined>(undefined);
   const [variant, setVariant] = useState<string>(ANY);
+  const [hideRemote, setHideRemote] = useState(false);
   const { project } = useParams<{ project: string }>();
 
   const loadImages = (file: string, server: string): Promise<RemoteImage[]> => {
@@ -105,15 +108,19 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
   const archSupported = getArchitectureAliases(
     settings?.environment?.architectures ?? [],
   );
-  const images = isLoading
+  let images = isLoading
     ? []
     : localImages
-        .filter((image) => !image.cached)
         .map(localLxdToRemoteImage)
-        .concat([...canonicalImages].reverse().sort(byLtsFirst))
-        .concat([...minimalImages].reverse().sort(byLtsFirst))
-        .concat([...imagesLxdImages])
-        .filter((image) => archSupported.includes(image.arch));
+        .sort((a, b) => Number(b.cached) - Number(a.cached));
+
+  if (!hideRemote) {
+    images = images
+      .concat([...canonicalImages].reverse().sort(byLtsFirst))
+      .concat([...minimalImages].reverse().sort(byLtsFirst))
+      .concat([...imagesLxdImages])
+      .filter((image) => archSupported.includes(image.arch));
+  }
 
   const archAll = [...new Set(images.map((item) => item.arch))]
     .filter((arch) => arch !== "")
@@ -210,19 +217,20 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
           : item.variant;
 
       const getSource = () => {
-        if (item.created_at) {
-          return "Local";
+        let source = "Custom";
+        if (!item.cached && item.created_at) {
+          source = "Local";
         }
         if (item.server === canonicalServer) {
-          return "Ubuntu";
+          source = "Ubuntu";
         }
         if (item.server === minimalServer) {
-          return "Ubuntu Minimal";
+          source = "Ubuntu Minimal";
         }
         if (item.server === imagesLxdServer) {
-          return "LXD Images";
+          source = "LXD Images";
         }
-        return "Custom";
+        return source;
       };
 
       return {
@@ -260,7 +268,12 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
             onClick: selectImage,
           },
           {
-            content: getSource(),
+            content: (
+              <>
+                {getSource()}
+                {item.cached && <span className="u-text--muted"> cached</span>}
+              </>
+            ),
             role: "cell",
             "aria-label": "Source",
             onClick: selectImage,
@@ -272,6 +285,11 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
                 type="button"
                 dense
                 className="u-no-margin--bottom"
+                appearance={
+                  item.cached || item.server === LOCAL_IMAGE
+                    ? "positive"
+                    : "default"
+                }
               >
                 Select
               </Button>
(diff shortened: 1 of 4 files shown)


src/pages/images/ImageSelector.tsx
```diff
@@ -403,6 +407,12 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
               ]}
               value={type ?? ""}
             />
+            <CheckboxInput
+              aria-label="Show cached images"
+              checked={showCached}
+              label="Show cached images"
+              onChange={() => setShowCached((prev) => !prev)}
+            />
```
edlerd, line 415: Do we need this, maybe it can be always on and doesn't need user control.

src/pages/images/ImageSelector.tsx
```diff
@@ -272,6 +275,7 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
                 type="button"
                 dense
                 className="u-no-margin--bottom"
+                appearance={item.cached ? "positive" : "default"}
```
edlerd, line 278: Should we also have the green one for local images?

src/util/images.tsx
```diff
@@ -48,9 +48,17 @@ export const localLxdToRemoteImage = (image: LxdImage): RemoteImage => {
     arch: image.architecture === "x86_64" ? "amd64" : image.architecture,
     os: image.properties?.os ?? "",
     created_at: new Date(image.uploaded_at).getTime(),
-    release: image.properties?.release ?? "",
-    server: LOCAL_IMAGE,
+    release:
+      image.properties?.os === "ubuntu"
+        ? (image.properties.version ?? "")
+        : (image.properties?.release ?? ""),
```
edlerd, line 54:
```suggestion
```
Why is this needed?
There is similar special handling of Ubuntu in the image selector component. The goal is to have it read nicely "Ubuntu xx.xx codename" as people are used to that ordering.

src/util/images.tsx
```diff
@@ -48,9 +48,17 @@ export const localLxdToRemoteImage = (image: LxdImage): RemoteImage => {
...
+    server: image.cached ? image.update_source?.server : LOCAL_IMAGE,
     type: image.type,
+    cached: image.cached,
+    variant:
+      image.properties?.os === "ubuntu"
+        ? image.properties?.release
+        : image.properties?.variant,
```
edlerd, line 61:
```suggestion
```
Seems not needed to me. Can it be removed?

src/pages/images/ImageSelector.tsx
mas-who: This was added after the discussion with @piperdeck . I think it could be useful to have the option to filter out cached images?

src/util/images.tsx
mas-who: This is added so that we can populate the "Release" column within the modal. The ternary condition is needed because for an `LxdImage` returned from the server, the data is not organised exactly the same as a `RemoteImage`. The field `image.properties.version` in this case for a cached `ubuntu` image aligns with the "Release" column content populated with the remote image data.

edlerd: Ah I see, this is handling the cached entries and transforming them into the same format as the ones we fetch from image servers.

edlerd: Still suspicious if we should have the special handling of Ubuntu here and in the modal.

edlerd: Suggesting to make this funciton more straight forward:
```
export const localLxdToRemoteImage = (image: LxdImage): RemoteImage => {
  return {
    aliases: image.update_source?.alias ?? image.aliases?.[0]?.name ?? "",
    fingerprint: image.fingerprint,
    arch: image.architecture === "x86_64" ? "amd64" : image.architecture,
    os: image.properties?.os ?? "",
    created_at: new Date(image.uploaded_at).getTime(),
    release: image.properties?.release ?? "",
    release_title: image.properties?.version ?? "",
    server: LOCAL_IMAGE,
    type: image.type,
  };
};
```
And keep the special handling of ubuntu in the ImageSelector.
If we pass the os here in this function through a converter, that ensures the first character is uppercase, it all should work well.

mas-who: yeah agreed, will implement that. I would keep the `cached` field though since we want to differentiate later on whether a local image is a cached remote image or uploaded image.

src/util/images.tsx
mas-who: closing this since the discussion is happening on the other comment.

src/pages/images/ImageSelector.tsx
edlerd: We can instead sort the cached images to the top of the list and keep the current behaviour of the "postive" and default button appearance.

src/pages/images/ImageSelector.tsx
mas-who: this is now removed

src/pages/images/ImageSelector.tsx
mas-who: implemented the sorting :+1:

src/pages/images/ImageSelector.tsx
```diff
@@ -210,19 +210,20 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
         if (item.server === minimalServer) {
-          return "Ubuntu Minimal";
+          source = "Ubuntu Minimal";
         }
         if (item.server === imagesLxdServer) {
-          return "LXD Images";
+          source = "LXD Images";
         }
-        return "Custom";
+        return source + (item.cached ? " (cached)" : "");
```
edlerd, line 226: I think we can better present this as a small muted text to not draw too much attention through the two lines with brackets

image

edlerd (commented):
QA and code LGTM

One idea to improve the design.

src/pages/images/ImageSelector.tsx
mas-who: great idea, done! :+1:

src/pages/images/ImageSelector.tsx
```diff
@@ -210,19 +210,25 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+          source = "LXD Images";
         }
-        return "Custom";
+        return (
+          <>
+            {source}{" "}
+            {item.cached ? <span className="u-text--muted">cached</span> : ""}
+          </>
+        );
       };
```
edlerd, line 232: Shall we keep this function to return the string as before. Then append the cached string based on the column below in the column? Seems a bit simpler to me.

edlerd (approved):
QA and code LGTM, just one comment on a possible change, but can also merge as is.

src/pages/images/ImageSelector.tsx
mas-who: done :+1:

src/pages/images/ImageSelector.tsx
```diff
@@ -260,7 +261,12 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
             onClick: selectImage,
           },
           {
-            content: getSource(),
+            content: (
+              <>
+                {getSource()}{" "}
+                {item.cached && <span className="u-text--muted">cached</span>}
```
edlerd, line 267: The space should be inside the conditional.

edlerd (approved):
All good, thanks for adding the cached image visibility.

src/pages/images/ImageSelector.tsx
mas-who: thanks! didn't pick that up as "cached" were all on a new line for me

edlerd (approved):

mas-who:
@edlerd had a discussion with @piperdeck and made the following updates:

1. Added a checkbox filter to "Show only cached images" if checked. The thinking here is that if the user for some reason have many cached images, and they want to sort those images for some reason (instead of search), then having the ability to sort cached images would be good. For remote images, we expect the user would  always want to do a search and filter really. I'm not entirely sure about this one TBH, but we agreed to implement it since it's quick and see what you think.

2. The "Select" button for local images (ones that we create in LXD or uploaded) also now have positive appearance.

piperdeck:
Please change the checkbox to "Only show cached images" instead of "Show only cached images". Then it's good to go!

PR #905, opened 2024-09-17
Ensure cache is cleared on image upload after the alias is created

edlerd:
## Done

- ensure cache is cleared on image upload after the alias is created
- this is to fix a ux issue and to make the image upload test more stable
- ensure we clear all the images caches, as we fetch images for all projects in permission components and need to clear those as well on update or deletion

## QA

1. Run the LXD-UI:
    - upload an image, ensure the new image appears in the image list with the alias specified.

### DIFF
--- a/src/pages/images/actions/forms/UploadImageForm.tsx
+++ b/src/pages/images/actions/forms/UploadImageForm.tsx
@@ -58,6 +58,12 @@ const UploadImageForm: FC<Props> = ({ close, project }) => {
     }
   };
 
+  const clearCache = () => {
+    void queryClient.invalidateQueries({
+      predicate: (query) => query.queryKey[0] === queryKeys.images,
+    });
+  };
+
   const formik = useFormik<{
     alias: string;
     isPublic: boolean;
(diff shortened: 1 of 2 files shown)


src/pages/images/actions/forms/UploadImageForm.tsx
```diff
@@ -58,6 +58,12 @@ const UploadImageForm: FC<Props> = ({ close, project }) => {
     }
   };
 
+  const clearCache = () => {
```
mas-who, line 61: great catch!

I just noticed that `fetchImageList` called in `EditGroupPanel` and `PermissionSelector` is cached with `queryKeys.images` only and not the project. So there would be a niche bug there since the permissions selector will not update correctly. Can we add `project` to the query keys for those api calls as part of this PR?

edlerd: Yes, good catch updated this to encounter for those two as well.

mas-who: aah yes, my suggestion is wrong, permissions operate outside of the scope of a single project so the cache keys there are correct so the invalidation logic should be updated. Thanks for the change :+1:

mas-who (approved):

PR #911, opened 2024-09-18
Show network parent in read and edit mode, when lxd is not clustered

edlerd:
## Done

- Show network parent in read and edit mode, when lxd is not clustered
- In clustered mode we can not read the parent, thus hiding it for now.

## QA

1. Run the LXD-UI:
    - edit a network that is of type physical in a clustered and non-clustered lxd.

### DIFF
--- a/src/pages/networks/forms/NetworkFormMain.tsx
+++ b/src/pages/networks/forms/NetworkFormMain.tsx
@@ -16,9 +16,10 @@ import { ensureEditMode } from "util/instanceEdit";
 interface Props {
   formik: FormikProps<NetworkFormValues>;
   project: string;
+  isClustered: boolean;
 }
 
-const NetworkFormMain: FC<Props> = ({ formik, project }) => {
+const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
   const getFormProps = (id: "network" | "name" | "description" | "parent") => {
     return {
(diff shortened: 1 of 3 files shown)


src/pages/networks/forms/NetworkFormMain.tsx
```diff
@@ -16,9 +16,10 @@ import { ensureEditMode } from "util/instanceEdit";
 interface Props {
   formik: FormikProps<NetworkFormValues>;
   project: string;
+  isClustered: boolean;
 }
 
-const NetworkFormMain: FC<Props> = ({ formik, project }) => {
+const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
```
mas-who, line 22: maybe we can check if lxd is clusterd directly in this component instead of passing it down as prop?

mas-who (approved):

src/pages/networks/forms/NetworkFormMain.tsx
edlerd: I did run into a "more hooks get rendered" error with the `useSettings` hook in the component and switching between network types. Hence, keeping the data loading in the parent to avoid it.

mas-who: I see, all good then

PR #913, opened 2024-09-18
Allow root disk device to have a custom name

edlerd:
## Done

- Allow root disk device to have a custom name

## QA

1. Run the LXD-UI:
    - create an instance where the root storage is overridden / set on the instance itself
    - edit instance in yaml editor, rename root device from "root" to another name, i.e. "rooty-root", save changes
    - ensure the guided forms (devices > disk) display the root device properly and custom volumes work as expected.

### DIFF
--- a/src/util/instanceValidation.tsx
+++ b/src/util/instanceValidation.tsx
@@ -18,8 +18,12 @@ export const hasNoRootDisk = (
   return missingRoot(values.devices) && !inheritsRoot(values, profiles);
 };
 
+export const isRootDisk = (device: FormDevice): device is FormDiskDevice => {
+  return device.type === "disk" && device.path === "/" && !device.source;
+};
+
 const missingRoot = (devices: FormDevice[]): boolean => {
-  return !devices.some((item) => item.type === "disk" && item.name === "root");
+  return !devices.some(isRootDisk);
 };
 
 const inheritsRoot = (
--- a/src/components/forms/DiskDeviceFormCustom.tsx
+++ b/src/components/forms/DiskDeviceFormCustom.tsx
@@ -15,7 +15,10 @@ import { getConfigurationRowBase } from "components/ConfigurationRow";
 import DetachDiskDeviceBtn from "pages/instances/actions/DetachDiskDeviceBtn";
 import classnames from "classnames";
 import { LxdStorageVolume } from "types/storage";
-import { isDiskDeviceMountPointMissing } from "util/instanceValidation";
+import {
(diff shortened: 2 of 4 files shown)


src/util/instanceValidation.tsx
```diff
@@ -18,8 +18,12 @@ export const hasNoRootDisk = (
   return missingRoot(values.devices) && !inheritsRoot(values, profiles);
 };
 
+export const isRootDisk = (device: FormDevice): device is FormDiskDevice => {
+  return device.type === "disk" && device.path === "/" && !device.source;
```
mas-who, line 22: how do we know this is the logic for determining root disk device?

edlerd: I think we only have two types of disks: root disk and custom storage volumes. The later have a source attached and the root disk is always on the path "/". I am adding the check that the source is not set, because when adding a new custom volume a user might enter /foo but for a short time, the item would match the path and type only check.

edlerd: Looking at the [docs](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_disk/#types-of-disk-devices), there are two other options device.ceph and device.cephfs All others use the device.source (except for the root storage). Will the two ceph fields to the filter as well.

edlerd: > Looking at the [docs](https://documentation.ubuntu.com/lxd/en/latest/reference/devices_disk/#types-of-disk-devices), there are two other options device.ceph and device.cephfs All others use the device.source (except for the root storage). Will the two ceph fields to the filter as well.

Actually not true, the two ceph types are also specified as the source, so this check should be fine.

mas-who: thanks for the clarification, this makes sense :+1:

mas-who:
Seeing this error when trying to update the name of the root

edlerd:
> Seeing this error when trying to update the name of the root
> It's weird though, I only have one device (root) with this instance, so not sure where the error message is coming from

Do you have a profile applied, that brings in a root device? If so, remove the profile first, then you can rename the one applied to the instance. Or rename the root device in the profile.

edlerd:
There was another problem: the inherited root device could have a custom name, and when overriding it, we should use the inherited root device name.

mas-who:
Okay all looking good now, like you said the issue was due to the default profile attached to my instance. LGTM :+1:

mas-who (approved):

PR #916, opened 2024-09-19
feat: [WD-14769] Move custom storage volume to another storage pool.

Kxiru:
## Done

- Implemented a button that opens a modal to migrate Custom storage volumes between different storage pools.
- Used Segmented-Controls css/ui for the Custom storage volume detail page.
- Created api endpoint 'migrateStorageVolume'.
- Added test for storage volume migration in storage.spec.ts.

## QA

1. Run the LXD-UI:
    - Navigate to a custom storage volume and view the Migrate button at the top-right of the page.
    - Attempt to change the storage pool of the volume.

## Screenshots
image
image
image

### DIFF
--- a/src/pages/storage/MigrateVolumeBtn.tsx
+++ b/src/pages/storage/MigrateVolumeBtn.tsx
@@ -1,0 +1,134 @@
+import { FC, useState } from "react";
+import { ActionButton, Icon } from "@canonical/react-components";
+import usePortal from "react-useportal";
+import { useQueryClient } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { useEventQueue } from "context/eventQueue";
+import ItemName from "components/ItemName";
+import { useToastNotification } from "context/toastNotificationProvider";
+import { LxdStorageVolume } from "types/storage";
+import MigrateVolumeModal from "./MigrateVolumeModal";
+import { migrateStorageVolume } from "api/storage-pools";
+import { useNavigate } from "react-router-dom";
+
+interface Props {
+  storageVolume: LxdStorageVolume;
+  project: string;
+  classname?: string;
+  onClose?: () => void;
+}
+
+const MigrateVolumeBtn: FC<Props> = ({
+  storageVolume,
+  project,
+  classname,
+  onClose,
+}) => {
+  const eventQueue = useEventQueue();
+  const toastNotify = useToastNotification();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const queryClient = useQueryClient();
+  const navigate = useNavigate();
+  const [isVolumeLoading, setVolumeLoading] = useState(false);
+
+  const handleSuccess = (
+    newTarget: string,
+    storageVolume: LxdStorageVolume,
+  ) => {
+    toastNotify.success(
+      <>
+        Volume <ItemName item={{ name: storageVolume.name }} bold />{" "}
+        successfully migrated to pool{" "}
+        <ItemName item={{ name: newTarget }} bold />
+      </>,
+    );
+
+    const oldVolumeUrl = `/ui/project/${storageVolume.project}/storage/pool/${storageVolume.pool}/volumes/${storageVolume.type}/${storageVolume.name}`;
+    const newVolumeUrl = `/ui/project/${storageVolume.project}/storage/pool/${newTarget}/volumes/${storageVolume.type}/${storageVolume.name}`;
+    if (window.location.pathname.startsWith(oldVolumeUrl)) {
+      navigate(newVolumeUrl);
+    }
+  };
+
+  const notifyFailure = (
+    e: unknown,
+    storageVolume: string,
+    targetPool: string,
+  ) => {
+    setVolumeLoading(false);
+    toastNotify.failure(
+      `Migration failed for volume ${storageVolume} to pool ${targetPool}`,
+      e,
+    );
+  };
+
+  const handleFailure = (
+    msg: string,
+    storageVolume: string,
+    targetPool: string,
+  ) => {
+    notifyFailure(new Error(msg), storageVolume, targetPool);
+  };
+
+  const handleFinish = () => {
+    void queryClient.invalidateQueries({
+      queryKey: [queryKeys.storage, storageVolume.name],
+    });
+    setVolumeLoading(false);
+  };
+
+  const handleMigrate = (targetPool: string) => {
+    setVolumeLoading(true);
+    migrateStorageVolume(storageVolume, targetPool)
+      .then((operation) => {
+        eventQueue.set(
+          operation.metadata.id,
+          () => handleSuccess(targetPool, storageVolume),
+          (err) => handleFailure(err, storageVolume.name, targetPool),
+          handleFinish,
+        );
+        toastNotify.info(
+          `Migration started for volume ${storageVolume.name} to pool ${targetPool}`,
+        );
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.storage, storageVolume.name, project],
+        });
+        handleClose();
+      })
+      .catch((e) => {
+        notifyFailure(e, storageVolume.name, targetPool);
+      });
+  };
+
+  const handleClose = () => {
+    closePortal();
+    onClose?.();
+  };
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <MigrateVolumeModal
+            close={handleClose}
+            migrate={handleMigrate}
+            storageVolume={storageVolume}
+          />
+        </Portal>
+      )}
+      <ActionButton
+        onClick={openPortal}
+        type="button"
+        className={classname}
+        loading={isVolumeLoading}
+        disabled={isVolumeLoading}
+        title="Migrate volume"
+      >
+        <Icon name="machines" />
+        <span>Migrate</span>
+      </ActionButton>
+    </>
+  );
+};
+
+export default MigrateVolumeBtn;
--- a/src/pages/storage/MigrateVolumeModal.tsx
+++ b/src/pages/storage/MigrateVolumeModal.tsx
@@ -1,0 +1,187 @@
+import { FC, KeyboardEvent, useState } from "react";
+import {
+  ActionButton,
+  Button,
+  MainTable,
+  Modal,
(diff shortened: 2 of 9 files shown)


src/pages/storage/forms/StorageVolumeFormMain.tsx
```diff
@@ -45,9 +32,26 @@ const StorageVolumeFormMain: FC<Props> = ({ formik, project }) => {
...
+          <StoragePoolSelector
+            project={project}
+            value={formik.values.pool}
+            setValue={(val) => void formik.setFieldValue("pool", val)}
+            hidePoolsWithUnsupportedDrivers
+            disabled={!formik.values.isCreating}
+            help={
+              formik.values.isCreating
+                ? undefined
+                : "Click 'Migrate' in the actions list to move the volume to a different storage pool."
```
edlerd, line 51:
```suggestion
                : "Use the migrate button in the header to move the volume to a different storage pool."
```
Maybe better? Users might not know what "action list" this refers to.

src/pages/storage/MigrateVolumeModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+      return;
+    }
+
+    close();
+  };
+
+  const headers = [
+    { content: "Name", sortKey: "name" },
+    { content: "Driver", sortKey: "driver" },
+    { content: "Size", sortKey: "size" },
```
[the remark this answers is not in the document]
Kxiru: I included this initially but nothing was inside the status lool. I'll include it again and do some more investigation.

src/pages/storage/forms/StorageVolumeFormMain.tsx
```diff
@@ -45,9 +32,26 @@ const StorageVolumeFormMain: FC<Props> = ({ formik, project }) => {
...
+
+          <Label forId="storage-pool-selector" required>
+            Storage pool
+          </Label>
+          <StoragePoolSelector
+            project={project}
+            value={formik.values.pool}
+            setValue={(val) => void formik.setFieldValue("pool", val)}
+            hidePoolsWithUnsupportedDrivers
+            disabled={!formik.values.isCreating}
```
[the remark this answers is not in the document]
Kxiru: There is no required flag set in the storagePoolSelector component and therefore I am assuming it is already false. Please could you clarify what you would like me to change?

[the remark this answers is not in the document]
edlerd: Sorry, that was not clear enough, the required is on the label above, not inside the selector component.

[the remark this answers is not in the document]
Kxiru: Right, but you said you want the selector component to have required set to false, I am asking, isn't that the default behaviour? As it has not been set at all in that component.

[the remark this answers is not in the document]
edlerd: I think this is still confused. What I meant is:

The label of the pool selector has a * in it, because it is a required label. It should keep the * during creation. Now that we also show it on edit as disabled, it should be the label without that *, meaning removing the required flag from the label in the edit case. The selector itself is fine and doesn't have a required flag, as it doesn't come with the label included.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -0,0 +1,137 @@
...
+  ) => {
+    toastNotify.success(
+      <>
+        Volume <ItemName item={{ name: storageVolume.name }} bold />{" "}
+        successfully migrated to <ItemName item={{ name: newTarget }} bold />
+      </>,
+    );
+
+    const newStorageVolumeURL = `/ui/project/${storageVolume.project}/storage/pool/${newTarget}/volumes/${storageVolume.type}/${storageVolume.name}`;
+    if (initialURL != newStorageVolumeURL) {
```
edlerd, line 49:
```suggestion
    const oldVolumeUrl = `/ui/project/${storageVolume.project}/storage/pool/${storageVolume.pool}/volumes/${storageVolume.type}/${storageVolume.name}`;
    const newVolumeUrl = `/ui/project/${storageVolume.project}/storage/pool/${newTarget}/volumes/${storageVolume.type}/${storageVolume.name}`;
    if (startsWith(location.pathname, oldVolumeUrl)) {
```
Two issues need solving:
1. We should compare the current active url, we don't need to pass the initial url around. That would be the url on submit of the request, but we want to know the current page the user is on.
2. We need to check the prefix of the url, because the user might be on the configuration or snapshots subpage with a nested url.

edlerd (commented):
One issue with the redirect and one question on changing the form field ordering. 

Also I think we should be more consistent in calling it "pool" vs "storage pool" and "volume" vs "storage volume".

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const handleSuccess = (
+    newTarget: string,
+    storageVolume: LxdStorageVolume,
+    initialURL: string,
+  ) => {
+    toastNotify.success(
+      <>
+        Volume <ItemName item={{ name: storageVolume.name }} bold />{" "}
+        successfully migrated to <ItemName item={{ name: newTarget }} bold />
```
edlerd, line 44:
```suggestion
        successfully migrated to pool <ItemName item={{ name: newTarget }} bold />
```

src/pages/storage/MigrateVolumeBtn.tsx
Kxiru: Apologies, this has been done within my local branch but I am having issues pushing them. I will fix this and have the old updates and new updates included.

tests/helpers/storageVolume.ts
```diff
@@ -53,3 +53,24 @@ export const saveVolume = async (page: Page, volume: string) => {
...
+    .locator("tr")
+    .filter({ hasText: targetPool })
+    .getByRole("button")
+    .click();
+  await page
+    .getByLabel("Confirm migration")
+    .getByRole("button", { name: "Migrate", exact: true })
+    .click();
+
+  await page.waitForSelector(`text=successfully migrated to ${targetPool}`);
```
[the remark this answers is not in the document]
Kxiru: The check on line 67 checks whether the pool is displayed on the page, which is kind of a check that the page refreshes (because it has to refresh to display the pool). Does that suffice or should I explicitly add a page check?

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -0,0 +1,137 @@
...
+        );
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.storage, storageVolume.name, project],
+        });
+      })
+      .catch((e) => {
+        notifyFailure(e, storageVolume.name, targetPool);
+      })
+      .finally(() => {
+        handleClose();
```
edlerd, line 102: I think we should not close on failure, the user might want to retry and we should keep the modal open. Only in the success case we should call handleClose.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <>
+        Volume <ItemName item={{ name: storageVolume.name }} bold />{" "}
+        successfully migrated to pool{" "}
+        <ItemName item={{ name: newTarget }} bold />
+      </>,
+    );
+
+    const oldVolumeUrl = `/ui/project/${storageVolume.project}/storage/pool/${storageVolume.pool}/volumes/${storageVolume.type}/${storageVolume.name}`;
+    const newVolumeUrl = `/ui/project/${storageVolume.project}/storage/pool/${newTarget}/volumes/${storageVolume.type}/${storageVolume.name}`;
+    if (location.pathname.startsWith(oldVolumeUrl)) {
```
edlerd, line 49:
```suggestion
    if (window.location.pathname.startsWith(oldVolumeUrl)) {
```
The location hook does not give us the current url at this point in time. So I do get redirected back to the new volume page, also in case I navigate quickly to the instances page after starting the migration process. But we can use the `window.location.pathname` instead and this should be fine.

edlerd (commented):
Two minor issues, then this should be good to go.

src/pages/storage/MigrateVolumeBtn.tsx
Kxiru: The success case, handleSuccess, doesn't need to handleClose in this case, right?

edlerd: Yeah, we'd close already on submitting the operation, no need to close again once the operation succeeds :)

edlerd: I might be overseeing, but don't we miss to close now in the success case? Like when the info notification gets submitted.

piperdeck:
Please see [my comment on PR 917](https://github.com/canonical/lxd-ui/pull/917#issuecomment-2372394464)

edlerd (commented):
QA and code LGTM mostly, just one thing left with giving a better experience in the error case.

src/pages/storage/MigrateVolumeBtn.tsx
edlerd: This is addressed be a new comment below.

Kxiru: i'm not sure I understand, but yes, there is no reference to handleClose or Close when the success criteria is met.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -0,0 +1,133 @@
...
+          () => handleSuccess(targetPool, storageVolume),
+          (err) => handleFailure(err, storageVolume.name, targetPool),
+          handleFinish,
+        );
+        toastNotify.info(
+          `Migration started for volume ${storageVolume.name} to pool ${targetPool}`,
+        );
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.storage, storageVolume.name, project],
+        });
```
edlerd, line 95:
```suggestion
        });
        handleClose();
```
Now the confirmation modal stays open, we should close it here.

edlerd (changes requested):

edlerd (approved):
LGTM

PR #917, opened 2024-09-20
feat: move instance root volume to another storage pool [WD-14768]

mas-who:
## Done

- for clustered LXD server, introduced a extra step for choosing the migration type between "storage pool" and "cluster member"
- for non-cluster LXD server, clicking the "Migrate" button will only show the storage pool select modal (new component)
- removed the storage pool selector for editing root disk device of an existing instance. The storage pool selector for root disk device will be present if the user is not editing an existing instance.
- also renamed `PanelFormLink` to `FormLink` and generalised its usage. Some of the permission specific logic had to be moved out in to `GroupForm`.
- extracted back linking logic from `GroupHeaderTitle` and generalised its usage into a component

## QA

1. Run the LXD-UI:
    - In a clustered LXD environment, try migrate the instance to a different cluster member
    - In a clustered LXD environment, try migrate the instance root volume to a different storage pool
    - In a non-clustered LXD environment, try migrate the instance root volume to a different storage pool

## Screenshots

Screenshot from 2024-09-20 15-03-10
Screenshot from 2024-09-20 15-03-40
Screenshot from 2024-09-20 15-03-52


src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,100 @@
...
-        archietecture: member.architecture.toLowerCase(),
-        status: member.status.toLowerCase(),
-      },
-    };
-  });
-
-  if (isLoading) {
-    return <Loader />;
+  let modalTitle: ReactNode = isClustered
+    ? `Choose migration method for instance ${instance.name}`
```
edlerd, line 31:
```suggestion
    ? `Choose migration method`
```

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,100 @@
...
+    : `Choose the target storage pool for instance ${instance.name}`;
+
+  const migrationEntity = isClustered
+    ? type?.split("-").join(" ")
+    : "storage pool";
+
+  if (["cluster-member", "storage-pool"].includes(type || "")) {
+    modalTitle = (
+      <BackLink
+        title={`Choose the target ${migrationEntity} for instance ${instance.name}`}
```
edlerd, line 41:
```suggestion
        title={`Choose new ${migrationEntity}`}
```

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,100 @@
...
-  migrate: (target: string, instance: LxdInstance) => void;
   instance: LxdInstance;
 }
 
-const MigrateInstanceModal: FC<Props> = ({ close, migrate, instance }) => {
-  const { data: members = [], isLoading } = useQuery({
-    queryKey: [queryKeys.cluster, queryKeys.members],
-    queryFn: fetchClusterMembers,
-  });
+type MigrationType = "cluster-member" | "storage-pool";
```
edlerd, line 16:
```suggestion
type MigrationType = "cluster member" | "root storage pool";
```

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,100 @@
...
-    return <Loader />;
+  let modalTitle: ReactNode = isClustered
+    ? `Choose migration method for instance ${instance.name}`
+    : `Choose the target storage pool for instance ${instance.name}`;
+
+  const migrationEntity = isClustered
+    ? type?.split("-").join(" ")
+    : "storage pool";
+
+  if (["cluster-member", "storage-pool"].includes(type || "")) {
```
edlerd, line 38:
```suggestion
  if (type !== "") {
```

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,100 @@
...
-  });
-
-  if (isLoading) {
-    return <Loader />;
+  let modalTitle: ReactNode = isClustered
+    ? `Choose migration method for instance ${instance.name}`
+    : `Choose the target storage pool for instance ${instance.name}`;
+
+  const migrationEntity = isClustered
+    ? type?.split("-").join(" ")
```
edlerd, line 35:
```suggestion
    ? type
```

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,100 @@
...
-    queryKey: [queryKeys.cluster, queryKeys.members],
-    queryFn: fetchClusterMembers,
-  });
+type MigrationType = "cluster-member" | "storage-pool";
 
-  const [selectedMember, setSelectedMember] = useState("");
+const MigrateInstanceModal: FC<Props> = ({ close, instance }) => {
+  const { data: settings } = useSettings();
+  const isClustered = isClusteredServer(settings);
+  const [type, setType] = useState<MigrationType | null>(null);
```
edlerd, line 21:
```suggestion
  const [type, setType] = useState<MigrationType | null>(isClustered ? null : "root storage pool");
```
Maybe this makes below simpler, we can hide the back link and remove the other special cases?

src/pages/instances/StoragePoolMigrationTable.tsx
edlerd, line 1: I think we should extract the table to a dedicated component, so we can reuse it in the custom storage volume migration.

src/pages/instances/StoragePoolMigrationTable.tsx
```diff
@@ -0,0 +1,244 @@
...
+        Instance <ItemName item={{ name: instanceName }} bold /> successfully
+        migrated to storage pool <ItemName item={{ name: newTarget }} bold />
+      </>,
+    );
+  };
+
+  const notifyFailure = (e: unknown, instanceName: string) => {
+    instanceLoading.setFinish(instance);
+    toastNotify.failure(
+      `Storage pool migration failed on instance ${instanceName}`,
```
edlerd, line 52:
```suggestion
      `Root storage migration failed for instance ${instanceName}`,
```

src/pages/instances/StoragePoolMigrationTable.tsx
```diff
@@ -0,0 +1,244 @@
...
+    migrateInstance(instance.name, instance.project, undefined, targetPool)
+      .then((operation) => {
+        eventQueue.set(
+          operation.metadata.id,
+          () => handleSuccess(targetPool, instance.name),
+          (err) => handleFailure(err, instance.name),
+          handleFinish,
+        );
+        toastNotify.info(
+          `Storage pool migration started for instance ${instance.name}`,
```
edlerd, line 79:
```suggestion
          `Root storage migration started for instance ${instance.name}`,
```

src/pages/instances/StoragePoolMigrationTable.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const { data: pools = [] } = useQuery({
+    queryKey: [queryKeys.storage],
+    queryFn: () => fetchStoragePools(instance.project),
+  });
+
+  const handleSuccess = (newTarget: string, instanceName: string) => {
+    toastNotify.success(
+      <>
+        Instance <ItemName item={{ name: instanceName }} bold /> successfully
+        migrated to storage pool <ItemName item={{ name: newTarget }} bold />
```
edlerd, line 44:
```suggestion
        Instance <ItemName item={{ name: instanceName }} bold /> root storage successfully
        migrated to pool <ItemName item={{ name: newTarget }} bold />
```

src/pages/instances/StoragePoolMigrationTable.tsx
```diff
@@ -0,0 +1,244 @@
...
+    { "aria-label": "Actions", className: "actions" },
+  ];
+
+  const rows = pools.map((pool) => {
+    const volumes =
+      pool.used_by?.filter((url) => !url.startsWith("/1.0/profiles")) ?? [];
+    const totalVolumeCount = volumes.length;
+    const rootDiskDevice = Object.values(instance.expanded_devices ?? {}).find(
+      (device) => isRootDisk(device as FormDevice),
+    ) as LxdDiskDevice;
```
edlerd, line 112: I think we should move this out of the loop.

src/pages/instances/StoragePoolMigrationTable.tsx
```diff
@@ -0,0 +1,244 @@
...
+        status: pool.status,
+        volumes: totalVolumeCount,
+      },
+    };
+  });
+
+  const summary = (
+    <div className="migrate-instance-summary">
+      <p>
+        This will migrate instance <strong>{instance.name}</strong> to storage
```
edlerd, line 194:
```suggestion
        This will migrate the instance <strong>{instance.name}</strong> root storage to
```

src/pages/instances/StoragePoolMigrationTable.tsx
```diff
@@ -0,0 +1,244 @@
...
+          </ScrollableTable>
+        </div>
+      )}
+      <footer id="migrate-instance-actions" className="p-modal__footer">
+        <Button
+          className="u-no-margin--bottom"
+          type="button"
+          aria-label="cancel migrate"
+          appearance="base"
+          onClick={close}
```
edlerd, line 227: This should have "go back" semantics

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,100 @@
...
+      {isClustered && !type && (
+        <div className="choose-migration-type">
+          <FormLink
+            icon="machines"
+            title="Migrate instance to a different cluster member"
+            onClick={() => setType("cluster-member")}
+          />
+          <FormLink
+            icon="pods"
+            title="Migrate instance to a different storage pool"
```
edlerd, line 74:
```suggestion
            title="Migrate instance root storage to a different pool"
```

src/pages/instances/ClusterMemberMigrationTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+          </ScrollableTable>
+        </div>
+      )}
+      <footer id="migrate-instance-actions" className="p-modal__footer">
+        <Button
+          className="u-no-margin--bottom"
+          type="button"
+          aria-label="cancel migrate"
+          appearance="base"
+          onClick={close}
```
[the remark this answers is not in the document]
mas-who: sounds good, I will change the button text to "Go back" as well to be clear :+1:

src/pages/instances/MigrateInstanceModal.tsx
mas-who: thanks this is much better! Just need to adjust the `MigrationType` type to include "" as well.

src/pages/instances/MigrateInstanceModal.tsx
mas-who: The simplest I can reduce the code below is something like 
```
const modalTitle = !type ? (
    "Choose migration method"
  ) : (
    <>
      {isClustered && (
        <BackLink
          title={target ? "Confirm migration" : `Choose new ${type}`}
          onClick={handleGoBack}
          linkText={target ? `Choose ${type}` : "Choose migration method"}
        />
      )}
      {!isClustered && (target ? "Confirm migration" : `Choose new ${type}`)}
    </>
  );
```
Is this what you had in mind? it's less code but I think it's still a little messy

src/pages/instances/StoragePoolMigrationTable.tsx
mas-who: will tackle this one, once #916 is merged

src/pages/instances/StoragePoolMigrationTable.tsx
mas-who: good catch thanks!

src/pages/instances/MigrateInstanceModal.tsx
mas-who: I double check the modal heading, summary copy and notification copy. It's consistently referring to "root storage" now.

src/pages/instances/StoragePoolMigrationTable.tsx
mas-who: @edlerd I added an extra commit moving logic out of `StoragePoolMigrationTable` and `ClusterMemberMigrationTable` into a custom hook `useInstanceMigration`. This should allow the table components to be reused easily for custom storage volume migration. The actual implementation might change a little bit depending on the final state for #916, let me know what you think?

edlerd: I am not sure if you need a hook for this. My idea was a pure rendering component for the pool selector. Removing the instance, and actions from the select logic:
```
const StoragePoolTableSelect: FC<Props> = ({
  onSelect,
}) => {
```

So the component only loads and renders the pools as a table with a select button next to it. It is purely to select a pool. You might need to pass in some info to disable the current row, but no more things should go into it. It can load the pools internally, render them and call a parent on click.

mas-who: right, this makes sense to me too. I created the `useInstanceMigration` custom hook for the reason that there is a fair bit of duplicated logic between `StoragePoolMigrationTable` and `ClusterMemberMigrationTable`. So the goal is a bit more than just splitting out the visual aspects of the storage pool selection table

mas-who: @edlerd made another change which splits out the storage pool selection into its own component `StoragePoolSelectTable`. I kept the custom hook `useInstanceMigration` for now though since I think it's quite nice to combine the repetitive stateful logic between cluster member and storage pool migration for instances.

src/pages/instances/MigrateInstanceModal.tsx
```diff
@@ -1,197 +1,120 @@
...
-  const handleMigrate = (instance: LxdInstance) => {
-    migrate(selectedMember, instance);
-    close();
-  };
+  const handleGoBack = () => {
+    // if lxd is not clustered, we close the modal
+    if (!isClustered) {
+      close();
+      return;
+    }
```
edlerd, line 41:
```suggestion
      close();
```
I think this should move to the bottom of the function and we don't need the comment, because we can also cancel on the 1st step in the clustered case and want to close under that condition.

src/util/instanceMigration.tsx
```diff
@@ -0,0 +1,134 @@
...
+    queryKey: [queryKeys.storage],
+    queryFn: () => fetchStoragePools(instance.project),
+    enabled: type === "root storage pool",
+  });
+
+  const { data: members = [] } = useQuery({
+    queryKey: [queryKeys.cluster, queryKeys.members],
+    queryFn: fetchClusterMembers,
+    enabled: type === "cluster member",
+  });
```
edlerd, line 46:
```suggestion
```
We should load the data, when it is needed, not unconditionally in this hook.

src/util/instanceMigration.tsx
```diff
@@ -0,0 +1,134 @@
...
+        );
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.instances, instance.name, instance.project],
+        });
+      })
+      .catch((e) => {
+        notifyFailure(e, instance.name);
+      })
+      .finally(() => {
+        close();
```
edlerd, line 121: I don't think we should call close on failure, probably only in the then block. Also I'd rename it to `onSuccess`, wdyt?

src/pages/instances/ClusterMemberMigrationTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { LxdInstance } from "types/instance";
+import ScrollableTable from "components/ScrollableTable";
+import { LxdClusterMember } from "types/cluster";
+
+interface Props {
+  instance: LxdInstance;
+  onSelect: (member: string) => void;
+  targetMember: string;
+  onCancel: () => void;
+  members: LxdClusterMember[];
```
edlerd, line 12:
```suggestion
```
We should load the members in here and add a loading indicator.

src/util/instanceMigration.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const toastNotify = useToastNotification();
+  const instanceLoading = useInstanceLoading();
+  const eventQueue = useEventQueue();
+  const queryClient = useQueryClient();
+  const { data: settings } = useSettings();
+  const isClustered = isClusteredServer(settings);
+  const [type, setType] = useState<MigrationType>(
+    isClustered ? "" : "root storage pool",
+  );
+  const [target, setTarget] = useState("");
```
edlerd, line 34: This state shouldn't live in the hook, we should pass in target and type. But the state should live in the parent caller make the interface of the hook smaller.

src/util/instanceMigration.tsx
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  return {
+    migrationType: type,
+    setMigrationType: setType,
+    target,
+    setTarget,
+    handleMigrate,
+    pools,
+    members,
```
edlerd, line 132:
```suggestion
    handleMigrate,
```

src/util/instanceMigration.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export type MigrationType = "cluster member" | "root storage pool" | "";
+
+type Props = {
+  instance: LxdInstance;
+  close: () => void;
+};
+
+export const useInstanceMigration = (props: Props) => {
+  const { instance, close } = props;
```
edlerd, line 24:
```suggestion
export const useInstanceMigration = ({ instance, close, type, target }: Props) => {
```
We tend to destructure the props in the header everywhere else in the repo.

src/util/instanceMigration.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { ReactNode, useState } from "react";
+import { useSettings } from "context/useSettings";
+import { isClusteredServer } from "./settings";
+import { capitalizeFirstLetter } from "./helpers";
+
+export type MigrationType = "cluster member" | "root storage pool" | "";
+
+type Props = {
+  instance: LxdInstance;
+  close: () => void;
```
edlerd, line 20:
```suggestion
  close: () => void;
  type: MigrationType;
  target: string;
```

src/pages/instances/StoragePoolMigrationTable.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface Props {
+  instance: LxdInstance;
+  onSelect: (pool: string) => void;
+  targetPool: string;
+  onCancel: () => void;
+  pools: LxdStoragePool[];
+  migrate: (pool: string) => void;
+}
+
+const StoragePoolMigrationTable: FC<Props> = ({
```
edlerd, line 19:
```suggestion
const InstanceStoragePoolMigration: FC<Props> = ({
```
Not a table anymore, and I think we should be more specific to avoid the name clash with the volume pool migration.

src/pages/instances/StoragePoolSelectTable.tsx
edlerd, line 1: This should probably live in the storage module and not under instances.

src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -105,16 +108,29 @@ const DiskDeviceFormRoot: FC<Props> = ({
...
+                    void formik.setFieldValue(
+                      `devices.${rootIndex}.pool`,
+                      value,
+                    )
+                  }
+                  selectProps={{
+                    className: restrictEditStoragePool
+                      ? ""
+                      : "u-no-margin--bottom",
+                    disabled: restrictEditStoragePool,
```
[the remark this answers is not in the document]
mas-who: sure, we still need to conditionally apply `u-no-margin--bottom` though, otherwise the help text will appear right against the selector's bottom border.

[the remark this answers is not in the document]
edlerd: > sure, we still need to conditionally apply `u-no-margin--bottom` though, otherwise the help text will appear right against the selector's bottom border.

Yes, that is why I kept in the suggestion above :)

src/pages/instances/StoragePoolSelectTable.tsx
```diff
@@ -0,0 +1,129 @@
+import { FC } from "react";
+import { Button, MainTable } from "@canonical/react-components";
+import ScrollableTable from "components/ScrollableTable";
+import StoragePoolSize from "pages/storage/StoragePoolSize";
+import { LxdStoragePool } from "types/storage";
+
+interface Props {
+  pools: LxdStoragePool[];
+  onSelect: (pool: string) => void;
+  disablePool?: (pool: LxdStoragePool) => string | undefined;
```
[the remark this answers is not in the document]
mas-who: that was my initial approach, however since this is a more general component, we should consider what if we want to disable multiple rows?  We could pass in a list of objects with your suggested structure, but I think it would be easier to allow the definition of a disable function since through the callback arguments we can get all the info we need for a storage pool wdyt?

src/components/forms/DiskDeviceFormRoot.tsx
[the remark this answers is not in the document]
mas-who: not seeing it :sweat_smile: `className` is assigned with "u-no-margin--bottom" without condition?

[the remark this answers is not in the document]
edlerd: I am confused, you are probably right :)

src/pages/instances/MigrateInstanceModal.tsx
mas-who: `migrationType` would be set to "root storage pool" initially if lxd is not clustered. Moving `close` to the bottom would result in the modal not being able to close.

src/pages/instances/StoragePoolSelectTable.tsx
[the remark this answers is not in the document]
edlerd: I was thinking about this too and as for the two use cases at hand we have only one row to disable, probably fine to only support that case for now. If we ever need more, we can enhance it.

src/pages/instances/MigrateInstanceModal.tsx
edlerd: Oh right, there is the special case for legacy. I don't really like that it increases complexity here, but don't have a better idea. Probably fine as is.

Do we need another `close()` at the end of the function for the first step in the clustered case?

src/pages/instances/StoragePoolSelectTable.tsx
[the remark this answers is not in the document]
mas-who: cool, done :+1:

src/pages/instances/MigrateInstanceModal.tsx
mas-who: The first step for the clustered case does not have a cancel button. The modal at that point will be closed by `close()` always.

src/util/instanceMigration.tsx
mas-who: 100%, I wasn't quite done with this part yet since I was planning on waiting for #916 to get merged. tbh I don't think we need to wait anymore though, I'm quite sure that `StoragePoolSelectTable` can be quite easily implemented in a follow up PR

mas-who: these calls will be moved into the respective table components

mas-who:
@edlerd one more change I'd like to make is to also split out the select table for cluster members. The code would be more consistent that way and also if we do want to add the feature for migrating custom volume to another cluster member later, we can reuse the table component.

src/util/instanceMigration.tsx
mas-who: yeah that sounds good :+1:

src/util/instanceMigration.tsx
mas-who: yes that's neat! :+1:

src/pages/instances/StoragePoolMigrationTable.tsx
mas-who: same goes for `ClusterMemberMigrationTable` -> `InstanceClusterMemberMigration`

src/pages/instances/StoragePoolSelectTable.tsx
mas-who: yes, same goes for `ClusterMemberSelectTable`, moving it to `/cluster`

src/util/instanceMigration.tsx
edlerd: The handle and on prefix is correct, I meant renaming the close prop (which doesn't fit any of the two oO) to onSuccess.

src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -105,16 +107,23 @@ const DiskDeviceFormRoot: FC<Props> = ({
...
+                  project={project}
+                  value={formRootDevice?.pool ?? ""}
+                  setValue={(value) =>
+                    void formik.setFieldValue(
+                      `devices.${rootIndex}.pool`,
+                      value,
+                    )
+                  }
+                  selectProps={{
+                    className: isEditingInstance ? "" : "u-no-margin--bottom",
```
edlerd, line 121: We need the class also in the non-edit/create case

image

src/pages/storage/StoragePoolSelectTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+    },
+    { "aria-label": "Actions", className: "actions" },
+  ];
+
+  const rows = pools.map((pool) => {
+    const volumes =
+      pool.used_by?.filter((url) => !url.startsWith("/1.0/profiles")) ?? [];
+    const totalVolumeCount = volumes.length;
+
+    const disableReason = disablePool?.name === pool.name && disablePool.reason;
```
edlerd, line 46: I think we should find a better way to visualize the disabled row. Maybe remove the hover effect for it, or we could try muted text in that row. Or hide it instead of disabling?

edlerd (commented):
This looks way cleaner, thanks for the changes. Two more comments below.

src/util/instanceMigration.tsx
mas-who: oh my bad for misunderstanding, yes that makes sense :+1:

src/components/forms/DiskDeviceFormRoot.tsx
mas-who: aah yes, actually we need to hide the help text if we are not editing an instance. `u-no-margin--bottom` is only needed when there is no help text in order to reduce vertical spacing between the pool selector and the size input

src/pages/storage/StoragePoolSelectTable.tsx
mas-who: going with muted text, show not-allowed cursor for a disabled row. wdyt?

edlerd: Sounds OK.
The not allowed cursor is barely used on the web, maybe just remove the hover effect and use the default pointer?

edlerd (dismissed):
LGTM

mas-who:
> LGTM

Thanks for the review! I'm quite happy about how clean the code has become :) Time for a design review please @piperdeck :pray:

piperdeck:
This modal has a primary CTA button at the bottom which is always disabled, and never able to be clicked. Could we simply remove the entire CTA footer from the modal, since the "Cancel" functionality is already provided by the close button at the top right?

CleanShot 2024-09-24 at 16 53 19@2x

I've also noticed that the modals for moving custom storage volumes and instance root volumes have different columns, different title formats, and slightly different dimensions. Could we make these modals match each other?


| Custom volume                                                                                                     | Instance                                                                                                      |
| ----------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| CleanShot 2024-09-24 at 16 55 05@2x | CleanShot 2024-09-24 at 16 55 00@2x |

For the title, I would go with:

| Custom volume                                       | Instance                                           |
| --------------------------------------------------- | -------------------------------------------------- |
| Choose storage pool for custom volume `volume name` | Choose storage pool for instance `<instance name>` |

and for the columns, I would keep the one being used by the root volume modal, and show those columns in that order for the instance modal as well.

mas-who:
> This modal has a primary CTA button at the bottom which is always disabled, and never able to be clicked. Could we simply remove the entire CTA footer from the modal, since the "Cancel" functionality is already provided by the close button at the top right?
> 
> CleanShot 2024-09-24 at 16 53 19@2x
> 

Yeah agreed, I will remove the action buttons for non-clustered scenario. In a clustered environment, the cancel button leads to prior steps in the modal flow.

> I've also noticed that the modals for moving custom storage volumes and instance root volumes have different columns, different title formats, and slightly different dimensions. Could we make these modals match each other?
> 
> Custom volume	Instance
> CleanShot 2024-09-24 at 16 55 05@2x	CleanShot 2024-09-24 at 16 55 00@2x
> For the title, I would go with:
> 
> Custom volume	Instance
> Choose storage pool for custom volume `volume name`	Choose storage pool for instance `<instance name>`
> and for the columns, I would keep the one being used by the root volume modal, and show those columns in that order for the instance modal as well.

@piperdeck agreed on both :) nice one catching the difference 

> This modal has a primary CTA button at the bottom which is always disabled, and never able to be clicked. Could we simply remove the entire CTA footer from the modal, since the "Cancel" functionality is already provided by the close button at the top right?
> 
> CleanShot 2024-09-24 at 16 53 19@2x
> 
> I've also noticed that the modals for moving custom storage volumes and instance root volumes have different columns, different title formats, and slightly different dimensions. Could we make these modals match each other?
> 
> Custom volume	Instance
> CleanShot 2024-09-24 at 16 55 05@2x	CleanShot 2024-09-24 at 16 55 00@2x
> For the title, I would go with:
> 
> Custom volume	Instance
> Choose storage pool for custom volume `volume name`	Choose storage pool for instance `<instance name>`

Sure I can update the copy text for the modal title. @edlerd this is contradictory to what's been changed in the code review, although I do agree with @piperdeck that having the instance/custom volume names here provides context to the resource that we are modifying, which is better IMO even if it makes the title a little longer. We should probably truncate the instance/volume name in the modal title though.
 
> and for the columns, I would keep the one being used by the root volume modal, and show those columns in that order for the instance modal as well.

@piperdeck do you mean keep the columns in the custom volume migration modal and use them in the instance root volume migration modal?

mas-who:
> > This modal has a primary CTA button at the bottom which is always disabled, and never able to be clicked. Could we simply remove the entire CTA footer from the modal, since the "Cancel" functionality is already provided by the close button at the top right?
> > CleanShot 2024-09-24 at 16 53 19@2x
> 
> Yeah agreed, I will remove the action buttons for non-clustered scenario. In a clustered environment, the cancel button leads to prior steps in the modal flow.
> 
> > I've also noticed that the modals for moving custom storage volumes and instance root volumes have different columns, different title formats, and slightly different dimensions. Could we make these modals match each other?
> > Custom volume	Instance
> > CleanShot 2024-09-24 at 16 55 05@2x	CleanShot 2024-09-24 at 16 55 00@2x
> > For the title, I would go with:
> > Custom volume	Instance
> > Choose storage pool for custom volume `volume name`	Choose storage pool for instance `<instance name>`
> > and for the columns, I would keep the one being used by the root volume modal, and show those columns in that order for the instance modal as well.
> 
> > I've also noticed that the modals for moving custom storage volumes and instance root volumes have different columns, different title formats, and slightly different dimensions. Could we make these modals match each other?

Yes we will make the modal consistent, these are two separate PRs for now though.

> > For the title, I would go with:
> > Custom volume	Instance
> > Choose storage pool for custom volume `volume name`	Choose storage pool for instance `<instance name>`
> 
> Sure I can update the copy text for the modal title. @edlerd this is contradictory to what's been changed in the code review, although I do agree with @piperdeck that having the instance/custom volume names here provides context to the resource that we are modifying, which is better IMO even if it makes the title a little longer. We should probably truncate the instance/volume name in the modal title though.
> 
> > and for the columns, I would keep the one being used by the root volume modal, and show those columns in that order for the instance modal as well.
> 
> @piperdeck do you mean keep the columns in the custom volume migration modal and use them in the instance root volume migration modal?

piperdeck:
+1 on the modal title.

Everything looks good to me!

mas-who:
@edlerd I have replaced the custom volume storage pool selection table with the `StoragePoolSelectTable` component in this PR. If all is good, would you approve again thanks :)

mas-who:
@edlerd I've also added a e2e test for this feature, relating to this [task](https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/930?assignee=712020%3Accb7b61f-3f6a-41ec-9935-a79519dada4e&selectedIssue=WD-15104). Note the test checks if the lxd environment is clustered since the instance root storage migration steps are different and we always have a lxd cluster in CI.

src/components/BackLink.tsx
```diff
@@ -0,0 +1,32 @@
+import { Button, Icon } from "@canonical/react-components";
+import { FC, ReactNode } from "react";
+
+interface Props {
+  linkText: string;
+  title: ReactNode;
+  onClick: () => void;
+}
+
+const BackLink: FC<Props> = ({ linkText, title, onClick }) => {
```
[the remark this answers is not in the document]
edlerd: That is not a good solution. Also the truncate doesn't work on the selector.

I think we should not have any custom CSS on the title and also remove the truncation. The heading will naturally wrap to multiple lines then.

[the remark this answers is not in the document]
mas-who: perfect, thanks that makes things a whole lot simpler :+1:

edlerd (approved):
QA and Code LGTM, good work :+1:

PR #918, opened 2024-09-23
fix: remove workaround for extracting instance name from snapshot operation [WD-14541]

mas-who:
## Done

- previously, an operation for creating an instance from a snapshot would include the snapshot name in the instances resource field within the operation response metadata. We had to include a workaround to filter out the snapshots in order to extract the instance name.
- with this upstream [PR](https://github.com/canonical/lxd/pull/14139) merged, we can now remove the workaround

## QA

1. Run the LXD-UI:
    - Create an instance from a snapshot
    - Make sure that the operation shows the new instance name

### DIFF
--- a/src/types/operation.d.ts
+++ b/src/types/operation.d.ts
@@ -15,6 +15,7 @@ export interface LxdOperation {
   may_cancel: boolean;
   resources?: {
     instances?: string[];
+    instances_snapshots?: string[];
   };
   status: LxdOperationStatus;
   status_code: string;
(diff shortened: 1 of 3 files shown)


mas-who:
@edlerd I tried refreshing my lxd snap on `latest/edge` channel and I see the commit isn't pointing at the latest on main branch yet. It could be getting released to the snap channel still, so we may have to wait a bit before testing again.

edlerd (approved):
LGTM

PR #923, opened 2024-09-24
feat: [WD-14774] Proxy Device Test

Kxiru:
## Done

- Added a test to create, edit and remove Proxy devices from a Profile.

## QA

1. Run the LXD-UI:
    - Run playwright test.

## Screenshots

N/A

### DIFF
--- a/tests/devices.spec.ts
+++ b/tests/devices.spec.ts
@@ -17,6 +17,7 @@ import {
 import { attachVolume, detachVolume } from "./helpers/devices";
 import { deleteVolume, randomVolumeName } from "./helpers/storageVolume";
 import { assertTextVisible } from "./helpers/permissions";
+import { expect } from "./fixtures/lxd-test";
 
 test("instance attach custom volumes and detach inherited volumes", async ({
   page,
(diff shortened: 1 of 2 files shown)


tests/devices.spec.ts
```diff
@@ -106,3 +106,61 @@ test("profile edit networks", async ({ page }) => {
...
+    .getByPlaceholder("127.0.0.1")
+    .fill("127.0.0.3");
+  await saveProfile(page, profile, 1);
+
+  //Edit and save Proxy Device
+  await visitProfile(page, profile);
+  await page
+    .getByRole("gridcell", { name: "127.0.0.3" })
+    .getByPlaceholder("127.0.0.1")
+    .isVisible();
```
edlerd, line 140: I think we should get by the label here as well, to ensure the right field has received the right value. Also Better to use expect, like below. This is true also for the other assertions below.

```
await expect(
    page....,
  ).toBeVisible();
```

tests/devices.spec.ts
```diff
@@ -106,3 +106,61 @@ test("profile edit networks", async ({ page }) => {
...
+  await page.getByText("Proxy", { exact: true }).click();
+  await page.getByRole("button", { name: "New Proxy Device" }).click();
+  await page.getByLabel("Bind").selectOption({ index: 2 });
+  await page.getByPlaceholder("127.0.0.1").first().fill("127.0.0.1");
+  await page.getByPlaceholder("[-00]").first().fill("3000");
+  await page
+    .locator('[id="devices\\.0\\.connect\\.address"]')
+    .fill("127.0.0.2");
+  await page.getByPlaceholder("[-00]").nth(1).fill("3001");
+  await finishProfileCreation(page, profile);
```
edlerd, line 124: I think we should assert the values from creation as well. We already assert the ones after create, then edit below. But the values directly after creation could be missing or wrong.

tests/devices.spec.ts
```diff
@@ -106,3 +107,63 @@ test("profile edit networks", async ({ page }) => {
...
+  await saveProfile(page, profile, 1);
+
+  //Edit and save Proxy Device (Refresh)
+  await visitProfile(page, profile);
+  await page.getByTestId("tab-link-Configuration").click();
+  await page.getByText("Proxy", { exact: true }).click();
+  await expect(
+    page
+      .getByRole("gridcell", { name: "127.0.0.3" })
+      .getByPlaceholder("127.0.0.1"),
```
edlerd, line 139: I think we should work with the label instead of the placeholder here and also in the other assertions below.

tests/devices.spec.ts
```diff
@@ -106,3 +107,63 @@ test("profile edit networks", async ({ page }) => {
...
+
+test("add, edit and remove a proxy device", async ({ page }) => {
+  const profile = randomProfileName();
+
+  //Add Proxy Device
+  await startProfileCreation(page, profile);
+  await page.getByText("Proxy", { exact: true }).click();
+  await page.getByRole("button", { name: "New Proxy Device" }).click();
+  await page.getByLabel("Bind").selectOption({ index: 2 });
+  await page.getByLabel("*Address").first().fill("127.0.0.1");
```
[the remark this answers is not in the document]
edlerd: Please also change the labels to not include the * as the label text in the Proxy device component. This slipped the review the proxy device PR.

edlerd (commented):
LGTM, please change the proxy device labels to use the `required` flag instead of adding the * to the label text directly. That way the * is not read by screen readers and the spacing of the * and the label text is better.

Kxiru:
> LGTM, please change the proxy device labels to use the `required` flag instead of adding the * to the label text directly. That way the * is not read by screen readers and the spacing of the * and the label text is better.

All done! Please let me know your thoughts :)

edlerd (approved):
LGTM, great how readable the test has become :+1:

PR #924, opened 2024-09-24
feat: duplicate custom storage volume [WD-15107]

mas-who:
## Done

- support duplicate custom storage volume

## QA

1. Run the LXD-UI:
    - Navigate to a custom storage volumes detail page, try duplicate the volume by using the "Duplicate" button.

## Screenshots
Screenshot from 2024-09-24 18-51-04

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -381,3 +381,38 @@ export const migrateStorageVolume = (
       .catch(reject);
   });
 };
+
+// Including project and target params if they did not change from source configs breaks the API call.
+// Therefore, we only include them if they are different from the source configs, that's why both project and target are optional inputs
+export const duplicateStorageVolume = (
+  volume: Partial<LxdStorageVolume>,
+  pool: string,
+  project?: string,
+  target?: string,
+): Promise<LxdOperationResponse> => {
+  return new Promise((resolve, reject) => {
+    const url = new URL(
+      `/1.0/storage-pools/${pool}/volumes/custom`,
+      window.location.origin,
+    );
+    const params = new URLSearchParams();
+
+    if (project) {
+      params.append("project", project);
+    }
+
+    if (target) {
+      params.append("target", target);
+    }
+
+    url.search = params.toString();
+
+    fetch(url.toString(), {
+      method: "POST",
+      body: JSON.stringify(volume),
+    })
+      .then(handleResponse)
+      .then(resolve)
+      .catch(reject);
+  });
+};
--- a/src/pages/storage/forms/DuplicateVolumeForm.tsx
+++ b/src/pages/storage/forms/DuplicateVolumeForm.tsx
@@ -1,0 +1,242 @@
+import { FC, useState } from "react";
+import { useFormik } from "formik";
+import { useToastNotification } from "context/toastNotificationProvider";
+import {
+  ActionButton,
+  Button,
+  Form,
+  Input,
+  Modal,
+  Select,
+} from "@canonical/react-components";
+import * as Yup from "yup";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { duplicateStorageVolume } from "api/storage-pools";
+import { Link, useNavigate } from "react-router-dom";
+import { fetchProjects } from "api/projects";
+import { useEventQueue } from "context/eventQueue";
+import { LxdStorageVolume } from "types/storage";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
+import { loadCustomVolumes } from "context/loadCustomVolumes";
+import ItemName from "components/ItemName";
+import StoragePoolSelector from "../StoragePoolSelector";
+import { checkDuplicateName, getUniqueResourceName } from "util/helpers";
+
+interface Props {
+  volume: LxdStorageVolume;
+  close: () => void;
+}
+
+export interface StorageVolumeDuplicate {
+  name: string;
+  project: string;
+  pool: string;
+  copySnapshots: boolean;
+}
+
+const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
+  const toastNotify = useToastNotification();
+  const controllerState = useState<AbortController | null>(null);
+  const navigate = useNavigate();
+  const eventQueue = useEventQueue();
+  const { hasStorageVolumesAll } = useSupportedFeatures();
+
+  const { data: projects = [], isLoading: projectsLoading } = useQuery({
+    queryKey: [queryKeys.projects],
+    queryFn: fetchProjects,
+  });
+
+  const { data: volumes = [], isLoading: volumesLoading } = useQuery({
+    queryKey: [queryKeys.customVolumes, volume.project],
+    queryFn: () => loadCustomVolumes(volume.project, hasStorageVolumesAll),
+  });
+
+  const notifySuccess = (volumeName: string, project: string, pool: string) => {
+    const message = (
+      <>
+        Created volume <strong>{volumeName}</strong>.
+      </>
+    );
+
+    const redirectUrl = `/ui/project/${project}/storage/pool/${pool}/volumes/custom/${volumeName}/configuration`;
+    const actions = [
+      {
+        label: "Configure",
+        onClick: () => navigate(redirectUrl),
+      },
+    ];
+
+    toastNotify.success(message, actions);
+  };
+
+  const getDuplicatedVolumeName = (volume: LxdStorageVolume): string => {
+    const newVolumeName = volume.name + "-duplicate";
+    return getUniqueResourceName(newVolumeName, volumes);
+  };
+
+  // this validation checks given changes in any one of the fields (name, project, pool)
+  // if the volume name already exists in the target project and storage pool
+  const validationSchema = Yup.object()
+    .shape({
+      name: Yup.string().required("Volume name is required"),
+      project: Yup.string().required(),
+      pool: Yup.string().required(),
+    })
+    .test("deduplicate", "", async function (values) {
+      const { name, project, pool } = values;
+      const notFound = await checkDuplicateName(
+        name,
+        project || "",
+        controllerState,
+        `storage-pools/${pool}/volumes/custom`,
+      );
+
+      if (notFound) {
+        return true;
+      }
+
+      return this.createError({
+        path: "name",
+        message:
+          "A volume with this name already exist in the target project and storage pool",
+      });
+    });
+
+  const formik = useFormik<StorageVolumeDuplicate>({
+    initialValues: {
+      name: getDuplicatedVolumeName(volume),
+      project: volume.project,
+      copySnapshots: true,
+      pool: volume.pool,
+    },
+    enableReinitialize: true,
+    validationSchema,
+    onSubmit: (values) => {
+      const payload: Partial<LxdStorageVolume> = {
+        name: values.name,
+        type: "custom",
+        config: volume.config,
+        description: volume.description,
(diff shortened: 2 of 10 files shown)


mas-who:
@edlerd @Kxiru please note the following for this PR:
1. The duplicate button should be grouped into a segmented control and contextual menu for smaller screen sizes. I will execute that once #916 is merged.
2. There is an issue with the API endpoint for duplicating custom storage volumes. If the target project and cluster member is set at the same time, it results in a `400` bad request error. See my [question](https://chat.canonical.com/canonical/pl/55ddjpynufgntmig5dsuuaighc) on mattermost for this as well. I think this is very likely a bug in the backend since the CLI has the same behaviour. In this case, I wonder maybe we should disallow the user to set both target project and cluster member at the same time in the UI? We can check that if either target project or cluster member changed from initial values, then disable one of the selectors wdyt?

mas-who:
@edlerd as discussed I've implemented the logic to disable either project or cluster member selector depending on which option has been set. For reference, this is related to this [issue](https://github.com/canonical/lxd/issues/14155). The one remaining thing to do here is to change the button into a segmented control, which as mentioned I will do when #916 is merged.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+      if (notFound) {
+        return true;
+      }
+
+      return this.createError({
+        path: "name",
+        message:
+          "A volume with this name already exist in the target project and storage pool",
+      });
```
edlerd, line 119: Usually we do this:
```
      .test("deduplicate", "A profile with this name already exists", (value) =>
        checkDuplicateName(value, project, controllerState, "profiles"),
      )
```
Not sure why this looks more complex?

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+        .catch((e) => {
+          toastNotify.failure(
+            "Volume duplication failed.",
+            e,
+            existingVolumeLink,
+          );
+        })
+        .finally(() => {
+          close();
+        });
```
edlerd, line 190:
```suggestion
```
We should close in the `then` branch to allow simple retry (without needing to specify all details again) in the error case.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+  const disableClusterMemberSelector =
+    formik.values.project !== volume.project
+      ? "Can not migrate cluster member when changing project"
+      : undefined;
+
+  const disableProjectSelector =
+    formik.values.member !== volume.location
+      ? "Can not change project when migrating cluster member"
+      : undefined;
```
edlerd, line 202: This doesn't work in the unclustered case, because it disables it there as well.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+          label="Target project"
+          options={projects.map((project) => {
+            return {
+              label: project.name,
+              value: project.name,
+            };
+          })}
+          disabled={!!disableProjectSelector}
+          help={disableProjectSelector}
+        />
```
edlerd, line 265: Shall we remove the target project selector to keep this form a bit simpler? Especially, the exclusive disable logic is hard to gasp for the user.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+        <Input
+          {...formik.getFieldProps("volumeOnly")}
+          type="checkbox"
+          label="Copy without volume snapshots"
```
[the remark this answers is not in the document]
mas-who: sounds good :+1:

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+    if (volumeNames.includes(newVolumeName)) {
+      let count = 1;
+      while (volumeNames.includes(`${newVolumeName}-${count}`)) {
+        count++;
+      }
+      return `${newVolumeName}-${count}`;
+    }
+
+    return newVolumeName;
+  };
```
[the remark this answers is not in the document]
mas-who: created a helper `getUniqueResourceName` which extracts the duplicate checking logic.

src/pages/storage/forms/DuplicateVolumeForm.tsx
mas-who: the example you've shown checks duplicate by making api call to a fixed project. In the current scenario, the target project and storage pool may change depending on the form inputs. This means we need to check duplication against the adjusted `project` and `pool` formik fields.

src/pages/storage/forms/DuplicateVolumeForm.tsx
mas-who: aah yes, some legacy code got copied here.

src/pages/storage/forms/DuplicateVolumeForm.tsx
mas-who: good catch, the logic should be `isClustered && formik.values.member !== volume.location`

src/pages/storage/forms/DuplicateVolumeForm.tsx
mas-who: sure I'm fine with that, let's then add a comment here with reference to the upstream bug issue so we can have a reminder to resolve this in the future.

mas-who: actually, on second thought, I think duplication to a different project could be a common use case. Also the project selector is available for instance duplication as well, I think we should keep the interaction somewhat consistent?  Maybe we only remove the project selector for the clustered case?

src/api/storage-pools.tsx
```diff
@@ -381,3 +381,38 @@ export const migrateStorageVolume = (
...
+  volume: Partial<LxdStorageVolume>,
+  pool: string,
+  project?: string,
+  target?: string,
+): Promise<LxdOperationResponse> => {
+  return new Promise((resolve, reject) => {
+    const url = new URL(
+      `/1.0/storage-pools/${pool}/volumes/custom`,
+      window.location.origin,
+    );
```
edlerd, line 397:
```suggestion
    const url = new URL(`/1.0/storage-pools/${pool}/volumes/custom`);
```
I think this is not needed.

edlerd (commented):
This doesn't work yet, I keep getting this error:

image

src/api/storage-pools.tsx
mas-who: Would you be able to give me a reproducer for the error you are seeing? I can replicate it locally or on the demo server. 

Edit: if I removed `window.location.origin` from the `URL` construction, then I see the above error :thinking:

edlerd: Doh, my bad. then all good as it is. Sorry for the confusion.

mas-who: all good!

src/pages/storage/forms/DuplicateVolumeForm.tsx
edlerd: Let's keep it in and simplify once the bug is fixed upstream

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,270 @@
...
+          type="text"
+          label="New volume name"
+          error={formik.touched.name ? formik.errors.name : null}
+        />
+        <ClusterMemberSelector
+          {...formik.getFieldProps("member")}
+          id="member"
+          label="Target cluster member"
+          disableReason={disableClusterMemberSelector}
+        />
```
edlerd, line 231: Depending on the storage pool driver the volume lives on, we should not show this selector. I.e. for a ceph backed storage, the cluster member has no effect.

I suspect the volume.location to be a good indicator -- if it is not an empty string, then we should show the selector.

edlerd: There is a related bug on the storage volume overview, if the volume.location is empty, we should render a "-" instead of nothing.

mas-who: This is a great observation! I think `volume.location` is likely a good indicator. Although, I'd like to do some additional testing by setting up a ceph pool, will report back.

edlerd: Yes it is. I encountered it on testing with a ceph pool on my local microcloud in VMs.

mas-who: Aaah okay I see, thanks for testing, then I think let's hide the selector if `volume.location` is not set for the existing volume, and the api query we send will only have the `project` param.

mas-who: @edlerd played around with this a little, I think we shouldn't hide the cluster member selector. In the case if we want to duplicate the volume to a non-remote storage pool, cluster member selection remains valid. Instead, I think we should disable the cluster member selector depending on the target pool selected. According to the [documentation](https://documentation.ubuntu.com/lxd/en/latest/explanation/storage/), we can determine if a pool is remote depending on the driver type i.e. all `ceph` types and `powerflex`. wdyt?

edlerd: Sounds good to me.

mas-who: great, that's already implemented :) let me know if you see anything wrong with it

edlerd: Should we then completely remove the cluster member selector from the duplicate form?

mas-who: yeah that's probably better, simpler :+1: . I don't think cluster location for a custom volume is as important as storage pool or project anyway :thinking:

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+            "Volume duplication failed.",
+            e,
+            existingVolumeLink,
+          );
+        });
+    },
+  });
+
+  // TODO: currently we don't support target cluster member selection due to a bug in lxd, we should enable it once the bug is fixed
+  // see issue:https://github.com/canonical/lxd/issues/14155
```
edlerd, line 173:
```suggestion
  // TODO: currently we don't support target cluster member selection due to a bug in lxd,
  // we should enable it once the bug is fixed. Note, that cluster member selection is not relevant
  // for remote storage pools, as they are not tied to a lxd cluster member.
  // see issue: https://github.com/canonical/lxd/issues/14155
```
Adding a bit more detail

edlerd (approved):
QA and code LGTM, one note on the comment to include the info we discovered in this discussion. Maybe that to-do should be a ticket and not included in the PR?

src/pages/storage/forms/DuplicateVolumeForm.tsx
mas-who: I removed the comment and created an [issue](https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/930?assignee=712020%3Accb7b61f-3f6a-41ec-9935-a79519dada4e&selectedIssue=WD-15524) for it.

edlerd (approved):

PR #927, opened 2024-09-26
chore: retry lazy load components when server is down [WD-13909]

mas-who:
## Done

- support retry lazy loading modules with increasing delay
- add more info for dynamic module loading error in the Error page

## QA

1. Run the LXD-UI:
    - Load LXD-UI initially, don't navigate anywhere so that the pages are not cached
    - Open dev tool, go to the Network tab and select Offline as network condition
    - Go to a different page in the UI, it may not be very noticeable but it will now take a little longer for the error page to show because we will now retry fetch the module a few times before throwing an error (this will be more noticeable if you increase the delayTime input for the `lazyWithRetry` function.
    - See that the error message includes details about lxd server being down if it's a dynamic module loading error

## Screenshots
Screenshot from 2024-09-26 10-27-29

### DIFF
--- a/src/components/ErrorPage.tsx
+++ b/src/components/ErrorPage.tsx
@@ -40,6 +40,35 @@ ${error?.stack ?? "No stack trace"}
   useEffect(updateHeight, []);
   useEventListener("resize", updateHeight);
 
+  const errorBlocks = [];
+  if (error?.message) {
+    errorBlocks.push({
+      title: "Error",
+      appearance: CodeSnippetBlockAppearance.NUMBERED,
+      wrapLines: true,
+      code: error.message,
+    });
+  }
+
+  if (error?.message.toLowerCase().includes("dynamically imported module")) {
+    errorBlocks.push({
+      title: "Possible causes",
+      appearance: CodeSnippetBlockAppearance.NUMBERED,
+      wrapLines: true,
+      code: `This might be due to a temporary network issue. Please try refreshing the page.
+If the problem continues, ensure your connection to the LXD server is active or try again later.`,
+    });
+  }
+
+  if (error?.stack) {
+    errorBlocks.push({
+      title: "Stack trace",
+      appearance: CodeSnippetBlockAppearance.NUMBERED,
+      wrapLines: true,
+      code: error.stack,
+    });
+  }
+
   return (
     <Strip className="u-no-padding--bottom">
       <Notification severity="negative" title="Error">
@@ -50,28 +79,7 @@ ${error?.stack ?? "No stack trace"}
       </Notification>
       <CodeSnippet
         className="error-info u-no-margin--bottom"
-        blocks={[
-          ...(error?.message
-            ? [
-                {
-                  title: "Error",
-                  appearance: CodeSnippetBlockAppearance.NUMBERED,
-                  wrapLines: true,
-                  code: error.message,
-                },
-              ]
-            : []),
-          ...(error?.stack
-            ? [
-                {
-                  title: "Stack trace",
-                  appearance: CodeSnippetBlockAppearance.NUMBERED,
-                  wrapLines: true,
-                  code: error.stack,
-                },
-              ]
-            : []),
-        ]}
+        blocks={errorBlocks}
       />
     </Strip>
   );
--- a/src/util/lazyWithRetry.tsx
+++ b/src/util/lazyWithRetry.tsx
@@ -1,0 +1,63 @@
+import { lazy } from "react";
+import { delay } from "./helpers";
+
+const RETRIES = 3;
+const DELAY_TIME = 250;
+
+// This function is a wrapper around React.lazy that will retry the import
+// keeping the function type signature the same as React.lazy
+// using "any" type here allows for components with any type interface to be imported
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const lazyWithRetry = <T extends React.ComponentType<any>>(
+  importFunction: () => Promise<{ default: T }>,
+) => {
+  const retryImport = async () => {
+    try {
+      const module = await importFunction();
+      return module;
+    } catch (error) {
+      let lastError = error as Error;
+      for (let attempt = 0; attempt < RETRIES; attempt++) {
+        await delay(DELAY_TIME * attempt);
+
+        // dynamic import i.e. import() unfortunately caches network response for fetching modules
+        // this is true even if a module fails to load
+        // if we simply just retry the import, the cached error response will be returned
+        // to work around this, we can invalidate the cache by adding a dynamic query parameter to the module URL
+        // the original module url can be potentially found in the error message
+        // HOWEVER, this is not a guaranteed solution since the error message may not contain the module URL depending on the browser used
+        // therefore, this is a best effort attempt, worst case scenario, the error will be thrown again causing the error boundary to render
+        const origin = window.location.origin;
+        const errorSegments = (error as Error).message.split(" ");
+        let moduleUrl = "";
+        for (const segment of errorSegments) {
+          if (segment.startsWith(origin)) {
+            moduleUrl = segment;
+            break;
+          }
+        }
+
+        const url = new URL(moduleUrl.trim());
+        // add a timestamp query parameter to invalidate the cache
+        url.searchParams.set("cacheBuster", `${+Date.now()}`);
+
+        try {
+          // vite does not support static checks with dynamic imports that takes in a non-relative path
+          // ref: https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations
+          const module = (await import(/* @vite-ignore */ url.href)) as {
+            default: T;
+          };
+          return module;
+        } catch (e) {
+          lastError = e as Error;
+        }
+      }
+
+      throw lastError;
+    }
+  };
+
+  return lazy(retryImport);
+};
+
+export default lazyWithRetry;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { FC, lazy, Suspense } from "react";
+import { FC, Suspense } from "react";
 import { Navigate, Route, Routes } from "react-router-dom";
 import Loader from "components/Loader";
 import ProjectRedirect from "pages/projects/ProjectRedirect";
@@ -10,6 +10,7 @@ import CustomLayout from "components/CustomLayout";
 import NoMatch from "components/NoMatch";
 import { logout } from "util/helpers";
 import NoProject from "components/NoProject";
+import lazy from "util/lazyWithRetry";
 
 const CertificateAdd = lazy(() => import("pages/login/CertificateAdd"));
 const CertificateGenerate = lazy(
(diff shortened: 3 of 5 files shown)


src/util/lazyWithRetry.tsx
```diff
@@ -0,0 +1,32 @@
+import { lazy } from "react";
+import { delay } from "./helpers";
+
+// This function is a wrapper around React.lazy that will retry the import
+// keeping the function type signature the same as React.lazy
+// using "any" type here allows for components with any type interface to be imported
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const lazyWithRetry = <T extends React.ComponentType<any>>(
+  importFunction: () => Promise<{ default: T }>,
+  retries = 3,
```
mas-who, line 10: @edlerd @Kxiru open to suggestions for reasonable retries and delay time. Note `delayTime` here essentially is the ceiling of how long we would wait to retry import again (for the final attempt)

src/util/lazyWithRetry.tsx
```diff
@@ -0,0 +1,32 @@
...
+      try {
+        const module = await importFunction();
+        return module;
+      } catch (error) {
+        // set err here so we can use it outside of the while loop context
+        err = error;
+      }
+
+      retries--;
+      await delay(delayTime / retries);
```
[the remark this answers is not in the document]
mas-who: retries is decreasing so multiplication would decrease the delay

src/util/lazyWithRetry.tsx
```diff
@@ -0,0 +1,32 @@
...
+// using "any" type here allows for components with any type interface to be imported
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const lazyWithRetry = <T extends React.ComponentType<any>>(
+  importFunction: () => Promise<{ default: T }>,
+  retries = 3,
+  delayTime = 500,
+) => {
+  return lazy<T>(async () => {
+    let err: unknown;
+    while (retries > 0) {
```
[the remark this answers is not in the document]
mas-who: aah okay, makes sense, please disregard the earlier comment for decreasing delay

src/components/ErrorPage.tsx
```diff
@@ -40,6 +40,25 @@ ${error?.stack ?? "No stack trace"}
...
+      modifiedError = new Error(
+        `${error.message}. This is likely due to a network issue or the lxd server being unreachable.`,
+      );
+      modifiedError.stack = error.stack;
+      modifiedError.name = error.name;
+    }
+  }
+
+  const displayError = modifiedError || error;
+
```
edlerd, line 61: Do we really need to modify the error? I think we can modify the rendering of the error message below and append the text there, which seems much simpler and more straight forward.

src/util/lazyWithRetry.tsx
```diff
@@ -0,0 +1,31 @@
...
+) => {
+  return lazy<T>(async () => {
+    let err: unknown;
+    for (let attempt = 0; attempt < retries; attempt++) {
+      try {
+        const module = await importFunction();
+        return module;
+      } catch (error) {
+        // set err here so we can use it outside of the loop context
+        err = error;
```
edlerd, line 21:
```suggestion
if (attempt === retries -1) {
  throw error;
}
```
Maybe simpler? This way we don't need another reference for the error and after the last error we avoid extra call to delay.

edlerd: Then let's keep it as is. I think it is actually easier to read than my suggestion.

edlerd (changes requested):
This seems to not work as expected. I blocked one of the js files manually with chrome dev tools and see only one request for the resource, where I would expect 3 network requests due to the retry.

edlerd:
image

edlerd:
> Edit: @edlerd so I pushed an approach, I would like to know what you think about it. Due to `React.lazy` caching success or failed promises for module import network calls, we need to invalidate the cache when retrying an import. To do this, we can try extract the module url from the error message and add a dynamic query param to force cache invalidation. This unfortunately is a best effort approach, since the error message may or may not contain the module url depending on the browser. However, I think this is acceptable since worst case scenario the error boundary will render. See this related [issue](https://github.com/facebook/react/issues/14254) regarding `React.lazy`, there are some approaches suggested which I think all look not so great. TBH, I'm also happy to not introduce this complexity, and maybe we just improve the error page rendered text? wdyt?

If we have a way "auto-heal" I think it is worth the complexity as it is contained within on file. I think there is a bit simpler approach using recursion as described in [this blog](https://dev.to/goenning/how-to-retry-when-react-lazy-fails-mb5). Not sure if this one still works, but maybe we can adopt it here?

mas-who:
> If we have a way "auto-heal" I think it is worth the complexity as it is contained within on file. I think there is a bit simpler approach using recursion as described in [this blog](https://dev.to/goenning/how-to-retry-when-react-lazy-fails-mb5). Not sure if this one still works, but maybe we can adopt it here?

Thanks for the resource :) I did see this one and tried it, most of the articles I've seen have similar recursive or iterative implementations. The problem is really that a dynamic import would cache successful or failed network requests i.e. `import(<module>)` would return a promise to `React.lazy`, if it failed, then future `import` calls for the same module, within the same browser session would fail because the cached failure response. Here is a good [reference](https://github.com/whatwg/html/issues/6768?source=post_page-----c7755a7d557a--------------------------------) about this issue, also I found the workaround with the query params there. The current approach in this PR works on chrome and firefox, but I'm not sure about safari. afaik, the only reliable way to get around the dynamic import cache issue is to do a `window.location.reload`, but I think this does not result in a good user experience.

edlerd:
> Thanks for the resource :) I did see this one and tried it, most of the articles I've seen have similar recursive or iterative implementations. The problem is really that a dynamic import would cache successful or failed network requests i.e. `import(<module>)` would return a promise to `React.lazy`, if it failed, then future `import` calls for the same module, within the same browser session would fail because the cached failure response. Here is a good [reference](https://github.com/whatwg/html/issues/6768?source=post_page-----c7755a7d557a--------------------------------) about this issue, also I found the workaround with the query params there. The current approach in this PR works on chrome and firefox, but I'm not sure about safari. afaik, the only reliable way to get around the dynamic import cache issue is to do a `window.location.reload`, but I think this does not result in a good user experience.

Yes, reload should be avoided. I suspected the recursive approach might avoid the caching behaviour.

mas-who:
> > Thanks for the resource :) I did see this one and tried it, most of the articles I've seen have similar recursive or iterative implementations. The problem is really that a dynamic import would cache successful or failed network requests i.e. `import(<module>)` would return a promise to `React.lazy`, if it failed, then future `import` calls for the same module, within the same browser session would fail because the cached failure response. Here is a good [reference](https://github.com/whatwg/html/issues/6768?source=post_page-----c7755a7d557a--------------------------------) about this issue, also I found the workaround with the query params there. The current approach in this PR works on chrome and firefox, but I'm not sure about safari. afaik, the only reliable way to get around the dynamic import cache issue is to do a `window.location.reload`, but I think this does not result in a good user experience.
> 
> Yes, reload should be avoided. I suspected the recursive approach might avoid the caching behaviour.

Unfortunately not, caching happens within `import()`, so recursion or iterating the retries does not make a difference. Either than `window.location.reload`, I think other methods will be "best effort" approaches as this is a dynamic import behaviour that affects all UI frameworks, particularly SPAs I think since poor network connectivity affects chunk loading and there is no reliable way to heal

src/components/ErrorPage.tsx
```diff
@@ -40,6 +40,35 @@ ${error?.stack ?? "No stack trace"}
...
+    });
+  }
+
+  if (error?.message.toLowerCase().includes("dynamically imported module")) {
+    errorBlocks.push({
+      title: "Possible causes",
+      appearance: CodeSnippetBlockAppearance.NUMBERED,
+      wrapLines: true,
+      code: `There might be a network issue preventing access to the LXD server.
+Please verify if the LXD server is operational and accessible.`,
```
edlerd, line 59: I think the easiest resolution in this case is to refresh the page -- they initially loaded the UI, so it should just work. So don't ask them to verify, but to refresh.

src/util/lazyWithRetry.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { delay } from "./helpers";
+
+// This function is a wrapper around React.lazy that will retry the import
+// keeping the function type signature the same as React.lazy
+// using "any" type here allows for components with any type interface to be imported
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const lazyWithRetry = <T extends React.ComponentType<any>>(
+  importFunction: () => Promise<{ default: T }>,
+  retries = 3,
+  delayTime = 1000,
```
edlerd, line 11:
```suggestion
  delayTime = 250,
```
I don't think we should set it this high.
Also I don't think the delay time and the retires should be configurable. We can just define constants for it.

src/util/lazyWithRetry.tsx
```diff
@@ -0,0 +1,62 @@
...
+        await delay(delayTime * attempt);
+
+        // React.lazy unfortunately caches network response for fetching modules
+        // this is true even if a module fails to load
+        // if we simply just retry the import, React.lazy will return the same error response from its cache
+        // to work around this, we can invalidate the cache by adding a dynamic query parameter to the module URL
+        // the original module url can be potentially found in the error message
+        // HOWEVER, this is not a guaranteed solution since the error message may not contain the module URL depending on the browser used
+        // therefore, this is a best effort attempt, worst case scenario, the error will be thrown again causing the error boundary to render
+        const protocol = window.location.protocol;
```
edlerd, line 29: Probably safer to use the `window.location.origin`, as the protocol might be used in the error message itself. The origin is at least a bit more specific.

src/util/lazyWithRetry.tsx
```diff
@@ -0,0 +1,62 @@
...
+        for (const segment of errorSegments) {
+          if (segment.startsWith(protocol)) {
+            moduleUrl = segment;
+            break;
+          }
+        }
+
+        const url = new URL(moduleUrl.trim());
+        // add a timestamp query parameter to invalidate the cache
+        url.searchParams.set("t", `${+new Date()}`);
```
edlerd, line 41: better use `Date.now()`, as it has higher precision so and easier format.

Also I would call the parameter `cacheBuster`, not `t`.

edlerd (commented):
QA LGTM, nice approach :)

Some comments on the code and copy. should be very easy to address and then this should be good to merge.

src/components/ErrorPage.tsx
mas-who: Sounds good, but I think it would still be good to include the copy for checking lxd connection since this is a problem that miona have ran into before. So we can have something like:

```
This might be due to a temporary network issue. Please try refreshing the page.
If the problem continues, ensure your connection to the LXD server is active or try again later.
```

src/util/lazyWithRetry.tsx
mas-who: thanks for the catch, the delay time was set a little high so it was easy to test if the page would load again if I turn on the network. I've made the above changes now :)

src/util/lazyWithRetry.tsx
mas-who: nice! done :+1:

edlerd (approved):
LGTM

fatso83:
I might add that you can drop a bit of code by reusing this: https://www.npmjs.com/package/@fatso83/retry-dynamic-import

PR #929, opened 2024-09-26
test: add e2e test for external instance file [WD-14777]

mas-who:
## Done

- add e2e for uploading external format instance file
- fixed a bug where uploading external instance file to a lxd cluster fails

## QA

1. Run the LXD-UI:
    - Ensure CI passes

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -131,10 +131,11 @@ jobs:
           sudo lxc cluster enable local
           sudo lxc config set user.show_permissions=true
 
-      - name: Create a custom image
+      - name: Create and download custom images
         shell: bash
         run: |
           set -x
+          curl -L -o ubuntu-minimal.qcow2 https://cloud-images.ubuntu.com/minimal/releases/jammy/release/ubuntu-22.04-minimal-cloudimg-amd64.img &
           sudo lxc launch ubuntu-minimal:22.04 my-instance
           sudo lxc publish my-instance --alias my-custom-image --force
           sudo lxc delete my-instance --force
--- a/tests/instances.spec.ts
+++ b/tests/instances.spec.ts
@@ -351,3 +351,35 @@ test("Export and Upload an instance backup", async ({ page }) => {
   await page.waitForSelector(`text=Created instance ${instance}-1`);
   await deleteInstance(page, `${instance}-1`);
 });
+
+test("Create instance from external instance file", async ({
+  page,
+  lxdVersion,
+}) => {
+  test.skip(
+    lxdVersion === "5.0-edge",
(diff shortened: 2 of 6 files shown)


mas-who:
@edlerd this one is ready for review :) picked up a bug for uploading an external instance file to a lxd cluster, that's also fixed in this PR

tests/instances.spec.ts
```diff
@@ -351,3 +351,36 @@ test("Export and Upload an instance backup", async ({ page }) => {
...
+
+  const INSTANCE_FILE = "ubuntu-minimal.qcow2";
+  await page.goto("/ui/");
+  await page.getByRole("button", { name: "Create instance" }).click();
+  await page.getByRole("button", { name: "Upload instance file" }).click();
+  await page.getByText("External format (.qcow2, .").click();
+  await page
+    .getByRole("textbox", { name: "External format (.qcow2, ." })
+    .setInputFiles(INSTANCE_FILE);
+  const instanceName = "ubuntu-minimal-import";
```
edlerd, line 377: This should be randomized to avoid failures on re-running this test and to allow parallel runs.

mas-who: This tests that the instance name is automatically set to the file name + `-import`. The file is downloaded in CI before the e2e tests and have a fixed name, so I don't think the test will fail?

edlerd: My point is: the instance should have a randomized name. Otherwise, retrying this test, if it fails in CI, might cause trouble. I.e. if a test run times out and doesn't finish deleting the instance, then the retry will fail, because it can't create a 2nd instance with the same name.

mas-who: I see, will update the test to input random instance name instead

.github/workflows/pr.yaml
```diff
@@ -131,10 +131,11 @@ jobs:
           sudo lxc cluster enable local
           sudo lxc config set user.show_permissions=true
 
-      - name: Create a custom image
+      - name: Create and download custom images
         shell: bash
         run: |
           set -x
+          curl -L -o ubuntu-minimal.qcow2 https://cloud-images.ubuntu.com/minimal/releases/jammy/release/ubuntu-22.04-minimal-cloudimg-amd64.img &
```
edlerd, line 138: this also needs to be added to coverage.yaml

edlerd (changes requested):

.github/workflows/pr.yaml
mas-who: yes, thanks for catching that :+1: I wonder if it's useful to create reusable actions for these jobs that's almost identical so that it's easier to maintain in the future?

edlerd (approved):
LGTM, thanks for adding the test and making it nice :)

PR #932, opened 2024-09-26
feat: [WD-15076] Instance Overview add devices

Kxiru:
## Done
Work Done:
- Added Pool to Instance Overview page
- Added Pool to Instance Detail Panel Content
- Added PID to Instance Detail Panel Content
- Added Devices to Instance Overview Page

## QA

1. Run the LXD-UI:
    - Navigate to an Instance and observe the changes in it's Overview and side panel.

## Screenshots

image


src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -85,6 +89,14 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
             {settings?.environment?.server_clustered ? instance.location : "-"}
           </td>
         </tr>
+        <tr>
+          <th className="u-text--muted">Pool</th>
+          <td>{(instance.expanded_devices?.root as LxdDiskDevice).pool}</td>
```
edlerd, line 94: Can we link to the storage pool detail page?

src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -85,6 +89,14 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
             {settings?.environment?.server_clustered ? instance.location : "-"}
           </td>
         </tr>
+        <tr>
+          <th className="u-text--muted">Pool</th>
```
edlerd, line 93: I think we need to be more specific
```suggestion
          <th className="u-text--muted">Root storage pool</th>
```

src/pages/instances/InstanceOverview.tsx
```diff
@@ -85,6 +87,12 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
                     : "-"}
                 </td>
               </tr>
+              <tr>
+                <th className="u-text--muted">Pool</th>
```
edlerd, line 91:
```suggestion
                <th className="u-text--muted">Root storage pool</th>
```

src/pages/instances/InstanceOverview.tsx
```diff
@@ -85,6 +87,12 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
...
+                <td>
+                  {(instance.expanded_devices?.root as LxdDiskDevice).pool}
```
edlerd, line 93: Add the link here as well.

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -0,0 +1,67 @@
...
+} from "types/device";
+
+interface Props {
+  instance: LxdInstance;
+}
+type InstanceOverviewDevices =
+  | LxdNoneDevice
+  | LxdGPUDevice
+  | LxdOtherDevice
+  | LxdProxyDevice;
```
edlerd, line 18:
```suggestion
```
I don't think we need a new type here (see below). I think we should not use the same name for a component and a type.

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
+  const instanceDevices = Object.entries(
+    instance.expanded_devices ?? {},
+  ).filter(([, device]) => {
+    return !(device.type === "nic" || device.type === "disk");
+  });
```
edlerd, line 25: I think this should be in a helper function `getDevices(instance)` that returns an `LxdDeviceValue[]`. That way we don't need the cast as `const device = dev as InstanceOverviewDevices` below and contain the cast to a function.

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -0,0 +1,67 @@
...
+      },
+    };
+  });
+
+  return (
+    <>
+      {hasDevices && (
+        <MainTable headers={deviceHeaders} rows={deviceRows} sortable />
+      )}
+      {!hasDevices && <>No devices available</>}
```
edlerd, line 62:
```suggestion
      {!hasDevices && <>-</>}
```
I think this is too verbose, we can have the same info with much less copy. This is also better in line how we deal with no networks.

src/sass/_instance_detail_overview.scss
```diff
@@ -20,6 +20,16 @@
...
+  .general,
+  .usage,
+  .devices {
+    border-bottom: 1px solid $color-mid-light;
+
+    table {
+      margin-bottom: $spv--large;
+    }
+  }
+
```
edlerd, line 32: We should not duplicate this from above?

src/sass/_instance_detail_overview.scss
```diff
@@ -54,6 +64,17 @@
...
+    th:nth-child(1) {
+      width: 30%;
+    }
+
+    th:first-of-type,
+    td:first-of-type {
+      padding-left: 0;
+    }
+  }
+
```
edlerd, line 77: We can reuse the definition for profiles below, no need to duplicate it.

edlerd (changes requested):

src/pages/instances/InstanceOverviewDevices.tsx
Kxiru: I understand not using the same name for the Component and type here, but removing the new types gives me the error that I was dealing with before. I cannot simply use the LxdDeviceValue type as we get another case of LxdISoDevice not having a name attribute, so I created a new type for all the devices but Iso and disks.

src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -85,6 +89,20 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
             {settings?.environment?.server_clustered ? instance.location : "-"}
           </td>
         </tr>
+        <tr>
+          <th className="u-text--muted">Root storage pool</th>
+          <td>
+            <Link
+              to={`/ui/project/${instance.project}/storage/pool/${(instance.expanded_devices?.root as LxdDiskDevice).pool}`}
+            >
+              {(instance.expanded_devices?.root as LxdDiskDevice).pool}
```
edlerd, line 98: Better assign a variable `const rootPool = ...` and reuse it, like we do for the pid.

Also the assumption here, that the root disk always has the name `root` is not correct. We should add a function to pick the root device by filtering all devices by the ones that are
1. type=disk
2. path=/
3. source is not set

Maybe put this into a function, so we can use it on the overview tab and in the side panel.

src/util/helpers.tsx
```diff
@@ -298,3 +299,15 @@ export const getFileExtension = (filename: string): string => {
...
   return `.${filename.split(".").pop()}` || "";
 };
+
+export const getDevices = (
+  instance: LxdInstance,
+): [string, LxdDeviceValue][] => {
+  const instanceDevices = Object.entries(
+    instance.expanded_devices ?? {},
+  ).filter(([, device]) => {
+    return !(device.type === "nic" || device.type === "disk");
```
edlerd, line 309:
```suggestion
    return (device.type !== "nic" && device.type !== "disk");
```
I think this is easier to read.

src/util/helpers.tsx
```diff
@@ -298,3 +299,15 @@ export const getFileExtension = (filename: string): string => {
 
   return `.${filename.split(".").pop()}` || "";
 };
+
+export const getDevices = (
```
edlerd, line 303: The name is misleading. Maybe return all devices here, and move the filter to the use case?

So the funciton body here becomes
```
  return Object.entries(
    instance.expanded_devices ?? {},
  );
```

src/pages/instances/InstanceOverviewDevices.tsx
edlerd: I think with the current code, we don't need to access device.name, but we only rely on devicename. So this should be fine?

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -0,0 +1,58 @@
+import { FC } from "react";
+import { MainTable } from "@canonical/react-components";
+import { LxdInstance } from "types/instance";
+import {
+  LxdGPUDevice,
+  LxdNoneDevice,
+  LxdOtherDevice,
+  LxdProxyDevice,
+} from "types/device";
```
edlerd, line 9:
```suggestion
```
seems unused

edlerd (commented):
Thanks for the iteration, some pointers to further improve this below.

src/pages/instances/InstanceOverviewDevices.tsx
Kxiru: Ahhh, I see. I understand where you are coming from now, thank you for clarifying.

src/pages/instances/InstanceDetailPanelContent.tsx
Kxiru: I understand now. I have extrapolated this into a function now, please let me know what you think :)

src/util/helpers.tsx
```diff
@@ -301,3 +302,21 @@ export const getFileExtension = (filename: string): string => {
...
 export const delay = (ms: number): Promise<void> =>
   new Promise((resolve) => setTimeout(resolve, ms));
+
+export const returnAllDevices = (
+  instance: LxdInstance,
+): [string, LxdDeviceValue][] => {
+  const instanceDevices = Object.entries(instance.expanded_devices ?? {});
+
+  return instanceDevices;
+};
```
edlerd, line 312:
```suggestion
export const getInstanceDevices = (
  instance: LxdInstance,
): [string, LxdDeviceValue][] => {
  return Object.entries(instance.expanded_devices ?? {});
};
```
Better naming and a bit simpler?

src/util/helpers.tsx
```diff
@@ -301,3 +302,21 @@ export const getFileExtension = (filename: string): string => {
...
+
+export const getRootPool = (instance: LxdInstance): string => {
+  const allDevices = Object.entries(instance.expanded_devices ?? {}).filter(
+    ([, device]) => {
+      return device.type == "disk" && (device as LxdDiskDevice).path == "/";
+    },
+  );
+  const rootPool = allDevices[0];
+  return (rootPool[1] as LxdDiskDevice).pool;
+};
```
edlerd, line 322:
```suggestion
export const getRootPool = (instance: LxdInstance): string => {
  const rootStorage = Object.values(instance.expanded_devices ?? {})
    .filter(isDiskDevice)
    .find((device) => {
      return isRootDisk(device as FormDevice);
    });

  return rootStorage ? rootStorage.pool : "";
};
```
A bit safer and a bit easier?

edlerd: There is some code in `InstanceStoragePoolMigration` that is doing this very calculation, I think we should use this function instead of the local `rootDiskDevice` variable in it. Do you mind simplifying that bit as well?

edlerd (commented):
QA LGTM, two tiny ideas to simplify the code, and there is one usage we can simplify with the new generic function you introduce here.

src/util/helpers.tsx
Kxiru: I had a look at InstanceStoragePoolMigration and switched it out! Let me know what you think :)

edlerd (approved):
LGTM, good work :+1:

PR #933, opened 2024-09-29
Fix various UI bugs

edlerd:
## Done

-  fix(image) avoid duplicate loading indicator
- fix(profile) show edit icon next to root storage pool on profile edit configuration
- fix(instance) reset nat for proxy device configuration, in case bind is set to instance
- fix(instance) avoid displaying root storage size as overridden, without an override present
- fix(operations) ensure operations cache is cleared to show correct amount of running operations in footer
- fix(storage) remove redundant disabled field on storage pool selector, that overrides the already used, nested disabled flag
- fix(instance) avoid crash on filtering only cached images with no cached images

## QA

1. Run the LXD-UI:
    - upload image and ensure 1 loading indicator
    - ensure edit icon next to root storage in configuration edit for profile, but not for instances
    - add a proxy device to instance configuration, set nat to enabled or disabled and set bind to instance, ensure the nat value is unset
    - check instance config that has no root storage override, ensure the size is not shown as overridden in configuration > disk
    - ensure the number of running operations in the footer is updating correctly for any action that triggers an operation -- also on operation finish the number of running operations should decrease
    - ensure storage pool is disabled on opening the upload custom iso flow
    - ensure storage pool is disabled on opening the upload instance flow with internal or external format
    - ensure checking the "only show cached images" filter is not crashing the interface, when no cached images are present

### DIFF
--- a/src/components/forms/ProxyDeviceForm.tsx
+++ b/src/components/forms/ProxyDeviceForm.tsx
@@ -250,7 +250,7 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
         device.bind,
         "Whether to bind the listen address to the instance or host",
         (value) => {
-          if (value === "instance" && device.nat !== undefined) {
+          if (value === "instance") {
             void formik.setFieldValue(`devices.${index}.nat`, undefined);
           }
         },
--- a/src/components/forms/ProxyDeviceForm.tsx
+++ b/src/components/forms/ProxyDeviceForm.tsx
@@ -250,8 +250,8 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
         device.bind,
         "Whether to bind the listen address to the instance or host",
         (value) => {
-          if (value === "instance" && device.nat === "true") {
-            void formik.setFieldValue(`devices.${index}.nat`, "false");
+          if (value === "instance" && device.nat !== undefined) {
+            void formik.setFieldValue(`devices.${index}.nat`, undefined);
           }
         },
       ),
--- a/src/context/eventQueue.tsx
+++ b/src/context/eventQueue.tsx
@@ -1,5 +1,7 @@
 import { createContext, FC, ReactNode, useContext } from "react";
 import { LxdEvent } from "types/event";
+import { useQueryClient } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
 
 export interface EventQueue {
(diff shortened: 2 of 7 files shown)


src/components/forms/ProxyDeviceForm.tsx
```diff
@@ -250,8 +250,8 @@ const ProxyDeviceForm: FC<Props> = ({ formik, project }) => {
         device.bind,
         "Whether to bind the listen address to the instance or host",
         (value) => {
-          if (value === "instance" && device.nat === "true") {
-            void formik.setFieldValue(`devices.${index}.nat`, "false");
+          if (value === "instance" && device.nat !== undefined) {
```
mas-who, line 253: I think `device.nat !== undefined` check is not necessary. We also wouldn't have to worry about unnecessary re-renders since React checks if state value has changed before re-rendering (for primitive values).

src/context/eventQueue.tsx
```diff
@@ -31,14 +33,24 @@ interface EventCallback {
...
 export const EventQueueProvider: FC<Props> = ({ children }) => {
+  const queryClient = useQueryClient();
+
   return (
     <EventQueueContext.Provider
       value={{
         get: (operationId) => eventQueue.get(operationId),
         set: (operationId, onSuccess, onFailure, onFinish) => {
           eventQueue.set(operationId, { onSuccess, onFailure, onFinish });
+          void queryClient.invalidateQueries({
```
mas-who, line 44: good catch! I see this was previously affecting instance creation operations from file uploads

src/components/forms/ProxyDeviceForm.tsx
edlerd: Good point, removing the extra check.

src/pages/images/ImageSelector.tsx
```diff
@@ -184,6 +180,12 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
         item.release.toLowerCase().includes(query)
       );
     })
+    .filter((item) => {
+      if (!hideRemote) {
+        return true;
+      }
+      return item.server === LOCAL_IMAGE;
```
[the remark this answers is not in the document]
edlerd: right, adding the check here.

mas-who:
LGTM, thanks for the fixes :)

mas-who (approved):

PR #935, opened 2024-10-01
feat(gpu) add gpu device test WD-14776

edlerd:
## Done

- add gpu device test 

Fixes WD-14776


mas-who (approved):
looks good, thanks for adding the test :+1:

PR #938, opened 2024-10-02
fix: [WD-15307] Truncate instance name in instance list (when creatin…

Kxiru:
…g instances)

## Done

- When creating images, their names are now truncated within the instance list.

Fixes: As above

## QA

1. Run the LXD-UI:
    - Attempt to create an instance and observe the name being truncated on the instance list page.

## Screenshots

image

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -258,6 +258,8 @@ const InstanceList: FC = () => {
         columns: [
           {
             content: getInstanceName(operation),
+            className: "u-truncate",
+            title: getInstanceName(operation),
             role: "rowheader",
             "aria-label": NAME,
             style: { width: `${COLUMN_WIDTHS[NAME]}px` },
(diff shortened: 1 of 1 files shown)


src/pages/instances/InstanceList.tsx
```diff
@@ -257,7 +257,14 @@ const InstanceList: FC = () => {
...
           {
-            content: getInstanceName(operation),
+            content: (
+              <div
+                className="u-truncate"
+                title={`Instance ${getInstanceName(operation)}`}
+              >
+                {getInstanceName(operation)}
+              </div>
+            ),
```
mas-who, line 267: I don't think you need to wrap the text in a div here. The column should have a `className` field that you can apply `u-truncate` to, would you please test that? Same story for `title`.

```suggestion
            content: getInstanceName(operation),
            className: "u-truncate",
            title: getInstanceName(operation),
```

mas-who (commented):
Thanks for the fix, I don't think we need the extra `div` though, please test the suggestion :)

src/pages/instances/InstanceList.tsx
Kxiru: Hey Mason,
For continuity purposes, I am using the same approach as is seen in like 342-344 where the content itself is truncated. Do you suggest I change both implementations to the one you proposed?

mas-who: Yes thanks, that would be an improvement since we are removing unnecessary html elements and therefore improving the semantic structure of our app. I'm sure there are other places in the codebase with similar issue, but we can fix them as we come across them.

mas-who (approved):

mas-who:
LGTM

PR #940, opened 2024-10-03
feat: improve permission selector [WD-15106]

mas-who:
## Done

- Improve permission resource selector
- Improve permission entitlement selector

## QA

1. Run the LXD-UI:
    - play around with the permission selector and make sure it works the same (functionally)

## Screenshots
Screenshot from 2024-10-03 21-02-02
Screenshot from 2024-10-04 18-38-37


src/components/select/CustomSelect.tsx
```diff
@@ -34,10 +35,16 @@ export type Props = PropsWithSpread<
...
     // The styling for the select toggle button
     toggleClassName?: ClassName;
+    // The styling for the select dropdown
+    dropdownClassName?: string;
     // Whether the select is searchable. Option "auto" is the default, the select will be searchable if it has 5 or more options.
     searchable?: "auto" | "always" | "never";
     // Whether to focus on the element on initial render.
     takeFocus?: boolean;
+    // Display the dropdown on the left or right side of the toggle button.
+    position?: Position;
```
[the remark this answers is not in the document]
mas-who: yes, I am doing that already below with `autoAdjust` set on `ContextualMenu`. I was thinking maybe for the `CustomSelect` component we should expose the `position` prop. Should I remove it rather?

src/util/permissions.tsx
```diff
@@ -342,3 +345,18 @@ export const enablePermissionsFeature = (): boolean => {
...
+export const resourceOptionColumns: Record<string, (keyof ResourceDetail)[]> = {
+  image: ["aliases", "fingerprint", "description", "imageType"],
+  image_alias: ["name", "project"],
+  instance: ["name", "project"],
+  network: ["name", "project"],
+  network_acl: ["name", "project"],
+  network_zone: ["name", "project"],
+  profile: ["name", "project"],
+  storage_bucket: ["name", "project"],
+  storage_volume: ["name", "project", "pool", "imageType"],
```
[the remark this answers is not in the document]
mas-who: only issue with that is image type combined with other columns here serves as a unique identifier for a image storage volume. Although the type is already embedded in the name of the volume, but it's not easily visible.

src/components/select/CustomSelect.tsx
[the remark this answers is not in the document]
edlerd: I don't see a use case where we can't rely on the auto positioning.

src/util/permissions.tsx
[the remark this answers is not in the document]
edlerd: I would expect for most of the volumes the image-type column would be empty, so probably better to remove it?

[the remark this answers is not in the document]
mas-who: makes sense, removed :+1:

piperdeck:
Link to a Figma Jam file where I've outlined the keyboard behaviour we talked about:

https://www.figma.com/board/UGpe7UDFDe2NgAxMlF61og/Permissions-dropdown-keyboard-controls?node-id=0-1&t=lLsSQVmiY4WmYmfA-1

mas-who:
> Link to a Figma Jam file where I've outlined the keyboard behaviour we talked about:
> 
> https://www.figma.com/board/UGpe7UDFDe2NgAxMlF61og/Permissions-dropdown-keyboard-controls?node-id=0-1&t=lLsSQVmiY4WmYmfA-1

Based on discussion with @piperdeck we will implement the following:
- [ ] do not auto focus resource type selector upon opening the panel
- [ ] tabbing with focus on any one of the selectors should open the dropdown automatically
- [ ] tabbing while a selector is open should focus the next selector without changing the current selection
- [ ] pressing enter should select an option
- [ ] the search functionality in the resource selector should search across all columns
- [ ] we will have search inputs always visible for all selectors. This is to replace the browser native select behaviour.
- [ ] we will skip tabbing to the search icon in a search input
- [ ] tabbing should skip over any disabled buttons/selectors within the edit panel

@piperdeck did I miss anything?

src/components/select/CustomSelect.tsx
edlerd, line 1: I get a visual glitch by following these steps:
1. Refresh page
2. Edit a permission group
3. Edit permissions
4. Select "server"

edlerd, line 1: When I select 3 things, then start changing the auto open next dropdown is not working.

edlerd: When in this "broken" state, this happens:

On opening the resource dropdown for images, the dropdown is initially rendered on the right, then it visually shifts to the left. This is probably resolved by fixing the auto-open, but this will be a problem if we want to use the Custom Select in other places.

edlerd, line 1: After selecting 3 things, on clicking "add", the 3rd dropdown opens.

edlerd (changes requested):
Found a couple of QA issues below.

src/components/select/CustomSelect.tsx
mas-who: > When I select 3 things, then start changing the auto open next dropdown is not working.

I'm not entirely sure what you meant by this. What were you trying to change? If you already have a selected value for all 3 selectors, which dropdown are you trying to open automatically next?

mas-who: > When in this "broken" state, this happens:
> 
> On opening the resource dropdown for images, the dropdown is initially rendered on the right, then it visually shifts to the left. This is probably resolved by fixing the auto-open, but this will be a problem if we want to use the Custom Select in other places.

Regarding this, I think we should definitely re-look at the auto-adjust logic in `ContextualMenu`. Specifically, at which point during the render cycle do we do the calculation and adjust styling for the dropdown menu should be considered for the above issue.

edlerd: > > When I select 3 things, then start changing the auto open next dropdown is not working.
> 
> I'm not entirely sure what you meant by this. What were you trying to change? If you already have a selected value for all 3 selectors, which dropdown are you trying to open automatically next?

I selected 3 things, then changed the selection of the 1st selector. From my understanding, that should open the 2nd selector, but it stays closed.

mas-who: got it, I assume you selected the option with the mouse? The auto open feature is for keyboard navigation only (tabbing), mostly for convenience. For mouse interactions, the behaviour is the same as before.

piperdeck:
This looks good to go!

edlerd:
As discussed,
1. the next dropdown should auto open if a selection is made with either mouse or keyboard, also on tab.
2. on add, we should not mess with the focus.

mas-who:
> As discussed,
> 
> 1. the next dropdown should auto open if a selection is made with either mouse or keyboard, also on tab.
> 2. on add, we should not mess with the focus.

I think I have finally found the issue within the `ContextualMenu`. See below logic:
Screenshot from 2024-10-10 18-52-31
The `onClick` of the toggle button gets triggered non-deterministically in its ordering relative to the `onFocus` event. This is relevant for the click to open scenario since both keyboard and mouse events will fire. One of them will end up closing the portal... going to now figure out a way to fix this.

mas-who:
@edlerd the discussed changes have been implemented, please have a look and see if you encounter any other issues. 

FYI, the solution I implemented for the `onClick` conflict with `onFocus` is to track a mouse down event on the toggle, and not handle the focus event while the mouse is down on the toggle.

src/util/permissions.tsx
```diff
@@ -342,3 +367,43 @@ export const enablePermissionsFeature = (): boolean => {
...
+      document.querySelectorAll<HTMLElement>(focusableSelectors);
+
+    // Find the index of the current element in the focusable elements list
+    // Then try to focus the next focusable element, if any
+    const index = Array.prototype.indexOf.call(focusableElements, element);
+    if (index > -1 && index < focusableElements.length - 1) {
+      focusableElements[index + (position === "next" ? 1 : -1)].focus();
+    }
+  }
+};
```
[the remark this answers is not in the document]
mas-who: the complexity comes from the tabbing interaction and I can't see another way of doing this since pressing tab while one of the select dropdown is open, we need to open the next dropdown. If we remove the tab to open functionality, then I agree with the above suggestion, we can keep things simple. 

Personally, I like the tabbing functionality as it makes the keyboard navigation a lot smoother for me. I think most of the complexity is in `moveToFocusable`, which is a general function that you would write once and not touch again. The logic in `handleTabbingForSelectors` seems also fairly straight forward no?

src/pages/permissions/panels/ResourceOptionLabel.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ResourceDetail } from "util/resourceDetails";
+import { resourceOptionColumns } from "util/permissions";
+
+interface Props {
+  resource: ResourceDetail;
+}
+
+const ResourceOptionLabel: FC<Props> = ({ resource }) => {
+  const columns =
+    resourceOptionColumns[resource.type] || resourceOptionColumns.default;
```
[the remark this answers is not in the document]
mas-who: good spot! thanks :)

src/pages/permissions/panels/ResourceOptionLabel.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const labelSegments = columns.map((column) => {
+    let value = resource[column] || "-";
+    if (Array.isArray(value)) {
+      value = value.join(", ") || "-";
+    }
+
+    return (
+      <span
+        key={`${column}-${value}`}
```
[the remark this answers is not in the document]
mas-who: aah yes I was thinking that there may be multiple row cells with the same column key, but here we are mapping inside the context of a single row

src/util/permissions.tsx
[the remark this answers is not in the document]
mas-who: @edlerd the changes as discussed have been implemented, I couldn't really implement a ref for the `add-entitlement` button as the `Button` component in react-component needs a ref input, which I think we can do outside the scope of this PR. Also made a few smaller adjustments/improvements:

1. When pressing `ESC` to close the select dropdown, keep focus on the select toggle.
2. After selecting a resource type, if there are no associated resources, then keep focus on the resource type selector.

Let me know if you agree with the above.

edlerd (commented):
Thanks for the adjustments, the code is getting much cleaner, and we are getting closer to standard compliance. Some more QA issues found, hope there is an easy, straight forward fix for those.

src/pages/permissions/panels/PermissionSelector.tsx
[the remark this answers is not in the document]
mas-who: Good point, nice catch, gonna fix that!

src/components/select/CustomSelect.tsx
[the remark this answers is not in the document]
edlerd: Actually, I think this is how it can be solved easily. From the active search input:

- Tab goes to the search result input clear/perform and then to the first item in the list. From the last item it can go back to the dropdown toggle
- shift + tab goes to the dropdown toggle

src/pages/permissions/panels/PermissionSelector.tsx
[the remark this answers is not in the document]
edlerd: That is confusing, now I also can't reproduce it any more oO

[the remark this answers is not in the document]
mas-who: Let's keep this comment open, I will try test a bit more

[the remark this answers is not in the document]
edlerd: Found another issue, but unrelated to this PR: When closing the panel with esc, we should unselect the group, as we do when we close the panel with the "cancel" button.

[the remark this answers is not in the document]
edlerd: This should better be handled in a follow up though.

src/components/select/CustomSelect.tsx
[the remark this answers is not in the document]
mas-who: I think it would be much easier to just move the selected option index up/down depending on tab or shift + tab combination? Technically, the toggle is always focused whilst the dropdown is open, and it makes sense to me to keyboard trap the tabbing sequence within the dropdown? I.e. tabbing or shift + tabbing will have the same effect as up/down keys, except when shift + tab is triggered when the selected option index is 0, then we close the dropdown and keep focus on the toggle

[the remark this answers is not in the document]
mas-who: I've just pushed an implementation, maybe give it a go and see if that makes sense?

src/pages/permissions/panels/PermissionSelector.tsx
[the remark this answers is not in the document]
mas-who: Will create an jira issue later for this, keeping the comment open so I don't forget

src/components/select/CustomSelectDropdown.tsx
```diff
@@ -86,47 +89,69 @@ const CustomSelectDropdown: FC<Props> = ({
 
   // track selected index from key board action and scroll into view if needed
   useEffect(() => {
-    if (selectedIndex !== -1 && optionsRef.current[selectedIndex]) {
+    if (optionsRef.current[selectedIndex]) {
```
edlerd, line 92: I think we can simplify now to `optionsRef.current[selectedIndex]?.scrollIntoView...`

src/components/select/CustomSelectDropdown.tsx
```diff
@@ -86,47 +89,69 @@ const CustomSelectDropdown: FC<Props> = ({
...
-        // consider lower bound for navigating up the list
-        return currIndex >= 0 ? currIndex : prevIndex;
+        return getNextOptionIndex(goingUp, prevIndex);
       });
     }
 
-    if (event.key === "Enter" && selectedIndex !== -1) {
-      onSelect(options[selectedIndex].value);
+    if (event.key === "Enter") {
+      onSelect(filteredOptions[selectedIndex].value);
```
edlerd, line 154: I think we have to double-check here if the selectedIndex is valid. Though for the permission selector this behaves correctly already.

src/components/select/CustomSelectDropdown.tsx
edlerd, line 1: I type something into the search input and accept with enter. The next time I enter the first letter into the same search input, I get suggested my previous typed words. I think there is a setting on an input to disable the browser auto suggest, that we should set here to prevent this.

src/pages/permissions/panels/PermissionSelector.tsx
```diff
@@ -62,39 +71,45 @@ const PermissionSelector: FC<Props> = ({ onAddPermission }) => {
...
-
   const handleResourceTypeChange = (value: string) => {
     const resourceType = value;
     setEntitlement("");
-    setResourceType(resourceType);
+    setResourceType((prev) => {
+      // in case if the same resource type is selected, open the resource dropdown
+      if (!!resourceType && prev === resourceType && permissions?.length) {
+        resourceRef.current?.open();
+      }
```
edlerd, line 109: This shouldn't be a special case. Probably better to remove the useeffect above and always handle the "focus next selector" here.
Because currently there is a special case for server we should go to the next-next selector and we shouldn't have to maintain that special case twice. So generalizing to have one place to handle it directly seems better than the indirect handling with the use effect.

src/util/permissions.tsx
```diff
@@ -342,3 +367,21 @@ export const enablePermissionsFeature = (): boolean => {
 
   return hasAccessManagement && (hasOIDCSettings || userShowPermissions);
 };
+
+// each resource type has specific columns to display, which should uniquely identify the resource
+export const getResourceOptionColumns = (type: string) => {
+  const resourceOptionColumns: Record<string, (keyof ResourceDetail)[]> = {
+    image: ["aliases", "fingerprint", "description", "imageType"],
```
edlerd, line 374: I think the description should come first, as it will be the label on the toggle after selection.

src/util/permissions.tsx
edlerd, line 1: Given resource type "image", for an image without description and alias.
The label after selection in the toggle is "undefined", so there seems to be something off in this case.

src/pages/permissions/panels/EntitlementOptionLabel.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const EntitlementOptionLabel: FC<Props> = ({ entitlement, description }) => {
+  return (
+    <div className="label">
+      <span title={entitlement} className="entitlement u-truncate">
+        {entitlement}
+      </span>
+      <span
+        title={description}
+        className="entitlement-description u-truncate u-text--muted"
```
edlerd, line 16: The muted doesn't work in case the row is selected. The contrast of text to background seems too low. There is probably an inverted class or something that can be set for muted text to be handling this case well.

src/components/select/CustomSelect.tsx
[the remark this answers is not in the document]
edlerd: this is pretty neat, yes.

edlerd (commented):
This looks very good.
Some probably easy to fix issues below, then this should be good to merge.

src/components/select/CustomSelectDropdown.tsx
mas-who: Since `selectedIndex` starts at 0 now, it will be a valid index. Note `filteredOptions` is computed before this part based on the search input state, and every time the search is updated we will reset the `selecteIndex` to 0.

src/components/select/CustomSelectDropdown.tsx
mas-who: right, it's the `autocomplete` prop on `SearchBox`, setting it to "off" resolves the issue.

src/pages/permissions/panels/PermissionSelector.tsx
mas-who: The `useEffect` is needed specifically for `resourceType` selection because we need to cater for the case when no permissions (and therefore resources) are available for a resource type. In that case, want to keep the focus on the resource type selector. The permissions for a resource type is fetched from the server when `resourceType` updates so it has to be dealt within a `useEffect`. 

In the case where the `resourceType` selection remains the same, we have to deal with it inside this function since we need to check if the same resource type was selected. Also the `useEffect` above will not trigger.

src/util/permissions.tsx
mas-who: It's supposed to be the fingerprint that gets displayed. Would you be able to share the response body for the specific image that you ran into the issue with?

Screenshot from 2024-10-13 10-42-56

src/pages/permissions/panels/PermissionSelector.tsx
edlerd: I see so we do need two entry points to cover both cases. Maybe refactor the handler into a function, so we can call it from both entry points.

src/util/permissions.tsx
edlerd: You can check on the demo server, it is the 2nd last one: 
image

mas-who: this should be resolved, will need to check on the demo server though since I can't figure out how to create a image without any properties in the response payload (saw this on the demo server).

src/components/select/CustomSelectDropdown.tsx
edlerd: What if there are no search value matches and `filteredOptions` is empty?

src/pages/permissions/panels/PermissionSelector.tsx
edlerd, line 1: QA issue: After selecting "change permission", the resource type selector is briefly selected and then gets unselected


https://github.com/user-attachments/assets/e43e7c1d-93b2-451c-9526-8f9b0c92abca

src/components/select/CustomSelectDropdown.tsx
mas-who: yes of course :+1: added a validation here now

src/pages/permissions/panels/PermissionSelector.tsx
edlerd, line 1: QA issue: When selecting resource type "server" and then selecting resource type "server" again, the middle menu opens, even though it is disabled.


https://github.com/user-attachments/assets/bbe4d970-512a-4148-800a-5c7bf24efadd

edlerd, line 1: QA issue: The order in the resource type dropdown seems random
image

edlerd: Also on the 1st selection of "server" resource type there is a visual glitch where the entitlement selector is jumping around.

edlerd, line 1: QA issue: When selecting resource type "server" and the entitlement dropdown is open and the first item in the list is active. Then shift+tab won't close the dropdown as with the other resource types.

edlerd: Also, we should think what to do when the user presses tab and the last item in the dropdown is active. The current behaviour is doing nothing, which seems wrong.

I think tab and shift+tab should both close the dropdown, to be compliant with the browser-native behaviour of a select component.

edlerd, line 1: QA issue: The height of the entitlement dropdown is not calculated correctly, in case the browser window is too low. When using the arrow keys and reaching an entry that would force a scroll down, the dropdown is causing a visual glitch and losing focus.

https://github.com/user-attachments/assets/dee73636-3400-43c7-873c-45354e70ce77

edlerd (changes requested):
Found a couple of QA issues with the recent changes.

src/pages/permissions/panels/PermissionSelector.tsx
edlerd: This is with a 6.1-78a3d8f backend. I think the select component should apply ordering.

mas-who: right, when the backend does not return metadata entities, we are using resource type options hardcoded in the client code. These options must be sorted too.

mas-who: Yes my logic is off for the case where the options consists of groups, will resolve. The behaviour of tabbing to close the dropdown at the end of the list sounds good :+1:

mas-who: The QA issue is resolved, but I can't seem to replicate the visual glitch for the entitlement selector?

[Screencast from 14-10-2024 17:27:33.webm](https://github.com/user-attachments/assets/eb84f949-b61b-4729-82bf-f05516500d3f)

mas-who: Previously I had the height adjustment logic in `CustomSelect`, but it should be placed inside the `CustomSelectDropdown` component instead since it's more reliable to adjust the dropdown menu height as the component renders instead of adjusting height based on open state in `CustomSelect`.

edlerd: it is resolved for me as well.

src/components/select/CustomSelectDropdown.tsx
edlerd, line 1: QA issue: When the dropdown is open and I switch focus to another tab or another application, the dropdown stays open. When I return to this tab the dropdown is still open, but there is no focus in it.

Looking at the browser dropdown, I think it should close when putting focus on another tab or application.

src/components/select/CustomSelectDropdown.tsx
```diff
@@ -84,49 +87,80 @@ const CustomSelectDropdown: FC<Props> = ({
     }, 100);
   }, [isSearchable]);
 
+  useLayoutEffect(() => {
+    adjustDropdownHeight(dropdownListRef.current, searchRef.current);
+  }, []);
+
+  useEventListener("resize", () =>
+    adjustDropdownHeight(dropdownListRef.current, searchRef.current),
+  );
```
edlerd, line 96:
```suggestion
  const handleResize = () => {
    adjustDropdownHeight(dropdownListRef.current, searchRef.current);
  };

  useLayoutEffect(handleResize, []);
  useEventListener("resize", handleResize);
```
A bit cleaner

src/pages/permissions/panels/PermissionSelector.tsx
```diff
@@ -62,39 +72,66 @@ const PermissionSelector: FC<Props> = ({ onAddPermission }) => {
...
+    if (!prevResourceType && currResourceType) {
+      // open the resource dropdown if there are permissions for the resource type
+      if (permissions?.length) {
+        resourceRef.current?.open();
+        return currResourceType;
+      }
+
+      // if there are no permissions, we can't select a resource
+      // keep focus on the resource type selector
+      resourceTypeRef.current?.focus();
```
edlerd, line 119:
```suggestion
  const focusAfterResourceTypeChange = (
    previous: string,
    current: string,
    permissions?: LxdPermission[],
  ) => {
    // for server, there will be no resources to select
    // skip to entitlement selector
    if (current === "server") {
      entitlementRef.current?.open();
      return current;
    }

    const wasChanged = !previous && current;
    const wasReselected = previous && current === previous;
    if (wasChanged || wasReselected) {
      // open the resource dropdown if there are permissions for the resource type
      if (permissions?.length) {
        resourceRef.current?.open();
        return current;
      }

      // if there are no permissions, we can't select a resource
      // keep focus on the resource type selector
      resourceTypeRef.current?.focus();
    }

    return current;
  };
```
We can remove the duplication and should aim for self documenting code in favour of too many comments.

edlerd (commented):
Found one more QA issue and two clean code suggestions. After that should be good to merge.

src/components/select/CustomSelectDropdown.tsx
mas-who: Aaah this is a interesting problem, after reading a bit I think we need to use specific events on the document and window objects. Will get this sorted :+1:

edlerd: I think you can add a listener on this event and close the dropdown based on it: https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event

mas-who: yes that's the one :+1:

mas-who: additionally I think the `blur` event on `window` object should be introduced for the case of focusing on another application.

src/components/select/CustomSelect.tsx
edlerd, line 1: QA issue: When using down arrow or tab to go through the items in the selector that is overflowing with a scrollbar, after the last item in the visible area, the first item in the visible area is selected in case the mouse is hovering that one. In the screencast attached, pressing down arrow all the time, while mouse is hovering first element.

https://github.com/user-attachments/assets/2123a354-a5c5-4353-a0e2-958659ee5795

edlerd (changes requested):

src/components/select/CustomSelect.tsx
edlerd: Expected is to go through the items in the list sequentially without jumping to the first one.

edlerd: A similar issue exists, when auto-opening the dropdown for resource/entitlement after the previous selection has been made. In case the mouse hovers any of the items, that item will be the first item selected on tab/down arrow. In that case I'd expect the first item to be selected, independent of mouse position.

mas-who: Will resolved the jumping issue.

Regarding the interaction between mouse and keyboard events inside a select dropdown. Looking at the browser native select element, the first highlighted option is the one that the mouse is hovering over, then keyboard navigation with up/down keys will start from that option. So I think the current behaviour is correct. Also I noticed that when pressing tab while a select dropdown is open, it actually closes the dropdown and keep focus on the toggle. I think we should stick to that behaviour for the custom select component, wdyt?

[Screencast from 15-10-2024 09:53:08.webm](https://github.com/user-attachments/assets/8df14368-9da6-4391-8d1a-6246f4489533)

mas-who: Okay sounds good, then I will remove the tabbed navigation of the options with the next push :+1:

mas-who: The jumping issue due to mouse hovering an option in a scrollable dropdown should be resolved now, please have a look.

edlerd: That is a rather complex solution. It also has a downside: When we now hover an item, then scroll down, then move the mouse, the item will not be selected, unless we stop hovering and start hovering it again.

Better remove the extra state and rely on the `onMouseMove` event instead of `onMouseEnter`

mas-who: purely using `onMouseMove` seems to not be sufficient since when opening the dropdown, it no longer detects if the mouse is hovering over an option without it moving first. I think we should combine `onMouseMove` and `onMouseEnter`, keep the state that tracks keyboard navigation for the `onMouseEnter` event. Wdyt?

edlerd: > purely using onMouseMove seems to not be sufficient since when opening the dropdown, it no longer detects if the mouse is hovering over an option without it moving first.

I think that is actually a good thing. Though it deviates from the native select behaviour, I think it is better UX.

mas-who: Yeah I agree, okay then I think the implementation is there, will resolve this comment once you have a look and everything is okay

src/pages/permissions/panels/PermissionSelector.tsx
```diff
@@ -1,15 +1,17 @@
 import { Button, useNotify } from "@canonical/react-components";
-import CustomSelect from "../../../components/select/CustomSelect";
+import CustomSelect, {
+  SelectRef,
+} from "../../../components/select/CustomSelect";
```
edlerd, line 4:
```suggestion
} from "components/select/CustomSelect";
```
Better use the non-relative imports

src/components/select/CustomSelectDropdown.tsx
edlerd: Reopening this one: When switching to another tab, the dropdown closes, when switching back, it opens again. Might be ok though, as the focus issue from earlier is gone.

edlerd (dismissed):
QA and code LGTM, thanks for driving this to perfection :)

Tiny nitpick below, but should be good as is to merge.

src/components/select/CustomSelectDropdown.tsx
mas-who: I can't replicate this on demo server (or locally) :sweat: 

[Screencast from 15-10-2024 17:42:27.webm](https://github.com/user-attachments/assets/58ccc818-b691-4354-9adc-333438d58d6f)

edlerd: It doesn't happen in all cases. For me, when choosing resource type "instance" and having instances to chose from, the resource dropdown will open when coming back to the tab. This is even the case if all 3 selections have been made. Or if the focus on leaving the tab is on the search input below the 3 selectors.

I think this is a side effect of the data refetching when bringing focus back to the tab. Use query will refresh in this case and might trigger the dropdown open and autofocus. This should be avoided.


https://github.com/user-attachments/assets/670c06d8-7cbd-4861-88ef-b48af68adfd9

mas-who: yes you are right, the permissions query are being executed on window focus, which triggers the `useEffect` that causes the resource selector to open. Setting `refetchOnWindowFocus` to `false` for the query prevents the above issue

edlerd: That's probably not the right solution. Because 1. it is not a pattern we'd want to have to apply to all cases using a drop-down with dynamically fetched content. And 2. It might lead to stale data if tabs stay open long in the background.

Can we avoid the autofocus and open on refetch of the data? Like why does it happen in the first place?

mas-who: Aaah sorry I should have mentioned, the `permissions` array does not have the same ordering when fetched from the API, so it causes the `useEffect` to trigger and therefore opens the selector. 

You are right about the stale data issue, I think in this case then we can leave this behaviour where the resource selector may stay open on window focus? Doing a deep comparison on `permissions` is not great due to the sheer quantity of data that may be returned by the endpoint. We could ask the backend to always sort this data, but I assume the reason why they don't do it is probably because sorting this amount of data may severely slow down the response speed for the endpoint.

edlerd: And we should probably open an issue on lxd to have the endpoint return a stable ordering.

mas-who: I'm not sure if we want to sort in the UI though. For context, 8 instances generates around 100 permissions, so this can turn into a really expensive operation wdyt?

edlerd: Ideally backend will do it for us, testing it locally, it behaves fine -- it takes for me half a millisecond with ~70 instances. I doubt this will be any problem:
```
        return resolve(
          data.metadata.sort((a, b) => {
            if (a.url === b.url) {
              if (a.entitlement === b.entitlement) {
                return a.entity_type.localeCompare(b.entity_type);
              }
              return a.entitlement.localeCompare(b.entitlement);
            }
            return a.url.localeCompare(b.url);
          }),
        );
```

edlerd (approved):
QA and code LGTM, good stuff :)

PR #943, opened 2024-10-08
Add links to grafana from instance detail pages, add grafana provision script

edlerd:
## Done

- add grafana base url to the settings
- add links to grafana from instance detail page if its base url is configured
- add script to setup grafana

## QA

1. Run the LXD-UI:
    - setup o11y stack for lxd: 
```
lxc launch ubuntu:24.04 grafana
curl https://localhost:8407/ui/assets/scripts/setup-grafana.sh --insecure > /tmp/setup-grafana.sh
bash /tmp/setup-grafana.sh grafana default
rm /tmp/setup-grafana.sh
```
    - create the dashboard:
      - open grafana with the link from the script above, then
      - login with admin/admin and chose a new password
      - after login, go to dashboards > new > import
      - enter 19131 and click load
      - confirm the options for prometheus and loki, the click import
    - go back to the lxd-ui and ensure the links from instance detail page to "metrics" work fine.

### DIFF
--- a/public/assets/scripts/setup-grafana.sh
+++ b/public/assets/scripts/setup-grafana.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <instance> <project>"
+    echo "Error: Both 'instance' and 'project' arguments are required."
+    exit 1
+fi
+
+INSTANCE=$1
+PROJECT=$2
+
+IS_LXD_CLUSTERED=$(lxc info | grep "server_clustered:" | grep "false")
+if [ -z "$IS_LXD_CLUSTERED" ]; then
+    echo "Error: LXD is clustered, this script only works for single node installations."
+    echo "See https://documentation.ubuntu.com/lxd/en/latest/metrics/ for more information."
+    exit 1
+fi
+
+CONTAINER_IP=$(lxc info "$INSTANCE" --project="$PROJECT" | grep inet: | grep "global" | head -n1 | cut -d ":" -f2 | cut -d " " -f3 | cut -d "/" -f1)
+CONTAINER_UPLINK_IP=$(echo "$CONTAINER_IP" | cut -d "." -f1,2,3)".1"
+echo "Found container IP as '$CONTAINER_IP' and uplink as '$CONTAINER_UPLINK_IP'"
+
+set -e
+set -x
+
+# upload server.crt to container
+lxc info | sed -n "/BEGIN CERTIFICATE/,/END CERTIFICATE/p" | sed 's/^[ \t]*//;s/[ \t]*$//' > /tmp/server.crt
+lxc file push /tmp/server.crt "$INSTANCE"/root/server.crt --project="$PROJECT"
+rm /tmp/server.crt
+
+# install and configure grafana and prometheus in container
+lxc exec "$INSTANCE" --project="$PROJECT" bash <<EOF
+set -x
+set -e
+# install grafana and prometheus
+apt-get update
+apt-get install -y apt-transport-https software-properties-common wget prometheus
+mkdir -p /etc/apt/keyrings/
+wget -q -O - https://apt.grafana.com/gpg.key | gpg --dearmor | tee /etc/apt/keyrings/grafana.gpg > /dev/null
+echo "deb [signed-by=/etc/apt/keyrings/grafana.gpg] https://apt.grafana.com stable main" | tee -a /etc/apt/sources.list.d/grafana.list
+apt-get update
+apt-get install -y grafana=11.4.0 loki=3.3.2 promtail=3.3.2
+systemctl daemon-reload
+systemctl start grafana-server
+systemctl enable grafana-server.service
+sed -ie '44d' /etc/loki/config.yml # fix the loki configuration, see https://github.com/grafana/loki/issues/15039
+systemctl start loki
+systemctl enable loki
+systemctl start promtail
+systemctl enable promtail
+
+# generate ssl key for grafana to serve via https
+openssl req -x509 -newkey rsa:4096 -keyout /etc/grafana/grafana.key -out /etc/grafana/grafana.crt -days 365 -nodes -subj "/CN=metrics.local"
+chown grafana:grafana /etc/grafana/grafana.crt
+chown grafana:grafana /etc/grafana/grafana.key
+chmod 400 /etc/grafana/grafana.key /etc/grafana/grafana.crt
+sed -i "s#;protocol = http#protocol = https#" /etc/grafana/grafana.ini
+cat <<EOT > /etc/grafana/provisioning/datasources/lxd-sources.yaml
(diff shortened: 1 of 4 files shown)


public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,112 @@
...
+set -x
+set -e
+# install grafana and prometheus
+apt-get update
+apt-get install -y apt-transport-https software-properties-common wget
+mkdir -p /etc/apt/keyrings/
+wget -q -O - https://apt.grafana.com/gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/grafana.gpg > /dev/null
+echo "deb [signed-by=/etc/apt/keyrings/grafana.gpg] https://apt.grafana.com stable main" | sudo tee -a /etc/apt/sources.list.d/grafana.list
+apt-get update
+apt-get install -y grafana prometheus loki promtail
```
mas-who, line 32: as discussed, I think we should fix the versions for the different components being installed here?

public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,112 @@
...
+systemctl daemon-reload
+systemctl start grafana-server
+systemctl enable grafana-server.service
+sed -ie '44d' /etc/loki/config.yml # fix the loki configuration
```
mas-who, line 36: what's the context for this fix? Potentially worth leaving a reference for the issue

public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,112 @@
...
+systemctl start loki
+systemctl enable loki
+systemctl start promtail
+systemctl enable promtail
+
+# generate ssl key for grafana to serve via https
+openssl req -x509 -newkey rsa:4096 -keyout /etc/grafana/grafana.key -out /etc/grafana/grafana.crt -days 365 -nodes -subj "/CN=metrics.local"
+sudo chown grafana:grafana /etc/grafana/grafana.crt
```
mas-who, line 44: I think `sudo` isn't necessary here since we are running as root in the lxd container

public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,112 @@
...
+sudo chmod 400 /etc/grafana/grafana.key /etc/grafana/grafana.crt
+sed -i "s#;protocol = http#protocol = https#" /etc/grafana/grafana.ini
+cat <<EOT > /etc/grafana/provisioning/datasources/lxd-sources.yaml
+apiVersion: 1
+
+datasources:
+  - name: prometheus
+    type: prometheus
+    access: proxy
+    url: http://localhost:9090
```
mas-who, line 55: Why don't we set these with the `CONTAINER_IP` variable as well? Functionally there is no difference but when I am in the grafana data sources section, these urls are slightly confusing since I first thought both prom and loki are running on my local host.

public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,112 @@
...
+chown -R prometheus /etc/prometheus/tls
+
+# configure prometheus
+cat <<EOT > /etc/prometheus/prometheus.yml
+global:
+  scrape_interval:     15s
+  evaluation_interval: 15s
+  scrape_timeout: 15s
+scrape_configs:
+  - job_name: lxd
```
mas-who, line 77: The current script works only for a standalone lxd installation. Is it out of scope to make this work with lxd clusters? If so, maybe we should display a message (potentially in a different color) at the beginning of the script to indicate that the current setup does not work with clusters.

public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,112 @@
...
+    scrape_interval: 15s
+    scrape_timeout: 15s
+    metrics_path: '/1.0/metrics'
+    scheme: 'https'
+    static_configs:
+      - targets: ['$CONTAINER_UPLINK_IP:8443']
```
mas-who, line 83: for my information, would you explain the concept behind `CONTAINER_UPLINK_IP`? So if we use the second address in the ipv4 range, then we can reach the host?

public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,112 @@
...
+systemctl enable prometheus.service
+EOF
+
+# download metrics.crt from container and add to host lxd trust store
+lxc file pull "$INSTANCE"/etc/prometheus/tls/metrics.crt /tmp/metrics.crt --project="$PROJECT"
+lxc config trust add /tmp/metrics.crt --type=metrics
+rm /tmp/metrics.crt
+
+# configure host lxd for loki and to link to grafana
+CONTAINER_IP=$(lxc info "$INSTANCE" --project="$PROJECT" | grep inet: | grep "global" | head -n1 | cut -d ":" -f2 | cut -d " " -f3 | cut -d "/" -f1)
```
mas-who, line 103: maybe we can set this as a variable in the beginning of the script, and reuse it for `CONTAINER_UPLINK_IP`?

mas-who (commented):
QA looks good :+1: just got a few comments

public/assets/scripts/setup-grafana.sh
edlerd: The loki config as it is delivered with the version at hand is broken. Good idea to reference the issue here.

public/assets/scripts/setup-grafana.sh
edlerd: Good idea.

public/assets/scripts/setup-grafana.sh
edlerd: Yes, this is assuming we have bridge networking ipv4 on the container. Probably something we should mention and/or check. With that in place, the networks .1 address is the host.

mas-who: I did not know that, thanks for explaining :+1:

public/assets/scripts/setup-grafana.sh
```diff
@@ -0,0 +1,129 @@
...
+lxc file push /tmp/server.crt "$INSTANCE"/root/server.crt --project="$PROJECT"
+rm /tmp/server.crt
+
+# install and configure grafana and prometheus in container
+lxc exec "$INSTANCE" --project="$PROJECT" bash <<EOF
+set -x
+set -e
+# install grafana and prometheus
+apt-get update
+apt-get install -y apt-transport-https software-properties-common wget prometheus
```
mas-who, line 37: shouldn't we fix version for `prometheus` as well?

mas-who (commented):
Mostly LGTM, just one question about dep version

public/assets/scripts/setup-grafana.sh
mas-who: aaah, thanks for clarifying :+1:

mas-who (approved):
LGTM :+1:

PR #944, opened 2024-10-08
feat: [WD-15296] Disable other devices for LXD 5.0

Kxiru:
## Done

- Disables the 'Other' tab in Profile and Instance configuration when using LXD 5.0
- Added a test to ensure this.

## QA

1. Run the LXD-UI:
    - Attempt to access the 'Other' tab when in Instance or Profile config.

## Screenshots

N/A

### DIFF
--- a/src/pages/instances/forms/InstanceFormMenu.tsx
+++ b/src/pages/instances/forms/InstanceFormMenu.tsx
@@ -3,6 +3,7 @@ import MenuItem from "components/forms/FormMenuItem";
 import { Button, useNotify } from "@canonical/react-components";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import useEventListener from "@use-it/event-listener";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
 
 export const MAIN_CONFIGURATION = "Main configuration";
 export const DISK_DEVICES = "Disk";
@@ -34,6 +35,7 @@ const InstanceFormMenu: FC<Props> = ({
 }) => {
   const notify = useNotify();
   const [isDeviceExpanded, setDeviceExpanded] = useState(true);
+  const { hasMetadataConfiguration } = useSupportedFeatures();
 
   const disableReason = isDisabled
     ? "Please select an image before adding custom configuration"
@@ -87,7 +89,9 @@ const InstanceFormMenu: FC<Props> = ({
               />
               <MenuItem label={GPU_DEVICES} {...menuItemProps} />
               <MenuItem label={PROXY_DEVICES} {...menuItemProps} />
-              <MenuItem label={OTHER_DEVICES} {...menuItemProps} />
+              {hasMetadataConfiguration && (
+                <MenuItem label={OTHER_DEVICES} {...menuItemProps} />
+              )}
             </ul>
           </li>
           <MenuItem label={RESOURCE_LIMITS} {...menuItemProps} />
--- a/src/pages/profiles/forms/ProfileFormMenu.tsx
+++ b/src/pages/profiles/forms/ProfileFormMenu.tsx
@@ -5,6 +5,7 @@ import { updateMaxHeight } from "util/updateMaxHeight";
 import useEventListener from "@use-it/event-listener";
 import { hasDiskError, hasNetworkError } from "util/instanceValidation";
 import { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
 
 export const MAIN_CONFIGURATION = "Main configuration";
 export const DISK_DEVICES = "Disk";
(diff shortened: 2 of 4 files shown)


src/pages/instances/forms/InstanceFormMenu.tsx
```diff
@@ -87,7 +88,9 @@ const InstanceFormMenu: FC<Props> = ({
               />
               <MenuItem label={GPU_DEVICES} {...menuItemProps} />
               <MenuItem label={PROXY_DEVICES} {...menuItemProps} />
-              <MenuItem label={OTHER_DEVICES} {...menuItemProps} />
+              {useSupportedFeatures().hasMetadataConfiguration ? (
```
mas-who, line 91: generally we use custom hooks outside of the return of a react component, let call `useSupportedFeatures()` above.

src/pages/instances/forms/InstanceFormMenu.tsx
```diff
@@ -87,7 +88,9 @@ const InstanceFormMenu: FC<Props> = ({
...
+                <MenuItem label={OTHER_DEVICES} {...menuItemProps} />
+              ) : null}
```
mas-who, line 93: A bit easier to read I think
```suggestion
              {hasMetadataConfiguration && <MenuItem label={OTHER_DEVICES} {...menuItemProps} />}
```

src/pages/profiles/forms/ProfileFormMenu.tsx
```diff
@@ -87,7 +88,9 @@ const ProfileFormMenu: FC<Props> = ({
               />
               <MenuItem label={GPU_DEVICES} {...menuItemProps} />
               <MenuItem label={PROXY_DEVICES} {...menuItemProps} />
-              <MenuItem label={OTHER_DEVICES} {...menuItemProps} />
+              {useSupportedFeatures().hasMetadataConfiguration ? (
```
mas-who, line 91: same comments as in `InstanceFormMenu.tsx`

tests/instances.spec.ts
```diff
@@ -398,3 +398,15 @@ test("Migrate instance root storage volume to a different pool", async ({
   await migrateInstanceRootStorage(page, instance, "default", serverClustered);
   await deletePool(page, targetPool);
 });
+
+test("'Other' tab is removed when config/creating Instances, on LXD Version 5.0", async ({
```
mas-who, line 402: This test is failing, and likely the other one will fail too. If you look at the playwright [video](https://canonical.github.io/lxd-ui/reports/pr-944/11241353075/1/#?testId=b980ab2a78df8d904d9c-9be92e23cd8b98d2fb66) for this failing test, you can see the "Other" tab is hidden, so likely `page.getByText("Other")` is resolving to some other components on the page.

To develop these tests locally, you have two options:
1. revert lxd to 5.0 and test
2. emulate the condition by just hiding the "Other" tab without using api extension, make this test pass locally without the `skip`, then it should pass in CI as well.

I personally prefer approach 2, since reverting to lxd 5.0 will require you to uninstall lxd and reinstall 5.0 plus doing all of the initial configs.

mas-who (commented):
Good draft, you got the gist of how api extensions work now :) The tests are however failing for lxd 5.0/edge. I left some suggestions on how to develop these tests locally, and also some suggestions on coding style :)

src/pages/profiles/forms/ProfileFormMenu.tsx
```diff
@@ -87,7 +89,9 @@ const ProfileFormMenu: FC<Props> = ({
               />
               <MenuItem label={GPU_DEVICES} {...menuItemProps} />
               <MenuItem label={PROXY_DEVICES} {...menuItemProps} />
-              <MenuItem label={OTHER_DEVICES} {...menuItemProps} />
+              {!hasMetadataConfiguration && (
```
edlerd, line 92:
```suggestion
              {hasMetadataConfiguration && (
```
I think this needs to be inversed.

edlerd (changes requested):

src/pages/profiles/forms/ProfileFormMenu.tsx
mas-who: oops, that's my bad!

tests/instances.spec.ts
```diff
@@ -398,3 +398,15 @@ test("Migrate instance root storage volume to a different pool", async ({
...
+  page,
+  lxdVersion,
+}) => {
+  test.skip(
+    lxdVersion != "5.0-edge",
+    "Unavailable API Extensions: metadata_configuration",
```
mas-who, line 408: the reason why we would skip the test here is because `metadata_configuration` api extension exists in lxd 5.21 and latest/edge. Please also make the same change for the other test.

```suggestion
    "Newer LXD versions has the metadata_configuration API extension.",
```

edlerd: I think the skip message should say why the test is skipped. In my opinion, the current message is expressing that just correctly, no?

edlerd: I mean the message could be a bit more verbose and explain "In 5.0 we are missing the extension ... "

edlerd: Oh got you, yes, all right.

Kxiru: I was trying to express the fact that 5.0 does NOT have the metadata_configuration, but I can replace it with the text that you provided, no worries.

edlerd (approved):
LGTM

PR #946, opened 2024-10-10
feat: resource link component for semiotics [WD-15623]

mas-who:
## Done

- Added a chip component for displaying lxd resources
- Added custom icons that we cannot get from vanilla
- Mock implementation for instance creation toast notifications

## QA

1. Run the LXD-UI:
    - TBC

## Screenshots
Screenshot from 2024-10-10 11-47-52


src/components/ResourceLink.tsx
```diff
@@ -0,0 +1,60 @@
...
+  type: ResourceType;
+  value: string;
+  to: string;
+}
+
+const ResourceLink: FC<Props> = ({ type, value, to }) => {
+  const icon = resourceIconMap[type];
+  const navigate = useNavigate();
+  return (
+    <button className="p-chip is-inline" onClick={() => navigate(to)}>
```
edlerd, line 50: I think button is the wrong component to use and this should be a `Link` to be semantically correct. Then the browser would display the target url correctly and you can do things like open in a new tab.

src/sass/_pattern_icon.scss
```diff
@@ -2,11 +2,83 @@
...
+}
+
+@mixin lxdui-icon-certificate() {
+  background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' width='16px' height='16px' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath%20fill-rule%3D%22evenodd%22%20clip-rule%3D%22evenodd%22%20d%3D%22M13%202C14.1046%202%2015%202.89543%2015%204V11C15%2012.1046%2014.1046%2013%2013%2013H12V11.5H13C13.2455%2011.5%2013.4496%2011.3231%2013.4919%2011.0899L13.5%2011V4C13.5%203.75454%2013.3231%203.55039%2013.0899%203.50806L13%203.5H3C2.75454%203.5%202.55039%203.67688%202.50806%203.91012L2.5%204V11C2.5%2011.2455%202.67688%2011.4496%202.91012%2011.4919L3%2011.5H7V13H3C1.89543%2013%201%2012.1046%201%2011V4C1%202.89543%201.89543%202%203%202H13ZM9.5%207C10.8807%207%2012%208.11929%2012%209.5C12%2010.3175%2011.6076%2011.0433%2011.001%2011.4995L11%2016L9.5%2014.9561L8%2016L8.00004%2011.5002C7.3928%2011.0441%207%2010.3179%207%209.5C7%208.11929%208.11929%207%209.5%207ZM8.5%209.5C8.5%208.94772%208.94772%208.5%209.5%208.5C10.0523%208.5%2010.5%208.94772%2010.5%209.5C10.5%2010.0523%2010.0523%2010.5%209.5%2010.5C8.94772%2010.5%208.5%2010.0523%208.5%209.5Z%22%20fill%3D%22black%22%2F%3E%3C%2Fsvg%3E");
+}
+
```
edlerd, line 36: We should update vanilla first, so we can import those instead of defining them here.

src/components/ResourceLink.tsx
edlerd, line 1: Can you add 1 or 2 more use cases in pages, maybe with the different types, so it is easier to see the component in action without having to create a new instance?

src/sass/_pattern_icon.scss
mas-who: still waiting for e2e to pass on #947

src/components/ResourceLink.tsx
mas-who: I added the new component to the instance detail overview page. I think with the new design from @piperdeck, where the border is removed from the chip, it looks decent. Side note this is after discussion between piper and lyubo. I'm not too sure about the status dots for storage pools (and volumes) though.

Screenshot from 2024-10-11 11-59-08

mas-who:
@edlerd @piperdeck I have applied the chip design across all floating notifications. Also implemented them on the instance detail overview page. Let's use the PR to help with our discussion later.

PS: also updated the side nav icons

src/components/Navigation.tsx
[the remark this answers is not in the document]
edlerd: i.e. the instance list https://lxd-ui-946.demos.haus/ui/project/No-Root-Disk-Project/instances should have the same icon in the empty state as the navigation icon.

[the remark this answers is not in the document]
mas-who: Ahh yes, I think safe bet is to do a global search for all `<Icon />` invocations in the codebase and see if there are any other missed scenarios

src/pages/cluster/ClusterGroupForm.tsx
```diff
@@ -85,7 +86,17 @@ const ClusterGroupForm: FC<Props> = ({ group }) => {
...
+            <>
+              Cluster group{" "}
+              <ResourceLink
+                type="cluster-member"
+                value={values.name}
+                to={`/ui/cluster/group/${values.name}`}
+              />{" "}
+              {verb}.
+            </>,
+          );
```
edlerd, line 99: I think we should always keep the resource type as part of the copy, like in this example. That makes it easy to learn the icons and reduces the cognitive load, i.e. when browsing notifications. Currently, you'd need to understand what each icon stands for to process a notification. Putting the type name as an explicit part of the copy makes it much easier to process.

mas-who: This is another design decision we need to align on with @piperdeck. I believe the purpose of using icons is to eliminate the need for resource type text in notifications. In last week’s discussion, we agreed that only storage pools and custom volumes should include their types in the notification copy, as we don’t yet have representative icons for them. Before proceeding with this change, I’d like to confirm this approach, as it will require updates to all the e2e tests.

edlerd: Sure, lets discuss with piper. When I was playing with this branch I found it hard to digest the notifications, hence my suggestion.

> Before proceeding with this change, I’d like to confirm this approach, as it will require updates to all the e2e tests.

I thought this would actually make the e2e test easier, because we had this in the past. So with this suggestion, the CI suite would need fewer adjustments.

mas-who: Aaah right I already made adjustments to all the e2e tests :sweat_smile: but it's easy enough to reset the commit and remove those changes, but let's first confirm the change with piper. I do agree that it's additional learning if we remove the resource type copy text from the notifications

mas-who: @edlerd after discussing with @piperdeck I've added the resource type back into all notifications. Also reverted all the e2e test changes.

mas-who: Related to this [comment](https://github.com/canonical/lxd-ui/pull/952#issuecomment-2414631370) from #952 . I have adjusted the instance side navigation tab and empty state to use the `pod` icon.

src/pages/cluster/ClusterGroupForm.tsx
```diff
@@ -85,7 +86,17 @@ const ClusterGroupForm: FC<Props> = ({ group }) => {
...
           navigate(`/ui/cluster/group/${values.name}`);
-          toastNotify.success(`Cluster group ${values.name} ${verb}.`);
+          toastNotify.success(
+            <>
+              Cluster group{" "}
+              <ResourceLink
+                type="cluster-member"
+                value={values.name}
+                to={`/ui/cluster/group/${values.name}`}
+              />{" "}
```
edlerd, line 96: A cluster group is a defined set of cluster members, not a cluster member itself. I think we do need a different resource type for it. Maybe also a different icon?

src/pages/images/actions/UploadCustomIsoBtn.tsx
```diff
@@ -20,7 +21,8 @@ const UploadCustomIsoBtn: FC<Props> = ({ className }) => {
   const handleFinish = (name: string) => {
     toastNotify.success(
       <>
-        Image <b>{name}</b> uploaded successfully
+        Image <ResourceLabel type="iso-volume" value={name} /> uploaded
```
edlerd, line 24:
```suggestion
        Custom ISO <ResourceLabel type="iso-volume" value={name} /> uploaded
```
I think this was a bug? Maybe check the rest of the file, it might be wrong in errors as well.

src/pages/instances/CreateInstance.tsx
```diff
@@ -172,7 +182,7 @@ const CreateInstance: FC = () => {
   const notifyCreated = (instanceLink: ReactNode, message?: ReactNode) => {
     toastNotify.success(
       <>
-        Created instance {instanceLink}.{message}
+        Created instance {instanceLink}. {message}
```
edlerd, line 185:
```suggestion
        Created instance {instanceLink}.{message}
```
There shouldn't be a dangling space here. Message might be null. The only case where it is set, it is wrapped with a p tag, so will jump into another line anyhow.

src/util/instances.tsx
```diff
@@ -1,34 +1,47 @@
...
+  const { project, instanceName, instanceType } = args;
   return (
-    <InstanceLink instance={{ name: instanceName, project: project || "" }} />
+    <ResourceLink
+      type={instanceType}
+      value={instanceName}
+      to={`/ui/project/${project}/instance/${instanceName}`}
+    />
   );
 };
```
edlerd, line 26: I think we should continue using instance link component (see below).

src/pages/instances/actions/AttachIsoBtn.tsx
```diff
@@ -46,14 +47,19 @@ const AttachIsoBtn: FC<Props> = ({ instance }) => {
...
           () =>
             toastNotify.success(
               <>
-                ISO <b>{attachedIso?.source ?? ""}</b> detached from{" "}
-                {instanceLink}
+                ISO{" "}
+                <ResourceLabel
+                  type="iso-volume"
+                  value={attachedIso?.source ?? ""}
+                />{" "}
```
edlerd, line 61: Should be bold after merging the new resource label. Applies to probably all other cases where we use the label in a notification.

src/util/instanceStart.tsx
```diff
@@ -34,25 +33,27 @@ export const useInstanceStart = (instance: LxdInstance) => {
     instanceLoading.setLoading(instance, "Starting");
     const mutation =
       instance.status === "Frozen" ? unfreezeInstance : startInstance;
+
+    const instanceLink = (
+      <ResourceLink
+        type={instance.type}
+        value={instance.name}
+        to={`/ui/project/${instance.project}/instance/${instance.name}`}
+      />
```
edlerd, line 42: I think we should keep the old `<InstanceLink instance={instance} />` helper to avoid repeating this bit over and over again.

edlerd (commented):
Managed to only get through ~25% of the files yet, but already found some ideas for simplification.

src/pages/cluster/ClusterGroupForm.tsx
mas-who: I think in this case we can just use the cluster icon instead? Will confirm with @piperdeck

src/pages/instances/actions/AttachIsoBtn.tsx
mas-who: yes exactly, that was the first task after I rebased main :)

src/util/instanceStart.tsx
mas-who: Yeah agreed :+1: `InstanceLink` also gets used in other UI components, would you be fine if this PR introduced bits of non-notification related changes? Alternatively, I could create another component `InstanceLinkChip`, and for other PRs I will replace `InstanceLink` with it until all updates are done?

mas-who: Going with a new `InstanceLinkChip` component. I don't want to introduce non-notification related changes into this PR, and it would be simple enough to replace the components later.

edlerd: Yeah, you are right, we want to keep instance link for the instance table use case anyway long term as is. Better introduce a new one.

src/pages/cluster/ClusterGroupForm.tsx
mas-who: Mapped cluster groups to the `cluster-host` icon for now as discussed.

src/pages/instances/CreateInstance.tsx
```diff
@@ -278,6 +293,7 @@ const CreateInstance: FC = () => {
     // we still need to keep the old way of create and start instances since users may be using an older version of lxd
     const instancePayload = {
       ...instance,
+      type: values.instanceType,
```
edlerd, line 296: Why add the type here? Can't we access `values.instanceType` below directly?

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -58,6 +57,7 @@ const InstanceDetailHeader: FC<Props> = ({
           const instanceLink = instanceLinkFromName({
             instanceName: values.name,
             project,
+            instanceType: instance?.type || "container",
```
edlerd, line 60:
```suggestion
            instanceType: instance?.type || "container",
            project,
```
Better keep this ordered?

src/pages/instances/forms/CreateImageFromInstanceSnapshotForm.tsx
```diff
@@ -30,14 +31,17 @@ const CreateImageFromInstanceSnapshotForm: FC<Props> = ({
   const eventQueue = useEventQueue();
   const toastNotify = useToastNotification();
   const queryClient = useQueryClient();
+  const snapshotLabel = (
+    <ResourceLabel bold type="snapshot" value={snapshot.name} />
+  );
 
   const notifySuccess = () => {
     const created = (
       <Link to={`/ui/project/${instance.project}/images`}>created</Link>
```
edlerd, line 40: Shouldn't we use the image chip with its name here?

src/pages/instances/forms/CreateImageFromInstanceSnapshotForm.tsx
```diff
@@ -30,14 +31,17 @@ const CreateImageFromInstanceSnapshotForm: FC<Props> = ({
   const eventQueue = useEventQueue();
   const toastNotify = useToastNotification();
   const queryClient = useQueryClient();
+  const snapshotLabel = (
+    <ResourceLabel bold type="snapshot" value={snapshot.name} />
```
edlerd, line 35: I think this should be a chip with a link to instance detail page, snapshots section.

src/pages/instances/forms/CreateInstanceSnapshotForm.tsx
```diff
@@ -72,8 +72,15 @@ const CreateInstanceSnapshotForm: FC<Props> = ({
               });
               onSuccess(
                 <>
-                  Snapshot <ItemName item={values} bold /> created for instance{" "}
-                  {instance.name}.
+                  Snapshot{" "}
+                  <ResourceLabel bold type="snapshot" value={values.name} />{" "}
+                  created for instance{" "}
+                  <ResourceLabel
```
edlerd, line 78: Why not a link for instance and snapshot?

src/pages/instances/forms/EditInstanceSnapshotForm.tsx
```diff
@@ -43,8 +43,9 @@ const EditInstanceSnapshotForm: FC<Props> = ({
     });
     onSuccess(
       <>
-        Snapshot <ItemName item={{ name: name }} bold /> saved for instance{" "}
-        {instance.name}.
+        Snapshot <ResourceLabel bold type="snapshot" value={name} /> saved for
+        instance{" "}
+        <ResourceLabel bold type={instance.type} value={instance.name} />.
```
edlerd, line 48: Better using links here.

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -60,19 +62,18 @@ const UploadInstanceBackupFileForm: FC<Props> = ({
   const { hasInstanceImportConversion } = useSupportedFeatures();
 
   const handleSuccess = (instanceName: string) => {
+    const url = `/ui/project/${project?.name}/instance/${instanceName}`;
     const message = (
       <>
-        Created instance <strong>{instanceName}</strong>.
+        Created instance{" "}
+        <ResourceLink type="virtual-machine" value={instanceName} to={url} />.
```
edlerd, line 69: This can be a container as well

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -107,7 +108,7 @@ const UploadInstanceBackupFileForm: FC<Props> = ({
         toastNotify.info(
           <>
             Upload completed. Now creating instance{" "}
-            <strong>{values.name}</strong>.
+            <ResourceLabel bold type="virtual-machine" value={values.name} />.
```
edlerd, line 111: type can be container

src/pages/permissions/actions/DeleteIdpGroupsModal.tsx
```diff
@@ -25,9 +26,15 @@ const DeleteIdpGroupsModal: FC<Props> = ({ idpGroups, close }) => {
       ? deleteIdpGroup(idpGroups[0].name)
       : deleteIdpGroups(idpGroups.map((group) => group.name));
 
-    const successMessage = hasOneGroup
-      ? `IDP group ${idpGroups[0].name} deleted.`
-      : `${idpGroups.length} IDP groups deleted.`;
+    const successMessage = hasOneGroup ? (
+      <>
+        IDP group{" "}
+        <ResourceLabel bold type="auth-group" value={idpGroups[0].name} />{" "}
```
edlerd, line 32: I think this should be a different type than auth-group.

src/pages/permissions/panels/CreateGroupPanel.tsx
edlerd, line 1: We are missing the error cases, I think also in other files, we should ensure we update the error cases as well.

src/pages/projects/CreateProject.tsx
```diff
@@ -111,7 +112,12 @@ const CreateProject: FC = () => {
       )
         .then(() => {
           navigate(`/ui/project/${values.name}/instances`);
-          toastNotify.success(`Project ${values.name} created.`);
+          toastNotify.success(
+            <>
+              Project <ResourceLabel bold type="project" value={values.name} />{" "}
```
edlerd, line 117: Should this be a link?

edlerd (changes requested):
Managed ~50% of the changes. Some comments below, I think there are two changes needed with a bigger scope:
1. Update the error case as well, in several places only the success case was updated
2. Apply links wherever possible. There are several components, that use a label, where we should use the link component. To ease this process, we should probably create helpers similar to the `InstanceLinkChip` component.

src/pages/instances/forms/CreateImageFromInstanceSnapshotForm.tsx
edlerd: Meaning a resource link for the image itself.

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
edlerd: Meaning the type assumption virtual-machine is not correct, this needs to be dynamic I think.

src/pages/permissions/actions/DeleteIdpGroupsModal.tsx
edlerd: Shall we introduce a new `idp-group` resource type?

src/pages/instances/forms/CreateImageFromInstanceSnapshotForm.tsx
mas-who: I might be wrong since I don't have the complete context for this feature, looking at the code though I think this was done before because we don't have a way to get the image name at this point?

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
mas-who: Potentially, we can also chain an API call to a successful event and fetch the instance based on the name returned by the operation.

src/pages/permissions/actions/DeleteIdpGroupsModal.tsx
mas-who: sounds good, done. Both `auth-group` and `idp-group` still map to the same icon though.

src/pages/projects/CreateProject.tsx
mas-who: I think it could be, do we link to the instance list for that project?

mas-who: perfect, I'm doing a project wide scan again to see if I can convert labels into links. So this will be part of that change :+1:

src/pages/permissions/panels/CreateGroupPanel.tsx
edlerd: I think we should do the same as in src/pages/instances/CreateInstance.tsx:164 adding the chip with the link as a 3rd argument into the body.

mas-who: Okay that aligns with what I suggested, that implementation concatenates the chip with the error message

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
edlerd: Yeah, I think a generic instance type would be fine here and in the other places.

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
mas-who: will resolve with generic `instance` type here

src/pages/instances/forms/CreateImageFromInstanceSnapshotForm.tsx
edlerd: Right, neither the form inputs nor the post response, nor the operation response contain the image description. Then let's keep it as simple as it is now.

src/pages/permissions/panels/CreateGroupPanel.tsx
edlerd: image

mas-who: yeah, `toastNotify.failure` uses the [messageBuilder](https://github.com/canonical/react-components/blob/d5ff954caa342ecbc0c141ed5a6d611c59cacf46/src/components/NotificationProvider/messageBuilder.tsx#L32) from react-components. The function concatenates the error message and the additional message input (in this case it would be the `instanceLink`.

mas-who:
@edlerd I've addressed the comments as well as:

1. Converted `ResourceLabel` to `ResourceLink` where appropriate.
2. Added links/labels to error messages, but only for the toast notifications. Thought about adding links to inline notifications, but I think that is not necessary since inline notifications are directly related to the context of a resource's detail page.

src/util/operations.tsx
```diff
@@ -13,6 +13,26 @@ export const getInstanceName = (operation?: LxdOperation): string => {
...
+    return instanceSnapshots[0].split("/")[5].split("?")[0];
+  }
+
+  if (storageVolumeSnapshots.length) {
+    return storageVolumeSnapshots[0].split("/")[8].split("?")[0];
+  }
+
+  return "";
+};
+
```
edlerd, line 35: please add a spec test for this one, similar to the other functions in this file.

src/util/instanceStart.tsx
edlerd: While the helper was introduced, this bit is not using it yet. Along with a couple of other places.

src/pages/projects/ProjectConfigurationHeader.tsx
```diff
@@ -47,28 +48,44 @@ const ProjectConfigurationHeader: FC<Props> = ({ project }) => {
         formik.setSubmitting(false);
         return;
       }
+      const projectLink = (
+        <ResourceLink
+          type="project"
+          value={values.name}
+          to={`/ui/project/${project.name}/configuration`}
```
edlerd, line 55: This is confusing, because it is linking to the old url. Maybe give the variable holding it a better name?

src/util/operations.tsx
edlerd: Also, it might be better to split this by instance and volume snapshots. Because the logic has no relevant overlap and the use cases are distinct.

src/pages/instances/SnapshotLinkChip.tsx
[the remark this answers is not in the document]
mas-who: great, done :+1:

src/util/operations.tsx
mas-who: sounds good, will do both. Let's stay consistent with the pattern to have separate functions/components for instance and volume snapshots

src/types/helpers.d.ts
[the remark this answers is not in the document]
mas-who: thanks this is good advice! :+1:

src/pages/projects/ProjectConfigurationHeader.tsx
mas-who: yeah you are right, calling it `oldProjectLink`.

edlerd (changes requested):
Reviewed all the files, some concerns below.

src/pages/storage/forms/EditStorageVolume.tsx
```diff
@@ -62,7 +63,17 @@ const EditStorageVolume: FC<Props> = ({ volume }) => {
               saveVolume.name,
             ],
           });
-          toastNotify.success(`Storage volume ${saveVolume.name} updated.`);
+          toastNotify.success(
+            <>
+              Storage volume{" "}
+              <ResourceLink
+                type="profile"
```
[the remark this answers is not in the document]
mas-who: Thanks for the spot, I did a full check across the entire codebase again, seems other places are fine.

src/types/instance.d.ts
```diff
@@ -102,6 +102,6 @@ export interface LxdInstance {
   state?: LxdInstanceState;
   stateful: boolean;
   status: LxdInstanceStatus;
-  type: string;
+  type: "container" | "virtual-machine" | "instance";
```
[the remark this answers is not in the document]
mas-who: removed, instead, I created a new `InstanceIconType` that gets used for these function definitions

src/util/instances.tsx
```diff
@@ -1,34 +1,30 @@
...
-  const { operation, project } = args;
+  const { operation, project, instanceType } = args;
   const linkText = getInstanceName(operation?.metadata);
   if (!linkText) {
     return;
   }
-  return <InstanceLink instance={{ name: linkText, project: project || "" }} />;
+  return (
+    <InstanceLinkChip
+      instance={{ name: linkText, project: project || "", type: instanceType }}
```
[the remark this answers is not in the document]
mas-who: fixed both places to default to "default"

src/pages/instances/forms/DuplicateInstanceForm.tsx
```diff
@@ -64,20 +66,22 @@ const DuplicateInstanceForm: FC<Props> = ({ instance, close }) => {
...
+  const notifySuccess = (
+    name: string,
+    project: string,
+    type: InstanceIconType,
+  ) => {
+    const url = `/ui/project/${project}/instance/${name}/configuration`;
     const message = (
       <>
-        Created instance <strong>{instanceName}</strong>.
+        Created instance <ResourceLink type={type} value={name} to={url} />.
```
edlerd, line 77: The chip should direct to the overview section. Maybe check other places, there might be more outliers like this one where we reuse the url.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -53,17 +53,18 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
   });
 
   const notifySuccess = (volumeName: string, project: string, pool: string) => {
+    const url = `/ui/project/${project}/storage/pool/${pool}/volumes/custom/${volumeName}/configuration`;
     const message = (
       <>
-        Created volume <strong>{volumeName}</strong>.
+        Created volume{" "}
+        <ResourceLink type="volume" value={volumeName} to={url} />.
```
edlerd, line 60: Here we also link the config tab, where the target url for the chip should be the overview page.

edlerd (changes requested):
Probably one last concern found below.

src/pages/instances/forms/DuplicateInstanceForm.tsx
mas-who: I remember this was discussed when we did the duplicate features, since the user can't configure the instance in the duplicate form, we want to create a notification that directs them to the config page. This is also indicated by the "Configure" action associated to the notification. I think that still makes sense and I'm not sure if we should now redirect to overview page?

edlerd: I am fine with the configure link going to the configure page. But I think the chips should always go to the overview page, so the user is confirmed in the mental modal of how the chips work.

src/pages/storage/forms/DuplicateVolumeForm.tsx
mas-who: resolved with the previous comment

edlerd (approved):
LGTM, thanks for the many iterations to drive this to perfection!

edlerd (approved):

PR #952, opened 2024-10-14
feat: Add resource link and label components [WD-15623]

mas-who:
## Done

- Added `ResourceLink` component for displaying lxd resources that have valid UI link
- Added `ResourceLabel` component for displaying lxd resources that does not have valid UI link 

## QA

1. Run the LXD-UI:
    - N/A


mas-who:
@edlerd I have split out just the resource link and label components from #946 here. If we can get this merged first, then I can work on applying these components to different UI areas in separate PRs.

src/sass/styles.scss
```diff
@@ -330,3 +343,29 @@ body {
...
+
+.resource-label {
+  // align font size with chip component
+  font-size: 0.875rem;
+  line-height: 1.25rem;
+
+  i {
+    margin-right: $sph--x-small;
+  }
+}
```
edlerd, line 371: Better extract those into dedicated files.

src/util/resourceIcons.tsx
edlerd, line 1: How about adding a ResourceIcon component, that contains these lists and renders an icon given a ResourceIconType as a prop?

src/sass/styles.scss
```diff
@@ -330,3 +343,29 @@ body {
...
+
+  .p-chip__value {
+    max-width: 8rem;
+  }
+}
+
+.resource-label {
+  // align font size with chip component
+  font-size: 0.875rem;
+  line-height: 1.25rem;
```
edlerd, line 366: I think there should be vanilla variables for this

src/components/ResoruceIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  certificate: "certificate",
+  "auth-group": "user-group",
+  device: "units",
+};
+
+interface Props {
+  type: ResourceIconType;
+}
+
+const ResoruceIcon: FC<Props> = ({ type }) => {
```
edlerd, line 43:
```suggestion
const ResourceIcon: FC<Props> = ({ type }) => {
```
typo

src/components/ResoruceIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  | "pool"
+  | "volume"
+  | "iso-volume"
+  | "image"
+  | "oidc-identity"
+  | "certificate"
+  | "auth-group"
+  | "device";
+
+export const resourceIcons: Record<ResourceIconType, string> = {
```
edlerd, line 21:
```suggestion
const resourceIcons: Record<ResourceIconType, string> = {
```
I think we should keep the map a private concept and not export it.

src/components/ResoruceIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  "auth-group": "user-group",
+  device: "units",
+};
+
+interface Props {
+  type: ResourceIconType;
+}
+
+const ResoruceIcon: FC<Props> = ({ type }) => {
+  return <Icon name={resourceIcons[type]} />;
```
edlerd, line 44: Don't we need to render null if the lookup fails?

edlerd (changes requested):

edlerd:
I can't QA this currently, maybe add two or three usages in this PR so we can see the components live?

src/components/ResoruceIcon.tsx
mas-who: I was doing it previously, but it shouldn't be necessary since we have strict type check for the `type` prop. Not too long ago, we implemented `tsc` for the `lint-js` script. So CI will fail if someone ignored the type error message.

mas-who:
> I can't QA this currently, maybe add two or three usages in this PR so we can see the components live?

I added an extra commit which I will revert after QA. Applied both label and link components on the instance detail overview page.

src/components/ResourceIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  certificate: "certificate",
+  "auth-group": "user-group",
+  device: "units",
+};
+
+interface Props {
+  type: ResourceIconType;
+}
+
+const ResoruceIcon: FC<Props> = ({ type }) => {
```
edlerd, line 43:
```suggestion
const ResourceIcon: FC<Props> = ({ type }) => {
```
still a typo :)

edlerd (changes requested):

src/components/ResourceIcon.tsx
mas-who: The only one I didn't change is the original one you mentioned... :sweat_smile:

mas-who:
@edlerd after discussing with @piperdeck , we will go with non-bolded text for resource labels displayed as standalone text. I added a `bold` optional prop to the `ResourceLabel` component for this.

mas-who:
@edlerd after a discussion with @piperdeck and also with Amy's input, we decided to use the `pod` icon for instance resources (VM or container). @piperdeck will submit request for new icons so that we can later on use them to differentiate between containers and VMs. In code, I kept both `container` and `virtual-machine` resource types and mapped them to the same `pod` icon. Once the new icons are ready, we can do a simple drop in replacement.

edlerd (approved):
QA and code LGTM 
very much looking forward to having this applied widely :)

mas-who:
@edlerd thanks for the review, just removed the QA implementation, so need another approval :)

edlerd (approved):

PR #953, opened 2024-10-15
Prevent local volume attach to instances on another cluster member

edlerd:
## Done

- Prevent local volume attach to instances on another cluster member
- disable volume selection on instance creation unless the target member is specified and the volume location match
- always allow selection of remote volumes
- handle target of local storage pools, when creating them on instance creation / instance edit.

## QA

1. Run the LXD-UI:
    - run clustered lxd with local volumes and remote volumes
    - attach disk volume on instance creation. any local volumes should be disabled
    - attach disk volume on instance creation. any local volumes should be enabled that match location after selecting a target member for the instance
    - attach disk volume on instance creation. table should display location
    - attach disk volume on instance creation in non-clustered environment: the location column should be hidden
    - create a disk volume on instance creation on a local storage pool, with a target for the instance selected. The volume should be created on the same target as chosen for the instance
    - create a disk volume on instance creation on a local storage pool, with a target for the instance *not* selected. The volume creation should be blocked for any local pool. It should succeed for remote pools.

### DIFF
--- a/src/pages/storage/CustomVolumeCreateModal.tsx
+++ b/src/pages/storage/CustomVolumeCreateModal.tsx
@@ -5,30 +5,40 @@ import {
   volumeFormToPayload,
 } from "pages/storage/forms/StorageVolumeForm";
 import { useFormik } from "formik";
-import { createStorageVolume } from "api/storage-pools";
+import { createStorageVolume, fetchStoragePools } from "api/storage-pools";
 import { queryKeys } from "util/queryKeys";
 import * as Yup from "yup";
-import { useQueryClient } from "@tanstack/react-query";
+import { useQuery, useQueryClient } from "@tanstack/react-query";
 import StorageVolumeFormMain from "pages/storage/forms/StorageVolumeFormMain";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import useEventListener from "@use-it/event-listener";
 import { testDuplicateStorageVolumeName } from "util/storageVolume";
 import { LxdStorageVolume } from "types/storage";
+import { useSettings } from "context/useSettings";
 
 interface Props {
   project: string;
+  instanceLocation?: string;
   onCancel: () => void;
   onFinish: (volume: LxdStorageVolume) => void;
 }
 
 const CustomVolumeCreateModal: FC<Props> = ({
   project,
+  instanceLocation,
   onCancel,
   onFinish,
 }) => {
   const notify = useNotify();
   const queryClient = useQueryClient();
   const controllerState = useState<AbortController | null>(null);
+
+  const { data: settings } = useSettings();
+
+  const { data: pools = [] } = useQuery({
+    queryKey: [queryKeys.storage],
+    queryFn: () => fetchStoragePools(project),
+  });
 
   const StorageVolumeSchema = Yup.object().shape({
     name: Yup.string()
@@ -37,6 +47,14 @@ const CustomVolumeCreateModal: FC<Props> = ({
       )
       .required("This field is required"),
   });
+
+  const isLocalPool = (poolName: string) => {
+    const pool = pools.find((pool) => pool.name === poolName);
+    const driverDetails = settings?.environment?.storage_supported_drivers.find(
+      (driver) => driver.Name === pool?.driver,
+    );
+    return !driverDetails?.Remote ?? false;
+  };
 
   const formik = useFormik<StorageVolumeFormValues>({
     initialValues: {
@@ -53,7 +71,9 @@ const CustomVolumeCreateModal: FC<Props> = ({
     validationSchema: StorageVolumeSchema,
     onSubmit: (values) => {
       const volume = volumeFormToPayload(values, project);
-      createStorageVolume(values.pool, project, volume)
+      const target = isLocalPool(values.pool) ? instanceLocation : undefined;
+
+      createStorageVolume(values.pool, project, volume, target)
         .then(() => {
           void queryClient.invalidateQueries({
             queryKey: [queryKeys.storage],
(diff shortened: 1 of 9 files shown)


mas-who:
Some QA issues:

1. When creating a new volume from the selection modal, after successful creation, the volume appears with a green select button. Is this the intended behaviour?
Screenshot from 2024-10-15 16-52-57

2. This is probably outside of the scope of the current PR. After an instance is created with a volume attached, we can then migrate the instance to a different cluster member. The operation succeeds, but when attempting to start the instance, an error is returned from the server stating that the attached volume cannot be found. Shall we block migration in the UI in this scenario, or maybe report this as an issue?

3. When duplicating an instance, we can select a target cluster member. If the cluster member does not correspond to the attached volume location, then the "volume not found" error is returned. I think we should block this in the UI in this case.

src/pages/storage/CustomVolumeSelectModal.tsx
```diff
@@ -51,20 +62,46 @@ const CustomVolumeSelectModal: FC<Props> = ({
...
-          const selectVolume = () => handleSelect(volume);
+          const isLocalVolume = !!volume.location; // if location is set, it's a local volume, otherwise a remote volume
+          const isDisabled =
+            isClustered &&
+            isInstance &&
+            isLocalVolume &&
+            instanceLocation !== volume.location;
+
+          const disableReason = isDisabled
+            ? `Mismatch of instance location (${instanceLocation}) and local volume location (${volume.location}). `
```
mas-who, line 90: maybe a bit easier to understand?
```suggestion
            ? `Instance location (${instanceLocation}) does not match local volume location (${volume.location}). `
```

edlerd: good idea.

edlerd:
Good catches on those three:

> * When creating a new volume from the selection modal, after successful creation, the volume appears with a green select button. Is this the intended behaviour? Also, when creating a new volume at this point, I assume the user intends to attach that volume to the instance. I think it could be better to create the volume on a specific cluster member if target location is set for the instance? (this is relevant for both instance creation and edit scenarios)

I think I found a way around this. On volume creation during instance creation, do not allow picking a local pool, unless the instance target is specified. Remote pools are always fine. You can check it with the recent changes.

> * This is probably outside of the scope of the current PR. After an instance is created with a volume attached, we can then migrate the instance to a different cluster member. The operation succeeds, but when attempting to start the instance, an error is returned from the server stating that the attached volume cannot be found. Shall we block migration in the UI in this scenario, or maybe report this as an issue?

We should not disable migration generally, because the attached volume might be on a remote pool. If it is on a local pool, I think LXD should trigger a migration of the volume internally. Something to discuss with the core team. Maybe file a bug for us to remember this -- or continue the discussion with the core team -- or both :)

> * When duplicating an instance, we can select a target cluster member. In this case, if the instance have any attached volumes, then the "volume not found" error will be returned. I think we should disable instance duplication if attached volumes are found (for lxd clusters only).

Similar to the 2nd one, due to the remote pool case, we should not prevent duplication in general. I think this is on the radar of the core team. We might want lxd to duplicate any attached volumes silently when launching an instance duplication. Food for thought and discussion. Similar to 2.

src/pages/storage/CustomVolumeCreateModal.tsx
```diff
@@ -5,31 +5,41 @@ import {
...
 import StorageVolumeFormMain from "pages/storage/forms/StorageVolumeFormMain";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import useEventListener from "@use-it/event-listener";
 import { testDuplicateStorageVolumeName } from "util/storageVolume";
 import { LxdStorageVolume } from "types/storage";
+import { useSettings } from "context/useSettings";
 
 interface Props {
   project: string;
+  instanceLocation: string | undefined;
```
mas-who, line 21:
```suggestion
  instanceLocation?: string;
```

src/pages/storage/CustomVolumeSelectModal.tsx
```diff
@@ -9,10 +9,12 @@ import { LxdStorageVolume } from "types/storage";
 import NotificationRow from "components/NotificationRow";
 import { renderContentType } from "util/storageVolume";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
+import classnames from "classnames";
 
 interface Props {
   project: string;
   primaryVolume?: LxdStorageVolume;
+  instanceLocation: string | undefined;
```
mas-who, line 17:
```suggestion
  instanceLocation?: string;
```

mas-who:
Thanks for the update, QA and code LGTM, just two very small nits :)

> I think I found a way around this. On volume creation during instance creation, do not allow picking a local pool, unless the instance target is specified. Remote pools are always fine. You can check it with the recent changes.

Yeah I think that's a good workaround, the error message is also clear :+1: 
 
> We should not disable migration generally, because the attached volume might be on a remote pool. If it is on a local pool, I think LXD should trigger a migration of the volume internally. Something to discuss with the core team. Maybe file a bug for us to remember this -- or continue the discussion with the core team -- or both :)

> Similar to the 2nd one, due to the remote pool case, we should not prevent duplication in general. I think this is on the radar of the core team. We might want lxd to duplicate any attached volumes silently when launching an instance duplication. Food for thought and discussion. Similar to 2.

I think on the above points, I'd like to test how the CLI behaves first before filing a bug. If the CLI is preventing these operations from happening, then I think the UI should implement the same in the interim. Regardless of the outcome, I think it'd be good to have a discussion around how the API endpoints work. I think in the context of this PR, there is no need to worry about the above behaviour.

mas-who (approved):

mas-who:
LGTM! Thanks for the fix

PR #954, opened 2024-10-16
Remove project parameter from storage pool api

edlerd:
## Done

- Remove project parameter from storage pool fetch as pools are global over all projects

## QA

1. Run the LXD-UI:
    - Check list and selectors of storage pool in various places in the app (storage pool list, custom volume creation, instance root device override, profile root disk selector, duplicate instance modal, etc)

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -17,25 +17,20 @@ import { LxdClusterMember } from "types/cluster";
 
 export const fetchStoragePool = (
   pool: string,
-  project: string,
   target?: string,
 ): Promise<LxdStoragePool> => {
   return new Promise((resolve, reject) => {
     const targetParam = target ? `&target=${target}` : "";
-    fetch(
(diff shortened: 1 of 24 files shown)


src/api/storage-pools.tsx
```diff
@@ -31,11 +31,9 @@ export const fetchStoragePool = (
   });
 };
 
-export const fetchStoragePools = (
-  project: string,
-): Promise<LxdStoragePool[]> => {
+export const fetchStoragePools = (): Promise<LxdStoragePool[]> => {
```
mas-who, line 34: wouldn't `fetchStoragePool` above also have the same issue? I.e. we should remove the project query param

mas-who (commented):
QA and code looks good, just one question about fetching single storage pool case

src/api/storage-pools.tsx
edlerd: Good point, actually all the pool API functions should have the project parameter removed. Amended the changes just now.

mas-who (approved):

PR #955, opened 2024-10-16
feat: semiotics update overview pages [WD-15624]

mas-who:
## Done

- Updated all resource detail overview pages with new icons i.e. instance overview, profile overview, network overview, storage pool overview, custom volume overview
- Adjusted instance overview devices section to hide `none` devices and show non-root disk devices.

## QA

1. Run the LXD-UI:
    - Go through instance overview, profile overview, network overview and storage pool overview and custom volume overview pages and make sure everything is working as expected.


src/components/ResourceLink.tsx
```diff
@@ -11,7 +11,7 @@ interface Props {
 const ResourceLink: FC<Props> = ({ type, value, to }) => {
   return (
     <Link
-      className="p-chip is-inline is-dense resource-link"
+      className={"p-chip is-inline is-dense resource-link"}
```
edlerd, line 14: Why? :grin:

src/pages/instances/InstanceOverview.tsx
```diff
@@ -84,19 +94,25 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
...
+                  )}
                 </td>
               </tr>
               <tr>
                 <th className="u-text--muted">Root storage pool</th>
                 <td>
-                  <Link
+                  <ResourceLink
+                    type="pool"
+                    value={rootPool}
```
edlerd, line 113: I think we should include the pool under devices below. Maybe mark it as type "disk (root)" or alike. But it seems wrong to have this separated out in the general section. Because in configuration, we treat the root disk like the other devices, and conceptually it is a device. So let's move it down?

src/pages/profiles/ProfileNetworkList.tsx
```diff
@@ -2,25 +2,27 @@ import { FC } from "react";
...
-                title={`Network ${device.network}`}
-              >
-                {device.network}
+              <div key={device.network}>
+                <ResourceLink
+                  type="network"
+                  value={device.network}
+                  to={`/ui/project/${project}/network/${device.network}`}
+                />
               </div>
```
edlerd, line 26: Can we strip the wrapping div? I think the networks can be displayed inline.

src/pages/profiles/ProfileStorageList.tsx
```diff
@@ -2,25 +2,27 @@ import { FC } from "react";
...
           items={Object.values(profile.devices)
             .filter(isDiskDevice)
             .map((device) => (
-              <div
-                key={device.path}
-                className="u-truncate list-item"
-                title={`Pool ${device.pool}`}
-              >
-                {device.pool}
+              <div key={device.path}>
```
edlerd, line 20: Strip the div?

src/pages/storage/StorageUsedBy.tsx
```diff
@@ -36,10 +34,13 @@ const StorageUsedBy: FC<Props> = ({ storage, project }) => {
...
               items={data[INSTANCES].map((item) => (
-                <div key={`${item.name}-${item.project}`}>
-                  <InstanceLink instance={item} />
-                  {item.project !== project && ` (project ${item.project})`}
-                </div>
+                <UsedByItem
+                  key={`${item.name}-${item.project}`}
+                  item={item}
+                  project={project}
+                  type="virtual-machine"
```
edlerd, line 41: Can be a container or vm.

src/components/UsedByItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  if (type === "snapshot") {
+    if (item.instance) {
+      label = `${item.instance} ${item.name}`;
+    }
+
+    if (item.volume) {
+      label = `${item.volume} ${item.name}`;
+    }
```
edlerd, line 35: I think we should follow the same paradigm as with the profile: show the instance (or volume) as a separate chip with a / next to it, to show the snapshot is a nested object.

src/pages/networks/NetworkDetailOverview.tsx
```diff
@@ -153,11 +153,13 @@ const NetworkDetailOverview: FC<Props> = ({ network }) => {
...
-                        <div key={item.name}>
-                          <InstanceLink instance={item} />
-                          {item.project !== project &&
-                            ` (project ${item.project})`}
-                        </div>
+                        <UsedByItem
+                          key={item.name}
+                          item={item}
+                          project={project}
+                          type="virtual-machine"
```
edlerd, line 160: This can also be a container.

edlerd (commented):
This looks very neat, I like it!

Some ideas to further refine this and issues, that seem to need corrections below.

src/components/ResourceLink.tsx
mas-who: was making some changes in here but reverted, seems like I missed some haha

src/pages/instances/InstanceOverview.tsx
mas-who: Yeah sounds good to me :+1:

src/pages/profiles/ProfileNetworkList.tsx
mas-who: I actually way prefer how this looks, I think we utilise the horizontal space efficiently like this. Let's do this across all detail pages?

Screenshot from 2024-10-17 11-48-29
image

mas-who: Never mind about that, the `project`/`resource` structure does not look good with this inline approach.

src/pages/storage/StorageUsedBy.tsx
mas-who: related to comments from #946, we can use a general `instance` type to generate the link here. Alternatively, we could also make an API call at this point to get instance type detail. Prefer the first approach as it's simple and less expensive

src/components/UsedByItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+            value={item.instance || item.volume || ""}
+            to={
+              item.instance
+                ? `/ui/project/${actualProject}/instance/${item.instance}`
+                : `/ui/project/${actualProject}/storage/pool/${item.pool}/volumes/custom/${item.volume}`
+            }
+          />{" "}
+          /{" "}
+        </>
+      )}
```
edlerd, line 55: I think branching out the instance/volume cases on a higher level, would make this more explicit and simpler to read. So basically `{type === "snapshot" && item.instance && (...)` and a second similar case for the volume below.

src/components/UsedByItem.tsx
mas-who: @edlerd do you think there is a bug with snapshots in used by? I can't seem to figure out how to get snapshots to showup in the `used_by` attributes for instances and volumes?

src/pages/profiles/ProfileStorageList.tsx
```diff
@@ -2,26 +2,27 @@ import { FC } from "react";
...
                 key={device.path}
-                className="u-truncate list-item"
-                title={`Pool ${device.pool}`}
-              >
-                {device.pool}
-              </div>
+                type="pool"
+                value={device.pool}
+                to={`/ui/project/${project}/storage/pool/${device.pool}`}
+              />
```
edlerd, line 25: There is a bug here, when assigning storage volumes to a profile, they will appear as storage pool.
I think the right solution is to use the same approach as on the instance overview page. So remove the storage list, and introduce a device list here.

Maybe this can be done with a shared component. If we can share code, it would be great.

This is a bit out of scope for the current task, might be better to create a task and follow up pr for this problem.

edlerd (commented):
This looks so nice, especially the new used by lists. Some small and probably easy simplification ideas below. Two suggestions, but I think those should be follow-ups.

src/components/UsedByItem.tsx
edlerd: Not sure if I understand the question. If you ask why the snapshots don't show up in the used by section of an instance or volume -- I think they shouldn't.

src/pages/instances/InstanceOverviewDevices.tsx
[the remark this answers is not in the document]
mas-who: Sure, do you mean using something like a tooltip?

src/components/UsedByItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+  type: ResourceIconType;
+  to: string;
+}
+
+const UsedByItem: FC<Props> = ({ item, project, type, to }) => {
+  const { data: image } = useQuery({
+    queryKey: [queryKeys.images, item.name],
+    queryFn: () => fetchImage(item.name, project),
+    enabled: type === "image",
+  });
```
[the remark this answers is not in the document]
mas-who: Good idea :+1:

src/pages/instances/InstanceOverviewDevices.tsx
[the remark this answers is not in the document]
edlerd: I was thinking of another generic column in the table. Maybe call it "details". It could contain a gist of the device.

src/components/UsedByItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+import ResourceLink from "./ResourceLink";
+import { LxdUsedBy } from "util/usedBy";
+import { ResourceIconType } from "./ResourceIcon";
+import { useQuery } from "@tanstack/react-query";
+import { fetchImage } from "api/images";
+import { queryKeys } from "util/queryKeys";
+
+interface Props {
+  item: LxdUsedBy;
+  project: string;
```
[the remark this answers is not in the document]
mas-who: I think `activeProject` is more correct here since `parentProject` is actually extracted from `item.project`?

src/components/UsedByItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+    enabled: type === "image",
+  });
+
+  let label = item.name;
+  if (type === "image") {
+    label = image?.properties?.description || item.name;
+  }
+
+  const fromAnotherProject = item.project !== project;
+  const actualProject = fromAnotherProject ? item.project : project;
```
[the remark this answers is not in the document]
mas-who: oh yeah you are right :+1:

src/components/UsedByItem.tsx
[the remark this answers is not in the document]
edlerd: sounds good

src/pages/profiles/ProfileStorageList.tsx
mas-who: Yes, I think let's keep this outside the scope of current PR. At the moment both volume and pool has the same icon so it's still okay, but will create a task for this.

src/components/UsedByItem.tsx
mas-who: So tried with a `dir` storage pool as suggested, still not seeing any snapshot (I have a snapshot created for the instance and  custom volume on the pool shown below). You mentioned snapshots shows up for you in this scenario right?

Screenshot from 2024-10-17 15-56-05

src/components/UsedByItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+    queryFn: () => fetchImageList(activeProject),
+    enabled: type === "image",
+  });
+
+  const image = images.find((image) => image.fingerprint === item.name);
+
+  let label = item.name;
+  if (type === "image") {
+    label = image?.properties?.description || item.name;
+  }
```
edlerd, line 28:
```suggestion
  const label = image?.properties?.description || item.name;
```
Can simplify this to one line.

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -2,15 +2,32 @@ import { FC } from "react";
...
+    .sort((a, b) => {
+      const nameA = a[0].toLowerCase();
+      const nameB = b[0].toLowerCase();
+      const typeA = a[1].type;
+      const typeB = b[1].type;
+      if (nameA === nameB) {
+        return typeA.localeCompare(typeB);
+      }
+
+      return nameA.localeCompare(nameB);
```
edlerd, line 29: I meant the other way around -- first type and then name, so the disk devices are on top. Because I think the disks are the most relevant ones and grouping by type seems better than sorting by name.

edlerd:
Thanks for improving, two very small things left, and the linter is complaining, maybe check how to resolve on that.

src/pages/instances/InstanceOverviewDevices.tsx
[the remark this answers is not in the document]
mas-who: Here's the [issue](https://warthogs.atlassian.net/browse/WD-15994?atlOrigin=eyJpIjoiMzVkZDMwZDFmNDhkNGI3ZmEzNzIyZTM3N2IxNTJjZGUiLCJwIjoiaiJ9)

src/components/UsedByItem.tsx
edlerd: Yes they do, using latest/stable and your branch here:

image

src/pages/profiles/ProfileStorageList.tsx
mas-who: Yeah was just doing it, here's the [issue](https://warthogs.atlassian.net/browse/WD-16007?atlOrigin=eyJpIjoiMjdmN2MwYzBlZTUzNDJlM2E0NjIwODJiNWY2Zjk3ZDQiLCJwIjoiaiJ9)

src/components/UsedByItem.tsx
mas-who: I don't understand..., it's not in the API response for me, I'm using latest/edge though, so maybe I will try latest/stable

mas-who: In any case, I don't think this issue should be a blocker for this PR, but I will play around with this a bit more outside of the discussion here as well

src/components/UsedByItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <div>
+      {item.project !== activeProject && (
```
edlerd, line 32: This check doesn't work, I see the project always there now. Which seems also fine, so maybe just remove the check here?

src/components/UsedByItem.tsx
edlerd: Can confirm on latest/edge the snapshots are missing. Please report this as a bug upstream.

mas-who: Will do! Thanks for checking :)

mas-who: keeping this comment open until the issue is created

src/pages/instances/InstanceOverviewDevices.tsx
mas-who: right, I read that wrong apologies

src/components/UsedByItem.tsx
mas-who: This is my bad, I neglected updating the prop input on this component. Should work now

src/components/UsedByItem.tsx
mas-who: [reported](https://github.com/canonical/lxd/issues/14291)

edlerd (approved):
QA and code LGTM, thanks for the implementation

edlerd (approved):

PR #961, opened 2024-10-20
chore(ci) cancel pr workflow on subsequent pushes to the same branch

edlerd:
## Done

- cancel pr workflow on subsequent pushes to the same branch


mas-who (approved):
LGTM, thanks for adding this. Maybe the PR title can be "cancel active pr workflow on subsequent pushes to the same branch"?

PR #962, opened 2024-10-21
fix(instance) remove read and write limits for custom storage volumes

edlerd:
## Done

- fix(instance) remove read and write limits for custom storage volumes
- the limits are not applied to all possible storage drivers a unit selector is missing in the ui for the limits, we wrongly advertise the number input as iops, but that is not correct, the default seems to be bytes/s when no unit is specified. Removing the limit support from the ui for now to resolve both issues as the feature seems not very important in the first place.


mas-who (approved):
LGTM :+1: Shall we create a ticket to add this feature back later?

edlerd:
> Shall we create a ticket to add this feature back later?

Not at this time, only if we put a focus on this in the core development, we'd create a task for the ui to bring it back.

PR #964, opened 2024-10-22
fix(cache) force cache refetch when lifecycle events come in

edlerd:
## Done

- force cache refetch when lifecycle events come in

## QA

1. Run the LXD-UI:
    - open the list of instances in the ui
    - start a stopped instance in a terminal, keep the ui open in the background
    - ensure the status of the instance in the ui updates automatically without having to change focus back to the ui

### DIFF
--- a/src/pages/instances/Events.tsx
+++ b/src/pages/instances/Events.tsx
@@ -30,6 +30,40 @@ const Events: FC = () => {
     }
   };
 
+  const getLifecycleRootQueryKey = (event: LxdEvent) => {
+    if (event.metadata.action.startsWith("auth-group-")) {
+      return queryKeys.authGroups;
+    }
+    if (event.metadata.action.startsWith("config-")) {
+      return queryKeys.settings;
+    }
+    if (event.metadata.action.startsWith("identity-provider-group-")) {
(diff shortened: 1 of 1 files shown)


edlerd:
> code and QA looks good, although I couldn't replicate the issue on an older branch? The instance status updates even when the UI is not in focus

That is confusing, I can reproduce it:


https://github.com/user-attachments/assets/5f0aef68-b249-4313-b313-e805865c1026

Whereas with this branch the row for the testy instance updates just fine:


https://github.com/user-attachments/assets/e12b8942-86e2-48cb-a019-ba83dfd9ee43

mas-who:
> > code and QA looks good, although I couldn't replicate the issue on an older branch? The instance status updates even when the UI is not in focus
> 
> That is confusing, I can reproduce it from the main branch:
> 
>  Screencast.from.2024-10-22.17-37-09.mp4 
> Whereas with this branch the row for the testy instance updates just fine:
> 
>  Screencast.from.2024-10-22.17-38-25.mp4

nvm! I just replicated the old issue, I think I was on this branch when I tried to do that, sorry my mistake

PR #967, opened 2024-10-22
feat: [WD-15994] Show Instance device details on Overview page.

Kxiru:
## Done

- Added device details to the overview page.

## QA

1. Run the LXD-UI:

## Screenshots

image

### DIFF
--- a/src/pages/instances/InstanceOverviewDeviceDetail.tsx
+++ b/src/pages/instances/InstanceOverviewDeviceDetail.tsx
@@ -1,0 +1,73 @@
+import { FC } from "react";
+import { isRootDisk } from "util/instanceValidation";
+import { FormDevice } from "util/formDevices";
+import { LxdDeviceValue } from "types/device";
+import ResourceLink from "components/ResourceLink";
+
+interface Props {
+  device: LxdDeviceValue;
+  project: string;
+}
+
+const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
+  if (device.type === "disk") {
+    if (isRootDisk(device as FormDevice)) {
+      return (
+        <>
+          <span>pool </span>
+          <ResourceLink
+            type={"pool"}
+            value={device.pool}
+            to={`/ui/project/${project}/storage/pool/${device.pool}`}
+          />
+        </>
+      );
+    }
+
+    return (
+      <>
+        <span>
+          volume&nbsp;
+          <ResourceLink
+            type={"volume"}
+            value={device.source as string}
+            to={`/ui/project/${project}/storage/pool/${device.pool}/volumes/custom/${device.source}`}
+          />{" "}
+        </span>
+        <span> on pool </span>
+        <ResourceLink
+          type={"pool"}
+          value={device.pool}
+          to={`/ui/project/${project}/storage/pool/${device.pool}`}
+        />
+      </>
+    );
+  }
+
+  if (device.type === "gpu") {
+    return (
+      <span title={device.pci} className="u-truncate">
+        {device.pci ? `pci: ${device.pci}` : `id: ${device.id}`}
+      </span>
+    );
+  }
+
+  if (device.type === "proxy") {
+    return (
+      <>
+        <span
+          title={device.listen}
+          className="u-truncate"
+        >{`listen: ${device.listen}`}</span>{" "}
+        <br />
+        <span
+          title={device.connect}
+          className="u-truncate"
+        >{`connect: ${device.connect}`}</span>
+      </>
+    );
+  }
+  return "-";
+};
+
+export default InstanceOverviewDeviceDetail;
--- a/src/pages/instances/InstanceOverviewDevices.tsx
+++ b/src/pages/instances/InstanceOverviewDevices.tsx
@@ -6,6 +6,7 @@ import { isRootDisk } from "util/instanceValidation";
 import { FormDevice } from "util/formDevices";
 import ResourceLink from "components/ResourceLink";
 import { isOtherDevice } from "util/devices";
+import InstanceOverviewDeviceDetail from "./InstanceOverviewDeviceDetail";
 
 interface Props {
   instance: LxdInstance;
@@ -34,6 +35,7 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
   const deviceHeaders = [
     { content: "Name", sortKey: "name", className: "u-text--muted" },
     { content: "Type", sortKey: "type", className: "u-text--muted" },
+    { content: "Details", className: "u-text--muted" },
   ];
 
   const deviceRows = instanceDevices.map(([devicename, device]) => {
@@ -51,6 +53,16 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
           content: `${device.type} ${isRootDisk(device as FormDevice) ? "(root)" : ""}`,
           role: "rowheader",
           "aria-label": "Type",
+        },
+        {
+          content: (
+            <InstanceOverviewDeviceDetail
+              device={device}
+              project={instance.project}
+            />
+          ),
+          role: "rowheader",
+          "aria-label": "Details",
         },
       ],
       sortData: {


src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
           role: "rowheader",
           "aria-label": "Type",
         },
+        {
+          content: (
```
mas-who, line 68: let's move this into a dedicated function to generate the detail column component. I find it difficult to read to have JSX directly embedded in JS code. So we can for example, have a function `getDeviceDetail` that takes in a device, then return some component.

Edit: actually, I think it would be easier to generate the entire detail column object, it would be easier for variable reuse for other column attributes aside from `content`

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -34,8 +35,19 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
   const deviceHeaders = [
     { content: "Name", sortKey: "name", className: "u-text--muted" },
     { content: "Type", sortKey: "type", className: "u-text--muted" },
+    { content: "Details", sortKey: "type", className: "u-text--muted" },
   ];
 
+  const deviceDetailRows = (lines: string[]) => {
```
mas-who, line 41: no need for a function here, this can be assigned to an array

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
           role: "rowheader",
           "aria-label": "Type",
         },
+        {
+          content: (
+            <div>
+              {device.type == "disk" ? (
+                isRootDisk(device as FormDevice) ? (
+                  <>
+                    <strong>Pool: </strong>
```
mas-who, line 73: as discussed, we can use the `ResourceLink` component to link to the storage pool here. The `type` prop for the `ResourceLink` component determines the icon to use

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
...
+                    <>{device.pool}</>
+                  </>
+                ) : (
+                  <>
+                    <strong>Storage Volume:</strong>
```
mas-who, line 78: for non-root disk devices, show two chips displaying storage pool and volume. Use `ResourceLink` for this

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
...
+                      {` ${device.source}`}
+                    </Link>
+                  </>
+                )
+              ) : (
+                ""
+              )}
+              {device.type === "gpu" && (
+                <>
+                  <strong>Pci: </strong>
```
mas-who, line 91: maybe instead of `field: device.value`, we can have `device.value (field)`?

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
...
+                  <>{device.pci}</>
+                </>
+              )}
+              {device.type === "proxy" &&
+                deviceDetailRows([
```
mas-who, line 96: I think it's simpler to just define two `div` elements here instead doing a function that internally is doing a map render.

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
...
+                  `Listen: ${device.listen}`,
+                  `Connect: ${device.connect}`,
+                ])}
+              {device.type === "usb" ? `uid: ${device.uid}` : ""}
```
mas-who, line 100: there is an issue here, `device.uid` may be undefined, in that case maybe just display `-`. Same comment for the other device values, please check all of the other cases

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
...
+              {device.type === "infiniband"
+                ? `Nic type: ${device.nictype}`
+                : ""}
+              {device.type === "pci" ? `Address: ${device.address}` : ""}
+              {device.type === "tpm" ? `Path: ${device.path}` : ""}
+              {device.type === "unix-block" ? `uid: ${device.uid}` : ""}
+              {device.type === "unix-char" ? `uid: ${device.uid}` : ""}
+              {device.type === "unix-hotplug" ? `uid: ${device.uid}` : ""}
+            </div>
```
mas-who, line 109: we should cater for the case that there may be a device type that the UI does not handle, maybe in the future more device types are added to the backend. In that case, we should display "-"

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
           role: "rowheader",
           "aria-label": "Type",
         },
+        {
+          content: (
+            <div>
```
mas-who, line 69: this `div` is not necessary, we can just use a fragment `<></>`

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -34,8 +35,19 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
   const deviceHeaders = [
     { content: "Name", sortKey: "name", className: "u-text--muted" },
     { content: "Type", sortKey: "type", className: "u-text--muted" },
+    { content: "Details", sortKey: "type", className: "u-text--muted" },
```
mas-who, line 38: `sortKey` here should be `details`

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +64,53 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
...
+                : ""}
+              {device.type === "pci" ? `Address: ${device.address}` : ""}
+              {device.type === "tpm" ? `Path: ${device.path}` : ""}
+              {device.type === "unix-block" ? `uid: ${device.uid}` : ""}
+              {device.type === "unix-char" ? `uid: ${device.uid}` : ""}
+              {device.type === "unix-hotplug" ? `uid: ${device.uid}` : ""}
+            </div>
+          ),
+          role: "rowheader",
+          "aria-label": "Details",
```
mas-who, line 112: we should handle the case if a value is too long for the table cell. You should be able to add `className` here with `u-truncate`. Then you would also need to include a `title` attribute to display device value on hover of the cell.

mas-who (commented):
The devices table is looking good with additional info :+1: here's some suggestions for improvements

src/pages/instances/InstanceOverviewDevices.tsx
edlerd: We can also avoid the sort key, so the column is not sortable. It might be hard and not relevant to sort by the details.

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+import ResourceLink from "components/ResourceLink";
+
+interface Props {
+  device: LxdDeviceValue;
+  instance: LxdInstance;
+}
+
+const InstanceOverviewDeviceList: FC<Props> = ({ device, instance }) => {
+  return (
+    <div className="u-truncate">
```
mas-who, line 15: This `div` is not necessary, you can add `className`  attributes to the column object properties i.e. where this component is being called.

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { LxdDeviceValue, LxdGPUDevice } from "types/device";
+import ResourceLink from "components/ResourceLink";
+
+interface Props {
+  device: LxdDeviceValue;
+  instance: LxdInstance;
+}
+
+const InstanceOverviewDeviceList: FC<Props> = ({ device, instance }) => {
+  return (
```
mas-who, line 14: In general I think using ternary is not very readable when it exceeds 2 conditions since it introduces cognitive load for someone reading the code e.g. trying to find the colons. Instead, we can try use the following structure:

```
if (device.type === "disk") {
    if (isRootDisk(device as FormDevice)) {
          return A;
    }
    
     return B;
}

if (device.type === "gpu) {
    return C;
}

etc...
```

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <div className="u-truncate">
+      {device.type == "disk" ? (
+        isRootDisk(device as FormDevice) ? (
+          <div title={`Pool: ${device.pool}`}>{`pool: ${device.pool}`}</div>
```
mas-who, line 18: This should use `ResourceLink` as well.

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+        ) : (
+          <>
+            <div title={`Storage Volume: ${device.source}`}>
```
mas-who, line 21: I think displaying the chips inline next to each other without prefixes is fine? Please can you check with @edlerd on this.

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+                value={device.pool}
+                to={`/ui/project/${instance.project}/storage/pool/${device.pool}`}
+              />
+            </div>
+          </>
+        )
+      ) : (
+        ""
+      )}
+      <div title={`UID: ${(device as LxdGPUDevice).pci}`}>
```
mas-who, line 42: instead of `div`, these should be `span` for better semantic html as it is meant to contain text.

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+        {device.type === "gpu" ? `uid: ${device.pci}` : ""}
```
mas-who, line 43: why are we using `device.pci` for `uid`? Also we should display "-" instead of ""

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <>
+          <div
+            title={`Listen: ${device.listen}`}
+          >{`listen: ${device.listen}`}</div>
+          <div
+            title={`Connect: ${device.connect}`}
+          >{`connect: ${device.connect}`}</div>
+        </>
+      )}
+      {device.type != "gpu" &&
```
mas-who, line 55: if we use the if statement structure for this component as mentioned above, then this can be a simple return statement at the end  of the component function

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -52,6 +54,13 @@ const InstanceOverviewDevices: FC<Props> = ({ instance }) => {
           role: "rowheader",
           "aria-label": "Type",
         },
+        {
+          content: (
+            <InstanceOverviewDeviceList device={device} instance={instance} />
```
mas-who, line 59: I think the component name should rather be `InstanceOverviewDeviceDetail`

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
+import { FC } from "react";
+import { LxdInstance } from "types/instance";
+import { isRootDisk } from "util/instanceValidation";
+import { FormDevice } from "util/formDevices";
+import { LxdDeviceValue, LxdGPUDevice } from "types/device";
+import ResourceLink from "components/ResourceLink";
+
+interface Props {
+  device: LxdDeviceValue;
+  instance: LxdInstance;
```
mas-who, line 10: since we only need the project, why not just pass `project: string` down as a prop instead of the entire instance? This is a general good practice in react, since prop changes causes components to re-render. Trying to limit the prop interface size is an optimisation.

mas-who (commented):
Some more code suggestions

src/pages/instances/InstanceOverviewDeviceList.tsx
mas-who: I think we can use the structure we use in the used by list i.e. `<PoolChip> / <VolumeChip>`

src/pages/instances/InstanceOverviewDeviceList.tsx
```diff
@@ -0,0 +1,63 @@
...
+              />
+            </div>
+            <div title={`Pool: ${device.pool}`}>
+              pool:
+              <ResourceLink
+                type={"pool"}
+                value={device.pool}
+                to={`/ui/project/${instance.project}/storage/pool/${device.pool}`}
+              />
+            </div>
```
edlerd, line 36: Better make this one line "volume (foo) on pool (bar)" where (foo) and (bar) are the chips

src/pages/instances/InstanceOverviewDeviceList.tsx
edlerd: We shoul display the uid if set, otherwise the id if it is set. We should never have an empty case here.

src/pages/instances/InstanceOverviewDeviceList.tsx
edlerd: also we can remove the : in front of the chips.

src/pages/instances/InstanceOverviewDeviceList.tsx
mas-who: see @edlerd comment below about this.

src/pages/instances/InstanceOverviewDeviceList.tsx
mas-who: Just checked the jira issue, for `gpu` device, we agreed to display pci address or id. Then the issue here is the `uid:` prefix for the column value. It should display `pci: ` or `id: ` depending on the value that's set on the gpu device

src/pages/instances/InstanceOverviewDeviceList.tsx
Kxiru: None of the other device types, such as gpu, have more than 2 conditions. Should I still change the ternary operator to an if statement?

Kxiru: As written below, I understand this is mostly due to an easier return for the Other device types.

src/pages/instances/InstanceOverviewDeviceDetail.tsx
```diff
@@ -0,0 +1,71 @@
...
+  device: LxdDeviceValue;
+  project: string;
+}
+
+const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
+  if (device.type === "disk") {
+    if (isRootDisk(device as FormDevice)) {
+      return (
+        <div title={`Pool: ${device.pool}`}>
+          pool
```
mas-who, line 17: add a space here between "pool" and chip please

mas-who (approved):
LGTM :+1:

mas-who (approved):

PR #968, opened 2024-10-23
feat: [WD-16007] Replace storage list with device list on profile det…

Kxiru:
…ail page.

Work done (So far):
- Separate Devices section into Networks and Other devices (Currently storage)
- Created a generic NetworkListTable file to generate a network list for instance and profile overview pages.
- Note, the file ProfileNetworkList still exists for the ProfileDetailPanelContent component.
- Create a generic DeviceListTable file to generate a device list for instance and profile overview pages.

## QA

1. Run the LXD-UI:

## Screenshots

image
image


src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+import ResourceLink from "components/ResourceLink";
+import { isOtherDevice } from "util/devices";
+import InstanceOverviewDeviceDetail from "../pages/instances/InstanceOverviewDeviceDetail";
+import { LxdProfile } from "types/profile";
+import { useParams } from "react-router-dom";
+import { LxdDeviceValue } from "types/device";
+
+interface Props {
+  instance?: LxdInstance;
+  profile?: LxdProfile;
```
edlerd, line 16: This is not a good interface, because both props are mutual exclusive. This leads to a lot of branching in this component. I think better would be to make this component agnostic of instances and profiles. So we pass in only two generic props: `devices` as an array and the `configBaseUrl` as a string.

src/components/NetworkListTable.tsx
```diff
@@ -7,31 +7,41 @@ import { LxdInstance } from "types/instance";
...
 import { isNicDevice } from "util/devices";
 import ResourceLink from "components/ResourceLink";
+import { LxdProfile } from "types/profile";
+import { useParams } from "react-router-dom";
 
 interface Props {
-  instance: LxdInstance;
   onFailure: (title: string, e: unknown) => void;
+  instance?: LxdInstance;
+  profile?: LxdProfile;
```
edlerd, line 16: Here we have the same issue, better find a generic interface that works for both cases in the same way. I think we can pass in devices here and that should work just fine. So we resolve the devices in the caller once, and actually we can reuse that for this component and the devices component.

edlerd (commented):
This looks quite nicely aligned. Some concerns to make the components more generic below.

src/components/DeviceListTable.tsx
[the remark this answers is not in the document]
mas-who: @Kxiru once #969 is merged, you will need to pull the updates to the devices table into this generic component.

src/components/NetworkListTable.tsx
```diff
@@ -7,31 +7,41 @@ import { LxdInstance } from "types/instance";
...
+    queryKey: [queryKeys.projects, overviewType.project, queryKeys.networks],
+    queryFn: () => fetchNetworks(overviewType.project as string),
   });
 
   if (error) {
     onFailure("Loading networks failed", error);
   }
 
-  const instanceNetworks = Object.values(instance.expanded_devices ?? {})
+  const overviewNetworks = Object.values(overviewType.devices ?? {})
```
mas-who, line 40: nit: I think a better name could be `networkDevices`

src/components/NetworkListTable.tsx
```diff
@@ -7,31 +7,41 @@ import { LxdInstance } from "types/instance";
...
   onFailure: (title: string, e: unknown) => void;
+  instance?: LxdInstance;
+  profile?: LxdProfile;
 }
 
-const InstanceOverviewNetworks: FC<Props> = ({ instance, onFailure }) => {
+const NetworkListTable: FC<Props> = ({ instance, profile, onFailure }) => {
+  const { project } = useParams<{ project: string }>();
+
+  const overviewType = {
```
mas-who, line 22: related to @edlerd previous comment, this object will not be needed.

src/components/NetworkListTable.tsx
```diff
@@ -65,7 +75,7 @@ const InstanceOverviewNetworks: FC<Props> = ({ instance, onFailure }) => {
               <ResourceLink
                 type="network"
                 value={network.name}
-                to={`/ui/project/${instance.project}/network/${network.name}`}
+                to={`/ui/project/${overviewType.project}/network/${network.name}`}
               />
             ),
             role: "rowheader",
```
mas-who, line 81: let's change the `role` for these columns to `cell` to improve a11y.

src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const typeA = a[1].type;
+    const typeB = b[1].type;
+    if (typeA === typeB) {
+      return nameA.localeCompare(nameB);
+    }
+
+    return typeA.localeCompare(typeB);
+  };
+
+  if (instance) {
```
mas-who, line 39: `getInstanceDevices` and `getProfileDevices` logic may be delegated to the parent. Leave the filter and sort inside this component

src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+      .sort(compareDevices);
+  } else {
+    profileDevices = getProfileDevices(profile as LxdProfile)
+      .filter(([, device]) => {
+        return device.type !== "nic" && device.type !== "none";
+      })
+      .sort(compareDevices);
+  }
+
+  const overviewType = {
```
mas-who, line 54: `overviewType` is not needed if we take in `devices` as a prop

src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const hasDevices = overviewType.devices.length > 0;
+
+  const deviceHeaders = [
+    { content: "Name", sortKey: "name", className: "u-text--muted" },
+    { content: "Type", sortKey: "type", className: "u-text--muted" },
+    { content: "Details", className: "u-text--muted" },
+  ];
+
+  const deviceRows = overviewType.devices.map(([devicename, device]) => {
+    const instanceConfigURL = `/ui/project/${overviewType.project}/instance/${instance?.name}/configuration/${isOtherDevice(device) ? "other" : device.type}`;
```
mas-who, line 68: related to comment from @edlerd, we just need to append `other` or `device.type` to `configUrlBase`

src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+      columns: [
+        {
+          content: (
+            <ResourceLink
+              type="device"
+              value={devicename}
+              to={instance ? instanceConfigURL : profileConfigURL}
+            />
+          ),
+          role: "rowheader",
```
mas-who, line 81: let's update `role` value to `cell` for all columns

mas-who (commented):
Some comments that extends on David's observations

mas-who:
One more observation from QA.

It seems that the network and device list tables on the instance overview page has row borders but not on the profile overview page. Would you double check the styling applied? i.e. identify differences between `_profile_detail_overview.scss` and `_instance_detail_overview.scss`

src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+        name: devicename.toLowerCase(),
+        type: device.type,
+      },
+    };
+  });
+
+  return (
+    <>
+      {hasDevices && (
+        <MainTable headers={deviceHeaders} rows={deviceRows} sortable />
```
edlerd, line 91: This table is missing the class "devices", probably from a wrong resolution of the merge conflicts? With the class the space among the column should be distributed in a better way, avoiding spread to two lines and aligning with of the first two columns with the tables above/below.

src/components/NetworkListTable.tsx
```diff
@@ -3,35 +3,38 @@ import { useQuery } from "@tanstack/react-query";
...
+    queryKey: [queryKeys.projects, project, queryKeys.networks],
+    queryFn: () => fetchNetworks(project as string),
   });
 
   if (error) {
     onFailure("Loading networks failed", error);
   }
 
-  const instanceNetworks = Object.values(instance.expanded_devices ?? {})
+  const overviewNetworks = Object.values(networkDevices ?? {})
```
edlerd, line 33: There is no concept of overview in this component. Yes, it is used in that context of overview pages. But only for now, and naming of variables should be optimal for the component it lives in. So probably better to call this just "networks" and drop the overview prefix.

src/components/OverviewPageDeviceDetails.tsx
```diff
@@ -9,7 +9,7 @@ interface Props {
   project: string;
 }
 
-const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
+const OverviewPageDeviceDetails: FC<Props> = ({ device, project }) => {
```
edlerd, line 12:
```suggestion
const DeviceDetails: FC<Props> = ({ device, project }) => {
```
A bit more concise and just as precise?

src/components/NetworkListTable.tsx
```diff
@@ -3,35 +3,38 @@ import { useQuery } from "@tanstack/react-query";
...
 import { fetchNetworks } from "api/networks";
 import { isNicDevice } from "util/devices";
 import ResourceLink from "components/ResourceLink";
+import { useParams } from "react-router-dom";
+import { LxdDevices } from "types/device";
 
 interface Props {
-  instance: LxdInstance;
   onFailure: (title: string, e: unknown) => void;
+  networkDevices: LxdDevices;
```
edlerd, line 14:
```suggestion
  devices: LxdDevices;
```
There are all devices, not just network devices. Better reflect that in the name not to confuse later on. Also it is shorter and thus simpler.

src/pages/instances/InstanceOverviewDevices.tsx
```diff
@@ -6,7 +6,7 @@ import { isRootDisk } from "util/instanceValidation";
 import { FormDevice } from "util/formDevices";
 import ResourceLink from "components/ResourceLink";
 import { isOtherDevice } from "util/devices";
-import InstanceOverviewDeviceDetail from "./InstanceOverviewDeviceDetail";
+import InstanceOverviewDeviceDetail from "../../components/OverviewPageDeviceDetails";
```
edlerd, line 9:
```suggestion
import InstanceOverviewDeviceDetail from "components/OverviewPageDeviceDetails";
```
Better not use relative imports, when crossing module boundaries.

src/pages/profiles/ProfileDetailOverview.tsx
```diff
@@ -10,21 +10,26 @@ import classnames from "classnames";
 import { CLOUD_INIT } from "./forms/ProfileFormMenu";
 import { slugify } from "util/slugify";
 import { getProfileInstances } from "util/usedBy";
-import ProfileNetworkList from "./ProfileNetworkList";
-import ProfileStorageList from "./ProfileStorageList";
+import NetworkListTable from "../../components/NetworkListTable";
```
edlerd, line 13:
```suggestion
import NetworkListTable from "components/NetworkListTable";
```

src/pages/profiles/ProfileDetailOverview.tsx
```diff
@@ -47,7 +52,7 @@ const ProfileDetailOverview: FC<Props> = ({ profile, featuresProfiles }) => {
     <div className="profile-overview-tab">
       {!featuresProfiles && (
         <Notification severity="caution" title="Inherited profile">
-          Modifications are only available in the{" "}
+          Modifications are only available in they
```
edlerd, line 55: this change seems wrong?

src/components/DeviceListTable.tsx
edlerd: Actually just applying the class is probably wrong, because the styling is applied only for instance overview page then and not for profile overview. 
I think best is to apply a "device-table" class here, and define a new css file. Then move the "devices" css block from the _instance_detail_overview.scss to the new file, to the root level, so it will be applied on both pages for instances and devices.

src/pages/instances/InstanceOverviewDevices.tsx
Kxiru: Thanks for letting me know. That's what I get for accepting imports to be automatically changed. I'll change them manually from now on.

src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+    return typeA.localeCompare(typeB);
+  };
+
+  // Identify non-NIC devices
+  const overviewDevices = Object.entries(devices ?? {})
+    .filter(([_key, device]) => {
+      return device.type !== "nic" && device.type !== "none";
+    })
+    // Sort devices by type and name
+    .sort(compareDevices);
```
edlerd, line 40:
```suggestion
    .sort(byTypeAndName);
```
By naming the function, you can make the code speaking and avoid the need for a comment

src/pages/instances/InstanceOverview.tsx
```diff
@@ -5,14 +5,15 @@ import { LxdInstance } from "types/instance";
...
 import useEventListener from "@use-it/event-listener";
 import { updateMaxHeight } from "util/updateMaxHeight";
-import InstanceOverviewNetworks from "./InstanceOverviewNetworks";
 import InstanceOverviewProfiles from "./InstanceOverviewProfiles";
 import InstanceOverviewMetrics from "./InstanceOverviewMetrics";
 import InstanceIps from "pages/instances/InstanceIps";
 import { useSettings } from "context/useSettings";
 import NotificationRow from "components/NotificationRow";
-import InstanceOverviewDevices from "./InstanceOverviewDevices";
+import DeviceListTable from "../../components/DeviceListTable";
```
edlerd, line 13:
```suggestion
import DeviceListTable from "/components/DeviceListTable";
```
Best avoid relative imports.

src/sass/_profile_detail_overview.scss
edlerd, line 1: We should remove the line below the last item in the devices table. I.e. in the screen below, there should not be a line under the usb device. The line splitting the devices and profiles section should stay.
image

edlerd: There should be a line between the devices and limits section on the profile overview page. Same as the line between other section, i.e. network and devices.

image

edlerd (commented):
Some concerns left, then this should be good to merge form my side.

src/components/DeviceListTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+    return typeA.localeCompare(typeB);
+  };
+
+  // Identify non-NIC devices
+  const overviewDevices = Object.entries(devices ?? {})
+    .filter(([_key, device]) => {
+      return device.type !== "nic" && device.type !== "none";
+    })
+    // Sort devices by type and name
+    .sort(byTypeAndName);
```
edlerd, line 40:
```suggestion
    .sort(byTypeAndName);
```
Now the comment seems verbose as the code speaks by itself.

src/sass/_profile_detail_overview.scss
```diff
@@ -12,6 +12,10 @@
     font-weight: 500;
   }
 
+  .row {
+    margin-top: $spv--x-large;
+  }
+
```
edlerd, line 18:
```suggestion
```
I might be missing something, but it seems this has no effect as it is overridden and never applied?

src/sass/_profile_detail_overview.scss
```diff
@@ -51,6 +55,23 @@
...
+    th:nth-child(1) {
+      width: 30%;
+    }
+
+    th:first-of-type,
+    td:first-of-type {
+      padding-left: 0;
+    }
+  }
```
edlerd, line 69: I think instead of duplicating the code from `_instance_detail_overview.scss`, better to extract them into a dedicated file -- as with the `_device_detail_list.scss`. We can have a `_network_detail_list.scss` as we introduced a reusable component for networks in the same way as we did previously for devices. So best to have the custom CSS also reused.

Also, I think the nth-child(1) and :first-of-type are the same, can we use one of the two and apply the two properties width and padding-left to that selector? That should simplify here. This is due to some legacy code, I think this simplification can be applied to other places in instance/profile overview as well, like in _device_detail_list and maybe other places as well.

edlerd (approved):
Thanks for the changes, QA LGTM, one nit and small ideas to simplify/reduce duplication below.

src/sass/_profile_detail_overview.scss
Kxiru: Hola!
As for that last paragraph, I don't think this can be condensed as nth child and first of type are different. In this case, nth child targets the first child of its parent, regardless of the type of element whilst first-of-type targets the first element of a specific type within its parent. Additionally, the latter is applied to table headers and table data cells and I don't think it would be useful to combine them.

edlerd: Ah, you are right!

src/sass/_network_detail_list.scss
```diff
@@ -0,0 +1,12 @@
+.networks {
```
edlerd, line 1: This is referencing a class that is not in the NetworkListTable component. We can define a dedicated `network-table` class in it and reference it here.

Then move the .networks and border-bottom to the respective usages in profile overview and instance overview scss files. Because the border is referencing element in those components.

The main idea is to keep the scss files aligned with the react component files and avoid referencing classes from other components.

edlerd (approved):
QA and code LGTM, tiny nitpick below, also fine to merge as is.

src/sass/_network_detail_list.scss
```diff
@@ -0,0 +1,12 @@
+.network-table {
+  // border-bottom: 1px solid $color-mid-light;
```
edlerd, line 2:
```suggestion
```

PR #969, opened 2024-10-23
fix(instance) layout adjustments for the instance overview tab

edlerd:
## Done

- layout adjustments for the instance overview tab
- consistent font size
- slightly improved spacing, captions and tag usage for the devices table.

## QA

1. Run the LXD-UI:
    - check instance overview tab

### DIFF
--- a/src/pages/instances/InstanceOverviewDeviceDetail.tsx
+++ b/src/pages/instances/InstanceOverviewDeviceDetail.tsx
@@ -14,7 +14,7 @@ const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
     if (isRootDisk(device as FormDevice)) {
       return (
         <>
-          <span>pool </span>
+          Pool{" "}
           <ResourceLink
             type={"pool"}
             value={device.pool}
@@ -26,15 +26,13 @@ const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
 
     return (
       <>
-        <span>
-          volume&nbsp;
-          <ResourceLink
-            type={"volume"}
-            value={device.source as string}
-            to={`/ui/project/${project}/storage/pool/${device.pool}/volumes/custom/${device.source}`}
-          />{" "}
-        </span>
-        <span> on pool </span>
+        Volume{" "}
+        <ResourceLink
+          type={"volume"}
+          value={device.source as string}
+          to={`/ui/project/${project}/storage/pool/${device.pool}/volumes/custom/${device.source}`}
+        />{" "}
+        on pool{" "}
         <ResourceLink
           type={"pool"}
           value={device.pool}
@@ -47,7 +45,7 @@ const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
   if (device.type === "gpu") {
     return (
       <span title={device.pci} className="u-truncate">
-        {device.pci ? `pci: ${device.pci}` : `id: ${device.id}`}
+        {device.pci ? `PCI  ${device.pci}` : `ID ${device.id}`}
       </span>
     );
   }
@@ -55,15 +53,12 @@ const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
   if (device.type === "proxy") {
     return (
       <>
-        <span
-          title={device.listen}
-          className="u-truncate"
-        >{`listen: ${device.listen}`}</span>{" "}
-        <br />
-        <span
-          title={device.connect}
-          className="u-truncate"
-        >{`connect: ${device.connect}`}</span>
+        <div title={device.listen} className="u-truncate">
+          Listen: {device.listen}
+        </div>
+        <div title={device.connect} className="u-truncate">
+          Connect: {device.connect}
+        </div>
       </>
     );
   }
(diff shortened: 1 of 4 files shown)


src/pages/instances/InstanceOverview.tsx
```diff
@@ -49,10 +49,10 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
                 <th className="u-text--muted">Base image</th>
                 <td>
                   {instance.config["image.description"] ? (
-                    <ResourceLabel
-                      type="image"
-                      value={instance.config["image.description"]}
-                    />
+                    <>
+                      <ResourceLabel type="image" value="" />
+                      {instance.config["image.description"]}
```
mas-who, line 54: oh interesting, I see what you did here. Maybe we should truncate the image description here in case if it gets too long?

src/pages/instances/InstanceOverviewDeviceDetail.tsx
```diff
@@ -46,24 +44,21 @@ const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
 
   if (device.type === "gpu") {
     return (
-      <span title={device.pci} className="u-truncate">
-        {device.pci ? `pci: ${device.pci}` : `id: ${device.id}`}
-      </span>
+      <div title={device.pci} className="u-truncate">
```
mas-who, line 47: why change these from `span` to `div`? I think semantically `span` makes sense?

src/pages/instances/InstanceOverviewDeviceDetail.tsx
```diff
@@ -26,15 +26,13 @@ const InstanceOverviewDeviceDetail: FC<Props> = ({ device, project }) => {
...
-        <span>
-          volume&nbsp;
-          <ResourceLink
-            type={"volume"}
-            value={device.source as string}
-            to={`/ui/project/${project}/storage/pool/${device.pool}/volumes/custom/${device.source}`}
-          />{" "}
-        </span>
-        <span> on pool </span>
+        Volume&nbsp;
```
mas-who, line 29: for some reason it seems like that `&nbsp;` isn't working on really small screen sizes?

src/pages/instances/InstanceOverview.tsx
edlerd: This is not the cleanest way to do it. Maybe we should add a "normal font size" option to the Resource Label component instead. So it gives us truncation and it might be useful in other places.

src/pages/instances/InstanceOverviewDeviceDetail.tsx
edlerd: I wanted to get rid of the explicit line break. As divs are block elements they appear in dedicated lines and we can remove it.

edlerd: I wanted to get rid of the explicit line break. As divs are block elements they appear in dedicated lines and we can remove it.

edlerd: We can always use the regular for it, I think.

src/pages/instances/InstanceOverview.tsx
edlerd: We can always use the regular font, I think.

src/pages/instances/InstanceOverviewDeviceDetail.tsx
edlerd: Replaced it with a regular space as there is no good solution if we run out of space having them in one line would hide the content, which is even worse than splitting by lines.

src/pages/instances/InstanceOverview.tsx
mas-who: If we use normal font size for `ResourceLabel`, then in notifications the font size will be different between `ResourceLabel` and `ResourceLink` (see image below). Although I think it actually looks okay given that `ResourceLabel` usually appear next to text with normal font size. wdyt?

Screenshot from 2024-10-24 09-39-45

src/pages/instances/InstanceOverview.tsx
```diff
@@ -49,10 +49,12 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
...
-                    <ResourceLabel
-                      type="image"
-                      value={instance.config["image.description"]}
-                    />
+                    <>
+                      <ResourceLabel
+                        type="image"
+                        value={instance.config["image.description"]}
+                      />
+                    </>
```
mas-who, line 57: nit
```suggestion
                      <ResourceLabel
                        type="image"
                        value={instance.config["image.description"]}
                      />
```

src/pages/instances/InstanceOverviewDeviceDetail.tsx
mas-who: nit: yes agreed with the approach for the `proxy` device section, since there are two lines. Although here it's a single line for the `gpu` device id, so `span` makes sense as it is semantically correct no?

mas-who (approved):
Thanks for the improvements, just some nits, not blocking, feel free to merge :)

src/pages/instances/InstanceOverview.tsx
edlerd: Yeah, I think this looks all right.

src/pages/instances/InstanceOverviewDeviceDetail.tsx
edlerd: ah right, will change this one :+1:

mas-who (approved):

PR #970, opened 2024-10-24
fix: remove publish to gh-pages from workflows

mas-who:
## Done

- Removed functionality to publish e2e reports to github pages in our workflows. The `gh-pages` branch history is causing bloat for the repo.

## QA

1. Run the LXD-UI:
    - Ensure CI passes


.github/workflows/coverage.yaml
```diff
@@ -110,60 +110,8 @@ jobs:
...
-            done
-        - name: Output Report URL as Worfklow Annotation
-          run: |
-            FULL_HTML_REPORT_URL=https://canonical.github.io/lxd-ui/coverage
-            echo "::notice title=Published Playwright Test Report::$FULL_HTML_REPORT_URL"
-
-
   tics-report:
     runs-on: ubuntu-latest
-    needs: publish-coverage-report
```
mas-who, line 166: @edlerd not sure why `tics-report` depended on `publish-coverage-report`. From what I can see, just removing this line in this job should be fine, let me know if I missed something here

.github/workflows/coverage.yaml
edlerd, line 1: We should keep the coverage report publishing, that is needed for the tics integration to fetch the latest coverage results.

edlerd (changes requested):
Thx for bringing this up quickly. On review, I realized we can't completely remove the gh-pages branch, because we need to keep it for the coverage. But the payload there is much smaller than for the e2e reports, so should be fine to keep.

edlerd (approved):
LGTM, thanks for bringing this up so quickly!

PR #974, opened 2024-10-28
feat(instance) add cluster member column and filter to instance list

edlerd:
## Done

- feat(instance) add cluster member column and filter to instance list

## QA

1. Run the LXD-UI:
    - check instance list in a clustered and non-clustered environment
    - in one the new cluster member column should show up and a filter on the instance list for it be available
    - in a non-clustered environment, the cluster member column and filter input should be invisible.

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -40,6 +40,7 @@ import {
   DESCRIPTION,
   IPV4,
   IPV6,
+  CLUSTER_MEMBER,
   NAME,
   SIZE_HIDEABLE_COLUMNS,
   SNAPSHOTS,
@@ -58,6 +59,8 @@ import useSortTableData from "util/useSortTableData";
 import PageHeader from "components/PageHeader";
 import InstanceDetailPanel from "./InstanceDetailPanel";
 import { useSmallScreen } from "context/useSmallScreen";
+import { useSettings } from "context/useSettings";
+import { isClusteredServer } from "util/settings";
 
 const loadHidden = () => {
   const saved = localStorage.getItem("instanceListHiddenColumns");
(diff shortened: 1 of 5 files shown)


src/pages/instances/InstanceList.tsx
```diff
@@ -178,10 +184,14 @@ const InstanceList: FC = () => {
       return false;
     }
     if (
```
mas-who, line 186: why don't we do something like 

```
const matchedProfiles =  filters.profiles.length > 0 && filters.profiles.every((profile) => item.profiles.includes(profile));
const matchedClusterMembers = filters.clusterMembers.length > 0 && filters.clusterMembers.includes(item.location);
...

return matchedProfiles || matchedClusterMembers || ...
```

mas-who (commented):
thanks for the improvements, just two very small nit :)

mas-who:
seems like CI is failing?

src/pages/instances/InstanceList.tsx
```diff
@@ -353,51 +372,67 @@ const InstanceList: FC = () => {
...
             "aria-label": TYPE,
             onClick: openSummary,
             className: "clickable-cell",
             style: { width: `${COLUMN_WIDTHS[TYPE]}px` },
           },
+          ...(isClustered
+            ? [
+                {
+                  content: (
+                    <div className="u-truncate" title={instance.location}>
```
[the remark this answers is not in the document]
edlerd: good idea.

src/pages/instances/InstanceList.tsx
edlerd: I think early access is a bit simpler to understand here and also a bit quicker. wdyt?

mas-who: yeah true, we won't have to do a filter for each condition then, please disregard

mas-who (approved):

PR #981, opened 2024-11-07
clarify ipv4 address label on network creation

edlerd:
## Done

- clarify ipv4 address label on network creation.
- additional help text for v4/v6 address on the network form

### DIFF
--- a/src/pages/networks/forms/NetworkFormMain.tsx
+++ b/src/pages/networks/forms/NetworkFormMain.tsx
@@ -71,11 +71,12 @@ const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
             getConfigurationRow({
               formik,
               name: "ipv4_address",
-              label: "IPv4 address",
+              label: "IPv4 address range",
               defaultValue: "auto",
               children: (
                 <IpAddressSelector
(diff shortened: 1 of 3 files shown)


src/pages/networks/forms/NetworkFormMain.tsx
```diff
@@ -109,11 +110,12 @@ const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
             getConfigurationRow({
               formik,
               name: "ipv6_address",
-              label: "IPv6 address",
+              label: "IPv6 address range",
               defaultValue: "auto",
               children: (
                 <IpAddressSelector
                   id="ipv6_address"
+                  family="IPv4"
```
Kxiru, line 118: Shouldn't this be IPv6? :)

Kxiru (changes requested):
Small copy error, but other than this, this is good to go! :+1:

src/pages/networks/forms/NetworkFormMain.tsx
edlerd: yes, good catch!

Kxiru (approved):
Looks good to me! :)

PR #982, opened 2024-11-07
feat: [WD-16398] Set default storage pool for new project...

Kxiru:
...on creation

## Done
- Create new project case
    - Used StoragePool Selector component
    - Disabled storage pool selector (sps) when profiles are shared.
    - Selected storage pool becomes project's default profile's storage pool when saved.
    - Error handling.
- Edit project case
    - Storage pool is disabled
    - Help text shown to direct the user on where to edit the storage pool
    - Initial value of storage pool is set to be the default profile's storage pool.

## QA

1. Run the LXD-UI:

## Screenshots

image

### DIFF
--- a/src/pages/projects/EditProject.tsx
+++ b/src/pages/projects/EditProject.tsx
@@ -1,7 +1,7 @@
 import { FC, useEffect } from "react";
 import { Button, useNotify } from "@canonical/react-components";
 import { updateProject } from "api/projects";
-import { useQueryClient } from "@tanstack/react-query";
+import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { PROJECT_DETAILS } from "pages/projects/forms/ProjectFormMenu";
 import { useFormik } from "formik";
@@ -23,6 +23,7 @@ import { useToastNotification } from "context/toastNotificationProvider";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
 import FormSubmitBtn from "components/forms/FormSubmitBtn";
 import ResourceLink from "components/ResourceLink";
+import { fetchProfile } from "api/profiles";
 
 interface Props {
   project: LxdProject;
@@ -38,6 +39,10 @@ const EditProject: FC<Props> = ({ project }) => {
   const { hasProjectsNetworksZones, hasStorageBuckets } =
     useSupportedFeatures();
 
+  const { data: profile } = useQuery({
+    queryKey: [queryKeys.profiles, "default", project.name],
+    queryFn: () => fetchProfile("default", project.name),
+  });
   const updateFormHeight = () => {
     updateMaxHeight("form-contents", "p-bottom-controls");
   };
@@ -48,7 +53,7 @@ const EditProject: FC<Props> = ({ project }) => {
     name: Yup.string().required(),
   });
 
-  const initialValues = getProjectEditValues(project);
+  const initialValues = getProjectEditValues(project, profile);
 
   const formik: FormikProps<ProjectFormValues> = useFormik({
     initialValues: initialValues,
--- a/src/pages/projects/forms/ProjectDetailsForm.tsx
+++ b/src/pages/projects/forms/ProjectDetailsForm.tsx
@@ -17,10 +17,13 @@ import ScrollableForm from "components/ScrollableForm";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
 import { LxdConfigPair } from "types/config";
 import { ensureEditMode } from "util/instanceEdit";
+import StoragePoolSelector from "pages/storage/StoragePoolSelector";
+import ResourceLink from "components/ResourceLink";
 
 export interface ProjectDetailsFormValues {
   name: string;
   description?: string;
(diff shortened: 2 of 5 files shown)


src/pages/projects/EditProject.tsx
```diff
@@ -38,6 +39,10 @@ const EditProject: FC<Props> = ({ project }) => {
   const { hasProjectsNetworksZones, hasStorageBuckets } =
     useSupportedFeatures();
 
+  const { data: profile } = useQuery({
+    queryKey: [queryKeys.profiles, "default", "default"],
+    queryFn: () => fetchProfile("default", "default"),
```
edlerd, line 44:
```suggestion
    queryKey: [queryKeys.profiles, "default", project.name],
    queryFn: () => fetchProfile("default", project.name),
```
We need the profile from the project we are editing here.

src/pages/storage/StoragePoolSelector.tsx
edlerd, line 1: I think we don't need the changes here. See the next comment below.

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -146,144 +149,183 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
-            </>
-          )}
+              </>
+            )}
+          </div>
+
           <hr />
+          <StoragePoolSelector
+            id="default_instance_storage_pool"
+            name="default_instance_storage_pool"
```
edlerd, line 303:
```suggestion
```
I think we don't need the id or name here, as it is not referenced. We can rely on the StoragePoolSelector to provide its default id and name.

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -17,10 +17,13 @@ import ScrollableForm from "components/ScrollableForm";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
 import { LxdConfigPair } from "types/config";
 import { ensureEditMode } from "util/instanceEdit";
+import StoragePoolSelector from "pages/storage/StoragePoolSelector";
+import ResourceLink from "components/ResourceLink";
 
 export interface ProjectDetailsFormValues {
   name: string;
   description?: string;
+  default_instance_storage_pool?: string;
```
edlerd, line 26:
```suggestion
  default_instance_storage_pool: string;
```
I think as we initialize this as string, we can make the type a bit stricter, and that will save some casts below.

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -146,144 +149,183 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
-          )}
+              </>
+            )}
+          </div>
+
           <hr />
+          <StoragePoolSelector
+            id="default_instance_storage_pool"
+            name="default_instance_storage_pool"
+            value={formik.values.default_instance_storage_pool as string}
```
edlerd, line 304:
```suggestion
            value={formik.values.default_instance_storage_pool}
```
With the stricter type, this cast should not be necessary.

edlerd (commented):
Looks great so far! Some comments below.
Maybe add a test for creating a project with custom pool and ensure it gets applied correctly in the new default profile?

src/pages/storage/StoragePoolSelector.tsx
Kxiru: Noted, thank you. I was used to id and name needing to be the same in order to make formik work...

edlerd: They are needed to be aligned, if we use the formik helpers. Here the custom on change handlers do the state propagation to formik, so we don't need to align id and name.

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -146,144 +149,181 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
+              label: "Default instance storage pool",
+              disabled:
+                (formik.values.features_profiles === false &&
+                  features === "customised") ||
+                isEdit,
+              help: isEdit ? (
+                <>
+                  Edit the storage pool in the{" "}
+                  <ResourceLink
+                    type="instance"
```
edlerd, line 316:
```suggestion
                    type="profile"
```
We should use the correct type, this is a profile, not an instance.

src/pages/storage/StoragePoolSelector.tsx
```diff
@@ -74,8 +74,6 @@ const StoragePoolSelector: FC<Props> = ({
 
   return (
     <Select
-      id="storage-pool-selector"
-      name="pool"
       options={getStoragePoolOptions()}
```
edlerd, line 77: I was probably commenting a bit ambiguous before. I meant customizing those for the profile form use case should be avoided. We should keep an id and a name here, otherwise we end up with trouble for a11y.

edlerd (changes requested):
Two tiny issues are left.

edlerd (approved):
LGTM, thanks for resolving this quickly.

PR #986, opened 2024-11-15
Fix warnings from TICS scan

edlerd:
## Done

- fixing a couple of warnings that showed up in the TICS dashboard


mas-who (approved):
LGTM, nice to see TICS provides some really relevant warnings :+1:

PR #988, opened 2024-11-18
simplify keys setup

edlerd:
## Done

- simplify keys setup, use ecc key which speeds up key generation, so we don't need to cache the key and can remove dhparams generation


simondeziel (commented):
LGTM (for what it's worth ;)

mas-who (approved):
Nice! checked the `Run LXD-UI` job against old action runs, no increase in run time even with out the cache. Will remember this for other projects :+1:

PR #989, opened 2024-11-18
github: save CI time by combining scss and JS linting since they use the same dependencies

simondeziel:


edlerd (approved):
Good idea, thanks for the improvement.

PR #991, opened 2024-11-18
feat: [WD-16399] Improve storage pool selector

Kxiru:
## Done

- Changed default Select to Custom Select component

## QA

1. Run the LXD-UI:
    - Review areas where StoragePoolSelector was previously used. Some examples are:
    -     New Project creation page
    -     Custom volume duplication page
    -     Permission Selector page

## Screenshots

image


edlerd:
@Kxiru please check the jira issue, from the screenshot it seems a couple of requirements have not been met yet.

mas-who:
> @Kxiru please check the jira issue, from the screenshot it seems a couple of requirements have not been met yet.

@Kxiru just adding to comment from @edlerd , I think what you want to do is to adjust the storage pool selector to have columns like the one shown below for example. Based on the Jira issue, the columns would be pool name, driver and usage. 

Screenshot from 2024-11-19 09-48-27

edlerd:
This looks much better, thx for the adjustments!

A couple of QA issues:

1. The tests need to be adjusted to the new selector.

2. The css of the custom select creates a ghost disk selector if you navigate to https://lxd-ui-991.demos.haus/ui/project/default/instance/aware-mackerel/configuration/disk and click "create image":
image

3. Please remove the "select option" line from the selector:
image

4. Spacing of the columns, I think the "driver" column could be smaller and the usage column a bit wider:
image

src/pages/storage/actions/DuplicateVolumeBtn.tsx
```diff
@@ -11,7 +11,9 @@ interface Props {
 }
 
 const DuplicateVolumeBtn: FC<Props> = ({ volume, className, onClose }) => {
-  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal({
+    closeOnOutsideClick: false,
+  });
```
edlerd, line 16: The custom select should contain the click events and avoid propagation. I think this (and similar) workarounds should be removed before merging. Again, we can wait for a fixing PR for the custom select to land until we come back to this one.

src/pages/storage/StoragePoolSelector.tsx
edlerd, line 1: When an option is selected, the pool selector should be active, but it looses focus instead. This might be a bug in custom select.

edlerd, line 1: The `DuplicateInstanceForm` should also use this selector.

edlerd (changes requested):

src/pages/storage/actions/DuplicateVolumeBtn.tsx
edlerd: I didn't find an easy way to avoid this. Maybe worth keeping the current approach from this PR.

src/pages/storage/StoragePoolSelector.tsx
edlerd: This is fixed in https://github.com/canonical/lxd-ui/pull/1002

src/sass/_custom_select.scss
```diff
@@ -75,6 +75,6 @@
   }
 }
 
-.p-custom-select__wrapper {
+.custom-select-dropdown-overlay {
```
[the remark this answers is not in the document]
edlerd: This PR makes this workaround obsolete: https://github.com/canonical/lxd-ui/pull/1002 after it landed, please rebase on main and remove this class and all its usages.

src/pages/storage/actions/DuplicateVolumeBtn.tsx
edlerd: I found a way to stop the event in the custom select as part of #1002 
please remove this workaround -- from here and also from other files, after that fix is merged and this is rebased on main.

src/pages/storage/StorageOptionHeader.tsx
edlerd, line 1: This file should be called StoragePoolOptionHeader

src/pages/storage/StorageOptionHeader.tsx
```diff
@@ -0,0 +1,20 @@
...
+      <span key={column} className="resource u-no-margin--bottom">
+        {column === "imageType" ? "image type" : column}
+      </span>
+    );
+  });
+
+  return <div className="header">{headerSegments}</div>;
+};
+
+export default ResourceOptionHeader;
```
edlerd, line 20: The naming here seems not aligned with the filename

src/pages/storage/StoragePoolOptionLabel.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <div className="label">
+      <span key="name" title="Name" className="resource u-truncate">
+        {pool.name || "-"}
+      </span>
+      <span key="driver" title="Driver" className="resource-driver u-truncate">
+        {pool.driver || "-"}
+      </span>
+      <span key="usage" title="Usage" className="resource-usage u-truncate">
+        <StoragePoolSize key={pool.name} pool={pool} shorthand_text={true} />
```
edlerd, line 19: There is a design problem with using the size graph here:

image

The highlight color and the usage color are matching and leaving no contrast to understand. Can you find a better solution?

edlerd (changes requested):
I created #1002 which should unblock and simplify this PR. A couple of other issues are highlighted above.

src/pages/storage/actions/DuplicateVolumeBtn.tsx
mas-who: @Kxiru #1002 has been merged

src/pages/storage/StoragePoolSelector.tsx
edlerd, line 1: The search input should also match on the driver.

Also the label in the custom select is `[object object]`, this should be a stringified version of the object instead I think this might be related to the search matching.

src/pages/storage/StoragePoolOptionLabel.tsx
Kxiru: Solution discussed in call: Conditionally removing the meter bar as having the same bar represented with different colour across the UI would be inconsistent, and the colour cannot be changed from Blue and is it a Vanilla colour.

src/pages/storage/StoragePoolSelector.tsx
Kxiru: Using String(), toString, and JSON.stringify do not return the object as a string here, but also, stringifying it will not show the driver either as the object only has the label, text and value, and the label - where the driver is stored - is a component. Is there another way we can access the driver from the option or have I misinterpreted something?

edlerd: I think there is a field on the custom select options prop to set the searchable text.

@mas-who should know the details, or I can check the details tomorrow and share it.

Kxiru: CustomSelectDropdown.tsx:
image
This is from where I am referencing the information above.

mas-who: The `text` attribute for a `CustomSelectOption` is used for search, sort and display in the toggle when `label` is a component. See [comment](https://github.com/canonical/lxd-ui/blob/b452c418c5ce083ef7dd21bb6427530bad1d6f62/src/components/select/CustomSelectDropdown.tsx#L17) regarding that.

In this case, if search must match on pool name and driver type, I'd suggest we can set option `text` to something like `${pool.name} (${pool.driver})`? This should then be displayed in the toggle as well once the option is selected.

mas-who: > Also the label in the custom select is [object object], this should be a stringified version of the object instead. I think this might be related to the search matching.

@edlerd would you be able to provide me with a screenshot of the option that you selected which resulted in the above behaviour? Looking at the code now the toggle label should display pool name and not some stringified object.

edlerd: @mas-who the problem is spreading the option directly in the dropdown, see the screenshot below. It leads to an invalid label value. I'll send a PR shortly to fix that.

image

edlerd: see https://github.com/canonical/lxd-ui/pull/1009 for a fix

mas-who: Oh I see what you mean, it's the `label` attribute for the `li` element and not the actual `label` prop for the options. Option should not be destructured into `li` actually.

mas-who: One more thing, the comment about setting the `text` prop for the options to have `pool.driver` still stands. This will allow search to match on driver type

src/pages/storage/StoragePoolOptionHeader.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const columns = ["name", "driver", "usage"];
+
+  if (columns.length < 2) {
+    return null;
+  }
+
+  const headerSegments = columns.map((column) => {
+    return (
+      <span key={column} className="resource u-no-margin--bottom">
+        {column === "imageType" ? "image type" : column}
```
edlerd, line 12: What does the image type mean here, can it be removed?

edlerd: I see this is duplicating ResourceOptionHeader. I think a better solution is to generalize that component, so we can reuse it here and for permissions.

src/pages/storage/StoragePoolSize.tsx
edlerd, line 1: Sorry to go back and forth on this one. Seeing it now in action, the usage text is rendered in small font, which looks unaligned in the table. I think the simpler solution is to make this component return the text value directly, if `hasMeterbar` is false. Then we don't need any changes to the Meter component and render the text in the correct size.

src/sass/_storage_pool_selector.scss
```diff
@@ -0,0 +1,43 @@
...
+    display: flex;
+    gap: $sph--small;
+    padding: $sph--x-small $sph--small;
+    position: sticky;
+    top: 0;
+
+    span {
+      @extend %table-header-label;
+
+      color: $colors--theme--text-default;
```
edlerd, line 19:
```suggestion
```
I think this can be removed or why was it added?

src/pages/storage/StoragePoolSelector.tsx
```diff
@@ -1,25 +1,32 @@
...
+import StoragePoolOptionLabel from "./StoragePoolOptionLabel";
+import StoragePoolOptionHeader from "./StoragePoolOptionHeader";
+import classNames from "classnames";
 
 interface Props {
   value: string;
   setValue: (value: string) => void;
   selectProps?: SelectProps;
   hidePoolsWithUnsupportedDrivers?: boolean;
+  dropdownClassName?: string;
```
edlerd, line 21:
```suggestion
```
I think this prop is not needed.

src/pages/storage/StoragePoolSelector.tsx
```diff
@@ -56,30 +63,47 @@ const StoragePoolSelector: FC<Props> = ({
...
+    setValue(value);
+  };
+
   return (
-    <Select
+    <CustomSelect
       id="storage-pool-selector"
       name="pool"
+      aria-label="Storage Pool"
+      searchable="always"
```
edlerd, line 97:
```suggestion
```
Let's go with the default of `auto` here, maybe? I think we can hide the search if there are less than 5 storage pools available.

src/pages/storage/StoragePoolSelector.tsx
edlerd: Yes, I am adding a new comment for the below for resolving this long thread.

edlerd (dismissed):
Thanks for the iteration, this has become much better and simpler.

Found one more QA edge case and an old one is still open. A couple of suggestions for simplifying the code. Then this should be good to merge.

src/pages/storage/StoragePoolSelector.tsx
```diff
@@ -56,30 +63,47 @@ const StoragePoolSelector: FC<Props> = ({
...
   };
 
+  const handleStoragePoolChange = (value: string) => {
+    setValue(value);
+  };
+
   return (
-    <Select
+    <CustomSelect
       id="storage-pool-selector"
```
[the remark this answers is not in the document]
edlerd: Sorry, that was a bit confusing. I think the `StoragePoolSelector` should not set an id in this line. In the use cases of the storage pool selector, that do need to reference it by ID, they should set a unique ID as `<StoragePoolSelector selectprops={id: "storage-pool-selector-:someUseCaseDescriptionThatIsUnique [...]`. In other use cases, that ID should be empty, and we can rely on the `CustomSelect` to generate a unique one. We can also have a quick chat on this if it is unclear.

src/pages/storage/StoragePoolSelector.tsx
```diff
@@ -56,30 +63,47 @@ const StoragePoolSelector: FC<Props> = ({
...
-      disabled: true,
-    });
+    const options: CustomSelectOption[] = [];
+    if (poolsToUse) {
+      poolsToUse.forEach((pool) => {
+        options.push({
+          label: <StoragePoolOptionLabel pool={pool} hasMeterBar={false} />,
+          value: pool.name,
+          disabled: false,
+          text: pool.name,
```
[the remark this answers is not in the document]
Kxiru: image
Is this really what we want?
image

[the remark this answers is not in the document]
edlerd: I think that is fine, probably don't need the comma, but just putting the driver behind the name makes the search match it, which is what I think users will expect.

src/sass/_storage_pool_selector.scss
Kxiru: I don't remember adding this, it must have been an accident, apologies.

src/pages/storage/StoragePoolSelector.tsx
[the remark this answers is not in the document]
Kxiru: Noted,
Removing the ID here has fixed the specific issue that you mentioned. Does it not count as one of those edge use cases?
Thank you for providing a solution for similar issues in the future.

src/components/Meter.tsx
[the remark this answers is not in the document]
edlerd: This is still open, I think.

src/pages/storage/StoragePoolSize.tsx
Kxiru: Sigh. I think my changes got lost somewhere. Thanks for pointing this out!

mas-who (commented):
Nicely done, all the remaining comments from @edlerd have been addressed and I tested all places where the new storage pool selector is used, seems to work as expected.

I've added a few small nits to clean up the code a bit, then it should be good to go. Although @edlerd will need to approve the PR still since he requested changes.

mas-who (approved):
LGTM, thanks for the changes :)

PR #999, opened 2024-11-22
Use shared tsconfig from design system

edlerd:
## Done

- Use shared tsconfig from design system
- Update eslint typescript packages
- Fix code issues to align with new rules.

Fixes [WD-17093](https://warthogs.atlassian.net/browse/WD-17093)

[WD-17093]: https://warthogs.atlassian.net/browse/WD-17093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -59,6 +58,10 @@
     "@babel/preset-env": "7.25.3",
     "@babel/preset-react": "7.24.7",
     "@babel/preset-typescript": "7.24.7",
+    "@biomejs/biome": "1.9.2",
+    "@canonical/biome-config": "0.1.1-experimental.0",
+    "@canonical/ds-react-core": "0.1.1-experimental.0",
```
jmuzina, line 63: This package will eventually have the new React components themselves, but it doesn't contain much that is ready for consumption right now.

src/types/cytoscapecomponent.d.ts
jmuzina, line 1: Was there some incompatibility that made creating these new type modules (cytoscape & spice) necessary?

package.json
```diff
@@ -59,20 +58,23 @@
     "@babel/preset-env": "7.25.3",
     "@babel/preset-react": "7.24.7",
     "@babel/preset-typescript": "7.24.7",
+    "@biomejs/biome": "1.9.2",
+    "@canonical/biome-config": "0.1.1-experimental.0",
```
jmuzina, line 62: By adding Biome, you should be able to replace Eslint & prettier, but I don't see that happening in this PR, and biome isn't being called.

If this PR is scoped to just using the tsconfig, I'd suggest dropping the Biome dependencies. 

It's also worth nothing that Biome doesn't currently support Sass (and I see some of that in this project), so Biome couldn't fully replace Eslint & Prettier for this project.

jmuzina (commented):
Very cool to see these configs getting some use, thanks for your work on this! A couple comments below to get the discussion going.

src/types/cytoscapecomponent.d.ts
edlerd: Yes:
- spice is a `js` library, hence we need to declare the types for it locally.
- `cytoscape` has a `@types/cytoscape`, but it created problems, so I created the local types and removed the types dependency.

tsconfig.json
```diff
@@ -1,42 +1,31 @@
...
-    "jsx": "react-jsx",
-    "strict": true,
-    "esModuleInterop": true,
     "skipLibCheck": true,
-    "forceConsistentCasingInFileNames": true,
-    "lib": ["dom", "dom.iterable", "esnext"],
-    "allowJs": true,
-    "allowSyntheticDefaultImports": true,
-    "noFallthroughCasesInSwitch": true,
-    "moduleResolution": "node",
```
mas-who, line 25: Can we maybe keep the `moduleResolution: node` setting so that we don't have to update all the file imports in the codebase?

mas-who: maybe an upstream suggestion?

src/Root.tsx
```diff
@@ -1,23 +1,23 @@
-import { FC } from "react";
-import Navigation from "components/Navigation";
+import React, { FC } from "react";
```
mas-who, line 1: interesting that we have to import `React` again now, I thought the jsx compiler is still set to `react-jsx`?

package.json
```diff
@@ -33,7 +33,6 @@
     "@canonical/react-components": "1.7.1",
     "@monaco-editor/react": "4.6.0",
     "@tanstack/react-query": "5.51.23",
-    "@use-it/event-listener": "0.1.7",
```
mas-who, line 36: why are we removing this dependency?

tsconfig.json
jmuzina: @mas-who Indeed we are working with ES modules for the new react component library. 

I don't see any issues with overriding the module settings in the project tsconfig. You'd also need to retain your existing `esModuleInterop` and `allowSyntheticDefaultImport` settings in this case.

```json
    "module": "esnext",
    "moduleResolution": "node",
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
```

CC @advl

edlerd: Yes, on 2nd thought we shouldn't mess with the imports or backporting anything will be a big pain. thx for the suggestion @mas-who  and thx for the new suggestion @jmuzina Adjusted the config and reverted the imports back to how they were before.

src/util/useEventListener.ts
```diff
@@ -0,0 +1,29 @@
...
+  WindowEventMap & {
+    ["spice-wheel"]: Event;
+  } & {
+    ["sfp-toggle"]: Event;
+  };
+
+function useEventListener<K extends keyof EventListenerEvents>(
+  eventName: K,
+  handler: (event: EventListenerEvents[K]) => void,
+  element: HTMLElement | Window | null = window,
```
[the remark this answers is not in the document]
edlerd: We don’t need them today and if we ever going to need it, we can extend at that point in time. Better to keep it simple IMHO.

src/Root.tsx
edlerd: removed those verbose imports

package.json
edlerd: The types for that library are incompatible, and it seemed simple enough to  implement  ourselves

edlerd:
Thx for the review and suggestions @jmuzina and @mas-who . Please have another look with the updated config and changes.

tsconfig.json
jmuzina: On Monday I'll open up a MM thread for more discussion about what sort of module settings & import syntax you'd like to see in the future - there are benefits to using `NodeNext` (increased strictness, less ambiguity, consistency between dev & production environments), but possibly we only need some of these for the library itself, and the base config could be slightly less strict. 

Looking forward to the discussion :)

src/util/useEventListener.ts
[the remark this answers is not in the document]
mas-who: sounds good :+1:

tsconfig.json
```diff
@@ -1,42 +1,35 @@
...
       "api/*": ["src/api/*"],
       "context/*": ["src/context/*"],
       "components/*": ["src/components/*"],
+      "lib/*": ["src/lib/*"],
       "pages/*": ["src/pages/*"],
       "types/*": ["src/types/*"],
       "util/*": ["src/util/*"],
       "*": ["types/*.d.ts"]
     },
-    "target": "ES2020",
```
mas-who, line 14: upstream config for `target` is set to `ES2023`, was a bit worried about browser compatibility but according to this [matrix](https://compat-table.github.io/compat-table/es2016plus/) I think it is okay.

tsconfig.json
```diff
@@ -1,42 +1,35 @@
...
     "esModuleInterop": true,
-    "skipLibCheck": true,
-    "forceConsistentCasingInFileNames": true,
-    "lib": ["dom", "dom.iterable", "esnext"],
-    "allowJs": true,
     "allowSyntheticDefaultImports": true,
-    "noFallthroughCasesInSwitch": true,
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
```
mas-who, line 27: I think we should have `isolatedModules` set to `true` still, since we are using `vite` which uses `esbuild` under the hood for transpilation (see [reference](https://v2.vitejs.dev/guide/features.html#typescript-compiler-options))

tsconfig.json
mas-who: Thanks @jmuzina :)

jmuzina: @edlerd @mas-who  I've tried to summarize the module settings to get a conversation going about what is the best way forward for the configurations and consuming apps. [Have a look](https://coda.io/d/Design-System_d5tlz-HCRxz/DS-12-Develop-new-Code-Architecture_sumBJY72#DS-12-Notes_tubUgmzW/r125&view=full) and let me know what you think, thanks :)

edlerd:
Added some custom tsconfig options back and updated the eslint typescript plugins to be compatible with the ts update.
@mas-who and @jmuzina please have another pass at this PR.

package.json
```diff
@@ -13,8 +13,8 @@
     "format-js-prettier": "prettier 'src/**/*.{json,jsx,tsx,ts}' 'tests/**/*.ts' --write",
     "format-js": "yarn format-js-eslint && yarn format-js-prettier",
     "lint-scss": "stylelint 'src/**/*.scss'",
-    "lint-js-eslint": "eslint 'src/**/*.{json,tsx,ts}' 'tests/**/*.ts' .eslintrc.cjs babel.config.js",
-    "lint-js-prettier": "prettier 'src/**/*.{json,tsx,ts}' 'tests/**/*.ts' .eslintrc.cjs babel.config.js --check",
```
mas-who, line 17: I assume the config files are removed from the script because we updated to the latest version of eslint, and also prettier does not require them?

tsconfig.json
```diff
@@ -1,42 +1,37 @@
...
-    "target": "ES2020",
     "module": "esnext",
-    "jsx": "react-jsx",
-    "strict": true,
     "esModuleInterop": true,
-    "skipLibCheck": true,
-    "forceConsistentCasingInFileNames": true,
-    "lib": ["dom", "dom.iterable", "esnext"],
-    "allowJs": true,
+    "skipLibCheck": true, // ignore types of the node_modules
```
mas-who, line 18: thanks for adding these comments! :+1:

mas-who (approved):
LGTM :+1: just had a query regarding the linting scripts, but not blocking at all

package.json
edlerd: In order to check them here, we would need to set `allowJs` in the typescript config, which I'd like to avoid. Hence removing linting for the config files.

PR #1002, opened 2024-11-23
fix custom select

edlerd:
## Done

- align dropdown width to custom select wrapper width
- focus wrapper after selection of an option in the dropdown
- align z-index of dropdown to custom select parents, so when the custom select is in a modal, its dropdown is rendered above the modal.
- avoid bubbling up mouse down events on the dropdown, so a modal stays open

## QA

1. Run the LXD-UI:
    - check custom selects in permission context

### DIFF
--- a/src/components/select/CustomSelectDropdown.tsx
+++ b/src/components/select/CustomSelectDropdown.tsx
@@ -29,6 +29,7 @@ interface Props {
   onSelect: (value: string) => void;
   onClose: () => void;
   header?: ReactNode;
+  toggleId: string;
 }
 
 export const getOptionText = (option: CustomSelectOption): string => {
@@ -62,6 +63,7 @@ const CustomSelectDropdown: FC<Props> = ({
   onSelect,
   onClose,
   header,
+  toggleId,
 }) => {
   const [search, setSearch] = useState("");
   // track selected option index for keyboard actions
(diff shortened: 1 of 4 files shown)


src/components/select/CustomSelectDropdown.tsx
```diff
@@ -77,6 +79,38 @@ const CustomSelectDropdown: FC<Props> = ({
     (searchable === "always" || (searchable === "auto" && options.length >= 5));
 
   useEffect(() => {
+    if (dropdownRef.current) {
+      const toggle = document.getElementById(toggleId);
+
+      // align width with wrapper toggle width
+      const toggleWidth = toggle?.getBoundingClientRect()?.width ?? 0;
+      dropdownRef.current.style.setProperty("min-width", `${toggleWidth}px`);
```
mas-who, line 87: nice one :+1:

src/components/select/CustomSelectDropdown.tsx
```diff
@@ -77,6 +79,38 @@ const CustomSelectDropdown: FC<Props> = ({
...
+
+      // align z-index: if in a modal context, take the next z-index we find above + 1
+      const getRecursiveZIndex = (element: HTMLElement | null): string => {
```
mas-who, line 90: I think this function can exist outside of the component

mas-who (approved):
Thanks for the fixes, only have one small nit, feel free to merge as is

mas-who:
One small QA issue, there seems to be a top border when hovering the select now

Screenshot from 2024-11-23 11-40-24

edlerd:
> One small QA issue, there seems to be a top border when hovering the select now

I don't see it 🙈  What exactly do you mean?

src/components/select/CustomSelectDropdown.tsx
edlerd: moved it into a dedicated file, true this might be handy in other places.

edlerd:
Found the issue with the border, that was also on main, probably came in with a vanilla update.

mas-who:
> Found the issue with the border, that was also on main, probably came in with a vanilla update.

Was about to record a video and then saw this, thanks for the catch! :+1:

mas-who (approved):
LGTM!

PR #1008, opened 2024-11-26
feat: [WD-16894] Add bulk deletion and group modification of TLS Users

Kxiru:
## Done

- [✔] Allow selection of tls identities in the permission- > identity list
- [✔] Use the new DELETE 1.0/auth/identities/tls/:id endpoint for the bulk delete
- [✔] Unhide “delete” and “modify groups” button for tls users of the new type in the identities list
- [✔] Modify bulk actions "delete" and "modify groups" on the identity list page to also allow inclusion of tls users
- [✔] Legacy TLS users should be disabled, not deleted.

## QA

1. Run the LXD-UI:

**Create "new" TLS Fine grained identities**
- To add a new TLS Identity Client (Pending) please follow the steps for creating a pending fine-grained TLS identity in the '[Authenticate with the LXD server](https://documentation.ubuntu.com/lxd/en/latest/howto/server_expose/#authenticate-with-the-lxd-server)' section of the LXD documentation.
- Be sure to use the API instructions to create pending fine grained identities, as the CLI instructions require a remote to be specified.
- Be sure to authenticate the client by following the instructions through step 2.
 
**Delete TLS Users**
- Navigate to the Permissions > Identities 
- Attempt to delete individual TLS Client (Pending) identities using the inline delete identity button.
- Attempt to delete bulk TLS Client (Pending) identities.
- Verify that Legacy TLS identities (Client Certificate (Unrestricted) identities) cannot be selected via the checkbox or modified/deleted via inline buttons.

**Add TLS users users to a group**
- Navigate to Permissions > Groups
- Create a new group and add permissions to the group.
- Navigate to the Permissions > Identities 
- Attempt to add an individual TLS identity to a group using the inline modify groups button.
- Attempt to add several TLS identities (Pending) to a group.

**Test TLS user permissions / Login as a TLS User**
- Concatenate the .key and .crt files to create a .pem file using the following command:
- ```cat <KEY-FILE> <CRT-FILE> > <PEM-FILE-NAME>.pem ```
- Change lines 11 and 22 of the _haproxy-dev.cfg_ file to the following, respectively.
- ``` bind 0.0.0.0:8407 ssl verify optional crt <PEM-FILE-PATH> ca-file <CRT-FILE-PATH> ```
- ``` server lxd_https LXD_UI_BACKEND_IP:8443 ssl verify none crt <PEM-FILE-PATH> ```
- TLS Fine grained identities  have been tested with the following permissions:
image

## Screenshots
image

### DIFF
--- a/src/components/IdentityResource.tsx
+++ b/src/components/IdentityResource.tsx
@@ -1,0 +1,22 @@
+import { FC } from "react";
+import { LxdIdentity } from "types/permissions";
+import ResourceLabel from "./ResourceLabel";
+
+interface Props {
+  identity: LxdIdentity;
+  truncate?: boolean;
+}
+
+const IdentityResource: FC<Props> = ({ identity, truncate }) => {
+  const identityIconType =
+    identity.authentication_method == "tls" ? "certificate" : "oidc-identity";
+
+  return (
+    <ResourceLabel
+      type={identityIconType}
+      value={identity.type}
+      truncate={truncate}
+    />
+  );
+};
+export default IdentityResource;
--- a/src/components/ResourceLabel.tsx
+++ b/src/components/ResourceLabel.tsx
@@ -1,16 +1,18 @@
 import { FC } from "react";
 import ResourceIcon, { ResourceIconType } from "./ResourceIcon";
+import classNames from "classnames";
 
 interface Props {
   type: ResourceIconType;
   value: string;
   bold?: boolean;
+  truncate?: boolean;
 }
 
-const ResourceLabel: FC<Props> = ({ type, value, bold }) => {
+const ResourceLabel: FC<Props> = ({ type, value, bold, truncate = true }) => {
   const ValueWrapper = bold ? "strong" : "span";
   return (
-    <span className="resource-label u-truncate">
+    <span className={classNames("resource-label", { "u-truncate": truncate })}>
       <ResourceIcon type={type} />
       <ValueWrapper>{value}</ValueWrapper>
     </span>
--- a/src/pages/permissions/PermissionIdentities.tsx
+++ b/src/pages/permissions/PermissionIdentities.tsx
@@ -33,6 +33,8 @@ import Tag from "components/Tag";
 import BulkDeleteIdentitiesBtn from "./actions/BulkDeleteIdentitiesBtn";
 import DeleteIdentityBtn from "./actions/DeleteIdentityBtn";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
+import { isUnrestricted } from "util/helpers";
+import IdentityResource from "components/IdentityResource";
 
 const PermissionIdentities: FC = () => {
   const notify = useNotify();
@@ -52,12 +54,10 @@ const PermissionIdentities: FC = () => {
   const { hasAccessManagementTLS } = useSupportedFeatures();
 
   useEffect(() => {
-    const validIdentities = new Set(
-      identities.map((identity) => identity.name),
-    );
+    const validIdentityIds = new Set(identities.map((identity) => identity.id));
 
     const validSelections = selectedIdentityIds.filter((identity) =>
-      validIdentities.has(identity),
+      validIdentityIds.has(identity),
     );
 
     if (validSelections.length !== selectedIdentityIds.length) {
@@ -114,9 +114,9 @@ const PermissionIdentities: FC = () => {
 
   const rows = filteredIdentities.map((identity) => {
     const isLoggedInIdentity = settings?.auth_user_name === identity.id;
-    const isTlsIdentity = identity.authentication_method === "tls";
+
     return {
-      name: isTlsIdentity ? "" : identity.id,
+      name: isUnrestricted(identity) ? "" : identity.id,
       key: identity.id,
       className: "u-row",
       columns: [
@@ -144,7 +144,7 @@ const PermissionIdentities: FC = () => {
           "aria-label": "Auth method",
         },
         {
-          content: identity.type,
+          content: <IdentityResource identity={identity} truncate={false} />,
           role: "cell",
           "aria-label": "Type",
           className: "u-truncate",
@@ -157,10 +157,11 @@ const PermissionIdentities: FC = () => {
           "aria-label": "Groups for this identity",
         },
         {
-          content: !isTlsIdentity && (
+          content: !isUnrestricted(identity) && (
             <>
               <Button
                 appearance="base"
+                className="u-no-margin--bottom"
                 hasIcon
                 dense
                 onClick={() => {
@@ -198,10 +199,8 @@ const PermissionIdentities: FC = () => {
     defaultSort: "name",
   });
 
-  // NOTE: tls user group membership cannot be modified, this will be supported in the future
-  const nonTlsUsers = identities.filter((identity) => {
-    const isTlsIdentity = identity.authentication_method === "tls";
-    return !isTlsIdentity;
+  const fineGrainedIdentities = identities.filter((identity) => {
+    return !isUnrestricted(identity);
   });
 
   if (isLoading) {
@@ -218,11 +217,11 @@ const PermissionIdentities: FC = () => {
     if (selectedIdentityIds.length > 0) {
       return (
         <SelectedTableNotification
-          totalCount={nonTlsUsers.length ?? 0}
-          itemName="OIDC identity"
+          totalCount={fineGrainedIdentities.length ?? 0}
+          itemName="identity"
           selectedNames={selectedIdentityIds}
           setSelectedNames={setSelectedIdentityIds}
-          filteredNames={nonTlsUsers.map((item) => item.id)}
+          filteredNames={fineGrainedIdentities.map((item) => item.id)}
           hideActions={!!panelParams.panel}
         />
       );
@@ -294,7 +293,9 @@ const PermissionIdentities: FC = () => {
                 selectedNames={selectedIdentityIds}
                 setSelectedNames={setSelectedIdentityIds}
                 processingNames={[]}
-                filteredNames={nonTlsUsers.map((identity) => identity.id)}
+                filteredNames={fineGrainedIdentities.map(
+                  (identity) => identity.id,
+                )}
                 disableSelect={!!panelParams.panel}
               />
             </TablePagination>
--- a/src/pages/permissions/actions/BulkDeleteIdentitiesBtn.tsx
+++ b/src/pages/permissions/actions/BulkDeleteIdentitiesBtn.tsx
@@ -1,11 +1,11 @@
-import { FC } from "react";
+import { FC, useState } from "react";
 import {
   ButtonProps,
   ConfirmationButton,
   useNotify,
 } from "@canonical/react-components";
 import { LxdIdentity } from "types/permissions";
-import { deleteOIDCIdentities } from "api/auth-identities";
+import { deleteIdentities } from "api/auth-identities";
 import { useQueryClient } from "@tanstack/react-query";
 import { useToastNotification } from "context/toastNotificationProvider";
 import { queryKeys } from "util/queryKeys";
@@ -24,10 +24,12 @@ const BulkDeleteIdentitiesBtn: FC<Props & ButtonProps> = ({
   const notify = useNotify();
   const toastNotify = useToastNotification();
   const buttonText = `Delete ${pluralize("identity", identities.length)}`;
+  const [isLoading, setLoading] = useState(false);
   const successMessage = `${identities.length} ${pluralize("identity", identities.length)} successfully deleted`;
 
   const handleDelete = () => {
-    deleteOIDCIdentities(identities)
+    setLoading(true);
+    deleteIdentities(identities)
       .then(() => {
         void queryClient.invalidateQueries({
           predicate: (query) => {
@@ -37,10 +39,12 @@ const BulkDeleteIdentitiesBtn: FC<Props & ButtonProps> = ({
           },
         });
         toastNotify.success(successMessage);
+        setLoading(false);
         close();
       })
       .catch((e) => {
         notify.failure(`Identity deletion failed`, e);
+        setLoading(false);
       });
   };
 
@@ -50,6 +54,7 @@ const BulkDeleteIdentitiesBtn: FC<Props & ButtonProps> = ({
       appearance=""
       aria-label="Delete identities"
       className={className}
+      loading={isLoading}
       confirmationModalProps={{
         title: "Confirm delete",
         children: (
--- a/src/pages/permissions/actions/DeleteIdentityBtn.tsx
+++ b/src/pages/permissions/actions/DeleteIdentityBtn.tsx
@@ -1,4 +1,4 @@
-import { FC } from "react";
+import { FC, useState } from "react";
 import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import {
@@ -9,8 +9,8 @@ import {
 import { useToastNotification } from "context/toastNotificationProvider";
 import { LxdIdentity } from "types/permissions";
 import ItemName from "components/ItemName";
-import { deleteOIDCIdentity } from "api/auth-identities";
-import ResourceLabel from "components/ResourceLabel";
+import { deleteIdentity } from "api/auth-identities";
+import IdentityResource from "components/IdentityResource";
 
 interface Props {
   identity: LxdIdentity;
@@ -20,9 +20,11 @@ const DeleteIdentityBtn: FC<Props> = ({ identity }) => {
   const queryClient = useQueryClient();
   const notify = useNotify();
   const toastNotify = useToastNotification();
+  const [isDeleting, setDeleting] = useState(false);
 
   const handleDelete = () => {
-    deleteOIDCIdentity(identity)
+    setDeleting(true);
+    deleteIdentity(identity)
       .then(() => {
         void queryClient.invalidateQueries({
           predicate: (query) => {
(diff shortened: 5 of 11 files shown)


src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -199,10 +198,10 @@ const PermissionIdentities: FC = () => {
...
 
   // NOTE: tls user group membership cannot be modified, this will be supported in the future
-  const nonTlsUsers = identities.filter((identity) => {
-    const isTlsIdentity = identity.authentication_method === "tls";
-    return !isTlsIdentity;
-  });
+  // const nonTlsUsers = identities.filter((identity) => {
+  //   const isTlsIdentity = identity.authentication_method === "tls";
+  //   return !isTlsIdentity;
+  // });
```
edlerd, line 204:
```suggestion
```

src/pages/permissions/actions/DeleteIdentityBtn.tsx
```diff
@@ -33,8 +33,8 @@ const DeleteIdentityBtn: FC<Props> = ({ identity }) => {
         });
         toastNotify.success(
           <>
-            Identity <ResourceLabel type={"idp-group"} value={identity.name} />{" "}
-            deleted.
+            {identity.authentication_method.toUpperCase()} Identity{" "}
+            <ResourceLabel type={"idp-group"} value={identity.name} /> deleted.
```
edlerd, line 37: Why add the auth method of the deleted identity to the message here? I think without the method, it is shorter to read and easier to understand.

src/util/instanceBulkActions.tsx
```diff
@@ -75,8 +75,8 @@ export const pluralize = (item: string, count: number): string => {
     return item;
   }
 
-  if (item.includes("identity")) {
-    return item.replace("identity", "identities");
+  if (item.toLowerCase().includes("identity")) {
+    return item.toLowerCase().replace("identity", "identities");
```
edlerd, line 79: This should not mess with the case of the input. When asked to pluralize "Identity", it should not return the lowercase plural, or it has potential to mess up the copy. Maybe a better solution is to add another conditional for the upper case option.

edlerd (commented):
Some comments below, I think this is still missing to address the requirement "Legacy TLS users should be disabled, not deleted.". But maybe that is why it is still in draft state.

src/pages/permissions/actions/DeleteIdentityBtn.tsx
Kxiru: The attached figma file only differentiates between OIDC and TLS identities, not between OIDC, TLS (Fine-grained) and TLS Legacy, which I feel may be more appropriate for this UI. That being said, I implemented your suggestion and I think it is a nice step in the right direction. Let me know your thoughts :)

mas-who:
Good work so far @Kxiru for figuring out the difference between legacy and fine-grained tls identities. I think before conducting a proper review for the PR, there are a few items I'd like to see covered first:

1. Please would you adjust the PR title to reflect that group modification for tls identities are included in this PR
2. Please tidy up the PR description with respect to formatting, correctness and clarity (If you are not sure about this one, please reach out to me). When we merge the PR, the description becomes part of the merge commit message so we should keep it neat.
3. Since this PR includes modifying groups for a tls identity, have you QA'd locally if permissions work for those identities? If not, then you should do so by getting the actual certificate for the tls identity from the lxd server, add it to our haproxy configs and "log in" as the tls identity to test group permissions (can chat about this if you need direction for what to do).
4. Please also include detailed QA steps in the PR description for testing tls identity deletion and group modification.

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -144,7 +146,17 @@ const PermissionIdentities: FC = () => {
           "aria-label": "Auth method",
         },
         {
-          content: identity.type,
+          content: (
+            <ResourceLink
```
mas-who, line 150: We can use the `ResourceLabel` component here since we are not linking each identity to another page.

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -114,9 +115,10 @@ const PermissionIdentities: FC = () => {
 
   const rows = filteredIdentities.map((identity) => {
     const isLoggedInIdentity = settings?.auth_user_name === identity.id;
-    const isTlsIdentity = identity.authentication_method === "tls";
+    const isLegacyIdentity =
```
mas-who, line 118:
```suggestion
    const isUnrestrictedIdentity =
```

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -198,12 +210,6 @@ const PermissionIdentities: FC = () => {
     defaultSort: "name",
   });
 
-  // NOTE: tls user group membership cannot be modified, this will be supported in the future
-  const nonTlsUsers = identities.filter((identity) => {
```
mas-who, line 202: we still need to do a filter here for fine-grained identities

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -218,11 +224,11 @@ const PermissionIdentities: FC = () => {
     if (selectedIdentityIds.length > 0) {
       return (
         <SelectedTableNotification
-          totalCount={nonTlsUsers.length ?? 0}
-          itemName="OIDC identity"
+          totalCount={identities.length ?? 0}
```
mas-who, line 227: We can only select fine-grained identities for bulk action, the count here should reflect that

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -294,7 +300,7 @@ const PermissionIdentities: FC = () => {
                 selectedNames={selectedIdentityIds}
                 setSelectedNames={setSelectedIdentityIds}
                 processingNames={[]}
-                filteredNames={nonTlsUsers.map((identity) => identity.id)}
+                filteredNames={identities.map((identity) => identity.id)}
```
mas-who, line 303: related to prior comments, this should reflect fine grained identities and not all of them

src/pages/permissions/actions/DeleteIdentityBtn.tsx
mas-who: TLS legacy and TLS fine-grained are both certificates, so they can be represented by the same icon. Furthermore, these messages will never show a legacy TLS identity since you can't delete them.

Unresolving this comment for now, I think we should handle the following issues:
1. We should remove the auth method from the message.
2. The resource label type should be either `oidc-identity` or `certificate` depending on identity type.

src/pages/permissions/panels/EditIdentitiesForm.tsx
```diff
@@ -36,8 +36,8 @@ const EditIdentitiesForm: FC<Props> = ({
     notify.failure("Loading details failed", error);
   }
 
-  const nonTlsIdentities = identities.filter(
-    (identity) => identity.authentication_method !== "tls",
+  const nonLegacyIdentities = identities.filter(
```
mas-who, line 39:
```suggestion
  const fineGrainedIdentities = identities.filter(
```

mas-who (commented):
Nice update on the PR details :fire: 

I've left some code related comments so far. Will QA tomorrow.

src/pages/permissions/PermissionIdentities.tsx
Kxiru: I have actually made the logic such that we can select in bulk irrespective of the identity type and still delete according to the relevant API. The logic for modifying groups for mixed certificate/oidc bulk selections also works too. I think this would be an easier approach for the User. Do you wish to only select one identity type in bulk at a time?

src/pages/permissions/PermissionIdentities.tsx
Kxiru: As above.

src/pages/permissions/PermissionIdentities.tsx
mas-who: I'm not sure if I follow, fine-grained means oidc entities + new TLS identities. We can't select legacy TLS identities and delete/modify them. The count here represents the total "selectable" identities, so it's not related to the API code which is fine

mas-who:
**QA comments**

1. When the screen is large enough, I think the link chip should show the full content and not truncate. Currently, the chip for tls certificate identities are truncated.
Screenshot from 2024-12-09 10-41-35

2. When bulk editing identity groups, upon completion, the tls identities are de-selected whereas the oidc identities remain selected. We should keep things consistent i.e. keep identities selected after modifying groups which corresponds to the original behaviour.
[Screencast from 09-12-2024 10:48:20.webm](https://github.com/user-attachments/assets/6dc50289-d8b8-4990-81a3-8fc9c4d18879)

3. The table header seems incorrect when selecting multiple identities e.g. "3 identities selected. Select all 5 identities" (see example below). The correct heading should mention "Select all 3 identities" which corresponds to the maximum number of selectable identities in the table.
Screenshot from 2024-12-09 10-52-17

4. The notification message for deleting an identity still includes the auth method, it should be removed. Furthermore, the identity I deleted in the example below is a tls certificate, the icon included in the notification is incorrect.
Screenshot from 2024-12-09 10-56-29

src/pages/permissions/PermissionIdentities.tsx
mas-who: nothing changed here

src/pages/permissions/PermissionIdentities.tsx
mas-who: nothing changed here

src/pages/permissions/PermissionIdentities.tsx
mas-who: nothing changed here.

mas-who (commented):
@Kxiru I think you didn't push the changes for the review comments. Please would you double check that. I have unresolved the comments for now.

Kxiru:
> @Kxiru I think you didn't push the changes for the review comments. Please would you double check that. I have unresolved the comments for now.

Apologies, you are right, I hadn't pushed the changes. I will review the QA comments and push them all together.

Kxiru:
> **QA comments**
> 
> 1. When the screen is large enough, I think the link chip should show the full content and not truncate. Currently, the chip for tls certificate identities are truncated.
>    ...

Comments 3 and 4 have been addressed in my last commit. Comment no.2 can not be recreated on my machine and therefore I believe that I have addressed this in my last commit as well. I'll address the dynamic identity chips now :).

src/pages/permissions/PermissionIdentities.tsx
mas-who: This is still using `ResourceLink` and not `ResourceLabel`?

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -294,7 +306,7 @@ const PermissionIdentities: FC = () => {
...
```
mas-who, line 309: I think this should also be `nonLegacyIdentities`

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -198,10 +210,10 @@ const PermissionIdentities: FC = () => {
     defaultSort: "name",
   });
 
-  // NOTE: tls user group membership cannot be modified, this will be supported in the future
-  const nonTlsUsers = identities.filter((identity) => {
-    const isTlsIdentity = identity.authentication_method === "tls";
-    return !isTlsIdentity;
+  const nonLegacyIdentities = identities.filter((identity) => {
```
mas-who, line 213: can we call this `fineGrainedIdentities` to be consistent?

mas-who (commented):
Almost there, some small nits

src/pages/permissions/PermissionIdentities.tsx
Kxiru: I think this comment applies to an earlier commit because this is reflected on my end.

src/pages/permissions/actions/DeleteIdentityBtn.tsx
```diff
@@ -33,7 +33,15 @@ const DeleteIdentityBtn: FC<Props> = ({ identity }) => {
         });
         toastNotify.success(
           <>
-            Identity <ResourceLabel type={"idp-group"} value={identity.name} />{" "}
+            Identity{" "}
+            <ResourceLabel
+              type={
+                identity.authentication_method == "tls"
```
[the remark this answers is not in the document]
Kxiru: Nice idea!

src/pages/permissions/PermissionIdentities.tsx
mas-who: restricted can be a bit confusing, since legacy certificates can still be restricted by projects.

src/components/ResourceLabel.tsx
```diff
@@ -1,16 +1,18 @@
 import { FC } from "react";
 import ResourceIcon, { ResourceIconType } from "./ResourceIcon";
+import classNames from "classnames";
 
 interface Props {
   type: ResourceIconType;
   value: string;
   bold?: boolean;
+  fullWidth?: boolean;
```
mas-who, line 9: better call this prop `truncate` instead `fullWidth` I rate, right now we would truncate the content if `fullWidth` is true and that's the opposite behaviour in my mind.

Also, the prop should be set to default `true`. When adjusting existing parameters for functions/components, we should always consider how it could affect the rest of the codebase. The current change will result in pre-existing truncation to be removed everywhere else in the UI.

mas-who:
Found another minor QA issue, you may need to remove the margin below the buttons for a fine-grained identity. Currently it's causing the rows in the identities table to have unequal height.

Screenshot from 2024-12-09 17-11-29

mas-who:
After completing group modification for mixed type identities, I am still seeing that tls identities gets de-selected.

[Screencast from 09-12-2024 17:20:37.webm](https://github.com/user-attachments/assets/096249ce-883a-49e2-bb6b-9f99066b9526)

edlerd (commented):
Two ideas to avoid duplication and make the code a bit cleaner.

Kxiru:
Addiitonal feature: Enable user feedback when the user is deleitng an identity such as a spinner.

src/components/IdentityResource.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { LxdIdentity } from "types/permissions";
+import ResourceLabel from "./ResourceLabel";
+
+interface Props {
+  identity: LxdIdentity;
+  truncate?: boolean;
+}
+
+const IdentityResource: FC<Props> = ({ identity, truncate }) => {
+  const authMethod =
```
mas-who, line 11:
```suggestion
  const identityIconType =
```
variable naming clearer

src/pages/permissions/actions/BulkDeleteIdentitiesBtn.tsx
mas-who, line 1: Based on the discussion from yesterday, the case for indicating loading state with bulk deletion is not yet covered here. We can adopt the same approach as the bulk group modification, add a loading state to the delete confirmation button

mas-who:
@Kxiru just noticed the tests are failing, looking at the failure, it's probably due to the change where we no longer reset identities selection after modifying their groups

src/pages/permissions/actions/BulkDeleteIdentitiesBtn.tsx
Kxiru: Yes, I believe I only added the loading state to the individual identity deletion button. Thank you for the reminder.

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -52,9 +54,7 @@ const PermissionIdentities: FC = () => {
   const { hasAccessManagementTLS } = useSupportedFeatures();
 
   useEffect(() => {
-    const validIdentities = new Set(
-      identities.map((identity) => identity.name),
-    );
+    const validIdentities = new Set(identities.map((identity) => identity.id));
```
edlerd, line 57:
```suggestion
    const validIdentityIds = new Set(identities.map((identity) => identity.id));
```
I think we should be precise on what this is containing -- not the identity objects, but ids.

src/pages/permissions/actions/DeleteIdentityBtn.tsx
```diff
@@ -20,9 +20,11 @@ const DeleteIdentityBtn: FC<Props> = ({ identity }) => {
   const queryClient = useQueryClient();
   const notify = useNotify();
   const toastNotify = useToastNotification();
+  const [isDeleting, setDeleting] = useState(false);
```
edlerd, line 23:
```suggestion
```
I think we should use `formik.isSubmitting` and avoid the custom state here.

src/pages/permissions/panels/EditIdentitiesForm.tsx
```diff
@@ -36,8 +36,8 @@ const EditIdentitiesForm: FC<Props> = ({
     notify.failure("Loading details failed", error);
   }
 
-  const nonTlsIdentities = identities.filter(
-    (identity) => identity.authentication_method !== "tls",
+  const fineGrainedIdentities = identities.filter(
+    (identity) => identity.type !== "Client certificate (unrestricted)",
```
edlerd, line 40: I think we should reuse the `isUnrestricted` function here as well to establish a single source of truth.

src/pages/permissions/actions/BulkDeleteIdentitiesBtn.tsx
edlerd: +1 discovered this as well :)

edlerd (commented):
Thanks for the iteration on this one. It seems very good already, a handful of comments for things that I discovered that can be a bit simpler or improved slightly. Most are tiny changes.

src/components/ResourceLabel.tsx
```diff
@@ -1,16 +1,18 @@
...
   bold?: boolean;
+  truncate?: boolean;
 }
 
-const ResourceLabel: FC<Props> = ({ type, value, bold }) => {
+const ResourceLabel: FC<Props> = ({ type, value, bold, truncate = true }) => {
   const ValueWrapper = bold ? "strong" : "span";
   return (
-    <span className="resource-label u-truncate">
+    <span className={classNames("resource-label", { "u-truncate": truncate })}>
```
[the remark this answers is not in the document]
Kxiru: Perhaps the title should be set regardless of truncation?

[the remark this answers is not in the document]
Kxiru: Even better, I have just checked, and there is a title set regardless!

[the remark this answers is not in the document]
edlerd: Yes, that title is better to always have.

src/pages/permissions/actions/DeleteIdentityBtn.tsx
edlerd: Right, this is just using a callback and no formik. All good then :+1:

PR #1011, opened 2024-11-26
volume snapshots can have an empty expiry

edlerd:
## Done

- fix volume snapshots, they can have an empty expiry, and we should show that correctly

## QA

1. Run the LXD-UI:
    - create a custom volume, create snapshots without a name or expiry on that volume, browse the snapshot list of that volume.

## Screenshots

Before:

image

After:

image


mas-who (approved):
LGTM :+1: interesting discrepancy in api response with instance snapshots

PR #1015, opened 2024-11-29
Add airgap support

edlerd:
## Done

- embed font, so in airgapped environments we don't try to load from assets.ubuntu.com
- avoid refetch of image index, so the image list appears immediately if the indexes can't be fetched
- remove unused icon

## QA

1. Run the LXD-UI:
    - check the ui for any third party requests. The font should be loaded locally. When blocking other domains, the image selector should load immediately


src/pages/images/ImageSelector.tsx
```diff
@@ -81,17 +81,20 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
   const { data: canonicalImages = [], isLoading: isCiLoading } = useQuery({
     queryKey: [queryKeys.images, canonicalServer],
     queryFn: () => loadImages(canonicalJson, canonicalServer),
+    retry: false, // avoid retry to ease experience in airgapped deployments
```
mas-who, line 84: Is this maybe not too drastic? not sure if there is a better way though, since ideally we should be aware if we are in an airgapped environment.

edlerd: I want to avoid the complexity of another configuration key. I think the UI should just work in airgap without needing to do anything by user or admin.

PR #1019, opened 2024-12-02
feat: add support for instance/profile host path devices [WD-17682]

### LINKED ISSUE #1018: path disk device shows undefined pool in gui
image

links to `/ui/project/default/storage/pool/undefined`

Path passthrough storage is not part of a pool and should not show `on pool <pool>`. Maybe it would be better to show `Host path /data/media/` instead of `Volume /data/media` when it is a path mount
image

yaml config for clarity:
```
devices:
  nfs-media:
    path: /data/media
    source: /data/media
    type: disk
```

mas-who:
## Done

- Adjusted devices table in instance/profile overview page to show host path disk devices
- Added support for create/edit host path disk devices in instance/profile configuration pages
- Adjusted the workflow for adding disk devices, now it is a multi-step flow, starting with selecting the type of disk device (volume or host path).
- Adjusted/added e2e tests to reflect the above changes.

## QA

1. Run the LXD-UI:
    - Create a new profile, add both custom volume and host path disk devices to it.
    - Create an instance and apply the profile to that instance, make sure all disk devices are inherited.
    - Create custom volume and host path disk devices for the above instance, using different volume and mount point.
    - Start the instance, go to its terminal and make sure that path corresponding to every disk device exists. Make sure that the mounted host path has the correct content (should be the same to the path on your machine).
    - Go to the instance/profile overview page, check the device list table and make sure disk devices are displayed correctly.
    - Try modifying the volume and host path disk devices on the profile and instance.

### DIFF
--- a/src/components/DeviceListTable.tsx
+++ b/src/components/DeviceListTable.tsx
@@ -3,7 +3,12 @@ import { MainTable } from "@canonical/react-components";
 import { isRootDisk } from "util/instanceValidation";
 import { FormDevice } from "util/formDevices";
 import ResourceLink from "components/ResourceLink";
-import { isOtherDevice } from "util/devices";
+import {
+  isDiskDevice,
+  isHostDiskDevice,
+  isOtherDevice,
+  isVolumeDevice,
+} from "util/devices";
 import DeviceDetails from "./DeviceDetails";
 import { useParams } from "react-router-dom";
 import { LxdDeviceValue, LxdDevices } from "types/device";
@@ -47,6 +52,21 @@ const DeviceListTable: FC<Props> = ({ configBaseURL, devices }) => {
   ];
 
   const deviceRows = overviewDevices.map(([devicename, device]) => {
+    let deviceType = device.type;
+    if (isDiskDevice(device)) {
+      if (isRootDisk(device as FormDevice)) {
+        deviceType += " (root)";
+      }
+
+      if (isVolumeDevice(device)) {
+        deviceType += " (volume)";
+      }
+
+      if (isHostDiskDevice(device)) {
+        deviceType += " (host)";
+      }
+    }
+
     return {
       columns: [
         {
@@ -61,8 +81,7 @@ const DeviceListTable: FC<Props> = ({ configBaseURL, devices }) => {
           "aria-label": "Name",
         },
         {
-          content: `${device.type} ${isRootDisk(device as FormDevice) ? "(root)" : ""}`,
-
+          content: deviceType,
           role: "cell",
           "aria-label": "Type",
         },
--- a/src/components/DeviceListTable.tsx
+++ b/src/components/DeviceListTable.tsx
@@ -63,7 +63,7 @@ const DeviceListTable: FC<Props> = ({ configBaseURL, devices }) => {
       }
 
       if (isHostDiskDevice(device)) {
-        deviceType += " (host)";
+        deviceType += " (host path)";
       }
     }
 
--- a/src/components/forms/DiskDeviceFormCustom.tsx
+++ b/src/components/forms/DiskDeviceFormCustom.tsx
@@ -16,7 +16,6 @@ import { MainTableRow } from "@canonical/react-components/dist/components/MainTa
 import { getConfigurationRowBase } from "components/ConfigurationRow";
 import DetachDiskDeviceBtn from "pages/instances/actions/DetachDiskDeviceBtn";
 import classnames from "classnames";
-import { LxdStorageVolume } from "types/storage";
 import {
   isDiskDeviceMountPointMissing,
   isRootDisk,
@@ -27,6 +26,7 @@ import { LxdProfile } from "types/profile";
 import { focusField } from "util/formFields";
 import AttachDiskDeviceBtn from "pages/storage/AttachDiskDeviceBtn";
 import { LxdDiskDevice } from "types/device";
+import { LxdStorageVolume } from "types/storage";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
@@ -38,22 +38,12 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
   const readOnly = (formik.values as EditInstanceFormValues).readOnly;
   const existingDeviceNames = getExistingDeviceNames(formik.values, profiles);
 
-  const addDiskDevice = (device: LxdStorageVolume | LxdDiskDevice) => {
+  const addDiskDevice = (device: LxdDiskDevice) => {
     const copy = [...formik.values.devices];
-    const isVolume = "content_type" in device;
     const newDevice: FormDevice = {
-      type: "disk",
+      ...device,
       name: deduplicateName("disk-device", 1, existingDeviceNames),
     };
-
-    if (isVolume) {
-      newDevice.pool = device.pool;
-      newDevice.source = device.name;
-      newDevice.path = device.content_type === "filesystem" ? "" : undefined;
-    } else {
-      newDevice.source = device.source;
-      newDevice.path = device.path;
-    }
 
     copy.push(newDevice);
     void formik.setFieldValue("devices", copy);
@@ -62,31 +52,25 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
     focusField(name);
   };
 
-  const changeDiskDevice = (
-    diskDevice: LxdStorageVolume | LxdDiskDevice,
-    formDiskDevice: FormDiskDevice,
+  const changeVolume = (
+    volume: LxdStorageVolume,
+    existingVolume: FormDiskDevice,
     index: number,
   ) => {
-    const isVolume = "content_type" in diskDevice;
-    if (isVolume) {
-      void formik.setFieldValue(`devices.${index}.pool`, diskDevice.pool);
-      void formik.setFieldValue(`devices.${index}.source`, diskDevice.name);
+    void formik.setFieldValue(`devices.${index}.pool`, volume.pool);
+    void formik.setFieldValue(`devices.${index}.source`, volume.name);
 
-      if (
-        diskDevice.content_type === "filesystem" &&
-        formDiskDevice.path === undefined
-      ) {
-        void formik.setFieldValue(`devices.${index}.path`, "");
-      }
-      if (diskDevice.content_type === "block") {
-        void formik.setFieldValue(`devices.${index}.path`, undefined);
-      }
-
-      return;
+    if (
+      volume.content_type === "filesystem" &&
+      existingVolume.path === undefined
+    ) {
+      void formik.setFieldValue(`devices.${index}.path`, "");
     }
 
-    void formik.setFieldValue(`devices.${index}.source`, diskDevice.source);
-    void formik.setFieldValue(`devices.${index}.path`, diskDevice.path);
+    // If path must not exist for the device, remote it even if it's set for the existing device
+    if (volume.content_type === "block") {
+      void formik.setFieldValue(`devices.${index}.path`, undefined);
+    }
   };
 
   const editButton = (fieldName: string) => (
@@ -138,75 +122,79 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
       }),
     );
 
-    const volumeDeviceSource = getConfigurationRowBase({
-      className: "no-border-top inherited-with-form",
-      configuration: (
-        <Label forId={`devices.${index}.pool`}>Pool / volume</Label>
-      ),
-      inherited: (
-        <div className="custom-disk-volume-source">
-          <div
-            className={classnames("mono-font", "u-truncate")}
-            title={`${item.pool} / ${item.source ?? ""}`}
-          >
-            <b>
-              {item.pool} / {item.source}
-            </b>
+    const volumeDeviceSource = () =>
+      getConfigurationRowBase({
+        className: "no-border-top inherited-with-form",
+        configuration: (
+          <Label forId={`devices.${index}.pool`}>Pool / volume</Label>
+        ),
+        inherited: (
+          <div className="custom-disk-volume-source">
+            <div
+              className={classnames("mono-font", "u-truncate")}
+              title={`${item.pool} / ${item.source ?? ""}`}
+            >
+              <b>
+                {item.pool} / {item.source}
+              </b>
+            </div>
+            <CustomVolumeSelectBtn
+              formik={formik}
+              project={project}
+              setValue={(volume) => {
+                ensureEditMode(formik);
+                changeVolume(volume, item, index);
+              }}
+              buttonProps={{
+                id: `devices.${index}.pool`,
+                appearance: "base",
+                className: "u-no-margin--bottom",
+                title: "Select storage volume",
+                dense: true,
+              }}
+            >
+              <Icon name="edit" />
+            </CustomVolumeSelectBtn>
           </div>
-          <CustomVolumeSelectBtn
-            formik={formik}
-            project={project}
-            setValue={(volume) => {
-              ensureEditMode(formik);
-              changeDiskDevice(volume, item, index);
-            }}
-            buttonProps={{
-              id: `devices.${index}.pool`,
-              appearance: "base",
-              className: "u-no-margin--bottom",
-              title: "Select storage volume",
-              dense: true,
-            }}
-          >
-            <Icon name="edit" />
-          </CustomVolumeSelectBtn>
-        </div>
-      ),
-      override: "",
-    });
+        ),
+        override: "",
+      });
 
-    const hostDeviceSource = getConfigurationRowBase({
-      className: "no-border-top inherited-with-form",
-      configuration: <Label forId={`devices.${index}.source`}>Host path</Label>,
-      inherited: readOnly ? (
-        <div className="custom-disk-read-mode">
-          <div className="mono-font custom-disk-value u-truncate">
-            <b>{item.source}</b>
+    const hostDeviceSource = () =>
+      getConfigurationRowBase({
+        className: "no-border-top inherited-with-form",
+        configuration: (
+          <Label forId={`devices.${index}.source`}>Host path</Label>
+        ),
+        inherited: readOnly ? (
+          <div className="custom-disk-read-mode">
+            <div className="mono-font custom-disk-value u-truncate">
+              <b>{item.source}</b>
+            </div>
+            {editButton(`devices.${index}.source`)}
           </div>
-          {editButton(`devices.${index}.source`)}
-        </div>
-      ) : (
-        <Input
-          id={`devices.${index}.source`}
-          name={`devices.${index}.source`}
-          onBlur={formik.handleBlur}
-          onChange={(e) => {
-            void formik.setFieldValue(
-              `devices.${index}.source`,
-              e.target.value,
-            );
-          }}
-          value={item.source}
-          type="text"
-          placeholder="Enter full host path (e.g. /data)"
-          className={!item.source ? undefined : "u-no-margin--bottom"}
-          error={!item.source ? "Host path is required" : undefined}
-        />
-      ),
-      override: "",
-    });
+        ) : (
+          <Input
+            id={`devices.${index}.source`}
+            name={`devices.${index}.source`}
+            onBlur={formik.handleBlur}
+            onChange={(e) => {
+              void formik.setFieldValue(
+                `devices.${index}.source`,
+                e.target.value,
+              );
+            }}
+            value={item.source}
+            type="text"
+            placeholder="Enter full host path (e.g. /data)"
+            className={!item.source ? undefined : "u-no-margin--bottom"}
+            error={!item.source ? "Host path is required" : undefined}
+          />
+        ),
+        override: "",
+      });
 
-    rows.push(isVolumeDevice(item) ? volumeDeviceSource : hostDeviceSource);
+    rows.push(isVolumeDevice(item) ? volumeDeviceSource() : hostDeviceSource());
 
     if (!isVolumeDevice(item) || item.path !== undefined) {
       const hasError = isDiskDeviceMountPointMissing(formik, index);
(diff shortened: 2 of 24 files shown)


edlerd (changes requested):
This looks good on the overview page, but on the [configuration page](https://lxd-ui-1019.demos.haus/ui/project/default/profile/other-device-profile/configuration/disk) it is still a bit broken:

image

I think the easiest solution is to filter the host path devices from displaying in the configuration form.

edlerd:
Edit: the resource link for the host path, should then direct to the yaml editor.

mas-who:
> This looks good on the overview page, but on the [configuration page](https://lxd-ui-1019.demos.haus/ui/project/default/profile/other-device-profile/configuration/disk) it is still a bit broken:
> 
> image
> 
> I think the easiest solution is to filter the host path devices from displaying in the configuration form.

Aaah yes didn't catch this one. I don't think filtering them is correct though, since these are still legit disk devices. Perhaps instead of "Pool/Volume", we can have "Host path" and remove Mount point. Also we can disable the edit functionality for a host path disk device for now. Wdyt?

edlerd:
> Aaah yes didn't catch this one. I don't think filtering them is correct though, since these are still legit disk devices. Perhaps instead of "Pool/Volume", we can have "Host path" and remove Mount point. Also we can disable the edit functionality for a host path disk device for now. Wdyt?

We could, then we'd half support them :) Ideally we allow adding them and editing as well, I wanted to keep the scope small, so hiding seemed like the easiest fix. Up to you how to solve it. Just having a read only would work as well, but then should add a way that guides to the YAML editor for editing it.

mas-who:
> > Aaah yes didn't catch this one. I don't think filtering them is correct though, since these are still legit disk devices. Perhaps instead of "Pool/Volume", we can have "Host path" and remove Mount point. Also we can disable the edit functionality for a host path disk device for now. Wdyt?
> 
> We could, then we'd half support them :) Ideally we allow adding them and editing as well, I wanted to keep the scope small, so hiding seemed like the easiest fix. Up to you how to solve it. Just having a read only would work as well, but then should add a way that guides to the YAML editor for editing it.

I think adding/editing host path disk devices may require some design input. For example, maybe we follow a similar approach as migrate instance modal where we have multiple steps, and the first step would be to choose the type of disk device to add. Furthermore, to add/edit a host path disk device, we can let the user to provide input using a directory picker. 

In any case, for now limiting the cope to read only seems appropriate with the redirection to the yaml editor. I can create an issue to do a follow up task for the new disk device type?

mas-who:
@edlerd please do wait a bit to review this PR, I'm not entirely happy with the implementation yet, will remove draft status once it's ready

src/pages/storage/CustomVolumeModal.tsx
```diff
@@ -34,14 +37,25 @@ const CustomVolumeModal: FC<Props> = ({
...
+    setContent(SELECT_VOLUME);
+  };
+
+  let modalTitle = title ?? "Choose custom volume";
+  if (content === CREATE_VOLUME) {
+    modalTitle = title ? (
+      <BackLink
+        title={"Create volume"}
+        onClick={handleGoBack}
+        linkText={"Attach custom volume device"}
```
edlerd, line 50:
```suggestion
        linkText="Attach custom volume"
```
This can be a bit shorter, and can you remove the `{` from all the string values?

src/pages/storage/HostPathDeviceModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+      title={title}
+      buttonRow={
+        <>
+          <Button
+            appearance="base"
+            className="u-no-margin--bottom"
+            type="button"
+            onClick={onCancel}
+          >
+            Cancel
```
edlerd, line 49:
```suggestion
            Back
```
I think this should say back, and we should also change the caption in the volume case, which has possibly 3 steps deep with volume creation. Because on cancel, I'd expect the modal to close, but we are going back to the previous step.

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+          close={close}
+          className="migrate-instance-modal"
+          title={modalTitle}
+          onKeyDown={handleEscKey}
+        >
+          {!type && (
+            <div className="choose-migration-type">
+              <FormLink
+                icon="add-logical-volume"
+                title="Attach custom volume device"
```
edlerd, line 64:
```suggestion
                title="Attach custom volume"
```
This can be shorter as well.

src/components/DeviceListTable.tsx
```diff
@@ -47,6 +52,21 @@ const DeviceListTable: FC<Props> = ({ configBaseURL, devices }) => {
...
+      if (isRootDisk(device as FormDevice)) {
+        deviceType += " (root)";
+      }
+
+      if (isVolumeDevice(device)) {
+        deviceType += " (volume)";
+      }
+
+      if (isHostDiskDevice(device)) {
+        deviceType += " (host)";
```
edlerd, line 66:
```suggestion
        deviceType += " (host path)";
```
Might be easier to understand. Wdyt?

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -32,40 +36,57 @@ interface Props {
...
-    .filter((item) => item.type === "disk" && !isRootDisk(item))
-    .map((device) => device as FormDiskDevice);
-
   const existingDeviceNames = getExistingDeviceNames(formik.values, profiles);
 
-  const addVolume = (volume: LxdStorageVolume) => {
+  const addDiskDevice = (device: LxdStorageVolume | LxdDiskDevice) => {
     const copy = [...formik.values.devices];
-    copy.push({
+    const isVolume = "content_type" in device;
```
edlerd, line 43: I think this should be a helper function to be reused, like `isVolume(device)`

src/pages/storage/HostPathDeviceModal.tsx
edlerd: I think we should distinguish in the logic between this back button and the x of the modal. The x should close it directly, this back button is a way to got to the previous step.

src/pages/instances/actions/DetachDiskDeviceBtn.tsx
```diff
@@ -10,15 +10,15 @@ const DetachDiskDeviceBtn: FC<Props> = ({ onDetach }) => {
     <ConfirmationButton
       appearance="base"
       type="button"
-      title="Detach volume"
+      title="Detach disk devicve"
       className="has-icon u-no-margin--bottom is-dense"
       confirmationModalProps={{
-        title: "Confirm volume detach",
+        title: "Confirm disk device detach",
```
edlerd, line 16:
```suggestion
        title: "Confirm disk detachment",
```

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const handleGoBack = () => {
+    if (type) {
+      setType("");
+      return;
+    }
+  };
+
+  const modalTitle = !type ? (
+    "Choose disk device type"
```
edlerd, line 42:
```suggestion
    "Choose type"
```
I think the options and the context make it clear this is about disk devices :)

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+  ) : (
+    <BackLink
+      title={`Attach ${type} device`}
```
edlerd, line 45:
```suggestion
      title={`Attach ${type}`}
```

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+      onClick={handleGoBack}
+      linkText={"Choose disk device type"}
```
edlerd, line 47:
```suggestion
      linkText="Choose type"
```

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+          {!type && (
+            <div className="choose-migration-type">
+              <FormLink
+                icon="add-logical-volume"
+                title="Attach custom volume device"
+                onClick={() => setType("custom volume")}
+              />
+              <FormLink
+                icon="mount"
+                title="Attach host path device"
```
edlerd, line 69:
```suggestion
                title="Mount host path"
```

src/pages/storage/HostPathDeviceModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+const HostPathDeviceModal: FC<Props> = ({
+  formik,
+  onFinish,
+  onCancel,
+  title,
+}) => {
+  const [source, setSource] = useState("");
+  const [path, setPath] = useState<string>();
```
edlerd, line 20:
```suggestion
  const [source, setSource] = useState("");
  const [path, setPath] = useState("");
```
Shouldn't they be the same?

src/pages/storage/HostPathDeviceModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const touchedRef = useRef({
+    source: false,
+    path: false,
+  });
```
edlerd, line 24: Why use a ref here and not a normal state?

src/pages/storage/HostPathDeviceModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }}
+        type="text"
+        label="Host path"
+        required
+        error={
+          !source && touchedRef.current.source
+            ? "Host path is required"
+            : undefined
+        }
+      />
```
edlerd, line 78: Can this one autofocus after render?

edlerd (commented):
This is a really nice interaction now. Some ideas to refine the UX a bit and some other ideas about code. Most of them are very small suggestions

src/pages/storage/AttachDiskDeviceModal.tsx
edlerd: On second thought: "Choose disk type" might be better.

src/components/DeviceListTable.tsx
mas-who: I was thinking the detail column displays a path, so seems a little redundant to have `(host path)` in the type column?

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -113,46 +138,77 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+          type="text"
+          placeholder="Enter full host path (e.g. /data)"
+          className={!item.source ? undefined : "u-no-margin--bottom"}
+          error={!item.source ? "Host path is required" : undefined}
+        />
+      ),
+      override: "",
+    });
+
+    rows.push(isVolumeDevice(item) ? volumeDeviceSource : hostDeviceSource);
```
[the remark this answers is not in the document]
mas-who: good point, turned them into functions as suggested :+1:

src/pages/storage/HostPathDeviceModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <Button
+            appearance="base"
+            className="u-no-margin--bottom"
+            type="button"
+            onClick={onCancel}
+          >
+            Cancel
+          </Button>
+          <Button
+            appearance=""
```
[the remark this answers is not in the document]
mas-who: My thinking is that we generally use green to indicate a create/update action will be performed in the backend. In this case we are only adding the device in the UI before saving the changes, so maybe green isn't the best fit here?

src/pages/storage/AttachDiskDeviceModal.tsx
edlerd: Yes, agreed.

src/pages/storage/HostPathDeviceModal.tsx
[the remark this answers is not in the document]
edlerd: true, let's keep it as is.

src/components/DeviceListTable.tsx
edlerd: My concern is this: Just because there is a path next to the type, it doesn't fully explain the path refers to the host. It could be understood as the mount path as well.

mas-who: Yes agreed, adding `path` :+1:

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { Modal } from "@canonical/react-components";
+import FormLink from "components/FormLink";
+import BackLink from "components/BackLink";
+import { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import CustomVolumeModal from "./CustomVolumeModal";
+import { LxdStorageVolume } from "types/storage";
+import { LxdDiskDevice } from "types/device";
+import HostPathDeviceModal from "./HostPathDeviceModal";
+
+type DiskDeviceType = "custom volume" | "host path" | "";
```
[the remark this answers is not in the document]
mas-who: yeah clearer, I've made the relevant changes

src/pages/storage/AttachDiskDeviceModal.tsx
mas-who: the back link text will also be `Choose disk type` to align with title from the previous step

src/pages/storage/HostPathDeviceModal.tsx
mas-who: They could be, used ref here since it does not cause unnecessary renders

src/util/helpers.tsx
```diff
@@ -333,5 +333,5 @@ export const getDefaultStoragePool = (profile: LxdProfile) => {
     .find((device) => {
       return isRootDisk(device as FormDevice);
     });
-  return rootStorage ? rootStorage.pool : "";
+  return rootStorage ? (rootStorage.pool ?? "") : "";
 };
```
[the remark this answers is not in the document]
mas-who: oops, thanks!

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const handleEscKey = (e: KeyboardEvent<HTMLElement>) => {
+    if (e.key === "Escape") {
+      close();
+    }
+  };
+
+  const handleGoBack = () => {
+    setType("choose type");
+    return;
```
edlerd, line 36:
```suggestion
```

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  const modalTitle =
+    type === "choose type" ? (
+      "Choose disk type"
+    ) : (
+      <BackLink
+        title={`Attach ${type}`}
```
edlerd, line 44: I think this should say "Mount host path" in one case and "Attach custom volume" in the other.

src/util/devices.tsx
```diff
@@ -98,3 +111,9 @@ export const getExistingDeviceNames = (
 
   return existingDeviceNames;
 };
+
+export const isVolume = (
```
edlerd, line 115: What is the difference between isVolume and isVolumeDevice? Can they be combined to a single function?

src/util/instanceValidation.tsx
```diff
@@ -18,7 +18,7 @@ export const hasNoRootDisk = (
   return missingRoot(values.devices) && !inheritsRoot(values, profiles);
 };
 
-export const isRootDisk = (device: FormDevice): device is FormDiskDevice => {
+export const isRootDisk = (device: FormDevice): boolean => {
```
edlerd, line 21: I think this should be reverted. This allowed for an implicit cast, so after `devices.filter(isRootDisk).map((device) => ...` typescript knows that device is of type `FormDiskDevice`.

src/pages/storage/CustomVolumeModal.tsx
```diff
@@ -21,6 +24,8 @@ const CustomVolumeModal: FC<Props> = ({
   project,
   onFinish,
   onCancel,
```
edlerd, line 26: I wonder if we should rename those to `onGoBack` or something that is avoiding the ambiguity between onCancel and onClose.

edlerd (commented):
Thanks for the quick iteration, just a handful of ideas then this should be good to merge.

src/util/devices.tsx
mas-who: slightly tricky from a typescript perspective, this function operates on `LxdStorageVolume` which is very different to the device types that `isVolumeDevice` works with

src/util/instanceValidation.tsx
mas-who: I don't think it's appropriate to have the implicit cast with `isRootDisk` since it's a sub variant of `FormDiskDevice`. I've created the function `isFormDiskDevice` for this purpose. With the old approach, the following condition may result in `device` to get casted with the `never` type.

```
if (device.type !== "disk" || isRootDisk(device)) {
      continue;
    }
```

src/pages/storage/CustomVolumeModal.tsx
mas-who: I think the logic originates from the parent, the modal does not always "go back" so `onCancel` seems like a better fit. For example, when the modal is opened by clicking on the edit icon for an existing volume device, clicking the cancel button will close the modal

src/util/devices.tsx
edlerd: Looking at the two usages of this function, I think we can simplify, leading to a removal of the function.

1. the `changeDiskDevice` function can only ever receive a `LxdStorageVolume` as the `diskDevice`. So we should remove the ambiguous other possible type of `LydDiskDevice` for the `diskDevice`. Then rename the parameter there from `diskDevice` to `volume` and we don't need to call `isVolume` inside.

2.  the `addDiskDevice` function should not receive a parameter `device: LxdStorageVolume | LxdDiskDevice`, because a volume is not a device. I think in `AttachDiskDeviceBtn` we should change the prop to `setValue: (device: LxdDiskDevice) => void`. Inside the `AttachDiskDeviceBtn::handleFinish` we can do the translation from volume to `LxdDiskDevice`. If we fill the parameters correctly, we have no branching any more in `addDiskDevice`. This removes the ambiguity and the second use case of `isVolume`.

Then we can remove `isVolume`.

edlerd: For 2. the translation needs to happen in `AttachDiskDeviceModal`, one more layer downwards, where we receive the volume from `CustomVolumeModal`. Only there it is not ambiguous. In the `AttachDiskDeviceBtn` we would also have to differentiate between disk and volume.

mas-who: Thanks this is great, regarding 1. I must have been out of it, not sure how it ended up like that!

mas-who: One note, I've made the change but moved the translation down into `CustomVolumeModal`. That way, both `HostPathDeviceModal` and `CustomVolumeModal` will pass a `LxdDiskDevice` back up to the `DiskDeviceFormCustom` component for both adding and editing disk devices regardless of the disk type. Let me know what you think.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -32,38 +35,38 @@ interface Props {
...
+    // otherwise keep the existing value
+    if (newDevice.path === "" && existingDevice.path === undefined) {
       void formik.setFieldValue(`devices.${index}.path`, "");
     }
-    if (volume.content_type === "block") {
+
+    // If path must not exist for the device, remote it even if it's set for the existing device
+    if (newDevice.path === undefined) {
       void formik.setFieldValue(`devices.${index}.path`, undefined);
     }
```
edlerd, line 71: This is a bit confusing. Can you describe in the comments why this is needed and avoid describing what is happening? I think for the what, the code is easy enough to gasp it with a glance.

src/util/devices.tsx
edlerd: I am not sure raising it to the CustomVolumeModal is the right choice. Maybe that component should not deal with the concept of devices. This makes the change device function a bit complicated to gasp. Maybe keep on drilling the volume down to the change device function, this might make it more obvious? Or there coudl be a way to make the change function easier to understand.

edlerd (commented):
Thanks for the change, I find the code still got a bit too complicated around change device. Maybe it can be improved?

src/components/forms/DiskDeviceFormCustom.tsx
mas-who: Yeah looking at it again, the fact that path === "" means storage volume has file system content type is completely lost here. Only way is to move the device translation up one more layer

src/util/devices.tsx
mas-who: I've moved the translation into `AttachDiskDeviceModal` and changed the function in `DiskDeviceFormCustom` to `changeVolume`.

src/components/forms/DiskDeviceFormCustom.tsx
mas-who: done, this function is basically reverted to its original state

src/util/devices.tsx
edlerd: Thanks for the changes! I'll have a closer look at this shortly. To answer your question: That was a continuation of the thought process of the implications of having the translation in the choose custom volume component. It should be resolved with your recent changes.

edlerd: Actually no, only in the AttachDiskDeviceModal we have two branches where we receive the volume or the host path as a device object. My thought was that there is a natural distinction in the modal as it is receiving the two different types in different callbacks for the different parents.

In the button component we would have condensed that I to a single callback. Hope this makes sense.

edlerd: Still recent changes should be all right.

mas-who: Thanks for the quick feedback :) The only thing that I wasn't sure about is the comment regarding the button component. With that I think all is in order, this comment can be resolved if you are happy with the changes

edlerd (approved):
Code lgtm! Thanks for the additional features and making this nice and round :)

Didn't QA yet, if you are confident on the QA feel free to merge right away.

mas-who:
> Code lgtm! Thanks for the additional features and making this nice and round :)
> 
> Didn't QA yet, if you are confident on the QA feel free to merge right away.

I'm confident about the functionality as I've tested it extensively. However, I think it's still a good idea to have a QA pass, perhaps  just for things that are not covered in the PR description?

PR #1023, opened 2024-12-09
fix: show meter bar in storage list and other places

edlerd:
## Done

- show meter bar in storage list and other places, only hide it in the storage selector
- simplify code a bit

## QA

1. Run the LXD-UI:
    - ensure the storage meter appears in the pool list, migrate modals for instances and volumes and the storage selector works as expected.

### DIFF
--- a/src/pages/storage/StoragePoolOptionLabel.tsx
+++ b/src/pages/storage/StoragePoolOptionLabel.tsx
@@ -4,10 +4,9 @@ import StoragePoolSize from "./StoragePoolSize";
 
 interface Props {
   pool: LxdStoragePool;
-  hasMeterBar?: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


Kxiru (approved):
Looking good to me :).
Thank you for the code cleanup.

mas-who (approved):
LGTM :+1: Thanks for the fix.

One small question. What is the default that we want for the `StoragePoolSize` component? Is it with the bar hidden? If not, then maybe we should add the optional prop `hideMeterBar` to `StoragePoolSize` component instead?

edlerd:
> One small question. What is the default that we want for the `StoragePoolSize` component? Is it with the bar hidden? If not, then maybe we should add the optional prop `hideMeterBar` to `StoragePoolSize` component instead?

I think it is okay to opt in to the meter bar.

PR #1026, opened 2024-12-10
feat: [WD-16972] TLS user management spike.

Kxiru:
## Done
- Created "Create Identity" Button
- Create modal to create new TLS fine grained identities.
- A few new functions and a new LXDTrustToken type.

## User Management Spike
[✔️] A user needs to be able to create a TLS user (Creates a pending user)
[✔️] A user needs to be able to destroy a TLS user ([WD-16894](https://github.com/canonical/lxd-ui/pull/1008))

Fixes:
- Inability to create TLS fine grained tokens through the UI.

## QA

1. Run the LXD-UI:

## Screenshots
image
Screenshot from 2024-12-10 22-47-10

[WD-16894]: https://warthogs.atlassian.net/browse/WD-16894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/auth-identities.tsx
+++ b/src/api/auth-identities.tsx
@@ -1,6 +1,6 @@
 import { handleResponse, handleSettledResult } from "util/helpers";
 import { LxdApiResponse } from "types/apiResponse";
-import { LxdIdentity } from "types/permissions";
+import { LxdIdentity, TlsIdentityTokenDetail } from "types/permissions";
 
 export const fetchIdentities = (): Promise<LxdIdentity[]> => {
   return new Promise((resolve, reject) => {
@@ -72,6 +72,25 @@ export const deleteIdentities = (identities: LxdIdentity[]): Promise<void> => {
     )
       .then(handleSettledResult)
       .then(resolve)
+      .catch(reject);
+  });
+};
+
+export const createFineGrainedTlsIdentity = (
+  clientName: string,
+): Promise<TlsIdentityTokenDetail> => {
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/auth/identities/tls`, {
+      method: "POST",
+      body: JSON.stringify({
+        name: clientName,
+        token: true,
+      }),
+    })
+      .then(handleResponse)
+      .then((data: LxdApiResponse<TlsIdentityTokenDetail>) =>
+        resolve(data.metadata),
+      )
       .catch(reject);
   });
 };
--- a/src/pages/permissions/CreateIdentityModal.tsx
+++ b/src/pages/permissions/CreateIdentityModal.tsx
@@ -1,0 +1,145 @@
+import { FC, useState } from "react";
+import {
+  ActionButton,
+  Button,
+  Icon,
+  Input,
+  Modal,
+  Notification,
+} from "@canonical/react-components";
+import { useFormik } from "formik";
+import { createFineGrainedTlsIdentity } from "api/auth-identities";
+import { base64EncodeObject } from "util/helpers";
+import { useQueryClient } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+
+interface Props {
+  onClose: () => void;
+}
+
+const CreateIdentityModal: FC<Props> = ({ onClose }) => {
+  const [token, setToken] = useState<string | null>(null);
+  const [error, setError] = useState(false);
+  const [copied, setCopied] = useState(false);
+  const queryClient = useQueryClient();
+
+  const formik = useFormik({
+    initialValues: {
+      identityName: "",
+    },
+    onSubmit: (values) => {
+      setError(false);
+      createFineGrainedTlsIdentity(values.identityName)
+        .then((response) => {
+          const encodedToken = base64EncodeObject(response);
+          setToken(encodedToken);
+
+          void queryClient.invalidateQueries({
+            queryKey: [queryKeys.identities],
+          });
+        })
+        .catch(() => {
+          setError(true);
+        });
+    },
+  });
+
+  const handleCopy = async () => {
+    if (token) {
+      try {
+        await navigator.clipboard.writeText(token);
+        setCopied(true);
+
+        setTimeout(() => {
+          setCopied(false);
+        }, 5000);
+      } catch (error) {
+        console.error(error);
+      }
+    }
+  };
+
+  return (
+    <Modal
+      close={onClose}
+      className="create-tls-identity"
+      title={token ? "Identity has been created" : "Generate trust token"}
+      buttonRow={
+        <>
+          {token && (
+            <>
+              <Button
+                aria-label={
+                  copied ? "Copied to clipboard" : "Copy to clipboard"
+                }
+                title="Copy token"
+                className="u-no-margin--bottom"
+                onClick={() => handleCopy()}
+                type="button"
+                hasIcon
+              >
+                <Icon name={copied ? "task-outstanding" : "copy"} />
+              </Button>
+
+              <Button
+                aria-label="Close"
+                className="u-no-margin--bottom"
+                onClick={onClose}
+                type="button"
+              >
+                Close
+              </Button>
+            </>
+          )}
+          {!token && (
+            <ActionButton
+              aria-label="Generate token"
+              appearance="positive"
+              className="u-no-margin--bottom"
+              onClick={() => void formik.submitForm()}
+              disabled={
+                formik.values.identityName.length === 0 || formik.isSubmitting
+              }
+              loading={formik.isSubmitting}
+            >
+              Generate token
+            </ActionButton>
+          )}
+        </>
+      }
+    >
+      {error && (
+        <Notification
+          severity="negative"
+          title="Token creation failed"
+        ></Notification>
+      )}
+      {!token && (
+        <Input
+          id="identityName"
+          type="text"
+          label="Identity Name"
+          onBlur={formik.handleBlur}
+          onChange={formik.handleChange}
+          value={formik.values.identityName}
+        />
+      )}
+      {token && (
+        <>
+          <p>The token below can be used to log in as the new identity.</p>
+
+          <Notification
+            severity="caution"
+            title="Make sure to copy the token now as it will not be shown again."
+          ></Notification>
+
+          <div className="token-code-block">
+            <code>{token}</code>
+          </div>
+        </>
+      )}
+    </Modal>
+  );
+};
+
+export default CreateIdentityModal;
--- a/src/pages/permissions/PermissionIdentities.tsx
+++ b/src/pages/permissions/PermissionIdentities.tsx
@@ -35,6 +35,7 @@ import DeleteIdentityBtn from "./actions/DeleteIdentityBtn";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
 import { isUnrestricted } from "util/helpers";
 import IdentityResource from "components/IdentityResource";
+import CreateTlsIdentityBtn from "./CreateTlsIdentityBtn";
 
 const PermissionIdentities: FC = () => {
   const notify = useNotify();
@@ -269,6 +270,9 @@ const PermissionIdentities: FC = () => {
                 />
               )}
             </PageHeader.Left>
+            <PageHeader.BaseActions>
+              <CreateTlsIdentityBtn />
+            </PageHeader.BaseActions>
           </PageHeader>
         }
       >
--- a/src/sass/_instance_detail_page.scss
+++ b/src/sass/_instance_detail_page.scss
@@ -7,13 +7,3 @@
     margin: 0 $sph--large;
   }
 }
-
-.create-instance-from-snapshot-modal,
-.duplicate-instances-modal,
-.create-image-from-instance-modal {
-  .p-modal__dialog {
-    @include large {
-      width: 35rem;
-    }
-  }
-}
--- a/src/types/permissions.d.ts
+++ b/src/types/permissions.d.ts
@@ -1,8 +1,13 @@
 export interface LxdIdentity {
   id: string; // fingerprint for tls and email for oidc
-  type: string;
+  type:
+    | "Client certificate"
+    | "Client certificate (pending)"
+    | "Client certificate (unrestricted)"
+    | "OIDC client";
   name: string;
   authentication_method: "tls" | "oidc";
+  tls_certificate: string;
   groups?: string[] | null;
   effective_groups?: string[];
   effective_permissions?: LxdPermission[];
@@ -29,4 +34,13 @@ export interface LxdPermission {
 export interface IdpGroup {
   name: string;
   groups: string[]; // these should be names of lxd groups
+}
+
+export interface TlsIdentityTokenDetail {
+  client_name: string;
+  addresses: string[];
+  expires_at: string;
+  fingerprint: string;
+  type: "Client certificate (pending)" | "Client certificate";
+  secret: string;
 }
--- a/src/util/helpers.tsx
+++ b/src/util/helpers.tsx
@@ -340,3 +340,14 @@ export const getDefaultStoragePool = (profile: LxdProfile) => {
 export const isUnrestricted = (identity: LxdIdentity) => {
   return identity.type === "Client certificate (unrestricted)";
 };
+
+export const isFineGrainedTls = (identity: LxdIdentity) => {
+  return ["Client certificate (pending)", "Client certificate"].includes(
+    identity.type,
+  );
+};
+
+export const base64EncodeObject = (data: object) => {
+  const jsonString = JSON.stringify(data);
+  return btoa(jsonString);
+};
--- a/src/pages/permissions/CreateTlsIdentityBtn.tsx
+++ b/src/pages/permissions/CreateTlsIdentityBtn.tsx
@@ -1,0 +1,31 @@
+import { Button, Icon } from "@canonical/react-components";
+import { useSmallScreen } from "context/useSmallScreen";
+import { FC } from "react";
+import usePortal from "react-useportal";
+import CreateIdentityModal from "./CreateIdentityModal";
+
+const CreateTlsIdentityBtn: FC = () => {
+  const isSmallScreen = useSmallScreen();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <CreateIdentityModal onClose={closePortal} />
+        </Portal>
+      )}
+      <Button
+        appearance="positive"
+        className="u-float-right u-no-margin--bottom"
+        onClick={openPortal}
+        hasIcon={!isSmallScreen}
+      >
+        {!isSmallScreen && <Icon name="plus" light />}
+        <span>Create TLS Identity</span>
+      </Button>
+    </>
+  );
+};
+
+export default CreateTlsIdentityBtn;
--- a/src/sass/_permission_confirm_modal.scss
+++ b/src/sass/_permission_confirm_modal.scss
@@ -48,3 +48,12 @@
     margin-top: $spv--large;
   }
 }
+
+.token-code-block {
+  background-color: rgb(0 0 0 / 5%);
+  margin-bottom: 1.5rem;
+
+  code {
+    background-color: transparent;
+  }
+}
--- a/src/sass/styles.scss
+++ b/src/sass/styles.scss
@@ -354,6 +354,17 @@ body {
   cursor: pointer;
 }
 
+.create-instance-from-snapshot-modal,
+.duplicate-instances-modal,
+.create-image-from-instance-modal,
+.create-tls-identity {
+  .p-modal__dialog {
+    @include large {
+      width: 35rem;
+    }
+  }
+}
+
 .host-path-device-modal {
   .p-modal__dialog {
     @include large {


src/api/auth-identities.tsx
```diff
@@ -74,3 +74,20 @@ export const deleteOIDCIdentities = (
       .catch(reject);
   });
 };
+
+export const fetchIdentityDetails = (
```
mas-who, line 78: I think `createFineGrainedTlsIdentity` might be a more appropriate name for this function, wdyt?

src/types/permissions.d.ts
```diff
@@ -30,3 +30,12 @@ export interface IdpGroup {
   name: string;
   groups: string[]; // these should be names of lxd groups
 }
+
+export interface LxdTrustToken {
```
mas-who, line 34: this type should be named `LxdFineGrainedTlsIdentity`, the token is what we will generate in the browser.

src/types/permissions.d.ts
```diff
@@ -30,3 +30,12 @@ export interface IdpGroup {
...
+  client_name: string;
+  addresses: string[];
+  expires_at: string;
+  fingerprint: string;
+  type: string;
```
mas-who, line 39: we can be more specific with the `type` field since it has limited number of possible values

src/util/helpers.tsx
```diff
@@ -335,3 +336,19 @@ export const getDefaultStoragePool = (profile: LxdProfile) => {
...
+    secret: tokenPreResponse.secret,
+    expires_at: tokenPreResponse.expires_at,
+    type: tokenPreResponse.type,
+  };
+
+  const jsonString = JSON.stringify(tokenDetails);
+  const base64Encoded = btoa(jsonString);
+
+  return base64Encoded;
+};
```
mas-who, line 354:
```suggestion
export const encodeBase64Token = (data: unknown) => {
  const jsonString = JSON.stringify(data);
  return btoa(jsonString);
};
```

We shouldn't couple a general concept like base64 encoding to the tls identities data.

src/sass/_instance_detail_page.scss
```diff
@@ -10,7 +10,8 @@
 
 .create-instance-from-snapshot-modal,
 .duplicate-instances-modal,
-.create-image-from-instance-modal {
+.create-image-from-instance-modal,
+.create-tls-identity {
```
mas-who, line 14: this feels a bit off adding identity related styling here, since the file is supposed to be dedicated for the instance detail page. Looking at the styles, it's a general modal dialog style override, I'd suggest move the styling into `styles.scss` and potentially come up with a more general class name. wdyt?

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -50,6 +53,12 @@ const PermissionIdentities: FC = () => {
   const [searchParams] = useSearchParams();
   const [selectedIdentityIds, setSelectedIdentityIds] = useState<string[]>([]);
   const { hasAccessManagementTLS } = useSupportedFeatures();
+  const isSmallScreen = useSmallScreen();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+
+  const handleClose = () => {
```
mas-who, line 59: no need to have this function here, we can just pass `closePortal` to `onClose` below

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+      fetchIdentityDetails(values.clientName)
+        .then((tokenPreResponse) => {
+          const tempToken = base64EncodeTokenDetails(tokenPreResponse);
+
+          setIsLoadingToken(false);
+          setToken(tempToken);
+        })
+        .catch((err) => {
+          setIsLoadingToken(false);
+          console.error("Error fetching token:", err);
```
mas-who, line 37: please remove the console log and use toast notification for error and success messages

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  return (
+    <Modal
+      close={onClose}
+      className="create-tls-identity"
+      title="Generate Trust Token"
+      buttonRow={
+        <>
+          {token ? (
+            <ActionButton
```
mas-who, line 50: We don't need a refresh button to update identities list, instead we should use react query invalidation

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+            </ActionButton>
+          ) : (
+            <ActionButton
+              appearance="positive"
+              className="u-no-margin--bottom"
+              onClick={() => void formik.submitForm()}
+              disabled={
+                formik.values.clientName.length === 0 ||
+                isLoadingToken ||
+                token != null
```
mas-who, line 65: Instead of two buttons, we can have one and adjust the onclick logic as well as the button label based on if a token has been generated

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+              }
+            >
+              Generate Token
+            </ActionButton>
+          )}
+        </>
+      }
+    >
+      {isLoadingToken && <Loader text="Loading token..." />}
+      {!isLoadingToken && !token && (
```
mas-who, line 75: We should disble the input instead of hiding it when making the API call, this could reduce layout shift in the modal. The loading state can be indicated by the action button instead of the dedicated loader

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <Input
+          id="clientName"
+          type="text"
+          label="Client Name"
+          onBlur={formik.handleBlur}
+          onChange={formik.handleChange}
+          value={formik.values.clientName}
+        />
+      )}
+      {!isLoadingToken && token && (
```
mas-who, line 85: as long as the token exist, we won't be in the loading state

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <div>
+          <p>
+            Please copy and paste your token for future reference as it will not
+            be available again.
+          </p>
+          <code>{token}</code>
+        </div>
```
mas-who, line 92: I think the div is not necessary here, we can just use a fragment

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+import Loader from "components/Loader";
+import { fetchIdentityDetails } from "api/auth-identities";
+import { base64EncodeTokenDetails } from "util/helpers";
+
+interface Props {
+  onClose: () => void;
+}
+
+const CreateIdentityModal: FC<Props> = ({ onClose }) => {
+  const [isLoadingToken, setIsLoadingToken] = useState(false);
```
mas-who, line 13: maybe this state can simply be `isLoading` as it is the only loading state for the modal

mas-who (commented):
Made some comments, I agree with @edlerd that this PR can actually be merged after some improvements. Although I'm also happy to not merge it and keep the PR as a reference for future, depending on if you have time to take it to completion

src/sass/_instance_detail_page.scss
Kxiru: Oh right, I see what you mean.
Yes, it is a general style. I have moved it into styles.scss for now, however do you think forms.scss would be appropriate? Not too sure, because although modals include forms, it is not directly relevant to forms themselves. Perhaps styles.scss is best as suggested.

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: Will we need a success message if the token is being presented right in front of us and we are already in a modal? I note this because if one exits the modal after generating a token by pressing refresh, we wouldn't see the toast notification anyway. Also a token would only be generated if the operation was successful.

mas-who: that's a interesting one, I think perhaps have some copy in the modal to indicate that an TLS identity was created. I'm not entirely confident that is the correct design though, probably a good topic to discuss during standup. The toast notification would be somewhat hidden when the modal is open anyway...

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: I think this gets addressed by invalidating the queries :)

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: Hmm. Won't this approach also result in layout shift because at one point the tokens are there and at the next, they aren't?
In my recent commit, I took out the loader state entirely, but I have still left the input box and token box there and moved the loading feedback to the action button. Let me know what you think?

mas-who: the layout shift would be during the loading state for the modal. If we don't show the loader and just disable the input, then during the loading state, the modal will not have its size adjusted

mas-who: so what you've done now seems good :+1:

src/types/permissions.d.ts
```diff
@@ -30,3 +30,12 @@ export interface IdpGroup {
   name: string;
   groups: string[]; // these should be names of lxd groups
 }
+
+export interface LxdFineGrainedTlsIdentity {
+  client_name: string;
+  addresses: string[];
+  expires_at: string;
+  fingerprint: string;
+  type: "oidc" | "Client certificate (pending)" | "Client certificate";
```
mas-who, line 39: I think `type` can't be `oidc` in this case since this data gets returned from the `.../tls` endpoint. On second thought, maybe we can change the type name to `LxdTlsIdentity` instead, it's less verbose

src/pages/permissions/CreateIdentityModal.tsx
mas-who, line 1: Code in this file looks pretty good now, thanks for the changes. I believe things will still change here based on our discussion earlier

Kxiru: Sounds good, thank you!

src/types/permissions.d.ts
Kxiru: Surely the type name should remain LxdFineGrainedTlsIdentity as "LxdTlsIdentity is too broad. It does not consider Unrestricted TLS identities.

mas-who: I've done some digging look at the API [docs](https://documentation.ubuntu.com/lxd/en/latest/api/#/), please refer to it for the following observation:

1. The naming for the type is wrong, since `LxdFineGrainedTlsIdentity` is a very specific response data struct only returned by making a request to `POST /1.0/auth/identities/tls`. The response does not represent an "official" identity data structure in LXD, and it includes details like `addresses`, `expires_at`, `fingerprint` and `secret` etc because those are required to authenticate a TLS identity later. Instead, we should think of a specific type name for the response data from this endpoint, perhaps what you had earlier was a closer fit e.g. `TlsIdentityTokenDetail`.

2. The official identity data structure from the LXD backend can be retrieved from `GET /1.0/auth/identities?recursion=1`, which is already represented by the existing `LxdIdentity` type. Can you please augment the existing type by looking at the response data from the documentation, add what is missing and also make the `type` field more strict (you will need to look at the network request for this since the documentation isn't all comprehensive).

3. After looking at the documentation for the `POST /1.0/auth/identities/tls` endpoint, I see we can set a `certificate` field in the request payload, instead of `token`. Please would you investigate what happens if we set `certificate` to `true`. I suspect we can potentially skip the create certificates in the UI step with this (for future reference).

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+          )}
+          <ActionButton
+            appearance="positive"
+            className="u-no-margin--bottom"
+            onClick={token ? onClose : () => void formik.submitForm()}
+            disabled={formik.values.clientName.length === 0 || isLoading}
+            loading={isLoading}
+          >
+            {token ? `Close` : `Generate Token`}
+          </ActionButton>
```
edlerd, line 80: I think the close should not have a positive appearance, and it doesn't need the disabling or loading props.

So there is no overlap in the props, it seems better to have two distinct buttons.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+          {token && (
+            <Button
+              className="u-no-margin--bottom"
+              onClick={() => {
+                navigator.clipboard.writeText(token);
+              }}
+              type="button"
+              hasIcon
+            >
+              <Icon name="copy" />
```
edlerd, line 69: This should have a copy label as well, to be discoverable to screen readers.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+          type="text"
+          label="Client Name"
+          onBlur={formik.handleBlur}
+          onChange={formik.handleChange}
+          value={formik.values.clientName}
+        />
+      )}
+      {token && (
+        <>
+          <p>Identity has been created.</p>
```
edlerd, line 102: Can this be the heading of the modal? Idea is replacing the "Generate trust token" copy once we are at this step.

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -264,6 +274,17 @@ const PermissionIdentities: FC = () => {
...
+              <Button
+                appearance="positive"
+                className="u-float-right u-no-margin--bottom"
+                onClick={openPortal}
+                hasIcon={!isSmallScreen}
+              >
+                {!isSmallScreen && <Icon name="plus" light />}
+                <span>Create TLS Identity</span>
+              </Button>
+            </PageHeader.BaseActions>
```
edlerd, line 287: All of this logic, with the button, mounting the modal and referencing the new CreateIdentityModal should move to a CreateTLSIdentityButton component, so we have contained the logic in a single place. This PermissionIdentities component responsibility should be the listing, not the action details.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <p>The token below can be used to log in as the new identity.</p>
+
+          <Notification
+            severity="caution"
+            title="Make sure to copy the token now as it will not be shown again."
+          ></Notification>
+
+          <div className="token-code-block">
+            <code>{token}</code>
+          </div>
```
edlerd, line 112: We should reuse the [CodeSnippet](https://canonical.github.io/react-components/?path=/story/components-codesnippet--wrap-lines) and avoid a custom styling for it. The custom styling creates more code and is missing margins.

src/util/helpers.tsx
```diff
@@ -335,3 +336,8 @@ export const getDefaultStoragePool = (profile: LxdProfile) => {
     });
   return rootStorage ? rootStorage.pool : "";
 };
+
+export const base64EncodeTokenDetails = (data: LxdFineGrainedTlsIdentity) => {
```
edlerd, line 340: I think we should make this a bit more generic, i.e. call it `base64EncodeObject(data: Object)` so it can be reused in other contexts.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [error, setError] = useState(false);
+  const toastNotify = useToastNotification();
+  const queryClient = useQueryClient();
+
+  const formik = useFormik({
+    initialValues: {
+      clientName: "",
+    },
+    onSubmit: (values) => {
+      setIsLoading(true);
```
edlerd, line 33: You can query `formik.isSubmitting`, no need for a custom state. The formik one is automatically toggled during submit.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+      createFineGrainedTlsIdentity(values.clientName)
+        .then((tokenPreResponse) => {
```
edlerd, line 36:
```suggestion
        .then((response) => {
```
What is a token pre response?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+          const tempToken = base64EncodeTokenDetails(tokenPreResponse);
```
edlerd, line 37:
```suggestion
          const encodedToken = base64EncodeTokenDetails(tokenPreResponse);
```

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+          >
+            {token ? `Close` : `Generate Token`}
+          </ActionButton>
+        </>
+      }
+    >
+      {error && (
+        <Notification
+          severity="negative"
+          title="Fine grained identity token creation failed"
```
edlerd, line 87:
```suggestion
          title="Token creation failed"
```

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+      )}
+      {!token && (
+        <Input
+          id="clientName"
+          type="text"
+          label="Client Name"
+          onBlur={formik.handleBlur}
+          onChange={formik.handleChange}
+          value={formik.values.clientName}
+        />
```
edlerd, line 98: We are creating an identity, this should be called "Identity name", where does the "client name" idea come from?

edlerd (commented):
This already works well. Some ideas below to streamline code and appearance.

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: Hmm...
Merging the close and submit funcitonality into one button was an earlier suggestion. Given you are not a fan of the "positive" appearance, do you recommend I use a normal button, or conditionally make the button appearance negative / null?

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: CodeSnippet does not currently support strings without breaks. I have raised this in the Vanilla channel. There are some workarounds however these also require custom styling. Please see (this thread)[https://chat.canonical.com/canonical/pl/44y5kzsagfnsmfyja8hi5wwt6e] and let me know your thoughts.

src/pages/permissions/CreateIdentityModal.tsx
edlerd: I think the default appearance, same as the copy button, would be fine for the close button.

src/pages/permissions/CreateIdentityModal.tsx
edlerd: I see, it doesn't wrap long strings. Then, let's add the default margin below the element to fix the spacing.

edlerd: image

Something like this, but not as hacky, better with a class and applying the styling via scss.

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: This is the terminology that the API and documentation uses. Yes, when the identity surfaces, it appears as the "Identity name", but I believe there is more to it than just the name and that staying true to the API may avoid potential issues in the future if the API changes against our mental model of "Identity name".

edlerd: Right good points. I see why the docs are doing it. They are tailoring towards a new command line client. I think in the UI, the target audience is a bit different. A user might create the token for a new client, but it could also be for another person (think an admin creating the token for a colleague). In the API layer we use, this is just "name". I agree, a "Name" label would be confusing here when creating a token. The entity that is created is referred to as "Identity" in the UI. So maybe "Identity name" is the better description to cover all target use cases. Wdyt?

Kxiru: Sure, I think for the sake of the end user we can change client name to identity name. I do think that "Client name" allows for the user to think about their nming decision a little more, whilst "name" just implies a casual identity. To me, it's like the difference between "James's token" and "token 1" or "Chrome browser" and "laptop", because "client" implies the user should think more about how, why and where they are using it, as opposed to just what to call it. Especially as there is no renaming function and poor naming can make for confusing identities. Thoughts?

edlerd: I think we should be consistent with the identity name as that is how the entity is presented after creation.

Kxiru: Alrighty! "Identity name" it is.

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -181,7 +182,9 @@ const PermissionIdentities: FC = () => {
                 <Icon name="user-group" />
               </Button>
               {hasAccessManagementTLS && (
-                <DeleteIdentityBtn identity={identity} />
+                <>
+                  <DeleteIdentityBtn identity={identity} />
+                </>
```
edlerd, line 187: Why this change? Can it be reverted to keep it simple?

src/util/permissionIdentities.tsx
```diff
@@ -5,6 +5,11 @@ export type ChangeSummary = Record<
   { added: Set<string>; removed: Set<string>; name: string }
 >;
 
+export type FineGrainedTLSFormValues = {
+  name: string;
+  publicity: boolean;
```
edlerd, line 10:
```suggestion
```
I think this is not a valid field and unused.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+              title="Copy token"
+              className="u-no-margin--bottom"
+              onClick={() => {
+                void navigator.clipboard.writeText(token);
+              }}
+              type="button"
+              hasIcon
+            >
+              <Icon name="copy" />
+            </Button>
```
edlerd, line 64: Can we add some feedback after this is used?
A [good example interaction is found in AMS](https://github.com/canonical/ams/blob/main/ui/templates/js/components/ui/CopyToClipboard.tsx#L10):

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+        .then((response) => {
+          const encodedToken = base64EncodeObject(response);
+          setToken(encodedToken);
+
+          void queryClient.invalidateQueries({
+            queryKey: [queryKeys.identities],
+          });
+        })
+        .catch(() => {
+          setError(true);
```
edlerd, line 40: This never gets unset to false. Maybe can do it on submit?

src/pages/permissions/CreateIdentityModal.tsx
edlerd: This is still not resolved.

Please do not resolve comments yourself. Better add comment with a reference to the commit that fixes the issue that was pointed out.

edlerd (changes requested):
A couple of issues remaining below.

src/util/permissionIdentities.tsx
Kxiru: I will remove the entire type as the rename functionality is not being introduced in this PR.

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: I have now added it on submit, but what about a handleClose function that sets it to false and calls onClose?

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: Hi David,
You asked for a label to be added for screen readers, and I added an aria-label. Was there anything else you wanted to add?
I resolve comments as I make code changes to keep track of my progress and minimise clutter on the screen which can be confusing.

Kxiru:
>This is still not resolved.
>Please do not resolve comments yourself. Better add comment with a reference to the commit that fixes the issue that was pointed out.


Hi David,
You asked for a label to be added for screen readers, and I added an aria-label. Was there anything else you wanted to add?
I resolve comments as I make code changes to keep track of my progress and minimise clutter on the screen which can be confusing.

src/pages/permissions/CreateIdentityModal.tsx
edlerd: I suspect on close/reopen the modal gets rendered with initial state. It should not keep state from being opened previously.

edlerd (approved):
QA and code LGTM, good stuff, thx for adding this feature :)

PR #1033, opened 2024-12-16
fix(permissions) link identity groups count to edit panel

edlerd:
## Done

- fix(permissions) link identity groups count to edit panel

## QA

1. Run the LXD-UI:
    - check permissions > identities list. The group count should be a shortcut to show and edit the linked groups for an identity.


Kxiru:
Screenshots for visibility:
image
image

edlerd:
> I understand the logic, but is it not redundant to have two buttons with the exact same functionlity, centimeters away from one another?

Yes, but I think it serves a purpose, because as a user you might read the number and wonder immediately what are more details for it. Having a link available unblocks your thinking.

Also, this is to create parity with the behaviour in the group list.

PR #1035, opened 2024-12-17
Overhaul network layout

edlerd:
## Done

- overhaul network layout
- add network topology map for network detail pages
- add network config key search
- combine network overview and configuration tabs into one

## QA

1. Run the LXD-UI:
    - explore network creation and editing


src/pages/networks/NetworkTopology.tsx
```diff
@@ -0,0 +1,136 @@
...
+          </div>
+        )}
+        <div
+          className={classNames("current-network", {
+            "has-descendents":
+              instances.length > 0 || downstreamNetworks.length > 0,
+            "has-parent": !!uplink,
+          })}
+        >
+          <div
```
mas-who, line 64: should we rather define a variant for the `ResourceLink` component instead of defining a new chip here?

src/pages/networks/NetworkTopology.tsx
```diff
@@ -0,0 +1,136 @@
...
+        <div className="downstream">
+          {downstreamNetworks
+            .slice(0, isCollapsed ? 5 : downstreamNetworks.length)
+            .map((item) => {
+              const networkUrl = `/ui/project/default/network/${item.name}`;
+              return (
+                <div
+                  key={networkUrl}
+                  className={classnames("downstream-item", {
+                    "has-descendents": instances.length > 0,
```
mas-who, line 83: do we want to show the lines extending from downstream networks even though we are not showing instance connected to them?

src/pages/networks/NetworkTopology.tsx
```diff
@@ -0,0 +1,136 @@
+import React, { FC, useState } from "react";
```
mas-who, line 1: very cool component, the simplicity is great! Related to the QA note, in the case where we have downstream networks and instances connected to the current uplink network, and some of the instances are connected to the downstream networks, do we want to show the instance connections to the downstream networks?

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -168,38 +170,123 @@ const NetworkForm: FC<Props> = ({
...
+    }
+    return filteredRows;
   };
-  useEffect(updateFormHeight, [notify.notification?.message, section]);
-  useEventListener("resize", updateFormHeight);
+
+  const isManagedNetwork = formik.values.bareNetwork?.managed ?? true;
+
+  const availableSections = [GENERAL];
+  if (formik.values.networkType !== "physical" && isManagedNetwork) {
```
mas-who, line 192: I think this could be a good place to leave a comment with link to lxd docs where we can find info relating to available configs for different network types. wdyt?

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -168,38 +170,123 @@ const NetworkForm: FC<Props> = ({
...
+    availableSections.push(DNS);
+  }
+  if (formik.values.networkType === "physical" && isManagedNetwork) {
+    availableSections.push(OVN);
+  }
+
+  useEffect(() => {
+    const wrapper = document.getElementById("content-details");
+
+    const activateSectionOnScroll = () => {
```
mas-who, line 211: should we add some logic here to check if the scroll position is at the bottom so that we can set the last available section? Currently, for a `physical` network, the DNS and OVN section have very few configs which results in the OVN section never getting highlighted (due to both sections being displayed together at the bottom scroll position).

src/pages/networks/forms/NetworkFormMain.tsx
```diff
@@ -35,122 +39,149 @@ const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
...
+      <div className="u-sv3">
+        <div className="general-field">
+          <div className="general-field-label">
+            <Label forId="networkType">Type</Label>
+          </div>
+          <div className="general-field-content">
+            {formik.values.isCreating ? (
+              <NetworkTypeSelector formik={formik} />
+            ) : (
+              capitalizeFirstLetter(formik.values.networkType)
```
mas-who, line 58: this would result in `Ovn` type label

mas-who (commented):
Love the network topology and how the existing form is reused. Still need to go through the css files though.

mas-who:
Please see some QA observations below:

1. Can we align the text for the network detail page general section? There is probably some margin that needs to be removed.
Screenshot from 2025-01-07 12-35-27

2. On smaller screens, on the create network page (edit page is fine), the config nav with the search bar seems to overlap with other page contents. Also maybe the nav can become collapsible on small screens?
Screenshot from 2025-01-07 12-55-03

3. When creating a network, if name is not entered we can switch to the yaml editor. Seems a little strange since we can set the network name in the yaml editor itself. Perhaps this is something to think about in the long run as it affects other creation forms too. Also we would have to figure out how to handle name field population in the yaml editor etc...
Screenshot from 2025-01-07 13-28-43

4. Should we show the downstream instances connected to the ovn networks in the following scenario? Probably a design question for @piperdeck . If not, we should probably get rid of the unconnected lines at the end of the topology.
Note that in this scenario, I have instance `i1` connected to the uplink as well as several ovn networks, the topology presented makes me think that `i1` is only connected to the uplink network. wdyt?
Screenshot from 2025-01-07 13-46-53

src/pages/networks/NetworkTopology.tsx
edlerd: Good idea. Changing this to indicate "there is more below" with a line and "..." without exposing it directly.

src/pages/networks/NetworkTopology.tsx
edlerd: I think that would be too confusing, see https://github.com/canonical/lxd-ui/pull/1035#discussion_r1905497867

src/pages/networks/forms/NetworkForm.tsx
edlerd: good idea

src/pages/networks/forms/NetworkForm.tsx
edlerd: That won't work, because the scroll position will be ambiguous, it the won't be clear which sections is active when scrolled down. But we can increase the bottom padding below the form, so the last section can get active.

src/pages/networks/forms/NetworkFormMain.tsx
edlerd: resolved with a helper function.

piperdeck:
@edlerd 

I've left some comments and feedback in a Figma page: https://www.figma.com/design/obSXJ4VkJUa2fHjt5lNf68/25.04-OVN-Networking?node-id=242-9623

src/pages/networks/forms/NetworkForm.tsx
mas-who: That's perfect, it works nicely now! :+1:

src/pages/networks/EditNetwork.tsx
```diff
@@ -113,32 +108,52 @@ const EditNetwork: FC<Props> = ({ network, project }) => {
...
         version={version}
       />
       <FormFooterLayout>
         <YamlSwitch
           formik={formik}
           section={section}
-          setSection={setSection}
+          setSection={() =>
+            setSection(
+              section === slugify(YAML_CONFIGURATION)
```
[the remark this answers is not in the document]
mas-who: Yes, agreed on keeping it simple. Let's keep the current behaviour :+1:

src/sass/_ip_address.scss
```diff
@@ -0,0 +1,9 @@
+.ip-address {
+  .configuration-help {
+    display: none;
+  }
+
+  .p-button--base[title="Clear override"] {
```
mas-who, line 6: how do we clear an override for the network ip address configs?

src/sass/_forms.scss
```diff
@@ -150,13 +150,8 @@
     align-items: flex-start;
 
     > :nth-child(2) {
-      margin-left: -1.5rem;
       margin-top: -5px;
```
mas-who, line 153: the `margin-top: -5px;` is causing the input and the custom raido button to not center align. Should we remove this or is there another reason this styling is here?

Screenshot from 2025-01-08 11-31-17

Not related to the CSS here, should we have all the radio button labels capitalised or converted to lowercase?

src/sass/_network_topology.scss
```diff
@@ -0,0 +1,73 @@
+.network-topology {
```
mas-who, line 1: very cool!

mas-who (commented):
It's working very nicely, just a few more observations.

src/sass/_network_topology.scss
lyubomir-popov: sorry to butt in, noticed a couple of things:
- we shouldn't have any "px" values anywhere in the css (other than defining single pixel lline thickness, if needed). This is because we have a setting that "zooms in" the entire content by changing the root em value for bigger screens. Anything defined in px will be unaware of that change. 
- our spacing strategy is based on multiples of .5rem. That means 8px before the zoom described above. So multiples of 5 will understandably lead to misalignments.

mas-who: not at all @lyubomir-popov , thanks for the input, this is good to keep in mind for future reviews! :+1:

src/pages/networks/NetworkTopology.tsx
```diff
@@ -0,0 +1,136 @@
...
+              );
+            })}
+          {downstreamNetworks.length > 5 && isCollapsed && (
+            <div className="downstream-item">
+              <Button
+                appearance="link"
+                onClick={() => setCollapsed(false)}
+                small
+              >
+                Show all
```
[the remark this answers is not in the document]
edlerd: Yes, agree that 2 is the better option, adjusted it accordignly.

src/sass/_ip_address.scss
edlerd: That is not possible, there is no network without an ip address. The default on create is "auto", which we should actually show, I'll change that.
On edit all you can do is set it to none, manually specify or set it to auto to get a new address. But you can't disable the override.

src/sass/_forms.scss
edlerd: Good idea with the label adjustments.

And yes, the margin-top is now removed. It is not needed any more and actually misaligns things now.

src/sass/_ip_address.scss
mas-who: okay got it, that makes sense :+1:

src/sass/_network_topology.scss
edlerd: Good point, adjusted our usages for this feature to rely on rem rather than pixel values.

mas-who:
LGTM from my end :+1:

edlerd:
@piperdeck gave design +1 yesterday. @mas-who if you have some time on Monday, please give another review of the updates.

src/sass/styles.scss
```diff
@@ -372,3 +373,10 @@ body {
     }
   }
 }
+
+.network-description {
```
mas-who, line 377: maybe this class can be `.table-description`? Seems like it could be used by other tables in the app.

mas-who (approved):
LGTM with one small non-blocking nit :+1:

src/sass/styles.scss
edlerd: good idea, that will also remove a bit of duplication as we can reuse this.

mas-who (approved):

PR #1049, opened 2025-01-09
chore(deps) update demo server dependencies

edlerd:
## Done

- update dependencies for demo server

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -54,7 +54,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
 
       - name: Install Playwright Browser
         run: npx playwright install --with-deps chromium
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -111,7 +111,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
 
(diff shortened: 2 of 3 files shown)


mas-who:
@edlerd I think we need to move the node setup step before the install dependency step in CI to resolve the failing tests. I'm not sure why this didn't surface in the CI for the update dependency PR though :thinking:

edlerd:
Node 23 is a bad choice as it is only supported until mid 2025: https://nodejs.org/en/about/previous-releases

mas-who:
> > @edlerd I think we need to move the node setup step before the install dependency step in CI to resolve the failing tests. I'm not sure why this didn't surface in the CI for the update dependency PR though 🤔
> 
> I think github updated something in between, so the node version changed.
> 
> Good hint with moving the setup node higher, I also updated the node version to 22 everywhere, as that is the LTS one.

Looking at the CI output now, I think something is setting up node either that the setup-node action. Maybe node even comes with ubuntu-latest?

.github/workflows/coverage.yaml
```diff
@@ -18,6 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-node@v4
```
mas-who, line 21: we can probably revert this change, since the node setup here is actually for installing playwright. Maybe add a comment for context?

mas-who (approved):
LGTM :+1:

PR #1058, opened 2025-01-15
fix: remove spaces from custom iso name

### LINKED ISSUE #1057: Error when trying to upload custom ISO
Image

Image

Ubuntu Server 24.10 all updates installed with LXD installed during setup ( Version 5.21.2-ui-0.8.4 )

mas-who:
## Done

- Remove spaces from uploaded custom iso file name
- Prevent custom iso alias with spaces

Fixes #1057 

## QA

1. Run the LXD-UI:
    - Upload a custom iso with spaces in its file name, make sure the auto populated alias does not contain spaces.
    - Try change the alias so that it contains spaces

### DIFF
--- a/src/pages/storage/UploadCustomIso.tsx
+++ b/src/pages/storage/UploadCustomIso.tsx
@@ -20,6 +20,8 @@ import { useToastNotification } from "context/toastNotificationProvider";
 import StoragePoolSelector from "./StoragePoolSelector";
 import { AxiosError } from "axios";
 import type { LxdSyncResponse } from "types/apiResponse";
+import { isValidISOAlias, sanitizeISOAlias } from "util/customISO";
+import classnames from "classnames";
 
 interface Props {
   onFinish: (name: string, pool: string) => void;
@@ -98,14 +100,18 @@ const UploadCustomIso: FC<Props> = ({ onCancel, onFinish }) => {
     if (e.target.files) {
       const file = e.target.files[0];
       setFile(file);
-      setName(file.name);
+      setName(sanitizeISOAlias(file.name));
     }
   };
 
   return (
     <>
       {notify.notification ? <NotificationRow /> : <UploadCustomImageHint />}
-      <div className={uploadState === null ? "" : "u-hide"}>
+      <div
+        className={classnames("custom-iso-form", {
+          "u-hide": !!uploadState,
+        })}
+      >
         <Input
           name="iso"
           type="file"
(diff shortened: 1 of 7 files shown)


src/pages/storage/UploadCustomIso.tsx
```diff
@@ -98,10 +98,12 @@ const UploadCustomIso: FC<Props> = ({ onCancel, onFinish }) => {
     if (e.target.files) {
       const file = e.target.files[0];
       setFile(file);
-      setName(file.name);
+      setName(file.name.replaceAll(" ", "-"));
```
edlerd, line 101: I think we should be more strict and only allow alphanumeric chars and -. Everything else should be replaced with "-". Similar to what `sanitizeInstanceName` is doing.

edlerd (approved):
Thanks for fixing this so quickly. One idea below, we might want to be more strict with the allowed characters, as filenames could come with random Unicode chars and I suspect they might cause issues as well.

src/pages/storage/UploadCustomIso.tsx
mas-who: I thought about using `sanitizeInstanceName` but it is too strict for custom iso alias name. I am not sure if there are any other restrictions either than spaces. Should we just apply the same rules as `sanitizeInstanceName`?

edlerd: so I think we can't reuse it, but a very similar check should do.

mas-who: done, please have a look

edlerd (approved):
LGTM, thanks for fixing this bug :+1:

edlerd:
Something odd came up in QA: When changing the alias to contain a space, the modal with jumps around.

mas-who:
> Something odd came up in QA: When changing the alias to contain a space, the modal width jumps around.
> 
>  Screencast.from.2025-01-15.17-55-12.mp4

Right, the modal is constrained by a 80rem max-width which leads to the horizontal shift when the error message appears. I've applied a max width to the error message itself which resolves the issue. The current behaviour is that the error message will wrap. We could also do truncation but I think wrapping is better as the error message becomes clearly visible wdyt?

src/pages/storage/UploadCustomIso.tsx
```diff
@@ -122,6 +128,14 @@ const UploadCustomIso: FC<Props> = ({ onCancel, onFinish }) => {
           value={name}
           onChange={(e) => setName(e.target.value)}
           disabled={file === null}
+          error={
+            name && !isValidISOAlias(name) ? (
+              <div className="alias-error">
+                Please enter only alphanumeric characters, periods (.) and
+                hyphens in this field
```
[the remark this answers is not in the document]
edlerd: Oh yes, please :)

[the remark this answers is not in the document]
mas-who: Done :+1:

edlerd (dismissed):
Thx for the adjustments!

edlerd:
The other two modals (instance and volume snapshot creation) have the same problem, that the width changes if the error gets visible.

mas-who:
> The other two modals (instance and volume snapshot creation) have the same problem, that the width changes if the error gets visible.

Those should be fixed now.

edlerd (approved):
All good now, thanks for making this great :)

PR #1060, opened 2025-01-17
chore: make vite dev server port configurable

mas-who:
## Done

- Make vite dev server port configurable. It is set to 3000 by default so we don't need to worry about creating an extra secret for the demo environment.
- Added to the docs about configuring `VITE_PORT`

## QA

1. Run the LXD-UI:
    - Ensure CI passes
    - Check starting the local dev environment and the UI loads
    - Check the demo server and make sure UI loads


package.json
```diff
@@ -18,7 +18,7 @@
     "lint-js": "tsc --noEmit && yarn lint-js-eslint && yarn lint-js-prettier",
     "hooks-add": "husky install",
     "hooks-remove": "husky uninstall",
-    "start": "concurrently --kill-others --raw 'vite --host | grep -v 3000' 'yarn serve'",
+    "start": "concurrently --kill-others --raw 'vite --host | grep -v http:' 'yarn serve'",
```
lorumic, line 21: Can we remove the `grep` from this command? I don't see why we would need it, and actually that was masking the issue of the "port already in use" for me today. For example, I just tried removing it from my system (where port 3000 is in use) and I got this output:
```
$ concurrently --kill-others --raw 'vite --host' 'yarn serve'
$ ./entrypoint
booting on https://localhost:8407
Port 3000 is in use, trying another one...

  VITE v6.0.7  ready in 107 ms

  ➜  Local:   http://localhost:3001/
...
```
If I had seen the filtered out line, that would have probably saved us one hour trying to figure out what was the issue today. :)
```suggestion
    "start": "concurrently --kill-others --raw 'vite --host' 'yarn serve'",
```

lorumic (dismissed):
Looks good, just a couple of minor ideas :)

package.json
lorumic: Actually, forget about this. Your edit makes the "Port already in use" line appear (which is desirable) and just filters out the misleading Vite URLs. So it's fine to keep it this way. Marking as resolved.

lorumic (approved):
Thanks for the change - LGTM :+1:

CONTRIBUTING.md
```diff
@@ -82,6 +82,9 @@
     You should enable a firewall as `dotrun` exposes an api to start or interact with unprivileged containers on your public
     ip via port `8407`. Ensure that the lxd API on port `8443` is open, so `dotrun` can access it.
 
+    For local development, the UI is served by the Vite development server, which uses port 3000 by default. 
+    If you have another process listening on this port on your system, you should set a different port by adding a custom `VITE_PORT` variable in the `.env.local` file.
```
edlerd, line 86:
```suggestion
    If you have another process listening on this port on your system, create `.env.local` as a copy of the `.env` file and change the `VITE_PORT` variable in it to any free port.
```

vite.config.ts
```diff
@@ -1,6 +1,15 @@
...
 import tsconfigPaths from "vite-tsconfig-paths";
+import * as dotenv from "dotenv";
+import * as fs from "fs";
+
+// Load .env.local if it exists, otherwise load .env
+if (fs.existsSync(".env.local")) {
+  dotenv.config({ path: ".env.local" });
+} else {
+  dotenv.config({ path: ".env" });
+}
```
edlerd, line 12: This is not exclusive, we should load the .env and then load .env.local if it exists to override some or all values

edlerd (changes requested):

CONTRIBUTING.md
edlerd: I think we need to guide how to create the .env.local file, as it won't be there for a fresh checkout.

edlerd:
Thanks for bringing this up so quickly, some suggestions from review above.

CONTRIBUTING.md
mas-who: yes, sounds good!

vite.config.ts
mas-who: Thanks that's a good catch, it should be resolved with the latest push. Tested by setting a different vite port and backend IP in `.env.local`

edlerd (commented):
Good work, thanks for adjusting it :)

PR #1064, opened 2025-01-21
Ensure vm is finished booting before connecting to it via terminal

### LINKED ISSUE #1063: Starting a VM on the terminal page will immediately show you an error
When I'm on the terminal page of a virtual machine that is turned off and I hit the big green button "Start instance":

Image

immediately results in:

Image

it should wait for the VM to start and the LXD agent to become available, just like the ui does with LXCs :)

edlerd:
## Done

- Ensure vm is finished booting before connecting to it via terminal

Fixes #1063

## QA

1. Run the LXD-UI:
    - start / stop vms and from the instance terminal page.

### DIFF
--- a/src/pages/instances/InstanceTerminal.tsx
+++ b/src/pages/instances/InstanceTerminal.tsx
@@ -44,9 +44,10 @@ export const defaultPayload: TerminalConnectPayload = {
 
 interface Props {
   instance: LxdInstance;
+  refreshInstance: () => Promise<unknown>;
 }
 
-const InstanceTerminal: FC<Props> = ({ instance }) => {
+const InstanceTerminal: FC<Props> = ({ instance, refreshInstance }) => {
   const { name, project } = useParams<{
     name: string;
     project: string;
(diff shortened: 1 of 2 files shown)


Kxiru (approved):
Looks good to me! :) Might wait for another perspective, also!

src/pages/instances/InstanceTerminal.tsx
```diff
@@ -143,16 +144,34 @@ const InstanceTerminal: FC<Props> = ({ instance }) => {
...
+  useEffect(() => {
+    if (isBooting && refreshTimerRef.current === null) {
+      const delay = 1000;
+      const triggerRefresh = () => {
+        void refreshInstance();
+        refreshTimerRef.current = null;
+        setVersion((old) => old + 1);
+      };
+      refreshTimerRef.current = setTimeout(triggerRefresh, delay);
+    }
```
mas-who, line 160: I think we should clear the timeout if it exists when the component unmounts.

mas-who (commented):
I think we should change the empty state text for the state where the instance is booting. Maybe "Instance starting" for the heading and "Terminal will be ready once the instance booting is complete." for the message?

Screenshot from 2025-01-21 16-39-45

mas-who (approved):
LGTM :+1: thanks for the fix!

PR #1070, opened 2025-01-22
feat: [WD-18264] CMS fields for storage pool source

Kxiru:
## Done
- Added intermediary StoragePoolClusteredSourceSelector file + a ClusterSpecificInput component.
- Added functionality to create storage pool source fields for specific cluster members.

## QA

1. Run the LXD-UI:
    - Within a clustered environment, navigate to the Create a storage pool page, and attempt to create a storage pool with a ZFS, LVM or Directory driver. (Directory is less prone to error and empty directories can be made for testing purposes by creating them in the terminal of the cluster node).
    - With a storage pool successfully created, navigate to it's configuration page and verify that
      - The page defaults to the correct setting (global source or single clusters sources)
      - The previously input sources persist.

## Screenshots
image
image

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -1,10 +1,12 @@
 import {
+  constructMemberError,
   handleEtagResponse,
   handleResponse,
   handleSettledResult,
 } from "util/helpers";
 import {
   LxdStoragePool,
+  LXDStoragePoolOnClusterMember,
   LxdStoragePoolResources,
   LxdStorageVolume,
   LxdStorageVolumeState,
@@ -14,6 +16,7 @@ import type { LxdApiResponse } from "types/apiResponse";
 import type { LxdOperationResponse } from "types/operation";
 import axios, { AxiosResponse } from "axios";
 import type { LxdClusterMember } from "types/cluster";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 
 export const fetchStoragePool = (
   pool: string,
@@ -66,7 +69,7 @@ export const createPool = (
   });
 };
 
-const getClusterAndMemberPools = (pool: Partial<LxdStoragePool>) => {
+const getClusterAndMemberPoolPayload = (pool: Partial<LxdStoragePool>) => {
   const memberSpecificConfigKeys = new Set([
     "source",
     "size",
@@ -85,27 +88,38 @@ const getClusterAndMemberPools = (pool: Partial<LxdStoragePool>) => {
     }
   }
 
-  const clusterPool = { ...pool, config: clusterConfig };
-  const memberPool = { ...pool, config: memberConfig };
+  const clusterPoolPayload = { ...pool, config: clusterConfig };
+  const memberPoolPayload = { ...pool, config: memberConfig };
 
   return {
-    clusterPool,
-    memberPool,
+    clusterPoolPayload,
+    memberPoolPayload,
   };
 };
 
 export const createClusteredPool = (
   pool: LxdStoragePool,
   clusterMembers: LxdClusterMember[],
+  sourcePerClusterMember?: ClusterSpecificValues,
 ): Promise<void> => {
-  const { memberPool, clusterPool } = getClusterAndMemberPools(pool);
+  const { memberPoolPayload, clusterPoolPayload } =
+    getClusterAndMemberPoolPayload(pool);
   return new Promise((resolve, reject) => {
     Promise.allSettled(
-      clusterMembers.map((item) => createPool(memberPool, item.server_name)),
+      clusterMembers.map((item) => {
+        const clusteredMemberPool = {
+          ...memberPoolPayload,
+          config: {
+            ...memberPoolPayload.config,
+            source: sourcePerClusterMember?.[item.server_name],
+          },
+        };
+        return createPool(clusteredMemberPool, item.server_name);
+      }),
     )
       .then(handleSettledResult)
       .then(() => {
-        return createPool(clusterPool);
+        return createPool(clusterPoolPayload);
       })
       .then(resolve)
       .catch(reject);
@@ -132,15 +146,16 @@ export const updateClusteredPool = (
   pool: Partial<LxdStoragePool>,
   clusterMembers: LxdClusterMember[],
 ): Promise<void> => {
-  const { memberPool, clusterPool } = getClusterAndMemberPools(pool);
+  const { memberPoolPayload, clusterPoolPayload } =
+    getClusterAndMemberPoolPayload(pool);
   return new Promise((resolve, reject) => {
     Promise.allSettled(
       clusterMembers.map(async (item) =>
-        updatePool(memberPool, item.server_name),
+        updatePool(memberPoolPayload, item.server_name),
       ),
     )
       .then(handleSettledResult)
-      .then(() => updatePool(clusterPool))
+      .then(() => updatePool(clusterPoolPayload))
       .then(resolve)
       .catch(reject);
   });
@@ -170,6 +185,37 @@ export const deleteStoragePool = (pool: string): Promise<void> => {
     })
       .then(handleResponse)
       .then(resolve)
+      .catch(reject);
+  });
+};
+
+export const fetchPoolFromClusterMembers = (
+  poolName: string,
+  clusterMembers: LxdClusterMember[],
+): Promise<LXDStoragePoolOnClusterMember[]> => {
+  return new Promise((resolve, reject) => {
+    Promise.allSettled(
+      clusterMembers.map((member) => {
+        return fetchStoragePool(poolName, member.server_name);
+      }),
+    )
+      .then((results) => {
+        const poolOnMembers: LXDStoragePoolOnClusterMember[] = [];
+        for (let i = 0; i < clusterMembers.length; i++) {
+          const memberName = clusterMembers[i].server_name;
+          const result = results[i];
+          if (result.status === "rejected") {
+            reject(constructMemberError(result, memberName));
+          }
+          if (result.status === "fulfilled") {
+            const promise = results[
+              i
+            ] as PromiseFulfilledResult<LxdStoragePool>;
+            poolOnMembers.push({ ...promise.value, memberName: memberName });
+          }
+        }
+        resolve(poolOnMembers);
+      })
       .catch(reject);
   });
 };
--- a/src/components/forms/ClusterSpecificInput.tsx
+++ b/src/components/forms/ClusterSpecificInput.tsx
@@ -1,0 +1,141 @@
+import { FC, Fragment, useEffect, useState } from "react";
+import { CheckboxInput, Input } from "@canonical/react-components";
+import ResourceLink from "components/ResourceLink";
+import FormEditButton from "components/FormEditButton";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
+
+interface Props {
+  id: string;
+  isReadOnly: boolean;
+  onChange: (value: ClusterSpecificValues) => void;
+  toggleReadOnly: () => void;
+  memberNames: string[];
+  values?: ClusterSpecificValues;
+  canToggleSpecific?: boolean;
+  isDefaultSpecific?: boolean;
+  clusterMemberLinkTarget?: (member: string) => string;
+  disabled?: boolean;
+  helpText?: string;
+}
+
+const ClusterSpecificInput: FC<Props> = ({
+  values,
+  id,
+  isReadOnly,
+  memberNames,
+  onChange,
+  toggleReadOnly,
+  canToggleSpecific = true,
+  isDefaultSpecific = null,
+  clusterMemberLinkTarget = () => "/ui/cluster",
+  disabled = false,
+  helpText,
+}) => {
+  const [isSpecific, setIsSpecific] = useState<boolean | null>(
+    isDefaultSpecific,
+  );
+  const firstValue = Object.values(values ?? {})[0];
+
+  useEffect(() => {
+    const rawValues = Object.values(values ?? {});
+    if (isSpecific === null && rawValues.length > 0) {
+      const newDefaultSpecific = rawValues.some(
+        (item) => item !== rawValues[0],
+      );
+      setIsSpecific(newDefaultSpecific);
+    }
+  }, [isSpecific, values]);
+
+  const setValueForAllMembers = (value: string) => {
+    const update: ClusterSpecificValues = {};
+    memberNames.forEach((member) => (update[member] = value));
+    onChange(update);
+  };
+
+  const setValueForMember = (value: string, member: string) => {
+    const update = {
+      ...values,
+      [member]: value,
+    };
+    onChange(update);
+  };
+
+  return (
+    <div className="u-sv3">
+      {canToggleSpecific && !isReadOnly && (
+        <CheckboxInput
+          id={`${id}-same-for-all-toggle`}
+          label="Same for all cluster members"
+          checked={!isSpecific}
+          onChange={() => {
+            setIsSpecific((val) => !val);
+          }}
+        />
+      )}
+      {isSpecific && (
+        <div className="cluster-specific-input">
+          {memberNames.map((item) => {
+            const activeValue = values?.[item];
+
+            return (
+              <Fragment key={item}>
+                <div className="cluster-specific-member">
+                  <ResourceLink
+                    type="cluster-member"
+                    value={item}
+                    to={clusterMemberLinkTarget(item)}
+                  />
+                </div>
+                <div className="cluster-specific-value">
+                  {isReadOnly ? (
+                    <>
+                      {activeValue}
+                      <FormEditButton toggleReadOnly={toggleReadOnly} />
+                    </>
+                  ) : (
+                    <Input
+                      id={
+                        memberNames.indexOf(item) === 0 ? id : `${id}-${item}`
+                      }
+                      type="text"
+                      className="u-no-margin--bottom"
+                      value={activeValue}
+                      onChange={(e) => setValueForMember(e.target.value, item)}
+                      disabled={disabled}
+                    />
+                  )}
+                </div>
+              </Fragment>
+            );
+          })}
+          {helpText && (
+            <div className="p-form-help-text cluster-specific-helptext">
+              {helpText}
+            </div>
+          )}
+        </div>
+      )}
+      {!isSpecific && (
+        <div>
+          {isReadOnly ? (
+            <>
+              {firstValue}
+              <FormEditButton toggleReadOnly={toggleReadOnly} />
+            </>
+          ) : (
+            <Input
+              id={id}
+              type="text"
+              value={firstValue}
+              onChange={(e) => setValueForAllMembers(e.target.value)}
+              disabled={disabled}
+              help={helpText}
+            />
+          )}
+        </div>
+      )}
+    </div>
+  );
+};
+
+export default ClusterSpecificInput;
--- a/src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
+++ b/src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
@@ -1,0 +1,40 @@
+import { Label } from "@canonical/react-components";
+import ClusterSpecificInput from "components/forms/ClusterSpecificInput";
+import { FormikProps } from "formik";
+import { FC } from "react";
+import { StoragePoolFormValues } from "./StoragePoolForm";
+import { useClusterMembers } from "context/useClusterMembers";
+
+interface Props {
+  formik: FormikProps<StoragePoolFormValues>;
+  helpText?: string;
+}
+
+const StoragePoolClusteredSourceSelector: FC<Props> = ({
+  formik,
+  helpText,
+}) => {
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const memberNames = clusterMembers.map((member) => member.server_name);
+
+  return (
+    <>
+      <Label forId="sourcePerClusterMember">Source</Label>
+      <ClusterSpecificInput
+        values={formik.values.sourcePerClusterMember}
+        id="sourcePerClusterMember"
+        isReadOnly={false}
+        onChange={(value) => {
+          void formik.setFieldValue("sourcePerClusterMember", value);
+        }}
+        canToggleSpecific={formik.values.isCreating}
+        toggleReadOnly={() => {}}
+        memberNames={memberNames}
+        disabled={!formik.values.isCreating}
+        helpText={helpText}
+      />
+    </>
+  );
+};
+
+export default StoragePoolClusteredSourceSelector;
--- a/src/pages/storage/forms/StoragePoolForm.tsx
+++ b/src/pages/storage/forms/StoragePoolForm.tsx
@@ -41,16 +41,10 @@ import StoragePoolFormZFS from "./StoragePoolFormZFS";
 import { useSettings } from "context/useSettings";
 import { ensureEditMode } from "util/instanceEdit";
 import StoragePoolFormCephFS from "pages/storage/forms/StoragePoolFormCephFS";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 
 export interface StoragePoolFormValues {
-  isCreating: boolean;
-  readOnly: boolean;
-  name: string;
-  description: string;
-  driver: string;
-  source: string;
-  entityType: "storagePool";
-  size?: string;
+  barePool?: LxdStoragePool;
   ceph_cluster_name?: string;
   ceph_osd_pg_num?: string;
   ceph_rbd_clone_copy?: string;
@@ -62,6 +56,11 @@ export interface StoragePoolFormValues {
   cephfs_osd_pg_num?: string;
   cephfs_path?: string;
   cephfs_user_name?: string;
+  description: string;
+  driver: string;
+  entityType: "storagePool";
+  isCreating: boolean;
+  name: string;
   powerflex_clone_copy?: string;
   powerflex_domain?: string;
   powerflex_gateway?: string;
@@ -75,11 +74,14 @@ export interface StoragePoolFormValues {
   pure_gateway?: string;
   pure_gateway_verify?: string;
   pure_mode?: string;
+  readOnly: boolean;
+  size?: string;
+  source: string;
+  sourcePerClusterMember?: ClusterSpecificValues;
+  yaml?: string;
   zfs_clone_copy?: string;
   zfs_export?: string;
   zfs_pool_name?: string;
-  yaml?: string;
-  barePool?: LxdStoragePool;
 }
 
 interface Props {
--- a/src/pages/storage/forms/StoragePoolFormMain.tsx
+++ b/src/pages/storage/forms/StoragePoolFormMain.tsx
@@ -20,6 +20,8 @@ import { useSettings } from "context/useSettings";
 import ScrollableForm from "components/ScrollableForm";
 import { ensureEditMode } from "util/instanceEdit";
 import { optionTrueFalse } from "util/instanceOptions";
+import StoragePoolClusteredSourceSelector from "./StoragePoolClusteredSourceSelector";
+import { isClusteredServer } from "util/settings";
 
 interface Props {
   formik: FormikProps<StoragePoolFormValues>;
@@ -46,6 +48,12 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
   const isPowerFlexDriver = formik.values.driver === powerFlex;
   const isPureDriver = formik.values.driver === pureStorage;
   const storageDriverOptions = getStorageDriverOptions(settings);
+  const hasClusterWideSource = isCephDriver || isCephFSDriver;
+  const hasSource = !isPureDriver && !isPowerFlexDriver;
+
+  const sourceHelpText = formik.values.isCreating
+    ? getSourceHelpForDriver(formik.values.driver)
+    : "Source can't be changed";
 
   return (
     <ScrollableForm>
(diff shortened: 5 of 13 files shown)


src/api/storage-pools.tsx
```diff
@@ -97,14 +98,26 @@ const getClusterAndMemberPools = (pool: Partial<LxdStoragePool>) => {
...
+            source: sourcePerClusterMember?.[item.server_name],
+          },
+        };
+        return createPool(clusteredMemberPool, item.server_name);
+      }),
     )
       .then(handleSettledResult)
       .then(() => {
+        // The pool source is cluster member specific, so we omit it on the cluster wide pool configuration.
+        delete clusterPool.config?.source;
```
edlerd, line 120:
```suggestion
```
I think we don't need this, because the member specific fields are filtered out by the function `getClusterAndMemberPools` that is generating the `clusterPool` object.

Btw. can you please rename `getClusterAndMemberPools` to `getClusterAndMemberPoolPayload`? We discovered yesterday the naming is off.

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
```diff
@@ -0,0 +1,49 @@
...
+          void formik.setFieldValue("sourcePerClusterMember", value);
+        }}
+        toggleReadOnly={() => {
+          ensureEditMode(formik);
+          focusField("sourcePerClusterMember");
+        }}
+        memberNames={memberNames}
+        isDefaultSpecific={currentValues.some(
+          (item) => item !== currentValues[0],
+        )}
```
edlerd, line 41: I think this calculation is better moved into the `ClusterSpecificInput`. It has all the data to do the calculation locally.

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -75,6 +76,7 @@ export interface StoragePoolFormValues {
   pure_gateway?: string;
   pure_gateway_verify?: string;
   pure_mode?: string;
+  sourcePerClusterMember?: ClusterSpecificValues;
```
edlerd, line 79: Please move the `source: string` next to this field. The fields were not properly sorted before.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -124,22 +129,29 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
-                  : "Source can't be changed"
-              }
-              label="Source"
-            />
-          )}
+          {!isPureDriver &&
+            !isPowerFlexDriver &&
+            (!isDirDriver &&
+            !isZfsDriver &&
+            !isLvmDriver &&
```
edlerd, line 136: I think the logic here is a bit off. For powerflex, pure, btrfs (and I think also for ZFS) we better hide the source field. Because it is not a valid config for those drivers.

src/util/storagePoolForm.tsx
```diff
@@ -1,9 +1,19 @@
...
+import type {
+  LxdStoragePool,
+  LXDStoragePoolOnClusterMember,
+} from "types/storage";
 import { StoragePoolFormValues } from "pages/storage/forms/StoragePoolForm";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 
 export const toStoragePoolFormValues = (
   pool: LxdStoragePool,
+  poolOnMembers?: LXDStoragePoolOnClusterMember[],
```
edlerd, line 10: The call in `EditStoragePool` needs to be adjusted to load and construct the `poolOnMembers`. I think with the current changes, we won't see different sources per member in the UI on edit.

edlerd (commented):
Good progress, some comments on the code below.

src/api/storage-pools.tsx
Kxiru: Sure thing :)

src/pages/storage/forms/StoragePoolFormMain.tsx
Kxiru: As previosuly discussed, I have condensed the logic above into a single variable, canClusterSource. Please review when I next push :)

src/util/storagePoolForm.tsx
Kxiru: On it! :computer:

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const StoragePoolClusteredSourceSelector: FC<Props> = ({ formik }) => {
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const memberNames: string[] = [];
+  clusterMembers.forEach((member) => memberNames.push(member.server_name));
+
+  return (
+    <>
+      <Label>Source</Label>
+      <ClusterSpecificInput
```
edlerd, line 22: We should also disable the `Same for all cluster members` checkbox in case of the edit form.

And I think we should preserve the help text `Source can't be changed` below the last input in the edit case.

src/api/storage-pools.tsx
```diff
@@ -97,11 +100,21 @@ const getClusterAndMemberPools = (pool: Partial<LxdStoragePool>) => {
 export const createClusteredPool = (
   pool: LxdStoragePool,
   clusterMembers: LxdClusterMember[],
+  sourcePerClusterMember?: ClusterSpecificValues,
 ): Promise<void> => {
-  const { memberPool, clusterPool } = getClusterAndMemberPools(pool);
+  const { memberPool, clusterPool } = getClusterAndMemberPoolPayload(pool);
```
mas-who, line 105: we should keep variable naming consistent with the function name change i.e. use `memberPoolPayload` and `clusterPoolPayload` instead

src/api/storage-pools.tsx
```diff
@@ -132,7 +145,7 @@ export const updateClusteredPool = (
   pool: Partial<LxdStoragePool>,
   clusterMembers: LxdClusterMember[],
 ): Promise<void> => {
-  const { memberPool, clusterPool } = getClusterAndMemberPools(pool);
+  const { memberPool, clusterPool } = getClusterAndMemberPoolPayload(pool);
```
mas-who, line 148: same comment about variable naming here

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+        values={formik.values.sourcePerClusterMember}
+        id="sourcePerClusterMember"
+        isReadOnly={false}
+        onChange={(value) => {
+          void formik.setFieldValue("sourcePerClusterMember", value);
+        }}
+        toggleReadOnly={() => {
+          ensureEditMode(formik);
+          focusField("sourcePerClusterMember");
```
edlerd, line 31:
```suggestion
```
We don't need this here, read only is hardcoded to false above.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -124,22 +130,26 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
-              help={
-                formik.values.isCreating
-                  ? getSourceHelpForDriver(formik.values.driver)
-                  : "Source can't be changed"
-              }
-              label="Source"
-            />
-          )}
+          {!isPureDriver &&
+            !isPowerFlexDriver &&
```
edlerd, line 134: Maybe add a `hasSource` constant above to make this easier to read?

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -124,22 +130,26 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
+            (canClusterSource && isClusteredServer(settings) ? (
+              <Input
+                {...getFormProps("source")}
+                type="text"
+                disabled={
+                  formik.values.driver === btrfsDriver ||
```
mas-who, line 140: can we confirm why the logic was put in place before? According to the [docs](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_btrfs/), the `btrfs` driver allows for the `source` config.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -43,9 +46,12 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
   const isCephDriver = formik.values.driver === cephDriver;
   const isCephFSDriver = formik.values.driver === cephFSDriver;
   const isDirDriver = formik.values.driver === dirDriver;
+  const isLvmDriver = formik.values.driver === lvmDriver;
   const isPowerFlexDriver = formik.values.driver === powerFlex;
   const isPureDriver = formik.values.driver === pureStorage;
+  const isZfsDriver = formik.values.driver == zfsDriver;
   const storageDriverOptions = getStorageDriverOptions(settings);
+  const canClusterSource = !isDirDriver && !isZfsDriver && !isLvmDriver;
```
mas-who, line 54: I think `hasSourceConfig` is a bit clearer wdyt?

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -124,22 +130,26 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
-                formik.values.isCreating
-                  ? getSourceHelpForDriver(formik.values.driver)
-                  : "Source can't be changed"
-              }
-              label="Source"
-            />
-          )}
+          {!isPureDriver &&
+            !isPowerFlexDriver &&
+            (canClusterSource && isClusteredServer(settings) ? (
```
mas-who, line 135: I don't think this logic here is correct, if it's not a cluster, it would show the `StoragePoolClusteredSourceSelector` component?

src/pages/storage/forms/StoragePoolFormMain.tsx
edlerd, line 54:
```suggestion
  const hasClusterWideSource = isCephDriver || isCephFSDriver;
```
I think defining this positively without negation is much easier to read.

src/pages/storage/forms/StoragePoolFormMain.tsx
edlerd, line 135:
```suggestion
            (hasClusterWideSource || !isClusteredServer(settings) ? (
```
Shouldn't this be negated, so we avoid going into the cluster specific branch in case of a non-clustered server?

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const StoragePoolClusteredSourceSelector: FC<Props> = ({ formik }) => {
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const memberNames: string[] = [];
+  clusterMembers.forEach((member) => memberNames.push(member.server_name));
+
+  return (
+    <>
+      <Label>Source</Label>
```
mas-who, line 21:
```suggestion
      <Label forId="sourcePerClusterMember">Source</Label>
```

We should also make this field required if we are creating the storage pool

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
edlerd: Or maybe better, set `canToggleSpecific` to false in case of editing, so the checkbox is not even visible.

edlerd (commented):
It is fully functional now, good progress. I found some QA and code issues below.

src/pages/storage/forms/StoragePoolFormMain.tsx
mas-who: Additionally, we should consider that for `ceph` storage pools, the `source` config exists but it is not member specific. I suggest go through the [docs](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_drivers/) on storage drivers to confirm if `source` config exists for each driver.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+          label="Same for all cluster members"
+          checked={!isSpecific}
+          onChange={() => {
+            isUserActionRef.current = true;
+            setIsSpecific((val) => !val);
+          }}
+        />
+      )}
+      {isSpecific && (
+        <div className="cluster-specific-input">
```
mas-who, line 76: where did we define the `cluster-specific-input` css class?

mas-who (commented):
Good work! Some code comments so far

mas-who:
@Kxiru I realise some of my comments may have overlapped with comments from @edlerd . Please resolve the issues if it's already addressed by your responses to @edlerd

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  useEffect(() => {
+    if (!isUserActionRef.current && values && Object.keys(values).length > 0) {
+      const newDefaultSpecific = !Object.values(values).every(
+        (item) => item === Object.values(values)[0],
+      );
+      setIsSpecific(newDefaultSpecific);
+    }
+    isUserActionRef.current = false;
+  }, [values]);
```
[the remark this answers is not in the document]
Kxiru: I see... The reasoning for this implementation was because the logic to display the correct version of the source (global or individual version of the source(s)) upon page load was not working properly because on page load, the values array - from which this is discerned - is not yet populatetd. 

The idea was to then set specificity whenever Values is updated, which worked for showing global values if I remember correctly, but when attempting to switch from individual to global values, the check button no longer worked as isSpecific is always evaluating to false at this point because Values is populated, thus overriding the user's intention to (un)check the box.

So then the idea was to be able to flick the checkbox as long as the user has been the one to initiate the action. The use effect basically:
- Checks if values are ready, by skipping the recalculation of isSpecific if values is undefined or incomplete.
- Respects user actions, it avoids overwriting isSpecific if the user has already toggled it.

It's a bit convoluted, but I wasn't sure how else to resolve the situation of loading the correct version of the CSI if the values array doesn't load when the page does. One way we could do this though is with an isLoading flag? And perhaps displaying a loader whilst values is being populated?

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+                </div>
+                <div className="cluster-specific-value">
+                  {isReadOnly ? (
+                    <>
+                      {activeValue}
+                      <FormEditButton toggleReadOnly={toggleReadOnly} />
+                    </>
+                  ) : (
+                    <Input
+                      id={`${id}-${item}`}
```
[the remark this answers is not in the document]
Kxiru: The global state input also has id={id} (Line 119), will this cause conflicts?

src/pages/storage/forms/StoragePoolFormMain.tsx
Kxiru: That was the intention of "canClusterSource" in line 54. PureDrivers and Powerflex drivers are excluded from this feature. Or do you mean something else?

src/pages/storage/forms/StoragePoolFormMain.tsx
Kxiru: @edlerd Should BTRFS be included in this feature?

src/pages/storage/forms/StoragePoolFormMain.tsx
Kxiru: I thought most drivers had source fields and the clusterability of them came down to whether they (the sources) could be clustered or not. I was not under the impression that just because a driver _has_ a source, it can automatically be clustered.
In any case, this was refactored to be ```  const hasClusterWideSource = isCephDriver || isCephFSDriver;``` as per David's suggestion :)

src/pages/storage/forms/StoragePoolFormMain.tsx
Kxiru: On second thought, refactoring `````` to ```  const hasClusterWideSource = isCephDriver || isCephFSDriver;``` automatically includes all of the drivers. (Pure and Powerflex are excluded later).

src/pages/storage/forms/StoragePoolFormMain.tsx
Kxiru: Do you mind having another look at this area in the amended commit?

src/components/forms/ClusterSpecificInput.tsx
Kxiru: I actually didn't create this, though it does have its own scss file.

src/components/forms/ClusterSpecificInput.tsx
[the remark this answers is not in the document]
edlerd: Only one of the two should be rendered at a time, so there shouldn't be a conflict

src/components/forms/ClusterSpecificInput.tsx
edlerd: We could extract a pure layout component displaying the member name next to some input that is referencing the CSS instead of copying the structure between the cluster member specific -select and -input components.

src/pages/storage/forms/StoragePoolFormMain.tsx
edlerd: It might be a mistake that earlier we disabled source for btrfs.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+  return (
+    <div className="u-sv3">
+      {canToggleSpecific && !isReadOnly && (
+        <CheckboxInput
+          id={`${id}-same-for-all-toggle`}
+          label="Same for all cluster members"
+          checked={!isSpecific}
+          onChange={() => {
+            isUserActionRef.current = true;
+            setIsSpecific((val) => !val);
```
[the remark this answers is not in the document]
edlerd: Sure, we can have a chat tomorrow. As a rule of thumb: user interaction resulting in a state change should better happen directly and not through indirection of use effects. I understand the concerns here are a bit mixed from your other comment.

src/components/forms/ClusterSpecificInput.tsx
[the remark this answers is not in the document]
edlerd: If I understand correctly, to solve the init case you could set is specific to null in the edit case. Then the use effect can check if values are loaded and is specific is null. Only in that case the use effect will call set is specific. User interaction with is specific should be directly handled and not be effected by this.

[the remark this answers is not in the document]
Kxiru: Ah, that's a great idea and a good approach, I'll implement that now and recommit.

src/components/forms/ClusterSpecificInput.tsx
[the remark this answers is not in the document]
Kxiru: I think I have implemented your above suggestion successfully, please tell me what you think :)

src/components/forms/ClusterSpecificInput.tsx
Kxiru: Hmm. Is there a reason why we would do this? Is there something wrong with the initial implementation or is it just a matter of further abstracting for each use case?

edlerd: I threw the idea out because that is one way to ensure CSS is related to exactly one piece of code. Though on second thought that component might complicate things and be the wrong abstraction to use. I'd be fine to accept the duplication in the two cluster member specific components and revisit this in a follow up.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+    isDefaultSpecific,
+  );
+  const firstValue = Object.values(values ?? {})[0];
+  const isUserActionRef = useRef(false);
+
+  useEffect(() => {
+    if (isSpecific === null && values) {
+      const newDefaultSpecific = !Object.values(values).every(
+        (item) => item === Object.values(values)[0],
+      );
```
edlerd, line 44:
```suggestion
      const newDefaultSpecific = Object.values(values).some(
        (item) => item !== Object.values(values)[0],
      );
```
Maybe bit more straight forward of the same logic?

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+  isDefaultSpecific = null,
+  clusterMemberLinkTarget = () => "/ui/cluster",
+  disabled = false,
+  helpText,
+}) => {
+  const [isSpecific, setIsSpecific] = useState<boolean | null>(
+    isDefaultSpecific,
+  );
+  const firstValue = Object.values(values ?? {})[0];
+  const isUserActionRef = useRef(false);
```
edlerd, line 38:
```suggestion
```
I think this is not needed any more.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  useEffect(() => {
+    if (isSpecific === null && values) {
+      const newDefaultSpecific = !Object.values(values).every(
+        (item) => item === Object.values(values)[0],
+      );
+      setIsSpecific(newDefaultSpecific);
+    }
+    isUserActionRef.current = false;
+  }, [values]);
```
edlerd, line 48:
```suggestion
  }, [isSpecific, values]);
```
`isSpecific` needs to be in the dependencies. Otherwise we risk evaluating an outdated value inside.

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <>
+      <Label forId="sourcePerClusterMember">Source</Label>
+      <ClusterSpecificInput
+        values={formik.values.sourcePerClusterMember}
+        id="sourcePerClusterMember"
+        isReadOnly={false}
+        onChange={(value) => {
+          void formik.setFieldValue("sourcePerClusterMember", value);
+        }}
+        canToggleSpecific={formik.values.isCreating ? true : false}
```
edlerd, line 27:
```suggestion
        canToggleSpecific={formik.values.isCreating}
```

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+                </div>
+                <div className="cluster-specific-value">
+                  {isReadOnly ? (
+                    <>
+                      {activeValue}
+                      <FormEditButton toggleReadOnly={toggleReadOnly} />
+                    </>
+                  ) : (
+                    <Input
+                      id={id}
```
edlerd, line 98: We can't set id for all memberName iterations, only for the first one, the others need to append the member name. Because the id needs to be unique.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+                      type="text"
+                      className="u-no-margin--bottom"
+                      value={activeValue}
+                      onChange={(e) => setValueForMember(e.target.value, item)}
+                      disabled={disabled}
```
edlerd, line 103: We should also set the help text for the last item in the list, otherwise the disable reason is invisible on a cluster specific input.

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
```diff
@@ -0,0 +1,37 @@
...
+        toggleReadOnly={() => {}}
+        memberNames={memberNames}
+        disabled={!formik.values.isCreating}
+        helpText="Source can't be changed"
```
edlerd, line 31: The help text should only be visible if the field is disabled / when editing.

edlerd (changes requested):

src/pages/storage/forms/StoragePoolFormMain.tsx
mas-who: I think there are 3 scenarios concretely:
1. Some drivers (`pure`, `powerflex`) does not have source config at all
2. Some drivers (`lvm`, `zfs`, `dir`, `btrfs`) have source config and can be member specific in a clustered environment.
3. Some drivers (`ceph`, `cephfs`) have source config, but the way you would set the source is exactly the same regardless of lxd being clustered or not.

The logic should reflect the above so that:
1. We should hide the source config
2. We should allow source config modification upon creation of the storage pool and it should be member specific if lxd is clustered. For storage pool edit page, we should display the source config (member specific if lxd is clustered) but disable modification. 
3. We should allow source config modification upon creation of the storage pool and it should be a single input regardless of lxd being clustered or not. For storage pool edit page, we should display the source config as a single input always and disable its modification.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -124,22 +128,25 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
-              label="Source"
-            />
-          )}
+          {hasSource &&
+            (hasClusterWideSource || !isClusteredServer(settings) ? (
+              <Input
+                {...getFormProps("source")}
+                type="text"
+                disabled={
+                  formik.values.driver === btrfsDriver ||
```
mas-who, line 137: @edlerd why is the source input disabled specifically for the `btrfs` driver? As far as I can see, the source config is configurable? (see [ref](https://documentation.ubuntu.com/lxd/en/latest/reference/storage_btrfs/#:~:text=in%20these%20pools.-,Storage%20pool%20configuration,-%C2%B6))

src/pages/storage/forms/StoragePoolFormMain.tsx
mas-who: looks good now :+1:

src/components/forms/ClusterSpecificInput.tsx
mas-who: sounds good to me :+1:

src/pages/storage/forms/StoragePoolFormMain.tsx
mas-who: aah sorry I missed that one, @Kxiru I think the condition here should be removed still since we should allow source input for the `btrfs` driver on creation.

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
Kxiru: Sigh. How did I not see this one :face_exhaling:

src/components/forms/ClusterSpecificInput.tsx
Kxiru: The help text is in line 108, however we can render it with the last item if you wish?

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+    isDefaultSpecific,
+  );
+  const firstValue = Object.values(values ?? {})[0];
+
+  useEffect(() => {
+    if (isSpecific === null && values) {
+      const newDefaultSpecific = !Object.values(values).some(
+        (item) => item !== Object.values(values)[0],
+      );
+      setIsSpecific(newDefaultSpecific);
```
edlerd, line 44:
```suggestion
      const rawValues = Object.values(values);
      const newDefaultSpecific = rawValues.some((item) => item !== rawValues[0]);
      setIsSpecific(newDefaultSpecific);
```
I think this should be inverted. Also a bit easier to read by introducing an interim variable.

src/pages/storage/forms/StoragePoolClusteredSourceSelector.tsx
```diff
@@ -0,0 +1,40 @@
+import { Label } from "@canonical/react-components";
+import ClusterSpecificInput from "components/forms/ClusterSpecificInput";
+import { FormikProps } from "formik";
+import { FC } from "react";
+import { StoragePoolFormValues } from "./StoragePoolForm";
+import { useClusterMembers } from "context/useClusterMembers";
+
+interface Props {
+  formik: FormikProps<StoragePoolFormValues>;
+  helpText: string;
```
edlerd, line 10:
```suggestion
  helpText?: string;
```
I think this should be optional.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -46,6 +48,12 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
   const isPowerFlexDriver = formik.values.driver === powerFlex;
   const isPureDriver = formik.values.driver === pureStorage;
   const storageDriverOptions = getStorageDriverOptions(settings);
+  const hasClusterWideSource = isCephDriver || isCephFSDriver;
+  const hasSource = !isPureDriver && !isPowerFlexDriver;
+
+  const helpText = formik.values.isCreating
```
edlerd, line 54:
```suggestion
  const sourceHelpText = formik.values.isCreating
```
This component handles other fields, better to be specific.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+                      value={activeValue}
+                      onChange={(e) => setValueForMember(e.target.value, item)}
+                      disabled={disabled}
+                    />
+                  )}
+                </div>
+              </Fragment>
+            );
+          })}
+          {helpText && <div className="p-form-help-text">{helpText}</div>}
```
edlerd, line 111: This should span the whole width and not be limited to the left side:

image

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -46,6 +48,12 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
+    ? getSourceHelpForDriver(formik.values.driver)
```
edlerd, line 55: Please update the `getSourceHelpForDriver` function for btrfs, as we enabled it in this PR, the current help text needs to be adjusted:

image

src/sass/_cluster_specific_input.scss
```diff
@@ -8,4 +8,8 @@
   .cluster-specific-member {
     overflow: hidden;
   }
+
+  .clustered-input-helptext {
```
edlerd, line 12:
```suggestion
  .cluster-specific-helptext {
```
Better keep the naming aligned here and in the usage

edlerd (commented):
QA and code LGTM, tiny non-blocking nitpick on the class naming, good to go as is.

PR #1074, opened 2025-01-27
feat: [WD-17724] CMS Storage Pool size field

Kxiru:
## Done

- Added ClusteredDiskSizeSelector

Fixes 
- Inability to host Cluster member specific storage pool sizes.

## QA

1. Run the LXD-UI:

## Screenshots
image

image

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -102,6 +102,7 @@ export const createClusteredPool = (
   clusterMembers: LxdClusterMember[],
   sourcePerClusterMember?: ClusterSpecificValues,
   zfsPoolNamePerClusterMember?: ClusterSpecificValues,
+  sizePerClusterMember?: ClusterSpecificValues,
 ): Promise<void> => {
   const { memberPoolPayload, clusterPoolPayload } =
     getClusterAndMemberPoolPayload(pool);
@@ -114,6 +115,7 @@ export const createClusteredPool = (
             ...memberPoolPayload.config,
             source: sourcePerClusterMember?.[item.server_name],
 
+            size: sizePerClusterMember?.[item.server_name],
             ...(zfsPoolNamePerClusterMember?.[item.server_name] && {
               "zfs.pool_name": zfsPoolNamePerClusterMember[item.server_name],
             }),
@@ -150,14 +152,22 @@ export const updatePool = (
 export const updateClusteredPool = (
   pool: Partial<LxdStoragePool>,
   clusterMembers: LxdClusterMember[],
+  sizePerClusterMember?: ClusterSpecificValues,
 ): Promise<void> => {
   const { memberPoolPayload, clusterPoolPayload } =
     getClusterAndMemberPoolPayload(pool);
   return new Promise((resolve, reject) => {
     Promise.allSettled(
-      clusterMembers.map(async (item) =>
-        updatePool(memberPoolPayload, item.server_name),
-      ),
+      clusterMembers.map(async (item) => {
+        const clusteredMemberPool = {
+          ...memberPoolPayload,
+          config: {
+            ...memberPoolPayload.config,
+            size: sizePerClusterMember?.[item.server_name],
+          },
+        };
+        return updatePool(clusteredMemberPool, item.server_name);
+      }),
     )
       .then(handleSettledResult)
       .then(() => updatePool(clusterPoolPayload))
--- a/src/components/forms/ClusteredDiskSizeSelector.tsx
+++ b/src/components/forms/ClusteredDiskSizeSelector.tsx
@@ -1,0 +1,116 @@
+import { FC, Fragment, useEffect, useState } from "react";
+import { CheckboxInput, Label } from "@canonical/react-components";
+import ResourceLink from "components/ResourceLink";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
+import { useClusterMembers } from "context/useClusterMembers";
+import DiskSizeSelector from "./DiskSizeSelector";
+
+interface Props {
+  id: string;
+  setValue: (value: ClusterSpecificValues) => void;
+  values?: ClusterSpecificValues;
+  helpText?: string;
+}
+
+const ClusteredDiskSizeSelector: FC<Props> = ({
+  id,
+  setValue,
+  values,
+  helpText,
+}) => {
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const memberNames = clusterMembers.map((member) => member.server_name);
+  const [isSpecific, setIsSpecific] = useState<boolean | null>(null);
+  const firstValue = Object.values(values ?? {})[0];
+
+  useEffect(() => {
+    const rawValues = Object.values(values ?? {});
+    if (isSpecific === null && rawValues.length > 0) {
+      const newDefaultSpecific = rawValues.some(
+        (item) => item !== rawValues[0],
+      );
+      setIsSpecific(newDefaultSpecific);
+    }
+  }, [isSpecific, values]);
+
+  const setValueForAllMembers = (value: string) => {
+    const update: ClusterSpecificValues = {};
+    memberNames.forEach((member) => (update[member] = value));
+    setValue(update);
+  };
+
+  const setValueForMember = (value: string, member: string) => {
+    const update = {
+      ...values,
+      [member]: value,
+    };
+    setValue(update);
+  };
+
+  return (
+    <div className="u-sv3">
+      <Label forId="sizePerClusterMember">Size</Label>
+      {
+        <CheckboxInput
+          id={`${id}-same-for-all-toggle`}
+          label="Same for all cluster members"
+          checked={!isSpecific}
+          onChange={() => {
+            setValueForAllMembers(firstValue);
+            setIsSpecific((val) => !val);
+          }}
+        />
+      }
+      {isSpecific && (
+        <div className="cluster-specific-input">
+          {memberNames.map((item) => {
+            const activeValue = values?.[item] ?? "";
+
+            return (
+              <Fragment key={item}>
+                <div className="cluster-specific-member">
+                  <ResourceLink
+                    type="cluster-member"
+                    value={item}
+                    to={"/ui/cluster"}
+                  />
+                </div>
+
+                <div className="cluster-specific-value">
+                  <DiskSizeSelector
+                    id={memberNames.indexOf(item) === 0 ? id : `${id}-${item}`}
+                    value={activeValue}
+                    setMemoryLimit={(value) => setValueForMember(value, item)}
+                    disabled={false}
+                    classname="u-no-margin--bottom"
+                  />
+                </div>
+              </Fragment>
+            );
+          })}
+          {helpText && (
+            <div className="p-form-help-text cluster-specific-helptext">
+              {helpText}
+            </div>
+          )}
+        </div>
+      )}
+
+      {!isSpecific && (
+        <div>
+          {
+            <DiskSizeSelector
+              id={id}
+              value={firstValue}
+              setMemoryLimit={(value) => setValueForAllMembers(value)}
+              disabled={false}
+              help={helpText}
+            />
+          }
+        </div>
+      )}
+    </div>
+  );
+};
+
+export default ClusteredDiskSizeSelector;
--- a/src/pages/storage/forms/StoragePoolForm.tsx
+++ b/src/pages/storage/forms/StoragePoolForm.tsx
@@ -72,6 +72,7 @@ export interface StoragePoolFormValues {
   powerflex_sdt?: string;
   powerflex_user_name?: string;
   powerflex_user_password?: string;
+  sizePerClusterMember?: ClusterSpecificValues;
   pure_api_token?: string;
   pure_gateway?: string;
   pure_gateway_verify?: string;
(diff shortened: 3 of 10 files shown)


src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -55,6 +60,15 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
     ? getSourceHelpForDriver(formik.values.driver)
     : "Source can't be changed";
 
+  const hasSize =
```
mas-who, line 63: probably easier to just set `hasSize = isZfsDriver || isLvmDriver || isBtrfsDriver`. We don't need the `canClusterSize` variable below because for the drivers that have size config, it's all member specific if lxd is clustered.

src/components/forms/ClusteredDiskSizeSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+}) => {
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const memberNames = clusterMembers.map((member) => member.server_name);
+
+  const [isSpecific, setIsSpecific] = useState<boolean | null>(
+    isDefaultSpecific,
+  );
+  const firstValue = Object.values(sizeValues ?? {})[0];
+
+  const limit = (val: string) => {
```
edlerd, line 47:
```suggestion
  const parseLimit = (val: string) => {
```
That should be easier to read.

src/components/forms/ClusteredDiskSizeSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+                          memberNames.indexOf(item) === 0 ? id : `${id}-${item}`
+                        }
+                        name="limits_disk"
+                        type="number"
+                        min="0"
+                        step="Any"
+                        placeholder="Enter value"
+                        onChange={(e) =>
+                          setValueForMember(
+                            e.target.value + limit(activeValue || "").unit,
```
edlerd, line 133: I think better to refactor the definition as `const activeValue = sizeValues?.[item] ?? "";` and then we can simplify all the use cases of it. Also the ones below.
```suggestion
                            e.target.value + limit(activeValue).unit,
```

src/components/forms/ClusteredDiskSizeSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+                            ? limit(activeValue || "").value
+                            : ""
+                        }
+                        className="u-no-margin--bottom"
+                      />
+                      <Select
+                        id={
+                          memberNames.indexOf(item) === 0
+                            ? "memUnitSelect"
+                            : `memUnitSelect-${item}`
```
edlerd, line 148: I think we should append the id prop from this component. We might use this component more than once per page and with the current code it would generate duplicate ids.

src/components/forms/ClusteredDiskSizeSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+                  onChange={(e) =>
+                    setValueForAllMembers(
+                      `${limit(firstValue || "").value ?? 0}${e.target.value}`,
+                    )
+                  }
+                  value={limit(firstValue || "").unit}
+                  disabled={disabled}
+                />
+              </div>
+              {helpText && <div className="p-form-help-text">{helpText}</div>}
```
edlerd, line 222: I think instead of repeating this in here twice, we should invoke the `DiskSizeSelector` component twice. That should make this component way slimmer, as we can also remove some of the helper functions.

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -70,14 +70,15 @@ export interface StoragePoolFormValues {
...
+  poolSizePerClusterMember?: ClusterSpecificValues;
   pure_api_token?: string;
   pure_gateway?: string;
   pure_gateway_verify?: string;
   pure_mode?: string;
   readOnly: boolean;
   size?: string;
   source: string;
-  sourcePerClusterMember?: ClusterSpecificValues;
+  poolSourcePerClusterMember?: ClusterSpecificValues;
```
edlerd, line 81: I think the pool prefix is redundant. This is the `StoragePoolFormValues` type after all.

src/pages/storage/forms/StoragePoolFormMain.tsx
edlerd: agreed.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -112,11 +126,8 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
             required
             disabled={!formik.values.isCreating}
           />
-          {!isCephDriver &&
-            !isCephFSDriver &&
-            !isDirDriver &&
-            !isPureDriver &&
-            !isPowerFlexDriver && (
+          {hasSize &&
+            (!canClusterSize || !isClusteredServer(settings) ? (
```
edlerd, line 130: After removing canClusterSize, we should revert the conditional to avoid the negation here and make this easier to read. Positive (non-negated) boolean usage is easier to read in general.

src/util/storagePoolForm.tsx
```diff
@@ -9,21 +9,16 @@ export const toStoragePoolFormValues = (
   pool: LxdStoragePool,
   poolOnMembers?: LXDStoragePoolOnClusterMember[],
 ): StoragePoolFormValues => {
-  const sourcePerClusterMember: ClusterSpecificValues = {};
-  poolOnMembers?.forEach(
-    (item) =>
-      (sourcePerClusterMember[item.memberName] = item.config?.source ?? ""),
-  );
+  const poolSourcePerClusterMember: ClusterSpecificValues = {};
+  const poolSizePerClusterMember: ClusterSpecificValues = {};
```
edlerd, line 13: I think the pool prefix is redundant.

src/api/storage-pools.tsx
edlerd, line 1: There is missing updates to the edit flow in the `EditStoragPool` component and also the api functions for it.

edlerd (changes requested):
Code mostly LGTM, two minor issues below. Two QA issue:

1. On pool edit, the size can't be changed from cluster member specific to non-specific or back. The checkbox is not rendered. 
2. Please check the failing tests.

edlerd:
> * On pool edit, the size can't be changed from cluster member specific to non-specific or back. The checkbox is not rendered.

@Kxiru did you also fix this one?

edlerd (changes requested):
Found QA issue, when switching from different sizes back to the same size, the change is not applied.

Scenario:
1. Create  zfs pool with different sizes per cluster member
2. edit the pool and click "Same for all cluster members" and edit the description
3. save the change
4. The pool sizes are still different per cluster member.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -112,11 +118,21 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
-            !isCephFSDriver &&
-            !isDirDriver &&
-            !isPureDriver &&
-            !isPowerFlexDriver && (
+          {hasSize &&
+            (isClusteredServer(settings) ? (
+              <ClusteredDiskSizeSelector
+                id="sizePerClusterMember"
+                sizeValues={formik.values.sizePerClusterMember}
+                toggleReadOnly={() => {}}
```
edlerd, line 126: This should do a bit more than an empty callback.

edlerd (changes requested):

src/pages/storage/forms/StoragePoolFormMain.tsx
Kxiru: What should be inside? I don't think it is necessary for this component.

Kxiru:
> Found QA issue, when switching from different sizes back to the same size, the change is not applied.
> 
> Scenario:
> 
> 1. Create  zfs pool with different sizes per cluster member
> 2. edit the pool and click "Same for all cluster members" and edit the description
> 3. save the change
> 4. The pool sizes are still different per cluster member.

Hey David!
So, one thing that can be done for this is to run setValueForAllMembers() upon clicking the checkbox input. This can take place in the onchange function in lines 73-73. Only issue with this, is that if the user changes their mind and reverts back to the cluster specific version, they would have to set all of the individual values all over again (Or refresh the page), which could be costly in the event of having too many clusters. What do you suggest?
This is necessary for the UI to register that the user has flipped from specific to non-specific and a change has to be made. Should we move forward with this approach?

image

src/pages/storage/forms/StoragePoolFormMain.tsx
edlerd: If the user wants to only edit the size, this should enable edit mode. Otherwise the save button won't appear on changes in the edit form to the size field.

edlerd:
> So, one thing that can be done for this is to run setValueForAllMembers() upon clicking the checkbox input. This can take place in the onchange function in lines 73-73. Only issue with this, is that if the user changes their mind and reverts back to the cluster specific version, they would have to set all of the individual values all over again

I think that is fine and expected behaviour. A user can cancel the edit to restore the previous state with all cluster specific settings.

src/util/storagePoolForm.tsx
```diff
@@ -4,26 +4,25 @@ import type {
...
-    (item) =>
-      (sourcePerClusterMember[item.memberName] = item.config?.source ?? ""),
-  );
+  const sizePerClusterMember: ClusterSpecificValues = {};
+  const hasSize = zfsDriver || lvmDriver || btrfsDriver;
+
+  poolOnMembers?.forEach((item) => {
+    if (item.driver === hasSize) {
+      sizePerClusterMember[item.memberName] = item.config?.size ?? "";
+    }
```
edlerd, line 20:
```suggestion
  const isStoragePoolWithSize = (driver: string) => {
    const driversWithSize = [zfsDriver, lvmDriver, btrfsDriver];
    return driversWithSize.includes(driver);
  }

  poolOnMembers?.forEach((item) => {
    if (isStoragePoolWithSize(item.driver)) {
      sizePerClusterMember[item.memberName] = item.config?.size ?? "";
    }
```
I think this might be what was meant here? 

Probably best to move the `isStoragePoolWithSize` function into a util file, to reuse it in `ClusteredDiskSizeSelector`. That avoids introducing two places that define which drivers support the size configuration.

src/util/storagePoolForm.tsx
```diff
@@ -35,6 +34,12 @@ export const toStoragePoolFormValues = (
     cephfs_osd_pg_num: pool.config?.["cephfs.osd_pg_num"],
     cephfs_path: pool.config?.["cephfs.path"],
     cephfs_user_name: pool.config?.["cephfs.user.name"],
+    description: pool.description,
+    driver: pool.driver,
+    entityType: "storagePool",
+    isCreating: false,
+    name: pool.name,
+    sizePerClusterMember,
```
edlerd, line 42: This can be next to `size` config field below.

src/components/forms/ClusteredDiskSizeSelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+  canToggleSpecific?: boolean;
+  id: string;
+  toggleReadOnly: () => void;
+  setMemoryLimit: (value: ClusterSpecificValues) => void;
+  sizeValues?: ClusterSpecificValues;
+  isReadOnly?: boolean;
+  isDefaultSpecific?: boolean;
+  clusterMemberLinkTarget?: (member: string) => string;
+  disabled?: boolean;
+  helpText?: string;
```
edlerd, line 19: Only 5 of those props are in use, we should probably remove the unused ones.

edlerd: Also probably best to align the naming, like `setMemoryLimit` => `setValue` and `sizeValues` => `values`.

edlerd (commented):
QA LGTM, some smaller ideas to clean up the code. Then this should be all good to merge :)

src/components/forms/ClusteredDiskSizeSelector.tsx
Kxiru: I have aligned the naming, however I'm not sure what you mean when you say only 5 of the props are in use. They are all referenced /used in the file at least once. Please could you clarify?

edlerd: There is only 1 usage of the component and that is not specifying `isDefaultSpecific`, there are other props are not in use.

edlerd:
Please rebase this on main, to resolve conflicts.

src/util/storagePoolForm.tsx
Kxiru: storagePoolForm is actually already a util file. Should it stay within?

src/components/forms/ClusteredDiskSizeSelector.tsx
Kxiru: So are you suggesting that for examples like canToggleSpecific, pictured below, it would be best to get rid of it as a prop and just have a const canToggleSpecific = true?

image

edlerd: Suggesting to removing the `canToggleSpecific` variable completely, and remove the unused code branches to simplify the interface and the component itself.

Kxiru: Unused code branches?
I think I understand the ask now, please see latest commit.

edlerd (commented):
This is nearly done, found one more issue with updating the values after edit and two tiny clean code suggestions.

edlerd (approved):
LGTM, thanks for adding the clustered size.

PR #1080, opened 2025-01-30
feat: [WD-18674] CMS Fields - ZFS.Poolname

Kxiru:
## Done

- Implemented ClusteredZfsNameSelector

## QA

1. Run the LXD-UI:

## Screenshots

image
image
image

### DIFF
--- a/src/components/forms/ClusterSpecificInput.tsx
+++ b/src/components/forms/ClusterSpecificInput.tsx
@@ -8,7 +8,7 @@ interface Props {
   id: string;
   isReadOnly: boolean;
   onChange: (value: ClusterSpecificValues) => void;
-  toggleReadOnly: () => void;
+  toggleReadOnly?: () => void;
   memberNames: string[];
   values?: ClusterSpecificValues;
   canToggleSpecific?: boolean;
@@ -16,6 +16,8 @@ interface Props {
   clusterMemberLinkTarget?: (member: string) => string;
   disabled?: boolean;
   helpText?: string;
+  placeholder?: string;
+  classname?: string;
 }
 
 const ClusterSpecificInput: FC<Props> = ({
@@ -24,12 +26,14 @@ const ClusterSpecificInput: FC<Props> = ({
   isReadOnly,
   memberNames,
   onChange,
-  toggleReadOnly,
+  toggleReadOnly = () => {},
   canToggleSpecific = true,
   isDefaultSpecific = null,
   clusterMemberLinkTarget = () => "/ui/cluster",
   disabled = false,
   helpText,
+  placeholder,
+  classname = "u-sv3",
 }) => {
   const [isSpecific, setIsSpecific] = useState<boolean | null>(
     isDefaultSpecific,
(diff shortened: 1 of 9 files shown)


edlerd (changes requested):

src/components/ConfigurationRow.tsx
```diff
@@ -59,7 +61,13 @@ export const getConfigurationRow = ({
...
   const value = values[name];
   const isOverridden = value !== undefined;
-  const overrideValue = readOnlyRenderer(value === "" ? "-" : value);
+  const overrideValue = readOnlyRenderer(
+    clusterSpecificDisplay
+      ? clusterSpecificDisplay
+      : value === ""
+        ? "-"
+        : value,
+  );
```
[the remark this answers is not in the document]
Kxiru: If the displayed value(s) are only generated and rendered within this file, what other approach from StoragePoolFormZFS could be done without needing to make changes to this file?

We could further abstract and introduce some kind of intermediary component for special cases, which could take the place of where clusterSpecificDisplay is now. I have gone with this method and introduced "overrideValueComponent" (please see the latest push). Is this the route you were going for?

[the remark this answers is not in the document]
edlerd: The read only render is exposed as a prop and should do what you need. We can discuss tomorrow in person if the usage is unclear.

src/pages/storage/forms/StoragePoolFormZFS.tsx
```diff
@@ -19,7 +24,20 @@ const StoragePoolFormZFS: FC<Props> = ({ formik }) => {
...
           defaultValue: "",
-          children: <Input type="text" placeholder="Enter ZFS pool name" />,
+          children: !isClusteredServer(settings) ? (
+            <Input type="text" placeholder="Enter ZFS pool name" />
+          ) : (
+            <ClusteredZfsNameSelector
+              formik={formik}
+              placeholder="Enter ZFS pool name"
+            />
+          ),
```
edlerd, line 34: Better to keep the conditional without negation. That tends to be much easier to read. Especially when you have also an else branch it is preferred to swap them and keep the conditional positive.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -130,6 +136,8 @@ const ClusterSpecificInput: FC<Props> = ({
               onChange={(e) => setValueForAllMembers(e.target.value)}
               disabled={disabled}
               help={helpText}
+              placeholder={placeholder}
+              className={classNames(classname)}
```
edlerd, line 140:
```suggestion
              className={classname}
```
Can simplify this.

src/components/forms/ClusterSpecificInput.tsx
```diff
@@ -16,6 +17,8 @@ interface Props {
   clusterMemberLinkTarget?: (member: string) => string;
   disabled?: boolean;
   helpText?: string;
+  placeholder?: string;
+  classname?: string;
```
edlerd, line 21: Naming is a bit imprecise, because this only applies to the input in the non-specific branch. Having it named `classname ` implies this is about the whole component and in all cases. Then again a more specific name might be a bit convoluted, so renaming might be not a good option either. Why do we need the additional classname on that specific input in the first place?

src/pages/storage/forms/ClusteredZfsNameSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+      values={formik.values.zfsPoolNamePerClusterMember}
+      id="zfsPoolNamePerClusterMember"
+      isReadOnly={!formik.values.isCreating}
+      onChange={(value) => {
+        void formik.setFieldValue("zfsPoolNamePerClusterMember", value);
+      }}
+      canToggleSpecific={formik.values.isCreating}
+      toggleReadOnly={() => {
+        ensureEditMode(formik);
+      }}
```
edlerd, line 35:
```suggestion
```
We don't need this, because the name can't be changed.

edlerd (approved):
QA and code LGTM, good work :+1: 

Tiny nitpick to clean up an unused prop.

edlerd (approved):
All good, thanks for the iterations!

PR #1082, opened 2025-01-30
feat: initial approach for restricted users permission checks [WD-18836]

mas-who:
## Done

- Check if user is restricted by using `/1.0/identities/current` endpoint
- Wrap `fetchInstances` and `fetchInstance` inside custom hooks to apply entitlement query params if the user is restricted
- Setup custom hook for checking instance entitlements
- Applied entitlement checks for instance state action buttons on the instance list page.

## QA

1. Run the LXD-UI:
    - Check that for a non-restricted user, the instance state action buttons are enabled
    - Check that for a restricted user, with limited permissions, the instance state action buttons are disabled.

## Screenshots
[Screencast from 30-01-2025 17:48:13.webm](https://github.com/user-attachments/assets/5abe9878-7da6-4ffb-b5be-e864a1c87344)

### DIFF
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -13,23 +13,33 @@ import type { LxdOperationResponse } from "types/operation";
 import { EventQueue } from "context/eventQueue";
 import axios, { AxiosResponse } from "axios";
 import type { UploadState } from "types/storage";
+import { withEntitlementsQuery } from "util/entitlements/api";
+
+export const instanceEntitlements = ["can_update_state"];
 
 export const fetchInstance = (
   name: string,
   project: string,
-  recursion = 2,
+  isFineGrained: boolean | null,
 ): Promise<LxdInstance> => {
+  const entitlements = `&${withEntitlementsQuery(isFineGrained, instanceEntitlements)}`;
   return new Promise((resolve, reject) => {
-    fetch(`/1.0/instances/${name}?project=${project}&recursion=${recursion}`)
+    fetch(
+      `/1.0/instances/${name}?project=${project}&recursion=2${entitlements}`,
+    )
       .then(handleEtagResponse)
       .then((data) => resolve(data as LxdInstance))
       .catch(reject);
   });
 };
 
-export const fetchInstances = (project: string): Promise<LxdInstance[]> => {
+export const fetchInstances = (
+  project: string,
+  isFineGrained: boolean | null,
+): Promise<LxdInstance[]> => {
+  const entitlements = `&${withEntitlementsQuery(isFineGrained, instanceEntitlements)}`;
   return new Promise((resolve, reject) => {
-    fetch(`/1.0/instances?project=${project}&recursion=2`)
+    fetch(`/1.0/instances?project=${project}&recursion=2${entitlements}`)
       .then(handleResponse)
       .then((data: LxdApiResponse<LxdInstance[]>) => resolve(data.metadata))
       .catch(reject);
--- a/src/context/auth.tsx
+++ b/src/context/auth.tsx
@@ -4,6 +4,8 @@ import { queryKeys } from "util/queryKeys";
 import { fetchCertificates } from "api/certificates";
 import { useSettings } from "context/useSettings";
 import { fetchProjects } from "api/projects";
+import { fetchCurrentIdentity } from "api/auth-identities";
+import { useSupportedFeatures } from "./useSupportedFeatures";
 
 interface ContextProps {
   isAuthenticated: boolean;
@@ -12,6 +14,7 @@ interface ContextProps {
   isRestricted: boolean;
   defaultProject: string;
   hasNoProjects: boolean;
+  isFineGrained: boolean | null;
 }
 
 const initialState: ContextProps = {
@@ -21,6 +24,7 @@ const initialState: ContextProps = {
   isRestricted: false,
   defaultProject: "default",
   hasNoProjects: false,
+  isFineGrained: null,
 };
 
 export const AuthContext = createContext<ContextProps>(initialState);
@@ -31,6 +35,9 @@ interface ProviderProps {
 
 export const AuthProvider: FC<ProviderProps> = ({ children }) => {
   const { data: settings, isLoading } = useSettings();
+
+  const { hasEntitiesWithEntitlements, isSettingsLoading } =
+    useSupportedFeatures();
 
   const { data: projects = [], isLoading: isProjectsLoading } = useQuery({
     queryKey: [queryKeys.projects],
@@ -51,11 +58,26 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
     enabled: isTls,
   });
 
+  const { data: currentIdentity } = useQuery({
+    queryKey: [queryKeys.currentIdentity],
+    queryFn: fetchCurrentIdentity,
+    retry: false, // avoid retry for older versions of lxd less than 5.21 due to missing endpoint
+  });
+
   const fingerprint = isTls ? settings.auth_user_name : undefined;
   const certificate = certificates.find(
     (certificate) => certificate.fingerprint === fingerprint,
   );
   const isRestricted = certificate?.restricted ?? defaultProject !== "default";
+  const isFineGrained = () => {
+    if (isSettingsLoading) {
+      return null;
+    }
+    if (hasEntitiesWithEntitlements) {
+      return currentIdentity?.fine_grained ?? null;
+    }
+    return false;
+  };
 
   return (
     <AuthContext.Provider
@@ -66,6 +88,7 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
         isRestricted,
         defaultProject,
         hasNoProjects: projects.length === 0 && !isProjectsLoading,
+        isFineGrained: isFineGrained(),
       }}
     >
       {children}
--- a/src/context/useInstances.tsx
+++ b/src/context/useInstances.tsx
@@ -1,0 +1,30 @@
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { UseQueryResult } from "@tanstack/react-query";
+import { fetchInstance, fetchInstances } from "api/instances";
+import { useAuth } from "./auth";
+import type { LxdInstance } from "types/instance";
+
+export const useInstances = (
+  project: string,
+): UseQueryResult<LxdInstance[]> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.instances, project],
+    queryFn: () => fetchInstances(project, isFineGrained),
+    enabled: !!project && isFineGrained !== null,
+  });
+};
+
+export const useInstance = (
+  name: string,
+  project: string,
+  enabled?: boolean,
+): UseQueryResult<LxdInstance> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.instances, name, project],
+    queryFn: () => fetchInstance(name, project, isFineGrained),
+    enabled: enabled && isFineGrained !== null,
+  });
+};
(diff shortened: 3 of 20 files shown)


src/api/instances.tsx
```diff
@@ -13,23 +13,32 @@ import type { LxdOperationResponse } from "types/operation";
 import { EventQueue } from "context/eventQueue";
 import axios, { AxiosResponse } from "axios";
 import type { UploadState } from "types/storage";
+import { withEntitlementsQuery } from "util/permissions";
 
 export const fetchInstance = (
   name: string,
   project: string,
   recursion = 2,
+  isFineGrained?: boolean,
```
edlerd, line 22: I think we shouldn't leak this all the up to an api caller. I suspect we can instead call `useAuth` in the `withEntitlementsQuery` function.

src/util/permissions.tsx
```diff
@@ -393,3 +394,29 @@ export const getResourceOptionColumns = (type: string) => {
...
+// These will need to take into consideration combined set of entitlements required for instance related actions in lxd-ui
+export const instanceEntitlementsQuery = [
+  "can_update_state",
+  // etc.
+];
+
+export const withEntitlementsQuery = (
+  isFineGrained: boolean,
+  entityType: EntityType,
+): string => {
```
edlerd, line 407:
```suggestion
export const withEntitlementsQuery = (entitlements: string[]): string => {
```
I think we can simplify this by having the a `const entitlements = [...];` in the `api/instances.tsx` (and a different list in `api/profiles.tsx`, and analogue for all the entity types. Then pass the relevant list in here. We don't need a generic `EntityType` and distinguish here again. Basically pushing the splitting by types up and then this function can be pure.

src/util/permissions.tsx
```diff
@@ -393,3 +394,29 @@ export const getResourceOptionColumns = (type: string) => {
...
+  if (entityType === "instance") {
+    entitlements = instanceEntitlementsQuery;
+  }
+
+  return `with-access-entitlements=${entitlements.join(",")}`;
+};
+
+export const canUpdateInstanceState = (instance: LxdInstance) => {
+  return instance.access_entitlements?.includes("can_update_state") ?? false;
+};
```
edlerd, line 422: Those functions need to also encounter for unrestricted clients. Maybe add a helper for this like `const hasEntitlement(entitlements?: string[], entitlement: string)` which fetches `useAuth` and holds the fallback to `false`. Then in here we only to `return hasEntitlement(instance.access_entitlements, "can_update_state");` keeping those helper functions simple to maintain.

src/util/permissions.tsx
edlerd, line 1: I would split out the new functionality into `entitlementChecks.tsx`. Or maybe even more granular, have `util/entitlements/api.tsx` for the `withEntitlementsQuery` and `util/entitilements/instances.tsx` for the `canUpdateInstanceState` and related functions. Then add one file per type to hold the permission check helpers.

src/api/instances.tsx
mas-who: I actually tried that but it actually resulted in api call errors that caused the UI to crash. I can look a bit deeper into the issue and see if I missed something there.

src/util/permissions.tsx
mas-who: yes definitely, sorry this was a very rough sketch for the approach, so I missed the unrestricted case :+1:

src/util/permissions.tsx
mas-who: sounds good, was thinking where to put those functions :+1:

src/api/instances.tsx
mas-who: actually, this is not possible because we can't invoke hooks inside non-react component functions according to the rules of hooks. We could call the `fetchCurrentIdentity` function directly here to find out if we need to append entitlements query, but that's not ideal since we would need to make that request for all GET requests to resources :thinking:

src/util/permissions.tsx
mas-who: the approach needs to be adjusted a bit since we can't call `useAuth` inside a regular function. Instead, for each entity type, we can create a custom hook like the one below:

```
export const useInstanceEntitlements = () => {
  const { isFineGrained } = useAuth();

  const canUpdateInstanceState = (instance: LxdInstance) =>
    !isFineGrained ||
    hasEntitlement(instance?.access_entitlements || [], "can_update_state");

  return {
    canUpdateInstanceState,
  };
};
```

In the custom hook we can call `useAuth` and we can export entitlement specific functions like `canUpdateInstanceState` to be consumed in UI components. wdyt?

src/api/instances.tsx
mas-who: maybe doing an api call each time isn't the worst, according to the backend [spec](https://docs.google.com/document/d/1GxWV5J57MLrjGEY5RDG7eS86J99A8RMmjKTpn0mEhgY/edit?tab=t.0) a per-request cache was implemented so the computation won't be heavy

mas-who: Yes, I got that locally just need to neaten it up and push it :+1:

src/context/auth.tsx
```diff
@@ -66,6 +74,7 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
         isRestricted,
         defaultProject,
         hasNoProjects: projects.length === 0 && !isProjectsLoading,
+        isFineGrained: currentIdentity?.fine_grained ?? false,
```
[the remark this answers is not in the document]
mas-who: yes this makes sense. Although I don't think changing the query keys is necessary because as you mentioned, the user's identity status will not really change unless if they re-authenticate with a different user. I think returning a null here for `isFineGrained` and use that to disable the api calls makes sense

src/api/instances.tsx
```diff
@@ -13,23 +13,34 @@ import type { LxdOperationResponse } from "types/operation";
...
+
+export const instanceEntitlements = ["can_update_state"];
 
 export const fetchInstance = (
   name: string,
   project: string,
   recursion = 2,
+  isFineGrained?: boolean | null,
 ): Promise<LxdInstance> => {
+  const entitlements = `&${withEntitlementsQuery(!!isFineGrained, instanceEntitlements)}`;
```
edlerd, line 26: We should throw if fine-grained is null, or reject it in the function definition. Same for the other function below. Maybe we should throw in withEntitlementsQuery to keep the query code simpler.

src/context/useInstances.tsx
```diff
@@ -0,0 +1,31 @@
...
+    queryFn: () => fetchInstances(project, isFineGrained),
+    enabled: !!project && isFineGrained !== null,
+  });
+};
+
+export const useInstance = (
+  name: string,
+  project: string,
+  enabled?: boolean,
+  recursion?: number,
```
edlerd, line 23:
```suggestion
```
I think we should not expose the recursion parameter. If we did it needs to be in the query key.

src/util/entitlements/helpers.tsx
```diff
@@ -0,0 +1,10 @@
+export const hasEntitlement = (
+  required: string,
+  entitlements?: string[],
```
edlerd, line 3:
```suggestion
  entitlement: string,
  grantedEntitlements?: string[],
```
A bit easier to read?

src/util/entitlements/helpers.tsx
```diff
@@ -0,0 +1,10 @@
...
+  isFineGrained: boolean | null = false,
```
edlerd, line 4:
```suggestion
  isFineGrained: boolean | null,
```
I think it is unexpected to apply a default here.

src/util/entitlements/helpers.tsx
```diff
@@ -0,0 +1,10 @@
...
+): boolean => {
+  if (!isFineGrained) {
```
edlerd, line 6:
```suggestion
  if (isFineGrained === false) {
```
I think we need to be specific here. Actually I am not sure what the right thing to do is in case isFineGrained is null.

edlerd (commented):
This looks much better, thanks for the iteration. Some further comments below.

src/context/useInstances.tsx
mas-who: looking at the usage across the app, I think we should remove the `recursion` parameter to keep things simple and function signature clean

src/util/entitlements/helpers.tsx
mas-who: if `isFineGrained` is `null`, that means the api call to get identity status did not yet complete. In that case we should not grant user access to restricted actions in the UI. The purpose of the `/1.0/identities/current` endpoint is that an identity should always be able to confirm its own "restricted" status, regardless of their permissions, which means `isFineGrained` should eventually be a `boolean` value. wdyt?

mas-who: So I think the function should look like:

```
export const hasEntitlement = (
  isFineGrained: boolean | null,
  entitlement: string,
  grantedEntitlements?: string[],
): boolean => {
  if (isFineGrained === false) {
    return true;
  }

  // if isFineGrained is null, we are still awaiting the api response to determine if the user has fine grained entitlements
  // in this case, we should not grant access to the user
  if (isFineGrained === null) {
    return false;
  }

  return (grantedEntitlements || []).includes(entitlement);
};
```

src/api/instances.tsx
mas-who: good idea :+1: will do it in `withEntitlementQuery`

src/context/auth.tsx
```diff
@@ -51,11 +57,20 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
...
+    queryFn: fetchCurrentIdentity,
+    enabled: hasEntitiesWithEntitlements,
+  });
+
   const fingerprint = isTls ? settings.auth_user_name : undefined;
   const certificate = certificates.find(
     (certificate) => certificate.fingerprint === fingerprint,
   );
   const isRestricted = certificate?.restricted ?? defaultProject !== "default";
+  const isFineGrained = hasEntitiesWithEntitlements
```
mas-who, line 71: @edlerd noticed an issue due to breaking tests on older versions of lxd. If the api extension for the dry-run endpoint is missing, we should revert UI behaviour to what it was i.e. broken for restricted users. This can be done by treating all users being non-restricted. wdyt?

src/util/entitlements/helpers.tsx
edlerd: > if (isFineGrained === null) {
>     return false;
>   }

I think we should return true in this case. In the end this decision is arbitrary, though I suspect most people accessing the UI to have server admin permissions.

src/context/auth.tsx
edlerd: Yes, I think handling users on older backends as having full access and getting a not-so-nice-experience seems fine.

edlerd: But wait, the initial value of `hasEntitiesWithEntitlements` will be false, because the query to 1.0 will not be fulfilled. We need a safe way to handle that case.

src/util/entitlements/helpers.tsx
mas-who: aah yes, logic was inverted here

src/context/auth.tsx
mas-who: I think it's already safe. If `hasEntitiesWithEntitlements` is false, then we will not call the `/1.0/auth/identities/current` endpoint (which may or may not exist). In addition, the auth check in the UI will assume old lxd behaviour.

edlerd: I think old behaviour is meaning we load without entitlements and cache the response. Later, we realize this is a user with fine-grained permissions. Wouldn't this be a problem?

mas-who: I think `hasEntitiesWithEntitlements` won't be cached, when the api call in `useSupportedFeatures` completes, the returned value will be updated?

mas-who: as discussed, if the settings api call is still loading, then we will set `isFineGrained` to `null` as well to avoid resource query caching.

edlerd (approved):
Good work, this turned out to be very nice :+1:

mas-who:
> Good work, this turned out to be very nice 👍

just the beginning of the journey :smile:

PR #1083, opened 2025-01-30
Take project resource limits into account for available limits

### LINKED ISSUE #1081: Bug: limits shown incorrect in a resource restricted project
Not a high priority bug or anything, just an oddity that I noticed !

Playing around with a restricted project, I set a limit on my project for max 5GB memory and 4 CPU cores.

**The "Create Instance" flow still shows the total cpu cores and max host memory as limits:**

Image

**Instance creation obviously fails:**

Image

Image

**The proper limits are also missing when editting an existing instance:**

Image

edlerd:
## Done

- Take project resource limits into account for available cpu and memory limits.

Fixes #1081

## QA

1. Run the LXD-UI:
    - set up a project with cpu and / or memory limit. (Project configuration > Resource Limits > "Max sum of CPU" and "Max sum of memory limits"
    - configure a profile or instance, ensure the total available is in sync with system and project limitation
    - play with the project and instance/project limits to hit all different cases: project limit applies, resource limit applies, no project limit.

### DIFF
--- a/src/components/forms/CpuLimitSelector.tsx
+++ b/src/components/forms/CpuLimitSelector.tsx
@@ -1,45 +1,21 @@
 import { FC } from "react";
-import { Input, RadioInput, useNotify } from "@canonical/react-components";
-import { fetchResources } from "api/server";
-import { useQuery } from "@tanstack/react-query";
-import { queryKeys } from "util/queryKeys";
+import { RadioInput } from "@canonical/react-components";
 import { CpuLimit, CPU_LIMIT_TYPE } from "types/limits";
-import Loader from "components/Loader";
+import CpuLimitInput from "components/forms/CpuLimitInput";
+import { useProject } from "context/project";
 
 interface Props {
   cpuLimit?: CpuLimit;
   setCpuLimit: (cpuLimit: CpuLimit) => void;
+  help?: string;
 }
 
-const CpuLimitSelector: FC<Props> = ({ cpuLimit, setCpuLimit }) => {
-  const notify = useNotify();
+const CpuLimitSelector: FC<Props> = ({ cpuLimit, setCpuLimit, help }) => {
+  const { project } = useProject();
+
   if (!cpuLimit) {
     return null;
   }
-
-  const {
-    data: resources,
-    error,
-    isLoading,
-  } = useQuery({
-    queryKey: [queryKeys.resources],
-    queryFn: fetchResources,
-  });
-
-  if (isLoading) {
-    return <Loader text="Loading resources..." />;
-  }
-
-  if (error) {
-    notify.failure("Loading resources failed", error);
-  }
-
-  const numberOfCores = resources?.cpu.total;
-  const helpText = numberOfCores && (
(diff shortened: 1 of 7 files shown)


mas-who:
## QA comments
1. Is it not worthwhile to show the host max CPU and memory for the corresponding project configs? I think it's possible right since we are showing those in the instance and profile pages.
2. Not sure if we should do anything about it, but this is an interesting behaviour. When setting the max cpu and memory configs for the project, if I have instances in the project with no limits applied (either directly or via profiles), I get the following error. This essentially blocks me from setting project limits before unless if I have limits set for all instances in the project.
Screenshot from 2025-01-31 09-47-22
3. Got the following error when I tried to set max CPU to 4, and error went away when I have set the limits to 12 as suggested by the error message. I'm not sure about the logic behind this though.
Screenshot from 2025-01-31 09-53-48

src/components/forms/CpuLimitSelector.tsx
```diff
@@ -34,7 +36,17 @@ const CpuLimitSelector: FC<Props> = ({ cpuLimit, setCpuLimit }) => {
     notify.failure("Loading resources failed", error);
   }
 
-  const numberOfCores = resources?.cpu.total;
+  const getNumberOfCores = () => {
```
mas-who, line 39: maybe we can have this as a utils function that take `project` and `resources` as inputs? We can then reuse this in the project settings page to display system limits

src/components/forms/MemoryLimitSelector.tsx
```diff
@@ -40,7 +43,19 @@ const MemoryLimitSelector: FC<Props> = ({ memoryLimit, setMemoryLimit }) => {
     notify.failure("Loading resources failed", error);
   }
 
-  const maxMemory = resources?.memory.total;
+  const getAvailableBytes = (): number => {
```
mas-who, line 46: same comment here, we could have this as a utils function? This will probably need to combine with `getAvailableMemory` below though

mas-who (commented):
Thanks for this, left some QA and code comments. Once the project limits are set (noting that it's blocked if any instances in the project have no limits), profile and instance creation works nicely with respect to resource limits.

src/components/forms/CpuLimitSelector.tsx
edlerd: That is a good idea, because on configuring the project you will want to know the system details.

edlerd:
@mas-who I accidentally pushed to the wrong branch this morning and requested a review. Pushed the right changes just now.

mas-who:
@edlerd code and QA all looks good :+1: I think the errors that I pointed out in my previous QA comments are inherent to lxd's behaviour, so it's not an UI issue. The e2e tests are failing probably because the new `CpuLimitInput` component.

edlerd:
Fixed the type of the cpu input to always contain text and number correctly. @mas-who can you have another look at this one?

mas-who (approved):
LGTM :+1:

PR #1091, opened 2025-02-03
chore: update react to v19

mas-who:
## Done

- Upgraded react to v19 with latest react-components 2.0.0 release

## QA

1. Run the LXD-UI:
    - Ensure CI passes
    - Spot check demo server app

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test-report-coverage": "cp coverage/unit/coverage-final.json coverage/playwright/ ; nyc report --reporter html --reporter cobertura --reporter text-summary --temp-dir coverage/playwright --report-dir coverage/playwright-report --exclude 'src/lib/**' --exclude 'src/types/**'"
   },
   "dependencies": {
-    "@canonical/react-components": "2.0.0",
+    "@canonical/react-components": "2.1.0",
     "@monaco-editor/react": "4.6.0",
     "@tanstack/react-query": "5.63.0",
     "axios": "1.7.9",
@@ -42,7 +42,6 @@
     "react": "19.0.0",
     "react-dom": "19.0.0",
     "react-router-dom": "7.1.1",
-    "react-useportal": "1.0.19",
     "serve": "14.2.4",
     "vanilla-framework": "4.20.3",
     "xterm": "5.3.0",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test-report-coverage": "cp coverage/unit/coverage-final.json coverage/playwright/ ; nyc report --reporter html --reporter cobertura --reporter text-summary --temp-dir coverage/playwright --report-dir coverage/playwright-report --exclude 'src/lib/**' --exclude 'src/types/**'"
   },
   "dependencies": {
-    "@canonical/react-components": "1.10.0",
+    "@canonical/react-components": "2.0.0",
     "@monaco-editor/react": "4.6.0",
     "@tanstack/react-query": "5.63.0",
     "axios": "1.7.9",
@@ -39,8 +39,8 @@
     "lodash.isequal": "4.5.0",
     "node-forge": "1.3.1",
     "parse-prometheus-text-format": "1.1.1",
-    "react": "18.3.1",
-    "react-dom": "18.3.1",
+    "react": "19.0.0",
+    "react-dom": "19.0.0",
     "react-router-dom": "7.1.1",
     "react-useportal": "1.0.19",
     "serve": "14.2.4",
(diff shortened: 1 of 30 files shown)


package.json
```diff
@@ -108,6 +108,7 @@
     "src/**/*.scss": "stylelint"
   },
   "resolutions": {
-    "path-to-regexp": "8.2.0"
+    "path-to-regexp": "8.2.0",
+    "wrap-ansi": "7.0.0"
```
mas-who, line 112: I was getting the error described in this [thread](https://stackoverflow.com/questions/77406363/error-err-require-esm-require-of-es-module-node-modules-wrap-ansi-index-js) after upgrading to react v19. So fixed `wrap-ansi` version here according to the suggestion there.

edlerd:
image
Though react-components announces compatibility to react 19, it brings some transitive dependencies that do not support react 19 yet. We should aim to update or replace those as well to avoid the warnings. This will also unblock downstream projects using npm. Because npm does not silence those incompatibilities with warnings like yarn does above.

mas-who:
> image Though react-components announces compatibility to react 19, it brings some transitive dependencies that do not support react 19 yet. We should aim to update or replace those as well to avoid the warnings. This will also unblock downstream projects using npm. Because npm does not silence those incompatibilities with warnings like yarn does above.

This is rather tricky. These dependencies are all quite old and seems like they are unlikely to get react v19 support. See [react-table](https://www.npmjs.com/package/react-table) for example, the latest version is the current version used in react-components. It seems we can use the [peerDependenciesMeta](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta) field to bypass the npm error checks. This seems like a less aggressive option than replacing dependencies, finding whole new frameworks to use. wdyt?

edlerd:
> This is rather tricky. These dependencies are all quite old and seems like they are unlikely to get react v19 support. See [react-table](https://www.npmjs.com/package/react-table) for example, the latest version is the current version used in react-components. It seems we can use the [peerDependenciesMeta](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta) field to bypass the npm error checks. This seems like a less aggressive option than replacing dependencies, finding whole new frameworks to use. wdyt?

Sounds like a good option to me, thanks for the suggestion.

mas-who:
> > This is rather tricky. These dependencies are all quite old and seems like they are unlikely to get react v19 support. See [react-table](https://www.npmjs.com/package/react-table) for example, the latest version is the current version used in react-components. It seems we can use the [peerDependenciesMeta](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta) field to bypass the npm error checks. This seems like a less aggressive option than replacing dependencies, finding whole new frameworks to use. wdyt?
> 
> Sounds like a good option to me, thanks for the suggestion.

I will do some testing with admin-ui locally as well. Will also look around to see if there is a way to not leak the peer dependency errors for these packages to a downstream project altogether.

edlerd:
> > This is rather tricky. These dependencies are all quite old and seems like they are unlikely to get react v19 support. See [react-table](https://www.npmjs.com/package/react-table) for example, the latest version is the current version used in react-components. It seems we can use the [peerDependenciesMeta](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta) field to bypass the npm error checks. This seems like a less aggressive option than replacing dependencies, finding whole new frameworks to use. wdyt?
> 
> Sounds like a good option to me, thanks for the suggestion.

Actually not sure if the peerdependenciesmeta field helps, because the docs mention to use it to mark peer dependency optional, but it doesn't mention version conflict resolution.

mas-who:
> > > This is rather tricky. These dependencies are all quite old and seems like they are unlikely to get react v19 support. See [react-table](https://www.npmjs.com/package/react-table) for example, the latest version is the current version used in react-components. It seems we can use the [peerDependenciesMeta](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta) field to bypass the npm error checks. This seems like a less aggressive option than replacing dependencies, finding whole new frameworks to use. wdyt?
> > 
> > 
> > Sounds like a good option to me, thanks for the suggestion.
> 
> Actually not sure if the peerdependenciesmeta field helps, because the docs mention to use it to mark peer dependency optional, but it doesn't mention version conflict resolution.

I think what this field does is that it would inform npm that don't enforce the peer dependencies. So npm does not error for those warnings. I will test that theory with admin-ui locally first though

edlerd (approved):
LGTM, thanks for the upstream resolution.

PR #1097, opened 2025-02-10
feat: [WD-19015] CMS Server Config for maas.machine

Kxiru:
## Done

- Introduced MaasMachineSettingFormInput.tsx

- Maas Machine Setting.
- Storage.*volume settings.
- .*_address settings.

## QA

1. Run the LXD-UI:
    - [PENDING]

## Screenshots

image
image

### DIFF
--- a/src/api/server.tsx
+++ b/src/api/server.tsx
@@ -1,27 +1,95 @@
-import { handleResponse, handleTextResponse } from "util/helpers";
-import type { LxdSettings } from "types/server";
+import {
+  constructMemberError,
+  handleResponse,
+  handleTextResponse,
+} from "util/helpers";
+import type { LXDSettingOnClusterMember, LxdSettings } from "types/server";
 import type { LxdApiResponse } from "types/apiResponse";
 import type { LxdMetadata, LxdConfigPair } from "types/config";
 import type { LxdResources } from "types/resources";
+import { LxdClusterMember } from "types/cluster";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 
-export const fetchSettings = (): Promise<LxdSettings> => {
+export const fetchSettings = (target?: string): Promise<LxdSettings> => {
   return new Promise((resolve, reject) => {
-    fetch("/1.0")
+    const targetQueryParam = target ? `?target=${target}` : "";
+    fetch(`/1.0${targetQueryParam}`)
       .then(handleResponse)
       .then((data: LxdApiResponse<LxdSettings>) => resolve(data.metadata))
       .catch(reject);
   });
 };
 
-export const updateSettings = (config: LxdConfigPair): Promise<void> => {
+export const fetchSettingsFromClusterMembers = (
+  clusterMembers: LxdClusterMember[],
+): Promise<LXDSettingOnClusterMember[]> => {
   return new Promise((resolve, reject) => {
-    fetch("/1.0", {
+    Promise.allSettled(
+      clusterMembers.map((member) => {
+        return fetchSettings(member.server_name);
+      }),
+    )
+      .then((results) => {
+        const settingOnMembers: LXDSettingOnClusterMember[] = [];
+        for (let i = 0; i < clusterMembers.length; i++) {
+          const memberName = clusterMembers[i].server_name;
+          const result = results[i];
+          if (result.status === "rejected") {
+            reject(constructMemberError(result, memberName));
+          }
+          if (result.status === "fulfilled") {
+            const promise = results[
+              i
+            ] as PromiseFulfilledResult<LXDSettingOnClusterMember>;
+            settingOnMembers.push({ ...promise.value, memberName: memberName });
+          }
+        }
+        resolve(settingOnMembers);
+      })
+      .catch(reject);
+  });
+};
+
+export const updateSettings = (
+  config: LxdConfigPair,
+  target?: string,
+): Promise<void> => {
+  const targetQueryParam = target ? `?target=${target}` : "";
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0${targetQueryParam}`, {
       method: "PATCH",
       body: JSON.stringify({
         config,
       }),
     })
       .then(handleResponse)
+      .then(resolve)
+      .catch(reject);
+  });
+};
+
+export const updateClusteredSettings = (
+  clusterValues: ClusterSpecificValues,
+  configName: string,
+): Promise<void> => {
+  return new Promise((resolve, reject) => {
+    Promise.allSettled(
+      Object.keys(clusterValues).map((memberName) => {
+        const config = {
+          [configName]: clusterValues[memberName],
+        };
+        return updateSettings(config, memberName);
+      }),
+    )
+      .then((results) => {
+        const error = results.find((res) => res.status === "rejected")
+          ?.reason as Error | undefined;
+
+        if (error) {
+          reject(error);
+          return;
+        }
+      })
       .then(resolve)
       .catch(reject);
   });
--- a/src/context/useSettings.tsx
+++ b/src/context/useSettings.tsx
@@ -1,12 +1,25 @@
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
-import { fetchSettings } from "api/server";
-import type { LxdSettings } from "types/server";
+import { fetchSettingsFromClusterMembers, fetchSettings } from "api/server";
+import type { LXDSettingOnClusterMember, LxdSettings } from "types/server";
 import { UseQueryResult } from "@tanstack/react-query";
+import { useClusterMembers } from "./useClusterMembers";
 
 export const useSettings = (): UseQueryResult<LxdSettings> => {
   return useQuery({
     queryKey: [queryKeys.settings],
-    queryFn: fetchSettings,
+    queryFn: () => fetchSettings(),
+  });
+};
+
+export const useClusteredSettings = (): UseQueryResult<
+  LXDSettingOnClusterMember[]
+> => {
+  const { data: clusterMembers = [] } = useClusterMembers();
+
+  return useQuery({
+    queryKey: [queryKeys.settings, queryKeys.cluster],
+    queryFn: () => fetchSettingsFromClusterMembers(clusterMembers),
+    enabled: clusterMembers.length > 0,
   });
 };
--- a/src/pages/settings/ClusteredSettingFormInput.tsx
+++ b/src/pages/settings/ClusteredSettingFormInput.tsx
@@ -1,0 +1,95 @@
+import { FC, useState } from "react";
+import { Button, Form, Icon } from "@canonical/react-components";
+import type { ConfigField } from "types/config";
+import { getConfigId } from "./SettingForm";
+import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
+import ClusterSpecificInput from "components/forms/ClusterSpecificInput";
+import { useClusterMembers } from "context/useClusterMembers";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
+
+interface Props {
+  initialValue: ClusterSpecificValues;
+  configField: ConfigField;
+  onSubmit: (newValue: ClusterSpecificValues) => void;
+  onCancel: () => void;
+  disableReason?: string;
+  readonly?: boolean;
+  toggleReadOnly?: () => void;
+}
+
+const ClusteredSettingFormInput: FC<Props> = ({
+  disableReason,
+  initialValue,
+  configField,
+  onSubmit,
+  onCancel,
+  readonly = false,
+  toggleReadOnly,
+}) => {
+  const [value, setValue] = useState<ClusterSpecificValues>(initialValue);
+
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const memberNames = clusterMembers.map((member) => member.server_name);
+
+  const canBeReset = Object.values(value).some(
+    (v) => v !== configField.default,
+  );
+
+  const resetToDefault = () => {
+    const defaultValues: { [key: string]: string } = {};
+    memberNames.forEach((name) => {
+      defaultValues[name] = configField.default;
+    });
+    setValue(defaultValues);
+  };
+
+  return (
+    <Form
+      onSubmit={(e) => {
+        e.preventDefault();
+        onSubmit(value);
+      }}
+    >
+      <ClusterSpecificInput
+        aria-label={configField.key}
+        disableReason={disableReason}
+        id={getConfigId(configField.key)}
+        values={value}
+        isReadOnly={readonly}
+        onChange={(value) => setValue(value)}
+        memberNames={memberNames}
+        toggleReadOnly={toggleReadOnly}
+        helpText={
+          <ConfigFieldDescription
+            description={configField.longdesc}
+            className="p-form-help-text"
+          />
+        }
+      />
+      {!readonly && (
+        <>
+          <Button appearance="base" onClick={onCancel}>
+            Cancel
+          </Button>
+          <Button appearance="positive" type="submit">
+            Save
+          </Button>
+          {canBeReset && (
+            <Button
+              className="reset-button"
+              appearance="base"
+              type="button"
+              onClick={resetToDefault}
+              hasIcon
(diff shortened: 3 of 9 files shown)


src/api/server.tsx
```diff
@@ -27,6 +64,33 @@ export const updateSettings = (config: LxdConfigPair): Promise<void> => {
   });
 };
 
+export const updateClusteredSettings = (
+  config: LxdConfigPair,
+  clusterMembers: LxdClusterMember[],
```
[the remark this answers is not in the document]
Kxiru: Whilst this function has not yet been fully fleshed out, it is my understanding that config will - at this stage - contain the object of key value pairs for each member.
But this is the focus of my next iteration so I will take these suggestions on board in any case.

src/pages/settings/MaasMachineSettingFormInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+interface Props {
+  initialValue: ClusterSpecificValues;
+  configField: ConfigField;
+  onSubmit: (newValue: ClusterSpecificValues) => void;
+  onCancel: () => void;
+  readonly?: boolean;
+}
+
+const MaasMachineSettingFormInput: FC<Props> = ({
```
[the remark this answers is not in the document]
Kxiru: I think you might be right :). I'll do that.

src/api/server.tsx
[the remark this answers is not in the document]
edlerd: Yes! I think you don't need key-value-pairs, because we are only changing on key at a time. So we can pass a key and  need member-value pairs. Because the value can be different per member. For this, we can use the existing `ClusterSpecificValues`, as it already represents member-value pairs.

src/api/server.tsx
```diff
@@ -27,6 +68,32 @@ export const updateSettings = (config: LxdConfigPair): Promise<void> => {
...
 };
 
+export const updateClusteredSettings = (
+  config: ClusterSpecificValues,
+  configName: string,
+): Promise<void> => {
+  return new Promise((resolve, reject) => {
+    Promise.allSettled(
+      Object.keys(config).map((memberName) => {
+        const memberNetwork = {
```
edlerd, line 78:
```suggestion
        const config = {
```
Better name? I'd rename the incoming parameter in the root of the function `clusterValues` and this one will be the `config`.

src/pages/settings/SettingForm.tsx
```diff
@@ -19,33 +23,50 @@ export const getConfigId = (key: string) => {
...
   );
 
-  const onSubmit = (newValue: string | boolean) => {
-    const config = {
-      [configField.key]: String(newValue),
-    };
-    updateSettings(config)
+  const onSubmit = (newValue: string | boolean | ClusterSpecificValues) => {
+    const isNotClustered =
+      typeof newValue !== "string" && typeof newValue !== "boolean";
```
edlerd, line 60: We can rely on `isClusteredInput` below directly. That variable is already defined above and will be more stable than defining a custom condition here based on the type of the value.

src/pages/settings/SettingForm.tsx
```diff
@@ -133,9 +164,19 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
                   : "You do not have permission to edit server configuration"
               }
             >
-              <div className="readmode-value u-truncate">
-                {getReadModeValue()}
-              </div>
+              {isClustered && isMaasMachine ? (
```
edlerd, line 167:
```suggestion
              {isClusteredInput ? (
```
We should simplify and generalize here.

src/pages/settings/SettingForm.tsx
```diff
@@ -133,9 +164,19 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
...
+                <ClusteredSettingFormInput
```
edlerd, line 168:
```suggestion
                <ClusteredSettingFormInput
                  key={JSON.stringify(clusteredValue)}
```
Problem with the read not working initially is, that the `clusteredValue` is updated on load from empty object to an object with values. But this component keeps an internal state as the `value` and doesn't reflect updates to the `clusteredValue`. A workaround is to use the json version of `clusteredValue` as a key, to enforce the component to re-initialize. This has the desired side effect of updating the internal state and displaying the correct values.

src/pages/settings/SettingForm.tsx
edlerd, line 1: When in read mode, this renders the `ClusteredSettingFormInput` inside a button, which leads to some wired layouting. Probably want to change that to have the button next to the values in that case and not wrap the values.

edlerd (commented):
Thanks for the iteration! Some hints how to get this finished and more generalized below.

src/pages/settings/SettingForm.tsx
Kxiru: Addressed this, although now both elements are in block display, I'll address this in my next push though let me know if you have any suggestions on how to approach this.

Kxiru:
@edlerd , Line 64 has a lint error for the use of Stirng(), however this cannot just be replaced to JSON.stringify because it then means that all the non-clustered values that take string input end up having their speech marks included in the Value field when in non-edit mode, and their default setting becomes ' "" ' instead of '-'. I would suggest trying this locally and testing with maas.api.url.
Do you have any ideas for how to get around this lint error?

src/context/useSettings.tsx
```diff
@@ -1,12 +1,23 @@
...
+import { fetchSettingsFromClusterMembers, fetchSettings } from "api/server";
+import type { LXDSettingOnClusterMember, LxdSettings } from "types/server";
 import { UseQueryResult } from "@tanstack/react-query";
+import { LxdClusterMember } from "types/cluster";
 
 export const useSettings = (): UseQueryResult<LxdSettings> => {
   return useQuery({
     queryKey: [queryKeys.settings],
-    queryFn: fetchSettings,
+    queryFn: () => fetchSettings(),
```
edlerd, line 11: This can stay as simple as it was before.

src/context/useSettings.tsx
```diff
@@ -1,12 +1,23 @@
...
+  });
+};
+
+export const useClusteredSettings = (
+  memberNames: LxdClusterMember[],
```
edlerd, line 16:
```suggestion
```
I think this should not get any input, we can move the `useClusterMembers` call into this function to increase cohesion as we don't need the members in the caller.

edlerd, line 16:
```suggestion
```
I think this should not get any input, we can move the `useClusterMembers` call into this function to increase cohesion as we don't need the members in the caller.

src/pages/settings/ClusteredSettingFormInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+  onCancel,
+  readonly = false,
+}) => {
+  const [value, setValue] = useState<ClusterSpecificValues>(initialValue);
+
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const memberNames = clusterMembers.map((member) => member.server_name);
+
+  const canBeReset =
+    JSON.stringify(configField.default) !== JSON.stringify(value);
```
edlerd, line 31: This seems wrong. We should compare the value per member with the default. Not the strigified object of all member to value pairs.

src/pages/settings/ClusteredSettingFormInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const resetToDefault = () => {
+    setValue({});
```
edlerd, line 34: This seems wrong, we should set an empty string for all members instead.

src/pages/settings/ClusteredSettingFormInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+import ClusterSpecificInput from "components/forms/ClusterSpecificInput";
+import { useClusterMembers } from "context/useClusterMembers";
+import { ClusterSpecificValues } from "components/ClusterSpecificSelect";
+
+interface Props {
+  initialValue: ClusterSpecificValues;
+  configField: ConfigField;
+  onSubmit: (newValue: ClusterSpecificValues) => void;
+  onCancel: () => void;
+  readonly?: boolean;
```
edlerd, line 15: We should add support for `toggleReadOnly` here, so get the nicely aligned edit buttons.

src/pages/settings/SettingForm.tsx
```diff
@@ -119,25 +148,39 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
...
+            <div>
+              {isClusteredInput && (
+                <ClusteredSettingFormInput
+                  key={JSON.stringify(clusteredValue)}
+                  initialValue={clusteredValue ?? {}}
+                  configField={configField}
+                  onSubmit={onSubmit}
+                  onCancel={onCancel}
+                  readonly={true}
+                />
```
edlerd, line 160: I think we should not be calling the component twice in here, with hardcoded readonly. I imagine we can rely on the `toggleReadOnly` of the `ClusterSpecificInput`, then we would avoid the edit button in here in case of clustered input, as that component will come with its own edit buttons.

This should also solve the layouting issue currently at hand.

edlerd (commented):
Good progress, the API layer seems all right. Some suggestions to fix the layout in read mode and to simplify a bit more.

src/context/useSettings.tsx
Kxiru: As fetch settings' input parameters have changed to now optionally include a target string, I believe the function now has to be invoked.

src/pages/settings/ClusteredSettingFormInput.tsx
Kxiru: Understandable. Previously, the JSON.stringify was String(), but this raises linting errors.

src/pages/settings/ClusteredSettingFormInput.tsx
Kxiru: The edit buttons I believe you are referring to (the first of the two pictures below) do not actually allow one to edit the values. I disabled these specifically on line 56 because they were impacting the UI as below. 
image

src/pages/settings/SettingForm.tsx
Kxiru: I think I see what you are trying to achieve here, however, the only previous use of toggleReadOnly in the codebase uses Formik, but formik is not used in the settings. Is there an ensureEditMode(formik) equivalent?

Kxiru: > I think we can rely on the edit buttons in the ClusteredInput, but we need to hook it to the right logic. You can pass a `toggleReadOnly` to it and it should be forwarded from `ClusteredSettingFormInput`, then we can let it do the same thing the edit button in `SettingForm` is doing. Basically calling `setEditMode(true)` to drive a local state. This state we can directly forward to `ClusteredSettingFormInput`.

Okay, let me mess around with this.

Kxiru: When using toggleReadOnly, we end up with this:
image

src/pages/settings/ClusteredSettingFormInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <Form
+      onSubmit={(e) => {
+        e.preventDefault();
+        onSubmit(value);
+      }}
+    >
+      <ClusterSpecificInput
+        aria-label={configField.key}
+        classname="input-wrapper"
```
edlerd, line 53:
```suggestion
```
I think this is unused.

src/pages/settings/ClusteredSettingFormInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+          </Button>
+          <Button appearance="positive" type="submit">
+            Save
+          </Button>
+          {canBeReset && (
+            <Button
+              className="reset-button"
+              appearance="base"
+              type="button"
+              onClick={() => resetToDefault(memberNames)}
```
edlerd, line 80:
```suggestion
              onClick={resetToDefault}
```
We don't need the argument and can use the variable in the parent scope directly from within the function.

src/pages/settings/ClusteredSettingFormInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+    (v) => v !== configField.default,
+  );
+
+  const resetToDefault = (memberNames: string[]) => {
+    const defaultValues: { [key: string]: string } = {};
+    memberNames.forEach((name) => {
+      defaultValues[name] = configField.default;
+    });
+    setValue(defaultValues);
+  };
```
edlerd, line 42:
```suggestion
  const resetToDefault = () => {
    const defaultValues: { [key: string]: string } = {};
    memberNames.forEach((name) => {
      defaultValues[name] = configField.default;
    });
    setValue(defaultValues);
  };
```
I think we can simplify this a bit.

src/pages/settings/SettingForm.tsx
```diff
@@ -19,33 +23,51 @@ export const getConfigId = (key: string) => {
...
     <ResourceLabel bold type="setting" value={configField.key} />
   );
 
-  const onSubmit = (newValue: string | boolean) => {
-    const config = {
-      [configField.key]: String(newValue),
-    };
-    updateSettings(config)
+  const onSubmit = (newValue: string | boolean | ClusterSpecificValues) => {
+    const mutationPromise =
```
edlerd, line 62:
```suggestion
    const mutation =
```
IMHO a cleaner name. Though we are not very consistent in other places with naming those.

src/pages/settings/SettingForm.tsx
```diff
@@ -119,25 +160,27 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
...
+                    ? ""
+                    : "You do not have permission to edit server configuration"
+                }
+              >
+                <div className="readmode-value u-truncate">
+                  {getReadModeValue()}
+                </div>
+                <Icon name="edit" className="edit-icon" />
+              </Button>
+            </div>
```
edlerd, line 183: I don't think we need the extra `div` wrapper.

src/pages/settings/Settings.tsx
```diff
@@ -52,7 +56,7 @@ const Settings: FC = () => {
   const getValue = (configField: ConfigField): string | undefined => {
     for (const [key, value] of Object.entries(settings?.config ?? {})) {
       if (key === configField.key) {
-        return value;
+        return value as string;
```
edlerd, line 59: I think casting here should not be necessary.

edlerd (commented):
Code and QA work nicely, thanks for the changes. Some suggestions to simplify below.

Ideally we make the empty read only state, when no value is set also display "-", to stay consistent with other parts of the app.

edlerd (approved):
LGTM, good work :+1:

src/pages/settings/ClusteredSettingFormInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    <Form
+      onSubmit={(e) => {
+        e.preventDefault();
+        onSubmit(value);
+      }}
+    >
+      <ClusterSpecificInput
```
mas-who, line 51: @Kxiru @edlerd I think one more change here will be needed. We should disable the input if user does not have permission to edit server settings. See how `canEditServerConfiguration` is used in `SettingForm`, we can do the same here using the `disabled` prop and maybe expose a prop for `disabledReason`?

edlerd: Yes, please. I forgot we already merged the permissions for server settings.

src/pages/settings/SettingForm.tsx
```diff
@@ -81,6 +106,28 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
     }
   }, [isEditMode]);
 
+  if (isClusteredInput) {
+    return (
+      <ClusteredSettingFormInput
+        key={JSON.stringify(clusteredValue)}
+        initialValue={clusteredValue ?? {}}
+        buttonTitle={
```
mas-who, line 114: I think the additional prop should rather be `disabledReason` instead of `buttonTitle` since it relates to a button element that is few levels down the component tree

Kxiru: Yep! I hadn't rebased at this point. Please see my next push.

PR #1099, opened 2025-02-11
feat: restricted permissions for server actions [WD-18907]

mas-who:
## Done

- Restricted user from editing server settings if they are missing required permissions

## QA

1. Run the LXD-UI:
    - Assign identity with `can_edit` entitlement on server, check that the identity can update server settings.
    - Remove the above permission from the user auth group, ensure edits are disabled for server settings

### DIFF
--- a/src/pages/settings/SettingForm.tsx
+++ b/src/pages/settings/SettingForm.tsx
@@ -28,7 +28,7 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
   const notify = useNotify();
   const toastNotify = useToastNotification();
   const queryClient = useQueryClient();
-  const { canEditServer } = useServerEntitlements();
+  const { canEditServerConfiguration } = useServerEntitlements();
 
   const editRef = useRef<HTMLDivElement | null>(null);
 
@@ -126,11 +126,11 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
                 setEditMode(true);
               }}
               hasIcon
-              disabled={!canEditServer()}
+              disabled={!canEditServerConfiguration()}
               title={
-                canEditServer()
+                canEditServerConfiguration()
                   ? ""
-                  : "You do not have permission to edit the server"
+                  : "You do not have permission to edit server configuration"
               }
             >
               <div className="readmode-value u-truncate">
--- a/src/util/entitlements/server.tsx
+++ b/src/util/entitlements/server.tsx
@@ -4,10 +4,11 @@ import { hasEntitlement } from "./helpers";
 export const useServerEntitlements = () => {
   const { isFineGrained, serverEntitlements } = useAuth();
 
-  const canEditServer = () =>
-    hasEntitlement(isFineGrained, "can_edit", serverEntitlements);
+  const canEditServerConfiguration = () =>
+    hasEntitlement(isFineGrained, "can_edit", serverEntitlements) ||
+    hasEntitlement(isFineGrained, "admin", serverEntitlements);
 
   return {
-    canEditServer,
+    canEditServerConfiguration,
   };
 };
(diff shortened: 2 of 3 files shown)


src/pages/settings/SettingForm.tsx
```diff
@@ -124,6 +126,12 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
                 setEditMode(true);
               }}
               hasIcon
+              disabled={!canEditServer()}
+              title={
+                canEditServer()
+                  ? ""
+                  : "You do not have permission to edit the server"
```
edlerd, line 133:
```suggestion
                  : "You do not have permission to edit server configuration"
```

src/util/entitlements/server.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { hasEntitlement } from "./helpers";
+
+export const useServerEntitlements = () => {
+  const { isFineGrained, serverEntitlements } = useAuth();
+
+  const canEditServer = () =>
+    hasEntitlement(isFineGrained, "can_edit", serverEntitlements);
+
+  return {
+    canEditServer,
```
edlerd, line 11:
```suggestion
    canEditServerConfiguration,
```
Maybe a bit better?

edlerd (approved):
LGTM, good work :+1:

edlerd (changes requested):
QA issue: I have a user with server admin but they can't edit the server actions.

mas-who:
> QA issue: I have a user with server admin but they can't edit the server configs.

Aaah I think I know why, we can't have entitlement query param for the `/1.0` endpoint. Which means we will need to check for all possible entitlements, will sort that out

mas-who:
> > QA issue: I have a user with server admin but they can't edit the server configs.
> 
> Aaah I think I know why, we can't have entitlement query param for the `/1.0` endpoint. Which means we will need to check for all possible entitlements, will sort that out

@edlerd thanks for catching this, we will need to take this into consideration for other server related permission restrictions as well

mas-who:
> QA issue: When the user has no permission to view server config, the 1.0 response will have config: null, see screen below
> 
> image
> 
> In this case the UI currently displays all the default server settings, even though that might be wrong. In my tests even the "server can_view" does not allow loading the server config. Only "server admin" allowed that. This seems a bit odd. Maybe we should disable the settings page if that permission is not granted entirely.

I don't think there is `can_view` for `server`, do you mean `can_edit`? afaik, A user would need either `server can_edit` or `server admin` to see and modify settings. Agreed on disabling the settings page, although should we disable the navigation settings link, or completely hide it?

edlerd:
> I don't think there is `can_view` for `server`, do you mean `can_edit`? afaik, A user would need either `server can_edit` or `server admin` to see and modify settings. Agreed on disabling the settings page, although should we disable the navigation settings link, or completely hide it?

I meant "server viewer", which maybe should allow showing server config? Might be a bug.

As for presentation, I think we should keep the navigation link active. The settings page can display a warning that the settings are not available due to permissions. Similar to what we do with `hasMetadataConfiguration` already.

mas-who:
> > I don't think there is `can_view` for `server`, do you mean `can_edit`? afaik, A user would need either `server can_edit` or `server admin` to see and modify settings. Agreed on disabling the settings page, although should we disable the navigation settings link, or completely hide it?
> 
> I meant "server viewer", which maybe should allow showing server config? Might be a bug.
> 
> As for presentation, I think we should keep the navigation link active. The settings page can display a warning that the settings are not available due to permissions. Similar to what we do with `hasMetadataConfiguration` already.

Oh interesting, according to the docs it says "Grants access to view all resources in the LXD server.". Maybe resources does not include configs? That doesn't make a whole lot of sense though, also documented it in the sheet but will bring it up with the core team.

edlerd (approved):
QA and code LGTM

PR #1100, opened 2025-02-11
feat: restricted permissions for image actions [WD-18905]

mas-who:
## Done

- handling of image related actions documented [here](https://docs.google.com/spreadsheets/d/1fXBNG9B8CL2iXP6kPEe4a3YM0q2QcOnj-I3essIUuh8/edit?gid=1540595032#gid=1540595032)

## QA

1. Run the LXD-UI:
    - Check that the UI handles images related permission restrictions according to actions documented [here](https://docs.google.com/spreadsheets/d/1fXBNG9B8CL2iXP6kPEe4a3YM0q2QcOnj-I3essIUuh8/edit?gid=1540595032#gid=1540595032)

### DIFF
--- a/src/api/images.tsx
+++ b/src/api/images.tsx
@@ -11,13 +11,20 @@ import { EventQueue } from "context/eventQueue";
 import type { LxdInstance } from "types/instance";
 import type { UploadState } from "types/storage";
 import axios, { AxiosResponse } from "axios";
+import { withEntitlementsQuery } from "util/entitlements/api";
 
-export const fetchImageList = (project?: string): Promise<LxdImage[]> => {
+const imageEntitlements = ["can_delete"];
+
+export const fetchImageList = (
+  isFineGrained: boolean | null,
+  project?: string,
+): Promise<LxdImage[]> => {
+  const entitlements = `&${withEntitlementsQuery(isFineGrained, imageEntitlements)}`;
   const url =
     "/1.0/images?recursion=1" +
     (project ? `&project=${project}` : "&all-projects=1");
   return new Promise((resolve, reject) => {
-    fetch(url)
+    fetch(url + entitlements)
       .then(handleResponse)
       .then((data: LxdApiResponse<LxdImage[]>) => resolve(data.metadata))
       .catch(reject);
--- a/src/api/images.tsx
+++ b/src/api/images.tsx
@@ -15,16 +15,25 @@ import { withEntitlementsQuery } from "util/entitlements/api";
 
 const imageEntitlements = ["can_delete"];
 
-export const fetchImageList = (
+export const fetchImagesInProject = (
+  project: string,
   isFineGrained: boolean | null,
-  project?: string,
 ): Promise<LxdImage[]> => {
   const entitlements = `&${withEntitlementsQuery(isFineGrained, imageEntitlements)}`;
-  const url =
-    "/1.0/images?recursion=1" +
-    (project ? `&project=${project}` : "&all-projects=1");
   return new Promise((resolve, reject) => {
-    fetch(url + entitlements)
+    fetch(`/1.0/images?recursion=1&project=${project}${entitlements}`)
+      .then(handleResponse)
+      .then((data: LxdApiResponse<LxdImage[]>) => resolve(data.metadata))
+      .catch(reject);
+  });
+};
+
+export const fetchImagesInAllProjects = (
+  isFineGrained: boolean | null,
+): Promise<LxdImage[]> => {
+  const entitlements = `&${withEntitlementsQuery(isFineGrained, imageEntitlements)}`;
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/images?recursion=1&all-projects=1${entitlements}`)
       .then(handleResponse)
       .then((data: LxdApiResponse<LxdImage[]>) => resolve(data.metadata))
       .catch(reject);
--- a/src/context/auth.tsx
+++ b/src/context/auth.tsx
@@ -2,7 +2,6 @@ import { createContext, FC, ReactNode, useContext } from "react";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { fetchCertificates } from "api/certificates";
-import { useSettings } from "context/useSettings";
 import { fetchProjects } from "api/projects";
 import { fetchCurrentIdentity } from "api/auth-identities";
 import { useSupportedFeatures } from "./useSupportedFeatures";
@@ -36,12 +35,10 @@ interface ProviderProps {
 }
 
 export const AuthProvider: FC<ProviderProps> = ({ children }) => {
-  const { data: settings, isLoading } = useSettings();
-
-  const { hasEntitiesWithEntitlements, isSettingsLoading } =
+  const { hasEntitiesWithEntitlements, isSettingsLoading, settings } =
     useSupportedFeatures();
 
-  const { data: currentIdentity } = useQuery({
+  const { data: currentIdentity, isLoading: isIdentityLoading } = useQuery({
     queryKey: [queryKeys.currentIdentity],
     queryFn: fetchCurrentIdentity,
     retry: false, // avoid retry for older versions of lxd less than 5.21 due to missing endpoint
@@ -91,7 +88,8 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
       value={{
         isAuthenticated: (settings && settings.auth !== "untrusted") ?? false,
         isOidc: settings?.auth_user_method === "oidc",
-        isAuthLoading: isLoading,
+        isAuthLoading:
+          isSettingsLoading || isIdentityLoading || isProjectsLoading,
         isRestricted,
         defaultProject,
         hasNoProjects: projects.length === 0 && !isProjectsLoading,
--- a/src/context/auth.tsx
+++ b/src/context/auth.tsx
@@ -41,10 +41,26 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
   const { hasEntitiesWithEntitlements, isSettingsLoading } =
     useSupportedFeatures();
 
+  const { data: currentIdentity } = useQuery({
+    queryKey: [queryKeys.currentIdentity],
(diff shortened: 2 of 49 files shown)


src/pages/images/ImageList.tsx
```diff
@@ -124,7 +119,8 @@ const ImageList: FC = () => {
 
     return {
       key: image.fingerprint,
-      name: image.fingerprint,
+      // disable image selection if user does not have entitlement to delete
+      name: canDeleteImage(image) ? image.fingerprint : "",
```
edlerd, line 123: can we use `deletableImages` here?

src/util/entitlements/images.tsx
```diff
@@ -0,0 +1,18 @@
+import { useAuth } from "context/auth";
+import { hasEntitlement } from "./helpers";
+import { LxdImage } from "types/image";
+
+export const useImageEntitlements = (image?: LxdImage) => {
+  const { isFineGrained } = useAuth();
+
+  const canDeleteImage = (imageOverride?: LxdImage) =>
```
edlerd, line 8: The `imageOverride` seems like an inconsistent approach, better have one way to feed the image in, not a 2nd one.

src/util/entitlements/projects.tsx
```diff
@@ -0,0 +1,38 @@
+import { useAuth } from "context/auth";
+import { hasEntitlement } from "./helpers";
+import { LxdProject } from "types/project";
+import { useProject } from "context/project";
+
+export const useProjectEntitlements = (project?: LxdProject) => {
+  const { isFineGrained } = useAuth();
+  const { project: currentProject } = useProject();
+
+  const validProject = project || currentProject;
```
edlerd, line 10: We should not have this other way of fetching the current project in here. Better keep the interface consistent and expect a project to be passed in. Callers then would need to fetch the current project.

src/api/images.tsx
```diff
@@ -11,13 +11,20 @@ import { EventQueue } from "context/eventQueue";
...
 import type { UploadState } from "types/storage";
 import axios, { AxiosResponse } from "axios";
+import { withEntitlementsQuery } from "util/entitlements/api";
 
-export const fetchImageList = (project?: string): Promise<LxdImage[]> => {
+const imageEntitlements = ["can_delete"];
+
+export const fetchImageList = (
+  isFineGrained: boolean | null,
+  project?: string,
```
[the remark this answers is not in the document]
mas-who: if project was missing previously, we add the `all-projects=1` query param to the url. If we make project mandatory, I don't think "default" represents this case well. Maybe project can be `string | null`?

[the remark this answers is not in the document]
edlerd: I see, that is a bit confusing. Maybe better to have two functions in that case to make the interface clear. One `fetchImagesInProject(project: string, ...)` and one `fetchImagesInAllProjects(...)`. Wdyt?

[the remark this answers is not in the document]
mas-who: yeah that sounds good! Then the corresponding hooks can be `useImagesInProject` and `useImagesInAllProjects`?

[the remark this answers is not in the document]
edlerd: sounds all right.

src/context/useProjects.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const useProjects = (): UseQueryResult<LxdProject[]> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.projects],
+    queryFn: () => fetchProjects(isFineGrained),
+    enabled: isFineGrained !== null,
+  });
+};
+
+export const useProjectFetch = (
```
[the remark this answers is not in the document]
mas-who: that's great :+1:

src/util/entitlements/images.tsx
mas-who: This is related to this [discussion](https://github.com/canonical/lxd-ui/pull/1094#discussion_r1950420370). The problem is that we need to have a way to check entitlements as we are iterating over a list of resources e.g. for bulk action permission checks etc. We can do it with the single hook input, because a hook can only be invoked at the top level of a component function. In the instance permissions PR, I've implemented a set function which serves the same purpose, the hook takes in a list of resources then generates a non-restricted resource set for lookup. I personally prefer the set method, not sure if you have an alternative approach?

edlerd: I think we can use `deletableImages` in the one use case that is passing in an image override.

edlerd: Oh wait that `deletableImages` also passes an override internally.

edlerd: Maybe after all, it is better to always pass the entity to the `canFoo` functions and make the `useFooEntitlements` hook agnostic of an entity.

mas-who: yeah I think that's better too, we didn't quite consider the bulk case in the first implementation PR

edlerd: Thanks, I think it is the better approach as it works in all cases, and we can remove inconsistent shortcuts.

src/pages/images/actions/UploadImageBtn.tsx
```diff
@@ -12,6 +13,7 @@ interface Props {
 const UploadImageBtn: FC<Props> = ({ project }) => {
   const { openPortal, closePortal, isOpen, Portal } = usePortal();
   const isSmallScreen = useSmallScreen();
+  const { project: currentProject } = useCurrentProject();
```
edlerd, line 16: I think we shouldn't have both the project and currentProject in this component. Now we have two competing project definitions. Better `useProject` here to fetch the project that is passed in via prop as a project name. Maybe rename the prop to "projectName".

edlerd (commented):
Thx for making the entitlement lookups consistent. Only two questions below left, then this should be good to go.

src/pages/images/actions/UploadImageBtn.tsx
mas-who: good point :+1:

src/context/auth.tsx
```diff
@@ -41,7 +41,7 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
 
   const { data: projects = [], isLoading: isProjectsLoading } = useQuery({
     queryKey: [queryKeys.projects],
-    queryFn: fetchProjects,
+    queryFn: () => fetchProjects(false),
```
edlerd, line 44: Hard-coding false seems to be wrong. Shouldn't we move this down and rely on the current identity here?

edlerd (changes requested):

src/context/auth.tsx
mas-who: right, we can just use current identity here to determine if we need to do a fine-grained fetch :+1:

src/context/auth.tsx
```diff
@@ -39,9 +39,25 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
...
+    }
+    return false;
+  };
+
   const { data: projects = [], isLoading: isProjectsLoading } = useQuery({
     queryKey: [queryKeys.projects],
-    queryFn: fetchProjects,
+    queryFn: () => fetchProjects(isFineGrained()),
     enabled: settings?.auth === "trusted",
   });
```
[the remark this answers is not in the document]
mas-who: aah sorry, was just about to push this logic

edlerd (approved):
code LGTM

mas-who:
> code LGTM

Thanks, are we not going to do QA?

mas-who:
> An open issue: Given a user has "project default image_manager", but not "project default viewer", then we fail to identify if the user has the image creation permission.

I believe if the user does not have `project default viewer` they will not be able to see anything in the UI for project default. This is related to this [task](https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/930?assignee=712020%3A312a6456-2dab-4385-a58a-61e033ad9fff&selectedIssue=WD-18837).

edlerd:
> I believe if the user does not have `project default viewer` they will not be able to see anything in the UI for project default. This is related to this [task](https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/930?assignee=712020%3A312a6456-2dab-4385-a58a-61e033ad9fff&selectedIssue=WD-18837).

Interestingly, the image list was visible without the project viewer permission. But the upload image button was disabled, because we fail to figure out the user has that permission.

edlerd:
> I think that's probably because in `App.tsx`, `ImageList` is not wrapped with `ProjectLoader`. It's interesting though, we can fetch images with the `all-project` query param, so it can be a non-project specific page. Currently, to upload an image, it is dependent on the current project that we have selected. Perhaps the upload image form should have a project selector, that way the list of project options would be governed by `project <project> can_view` permission. wdyt? Should it be part of this PR?

I wouldn't change it now. I am also not sure if it is the right thing to do. Images are project specific, after all.

The project viewer permission is mandatory to have image permissions correctly applied. Maybe mention it as an issue in the doc you created. I am fine with keeping things as they are now.

edlerd (approved):
Apart from the issue mentioned above, QA LGTM as well.

mas-who:
> > I think that's probably because in `App.tsx`, `ImageList` is not wrapped with `ProjectLoader`. It's interesting though, we can fetch images with the `all-project` query param, so it can be a non-project specific page. Currently, to upload an image, it is dependent on the current project that we have selected. Perhaps the upload image form should have a project selector, that way the list of project options would be governed by `project <project> can_view` permission. wdyt? Should it be part of this PR?
> 
> I wouldn't change it now. I am also not sure if it is the right thing to do. Images are project specific, after all.
> 
> The project viewer permission is mandatory to have image permissions correctly applied. Maybe mention it as an issue in the doc you created. I am fine with keeping things as they are now.

Okay sure, when it comes to permissions, I think we may run into a few places where the current UI structure may need to be updated/changed/improved. I've documented the above issue in the [sheet](https://docs.google.com/spreadsheets/d/1fXBNG9B8CL2iXP6kPEe4a3YM0q2QcOnj-I3essIUuh8/edit?gid=1540595032#gid=1540595032) for reference.

edlerd (approved):
LGTM :+1:

PR #1104, opened 2025-02-12
fix: fix network failing e2e tests

mas-who:
## Done

-  fix network failing e2e tests
- Instead of checking floating notifications, we check if the network exists on the list page after creation. This circumvents the problem of page reloading due to ERR_NETWORK_CHANGED.

## QA

1. Run the LXD-UI:
    - Check that CI passes

### DIFF
--- a/tests/helpers/network.ts
+++ b/tests/helpers/network.ts
@@ -3,6 +3,7 @@ import { randomNameSuffix } from "./name";
 import type { LxdNetworkType } from "types/network";
 import { activateAllTableOverrides } from "./configuration";
 import { gotoURL } from "./navigate";
+import { expect } from "../fixtures/lxd-test";
 
 export const randomNetworkName = (): string => {
   return `test-${randomNameSuffix()}`;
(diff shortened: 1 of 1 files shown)


tests/helpers/network.ts
```diff
@@ -95,3 +95,22 @@ export const createNetworkForward = async (page: Page, network: string) => {
 
   await deleteNetwork(page, network);
 };
+
+export const networkExist = async (
```
edlerd, line 99: This is an interesting workaround. Maybe a better interface is to get the network link and ensure it to be visible / to not be visible in the caller. this gets rid of the visible parameter to the function.

edlerd: Meant to return the `networkLink`.

mas-who: yeah it feels a bit hacky, but the problem is that the page reloads in CI causes notification state to get wiped causing teh tests to fail. will update the interface for the function as suggested

tests/helpers/network.ts
```diff
@@ -24,8 +25,9 @@ export const createNetwork = async (
     await page.getByLabel("Parent").selectOption({ index: 1 });
   }
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForSelector(`text=Network ${network} created.`);
-  await page.getByRole("button", { name: "Close notification" }).click();
+  const networkLink = await getNetworkLink(page, network);
+  await expect(networkLink).toBeVisible();
+  await expect(networkLink).toHaveText(network);
```
edlerd, line 30:
```suggestion
```
Isn't this redundant with the way this is requested?

edlerd (dismissed):
LGTM, thanks for finding this solution.

tests/helpers/network.ts
mas-who: removed

edlerd (approved):
LGTM

PR #1107, opened 2025-02-13
Ensure the size is unset, when selecting a storage pool driver without it, fix network form validation

### LINKED ISSUE #1101: Storage pool creation issue
Hey,

Tried creating a storage pool with the btrfs driver. It failed because the filesystem is ext4 not btrfs, so I switched to "Directory" instead. But creation failed because the UI still remembers the `Size` field I set to 5GB from the btrfs options.

It should reset all fields if you switch Driver to prevent this from happening. Clearing them myself did not work, I had to actually refresh the page.

Image

edlerd:
## Done

- Ensure the size is unset, when selecting a storage pool driver without it
- Simplify and unify the storage pool field reset on driver changes, decouple this from the api layer.
- Do not rely on formik for validating an uplink is selected for ovn networks. there is a custom validation in place.

Fixes #1101

## QA

1. Run the LXD-UI:
    - create a storage pool, configure some fields like size, then switch to another driver that has no size and ensure it gets created fine. Especially test in cluster and non-cluster backends.
    - create a network of type bridge, enter name, select type ovn, then select type bridge. Ensure the create button is active.

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -114,11 +114,8 @@ export const createClusteredPool = (
           config: {
             ...memberPoolPayload.config,
             source: sourcePerClusterMember?.[item.server_name],
-
             size: sizePerClusterMember?.[item.server_name],
-            ...(zfsPoolNamePerClusterMember?.[item.server_name] && {
-              "zfs.pool_name": zfsPoolNamePerClusterMember[item.server_name],
-            }),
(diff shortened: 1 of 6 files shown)


mas-who:
code and QA LGTM for cluster and non-cluster envs :+1: thanks for the improvements

mas-who (approved):

PR #1109, opened 2025-02-13
Ensure to fetch instance metrics from the right cluster member

### LINKED ISSUE #1102: resource usage view doesn't work properly in a cluster
Hey,

Couldn't wait for 6.3 and wanted to see what the new resource usage graphs looks like : )

Viewing the UI from `lxd1`, the screenshot speaks for itself, resource usage is only shown on the locally running instances:

Image

edlerd:
## Done

- Ensure to fetch instance metrics from the right cluster member
- Add permission handling of the metrics query

Fixes #1102 

## QA

1. Run the LXD-UI:
    - run a clustered backend with instances running on different members
    - ensure the cpu and disk metrics in the instance list and instance detail page show properly
    - test the same on a non-clustered backend

### DIFF
--- a/src/api/metrics.tsx
+++ b/src/api/metrics.tsx
@@ -2,6 +2,8 @@ import parsePrometheusTextFormat from "parse-prometheus-text-format";
 import type { LxdMetricGroup } from "types/metrics";
 
 export const fetchMetrics = (target: string): Promise<LxdMetricGroup[]> => {
+  // in a simple and non-clustered environment, the LXD api responds
+  // with instance.location as "none". Handle it, to avoid sending an invalid target
   const params = target === "none" ? "" : `?target=${target}`;
 
   return new Promise((resolve, reject) => {
--- a/src/api/metrics.tsx
+++ b/src/api/metrics.tsx
@@ -1,9 +1,11 @@
 import parsePrometheusTextFormat from "parse-prometheus-text-format";
 import type { LxdMetricGroup } from "types/metrics";
 
-export const fetchMetrics = (): Promise<LxdMetricGroup[]> => {
+export const fetchMetrics = (target: string): Promise<LxdMetricGroup[]> => {
+  const params = target === "none" ? "" : `?target=${target}`;
+
   return new Promise((resolve, reject) => {
-    fetch("/1.0/metrics")
+    fetch(`/1.0/metrics${params}`)
       .then((response) => {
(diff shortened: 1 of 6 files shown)


src/context/useMetrics.tsx
```diff
@@ -0,0 +1,18 @@
+import { useQuery, UseQueryResult } from "@tanstack/react-query";
+import { useAuth } from "./auth";
+import { queryKeys } from "util/queryKeys";
+import { fetchMetrics } from "api/metrics";
+import { LxdMetricGroup } from "types/metrics";
+
+export const useMetrics = (
+  location: string,
+): UseQueryResult<LxdMetricGroup[]> => {
+  const { isRestricted } = useAuth();
```
mas-who, line 10: we can use `useServerEntitlements` here to determine if the user has relevant permissions to view metrics i.e. `server admin` || `server can_view_metrics` || `server viewer`.

Shouldn't we also disable the fetch if `isFineGrained` is null?

mas-who:
QA looks good :+1:

src/context/useMetrics.tsx
edlerd: This PR is aimed to fix cluster case. To make sure metrics are loaded correctly in a clustered environment, because that was broken before. I didn't aim to resolve the new requirements for permission just yet. Even though this hook is already half of the way done into that direction.

edlerd: Thinking of this, probably best to solve the permission issue in this PR directly. It should be fairly easy, as you suggested. Will update it shortly.

mas-who: Aaah didn't see the second comment, yeah that makes sense, I guess it's easier to deal with it now while the context is there

src/api/metrics.tsx
```diff
@@ -1,9 +1,11 @@
 import parsePrometheusTextFormat from "parse-prometheus-text-format";
 import type { LxdMetricGroup } from "types/metrics";
 
-export const fetchMetrics = (): Promise<LxdMetricGroup[]> => {
+export const fetchMetrics = (target: string): Promise<LxdMetricGroup[]> => {
```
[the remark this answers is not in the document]
mas-who: oh I see, all good then

mas-who (approved):
LGTM thanks for the fix :+1:

PR #1114, opened 2025-02-18
feat: restricted permissions for networks [WD-18903]

mas-who:
## Done

- Graceful handling of networks in the UI
- covered actions documented in the [permissions sheet](https://docs.google.com/spreadsheets/d/1fXBNG9B8CL2iXP6kPEe4a3YM0q2QcOnj-I3essIUuh8/edit?gid=1505273394#gid=1505273394)

## QA

1. Run the LXD-UI:
    - Check through the list of actions documented [here](https://docs.google.com/spreadsheets/d/1fXBNG9B8CL2iXP6kPEe4a3YM0q2QcOnj-I3essIUuh8/edit?gid=1505273394#gid=1505273394)

### DIFF
--- a/src/api/networks.tsx
+++ b/src/api/networks.tsx
@@ -12,14 +12,46 @@ import type { LxdApiResponse } from "types/apiResponse";
 import { areNetworksEqual } from "util/networks";
 import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 import type { LxdClusterMember } from "types/cluster";
+import { withEntitlementsQuery } from "util/entitlements/api";
+
+const networkEntitlements = ["can_edit", "can_delete"];
+
+export const fetchNetworksOnMember = (
+  project: string,
+  target: string,
+  isFineGrained: boolean | null,
+): Promise<LxdNetwork[]> => {
+  const entitlements = `&${withEntitlementsQuery(
+    isFineGrained,
+    networkEntitlements,
+  )}`;
+  return new Promise((resolve, reject) => {
+    fetch(
+      `/1.0/networks?project=${project}&recursion=1&target=${target}${entitlements}`,
+    )
+      .then(handleResponse)
+      .then((data: LxdApiResponse<LxdNetwork[]>) => {
+        const filteredNetworks = data.metadata.filter(
+          // Filter out loopback and unknown networks, both are not useful for the user.
+          // this is in line with the filtering done in the LXD CLI
+          (network) => !["loopback", "unknown"].includes(network.type),
+        );
+        resolve(filteredNetworks);
+      })
+      .catch(reject);
+  });
+};
 
 export const fetchNetworks = (
   project: string,
-  target?: string,
+  isFineGrained: boolean | null,
 ): Promise<LxdNetwork[]> => {
-  const targetParam = target ? `&target=${target}` : "";
+  const entitlements = `&${withEntitlementsQuery(
+    isFineGrained,
+    networkEntitlements,
+  )}`;
   return new Promise((resolve, reject) => {
-    fetch(`/1.0/networks?project=${project}&recursion=1${targetParam}`)
+    fetch(`/1.0/networks?project=${project}&recursion=1${entitlements}`)
       .then(handleResponse)
       .then((data: LxdApiResponse<LxdNetwork[]>) => {
         const filteredNetworks = data.metadata.filter(
@@ -36,11 +68,16 @@ export const fetchNetworks = (
 export const fetchNetworksFromClusterMembers = (
   project: string,
   clusterMembers: LxdClusterMember[],
+  isFineGrained: boolean | null,
 ): Promise<LXDNetworkOnClusterMember[]> => {
   return new Promise((resolve, reject) => {
     Promise.allSettled(
       clusterMembers.map((member) => {
-        return fetchNetworks(project, member.server_name);
+        return fetchNetworksOnMember(
+          project,
+          member.server_name,
+          isFineGrained,
+        );
       }),
     )
       .then((results) => {
@@ -63,14 +100,37 @@ export const fetchNetworksFromClusterMembers = (
   });
 };
 
+export const fetchNetworkOnMember = (
+  name: string,
+  project: string,
+  target: string,
+  isFineGrained: boolean | null,
+): Promise<LxdNetwork> => {
+  const entitlements = `&${withEntitlementsQuery(
+    isFineGrained,
+    networkEntitlements,
+  )}`;
+  return new Promise((resolve, reject) => {
+    fetch(
+      `/1.0/networks/${name}?project=${project}&target=${target}${entitlements}`,
+    )
+      .then(handleEtagResponse)
+      .then((data) => resolve(data as LxdNetwork))
+      .catch(reject);
+  });
+};
+
 export const fetchNetwork = (
   name: string,
   project: string,
-  target?: string,
+  isFineGrained: boolean | null,
 ): Promise<LxdNetwork> => {
-  const targetParam = target ? `&target=${target}` : "";
+  const entitlements = `&${withEntitlementsQuery(
+    isFineGrained,
+    networkEntitlements,
+  )}`;
   return new Promise((resolve, reject) => {
-    fetch(`/1.0/networks/${name}?project=${project}${targetParam}`)
+    fetch(`/1.0/networks/${name}?project=${project}${entitlements}`)
       .then(handleEtagResponse)
       .then((data) => resolve(data as LxdNetwork))
       .catch(reject);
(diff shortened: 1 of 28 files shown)


src/api/networks.tsx
```diff
@@ -12,14 +12,46 @@ import type { LxdApiResponse } from "types/apiResponse";
...
+        const filteredNetworks = data.metadata.filter(
+          // Filter out loopback and unknown networks, both are not useful for the user.
+          // this is in line with the filtering done in the LXD CLI
+          (network) => !["loopback", "unknown"].includes(network.type),
+        );
+        resolve(filteredNetworks);
+      })
+      .catch(reject);
+  });
+};
```
edlerd, line 43: Why duplicate this? I think it's fine to keep this as part of the `fetchNetworks`. It is also what we do on other entities API layer.

edlerd (commented):
Reviewed some of the files, some preliminary comments below.

src/api/networks.tsx
edlerd: I see the usenetworks gets more complex due to the split, because we have to dynamically build a query key. I think a combined approach would benefit simplicity there and avoid duplication here.

mas-who: sure that's a valid point, I think then let's have `isFineGrained` as the first parameter to `fetchNetworks` in this case wdyt?

mas-who: the same was done in the storage-pools PR btw, once we confirm the change then I will apply it there too

src/pages/networks/NetworkForwards.tsx
```diff
@@ -89,13 +93,27 @@ const NetworkForwards: FC<Props> = ({ network, project }) => {
...
+                    `/ui/project/${project}/network/${network.name}/forwards/${forward.listen_address}/edit`,
+                  )
+                }
+                type="button"
+                title={
+                  canEditNetwork(network)
+                    ? "Edit network forward"
+                    : "You do not have permission to edit forwards for this network"
+                }
+                disabled={!canEditNetwork(network)}
```
edlerd, line 113: This should stay a link so we can open it in a new tab and it behaves like a native link. Better switch it out for a disabled button in the read only case.

src/pages/networks/NetworkForwards.tsx
```diff
@@ -123,12 +141,23 @@ const NetworkForwards: FC<Props> = ({ network, project }) => {
...
+        title={
+          canEditNetwork(network)
+            ? ""
+            : "You do not have permission to create network forwards for this network"
+        }
       >
-        Create forward
-      </Link>
+        <span>Create forward</span>
+      </Button>
```
edlerd, line 160: We should keep the link semantics here as well.

src/pages/networks/NetworkList.tsx
```diff
@@ -80,8 +80,9 @@ const NetworkList: FC = () => {
     isLoading: isClusterNetworksLoading,
   } = useQuery({
     queryKey: [queryKeys.networks, project, queryKeys.cluster],
-    queryFn: () => fetchNetworksFromClusterMembers(project, clusterMembers),
-    enabled: clusterMembers.length > 0,
+    queryFn: () =>
+      fetchNetworksFromClusterMembers(project, clusterMembers, isFineGrained),
+    enabled: clusterMembers.length > 0 && isFineGrained !== null,
```
edlerd, line 85: I think this should be part of the hook as well. Then we can contain the `isFineGrained` call in it.

src/pages/networks/actions/DeleteNetworkForwardBtn.tsx
```diff
@@ -46,7 +48,11 @@ const DeleteNetworkForwardBtn: FC<Props> = ({ network, forward, project }) => {
   return (
     <ConfirmationButton
       appearance="base"
-      onHoverText="Delete network forward"
+      onHoverText={
+        canDeleteNetwork(network)
+          ? "Delete network forward"
+          : "You do not have permission to delete this forwards for this network"
```
edlerd, line 54:
```suggestion
          : "You do not have permission to delete this network forward"
```

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -321,6 +321,8 @@ const NetworkForm: FC<Props> = ({
                 ensureEditMode(formik);
                 void formik.setFieldValue("yaml", yaml);
               }}
+              readOnly={!!formik.values.editRestriction}
+              readOnlyMessage={{ value: formik.values.editRestriction ?? "" }}
```
edlerd, line 325: It is a bit ugly to leak building of this object to all use cases that want to set a read only message. Can we change the prop to receive an optional string and build the object inside the `YamlForm` component instead? I think that will better contain the logic.

src/pages/networks/forms/NetworkParentSelector.tsx
```diff
@@ -57,8 +57,9 @@ const NetworkParentSelector: FC<Props> = ({ props, formik, isClustered }) => {
     isLoading: isClusterNetworksLoading,
   } = useQuery({
     queryKey: [queryKeys.networks, "default", queryKeys.cluster],
-    queryFn: () => fetchNetworksFromClusterMembers("default", clusterMembers),
-    enabled: clusterMembers.length > 0,
+    queryFn: () =>
+      fetchNetworksFromClusterMembers("default", clusterMembers, isFineGrained),
+    enabled: clusterMembers.length > 0 && isFineGrained !== null,
```
edlerd, line 62: Should use a hook.

edlerd (commented):
Finished the review just now.

src/pages/networks/forms/NetworkForm.tsx
mas-who: sounds good :+1:

src/context/useNetworks.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return useQuery({
+    queryKey,
+    queryFn: () => fetchNetwork(network, project, isFineGrained, target),
+    enabled: (enabled ?? true) && isFineGrained !== null,
+  });
+};
+
+export const useNetworksFromClusterMembers = (
+  project: string,
+  clusterMembers: LxdClusterMember[],
```
edlerd, line 57: Shall we remove the cluster members from the interface here (and in the function below) and call the `useClusterMembers` in here directly? It seems simpler to do that and also to have the enable/disable logic for the query inside here.

src/context/useNetworks.tsx
```diff
@@ -0,0 +1,95 @@
...
+    queryFn: () =>
+      fetchNetworksFromClusterMembers(project, clusterMembers, isFineGrained),
+    enabled: (enabled ?? true) && isFineGrained !== null,
+  });
+};
+
+export const useNetworkFromClusterMembers = (
+  network: string,
+  project: string,
+  clusterMembers: LxdClusterMember[],
```
edlerd, line 73:
```suggestion
```
Can use the same approach as the comment above.

src/api/networks.tsx
```diff
@@ -174,7 +193,11 @@ export const createNetwork = (
         // when creating a network on localhost the request will get cancelled
         // check manually if creation was successful
         if (e.message === "Failed to fetch") {
-          const newNetwork = await fetchNetwork(network.name ?? "", project);
+          const newNetwork = await fetchNetwork(
+            network.name ?? "",
+            project,
+            false,
```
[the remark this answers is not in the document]
mas-who: oh nice, sorry I should have picked that up

src/context/useNetworks.tsx
mas-who: yeah that sounds pretty neat, although looking at where these hooks are currently being called, it seems like `clusterMembers` are used elsewhere in the components as well. It feels like we are making redundant queries, but the interface does become cleaner. Do you think it's worth the change?

edlerd: I think it is fine as the queries should be fetched only once due to caching.

src/pages/networks/actions/DeleteNetworkForwardBtn.tsx
```diff
@@ -21,6 +22,7 @@ const DeleteNetworkForwardBtn: FC<Props> = ({ network, forward, project }) => {
   const toastNotify = useToastNotification();
   const queryClient = useQueryClient();
   const [isLoading, setLoading] = useState(false);
+  const { canDeleteNetwork } = useNetworkEntitlements();
```
edlerd, line 25: this is not the right permission, i think this should check canEditNetwork

edlerd (commented):
Code LGTM, just one QA issue, see comment below.

I found a QA issue with instances: When browsing snapshots in default projects instance. User has no no create instance permission in default project, but has the permission in project foo. Then the "Create instance from snapshot" button is correctly enabled and in the modal we have only one option as target project: "foo". So far all correct. But on submit the instance is created in the default project. I think the form is not initialized with the correct target project value.

mas-who:
> Code LGTM, just one QA issue, see comment below.
> 
> I found a QA issue with instances: When browsing snapshots in default projects instance. User has no no create instance permission in default project, but has the permission in project foo. Then the "Create instance from snapshot" button is correctly enabled and in the modal we have only one option as target project: "foo". So far all correct. But on submit the instance is created in the default project. I think the form is not initialized with the correct target project value.

@edlerd I have created a separate [PR](https://github.com/canonical/lxd-ui/pull/1117) to fix this as it is not related to the current context. Thanks for the catch on this one, this is a great spot! :clap:

edlerd (approved):
LGTM, good stuff :)

PR #1123, opened 2025-02-24
feat: generalise bulk delete button across UI pages [WD-19546]

mas-who:
## Done

- Create a generalised bulk delete button component to standardise usage across UI pages

## QA

1. Run the LXD-UI:
    - Test Bulk delete instances, images, instance and volume snapshots

### DIFF
--- a/src/components/BulkDeleteButton.tsx
+++ b/src/components/BulkDeleteButton.tsx
@@ -1,0 +1,87 @@
+import { FC } from "react";
+import { pluralize } from "util/instanceBulkActions";
+import {
+  ConfirmationButton,
+  ConfirmationButtonProps,
+  Icon,
+} from "@canonical/react-components";
+import classnames from "classnames";
+
+interface Props {
+  entities: unknown[];
+  deletableEntities: unknown[];
+  entityType: string;
+  onDelete: () => void;
+  disabledReason?: string;
+  bulkDeleteBreakdown?: string[];
+  confirmationButtonProps?: Partial<ConfirmationButtonProps>;
+  buttonLabel?: React.ReactNode;
+  className?: string;
+}
+
+const BulkDeleteButton: FC<Props> = ({
+  entities,
+  deletableEntities,
+  disabledReason,
+  entityType,
+  bulkDeleteBreakdown,
+  confirmationButtonProps,
+  onDelete,
+  buttonLabel = "Delete",
+  className,
+}) => {
+  const totalCount = entities.length;
+  const deleteCount = deletableEntities.length;
+
+  const breakdown =
+    bulkDeleteBreakdown?.map((action) => {
+      return (
+        <li key={action} className="p-list__item">
+          - {action}
+        </li>
+      );
+    }) || [];
+
+  const modalContent = (
+    <>
+      {breakdown.length > 0 && (
+        <>
+          <p>
+            <b>{totalCount}</b> {pluralize(entityType, entities.length)}{" "}
+            selected:
+          </p>
+          <ul className="p-list">{breakdown}</ul>
+        </>
+      )}
+      <p className="u-no-padding--top">
+        This will permanently delete <b>{deleteCount}</b>{" "}
+        {pluralize(entityType, deleteCount)}.{"\n"}This action cannot be undone,
+        and can result in data loss.
+      </p>
+    </>
+  );
+
+  return (
+    <ConfirmationButton
+      className={classnames("has-icon", className)}
+      onHoverText={
+        disabledReason ?? `Delete ${pluralize(entityType, entities.length)}`
+      }
+      disabled={deleteCount === 0 || !!disabledReason}
+      shiftClickEnabled
+      showShiftClickHint
+      {...confirmationButtonProps}
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: modalContent,
+        confirmButtonLabel: "Delete",
+        onConfirm: onDelete,
+      }}
+    >
+      <Icon name="delete" />
+      <span>{buttonLabel}</span>
+    </ConfirmationButton>
+  );
+};
+
+export default BulkDeleteButton;
--- a/src/pages/images/actions/BulkDeleteImageBtn.tsx
+++ b/src/pages/images/actions/BulkDeleteImageBtn.tsx
@@ -2,21 +2,23 @@ import { FC, useState } from "react";
 import { deleteImageBulk } from "api/images";
 import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
-import { ConfirmationButton } from "@canonical/react-components";
 import { useEventQueue } from "context/eventQueue";
 import { getPromiseSettledCounts } from "util/helpers";
 import { pluralize } from "util/instanceBulkActions";
 import { useToastNotification } from "context/toastNotificationProvider";
+import BulkDeleteButton from "components/BulkDeleteButton";
+import { useImageEntitlements } from "util/entitlements/images";
+import { LxdImage } from "types/image";
 
 interface Props {
-  fingerprints: string[];
+  images: LxdImage[];
   project: string;
   onStart: () => void;
   onFinish: () => void;
 }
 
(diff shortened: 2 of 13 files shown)


src/components/BulkDeleteButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  confirmationButtonProps?: Partial<ConfirmationButtonProps>;
+  buttonLabel?: React.ReactNode;
+}
+
+const BulkDeleteButton: FC<Props> = ({
+  entities,
+  deletableEntities,
+  disabledReason,
+  entityType,
+  extraActions,
```
edlerd, line 25: Extra action seems like the wrong naming as this is used to communicate items that are not going to receive any action. Maybe better call this `actionSummary` or `bulkActionBreakdown`?

src/components/BulkDeleteButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+        children: (
+          <p>
+            {actions.length > 0 && (
+              <>
+                <b>{totalCount}</b> {pluralize(entityType, entities.length)}{" "}
+                selected:
+                <br />
+                <br />
+                {actions}
+                <br />
```
edlerd, line 65: Instead of those extra line breaks, we should probably remove the wrapping paragraph and make the sections here individual paragraphs, so equal spacing applies.

src/components/BulkDeleteButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+      }}
+    >
+      {buttonLabel ? (
+        buttonLabel
+      ) : (
+        <>
+          <Icon name="delete" />
+          <span>Delete</span>
+        </>
+      )}
```
edlerd, line 84: I think 
```suggestion
          <Icon name="delete" />
          <span>{buttonLabel}</span>
```
I think this should always render the icon and have the `buttonLabel` default to "Delete", but allow override for i.e. writing "Delete snapshot" or "Delete image". I'd suggest removing the loading states in the two snapshot cases to make the button label a string.

src/components/BulkDeleteButton.tsx
edlerd, line 1: Shouldn't this be used for the permission entities as well?

edlerd (commented):
Good stuff, some comments below to improve the generic component.

src/components/BulkDeleteButton.tsx
mas-who: Aah yes, I should have marked this PR as draft, was waiting for the permissions [PR](https://github.com/canonical/lxd-ui/pull/1113) to get merged so I won't have to deal with a ton of merge conflicts. Will apply this component for permission entities too

src/components/BulkDeleteButton.tsx
mas-who: yes I was struggling with this prop naming a bit, I think the suggestions are great, maybe a bit more specific for this component? e.g. `bulkdDeleteBreakdown`?

src/components/BulkDeleteButton.tsx
mas-who: I think probably better to also convert the breakdown into a `ul` wdyt?

Screenshot from 2025-02-25 08-21-35

src/components/BulkDeleteButton.tsx
mas-who: I have applied the new `BulkDeleteButton` to all permission entities except groups. The structure of the delete group modal is quite different. However, I have adopted the same bulk delete warning approach, let me know what you think:

User does not have permission to delete any of the selected groups (button disabled):
Screenshot from 2025-02-25 09-31-34

User has permission to delete all selected groups:
Screenshot from 2025-02-25 09-32-49

User selected only one group and can delete the goup:
Screenshot from 2025-02-25 09-34-04

User selected multiple groups and can delete only one group:
Screenshot from 2025-02-25 09-35-35

User selected multiple groups and can delete more than one but not all groups:
Screenshot from 2025-02-25 09-36-54

src/components/BulkDeleteButton.tsx
edlerd: Yes, that is even better.

src/components/BulkDeleteButton.tsx
edlerd: Yeah, this looks all right :+1:

src/components/BulkDeleteButton.tsx
edlerd: sounds good to me.

src/pages/instances/actions/InstanceBulkDelete.tsx
```diff
@@ -79,90 +79,53 @@ const InstanceBulkDelete: FC<Props> = ({ instances, onStart, onFinish }) => {
     });
   };
 
-  const getStoppedInstances = () => {
-    if (!deleteCount) {
-      return null;
+  const getBulkDeleteBreakdown = () => {
+    if (!(ignoredCount + restrictedCount)) {
```
edlerd, line 83:
```suggestion
    if (ignoredCount + restrictedCount === 0) {
```
This seems easier to read.

src/pages/permissions/actions/BulkDeleteIdentitiesBtn.tsx
```diff
@@ -1,27 +1,19 @@
...
 interface Props {
   identities: LxdIdentity[];
-  className?: string;
 }
 
-const BulkDeleteIdentitiesBtn: FC<Props & ButtonProps> = ({
-  identities,
-  className,
-}) => {
+const BulkDeleteIdentitiesBtn: FC<Props & ButtonProps> = ({ identities }) => {
```
edlerd, line 16:
```suggestion
const BulkDeleteIdentitiesBtn: FC<Props> = ({ identities }) => {
```
Seems to be ignored.

src/pages/images/actions/BulkDeleteImageBtn.tsx
```diff
@@ -70,32 +77,37 @@ const BulkDeleteImageBtn: FC<Props> = ({
...
 
+  const getBulkDeleteBreakdown = () => {
+    if (deleteCount === totalCount) {
+      return undefined;
+    }
+
+    const restrictedCount = totalCount - deleteCount;
+    return [
+      `${restrictedCount} ${pluralize("image", restrictedCount)} that you do not have permission to delete will be ignored.`,
+    ];
```
edlerd, line 88: This should include a recap of "this amount will be deleted", like in the instance case.

Same for the other cases. I think the breakdown should always include the "this amount will be handled" case as a first item.

src/pages/permissions/actions/BulkDeleteIdentitiesBtn.tsx
```diff
@@ -61,51 +53,35 @@ const BulkDeleteIdentitiesBtn: FC<Props & ButtonProps> = ({
...
-            ) : null}
-            This action cannot be undone, and can result in data loss.
-          </p>
-        ),
-        confirmButtonLabel: "Delete",
-        onConfirm: handleDelete,
+      className="u-no-margin--bottom"
+      confirmationButtonProps={{
+        loading: isLoading,
+        "aria-label": "Delete identities",
```
edlerd, line 80:
```suggestion
```
We probably don't need this as there is already a label on the button.

src/pages/permissions/actions/DeleteGroupModal.tsx
edlerd, line 1: The self modification warning should not be displayed, if a group with myself is selected, but it will be ignored because I do not have permission to delete it.

edlerd (commented):
QA seems mostly all right, just the breakdowns can be tweaked a bit and some minor code quality suggestions below.

src/pages/images/actions/BulkDeleteImageBtn.tsx
mas-who: will apply through out

src/pages/permissions/actions/DeleteGroupModal.tsx
mas-who: makes sense, done :+1:

edlerd (approved):
QA and code LGTM :+1:

PR #1124, opened 2025-02-25
feat: [WD-18263] CMS Storage pool usage

Kxiru:
## Done
- Displaying multiple pools depending on Cluster

## QA

1. Run the LXD-UI:
    - Review the Storage Pool list page on a clustered and non-clustered environment.

## Screenshots

image

### DIFF
--- a/src/pages/storage/StoragePoolClusterMember.tsx
+++ b/src/pages/storage/StoragePoolClusterMember.tsx
@@ -1,0 +1,38 @@
+import ResourceLink from "components/ResourceLink";
+import { useClusterMembers } from "context/useClusterMembers";
+import type { FC } from "react";
+import type { LxdStoragePool } from "types/storage";
+import { hasPoolMemberSpecificSize } from "util/storagePool";
+
+interface Props {
+  pool: LxdStoragePool;
+}
+
+export const StoragePoolClusterMember: FC<Props> = ({ pool }) => {
+  const hasMemberSpecificSize = hasPoolMemberSpecificSize(pool.driver);
+  const { data: clusterMembers = [] } = useClusterMembers();
+
+  return (
+    <div>
+      {hasMemberSpecificSize ? (
+        clusterMembers.map((member) => {
+          return (
+            <div className="clustered-resource-link" key={member.server_name}>
+              <ResourceLink
+                to="/ui/cluster"
+                type="cluster-member"
+                value={member.server_name}
+              />
+            </div>
+          );
+        })
+      ) : (
+        <ResourceLink
+          to="/ui/cluster"
+          type="cluster-group"
+          value="Cluster wide"
+        />
+      )}
+    </div>
+  );
+};
--- a/src/pages/storage/StoragePoolOverview.tsx
+++ b/src/pages/storage/StoragePoolOverview.tsx
@@ -6,6 +6,10 @@ import StorageUsedBy from "pages/storage/StorageUsedBy";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import useEventListener from "util/useEventListener";
 import type { LxdStoragePool } from "types/storage";
+import { StoragePoolClusterMember } from "./StoragePoolClusterMember";
+import { isClusteredServer } from "util/settings";
+import { useSettings } from "context/useSettings";
+import { hasPoolMemberSpecificSize } from "util/storagePool";
 
 interface Props {
   pool: LxdStoragePool;
@@ -18,6 +22,9 @@ const StoragePoolOverview: FC<Props> = ({ pool, project }) => {
   };
   useEffect(updateContentHeight, [project, pool]);
   useEventListener("resize", updateContentHeight);
+  const { data: settings } = useSettings();
+  const isClustered = isClusteredServer(settings);
+  const hasMemberSpecificSize = hasPoolMemberSpecificSize(pool.driver);
 
   return (
     <div className="storage-overview-tab">
@@ -38,7 +45,10 @@ const StoragePoolOverview: FC<Props> = ({ pool, project }) => {
               </tr>
               <tr>
                 <th className="u-text--muted">Size</th>
-                <td>
+                <td className="cluster-member-with-meters">
+                  {hasMemberSpecificSize && isClustered && (
+                    <StoragePoolClusterMember pool={pool} />
+                  )}
                   <StoragePoolSize pool={pool} hasMeterBar />
                 </td>
               </tr>
--- a/src/pages/storage/StoragePoolSelectTable.tsx
+++ b/src/pages/storage/StoragePoolSelectTable.tsx
@@ -4,6 +4,9 @@ import ScrollableTable from "components/ScrollableTable";
 import StoragePoolSize from "pages/storage/StoragePoolSize";
 import classnames from "classnames";
 import { useStoragePools } from "context/useStoragePools";
+import { StoragePoolClusterMember } from "./StoragePoolClusterMember";
+import { isClusteredServer } from "util/settings";
+import { useSettings } from "context/useSettings";
 
 interface Props {
   onSelect: (pool: string) => void;
@@ -15,11 +18,14 @@ interface Props {
 
 const StoragePoolSelectTable: FC<Props> = ({ onSelect, disablePool }) => {
   const { data: pools = [], isLoading } = useStoragePools();
+  const { data: settings } = useSettings();
+  const isClustered = isClusteredServer(settings);
 
   const headers = [
     { content: "Name", sortKey: "name" },
     { content: "Driver", sortKey: "driver" },
     { content: "Status", sortKey: "status" },
+    ...(isClustered ? [{ content: "Cluster member" }] : []),
     { content: "Size", className: "size" },
     { "aria-label": "Actions", className: "actions" },
   ];
@@ -63,6 +69,15 @@ const StoragePoolSelectTable: FC<Props> = ({ onSelect, disablePool }) => {
           "aria-label": "Status",
           onClick: selectPool,
         },
+        ...(isClustered
+          ? [
+              {
+                content: <StoragePoolClusterMember pool={pool} />,
+                role: "cell",
+                "aria-label": "Cluster member",
+              },
+            ]
+          : []),
         {
           content: <StoragePoolSize pool={pool} hasMeterBar />,
           role: "cell",
--- a/src/pages/storage/StoragePoolSize.tsx
+++ b/src/pages/storage/StoragePoolSize.tsx
@@ -1,10 +1,14 @@
-import type { FC } from "react";
-import { fetchStoragePoolResources } from "api/storage-pools";
-import { useQuery } from "@tanstack/react-query";
-import { queryKeys } from "util/queryKeys";
+import { Fragment, type FC } from "react";
 import type { LxdStoragePool } from "types/storage";
 import { humanFileSize } from "util/helpers";
 import Meter from "components/Meter";
+import { useClusteredStoragePoolResources } from "context/useStoragePools";
+import { isClusteredServer } from "util/settings";
+import { useSettings } from "context/useSettings";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { fetchStoragePoolResources } from "api/storage-pools";
+import { hasPoolMemberSpecificSize } from "util/storagePool";
 
 interface Props {
   pool: LxdStoragePool;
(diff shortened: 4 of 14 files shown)


edlerd (commented):
I think we can improve the approach for the data loading. See comment below.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,89 @@
...
 interface Props {
   pool: LxdStoragePool;
   hasMeterBar?: boolean;
 }
 
 const StoragePoolSize: FC<Props> = ({ pool, hasMeterBar }) => {
+  const { data: clusteredPoolResources = [] } =
+    useClusteredStoragePoolResources(pool.name);
+  const { data: settings } = useSettings();
+  const sizeSpecificDrivers = ["zfs", "btrfs", "dir", "lvm"];
```
edlerd, line 25: Better keep this sorted, also better to use the constants, so we can find the reference more easily.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,89 @@
...
+  const hasMemberSpecificSize =
+    sizeSpecificDrivers.includes(pool.driver) && isClusteredServer(settings);
+
   const { data: resources } = useQuery({
     queryKey: [queryKeys.storage, pool.name, queryKeys.resources],
     queryFn: () => fetchStoragePoolResources(pool.name),
+    enabled: !hasMemberSpecificSize,
   });
+  const isClustered = isClusteredServer(settings);
```
edlerd, line 34: Better move this line next to the `useSettings` call above, so the loading and usage is closer. It will be easier to understand that way. Introducing and using a variable should ideally be in consecutive lines.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,89 @@
...
+  const poolsToRender = hasMemberSpecificSize
+    ? clusteredPoolResources
+    : [resources];
```
edlerd, line 37:
```suggestion
  const resourceList = hasMemberSpecificSize
    ? clusteredPoolResources
    : [poolResources];
```
Maybe a bit better naming?

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,89 @@
...
+            {isClustered && (
+              <ResourceLink
+                to="/ui/cluster"
+                type={
+                  hasMemberSpecificSize ? "cluster-member" : "cluster-group"
+                }
+                value={
+                  hasMemberSpecificSize
+                    ? (poolResource as LxdClusteredStoragePoolResources)
+                        .memberName
```
edlerd, line 72: Ideally, we shouldn't need any type casting. Maybe a better way is to add the `memberName` as an optional field to the `LxdStoragePoolResources` type, and remove the new `LxdClusteredStoragePoolResources`. Then we can use the optional chaining here and remove the cast. Wdyt?

src/sass/_storage.scss
```diff
@@ -11,6 +11,10 @@
     width: 14rem;
   }
 
+  .size.clustered {
+    width: 18rem;
+  }
```
edlerd, line 16: This should be responsive.

src/pages/storage/StoragePools.tsx
```diff
@@ -86,19 +94,19 @@ const StoragePools: FC = () => {
...
+          role: "cell",
           "aria-label": "Driver",
         },
         {
           content: <StoragePoolSize pool={pool} hasMeterBar />,
-          role: "rowheader",
+          role: "cell",
           "aria-label": "Size",
-          className: "size",
+          className: classNames("size", { clustered: isClustered }),
```
edlerd, line 109: Please adjust the layout, so the cluster names appear like a column in the storage pool list next to the size. It can be all within this cell, but the member label should be placed be next to the meter.

src/pages/storage/StoragePoolSize.tsx
Kxiru: I think that makes sense. Please see my implementation in the latest commit (Pending)

src/components/ResourceLink.tsx
```diff
@@ -11,7 +12,9 @@ interface Props {
 const ResourceLink: FC<Props> = ({ type, value, to }) => {
   return (
     <Link
-      className="p-chip is-inline is-dense resource-link"
+      className={classNames("p-chip is-inline is-dense resource-link", {
+        "storage-pool-table": true,
```
edlerd, line 16: This shouldn't be in the generic component

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,91 @@
...
+        if (!hasMeterBar) {
+          return (
+            <div key={index}>
+              {`${humanFileSize(used)} of ${humanFileSize(total)} used`}
+            </div>
+          );
+        }
+
+        return (
+          <div className="storage-pool-meter-container" key={index}>
```
edlerd, line 65: This class seems unused.

src/sass/_storage.scss
```diff
@@ -83,3 +93,16 @@
...
+  display: flex;
+}
+
+.resource-link {
+  &.storage-pool-table {
+    justify-content: center;
+    min-width: 6rem;
+    width: 6rem;
+  }
+}
```
edlerd, line 108: Those chips should match the content width, the text should not be centred, but left aligned. The space that is reserved for them should be equal. Like there should be some space between the chip and the meter if a name is very short. And the Name chip should go all the way next to the meter if the name is long.

image

src/sass/_storage.scss
```diff
@@ -11,6 +11,10 @@
     width: 14rem;
   }
 
+  .size.clustered {
+    width: 20rem;
+  }
+
```
edlerd, line 17: This distribution of the column widths should be optimized. In the example below, the driver and two volumes columns can give space to allow showing the full content of the member labels.

image

edlerd: This applies also to the detail page. We should allow the member names to grow wider in order not to truncate if the space is available.

image

edlerd: We might want to do some calculation based on the longest member name to find the best max width for them.

src/sass/_storage.scss
edlerd: In the horizontal axis the names of the chips should be aligned with the other columns, they are a bit too low. The meters seem to be okay, but the chip content should be aligned with the meter. I think if the chips are a bit higher, their text will align also in secondary lines within the cell with the meter.

image

src/components/Meter.tsx
```diff
@@ -15,7 +15,7 @@ const Meter: FC<Props> = ({
   hoverText,
 }: Props) => {
   return (
-    <>
+    <div className="p-meter-container">
```
edlerd, line 18: Ideally we avoid those wrapper classes and find a way to style without having to add markup just for layouting.

edlerd (commented):
This looks already very neat. Data loading code LGTM, QA of the data part as well.

The layout needs a bit of refinement, I think there are some improvements we should aim for.

src/pages/storage/StoragePoolSize.tsx
edlerd, line 1: In the nonclustered case the meter is shifted a bit to the right, it should be aligned with the other items. In the detail page and in the overview table.

image

edlerd: In the overview table the left part of the meter should be aligned with the heading.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,84 @@
...
+                type={
+                  hasMemberSpecificSize ? "cluster-member" : "cluster-group"
+                }
+                value={
+                  hasMemberSpecificSize && poolResource.memberName
+                    ? poolResource.memberName
+                    : "Cluster wide"
+                }
+              />
+            )} */}
```
edlerd, line 71:
```suggestion
```

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,84 @@
...
+        if (!hasMeterBar) {
+          return (
+            <div key={index}>
+              {`${humanFileSize(used)} of ${humanFileSize(total)} used`}
+            </div>
+          );
+        }
+
+        return (
+          <div className="storage-pool-meter-container" key={index}>
```
edlerd, line 58:
```suggestion
          <Fragment key={index}>
```
I think this is not needed.

src/sass/_storage.scss
```diff
@@ -83,3 +93,12 @@
     margin-top: -7px;
   }
 }
+
+.storage-pool-meter-container {
+  align-items: center;
+  display: flex;
+}
```
edlerd, line 100:
```suggestion
```
Seems not needed.

src/sass/_storage.scss
```diff
@@ -83,3 +93,12 @@
...
+
+.storage-pool-overview-meter {
+  display: flex;
+}
```
edlerd, line 104:
```suggestion
.cluster-member-with-meters {
  display: flex;
}
```
Better name and move this into the namespace for `storage-overview-tab`. This doesn't need to be on the root namespace.

src/pages/storage/StoragePoolOverview.tsx
```diff
@@ -38,7 +45,10 @@ const StoragePoolOverview: FC<Props> = ({ pool, project }) => {
               </tr>
               <tr>
                 <th className="u-text--muted">Size</th>
-                <td>
+                <td className="storage-pool-overview-meter">
```
edlerd, line 48:
```suggestion
                <td className="cluster-member-with-meters">
```
The styling of this one needs adjustments. The height of the member names should be aligned with the meters and there needs to be some gap between the chips and the meters.

image

src/pages/storage/StoragePoolClusterMember.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <div>
+      {hasMemberSpecificSize ? (
+        clusterMembers.map((member) => {
+          return (
+            <div className="clustered-resource-link" key={member.server_name}>
+              <ResourceLink
+                to="/ui/cluster"
+                type={"cluster-member"}
```
edlerd, line 23:
```suggestion
                type="cluster-member"
```
We should avoid the `{` where not needed. Same applies below in the file.

src/pages/storage/StoragePoolSelectTable.tsx
```diff
@@ -63,6 +69,15 @@ const StoragePoolSelectTable: FC<Props> = ({ onSelect, disablePool }) => {
           "aria-label": "Status",
           onClick: selectPool,
         },
+        ...(isClustered
+          ? [
+              {
+                content: <StoragePoolClusterMember pool={pool} />,
+                role: "rowheader",
```
edlerd, line 76:
```suggestion
                role: "cell",
```

src/components/ResourceLink.tsx
edlerd: This should be reverted.

src/components/Meter.tsx
edlerd: Can we undo the changes here and in the `meter.scss` file? I assume with the two column solution most of it is not needed.

src/sass/_resources.scss
```diff
@@ -24,3 +24,7 @@
     margin-right: $sph--x-small;
   }
 }
+
+.clustered-resource-link {
+  height: 2.25rem;
+}
```
edlerd, line 30: This is only for the `storage-pool-table` and the `migrate-instance-table > storage-pools`. On the storage pool overview page, the height should be adjusted to match its context.

I suggest moving this into the SCSS files for the respective use cases instead of the global size in here.

src/pages/storage/StoragePools.tsx
edlerd, line 1: Maybe hide the usage and cluster member columns on smaller screen sizes?

edlerd (commented):
The table is great. Some ideas for refinement below.

src/pages/storage/StoragePools.tsx
edlerd: Because on small screen sizes, the columns and headings become unreadable. We can hide the columns with css and a media query below a reasonable breakpoint.

src/sass/_resources.scss
Kxiru: Sure thing. As the migrate instance table and storage pool table both use StoragePoolSelectTable as eventual children, I have moved this to storage-pool-table which addresses both.

src/pages/storage/StoragePoolClusterMember.tsx
edlerd: This seems not applied yet.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,70 @@
...
+        const used = poolResource.space.used || 0;
 
-  return (
-    <Meter
-      percentage={(100 / total) * used || 0}
-      text={`${humanFileSize(used)} of ${humanFileSize(total)} used`}
-    />
+        if (!hasMeterBar) {
+          return (
+            <div key={index}>
```
edlerd, line 51:
```suggestion
            <div key={poolResource.memberName}>
```
The key should always be coming from the data. The key is used to decide if a component has to be re-rendered on data change. If the key is not connected to the data, a row might show stale data. Here we have the member name as a unique string in the array available and should use it instead.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -1,38 +1,70 @@
...
+        }
+
+        return (
+          <Fragment key={index}>
+            <Meter
+              key={index}
+              percentage={(100 / total) * used || 0}
+              text={`${humanFileSize(used)} of ${humanFileSize(total)} used`}
+            />
+          </Fragment>
```
edlerd, line 64:
```suggestion
            <Meter
              key={poolResource.memberName}
              percentage={(100 / total) * used || 0}
              text={`${humanFileSize(used)} of ${humanFileSize(total)} used`}
            />
```
Can simplify and use a stable key.

src/sass/_storage.scss
```diff
@@ -22,6 +26,20 @@
   .p-meter {
     margin-top: 4px;
   }
+
+  @media screen and (width < 1200px) {
+    .size.clustered {
+      display: none;
+    }
+
+    .cluster-member-header {
```
edlerd, line 35:
```suggestion
    .cluster-member {
```
I think the `header` suffix is confusing, because in a table that is assumed to be relating to the table header. Maybe we can just remove it and naming works well. Wdyt?

edlerd (commented):
Thanks for the iteration, this is very close to perfect. Three small comments below, one above and the horizontal placement of meters / member names needs a small adjustment. After that this should be ready to merge :+1:

edlerd (commented):
This seems mostly good to merge, just one [previous comment](https://github.com/canonical/lxd-ui/pull/1124/files#r1973938707) left.

edlerd (approved):
QA and Code LGTM :+1: good work!

PR #1128, opened 2025-02-26
Remove redundant void calls

edlerd:
## Done

solved two issues surfaced by the TICS check:

- remove redundant void calls
- fix copy paste error in permissions


Kxiru (commented):
Thanks for being so meticulous in addressing the void function issue(s). What would be the best way to test the copy/paste permissions fix?

edlerd:
> Thanks for being so meticulous in addressing the void function issue(s). What would be the best way to test the copy/paste permissions fix?

the copy paste error was this one: https://github.com/canonical/lxd-ui/pull/1128/commits/b76a48ba82fa39e3768a048ab3d5bd5e0f9d23be#diff-bc94b30448b316e2f795f37eb93fd41b9a33670e82160c55fdc713a85f3a6021R101

I added more ESLint rules (the ones that TICS is checking for and have an automated fix) to this one after creating the PR, you can see the new rules as individual commits.

PR #1129, opened 2025-02-26
feat: [WD-19698] Custom ISO can_delete permission check.

Kxiru:
## Done

- Custom ISO - can delete permission implemented.
- Codebase refactor - Inkeeping with the fact that the there is a disassociation between Storage pools and volumes within the codebase in Context.
  - Split /api/storage-pools into /api/storage-pools and /api/storage-volumes and updated all imports.
  - Consequently, storageVolumes now have their own entitlements file and api file.

## QA

1. Run the LXD-UI:
    - Attempt to delete a Custom ISO volume with a user that has the pemissions to do so.
    - Attempt to delete a Custom ISO volume with a user that does not have the pemissions to do so.
    - Observe the expected behaviour.

## Screenshots
Custom ISOs, one of which the user has permissions to delete, the other, they do not.
image


src/api/storage-volumes.tsx
edlerd, line 1: Good idea to extract those into a dedicated file.

src/api/storage-volumes.tsx
```diff
@@ -0,0 +1,272 @@
...
+export const deleteStorageVolume = (
+  volume: string,
+  pool: string,
+  project: string,
+  isFineGrained: boolean | null,
+): Promise<void> => {
+  const entitlements = withEntitlementsQuery(
+    isFineGrained,
+    storageVolumeEntitlements,
+  );
```
edlerd, line 203: I don't think we should fetch entitlements on deletion.

Kxiru: I'll remove this. Just setting the stage for what's next.

edlerd: We add the entitlements param to the url when fetching entities. If we delete, there is a void return value, so no entitlements needed.

src/pages/storage/actions/DeleteStorageVolumeBtn.tsx
```diff
@@ -103,7 +105,12 @@ const DeleteStorageVolumeBtn: FC<Props> = ({
       className={classname}
       shiftClickEnabled
       showShiftClickHint
-      disabled={Boolean(disabledReason)}
+      disabled={!canDeleteVolume(volume) || Boolean(disabledReason)}
+      title={
```
edlerd, line 109:
```suggestion
      onHoverText={
```
The title prop gets ignored and we should use on hover text instead to override it.

edlerd (changes requested):

src/pages/storage/actions/DeleteStorageVolumeBtn.tsx
```diff
@@ -104,8 +106,12 @@ const DeleteStorageVolumeBtn: FC<Props> = ({
       className={classname}
       shiftClickEnabled
       showShiftClickHint
-      disabled={Boolean(disabledReason)}
-      onHoverText={disabledReason}
+      disabled={!canDeleteVolume(volume) || Boolean(disabledReason)}
+      onHoverText={
+        canDeleteVolume(volume)
+          ? ""
+          : "You do not have permission to delete storage volumes."
```
edlerd, line 113:
```suggestion
          : "You do not have permission to delete this volume."
```
Better be precise, permissions can be granted on a entity level.

edlerd (approved):
QA and Code LGTM, good work :+1:

PR #1130, opened 2025-03-03
fix(events): close ws connection on unmount, catch and retry logic

lorumic:
## Done

Refactored events code:
- Close ws connection on unmount
- Add catch and retry logic for ws connection failures
- Refactored promises code

### DIFF
--- a/src/pages/instances/Events.tsx
+++ b/src/pages/instances/Events.tsx
@@ -6,11 +6,17 @@ import { useAuth } from "context/auth";
 import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { useOperations } from "context/operationsProvider";
+import { useNotify } from "@canonical/react-components";
+
+const EVENT_HANDLER_DELAY = 250;
+const WS_RETRY_DELAY_MULTIPLIER = 250;
+const MAX_WS_CONNECTION_RETRIES = 5;
 
 const Events: FC = () => {
   const { isAuthenticated } = useAuth();
   const eventQueue = useEventQueue();
   const queryClient = useQueryClient();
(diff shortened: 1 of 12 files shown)


edlerd (approved):
This is great, thanks for adding it to LXD! Code LGTM as is.

I need to do more QA tomorrow, see the comment below. I'd like to check those concerns, it might not be an issue but we should be sure before merging.

src/pages/instances/Events.tsx
```diff
@@ -65,45 +71,60 @@ const Events: FC = () => {
...
+          });
+        }
+        // ensure open requests that reply with an operation and register
+        // new handlers in the eventQueue are closed before handling the event
+        setTimeout(() => {
+          handleEvent(event);
+        }, EVENT_HANDLER_DELAY);
+      };
+    } catch (e) {
+      if (retryCount < MAX_WS_CONNECTION_RETRIES) {
```
edlerd, line 109: Need to ensure this won't delay surfacing connection error for users with restricted permissions.

Also the need for reconnect in LXD is slightly lower than in anbox. Because here we only ever connect to a single endpoint. Unlike anbox connecting to multiple sub clusters where some might be unavailable.

lorumic: Ok, let me know and in case of any problem I can revert this part. But if QA works with no problems, I think having a reconnection mechanism is good even for the single endpoint :)

edlerd: All good after QA check.

lorumic: Ok, if there are no changes needed feel free to merge whenever you want - I don't have the authorisation to merge myself :)

PR #1132, opened 2025-03-04
Use cla v2

edlerd:
## Done

- update cla check in PR workflow to v2


Kxiru (approved):
Thanks for this. What exactly does it do?

edlerd:
> Thanks for this. What exactly does it do?

This is for external contributions. Checking they have signed the content licence agreement. A new version of that canonical service was released recently, so we update our action to it.

PR #1133, opened 2025-03-04
Add ssh key configuration for instance edit WD-19727

edlerd:
## Done

- Add ssh key configuration for instance edit

Fixes WD-19727

## QA

1. Run the LXD-UI:
    - create an instance and apply ssh keys
    - create a profile with ssh keys
    - ensure the profiles ssh keys get applied

Todo:
* profile edit and create saving
* show inherited ssh keys
* move to dedicated section in config form


piperdeck:
This is looking really polished! I just have a couple notes:

---

Can we hold onto text that's been typed into this field when you switch to another source and then switch back?

https://github.com/user-attachments/assets/2c8a8b65-43d2-4289-b41a-52beaa7167a1

---

Can we please have the file picker filter for `.pub` files?

CleanShot 2025-03-24 at 22 09 22@2x

---

After I create an ssh key and confirm, when the modal disappears the key I've just created ends up below the fold, so I can't see confirmation of what I've just done without scrolling down. It also hides the new ssh key button, so if I have a few keys to create in a row, I need to scroll down in between each one to click the button.

Can we please have it so that the page scrolls up to reveal the bottom of the ssh keys table and button after confirming key creation in the modal?

https://github.com/user-attachments/assets/66cf0ec0-cdbb-4e03-8ace-1e0d31ac738b

---

This feels like a small thing, but because ssh keys have this long string of text in them, the text wrapper ends up breaking them on the space after `ssh-rsa`, creating this weird white space that makes it look like you've mistakenly pasted something with two lines. Could we add `word-break: break-all` to this textarea so that it breaks the text inside the hash rather than on the space before the hash?

| Current | With word-break: break-all |
| --- | --- |
| CleanShot 2025-03-24 at 22 32 39@2x | CleanShot 2025-03-24 at 22 32 30@2x |

---

On safari, when I upload a public key file, the attached file has this weird dotted rectangle next to the file name. Is this normal? Is it supposed to be a file icon?

CleanShot 2025-03-24 at 22 40 22@2x

edlerd:

Addressed all issues mentioned. Please have another pass at the updated PR for design review.

Except for the "filter .pub files only", that was already set. I see on Ubuntu, if you walk up in the directory structure, the file filter is reset. But that is an OS behavior we can't influence. On macOS, it grays out the other files, works fine. On Ubuntu, it also works fine, as long as you stay in the directory or only walk down directories.

Kxiru (approved):
Thanks for this, David!

**Screenshots**:
1) Instance creation:
image
image
image


2) Profile creation:
image

3) New instance inheriting Profile SSH key:
image
image

Perhaps in a new PR we can display SSH keys on the Instance and Profile overview pages and maybe even the side panels?

edlerd:
> Perhaps in a new PR we can display SSH keys on the Instance and Profile overview pages and maybe even the side panels?

Oh yes, we could. At least an indication if keys are attached and how many.

PR #1134, opened 2025-03-04
feat: [WD-19336] Storage volumes permission checks

Kxiru:
## Done

**Storage Volumes page**
- if instance view permission is missing we should disable linking to the instance
- if image view permission is missing we should disable linking to the image list
- disable create volume button if user does not have permission
- disable add snapshot button if user does not have permission
- disable button if user does not have permission to delete a SV (DeleteStorageVolumeComponent x2)
- disable rename header if permission is missing
- disable SV migrate button if user does not have permission
- disable SV duplicate button if user does not have permission for any projects
- disable Add SV Snapshot button if user does not have permission
- disable button if user does not have permission
- disable configure volume snapshot configs button if user does not have permission
- disable bulk delete volume snapshots button if user does not have permission
- disable edit volume snapshot button if user does not have permission
- disable restore volume from snapshot button if user does not have permission
- disable delete volume snapshot button if user does not have permission
- disable all configs if permission is missing
- bugfix: [Permission Checks] Disk Size select component not mirroring edi…

See [Storage Volume Permission Actions](https://docs.google.com/spreadsheets/d/1fXBNG9B8CL2iXP6kPEe4a3YM0q2QcOnj-I3essIUuh8/edit?gid=954494754#gid=954494754) for full list of implemented / pending actions to permission check.

## QA

1. Run the LXD-UI:

## Screenshots
image

### DIFF
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -21,6 +21,7 @@ export const instanceEntitlements = [
   "can_delete",
   "can_edit",
   "can_exec",
+  "can_view",
   "can_manage_backups",
   "can_manage_snapshots",
   "can_update_state",
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -21,7 +21,6 @@ export const instanceEntitlements = [
   "can_delete",
   "can_edit",
   "can_exec",
-  "can_view",
   "can_manage_backups",
   "can_manage_snapshots",
   "can_update_state",
--- a/src/pages/storage/MigrateVolumeBtn.tsx
+++ b/src/pages/storage/MigrateVolumeBtn.tsx
@@ -12,6 +12,7 @@ import { useNavigate } from "react-router-dom";
 import ResourceLabel from "components/ResourceLabel";
 import ResourceLink from "components/ResourceLink";
 import { migrateStorageVolume } from "api/storage-volumes";
+import { useStorageVolumeEntitlements } from "util/entitlements/storage-volumes";
 
 interface Props {
   storageVolume: LxdStorageVolume;
@@ -32,6 +33,7 @@ const MigrateVolumeBtn: FC<Props> = ({
   const queryClient = useQueryClient();
   const navigate = useNavigate();
   const [isVolumeLoading, setVolumeLoading] = useState(false);
+  const { canEditVolume } = useStorageVolumeEntitlements();
 
   const handleSuccess = (
     newTarget: string,
@@ -157,7 +159,7 @@ const MigrateVolumeBtn: FC<Props> = ({
         type="button"
         className={classname}
         loading={isVolumeLoading}
-        disabled={isVolumeLoading}
+        disabled={!canEditVolume(storageVolume) || isVolumeLoading}
         title="Migrate volume"
       >
         <Icon name="machines" />
--- a/src/pages/storage/MigrateVolumeBtn.tsx
+++ b/src/pages/storage/MigrateVolumeBtn.tsx
@@ -161,9 +161,9 @@ const MigrateVolumeBtn: FC<Props> = ({
         loading={isVolumeLoading}
         disabled={!canEditVolume(storageVolume) || isVolumeLoading}
         title={
-          !canEditVolume(storageVolume)
-            ? "You do not have permission to migrate this volume"
-            : "Migrate volume"
+          canEditVolume(storageVolume)
+            ? "Migrate volume"
+            : "You do not have permission to migrate this volume"
         }
       >
         <Icon name="machines" />
--- a/src/pages/storage/MigrateVolumeBtn.tsx
+++ b/src/pages/storage/MigrateVolumeBtn.tsx
@@ -160,7 +160,11 @@ const MigrateVolumeBtn: FC<Props> = ({
         className={classname}
         loading={isVolumeLoading}
         disabled={!canEditVolume(storageVolume) || isVolumeLoading}
-        title="Migrate volume"
+        title={
+          !canEditVolume(storageVolume)
+            ? "You do not have permission to migrate this volume"
+            : "Migrate volume"
+        }
       >
         <Icon name="machines" />
         <span>Migrate</span>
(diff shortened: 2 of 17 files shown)


src/api/instances.tsx
```diff
@@ -21,6 +21,7 @@ export const instanceEntitlements = [
   "can_delete",
   "can_edit",
   "can_exec",
+  "can_view",
```
edlerd, line 24:
```suggestion
```
We never check view permission explicitly. If a user knows an entity exists, they can view it. We can always infer this permission.

src/pages/storage/StorageVolumeNameLink.tsx
```diff
@@ -23,21 +25,44 @@ const StorageVolumeNameLink: FC<Props> = ({
...
+      ? overrideName
+      : ""
+    : volume.name;
 
   return (
     <div className={classnames("u-flex", className)}>
       <div
         className={classnames("u-truncate", "volume-name-link")}
-        title={caption}
+        title={volume.name}
```
edlerd, line 53: The title should match the truncated contents.
```suggestion
        title={caption}
```

src/pages/storage/StorageVolumeNameLink.tsx
```diff
@@ -23,21 +25,44 @@ const StorageVolumeNameLink: FC<Props> = ({
...
+    isImage && images.find((image) => image.fingerprint === volume.name);
+
+  const isCustom = volume.type === "custom";
+  const displayLink = instance || image || isCustom;
+
+  const caption = overrideName
+    ? displayLink
+      ? overrideName
+      : ""
+    : volume.name;
```
edlerd, line 47: If I understand correctly, this is to not render anything in the right-hand column on the volume list. Currently, we still render some markup with titles and empty string as content in the case that link should be hidden. Better simplify and avoid the empty markup, as suggested below?

Might be worth adding a comment or solving this in a more straight forward way to make it easy to understand.

```suggestion
  const caption = overrideName ? overrideName : volume.name;
  
  if (overrideName && !displayLink) {
    return null;
  }
```

src/pages/storage/actions/CreateVolumeBtn.tsx
```diff
@@ -27,6 +35,7 @@ const CreateVolumeBtn: FC<Props> = ({ project, className, defaultPool }) => {
       hasIcon={!isSmallScreen}
       onClick={handleAdd}
       className={className}
+      disabled={!canCreateStorageVolumes(project)}
```
edlerd, line 38: Please add a title.

src/pages/storage/actions/DuplicateVolumeBtn.tsx
```diff
@@ -31,6 +33,7 @@ const DuplicateVolumeBtn: FC<Props> = ({ volume, className, onClose }) => {
         className={className}
         onClick={openPortal}
         title="Duplicate volume"
+        disabled={!canCreateStorageVolumes(volume)}
```
edlerd, line 36: Title please

src/pages/storage/actions/snapshots/VolumeAddSnapshotBtn.tsx
```diff
@@ -37,9 +39,11 @@ const VolumeAddSnapshotBtn: FC<Props> = ({
           title={
             isDisabled
               ? `Snapshot creation is blocked for project ${volume.project}`
-              : "Add Snapshot"
+              : !canManageStorageVolumeSnapshots(volume)
+                ? "You do not have permission to create or delete snapshots of this volume."
+                : "Add Snapshot"
```
edlerd, line 44: Probably better to extract this into a function with early exits as well. Then the disabled check below can use that same function. See how `DeleteStoragePoolBtn` is doing it for an example.

edlerd (commented):
This is a good direction. Some comments on the draft below.

edlerd (commented):
Thanks for adding the checks for volumes. QA looks mostly good. Some refinement on the copy is need. Also, some small comments on the code.

This covers everything from our permission tracking spreadsheet for volumes.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -157,8 +159,12 @@ const MigrateVolumeBtn: FC<Props> = ({
         type="button"
         className={classname}
         loading={isVolumeLoading}
-        disabled={isVolumeLoading}
-        title="Migrate volume"
+        disabled={!canEditVolume(storageVolume) || isVolumeLoading}
+        title={
+          !canEditVolume(storageVolume)
+            ? "You do not have permission to migrate this volume"
+            : "Migrate volume"
```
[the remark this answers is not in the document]
Kxiru: I found 6 instances of this, thank you for the heads up!

src/pages/storage/actions/DuplicateVolumeBtn.tsx
```diff
@@ -30,7 +32,12 @@ const DuplicateVolumeBtn: FC<Props> = ({ volume, className, onClose }) => {
         aria-label="Duplicate volume"
         className={className}
         onClick={openPortal}
-        title="Duplicate volume"
+        title={
+          canCreateStorageVolumes(volume)
```
edlerd, line 36: I think we need to pass the project here, not the volume.

src/pages/storage/actions/snapshots/VolumeSnapshotActions.tsx
```diff
@@ -43,6 +45,12 @@ const VolumeSnapshotActions: FC<Props> = ({ volume, snapshot }) => {
     />
   );
 
+  const getTitle = (action: string) => {
+    return !canManageStorageVolumeSnapshots(volume)
+      ? `You do not have permission to ${action.toLowerCase()} this snapshot`
+      : `${action} snapshot`;
+  };
+
```
edlerd, line 53: Good idea to make this reusable! Would suggest using a positive check here as well.

edlerd (commented):
Two tiny comments below.

src/pages/storage/actions/snapshots/VolumeSnapshotActions.tsx
Kxiru: Haha, thank you, and done!

edlerd (approved):
LGTM, thanks for the iteration and improvements.

PR #1142, opened 2025-03-17
fix custom iso volumes in clusters

edlerd:
## Done

- fix custom iso volumes in clusters: ISO volumes stored on a local storage pool in a cluster can only be used for instances on the member they are saved on

## QA

1. Run the LXD-UI:
    - in a clustered backend:
      - upload a custom ISO to a storage pool that is local to each cluster member
      - use the custom ISO, ensure only the member that the ISO resides on can be selected for the instance target
      - create an instance, ensure you can place it on any cluster member.
      - upload a custom ISO to a storage pool that is cluster wide (todo: this is failing, might be a backend bug).
    - in a non-clustered backend:
      - create an instance

### DIFF
--- a/src/pages/instances/forms/InstanceLocationSelect.tsx
+++ b/src/pages/instances/forms/InstanceLocationSelect.tsx
@@ -1,6 +1,6 @@
 import type { FC } from "react";
 import { useState } from "react";
-import { Select } from "@canonical/react-components";
+import { Input, Select } from "@canonical/react-components";
 import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import Loader from "components/Loader";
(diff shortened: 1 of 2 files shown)


src/pages/instances/forms/InstanceLocationSelect.tsx
```diff
@@ -68,6 +68,22 @@ const InstanceLocationSelect: FC<Props> = ({ formik }) => {
...
+    .targetSelectedByVolume;
+
+  if (isPreselected) {
+    return (
+      <Input
+        id="clusterMember"
+        label="Cluster member"
+        value={formik.values.target}
+        disabled
+        help="Member is determinded by the selected ISO volume and can't be changed."
```
Kxiru, line 81: determined*

Kxiru (changes requested):
LGTM. Minor copy check.

src/pages/instances/forms/InstanceLocationSelect.tsx
edlerd: good catch, have fixed this below.

Kxiru (approved):
Thanks for the copy change, and the code ofc!

PR #1150, opened 2025-03-20
Update vite and vitest

edlerd:
## Done

- Update vite and vitest to fix security problems


Kxiru (approved):
LGTM. Was there a necessary reason for this update? :)

PR #1151, opened 2025-03-20
Ensure instances in foreign projects are shown with project in the network topology. Avoid mixup of projects for instance metrics.

### LINKED ISSUE #1149: memory/disk column don't disappear after stopping an instance
When i stop my instance the memory and disk column don't disappear like the ipv4/ipv6 columns do

Image

At least memory should disappear, disk is still in use when it's stopped so that is a good candidate to stay visible :)

edlerd:
## Done

- Ensure instances in foreign projects are shown with a parent project in the network topology
- Ensure to pick instance metrics from the right project. do not confuse instances with the same name on different projects

Fixes #1149

## QA

1. Run the LXD-UI:
    - create instances in different projects, assign them to the same network
    - browse the network detail page, ensure the instances in foreign projects show up with their project next to them.
    - With two instances having the same name from different projects, ensure the metrics displayed are belonging to the right instance.

### DIFF
--- a/src/pages/networks/NetworkTopology.tsx
+++ b/src/pages/networks/NetworkTopology.tsx
@@ -165,8 +165,20 @@ const NetworkTopology: FC<Props> = ({ formik, project, isServerClustered }) => {
             .slice(0, isInstancesCollapsed ? 5 : instances.length)
             .map((item) => {
               const instanceUrl = `/ui/project/${item.project}/instance/${item.name}`;
+              const projectUrl = `/ui/project/${item.project}`;
+              const isExternalProject = item.project !== project;
               return (
                 <div key={instanceUrl} className="downstream-item">
+                  {isExternalProject && (
+                    <>
+                      <ResourceLink
+                        type="project"
+                        value={item.project}
+                        to={projectUrl}
+                      />{" "}
+                      /{" "}
+                    </>
+                  )}
                   <ResourceLink
                     type="instance"
                     value={item.name}
--- a/src/util/formChangeCount.spec.ts
+++ b/src/util/formChangeCount.spec.ts
@@ -244,4 +244,85 @@ describe("formChangeCount", () => {
 
     expect(result).toBe(1);
   });
+
+  it("counts rename of a device as one", () => {
+    const formik = {
+      initialValues: {
+        devices: [
+          {
+            name: "eth0",
+            type: "nic",
+          },
+        ],
+      },
+      values: {
+        devices: [
+          {
+            name: "eth1",
+            type: "nic",
+          },
+        ],
+      },
+    } as ConfigurationRowFormikProps;
+
+    const result = getFormChangeCount(formik);
+
+    expect(result).toBe(1);
+  });
+
+  it("counts rename and detail change of a device as two", () => {
+    const formik = {
+      initialValues: {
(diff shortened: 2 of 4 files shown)


Kxiru (commented):
Network that *should* include 3 instances:
image

1 instance in default project (test1)
2 instances in Project1 (test1, test2)
Each instance has the following in their network section in the instance overview.
image

Yet Test 2 does not seem to show in the network display.

Troubleshooting:
- Pehaps because all of the instances are on teh same interfaces, eth-1, but in that case, why are test1 and test1(project1) showing?
 
In trying to change the interface, 2 changes needed to be save when I only edited one field: 
image

Additionally, unable to save a new device name:
image

edlerd:
I think you found another bug upstream in LXD. With same instance names in different projects connected to the same network. It seems the network to instance link should contain the project as identifier next to the name, but is not. Let's further investigate and report this to the core team as well to verify.

edlerd:
> Pehaps because all of the instances are on teh same interfaces, eth-1, but in that case, why are test1 and test1(project1) showing?

The interface is per instance local. I think the network won't know about it's name.

edlerd:
> In trying to change the interface, 2 changes needed to be save when I only edited one field:

This seems like another bug in the UI.

edlerd:
> I think you found another bug upstream in LXD. With same instance names in different projects connected to the same network. It seems the network to instance link should contain the project as identifier next to the name, but is not. Let's further investigate and report this to the core team as well to verify.

This is wrong, the same instance name in different projects works fine, also in your report, the test1 instance exists twice in different projects and that works well.

edlerd:
> Additionally, unable to save a new device name:

This seems like a bug. When you rename the network interface on an instance, this error pops up. I am not sure how to rename a network device, there might be a special endpoint to rename devices. We should investigate and maybe disable rename of network interfaces on instance edit in the UI.

Kxiru:
> Refere

Kxiru:
> > Yet Test 2 does not seem to show in the network display.
> 
> That sounds wired, might be a bug. But I can't reproduce it yet.
> 
> @Kxiru Can you please check in the dev consoles requests what the response to `GET /1.0/networks/testnet` is? In the `metadata.used_by` list of the response, there should be all three instances listed. That is the case for me when trying to reproduce it.

Discussed together in meeting

Kxiru:
> > Additionally, unable to save a new device name:
> 
> This seems like a bug. When you rename the network interface on an instance, this error pops up. I am not sure how to rename a network device, there might be a special endpoint to rename devices. We should investigate and maybe disable rename of network interfaces on instance edit in the UI.
> 
> ~I think there is an issue with network device rename on instances. This might not be allowed, then the UI should disable the field in edit mode. This might also be a bug in LXD. I see rename of devices works fine for others, such as disk devices. This might also be related to only some network types.~
> 
> Reported as [canonical/lxd#15300](https://github.com/canonical/lxd/issues/15300)

As discussed, this also appears when adding a NIC device to an instance where the instance's name is already taken by an instance in another project, and both are connected on the same network.

Kxiru (approved):
Thank you for your work on this, David! Let's keep an eye out on the renaming network interface bug :)

Screenshots:
image

PR #1158, opened 2025-03-24
Detect custom image for operation relating to an image in a non-default project

### LINKED ISSUE #1154: Invalid operation representation
lxd version: `5.21/stable`

In cli my current project is `lknkljl` and when I do `lxc operation ls` then I get the operations

Image

But When I switch from `lknkljl` to `default` project and do `lxc operation ls` then there is no operation in the `default` project. then why GUI is show that this operation belongs to `default` project?

Image

Image

These are the cached images in my `lknkljl` project.

Image

issue is in this `func` [getProjectName](https://github.com/canonical/lxd-ui/blob/main/src/util/operations.tsx) it does not check the operations of images 

```
                "resources": {
                    "images": [
                        "/1.0/images/333449f566531c586d405772afaf9ced7eb9c2ca2f191d487c63b170f62b3172?project=lknkljl"
                    ]
                },
```

Image

> If I understand correctly, in your case the operations from projects `lknkljl` are displayed under `default` project in LXD UI, but not in the `lxc`?
>  

 _Originally posted by @MusicDin in [#13509](https://github.com/canonical/lxd/issues/13509#issuecomment-2742812976)_

edlerd:
## Done

- Detect custom image for operation relating to an image in a non-default project

Fixes #1154

### DIFF
--- a/src/types/operation.d.ts
+++ b/src/types/operation.d.ts
@@ -14,6 +14,7 @@ export interface LxdOperation {
   metadata?: Record<string, string>;
   may_cancel: boolean;
   resources?: {
+    images?: string[];
     instances?: string[];
     instances_snapshots?: string[];
     storage_volume_snapshots?: string[];
--- a/src/util/operations.spec.ts
+++ b/src/util/operations.spec.ts
@@ -7,6 +7,7 @@ import {
 import type { LxdOperation } from "types/operation";
 
 const craftOperation = (...url: string[]) => {
+  const images: string[] = [];
   const instances: string[] = [];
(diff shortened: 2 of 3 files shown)


Kxiru (approved):
Hey David!
The code for this looks sound but I wonder- in what circumstances can/will this logic be used?
Thank you also for the test!

edlerd:
> The code for this looks sound but I wonder- in what circumstances can/will this logic be used?

See the issue that was reported with a reproducer: https://github.com/canonical/lxd-ui/issues/1154

PR #1160, opened 2025-03-24
Use link instead of anchor to avoid full page load

edlerd:
## Done

- Use link instead of anchor to avoid full page load in storage and login section.

## QA

1. Run the LXD-UI:
    - go to storage pool detail page, click "Volumes". Ensure the page does not refresh fully, but the link works.
    - go to login page, click "Login with TLS" and ensure the page does not refresh fully, but the link works.

### DIFF
--- a/src/pages/login/Login.tsx
+++ b/src/pages/login/Login.tsx
@@ -1,6 +1,6 @@
 import type { FC } from "react";
 import { Icon } from "@canonical/react-components";
-import { Navigate } from "react-router-dom";
+import { Link, Navigate } from "react-router-dom";
 import { useAuth } from "context/auth";
 import Loader from "components/Loader";
 import { useSettings } from "context/useSettings";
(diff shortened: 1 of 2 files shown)


Kxiru (approved):
A bit hard to test because I'm not sure what "ensure the page does not refresh fully", means, especially on the Login page - the experience has little to no difference for me.
I think I see it more clearly when testing the storage volumes page. There is no "loading" icon as there was before. I did have a little lag when clicking on a volume from the Volumes page at first but it disappeared after 1-3 interactions and could not be replicated.

edlerd:
> A bit hard to test because I'm not sure what "ensure the page does not refresh fully", means, especially on the Login page - the experience has little to no difference for me.
> I think I see it more clearly when testing the storage volumes page. There is no "loading" icon as there was before. I did have a little lag when clicking on a volume from the Volumes page at first but it disappeared after 1-3 interactions and could not be replicated.

You can see it in the dev console -- it should only fetch additional js resources, not navigate to a new page and load all resources again. It is not a big difference, apart from what you already described on the storage volume link. Especially through caching this should be fast as well. But a tiny optimization.

PR #1164, opened 2025-03-27
feat: [WD-19726] TLS Identity creation with groups

Kxiru:
## Done

- Allow group selection into the userflow for creating TLS identities

## QA

1. Run the LXD-UI:
    - Navigate to the "Identities" page in the UI and attempt to create an identity (with the correct permissions). Observe that a panel should appear and a token should be able to be generated by creating an identity with a name and 0 or more groups.

## Screenshots

image

### DIFF
--- a/src/api/auth-identities.tsx
+++ b/src/api/auth-identities.tsx
@@ -108,13 +108,15 @@ export const deleteIdentities = async (
 };
 
 export const createFineGrainedTlsIdentity = async (
-  clientName: string,
+  name: string,
+  groups: string[],
 ): Promise<TlsIdentityTokenDetail> => {
   return new Promise((resolve, reject) => {
     fetch(`/1.0/auth/identities/tls`, {
       method: "POST",
       body: JSON.stringify({
-        name: clientName,
+        name: name,
+        groups: groups,
         token: true,
       }),
     })
--- a/src/pages/permissions/CreateIdentityModal.tsx
+++ b/src/pages/permissions/CreateIdentityModal.tsx
@@ -1,49 +1,15 @@
 import type { FC } from "react";
 import { useState } from "react";
-import {
-  ActionButton,
-  Button,
-  Icon,
-  Input,
-  Modal,
-  Notification,
-} from "@canonical/react-components";
-import { useFormik } from "formik";
-import { createFineGrainedTlsIdentity } from "api/auth-identities";
-import { base64EncodeObject } from "util/helpers";
-import { useQueryClient } from "@tanstack/react-query";
-import { queryKeys } from "util/queryKeys";
+import { Button, Icon, Modal } from "@canonical/react-components";
 
 interface Props {
   onClose: () => void;
+  token: string;
+  identityName: string;
 }
 
-const CreateIdentityModal: FC<Props> = ({ onClose }) => {
-  const [token, setToken] = useState<string | null>(null);
-  const [error, setError] = useState(false);
+const CreateIdentityModal: FC<Props> = ({ onClose, token, identityName }) => {
   const [copied, setCopied] = useState(false);
-  const queryClient = useQueryClient();
-
-  const formik = useFormik({
-    initialValues: {
-      identityName: "",
-    },
-    onSubmit: (values) => {
-      setError(false);
-      createFineGrainedTlsIdentity(values.identityName)
-        .then((response) => {
-          const encodedToken = base64EncodeObject(response);
-          setToken(encodedToken);
-
-          queryClient.invalidateQueries({
-            queryKey: [queryKeys.identities],
-          });
-        })
-        .catch(() => {
-          setError(true);
-        });
-    },
-  });
 
   const handleCopy = async () => {
     if (token) {
@@ -64,7 +30,7 @@ const CreateIdentityModal: FC<Props> = ({ onClose }) => {
     <Modal
       close={onClose}
       className="create-tls-identity"
-      title={token ? "Identity has been created" : "Generate trust token"}
+      title={`Identity ${identityName} created`}
       buttonRow={
         <>
           {token && (
@@ -80,6 +46,7 @@ const CreateIdentityModal: FC<Props> = ({ onClose }) => {
                 hasIcon
               >
                 <Icon name={copied ? "task-outstanding" : "copy"} />
+                <span>Copy trust token</span>
               </Button>
 
               <Button
@@ -92,47 +59,19 @@ const CreateIdentityModal: FC<Props> = ({ onClose }) => {
               </Button>
             </>
           )}
-          {!token && (
-            <ActionButton
-              aria-label="Generate token"
-              appearance="positive"
-              className="u-no-margin--bottom"
-              onClick={() => void formik.submitForm()}
-              disabled={
-                formik.values.identityName.length === 0 || formik.isSubmitting
-              }
-              loading={formik.isSubmitting}
-            >
-              Generate token
-            </ActionButton>
-          )}
         </>
       }
     >
-      {error && (
-        <Notification
-          severity="negative"
-          title="Token creation failed"
-        ></Notification>
-      )}
-      {!token && (
-        <Input
-          id="identityName"
-          type="text"
-          label="Identity Name"
-          onBlur={formik.handleBlur}
-          onChange={formik.handleChange}
-          value={formik.values.identityName}
-        />
-      )}
       {token && (
         <>
-          <p>The token below can be used to log in as the new identity.</p>
-
-          <Notification
-            severity="caution"
-            title="Make sure to copy the token now as it will not be shown again."
-          ></Notification>
+          <p>
+            The trust token below can be used to log in with the newly created
+            identity.{" "}
+            <b>
+              Once this modal is closed, the trust token can’t be generated
+              again.
+            </b>
+          </p>
 
           <div className="token-code-block">
             <code>{token}</code>
--- a/src/pages/permissions/CreateTLSIdentity.tsx
+++ b/src/pages/permissions/CreateTLSIdentity.tsx
@@ -1,0 +1,40 @@
+import { useState, type FC } from "react";
+import CreateIdentityModal from "./CreateIdentityModal";
+import CreateTLSIdentityPanel from "./panels/CreateTLSIdentityPanel";
+import { usePortal } from "@canonical/react-components";
+import usePanelParams, { panels } from "util/usePanelParams";
+
+const CreateTLSIdentity: FC = () => {
+  const panelParams = usePanelParams();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal({
+    programmaticallyOpen: true,
+  });
+  const [token, setToken] = useState("");
+  const [identityName, setIdentityName] = useState("");
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <CreateIdentityModal
+            onClose={closePortal}
+            token={token}
+            identityName={identityName}
+          />
+        </Portal>
+      )}
+
+      {panelParams.panel === panels.createTLSIdentity && (
+        <CreateTLSIdentityPanel
+          onSuccess={(identityName: string, token: string) => {
+            setIdentityName(identityName);
+            setToken(token);
+            openPortal();
+          }}
+        />
+      )}
+    </>
+  );
+};
+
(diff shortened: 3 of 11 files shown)


edlerd:
Thanks for putting this up!

The process seems broken, after submitting the side panel, the identity should be directly created with the groups selected and the result modal with the token presented.

There is a step in between that shouldn't be there asking again for a name and the identity is created without groups at the moment.

Kxiru:
> Thanks for putting this up!
> 
> The process seems broken, after submitting the side panel, the identity should be directly created with the groups selected and the result modal with the token presented.
> 
> There is a step in between that shouldn't be there asking again for a name and the identity is created without groups at the moment.

Hey David!
As discussed, at the time of posting this comment, the PR was a draft, but all the features you are looking for are now apparent in this full PR.
Please provide a code and QA review if you have the time and let me know your thoughts. :)

src/api/auth-identities.tsx
```diff
@@ -107,14 +107,16 @@ export const deleteIdentities = async (
   });
 };
 
-export const createFineGrainedTlsIdentity = async (
-  clientName: string,
-): Promise<TlsIdentityTokenDetail> => {
+export const createFineGrainedTlsIdentity = async (identity: {
+  name: string;
+  groups: string[];
+}): Promise<TlsIdentityTokenDetail> => {
```
edlerd, line 113:
```suggestion
export const createFineGrainedTlsIdentity = async (name: string, groups: string[]): Promise<TlsIdentityTokenDetail> => {
```
Maybe a bit simpler to unpack this outside of the api layer?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -60,11 +26,15 @@ const CreateIdentityModal: FC<Props> = ({ onClose }) => {
     }
   };
 
+  const handleClose = () => {
+    onClose();
+  };
```
edlerd, line 31:
```suggestion
```
Can we directly call `onClose` instead? I don't think we need the extra callback here.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -60,11 +26,15 @@ const CreateIdentityModal: FC<Props> = ({ onClose }) => {
...
+
   return (
     <Modal
-      close={onClose}
+      close={handleClose}
       className="create-tls-identity"
-      title={token ? "Identity has been created" : "Generate trust token"}
+      title={"Identity created"}
```
edlerd, line 37: Maybe best to include the name of the identity here?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -92,47 +63,25 @@ const CreateIdentityModal: FC<Props> = ({ onClose }) => {
...
-            title="Make sure to copy the token now as it will not be shown again."
-          ></Notification>
+          <p>
+            The trust token below can be used to log in with the newly created
+            identity.{" "}
+            <b>
+              Once this modal is closed, the trust token can’t be generated
+              again.
+            </b>
+          </p>
```
edlerd, line 84: Good idea to change the wording to "trust token" to match the new onboarding! Why remove the notification with the previous copy "Make sure to copy the token now as it will not be shown again"? In my opinion, that was slightly better. Wdyt?

src/pages/permissions/CreateTLSIdentity.tsx
```diff
@@ -0,0 +1,36 @@
...
+            error={error}
+          />
+        </Portal>
+      )}
+
+      {panelParams.panel === panels.createTLSIdentity && (
+        <CreateTLSIdentityPanel
+          openPortal={openPortal}
+          setToken={setToken}
+          setError={setError}
```
edlerd, line 29: With moving the error handling into the panel, the interface can slightly simplify to decouple the two components:
```suggestion
          onSuccess={setToken}
```

src/pages/permissions/forms/IdpGroupForm.tsx
```diff
@@ -5,9 +5,12 @@ import type { FormikProps } from "formik/dist/types";
 export interface IdpGroupFormValues {
   name: string;
 }
+export interface TLSIdentityFormValues {
+  name: string;
+}
```
edlerd, line 10: The type should also include the groups.
```suggestion
export interface TLSIdentityFormValues {
  name: string;
  groups: string[];
}
```

src/pages/permissions/panels/CreateTLSIdentityPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+      .then((response) => {
+        const encodedToken = base64EncodeObject(response);
+        setToken(encodedToken);
+
+        queryClient.invalidateQueries({
+          queryKey: [queryKeys.identities],
+        });
+      })
+      .catch(() => {
+        setError(true);
```
edlerd, line 90: Better show the api error in a notification in this panel.

src/pages/permissions/panels/CreateTLSIdentityPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const handleSubmit = (values: TLSIdentityFormValues) => {
+    const newIdentity = {
+      name: values.name,
+      groups: Array.from(desiredState.groupsAdded),
+    };
+    setToken("");
+
+    closePanel();
+    setError(false);
```
edlerd, line 79:
```suggestion
```

src/pages/permissions/panels/CreateTLSIdentityPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+    openPortal();
+  };
+
+  const groupSchema = Yup.object().shape({
+    name: Yup.string().required("Identity name is required"),
+  });
+
+  const formik = useFormik<TLSIdentityFormValues>({
+    initialValues: {
+      name: "",
```
edlerd, line 102:
```suggestion
      name: "",
      groups: [],
```
I think we should init the groups as well.

src/pages/permissions/panels/CreateTLSIdentityPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+    onSubmit: handleSubmit,
+  });
+
+  return (
+    <>
+      <SidePanel isOverlay loading={isLoading} hasError={!groups}>
+        <SidePanel.Header>
+          <SidePanel.HeaderTitle>Create identity</SidePanel.HeaderTitle>
+        </SidePanel.Header>
+        <NotificationRow className="u-no-padding" />
```
edlerd, line 114: We already have the notification row here, just need to set the state in the error case to have it displayed :)

src/pages/permissions/forms/IdpGroupForm.tsx
```diff
@@ -5,9 +5,12 @@ import type { FormikProps } from "formik/dist/types";
...
 
 interface Props {
-  formik: FormikProps<IdpGroupFormValues>;
+  formik: FormikProps<IdpGroupFormValues | TLSIdentityFormValues>;
 }
 
 const IdpGroupForm: FC<Props> = ({ formik }) => {
```
edlerd, line 16: The name is now wrong. Maybe rename this to `NameWithGroupFrom`, so it matches the new use case and the old one?

edlerd (commented):
This looks great and works well already. A couple of QA comments:

1. After submitting, the panel should stay open and the submit button show the loading state. On error, show the message in the panel, same as the other permission panels do, you can explore their implementation for how it can be done or we can discuss it further. Currently, there is the modal shown immediately with an error, which seems wrong.

2. The confirmation modal content, maybe it should mention the identity name entered? I also suggest showing the groups in the modal. Or the group count to keep it short. What do you think?

> at the time of posting this comment, the PR was a draft, but all the features you are looking for are now apparent in this full PR. 

thanks, I missed the draft status on the previous comment.

src/pages/permissions/CreateIdentityModal.tsx
Kxiru: Well,
1) I was going off of Piper's design, and I found that the interface looked a lot simpler and sleek without the notification, which IMO just added a bit too much to the page. It can certainly come back, though the current design also does a good job of highlighting the fact that the token will not be shown again. Happy to go in either direction!

src/pages/permissions/forms/IdpGroupForm.tsx
Kxiru: Should it? The way I understood it, the formvalues referred to the first formik form that only contained the name input in it.

The logic of adding groups was not changed, and the IDPGroupFormValues did not include a string[] interface for the groups. I believe the two areas are being addressed separately. Am I misunderstanding something here?

Kxiru: I see that this suggestion is based off of a later change/suggestion that you have made. Noted!

src/pages/permissions/forms/IdpGroupForm.tsx
Kxiru: I'm not sure I understand this one. Do you mind revisiting my new logic and checking whether it still applies?

src/pages/permissions/CreateTLSIdentity.tsx
```diff
@@ -0,0 +1,36 @@
...
+            identityName={identityName}
+          />
+        </Portal>
+      )}
+
+      {panelParams.panel === panels.createTLSIdentity && (
+        <CreateTLSIdentityPanel
+          openPortal={openPortal}
+          onSuccess={setToken}
+          setIdentityName={setIdentityName}
```
edlerd, line 29:
```suggestion
          openPortal={openPortal}
          onSuccess={(token, identityName) => { setToken ... setIdentityName ... openPortal}}
```
I think the interface can be much smaller and more concise, as all the updates relate to a single event.

edlerd (commented):
Code looks mostly polished, good stuff :)

Just two small and probably quick suggestions below.

src/pages/permissions/forms/IdpGroupForm.tsx
Kxiru: Ooh, right I understand. I think NameAndGroupForm will suffice!

src/pages/permissions/panels/CreateTLSIdentityPanel.tsx
```diff
@@ -0,0 +1,151 @@
...
+    }
+  };
+
+  const closePanel = () => {
+    panelParams.clear();
+    notify.clear();
+  };
+
+  const handleSubmit = (values: TLSIdentityFormValues) => {
+    closePanel();
```
edlerd, line 66: This is too early to close the panel. We should do this only below in the `then` branch. Otherwise, in the error case the user input will be lost. By postponing the panel close, the user can resubmit with the previous inputs.

edlerd (commented):
QA and code 99% LGMT. One tiny issue left with the panel closing, then this should be good to merge.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -64,7 +30,7 @@ const CreateIdentityModal: FC<Props> = ({ onClose }) => {
     <Modal
       close={onClose}
       className="create-tls-identity"
-      title={token ? "Identity has been created" : "Generate trust token"}
+      title={`Identity '${identityName}' created`}
```
edlerd, line 33:
```suggestion
      title={`Identity ${identityName} created`}
```
We don't put user provided values in quotes, usually. We sometimes use bold text, but maybe here we can inline it plainly.

edlerd (approved):
LGTM, tiny nit, but also ok to merge as is.

piperdeck:
Hi Nkeiru, I've found a couple things that can be tweaked in this implementation.

There doesn't seem to be enough space between the input and the table. Please make sure both elements have the padding their supposed to have

CleanShot 2025-04-04 at 15 13 40@2x

It seems like this is also an issue in the edit identity side panel.

CleanShot 2025-04-04 at 15 32 55@2x

---

I think this label works better as a simple title "Groups"

CleanShot 2025-04-04 at 15 38 59@2x

---

I think that this change confirmation is useful for editing existing entities, but not really necessary in identity creation.

CleanShot 2025-04-04 at 15 40 36@2x

PR #1167, opened 2025-03-28
fix(instance) mouse position should be scaled if the guest os is not responsive

### LINKED ISSUE #1140: Error repot on mouse position problem
Host OS: Ubuntu 24.04.2
Guest OS: Raspberry Pi Desktop (For PC and Mac)

Mouse did not function right at the position that it should be. It seems it shift about 1 ~ 1.5 cm left what I really point to. Like the photo showing below, when I try to open a Terminal with mouse the File Manager was opened instead.

Image

edlerd:
## Done

- fix mouse position should be scaled if the guest os is not responsive i.e. for custom images like raspberry os or windows.
- When the browser window is smaller than the guest, we now have correct mouse coordinates. fix #1140

Fixes #1140 

## QA

1. Run the LXD-UI:
    - in a small browser window, ensure input events are rightly aligned between your host and the guest os

### DIFF
--- a/src/lib/spice/src/spicemsg.js
+++ b/src/lib/spice/src/spicemsg.js
@@ -944,6 +944,21 @@ SpiceMsgCursorSet.prototype =
     },
 }
 
+function getScaleFactor(sc) {
+    const screen = document.getElementById(sc.screen_id);
+    const screenWidth = screen?.clientWidth;
+    const canvasWidth = screen?.firstChild.width;
+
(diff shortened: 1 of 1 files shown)


Kxiru (approved):
A bit hard to test. I read through the OG issue. On this branch I do not seem to have any offset with mouse clicking. LGTM (?) Perhaps worth double checking in a call.

PR #1171, opened 2025-03-31
feat(acl) add network acl management

edlerd:
## Done

- feat(acl) add network acl management
- feat(network) allow assigning acls to networks

## QA

1. Run the LXD-UI:
    - create network acls

# Screenshots

image

image

image

image

image

image

image

image

image

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -53,10 +53,10 @@ const NetworkAclDetail = lazy(
   async () => import("pages/networks/NetworkAclDetail"),
 );
 const NetworkAclList = lazy(
-  async () => import("./pages/networks/NetworkAclList"),
+  async () => import("pages/networks/NetworkAclList"),
 );
 const NetworkDetail = lazy(async () => import("pages/networks/NetworkDetail"));
-const NetworkList = lazy(async () => import("./pages/networks/NetworkList"));
+const NetworkList = lazy(async () => import("pages/networks/NetworkList"));
 const OperationList = lazy(
   async () => import("pages/operations/OperationList"),
 );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,6 +26,9 @@ const CreateInstance = lazy(
   async () => import("pages/instances/CreateInstance"),
 );
 const CreateNetwork = lazy(async () => import("pages/networks/CreateNetwork"));
+const CreateNetworkAcl = lazy(
+  async () => import("pages/networks/CreateNetworkAcl"),
+);
 const CreateNetworkForward = lazy(
   async () => import("pages/networks/CreateNetworkForward"),
 );
@@ -46,6 +49,12 @@ const InstanceDetail = lazy(
 );
 const InstanceList = lazy(async () => import("pages/instances/InstanceList"));
 const Login = lazy(async () => import("pages/login/Login"));
+const NetworkAclDetail = lazy(
+  async () => import("pages/networks/NetworkAclDetail"),
+);
+const NetworkAclList = lazy(
+  async () => import("./pages/networks/NetworkAclList"),
+);
 const NetworkDetail = lazy(async () => import("pages/networks/NetworkDetail"));
 const NetworkList = lazy(async () => import("./pages/networks/NetworkList"));
 const OperationList = lazy(
@@ -285,6 +294,30 @@ const App: FC = () => {
           element={
             <ProtectedRoute
               outlet={<ProjectLoader outlet={<EditNetworkForward />} />}
+            />
+          }
+        />
+        <Route
+          path="/ui/project/:project/network-acls"
+          element={
+            <ProtectedRoute
+              outlet={<ProjectLoader outlet={<NetworkAclList />} />}
+            />
+          }
+        />
+        <Route
+          path="/ui/project/:project/network-acls/create"
+          element={
+            <ProtectedRoute
+              outlet={<ProjectLoader outlet={<CreateNetworkAcl />} />}
+            />
+          }
+        />
+        <Route
+          path="/ui/project/:project/network-acl/:name"
+          element={
+            <ProtectedRoute
+              outlet={<ProjectLoader outlet={<NetworkAclDetail />} />}
             />
           }
         />
--- a/src/api/network-acls.tsx
+++ b/src/api/network-acls.tsx
@@ -45,7 +45,7 @@ export const fetchNetworkAcl = async (
 };
 
 export const createNetworkAcl = async (
-  networkAcl: Partial<LxdNetworkAcl>,
+  networkAcl: LxdNetworkAcl,
   project: string,
 ): Promise<void> => {
   return new Promise((resolve, reject) => {
@@ -78,11 +78,11 @@ export const renameNetworkAcl = async (
 };
 
 export const updateNetworkAcl = async (
-  networkAcl: Partial<LxdNetworkAcl>,
+  networkAcl: LxdNetworkAcl,
   project: string,
 ): Promise<void> => {
   return new Promise((resolve, reject) => {
-    fetch(`/1.0/network-acls/${networkAcl.name ?? ""}?project=${project}`, {
+    fetch(`/1.0/network-acls/${networkAcl.name}?project=${project}`, {
       method: "PUT",
       body: JSON.stringify(networkAcl),
       headers: {
--- a/src/api/network-acls.tsx
+++ b/src/api/network-acls.tsx
@@ -1,0 +1,110 @@
+import { handleEtagResponse, handleResponse } from "util/helpers";
+import type { LxdNetworkAcl } from "types/network";
+import type { LxdApiResponse } from "types/apiResponse";
+import { withEntitlementsQuery } from "util/entitlements/api";
+
+const networkAclEntitlements = ["can_edit", "can_delete"];
+
+export const fetchNetworkAcls = async (
+  project: string,
+  isFineGrained: boolean | null,
+): Promise<LxdNetworkAcl[]> => {
+  const entitlements = withEntitlementsQuery(
+    isFineGrained,
+    networkAclEntitlements,
+  );
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/network-acls?project=${project}&recursion=1${entitlements}`)
+      .then(handleResponse)
+      .then((data: LxdApiResponse<LxdNetworkAcl[]>) => {
+        resolve(data.metadata);
+      })
+      .catch(reject);
+  });
+};
+
+export const fetchNetworkAcl = async (
+  name: string,
+  project: string,
+  isFineGrained: boolean | null,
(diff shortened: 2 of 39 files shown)


src/pages/networks/NetworkAclUsedBy.tsx
anusha-c18, line 1: We can create a small array to define each category and map over it instead of repeating the same `<tr> ` block

src/pages/networks/forms/EditNetworkAcl.tsx
anusha-c18, line 1: Here we're checking for permissions in two places: using `canEditNetworkAcl` to define `editRestriction` and also defining `readOnly` in the initial values. Can we extract `const canEdit = canEditNetworkAcl(networkAcl)` at the top and reuse it for both `initialValues` and conditional rendering?

src/pages/networks/forms/NetworkAclForm.tsx
```diff
@@ -0,0 +1,265 @@
...
+                  name="description"
+                  label="Description"
+                  placeholder="Enter description"
+                  onBlur={formik.handleBlur}
+                  onChange={(e) => {
+                    ensureEditMode(formik);
+                    formik.handleChange(e);
+                  }}
+                  value={formik.values.description}
+                  disabled={!!formik.values.editRestriction}
```
anusha-c18, line 179: Since we're reusing this condition several times, we could extract it at the top and use it everywhere for better readability:
`const isReadOnly = !!formik.values.editRestriction;`

src/pages/networks/forms/NetworkAclForm.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+  const getRules = (direction: RuleDirection) => {
+    return direction === "ingress"
+      ? formik.values.ingress
+      : formik.values.egress;
+  };
+
+  const getField = (direction: RuleDirection) => {
+    return direction === "ingress" ? "ingress" : "egress";
+  };
```
anusha-c18, line 75: Isn't this function redundant? We're getting back the value we pass to the function, right?

src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+const NetworkAclRuleModal: FC<Props> = ({
+  direction,
+  onClose,
+  onAdd,
+  editRule,
+}) => {
+  const AclRuleSchema = Yup.object().shape({
+    password: Yup.string(),
+  });
```
anusha-c18, line 42: There is no password field in this form, right? Can we remove this schema if it isn't validating any inputs?

src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -0,0 +1,212 @@
...
+          <>
+            <Input
+              id="icmp_code"
+              label="ICMP code"
+              placeholder="Enter ICMP code"
+              type="text"
+              help="Specify the ICMP code number, or leave the value empty for any."
+              onBlur={formik.handleBlur}
+              onChange={formik.handleChange}
+              value={formik.values.icmp_code}
```
anusha-c18, line 193: We could extract shared formik props

src/pages/networks/forms/NetworkAclRuleTable.tsx
```diff
@@ -0,0 +1,120 @@
...
+        { content: "" },
+      ]}
+      rows={rules.map((rule, index) => {
+        const source = rule.source_port
+          ? `${rule.source ?? "*"}:${rule.source_port}`
+          : `${rule.source ?? "*"}`;
+        const destination = rule.destination_port
+          ? `${rule.destination ?? "*"}:${rule.destination_port}`
+          : `${rule.destination ?? "*"}`;
+
```
anusha-c18, line 38: The logic here is repeated in source and destination. We can extract it into a helper function and call it here.

src/pages/networks/forms/NetworkAclSelector.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <MultiSelect
+      items={toOptionList(availableAcls.map((acl) => acl.name))}
+      disabled={!hasAcls}
+      selectedItems={toOptionList(selectedAcls)}
+      variant={hasAcls ? "condensed" : "search"}
+      placeholder={hasAcls ? "Select ACLs" : "No ACLs available"}
+      onItemsUpdate={(items) => {
+        setSelectedAcls(items.map((item) => item.label));
```
anusha-c18, line 37:
```suggestion
        setSelectedAcls(items.map((item) => item.value));
```

Although it's the same it's probably more clearer?

src/App.tsx
```diff
@@ -46,6 +49,12 @@ const InstanceDetail = lazy(
 );
 const InstanceList = lazy(async () => import("pages/instances/InstanceList"));
 const Login = lazy(async () => import("pages/login/Login"));
+const NetworkAclDetail = lazy(
+  async () => import("pages/networks/NetworkAclDetail"),
+);
+const NetworkAclList = lazy(
+  async () => import("./pages/networks/NetworkAclList"),
+);
```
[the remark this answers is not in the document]
edlerd: Good catch, yes we should use absolute ones here.

src/api/network-acls.tsx
```diff
@@ -0,0 +1,110 @@
...
+  );
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/network-acls?project=${project}&recursion=1${entitlements}`)
+      .then(handleResponse)
+      .then((data: LxdApiResponse<LxdNetworkAcl[]>) => {
+        resolve(data.metadata);
+      })
+      .catch(reject);
+  });
+};
```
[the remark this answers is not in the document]
edlerd: I agree this can be simplified. All the api functions in this folder use this approach of wrapping the fetch call with a custom promise. It is handy especially when dealing with bulk actions, or to add custom error handling, such as in `networks.tsx`.

Maybe we should change this in files like this one, where there is no need for the wrapping promise. For now, I'd keep it as is for consistency in the api layer. I'll log a ticket to have this pattern reviewed and eventually simplified.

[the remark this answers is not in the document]
edlerd: https://warthogs.atlassian.net/browse/WD-21052

src/api/network-acls.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+export const updateNetworkAcl = async (
+  networkAcl: Partial<LxdNetworkAcl>,
+  project: string,
+): Promise<void> => {
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/network-acls/${networkAcl.name ?? ""}?project=${project}`, {
+      method: "PUT",
```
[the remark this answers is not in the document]
edlerd: Good point, better solution is IMHO to have a stricter the input type, removing the partial. Doing that below.

src/components/Navigation.tsx
```diff
@@ -32,6 +32,7 @@ const isSmallScreen = () => isWidthBelow(620);
 const initialiseOpenNavMenus = (location: Location) => {
   const openPermissions = location.pathname.includes("permissions");
   const openStorage = location.pathname.includes("storage");
+  const openNetwork = location.pathname.includes("network");
```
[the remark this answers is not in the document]
edlerd: Yeah, true, I can make the check a bit stricter for network and the other two as well.

src/pages/networks/NetworkAclUsedBy.tsx
edlerd: I think keeping this explicit is slightly more readable that making it generic.

src/pages/networks/forms/EditNetworkAcl.tsx
edlerd: I am not sure if I understand correctly. I think `canEditNetworkAcl` is only evaluated once, why add it to the top?
There is a difference between the `readOnly` and `editRestriction`:
- `readOnly` is for the edit mode, when a user lands on the form. They will see the values with pencil icon next to them. Once any pencil is clicked, they are in edit mode and see form elements as well as submit/cancel buttons in the footer.
- `editRestriction` is a permission check. If a user has no permission, they can never flip into write mode, hence we disable the pencil icons or other elements to manipulate the config for them. They can still browse the read only presentation of the config.

src/pages/networks/forms/NetworkAclForm.tsx
edlerd: Yeah, we can do that, good idea.

src/pages/networks/forms/NetworkAclForm.tsx
edlerd: Yes, absolutely. I was running into type error while developing, but now it works well without.

src/pages/networks/forms/NetworkAclRuleModal.tsx
edlerd: Yes, this can be removed.

src/pages/networks/forms/NetworkAclRuleModal.tsx
edlerd: Can you please clarify? I don't understand what you mean here.

src/pages/networks/forms/NetworkAclRuleTable.tsx
edlerd: Yes, very good idea.

src/pages/networks/forms/NetworkAclSelector.tsx
edlerd: The typesystem is bamboozled by value, because it can be numeric. But we can cast it here. Agree value is the better option.

edlerd:
Thanks for the review @anusha-c18 All issues should have been resolved or a discussion started. Please have another pass at the changes.

src/pages/networks/forms/NetworkAclRuleModal.tsx
anusha-c18: We can use `getFieldProps()` in these inputs to reduce boilerplate code. Here's an example of how it is used in [anbox](https://github.com/canonical/ams/blob/172d065cd52f8c8432f80075183bfb7e9c491b75/ui/templates/js/pages/instances/form/InstanceMainConfig.tsx#L247).

edlerd: Ah right, got it, that's a good idea.

edlerd:
Added some tests to this one.

src/pages/networks/forms/EditNetworkAcl.tsx
anusha-c18: This clears it up, thanks!

anusha-c18 (dismissed):
Thanks for the changes! LGTM 👍

PR #1174, opened 2025-04-03
fix(group) allow to add permissions when creating a group

edlerd:
## Done

- fix(group) allow to add permissions when creating a group

## QA

1. Run the LXD-UI:
    - Create a group and add permissions right while creating it.

### DIFF
--- a/src/pages/permissions/panels/EditGroupPermissionsForm.tsx
+++ b/src/pages/permissions/panels/EditGroupPermissionsForm.tsx
@@ -43,8 +43,9 @@ const EditGroupPermissionsForm: FC<Props> = ({
   const [search, setSearch] = useState("");
   const { canViewPermissions } = useServerEntitlements();
   const { canEditGroup } = useGroupEntitlements();
+
   const getEditRestriction = () => {
-    if (!canEditGroup(group)) {
+    if (group && !canEditGroup(group)) {
       return "You do not have permission to edit this group";
     }
 
--- a/tests/helpers/permission-groups.ts
+++ b/tests/helpers/permission-groups.ts
@@ -130,14 +130,21 @@ export const createGroup = async (
   page: Page,
   groupName: string,
   description: string,
+  withPermissions = false,
 ) => {
   await visitGroups(page);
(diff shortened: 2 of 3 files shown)


tests/helpers/permission-groups.ts
```diff
@@ -130,15 +130,23 @@ export const createGroup = async (
...
+
+  if (withPermissions) {
+    await page.getByRole("button", { name: "Add permissions" }).click();
+    await addPermission(page, "Server", "server", "admin");
+  }
+
   await page
     .getByLabel("Side panel")
     .getByRole("button", { name: "Create group" })
+    .last()
```
anusha-c18, line 149: We should probably use a more specific selector as this comes with the risk of breaking if the contents change in the future

src/pages/permissions/panels/EditGroupPermissionsForm.tsx
```diff
@@ -44,7 +44,7 @@ const EditGroupPermissionsForm: FC<Props> = ({
   const { canViewPermissions } = useServerEntitlements();
   const { canEditGroup } = useGroupEntitlements();
   const getEditRestriction = () => {
-    if (!canEditGroup(group)) {
+    if (group && !canEditGroup(group)) {
       return "You do not have permission to edit this group";
```
anusha-c18, line 48: The return message could be updated to reflect both conditions

anusha-c18 (changes requested):

tests/helpers/permission-groups.ts
edlerd: Yes, right. We can reference the side panel footer here to be more specific.

src/pages/permissions/panels/EditGroupPermissionsForm.tsx
edlerd: I think the message is correct. This is to guard this check against creation of a group. In that case the group object is undefined, and we shouldn't check if we can edit at all. Hence, introducing the 2nd condition to fix this.

anusha-c18 (approved):
LGTM 👍

PR #1175, opened 2025-04-04
fix(layout) adjust pages for responsiveness

edlerd:
## Done

- make list pages look okay on phone and tablet sizes, by hiding columns
- fix loading flickering on small screens
- remember menu open state on small screens with navigation drawer
- hide search input on very small screen sizes

## QA

1. Run the LXD-UI:
    - check list pages from main navigation on phone sized screen.


anusha-c18 (changes requested):
Are we planning to fully replace `Loader` with `ContentPageLoader` for early returns in pages? If that's the case, there are still quite a few components where `Loader` is being used and could be updated. 

From a quick search, I found a couple of examples in components like NetworkListTable, CpuLimitInput, DiskDeviceForm, GPUDeviceForm, etc.

edlerd:
> Are we planning to fully replace `Loader` with `ContentPageLoader` for early returns in pages? If that's the case, there are still quite a few components where `Loader` is being used and could be updated.
> 
> From a quick search, I found a couple of examples in components like NetworkListTable, CpuLimitInput, DiskDeviceForm, GPUDeviceForm, etc.

The differentiation is intentional. Only components that are the main part of the application layout need the content page wrapper. This is a fix for the mobile view, without it, the loader is moving into the main navigation bar at the top leading to a broken layout for small screens.

For the other cases where we load in nested components, we should use the loader directly and don't need the wrapper around it.

I am not super happy with the current solution either, for exactly the reason you mentioned. It is not very clear when to use which loader type. But I am lacking an idea how to structure it better. Maybe you have an idea how to improve it?

anusha-c18:
> Maybe you have an idea how to improve it?

One option could be to pass a prop to the existing `Loader` component that conditionally applies a class. This class could set `position: relative` and adjust its layout behavior for smaller screen sizes, wdyt?

edlerd:
> > Maybe you have an idea how to improve it?
> 
> One option could be to pass a prop (something like `isPartOfMainLayout`) to the existing `Loader` component. This prop could either:
> 
>     1. Conditionally apply a class. This class could set `position: relative` and adjust its layout for mobile view.
> 
>     2. Or alternatively return the loader wrapped in the `CustomLayout` component when the prop is true.
> 
> 
> Wdyt?

Yeah, keeping one component with additional prop makes it clearer. I'll adjust it .

edlerd:
Adjusted the loader component to align with your idea @anusha-c18

anusha-c18 (approved):
We could also add a comment to explain the prop but besides that, LGTM!

anusha-c18 (approved):
Thanks, lgtm 😄

PR #1182, opened 2025-04-09
feat: [WD-20438] Select network on project creation

Kxiru:
## Done

- On project creation, add a network selector for the “default network” to be set on the default profile in the new project. Similar to the default storage pool.
- Options should be all networks in the default project,
- A a “none” option should be includued.
- Default selected should be the network in the default profile of the default project.
- If the user choses features > custom > networks to isolate networks, set default network in the form to none and disable the selector.

## QA

1. Run the LXD-UI:

## Screenshots
On creation (form)
image
On creation (overview)
image

### DIFF
--- a/src/pages/projects/CreateProject.tsx
+++ b/src/pages/projects/CreateProject.tsx
@@ -5,7 +5,11 @@ import { useFormik } from "formik";
 import * as Yup from "yup";
 import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
-import { checkDuplicateName, getDefaultStoragePool } from "util/helpers";
+import {
+  checkDuplicateName,
+  getDefaultNetwork,
+  getDefaultStoragePool,
+} from "util/helpers";
 import { useNavigate } from "react-router-dom";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import useEventListener from "util/useEventListener";
@@ -101,6 +105,9 @@ const CreateProject: FC = () => {
       default_instance_storage_pool: defaultProjectDefaultProfile
         ? getDefaultStoragePool(defaultProjectDefaultProfile)
         : "",
+      default_project_network: defaultProjectDefaultProfile
+        ? getDefaultNetwork(defaultProjectDefaultProfile)
+        : "",
     },
     enableReinitialize: true,
     validationSchema: ProjectSchema,
@@ -122,6 +129,8 @@ const CreateProject: FC = () => {
         values.features_storage_buckets = undefined;
       }
 
+      const hasNetwork = values.default_project_network !== "none";
+
       createProject(
         JSON.stringify({
           ...projectDetailPayload(values),
@@ -134,7 +143,7 @@ const CreateProject: FC = () => {
       )
         .then(async () => {
           if (
-            !values.default_instance_storage_pool ||
+            (!values.default_instance_storage_pool && !hasNetwork) ||
             values.features_profiles === false
           ) {
             notifySuccess(values);
@@ -151,6 +160,13 @@ const CreateProject: FC = () => {
               pool: values.default_instance_storage_pool,
               type: "disk",
             },
+            ...(hasNetwork && {
+              eth0: {
+                name: "eth0",
+                network: values.default_project_network,
+                type: "nic",
+              },
+            }),
           };
 
           updateProfile(profile, values.name)
@@ -160,7 +176,7 @@ const CreateProject: FC = () => {
             .catch((e: Error) => {
               navigate(`/ui/project/${values.name}/instances`);
               toastNotify.failure(
-                `Successfully created ${profile.name}, Failed to attach storage pool`,
+                `Successfully created ${values.name} project. Failed to attach storage pool${hasNetwork ? " and network" : ""}.`,
                 e,
               );
             });
--- a/src/pages/projects/forms/NetworkSelector.tsx
+++ b/src/pages/projects/forms/NetworkSelector.tsx
@@ -1,0 +1,61 @@
+import { Select } from "@canonical/react-components";
+import type { SelectProps } from "@canonical/react-components";
+import type { FC } from "react";
+import { useNetworks } from "context/useNetworks";
+
+interface Props {
+  value: string;
+  project: string;
+  setValue: (value: string) => void;
+  onBlur?: (e: React.FocusEvent) => void;
+  hasNoneOption?: boolean;
+}
+
+const NetworkSelector: FC<Props & SelectProps> = ({
+  value,
+  project,
+  setValue,
+  onBlur,
+  hasNoneOption = false,
+  ...selectProps
+}) => {
+  const { data: networks = [] } = useNetworks(project);
+
+  const managedNetworks = networks.filter((network) => network.managed);
+
+  const getNetworkOptions = () => {
+    const options = managedNetworks.map((network) => {
+      return {
+        label: network.name,
+        value: network.name,
+        disabled: false,
+      };
+    });
+
+    options.unshift({
+      label: options.length === 0 ? "No networks available" : "Select option",
+      value: "",
+      disabled: true,
+    });
+
+    if (hasNoneOption) {
+      options.push({ label: "No network", value: "none", disabled: false });
+    }
+
+    return options;
+  };
+  return (
+    <Select
+      label="Network"
+      onChange={(e) => {
+        setValue(e.target.value);
+      }}
+      onBlur={onBlur}
+      value={value}
+      {...selectProps}
+      options={getNetworkOptions()}
+    />
+  );
+};
+
+export default NetworkSelector;
--- a/src/pages/projects/forms/ProjectDetailsForm.tsx
+++ b/src/pages/projects/forms/ProjectDetailsForm.tsx
@@ -20,6 +20,7 @@ import type { LxdConfigPair } from "types/config";
 import { ensureEditMode } from "util/instanceEdit";
 import StoragePoolSelector from "pages/storage/StoragePoolSelector";
 import ResourceLink from "components/ResourceLink";
+import NetworkSelector from "./NetworkSelector";
 
 export interface ProjectDetailsFormValues {
   name: string;
@@ -35,6 +36,7 @@ export interface ProjectDetailsFormValues {
   readOnly: boolean;
   entityType: "project";
   editRestriction?: string;
+  default_project_network: string;
 }
 
 export const projectDetailPayload = (
@@ -117,6 +119,10 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
   const hadFeaturesNetwork = project?.config["features.networks"] === "true";
   const hadFeaturesNetworkZones =
     project?.config["features.networks.zones"] === "true";
+  const hasNoProfiles =
+    features === "customised" && !formik.values.features_profiles;
+  const hasIsolatedNetworks =
+    features === "customised" && formik.values.features_networks;
 
   return (
     <ScrollableForm>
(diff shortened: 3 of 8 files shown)


src/util/projectEdit.tsx
```diff
@@ -108,7 +111,7 @@ export const getProjectPayload = (
   values: ProjectFormValues,
 ): Partial<LxdProject> => {
   const handledConfigKeys = getProjectConfigKeys();
-  const handledKeys = new Set(["name", "description", "config"]);
+  const handledKeys = new Set(["name", "description", "network", "config"]);
```
edlerd, line 114: We shouldn't need to add the network here. It is stored on the profile entity in the project, not in the project itself.

src/pages/projects/CreateProject.tsx
```diff
@@ -135,6 +142,7 @@ const CreateProject: FC = () => {
         .then(async () => {
           if (
             !values.default_instance_storage_pool ||
+            !values.default_project_network ||
             values.features_profiles === false
```
edlerd, line 146: This condition is not correct. We should ensure both -- storage and network -- are unset, not one or the other for the early exit.
```suggestion
            (!values.default_instance_storage_pool &&
            !values.default_project_network) ||
            values.features_profiles === false
```

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -179,6 +181,34 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
+                      to={`/ui/project/${project?.name}/profile/default`}
+                    />{" "}
+                    {" profile"}
+                  </>
+                ) : (
+                  ""
+                ),
+              }}
+            />
+          </div>
```
edlerd, line 211: The extra div wrapper should not be necessary.
```suggestion
            <NetworkSelectorForm
              value={formik.values.default_project_network}
              setValue={(value) =>
                void formik.setFieldValue("default_project_network", value)
              }
              selectProps={{
                label: "Default network",
                disabled:
                  (formik.values.features_profiles === false &&
                    features === "customised") ||
                  isEdit,
                help: isEdit ? (
                  <>
                    Edit the network in the{" "}
                    <ResourceLink
                      type="profile"
                      value="default"
                      to={`/ui/project/${project?.name}/profile/default`}
                    />{" "}
                    {" profile"}
                  </>
                ) : (
                  ""
                ),
              }}
            />
```

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -179,6 +181,34 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
+              value={formik.values.default_project_network}
+              setValue={(value) =>
+                void formik.setFieldValue("default_project_network", value)
+              }
+              selectProps={{
+                label: "Default network",
+                disabled:
+                  (formik.values.features_profiles === false &&
+                    features === "customised") ||
+                  isEdit,
```
edlerd, line 195: This should also disable, if features_networks is selected. Because then the project will not have access to the networks in the default project, so we should fall back to no network in the new project.

src/pages/projects/forms/ProjectDetailsForm.tsx
edlerd, line 1: Found the unrelated bug on this page: The "Click the name in the header to rename the project" help text on the project name input should not show during creation. Can you please make it conditional?

src/pages/projects/forms/NetworkSelectorForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+import type { FC } from "react";
+import { useNetworks } from "context/useNetworks";
+
+interface Props {
+  value: string;
+  setValue: (value: string) => void;
+  selectProps?: SelectProps;
+}
+
+const NetworkSelectorForm: FC<Props> = ({ value, setValue, selectProps }) => {
```
edlerd, line 12:
```suggestion
const NetworkSelector: FC<Props> = ({ value, setValue, selectProps }) => {
```
I think `NetworkSelector` is more aligned with other component names and what this component is intended for.

src/pages/projects/forms/NetworkSelectorForm.tsx
edlerd, line 1: Can we use this component in `NetworkDevicesForm` and remove the duplication from there?

edlerd (commented):
Code looks decent already, and testing it works well.

As you mentioned already, some edge cases need to be addressed. I added comments on that and some other aspects.

src/util/projectEdit.tsx
Kxiru: Yep, thanks, this was a remnant of some other experimentation. Thank you for pointing this out :sob:

src/pages/projects/CreateProject.tsx
edlerd: Best adjust the logic below to cater for the case that network can be "none" and avoid setting it in the payload in that case.

src/pages/projects/forms/ProjectDetailsForm.tsx
Kxiru: Nice spot

src/pages/projects/CreateProject.tsx
Kxiru: Ah, that's a great solution to what we were discussing earlier, thank you!

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -179,6 +182,34 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
+            setValue={(value) =>
+              void formik.setFieldValue("default_project_network", value)
+            }
+            hasNone
+            selectProps={{
+              label: "Default profile network",
+              disabled:
+                (features === "customised" &&
+                  (!formik.values.features_profiles ||
+                    formik.values.features_networks)) ||
```
edlerd, line 196: Can we isolate this into a const? Something like
```
const hasNoProfiles = features === "customised" && !formik.values.features_profiles;
const hasIsolatedNetworks = features === "customised" && formik.values.features_networks;
```
Then in here we can simplify the check to `hasNoProfiles || hasIsolatedNetworks || isEdit` which is much more readable.

edlerd (commented):
Functionality works great, good stuff.

Some ideas to simplify or make things easier to read and improve on the error messages wording below.

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -179,6 +186,31 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
               ),
             }}
           />
+          <NetworkSelector
+            value={formik.values.default_project_network}
+            project={project ? project.name : "default"}
```
edlerd, line 191:
```suggestion
            project={"default"}
```
We should always use the networks from the default project for project creation.

src/pages/projects/CreateProject.tsx
```diff
@@ -134,12 +141,14 @@ const CreateProject: FC = () => {
       )
         .then(async () => {
           if (
-            !values.default_instance_storage_pool ||
+            (!values.default_instance_storage_pool &&
+              !values.default_project_network) ||
```
edlerd, line 145:
```suggestion
            (!values.default_instance_storage_pool && !hasNetwork) ||
```

I suspect we should expect the "none" network value here, not empty value.

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const handleNetworkChange = (value: string) => {
+    setValue(value);
+  };
+
+  return (
+    <Select
+      label="Network"
+      onChange={(e) => {
+        handleNetworkChange(e.target.value);
```
edlerd, line 57:
```suggestion
        setValue(e.target.value);
```
We can remove the extra callback.

edlerd (commented):
QA LGTM, some small comments below.

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -179,6 +186,31 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
               ),
             }}
           />
+          <NetworkSelector
+            value={formik.values.default_project_network}
+            project={"default"}
```
edlerd, line 191:
```suggestion
            project="default"
```
A bit simpler.

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -179,6 +186,31 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
+              disabled: hasNoProfiles || hasIsolatedNetworks || isEdit,
+              help: isEdit ? (
+                <>
+                  Configure networks in the{" "}
+                  <ResourceLink
+                    type="profile"
+                    value="default"
+                    to={`/ui/project/${project?.name}/profile/default`}
+                  />{" "}
+                  {" profile"}
```
edlerd, line 207:
```suggestion
                  profile
```
this would create two spaces.

tests/helpers/projects.ts
```diff
@@ -15,6 +15,9 @@ const openProjectCreationForm = async (page: Page) => {
 
   await expect(page.getByText("Project name")).toBeVisible();
   await expect(page.getByText("Loading storage pools")).not.toBeVisible();
+
+  //Set default profile network to None
+  await page.getByLabel("Default profile network").selectOption("No network");
```
edlerd, line 20:
```suggestion
  await page.getByLabel("Default profile network").selectOption("No network");
```
I think the comment is redundant considering the line of code below, wdyt?

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -280,6 +312,10 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
                       "features_networks",
                       !formik.values.features_networks,
                     );
+
+                    if (!formik.values.features_networks && !isEdit) {
+                      formik.setFieldValue("default_project_network", "none");
+                    }
```
edlerd, line 318:
```suggestion
                    const featuresNetworks = !formik.values.features_networks;
                    formik.setFieldValue("features_networks", featuresNetworks);
                    if (featuresNetworks && !isEdit) {
                      formik.setFieldValue("default_project_network", "none");
                    }
```
I think we are introducing a race here with updating the value and immediately reading it in the line after. Better capture the new value and use it twice to avoid the race. This seems also easier to read.

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const NetworkSelector: FC<Props> = ({
+  value,
+  project,
+  setValue,
+  selectProps,
+  onBlur,
+  hasNoneOption = false,
+}) => {
```
edlerd, line 22:
```suggestion
  setValue: (value: string) => void;
  onBlur?: (e: React.FocusEvent) => void;
  hasNoneOption?: boolean;
}

const NetworkSelector: FC<Props & SelectProps> = ({
  value,
  project,
  setValue,
  onBlur,
  hasNoneOption = false,
  ...selectProps
}) => {
```
I think if we inline the select props, the usage will be simpler, we don't need to nest the select props, and can directly use:

```
                      <NetworkSelector
                        ...
                        id={`devices.${index}.network`}
                        name={`devices.${index}.network`}
                        ...
```

edlerd (commented):
QA LGTM, good work :)

Some ideas to simplify and tiny nitpicks.

src/pages/projects/forms/NetworkSelector.tsx
Kxiru: This took me a while to understand but I think I've got it now, please review the changes!

edlerd (dismissed):
code LGTM, thanks for the iteration :)

edlerd:
Tiny QA issue: When a project has no network on the default project, the selector says "select a network", but should in that case say "No network" instead. Because the selector is disabled.

image

edlerd (approved):
LGTM, thanks for the QA improvement :)

PR #1198, opened 2025-04-22
fix(test) ensure project test is setting a valid cluster group targeting value

edlerd:
## Done

- fix(test) ensure project test is setting a valid cluster group targeting value


Kxiru (approved):

Kxiru (approved):
LGTM, Thank you for finding the cause.

PR #1200, opened 2025-04-23
fix(instances) disable ssh key buttons for users without edit instance permission

edlerd:
## Done

- fix(instances) disable ssh key buttons for users without edit instance permission

## QA

1. Run the LXD-UI:
    - given a user that has only view permission on an instance, but not edit view the instance detail page and ensure the ssh key buttons are disabled.


Kxiru (approved):
LGTM!

PR #1202, opened 2025-04-23
fix(forms) show browser warning when leaving an edit form with pending changes

edlerd:
## Done

- fix(forms) show browser warning when leaving an edit form with pending changes

Fixes WD-20455

## QA

1. Run the LXD-UI:
    - edit a resource (instance, profile, network, network acl, project, storage pool, storage volume) then leave the site to another page or reload the page. A warning should appear, prompting to confirm losing pending changes.

## Screenshots

image

### DIFF
--- a/src/components/forms/FormSubmitBtn.tsx
+++ b/src/components/forms/FormSubmitBtn.tsx
@@ -3,15 +3,37 @@ import { ActionButton } from "@canonical/react-components";
 import { pluralize } from "util/instanceBulkActions";
 import type { ConfigurationRowFormikProps } from "components/ConfigurationRow";
 import { getFormChangeCount } from "util/formChangeCount";
+import { unstable_usePrompt as usePrompt } from "react-router";
+import useEventListener from "util/useEventListener";
 
 interface Props {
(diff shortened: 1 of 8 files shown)


Kxiru (approved):
Ah, we definitely needed one of these! LGTM!
image

Not sure if this can be done here, but I have seen on some websites that this can also be triggered when a user tries to go to a new tab or even touches the bookmarks bar. Although this implementation is fine (triggers when the windows attempts to close or when the page is navigated away from, internally), I wonder if this is possible?

Perhaps not worth looking into though as eventually it can become obstructive to the user. Seems good to go to me!

edlerd:
> Not sure if this can be done here, but I have seen on some websites that this can also be triggered when a user tries to go to a new tab or even touches the bookmarks bar. Although this implementation is fine (triggers when the windows attempts to close or when the page is navigated away from, internally), I wonder if this is possible?
> 
> Perhaps not worth looking into though as eventually it can become obstructive to the user. Seems good to go to me!

Yeah, I wouldn't want to bug users that open a new tab or bookmarks. They might just search for information to fill the form and the form state will be preserved until they come back.

PR #1204, opened 2025-04-23
feat(metrics) show all disk metrics and label main disk metric to root disk

### LINKED ISSUE #1194: disk metrics show usage&size of root partition, not the full disk
The disk metrics in overview page and instance list only show the usage and size of the root partition of the guest, not the full disk usage.

This looks real silly if you have a lot of partitions in your guest, showing the disk is 100% full and only 229MB

Image

while in reality it has a lot of partitions:

Image

and the zfs block device is 36G big with 11G in use

```
zfs get all lxd/virtual-machines/HomeassistantOS.block | grep -e volsize -e used
lxd/virtual-machines/HomeassistantOS.block  used                  11.3G  
lxd/virtual-machines/HomeassistantOS.block  volsize               36G 
lxd/virtual-machines/HomeassistantOS.block  usedbysnapshots       9.08G  
lxd/virtual-machines/HomeassistantOS.block  usedbydataset         2.18G   
lxd/virtual-machines/HomeassistantOS.block  usedbychildren        0B  
lxd/virtual-machines/HomeassistantOS.block  usedbyrefreservation  0B   
lxd/virtual-machines/HomeassistantOS.block  logicalused           23.3G    
```

I noticed this issue with my HomeAssistant VM so I created a new ubuntu 24.04 VM through custom iso with the default 30GB disk. Gave it a 15G root partition, finished install, installed lxd agent through virtiofs mount and can confirm it shows as 15G in the ui

Image

Image

edlerd:
## Done

- feat(metrics) show all disk metrics and label main disk metric to root disk
- feat(metrics) add cpu metrics, rework metric parsing functions and types

Fixes #1194

## QA

1. Run the LXD-UI:
    - browse instance detail page for a running instance, ensure the disk metrics can be expanded and are available. ensure cpu metrics are available and seem reasonable.

## Screenshots

instance container

image

instance container with custom volume

image

virtual machine with expanded filesystems

image

instance overview

image


Kxiru (changes requested):
I see this odd behaviour where if you wait on the instance overview page, the CPU meter will complete it's countdown and the meter will display, however if you navigate to a new tab or page from there, and go back to the overviewpage, the CPU  meter will load again, "loading, X seconds remaining to **first** CPU snapshot". Effectively, the meter is not persistent and/or the values of the meter are not being updated correctly. Its as though the component is being rendered every X seconds as opposed to updated every X seconds.
 
image
image

Kxiru:
Screenshots supporting work done:
Before expansion:
image

After expansion:
image

edlerd:
> I see this odd behaviour where if you wait on the instance overview page, the CPU meter will complete it's countdown and the meter will display, however if you navigate to a new tab or page from there, and go back to the overviewpage, the CPU meter will load again with the line "loading, X seconds remaining to **first** CPU snapshot".
> 
> Effectively, the meter is not persistent and/or the values of the meter are not being updated correctly. It is as though the component is being rendered every X seconds as opposed to being updated every X seconds.

Yeah, fair point. This was due to the CPU metric history being local to that CPU metric component. So each time it got unmounted we lost all history. I now introduced a global metric history, so we don't have to wait when switching between instances or between the instance detail page tabs.

@Kxiru please check QA for the revised logic.

edlerd:
> Thank you David, the CPU usage now maintains, but I have also seen this inconsistency in spacing between the first and remaining meters. Could this also be addressed?

That was intentional between the three types memory, disk and CPU. The disk meters are a bit closer as they are one group.

edlerd:
After second thought, I think the unified spacing looks better indeed. Adjusted it according to your suggestion @Kxiru

Kxiru (approved):
I think this looks, feels and executes very well, now. Thanks, David!

PR #1209, opened 2025-04-25
feat: [WD-15524] Add target member selection for volume duplication

Kxiru:
## Done

- Added cluster member selector which conditionally renders whenever the storage pool is not remote.
- To do: (Testing) selecting a cluster member does not duplicate the volume in the correct cluster. Troubleshoot why.
- To do: (Discovery) Identify all remote storage drivers to be included in the list.

## QA

1. Run the LXD-UI:
    - Navigate to the storage volumes page and identify a custom storage volume.
    - Navigate to custom storage volume overview page and attempt to duplicate the volume using the button in the top right.
    - 

## Screenshots

image

### DIFF
--- a/src/pages/storage/forms/DuplicateVolumeForm.tsx
+++ b/src/pages/storage/forms/DuplicateVolumeForm.tsx
@@ -20,6 +20,9 @@ import ResourceLink from "components/ResourceLink";
 import { useProjects } from "context/useProjects";
 import { useLoadCustomVolumes } from "context/useVolumes";
 import { duplicateStorageVolume } from "api/storage-volumes";
+import ClusterMemberSelector from "pages/cluster/ClusterMemberSelector";
+import { useStoragePool } from "context/useStoragePools";
+import { isRemoteStorage } from "util/storageOptions";
 
 interface Props {
   volume: LxdStorageVolume;
@@ -31,6 +34,7 @@ export interface StorageVolumeDuplicate {
   project: string;
   pool: string;
   copySnapshots: boolean;
+  location: string;
 }
 
 const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
@@ -40,7 +44,7 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
   const eventQueue = useEventQueue();
 
   const { data: projects = [], isLoading: projectsLoading } = useProjects();
-
+  const { data: pool } = useStoragePool(volume.pool);
   const { data: volumes = [], isLoading: volumesLoading } =
     useLoadCustomVolumes(volume.project);
 
@@ -102,6 +106,7 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
       project: volume.project,
       copySnapshots: true,
       pool: volume.pool,
+      location: isRemoteStorage(pool?.driver ?? "") ? "" : volume.location,
     },
     enableReinitialize: true,
     validationSchema,
@@ -133,10 +138,12 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
         />
       );
 
-      const targetProject =
-        values.project === formik.initialValues.project ? "" : values.project;
-
-      duplicateStorageVolume(payload, values.pool, targetProject)
+      duplicateStorageVolume(
+        payload,
+        values.pool,
+        values.project,
+        values.location,
+      )
         .then((operation) => {
           toastNotify.info(
             <>Duplication of volume {existingVolumeLink} started.</>,
@@ -207,6 +214,9 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
             label: "Storage pool",
           }}
         />
+        {!isRemoteStorage(pool?.driver ?? "") && (
+          <ClusterMemberSelector {...formik.getFieldProps("location")} />
+        )}
         <Select
           {...formik.getFieldProps("project")}
           id="project"
--- a/src/util/storageOptions.tsx
+++ b/src/util/storageOptions.tsx
@@ -70,3 +70,7 @@ export const driversWithFilesystemSupport = [
   cephDriver,
   pureStorage,
 ];
+
+export const isRemoteStorage = (driver: string) => {
+  return [cephDriver, cephFSDriver, powerFlex].includes(driver);
+};


src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -40,10 +42,15 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
   const eventQueue = useEventQueue();
 
   const { data: projects = [], isLoading: projectsLoading } = useProjects();
-
+  const { data: pool } = useStoragePool(volume.pool);
   const { data: volumes = [], isLoading: volumesLoading } =
     useLoadCustomVolumes(volume.project);
 
+  const isRemoteStorage = ["ceph", "cephfs", "powerflex"].includes(
```
edlerd, line 49: This should reuse constants from `src/util/storageOptions.tsx`. Also might be good to have this array and check in the util file instead of keeping it local in this component.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -40,10 +42,15 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
...
+    pool?.driver ?? "",
+  );
+  const [member, setMember] = useState("");
```
edlerd, line 52: This should be kept in formik state, not in local state.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -136,7 +143,7 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
       const targetProject =
         values.project === formik.initialValues.project ? "" : values.project;
 
-      duplicateStorageVolume(payload, values.pool, targetProject)
+      duplicateStorageVolume(payload, values.pool, targetProject, member)
```
edlerd, line 146: There is a comment in the api function
```
// Including project and target params if they did not change from source configs breaks the API call.
// Therefore, we only include them if they are different from the source configs, that's why both project and target are optional inputs
```
Can you validate the statement in the comment is still correct?

If yes, I think we should set the member here to empty string or null if the value is equal to the initial value. Kind of like the `targetProject` calculation above.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -207,6 +214,13 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
             label: "Storage pool",
           }}
         />
+        {!isRemoteStorage && (
+          <ClusterMemberSelector
+            setMember={(member) => {
+              setMember(member);
```
edlerd, line 220: this should update the formik state, also initialize it to the location of the volume when the modal opens.

edlerd (changes requested):
When selecting a member, the selection is not reflected in the API call and as such also not in the duplicated volume

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -102,6 +106,7 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
       project: volume.project,
       copySnapshots: true,
       pool: volume.pool,
+      location: "",
```
edlerd, line 109:
```suggestion
      location: isRemoteStorage(pool?.driver ?? "") ? "" : volume.location,
```
We need to initialize this for cluster member local volumes. Otherwise the cluster member selector will show the wrong default value on opening the duplication modal.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -118,7 +123,7 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
           pool: volume.pool,
           volume_only: !values.copySnapshots,
           // logic from the lxc source code.
-          // We should not set source.project if target project is the same as the source project
+          // We should set source.project or source.location only if the target project/location is NOT the same as the source project
           project:
             values.project !== volume.project ? volume.project : undefined,
           location: volume.location,
```
edlerd, line 129: The changed comment seems to be wrong. We set the location regardless of a change below.

edlerd (changes requested):
Tested it, and it works well now, good work :+1: 

Code LGTM to be, just one comment on a comment that seems not aligned.

Found one QA issue below, where initialization is missing.

src/pages/storage/forms/DuplicateVolumeForm.tsx
Kxiru: I don't think this is necessary as the selector already only renders when the Pool driver is not remote storage, so rendering a conditional based on this will never be visible. Am I missing something else?

edlerd: The issue is with a custom volume on a non-remote pool. So the cluster member selector will be visible and show the first cluster member alphabetically. But the volume could live on the 2nd or 3rd or any other cluster member. When submitting the form without changing the member, the member of the volume will be used. So the selected member when launching the modal must match the volume node.

edlerd: https://github.com/user-attachments/assets/4018c915-d709-4b5c-9ce9-c548875657d3

Kxiru: In fact, most cases having anything other than "" in this field results in the following error when attempting to select the default value:

image

Whilst the code suggestion works for other values, it does not work when selecting the default value. I believe this has something to do with the case where the source location should only be set if the target member is not the same as the source, otherwise conflicts arise. Therefore, having a starting value of "" eliminates this.

Please advise how you would like me to proceed.

edlerd: The suggestions above are correct. The issue is with the lines below, simplify them to this to make it succeed:
```
      const targetProject = values.project;
      const targetMember = values.location;
```

Kxiru: This works! Thank you for the correction.

edlerd (changes requested):
When doing QA I found [this bug](https://github.com/canonical/lxd/issues/15475), but it shouldn't block this PR.

QA otherwise LGTM, code as well, just a tiny issue with the changed comment below.

src/pages/storage/forms/DuplicateVolumeForm.tsx
```diff
@@ -118,7 +123,7 @@ const DuplicateVolumeForm: FC<Props> = ({ volume, close }) => {
           pool: volume.pool,
           volume_only: !values.copySnapshots,
           // logic from the lxc source code.
-          // We should not set source.project if target project is the same as the source project
+          // We should set source.project if target project is the same as the source project
```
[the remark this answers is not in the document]
Kxiru: Mistake on my part, I thought I had addressed this reversion in full.

edlerd (approved):
LGTM, thanks for driving this forward to completion :)

PR #1211, opened 2025-04-25
chore(permissions) add permissions to all workflows

edlerd:
## Done

- chore(permissions) add permissions to all workflows


Kxiru (approved):
Thanks for this added security.

PR #1215, opened 2025-04-29
fix(volume) handle volumes local to cluster members correctly

edlerd:
## Done

- fix(volume) handle volumes local to cluster members correctly in url, snapshots and updates

## QA

1. Run the LXD-UI:
    - check volume interactions on a cluster, especially duplicate a volume on a local pool (like backed by directory or zfs) and have a volume with the same name on different cluster members. Then try all the interactions with a duplicate in place: (snapshot create, restore, remove as well as volume duplicate, migrate, delete).

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -394,6 +394,18 @@ const App: FC = () => {
           element={<ProtectedRoute outlet={<StorageVolumeDetail />} />}
         />
         <Route
+          path="/ui/project/:project/storage/pool/:pool/member/:member/volumes/:type/:volume"
+          element={<ProtectedRoute outlet={<StorageVolumeDetail />} />}
+        />
+        <Route
+          path="/ui/project/:project/storage/pool/:pool/member/:member/volumes/:type/:volume/:activeTab"
+          element={<ProtectedRoute outlet={<StorageVolumeDetail />} />}
+        />
+        <Route
+          path="/ui/project/:project/storage/pool/:pool/member/:member/volumes/:type/:volume/:activeTab/:section"
+          element={<ProtectedRoute outlet={<StorageVolumeDetail />} />}
+        />
+        <Route
           path="/ui/project/:project/images"
           element={<ProtectedRoute outlet={<ImageList />} />}
         />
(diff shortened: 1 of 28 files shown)


edlerd:
While testing, surfaced this issue in LXD: https://github.com/canonical/lxd/issues/15480

Kxiru (commented):
- Please address merge conflicts

Tested:
- Copy volume on cluster
- Migrate volume to new local pool (zfs backed)- This happened when attempting to change the storage pool of a volume. Not the timestamps of the notifications. Also the migrate button takes about 10 seconds to react and close the modal.
image.
When trying this again with a new volume, migrating from zfs to zfs pool, there was no issue.
zfs -> ceph: 
image
- Having a volume with the same name on different cluster members - "test" is on cluster member micro 1, and attempted to make "test" (duplicate name) on micro3, however the UI stopped me saying "this field is required", when actually the error was a name duplication error because I had selected the SAME pool: 
image
Somehow, I changed tabs to write up this review and when I went back, the red error was gone and I was able to proceed with creation. **It took a long time,** but eventually test(2) was created on another cluster, in the same storage pool: 
image

**Please let me know if there is anything else you would like me to test**

Other notes:
- Not a part of this PR I don't think, but why is the Name field s3rd in the list? Especially after the optional cluster member dropdown.
image
- I made "test" on cluster member micro1, and test2 on micro2, HOWEVER, the API did not seem to capture this, as test2 was put on Micro1: 
image

edlerd:
Thanks for the review
- addressed the merge conflicts
- thanks for the note with the field ordering, cluster member should be second so the user choses that first and does not run into "already exists" error which is not right, because they want to target another member. But the field was missing a mark as being required on creation
- most of the errors described I can't reproduce locally
- You surfaced an interesting issue with migration. I think we should select a pool and target for the migration, in case the target pool is member-specific. But the API seems not to allow setting a different target for the migration than the one the instance already resides in. I'll try to figure out if this is a shortcoming of the API.

Kxiru (approved):
Looking good now, thanks, David!

Kxiru:
There is some issue about attempting to delete cuplicate pools that have the same name but are on different clusters, but I believe you mentioned you would address this in a separate PR.

To clarify, if you have "test" on cluster 1 and "test" on cluster 2, you get an error when you try to delete either volume from the Volume page, and may also not be able to access the overview page because of the URL generation.

PR #1216, opened 2025-04-29
Simplify api calls

edlerd:
## Done

- Simplify api calls, remove wrapping promises where they are redundant.


Kxiru (approved):
A very thorough implementation, thanks, David!
I may need to go back into the Ceph PR and reimplement some things for consistency. LGTM!

PR #1217, opened 2025-04-30
fix(modals) ensure errors from modals are visible

edlerd:
## Done

- fix(modals) ensure errors from modals are visible, ensure modals are closed on errors as well as on success.


anusha-c18 (approved):
LGTM 👍

PR #1224, opened 2025-05-02
fix(instance) combined instance migration

edlerd:
## Done

- fix(instance) combined instance migration
- this allows changing root storage, project and cluster member of an instance at the same time.
- draft because this is not working as expected in the api layer yet. The cluster member change is ignored if it is mixed with pool or project change.


edlerd:
This is blocked by https://github.com/canonical/lxd/issues/15525

edlerd:
Closing this as the api does not support this mix and match of various migrations and the current flow is much easier to gasp for a user.

PR #1244, opened 2025-05-11
fix(image) enable image export detection 

edlerd:
## Done

- enable image export detection for empty update source server after change in lxd led to image update source always being set
- Before the image.update_source would be null for images created from an instance. Now the field is filled and looks like this: 
image


edlerd:
This will not be needed as the old empty field behaviour is restored in https://github.com/canonical/lxd/pull/15555

PR #1249, opened 2025-05-15
fix(instance) disable cluster member and cluster group targeting

### LINKED ISSUE #1248: Project cluster member/group limiting not repsected on instance creation
> Hello @edlerd ,
> 
> I'm not sure if I should add this in the same thread or create another one, but basically, if I only assign a cluster group to a project and block any cluster member targeting, in the UI, I still  show the clusters/ cluster members when  trying to create an instance. It's a bit concerning since usually this is not a cloud user experience where the user doesnt know anything about the underlay infrastructure. 
> 
> Image
> 
> Thank you 

 _Originally posted by @aym-frikha in [#1197](https://github.com/canonical/lxd-ui/issues/1197#issuecomment-2869525815)_

edlerd:
## Done

- fix(instance) disable cluster member and cluster group targeting on instance creation if the project settings are limiting.

Fixes #1248

## QA

1. Run the LXD-UI:
    - in a clustered environment:
    - create a dedicated project
    - enable project restrictions
    - go to instance creation and ensure the "Cluster member" selection is disabled
    - set project restrictions "Cluster>Direct cluster targeting" to "allow"
    - ensure the Cluster member selection gets enabled
    - create a cluster group with some only one member
    - set project config "Cluster>Cluster groups targeting" to that new group
    - go to instance creation and ensure the "Cluster group" input only contains the restricted group

# Screenshots

image

### DIFF
--- a/src/pages/instances/forms/InstanceLocationSelect.tsx
+++ b/src/pages/instances/forms/InstanceLocationSelect.tsx
@@ -8,6 +8,7 @@ import { fetchClusterGroups } from "api/cluster";
 import type { FormikProps } from "formik/dist/types";
 import type { CreateInstanceFormValues } from "pages/instances/CreateInstance";
 import { useIsClustered } from "context/useIsClustered";
+import { useCurrentProject } from "context/useCurrentProject";
 
 interface Props {
   formik: FormikProps<CreateInstanceFormValues>;
@@ -38,6 +39,7 @@ const InstanceLocationSelect: FC<Props> = ({ formik }) => {
   const [selectedGroup, setSelectedGroup] = useState(defaultGroup);
   const defaultMember = figureDefaultMember(formik.values.target);
   const [selectedMember, setSelectedMember] = useState(defaultMember);
+  const { project } = useCurrentProject();
 
   const { data: clusterGroups = [], isLoading } = useQuery({
     queryKey: [queryKeys.cluster, queryKeys.groups],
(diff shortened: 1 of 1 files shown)


edlerd:
Related issue https://github.com/canonical/lxd/issues/15586

src/pages/instances/forms/InstanceLocationSelect.tsx
```diff
@@ -91,13 +99,23 @@ const InstanceLocationSelect: FC<Props> = ({ formik }) => {
...
-            value: group.name,
-            disabled: group.members.length < 1,
-          };
-        })}
+        options={clusterGroups
+          .filter((group) => {
+            const restrictedGroups =
+              project?.config["restricted.cluster.groups"];
+            if (isProjectRestricted && restrictedGroups) {
+              return restrictedGroups.includes(group.name);
```
Kxiru, line 107: I don't understand the point of this If-statement. It either returns true/false if the restrictedGroups array contains the group name, or True?
Is this supposed to be a kind of filter, and if so, why can't we use .filter() here?

edlerd: There is a wrapping filter function, depending on the return the items are filtered if the value returned is false.

The if statement here is to handle the case where no restrictions are applied on the project level. In that case we return true to keep all the options available.
In the other branch -- if restriction apply -- we only keep the groups that are in the allowed/restricted list from the project.

lorumic (commented):
> Perform the following QA steps:
> * in a clustered environment:
> * create a dedicated project
> * enable project restrictions
> * go to instance creation and ensure the "Cluster member" selection is disabled
> * set project restrictions "Cluster>Direct cluster targeting" to "allow"
> * ensure the Cluster member selection gets enabled
> * create a cluster group with some only one member
> * set project config "Cluster>Cluster groups targeting" to that new group
> * go to instance creation and ensure the "Cluster group" input only contains the restricted group

Only the cluster group targeting works for me. Direct cluster selection stays disabled - see screen capture below:

Peek 2025-05-20 14-51

I think the reason might be that I only have 1 member in the cluster, and there might be a check that disables the dropdown in that case. Does such check exist, and if so, should I relax it to test that the direct cluster selection works?

src/pages/instances/forms/InstanceLocationSelect.tsx
lorumic, line 107: What about:
```suggestion
            if (isProjectRestricted) {
              return (project?.config["restricted.cluster.groups"] ?? []).includes(group.name);
            }
            return true;
```

lorumic (approved):
Thanks for the clarification. Just one small improvement idea, but LGTM overall :+1:

PR #1252, opened 2025-05-16
feat(warnings) allow bulk delete of server warnings

edlerd:
## Done

- feat(warnings) allow bulk delete of server warnings

## QA

1. Run the LXD-UI:
    - open server warning page
    - ensure warnings page is responsive in smaller resolutions
    - bulk select / delete warnings

## Screenshots

image

### DIFF
--- a/src/api/warnings.tsx
+++ b/src/api/warnings.tsx
@@ -2,6 +2,7 @@ import { handleResponse } from "util/helpers";
 import type { LxdWarning } from "types/warning";
 import type { LxdApiResponse } from "types/apiResponse";
 import { withEntitlementsQuery } from "util/entitlements/api";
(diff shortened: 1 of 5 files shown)


Kxiru (approved):
- Wondering if the Delete Warnings actions should go in the top right as opposed to top left? I know we have a paradigm for both. Maybe it's better as is.
- Code QA looks good to me!

Screenshots@
image
image
image

edlerd:
> * Wondering if the Delete Warnings actions should go in the top right as opposed to top left? I know we have a paradigm for both. Maybe it's better as is.

I think left alignment is better as it is closer to the select and consistent with the other select tables. We tend to put the creation actions on the right side in the header, so probably fine to keep that side empty for warnings that are not user created.

PR #1258, opened 2025-05-20
fix(network) remove 'range' from ipv4 and ipv6 address labels in network form

### LINKED ISSUE #1253: Unclear description at network creation
Took me longer to figure out than i'd like to admit haha, but I came across the following issue

Image

the above is invalid. Eventhough it's a perfectly valid **IPv4 address range**. 

It only accepts host addresses inside the subnet, not the network address. `10.30.0.1/24` works :). Perhaps you could put this in the description to use the first host address, or change the name from `IPv4 address range` to `IPv4 address` to match the yaml config `ipv4.address`.

As a network engineer, changing it to `IPv4 address`  (And of course IPv6 as well) makes more sense to me. The /16 already indicates that it is a range and this is the address of the bridge on your host.

edlerd:
## Done

- fix(network) remove 'range' from ipv4 and ipv6 address labels in network form

Fixes #1253

## QA

1. Run the LXD-UI:
    - open network detail page, check ipv4 and ipv6 address field labels. they should be without the "range" suffix.

## Screenshots

image

### DIFF
--- a/src/pages/networks/forms/NetworkFormMain.tsx
+++ b/src/pages/networks/forms/NetworkFormMain.tsx
@@ -108,7 +108,7 @@ const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
               row={getConfigurationRow({
                 formik,
                 name: "ipv4_address",
-                label: "IPv4 address range",
+                label: "IPv4 address",
                 defaultValue: "auto",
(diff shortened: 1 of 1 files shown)


edlerd:
Demo: https://lxd-ui-1258.demos.haus/

PR #1260, opened 2025-05-20
chore(pr) check for non-ascii characters in code files

edlerd:
## Done

- chore(pr) check for non-ascii characters in code files


src/pages/permissions/panels/LoggedInUserNotification.tsx
```diff
@@ -26,7 +26,7 @@ const LoggedInUserNotification: FC<Props> = ({ isVisible }) => {
         <Tag className="u-no-margin--left" isVisible={isVisible}>
           You
         </Tag>{" "}
-        might not be able to reverse this change once you’ve made it.
+        might not be able to reverse this change once you&rsquo;ve made it.
```
Kxiru, line 29: All the other apostrophes use &rsquo;. Is it not necessary here because it is a comment?

Kxiru (approved):
Nice way to maintain code integrity. LGTM!

tests/operations.spec.ts
```diff
@@ -29,7 +29,7 @@ test("instance operations are recognised on the Operations page", async ({
   await checkNotificationExists(page);
   await assertTextVisible(page, "Instance " + instance + " started");
 
-  // click on the “x operations” link in the status bar
+  // click on the 'x operations' link in the status bar
```
Kxiru, line 32: All the other apostrophes use ’. Is it not necessary here because it is a comment?

Kxiru: I suppose that is exactly the reason because it does not get rendered. Looks good!

PR #1267, opened 2025-05-22
fix(helpers): fix calculation of element absolute height

lorumic:
## Done

- Fixed calculation of element absolute height

### DIFF
--- a/src/util/helpers.tsx
+++ b/src/util/helpers.tsx
@@ -242,9 +242,7 @@ export const getElementAbsoluteHeight = (element: HTMLElement) => {
   }
   const style = window.getComputedStyle(element);
   const margin = parseFloat(style.marginTop) + parseFloat(style.marginBottom);
-  const padding =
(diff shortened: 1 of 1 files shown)


edlerd (changes requested):
The CSS in many places of the UI relies on the old behavior of the function. With the change, scrollbar appear in many places where they should be omitted.

I wonder if it will be better to keep the old behavior.

# Profile YAML editor
image

# Instance config
image

# Instance creation
image

# Network creation (likely all other creation pages as well)
image

# Storage volume config
image

lorumic:
> I wonder if it will be better to keep the old behavior.

The old behaviour is wrong - it is adding padding twice. If correctness has some importance, it should not be kept.

Since changes in multiple places are needed, I have filed [a Jira issue](https://warthogs.atlassian.net/browse/WD-22275) for you to work on in the future, and will close this PR for now.

PR #1277, opened 2025-05-27
feat: Import custom storage volumes [WD-22291]

Kxiru:
## Done

- Import Custom storage volumes button

## QA

1. Run the LXD-UI:
    - Navigate to create a custom storage volume page.
    - Attempt to import a custom storage volume (Having already exported / backed up a volume)

## Screenshots

image

### DIFF
--- a/src/api/storage-volumes.tsx
+++ b/src/api/storage-volumes.tsx
@@ -251,3 +251,35 @@ export const createVolumeBackup = async (
       return data;
     });
 };
+
+export const uploadVolume = async (
+  file: File | null,
+  name: string,
+  project: string | undefined,
+  pool: string | undefined,
+  setUploadState: (value: UploadState) => void,
+  uploadController: AbortController,
+  target: string,
+): Promise<LxdOperationResponse> => {
+  const targetParam = target ? `&target=${target}` : "";
+  return axios
+    .post(
+      `/1.0/storage-pools/${pool}/volumes/custom?project=${project}${targetParam}`,
+      file,
+      {
+        headers: {
+          "Content-Type": "application/octet-stream",
+          "X-LXD-name": name,
+        },
+        onUploadProgress: (event) => {
+          setUploadState({
+            percentage: event.progress ? Math.floor(event.progress * 100) : 0,
+            loaded: event.loaded,
+            total: event.total,
+          });
+        },
+        signal: uploadController.signal,
+      },
+    )
+    .then((response: AxiosResponse<LxdOperationResponse>) => response.data);
+};
--- a/src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
+++ b/src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
@@ -9,48 +9,49 @@ import { useCurrentProject } from "context/useCurrentProject";
 import StoragePoolSelector from "pages/storage/StoragePoolSelector";
 import type { ChangeEvent, FC } from "react";
 import { useCallback, useState } from "react";
-import { instanceNameValidation } from "util/instances";
-import type { UploadState } from "types/storage";
+import { fileToSanitisedName } from "util/helpers";
+import type { LxdStorageVolume, UploadState } from "types/storage";
 import { useEventQueue } from "context/eventQueue";
 import { useToastNotification } from "context/toastNotificationProvider";
 import { useQueryClient } from "@tanstack/react-query";
 import { useNavigate } from "react-router-dom";
 import { queryKeys } from "util/queryKeys";
-import { uploadInstance } from "api/instances";
 import { useFormik } from "formik";
 import type { AxiosError } from "axios";
 import type { LxdSyncResponse } from "types/apiResponse";
 import * as Yup from "yup";
 import classnames from "classnames";
-import { useSupportedFeatures } from "context/useSupportedFeatures";
-import type { InstanceFileType } from "./InstanceFileTypeSelector";
-import InstanceFileTypeSelector from "./InstanceFileTypeSelector";
 import ResourceLink from "components/ResourceLink";
 import ResourceLabel from "components/ResourceLabel";
-import { fileToSanitisedName } from "util/helpers";
+import { uploadVolume } from "api/storage-volumes";
+import {
+  linkForVolumeDetail,
+  testDuplicateStorageVolumeName,
+} from "util/storageVolume";
+import ClusterMemberSelector from "pages/cluster/ClusterMemberSelector";
+import { useStoragePools } from "context/useStoragePools";
+import { isClusterLocalDriver } from "util/storagePool";
+import { useClusterMembers } from "context/useClusterMembers";
 
-export interface UploadInstanceBackupFileFormValues {
-  instanceFile: File | null;
+export interface UploadVolumeBackupFileFormValues {
+  volumeFile: File | null;
   name: string;
   pool: string;
+  clusterMember: string;
 }
 
 interface Props {
   close: () => void;
   uploadState: UploadState | null;
   setUploadState: (state: UploadState | null) => void;
-  fileType: InstanceFileType;
-  setFileType: (value: InstanceFileType) => void;
-  defaultInstanceName?: string;
+  defaultVolumeName?: string;
 }
 
-const UploadInstanceBackupFileForm: FC<Props> = ({
+const UploadVolumeBackupFileForm: FC<Props> = ({
   close,
   uploadState,
   setUploadState,
-  fileType,
-  setFileType,
-  defaultInstanceName,
+  defaultVolumeName,
 }) => {
   const { project, isLoading } = useCurrentProject();
   const eventQueue = useEventQueue();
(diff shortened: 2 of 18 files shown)


src/api/storage-volumes.tsx
```diff
@@ -232,3 +232,34 @@ export const copyStorageVolume = async (
...
+  file: File | null,
+  name: string,
+  project: string | undefined,
+  pool: string | undefined,
+  setUploadState: (value: UploadState) => void,
+  uploadController: AbortController,
+): Promise<LxdOperationResponse> => {
+  return axios
+    .post(
+      `/1.0/storage-pools/${pool}/volumes/custom/`,
```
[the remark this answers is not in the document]
edlerd: We also need to add the project as a parameter here to import to the right current project.

edlerd (changes requested):
Thanks for adding the cluster member selector! Some observations to make this perfect below.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+    setUploadAbort(uploadController);
+
+    uploadVolume(
+      values.volumeFile,
+      values.name,
+      project?.name,
+      values.pool,
+      setUploadState,
+      uploadController,
+      values.targetClusterMember,
```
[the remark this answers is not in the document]
Kxiru: 1. If selected a CMS pool, then the target should be forwarded,
2. If not CMS, then the target should not be forwarded.
3. OnSubmit, we can check if the pool is CMS. If not, then we can forward a null or empty string.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+      <>
+        Created volume{" "}
+        <ResourceLink type="volume" value={volumeName} to={volumeURL} />.
+      </>
+    );
+
+    const actions = [
+      {
+        label: "Configure",
+        onClick: async () => navigate(`${volumeURL}/configuration`),
```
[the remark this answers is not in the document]
Kxiru: Redefine VolumeURL locally.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+            title: noFileSelectedMessage,
+          }}
+        />
+        <ClusterMemberSelector
+          {...formik.getFieldProps("targetClusterMember")}
+          id="targetClusterMember"
+          label="Target cluster member"
+          disabled={!!noFileSelectedMessage || !isCMSDriver}
+          disableReason={
+            !isCMSDriver ? "The selected pool is not cluster specific" : ""
```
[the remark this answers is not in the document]
Kxiru: I think the negative version better describes _why_ the button is disabled. Although it may be logically implied, the suggestion you provided does not tell the User why the element is disabled or imply what to do about it.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+        setUploadState(null);
+      });
+  };
+
+  const formik = useFormik<UploadVolumeBackupFileFormValues>({
+    initialValues: {
+      name: defaultVolumeName || "",
+      pool: "",
+      volumeFile: null,
+      targetClusterMember: "",
```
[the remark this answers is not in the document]
edlerd: We could add the `  const { data: clusterMembers = [], isLoading } = useClusterMembers();` to the top of the file and use `clusterMembers?.[0]?.server_name` as a default here.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,284 @@
...
+        setUploadState(null);
+      });
+  };
+
+  const formik = useFormik<UploadVolumeBackupFileFormValues>({
+    initialValues: {
+      name: defaultVolumeName || "",
+      pool: pools?.[0]?.name,
+      volumeFile: null,
+      targetClusterMember: clusterMembers?.[0]?.server_name,
```
edlerd:
```suggestion
      targetClusterMember: clusterMembers?.[0]?.server_name ?? "",
```
This needs a fallback, otherwise this gets initialized with null in the non-cluster case and the success notification branch produces an error.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,284 @@
...
+import ClusterMemberSelector from "pages/cluster/ClusterMemberSelector";
+import { useStoragePools } from "context/useStoragePools";
+import { isClusterLocalDriver } from "util/storagePool";
+import { useClusterMembers } from "context/useClusterMembers";
+
+export interface UploadVolumeBackupFileFormValues {
+  volumeFile: File | null;
+  name: string;
+  pool: string;
+  targetClusterMember: string;
```
edlerd:
```suggestion
  clusterMember: string;
```
Let's go with a standard name.

The current custom name breaks the check in `testDuplicateStorageVolumeName`. Because on a cluster, an import with the name that already exists on another cluster member is wrongfully detected as a duplicate. I.e. when importing to another cluster member where a volume with that name does not exist yet.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,284 @@
...
+          disabled={!!noFileSelectedMessage}
+          title={noFileSelectedMessage}
+        />
+        <StoragePoolSelector
+          value={formik.values.pool}
+          setValue={(value) => {
+            void formik.setFieldValue("pool", value);
+          }}
+          selectProps={{
+            id: "pool",
```
edlerd:
```suggestion
            id: "volume-import-pool",
```
This must be a unique id per page. It clashes with the pool id as we are mounting this modal on the create a volume form page.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,284 @@
...
+            "custom",
+            volumeNameAbort,
+          ),
+        )
+        .optional(),
+    }),
+    onSubmit: handleUpload,
+  });
+
+  const pool = pools.find((pool) => (pool.name = formik.values.pool));
```
edlerd:
```suggestion
  const pool = pools.find((pool) => (pool.name === formik.values.pool));
```
This should be a comparison, not an assignment.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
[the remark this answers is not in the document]
edlerd: We are still sending the cluster member to a non-local pool on a cluster.

edlerd (changes requested):

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
Kxiru: Omg, embarassed at having missed that. Thanks for pointing it out!

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,286 @@
...
+            "custom",
+            volumeNameAbort,
+          ),
+        )
+        .optional(),
+    }),
+    onSubmit: handleUpload,
+  });
+
+  const pool = pools.find((pool) => pool.name == formik.values.pool);
```
edlerd:
```suggestion
  const pool = pools.find((pool) => pool.name === formik.values.pool);
```
Better use the strict type comparison.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,286 @@
...
+          }}
+          selectProps={{
+            id: "volume-import-pool",
+            label: "Storage pool",
+            disabled: isLoading || !!noFileSelectedMessage,
+            title: noFileSelectedMessage,
+          }}
+        />
+        <ClusterMemberSelector
+          {...formik.getFieldProps("targetClusterMember")}
```
edlerd:
```suggestion
          {...formik.getFieldProps("clusterMember")}
```
This is now updating an outdated field name.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,286 @@
...
+  };
+
+  const formik = useFormik<UploadVolumeBackupFileFormValues>({
+    initialValues: {
+      name: defaultVolumeName || "",
+      pool: pools?.[0]?.name,
+      volumeFile: null,
+      clusterMember: clusterMembers?.[0]?.server_name ?? "",
+    },
+    validateOnMount: true,
```
edlerd, line 159:
```suggestion
    validateOnMount: true,
    enableReinitialize: true,
```
We should allow reinitialize in case the pools or cluster members are loaded async. Currently not an issue, because the modal is mounted on the create volume page so both are hitting the cache. But we might move the entry point to this form to a different place and then this would break.

src/pages/storage/forms/UploadVolumeBackupFileForm.tsx
```diff
@@ -0,0 +1,286 @@
...
+            title: noFileSelectedMessage,
+          }}
+        />
+        <ClusterMemberSelector
+          {...formik.getFieldProps("targetClusterMember")}
+          id="targetClusterMember"
+          label="Target cluster member"
+          disabled={!!noFileSelectedMessage || !isCMSDriver}
+          disableReason={
+            !isCMSDriver ? "The selected pool is not cluster specific" : ""
```
edlerd: We are missing to set the noFileSelectedMessage here if it is set.

edlerd (changes requested):
Looks good, thanks for the fixes. Only very minor adjustments, then this should be good to merge.

edlerd (approved):
QA and Code LGTM, thanks for adding the import volume! :+1:

PR #1279, opened 2025-05-30
fix(ceph) add ceph.rbd.du setting to storage pools

edlerd:
## Done

- fix(ceph) add ceph.rbd.du setting to storage pools
- filter out volatile.initial_source as a member local storage pool setting

## QA

1. Run the LXD-UI:
    - Have a ceph pool on a clustered backend
    - Check the pool config, enable the `Ceph disk usage` option
    - Create an instance and/or a custom volume on the ceph pool.
    - Go to the volume list and see the "size" column filled.
    - Disable the config and ensure you can't read the sizes anymore on the volume list.

# Screenshot
image

### DIFF
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -109,6 +109,7 @@ const getClusterAndMemberPoolPayload = (pool: Partial<LxdStoragePool>) => {
     "zfs.pool_name",
     "lvm.thinpool_name",
     "lvm.vg_name",
+    "volatile.initial_source",
   ]);
   const configKeys = Object.keys(pool.config || {});
   const memberConfig: LxdStoragePool["config"] = {};
--- a/src/pages/storage/EditStoragePool.tsx
+++ b/src/pages/storage/EditStoragePool.tsx
@@ -25,7 +25,7 @@ import {
 import { slugify } from "util/slugify";
 import { yamlToObject } from "util/yaml";
 import { useSettings } from "context/useSettings";
-import { getSupportedStorageDrivers } from "util/storageOptions";
+import { cephDriver, getSupportedStorageDrivers } from "util/storageOptions";
 import YamlSwitch from "components/forms/YamlSwitch";
 import FormSubmitBtn from "components/forms/FormSubmitBtn";
 import ResourceLink from "components/ResourceLink";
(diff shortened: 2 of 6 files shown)


Kxiru (commented):
## **Screenshots:**
**Config row**
image

**Size column filled having created a custom volume:**
image

**Size column filled having created a custom instance:**
image

**Size column(s) no longer visible after making config false:**
image


**QA Notes:**
- Everything worked as expected, however after changed the config back to false and going to the volume page, the sizes were still visible and I had to refresh the page to have the sizes disappear.
  - This could either be a network issue, or a need to invalidate cache somewhere. Wdyt?
  - I just turned the config off and on again to see if I could recreate the issue, and went back to the volume page. Sure enough, the Sizes have not disappeared yet 
image. Upon refreshing the pages, they are gone.
.

edlerd:
> Everything worked as expected, however after changed the config back to false and going to the volume page, the sizes were still visible and I had to refresh the page to have the sizes disappear.
> * This could either be a network issue, or a need to invalidate cache somewhere. Wdyt?

Good catch, I added a more vague invalidation, so it also clears the cached volume and volume state queries after editing a storage pool.

edlerd:
> Hmm. I tested this three times, but unfortunately this still does not behave as expected as the Size values still remain after I have set the rbd value to false from true. Happy to Show/Demo this for you.

I can reproduce it locally. The problem is that the newly fetched sizes result in an error response after disabling the `rbd.du` feature. The invalidation of the cache does not clear the cache. So the volume sizes are still served from the stale cache. We could remove the queries from the cache, but that leads to a slightly worse UX. I think we should better keep it as is and have this slight inconsistency in cache behavior.

Kxiru:
> > Hmm. I tested this three times, but unfortunately this still does not behave as expected as the Size values still remain after I have set the rbd value to false from true. Happy to Show/Demo this for you.
> 
> I can reproduce it locally. The problem is that the newly fetched sizes result in an error response after disabling the `rbd.du` feature. The invalidation of the cache does not clear the cache. So the volume sizes are still served from the stale cache. We could remove the queries from the cache, but that leads to a slightly worse UX. I think we should better keep it as is and have this slight inconsistency in cache behavior.

Moving the conversation to Mattermost for further discussion.

edlerd:
> Moving the conversation to Mattermost for further discussion.

Added [this](https://github.com/canonical/lxd-ui/pull/1279/files#diff-e8fa5f29699cd8f99f74279ddc6bb51a7d8e57f14516b7f77498384c4b684412R132-R141) to remove queries from the cache. It forces the whole page to refetch, so trying to limit the cases that lead to the invalidation.

PR #1280, opened 2025-05-30
fix(instance) use sh instead of bash on nixos and alpine images for new terminal

edlerd:
## Done

- fix(instance) use sh instead of bash on nixos and alpine images for new terminal

## QA

1. Run the LXD-UI:
    - create three new instances, based on ubuntu, alpine nixos base image
    - start the instances and open the terminal page on the three detail pages
    - ensure the terminal connects.

### DIFF
--- a/src/pages/instances/InstanceTerminal.tsx
+++ b/src/pages/instances/InstanceTerminal.tsx
@@ -30,20 +30,27 @@ const XTERM_OPTIONS = {
   },
 };
 
-export const defaultPayload: TerminalConnectPayload = {
-  command: "bash",
-  environment: [
-    {
-      key: "TERM",
-      value: "xterm-256color",
-    },
-    {
-      key: "HOME",
-      value: "/root",
-    },
-  ],
-  user: 0,
-  group: 0,
+const getDefaultPayload = (instance: LxdInstance): TerminalConnectPayload => {
+  const imageDescription = instance.config["image.description"];
+  const isAlpine = imageDescription?.toLowerCase().includes("alpine");
+  const isNixOs = imageDescription?.toLowerCase().includes("nixos");
+  const command = isAlpine || isNixOs ? "sh" : "bash";
+
+  return {
+    command: command,
+    environment: [
+      {
+        key: "TERM",
+        value: "xterm-256color",
+      },
+      {
+        key: "HOME",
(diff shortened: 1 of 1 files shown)


edlerd:
> The [build and run as described in the docs](https://github.com/canonical/lxd-ui/CONTRIBUTING.md#setting-up-for-development) link in the PR description is broken.

Woops, will update this in a follow up PR.

edlerd:
> Why did NixOS get downgraded to sh? We have bash in our base images.

When I tested this back then the console wouldn't connect with bash on NixOS. I'll check again, though this depends on the LXD provided images for nixos. Not sure if those are matching the base images you refer to or or are minimal with sh only. If the latter, we might also adjust the image build to include bash.

adamcstephens:
Y'all aren't building the images, the NixOS project is. :) Admittedly I just realized this is the LXD image, which is not as well maintained any longer. It should still have bash installed, if not in /bin then it's definitely in /run/current-system/sw/bin

https://github.com/canonical/lxd-imagebuilder/blob/4fcc0ae4828c50d4f5375d4c5033f157dc15f904/sources/nixos-http.go#L21

adamcstephens:
```
root@ethical-rat:~# lxc launch images:nixos/unstable/default c2
Launching c2
root@ethical-rat:~# lxc exec c2 -- sh -l

[root@nixos:~]# ls /run
sh: ls: command not found

[root@nixos:~]# echo $PATH
/run/wrappers/bin:/root/.nix-profile/bin:/nix/profile/bin:/root/.local/state/nix/profile/bin:/etc/profiles/per-user/root/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin
```

This image is broken and `/sbin/init` is not being properly executed which would setup /run/current-system. The correct fix is not to disable bash, but to fix the image.

adamcstephens:
I actually tracked down the issue. NixOS images need to be launched with nesting.

```
root@growing-longhorn:~# lxc launch images:nixos/unstable --ephemeral -c security.nesting=true
Launching the instance
Instance name is: outgoing-grizzly                 

root@growing-longhorn:~# lxc exec outgoing-grizzly bash  

[root@nixos:~]# exit
```

I agree with the switch to `su -l`, thanks for looking into this again!

PR #1294, opened 2025-06-10
feat(operations) add pagination and other improvements for operation list

edlerd:
## Done

- feat(operations) add pagination and other improvements for operation list

Fixes WD-22684

## QA

1. Run the LXD-UI:
    - check operation list. open a couple of instance terminals in other tabs to have running operations.

## Screenshots

image


lorumic (commented):
Some other small ideas that I noticed while working on the implementation in AMS:

- This [line](https://github.com/edlerd/lxd-ui/blob/operation-improvements/src/pages/operations/OperationList.tsx#L83) should be changed to `{operation.err && <div>{operation.err}</div>}` to avoid rendering an empty `div` when `operation.err` is not defined.
- In this [line](https://github.com/edlerd/lxd-ui/blob/operation-improvements/src/components/OperationStatus.tsx#L19) the word "Ops" should be passed to the `pluralize` method, to avoid rendering "1 Ops".

- Search and filter is reported in the ticket, but seems to be missing from the changes in this PR.

You might want to have a look at @anusha-c18's review in AMS for a few other improvements: https://github.com/canonical/ams/pull/1213#pullrequestreview-2910210674

src/pages/operations/OperationList.tsx
```diff
@@ -42,7 +49,14 @@ const OperationList: FC = () => {
     }[status];
   };
 
-  const rows = operations.map((operation) => {
+  const filteredOperations = operations.filter(
+    (operation) =>
+      operation.description.toLowerCase().includes(query.toLowerCase()) ||
+      getProjectName(operation).toLowerCase().includes(query.toLowerCase()) ||
+      operation.status.toLowerCase().includes(query.toLowerCase()),
+  );
```
lorumic: A couple of considerations here:

1. We could declare `const lowerCaseQuery = query.toLowerCase()` once, and then reuse the const in the 3 conditions instead of calling `query.toLowerCase()` thrice.
2. While QAing this I found a bit surprising that I could not filter by instance name. I think that is what most users would do, so we should probably include that in the filter condition.

src/components/OperationStatus.tsx
```diff
@@ -14,10 +14,9 @@ const OperationStatus = () => {
...
 
-  let operationsStatus = `${runningOperations.length} ${pluralize("operation", runningOperations.length)} in progress...`;
-  if (isSmallScreen) {
-    operationsStatus = `${runningOperations.length} Ops`;
-  }
+  const operationsStatus = isSmallScreen
+    ? `${runningOperations.length} ${pluralize("op", runningOperations.length)}...`
+    : `${runningOperations.length} ${pluralize("operation", runningOperations.length)} in progress...`;
 
   return runningOperations.length ? (
```
lorumic: We can also have an early exit before computing the `operationStatus` value, see https://github.com/canonical/ams/pull/1213#discussion_r2135835194

src/pages/operations/OperationList.tsx
```diff
@@ -122,30 +134,68 @@ const OperationList: FC = () => {
...
+        header={
+          <PageHeader>
+            <PageHeader.Left>
+              <PageHeader.Title>Ongoing operations</PageHeader.Title>
+              {operations.length > 0 && (
+                <PageHeader.Search>
+                  <SearchBox
+                    className="search-box margin-right u-no-margin--bottom"
+                    name="search-operations"
+                    type="text"
```
lorumic: Do we need this? Looks like it can be removed.
```suggestion
```

src/pages/operations/OperationList.tsx
```diff
@@ -122,30 +134,68 @@ const OperationList: FC = () => {
...
+                    onChange={(value) => {
+                      setQuery(value);
+                    }}
```
lorumic: Maybe shorter?
```suggestion
                    onChange={setQuery}
```

src/pages/operations/OperationList.tsx
```diff
@@ -122,30 +134,68 @@ const OperationList: FC = () => {
...
+                  id="operation-table"
+                  headers={headers}
+                  sortable
+                  onUpdateSort={updateSort}
+                  className="operation-list"
+                  emptyStateMsg={
+                    isLoading ? (
+                      <Loader text="Loading operations..." />
+                    ) : (
+                      "No matching operations found for this search"
```
lorumic: I think that the word "matching" already implies that there is a search involved:
```suggestion
                      "No matching operations found"
```

src/pages/operations/OperationList.tsx
```diff
@@ -122,30 +134,68 @@ const OperationList: FC = () => {
...
+                itemName="operation"
+                className="u-no-margin--top"
+                aria-label="Table pagination control"
+              >
+                <MainTable
+                  id="operation-table"
+                  headers={headers}
+                  sortable
+                  onUpdateSort={updateSort}
+                  className="operation-list"
```
lorumic: Was the `responsive` prop removed intentionally from here? (it used to be there before)

lorumic (commented):
Thanks for the quick iteration! Adding a few other ideas.

src/pages/operations/OperationList.tsx
```diff
@@ -122,30 +138,66 @@ const OperationList: FC = () => {
...
-      <BaseLayout
-        title="Ongoing operations"
-        controls={<RefreshOperationsBtn />}
+      <CustomLayout
+        header={
+          <PageHeader>
+            <PageHeader.Left>
+              <PageHeader.Title>Ongoing operations</PageHeader.Title>
+              {operations.length > 0 && (
+                <PageHeader.Search className="u-hide--small u-hide--medium">
```
lorumic: I'm surprised to see this - since we have the search box in other places, I would have expected to see this applied upstream. Then I investigated and found that, in other pages, this is handled with media queries individually - e.g. [in the images page](https://github.com/edlerd/lxd-ui/blob/operation-improvements/src/sass/_images.scss#L59-L63):
```
  @media screen and (width < 650px) {
    .page-header__search {
      display: none;
    }
  }
```
No strong opinions, but I think it would be better to adopt a consistent approach:

- One can use the same approach used in other pages, i.e. the media query as in the example linked above, or
- The utility classes `u-hide--small u-hide--medium` can be applied directly [here](https://github.com/edlerd/lxd-ui/blob/operation-improvements/src/components/PageHeader.tsx#L18-L23) so that we don't need to define custom CSS for each page individually. I personally favour this one, as I think we would always want to hide the search box on small and medium screens.

edlerd: Good catch, going with the established media query.

src/pages/operations/OperationList.tsx
```diff
@@ -56,30 +74,28 @@ const OperationList: FC = () => {
...
-                <OperationInstanceName operation={operation} />
-              </div>
+              <OperationInstanceName operation={operation} />
               <div className="u-text--muted u-truncate" title={projectName}>
                 Project: {projectName}
               </div>
             </>
           ),
-          role: "cell",
+          role: "rowheader",
```
lorumic, line 91: What happened here? :sweat_smile: I think this was correct before?
```suggestion
          role: "cell",
```

edlerd: We do want one cell to be the header of the row.

edlerd: See also https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/rowheader_role

lorumic (approved):
Optionally, you could search also in the `metadata` - but non-blocking.

PR #1301, opened 2025-06-13
chore(volume) add volume upload test

edlerd:
## Done

- chore(volume) add volume upload test

### DIFF
--- a/tests/storage.spec.ts
+++ b/tests/storage.spec.ts
@@ -24,6 +24,7 @@ import {
   deleteProject,
   randomProjectName,
 } from "./helpers/projects";
+import { gotoURL } from "./helpers/navigate";
 
 let volume = randomVolumeName();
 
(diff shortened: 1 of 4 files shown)


tests/storage.spec.ts
```diff
@@ -200,4 +201,24 @@ test("Export a volume backup", async ({ page }) => {
   await page.waitForSelector(`text=Volume ${volume} download started`);
   const VOLUME_FILE = "tests/fixtures/volume.tar.gz";
   await download.saveAs(VOLUME_FILE);
+
+  //Upload a volume
+  const uploadVolume = randomVolumeName();
+  await gotoURL(page, "/ui/");
```
Kxiru, line 207: Getting this weird error over here, but it doesn't stop the flow of the test because the next few lines rely on navigation via the omni-present side navigation.
Thoughts? Not sure what exactly makes this an "invalid" URL, and we do use this approach in other areas of the CB, without issue (albeit, as a fallback).
image

edlerd: The lines in the error relate to the `createVolume` call in the `test.beforeAll` block. Maybe your `dotrun` was still booting up and the test retried until it worked?

PR #1306, opened 2025-06-16
fix(operations) ensure date wraps well on small screens

edlerd:
## Done

- fix(operations) ensure date wraps well on small screens

## QA

1. Run the LXD-UI:
    - have a running operation (i.e. open a terminal for a running instance in antoher tab)
    - browse operation list with a small screen, like mobile format
    - ensure the date wraps correctly and is readable

### DIFF
--- a/src/pages/operations/actions/RefreshOperationsBtn.tsx
+++ b/src/pages/operations/actions/RefreshOperationsBtn.tsx
@@ -4,10 +4,13 @@ import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { ActionButton, Icon } from "@canonical/react-components";
 import { useOperations } from "context/operationsProvider";
+import { isWidthBelow } from "util/helpers";
+import classnames from "classnames";
 
 const RefreshOperationsBtn: FC = () => {
   const { isFetching } = useOperations();
   const queryClient = useQueryClient();
+  const isSmallScreen = () => isWidthBelow(620);
 
   const handleRefresh = () => {
     queryClient.invalidateQueries({ queryKey: [queryKeys.operations] });
@@ -18,12 +21,14 @@ const RefreshOperationsBtn: FC = () => {
 
   return (
     <ActionButton
-      className="u-no-margin--bottom has-icon"
+      className={classnames("u-no-margin--bottom", {
+        "has-icon": !isSmallScreen(),
+      })}
       onClick={handleRefresh}
       loading={isFetching}
       disabled={isFetching}
     >
-      <Icon name="restart" />
+      {!isSmallScreen() && <Icon name="restart" />}
       <span>Refresh</span>
     </ActionButton>
   );
--- a/src/sass/_operation_list.scss
+++ b/src/sass/_operation_list.scss
@@ -9,6 +9,10 @@
     }
   }
 
+  .date-pair {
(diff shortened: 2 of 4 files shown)


src/util/helpers.tsx
```diff
@@ -55,6 +55,10 @@ export const getBrowserFormatDate = (d: Date): string =>
     d.getHours(),
   )}:${pad(d.getMinutes())}`;
 
+export const nonBreakingSpaces = (text: string): string => {
+  return text.replace(/ /g, "\u00A0");
```
bartaz, line 59: I wonder if it would make sense to use `\s` to match any whitespace character, but this would include new lines, so maybe safer without it.

src/sass/_operation_list.scss
```diff
@@ -9,6 +9,10 @@
     }
   }
 
+  .date-pair {
+    text-wrap: auto;
```
bartaz: Isn't that the default? You need to override it?
Oh, you are overriding Vanilla's `pretty`? Maybe it would be worth mentioning in code comment for future context.

bartaz:
QA +1 on the date wrapping, but there is a bunch of other things that get bit wonky on such small screen:
- refresh button label wrapped under icon
- pagination wraps weirdly
- there is some weird whitespace above bottom bar, making the scrollable area with table smaller than it could be

image

bartaz (dismissed):
Looks good, works as advertised.

Small suggestion to add a comment in CSS override for context.

src/sass/_operation_list.scss
edlerd: Added a comment.

edlerd:
Thanks for the thorough review @bartaz 

Those have been addressed now as well:

> * refresh button label wrapped under icon
> * there is some weird whitespace above bottom bar, making the scrollable area with table smaller than it could be

This one should be fixed upstream in react-components, I took a note to look into it shortly.

> * pagination wraps weirdly

src/pages/operations/actions/RefreshOperationsBtn.tsx
```diff
@@ -4,10 +4,13 @@ import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { ActionButton, Icon } from "@canonical/react-components";
 import { useOperations } from "context/operationsProvider";
+import { isWidthBelow } from "util/helpers";
+import classnames from "classnames";
 
 const RefreshOperationsBtn: FC = () => {
   const { isFetching } = useOperations();
   const queryClient = useQueryClient();
+  const isSmallScreen = () => isWidthBelow(620);
```
bartaz, line 13: This only seems to work when page is loaded (component initially rendered), and not when the screen is resized.
So, if screen was loaded large and resized to small, the icon will still be there. And vice versa, if loaded on small (without icon), the icon will not be there after resize.

But, as it's likely not a common case for users to resize on small screens, it should be OK to keep it as is for now.

Maybe in future it could be improved with some resize handling.

bartaz (approved):
LGTM, thanks

src/pages/operations/actions/RefreshOperationsBtn.tsx
edlerd: I agree. We should probably rewrite the function that is called to be a hook that calculates again on resize. I'll file a ticket to look into this.

PR #1308, opened 2025-06-17
feat: [WD-22736] Include buckets in pool used by list

Kxiru:
## Done

- Includes buckets in pool used-by list.

## QA

1. Run the LXD-UI:
    - Navigate to a storage pool (cephobject) that contains at least one Storage bucket and verify that this is reflected in the Used By list in the detail page.

## Screenshots

image


src/pages/storage/StorageUsedBy.tsx
```diff
@@ -127,6 +129,22 @@ const StorageUsedBy: FC<Props> = ({ storage, project }) => {
...
+          <th className="u-text--muted">Buckets ({data[BUCKETS].length})</th>
+          <td>
+            <ExpandableList
+              items={data[BUCKETS].map((item) => (
+                <UsedByItem
+                  key={`${item.name}-${item.project}`}
+                  item={item}
+                  activeProject={project}
+                  type="bucket"
+                  to={`/ui/project/${item.project}/storage/bucket/${item.name}`}
```
edlerd: This is not a valid url yet.  I think the bucket detail url would also need the pool along with the name and location to uniquely identify a bucket.

edlerd (changes requested):

src/pages/storage/StorageUsedBy.tsx
Kxiru: I extrapolated this logic into a function (getStorageBucketURL) as it can also be used in Create and Edit. That way the URLs can be changed from a central location when it is decided.

src/util/storageBucket.tsx
```diff
@@ -27,3 +27,13 @@ export const testDuplicateStorageBucketName = (
...
 };
+
+export const getStorageBucketURL = (
+  bucketName: string,
+  pool: string,
+  location?: string,
+  project?: string,
+) => {
+  const bucketLocation = location ? `member/${location}` : ``;
+  return `/ui/project/${project}/storage/pools/${pool}/${bucketLocation}/bucket/${bucketName}`;
```
edlerd: This seems to be broken, one bucket creates this url for me: `/ui/project/undefined/storage/pools//member/default/bucket/aaa`.

We don't have a detail url for buckets at this point.

Kxiru: Implementing approach as discussed in meeting.

edlerd (approved):
QA and Code LGTM, thanks for adding this :)

PR #1310, opened 2025-06-19
feat: [WD-22253] Manage storage bucket keys

Kxiru:
## Done
Storage bucket keys can be:
[ :+1: ] created,
[ :+1: ]  edited
[ :+1: ] deleted 
[ :+1: ] bulk delete
[ :+1: ] explored in the UI.

Additionally
[ :+1: ] URL routing

> Due another round of testing with a fresh backend before marking as ready for final review(s).
- Test bulk delete behaviour :heavy_check_mark: 
- Test update Key behaviour :heavy_check_mark: 
- Verify update bucket behaviour :heavy_check_mark: 

## QA

1. Run the LXD-UI:
    - Verify CRUD operations on Storage bucket keys
    - Verify CRUD operations on Storage buckets

## Screenshots

image
image
image

### DIFF
--- a/src/api/storage-buckets.tsx
+++ b/src/api/storage-buckets.tsx
@@ -1,5 +1,5 @@
 import { handleResponse } from "util/helpers";
-import type { LxdStorageBucket } from "types/storage";
+import type { LxdStorageBucket, LxdStorageBucketKey } from "types/storage";
 import type { LxdApiResponse } from "types/apiResponse";
 import { addEntitlements } from "util/entitlements/api";
 import { fetchStoragePools } from "./storage-pools";
@@ -45,6 +45,27 @@ export const fetchAllStorageBuckets = async (
 
   const allBuckets = await Promise.all(fetches);
   return allBuckets.flat();
+};
+
+export const fetchStorageBucket = async (
+  pool: string,
+  project: string,
+  bucketName: string,
+  isFineGrained: boolean | null,
+  target?: string | null,
+): Promise<LxdStorageBucket> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+  addTarget(params, target);
+  addEntitlements(params, isFineGrained, storageBucketEntitlements);
+
+  return fetch(
+    `/1.0/storage-pools/${encodeURIComponent(pool)}/buckets/${encodeURIComponent(bucketName)}?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdStorageBucket>) => {
+      return { ...data.metadata, pool } as LxdStorageBucket;
+    });
 };
 
 export const createStorageBucket = async (
@@ -127,6 +148,131 @@ export const deleteStorageBucketBulk = async (
           .catch((e) => {
             pushFailure(results, e instanceof Error ? e.message : "");
             continueOrFinish(results, buckets.length, resolve);
+          });
+      }),
+    ).catch(reject);
+  });
+};
+
+export const fetchStorageBucketKeys = async (
+  bucket: LxdStorageBucket,
+  isFineGrained: boolean | null,
+  project: string,
+): Promise<LxdStorageBucketKey[]> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+  params.set("recursion", "1");
+  addEntitlements(params, isFineGrained, storageBucketEntitlements);
+
+  return fetch(
+    `/1.0/storage-pools/${encodeURIComponent(bucket.pool)}/buckets/${encodeURIComponent(bucket.name)}/keys?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdStorageBucketKey[]>) => {
+      return data.metadata;
+    });
+};
+
+export const fetchStorageBucketKey = async (
+  bucket: LxdStorageBucket,
+  keyName: string,
+  isFineGrained: boolean | null,
+  project: string,
+): Promise<LxdStorageBucketKey[]> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+  addEntitlements(params, isFineGrained, storageBucketEntitlements);
+
+  return fetch(
+    `/1.0/storage-pools/${encodeURIComponent(bucket.pool)}/buckets/${encodeURIComponent(bucket.name)}/keys/${encodeURIComponent(keyName)}?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdStorageBucketKey[]>) => {
+      return data.metadata;
+    });
+};
+
+export const createStorageBucketKey = async (
+  body: string,
+  project: string,
+  pool: string,
+  bucket: string,
+): Promise<LxdOperationResponse> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+
+  return fetch(
+    `/1.0/storage-pools/${encodeURIComponent(pool)}/buckets/${encodeURIComponent(bucket)}/keys?${params.toString()}`,
+    {
+      method: "POST",
+      body: body,
+    },
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
+};
+
+export const updateStorageBucketKey = async (
+  bucket: string,
+  key: LxdStorageBucketKey,
+  pool: string,
+  project: string,
+  target?: string,
+): Promise<void> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+  addTarget(params, target);
+
+  await fetch(
+    `/1.0/storage-pools/${encodeURIComponent(pool)}/buckets/${encodeURIComponent(bucket)}/keys/${encodeURIComponent(key.name)}?${params.toString()}`,
+    {
+      method: "PUT",
+      body: JSON.stringify(key),
+    },
+  ).then(handleResponse);
+};
+
+export const deleteStorageBucketKey = async (
+  bucket: string,
+  key: string,
+  pool: string,
+  project: string,
+): Promise<void> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+
+  await fetch(
+    `/1.0/storage-pools/${encodeURIComponent(pool)}/buckets/${encodeURIComponent(bucket)}/keys/${encodeURIComponent(key)}?${params.toString()}`,
+    {
+      method: "DELETE",
+    },
+  ).then(handleResponse);
+};
+
+export const deleteStorageBucketKeyBulk = async (
+  bucket: LxdStorageBucket,
+  keys: LxdStorageBucketKey[],
+  project: string,
+): Promise<PromiseSettledResult<void>[]> => {
+  const results: PromiseSettledResult<void>[] = [];
+  return new Promise((resolve, reject) => {
+    Promise.allSettled(
+      keys.map(async (key) => {
+        return deleteStorageBucketKey(
+          bucket.name,
+          key.name,
+          bucket.pool,
+          project,
+        )
+          .then(() => {
+            pushSuccess(results);
+            continueOrFinish(results, keys.length, resolve);
+          })
+          .catch((e) => {
+            pushFailure(results, e instanceof Error ? e.message : "");
+            continueOrFinish(results, keys.length, resolve);
           });
       }),
     ).catch(reject);
--- a/src/context/useBuckets.tsx
+++ b/src/context/useBuckets.tsx
@@ -2,15 +2,86 @@ import { queryKeys } from "util/queryKeys";
 import { useAuth } from "./auth";
 import type { UseQueryResult } from "@tanstack/react-query";
 import { useQuery } from "@tanstack/react-query";
-import type { LxdStorageBucket } from "types/storage";
-import { fetchAllStorageBuckets } from "api/storage-buckets";
+import type { LxdStorageBucket, LxdStorageBucketKey } from "types/storage";
+import {
+  fetchAllStorageBuckets,
+  fetchStorageBucket,
+  fetchStorageBucketKey,
+  fetchStorageBucketKeys,
+} from "api/storage-buckets";
 
-export const useLoadBuckets = (
+export const useBuckets = (
   project: string,
 ): UseQueryResult<LxdStorageBucket[]> => {
   const { isFineGrained } = useAuth();
   return useQuery({
-    queryKey: [queryKeys.buckets, project],
+    queryKey: [queryKeys.storage, project, queryKeys.buckets],
     queryFn: async () => fetchAllStorageBuckets(isFineGrained, project),
+  });
+};
+
+export const useBucket = (
+  bucketName: string,
+  pool: string,
+  project: string,
+  target?: string,
+): UseQueryResult<LxdStorageBucket> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [
+      queryKeys.storage,
+      pool,
+      project,
+      queryKeys.buckets,
+      bucketName,
+      target,
+    ],
+    queryFn: async () =>
+      fetchStorageBucket(
+        pool,
+        project,
+        bucketName,
+        isFineGrained,
+        target ?? null,
+      ),
+  });
+};
+
+export const useBucketKey = (
+  bucket: LxdStorageBucket,
+  keyName: string,
+  project: string,
+): UseQueryResult<LxdStorageBucketKey> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [
+      queryKeys.storage,
+      bucket.pool,
+      project,
+      queryKeys.buckets,
+      bucket.name,
+      queryKeys.keys,
+      keyName,
+    ],
+    queryFn: async () =>
+      fetchStorageBucketKey(bucket, keyName, isFineGrained, project),
+  });
+};
+
+export const useBucketKeys = (
+  bucket: LxdStorageBucket,
+  project: string,
+): UseQueryResult<LxdStorageBucketKey[]> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [
+      queryKeys.storage,
+      bucket.pool,
+      project,
+      queryKeys.buckets,
+      bucket.name,
+      queryKeys.keys,
+    ],
+    queryFn: async () => fetchStorageBucketKeys(bucket, isFineGrained, project),
   });
 };
--- a/src/pages/storage/StorageBucketDetail.tsx
+++ b/src/pages/storage/StorageBucketDetail.tsx
@@ -1,0 +1,79 @@
+import type { FC } from "react";
+import { useParams } from "react-router-dom";
+import { Row, useNotify } from "@canonical/react-components";
+import Loader from "components/Loader";
+import CustomLayout from "components/CustomLayout";
+import { useBucket } from "context/useBuckets";
+import StorageBucketHeader from "./StorageBucketHeader";
+import StorageBucketKeys from "./StorageBucketKeys";
+import usePanelParams, { panels } from "util/usePanelParams";
+import CreateStorageBucketKeyPanel from "./panels/CreateStorageBucketKeyPanel";
+import EditStorageBucketPanel from "./panels/EditStorageBucketPanel";
+import EditStorageBucketKeyPanel from "./panels/EditStorageBucketKeyPanel";
+
+const StorageBucketDetail: FC = () => {
+  const notify = useNotify();
+  const {
+    pool,
+    project,
+    member,
+    bucket: bucketName,
+  } = useParams<{
+    bucket: string;
+    pool: string;
+    project: string;
+    member?: string;
+  }>();
+
+  if (!pool) {
+    return <>Missing storage pool</>;
+  }
+  if (!project) {
+    return <>Missing project</>;
+  }
+  if (!bucketName) {
+    return <>Missing bucket</>;
+  }
+  const {
+    data: bucket,
+    error,
+    isLoading,
+  } = useBucket(bucketName, pool, project, member);
+
+  const panelParams = usePanelParams();
+
+  if (error) {
+    notify.failure("Loading storage bucket failed", error);
+  }
+
+  if (isLoading) {
+    return <Loader isMainComponent />;
+  } else if (!bucket) {
+    return <>Loading storage bucket failed</>;
+  }
+
+  return (
+    <>
+      <CustomLayout
+        header={<StorageBucketHeader bucket={bucket} project={project} />}
+        contentClassName="detail-page u-no-padding--bottom"
+      >
+        <Row>
+          <StorageBucketKeys bucket={bucket} />
+        </Row>
+      </CustomLayout>
+
+      {panelParams.panel === panels.editStorageBucket && (
+        <EditStorageBucketPanel bucket={bucket} />
+      )}
+      {panelParams.panel === panels.createStorageBucketKey && (
+        <CreateStorageBucketKeyPanel bucket={bucket} />
+      )}
+      {panelParams.panel === panels.editStorageBucketKey && (
+        <EditStorageBucketKeyPanel bucket={bucket} />
+      )}
+    </>
+  );
+};
+
+export default StorageBucketDetail;
--- a/src/pages/storage/StorageBucketHeader.tsx
+++ b/src/pages/storage/StorageBucketHeader.tsx
@@ -1,0 +1,33 @@
+import type { FC } from "react";
+import { Link } from "react-router-dom";
+import RenameHeader from "components/RenameHeader";
+import type { LxdStorageBucket } from "types/storage";
+import StorageBucketActions from "./actions/StorageBucketActions";
+
+interface Props {
+  bucket: LxdStorageBucket;
+  project: string;
+}
+
+const StorageBucketHeader: FC<Props> = ({ bucket, project }) => {
+  return (
+    <RenameHeader
+      name={bucket.name}
+      parentItems={[
+        <Link
+          to={`/ui/project/${encodeURIComponent(project)}/storage/buckets`}
+          key={1}
+        >
+          Storage buckets
+        </Link>,
+      ]}
+      controls={
+        bucket ? <StorageBucketActions bucket={bucket} isDetailPage /> : null
+      }
+      isLoaded={true}
+      renameDisabledReason="Storage buckets cannot be renamed"
+    />
+  );
+};
+
+export default StorageBucketHeader;
--- a/src/pages/storage/StorageBucketKeys.tsx
+++ b/src/pages/storage/StorageBucketKeys.tsx
@@ -1,0 +1,263 @@
+import type { FC } from "react";
+import { useEffect, useState } from "react";
+import {
+  EmptyState,
+  Icon,
+  SearchBox,
+  TablePagination,
+  useNotify,
+} from "@canonical/react-components";
+import ItemName from "components/ItemName";
+import SelectableMainTable from "components/SelectableMainTable";
+import { useCurrentProject } from "context/useCurrentProject";
+import ScrollableTable from "components/ScrollableTable";
+import SelectedTableNotification from "components/SelectedTableNotification";
+import { useDocs } from "context/useDocs";
+import useSortTableData from "util/useSortTableData";
+import NotificationRow from "components/NotificationRow";
+import CreateStorageBucketKeyBtn from "./actions/CreateStorageBucketKeyBtn";
+import type { LxdStorageBucket } from "types/storage";
+import { useBucketKeys } from "context/useBuckets";
+import StorageBucketKeyActions from "./actions/StorageBucketKeyActions";
+import StorageBucketKeyBulkDelete from "./actions/StorageBucketKeyBulkDelete";
+import { capitalizeFirstLetter } from "util/helpers";
+
+interface Props {
+  bucket: LxdStorageBucket;
+}
+
+const StorageBucketKeys: FC<Props> = ({ bucket }) => {
+  const docBaseLink = useDocs();
+  const [query, setQuery] = useState<string>("");
+  const notify = useNotify();
+  const [selectedNames, setSelectedNames] = useState<string[]>([]);
+  const [processingNames, setProcessingNames] = useState<string[]>([]);
+  const { project } = useCurrentProject();
+  const { data: keys = [] } = useBucketKeys(bucket, project?.name ?? "");
+
+  useEffect(() => {
+    const validNames = new Set(keys?.map((key) => key.name));
+    const validSelections = selectedNames.filter((name) =>
+      validNames.has(name),
+    );
+    if (validSelections.length !== selectedNames.length) {
+      setSelectedNames(validSelections);
+    }
+  }, [keys]);
+
+  const filteredKeys =
+    keys?.filter((key) => {
+      if (query) {
+        const q = query.toLowerCase();
+
+        return (
+          key.name.toLowerCase().includes(q) ||
+          key.role.toLowerCase().includes(q) ||
+          key.description.toLowerCase().includes(q) ||
+          key["access-key"].toLowerCase().includes(q) ||
+          key["secret-key"].toLowerCase().includes(q)
+        );
+      }
+      return true;
+    }) ?? [];
+
+  const hasKeys = keys && keys.length > 0;
+
+  const headers = [
+    {
+      content: "Name",
+      sortKey: "name",
+      className: "name",
+    },
+    {
+      content: "Role",
+      sortKey: "role",
+      className: "role",
+    },
+    {
+      content: "Description",
+      sortKey: "description",
+      className: "description",
+    },
+    { content: "Access key", className: "key-field", sortKey: "access-key" },
+    { content: "Secret key", className: "key-field", sortKey: "secret-key" },
+    { "aria-label": "Actions", className: "actions" },
+  ];
+
+  const rows = filteredKeys.map((key) => {
+    return {
+      key: key.name,
+      className: "u-row",
+      name: key.name,
+      columns: [
+        {
+          content: (
+            <div className="u-truncate" title={`Key ${key.name}`}>
+              <ItemName item={key} />
+            </div>
+          ),
+          role: "rowheader",
+          "aria-label": "Name",
+          className: "name",
+        },
+        {
+          content: capitalizeFirstLetter(key.role),
+          role: "cell",
+          "aria-label": "Role",
+          className: "role",
+        },
+        {
+          content: key.description || "-",
+          title: key.description,
+          role: "cell",
+          "aria-label": "Description",
+          className: "description u-truncate",
+        },
+        {
+          content: key["access-key"],
+          role: "cell",
+          "aria-label": "Access key",
+          className: "key-field",
+        },
+        {
+          content: key["secret-key"],
+          role: "cell",
+          "aria-label": "Secret key",
+          className: "key-field",
+        },
+        {
+          content: <StorageBucketKeyActions bucketKey={key} bucket={bucket} />,
+          role: "cell",
+          "aria-label": "Actions",
+          className: "u-align--right actions",
+        },
+      ],
+      sortData: {
+        name: key.name.toLowerCase(),
+        role: key.role,
+        description: key.description.toLowerCase(),
+        "access-key": key["access-key"],
+        "secret-key": key["secret-key"],
+      },
+    };
+  });
+
+  const selectedKeys = keys.filter((key) => {
+    return selectedNames.includes(key.name);
+  });
+
+  const { rows: sortedRows, updateSort } = useSortTableData({
+    rows,
+    defaultSort: "name",
+    defaultSortDirection: "ascending",
+  });
+
+  return (
+    <div className="storage-bucket-key-list">
+      {hasKeys && (
+        <div className="upper-controls-bar">
+          {selectedNames.length === 0 ? (
+            <>
+              <div className="search-box-wrapper">
+                <SearchBox
+                  name="search-snapshot"
+                  className="search-box margin-right"
+                  type="text"
+                  onChange={(value) => {
+                    setQuery(value);
+                  }}
+                  placeholder="Search for keys"
+                  value={query}
+                  aria-label="Search for keys"
+                />
+              </div>
+              <CreateStorageBucketKeyBtn />
+            </>
+          ) : (
+            <div className="p-panel__controls">
+              <StorageBucketKeyBulkDelete
+                keys={selectedKeys}
+                bucket={bucket}
+                onStart={() => {
+                  setProcessingNames(selectedNames);
+                }}
+                onFinish={() => {
+                  setProcessingNames([]);
+                }}
+              />
+            </div>
+          )}
+        </div>
+      )}
+      <NotificationRow />
+      {hasKeys ? (
+        <>
+          <ScrollableTable
+            dependencies={[filteredKeys, notify.notification]}
+            tableId="buckets-key-table"
+            belowIds={["status-bar"]}
+          >
+            <TablePagination
+              data={sortedRows}
+              id="pagination"
+              itemName="key"
+              className="u-no-margin--top"
+              aria-label="Table pagination control"
+              description={
+                selectedNames.length > 0 && (
+                  <SelectedTableNotification
+                    totalCount={keys?.length ?? 0}
+                    itemName="key"
+                    parentName="bucket"
+                    selectedNames={selectedNames}
+                    setSelectedNames={setSelectedNames}
+                    filteredNames={filteredKeys.map((item) => item.name)}
+                  />
+                )
+              }
+            >
+              <SelectableMainTable
+                id="buckets-key-table"
+                headers={headers}
+                rows={sortedRows}
+                sortable
+                emptyStateMsg="No key found matching this search"
+                itemName="key"
+                parentName="bucket"
+                selectedNames={selectedNames}
+                setSelectedNames={setSelectedNames}
+                disabledNames={processingNames}
+                filteredNames={filteredKeys.map((item) => item.name)}
+                onUpdateSort={updateSort}
+                defaultSort="name"
+                defaultSortDirection="ascending"
+                responsive
+              />
+            </TablePagination>
+          </ScrollableTable>
+        </>
+      ) : (
+        <EmptyState
+          className="empty-state"
+          image={<Icon name="private-key" className="empty-state-icon" />}
+          title="No keys"
+        >
+          <p>This bucket does not contain any keys.</p>
+          <p>
+            <a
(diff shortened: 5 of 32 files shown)


src/context/useBuckets.tsx
```diff
@@ -14,3 +14,24 @@ export const useLoadBuckets = (
...
+  project: string,
+  type: string,
+  volume: string,
+  target?: string,
+): UseQueryResult<LxdStorageBucket> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.storage, pool, project, type, volume, target],
+    queryFn: async () =>
+      fetchStorageVolume(
```
edlerd: should fetch a bucket

src/context/useBuckets.tsx
```diff
@@ -14,3 +14,24 @@ export const useLoadBuckets = (
     queryFn: async () => fetchAllStorageBuckets(isFineGrained, project),
   });
 };
+export const useStorageBucket = (
+  pool: string,
+  project: string,
+  type: string,
+  volume: string,
```
edlerd:
```suggestion
  bucket: string,
```

src/context/useBuckets.tsx
```diff
@@ -14,3 +14,24 @@ export const useLoadBuckets = (
...
+      fetchStorageVolume(
+        pool,
+        project,
+        type,
+        volume,
+        target ?? null,
+        isFineGrained,
+      ),
+  });
+};
```
edlerd:
```suggestion
};

```

src/pages/storage/StorageBucketDetail.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  if (!pool) {
+    return <>Missing storage pool</>;
+  }
+  if (!project) {
+    return <>Missing project</>;
+  }
+  if (!type) {
+    return <>Missing type</>;
+  }
```
edlerd:
```suggestion
```
Buckets don't have a type.

src/pages/storage/StorageBucketLink.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { LxdStorageBucket } from "types/storage";
+
+interface Props {
+  bucket: LxdStorageBucket;
+}
+
+const StorageBucketLink: FC<Props> = ({ bucket }) => {
+  return (
+    <Link
+      to={`/ui/project/${bucket.project}/storage/pool/${bucket.pool}/bucket/${bucket.name}`}
```
edlerd: This should use the shared function to generate the link from the bucket object.

src/pages/storage/StorageBucketKeys.tsx
edlerd, line 1: This needs adjustments

edlerd (changes requested):
some early review notes below.

src/context/useBuckets.tsx
```diff
@@ -14,3 +18,34 @@ export const useLoadBuckets = (
...
+        pool,
+        project,
+        bucketName,
+        isFineGrained,
+        target ?? null,
+      ),
+  });
+};
+
+export const useLoadBucketKeys = (
```
edlerd:
```suggestion
export const useBucketKeys = (
```
More concise, the loading is implied.

src/context/useBuckets.tsx
```diff
@@ -14,3 +18,34 @@ export const useLoadBuckets = (
...
+  bucket: LxdStorageBucket,
+  project: string,
+): UseQueryResult<LxdStorageBucketKey[]> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.buckets, project],
```
edlerd: This key clashes with others. It needs to uniquely identify the query done below. So we need all things that identify the `fetchStorageBucketKeys` call uniquely. That is `bucket`, `isFineGrained` (though this one won't change ever, so we can omit it) and `project` (which you already have).

It would also be nice to add a new `queryKeys.key` constant and use it here, so we can identify all keys in the cache.

Kxiru: I have it already actually, I just forgot to add it here. It is called bucketKeys. Thank you for reminding me!

src/context/useBuckets.tsx
```diff
@@ -14,3 +18,34 @@ export const useLoadBuckets = (
...
+};
+
+export const useBucketKeys = (
+  bucket: LxdStorageBucket,
+  project: string,
+): UseQueryResult<LxdStorageBucketKey[]> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.buckets, project, queryKeys.bucketKeys],
+    queryFn: async () => fetchStorageBucketKeys(bucket, isFineGrained, project),
```
edlerd: The bucket name and bucket pool are missing from the query key. So if we call this with different buckets, the cache would overlap instead of identify each buckets keys uniquely.

src/context/useBuckets.tsx
```diff
@@ -14,3 +18,34 @@ export const useLoadBuckets = (
...
+        pool,
+        project,
+        bucketName,
+        isFineGrained,
+        target ?? null,
+      ),
+  });
+};
+
+export const useBucketKeys = (
```
edlerd: Shall we unify the naming a bit? I'd propose `useBucket` and `useBucketKeys`.

src/pages/storage/actions/DeleteStorageBucketKeyBtn.tsx
```diff
@@ -0,0 +1,90 @@
...
+      projectName,
+    )
+      .then(onFinish)
+      .catch((e) => {
+        notify.failure("Bucket key deletion failed", e);
+      })
+      .finally(() => {
+        setLoading(false);
+        queryClient.invalidateQueries({
+          queryKey: [queryKeys.buckets, project, queryKeys.bucketKeys],
```
edlerd:
```suggestion
          queryKey: [queryKeys.buckets, projectName, queryKeys.bucketKeys],
```
This should be the name to clear the right thing.

src/pages/storage/panels/CreateStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const keySchema = Yup.object().shape({
+    name: Yup.string().required("Key name is required"),
+  });
+
+  const handleSuccess = (keyName: string) => {
+    toastNotify.success(
+      <>
+        Key{" "}
+        <ResourceLink
+          type="bucket" //Change to what?
```
edlerd: We should define a new type bucket-key. I guess we have a good matching icon with private key:

image

src/pages/storage/panels/CreateStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,130 @@
...
+          value={keyName}
+          to={``} //Open edit panel for the identity.
+        />{" "}
+        created.
```
edlerd: This should better say "Key $name created for bucket $bucket".

src/pages/storage/panels/CreateStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,130 @@
...
+  });
+
+  const handleSuccess = (keyName: string) => {
+    toastNotify.success(
+      <>
+        Key{" "}
+        <ResourceLink
+          type="bucket" //Change to what?
+          value={keyName}
+          to={``} //Open edit panel for the identity.
```
edlerd: This should link to the key list for the bucket.

src/pages/storage/actions/StorageBucketKeyActions.tsx
```diff
@@ -0,0 +1,53 @@
...
+      classname={classnames(className, "has-icon")}
+      bucket={bucket}
+      onFinish={() => {
+        toastNotify.success(
+          <>
+            Bucket key{" "}
+            <ResourceLabel bold type="bucket" value={bucketKey.name} /> deleted.
+          </>,
+        );
+      }}
```
edlerd: I don't think we should expose this in the interface. We do it in other cases, because we want to use the delete button in different contexts. Here we only use it on the list page, so the on finish logic can be embedded in the `DeleteStorageBucketKeyBtn` component.

Copy should be updated to "Key $name deleted for bucket $bucket".

src/pages/storage/actions/StorageBucketActions.tsx
```diff
@@ -1,44 +1,101 @@
...
+
+  const menuElements = [
+    <EditStorageBucketBtn
+      key="edit"
+      classname={classnames(classname, "has-icon", {
+        "is-dense": !isDetailPage,
+      })}
+      bucket={bucket}
+      isDetailPage={isDetailPage}
+    />,
```
edlerd, line 41: There is a bug with editing from the detail page. When update size and/or description and saving, then again opening the edit panel, the saved changes are not updated in the form -- but the previous settings are displayed instead.

src/pages/storage/actions/StorageBucketKeyBulkDelete.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { project } = useCurrentProject();
+  const projectName = project?.name || "";
+  const totalCount = keys.length;
+
+  const buttonText = `Delete ${keys.length} ${pluralize("key", keys.length)}`;
+
+  const handleDelete = () => {
+    setLoading(true);
+    onStart();
+    const successMessage = `${keys.length} ${pluralize("key", keys.length)} successfully deleted`;
```
edlerd: Copy should include "for bucket $bucket".

src/pages/storage/actions/StorageBucketKeyBulkDelete.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+        queryClient.invalidateQueries({
+          queryKey: [queryKeys.buckets, projectName, queryKeys.bucketKeys],
+        });
+        setLoading(false);
+        onFinish();
+      })
+      .catch((e) => {
+        setLoading(false);
+        toastNotify.failure("Bucket bulk deletion failed", e);
```
edlerd: Should be "Key bulk deletion failed for bucket $bucket"

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+  });
+
+  const selectedKeys = bucketKeys.filter((key) => {
+    const identifierKey = getKeyIdentifier(key);
+    return selectedNames.includes(identifierKey);
+  });
+
+  const { rows: sortedRows, updateSort } = useSortTableData({
+    rows,
+    defaultSort: "role",
```
edlerd: I think this is unexpected. Shouldn't we sort by name?

src/api/storage-buckets.tsx
edlerd, line 1: This needs to be adjusted with the new patterns from https://github.com/canonical/lxd-ui/pull/1335

Though we should probably let the other PR land first so we can use the functions.

src/pages/storage/actions/StorageBucketActions.tsx
edlerd: The reason is we invalidate only the `[queryKeys.buckets, project?.name]` key in the edit storage bucket panel. But from the detail page we load a single bucket with `useBucket`, which is another query key for just the single bucket fetch. We should also invalidate that key from the edit panel.

src/context/useBuckets.tsx
```diff
@@ -14,3 +19,53 @@ export const useLoadBuckets = (
...
+
+export const useBucket = (
+  bucketName: string,
+  pool: string,
+  project: string,
+  target?: string,
+): UseQueryResult<LxdStorageBucket> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.storage, pool, project, bucketName, target],
```
edlerd: I think the query keys in this file should be a bit more persistent. Suggesting below.
```suggestion
    queryKey: [queryKeys.storage, pool, project, queryKeys.bucket, bucketName, target],
```

src/context/useBuckets.tsx
```diff
@@ -14,3 +19,53 @@ export const useLoadBuckets = (
...
+};
+
+export const useBucketKey = (
+  bucket: LxdStorageBucket,
+  keyName: string,
+  project: string,
+): UseQueryResult<LxdStorageBucketKey> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [bucket.name, bucket.pool, keyName],
```
edlerd:
```suggestion
    queryKey: [queryKeys.storage, pool, project, queryKeys.bucket, bucketName, target, queryKeys.bucketKeys, keyName],
```

src/context/useBuckets.tsx
```diff
@@ -14,3 +19,53 @@ export const useLoadBuckets = (
...
+): UseQueryResult<LxdStorageBucketKey[]> => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [
+      queryKeys.buckets,
+      project,
+      queryKeys.bucketKeys,
+      bucket.name,
+      bucket.pool,
+    ],
```
edlerd:
```suggestion
    queryKey: [queryKeys.storage, pool, project, queryKeys.bucket, bucketName, target, queryKeys.bucketKeys],
```

src/pages/storage/StorageBucketHeader.tsx
```diff
@@ -0,0 +1,30 @@
...
+      parentItems={[
+        <Link to={`/ui/project/${project}/storage/buckets`} key={1}>
+          Storage buckets
+        </Link>,
+      ]}
+      controls={
+        bucket ? <StorageBucketActions bucket={bucket} isDetailPage /> : null
+      }
+      isLoaded={true}
+      renameDisabledReason={"Storage buckets cannot be renamed"}
```
edlerd:
```suggestion
      renameDisabledReason="Storage buckets cannot be renamed"
```

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+const StorageBucketKeys: FC<Props> = ({ bucket }) => {
+  const docBaseLink = useDocs();
+  const [query, setQuery] = useState<string>("");
+  const notify = useNotify();
+  const [selectedNames, setSelectedNames] = useState<string[]>([]);
+  const [processingNames, setProcessingNames] = useState<string[]>([]);
+  const [isSmallScreen, setSmallScreen] = useState(figureCollapsedScreen());
+  const { project } = useCurrentProject();
+  const { data: bucketKeys = [] } = useBucketKeys(bucket, project?.name ?? "");
```
edlerd:
```suggestion
  const { data: keys = [] } = useBucketKeys(bucket, project?.name ?? "");
```
Might be simpler.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+  }, [bucketKeys]);
+
+  const filteredKeys =
+    bucketKeys?.filter((key) => {
+      if (query) {
+        if (!key.name.toLowerCase().includes(query.toLowerCase())) {
+          return false;
+        }
+      }
+      return true;
```
edlerd: I think we should also match the description and the secret/access key as well as the role.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+    }) ?? [];
+
+  const getKeyIdentifier = (key: LxdStorageBucketKey) => {
+    return `${key.name}-${bucket.name || ""}`;
+  };
```
edlerd:
```suggestion
```
We don't need this, as we only ever have keys from one bucket per this component. So we can use `key.name` directly instead.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+  bucket: LxdStorageBucket;
+}
+
+const StorageBucketKeys: FC<Props> = ({ bucket }) => {
+  const docBaseLink = useDocs();
+  const [query, setQuery] = useState<string>("");
+  const notify = useNotify();
+  const [selectedNames, setSelectedNames] = useState<string[]>([]);
+  const [processingNames, setProcessingNames] = useState<string[]>([]);
+  const [isSmallScreen, setSmallScreen] = useState(figureCollapsedScreen());
```
edlerd: Responsiveness is broken on this page

image

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+          content: actions,
+          role: "cell",
+          "aria-label": "Actions",
+          className: "u-align--right actions",
+        },
+      ],
+      sortData: {
+        name: key.name.toLowerCase(),
+        role: key.role,
+      },
```
edlerd: We should also support sort by Description	Access key	Secret key.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+            </TablePagination>
+          </ScrollableTable>
+        </>
+      ) : (
+        <EmptyState
+          className="empty-state"
+          image={<Icon name="private-key" className="empty-state-key" />}
+          title="No keys"
+        >
+          <p>This bucket doesn&apos;t have any keys yet.</p>
```
edlerd:
```suggestion
          <p>There are no keys for this bucket.</p>
```
Slightly simpler to read?

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,289 @@
...
+          <p>
+            <a
+              href={`${docBaseLink}/explanation/storage/`}
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              Read more
```
edlerd:
```suggestion
              Learn more about storage
```

src/pages/storage/StorageBuckets.tsx
edlerd, line 1: This needs to be responsive, currently it hides when the screen width is too small

image

edlerd (commented):
Thanks for the new features, some comments as discussed below :)

src/pages/storage/StorageBucketKeys.tsx
Kxiru: How about "This bucket does not contain any keys"?

src/pages/storage/StorageBucketKeys.tsx
Kxiru: I too prefer this, although it is a deviation from the figma design.

src/pages/storage/panels/EditStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+      role: bucketKey?.role,
+      description: bucketKey?.description,
+    },
+    enableReinitialize: true,
+    onSubmit: (values) => {
+      const keyPayload = {
+        name: values.name,
+        role: values.role,
+        description: values.description,
+      } as LxdStorageBucketKey;
```
edlerd:
```suggestion
      const keyPayload = {
        name: values.name,
        role: values.role,
        description: values.description,
        "access-key": bucketKey?.["access-key"] ?? "",
        "secret-key": bucketKey?.["secret-key"] ?? "",
      } as LxdStorageBucketKey;
```
Let's add the keys here, so they don't get regenerated on edit. 

Acutally I think we should add two fields for the keys to the form, so they can be manually changed.

src/pages/storage/panels/EditStorageBucketPanel.tsx
```diff
@@ -73,7 +73,13 @@ const EditStorageBucketPanel: FC<Props> = ({ bucket }) => {
...
           queryClient.invalidateQueries({
-            queryKey: [queryKeys.buckets, project?.name],
+            queryKey: [
+              queryKeys.storage,
+              bucket.pool,
+              project?.name ?? "",
+              queryKeys.buckets,
+              bucket.name,
+            ],
           });
```
edlerd: We need both, the key for the individual bucket and the list here. As the details that are updated are also part of the list query response:

```

          queryClient.invalidateQueries({
            queryKey: [
              queryKeys.storage,
              project?.name ?? "",
              queryKeys.buckets,
            ],
          });
```

src/pages/storage/actions/DeleteStorageBucketBtn.tsx
edlerd, line 1: We need to invalidate the right query key now that it is changed. Should invalidate `queryKeys.storage, projectName,queryKeys.buckets` now, instead of `queryKeys.buckets, projectName`.

src/pages/storage/panels/CreateStorageBucketPanel.tsx
edlerd, line 1: Should invalidate `queryKeys.storage, projectName,queryKeys.buckets` now, instead of `queryKeys.buckets, projectName`.

src/pages/storage/panels/CreateStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+        panelParams.project,
+        bucket.pool,
+        bucket.name,
+      )
+        .then(() => {
+          queryClient.invalidateQueries({
+            queryKey: [
+              queryKeys.buckets,
+              panelParams.project,
+              queryKeys.bucketKeys,
```
edlerd: This should match the key list for the bucket from `useBucketKeys`:
```suggestion
      queryKeys.storage,
      bucket.pool,
      panelParams.project,
      queryKeys.buckets,
      bucket.name,
      queryKeys.bucketKeys
```

src/sass/_empty_state.scss
```diff
@@ -8,6 +8,12 @@
     width: 2.5rem;
   }
 
+  .empty-state-key {
+    height: 4rem;
+    margin-top: $spv--medium;
+    width: 4rem;
+  }
+
```
edlerd: We should reuse the existing `empty-state-icon` instead. It looks a little different but should be more consistent.

src/util/queryKeys.tsx
```diff
@@ -1,5 +1,6 @@
 export const queryKeys = {
   buckets: "buckets",
+  bucketKeys: "bucket-keys",
```
edlerd:
```suggestion
  keys: "keys",
```
I think this should be just keys, we always use together with buckets so it works fine context as a bucket key.

src/util/usePanelParams.tsx
```diff
@@ -40,6 +43,8 @@ export const panels = {
   createTLSIdentity: "create-tls-identity",
   createStorageBucket: "create-storage-bucket",
   editStorageBucket: "edit-storage-bucket",
+  createStorageBucketKey: "create-bucket-key",
+  editStorageBucketKey: "edit-bucket-key",
```
edlerd: I think we should shorten the ones on top to be `create-bucket` and `edit-bucket`. Then they are more concise in the url and aligned with the keys pattern.

src/pages/storage/panels/EditStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+            queryKey: [
+              queryKeys.buckets,
+              project,
+              queryKeys.bucketKeys,
+              bucket.name,
+              bucket.pool,
+            ],
+          });
+          handleSuccess(bucket);
+        })
```
edlerd: This needs to invalidate 1. the new key list and 2. the new key detail query key.

src/pages/storage/panels/EditStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const handleSuccess = (bucket: LxdStorageBucket) => {
+    toastNotify.success(
+      <>
+        Key{" "}
+        <ResourceLink
+          type="bucket"
+          value={bucket.name}
+          to={`/ui/project/${project}/storage/pool/${bucket.pool}/bucket/${bucket.name}`}
+        />{" "}
+        updated.
```
edlerd: this should be "Key $key updated for bucket $bucket".

src/pages/storage/actions/DeleteStorageBucketKeyBtn.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { canEditBucket } = useStorageBucketEntitlements();
+  const { project } = useCurrentProject();
+  const projectName = project?.name || "";
+  const toastNotify = useToastNotification();
+
+  const onFinish = () => {
+    toastNotify.success(
+      <>
+        Bucket key <ResourceLabel bold type="bucket" value={bucketKey.name} />{" "}
+        deleted for bucket {bucket.name}.
```
edlerd: The bucket name should be a resource link.

src/pages/storage/panels/CreateStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+    notify.clear();
+  };
+  const bucketURL = getStorageBucketURL(
+    bucket.name,
+    bucket.pool,
+    project?.name ?? "",
+  );
+
+  const keySchema = Yup.object().shape({
+    name: Yup.string().required("Key name is required"),
```
edlerd: This should also warn on duplicate keys.

src/pages/storage/actions/StorageBucketKeyBulkDelete.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const buttonText = `Delete ${keys.length} ${pluralize("key", keys.length)}`;
+
+  const handleDelete = () => {
+    setLoading(true);
+    onStart();
+    const successMessage = (
+      <>
+        {keys.length} {pluralize("key", keys.length)} successfully deleted for
+        bucket{" "}
```
edlerd:
```suggestion
        {keys.length} {pluralize("key", keys.length)} deleted for
        bucket{" "}
```
In a success notification we can leave out the "successfully". Otherwise it is a bit redundant:
image

src/pages/storage/actions/StorageBucketKeyActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+import classnames from "classnames";
+import { List } from "@canonical/react-components";
+import type { LxdStorageBucket, LxdStorageBucketKey } from "types/storage";
+import DeleteStorageBucketKeyBtn from "./DeleteStorageBucketKeyBtn";
+import EditStorageBucketKeyBtn from "./EditStorageBucketKeyBtn";
+
+interface Props {
+  bucketKey: LxdStorageBucketKey;
+  bucket: LxdStorageBucket;
+  className?: string;
```
edlerd:
```suggestion
```
This seems unused, we should probably avoid it to simplify?

src/pages/storage/actions/StorageBucketKeyActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+const StorageBucketKeyActions: FC<Props> = ({
+  bucketKey,
+  bucket,
+  className,
+}) => {
+  const menuElements = [
+    <EditStorageBucketKeyBtn
+      key="edit"
+      bucketKey={bucketKey}
+      classname={classnames(className, "has-icon")}
```
edlerd:
```suggestion
```
I think this class can move into the `EditStorageBucketKeyBtn`, no need to have it here on the interface.

src/pages/storage/actions/StorageBucketKeyActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+      bucket={bucket}
+    />,
+    <DeleteStorageBucketKeyBtn
+      key="delete"
+      bucketKey={bucketKey}
+      classname={classnames(className, "has-icon")}
```
edlerd:
```suggestion
```
I think this class can move into the `DeleteStorageBucketKeyBtn`, no need to have it here on the interface.

src/pages/storage/actions/EditStorageBucketKeyBtn.tsx
```diff
@@ -0,0 +1,44 @@
...
+  bucketKey,
+  classname,
+  isDetailPage,
+}) => {
+  const panelParams = usePanelParams();
+  const { canEditBucket } = useStorageBucketEntitlements();
+
+  return (
+    <Button
+      key={`${bucket.name}-${bucketKey.name}-edit`}
```
edlerd:
```suggestion
```
We probably don't need the key here. This is just a single component. Keys are mostly needed if we have multiple siblings inside a component.

src/pages/storage/actions/EditStorageBucketKeyBtn.tsx
```diff
@@ -0,0 +1,44 @@
...
+import type { LxdStorageBucket, LxdStorageBucketKey } from "types/storage";
+import { Button, Icon } from "@canonical/react-components";
+import { useStorageBucketEntitlements } from "util/entitlements/storage-buckets";
+import usePanelParams from "util/usePanelParams";
+
+interface Props {
+  bucketKey: LxdStorageBucketKey;
+  bucket: LxdStorageBucket;
+  classname?: string;
+  isDetailPage?: boolean;
```
edlerd:
```suggestion
```
This seems unused, best to remove it.

src/pages/storage/actions/DeleteStorageBucketKeyBtn.tsx
```diff
@@ -0,0 +1,107 @@
...
+  classname?: string;
+  isDetailPage?: boolean;
+}
+
+const DeleteStorageBucketKeyBtn: FC<Props> = ({
+  bucket,
+  bucketKey,
+  hasIcon = true,
+  classname,
+  isDetailPage,
```
edlerd:
```suggestion
```
We only have one use case, where all three of these can be defined, no need to make it too configurable from the outside. Let's keep the complexity low and this component interface simple.

src/pages/storage/actions/DeleteStorageBucketBtn.tsx
```diff
@@ -7,34 +7,45 @@ import {
...
-  label?: string;
   classname?: string;
+  isDetailPage?: boolean;
 }
 
 const DeleteStorageBucketBtn: FC<Props> = ({
   bucket,
-  onFinish,
-  appearance = "base",
   hasIcon = true,
```
edlerd:
```suggestion
```
Seems unused.

src/pages/storage/actions/CreateStorageBucketKeyBtn.tsx
```diff
@@ -0,0 +1,39 @@
...
+      hasIcon={!isSmallScreen}
+      onClick={() => {
+        panelParams.openCreateStorageBucketKey(project?.name || "");
+      }}
+      className={className}
+      disabled={!canCreateStorageBuckets(project)}
+      title={
+        canCreateStorageBuckets(project)
+          ? "Create bucket key"
+          : "You do not have permission to create keys in this project"
```
edlerd:
```suggestion
          : "You do not have permission to create keys for this bucket"
```
Permissions can be granular, so better be precise here.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,295 @@
...
+                emptyStateMsg="No key found matching this search"
+                itemName="key"
+                parentName="bucket"
+                selectedNames={selectedNames}
+                setSelectedNames={setSelectedNames}
+                disabledNames={processingNames}
+                filteredNames={filteredKeys.map((item) => item.name)}
+                onUpdateSort={updateSort}
+                defaultSort="role"
+                defaultSortDirection="descending"
```
edlerd:
```suggestion
                defaultSort="name"
                defaultSortDirection="ascending"
```
I think this is a more sensible default and should be aligned with the default sort above.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,295 @@
...
+          content: actions,
+          role: "cell",
+          "aria-label": "Actions",
+          className: "u-align--right actions",
+        },
+      ],
+      sortData: {
+        name: key.name.toLowerCase(),
+        role: key.role,
+        description: key.description,
```
edlerd:
```suggestion
        description: key.description.toLowerCase(),
```
Should be case-insensitive sorting.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,295 @@
...
+      className: "description",
+      sortKey: "description",
+    },
+    { content: "Access key", className: "key-field", sortKey: "access-key" },
+    { content: "Secret key", className: "key-field", sortKey: "secret-key" },
+    { "aria-label": "Actions", className: "actions" },
+  ];
+
+  const rows = filteredKeys.map((key) => {
+    const actions = <StorageBucketKeyActions bucketKey={key} bucket={bucket} />;
```
edlerd: This can be inlined below, probably simpler.

src/context/useBuckets.tsx
```diff
@@ -2,15 +2,86 @@ import { queryKeys } from "util/queryKeys";
...
-import { fetchAllStorageBuckets } from "api/storage-buckets";
+import type { LxdStorageBucket, LxdStorageBucketKey } from "types/storage";
+import {
+  fetchAllStorageBuckets,
+  fetchStorageBucket,
+  fetchStorageBucketKey,
+  fetchStorageBucketKeys,
+} from "api/storage-buckets";
 
 export const useLoadBuckets = (
```
edlerd:
```suggestion
export const useBuckets = (
```
Shall we rename this as well to unify the names?

edlerd (commented):
Very good progress on this one. Did a detailed code review with some ideas for simplification below. Found some more QA issues, mostly fallout from the query key adjustments. The responsiveness is also still open, we can collaborate on that one.

src/pages/storage/actions/DeleteStorageBucketBtn.tsx
Kxiru: Why does queryKeys.buckets need to be repeated twice?

src/pages/storage/panels/EditStorageBucketKeyPanel.tsx
edlerd: So call `queryClient.invalidateQueries` twice with the query key matching `useBucketKey` for one and `useBucketKeys` for the other.

src/pages/storage/actions/DeleteStorageBucketBtn.tsx
edlerd: I mixed up the query key before, fixed it in the message above just now.

src/pages/storage/actions/DeleteStorageBucketKeyBtn.tsx
Kxiru: I think isDetailPage needs to remain to decide which button appearance to have, and also whether to show the "Delete" copy, as this component serves for both these delete buttons:

image

and

image

edlerd: This is for deleting the key -- I think it is only ever used with the icon and no copy/label.

src/pages/storage/StorageBucketKeys.tsx
edlerd: This has improved, but there are some screen ranges, where the actions are not available. Seems like the breakpoints need a little bit more tweaking:

image

image

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,295 @@
...
+      className: "u-row",
+      name: key.name,
+      columns: [
+        {
+          content: (
+            <>
+              <div className="u-truncate" title={`Key ${key.name}`}>
+                <ItemName item={key} />
+              </div>
+              {isSmallScreen && key.role}
```
edlerd: Can we add the muted text css class to the role, to make it easier to scan by name in the collapsed form?

Also, we probably shouldn't collapse in the mobile view with cards.

image

src/pages/storage/panels/EditStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  if (isLoading) {
+    return <Loader isMainComponent />;
+  } else if (!bucketKey) {
+    return <>Loading key failed</>;
+  }
+  const changeCount =
+    (formik.values.description !== bucketKey.description ? 1 : 0) +
+    (formik.values.role !== bucketKey.role ? 1 : 0);
```
edlerd: We should add the two new inputs for secrets here.

src/pages/storage/forms/StorageBucketKeyForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+      placeholder: `Enter ${id.replaceAll("-", " ")}`,
+    };
+  };
+
+  const isEditing = !!bucket;
+
+  const bucketEditRestriction =
+    !isEditing || canEditBucket(bucket)
+      ? ""
+      : "You do not have permission to modify this bucket";
```
edlerd:
```suggestion
      : "You do not have permission to edit this bucket";
```
I think we should be consistent with "edit".

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,295 @@
...
+          ),
+          role: "rowheader",
+          "aria-label": "Name",
+          className: "name",
+        },
+        ...(isSmallScreen
+          ? []
+          : [
+              {
+                content: key.role,
```
edlerd: Here and in the collapsed above, we should call `capitalizeFirstLetter`. So the presentation is the same as in the edit/create form.
```suggestion
                content: capitalizeFirstLetter(key.role),
```

src/util/storageBucket.tsx
```diff
@@ -28,6 +29,34 @@ export const testDuplicateStorageBucketName = (
...
+    "A key with this name already exists",
+    async (value?: string, context?: TestContext) => {
+
+      return (
+        (excludeName && value === excludeName) ||
+        checkDuplicateName(
+          value,
+          project,
+          controllerState,
+          `storage/pool/${encodeURIComponent(bucket.pool)}/bucket/${encodeURIComponent(bucket.name)}`,
```
edlerd:
```suggestion
          `storage-pools/${encodeURIComponent(bucket.pool)}/buckets/${encodeURIComponent(bucket.name)}/keys`,
```
The test url is not a valid path.

src/util/storageBucket.tsx
```diff
@@ -28,6 +29,34 @@ export const testDuplicateStorageBucketName = (
...
+  project: string,
+  bucket: LxdStorageBucket,
+  controllerState: AbortControllerState,
+  excludeName?: string,
+): [string, string, TestFunction<string | undefined, AnyObject>] => {
+  return [
+    "deduplicate",
+    "A key with this name already exists",
+    async (value?: string, context?: TestContext) => {
+
```
edlerd:
```suggestion
    async (value?: string) => {
```
We can clean up the unused context and new line here.

edlerd (commented):
Good progress, 85% of the files are perfect. Found two remaining issues with responsiveness and the change counter on the now newly added fields among some tiny suggestions.

src/pages/storage/StorageBucketKeys.tsx
Kxiru: The cards come from the "responsive" attribute. Are you asking that I edit this set of classes?

edlerd: The cards are fine, I think they should stay. My concern was on combining name and role when the card view is active. In the card, it is better to not combine them, but keep them as two different fields/columns. Because we don't have the problem with limited space for the columns there.

In the screen above we see "admin admin", one is the name, the other one the role. The "The admin key" description gets labelled as role wrongly. Better would be to have a heading "role" with value "admin" and another separate field with the name "admin".

src/pages/storage/panels/EditStorageBucketKeyPanel.tsx
```diff
@@ -0,0 +1,183 @@
...
+  };
+
+  const {
+    data: bucketKey,
+    error,
+    isLoading,
+  } = useBucketKey(bucket, key ?? "", project);
+
+  const handleSuccess = (bucket: LxdStorageBucket) => {
+    const bucketURL = `/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(bucket.pool)}/bucket/${encodeURIComponent(bucket.name)}`;
```
edlerd: This should use getStorageBucketURL

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,300 @@
...
+        ...(isSmallScreen
+          ? []
+          : [
+              {
+                content: capitalizeFirstLetter(key.role),
+                role: "cell",
+                "aria-label": "Role",
+                className: "role",
+              },
+            ]),
```
edlerd: We should not combine the name and role columns in the card view. Instead of `isSmallScreen` we should only consider tablet size as a condition for the column collapse of the two. Or get rid of the collapse completely, maybe there is a way to make them fit without, which would be simpler.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,266 @@
...
+        }
+        if (!key.description.toLowerCase().includes(query.toLowerCase())) {
+          return false;
+        }
+        if (!key["access-key"].toLowerCase().includes(query.toLowerCase())) {
+          return false;
+        }
+        if (!key["secret-key"].toLowerCase().includes(query.toLowerCase())) {
+          return false;
+        }
```
edlerd: The matching logic here is off. We should allow matching any of the name, description, etc fields. Not expect the query to be present in all of them.

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,266 @@
...
+          className: "name",
+        },
+        {
+          content: capitalizeFirstLetter(key.role),
+          role: "cell",
+          "aria-label": "Role",
+          className: "role",
+        },
+        {
+          content: key.description,
```
edlerd: At least on the small screen with cards we should add a placeholder like minus for the empty value. Without the description heading looks a bit off. For simplicity, we can always render a minus in case the value is empty.

image

src/pages/storage/StorageBucketKeys.tsx
```diff
@@ -0,0 +1,262 @@
...
+          key.name.toLowerCase().includes(query.toLowerCase()) ||
+          key.description.toLowerCase().includes(query.toLowerCase()) ||
+          key["access-key"].toLowerCase().includes(query.toLowerCase()) ||
+          key["secret-key"].toLowerCase().includes(query.toLowerCase())
+        ) {
+          return true;
+        }
+      }
+      return true;
+    }) ?? [];
```
edlerd: The logic is broken, this matches any input.

edlerd (approved):

PR #1326, opened 2025-06-23
fix(settings) surface error on loading cluster member specific values in server settings page

### LINKED ISSUE #1319: node specific options are empty in settings when node is offline
Image

I was rebooting node2 and i wanted to update a setting and i noticed the setting was empty

in my console it shows a http 503 for node2. As soon as node2 came back online the value showed.

Image

I think this can  cause issues and perhaps there should be a notification in the ui if a cluster member is offline and you;re trying to edit node specific settings.

edlerd:
## Done

- fix(settings) surface error on loading cluster member specific values in server settings page

Fixes #1319

## QA

1. Run the LXD-UI:
    - have a clustered backend
    - switch off one of the cluster nodes (except the one you are using the api from to still be able to load it)
    - go to server settings
    - ensure an error message is displayed informing about the unavailable member.

## Screenshots

image

### DIFF
--- a/src/context/useSettings.tsx
+++ b/src/context/useSettings.tsx
@@ -21,5 +21,6 @@ export const useClusteredSettings = (): UseQueryResult<
     queryKey: [queryKeys.settings, queryKeys.cluster],
     queryFn: async () => fetchSettingsFromClusterMembers(clusterMembers),
     enabled: clusterMembers.length > 0,
+    retry: false,
   });
 };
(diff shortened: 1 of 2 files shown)


Kxiru (approved):
I can see the warning when on the page, as below, however it either takes a long time to load, or is only visible when I refresh the (settings) page. This may be a performance or network issue on my end, and may not actually be an issue when live.

image

What does `retry: false` mean in line 24 of useSettings?

PR #1328, opened 2025-06-23
fix(events) treat events ws stale after on hour and force a reconnect on returning back to the tab if it is stale

### LINKED ISSUE #1307: UI does not handle network disconnects
Every day I take my laptop out of sleep mode or after hours of not using the web ui I have to F5/Refresh the LXD UI before I can use it properly.

I can perform actions like creating a snapshot, stopping/starting a VM etc, but I won't see the result of these actions until I perform a refresh. This causes for some weird scenarios.

For example: 
Image

This just keeps spinning forever. I do see the snapshot show up in the background behind the popup and it is increated on the server :) But I need to refresh my browser to continue to use the ui.

I am running 6.4 with ui 0.17, behind haproxy. Here's a snippet of my haproxy config. Still has the raised timeouts you suggested from when I reported issues with the 30 sec timeouts of the graphic terminal

```
backend lxd-nodes
  mode tcp
  timeout client 24h
  timeout server 24h

  option tcp-check
```

edlerd:
## Done

- fix(events) treat events ws stale after on hour and force a reconnect on returning back to the tab if it is stale

Fixes #1307 WD-23070


lorumic (commented):
Have you considered using something like [`ReconnectingWebSocket`](https://github.com/joewalnes/reconnecting-websocket) instead? Seems like it could be a more robust way to handle unexpected websocket closures - which might not happen exactly after one hour.

src/pages/instances/Events.tsx
```diff
@@ -7,19 +7,46 @@ import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { useOperations } from "context/operationsProvider";
 import { useNotify } from "@canonical/react-components";
+import useEventListener from "util/useEventListener";
 
 const EVENT_HANDLER_DELAY = 250;
 const WS_RETRY_DELAY_MULTIPLIER = 250;
 const MAX_WS_CONNECTION_RETRIES = 5;
+const EVENT_WS_STALE_TIMEOUT_MS = 3600_000;
```
lorumic: If we want to use the underscore separator correctly I think this should be:
```suggestion
const EVENT_WS_STALE_TIMEOUT_MS = 3_600_000;
```

src/pages/instances/Events.tsx
```diff
@@ -7,19 +7,46 @@ import { useQueryClient } from "@tanstack/react-query";
...
+      setReconnectWsId((prevId) => prevId + 1);
+      return;
+    }
+    // If the ws was open for more than EVENT_WS_STALE_TIMEOUT_MS, close it and reconnect.
+    const timeSinceOpen = getCurrentTime() - wsOpenTime;
+    const isStale = timeSinceOpen > EVENT_WS_STALE_TIMEOUT_MS;
+    if (isStale && eventWs) {
+      eventWs.close();
+      setEventWs(null);
+      setReconnectWsId((prevId) => prevId + 1);
```
lorumic: Looks like doing both of these would trigger the effect below, that depends on `[eventWs, isAuthenticated, reconnectWsId]`, twice. Maybe we can just do `setEventWs(null)`? And this makes me wonder: do we actually need the `reconnectWsId` state? Can't we just use the `eventWs` and toggle it between null and ws?
In the check above, where we do:
```
    if (!eventWs) {
      setReconnectWsId((prevId) => prevId + 1);
      return;
    }
```
we could instead directly call the function that is currently inlined with the `useEffect`.

lorumic: Actually, we don't even need to call `setEventWs(null)` here. Because we do `eventWs.close();` and the close handler will already do `setEventWs(null);`.

src/pages/instances/Events.tsx
```diff
@@ -77,10 +104,14 @@ const Events: FC = () => {
...
       ws.onopen = () => {
         setEventWs(ws);
+        setWsOpenTime(getCurrentTime());
       };
       ws.onclose = () => {
         setEventWs(null);
       };
+      ws.onerror = () => {
+        setEventWs(null);
+      };
```
lorumic, line 116: Maybe we can extract `setEventWs(null);` into a reusable `resetEventWs` function, ad use as `onclose`, `onerror` handler and everywhere else we need to reset the event ws to null?

lorumic (commented):
Looks nice, adding a few ideas for code and logic improvements.

src/pages/instances/Events.tsx
edlerd: > we could instead directly call the function that is currently inlined with the useEffect.

That won't work, we need the cleanup of the use effect.

> Actually, we don't even need to call setEventWs(null) here. Because we do eventWs.close(); and the close handler will already do setEventWs(null);.

True and the close call is guaranteed to trigger either onclose or onerror. Will simplify this bit.

src/pages/instances/Events.tsx
edlerd: I think it is simpler as is.

src/pages/instances/Events.tsx
lorumic: > > we could instead directly call the function that is currently inlined with the useEffect.
> 
> That won't work, we need the cleanup of the use effect.

```
  const myEffectHandler = () => {
    if (!eventWs && isAuthenticated) {
      connectEventWs();
    }
    return () => {
      if (eventWs) {
        eventWs.close();
      }
    };
  };

  const reconnectWsOnFocusTab = () => {
    if (document.visibilityState !== "visible") {
      return;
    }
    // Force reconnect if the ws is not open.
    if (!eventWs) {
      myEffectHandler();
      return;
    }

...

useEffect(myEffectHandler, [eventWs, isAuthenticated]);
```

This is what I had in mind. Alternatively, keep the effect handler inline and just change the check here:
```
    // Force reconnect if the ws is not open.
    if (!eventWs) {
      connectEventWs();
      return;
    }
```
And I still think the `reconnectWsId` state is superfluous.

src/pages/instances/Events.tsx
lorumic: DRY disagrees, but as you prefer.

src/pages/instances/Events.tsx
edlerd: The use effect calls the returned function on unmount. We can't do that with calling `myEffectHandler` directly. The returned function is ignored.

src/pages/instances/Events.tsx
edlerd: It is a tradeoff between keeping it simple. As it is about a single trival line seems ok.

src/pages/instances/Events.tsx
lorumic: useEffect is not calling the function directly. **<ins>It's passing the function as a parameter instead of inlining it</ins>**. From my comment above:
```
useEffect(myEffectHandler, [eventWs, isAuthenticated]);
```
Also quoting again from the comment above:
> Alternatively, keep the effect handler inline and just change the check here:
```
    // Force reconnect if the ws is not open.
    if (!eventWs) {
      connectEventWs();
      return;
    }
```
> And I still think the reconnectWsId state is superfluous.

src/pages/instances/Events.tsx
lorumic: Having a `resetEventWs` function that can be reused (and needs to be reused already) doesn't seem this absurd complication to me, but again - YMMV, so as you prefer.

src/pages/instances/Events.tsx
edlerd: I understand you suggested to call `myEffectHandler` directly, right? That is a problem, because the function returns a callback that will be ignored when calling it directly.

edlerd:
```
  const reconnectWsOnFocusTab = () => {
    if (document.visibilityState !== "visible") {
      return;
    }
    // Force reconnect if the ws is not open.
    if (!eventWs) {
      myEffectHandler(); << this line
      return;
    }
```

lorumic: Correct, it is called directly only in the `reconnectWsOnFocusTab` context and you don't need the return value there, so you just ignore it there and all is good. Or what's the harm of having a return value that is not used?

Anyway, I think we should focus on my second idea, which looks cleaner/better, doesn't touch the inlined useEffect, and still doesn't require the additional state.

edlerd: > Correct, it is called directly only in the reconnectWsOnFocusTab context and you don't need the return value there, so you just ignore it there and all is good. Or what's the harm of having a return value that is not used?

We need the callback to be called once the component unmounts to close the websocket. That is why the use effect is in place and needed.

> Anyway, I think we should focus on my second idea, which looks cleaner/better, doesn't touch the inlined useEffect, and still doesn't require the additional state.

That would leave us with multiple connected websockets. We shouldn't consume more than one at a time.

lorumic: > We need the callback to be called once the component unmounts to close the websocket. That is why the use effect is in place and needed.

Answered already: https://github.com/canonical/lxd-ui/pull/1328#discussion_r2166536850

> That would leave us with multiple connected websockets. We shouldn't consume more than one at a time.

```
    if (!eventWs) {
      connectEventWs();
      return;
    }
```
We call `connectEventWs` only when `eventWs` is null. So how can we have multiple connected websockets?

src/pages/instances/Events.tsx
```diff
@@ -7,19 +7,48 @@ import { useQueryClient } from "@tanstack/react-query";
...
+
+  const reconnectWsOnFocusTab = () => {
+    if (document.visibilityState !== "visible") {
+      return;
+    }
+    // Force reconnect if the ws is not open.
+    if (!eventWs) {
+      setReconnectWsId((prevId) => prevId + 1);
+      return;
+    }
```
lorumic, line 39: We probably don't need this check at all, do we? Because the websocket lifecycle comprises the following:

> 1. `onerror`/`onclose` handler called
> 2. `eventWs` set to `null`
> 3. `useEffect` called (because `eventWs` is a dependency)
> 4. `connectWs` called from `useEffect` (because `eventWs` is `null`, so the condition inside `useEffect` is satisfied)

So we don't need to trigger the effect again by doing the following:
```suggestion
```

And the `reconnectWsId` state should be removed.

src/pages/instances/Events.tsx
lorumic: Moving to https://github.com/canonical/lxd-ui/pull/1328#discussion_r2166903507, resolving this to avoid further confusion.

src/pages/instances/Events.tsx
edlerd: The connection could fail after retrying up to 5 times and leave the eventWs as null. This ensures we try to reconnect when the browser tab gets focus later on.

lorumic: Ok, I think it would be nice to hint this in a comment, to make things clearer for future reference :)

lorumic, line 39: This should be enough still - no need for any additional state. Just check whether it's `null` on `visibilitychange` and call the `connectEventWs` function in that case:
```suggestion
    if (!eventWs) {
      connectEventWs();
      return;
    }
```

edlerd: This goes back to the problem that someone needs to close the socket on unmount, only a use effect can do it :D

lorumic: The useEffect is there, untouched and unaffected by this change.

lorumic: The fact that we call connectEventWs from here does not mean that the cleanup function of the useEffect will not be called. You are setting a dedicated state to trigger an effect that calls connectEventWs, when you could have just called connectEventWs directly here.

PR #1330, opened 2025-06-24
chore(tests) update oidc config

edlerd:
## Done

- update oidc config


edlerd:
Adjusted the approach, to rely on the api extension -- if available we set the client secret, if not we skip that step.

PR #1343, opened 2025-06-27
feat: [WD-22926] Volume Bulk Delete

Kxiru:
## Done

- Added Bulk volume deletion capabilities

## QA

1. Run the LXD-UI:
    - Navigate tot Storage > Volumes
    - Verify that a Selectable Main table is in use (You can select one, many or all volumes)
    - Attempt to delete several volumes at once (Please create a few volumes to delete, too!)
    - Verify the success or failure messaging is appropriate and the volumes have been deleted.
    - Verify the table wording accurately reflects the no. of remaining volumes in the table.

## Screenshots
image

image

### DIFF
--- a/src/api/storage-volumes.tsx
+++ b/src/api/storage-volumes.tsx
@@ -10,6 +10,7 @@ import axios from "axios";
 import type { LxdApiResponse } from "types/apiResponse";
 import { addEntitlements } from "util/entitlements/api";
 import { addTarget } from "util/target";
+import { continueOrFinish, pushFailure, pushSuccess } from "util/promises";
 
 export const volumeEntitlements = [
   "can_delete",
@@ -216,6 +217,33 @@ export const deleteStorageVolume = async (
       method: "DELETE",
     },
   ).then(handleResponse);
+};
+
+export const deleteStorageVolumeBulk = async (
+  volumes: LxdStorageVolume[],
+  project: string,
+): Promise<PromiseSettledResult<void>[]> => {
+  const results: PromiseSettledResult<void>[] = [];
+  return new Promise((resolve, reject) => {
+    Promise.allSettled(
+      volumes.map(async (volume) => {
+        return deleteStorageVolume(
+          volume.name,
+          volume.pool,
+          project,
+          volume.location,
+        )
+          .then(() => {
+            pushSuccess(results);
+            continueOrFinish(results, volumes.length, resolve);
+          })
+          .catch((e) => {
+            pushFailure(results, e instanceof Error ? e.message : "");
+            continueOrFinish(results, volumes.length, resolve);
+          });
+      }),
+    ).catch(reject);
+  });
 };
 
 export const migrateStorageVolume = async (
--- a/src/pages/storage/StorageVolumes.tsx
+++ b/src/pages/storage/StorageVolumes.tsx
@@ -1,10 +1,9 @@
 import type { FC } from "react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useParams, useSearchParams } from "react-router-dom";
 import {
   EmptyState,
   Icon,
-  MainTable,
   Row,
   TablePagination,
   useNotify,
@@ -53,6 +52,10 @@ import NotificationRow from "components/NotificationRow";
 import { useLoadVolumes } from "context/useVolumes";
 import { useIsClustered } from "context/useIsClustered";
 import ResourceLink from "components/ResourceLink";
+import SelectableMainTable from "components/SelectableMainTable";
+import SelectedTableNotification from "components/SelectedTableNotification";
+import type { LxdStorageVolume } from "types/storage";
+import StorageVolumeBulkDelete from "./actions/StorageVolumeBulkDelete";
 
 const StorageVolumes: FC = () => {
   const docBaseLink = useDocs();
@@ -86,9 +89,23 @@ const StorageVolumes: FC = () => {
 
   const { data: volumes = [], error, isLoading } = useLoadVolumes(project);
 
+  const getVolumeKey = (volume: LxdStorageVolume) => {
+    return `${volume.name}-${volume.pool}-${volume.location || ""}`;
+  };
+  const [selectedNames, setSelectedNames] = useState<string[]>([]);
+  const [processingVolumes, setProcessingVolumes] = useState<string[]>([]);
+
   if (error) {
     notify.failure("Loading storage volumes failed", error);
   }
+
+  useEffect(() => {
+    const validKeys = new Set(volumes.map(getVolumeKey));
+    const validSelections = selectedNames.filter((name) => validKeys.has(name));
+    if (validSelections.length !== selectedNames.length) {
+      setSelectedNames(validSelections);
+    }
+  }, [volumes]);
 
   const headers = [
     {
@@ -210,14 +227,15 @@ const StorageVolumes: FC = () => {
 
   const snapshotsPerVolume = getSnapshotsPerVolume(volumes);
   const rows = filteredVolumes.map((volume) => {
+    const key = getVolumeKey(volume);
     const volumeType = renderVolumeType(volume);
     const contentType = renderContentType(volume);
-
-    const key = `${volume.name}-${volume.pool}-${volume.location}`;
     const snapshotCount = snapshotsPerVolume[key]?.length ?? 0;
+    const canSelect = hasVolumeDetailPage(volume);
 
     return {
       key,
+      name: canSelect ? key : undefined,
       className: "u-row",
       columns: [
         {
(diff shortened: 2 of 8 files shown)


src/pages/storage/StorageVolumes.tsx
```diff
@@ -402,20 +419,45 @@ const StorageVolumes: FC = () => {
...
-            className="storage-volume-table"
+            itemName="volume"
+            parentName="project"
+            selectedNames={selectedNames}
+            setSelectedNames={setSelectedNames}
+            disabledNames={processingVolumes}
+            filteredNames={filteredVolumes.map(getVolumeKey)}
             onUpdateSort={updateSort}
+            defaultSortDirection="descending"
           />
```
edlerd, line 458: The layout on large screens is bamboozled, the first column with the selector has too much space:

image

src/pages/storage/StorageVolumes.tsx
```diff
@@ -210,14 +227,14 @@ const StorageVolumes: FC = () => {
...
+    const key = getVolumeKey(volume);
     const volumeType = renderVolumeType(volume);
     const contentType = renderContentType(volume);
-
-    const key = `${volume.name}-${volume.pool}-${volume.location}`;
     const snapshotCount = snapshotsPerVolume[key]?.length ?? 0;
 
     return {
       key,
+      name: key,
```
edlerd: If the volume relates to an instance root disk, it can't be deleted from here.

I think by not setting the name on such instances, the selectable table components will react correctly and 1. disable the select checkbox and 2. exclude the row from the select all actions.

To conclude: I think we should only set the key on volumes that can be deleted.

src/pages/storage/StorageVolumes.tsx
```diff
@@ -402,20 +419,45 @@ const StorageVolumes: FC = () => {
...
+          description={
+            selectedNames.length > 0 && (
+              <SelectedTableNotification
+                totalCount={volumes.length ?? 0}
+                itemName="volume"
+                parentName="project"
+                selectedNames={selectedNames}
+                setSelectedNames={setSelectedNames}
+                filteredNames={filteredVolumes.map(getVolumeKey)}
+              />
```
edlerd: This currently shows the "select all filtered volumes" even when no filters are applied (see below). I think the reason is that, the `totalCount={volumes.length ?? 0}` is including snapshots and in the `filteredVolumes` we exclude snapshots. So to fix this, we should exclude snapshots also for the total count we pass in here.


image

src/pages/storage/actions/StorageVolumeBulkDelete.tsx
```diff
@@ -0,0 +1,115 @@
...
+              <br />
+              <b>{rejectedCount}</b> {pluralize("volume", rejectedCount)} could
+              not be deleted.
+            </>,
+          );
+        }
+
+        queryClient.invalidateQueries({
+          queryKey: [queryKeys.volumes, projectName],
+        });
```
edlerd, line 71: We need to invalidate a couple of more keys here. Because volume delete might also affect ISO volumes, among others. See `DeleteStorageVolumeBtn` handler for a full list.

src/pages/storage/StorageVolumes.tsx
Kxiru: Hmm... I cannot reproduce this on my large monitor.
image

The closest I can get is when I deliberately resize to a very awkward size. Can look into this:
image

edlerd: I experience it [on the demo server](https://lxd-ui-1343.demos.haus/ui/project/default/storage/volumes) with 1875px width.

Kxiru: image
I can't recreate the problem very well. In this image, I have zoomed out substantially and there is no distortion. In any case, how would you like this to be approached?

edlerd: But also with less, such as this one (also demo server from this PR)

image

edlerd: I think the problem is with the other columns have percentage values for the width. In other selectable tables such as for warnings, we use `rem` as a unit.

Kxiru: Please let me know your thoughts now.

edlerd: The width of columns looks much better aligned now, thanks for fixing it :)

src/pages/storage/StorageVolumes.tsx
edlerd: This is now using the permissions, but there are still the instance root volumes enabled. Deleting an instance root volume will result in an error. The user has to delete the instance instead. So my suggestion is to disable the selector by not setting a name in here for the volumes which are instance root disks. We might need to add a comment or introduce a function for this to keep this knowledge in the code. Because I assume this is not obvious from readign the code.

src/pages/storage/StorageVolumes.tsx
```diff
@@ -430,11 +474,22 @@ const StorageVolumes: FC = () => {
                 Volumes
               </HelpLink>
             </PageHeader.Title>
-            {hasVolumes && (
+            {!selectedNames.length && hasVolumes && (
```
edlerd, line 483: There is a jump in the page content when selecting the first volume. I think that is from hiding this search input. We should ensure the table content does not jump up in that case.


https://github.com/user-attachments/assets/7c3701ba-faaa-4b04-b4ec-64ee9a724e6a

src/pages/storage/actions/StorageVolumeBulkDelete.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+const StorageVolumeBulkDelete: FC<Props> = ({ volumes, onStart, onFinish }) => {
+  const toastNotify = useToastNotification();
+  const queryClient = useQueryClient();
+  const [isLoading, setLoading] = useState(false);
+  const { canDeleteVolume } = useStorageVolumeEntitlements();
+  const { project } = useCurrentProject();
+  const projectName = project?.name || "";
+
+  const deleteableVolumes = volumes.filter((volume) => canDeleteVolume(volume));
```
edlerd:
```suggestion
  const deletableVolumes = volumes.filter((volume) => canDeleteVolume(volume));
```
small typo

src/pages/storage/StorageVolumes.tsx
```diff
@@ -402,20 +422,45 @@ const StorageVolumes: FC = () => {
           itemName="volume"
           className="u-no-margin--top"
           aria-label="Table pagination control"
+          description={
+            selectedNames.length > 0 && (
+              <SelectedTableNotification
+                totalCount={filteredVolumes.length ?? 0}
```
edlerd:
```suggestion
                totalCount={volumes.filter((volume) => !isSnapshot(volume)).length}
```
This should be the unfiltered list, so we get "select all n filtered volumes" copy in case filters are applied:

image

src/pages/storage/StorageVolumes.tsx
```diff
@@ -210,14 +229,15 @@ const StorageVolumes: FC = () => {
 
   const snapshotsPerVolume = getSnapshotsPerVolume(volumes);
   const rows = filteredVolumes.map((volume) => {
+    const key = getVolumeKey(volume);
     const volumeType = renderVolumeType(volume);
     const contentType = renderContentType(volume);
-
-    const key = `${volume.name}-${volume.pool}-${volume.location}`;
     const snapshotCount = snapshotsPerVolume[key]?.length ?? 0;
+    const canSelect = canDeleteVolume(volume) && hasVolumeDetailPage(volume);
```
edlerd:
```suggestion
    const canSelect = hasVolumeDetailPage(volume);
```
Let's allow selecting also volumes without permissions. We can rely on the `StorageVolumeBulkDelete` component to surface the missing permission.

src/pages/storage/StorageVolumes.tsx
edlerd: This should be the right value for it:

```
              <SelectedTableNotification
                totalCount={
                  volumes.filter(
                    (volume) =>
                      !isSnapshot(volume) && hasVolumeDetailPage(volume),
                  ).length
                }
                itemName="volume"
                parentName="project"
                selectedNames={selectedNames}
                setSelectedNames={setSelectedNames}
                filteredNames={filteredVolumes
                  .filter(hasVolumeDetailPage)
                  .map(getVolumeKey)}
              />
```

src/api/storage-volumes.tsx
```diff
@@ -218,6 +219,28 @@ export const deleteStorageVolume = async (
...
 
+export const deleteStorageVolumeBulk = async (
+  volumes: LxdStorageVolume[],
+  project: string,
+): Promise<PromiseSettledResult<void>[]> => {
+  const results: PromiseSettledResult<void>[] = [];
+  return new Promise((resolve, reject) => {
+    Promise.allSettled(
+      volumes.map(async (volume) => {
+        return deleteStorageVolume(volume.name, volume.pool, project)
```
edlerd:
```suggestion
        return deleteStorageVolume(volume.name, volume.pool, project, volume.location)
```
For a cluster we need to specify the target when deleting a volume as well to uniquely identify it.

edlerd (approved):
LGTM, thanks for the iteration.

PR #1348, opened 2025-06-30
feat(theme) add dark mode

### LINKED ISSUE #1341: Feature Request: Dark Mode
It would be a good addition to have a dark mode. With the popularity of dark mode I would guess this is on the road map, but I don't see an issue here.

 [Branding colors](https://design.ubuntu.com/brand/colour-palette)

edlerd:
## Done

feat(theme) add dark mode

Fixes #1341

## QA

1. Run the LXD-UI:
    - go to settings and switch the theme to dark mode
    - check if pages and interactions work just as before in dark and light mode.

## Screenshots

image

image

image

### DIFF
--- a/src/components/ResourceLink.tsx
+++ b/src/components/ResourceLink.tsx
@@ -20,10 +20,8 @@ const ResourceLink: FC<Props> = ({ type, value, to, disabled }) => {
       to={to}
       title={value}
     >
-      <span className="p-chip__value">
-        <ResourceIcon type={type} />
-        {value}
-      </span>
+      <ResourceIcon type={type} />
+      <span className="p-chip__value">{value}</span>
     </Link>
   );
 };
--- a/src/pages/settings/SettingForm.tsx
+++ b/src/pages/settings/SettingForm.tsx
@@ -19,6 +19,7 @@ import { useServerEntitlements } from "util/entitlements/server";
 import ClusteredSettingFormInput from "./ClusteredSettingFormInput";
 import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 import { useIsClustered } from "context/useIsClustered";
+import SettingThemeSwitcher from "pages/settings/SettingThemeSwitcher";
 
 export const getConfigId = (key: string) => {
   return key.replace(".", "___");
@@ -52,6 +53,7 @@ const SettingForm: FC<Props> = ({
   const isLokiAuthPassword = configField.key === "loki.auth.password";
   const isSecret = isTrustPassword || isLokiAuthPassword;
   const isClusteredInput = isClustered && configField.scope === "local";
+  const isThemeSelector = configField.key === "user.ui_theme";
 
   const settingLabel = (
     <ResourceLabel bold type="setting" value={configField.key} />
@@ -124,6 +126,10 @@ const SettingForm: FC<Props> = ({
         }}
       />
     );
+  }
+
+  if (isThemeSelector) {
+    return <SettingThemeSwitcher />;
   }
 
   return (
--- a/src/pages/settings/SettingThemeSwitcher.tsx
+++ b/src/pages/settings/SettingThemeSwitcher.tsx
@@ -1,0 +1,67 @@
+import type { FC } from "react";
+import { useState } from "react";
+
+const LOCAL_STORAGE_KEY = "theme";
+const THEME_SYSTEM = "system";
+const THEME_DARK = "dark";
+const THEME_LIGHT = "light";
+
+export const loadTheme = (): string => {
+  const saved = localStorage.getItem(LOCAL_STORAGE_KEY);
+  return saved || THEME_SYSTEM;
+};
+
+const saveTheme = (theme: string): void => {
+  localStorage.setItem(LOCAL_STORAGE_KEY, theme);
+};
+
+export const isDarkTheme = (theme: string) => {
+  if (theme === THEME_SYSTEM) {
+    return (
+      window.matchMedia &&
+      window.matchMedia("(prefers-color-scheme: dark)").matches
+    );
+  }
+  return theme === THEME_DARK;
+};
+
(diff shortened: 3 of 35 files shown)


piperdeck:
Is this what bar graphs are supposed to look like? Looks a little weird to me, but if this is what the upstream darkmode calls for then I can live with it

CleanShot 2025-06-30 at 15 28 16@2x

---

In light mode, we use dark mode to indicate the current navigation of a network diagram. So conversely in dark mode we should use light mode to indicate the navigation (current node in the diagram)

CleanShot 2025-06-30 at 15 29 21@2x

---

Looks like code snippets are getting a bit messed up

|  Light    |   Dark  |
| --- | --- |
| CleanShot 2025-06-30 at 15 37 00@2x | CleanShot 2025-06-30 at 15 36 55@2x |
| CleanShot 2025-06-30 at 15 37 40@2x | CleanShot 2025-06-30 at 15 37 44@2x |

piperdeck:
Maybe the switch should go over to the left, next to the version info?

edlerd:
> Is this what bar graphs are supposed to look like? Looks a little weird to me, but if this is what the upstream darkmode calls for then I can live with it

updated the bar graphs in light mode and also in dark mode, so they rely on themed colors.

> In light mode, we use dark mode to indicate the current navigation of a network diagram. So conversely in dark mode we should use light mode to indicate the navigation (current node in the diagram)

good catch, updated the network topology styles to be theme based.

> Looks like code snippets are getting a bit messed up

this was another override on our side, adjusted it now.

> Maybe the switch should go over to the left, next to the version info?

I experimented with it, using a smaller switch, left aligning it or using a segmented control with "light, dark, system" options. All looked a bit too intrusive in the footer. So I moved the switch to the settings page. I don't think it should be that prominent in our footer.

edlerd:
> So if I was a user who had heard that LXD-UI had dark mode now, the first thing I would do is search for "dark mode" in the settings. But "dark" doesn't return any matches, because the name of the setting doesn't contain that string. Is there some way for these configurations to have a hidden field that we can search for alternate terms like this? Or alternatively, search over values somehow?

Good point, I gave the setting a description and made the search also match on description text. That way the query "dark" matches the theme setting.

piperdeck:
Can we use this copy instead:

> Set UI to dark theme, light theme, or match the user’s system theme

Other than that, this is good to go

src/pages/settings/SettingForm.tsx
```diff
@@ -52,6 +53,7 @@ const SettingForm: FC<Props> = ({
   const isLokiAuthPassword = configField.key === "loki.auth.password";
   const isSecret = isTrustPassword || isLokiAuthPassword;
   const isClusteredInput = isClustered && configField.scope === "local";
+  const isTheme = configField.key === "user.ui_theme";
```
Kxiru: More descriptive naming? Is this to say whether the current version HAS a theme? Perhaps hasTheme?

src/pages/settings/SettingThemeSwitcher.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const loadTheme = (): string => {
+  const saved = localStorage.getItem(LOCAL_STORAGE_KEY);
+  return saved || THEME_SYSTEM;
+};
+
+const saveTheme = (theme: string): void => {
+  localStorage.setItem(LOCAL_STORAGE_KEY, theme);
+};
+
+export const isDarkTheme = (theme: string) => {
```
Kxiru, line 18: Nice function logic. I had to really unpack this one line by line.

src/components/ResourceLink.tsx
```diff
@@ -20,10 +20,8 @@ const ResourceLink: FC<Props> = ({ type, value, to, disabled }) => {
       to={to}
       title={value}
     >
-      <span className="p-chip__value">
-        <ResourceIcon type={type} />
-        {value}
-      </span>
+      <ResourceIcon type={type} />
```
Kxiru, line 23: Thanks for this fix!

Kxiru (commented):
I made a mock of what a nav-bar toggle could look like as I thought it would be cute and quaint. What do you think? We could have this be a more accessible toggle.

image
image
image

Just using the [Switch Component](https://canonical.github.io/react-components/?path=/docs/components-switch--docs). It's definitely quite out of the way to keep going to Settings to change this.


Tested edges of the UI such as:
- Panels
- Confirmation/upload modals
- Inline actions
- Search and filter

1. I think the sub-menus should be in white for higher contrast, accessibility and visibility:
image

src/pages/settings/SettingForm.tsx
edlerd: This is describing what the currently handled field is in terms of input. I hope `isThemeSelector` is a bit more clear, changing to that name.

edlerd:
> I made a mock of what a nav-bar toggle could look like as I thought it would be cute and quaint. What do you think? We could have this be a more accessible toggle.
> Just using the [Switch Component](https://canonical.github.io/react-components/?path=/docs/components-switch--docs). It's definitely quite out of the way to keep going to Settings to change this.

I tested the switch in an earlier iteration in the status bar, next to the version number. Concerns were raised on supporting the "system" default among the dark/light choices. So the binary switch input won't work for the theme. Hence, the move of the theme picker into the settings. I think that is ok, this is not the most important setting that needs to be accessed frequently. I agree, putting it more prominently would advertise the feature nicely -- though I am not sure if we should. I think putting it into settings is an okay tradeoff.

> 1. I think the sub-menus should be in white for higher contrast, accessibility and visibility:

Good point, though this is an edge case of instance creation, where the navigation is disabled. After picking a base image, the text in the nav is not muted anymore and gets a higher contrast. So I think this works as intended.

Kxiru (approved):
A nice new feature to the UI. Thank you for this addition.

PR #1365, opened 2025-07-07
feat(placement-groups) add placement group management

edlerd:
## Done

- feat(placement-groups) add placement group management

## QA

1. Run the LXD-UI:
    - go to profiles > placement groups
    - create a placement group
    - edit a placement group
    - delete a placement group

## Screenshots

image

image

image

image

image


src/pages/instances/CreateInstance.tsx
edlerd, line 1: @piperdeck there is a design challenge here. How to properly select the placement of the new instance. We have too many different options. The spec lists them all. Can you find an easy and intuitive way that communicates the implications of each option well, but also doesn't take too much space?

PR has an engineering drafted way as show below. Happy to adjust it to something better.

image

image

image


image

image

src/pages/placement-groups/PlacementGroupList.tsx
```diff
@@ -0,0 +1,219 @@
...
+  return (
+    <>
+      <CustomLayout
+        mainClassName="placement-groups-list"
+        header={
+          <PageHeader>
+            <PageHeader.Left>
+              <PageHeader.Title>
+                <HelpLink
+                  href={`${docBaseLink}/placement-groups/`}
```
edlerd: Todo: fix this link

src/components/ResourceIcon.tsx
```diff
@@ -54,6 +55,7 @@ const resourceIcons: Record<ResourceIconType, string> = {
   bucket: "status-queued-small",
   "bucket-key": "private-key",
   metric: "statistics",
+  "placement-group": "repository",
```
edlerd, line 58: Which icon to use @piperdeck ?

src/pages/profiles/ProfileList.tsx
```diff
@@ -231,6 +231,17 @@ const ProfileList: FC = () => {
...
+                  appearance=""
+                  className="u-no-margin--bottom u-float-right"
+                  onClick={async () =>
+                    navigate(
+                      `/ui/project/${encodeURIComponent(projectName)}/placement-groups`,
+                    )
+                  }
+                >
+                  <span>Placement groups</span>
+                </Button>
```
edlerd: Is the profile list header the right entry point for this @piperdeck ?

src/pages/placement-groups/PlacementGroupList.tsx
```diff
@@ -0,0 +1,219 @@
...
+  return (
+    <>
+      <CustomLayout
+        mainClassName="placement-groups-list"
+        header={
+          <PageHeader>
+            <PageHeader.Left>
+              <PageHeader.Title>
+                <HelpLink
+                  href={`${docBaseLink}/howto/cluster_placement_groups/`}
```
edlerd: Double check this is the right link.

edlerd:
Draft state as this is blocked until https://github.com/canonical/lxd/pull/16781 is merged

src/components/ResourceIcon.tsx
kimanhou: Are we sure this is the right icon ? Since the page is under Clustering in the navigation, would it make sense to use `cluster-host` icon?

edlerd: I think `repository` is the better option as it resembles a type of reusable template, which the placement group is. Though I see you could also say the same about `cluster-host` as it resembles placement in a cluster. Then again, I think placement is already inherent in the name.

edlerd:
> When creating an instance, the UI checks if there already is an instance with the same name in `instanceNameValidation`. Should we have the same for placement group ?
> 
> Should we have basic validation to avoid this error from back end: `Name can only contain alphanumeric, forward slash, hyphen, colon, underscore and full stop characters`? It might also be worth trimming the name ,to allow `"unicorn "` for example.

Good idea about validation, added those. I wouldn't do trimming at the component level. Might be something to look into more holistically for all name inputs. Though not even sure if this is necessary.

src/pages/placement-groups/PlacementGroupUsedBy.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+interface Props {
+  placementGroup?: LxdPlacementGroup;
+}
+
+const PlacementGroupUsedBy: FC<Props> = ({ placementGroup }) => {
+  return (
+    <table>
+      <tbody>
+        <UsedByRow entityType="instance" usedBy={placementGroup?.used_by} />
```
kimanhou: Yes ! 🥳

src/pages/placement-groups/PlacementGroupUsedBy.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <UsedByRow entityType="profile" usedBy={placementGroup?.used_by} />
```
kimanhou: How do you get a placement group used by a profile ? I can't find a way to add a placement group to a profile, am I missing something ? 🤔

src/pages/placement-groups/PlacementGroupList.tsx
```diff
@@ -0,0 +1,223 @@
...
+              >
+                <TablePagination
+                  id="pagination"
+                  data={sortedRows}
+                  itemName="placement group"
+                  className="u-no-margin--top"
+                  aria-label="Table pagination control"
+                >
+                  <MainTable
+                    id="profile-table"
```
kimanhou:
```suggestion
                    id="placement-table"
```

Do we even need an id ?

src/pages/instances/CreateInstance.tsx
```diff
@@ -433,6 +433,12 @@ const CreateInstance: FC = () => {
         ...sshKeyPayload(values),
       },
     };
+
```
kimanhou: Would this work ?
```suggestion
...(values.placementGroup && {
          "placement.group": values.placementGroup,
        }),
```

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -0,0 +1,300 @@
...
+        label="Cluster member"
+        value={formik.values.target}
+        disabled
+        help="Member is determined by the selected ISO volume and can't be changed."
+        type="text"
+      />
+    );
+  }
+
+  const missingImageTitle = formik.values.image
```
kimanhou, line 176: Just an observation: on instance creation, if you select an image, then change target, then remove the image, the target is disabled but not reset to default.

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -0,0 +1,300 @@
...
+import { useClusterMembers } from "context/useClusterMembers";
+import type { SelectRef } from "@canonical/react-components/dist/components/CustomSelect/CustomSelect";
+import PlacementGroupSelect from "pages/instances/forms/PlacementGroupSelect";
+
+interface Props {
+  formik: FormikProps<CreateInstanceFormValues>;
+}
+
+const selectedType = (values: CreateInstanceFormValues) => {
+  if (values.target?.startsWith("@")) {
```
kimanhou: Should these be in constants ?

```
export const TARGET_TYPES = {
  AUTO: "auto",
  CLUSTER_GROUP: "clusterGroup", 
  CLUSTER_MEMBER: "clusterMember",
  PLACEMENT_GROUP: "placementGroup",
}

export const CLUSTER_GROUP_PREFIX = "@";
```

src/pages/placement-groups/PlacementGroupUsedBy.tsx
edlerd: there is a placement group selector on profile edit and create

image

src/pages/placement-groups/PlacementGroupList.tsx
edlerd: Good catch! Yes, we reference the id from the `ScrollableTable` right above. But the ids did not match, now they do.

src/pages/instances/CreateInstance.tsx
edlerd: Yes, good idea!

src/pages/instances/forms/InstanceTargetSelect.tsx
edlerd: The idea is we disable all other inputs until a base image is selected so users don't get stuck configuring the instance and forgetting the image.

On clearing the image I think it is fair to assume other configuration changes that have been set are expected to be kept. Clearing them too on instance change seems unexpected.

src/pages/instances/forms/InstanceTargetSelect.tsx
edlerd: good idea :+1:

src/pages/placement-groups/PlacementGroupUsedBy.tsx
kimanhou: Found it, thank you !

When I add a placement group to a profile, then create an instance with that profile, the inherited placement group is not visible on the instance. It might the case already for other properties but should we display them somewhere ?


In this screenshot, the profile `poney` has a placement group.
image

edlerd: Oh, good point, the inheritance is not showing on the configuration screen. It should show on the overview tab though. I think I'll show the inherited value instead of "None" in the placement group selector with some copy of the originating profile.

src/pages/placement-groups/PlacementGroupForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+        autoFocus={isEdit}
+        label="Description"
+        placeholder="Enter description"
+        error={formik.touched.description ? formik.errors.description : null}
+      />
+      <Select
+        {...formik.getFieldProps("policy")}
+        label="Policy"
+        options={[
+          { value: "compact", label: "Compact" },
```
kimanhou: It might be worth putting these in a constant and reuse them in `canAnchorPlacementGroup` of `InstanceTargetSelect`, `CreatePlacementGroupPanel` and maybe the type ?

src/pages/placement-groups/PlacementGroupForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <Form onSubmit={formik.handleSubmit}>
+      {/* hidden submit to enable enter key in inputs */}
+      <Input type="submit" hidden value="Hidden input" />
+      <Input
+        {...formik.getFieldProps("name")}
+        type="text"
+        label="Name"
+        required
+        autoFocus={!isEdit}
+        error={formik.touched.name ? formik.errors.name : null}
```
kimanhou: Just a suggestion

```
const getFieldError = (fieldName: keyof PlacementGroupFormValues) =>
    formik.touched[fieldName] ? formik.errors[fieldName] : undefined;
```

```suggestion
        error={getFieldError("name")}
```

src/pages/placement-groups/PlacementGroupList.tsx
```diff
@@ -0,0 +1,223 @@
...
+    data: placementGroups = [],
+    error,
+    isLoading,
+  } = usePlacementGroups(projectName);
+
+  if (error) {
+    notify.failure("Loading placement groups failed", error);
+  }
+
+  const headers = [
```
kimanhou, line 48:
```suggestion
  const TABLE_HEADERS = [
```
Since it's a constant, it should be all caps.

src/pages/placement-groups/PlacementGroupList.tsx
```diff
@@ -0,0 +1,223 @@
...
+                <HelpLink
+                  docPath="/howto/cluster_placement_groups/"
+                  title="Learn how to use placement groups"
+                >
+                  Placement groups
+                </HelpLink>
+              </PageHeader.Title>
+            </PageHeader.Left>
+            <PageHeader.BaseActions>
+              {placementGroups.length > 0 && (
```
kimanhou:
```
const isEmpty = placementGroups.length === 0;
```

```suggestion
              {!isEmpty && (
```

You can then re-use `isEmpty` on lines 169 and 187

src/pages/placement-groups/PlacementGroupUsedBy.tsx
edlerd: Added the inherited placement group show up correctly on instance edit.

src/pages/placement-groups/PlacementGroupUsedBy.tsx
kimanhou: Suggestion: early return in case placementGroup doesn't exist or is empty. Then we could get rid of the ?

src/pages/placement-groups/PlacementGroupUsedBy.tsx
kimanhou: Cool, I see it now !

image

Are we happy with the UI tho ? `From` with a capital F looks a bit weird. I know other "from profile blabla" are displayed like this but should we put that in parenthesis ?

src/pages/placement-groups/actions/CreatePlacementGroupBtn.tsx
```diff
@@ -0,0 +1,31 @@
...
+  smallScreenBreakpoint,
+  useIsScreenBelow,
+} from "context/useIsScreenBelow";
+import usePanelParams from "util/usePanelParams";
+
+interface Props {
+  className?: string;
+}
+
+const DeletePlacementGroupBtn: FC<Props> = ({ className }) => {
```
kimanhou: Is it Delete or Create ? The name doesn't match the file name

src/pages/placement-groups/actions/DeletePlacementGroupBtn.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return (
+    <ConfirmationButton
+      appearance="base"
+      onHoverText={
+        isUsed
+          ? "Remove all usages to delete this placement group"
+          : "Delete placement group"
+      }
+      className={classnames("u-no-margin--bottom", {
```
kimanhou:
```suggestion
      className="u-no-margin--bottom has-icon"
```

Do we need `classnames` here ?

src/pages/placement-groups/actions/CreatePlacementGroupBtn.tsx
edlerd: whoopsie

src/pages/placement-groups/actions/DeletePlacementGroupBtn.tsx
edlerd: good spot, yes this can be much simpler.

src/pages/placement-groups/PlacementGroupList.tsx
edlerd: I think this is fine, we tend to use `camlCase` for inline `const`, only exported or global `const` are all caps.

src/pages/placement-groups/PlacementGroupUsedBy.tsx
edlerd: Good point. Using parenthesis with lowercase "from" looks better indeed.

src/pages/placement-groups/actions/CreatePlacementGroupBtn.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+const CreatePlacementGroupBtn: FC<Props> = ({ className }) => {
+  const isSmallScreen = useIsScreenBelow(smallScreenBreakpoint);
+  const panelParams = usePanelParams();
+  return (
+    <Button
+      appearance="positive"
+      className={className}
+      onClick={() => {
```
kimanhou: Would this work ?

```suggestion
      onClick={panelParams.openCreatePlacementGroup}
```

edlerd:
> Quick question: Did you remember to include support for search? Like we can do with profiles and cluster members

@vosdev I was expecting not a massive use of the feature, so having maybe a handful of placement groups. Do you think there is a reason to have many such placement groups, so we need a search for them?

src/pages/placement-groups/actions/CreatePlacementGroupBtn.tsx
edlerd: Yes, good idea.

src/pages/placement-groups/panels/CreatePlacementGroupPanel.tsx
```diff
@@ -0,0 +1,146 @@
...
+          <Button
+            appearance="base"
+            onClick={closePanel}
+            className="u-no-margin--bottom"
+          >
+            Cancel
+          </Button>
+          <ActionButton
+            appearance="positive"
+            onClick={() => void formik.submitForm()}
```
kimanhou, line 131:
```suggestion
            onClick={formik.submitForm}
```
A bit simpler ?

src/sass/_instance_target_selector.scss
```diff
@@ -0,0 +1,60 @@
+.instance-target-selector {
+  display: flex;
+  gap: 1rem;
```
kimanhou:
```suggestion
  gap: $sph--large;
```

src/sass/_instance_target_selector.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  @media screen and (width <= 860px) {
+    flex-direction: column;
```
kimanhou, line 6:
```suggestion
    flex-direction: column;
    gap: $sph--small;
```

Just a suggestion

edlerd: Good point! I think this should even be gap of 0 to match the spacing of other fields

src/pages/placement-groups/panels/CreatePlacementGroupPanel.tsx
edlerd: This throws a type error because submit form is returning a promise.

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -0,0 +1,311 @@
...
+    return TARGET.PLACEMENT_GROUP;
+  }
+  if (values.target) {
+    return TARGET.CLUSTER_MEMBER;
+  }
+  return TARGET.AUTO;
+};
+
+const selectedClusterGroup = (target?: string) => {
+  if (!target?.startsWith("@")) {
```
kimanhou: Why not reuse `CLUSTER_GROUP_PREFIX` on lines 51, 54, and 58 ?

src/pages/instances/forms/EditInstanceDetails.tsx
```diff
@@ -76,6 +86,21 @@ const EditInstanceDetails: FC<Props> = ({ formik, project }) => {
...
+              setValue={(value) => {
+                ensureEditMode(formik);
+                formik.setFieldValue("placement_group", value || undefined);
+              }}
+              project={project}
+              profileNames={formik.values.profiles}
+              hasNoneOption
+              help={
+                placementGroups.length === 0
+                  ? "No placement groups found for this project. "
```
kimanhou: This string appears 3 times in the code. Is it worth creating a constant ?

src/pages/placement-groups/PlacementGroupForm.tsx
```diff
@@ -0,0 +1,73 @@
+import { Form, Input, Select } from "@canonical/react-components";
+import type { FC } from "react";
+import type { FormikProps } from "formik/dist/types";
+
+export const PLACEMENT_GROUP_POLICY_COMPACT = "compact";
```
kimanhou, line 5: I know this is very late for this but what do you think of affinity/anti-affinity instead of compact/spread ?
Also what about placement strategy or affinity rule instead of policy ?

For rigor, I was thinking of enforcement level but it's very long. What do you think ?

edlerd: I think we should match with the CLI here.

This was discussed a while ago with @markylaing and I think there were good reasons to settle on compact/spread.

src/pages/instances/forms/EditInstanceDetails.tsx
edlerd: Ideally, this should generalize and move into the `PlacementGroupSelect` component. So we don't need to duplicate at all. Maybe this can be combined with the `hasNoneOption` to a more generic concept.

PR #1370, opened 2025-07-09
fix(permission) adjust design for removing a permission from a group on mobile

edlerd:
## Done

- fix(permission) adjust design for removing a permission from a group on mobile

## QA

1. Run the LXD-UI:
    - go to permissions > groups > create group > edit permissions and add some permissions. save the group
    - edit the group you created > edit permissions > remove a permission
    - check on smaller screens how the removed permission is presented

## Screenshots

image

image


anusha-c18:
This removes the strikeout even when it isn't in the cards view
Is it better to only change it in the cards view?

edlerd:
> This removes the strikeout across the entire row, even when it isn't in the cards view. Is it better to only change it in the cards view?

The text of a removed permission in the large screen is still appearing with a line through it. Not sure if I understand your concern correctly. Can you please detail a bit more?

anusha-c18:
We can keep the strikeout across the entire row when it isn't in cards view and only apply `text-decoration: line-through` in the cards view:

https://github.com/user-attachments/assets/72796091-fec7-4691-b3fc-d3a6f6c22d7a

src/sass/_permission_groups.scss
```diff
@@ -140,16 +140,25 @@
...
-      td:not(.actions)::before {
-        border-bottom: 1px solid black;
-        content: "";
-        left: 0;
-        position: absolute;
-        top: 1.2rem;
-        width: 100%;
+      @media screen and (width > $breakpoint-large) {
+        td:not(.actions)::before {
+          border-bottom: 1px solid black;
```
anusha-c18: I think this would better suit the dark theme:

```suggestion
          border-bottom: 1px solid $colors--theme--border-high-contrast;
```

with `black`:

Screenshot 2025-07-09 at 3 10 34 PM

with `$colors--theme--border-high-contrast`:

Screenshot 2025-07-09 at 3 09 33 PM

src/sass/_permission_groups.scss
```diff
@@ -140,16 +140,25 @@
 
     tr.strikeout {
       td {
+        opacity: 0.6;
         position: relative;
       }
```
anusha-c18: The opacity shouldn't apply to the actions:

```suggestion
      td:not(.actions) {
        opacity: 0.6;
        position: relative;
      }
```

src/sass/_permission_groups.scss
anusha-c18, line 1: I think this can be cleaned up a bit:
```
  .strikeout {
      td:not(.actions) {
        text-decoration: line-through;
        opacity: 0.6;
      }

      @media screen and (width > $breakpoint-large) {
        td {
          position: relative;
          text-decoration: none;
        }

        td:not(.actions)::before {
          border-bottom: 1px solid $colors--theme--border-high-contrast;
          content: "";
          left: 0;
          position: absolute;
          top: 1rem;
          width: 100%;
        }
      }
    }
```

anusha-c18 (changes requested):
A couple of suggestions

src/sass/_permission_groups.scss
edlerd: agreed, good suggestion.

src/sass/_permission_groups.scss
edlerd: right, this is better indeed.

edlerd:
> A couple of suggestions

All good ideas, thanks for the suggestions. Please see the revised changeset.

src/sass/_permission_groups.scss
```diff
@@ -143,13 +143,24 @@
         position: relative;
```
anusha-c18: Don't we only need this when the `width > $breakpoint-large`?

anusha-c18 (changes requested):
Thanks for the fixes, one last question, and then we're good to merge

src/sass/_permission_groups.scss
edlerd: Yes, that simplifies this a bit more as well.

PR #1372, opened 2025-07-09
fix: [WD-22935] Add hook for small screens

Kxiru:
## Done

- Amended the existing useSmallScreen hook to be more general and persisted this throughout the code base (most uses of isWidthBelow)

## QA

1. Run the LXD-UI:
    - Ensure that changed files still work as intended.

## Screenshots

N/A


src/components/Navigation.tsx
```diff
@@ -32,8 +31,9 @@ import ProjectPermissionWarning from "pages/projects/ProjectPermissionWarning";
 import { useSettings } from "context/useSettings";
 import type { LxdProject } from "types/project";
 import { isDarkTheme, loadTheme } from "pages/settings/SettingThemeSwitcher";
+import { useSmallScreen } from "context/useSmallScreen";
 
-const isSmallScreen = () => isWidthBelow(620);
+const isSmallScreen = useSmallScreen();
```
edlerd: We can't invoke the hook here on the top level, it must be in a component. This is probably what is breaking CI.

edlerd (changes requested):

src/context/useSmallScreen.tsx
```diff
@@ -2,12 +2,20 @@ import useEventListener from "util/useEventListener";
...
 import { isWidthBelow } from "util/helpers";
 
-export const useSmallScreen = (): boolean => {
-  const [isSmallScreen, setIsSmallScreen] = useState(isWidthBelow(620));
+export const smallScreenSize = 620;
+export const mediumScreenSize = 820;
+export const largeScreenSize = 1200;
+
+export const useSmallScreen = (
+  currentScreenSize = smallScreenSize,
```
edlerd: Naming is a bit confusing. The parameter we pass in here is the breakpoint between small and not small, not really a current size.

src/context/useSmallScreen.tsx
```diff
@@ -2,12 +2,20 @@ import useEventListener from "util/useEventListener";
 import { useState } from "react";
 import { isWidthBelow } from "util/helpers";
 
-export const useSmallScreen = (): boolean => {
-  const [isSmallScreen, setIsSmallScreen] = useState(isWidthBelow(620));
+export const smallScreenSize = 620;
+export const mediumScreenSize = 820;
+export const largeScreenSize = 1200;
```
edlerd: I think we should call these breakpoints as well.

src/pages/instances/InstanceList.tsx
```diff
@@ -83,7 +82,7 @@ const saveHidden = (columns: string[]) => {
   localStorage.setItem("instanceListHiddenColumns", JSON.stringify(columns));
 };
 
-const isMediumScreen = () => isWidthBelow(820);
+const isMediumScreen = useSmallScreen(mediumScreenSize);
```
edlerd: This is what is breaking the instance list. We can't call the hook outisde a component according to the rules for hooks. See also https://react.dev/reference/rules/rules-of-hooks

src/context/useSmallScreen.tsx
```diff
@@ -2,12 +2,20 @@ import useEventListener from "util/useEventListener";
...
+
+export const useSmallScreen = (
```
edlerd: The hook name should be adjusted. This is now also used to figure out if the screen is medium. Maybe something like `useIsScreenWidthBelow`? Or find a better suiting name, my suggestion is a bit longish.

src/context/useSmallScreen.tsx
```diff
@@ -2,12 +2,20 @@ import useEventListener from "util/useEventListener";
 import { useState } from "react";
 import { isWidthBelow } from "util/helpers";
```
edlerd: I would recommend putting this function directly in here and making it local, so we ensure using the hook with updates on resize and avoid using the raw function.

Kxiru: As discussed, we won't be moving the function into this file due to other usages.

src/context/useIsScreenBelow.tsx
```diff
@@ -0,0 +1,23 @@
...
+    isWidthBelow(currentBreakpoint),
+  );
+  const handleResize = () => {
+    const newSize = isWidthBelow(currentBreakpoint);
+    if (newSize !== isSmallScreen) {
+      setIsSmallScreen(newSize);
+    }
+  };
+  useEventListener("resize", handleResize);
+  return isSmallScreen;
```
edlerd: Suggesting a slight adjusting in naming and simplification of the hook logic.

```suggestion
  breakpoint = smallScreenBreakpoint,
): boolean => {
  const [isScreenBelow, setScreenBelow] = useState(isWidthBelow(breakpoint));

  const updateIsBelow = () => {
    if (isWidthBelow(breakpoint) !== isScreenBelow) {
      setScreenBelow(!isScreenBelow);
    }
  };
  useEventListener("resize", updateIsBelow);

  return isScreenBelow;
```

edlerd (dismissed):
QA looks all right, thanks for adjusting this and rolling it out consistently. Small nitpick suggestions on the hook internal code. Also ok as is.

src/context/useIsScreenBelow.tsx
Kxiru: Thanks David! Please re-approve when you can.

PR #1377, opened 2025-07-15
feat(tests): [WD-22649] Screenshot automation tests

Kxiru:
## Done

- Captures screenshots for the LXD docs (UI sections)
  - Instances (6) :heavy_check_mark: 
  - Networks (9) :heavy_check_mark: 
  - Storage (18)
    - Pools (4) :heavy_check_mark: 
    - Volumes (8) :heavy_check_mark: 
    - Volume snapshots (6) :heavy_check_mark: 
  - Tutorial (13) :heavy_check_mark: 
  - UI (20) :heavy_check_mark: 

Fixes lack of screenshot automation in the UI.

## QA

1. Run the LXD-UI:
    - Locally run tests in docs-screenshots.spec.ts and observe the generated screenshots in screenshots/doc/images/...

## Screenshots

Locally run tests in docs-screenshots.spec.ts and observe the generated screenshots in screenshots/doc/images/...

### DIFF
--- a/tests/docs-screenshots-clustered.spec.ts
+++ b/tests/docs-screenshots-clustered.spec.ts
@@ -1,4 +1,5 @@
 import { test } from "./fixtures/lxd-test";
+import { getClipPosition } from "./helpers/doc-screenshots";
 import { gotoURL } from "./helpers/navigate";
 import { createPool, deletePool } from "./helpers/storagePool";
 
@@ -9,20 +10,9 @@ test.beforeEach(() => {
   );
 });
 
-// x,y is top left coordinate, xx,yy is bottom right coordinate
-// gimp provides the coordinates of the area easily
-const getClipPosition = (x: number, y: number, xx: number, yy: number) => {
-  return {
-    x: x,
-    y: y,
-    width: xx - x,
-    height: yy - y,
-  };
-};
-
 //Run in a clustered backend
 
-test("clustered storage pools", async ({ page }) => {
+test("Clustered storage pools", async ({ page }) => {
   //Clustered storage pool creation
   const poolName = "clustered-pool";
   page.setViewportSize({ width: 1440, height: 1000 });
@@ -63,4 +53,26 @@ test("Clustered storage volumes", async ({ page }) => {
     clip: getClipPosition(240, 0, 1420, 750),
   });
   await deletePool(page, poolName);
+});
+
+test("LXD - UI Folder - Clustered", async ({ page }) => {
+  //This test assumes that there is a cluster member named micro2 within the environment. This is the value within the original screenshot, but it can also be changed to accomodate alternative names.
+  const clusterMemberName = "micro2";
+  page.setViewportSize({ width: 1440, height: 800 });
+  await gotoURL(page, "/ui/");
+  await page.getByRole("link", { name: "Instances", exact: true }).click();
+  await page.getByText("Create instance").click();
+  await page.getByPlaceholder("Enter name").fill("Ubuntu-vm-server2");
+  await page.getByRole("button", { name: "Browse images" }).click();
+  await page
+    .locator("tr")
+    .filter({ hasText: "Ubuntu24.04 LTSnoblealldefaultUbuntuSelect" })
+    .getByRole("button")
+    .click();
+  await page.getByLabel("Cluster member").selectOption(clusterMemberName);
+  await page.getByLabel("Cluster member").dblclick();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/ui/create_instance_ex4.png",
+    clip: getClipPosition(240, 0, 1440, 760),
+  });
 });
--- a/tests/docs-screenshots.spec.ts
+++ b/tests/docs-screenshots.spec.ts
@@ -1,9 +1,16 @@
 import { expect, test } from "./fixtures/lxd-test";
 import { gotoURL } from "./helpers/navigate";
-import { deleteInstance, visitInstance } from "./helpers/instances";
+import {
+  deleteInstance,
+  forceStopInstance,
+  visitInstance,
+} from "./helpers/instances";
 import { createPool, deletePool } from "./helpers/storagePool";
-import { deleteNetwork, visitNetwork } from "./helpers/network";
+import { createNetwork, deleteNetwork, visitNetwork } from "./helpers/network";
 import { createVolume, deleteVolume } from "./helpers/storageVolume";
+import { setOption } from "./helpers/configuration";
+import { getClipPosition } from "./helpers/doc-screenshots";
+import { openInstancePanel } from "./helpers/instancePanel";
 
 test.beforeEach(() => {
   test.skip(
@@ -12,25 +19,13 @@ test.beforeEach(() => {
   );
 });
 
-// x,y is top left coordinate, xx,yy is bottom right coordinate
-// gimp provides the coordinates of the area easily
-const getClipPosition = (x: number, y: number, xx: number, yy: number) => {
-  return {
-    x: x,
-    y: y,
-    width: xx - x,
-    height: yy - y,
-  };
-};
-
-//Run in a non-clustered backend
-
 test("instances", async ({ page }) => {
   const pool = "other-pool";
-  const volume = "CustomVol";
-  await createPool(page, pool, "dir");
-
+  const volume = "CustomVolume";
   const instance = "comic-glider";
+
+  await page.setViewportSize({ width: 1440, height: 800 });
+  await createPool(page, pool, "dir");
   await createVolume(page, volume);
   await gotoURL(page, "/ui/");
   await page.getByText("Instances", { exact: true }).click();
@@ -57,20 +52,24 @@ test("instances", async ({ page }) => {
   await page.getByRole("button", { name: "Create override" }).click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/instances/create_instance_form.png",
-    clip: getClipPosition(240, 0, 1280, 675),
+    clip: getClipPosition(240, 0, 1360, 520),
   });
 
   await page.getByRole("button", { name: "Attach disk device" }).click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_attach_to_instance_2.png",
-    clip: getClipPosition(310, 230, 900, 480),
+    clip: getClipPosition(390, 270, 950, 520),
   });
 
+  await page.setViewportSize({ width: 1440, height: 500 });
   await page.getByRole("button", { name: "Attach custom volume" }).click();
+  await page.waitForSelector("text=Select");
+  await page.waitForTimeout(1000);
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_attach_to_instance_3.png",
-    clip: getClipPosition(20, 20, 1300, 680),
+    clip: getClipPosition(85, 20, 1350, 470),
   });
+  await page.setViewportSize({ width: 1440, height: 800 });
   await page.getByLabel("Close active modal").click();
 
   await page.getByText("Network", { exact: true }).click();
@@ -92,13 +91,13 @@ test("instances", async ({ page }) => {
 
   await page.screenshot({
     path: "tests/screenshots/doc/images/instances/instance_overview_page.png",
-    clip: getClipPosition(240, 0, 1280, 600),
+    clip: getClipPosition(240, 0, 1420, 600),
   });
 
   await page.getByRole("button", { name: "Migrate", exact: true }).click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/instances/move_instance_modal.png",
-    clip: getClipPosition(305, 210, 974, 500),
+    clip: getClipPosition(390, 250, 1040, 550),
   });
 
   await page
@@ -107,13 +106,13 @@ test("instances", async ({ page }) => {
   await page.waitForTimeout(1000);
   await page.screenshot({
     path: "tests/screenshots/doc/images/instances/move_instance_modal_2.png",
-    clip: getClipPosition(17, 80, 1262, 630),
+    clip: getClipPosition(80, 80, 1340, 650),
   });
 
   await page.getByText(pool).click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/instances/move_confirmation_modal.png",
-    clip: getClipPosition(264, 262, 1015, 458),
+    clip: getClipPosition(350, 300, 1090, 490),
   });
 
   await deleteVolume(page, volume.toLowerCase());
@@ -124,6 +123,8 @@ test("instances", async ({ page }) => {
 test("networks", async ({ page }) => {
   const network = "BridgeNetwork";
   const networkACL = network + "-ACL";
+  await page.setViewportSize({ width: 1440, height: 800 });
+
   await gotoURL(page, "/ui/");
   await page.getByText("Networking").click();
   await page.getByText("Networks").click();
@@ -195,7 +196,7 @@ test("networks", async ({ page }) => {
 
 test("storage pools", async ({ page }) => {
   const poolName = "pool1";
-  page.setViewportSize({ width: 1440, height: 800 });
+  await page.setViewportSize({ width: 1440, height: 800 });
   await gotoURL(page, "/ui/");
   await page.getByText("Storage").click();
   await page.getByRole("link", { name: "Pools" }).click();
@@ -220,7 +221,7 @@ test("storage pools", async ({ page }) => {
 test("storage volumes", async ({ page }) => {
   const poolName = "pool1";
   const volumeName = "CustomVol1";
-  page.setViewportSize({ width: 1440, height: 800 });
+  await page.setViewportSize({ width: 1440, height: 800 });
   await createPool(page, poolName);
 
   await page.getByRole("link", { name: "Volumes", exact: true }).click();
@@ -281,7 +282,7 @@ test("storage volume snapshots", async ({ page }) => {
   const volumeName = "CustomVol1";
   const snapshot = "snapshot1";
 
-  page.setViewportSize({ width: 1440, height: 800 });
+  await page.setViewportSize({ width: 1440, height: 800 });
   await createPool(page, poolName);
   await createVolume(page, volumeName);
 
@@ -291,7 +292,6 @@ test("storage volume snapshots", async ({ page }) => {
   await expect(
     page.getByText("There are no snapshots for this volume."),
   ).toBeVisible();
-
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_tab.png",
     clip: getClipPosition(240, 0, 1420, 350),
@@ -300,7 +300,6 @@ test("storage volume snapshots", async ({ page }) => {
   await page.getByRole("button", { name: "Create snapshot" }).click();
   await page.getByLabel("Snapshot name").click();
   await page.getByLabel("Snapshot name").fill(snapshot);
-
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_create.png",
     clip: getClipPosition(480, 250, 950, 560),
@@ -310,29 +309,25 @@ test("storage volume snapshots", async ({ page }) => {
   await page.waitForSelector(
     `text=Snapshot ${snapshot} created for volume ${volumeName}.`,
   );
-
+  await page.reload();
+  await page.waitForSelector("text=Showing 1 out of 1 snapshot");
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_list.png",
     clip: getClipPosition(240, 0, 1420, 320),
   });
 
-  await page
-    .getByRole("row", { name: "Name" })
-    .filter({ hasText: snapshot })
-    .hover();
-  await page.getByLabel("Edit snapshot").click();
-
+  await page.getByRole("button", { name: "See configuration" }).click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_configuration.png",
-    clip: getClipPosition(480, 250, 950, 560),
+    clip: getClipPosition(280, 190, 1165, 640),
   });
 
   await deleteVolume(page, volumeName.toLowerCase());
   await deletePool(page, poolName);
 });
 
-test("LXD - tutorial", async ({ page }) => {
-  page.setViewportSize({ width: 1440, height: 800 });
+test("LXD - Tutorial folder", async ({ page }) => {
+  await page.setViewportSize({ width: 1440, height: 800 });
 
   const instance = "Instance1";
   await gotoURL(page, "/ui/");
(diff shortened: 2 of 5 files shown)


tests/docs-screenshots-clustered.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+// x,y is top left coordinate, xx,yy is bottom right coordinate
+// gimp provides the coordinates of the area easily
+const getClipPosition = (x: number, y: number, xx: number, yy: number) => {
+  return {
+    x: x,
+    y: y,
+    width: xx - x,
+    height: yy - y,
+  };
+};
```
edlerd: Better move this to helpers instead of duplicating.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +117,592 @@ test("instances", async ({ page }) => {
...
+  // await page.screenshot({
+  //   path: "tests/screenshots/doc/images/tutorial/desktop_console.png",
+  //   clip: getClipPosition(0, 0, 1440, 760),
+  // });
+
+  // // Hello world desktop
+  // await page.screenshot({
+  //   path: "tests/screenshots/doc/images/tutorial/hello_world_desktop.png",
+  //   clip: getClipPosition(0, 0, 1440, 760),
+  // });
```
edlerd: Can this be removed? Or should it be uncommented?

Kxiru: @edlerd , I wanted to get your opinion on this line "desktop_console.png and hello_world_desktop.png may need to be screenshotted automatically as it take too long for the instance [graphical console] to start up by itself. The test times out.". What do you think?

edlerd: How about putting these two into a dedicated test suite? Then the timeout of 2 minutes might not be hit.

Also, I think we should have the click path as part of the test that opens the file browser and goes to the folder as well as creating necessary files in the screen

Kxiru: Hey @edlerd ,
I've spent all day trying to fine tune these graphical console tests and get them to work. The problem with these two tests specifically is that this graphical console does not load within 2 minutes even when isolated. I am not sure how to circumvent that.

image

I'm pushing a commit amend where a separate test suite ("LXD - Tutorial - Graphical consoles") is solely responsible for setting up these two tests. Perhaps you can take a look at it and see if it is feasible (Perhaps my hardware is just incapable of running this any faster). But if you end up agreeing that these may need to be left out, I will incorporate the other screenshots (There are some screenshots taken as part of the setup) back into "LXD - Tutorial folder".

edlerd: Let's isolate this into a dedicated test, and then we can decide how to run it. Either manually running with the prerequisite of increasing the timeout before. Or we could in CI script the timeout extension for this very test.

edlerd: Did some testing and with a cached desktop image the boot only takes ~40 seconds. So should be within the timeout of two minutes if the image is cached. I think we can make it a prerequisite, the desktop image is cached. And as we schedule the job later in CI, we can add a pre-run command that is downloading the image.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  await page
+    .getByRole("row", { name: "Exposed CPU limit Which CPUs" })
+    .getByRole("button")
+    .click();
+  await page.getByPlaceholder("Number of exposed cores").fill("1");
+  await page
+    .getByRole("row", { name: "Memory limit Usage limit for" })
+    .getByRole("button")
+    .click();
+  await page.getByPlaceholder("Enter value").fill("4");
```
edlerd: We could also use the `setCpuLimit` and `setMemLimit` helpers for this. Might make these lines a bit cleaner.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  await page
+    .getByRole("row", { name: "Memory limit Usage" })
+    .getByRole("button")
+    .click();
+  await page
+    .getByRole("row", { name: "Exposed CPU limit" })
+    .getByRole("button")
+    .click();
+  await page.getByPlaceholder("Number of exposed cores").fill("4");
+  await page.getByPlaceholder("Enter value").fill("8");
```
edlerd, line 470: Here we could use the helpers as well.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  await page.getByText("Security policies").click();
+  await setOption(page, "Enable secureboot", "false");
```
edlerd: I think this should also work with secureboot enabled. We found an issue with a setting in your default profile last week. Can you check if the VM also boots with the default settings, now we restored the default profile for your environment? If yes, we should remove those two lines and the comment above.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+    .getByRole("navigation", { name: "Project form navigation" })
+    .getByText("Instances")
+    .click();
+  await setOption(page, "Snapshot creation", "allow");
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/ui/create_restr_project2.png",
+    clip: getClipPosition(240, 0, 1440, 760),
+  });
+
+  //Instance Screenshots
```
edlerd, line 563: Shall we start a new test definition here?

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  await page
+    .getByRole("row", { name: "Exposed CPU limit" })
+    .getByRole("button")
+    .click();
+  await page.getByPlaceholder("Number of exposed cores").fill("1");
+  await page
+    .getByRole("row", { name: "Memory limit Usage" })
+    .getByRole("button")
+    .click();
+  await page.getByPlaceholder("Enter value").fill("8");
```
edlerd, line 612: Can use the memory / cpu helpers.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+});
+
+test("LXD - Tutorial folder", async ({ page }) => {
+  page.setViewportSize({ width: 1440, height: 800 });
+
+  const instance = "Instance1";
+  await gotoURL(page, "/ui/");
+  await page.waitForTimeout(1000);
+  await page.getByRole("button", { name: "default", exact: true }).click();
+  await page.getByRole("button", { name: "Create project" }).click();
```
edlerd, line 336: This is a bit flaky. Better reuse the `openProjectCreationForm` helper for this.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+
+  await page.getByRole("button", { name: "Create and start" }).click();
+  await page.getByTestId("notification-close-button").click();
+
+  await page
+    .getByRole("row", {
+      name: `Select ${instance} default Name Type Description IPv4 Status Actions`,
+    })
+    .getByLabel("Name")
+    .click();
```
edlerd: The test always fails on this point for me. I think we have to wait for the message "Created and started instance ...". And only close the notification then.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+    path: "tests/screenshots/doc/images/ui/snapshots_cron.png",
+    clip: getClipPosition(280, 50, 1165, 770),
+  });
+  await page.getByText("Choose interval").click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/ui/snapshots_daily.png",
+    clip: getClipPosition(280, 110, 1165, 715),
+  });
+  await page.getByLabel("Close active modal").click();
+  await page.getByRole("button", { name: "Delete" }).click();
```
edlerd: The instance is running at this point, so the test always failed for me at this line. We have to stop it (can force stop for speed) before.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+
+test("LXD - UI Folder - Networks", async ({ page }) => {
+  page.setViewportSize({ width: 1440, height: 800 });
+
+  // Network forwards screenshots
+  // Environment must have OVN configured
+
+  const network1 = "lxdbr0";
+  const network2 = "ovntest";
+  await createNetwork(page, network2, "ovn");
```
edlerd: I think we can call this ovntest, but we should just use a bridge for the screen. The output is (as far as I am aware) the same and bridges are way simpler to setup. We won't have OVN by default when running this test suite.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  let networkSubnet = await page.inputValue("input#ipv4_address");
+  let listenAddress = networkSubnet.replace("1/24", "1");
+  let targetAddress = networkSubnet.replace("1/24", "3");
+  await page.getByTestId("tab-link-Forwards").click();
+  await page.getByRole("link", { name: "Create forward" }).click();
+  await page.getByText("Manually enter address").click();
+  await page.getByLabel("Listen address").fill(listenAddress);
+  await page.getByLabel("Default target address").fill(targetAddress);
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/ui/forward_create_ovn.png",
```
edlerd:
```suggestion
    path: "tests/screenshots/doc/images/UI/forward_create_ovn.png",
```
This and all the other `/UI/` paths need to be uppercase to match the one in LXD docs.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  await page.getByLabel("Copy volume").click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/storage/storage_volumes_copy_modal.png",
+    clip: getClipPosition(445, 130, 1000, 680),
+  });
+  await page.getByLabel("Close active modal").click();
+
+  await page.getByRole("button", { name: "Export" }).click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/storage/storage_volumes_export.png",
```
edlerd, line 271:
```suggestion
    path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_export.png",
```
Should match LXD name.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  const volumeName = "CustomVol1";
+  page.setViewportSize({ width: 1440, height: 800 });
+  await createPool(page, poolName);
+
+  await page.getByRole("link", { name: "Volumes", exact: true }).click();
+  await page.getByText("Create volume").click();
+
+  await page.getByRole("button", { name: "Upload volume file" }).click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/storage/storage_volumes_import.png",
```
edlerd, line 232:
```suggestion
    path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_import.png",
```
This should match the LXD name.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+  // Project Screenshots
+  await page.waitForTimeout(1000);
+  await page.getByRole("button", { name: "default", exact: true }).click();
+  await page.getByRole("button", { name: "Create project" }).click();
+  await page.getByPlaceholder("Enter name").fill("my-project");
+  await page
+    .getByRole("combobox", { name: "Features" })
+    .selectOption("customised");
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/ui/create_project.png",
```
edlerd: This seems like the wrong screenshot. The original is the initial create project form, without the features selector highlighted. Is this an intended change?

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +107,623 @@ test("instances", async ({ page }) => {
...
+    clip: getClipPosition(240, 0, 1420, 320),
+  });
+
+  await page
+    .getByRole("row", { name: "Name" })
+    .filter({ hasText: snapshot })
+    .hover();
+  await page.getByLabel("Edit snapshot").click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_configuration.png",
```
edlerd, line 321: This is a different screen in LXD.

edlerd (commented):
This is really nice, well done! Some nitpicks on code and a couple of screens or paths don't match what is in the LXD docs yet. Especially in the networking / forward, there were some more screens with differences.

tests/docs-screenshots.spec.ts
Kxiru: This has now been changed following this PR: https://github.com/canonical/lxd/pull/16110

tests/docs-screenshots.spec.ts
Kxiru: As above.

tests/docs-screenshots.spec.ts
Kxiru: I cannot use this helper, as the screenshot must be taken with the project dropdown open, and this helper clicks on Create Project, thus closing the dropdown.

tests/docs-screenshots.spec.ts
Kxiru: I'm not sure what you mean, line 379 does that (as highlighted in your selection). Are you referring to something else?

tests/docs-screenshots.spec.ts
Kxiru: Unfortunately these helpers are only for setting limits on instances that have already been created, and are not appropriate for use _during_ instance creation. I will use them where I believe this can work :)
For example, the function itself requires a Configuration tab to be present, which is not applicable upon instance creation.

tests/docs-screenshots.spec.ts
edlerd: ah right, got it.

tests/docs-screenshots.spec.ts
edlerd: The problem I encounter is there is an info notification "your instance is being created" right after submitting the create form. But before closing we have to wait for another notification that the instance creation is completed.

tests/docs-screenshots.spec.ts
edlerd: Maybe they can be adjusted to be reusable here? Also fine to keep as is, resolving the comment.

tests/docs-screenshots.spec.ts
edlerd: This only failed for me, because I customized my columns, so should be good as is.

edlerd: Maybe we should make this agnostic to the actual columns displayed:
```
  await page
    .getByRole("row", {
      name: `Select ${instance}`,
    })
    .first()
    .getByLabel("Name")
    .click();
```

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +111,617 @@ test("instances", async ({ page }) => {
...
+  await page
+    .getByRole("navigation", { name: "Project form navigation" })
+    .getByText("Instances")
+    .click();
+  await setOption(page, "Snapshot creation", "allow");
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/UI/create_restr_project2.png",
+    clip: getClipPosition(240, 0, 1440, 760),
+  });
+});
```
edlerd: In my test run the storage pools and networks were still loading, maybe add something before these two to ensure both are finished loading.

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +111,617 @@ test("instances", async ({ page }) => {
...
+  await page.getByPlaceholder("Enter value").fill("8");
+  await page
+    .getByRole("button", { name: "Create and start", exact: true })
+    .click();
+  await page.getByTestId("notification-close-button").click();
+
+  await visitInstance(page, vminstance);
+  await page.waitForSelector("text=General");
+  await page.getByTestId("tab-link-Console").click();
+  await page.waitForTimeout(35000);
```
edlerd: This wasn't enough in my local tests, can you raise this to 45 maybe?

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +111,617 @@ test("instances", async ({ page }) => {
...
+  await page.getByLabel("Port 0 target address").fill(targetAddress);
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/UI/forward_edit_ex2.png",
+    clip: getClipPosition(240, 0, 1440, 760),
+  });
+
+  await page.getByRole("button", { name: "Update" }).click();
+  await page.getByTestId("notification-close-button").click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/UI/forward_edit_ex1.png",
```
edlerd, line 728: This screen is very different. I think we need to wait after the update to ensure we are on the right page:

image

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +111,617 @@ test("instances", async ({ page }) => {
...
+  await page.getByTestId("notification-close-button").click();
+
+  await page
+    .getByRole("row", {
+      name: `Select ${instance} default Name Type Description IPv4 Status Actions`,
+    })
+    .getByLabel("Name")
+    .click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/tutorial/instance_summary.png",
```
edlerd, line 394: This should wait for the side panel content to be loaded

image

tests/docs-screenshots.spec.ts
```diff
@@ -38,17 +39,35 @@ test("instances", async ({ page }) => {
...
 
+  await page.getByRole("button", { name: "Attach disk device" }).click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/storage/storage_volumes_attach_to_instance_2.png",
+    clip: getClipPosition(310, 230, 900, 480),
+  });
+
+  await page.getByRole("button", { name: "Attach custom volume" }).click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/storage/storage_volumes_attach_to_instance_3.png",
```
edlerd, line 69: This needs to ensure the content is loaded

image

tests/docs-screenshots.spec.ts
```diff
@@ -91,20 +111,617 @@ test("instances", async ({ page }) => {
...
+  await page.getByRole("button", { name: "Attach", exact: true }).click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/tutorial/add_disk_device.png",
+    clip: getClipPosition(0, 0, 1440, 760),
   });
+
+  await page.locator("#form-footer span").first().click();
+  await page.waitForTimeout(1000);
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/tutorial/yaml_configuration.png",
```
edlerd, line 418: Need to wait a little longer, so the syntax highlighting kicks in

image

edlerd (commented):
Good progress, I really like the automated screens. Some tuning seems needed to get the timings right, then this seems good to merge.

tests/docs-screenshots.spec.ts
edlerd: This is still unresolved.

tests/docs-screenshots.spec.ts
```diff
@@ -68,43 +90,646 @@ test("instances", async ({ page }) => {
...
+    path: "tests/screenshots/doc/images/UI/create_project.png",
+    clip: getClipPosition(240, 0, 1440, 960),
+  });
+
+  await page.getByPlaceholder("Enter name").fill("my-restricted-project");
+  await page.getByText("Allow custom restrictions on").click();
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/UI/create_restr_project1.png",
+    clip: getClipPosition(240, 0, 1440, 760),
+  });
```
edlerd: Features should be set to default lxd for this screenshot

image

tests/docs-screenshots.spec.ts
```diff
@@ -68,43 +90,646 @@ test("instances", async ({ page }) => {
...
+
+  await page
+    .getByRole("row", {
+      name: `Select ${instance} default Name Type Description IPv4 Status Actions`,
+    })
+    .getByLabel("Name")
+    .click();
+  await page.waitForSelector(`text=Instance summary`);
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/tutorial/instance_summary.png",
```
edlerd, line 394: The info box should be closed on this screen

image

tests/docs-screenshots.spec.ts
```diff
@@ -68,43 +90,646 @@ test("instances", async ({ page }) => {
...
+  });
+
+  await page.getByRole("button", { name: "Create snapshot" }).last().click();
+  await page.waitForSelector(
+    `text=Snapshot ${snapshot} created for volume ${volumeName}.`,
+  );
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/storage/storage_volumes_snapshots_list.png",
+    clip: getClipPosition(240, 0, 1420, 320),
+  });
```
edlerd: The snapshot is missing in the screenshot. This might need some logic to ensure the cache reloaded and the snapshot list is visible. On subsequent runs, the snapshot was visible.

image

Kxiru:
> This is still unresolved.
The instance summary appears in my local screenshots.image

I have replaced this with OpenInstanceDetailPanel for potentially more consistency, but as for the summary page appearing at all, this should have been fixed in my last push.

tests/docs-screenshots.spec.ts
```diff
@@ -68,43 +91,647 @@ test("instances", async ({ page }) => {
...
+    path: "tests/screenshots/doc/images/tutorial/yaml_configuration.png",
+    clip: getClipPosition(0, 0, 1440, 760),
   });
+  await page.getByRole("button", { name: "Cancel" }).click();
+
+  await page.getByTestId("tab-link-Terminal").click();
+  await page.waitForTimeout(2000);
+  await page.keyboard.type("rm /usr/bin/bash");
+  await page.keyboard.press("Enter");
+  await page.reload();
```
edlerd:
```suggestion
  await page.keyboard.type("exit");
  await page.keyboard.press("Enter");
```
We can exit the terminal to produce the next screenshot. This saves us an extra page reload and it doesn't break the instance.

tests/docs-screenshots.spec.ts
```diff
@@ -68,43 +91,647 @@ test("instances", async ({ page }) => {
...
+  await page.waitForSelector(`text=The connection was closed abnormally`);
+  await page.screenshot({
+    path: "tests/screenshots/doc/images/tutorial/broken_terminal.png",
+    clip: getClipPosition(0, 0, 1440, 760),
+  });
+
+  await deleteInstance(page, instance);
```
edlerd, line 434: Somehow this always failed for me locally. Does this pass for your setup?

Kxiru: Yes, this is a generated screenshot:

Perhaps it will be more resilient now that we have adopted your approach, but all screenshots in this suite generate appropriately for me.

image

edlerd (approved):
LGTM, thanks for the automated doc screenshots :)

PR #1378, opened 2025-07-15
fix(panels) ensure side panels for instances and profiles are sticky when scrolling

edlerd:
## Done

- fix(panels) ensure side panels for instances and profiles are sticky when scrolling

## QA

1. Run the LXD-UI:
    - open instance list and click on any row of instances
    - see detail panel next to the list
    - shrink browser window height, ensure the top part of the detail panel stays sticky
    - repeat with profile detail panel on profile list

## Screenshots

image


edlerd:
> It looks like SidePanel.Container was hiding this functionality, is that correct? Nice catch!

I think this is a regression that was introduced with the generic side panel component about 18 months ago. I just realized it yesterday when I was preparing the upstream PR for the side panel.

PR #1411, opened 2025-07-31
feat: [WD-24084] Fix error messaging for creation a SB with no valid …

Kxiru:
…pools available.

## Done

- [List of work items including drive-bys]

## QA

1. Run the LXD-UI:

## Screenshots


anusha-c18 (approved):
code lgtm 👍

anusha-c18 (changes requested):
One small issue on QA
The error is shown even before I interact with the custom select 

https://github.com/user-attachments/assets/29f97984-b386-4aed-b377-d56c4d413b85

Kxiru:
> One small issue on QA The error is shown even before I interact with the custom select
> 

Oh, that's fine, the base help text is supposed to be the same. I think @edlerd just wanted the error styling to apply so it can bring attention to enviros where there are no CephObject pools.

edlerd:
The interaction in the error case is a little wired. As in the screenshot in the description, two fields have an error status, where the 2nd error is a blocking one.

I think we should put the pool field first in the form and autofocus on it. Then on leaving that field, the error becomes prominent and we avoid showing too much red text at once.

Kxiru:
Thank you, @anusha-c18 , @edlerd . I agree with your suggestions and have implemented them. Please have another look when you can.

src/pages/storage/forms/StorageBucketForm.tsx
```diff
@@ -46,32 +46,39 @@ const StorageBucketForm: FC<Props> = ({ formik, bucket }) => {
...
           disabled: !!bucketEditRestriction || isEditing,
           help: isEditing
             ? "Storage bucket pool can't be changed"
-            : "Pool must have a Ceph Object driver",
+            : formik.errors.pool
+              ? null
+              : "Pool must have a Ceph Object driver",
+
+          onBlur: formik.handleBlur,
+          autoFocus: true,
```
edlerd:
```suggestion
          takeFocus: true,
```
The field is differently named on this one.

src/pages/storage/StoragePoolSelector.tsx
```diff
@@ -8,19 +8,25 @@ import { cephObject } from "util/storageOptions";
 import StoragePoolOptionLabel from "./StoragePoolOptionLabel";
 import StoragePoolOptionHeader from "./StoragePoolOptionHeader";
 import { useStoragePools } from "context/useStoragePools";
+import type { FormikProps } from "formik";
+import type { StorageBucketFormValues } from "./forms/StorageBucketForm";
 
 interface Props {
   value: string;
   setValue: (value: string) => void;
+  formik?: FormikProps<StorageBucketFormValues>;
```
edlerd: This is a generic component and shouldn't have direct dependencies to this one use case in storage buckets.

I think we should revert all changes to this file. Instead, we can use the `selectProps` which are forwarded to the `CustomSelect`. Setting the `required`, `error` and `takeFocus` fields on that object would be a simpler solution.

Kxiru: We cannot get the error message and validation from bucketSchema (CreateStorageBucketPanel) without passing down the formik through CSBP -> StorageBucketForm -> StoragePoolSelector.
Similarly, moving the validation from CreateStorageBucketPanel to StoragePoolSelector (to pass through to Custom Select) now brings the context of storage buckets into this otherwise unrelated component.

How do you suggest we proceed.

edlerd: I suggest moving the logic for `error`, `required` and `takeFocus` into the `StorageBucketForm`. Setting the three as `selectProps` on the usage of `StoragePoolSelector`. That way, the `StoragePoolSelector` doesn't need changes and stays a generic component.

I don't fully understand your concern, maybe we should have a quick sync on this to clarify.

Kxiru: I can sync whenever you're free :+1:

Kxiru: @edlerd , I think I understand what you were trying to say now. Please can you review and let me know if the latest changes are aligned with what you were saying?

src/pages/storage/forms/StorageBucketForm.tsx
edlerd: This needs to match the name on the `CustomSelect` component that is evaluating it.

edlerd (commented):
Yes, exactly what I meant, thanks for adjusting. One tiny older issue now popped up again though :)

Kxiru:
> This needs to match the name on the CustomSelect component that is evaluating it.

Got it, thank you, @edlerd , let me know what you think, now.

src/pages/storage/panels/CreateStorageBucketPanel.tsx
```diff
@@ -40,6 +40,9 @@ const CreateStorageBucketPanel: FC = () => {
         ...testDuplicateStorageBucketName(panelParams.project, controllerState),
       )
       .required("Bucket name is required"),
+    pool: Yup.string()
+      .notOneOf([""])
```
edlerd:
```suggestion
```
This seems not needed, why would the pool become an array with empty string in it?

src/pages/storage/forms/StorageBucketForm.tsx
```diff
@@ -69,9 +58,25 @@ const StorageBucketForm: FC<Props> = ({ formik, bucket }) => {
           disabled: !!bucketEditRestriction || isEditing,
           help: isEditing
             ? "Storage bucket pool can't be changed"
-            : "Pool must have a Ceph Object driver",
+            : formik.errors.pool
+              ? null
+              : "Pool must have a Ceph Object driver",
+          error: formik.touched.pool ? formik.errors.pool : null,
```
edlerd:
```suggestion
          error: formik.errors.pool,
          onBlur: formik.handleBlur,
```
Currently the error never shows. Suggesting this to ensure the error is highlighted on switching to another field.

src/pages/storage/panels/CreateStorageBucketPanel.tsx
Kxiru: Empty string is the value of "No Storage pool available". There doesnt seem to b a yup version of .notequalto. The syntax is .notOneOf([Array of things that the value should not be one of]). Not that it shouldn't be an empty string.

Eg; .notOneOf(["x", "y"] means that the value of Pool cannot be x or y, not that it is not ["x","y"].

Also, as a reminder:
image

src/pages/storage/forms/StorageBucketForm.tsx
Kxiru: Addressed! :)

src/pages/storage/panels/CreateStorageBucketPanel.tsx
edlerd: Ah right got it. The `string.required` is already covering this as well, so the `notOneOf` to be redundant.

Kxiru: Gotcha :')

src/pages/storage/forms/StorageBucketForm.tsx
edlerd: The onblur is missing and prevents the error from showing at the moment.

edlerd (approved):
Thanks for the many iterations. QA and code LGTM!

PR #1415, opened 2025-08-05
fix(forms) update wrong detach message

ahresse:
## Done

- Fix wrong message for detaching devices.


edlerd (approved):
Good catch, thanks for the fix :+1:

PR #1419, opened 2025-08-06
feat(tests): Add e2e view-only permissions tests [WD-22228]

Kxiru:
## Done

- Adds tests for interactions with different LXD entities given a Read-only user.

## QA

1. Run the LXD-UI:
    - Create a TLS identity (Easier to run playwright tests with a TLS identity than OIDC) with Server Viewer permissions, and run the test suite locally.

## Screenshots

N/A

### DIFF
--- a/src/pages/cluster/actions/EnableClusteringBtn.tsx
+++ b/src/pages/cluster/actions/EnableClusteringBtn.tsx
@@ -1,15 +1,27 @@
 import type { FC } from "react";
 import { Button, Icon, usePortal } from "@canonical/react-components";
 import EnableClusteringModal from "pages/cluster/EnableClusteringModal";
+import { useServerEntitlements } from "util/entitlements/server";
 
 const EnableClusteringBtn: FC = () => {
+  const { canEditServerConfiguration } = useServerEntitlements();
   const { openPortal, closePortal, isOpen, Portal } = usePortal({
     programmaticallyOpen: true,
   });
 
+  const editRestriction = canEditServerConfiguration()
+    ? "Enable clustering"
+    : "You do not have permission to edit the server";
+
   return (
     <>
-      <Button appearance="positive" hasIcon onClick={openPortal}>
+      <Button
+        appearance="positive"
+        hasIcon
+        onClick={openPortal}
+        disabled={!!editRestriction}
+        title={editRestriction}
+      >
         <Icon name="plus" light />
         <span>Enable clustering</span>
       </Button>
--- a/src/pages/permissions/actions/GroupActions.tsx
+++ b/src/pages/permissions/actions/GroupActions.tsx
@@ -12,7 +12,7 @@ interface Props {
 const GroupActions: FC<Props> = ({ group }) => {
   const panelParams = usePanelParams();
   const { openPortal, closePortal, isOpen, Portal } = usePortal();
-  const { canDeleteGroup } = useGroupEntitlements();
+  const { canDeleteGroup, canEditGroup } = useGroupEntitlements();
 
   return (
     <>
@@ -29,7 +29,12 @@ const GroupActions: FC<Props> = ({ group }) => {
               panelParams.openEditGroup(group.name);
             }}
             type="button"
-            title="Edit group"
+            title={
+              canEditGroup(group)
+                ? "Edit group"
+                : "Edit group - You do not have permission to edit this group"
+            }
+            disabled={!canEditGroup(group)}
           >
             <Icon name="edit" />
           </Button>,
@@ -43,7 +48,7 @@ const GroupActions: FC<Props> = ({ group }) => {
             title={
               canDeleteGroup(group)
                 ? "Delete group"
-                : "You do not have permission to delete this group"
+                : "Delete group - You do not have permission to delete this group"
             }
             disabled={!canDeleteGroup(group)}
           >
--- a/tests/permissions-read-only.spec.ts
+++ b/tests/permissions-read-only.spec.ts
@@ -1,0 +1,477 @@
+import { test, expect } from "./fixtures/lxd-test";
+import { randomInstanceName, visitInstance } from "./helpers/instances";
+import { gotoURL } from "./helpers/navigate";
+import { execSync } from "child_process";
+import { randomProfileName, visitProfile } from "./helpers/profile";
+import { randomNetworkName, visitNetwork } from "./helpers/network";
+import { randomPoolName, visitPool } from "./helpers/storagePool";
+import {
+  randomIsoName,
+  randomVolumeName,
+  visitVolume,
+} from "./helpers/storageVolume";
+import { randomNetworkAclName, visitNetworkAcl } from "./helpers/network-acls";
+import { randomGroupName } from "./helpers/permission-groups";
+import { randomIdentityName } from "./helpers/permission-identities";
+import { openInstancePanel } from "./helpers/instancePanel";
+
+test.beforeEach(() => {
+  test.skip(
+    Boolean(process.env.CI),
+    "This suite is currently only run manually to test View-Only user permissions.",
+  );
+});
+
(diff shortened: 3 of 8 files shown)


tests/e2e-permissions.spec.ts
```diff
@@ -0,0 +1,210 @@
+import { test, expect } from "./fixtures/lxd-test";
+import { visitInstance } from "./helpers/instances";
+import { gotoURL } from "./helpers/navigate";
+import { execSync } from "child_process";
+
+test.beforeEach(() => {
+  test.skip(
+    Boolean(process.env.CI),
+    "This suite is only run manually to create screenshots for the documentation",
```
edlerd: The description seems  off.

tests/e2e-permissions.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+  const instanceName = "test-instance";
+  const profileName = "test-profile";
+  const networkName = "test-network";
+  const poolName = "test-pool";
+  const volumeName = "test-volume";
+  //   const customISOName = "test-custom-iso"; // Need ISO file
+  //   const bucketName = "test-bucket"; // Need CephObject
+  const aclName = "test-acl";
+  const identityName = "test-identity";
+  const groupName = "test-group";
```
edlerd: We should use the randomized names, to allow retries on failures. I think we have helpers for all of these entities to generate the random names.

tests/e2e-permissions.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      console.log("Storage pool created");
+
+      execSync(`lxc storage volume create ${poolName} ${volumeName}`);
+      console.log("Custom volume created");
+
+      // execSync(`lxc storage volume import ${poolName} /tmp/${customISOName}.iso ${customISOName} --type iso`);
+      // console.log("Custom ISO imported");
+
+      // execSync(`lxc bucket create ${bucketName}`);
+      // console.log("Bucket created");
```
edlerd: We probably should skip this for now, as the setup for it is too complex.

tests/e2e-permissions.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      console.log("Network created");
+
+      execSync(`lxc storage create ${poolName} dir`);
+      console.log("Storage pool created");
+
+      execSync(`lxc storage volume create ${poolName} ${volumeName}`);
+      console.log("Custom volume created");
+
+      // execSync(`lxc storage volume import ${poolName} /tmp/${customISOName}.iso ${customISOName} --type iso`);
+      // console.log("Custom ISO imported");
```
edlerd: I think you can import any random file here, we could do an exec to create a file in tmp that we import in the next exec.

edlerd (commented):
Good start, the instance tests seem reasonable, clear and simple. Please extend the other entities as well.

src/pages/cluster/actions/EnableClusteringBtn.tsx
edlerd, line 1: Ah great, thanks for catching and fixing this as a drive by!

src/pages/permissions/actions/GroupActions.tsx
```diff
@@ -29,7 +29,12 @@ const GroupActions: FC<Props> = ({ group }) => {
               panelParams.openEditGroup(group.name);
             }}
             type="button"
-            title="Edit group"
+            title={
+              canEditGroup(group)
+                ? "Edit group"
+                : "Edit group - You do not have permission to edit this group"
+            }
+            disabled={!canDeleteGroup(group)}
```
edlerd:
```suggestion
            disabled={!canEditGroup(group)}
```
I think this is a mixup.

tests/e2e-permissions.spec.ts
```diff
@@ -0,0 +1,375 @@
...
+    await expect(
+      page.getByRole("button", { name: "Upload image" }),
+    ).toBeDisabled();
+  });
+
+  test("Cannot interact with the Project", async ({ page }) => {
+    await gotoURL(page, "/ui/");
+    await page.getByText("Configuration", { exact: true }).click();
+    await expect(page.getByLabel("Description")).toBeDisabled();
+  });
```
edlerd: Please add the "Create project" button is disabled in the navigations project dropdown.

tests/e2e-permissions.spec.ts
```diff
@@ -0,0 +1,375 @@
...
+
+  test("Cannot interact with Server settings", async ({ page }) => {
+    await gotoURL(page, "/ui/");
+    await page.getByText("Server", { exact: true }).click();
+    await page.getByTestId("tab-link-Clustering").click();
+    await expect(
+      page.getByRole("button", { name: "Enable clustering" }),
+    ).toBeDisabled();
+  });
```
edlerd: This covers the new server page. Can you add the settings page as well please?

tests/e2e-permissions.spec.ts
```diff
@@ -0,0 +1,375 @@
...
+        exact: true,
+      }),
+    ).toBeDisabled();
+    await expect(
+      page.getByRole("button", {
+        name: "You do not have permission to stop this instance",
+        exact: true,
+      }),
+    ).toBeDisabled();
+  });
```
edlerd: I think instance actions have a bit more scope:
1. There are the migrate/create image/copy/export/delete actions in the header of the instance detail page
2. On the instance list there are the start/stop/etc actions and bulk actions 
3. On the instance list > side panel there are also the start/stop/etc actions.

I understand this is quite a scope increase. Would be fine with creating a future task to extend this later or doing it right now.

tests/e2e-permissions.spec.ts
edlerd, line 1: All tests in here are e2e, no need to put it in the name. Better call this permissions-read-only or alike.

edlerd (commented):
Great stuff, could also run it locally and all tests passed :+1: 

Small suggestions then this could land and one bigger item below.

tests/permissions-read-only.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+      page.getByRole("button", { name: "Enable clustering" }),
+    ).toBeDisabled();
+
+    await page.getByText("Settings", { exact: true }).click();
+    await expect(
+      page
+        .locator("tr")
+        .filter({ hasText: "acmeacme.agree_tos" })
+        .getByRole("button"),
+    ).toBeDisabled();
```
edlerd: This should check for the notification, that server settings are unavailable, instead. This setting will not be visible to a read only user

tests/permissions-read-only.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+    await expect(
+      page.getByRole("button", { name: "Stop", exact: true }),
+    ).toBeDisabled();
+    await page
+      .locator("#instances-table span")
+      .filter({ hasText: "Select all" })
+      .click();
+
+    // Cannot start, restart, pause or stop instances from the instance page.
```
edlerd:
```suggestion
    // Cannot start, restart, pause or stop instances from the instance list page on hover.
```

tests/permissions-read-only.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+        execSync(
+          `lxc auth identity-provider-group delete ${idpGroupName}`,
+        ).toString(),
+      );
+    } catch (err) {
+      console.error("Cleanup error:", err);
+    }
+  });
+
+  test("Cannot interact with instances", async ({ page }) => {
```
edlerd, line 155: We should also check the actions in the instance detail panel. On the instance list page click on the table row for an instance and then the side panel with instance information holds the start/stop/etc actions as well. We should check them too.

tests/permissions-read-only.spec.ts
Kxiru: Apologies, it seems that my local changes weren't pushed. Please review the next push.

PR #1420, opened 2025-08-07
fix(onboard) clearify on creating admins group on upgrades

### LINKED ISSUE #1410: One or more groups were not found: "admins"
I'm unsure if this is an issue with lxd or lxd-ui here. There was a recent LXD refresh which caused the LXD UI to update.
 
```
Name  Version      Rev    Tracking  Publisher   Notes
lxd   6.4-1c216d4  34285  6/stable  canonical✓  -
```

When using the same browser certificate I previously had I encounter the behavior described below. 

> Client certificate already present
> It looks like you you already have a client certificate installed and selected, skip to the next step.

Continuing to Step 2.

> Run the following command on the LXD server to create a new identity and generate a trust token.
> `lxc auth identity create tls/lxd-ui --group admins`
> You can replace lxd-ui with the name you want to give to the identity.

When running this command I get the following. I've gotten this error on two different machines already.

```
lxc auth identity create tls/lxd-ui --group admins
Error: Failed to create pending TLS identity: One or more groups were not found: "admins"
```

Fix, create the group and give it server permissions.

```
lxc auth group create admins
lxc auth group permission add admins server admin
lxc auth identity create tls/lxd-ui --group admins
```

Again, not sure if this is an issue with lxd or if these steps should be mentioned in lxd-ui.

edlerd:
## Done

- fix(onboard) clearify on creating admins group on upgrades

Fixes #1410

## QA

1. Run the LXD-UI:
    - open the lxd ui without being authenticated and chose TLS, step 2
    - ensure the admins group creation step is clearly labelled

## Screenshots

image


edlerd:
> Can we change this to "What does this command do?"

Sure, changed that just now.

> Should we consider the case where an `admins` group exists but doesn't have the `admin` permission? Like for example if the user has completed some of these steps manually but then stopped partway? I feel like we should put `lxc auth group permission add admins server admin` outside of the `if` block.

I assume if the group exists, it has server admin permissions. All ways to have the group include the permission. Apart from a user shooting themselves in the foot and messing things up. But I'd consider that an edge case to ignore. The problem with moving this bit out of the conditional is, that it produces an error if the permission is assigned to the group.

src/pages/login/CertificateAdd.tsx
```diff
@@ -64,66 +65,94 @@ const CertificateAdd: FC = () => {
...
+              Paste the following commands into the console for the machine
+              where LXD is running:
+            </p>
             <CodeSnippet
               className="u-no-margin--bottom"
               blocks={[
                 {
-                  code: `lxc auth identity create tls/lxd-ui --group admins`,
-                  wrapLines: true,
+                  code: `if ! lxc auth group show admins ; then lxc auth group create admins ; lxc auth group permission add admins server admin ; fi ; lxc auth identity create tls/lxd-ui --group admins`,
```
Kxiru, line 75: Is "fi" a spelling mistake?

src/pages/login/CertificateAdd.tsx
```diff
@@ -64,66 +65,94 @@ const CertificateAdd: FC = () => {
             </Row>
           )}
           <div className="p-stepped-list__content">
-            <div>
-              Run the following command on the LXD server to create a new
-              identity and generate a trust token.
-            </div>
+            <p>
+              Paste the following commands into the console for the machine
+              where LXD is running:
```
Kxiru:
```suggestion
              Paste the following commands into the console of the machine
              where LXD is running:
```

src/pages/login/CertificateAdd.tsx
edlerd: Sadly, that is the bash syntax for ending an if statement.

Kxiru: Haha, I see

PR #1423, opened 2025-08-12
chore(test) make instance terminal test more resilient

edlerd:
## Done

- chore(test) make instance terminal test more resilient


Kxiru (approved):
Thanks for making our tests more robust!

PR #1432, opened 2025-08-18
github: remove the CLA check now there is a Org level one

simondeziel:


edlerd (approved):
LGTM, thanks for cleaning the double check up.

PR #1436, opened 2025-08-19
fix(caching) settings and documentation objects files can be considered fresh in cache for up to 1 minute

edlerd:
## Done

- fix(caching) settings and documentation objects files can be considered fresh in cache for up to 1 minute
- This is to avoid excessive re-fetching of the same data.
- This should also should fix the e2e tests for 5.0 on Firefox, that triggered a race based on the queries from refetching the data

## QA

1. Run the LXD-UI:
    - ensure CI passes, no additional QA needed.


Kxiru (approved):
Thank you for identifying this fix, David!

PR #1437, opened 2025-08-19
fix(buckets) ensure storage bucket detail pages highlight the main navigation item

edlerd:
## Done

- fix(buckets) ensure storage bucket detail pages highlight the main navigation item

## QA

1. Run the LXD-UI:
    - open a storage bucket detail page
    - ensure the buckets link in the main navigation is active and not the pools link in it.

## Screenshots

image


edlerd:
> Great spot! But why is /buckets/ also put in the "ignoreUrlMatches" array when it is also the active URL?

It is ignored for the pools entry as that one matches the bucket detail page URL otherwise. So we'd end up with two active menu items.

PR #1444, opened 2025-08-20
feat: [WD-24077] Enable clustering tests + workflow integration

Kxiru:
## Done

- Added cluster-enabling check in server.spec.ts (Needs to be outside of the test suites that use clustered environments)
- Added cluster differentiation to the github workflow.

## QA

1. Run the LXD-UI:

## Screenshots

N/A

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -85,6 +85,7 @@ jobs:
       matrix:
         lxd_channel: ["5.0/edge", "5.21/edge", "latest/edge"]
         browser: ["chromium", "firefox"]
+        deployment: ["unclustered", "clustered"]
     outputs:
       job_status: ${{job.status}}
     env:
@@ -131,7 +132,6 @@ jobs:
           sudo lxc config set core.https_address "[::]:8443"
           sudo lxc config trust add keys/lxd-ui.crt
           sudo lxc config set cluster.https_address "127.0.0.1"
-          sudo lxc cluster enable local
 
       - uses: actions/setup-node@v4
         with:
@@ -153,17 +153,17 @@ jobs:
         run: sudo -E ./tests/scripts/setup_test
 
       - name: Run Playwright tests
-        run: npx playwright test --project ${{ matrix.browser }}:lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}
+        run: npx playwright test --project ${{ matrix.browser }}:lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}:${{ matrix.deployment }}
 
       - name: Rename Playwright report
         if: always()
-        run: mv blob-report/report.zip blob-report/${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}-report.zip
+        run: mv blob-report/report.zip blob-report/${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}-${{ matrix.deployment }}-report.zip
 
-      - name: Upload ${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }} blob reports to be merged
+      - name: Upload ${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}-${{ matrix.deployment }} blob reports to be merged
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: blob-report-${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}
+          name: blob-report-${{ matrix.browser }}-lxd-${{ steps.lxd-env.outputs.LXD_CHANNEL }}-${{ matrix.deployment }}
           path: blob-report
           retention-days: 1
 
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -68,21 +68,33 @@ const config: PlaywrightTestConfig<TestOptions> = {
       testMatch: "login.spec.ts",
     },
     {
-      name: "chromium:lxd-5.0-edge",
+      name: "enable-clustering-chrome",
+      // enable clustering for subsequent tests
+      use: { ...devices["Desktop Chrome"]},
+      testMatch: "enable-clustering.spec.ts",
+    },
+    {
+      name: "enable-clustering-firefox",
+      // enable clustering for subsequent tests
+      use: { ...devices["Desktop Firefox"]},
+      testMatch: "enable-clustering.spec.ts",
+    },
+    {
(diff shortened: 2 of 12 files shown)


.github/workflows/pr.yaml
```diff
@@ -131,14 +132,13 @@ jobs:
...
           sudo lxc config set cluster.https_address "127.0.0.1"
-          sudo lxc cluster enable local
 
       - uses: actions/setup-node@v4
         with:
           node-version: 22
 
       - name: Install Playwright Browsers
-        run: npx playwright install --with-deps ${{ matrix.browser }}
+        run: npx playwright install --with-deps ${{ matrix.browser }}:${{ matrix.deployment }}
```
edlerd: This is installing the browser, so we shouldn't add the deployment here.

tests/helpers/cluster.ts
```diff
@@ -11,6 +11,10 @@ export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
   );
 };
 
+export const skipIfNotClustered = (projectName: string) => {
+  test.skip(projectName.includes("clustered"));
```
edlerd:
```suggestion
  test.skip(projectName.includes("-clustered"));
```
Without the prefix, `unclustered` matches as well.

Kxiru: Ah of course, thank you.

playwright.config.ts
```diff
@@ -68,51 +68,103 @@ const config: PlaywrightTestConfig<TestOptions> = {
...
       dependencies: ["login-firefox"],
     },
+    {
+      name: "chromium:lxd-5.0-edge:clustered",
+      use: {
+        ...devices["Desktop Chrome"],
+        lxdVersion: "5.0-edge",
+      },
+      testMatch: "*-clustered.spec.ts",
+    },
```
edlerd:
```suggestion
    {
      name: "chromium:lxd-5.0-edge:clustered",
      use: {
        ...devices["Desktop Chrome"],
        lxdVersion: "5.0-edge",
      },
      testMatch: "*-clustered.spec.ts",
      dependencies: ["enable-clusterinng-chrome"],
    },
```
We need to enable clustering before running the new `:clustered` suites. So maybe add another suite similar to login tests just to click the enable clustering button?

playwright.config.ts
```diff
@@ -68,51 +68,118 @@ const config: PlaywrightTestConfig<TestOptions> = {
       testMatch: "login.spec.ts",
     },
     {
-      name: "chromium:lxd-5.0-edge",
+      name: "enable-clustering-chrome",
+      // enable clustering for subsequent tests
+      use: { ...devices["Desktop Firefox"], trace: "off" },
```
edlerd:
```suggestion
      use: { ...devices["Desktop Chrome"] },
```
This is chrome, and we only disable tracing for login to not leak the password.

playwright.config.ts
```diff
@@ -68,51 +68,118 @@ const config: PlaywrightTestConfig<TestOptions> = {
...
       dependencies: ["login-firefox"],
     },
+    {
+      name: "chromium:lxd-5.0-edge:clustered",
+      use: {
+        ...devices["Desktop Chrome"],
+        lxdVersion: "5.0-edge",
+      },
+      testMatch: "*-clustered.spec.ts",
+      dependencies: ["enable-clustering-chrome"],
```
edlerd, line 136: We should have this dependency for all other :clustered suites and remove the login dependency for those. Best to have two versions of this for chrome and firefox.

playwright.config.ts
```diff
@@ -122,6 +189,16 @@ const config: PlaywrightTestConfig<TestOptions> = {
...
       dependencies: ["login-chromium"],
     },
+    {
+      name: "coverage-clustered",
+      use: {
+        ...devices["Desktop Chrome"],
+        lxdVersion: "latest-edge",
+        hasCoverage: true,
+      },
+      dependencies: ["login-chromium"],
```
edlerd: Here we also need the "enable clustering" dependency instead of the login one.

tests/enable-clustering.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { test, expect } from "./fixtures/lxd-test";
+import { skipIfNotSupported } from "./helpers/cluster";
+import { gotoURL } from "./helpers/navigate";
+
+test("check enabling clustering", async ({ page, lxdVersion }) => {
+  skipIfNotSupported(lxdVersion);
```
edlerd, line 6: this test should only run in the `enable-clustering` suites. Similar to the login spec test.

playwright.config.ts
Kxiru: Do we need a Firefox version as well?

tests/enable-clustering.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+test("check enabling clustering", async ({ page, lxdVersion }, testInfo) => {
+  skipIfNotSupported(lxdVersion);
+  test.skip(!testInfo.project.name.includes("enable-clustering"));
+
+  await gotoURL(page, "/ui/");
+  await page.getByText("Server", { exact: true }).click();
+  await page.getByTestId("tab-link-Clustering").click();
+  await expect(page.getByText("This server is not clustered")).toBeVisible();
+  await expect(page.getByText("Enable clustering")).toBeVisible();
+  await page.getByText("Enable clustering").click();
```
edlerd: This test also needs to fill and submit the modal.

tests/clusters-clustered.spec.ts
edlerd, line 1: Naming of the file is a bit strange. Maybe call it groups-clustered.spec.ts and split off the members-clustered.spec.ts -- or maybe you have a better idea for it.

tests/helpers/cluster.ts
```diff
@@ -11,6 +11,10 @@ export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
   );
 };
 
+export const skipIfNotClustered = (projectName: string) => {
+  test.skip(projectName.includes("-clustered"));
```
edlerd:
```suggestion
  test.skip(!projectName.includes("-clustered"));
```
The inverse should be right here.

tests/helpers/cluster.ts
edlerd, line 1: Now that we run the tests against an unclustered backend, a couple of suites need fixing, such as this function, as the new main navigation entry is server and the "clustering" link can't be found. So we need to change the condition. If clustering is visible we are on a clustered server if server is visible we are on an unclustered one.

image

playwright.config.ts
edlerd, line 1: Another issue in instances.spec.ts that has the assumption of the "cluster member" column being visible. That has changed since the backend is now unclustered. We have to find a way to be agnostic about the columns:

image

edlerd (dismissed):
LGTM, good work :+1:

PR #1448, opened 2025-08-21
feat: [WD-24083] Run low permissions tests in github

Kxiru:
## Done

- First attempt at automating e2e permissions tests in CI.

## QA

1. Run the LXD-UI:

## Screenshots

N/A

### DIFF
--- a/tests/helpers/permissions.ts
+++ b/tests/helpers/permissions.ts
@@ -1,3 +1,4 @@
+import { execSync } from "child_process";
 import type { LxdVersions } from "../fixtures/lxd-test";
 import { test, expect } from "../fixtures/lxd-test";
 import type { Page } from "@playwright/test";
@@ -50,4 +51,11 @@ export const confirmGroupsModifiedForIdentity = async (
   for (const group of groups) {
     await expect(identityRow.getByText(`${symbol} ${group}`)).toBeVisible();
   }
+};
+
+export const runCommand = (command: string) => {
+  console.log("Running command: ", command);
+  const result = execSync(`sudo -E ${command}`).toString();
+  console.log("Result: ", result);
+  return result;
 };
--- a/tests/permissions-read-only.spec.ts
+++ b/tests/permissions-read-only.spec.ts
@@ -1,7 +1,10 @@
 import { test, expect } from "./fixtures/lxd-test";
-import { randomInstanceName, visitInstance } from "./helpers/instances";
+import {
+  randomInstanceName,
+  selectAllInstances,
+  visitInstance,
+} from "./helpers/instances";
 import { gotoURL } from "./helpers/navigate";
-import { execSync } from "child_process";
 import { randomProfileName, visitProfile } from "./helpers/profile";
 import { randomNetworkName, visitNetwork } from "./helpers/network";
 import { randomPoolName, visitPool } from "./helpers/storagePool";
@@ -14,18 +17,13 @@ import { randomNetworkAclName, visitNetworkAcl } from "./helpers/network-acls";
 import { randomGroupName } from "./helpers/permission-groups";
 import { randomIdentityName } from "./helpers/permission-identities";
 import { openInstancePanel } from "./helpers/instancePanel";
-
-test.beforeEach(() => {
-  test.skip(
-    Boolean(process.env.CI),
-    "This suite is currently only run manually to test View-Only user permissions.",
-  );
-});
+import { runCommand, skipIfNotSupported } from "./helpers/permissions";
 
 test.describe("Given a user with Viewer Server permissions...", () => {
   const ISO_FILE = "./tests/fixtures/foo.iso";
 
-  const instanceName = randomInstanceName();
+  const instanceName1 = randomInstanceName() + "-1";
+  const instanceName2 = randomInstanceName() + "-2";
   const profileName = randomProfileName();
   const networkName = randomNetworkName();
   const poolName = randomPoolName();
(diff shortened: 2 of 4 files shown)


tests/permissions-read-only.spec.ts
```diff
@@ -97,6 +90,29 @@ test.describe("Given a user with Viewer Server permissions...", () => {
...
+          `lxc auth identity group add tls/lxd-ui test-viewers`,
+        ).toString(),
+      );
+
+      console.log(
+        "Grant viewer entitlement to test-viewers group",
+        execSync(
+          `lxc auth group permission add test-viewers server viewer`,
+        ).toString(),
+      );
```
edlerd: We should remove the current user from the admin group here

tests/permissions-read-only.spec.ts
```diff
@@ -150,6 +166,34 @@ test.describe("Given a user with Viewer Server permissions...", () => {
...
+        "Grant admin entitlement to test-admins group",
+        execSync(
+          `lxc auth group permission add test-admins server admin`,
+        ).toString(),
+      );
+
+      console.log(
+        "Delete test-viewers group",
+        execSync(`lxc auth group delete test-viewers`).toString(),
+      );
```
edlerd: We should only delete the test-viewers group here and add the user back to the admins group.

edlerd: Here we should restore the unrestricted user
1. delete restricted user `lxc auth identity delete tls/lxd-ui`
2. Add the unrestricted user. Again best to extract this for reuse.
```
  execSync("sudo -E lxc config trust add keys/lxd-ui.crt");
```

tests/permissions-read-only.spec.ts
edlerd: 1. remove the unrestricted user. Same as in login.spec.tsx. Maybe extract this logic to reuse
```
  const fingerprint = execSync(
    "sudo -E lxc config trust list | grep lxd-ui.crt | awk '{print $8}'",
  ).toString();
  execSync(`sudo -E lxc config trust remove ${fingerprint}`);

```
2. Add a new user with the used certificate with restricted permissions using your new group.
```
lxc auth identity create tls/lxd-ui --group test-viewers keys/lxd-ui.crt
```

tests/permissions-read-only.spec.ts
edlerd, line 1: A couple of test failures in CI based on this suite, see https://github.com/canonical/lxd-ui/actions/runs/17439055591/job/49516919430?pr=1448

tests/permissions-read-only.spec.ts
```diff
@@ -166,20 +214,36 @@ test.describe("Given a user with Viewer Server permissions...", () => {
...
+        name: "select Name Type Description Status Actions",
+      })
+      .getByLabel("multiselect rows")
       .click();
+    await page.getByRole("button", { name: "Select all instances" }).click();
     await expect(
-      page.getByRole("button", { name: "Start", exact: true }),
+      page.getByTitle(
+        "You do not have permission to start the selected instances",
+      ),
```
edlerd, line 127: This fails, because we are only selecting one instance here. Maybe use a more robust way to select either only one instance or make sure the logic works with both one or many instances. We can encounter both depending on previous test runs, that can crash and leave behind instances.

edlerd (changes requested):
We need to disalbe the suite (and the before all and after all) for lxd 5.0

tests/permissions-read-only.spec.ts
```diff
@@ -36,123 +30,199 @@ test.describe("Given a user with Viewer Server permissions...", () => {
   const groupName = randomGroupName();
   const idpGroupName = "idp-" + randomGroupName();
 
-  test.beforeAll(() => {
+  test.beforeAll(({ lxdVersion }) => {
+    if (lxdVersion === "5.0-edge") {
+      return;
+    }
```
edlerd: Please add a log message here why we are skipping. This might simplify future debugging.

tests/permissions-read-only.spec.ts
```diff
@@ -188,12 +274,12 @@ test.describe("Given a user with Viewer Server permissions...", () => {
```
edlerd, line 150: This bit here is still broken. Maybe put the select all instances in a helper funciton to make it easier maintainable.

edlerd: The reference here got bamboozled. this refers to the second time we do "select all" to unselect instances in the permission instance state change test.

tests/permissions-read-only.spec.ts
```diff
@@ -440,11 +558,17 @@ test.describe("Given a user with Viewer Server permissions...", () => {
     await page.getByPlaceholder("Filter").click();
     await page.getByPlaceholder("Filter").fill(identityName);
     await page.keyboard.press("Enter");
-    await expect(page.getByLabel("Manage groups")).toBeDisabled();
-    await expect(page.getByLabel("Delete identity")).toBeDisabled();
+    await page.waitForTimeout(200);
+    await expect(page.getByLabel("Manage groups").first()).toBeDisabled();
+    await expect(page.getByLabel("Delete identity").first()).toBeDisabled();
```
edlerd:
```suggestion
    await expect(page.getByLabel("Manage groups")).toBeDisabled();
    await expect(page.getByLabel("Delete identity")).toBeDisabled();
```
Can we target a specific row? Picking the first could pick a random identity from another test run that just happens to be the first one in the list. I think the filter above should ensure we just have one row available.

tests/permissions-read-only.spec.ts
```diff
@@ -36,123 +34,200 @@ test.describe("Given a user with Viewer Server permissions...", () => {
...
       console.log(
-        "[Instance Created]",
-        execSync(`lxc init ubuntu:24.04 ${instanceName}`).toString(),
+        "[Instance 1 Created]",
+        execSync(`sudo -E lxc init ubuntu:24.04 ${instanceName1}`).toString(),
+      );
+      console.log(
+        "[Instance 2 Created]",
+        execSync(`sudo -E lxc init ubuntu:24.04 ${instanceName2}`).toString(),
       );
```
edlerd: The commands here are fairly verbose. How about a helper function like below and dropping the description of the command?
```
  const runCommand = (command: string): string => {
    console.log("run command: ", command);
    const result = execSync(`sudo -E ${command}`).toString();
    console.log("got result: ", result);
    return result;
  };
```
Then the calls simplify to
```suggestion
      runCommand(`lxc init ubuntu:24.04 ${instanceName2}`);
```

tests/permissions-read-only.spec.ts
```diff
@@ -36,123 +34,200 @@ test.describe("Given a user with Viewer Server permissions...", () => {
...
+      console.error("Error occurred during afterAll cleanup:", err);
+    }
   });
 
-  test("Cannot interact with instances", async ({ page }) => {
+  test("Cannot interact with Instances", async ({ page, lxdVersion }) => {
+    test.skip(
+      lxdVersion === "5.0-edge",
+      "Lxd v5.0/edge does not support fine-grained permissions",
+    );
```
edlerd: We should extract a function similar to `skipIfNotSupported` in the cluster tests to not repeat this below.

tests/permissions-read-only.spec.ts
```diff
@@ -440,11 +559,17 @@ test.describe("Given a user with Viewer Server permissions...", () => {
     await page.getByPlaceholder("Filter").click();
     await page.getByPlaceholder("Filter").fill(identityName);
     await page.keyboard.press("Enter");
+    await page.waitForTimeout(200);
```
edlerd: Please add a comment, explaining that we are waiting for the filter to get applied here.

edlerd (commented):
Good progress on this one! Passing at last :)

Some small nitpicks and suggestions to simplify, then this should be good to merge.

tests/helpers/permissions.ts
```diff
@@ -51,3 +52,7 @@ export const confirmGroupsModifiedForIdentity = async (
     await expect(identityRow.getByText(`${symbol} ${group}`)).toBeVisible();
   }
 };
+
+export const runCommand = (title: string, command: string) => {
+  console.log(title, execSync(`sudo -E ${command}`).toString());
+};
```
edlerd:
```suggestion
  const runCommand = (command: string): string => {
    console.log("run command: ", command);
    const result = execSync(`sudo -E ${command}`).toString();
    console.log("got result: ", result);
    return result;
  };
```
I'd suggest dropping the title and log the raw command. Lxc commands are fairly speaking and easy to read. Also add a return of the result. So in the cases like producing the fingerprint we can use the function as well. There are some more cases that call lxc commands which could reuse this function too, making it generic.

edlerd (dismissed):
LGTM, thanks for scheduling the permission tests. One more suggestion on the new function, but we can also keep that as a follow-up task to refactor and use it everywhere, as it increases the scope a bit.

tests/permissions-read-only.spec.ts
```diff
@@ -36,123 +35,81 @@ test.describe("Given a user with Viewer Server permissions...", () => {
...
-      );
-      console.log(
-        "[IDP Group Deleted]",
-        execSync(
-          `lxc auth identity-provider-group delete ${idpGroupName}`,
-        ).toString(),
+      const fingerprint = execSync(
+        "sudo -E lxc config trust list | grep lxd-ui.crt | awk '{print $8}'",
+      ).toString();
+
```
edlerd:
```suggestion
      const fingerprint = runCommand("lxc config trust list | grep lxd-ui.crt | awk '{print $8}'");
```
Now we can use the helper function here as well.

edlerd (dismissed):
LGTM, thanks :)

PR #1457, opened 2025-08-27
feat: [WD-24078] Networking tests in clustered env

Kxiru:
## Done

- Relies on https://github.com/canonical/lxd-ui/pull/1444 to be merged first, to pass
- Adds CRUD testing for networking operations that use clustered functionality.

## QA

1. Run the LXD-UI:
    - N/A

## Screenshots

N/A


tests/helpers/network.ts
```diff
@@ -68,8 +68,11 @@ export const visitNetworkConfiguration = async (page: Page, tab: string) => {
   await activateAllTableOverrides(page);
 };
 
-export const createNetworkForward = async (page: Page, network: string) => {
-  await createNetwork(page, network);
+export const createNetworkForward = async (
+  page: Page,
+  network: string,
+  serverClustered?: boolean,
```
edlerd:
```suggestion
```
I think it will be simpler to make the call to `isServerClustered` inside this function instead of expanding the interface here and expect the caller to pass this in.

tests/networks-clustered.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  createNetwork,
+  createNetworkForward,
+  deleteNetwork,
+  getNetworkLink,
+  randomNetworkName,
+} from "./helpers/network";
+
+let network = randomNetworkName();
+
+test.describe("Network operations in a clustered environment with multiple nodes", () => {
```
edlerd: the description is off, we only have one node.

tests/networks-clustered.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  test.beforeAll(({ browserName }) => {
+    // network names can only be 15 characters long
+    network = `${browserName.substring(0, 2)}-${network}`;
```
edlerd: I think we should create the name inside each test, wdyt?

tests/networks-clustered.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    await page.getByRole("button", { name: "Networking" }).click();
+    await page.getByRole("link", { name: "Networks", exact: true }).click();
+    await page.getByRole("button", { name: "Create network" }).click();
+    await page.getByRole("heading", { name: "Create a network" }).click();
+    await page.getByLabel("Type").selectOption("physical");
+    await page.getByLabel("Name", { exact: true }).click();
+    await page.getByLabel("Name", { exact: true }).fill(network);
+    await page.getByText("Same for all cluster members").click();
+    await page.getByLabel("Parent").selectOption({ index: 1 });
+    await page.getByRole("button", { name: "Create", exact: true }).click();
```
edlerd: Maybe a better approach is to expand or decompose the `createNetwork` helper function. Most of the logic in here is very similar and also repeated below for macvlan. so if we refactor the `createNetwork` steps, we might save a lot of code and make this test file more clear.

edlerd (commented):
Some suggestions to simplify the code below.

tests/networks-clustered.spec.ts
Kxiru: Happy to do so to remove bloat.

tests/networks-clustered.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { test } from "./fixtures/lxd-test";
+import { skipIfNotClustered, skipIfNotSupported } from "./helpers/cluster";
+import {
+  createNetwork,
+  createNetworkForward,
+  deleteNetwork,
+  randomNetworkName,
+} from "./helpers/network";
+
+let network = randomNetworkName();
```
edlerd:
```suggestion
```
I think we can define a `const` below in the three different use cases. That should be simpler and a bit more robust.

tests/networks-clustered.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  }, testInfo) => {
+    network = `${browserName.substring(0, 2)}-${network}`;
+    skipIfNotSupported(lxdVersion);
+    skipIfNotClustered(testInfo.project.name);
+
+    await createNetwork(page, network, "physical", true);
+    await deleteNetwork(page, network);
+  });
+
+  test("create a macvlan network with different parents", async ({
```
edlerd: Here as well, we only have one parent being selected.

tests/networks-clustered.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    network = `${browserName.substring(0, 2)}-${network}`;
+    skipIfNotSupported(lxdVersion);
+    skipIfNotClustered(testInfo.project.name);
+
+    await createNetwork(page, network);
+    await createNetworkForward(page, network);
+    await deleteNetwork(page, network);
+  });
+
+  test("create a physical network with different parents", async ({
```
edlerd: The title is slightly off, because only have one node available. Maybe find something a bit more precise.

edlerd (approved):
Code and CI results LGTM, thanks for adding the clustered network tests!

edlerd:
@Kxiru please squash the two commits into one and sign that one.

PR #1470, opened 2025-09-03
feat: [WD-26013] Hide acknowledged warnings by default

Kxiru:
## Done

- Implement Search functionality [ :+1: ]
- Implement Filter functionality [ :+1: ]

## QA

1. Run the LXD-UI:
    - Navigate to the Warnings page and test filtering and searching functionality.

## Screenshots

image

image

### DIFF
--- a/src/pages/warnings/WarningList.tsx
+++ b/src/pages/warnings/WarningList.tsx
@@ -17,12 +17,17 @@ import { useDocs } from "context/useDocs";
 import BulkDeleteWarningBtn from "pages/warnings/actions/BulkDeleteWarningBtn";
 import SelectableMainTable from "components/SelectableMainTable";
 import PageHeader from "components/PageHeader";
+import WarningSearchFilter from "./WarningSearchFilter";
+import { useSearchParams } from "react-router-dom";
+import type { LxdWarningSeverity, LxdWarningStatus } from "types/warning";
+import type { WarningFilters } from "util/warningFilter";
 
 const WarningList: FC = () => {
   const docBaseLink = useDocs();
   const notify = useNotify();
   const [selectedNames, setSelectedNames] = useState<string[]>([]);
   const [processingNames, setProcessingNames] = useState<string[]>([]);
+  const [searchParams] = useSearchParams();
 
   const {
     data: warnings = [],
@@ -38,15 +43,50 @@ const WarningList: FC = () => {
     notify.failure("Loading warnings failed", error);
   }
 
+  const hasWarnings = isLoading || warnings.length > 0;
+
+  const filters: WarningFilters = {
+    queries: searchParams.getAll("query"),
+    statuses: searchParams.getAll("status") as LxdWarningStatus[],
+    severities: searchParams.getAll("severity") as LxdWarningSeverity[],
+  };
+
+  const filteredWarnings = warnings.filter((item) => {
+    if (
+      !filters.queries.every(
+        (q) =>
+          item.type.toLowerCase().includes(q) ||
+          item.last_message.toLowerCase().includes(q),
+      )
+    ) {
+      return false;
+    }
+    if (
+      filters.statuses.length > 0 &&
+      !filters.statuses.includes(item.status)
+    ) {
+      return false;
+    }
+    if (
+      filters.severities.length > 0 &&
+      !filters.severities.includes(item.severity)
+    ) {
+      return false;
+    }
+    return true;
+  });
+
   useEffect(() => {
-    const validNames = new Set(warnings?.map((warning) => warning.uuid));
+    const validNames = new Set(
+      filteredWarnings?.map((warning) => warning.uuid),
+    );
     const validSelections = selectedNames.filter((name) =>
       validNames.has(name),
     );
     if (validSelections.length !== selectedNames.length) {
       setSelectedNames(validSelections);
     }
-  }, [warnings]);
+  }, [filteredWarnings]);
 
   const headers = [
     { content: "Type", sortKey: "type", className: "type" },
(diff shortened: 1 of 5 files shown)


src/pages/warnings/WarningSearchFilter.tsx
```diff
@@ -0,0 +1,94 @@
...
+export const SEVERITY = "severity";
+
+const QUERY_PARAMS = [QUERY, STATUS, SEVERITY];
+
+interface Props {
+  warnings: LxdWarning[];
+  hasProjectFilter: boolean;
+}
+
+const WarningSearchFilter: FC<Props> = ({ warnings, hasProjectFilter }) => {
```
edlerd:
```suggestion
const WarningSearchFilter: FC = () => {
```
I think we don't need any input here without the project filter.

src/pages/warnings/WarningList.tsx
```diff
@@ -145,6 +194,15 @@ const WarningList: FC = () => {
                 Warnings
               </HelpLink>
             </PageHeader.Title>
+            {hasWarnings && selectedNames.length === 0 && (
+              <PageHeader.Search>
+                <WarningSearchFilter
+                  key={`warning-${searchParams.get("search")}`}
+                  warnings={warnings}
+                  hasProjectFilter={isAllProjects}
```
edlerd: `isAllProjects` is always be false, as the warnings page is server-wide, across all projects. Maybe we should remove the project filter completely from this page? Would simplify the things quite a bit.

src/types/warning.d.ts
edlerd, line 1: Good stuff, thanks for making the types more precise :)

Kxiru:
@edlerd , this surfaces another error wherein the Warnings are not returning their projects.
image

Can we discuss?

src/pages/warnings/WarningSearchFilter.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const { data: warnings = [] } = useQuery({
+    queryKey: [queryKeys.warnings],
+    queryFn: async () => fetchWarnings(),
+    retry: false, // the api returns a 403 for users with limited permissions, surface the error right away
+  });
+
+  const projectSet = [
+    ...new Set(warnings.flatMap((warning) => warning.project)),
+  ];
```
edlerd: I'd suggest to completely drop the project filter, that simplifies this component and we avoid the issue with no projects being available.

src/util/warningFilter.tsx
```diff
@@ -0,0 +1,20 @@
+import type { LxdWarningSeverity, LxdWarningStatus } from "types/warning";
+
+export interface WarningFilters {
+  queries: string[];
+  statuses: LxdWarningStatus[];
+  severities: LxdWarningSeverity[];
+  projects: string[];
```
edlerd:
```suggestion
```

src/pages/warnings/WarningList.tsx
```diff
@@ -38,15 +43,57 @@ const WarningList: FC = () => {
...
+      !filters.severities.includes(item.severity)
+    ) {
+      return false;
+    }
+    if (
+      filters.projects.length > 0 &&
+      !filters.projects.includes(item.project)
+    ) {
+      return false;
+    }
```
edlerd:
```suggestion
```

src/pages/warnings/WarningList.tsx
```diff
@@ -38,15 +43,57 @@ const WarningList: FC = () => {
     notify.failure("Loading warnings failed", error);
   }
 
+  const hasWarnings = isLoading || warnings.length > 0;
+
+  const filters: WarningFilters = {
+    queries: searchParams.getAll("query"),
+    statuses: searchParams.getAll("status") as LxdWarningStatus[],
+    severities: searchParams.getAll("severity") as LxdWarningSeverity[],
+    projects: searchParams.getAll("project"),
```
edlerd:
```suggestion
```

src/pages/warnings/WarningList.tsx
```diff
@@ -185,7 +239,7 @@ const WarningList: FC = () => {
             parentName="server"
             selectedNames={selectedNames}
             setSelectedNames={setSelectedNames}
-            filteredNames={rows.map((item) => item.name)}
+            filteredNames={filteredWarnings.map((item) => item.type)}
```
edlerd: Why is this changing from `item.name` to `item.type`? I think that should stay as before.

edlerd (commented):
Two more mentions of the project filter can be cleared and one question below.

src/pages/warnings/WarningList.tsx
Kxiru: Warnings do not have a name :). They do have a type though!

Kxiru: image

edlerd: This is a list of rows, not a list of warnings. So I think this change is breaking select all behavior from the table header. Should this be reverted?

Kxiru: Ah yes, I realise this didn't need to be done as I have already amended Rows to use the filteredWarnings. :+1:

src/pages/warnings/WarningList.tsx
```diff
@@ -163,7 +210,7 @@ const WarningList: FC = () => {
       <NotificationRow />
       <Row>
         <ScrollableTable
-          dependencies={[warnings]}
+          dependencies={[filteredWarnings]}
```
edlerd, line 213: We must use the `filteredWarnings` instead of `row` to construct the `filteredNames` parameter for this component. The undo is not correct and leaves wrong messaging on selecting warnings with active filters. My point was we should only change the variable, not the map logic applied to it.

edlerd (changes requested):

src/pages/warnings/WarningList.tsx
edlerd: Ah never mind, this is correct as is.

PR #1474, opened 2025-09-04
WD-25876 - Hide cluster member on instance detail and overview

kimanhou:
## Done

- Add useIsClustered in InstanceOverview and InstanceDetailPanelContent
- Hide "Cluster member" if instance is not clustered

## QA

1. Run the LXD-UI:

## Screenshots

image
image


src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -97,16 +100,18 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
...
+            <th className="u-text--muted">Cluster member</th>
+            <td>
+              {settings?.environment?.server_clustered ? (
+                <InstanceClusterMemberChip instance={instance} />
+              ) : (
+                "-"
+              )}
+            </td>
+          </tr>
+        )}
```
Kxiru: This works well, though now that this row is only rendered when in a clustered environment, we no longer need the ternary check in line 107, and can just display the chip as is.

src/pages/instances/InstanceOverview.tsx
```diff
@@ -82,21 +84,23 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
...
+                  <td>
+                    {settings?.environment?.server_clustered &&
+                    instance.location ? (
+                      <ResourceLink
+                        type="cluster-member"
+                        value={instance.location}
+                        to={`/ui/cluster/member/${encodeURIComponent(instance.location)}`}
+                      />
+                    ) : (
+                      "-"
```
Kxiru: Similar to the above but slightly different. In this case we can keep the ternary evaluator as it displays only when there is a specific node for the instance, but lets get rid of line 91 that checks the server clustered status as this is newly covered in your line 87.

Kxiru (changes requested):

edlerd (approved):
Code and QA LGTM :+1:

PR #1475, opened 2025-09-05
fix(instance) default disk size limit of vm is 10GiB and only containers are unlimited

### LINKED ISSUE #1464: Root storage size for a VM is "unlimited" (will not have a size limit) but it's limited to 10GiB
```
$ snap list microcloud lxd
Name        Version         Rev    Tracking     Publisher   Notes
lxd         5.21.3-c5ae129  33110  5.21/stable  canonical✓  in-cohort
microcloud  2.1.1-202e275   1781   2/stable     canonical✓  in-cohort
```

By following the tutorial such as:
https://documentation.ubuntu.com/microcloud/latest/microcloud/how-to/initialize/
We get remote storage (Ceph RBD) in the default profile.

When creating a new instance, the size of the volume is "unlimited". And the description of the config states `Size of root storage. If empty, root storage will not have a size limit.`. The root storage is actually limited to 10GiB if I'm not mistaken. I think it's just meant to be "unset" and "if unset, the default size of the root storage will be 10GiB" instead at least for block storage.

If the backend is `dir` and it's a container, it can truly mean "unlimited" (up to the whole size of the host filesystem) though.

Image

```
ubuntu@mc-1:~$ sudo rbd info --pool lxd_remote virtual-machine_my-test-vm.block
rbd image 'virtual-machine_my-test-vm.block':
        size 10 GiB in 2560 objects
        order 22 (4 MiB objects)
        snapshot_count: 0
        id: 3918a1f806b2
        block_name_prefix: rbd_data.3918a1f806b2
        format: 2
        features: layering, exclusive-lock, object-map, fast-diff, deep-flatten
        op_features: 
        flags: 
        create_timestamp: Mon Sep  1 06:33:22 2025
        access_timestamp: Mon Sep  1 08:29:01 2025
        modify_timestamp: Mon Sep  1 11:56:42 2025
        parent: lxd_remote/image_01cbaaaa4d04ae2433dc4402ab61a4d9638637888d4759397e4fb036aa98f85b_ext4.block@readonly
        overlap: 10 GiB
```

edlerd:
## Done

- fix(instance) default disk size limit of vm is 10GiB and only containers are unlimited

Fixes #1464

## QA

1. Run the LXD-UI:
    - create a container, ensure the disk size default is unlimited. Also take note of the help text for an override root disk.
    - create a vm, ensure the disk size default is mentioned as "10GiB". Also take note of the help text for an override root disk.

## Screenshots

Creating a VM

image

Creating a container

image

### DIFF
--- a/src/components/forms/DiskDeviceFormRoot.tsx
+++ b/src/components/forms/DiskDeviceFormRoot.tsx
@@ -31,6 +31,10 @@ const DiskDeviceFormRoot: FC<Props> = ({ formik, pools, profiles }) => {
   ] as LxdDiskDevice | null;
   const isEditingInstance =
     formik.values.entityType === "instance" && !formik.values.isCreating;
+  const isVirtualMachine =
+    formik.values.entityType === "instance" &&
+    formik.values.instanceType === "virtual-machine";
+  const defaultSize = isVirtualMachine ? "10GiB" : "unlimited";
 
   const [inheritValue, inheritSource] = getInheritedRootStorage(
     formik.values,
@@ -149,7 +153,7 @@ const DiskDeviceFormRoot: FC<Props> = ({ formik, pools, profiles }) => {
             id: "limits_disk",
             className: "override-with-form",
             inheritValue:
-              inheritValue?.size ?? (inheritValue ? "unlimited" : ""),
+              inheritValue?.size ?? (inheritValue ? defaultSize : ""),
             inheritSource,
             readOnly: readOnly,
             disabledReason: formik.values.editRestriction,
@@ -181,8 +185,8 @@ const DiskDeviceFormRoot: FC<Props> = ({ formik, pools, profiles }) => {
                   }
                 />
                 <p className="p-form-help-text">
-                  Size of root storage. If empty, root storage will not have a
-                  size limit.
+                  Size of root storage. If empty, root storage will{" "}
+                  {isVirtualMachine ? "be 10GiB." : "not have a size limit."}
                 </p>
               </>
             ),


src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -149,7 +153,7 @@ const DiskDeviceFormRoot: FC<Props> = ({ formik, pools, profiles }) => {
             id: "limits_disk",
             className: "override-with-form",
             inheritValue:
-              inheritValue?.size ?? (inheritValue ? "unlimited" : ""),
+              inheritValue?.size ?? (inheritValue ? defaultSize : ""),
```
Kxiru, line 156: Can this be simplified to `inheritValue?.size ?? defaultSize ?? ""`

src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -181,8 +185,8 @@ const DiskDeviceFormRoot: FC<Props> = ({ formik, pools, profiles }) => {
                   }
                 />
                 <p className="p-form-help-text">
-                  Size of root storage. If empty, root storage will not have a
-                  size limit.
+                  Size of root storage. If empty, root storage will{" "}
+                  {isVirtualMachine ? "be 10GiB." : "not have a size limit."}
```
Kxiru, line 189:
```suggestion
                  Size of root storage. If empty, root storage will be {" "} {defaultSize}
```

Wdyt?

Kxiru (changes requested):

src/components/forms/DiskDeviceFormRoot.tsx
edlerd: We have the 3rd case of no profile assigned defining any storage and then this should be empty. That's why we need the inner ternary with checking if any inheritValue is defined and fall back to empty.

src/components/forms/DiskDeviceFormRoot.tsx
edlerd: I think the suggested "root storage is unlimited" is a bit different from the current "root storage will not have a size limit." 

Maybe the current version is a bit more clear?

Kxiru: Hmm... I find them to be semantically the same, but I'm happy to go forward with either option. I thought it would be a good reuse of the variable we had already defined.

Kxiru (approved):
Thank you for adding this more fine-grained information!

PR #1482, opened 2025-09-09
feat(InstanceList, InstanceIps): sort IPv6 addresses in InstanceIps and display 2 IPv6 addresses in InstanceList (+n more if necessary)

kimanhou:
## Done

- in InstanceIps (used in InstanceOverview and InstanceDetailPanelContent), sort IPv6 addresses so that local addresses are at the end
- in InstanceList, display 2 addresses and "+ n more" if necessary

## QA

1. Run the LXD-UI:

## Screenshots

Screenshot from 2025-09-09 09-40-04
Screenshot from 2025-09-09 09-41-06
Screenshot from 2025-09-09 14-40-51
Screenshot from 2025-09-09 14-41-32

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -61,6 +61,7 @@ import type { LxdInstanceStatus } from "types/instance";
 import useSortTableData from "util/useSortTableData";
 import PageHeader from "components/PageHeader";
 import InstanceDetailPanel from "./InstanceDetailPanel";
+import InstanceListIps from "./InstanceListIps";
 import {
   mediumScreenBreakpoint,
   useIsScreenBelow,
@@ -492,7 +493,7 @@ const InstanceList: FC = () => {
           },
           {
             key: `ipv4-${ipv4.length}`,
-            content: ipv4.length > 1 ? `${ipv4.length} addresses` : ipv4,
+            content: <InstanceListIps ips={ipv4} />,
             role: "cell",
             className: "u-align--right clickable-cell",
             "aria-label": IPV4,
@@ -501,7 +502,7 @@ const InstanceList: FC = () => {
           },
           {
             key: `ipv6-${ipv6.length}`,
-            content: ipv6.length > 1 ? `${ipv6.length} addresses` : ipv6,
+            content: <InstanceListIps ips={ipv6} />,
             role: "cell",
             "aria-label": IPV6,
             onClick: openSummary,
--- a/src/pages/instances/InstanceListIps.tsx
+++ b/src/pages/instances/InstanceListIps.tsx
@@ -1,0 +1,23 @@
+import type { FC } from "react";
+
+interface Props {
+  ips: string[];
+}
+
+const InstanceListIps: FC<Props> = ({ ips }) => {
+  if (ips.length < 3) {
+    return ips.map((ip) => <div key={ip}>{ip}</div>);
+  }
+
+  return (
+    <>
+      <div>{ips[0]}</div>
+      <div>{ips[1]}</div>
+      <div className="p-text--x-small u-text--muted">
+        + {ips.length - 2} more
+      </div>
+    </>
+  );
+};
+
+export default InstanceListIps;
--- a/src/util/networks.tsx
+++ b/src/util/networks.tsx
@@ -1,4 +1,4 @@
-import type { LxdInstance } from "types/instance";
+import type { IpAddress, IpFamily, LxdInstance } from "types/instance";
 import type { LxdNetwork, LxdNetworkConfig } from "types/network";
 import type { LxdConfigOptionsKeys } from "types/config";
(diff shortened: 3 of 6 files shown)


src/pages/instances/InstanceList.tsx
```diff
@@ -501,7 +502,7 @@ const InstanceList: FC = () => {
           },
           {
             key: `ipv6-${ipv6.length}`,
-            content: ipv6.length > 1 ? `${ipv6.length} addresses` : ipv6,
+            content: <InstanceListIpv6Addresses ips={ipv6} />,
```
edlerd: Can we have the same change for ipv4 addresses?

src/util/networks.tsx
```diff
@@ -38,6 +38,40 @@ export const getIpAddresses = (
...
+};
+
+export const sortIpv6Addresses = (ipv6Addresses: IpAddress[]): IpAddress[] => {
+  if (ipv6Addresses.length === 0) return [];
+
+  // Set local addresses at the end
+  return ipv6Addresses.sort((a, b) => {
+    const isA_local = isLocalIPv6(a.address);
+    const isB_local = isLocalIPv6(b.address);
+    console.log(a.address, isA_local ? "true" : "false");
```
edlerd:
```suggestion
```

src/util/networks.tsx
```diff
@@ -38,6 +38,40 @@ export const getIpAddresses = (
...
+    normalizedAddress.startsWith("fe8") ||
+    normalizedAddress.startsWith("fe9") ||
+    normalizedAddress.startsWith("fea") ||
+    normalizedAddress.startsWith("feb")
+  ) {
+    const firstHextet = parseInt(normalizedAddress.split(":")[0], 16);
+    if (firstHextet >= 0xfe80 && firstHextet <= 0xfebf) {
+      return true;
+    }
+  }
```
edlerd:
```suggestion
    const firstHextet = parseInt(normalizedAddress.split(":")[0], 16);
    if (firstHextet >= 0xfe80 && firstHextet <= 0xfebf) {
      return true;
    }
```
Do we need the wrapping check? Or can we directly to the inner check? Seems a bit simpler.

src/pages/instances/InstanceListIpv6Addresses.tsx
```diff
@@ -0,0 +1,25 @@
+import type { FC } from "react";
+
+interface Props {
+  ips: string[];
+}
+
+const InstanceListIpv6Addresses: FC<Props> = ({ ips }) => {
```
edlerd: Can we make this agnostic of ipv4/6 so we can use this component for both? Maybe call it `InstanceListIps`?

src/pages/instances/InstanceListIpv6Addresses.tsx
```diff
@@ -0,0 +1,25 @@
...
+        {ip}
+      </p>
+    ));
+  }
+
+  return (
+    <>
+      <p className="u-no-margin--bottom">{ips[0]}</p>
+      <p className="u-no-margin--bottom">{ips[1]}</p>
+      <p className="p-text--x-small u-text--muted">+ {ips.length - 2} more</p>
```
edlerd: like this approach, it works realy nicely.

src/pages/instances/InstanceListIpv6Addresses.tsx
```diff
@@ -0,0 +1,25 @@
+import type { FC } from "react";
+
+interface Props {
+  ips: string[];
+}
+
+const InstanceListIpv6Addresses: FC<Props> = ({ ips }) => {
+  if (ips.length < 3) {
+    return ips.map((ip) => (
+      <p className="u-no-margin--bottom" key={ip}>
```
edlerd:
```suggestion
      <div className="u-no-margin--bottom" key={ip}>
```
I think the paragraph adds some spacing that messes up alignment. Maybe we can use div instead? Also in the multi case below.

image

edlerd (commented):
Good start, some obeservations below.

src/pages/instances/InstanceList.tsx
kimanhou: Yes !

src/pages/instances/InstanceListIps.tsx
```diff
@@ -0,0 +1,27 @@
+import type { FC } from "react";
+
+interface Props {
+  ips: string[];
+}
+
+const InstanceListIps: FC<Props> = ({ ips }) => {
+  if (ips.length < 3) {
+    return ips.map((ip) => (
+      <div className="u-no-margin--bottom" key={ip}>
```
edlerd:
```suggestion
      <div key={ip}>
```
I think now with the `div` tag we don't need this here and below either.

edlerd (approved):
QA and code LGTM, thanks for the iterations!

PR #1487, opened 2025-09-10
feat: [WD-17985] Add HPE Storage Driver

Kxiru:
## Done

- Adds HPE Driver integration to the UI.
- Create Operations :+1: 
- Read Operations :+1: 
- Update Operations :+1: 
- Delete Operations :+1: 

## QA

1. Run the LXD-UI:
    - Within a HPE enabled backend, attempt CRUD Operations using a HPE Driver.
     
## Screenshots

image
image
image

### DIFF
--- a/src/pages/storage/forms/StoragePoolFormAlletra.tsx
+++ b/src/pages/storage/forms/StoragePoolFormAlletra.tsx
@@ -1,0 +1,43 @@
+import type { FormikProps } from "formik";
+import type { FC } from "react";
+import type { StoragePoolFormValues } from "./StoragePoolForm";
+import { getConfigurationRow } from "components/ConfigurationRow";
+import { Input, Select } from "@canonical/react-components";
+import { optionIscsiNvme, optionTrueFalse } from "util/instanceOptions";
+import ScrollableConfigurationTable from "components/forms/ScrollableConfigurationTable";
+
+interface Props {
+  formik: FormikProps<StoragePoolFormValues>;
+}
+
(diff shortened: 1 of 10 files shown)


src/pages/storage/forms/StoragePoolFormAlletra.tsx
edlerd, line 1: I see some additional fields: `rsync.bwlimit`, `rsync.compression` and `volume.size`. Should those be visible in the UI as well?

src/util/storagePool.tsx
```diff
@@ -103,6 +117,7 @@ const storagePoolDriverToOptionKey: Record<string, LxdConfigOptionsKeys> = {
   powerflex: "storage-powerflex",
   pure: "storage-pure",
   cephobject: "storage-cephobject",
+  alletra: "storage-alletra",
```
edlerd, line 120: I think we should use the constants here instead. Maybe you can update the other ones as well to use the referenced constant here? Similar to the `storageDriverLabels` definition.
```suggestion
  [alletraDriver]: "storage-alletra",
```

src/pages/storage/forms/StoragePoolFormAlletra.tsx
Kxiru: To my knowledge, these are not Alletra fields, and the only alletra fields are the 7 included.

edlerd: I found those three in the config API mentioned under the alletra driver. Maybe check with Aleksandr if we should surface those settings in the UI as well?

PR #1500, opened 2025-09-15
feat(project) add restricted.backups input for project configuration

edlerd:
## Done

- feat(project) add restricted.backups input for project configuration

## QA

1. Run the LXD-UI:
    - go to project > configuration > enable "Allow custom restrictions on a project level" > restrictions > instances > Backup restriction and set it to allow or block.

## Screenshots

image

### DIFF
--- a/src/pages/instances/actions/ExportInstanceBtn.tsx
+++ b/src/pages/instances/actions/ExportInstanceBtn.tsx
@@ -4,6 +4,8 @@ import { Button, Icon, usePortal } from "@canonical/react-components";
 import classNames from "classnames";
 import { useInstanceEntitlements } from "util/entitlements/instances";
 import ExportInstanceModal from "pages/instances/forms/ExportInstanceModal";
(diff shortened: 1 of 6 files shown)


edlerd:
> The wording felt a little trippy to me, like having double negatives mean a positive. I had to tell myself, "If I BLOCK RESTRICTIONS on backups, then I will be able to allow backups" to really get it, but perhaps that is to be expected?

Good point, the label should better be "Backup creation". Then the allow and block values make more sense and become easier to parse for our human brains.

edlerd:
Also fair point about the export buttons on instances and volumes. Disabling them now, so users won't run into the error.

edlerd:
> On the demo server, I am able to create backup volumes just fine, which was also the case when I attempted previously:

This only relates to a project with active restrictions and the "backup creation" set to block. Testing it seems correct to me. Though none of the projects on demo had this setting before, only after adjusting it for pr2 there I could see it.

PR #1503, opened 2025-09-16
feat(instance) loading status for instances consistency update

edlerd:
## Done

- feat(instance) loading status for instances consistency update
- Instance loading triggers from cli or other updates to instance state outside the currently open ui tab.
- Allow resubmit of stop instance action for triggering force-stop while an instance is stopping.
- Bootstrap instance loading states on initial load based on running operations

Fixes WD-26638

## QA

1. Run the LXD-UI:
    - Open UI and stop/start also bulk stop/start instance, ensure the loading state is consistent
    - Stop a VM and then force stop it to stop it faster. Ensure loading state is consistent.
    - Open UI and start / stop an instance from CLI. Ensure the loading status for that instance updates in the UI without giving it focus.

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -32,9 +32,9 @@ const Root: FC = () => {
         <QueryClientProvider client={queryClient}>
           <AuthProvider>
             <ProjectProvider>
-              <InstanceLoadingProvider>
-                <MemberLoadingProvider>
-                  <OperationsProvider>
+              <OperationsProvider>
+                <InstanceLoadingProvider>
+                  <MemberLoadingProvider>
                     <EventQueueProvider>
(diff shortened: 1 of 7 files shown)


Kxiru (approved):
Not sure what exactly to test/provide evidence of, but everything stops/starts withi a timeley fashion. Here is an example of me stopping 3 instances. The Start button reappears before  all instances have stopped, but I wonder whether that is to be expected:
image
image
image

Force stopping behaves similarly but resolves faster.

Starting from the CLI was instantaneous
image

Stopping displayed the loading bar after a few seconds:
image

I'm not sure if this correlates with the expected behaviour. Approving to remove it as a blocker but do iterate on it if it needs fine-tuning.

PR #1505, opened 2025-09-16
fix(cluster) permission checks for cluster groups and members

edlerd:
## Done

- fix(cluster) permission checks for cluster groups and members
- disable buttons if user has no permissions

## QA

1. Run the LXD-UI:
    - use a user with limited permissions, as in only having project viewer on the default project
    - go to clustering > members and clustering > groups and ensure the interactive elements are disabled (like edit and delete or evacuate)

### DIFF
--- a/src/components/SelectableMainTable.tsx
+++ b/src/components/SelectableMainTable.tsx
@@ -194,7 +194,7 @@ const SelectableMainTable: FC<Props> = ({
             indeterminate={isRowIndeterminate && !isRowSelected}
           />
         ),
-        role: "rowheader",
+        role: "cell",
         className: "select",
       },
       ...(row.columns ?? []),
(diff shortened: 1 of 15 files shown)


Kxiru (dismissed):
LGTM! It's fun to also identify areas where permission checks are missing. Thank you for adding these!

This doesn't quite work on my machine, but I suppose that as long as it works in CI, that's what matters. I tested this under chromium:lxd-latest-edge:clustered.
image

edlerd:
> LGTM! It's fun to also identify areas where permission checks are missing. Thank you for adding these!
> 
> This doesn't quite work on my machine, but I suppose that as long as it works in CI, that's what matters. I tested this under chromium:lxd-latest-edge:clustered. image

This happened because our `getFirstClusterMember` is not waiting for the switch to the cluster member page to finish, then picking content from the instance list instead of a cluster member. Also, there is a second bug here, with selectable table marking the column with the checkbox as rowheader instead of cell. Fixed both issues with recent push.

edlerd:
Test should be more stable now.

PR #1506, opened 2025-09-17
fix(instance) column responsiveness for a non-clustered environment for the instance list should not consider cluster member column

edlerd:
## Done

- fix(instance) column responsiveness for a non-clustered environment for the instance list should not consider cluster member column

## QA

1. Run the LXD-UI:
    - open instance list on a non-clustered backend. Resize the window width and ensure the columns get hidden correctly.
    - test instance list with side panel open and resizing window from mobile to desktop and back.

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -126,6 +126,13 @@ const InstanceList: FC = () => {
     return <>Missing project</>;
   }
 
+  useEffect(() => {
+    if (isClustered || !userHidden.includes(CLUSTER_MEMBER)) {
+      return;
+    }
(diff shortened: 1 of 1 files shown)


Kxiru (approved):
Columns looks appropriate to me!
image

Just double checking, is IPv4 supposed to be right aligned as opposed to left-aligned?

edlerd:
> Just double checking, is IPv4 supposed to be right aligned as opposed to left-aligned?

Yes it is. The default view is ipv4 and next to it ipv6 and booth look best with ipv4 is right aligned to the left of it -- making it look a bit nicer. Also numbers are best right aligned to make them comparable. For ipv4 we expect them to have same prefixes in simple setups with single ip addresses, where right alignment makes them easy to compare.

Issue #1334, opened 2025-06-26
Feature Request: Assign ACLs to NICs


vosdev:
Hey, currently the UI does not allow us to assign Network ACLs to NICs (Instances or Profiles) but only directly to networks.

vosdev:
I noticed that I forgot to mention that the default actions cannot be modified through the UI

https://documentation.ubuntu.com/lxd/latest/howto/network_acls/#configure-default-actions

Is that something you want to support through the UI as well or is it too advanced?

Perhaps a warning could be shown that if an ACL is applied that the default ingress and egress are set to reject by default.

edlerd:
> I noticed that I forgot to mention that the default actions cannot be modified through the UI
> 
> https://documentation.ubuntu.com/lxd/latest/howto/network_acls/#configure-default-actions
> 
> Is that something you want to support through the UI as well or is it too advanced?

A fine question!

In my opinion, the default to reject unmatched traffic is quite reasonable. I am not sure how common it is to change this behavior. As the CLI allows it, we might add it to the UI as well. At least on the network level. For the NIC device level, I am not so sure.

Do you want to customize and use this option? Which reasons do you see to change the default?

vosdev:
> > I noticed that I forgot to mention that the default actions cannot be modified through the UI
> > https://documentation.ubuntu.com/lxd/latest/howto/network_acls/#configure-default-actions
> > Is that something you want to support through the UI as well or is it too advanced?
> 
> A fine question!
> 
> In my opinion, the default to reject unmatched traffic is quite reasonable. I am not sure how common it is to change this behavior. As the CLI allows it, we might add it to the UI as well. At least on the network level. For the NIC device level, I am not so sure.
> 
> Do you want to customize and use this option? Which reasons do you see to change the default?

>  In my opinion, the default to reject unmatched traffic is quite reasonable.

Agreed. It's just that I didn't know about default reject all ingress _and_ egress when I added a single ACL for ingress SSH. I spent some time debugging why my traffic wasn't working. It's hidden at the bottom of the documentation.

I ended up solving this with an ACL that simply allows all egress. Which is a lot easier and friendlier to do than diverge from the default configuration.

> At least on the network level. For the NIC device level, I am not so sure.

If you're adding support for this option to the network level then that would probably be a clean solution. I don't expect many people wanting to change this default behavior on NICs and those who do can always revert back to the yaml. Less clutter in the UI :)


[...]

PR #1513, opened 2025-09-19
WD-26012: Add ACLs in Instance/Profile Overview and Configuration

kimanhou:
## Done

- In Instance/Profile Overview > Networks section: display ACLs count. This count includes ACLs attached to the network + ACLs manually attached to the device. If the same ACL is attached to the network and to the device, it will be counted once (duplicates are removed). The count is a link, going to Instance/Profile > Configuration > Devices > Network.
- In Instance/Profile Configuration > Devices > Network: 
    - in read only mode: display ACLs (ACLs from network and manual ACLs are displayed in the same list) as a list of chips.
    - in edit mode: display ACLs attached to the network and display ACL selector when selected network is OVN
    - inherited networks appear as chip (and not bold name anymore)

Fixes [#1334](https://github.com/canonical/lxd-ui/issues/1334)

## QA

1. Run the LXD-UI:

## Screenshots

Overview
Screenshot from 2025-09-19 17-37-26

Configuration - read only
image

Configuration - edit mode
Network selector improved with Name, Type and ACLs (count)
image

ACLs attached to network are automatically selected and disabled
image

ACLs selector disabled if selected network is not type = ovn
image

ACLs selector displays a helper text underneath itself
If network is not OVN: `Network must be of type OVN to customize ACLs.` (see screenshot above)
If network is OVN with no ACL attached to it: no helper text
image

If network is OVN with ACLs attached to it `Some ACLs are inherited from the network. They cannot be deselected here.`
image


src/components/NetworkListTable.tsx
```diff
@@ -45,7 +52,39 @@ const NetworkListTable: FC<Props> = ({ onFailure, devices, instance }) => {
...
+        return null;
+      }
+
+      return network;
+    })
+    .filter((network) => network !== null);
+
+  const shouldDisplayAclsColumn = lxdNetworks.some(
+    (t) => t.config["security.acls"],
+  );
```
edlerd: We should check if any network device has an acl or any network attached to the instance has an acl here. Currently checking  if any network has an acl, that could also be one that is not attached. Or we could have no network with acls, but one instance device having an acl.

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,111 @@
...
+          </div>
+        </div>
+      )}
+      {(deviceAcls.length > 0 ||
+        (!readOnly &&
+          formik &&
+          index !== undefined &&
+          canSelectManualAcls)) && (
+        <div>
+          Manual ACLs
```
edlerd: In edit mode, this should be a label for the input below.

src/components/NetworkListTable.tsx
```diff
@@ -45,7 +52,39 @@ const NetworkListTable: FC<Props> = ({ onFailure, devices, instance }) => {
...
+
+  if (shouldDisplayAclsColumn) {
+    networksHeaders.push({
+      content: "ACLs",
+      sortKey: "acls",
+      className: "u-text--muted u-hide--small u-hide--medium",
+    });
+  }
```
edlerd: Adding this column leaves not enough space for the mac address to be fully visible. I think we can reduce the width of the Interface column slightly to make enough space for it.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,127 @@
...
+  useEffect(() => {
+    setIsAclsDisplayed(shouldDisplayAcls());
+  }, [(formik.values.devices[index] as LxdNicDevice).network]);
+
+  return (
+    <div className="network-device" key={index}>
+      <div>
+        {readOnly ? (
+          <div>
+            {(formik.values.devices[index] as LxdNicDevice).network}
```
edlerd: I think in read mode, this can be an entity chip linking to the network.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,127 @@
...
+            <Acls
+              project={project}
+              managedNetworks={managedNetworks}
+              device={device}
+              readOnly
+            />
```
edlerd: I think in read mode this could be only one heading ACLs with a combined list of chips. We might not need the source information in read mode.

src/components/forms/NetworkDevicesForm/Acls.tsx
edlerd: This should be hidden when going from a network with acl support and manual acls selected to a network without acl support. See screencast below


https://github.com/user-attachments/assets/55e1b2fe-11e0-4f0a-8991-9e7711df779b

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,111 @@
...
+  project: string;
+  managedNetworks: LxdNetwork[];
+  device: LxdNicDevice | CustomNetworkDevice;
+  readOnly?: boolean;
+  formik?: InstanceAndProfileFormikProps;
+  index?: number;
+  canSelectManualAcls?: boolean;
+}
+
+const Acls: FC<Props> = ({
```
edlerd: I think the spacing in edit mode can be improved, maybe get input from @piperdeck 

image

edlerd (commented):
Good progress, QA seems fully functional from testing. Some observations on presentation, see comments below.

src/components/NetworkListTable.tsx
kimanhou: Good catch ! Fixed

src/components/forms/NetworkDevicesForm/Acls.tsx
kimanhou: Yes, I thought of it but MultiSelect doesn't have label as argument. I can add the label manually but MultiSelect doesn't have id as argument either so whatever I put in for in label will not match the button's id. Should I update react components to add label and id in MultiSelect ? Or use a label without for ?

And then what should "Manual ACLs" be in read only mode ? A span ?

src/components/NetworkListTable.tsx
kimanhou: In this table, only the first column (Name) has a set width of 30%. I don't want to change that.
The other columns are calculated automatically, so I set the width of the interface column to 6rem (I prefer using rem to %) and the width of ACLs column to 3rem.
image

It looks like network_detail_list.scss file is not useful so I deleted it.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
kimanhou: Okay, this would be an improvement from the original version (I just moved the code to its own component, it was a string already before).
It's a good improvement, done !

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
kimanhou: Oh okay. Done

src/components/forms/NetworkDevicesForm/Acls.tsx
kimanhou: Fixed with above comment

src/components/forms/NetworkDevicesForm/Acls.tsx
kimanhou: @piperdeck Please help 🙏

src/components/forms/NetworkDevicesForm/Acls.tsx
edlerd: > MultiSelect doesn't have label as argument

Good point, this is an issue also in the network config page, where the label for ACLs is missing the connection. Yes, I think having a way to set the `id` of the `MultiSelect` is a good improvement and would unblock the fix here and in the network config.

> And then what should "Manual ACLs" be in read only mode ? A span ?

Yes, a span in read mode. For read mode I was thinking we just present one heading and listing all ACLs, so not communicating the split between "from network" and "from nic device" as a source -- to simplify read mode.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -110,14 +112,31 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
               </div>
             ),
             override: (
-              <Tooltip
-                message="This network is inherited from a profile or project.
+              <>
+                <Tooltip
+                  message="This network is inherited from a profile or project.
```
edlerd: I think we should change the inherited device as well to use the chip for the network reference instead of just a text label in bold.
Also, we should show the ACLs for the networks inherited from a profile. For those we can have an acl again on the network or in the profile nic device.

edlerd: Example of an instance with inherited network devices:

image

The profile applied in the screen above also defines ACLs:

image

kimanhou: Sure !

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -101,23 +117,45 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
             inherited: (
               <div>
-                <div className="mono-font">
-                  <b>{item.network?.network}</b>
-                </div>
+                <ResourceLink
+                  type="network"
+                  value={item.network?.network || ""}
+                  to={`/ui/project/${encodeURIComponent(project ?? "")}/network/${encodeURIComponent(item.network?.network || "")}`}
+                />
```
edlerd, line 114: We should display the acls for inherited networks below the network (in the inherited column).
The override column is for overriding the device, which is currently not possible, but we will enable that soon with a follow-up task.

See screen below, the first two networks are inherited and the ACLs should show up below them.

Another thought: Should display just one line for acls in read mode? So it becomes "ACLs: chip1 chip2 ...", removing the line break. I think that would make the read mode a bit cleaner.

image

kimanhou: Sorry, I misunderstood. Fixed now.

I put the ACLs chips on the same line as "ACLs" header in read only but left it as it is in edit mode because it looked weird.

src/components/forms/NetworkDevicesForm/Acls.tsx
kimanhou: > having a way to set the id of the MultiSelect is a good improvement

I had a look in react-components and it seems difficult to implement. MultiSelect uses ContextualMenu, which creates its own id with useId.
An easy fix would be to wrap the MultiSelect in a span with an id, see below picture ⬇️

image

edlerd: > I had a look in react-components and it seems difficult to implement. MultiSelect uses ContextualMenu, which creates its own id with useId.

The wrapper is an option, I think we would then highlight the wrong element when clicking on the label.

I think best is if we allow passing in an ID to the `MulstiSelect`. The ID gets set to the target of the contextual menu. So that is either the `button` or the `SearchBox`, depending on the active variant. If the ID prop is empty, we can still call `useId` to generate one as `inputId` to remain backwards compatible.

Wdyt?

kimanhou: Okay okay let's do that.

https://github.com/canonical/react-components/pull/1256

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,132 @@
+import { useEffect, useState, type FC } from "react";
+import type { LxdNicDevice } from "types/device";
+import type { LxdNetwork } from "types/network";
+import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import type { CustomNetworkDevice } from "util/formDevices";
+import NetworkSelector from "pages/projects/forms/NetworkSelector";
```
edlerd, line 5: The changes to the `NetworkSelector` are not visible in this PR yet.

src/components/NetworkListTable.tsx
```diff
@@ -45,7 +52,39 @@ const NetworkListTable: FC<Props> = ({ onFailure, devices, instance }) => {
...
+        return null;
+      }
+
+      return network;
+    })
+    .filter((network) => network !== null);
+
+  const shouldDisplayAclsColumn =
+    lxdNetworks.some((t) => t.config["security.acls"]) ||
+    networkDevices.some((t) => t["security.acls"]);
```
edlerd: I am having second thoughts on this. Usually, it is best to either always show a UI component or never show it, but avoid the specific configurations from having an impact on the displayed elements. The reason is to improve the learnability, as if a column is always there, the user understands it easily and that supports building of the mental model of how the software works. So maybe we should always show the ACLs column here. Wdyt?

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -22,38 +22,89 @@ const NetworkSelector: FC<Props & SelectProps> = ({
   const { data: networks = [] } = useNetworks(project);
 
   const managedNetworks = networks.filter((network) => network.managed);
+  const hasAcls = managedNetworks.some((t) => t.config["security.acls"]);
 
   const getNetworkOptions = () => {
     const options = managedNetworks.map((network) => {
       return {
-        label: network.name,
+        label: (
```
kimanhou, line 29: @edlerd Are you talking about this change ?

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,127 @@
...
+      if (networkAcls.length === 2) {
+        return `${networkAcls[0]} and ${networkAcls[1]} are inherited from the network. They cannot be deselected here.`;
+      }
+      if (networkAcls.length === 3) {
+        return `${networkAcls[0]}, ${networkAcls[1]} and ${networkAcls[2]} are inherited from the network. They cannot be deselected here.`;
+      }
+      return `${networkAcls[0]}, ${networkAcls[1]} and ${networkAcls.length - 2} more are inherited from the network. They cannot be deselected here.`;
+    }
+
+    return "No ACLs can be added on a network that is not of type OVN.";
```
edlerd:
```suggestion
    return "Network must be of type OVN to customize ACLs.";
```
Better use a positive formulation.

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,127 @@
...
+      if (networkAcls.length === 1) {
+        return `${networkAcls[0]} is inherited from the network. It cannot be deselected here.`;
+      }
+      if (networkAcls.length === 2) {
+        return `${networkAcls[0]} and ${networkAcls[1]} are inherited from the network. They cannot be deselected here.`;
+      }
+      if (networkAcls.length === 3) {
+        return `${networkAcls[0]}, ${networkAcls[1]} and ${networkAcls[2]} are inherited from the network. They cannot be deselected here.`;
+      }
+      return `${networkAcls[0]}, ${networkAcls[1]} and ${networkAcls.length - 2} more are inherited from the network. They cannot be deselected here.`;
```
edlerd: We probably shouldn't start with the ACL name. It looks wired starting with a lowercase character. Maybe start with "The ACL" or "The ACLs" prefix on the current text?

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,127 @@
...
+  device: LxdNicDevice | CustomNetworkDevice;
+  readOnly?: boolean;
+  formik?: InstanceAndProfileFormikProps;
+  index?: number;
+  canSelectManualAcls?: boolean;
+}
+
+const Acls: FC<Props> = ({
+  project,
+  managedNetworks,
```
edlerd: I think we should pass in only the relevant network to avoid the filter logic in here. This component should not be concerned about other networks.

package.json
```diff
@@ -31,7 +31,7 @@
     "test-report-coverage": "cp coverage/unit/coverage-final.json coverage/playwright/ ; nyc report --reporter html --reporter cobertura --reporter text-summary --temp-dir coverage/playwright --report-dir coverage/playwright-report --exclude 'src/lib/**' --exclude 'src/types/**'"
   },
   "dependencies": {
-    "@canonical/react-components": "3.0.1",
+    "@canonical/react-components": "3.2.0",
```
edlerd: Please rebase on main, then this change should be pulled in automatically.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -167,66 +205,16 @@ or remove the originating item"
...
+                <NetworkDevice
+                  index={index}
+                  readOnly={readOnly}
+                  formik={formik}
+                  project={project}
+                  focusNetwork={focusNetwork}
+                  removeNetwork={removeNetwork}
+                  managedNetworks={managedNetworks}
+                  device={device}
+                />
```
edlerd, line 195: We should have a bit more spacing below the content of the override contents

Instead of this:

image


We should have some spacing below, maybe like this:

image

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -101,23 +117,45 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
             ),
             inherited: (
               <div>
-                <div className="mono-font">
-                  <b>{item.network?.network}</b>
-                </div>
+                <ResourceLink
+                  type="network"
```
edlerd, line 111: We should also have the "Network" heading for inherited devices. And have the ACLs header alike to the override read mode.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -101,23 +117,45 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
+                              type="network-acl"
+                              value={acl}
+                              to={`/ui/project/${encodeURIComponent(project || "default")}/network-acl/${encodeURIComponent(acl)}`}
+                            />
+                          </div>
+                        ))}
+                      />
+                    </div>
+                  </div>
+                )}
```
edlerd: The qualifier from which profile this is inherited should not be between the network and acl. Maybe the qualifier should be below the ACLs or on top of the network.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+
+const NetworkDevice: FC<Props> = ({
+  index,
+  readOnly,
+  project,
+  formik,
+  focusNetwork,
+  removeNetwork,
+  managedNetworks,
```
edlerd: I think here we should only receive the `managedNetwork` instead of the list of all networks. So upstream can do the filtering.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,133 @@
...
+    }
+    return false;
+  };
+
+  const [isAclsDisplayed, setIsAclsDisplayed] =
+    useState<boolean>(shouldDisplayAcls());
+
+  useEffect(() => {
+    setIsAclsDisplayed(shouldDisplayAcls());
+  }, [(formik.values.devices[index] as LxdNicDevice).network]);
```
edlerd: This can be calculated on the fly, no need for keeping a state for this with a use effect.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,133 @@
...
+  index,
+  readOnly,
+  project,
+  formik,
+  focusNetwork,
+  removeNetwork,
+  managedNetworks,
+  device,
+}) => {
+  const shouldDisplayAcls = () => {
```
edlerd: Please align the naming here. I think we iterated away from hiding ACLs.

Maybe with the other changes we can directly pass `canSelectManualAcls={managedNetwork.type === "ovn"}` below where this is used as a prop for the `Acls` component

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -93,6 +95,20 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
+          const parentProfile = profiles.find(
+            (t) => t.name === item.source.split(" ")[0],
+          );
+          const profileDevice = parentProfile?.devices[
+            item.key
+          ] as LxdNicDevice;
+          const deviceAcls = profileDevice
+            ? profileDevice["security.acls"]?.split(",") || []
+            : [];
+          const allAcls = Array.from(new Set(networkAcls.concat(deviceAcls)));
```
edlerd: There is a mixup here:
- `item.network["security.acls"]` are the device acls
- the network acls are missing, we need to pick them out from the matching lxd network. The network name is `item.network.network` and we can match it to the lxd network list to get the network configuration. The network configuration can have acls.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -101,23 +117,45 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
+                        items={allAcls.map((acl) => (
+                          <div key={acl} className="u-whitespace-nowrap">
+                            <ResourceLink
+                              type="network-acl"
+                              value={acl}
+                              to={`/ui/project/${encodeURIComponent(project || "default")}/network-acl/${encodeURIComponent(acl)}`}
+                            />
+                          </div>
+                        ))}
+                      />
```
edlerd: A long list of ACLs is breaking the layout here

image

with expansion:

image

src/pages/networks/forms/NetworkAclSelector.tsx
```diff
@@ -28,15 +36,23 @@ const NetworkAclSelector: FC<Props> = ({
...
-      disabled={!hasAcls}
+      items={toOptionList(
+        availableAcls.map((acl) => acl.name),
+        disabledItems,
+      )}
+      disabled={!hasAcls || disabled}
       selectedItems={toOptionList(selectedAcls)}
       variant={hasAcls ? "condensed" : "search"}
-      placeholder={hasAcls ? "Select ACLs" : "No ACLs available"}
+      placeholder={hasAcls && !disabled ? "Select ACLs" : "No ACLs available"}
```
edlerd: I think we need a bit more granularity here. The copy is correct if we have no acls available, but wrong if we have acls and this is disabled.

Instead of a `disabled` prop name, I'd suggest keeping the more specific `canSelectManualAcls` name from upstream. Then we can adjust the copy here for the `canSelectManualAcls=false` case to be something better fitting. Because "No ACLs available" seems to imply there are no ACLs defined, which can be wrong in that case.

src/pages/projects/forms/NetworkSelector.tsx
edlerd: Yes, it is all here now, thanks :)

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -22,38 +22,89 @@ const NetworkSelector: FC<Props & SelectProps> = ({
...
+                {network.config["security.acls"]?.split(",").length}
+              </span>
+            ) : null}
+          </div>
+        ),
         value: network.name,
+        text: (
+          <span>
+            {network.name}&nbsp;
+            <span className="u-text--muted">&#40;{network.type}&#41;</span>
```
edlerd, line 51: this looks very nice with the muted text on the addition.

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -22,38 +22,89 @@ const NetworkSelector: FC<Props & SelectProps> = ({
...
+            </span>
+            <span title={network.type} className="network-option u-truncate">
+              {network.type}
+            </span>
+            {hasAcls ? (
+              <span
+                title={network.type}
+                className="network-option acls u-truncate u-align--right"
+              >
+                {network.config["security.acls"]?.split(",").length}
```
edlerd: We should have a fallback here for when no ACLs are on the network. The fallback should be either "0" or a "-".

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -22,38 +22,89 @@ const NetworkSelector: FC<Props & SelectProps> = ({
   const { data: networks = [] } = useNetworks(project);
 
   const managedNetworks = networks.filter((network) => network.managed);
+  const hasAcls = managedNetworks.some((t) => t.config["security.acls"]);
```
edlerd:
```suggestion
```
I think we should always have the ACLs visible in here, thus can remove the const. See https://github.com/canonical/lxd-ui/pull/1513#discussion_r2387678174 for the reasoning.

src/pages/networks/forms/NetworkAclSelector.tsx
kimanhou: What do you suggest in case `canSelectManualAcls=false` ?
I've tried two options but it feels weird with the help text underneath. We should either combine both, or not display the dropdown.

Screenshot from 2025-09-30 14-26-10
Screenshot from 2025-09-30 14-25-36

edlerd: We could use a dash "-" or write "None" into it if the ACLs are empty and can't be applied. Wdyt?

edlerd: I think if ACLs are applied to the network we'd not see the placeholder, so it would be correct to write "None" here, right?

src/components/NetworkListTable.tsx
kimanhou: I would prefer not to display the ACLs column if not needed so that people who don't use ACLs don't see any change. 
But I don't have a strong opinion on this, I'm happy either way.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
kimanhou: I tried to maximise the width of `acl-list` div. But I had to put in a magic number, that works well with my screen. I also maximised the width of each ACL chip in case of a very long name.
I am welcome to try another solution if you have any suggestion.

src/pages/networks/forms/NetworkAclSelector.tsx
kimanhou: I like "-", it's a sleek design
image

src/pages/projects/forms/NetworkSelector.tsx
kimanhou: However it throws an error, since text is supposed to be a `string`. 

`// text used for search, sort and display in toggle button
  // text must be provided if label is not a string`

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,129 @@
...
+    return [];
+  };
+
+  const getDeviceAcls = (device: LxdNicDevice | CustomNetworkDevice) => {
+    return (device as LxdNicDevice)["security.acls"]?.split(",") || [];
+  };
+
+  const networkAcls = getNetworkAcls(network);
+  const deviceAcls = getDeviceAcls(device);
+  const allAcls = Array.from(new Set(networkAcls.concat(deviceAcls)));
```
edlerd: Some rendering bug is observed here when removing the ACLs. I think this happens when `security.acls` has a value of empty string. Then we end up with an array with one empty string element from the split call.


https://github.com/user-attachments/assets/05d51d05-3008-4a17-b758-5a8c1f2d5d19

kimanhou: I saw it and I fixed it... in only one place :woman-facepalming: 
I'll create 2 functions to get ACLs (one from network and one from device) and reuse them everywhere.

src/components/NetworkListTable.tsx
edlerd: I opened the comment because I was confused on testing this PR and not seeing the ACL column due to this logic. I am leaning to always show it to avoid surprising the user.

src/components/NetworkListTable.tsx
```diff
@@ -61,43 +100,77 @@ const NetworkListTable: FC<Props> = ({ onFailure, devices, instance }) => {
         return null;
       }
 
+      const device = networkDevices.find((t) => t.network === network.name);
+      const deviceAcls =
+        device && device["security.acls"] ? device["security.acls"] : "";
+      const networkAcls = network.config["security.acls"] ?? "";
```
edlerd: We can use the helper function here as well to align and simplify below.

src/components/NetworkListTable.tsx
```diff
@@ -61,43 +100,77 @@ const NetworkListTable: FC<Props> = ({ onFailure, devices, instance }) => {
...
+        },
+      ];
+
+      const baseUrl = `/ui/project/${encodeURIComponent(project ?? "")}/`;
+      const urlSuffix = "/configuration/network";
+      const linkToConfiguration = instance
+        ? `${baseUrl}instance/${encodeURIComponent(instance.name)}${urlSuffix}`
+        : profileName
+          ? `${baseUrl}profile/${encodeURIComponent(profileName)}${urlSuffix}`
+          : "";
```
edlerd: This can be a relative link to `configuration/network`, so we avoid the complex link building.

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,118 @@
...
+                  type="network-acl"
+                  value={acl}
+                  to={`/ui/project/${encodeURIComponent(project || "default")}/network-acl/${encodeURIComponent(acl)}`}
+                  className="acl-chip"
+                />
+              </div>
+            ))}
+          />
+        </div>
+      </div>
```
edlerd: I think this block is duplicated in `NetworkDevicesForm`. Can we extract a component for reuse and avoid duplication?

src/sass/_forms.scss
```diff
@@ -62,6 +62,21 @@
...
+      margin-top: $spv--small;
+
+      > .acls-list {
+        display: flex;
+        flex-wrap: wrap;
+        max-width: 15rem;
+
+        .acl-chip {
+          max-width: 12rem;
+        }
```
edlerd: To avoid the magic number and to ease the layout a bit, how about giving the acl chips a line, so they appear underneath each other instead of lined up next until they wrap? That would take a bit more space vertically, but might be easier to read and ease the custom css and magic numbers here.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,114 @@
...
+            />
+          </div>
+        ) : (
+          <>
+            <NetworkSelector
+              value={(formik.values.devices[index] as LxdNicDevice).network}
+              project={project}
+              onBlur={formik.handleBlur}
+              setValue={(value) =>
+                void formik.setFieldValue(`devices.${index}.network`, value)
```
edlerd: On set value here we should clear the acls if any were selected if the network does not support custom acls.

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
```diff
@@ -0,0 +1,118 @@
...
+      </div>
+    );
+  }
+
+  const id = useId();
+
+  const userSelectedAcls =
+    formik && index !== undefined && canSelectManualAcls
+      ? getDeviceAcls(formik.values.devices[index] as LxdNicDevice)
+      : [];
```
edlerd:
```suggestion
  const userSelectedAcls =
    formik && index !== undefined
      ? getDeviceAcls(formik.values.devices[index] as LxdNicDevice)
      : [];
```
I don't think we need the guard here, the form state should be corrected instead (see other comment).

src/components/forms/NetworkDevicesForm/Acls.tsx
```diff
@@ -0,0 +1,118 @@
...
+  project: string;
+  network?: LxdNetwork;
+  device: LxdNicDevice | CustomNetworkDevice;
+  readOnly?: boolean;
+  formik?: InstanceAndProfileFormikProps;
+  index?: number;
+  canSelectManualAcls?: boolean;
+}
+
+const Acls: FC<Props> = ({
```
edlerd: Let's use a more descriptive name like `NetworkDeviceAcls`, so it is clear from a quick usage search where this component comes from.

src/pages/networks/forms/NetworkAclSelector.tsx
```diff
@@ -1,42 +1,73 @@
...
+    if (canSelectManualAcls) {
+      if (hasAcls) {
+        return "Select ACLs";
+      }
+      if (!hasAcls) {
+        return "No ACLs available";
+      }
+    }
+
+    return "-";
```
edlerd:
```suggestion
    if (!canSelectManualAcls) {
      return "-";
    }
    if (!hasAcls) {
      return "No ACLs available";
    }
    if (hasAcls) {
      return "Select ACLs";
    }
```
A bit less complex.

src/pages/networks/forms/NetworkAclSelector.tsx
```diff
@@ -1,42 +1,73 @@
...
 
 const NetworkAclSelector: FC<Props> = ({
   project,
   setSelectedAcls,
   selectedAcls,
+  id,
+  disabledItems,
+  disabled,
+  help,
+  canSelectManualAcls,
```
edlerd: I think props are now a bit bamboozled. We should probably remove disabled in favor of the new canSelectManualAcls, The disabledItems we should probably rename to avoid surprises with the added "from network" suffix. Maybe call it inheritedAcls instead.

src/pages/networks/forms/NetworkAclSelector.tsx
```diff
@@ -1,42 +1,73 @@
...
 }) => {
   const { data: availableAcls = [] } = useNetworkAcls(project);
 
-  const toOptionList = (list: string[]) => {
+  const toOptionList = (list: string[], disabledItems?: string[]) => {
```
edlerd: We should avoid shadowing the `disabledItems` name from the upper scope. This can lead to confusion and bugs due to that. But with renaming the incoming prop as suggested above, it should already be avoided.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -40,28 +48,26 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
...
-              }}
-            />
-          </div>
+          <NetworkAclSelector
+            project={project}
+            selectedAcls={formik.values.security_acls}
+            setSelectedAcls={(selectedItems) => {
+              formik.setFieldValue("security_acls", selectedItems);
+            }}
+            id={formik.values.readOnly ? undefined : networlAclSelectorId}
```
edlerd:
```suggestion
            id={networlAclSelectorId}
```
I think the conditional is redundant here.

src/pages/profiles/ProfileDetailOverview.tsx
```diff
@@ -94,7 +94,11 @@ const ProfileDetailOverview: FC<Props> = ({ profile, featuresProfiles }) => {
           <h2 className="p-heading--5">Networks</h2>
         </Col>
         <Col size={7}>
-          <NetworkListTable devices={profile.devices} onFailure={onFailure} />
+          <NetworkListTable
+            devices={profile.devices}
+            onFailure={onFailure}
+            profileName={profile.name}
```
edlerd: This can probably be reverted with the relative link (see above comment)

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -17,6 +18,7 @@ const NetworkSelector: FC<Props & SelectProps> = ({
   setValue,
   onBlur,
   hasNoneOption = false,
+  onChange: _ignoredOnChange,
```
edlerd:
```suggestion
```
Not sure why is this added here?

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -26,34 +28,77 @@ const NetworkSelector: FC<Props & SelectProps> = ({
...
+            </span>
+            <span title={network.type} className="network-option u-truncate">
+              {network.type}
+            </span>
+            <span
+              title="network ACLs"
+              className="network-option u-truncate u-align--right"
+            >
+              {getNetworkAcls(network).length || "-"}
+            </span>
```
edlerd, line 42: I think a bit more spacing on the right would be nice, so the numbers don't sit next to the end. Or maybe not right aligning. Though right alignment is usually good on numbers, but this one probably can be assumed to be single digit.

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -26,34 +28,77 @@ const NetworkSelector: FC<Props & SelectProps> = ({
...
+        <span className="network-option u-no-margin--bottom">Name</span>
+        <span className="network-option u-no-margin--bottom">Type</span>
+        <span className="network-option u-no-margin--bottom">ACLs</span>
+      </div>
+    );
+  };
+
   return (
-    <Select
+    <CustomSelect
```
edlerd, line 106: There is a warning here, when opening this one:
```
Warning: Formik called `handleBlur`, but you forgot to pass an `id` or `name` attribute to your input:
```
I think to resolve it, we should remove the `onBlur` when calling the `NetworkSelector` in `NetworkDevice.tsx`.

src/sass/_forms.scss
```diff
@@ -482,3 +501,53 @@
     margin-right: $sph--x-small;
   }
 }
+
+.network-select-dropdown {
+  div.label {
```
edlerd:
```suggestion
  .label {
```
When targeting classes, we usually avoid the specific element. This also applies to other css below.

src/sass/_forms.scss
```diff
@@ -482,3 +501,53 @@
     margin-right: $sph--x-small;
   }
 }
+
+.network-select-dropdown {
```
edlerd: This should probably live in a dedicated file for the network selector, similar to `_storage_pool_selector.scss`.

src/util/formDevices.tsx
```diff
@@ -98,11 +98,25 @@ export const formDeviceToPayload = (devices: FormDevice[]) => {
...
+        (item as LxdNicDevice)["security.acls"] !== undefined
+      ) {
+        return {
+          ...obj,
+          [name]: {
+            ...item,
+            "security.acls": (item as LxdNicDevice)["security.acls"],
+          },
+        };
+      }
```
edlerd:
```suggestion
```
This should not be needed, I think this function can stay as it was. I briefly tested commenting this out and it worked for me on editing an instance also without.

src/pages/projects/forms/NetworkSelector.tsx
kimanhou: I had a lint issue without it, because `selectProps` might include an `onChange` and then `CustomSelect` would not know which one to take.

edlerd: I see, thanks for the explainer. Can we update `Props & SelectProps` in the component type to `Props & CustomSelectProps`? Then the types should match.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,121 @@
...
+              project={project}
+              setValue={(value) => {
+                formik.setFieldValue(`devices.${index}.network`, value);
+
+                if (!isTypeOvn(network)) {
+                  formik.setFieldValue(
+                    `devices.${index}.network["security.acls"]`,
+                    undefined,
+                  );
+                }
```
edlerd, line 74: I think there is a race condition here. We are updating the formik state. The network that goes into the condition check may not be updated. This is a bit more complicated than I thought, because we don't have the network object here at hand. I think we need to filter the network from the full list of networks here based on the `value` coming in as network name.

src/components/NetworkListTable.tsx
```diff
@@ -61,43 +68,60 @@ const NetworkListTable: FC<Props> = ({ onFailure, devices, instance }) => {
...
+        },
+        {
+          content: instance?.config?.[`volatile.${deviceName}.hwaddr`] || "-",
+          role: "cell",
+          "aria-label": "MAC address",
+        },
+        {
+          content:
+            aclsCount > 0 ? (
+              <Link to={"configuration/network"}>{aclsCount}</Link>
```
edlerd:
```suggestion
              <Link to="configuration/network">{aclsCount}</Link>
```
A bit simpler.

src/pages/networks/forms/NetworkAclSelector.tsx
```diff
@@ -1,42 +1,70 @@
...
 }
 
 const NetworkAclSelector: FC<Props> = ({
   project,
   setSelectedAcls,
   selectedAcls,
+  id,
+  inheritedAcls,
+  help,
+  canSelectManualAcls,
```
edlerd:
```suggestion
  canSelectManualAcls = true,
```
This must have a default not to disable the acl selector on the network create / edit.

src/components/forms/NetworkDevicesForm/ReadOnlyAclsList.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  if (!allAcls.length) return null;
+
+  return (
+    <div className="acls-from-network">
+      <div>ACLs</div>
+      <div className="acls-list">
+        <ExpandableList
+          items={allAcls.map((acl) => (
+            <div key={acl} className="u-whitespace-nowrap">
```
edlerd:
```suggestion
            <div key={acl}>
```
I think with the line by line layout, we don't need this anymore.

src/sass/styles.scss
```diff
@@ -394,3 +394,7 @@ body {
 .u-break-all {
   word-break: break-all;
 }
+
+.u-whitespace-nowrap {
+  white-space: nowrap;
+}
```
```suggestion
```
With the comment above, this will become unused.

src/util/networks.tsx
```diff
@@ -196,3 +196,14 @@ export const renderNetworkType = (type: LxdNetwork["type"]) => {
...
 };
+
+export const getNetworkAcls = (network?: LxdNetwork) => {
+  if (network) {
+    return network.config["security.acls"]?.split(",").filter((t) => t) || [];
+  }
+  return [];
+};
+
+export const isTypeOvn = (network?: LxdNetwork) => {
```
edlerd, line 208: Excellent idea to extract this as a function and applying it throughout.

I'd suggest using the approach above in this file like with `typesWithAcls` defining a new `typesWithNicDeviceAcls` and using that in a dedicated function `supportsNicDeviceAcls`. So we extract the knowledge of which types support acls into a single place. To avoid leaking this definition into all the callers. In case we get other types with custom acl support, this will be easier to extend, and the code will be easier to read in the callers.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,121 @@
...
+}) => {
+  return (
+    <div className="network-device" key={index}>
+      <div>
+        {readOnly ? (
+          <div>
+            <div>Network</div>
+            <ResourceLink
+              type="network"
+              value={(formik.values.devices[index] as LxdNicDevice).network}
```
edlerd:
```suggestion
              value={device.network}
```
Some more occurrences of this can be simplified below.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { isTypeOvn } from "util/networks";
+
+interface Props {
+  index: number;
+  readOnly: boolean;
+  project: string;
+  formik: InstanceAndProfileFormikProps;
+  focusNetwork: (id: number) => void;
+  removeNetwork: (id: number) => void;
+  device: LxdNicDevice | CustomNetworkDevice;
```
edlerd:
```suggestion
  device: LxdNicDevice;
```

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
```diff
@@ -0,0 +1,90 @@
...
+import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import NetworkAclSelector from "pages/networks/forms/NetworkAclSelector";
+import { getDeviceAcls } from "util/devices";
+import { getNetworkAcls } from "util/networks";
+import ReadOnlyAclsList from "./ReadOnlyAclsList";
+
+interface Props {
+  project: string;
+  network?: LxdNetwork;
+  device: LxdNicDevice | CustomNetworkDevice;
```
edlerd:
```suggestion
  device: LxdNicDevice;
```

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
```diff
@@ -0,0 +1,90 @@
...
+  formik,
+  index,
+  canSelectManualAcls,
+}) => {
+  if (readOnly) {
+    return (
+      <ReadOnlyAclsList
+        project={project}
+        network={network}
+        device={device as LxdNicDevice}
```
edlerd:
```suggestion
        device={device}
```

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
```diff
@@ -0,0 +1,90 @@
...
+    );
+  }
+
+  const id = useId();
+  const networkAcls = getNetworkAcls(network);
+
+  const userSelectedAcls =
+    formik && index !== undefined
+      ? getDeviceAcls(formik.values.devices[index] as LxdNicDevice)
+      : [];
```
edlerd:
```suggestion
  const userSelectedAcls = device ? getDeviceAcls(device) : [];
```

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
```diff
@@ -0,0 +1,90 @@
...
+  );
+
+  const getHelperText = (): ReactNode => {
+    if (canSelectManualAcls) {
+      if (networkAcls.length === 0) return undefined;
+
+      return "Some ACLs are inherited from the network. They cannot be deselected here.";
+    }
+
+    return "Network must be of type OVN to customize ACLs.";
```
edlerd:
```suggestion
    if (!canSelectManualAcls) {
      return "Network must be of type OVN to customize ACLs.";
    }
    if (networkAcls.length === 0) return undefined;
    return "Some ACLs are inherited from the network. They cannot be deselected here.";
```

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  return (
+    <>
+      <label
+        className={!canSelectManualAcls ? "u-text--muted" : ""}
```
edlerd:
```suggestion
        className={canSelectManualAcls ? "" : "u-text--muted"}
```
Let's keep this positive as it is easier to read.

src/sass/_forms.scss
```diff
@@ -62,6 +62,20 @@
...
+        max-width: 19rem;
+      }
+    }
+
+    .acls-from-network {
+      margin-top: $spv--small;
+
+      > .acls-list {
+        display: flex;
+        flex-direction: column;
```
edlerd:
```suggestion
```
I think now with the line by line acls this can be removed. Maybe we can even remove the acl-list class?

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
kimanhou: `device` should always exist

src/pages/projects/forms/NetworkSelector.tsx
kimanhou: Should be resolved when react-components will be updated. See [PR](https://github.com/canonical/react-components/pull/1260)

src/pages/projects/forms/NetworkSelector.tsx
kimanhou: That does't work because then all callers of `<NetworkSelector>` then expect an `onChange`.
I specifically omitted problematic props: `Props & Omit<CustomSelectProps, "onChange" | "options" | "value">`.
And removed onBlur, which was not used. I hope I didn't miss anything.

edlerd: Yes, this is much better, as the omitted properties will anyway be overridden by this component. So an upstream user should net be setting those, as they won't be applied.

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -1,59 +1,102 @@
...
-    options.unshift({
-      label: options.length === 0 ? "No networks available" : "Select option",
-      value: "",
-      disabled: true,
-    });
+    if (options.length === 0) {
+      options.unshift({
+        label: <span>No networks available</span>,
+        value: "",
+        text: "",
```
edlerd:
```suggestion
        text: "None",
```
This is causing the project creation page to crash, because no string value is causing the custom select to throw. I think this is also what is causing the test failures.

src/pages/projects/forms/NetworkSelector.tsx
```diff
@@ -1,59 +1,102 @@
...
+    }
 
     if (hasNoneOption) {
-      options.push({ label: "No network", value: "none", disabled: false });
+      options.push({
+        label: <span>No network</span>,
+        value: "none",
+        text: "none",
+        disabled: false,
+      });
```
edlerd, line 87: This option in the project creation looks a bit odd. I think we should use the same tabular structure here and render "-" for the other columns. Wdyt?

image

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -89,6 +90,9 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
   const { hasProjectsNetworksZones, hasStorageBuckets } =
     useSupportedFeatures();
 
+  const { data: networks = [] } = useNetworks(project?.name || "");
```
edlerd:
```suggestion
  const { data: networks = [] } = useNetworks(project?.name || "default");
```
We must not use an empty project here, falling back to the `default` one as before seems reasonable.

src/sass/_network_selector.scss
```diff
@@ -0,0 +1,50 @@
...
+      }
+
+      &:nth-child(2) {
+        width: 8rem;
+      }
+
+      &:nth-child(3) {
+        padding-right: 8px;
+        width: 4rem;
+      }
```
edlerd, line 18: Spacing and margin for the ACL column contents looks good. Please apply the same spacing and right alignment to the ACL header so it is aligned with the numbers. Currently the heading sits a bit too much on the left.

image

src/util/formChangeCount.tsx
```diff
@@ -95,6 +95,16 @@ const getLimitChanges = (formik: ConfigurationRowFormikProps): number => {
...
+  const aKeys = Object.keys(a);
+  const bKeys = Object.keys(b);
+  if (
+    bKeys.length - aKeys.length === 1 &&
+    !aKeys.includes("security.acls") &&
+    bKeys.includes("security.acls")
+  ) {
+    changeCount += 1;
+  }
+
```
edlerd:
```suggestion
```
I don't think this is needed. Or which condition is this counting?

kimanhou: It was needed when adding/removing ACLs, eg when adding an ACL to a device that didn't have any or removing the unique ACL on a device resulting in no ACL.
Now `security.acls` is always here but is `undefined` so this bit is not needed anymore.

package.json
```diff
@@ -31,7 +31,7 @@
     "test-report-coverage": "cp coverage/unit/coverage-final.json coverage/playwright/ ; nyc report --reporter html --reporter cobertura --reporter text-summary --temp-dir coverage/playwright --report-dir coverage/playwright-report --exclude 'src/lib/**' --exclude 'src/types/**'"
   },
   "dependencies": {
-    "@canonical/react-components": "3.2.0",
+    "@canonical/react-components": "3.3.0",
```
edlerd, line 34: Please rebase on main, then changes to this file should vanish.

src/sass/_network_selector.scss
```diff
@@ -0,0 +1,50 @@
...
+      &:nth-child(1) {
+        width: 12rem;
+      }
+
+      &:nth-child(2) {
+        width: 8rem;
+      }
+
+      &:nth-child(3) {
+        padding-right: 8px;
```
edlerd, line 16: Please add the same padding to the header to fix the slight misalignment.

PR #1527, opened 2025-10-06
feat(SSO login): Advertise SSO in login page and in Permissions/Identities screen

kimanhou:
## Done

- In Login screen, display SSO option
- In Permissions > Identities screen, display a notification that advertises SSO

## QA

1. Run the LXD-UI:

## Screenshots

SSO-permissions-identities
SSO-provider-selector
https://github.com/user-attachments/assets/239a2308-4dea-4a1b-a590-27db2988039e


edlerd:
> Did I break something ? Unit tests fail when installing packages

Looks like a temporary network error, requested a re-run manually. HTH.

src/components/SsoProviderSelector/assets/Auth0Logo.tsx
```diff
@@ -0,0 +1,45 @@
+import type { FC } from "react";
+
+const Auth0Logo: FC = () => {
```
edlerd: Please put the SVGs into `public/assets/img` folder and use them like the existing assets.

Also, I am not 100% confident we are allowed to use those logos. We should best clarify on that before merging. Though I think the identity team is also using logos for a similar use case and it probably is ok.

src/pages/permissions/SsoNotification.tsx
```diff
@@ -0,0 +1,51 @@
...
+      actions={[
+        {
+          label: "Show me how",
+          onClick: () => {
+            openSsoModal();
+          },
+        },
+      ]}
+    >
+      LXD can be configured to log in using a single-signon provider instead of
```
edlerd:
```suggestion
      LXD can be configured to log in using a single-sign on provider instead of
```
I think this should be with a space. Wdyt?

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -365,6 +386,16 @@ const PermissionIdentities: FC = () => {
...
+
+      {isSsoModal && (
+        <Modal
+          close={closeSsoModal}
+          title="SSO setup"
+          className="choose-sso-provider-modal"
+        >
+          <SsoProviderSelector />
+        </Modal>
+      )}
```
edlerd: We should best contain this logic within the `SsoNotification` component, so it triggers the modal, and also renders it within a portal. So we don't leak details like state handling for the modal and import of components into this component.

src/components/SsoProviderSelector/SsoProviderSelector.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const SsoProviderSelector: FC = () => {
+  return (
+    <div className="sso-provider-selector u-flex-column">
+      <p>Choose an SSO provider</p>
+      <div className="sso-provider-column u-flex-column">
+        <div className="sso-provider-row u-flex">
+          <SsoProviderButton
+            providerName="Entra ID"
+            to="https://documentation.ubuntu.com/lxd/latest/howto/oidc_entra_id/#"
```
edlerd: I think we should introduce the `useDocs` helper here. And then link to the local documentation if available with it.

src/pages/login/SetupSsoLogin.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  if (isAuthenticated) {
+    return <Navigate to="/ui" replace={true} />;
+  }
+
+  return (
+    <CustomLayout
+      contentClassName="login"
+      header={
+        <RenameHeader
```
edlerd: This seems like a wrong usage of the component, as there is no renaming here. Maybe we can use the same modal interaction as on the identity page?

src/pages/login/Login.tsx
```diff
@@ -46,8 +47,27 @@ const Login: FC = () => {
...
+                to="/ui/login/set-up-sso-login"
+              >
+                <Icon
+                  name="security"
+                  className={classnames("auth-button-icon", {
+                    "is-light": false,
+                  })}
+                />
+                <span>Set up SSO login</span>
+              </Link>
```
edlerd, line 63: I think this is highly confusing. As we switch the functionality, while the design doesn't change: Big green button above big white button. We should probably advertise it differently. We could reuse the notification approach from the identities page. Including the close logic to remember this is not showing again.

without sso:

image

with sso:

image

src/components/SsoProviderSelector/SsoProviderSelector.tsx
```diff
@@ -0,0 +1,42 @@
...
+import EntryIdLogo from "./assets/EntryIdLogo";
+import KeycloakLogo from "./assets/KeycloakLogo";
+import OryHydraLogo from "./assets/OryHydraLogo";
+import Auth0Logo from "./assets/Auth0Logo";
+import SsoProviderButton from "./SsoProviderButton";
+
+const SsoProviderSelector: FC = () => {
+  return (
+    <div className="sso-provider-selector u-flex-column">
+      <p>Choose an SSO provider</p>
```
edlerd: In the modal case, we can probably raise this as the modal title and remove it here. Similar to the image below.

image

src/pages/login/Login.tsx
```diff
@@ -9,6 +9,7 @@ const Login: FC = () => {
   const { isAuthenticated, isAuthLoading } = useAuth();
   const { data: settings } = useSettings();
   const hasOidc = settings?.auth_methods?.includes("oidc");
+  const hasTls = settings?.auth_methods?.includes("tls");
```
edlerd:
```suggestion
```
TLS will always be available here, see also [server.go](https://github.com/canonical/lxd/blob/main/shared/api/server.go#L212). All references to the variable in LXD that write to it include TLS.

edlerd (changes requested):
The notification on the identities list seems good, just tiny suggestions to improve it. The login is a bit confusing now. Some other random thoughts on the changes below.

src/components/SsoProviderSelector/assets/Auth0Logo.tsx
kimanhou: @piperdeck Would you know if/how we are allowed to use external logos ?

src/pages/permissions/SsoNotification.tsx
kimanhou: I am not sure, I just copy pasted what's on @piperdeck 's Figma. Gemini says "**single sign-on**"

src/pages/login/SetupSsoLogin.tsx
kimanhou: @piperdeck What do you think ?
Technically, it's easier to use the modal solution in both cases and it simplifies the code. I'll implement that while waiting for Piper's opinion

src/pages/login/Login.tsx
kimanhou: @piperdeck ?

src/pages/permissions/SsoNotification.tsx
```diff
@@ -0,0 +1,64 @@
...
+          {
+            label: "Show me how",
+            onClick: () => {
+              openModal();
+            },
+          },
+        ]}
+      >
+        LXD can be configured to log in using a single sign-on provider instead
+        of TLS certificates.
```
edlerd:
```suggestion
        LXD can be configured to log in using a single sign-on provider.
```
A bit more concise?

src/components/SsoProviderSelector/SsoProviderButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+const SsoProviderButton: FC<Props> = ({ providerName, to, imgSrc }: Props) => {
+  return (
+    <Link to={to} target="_blank" className="sso-provider-link">
+      <Button hasIcon title={providerName} className="sso-provider-button">
+        <img src={imgSrc} alt={`${providerName} logo"`} className="sso-logo" />
+        <span>{providerName}</span>
+      </Button>
+    </Link>
```
edlerd:
```suggestion
    <Link
      to={to}
      target="_blank"
      rel="noopener noreferrer"
      className="p-button has-icon"
    >
      <img src={imgSrc} alt={`${providerName} logo"`} className="p-icon" />
      <span>{providerName}</span>
    </Link>
```
We can use vanilla classes to avoid most of the custom styling and avoid the button in link pattern. The suggestion above is failing to provide a fixed width for the button, so we might still need a custom class for that.


 We should add the `rel="noopener noreferrer"` because it might be an external link.

src/components/SsoProviderSelector/SsoProviderSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+              imgSrc="/ui/assets/img/ory-hydra-logo.svg"
+            />
+            <SsoProviderButton
+              providerName="Auth0"
+              to={`${docBaseLink}/howto/oidc_auth0/`}
+              imgSrc="/ui/assets/img/auth0-logo.svg"
+            />
+          </div>
+        </div>
+      </div>
```
edlerd: I think we could reduce the structure to a single `div` wrapping the buttons. Then applying a grid layout to produce the two columns and switch to a single column on smaller screens. That way, the DOM structure is simpler, and the layout is responsive.

src/pages/login/Login.tsx
edlerd: How about showing the same notification as on the identities page on the login screen if SSO is not configured? Would it look okay on the login screen?

src/pages/permissions/SsoNotification.tsx
edlerd: > single sign-on

That looks good to me.

src/pages/login/Login.tsx
kimanhou: Like this without SSO ?
image

And this with SSO ?
image

edlerd: > Like this without SSO ?

Yes, that was the idea. Maybe move the notification to the bottom of the screen? Also it might need a max-width to not span the whole screen on large/very large screens. Wdyt?

src/components/SsoProviderSelector/SsoProviderSelector.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <Modal
+      close={closeModal}
+      title="Choose Single Sign-On provider"
+      className="choose-sso-provider-modal"
+    >
+      <p>
+        Single Sign-On allows you to log into LXD using an external identity
+        (like Google or Microsoft) rather than a custom identity defined by LXD.
```
edlerd:
```suggestion
        like Google or Microsoft.
```
This is much better with the intro text! Can we make it a little bit more concise?

src/components/SsoProviderSelector/SsoProviderSelector.tsx
```diff
@@ -0,0 +1,54 @@
...
+            to={`${docBaseLink}/howto/oidc_ory/`}
+            imgSrc="/ui/assets/img/ory-hydra-logo.svg"
+          />
+          <SsoProviderButton
+            providerName="Auth0"
+            to={`${docBaseLink}/howto/oidc_auth0/`}
+            imgSrc="/ui/assets/img/auth0-logo.svg"
+          />
+        </div>
+      </div>
```
edlerd: Can we only use one div as a wrapper? Also this should be responsive, so if we shrink the screen width, the buttons are in one column.

src/pages/login/Login.tsx
```diff
@@ -19,39 +34,59 @@ const Login: FC = () => {
...
+                  onClick={() => {
+                    openModal();
+                  }}
+                  appearance="positive"
+                >
+                  <Icon
+                    name="security"
+                    className={classnames("auth-button-icon", {
+                      "is-light": false,
+                    })}
```
edlerd:
```suggestion
                    light
```
We probably don't need any custom class here anymore. This seems like an oversight from previous code that it wasn't removed.

src/pages/login/Login.tsx
```diff
@@ -19,39 +34,59 @@ const Login: FC = () => {
...
+                to="/ui/login/certificate-generate"
+              >
+                <Icon
+                  name="certificate"
+                  className={classnames("auth-button-icon", {
+                    "is-light": !hasOidc,
+                  })}
+                />
+                <span>Set up TLS login</span>
+              </Link>
```
edlerd, line 63:
```suggestion
              <Link
                className="has-icon p-button"
                to="/ui/login/certificate-generate"
              >
                <Icon name="certificate" />
                <span>Set up TLS login</span>
              </Link>
```
This can be much simpler now as it is always the secondary button.

src/sass/styles.scss
```diff
@@ -288,6 +289,11 @@ body {
   display: flex;
 }
 
+.u-flex-column {
+  display: flex;
+  flex-direction: column;
+}
+
```
```suggestion
```
This seems unused with latest changes.

src/pages/login/Login.tsx
```diff
@@ -19,39 +33,46 @@ const Login: FC = () => {
...
+                </a>
+              )}
+              {!hasOidc && (
+                <Button
+                  hasIcon
+                  onClick={() => {
+                    openModal();
+                  }}
+                  appearance="positive"
+                >
```
edlerd:
```suggestion
                <Button hasIcon onClick={openModal} appearance="positive">
```
A bit simpler

src/sass/_sso_provider_selector.scss
```diff
@@ -0,0 +1,36 @@
...
+      width: 10rem;
+
+      .sso-logo {
+        align-items: center;
+        display: flex;
+        height: 16px;
+        margin-left: -0.5rem;
+        margin-right: 0.5rem;
+        width: 16px;
+      }
```
edlerd:
```suggestion
```
This seems unused after latest changes.

src/sass/_sso_provider_selector.scss
```diff
@@ -0,0 +1,36 @@
...
+    align-items: center;
+    display: grid;
+    gap: $spv--large;
+    grid-template-columns: 1fr 1fr;
+    margin: 0 auto;
+    padding: $spv--large 0;
+    width: 22rem;
+
+    .sso-provider-link {
+      align-items: center;
```
edlerd, line 12:
```suggestion
```
This is the vanilla default, now that we are using `p-button`.

public/assets/img/ory-hydra-logo.svg
edlerd, line 1: This one seems smaller than the other ones on screen. I think this is because this one has a bit too much empty space around the logo content.

edlerd (commented):
QA LGTM, very small nitpicks on the code to simplify. Then we can merge it from my side.
Only pending the green light on using the logos.

PR #1534, opened 2025-10-08
feat: [WD-24082] Schedule doc screenshots

Kxiru:
## Done

- Created screenshot_automation.yaml, a new workflow to automate documentation screenshots.

Fixes inability to generate documentation images.

## QA

1. Run the LXD-UI:
    - Manually start the workflow from Github actions and verify that screenshots are generated as behaviour suggests.

## Screenshots

N/A

### DIFF
--- a/.github/workflows/screenshot_automation.yaml
+++ b/.github/workflows/screenshot_automation.yaml
@@ -1,0 +1,88 @@
+name: Screenshot automation
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 0'  # Weekly on Sundays at midnight
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  browser-e2e-test:
+    name: generate-screenshots
+    runs-on: ubuntu-latest
+    outputs:
+      job_status: ${{job.status}}
+    env:
+      LXD_OIDC_CLIENT_ID: "RYDnMpkygLAMfeo17lU7LYwWGxisRuRR"
+      LXD_OIDC_CLIENT_SECRET: "CNKX4UmrZKZJq5rJy5VM_JfcNPqkws1rwWWQk_q0oyZ8gABARr19ic7xrhPssGA1"
+      LXD_OIDC_ISSUER: "https://dev-h6c02msuggpi6ijh.eu.auth0.com/"
+      LXD_OIDC_AUDIENCE: "https://dev-h6c02msuggpi6ijh.eu.auth0.com/api/v2/"
+      LXD_OIDC_USER: "lxd-ui-e2e-tests@example.org"
+      LXD_OIDC_PASSWORD: "lxd-ui-e2e-password"
+      LXD_OIDC_GROUPS_CLAIM: "lxd-idp-groups"
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Install Dotrun
+        run: |
+          sudo pip3 install dotrun
+
+      - name: Install LXD-UI dependencies
+        run: |
+          set -x
+          sudo chmod 0777 ../lxd-ui
+          dotrun install
+
+      - name: Run LXD-UI
+        env:
+          ENVIRONMENT: devel
+          PORT: 8407
+          LXD_UI_BACKEND_IP: 172.17.0.1
+          VITE_PORT: 3000
+        run: |
+          dotrun &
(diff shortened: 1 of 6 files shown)


edlerd (dismissed):
The workflow is [currently failing](https://github.com/Kxiru/lxd-ui/actions/runs/18348162108/job/52260381260) and is skipping most of the tests, because `CI=true` needs to be set in the flow.

Edit: I meant `CI=false` needs to be set, the current default is leading to the skipped tests.

.github/workflows/screenshot_automation.yaml
```diff
@@ -0,0 +1,86 @@
...
+          npx playwright test --project screenshots
+          npx playwright test --project screenshots-clustered
+
+      - name: Upload screenshot-automation-report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: report
+          path: blob-report
+          retention-days: 1
```
edlerd:
```suggestion
          retention-days: 1

```
Nit: let's format with a trailing newline.

.github/workflows/screenshot_automation.yaml
```diff
@@ -0,0 +1,86 @@
+name: Screenshot automation
+on: 
+  workflow_dispatch:
```
edlerd: Now this is stable, I think we should also schedule this to be run weekly.

tests/helpers/network.ts
```diff
@@ -48,12 +48,14 @@ export const deleteNetwork = async (page: Page, network: string) => {
...
     .getByRole("button", { name: "Delete" })
     .click();
+  await page.waitForTimeout(2500);
   const networkLink = await getNetworkLink(page, network);
   await expect(networkLink).not.toBeVisible();
 };
 
 export const visitNetwork = async (page: Page, network: string) => {
   await gotoURL(page, "/ui/");
+  await page.waitForTimeout(1500);
```
edlerd:
```suggestion
```
This seems redundant. If it is needed, please explain why this is here with a comment.

tests/docs-screenshots.spec.ts
```diff
@@ -646,15 +644,25 @@ test("LXD - UI Folder - Instances", async ({ page }) => {
...
+  await gotoURL(page, "/ui/");
+  await page.waitForTimeout(1000);
+  await page.getByText("Networking").click();
+  await page.getByText("Networks").click();
+  await page.getByText("Create network").click();
+  await page.getByPlaceholder("Enter name").fill(network2);
+  await page.getByLabel("Type").selectOption("bridge");
+  await page.waitForTimeout(2500);
+  await page.getByRole("button", { name: "Create", exact: true }).click();
+  await page.waitForTimeout(2500);
```
edlerd: I think this whole block is redundant and can use the `createNetwork` helper instead.

tests/docs-screenshots.spec.ts
```diff
@@ -136,7 +137,7 @@ test("networks", async ({ page }) => {
     clip: getClipPosition(240, 0, 1420, 750),
   });
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForSelector(`text=Network ${network} created.`);
+  await page.waitForTimeout(2500);
```
edlerd: Maybe instead of duplicating we can define a `submitCreateNetwork` helper that extracts those lines from the `createNetwork` helper, so we can reuse it here and avoid the explicit timeout in the test:

```
  await page.waitForTimeout(750);
  await page.waitForTimeout(2500);
  const networkLink = await getNetworkLink(page, network);
  await expect(networkLink).toBeVisible();
  ```

File: tests/helpers/network.ts [hunk 6]
```diff
@@ -152,7 +154,15 @@ export const getNetworkLink = async (page: Page, network: string) => {
...
   return networkLink;
 };
+
+export const submitCreateNetwork = async (page: Page, network: string) => {
+  await page.waitForTimeout(750);
+  await page.getByRole("button", { name: "Create", exact: true }).click();
+  await page.waitForTimeout(2500);
+  const networkLink = await getNetworkLink(page, network);
+  await expect(networkLink).toBeVisible();
+};
```
edlerd, line 164: This is duplicating parts of `createNetwork` above in this file, lets replace it with calling this function instead.

Kxiru: Such that this SubmitCreateNetwork now needs to be called after all instances of createNetwork?

edlerd: I thought createNetwork can call this function directly, as a small refactor, to reduce code duplication.

Kxiru: Good plan

edlerd (approved):
LGTM, thanks :)

PR #1539, opened 2025-10-10
feat(NetworkDevicesForm): override inherited network

### LINKED ISSUE #1547: Feature: deeplink to inherited profile
Image

Do you think it's a nice feature to make the `network-infra` profile clickable and link to the profile's page?

kimanhou:
## Done

- Handle overrides: if a device is overriden, display both on the same line.
- An inherited network can be overridden from the ui with one click.
- An inherited network, that is **not** overriden, can be removed from the ui with one click (a device with `type = none` and same name is created)
- An inherited network, that **is** overriden, can become not overriden from the ui with one click (the override device is removed)
- Profile chip in inherited network

Fixes #1547 

## QA

1. Run the LXD-UI:
    - Step 0: Create or use an instance with an inherited network.
image

- Step 1: In Instance > Configuration > Network > in the column override, make sure that `Edit` and `Detach` buttons exist.
    
- Step 2: Click on `Detach`. Everything in Inherited column should be striked through. Save.
image

- Step 3: The inherited device is now removed. To put it back, remove the override by clicking `Clear` and save. You should see the original UI of Step 0.
- Step 4: Click on `Edit`. Select any network and save. The inherited network should be overriden.
image

- Step 5: Click on `Clear` and save. You should see the original UI of Step 0.

## Screenshots

Before
before

### Inherited
Not overridden, read mode
image

Not overridden, edit mode after clicking on "Detach"
image

Overridden, read mode
image

Overridden, edit mode
image

### DIFF
--- a/src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
+++ b/src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
@@ -1,0 +1,100 @@
+import ResourceLink from "components/ResourceLink";
+import { getConfigurationRowBase } from "components/ConfigurationRow";
+import ReadOnlyAclsList from "components/forms/NetworkDevicesForm/ReadOnlyAclsList";
+import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import type { InheritedNetwork } from "util/configInheritance";
+import type { LxdNetwork } from "types/network";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import classnames from "classnames";
+import NetworkDevice from "./NetworkDevice";
+import type { LxdNicDevice, LxdNoneDevice } from "types/device";
+import { isNoneDevice } from "util/devices";
+
+interface Props {
+  device: InheritedNetwork;
+  project: string;
+  managedNetworks: LxdNetwork[];
+  formik: InstanceAndProfileFormikProps;
+}
+
+export const getInheritedNetworkRow = ({
+  device,
+  project,
+  managedNetworks,
+  formik,
+}: Props): MainTableRow => {
+  const overrideDevice = formik.values.devices.find(
+    (t) => t.name === device.key,
+  ) as LxdNicDevice | LxdNoneDevice | undefined;
+  const isOverridden = overrideDevice !== undefined;
+  const isDetached = overrideDevice && isNoneDevice(overrideDevice);
+
+  const overrideNetwork = managedNetworks.find(
+    (t) => t.name === (overrideDevice as LxdNicDevice)?.network,
+  );
+
+  return getConfigurationRowBase({
+    configuration: (
+      <b
+        className={classnames({
+          "u-text--muted": isDetached,
+          "u-text--line-through": isDetached,
+        })}
+      >
+        {device.key}
+      </b>
+    ),
+    inherited: (
+      <div>
+        <div
+          className={classnames("p-text--small", "u-text--muted", {
+            "u-text--line-through": isOverridden,
+          })}
+        >
+          From profile{" "}
+          <ResourceLink
+            type="profile"
+            value={device.sourceProfile}
+            to={`/ui/project/${encodeURIComponent(project)}/profile/${encodeURIComponent(device.sourceProfile)}`}
+            className={classnames({
+              "u-text--line-through": isOverridden,
+            })}
+          />
+        </div>
+        <div
+          className={classnames({
+            "u-text--muted": isOverridden,
+            "u-text--line-through": isOverridden,
+          })}
+        >
+          Network
+        </div>
+        <ResourceLink
+          type="network"
+          value={device.network?.network || ""}
+          to={`/ui/project/${encodeURIComponent(project ?? "")}/network/${encodeURIComponent(device.network?.network || "")}`}
+          className={classnames({
+            "u-text--line-through": isOverridden,
+          })}
+        />
+        <ReadOnlyAclsList
+          project={project}
+          network={managedNetworks.find(
+            (t) => t.name === device.network?.network,
+          )}
+          device={device.network}
+          isOverridden={isOverridden}
+        />
+      </div>
+    ),
+    override: (
+      <NetworkDevice
+        formik={formik}
+        project={project}
+        device={overrideDevice}
+        network={overrideNetwork}
+        inheritedDevice={device}
+      />
+    ),
+  });
+};
--- a/src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
+++ b/src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
@@ -1,129 +1,56 @@
 import { type FC } from "react";
-import type { LxdNicDevice } from "types/device";
+import type { LxdNicDevice, LxdNoneDevice } from "types/device";
 import type { LxdNetwork } from "types/network";
 import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
-import NetworkSelector from "pages/projects/forms/NetworkSelector";
-import { Button, Icon } from "@canonical/react-components";
-import { ensureEditMode } from "util/instanceEdit";
-import NetworkDeviceAcls from "components/forms/NetworkDevicesForm/NetworkDeviceAcls";
-import ResourceLink from "components/ResourceLink";
-import { supportsNicDeviceAcls } from "util/networks";
 import { useNetworks } from "context/useNetworks";
+import NetworkDeviceActionButtons from "./NetworkDeviceActionButtons";
+import NetworkDeviceContent from "./NetworkDeviceContent";
+import { getIndex } from "util/devices";
+import type { InheritedNetwork } from "util/configInheritance";
+import type { EditInstanceFormValues } from "pages/instances/EditInstance";
 
 interface Props {
-  index: number;
-  readOnly: boolean;
   project: string;
   formik: InstanceAndProfileFormikProps;
-  focusNetwork: (id: number) => void;
-  removeNetwork: (id: number) => void;
-  device: LxdNicDevice;
+  device?: LxdNicDevice | LxdNoneDevice;
   network?: LxdNetwork;
+  inheritedDevice?: InheritedNetwork;
 }
 
 const NetworkDevice: FC<Props> = ({
-  index,
-  readOnly,
   project,
   formik,
-  focusNetwork,
-  removeNetwork,
-  network,
   device,
+  network,
+  inheritedDevice,
 }) => {
+  const readOnly = (formik.values as EditInstanceFormValues).readOnly;
   const { data: networks = [] } = useNetworks(project);
   const managedNetworks = networks.filter((network) => network.managed);
+  const index = getIndex(device?.name || "", formik);
 
   return (
     <div className="network-device" key={index}>
-      <div>
-        {readOnly ? (
-          <div>
-            <div>Network</div>
-            <ResourceLink
-              type="network"
-              value={device.network}
-              to={`/ui/project/${encodeURIComponent(project ?? "")}/network/${encodeURIComponent(device.network)}`}
-            />
-            <NetworkDeviceAcls
-              project={project}
-              network={network}
-              device={device}
-              readOnly
-            />
-          </div>
-        ) : (
-          <>
-            <NetworkSelector
-              value={device.network}
-              setValue={(value) => {
-                formik.setFieldValue(`devices.${index}.network`, value);
-
-                const selectedNetwork = managedNetworks.find(
-                  (t) => t.name === value,
-                );
-
-                if (
-                  selectedNetwork &&
-                  !supportsNicDeviceAcls(selectedNetwork)
-                ) {
-                  formik.setFieldValue(
-                    `devices.${index}["security.acls"]`,
-                    undefined,
-                  );
-                }
-              }}
-              id={`devices.${index}.network`}
-              name={`devices.${index}.network`}
-              managedNetworks={managedNetworks}
-            />
-            <NetworkDeviceAcls
-              project={project}
-              network={network}
-              device={device}
-              readOnly={readOnly}
-              formik={formik}
-              index={index}
-              canSelectManualAcls={supportsNicDeviceAcls(network)}
-            />
-          </>
+      <div className="network-device-content">
+        {device && (
+          <NetworkDeviceContent
+            readOnly={readOnly}
+            project={project}
+            device={device}
+            formik={formik}
+            index={index}
+            managedNetworks={managedNetworks}
+            network={network}
+          />
         )}
       </div>
-      <div>
-        {readOnly && (
-          <Button
-            onClick={() => {
-              ensureEditMode(formik);
-              focusNetwork(index);
-            }}
-            type="button"
-            appearance="base"
-            title={formik.values.editRestriction ?? "Edit network"}
-            className="u-no-margin--top"
-            hasIcon
-            dense
-            disabled={!!formik.values.editRestriction}
-          >
-            <Icon name="edit" />
-          </Button>
-        )}
-        <Button
-          className="delete-device u-no-margin--top"
-          onClick={() => {
-            ensureEditMode(formik);
-            removeNetwork(index);
-          }}
-          type="button"
-          appearance="base"
-          hasIcon
-          dense
-          title={formik.values.editRestriction ?? "Detach network"}
-          disabled={!!formik.values.editRestriction}
-        >
-          <Icon name="disconnect" />
-          <span>Detach</span>
-        </Button>
-      </div>
+      <NetworkDeviceActionButtons
+        readOnly={readOnly}
+        formik={formik}
+        index={index}
+        device={device}
+        inheritedDevice={inheritedDevice}
+      />
     </div>
   );
 };
--- a/src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
+++ b/src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
@@ -3,7 +3,7 @@ import type { LxdNicDevice } from "types/device";
 import type { LxdNetwork } from "types/network";
 import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
 import NetworkAclSelector from "pages/networks/forms/NetworkAclSelector";
-import { getDeviceAcls } from "util/devices";
+import { getDeviceAcls, getIndex } from "util/devices";
 import { getNetworkAcls } from "util/networks";
 import ReadOnlyAclsList from "./ReadOnlyAclsList";
 
@@ -13,7 +13,6 @@ interface Props {
   device: LxdNicDevice;
   readOnly?: boolean;
   formik?: InstanceAndProfileFormikProps;
-  index?: number;
   canSelectManualAcls?: boolean;
 }
 
@@ -23,7 +22,6 @@ const NetworkDeviceAcls: FC<Props> = ({
   device,
   readOnly,
   formik,
-  index,
   canSelectManualAcls,
 }) => {
   if (readOnly) {
@@ -62,7 +60,7 @@ const NetworkDeviceAcls: FC<Props> = ({
           selectedAcls={selectedAcls}
           setSelectedAcls={(selectedItems) => {
             formik.setFieldValue(
-              `devices.${index}["security.acls"]`,
+              `devices.${getIndex(device.name || "", formik)}["security.acls"]`,
               selectedItems.join(","),
             );
           }}
--- a/src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
+++ b/src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
@@ -1,0 +1,175 @@
+import { Button, Icon } from "@canonical/react-components";
+import type { FC } from "react";
+import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import { ensureEditMode } from "util/instanceEdit";
+import {
+  addNicDevice,
+  addNoneDevice,
+  focusNicDevice,
+  removeNicDevice,
+} from "util/formDevices";
+import type { LxdNicDevice, LxdNoneDevice } from "types/device";
+import { isNicDevice, isNoneDevice } from "util/devices";
+import type { InheritedNetwork } from "util/configInheritance";
+
+interface Props {
+  readOnly: boolean;
+  formik: InstanceAndProfileFormikProps;
+  index: number;
+  device?: LxdNicDevice | LxdNoneDevice;
+  inheritedDevice?: InheritedNetwork;
+}
+
+const NetworkDeviceActionButtons: FC<Props> = ({
+  readOnly,
+  formik,
+  index,
+  device,
+  inheritedDevice,
+}: Props) => {
+  const isPurelyInherited = inheritedDevice && !device;
+  const hasNicOverride = inheritedDevice && device && isNicDevice(device);
+  const hasNoneOverride = inheritedDevice && device && isNoneDevice(device);
+  const isLocal = !inheritedDevice && device;
+
+  const isDisabled = !!formik.values.editRestriction;
+
+  const getEditTitle = () => {
+    if (formik.values.editRestriction) return formik.values.editRestriction;
+    if (isPurelyInherited || hasNoneOverride) return "Create override";
+    if (hasNicOverride) return "Edit override";
+    return "Edit network";
+  };
+  const editTitle = getEditTitle();
+
+  const onEdit = () => {
+    ensureEditMode(formik);
+    if (isPurelyInherited || hasNoneOverride) {
+      const newDeviceIndex = formik.values.devices.length;
+      addNicDevice({
+        formik,
+        deviceName: inheritedDevice.key,
+        deviceNetworkName: inheritedDevice.network?.network ?? "",
+      });
+      focusNicDevice(newDeviceIndex);
+    } else {
+      focusNicDevice(index);
+    }
+  };
+
+  const clearOverride = () => {
+    ensureEditMode(formik);
+    removeNicDevice({ formik, deviceName: device?.name || "" });
+  };
+
+  const detachInherited = () => {
+    ensureEditMode(formik);
+    addNoneDevice(inheritedDevice?.key || "", formik);
+  };
+
+  const detachOverridden = () => {
+    ensureEditMode(formik);
+    addNoneDevice(device?.name || "", formik);
+  };
+
+  return (
+    <div className="network-device-actions">
+      {(readOnly || isPurelyInherited) && (
+        <Button
+          onClick={onEdit}
+          type="button"
+          appearance="base"
+          title={editTitle}
+          className="u-no-margin--top"
+          hasIcon
+          dense
+          disabled={isDisabled}
+        >
+          <Icon name="edit" />
+          <span>Edit</span>
+        </Button>
+      )}
+
+      {isPurelyInherited && (
+        <Button
+          className="u-no-margin--top"
+          onClick={detachInherited}
+          type="button"
+          appearance="base"
+          hasIcon
+          dense
+          title={formik.values.editRestriction || "Detach network"}
+          disabled={isDisabled}
+        >
+          <Icon name="disconnect" />
+          <span>Detach</span>
+        </Button>
+      )}
+
+      {hasNicOverride && (
+        <>
+          <Button
+            className="u-no-margin--top"
+            onClick={clearOverride}
+            type="button"
+            appearance="base"
+            hasIcon
+            dense
+            title={formik.values.editRestriction || "Clear override"}
+            disabled={isDisabled}
+          >
+            <Icon name="close" />
+            <span>Clear</span>
+          </Button>
+          <Button
+            className="u-no-margin--top"
+            onClick={detachInherited}
+            type="button"
+            appearance="base"
+            hasIcon
+            dense
+            title={formik.values.editRestriction || "Detach network"}
+            disabled={isDisabled}
+          >
+            <Icon name="disconnect" />
+            <span>Detach</span>
+          </Button>
+        </>
+      )}
+
+      {hasNoneOverride && (
+        <Button
+          className="u-no-margin--top"
+          onClick={clearOverride}
+          type="button"
+          appearance="base"
+          hasIcon
+          dense
+          title={formik.values.editRestriction || "Reattach inherited network"}
+          disabled={isDisabled}
+        >
+          <Icon name="connected" />
+          <span>Reattach</span>
+        </Button>
+      )}
+
+      {isLocal && (
+        <Button
+          className="u-no-margin--top"
+          onClick={detachOverridden}
+          type="button"
+          appearance="base"
+          hasIcon
+          dense
+          title={formik.values.editRestriction || "Detach network"}
+          disabled={isDisabled}
+        >
+          <Icon name="disconnect" />
+          <span>Detach</span>
+        </Button>
+      )}
+    </div>
+  );
+};
+
+export default NetworkDeviceActionButtons;
--- a/src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
+++ b/src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
@@ -1,0 +1,89 @@
+import ResourceLink from "components/ResourceLink";
+import type { FC } from "react";
+import type { LxdNicDevice, LxdNoneDevice } from "types/device";
+import NetworkDeviceAcls from "./NetworkDeviceAcls";
+import type { LxdNetwork } from "types/network";
+import NetworkSelector from "pages/projects/forms/NetworkSelector";
+import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import { supportsNicDeviceAcls } from "util/networks";
+import { isNoneDevice } from "util/devices";
+
+interface Props {
+  readOnly: boolean;
+  project: string;
+  formik: InstanceAndProfileFormikProps;
+  device: LxdNicDevice | LxdNoneDevice;
+  index: number;
+  managedNetworks: LxdNetwork[];
+  network?: LxdNetwork;
+}
+
+const NetworkDeviceContent: FC<Props> = ({
+  readOnly,
+  project,
+  formik,
+  device,
+  index,
+  managedNetworks,
+  network,
+}) => {
+  if (isNoneDevice(device)) {
+    return (
+      <span className="u-text--muted">
+        <i>detached</i>
+      </span>
+    );
+  }
+
+  if (readOnly) {
+    return (
+      <>
+        <div>Network</div>
+        <ResourceLink
+          type="network"
+          value={device.network}
+          to={`/ui/project/${encodeURIComponent(project ?? "")}/network/${encodeURIComponent(device.network)}`}
+        />
+        <NetworkDeviceAcls
+          project={project}
+          network={network}
+          device={device}
+          readOnly
+        />
+      </>
+    );
+  }
+
+  return (
+    <>
+      <NetworkSelector
+        value={device.network}
+        setValue={(value) => {
+          formik.setFieldValue(`devices.${index}.network`, value);
+
+          const selectedNetwork = managedNetworks.find((t) => t.name === value);
+
+          if (selectedNetwork && !supportsNicDeviceAcls(selectedNetwork)) {
+            formik.setFieldValue(
+              `devices.${index}["security.acls"]`,
+              undefined,
+            );
+          }
+        }}
+        id={`devices.${index}.network`}
+        name={`devices.${index}.network`}
+        managedNetworks={managedNetworks}
+      />
+      <NetworkDeviceAcls
+        project={project}
+        network={network}
+        device={device}
+        readOnly={readOnly}
+        formik={formik}
+        canSelectManualAcls={supportsNicDeviceAcls(network)}
+      />
+    </>
+  );
+};
+
+export default NetworkDeviceContent;
--- a/src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
+++ b/src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
@@ -15,16 +15,18 @@ import type { EditInstanceFormValues } from "pages/instances/EditInstance";
 import { getConfigurationRowBase } from "components/ConfigurationRow";
 import { getInheritedNetworks } from "util/configInheritance";
 import type { CustomNetworkDevice } from "util/formDevices";
-import { deduplicateName } from "util/formDevices";
+import {
+  addNicDevice,
+  deduplicateName,
+  focusNicDevice,
+} from "util/formDevices";
 import { isNicDeviceNameMissing } from "util/instanceValidation";
 import { ensureEditMode } from "util/instanceEdit";
 import { getExistingDeviceNames } from "util/devices";
-import { focusField } from "util/formFields";
 import { useNetworks } from "context/useNetworks";
 import { useProfiles } from "context/useProfiles";
 import NetworkDevice from "components/forms/NetworkDevicesForm/NetworkDevice";
-import ResourceLink from "components/ResourceLink";
-import ReadOnlyAclsList from "./ReadOnlyAclsList";
+import { getInheritedNetworkRow } from "components/forms/NetworkDevicesForm/InheritedNetworkRow";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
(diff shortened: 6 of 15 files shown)


src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,138 @@
...
+    override: (
+      <div>
+        <Tooltip
+          message="This network is inherited from a profile or project.
+To change it, edit it in the profile or project it originates from,
+or remove the originating item"
+          position="btm-left"
+        >
+          <Icon name="information" />
+        </Tooltip>
```
edlerd:
```suggestion
```
Now that we offer override, I think we don't need this tooltip any more.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,138 @@
...
+            appearance="base"
+            title={formik.values.editRestriction ?? "Override network"}
+            className="u-no-margin--top"
+            hasIcon
+            dense
+            disabled={!!formik.values.editRestriction}
+          >
+            <Icon name="edit" />
+          </Button>
+        )}
```
edlerd: I think we need some more interactions on inherited networks 
1. Without an existing override, create one. This is available with your current change
2. With an existing override, allow clearing the override. This is available with your current change. But we should label it "clear override" instead of "detach".
3. Detach the device completely. This creates a device with the same name as the inherited network, but using `type: none` and no other device config. We should have this when there is no override.

We might also want 3. on inherited device with an existing override. Then it would remove the override device and create the non device instead. Though, I am not sure if we should allow this shortcut or if the pattern clear override and then detach is easy enough.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -94,48 +94,25 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
-To change it, edit it in the profile or project it originates from,
-or remove the originating item"
-                position="btm-left"
-              >
-                <Icon name="information" />
-              </Tooltip>
-            ),
-          });
-        }),
+        ...inheritedNetworks.map((item) =>
```
edlerd, line 88: On QA I realized there is a slightly related bug with calculating the inherited devices, which this PR fixes: https://github.com/canonical/lxd-ui/pull/1540

src/components/forms/NetworkDevicesForm/NetworkDeviceAcls.tsx
```diff
@@ -48,6 +46,8 @@ const NetworkDeviceAcls: FC<Props> = ({
     return "Some ACLs are inherited from the network. They cannot be deselected here.";
   };
 
+  const index = formik?.values.devices.findIndex((t) => t.name === device.name);
```
edlerd: Should this be shared in `util/devices.tsx`?

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
kimanhou: 2. It now looks clunky. I tried putting Clear override on 2 lines but since it doesn't look like a button, it doesn't work. Any ideas ? @piperdeck ?
override-clear-override

3. I implemented it, however it led to a massive refactoring and I feel like I might have missed some things. I am not sure if the `Edit` button is correctly displayed when it should. Could you please help ? 🙏

edlerd: For 2: We can follow the root storage pattern. Check how create / clear override works on instances > disk. Use an "X" button to clear override without label. Then it should always fit into the available space.

For 3: When a network device is inherited but not overridden and we are in edit mode. Then we should display the pencil button to create override. Currently we can only create an override when we are in read mode of the instance.

I think visually the edit/detach buttons for inherited and non-inherited overrides should always be aligned and not move when opening edit mode or enabling/clearing an override for a inherited device.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+          onClickDetach={() => {
+            ensureEditMode(formik);
+            addNoneDevice(device.key, formik);
+            focusNetwork(
+              formik?.values.devices.findIndex((t) => t.name === device.key),
+            );
+          }}
+          isOverridden={false}
+          isInheritedRow
+        />
```
edlerd: If the network is inherited and detached (overridden with a `none` device). then we should use the button label `Reattach` and the same icon as on the disk section instead of the `Clear` label with the x icon.

Current clear icon:

image

reattach icon and label:

image

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+        project={project}
+        focusNetwork={focusNetwork}
+        removeNetwork={removeNetwork}
+        device={overrideDevice}
+        network={overrideNetwork}
+        isOverridden
+        isInheritedRow
+      />
+    ) : (
+      <div className="network-device">
```
edlerd: The layout looks nicely aligned now on desktop sizes. When shrinking the screen width, I see the buttons are being hidden in some intermediate widths, such as at `1000px` screen width. In those we don't need the screens to look super nice, but there should be a way to access all the buttons, where currently they are clipped away. I think this is an older bug that also occurs on the current LXD 6.5. do you think there is a way to make the vertical scroll also able to access those controls?

src/util/formDevices.tsx
```diff
@@ -232,3 +232,21 @@ export const deduplicateName = (
   }
   return candidate;
 };
+
+export const addDevice = ({
```
edlerd:
```suggestion
export const addNicDevice = ({
```
Maybe better as this is only for network devices?

src/sass/_forms.scss
```diff
@@ -64,6 +64,14 @@
         margin-right: $sph--large;
         max-width: 19rem;
       }
+
+      .action-buttons-container {
+        align-items: flex-start;
+        display: flex;
+        flex-direction: column;
+        max-width: 151px;
+        min-width: 151px;
```
edlerd:
```suggestion
        max-width: 8rem;
        min-width: 8rem;
```
I think this can be a bit smaller, and we should better use `rem`, avoid using pixel values.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -94,48 +92,25 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
+          getInheritedNetworkRow({
+            device: item,
+            project: project,
+            managedNetworks: managedNetworks,
+            formik: formik,
+            focusNetwork: focusNetwork,
```
edlerd: I think the focus is broken. When I create an override for an inherited network, the network selector should be selected, but doesn't receive focus.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,69 @@
+import { Button, Icon } from "@canonical/react-components";
+import type { FC } from "react";
+import type { InstanceAndProfileFormikProps } from "../instanceAndProfileFormValues";
+
+interface Props {
+  readOnly: boolean;
+  formik: InstanceAndProfileFormikProps;
+  onClickEdit: () => void;
+  onClickDetach: () => void;
```
edlerd:
```suggestion
  onEdit: () => void;
  onDetach: () => void;
```
A bit simpler naming.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <div />
+        <NetworkDeviceActionButtons
+          readOnly={readOnly}
+          formik={formik}
+          onClickEdit={() => {
+            ensureEditMode(formik);
+            addDevice({
+              formik,
+              deviceName: device.key,
+              deviceNetworkName: managedNetworks[0]?.name ?? "",
```
edlerd: I think instead of taking the first network, we should probably better take the network from the inherited device we create an override for.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+            focusNetwork(
+              formik?.values.devices.findIndex((t) => t.name === device.key),
+            );
+          }}
+          onClickDetach={() => {
+            ensureEditMode(formik);
+            addNoneDevice(device.key, formik);
+            focusNetwork(
+              formik?.values.devices.findIndex((t) => t.name === device.key),
+            );
```
edlerd:
```suggestion
```
The none device has no inputs, so we probably don't need to focus here.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+            ensureEditMode(formik);
+            addDevice({
+              formik,
+              deviceName: device.key,
+              deviceNetworkName: managedNetworks[0]?.name ?? "",
+            });
+
+            focusNetwork(
+              formik?.values.devices.findIndex((t) => t.name === device.key),
+            );
```
edlerd: Similar to the other comment, I think this should focus the network selector, currently does not work as I'd expect.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const readOnly = (formik.values as EditInstanceFormValues).readOnly;
+  const overrideDevice = formik.values.devices.find(
+    (t) => t.name === device.key,
+  ) as LxdNicDevice | LxdNoneDevice;
+
+  const overrideNetwork = managedNetworks.find(
+    (t) =>
+      t.name ===
+      (formik.values.devices.find((t) => t.name === device.key) as LxdNicDevice)
+        ?.network,
```
edlerd:
```suggestion
  const overrideNetwork = managedNetworks.find(
    (t) =>
      t.name === overrideDevice?.network,
```
I think we can use the variable from above here to simplify.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+  );
+
+  const getProfileName = (device: InheritedNetwork) => {
+    if (device.source.endsWith(" profile")) {
+      return device.source.substring(0, device.source.indexOf(" profile"));
```
edlerd: I think we should better enhance the `InheritedDevice` and `InheritedNetwork` types and functions to write the profile value into a dedicated field. So we avoid string matching here and can directly pick from `device.sourceProfile` or alike.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    return device.source;
+  };
+
+  return getConfigurationRowBase({
+    configuration: (
+      <>
+        <b>{device.key}</b>
```
edlerd: This should also be muted (and maybe strike through) if we have detached / the none device active. The device name will then be unavailable in the instance.

edlerd (commented):
This works really nicely, thanks for improving on network devices :+1: 

Did a full pass of QA and Code review. One potentially harder fix with the responsiveness, most of the other comments should be easy to resolve.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
kimanhou: The focus is indeed broken. But even if I fix it, it doesn't focus on network selector but on device name.
```
const focusNetwork = (id: number) => {
    focusField(`devices.${id}.name`);
  };
```

I have troubles fixing it because formik has another device, that I did not expected. As a result, indexes are off by 1. Do you know what this proxy device is ?
image

edlerd: I think we can change this to focus the network selector and not the device name. The device name is either auto-generated, so I suspect users are unlikely to want to change it. In this case of inherited devices, they can't change it anyway. So probably better to always focus on the network selector.

As for the mix of device types: All device types are in the formik state. Proxy devices are just one other possible type. I suspect you can use the `isNicDevice` helper to filter the array for relevant ones.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
kimanhou: I added this on network-device:
```
@include mobile-and-tablet {
  flex-direction: column-reverse;
}

```
It puts the action buttons at the top of the cell on small screens. 
image

There is a weird moment just before the change of layout when the buttons are barely accessible.
image

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
kimanhou: It seems `source` is not used 🤔

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
kimanhou: I created a map linking the device name with its index. It's clunky and I still don't fully understand what's going on but it seems to work (except when I add a new network but I'm working on it). Do you have a better idea ?

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
kimanhou: image

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
edlerd: I think `source` should have use cases. Like in the `Inherited disk devices` section.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
edlerd: Using `flex-direction: column-reverse` as suggested seems good solution to me. Though we should find a solution to prevent the edge case where the buttons are only half visible.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
edlerd: Yes, there is some bug in the current state of the PR when opening [this instance](https://lxd-ui-1539.demos.haus/ui/project/default/instance/DemoInstance/configuration/network) on the demo and then click `Reattache` `Edit` to override the inherited network and then `Attach network` to add another device. At this point, changing the network for either the override or the custom device don't get updated correctly. Also changing the device name is not reflected in the UI. I didn't look into the code how or what is mixed up. Suspecting the map that you mention might not get updated correctly.

kimanhou: I lost access to the demo server 😭
image

> At this point, changing the network for either the override or the custom device don't get updated correctly. Also changing the device name is not reflected in the UI.

Could you please re-explain this part ?


https://github.com/user-attachments/assets/ce171d21-436d-4143-b4ee-1ec86a48219a

edlerd: I think the order of interactions matter to reproduce the issue, see screencast below.

At 5s I'd expect the foo network to be chosen and at 8s I'd expect the selection works, but the focus jumps to the other interface.

https://github.com/user-attachments/assets/ba1f536a-98e3-4b68-b0aa-b3ff76745a75

kimanhou: Oooh I see, let me fix that

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
kimanhou: https://github.com/user-attachments/assets/a9c383cc-2109-4139-9197-a885cfeb008f

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -30,100 +31,59 @@ const NetworkDevice: FC<Props> = ({
...
-            />
-          </div>
-        ) : (
-          <>
-            <NetworkSelector
-              value={device.network}
-              setValue={(value) => {
-                formik.setFieldValue(`devices.${index}.network`, value);
+  useEffect(() => {
+    const observer = new ResizeObserver((entries) => {
```
edlerd: Better use the existing pattern with `useListener(window, someCallback, "resize", true);`. It has the advantage of including a debounce of the resize events.

Another idea: Can we capture this with a CSS breakpoint instead, using a media query?

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,137 @@
...
+          device={device.network}
+        />
+      </div>
+    ),
+    override: isOverridden ? (
+      <NetworkDevice
+        readOnly={readOnly}
+        formik={formik}
+        project={project}
+        focusNetwork={focusNetwork}
```
edlerd: The focus now works when going from read to edit mode. But when creating an override this should also focus the network selector. See last interaction in the screencast:


https://github.com/user-attachments/assets/46ebf88c-be9d-4c67-a495-089a8bf92b96

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,137 @@
...
+        device={overrideDevice}
+        index={index}
+        network={overrideNetwork}
+        isOverridden
+        isInheritedRow
+      />
+    ) : (
+      <div className="network-device">
+        <div />
+        <NetworkDeviceActionButtons
```
edlerd: Shall we add the detach button when an override exists? Currently, when an override exists, I have to first clear, and then I can detach the device. I think it would be nice to do that in one click.


https://github.com/user-attachments/assets/962228a1-3637-4d5e-a349-c427326bd163

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -30,100 +31,59 @@ const NetworkDevice: FC<Props> = ({
   removeNetwork,
   network,
   device,
+  isOverridden,
+  isInheritedRow,
 }) => {
+  const [isNarrow, setIsNarrow] = useState(false);
```
edlerd: The calculation for this value seems to be missing some edge cases. I think in the screencast below it shouldn't go to narrow.

https://github.com/user-attachments/assets/e703089f-e9f9-4f11-8a11-e52d53a1f199

src/sass/_forms.scss
```diff
@@ -59,11 +59,23 @@
...
+        flex-direction: column-reverse;
+      }
+
       > :first-child {
         flex-grow: 1;
         margin-right: $sph--large;
         max-width: 19rem;
       }
+
+      .action-buttons-container {
```
edlerd: Currently, can lead to slight misalignments like below. I think the ACL input is growing the left side.

image

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,86 @@
...
+import type { InstanceAndProfileFormikProps } from "../instanceAndProfileFormValues";
+
+interface Props {
+  readOnly: boolean;
+  formik: InstanceAndProfileFormikProps;
+  onEdit: () => void;
+  onDetach: () => void;
+  isOverridden?: boolean;
+  isInheritedRow?: boolean;
+  isInehritedAndDetached?: boolean;
```
edlerd:
```suggestion
  isInheritedAndDetached?: boolean;
```
Small typo

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,86 @@
...
+              : "disconnect"
+          }
+        />
+        <span>
+          {isOverridden
+            ? isInehritedAndDetached
+              ? "Reattach"
+              : "Clear"
+            : "Detach"}
+        </span>
```
edlerd: This is fairly hard to read and parse as a human. Same with the `getDetachTitle` and `getEditTitle` functions above. I have no immediate idea how to simplify. Maybe making the buttons more explicit with icon and title. Then apply the conditions to render or not render them. Like in [this file](https://github.com/canonical/ams/blob/e8a996451dc478ee3fcbd11197cec1bce4265de4/ui/templates/js/pages/instances/InstancesPage.tsx#L234-L302).  Though I see a slight disadvantage of it becoming more repetitive.

edlerd (commented):
Thanks for the changes, nice improvements :)

Some concerns and suggestions below.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
kimanhou: Both `useListener(window, ...)` and a CSS media query are tied to the viewport width, meaning that it will not be responsive to its immediate container. With this approach, if you reduce the main navigation sidebar, without the window resizing, the layout will not change.

However, I can wrap the `setIsNarrow` call inside my `ResizeObserver` with a debounce or throttle utility to prevent excessive re-renders on rapid size changes.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
kimanhou: https://github.com/user-attachments/assets/0ae23606-d3f0-4607-bf66-6f4701d1dd07

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
edlerd: The CSS media query could set a breakpoint where the table gets too narrow. I think all widths are static and independent of the content so we could define a fixed breakpoint where the flex direction changes to a column if we are below it. I didn't verify or test if this actually is the case, but if true this would be much simpler.

src/util/formDevices.tsx
```diff
@@ -232,3 +232,36 @@ export const deduplicateName = (
...
+  formik: InstanceAndProfileFormikProps;
+  deviceName: string;
+}) => {
+  const copy = [...formik.values.devices].filter((t) => t.name !== deviceName);
+  copy.push({
+    type: "none",
+    name: deviceName,
+  });
+  formik.setFieldValue("devices", copy);
+};
```
edlerd: I think we could enhance the `addNoneDevice` function above with the filter logic and use it instead. Then we'd not need this new function.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -62,28 +60,17 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
+    ensureEditMode(formik);
+    addNicDevice({
+      formik,
+      deviceName: deduplicateName("eth", 1, existingDeviceNames),
+      deviceNetworkName: managedNetworks[0]?.name ?? "",
     });
-    formik.setFieldValue("devices", copy);
 
-    focusNetwork(copy.length - 1);
+    focusNetwork(formik.values.devices.length);
```
edlerd: There is a race condition between the add function changing formik state and reading the device length. I think it would be safer to return the index from `addNicDevice`, as it knows the correct and updated length internally.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return getConfigurationRowBase({
+    configuration: (
+      <b
+        className={classnames({
+          "u-text--muted": isOverridden,
+          "u-text--line-through": isOverridden,
+        })}
+      >
+        {device.key}
+      </b>
```
edlerd, line 45: We should only mute and line through if this is overridden with the none device. When overriding with another network, this should be displayed normally.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+  formik,
+  index,
+}: Props): MainTableRow => {
+  const readOnly = (formik.values as EditInstanceFormValues).readOnly;
+  const isOverridden = formik.values.devices
+    .map((t) => t.name)
+    .includes(device.key);
+  const overrideDevice = formik.values.devices.find(
+    (t) => t.name === device.key,
+  ) as LxdNicDevice | LxdNoneDevice | undefined;
```
edlerd:
```suggestion
  const overrideDevice = formik.values.devices.find(
    (t) => t.name === device.key,
  ) as LxdNicDevice | LxdNoneDevice | undefined;
  const isOverridden = overrideDevice !== undefined;
```
A bit simpler.

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -1,129 +1,57 @@
...
-  focusNetwork: (id: number) => void;
-  removeNetwork: (id: number) => void;
-  device: LxdNicDevice;
+  device?: LxdNicDevice | LxdNoneDevice;
   network?: LxdNetwork;
+  inheritedDevice?: InheritedNetwork;
 }
 
 const NetworkDevice: FC<Props> = ({
   index,
```
edlerd: Can we remove the index here and in the children? We already pass the device, and I think in the nested children we can use the `getIndex` function instead of prop drilling to simplify the interfaces in between.
```suggestion
```

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,160 @@
...
+      addNoneDevice(inheritedDevice.key, formik);
+    } else {
+      removeNetwork({ formik, deviceName: device?.name || "" });
+    }
+  };
+
+  const onDetachOverridden = () => {
+    ensureEditMode(formik);
+    addNoneDevice(device?.name || "", formik);
+  };
```
edlerd:
```suggestion
  const onDetach = () => {
    ensureEditMode(formik);
    if (inheritedDevice) {
      addNoneDevice(inheritedDevice.key, formik);
    } else {
      removeNetwork({ formik, deviceName: device?.name || "" });
    }
  };
```
Now that the `addNoneDevice` covers both cases, I think we can generalize `onDetach` and remove the `onDetachOverridden`.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,160 @@
...
+    icon: "close",
+    text: "Clear",
+    title: "Clear override",
+  },
+  reattach: {
+    icon: "connected",
+    text: "Reattach",
+    title: "Reattach inherited network",
+  },
+};
```
edlerd: Naming is a bit unclear, `detachButtonConfig` does not imply this is the clear or reattach button. Same for the `getDetachButtonProps` and `detachIcon` below. We should find a name that covers all cases.

src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const NetworkDeviceContent: FC<Props> = ({
+  readOnly,
+  project,
+  formik,
+  device,
+  index,
+  managedNetworks,
+  network,
+}: Props) => {
```
edlerd:
```suggestion
}) => {
```
A bit shorter. I see some occasions of this, but it shouldn't be needed.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -94,50 +81,24 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
-                <Icon name="information" />
-              </Tooltip>
-            ),
-          });
-        }),
+        ...inheritedNetworks.map((item) =>
+          getInheritedNetworkRow({
+            device: item,
+            project,
+            managedNetworks: managedNetworks,
```
edlerd:
```suggestion
            managedNetworks,
```

src/sass/_forms.scss
```diff
@@ -59,21 +59,43 @@
...
         flex-grow: 1;
         margin-right: $sph--large;
-        max-width: 19rem;
+        min-width: 19rem;
       }
-    }
 
-    .acls-from-network {
-      margin-top: $spv--small;
+      .action-buttons-container {
```
edlerd:
```suggestion
      .network-device-actions {
```
A bit more concise naming.

src/sass/_forms.scss
```diff
@@ -59,21 +59,43 @@
...
+      }
 
-      .acl-chip {
-        max-width: 15rem;
+      &.narrow {
+        flex-direction: column-reverse;
+
+        > .action-buttons-container {
+          margin-bottom: $spv--medium;
+        }
```
edlerd:
```suggestion
```
Seems unused. We might need the bottom margin with the `@media screen and (width <= 1420px)` condition.

src/util/formDevices.tsx
```diff
@@ -232,3 +233,36 @@ export const deduplicateName = (
...
+  const copy = [...formik.values.devices];
+  copy.push({
+    type: "nic",
+    name: deviceName,
+    network: deviceNetworkName,
+  });
+  formik.setFieldValue("devices", copy);
+};
+
+export const focusNetwork = (id: number) => {
```
edlerd:
```suggestion
export const focusNicDevice = (id: number) => {
```
A bit more consistent naming.

src/util/formDevices.tsx
```diff
@@ -232,3 +233,36 @@ export const deduplicateName = (
...
+  focusField(`devices.${id}.network`);
+};
+
+export const removeNetwork = ({
```
edlerd:
```suggestion
export const removeNicDevice = ({
```
A bit more consistent.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
kimanhou: `removeNicDevice` ? In some cases, it removes a device from the formik. In other cases, it will add a `none` device.

edlerd: Yes, `removeNicDevice` sounds good to me :+1:

edlerd: Oh but wait, the reattach is not covered by that name, is it?

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
kimanhou: But how do you tell the difference between `Clear override` and `Detach inherited` ?

edlerd: Oh, that is a surprise, it is unexpected for the `onDetach` to trigger on clear override and on reattach. Maybe better to split them into dedicated buttons and config? Also in light of the trouble naming the `detachButtonConfig` might be more clear.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -0,0 +1,103 @@
...
+  formik: InstanceAndProfileFormikProps;
+}
+
+export const getInheritedNetworkRow = ({
+  device,
+  project,
+  managedNetworks,
+  formik,
+}: Props): MainTableRow => {
+  const readOnly = (formik.values as EditInstanceFormValues).readOnly;
```
edlerd:
```suggestion
```
I think this is only used in `NetworkDevice`, better move it over and simplify its interface.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
kimanhou: reattach is removing the `none` device 😅

src/components/forms/NetworkDevicesForm/NetworkDevice.tsx
```diff
@@ -1,129 +1,56 @@
...
+        {device && isNicDevice(device) && (
+          <NetworkDeviceContent
+            readOnly={readOnly}
+            project={project}
+            device={device}
+            formik={formik}
+            index={index}
+            managedNetworks={managedNetworks}
+            network={network}
+          />
```
edlerd: We discussed to display something here in case there is a none device, so the override column is not empty.

kimanhou: Yes, what do you think ?
"eth0 is detached" ?

edlerd: I think a bit shorter like "detached" in a muted font might work as well. Maybe consult with @piperdeck

kimanhou: What do you guys think ?

image

edlerd: Yes, this looks good to me and the experience on detaching is much more clear to me. Like you click detach on the inherited device and something happens that lets me think "this worked" :)

src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+  if (isNoneDevice(device)) {
+    return (
+      <span className="u-text--muted">
+        <i>detached</i>
+      </span>
+    );
+  }
+
+  if (!isNicDevice(device)) return null;
+
```
edlerd:
```suggestion
```
Is this needed? I think all cases of other devices than NIC type won't end up here.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+    return removeNicDeviceButtonConfig.detach;
+  };
+
+  const {
+    icon: detachIcon,
+    text: detachText,
+    title: detachTitle,
+  } = getRemoveNicDeviceButtonProps();
+
```
edlerd: This selecting of the props depending on the context with the `removeNicDeviceButtonConfig` seems like a lot of indirections. Defining this explicitly as 3 buttons with directly applied label / handlers / icon will be simpler and thus easier to maintain and understand?

We could define helper like `getButton = (label, icon, onClick) => {...}` to reduce the duplication, but probably also okay without it and accepting some duplication here.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,160 @@
...
+        title={formik.values.editRestriction || detachTitle}
+        disabled={!!formik.values.editRestriction}
+      >
+        <Icon name={detachIcon} />
+        <span>{detachText}</span>
+      </Button>
+
+      {isLocalNicOverride && (
+        <Button
+          className="delete-device u-no-margin--top"
```
edlerd: I think this `delete-device` class is not needed here and above either. The no margin top class overrides its effect.
```suggestion
          className="u-no-margin--top"
```

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+const NetworkDeviceActionButtons: FC<Props> = ({
+  readOnly,
+  formik,
+  index,
+  device,
+  inheritedDevice,
+}: Props) => {
+  const isInheritedNoOverride = inheritedDevice && !device;
+  const isLocalNicOverride = inheritedDevice && device && isNicDevice(device);
```
edlerd:
```suggestion
  const hasNicOverride = inheritedDevice && device && isNicDevice(device);
```

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const isLocalNoneOverride = inheritedDevice && device && isNoneDevice(device);
```
edlerd:
```suggestion
  const hasNoneOverride = inheritedDevice && device && isNoneDevice(device);
```

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const isLocalOnlyDevice = !inheritedDevice && device;
```
edlerd:
```suggestion
  const isLocal = !inheritedDevice && device;
```

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,179 @@
...
+}
+
+const NetworkDeviceActionButtons: FC<Props> = ({
+  readOnly,
+  formik,
+  index,
+  device,
+  inheritedDevice,
+}: Props) => {
+  const isInheritedNoOverride = inheritedDevice && !device;
```
edlerd:
```suggestion
  const isPurelyInherited = inheritedDevice && !device;
```

edlerd (dismissed):
LGTM, thanks for driving this to completion :)

Some non blocking suggestions for naming below.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const clearOverride = () => {
+    ensureEditMode(formik);
+    removeNicDevice({ formik, deviceName: device?.name || "" });
+  };
+
+  const detachInherited = () => {
+    ensureEditMode(formik);
+    if (inheritedDevice) {
+      addNoneDevice(inheritedDevice.key, formik);
+    }
```
edlerd:
```suggestion
    addNoneDevice(inheritedDevice.key, formik);
```
I think we don't need the conditional here?

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,179 @@
...
+            hasIcon
+            dense
+            title={formik.values.editRestriction || "Detach network"}
+            disabled={isDisabled}
+          >
+            <Icon name="disconnect" />
+            <span>Detach</span>
+          </Button>
+        </>
+      )}
```
edlerd, line 138:
```suggestion
      {hasNicOverride && (
          <Button
            className="u-no-margin--top"
            onClick={clearOverride}
            type="button"
            appearance="base"
            hasIcon
            dense
            title={formik.values.editRestriction || "Clear override"}
            disabled={isDisabled}
          >
            <Icon name="close" />
            <span>Clear</span>
          </Button>
        )}

      {(isPurelyInherited || hasNicOverride) && (
          <Button
            className="u-no-margin--top"
            onClick={detachInherited}
            type="button"
            appearance="base"
            hasIcon
            dense
            title={formik.values.editRestriction || "Detach network"}
            disabled={isDisabled}
          >
            <Icon name="disconnect" />
            <span>Detach</span>
          </Button>
      )}
```
A bit simpler

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
```diff
@@ -0,0 +1,179 @@
...
+  };
+
+  const detachOverridden = () => {
+    ensureEditMode(formik);
+    addNoneDevice(device?.name || "", formik);
+  };
+
+  return (
+    <div className="network-device-actions">
+      {showEditButton && (
```
edlerd:
```suggestion
      {(readOnly || isPurelyInherited) && (
```
We could inline the variable to be a bit more straight forward.

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
kimanhou: No, because `detachInherited` is always called when `inheritedDevice` exists. But technically, TypeScript wants to be sure `inheritedDevice` exists, so then it looks like this:
```
const detachInherited = () => {
    ensureEditMode(formik);
    addNoneDevice(inheritedDevice?.key || "", formik);
  };
```

Is it simpler ?

src/components/forms/NetworkDevicesForm/NetworkDeviceActionButtons.tsx
kimanhou: I like the format "Case 1: this button, Case 2: these 2 buttons ...". That way you know exactly what buttons should be there in each case, specifically when an inherited network is overridden by a nic device. It's clearer that if `hasNicOverride`, then there are 2 buttons (Clear and Detach)

edlerd (approved):
LGTM, thank you :)

PR #1332, opened 2025-06-25
feat(cluster-link) add cluster link crud WD-23025

edlerd:
## Done

- Interaction for cluster link crud

Fixes WD-23025

Draft state as this is still pending design and relies on https://github.com/canonical/lxd/pull/14884

## QA

1. Run the LXD-UI:
    - On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @Kxiru or @edlerd for access.
    - With a local copy of this branch, [build and run as described in the docs](https://github.com/canonical/lxd-ui/blob/main/CONTRIBUTING.md#setting-up-for-development).
2. Perform the following QA steps:
    - open the cluster link list
    - create a cluster link
    - delete a cluster link

### Cluster links

image

image

image

image

image

PR #1563, opened 2025-10-22
feat(cluster-link) add cluster link crud

edlerd:
## Done

- feat(placement-groups) add placement group management
- Follow up to https://github.com/canonical/lxd-ui/pull/1332

QA steps
* go to cluster > links
* create a cluster link token
* view cluster link list
* delete a cluster link

### DIFF
--- a/src/api/cluster-links.tsx
+++ b/src/api/cluster-links.tsx
@@ -1,0 +1,89 @@
+import { handleResponse } from "util/helpers";
+import type {
+  LxdClusterLink,
+  LxdClusterLinkCreated,
+  LxdClusterLinkState,
+} from "types/cluster";
+import type { LxdApiResponse } from "types/apiResponse";
+import { ROOT_PATH } from "util/rootPath";
+import { addEntitlements } from "util/entitlements/api";
+
+const clusterLinkEntitlements = ["can_edit", "can_delete"];
+
+export const fetchClusterLinks = async (
+  isFineGrained: boolean | null,
+): Promise<LxdClusterLink[]> => {
+  const params = new URLSearchParams();
+  params.set("recursion", "2");
+  addEntitlements(params, isFineGrained, clusterLinkEntitlements);
+
+  return fetch(`${ROOT_PATH}/1.0/cluster/links?${params.toString()}`)
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdClusterLink[]>) => {
+      return data.metadata;
+    });
+};
+
+export const fetchClusterLink = async (
+  link: string,
+  isFineGrained: boolean | null,
+): Promise<LxdClusterLink> => {
+  const params = new URLSearchParams();
+  params.set("recursion", "2");
+  addEntitlements(params, isFineGrained, clusterLinkEntitlements);
+
+  return fetch(
+    `${ROOT_PATH}/1.0/cluster/links/${encodeURIComponent(link)}?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdClusterLink>) => {
+      return data.metadata;
+    });
+};
+
+export const fetchClusterLinkState = async (
+  link: string,
+): Promise<LxdClusterLinkState> => {
+  return fetch(
+    `${ROOT_PATH}/1.0/cluster/links/${encodeURIComponent(link)}/state`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdClusterLinkState>) => {
+      return data.metadata;
+    });
+};
+
+export const createClusterLink = async (
+  body: string,
+): Promise<LxdClusterLinkCreated | null> => {
+  return fetch(`${ROOT_PATH}/1.0/cluster/links`, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body,
+  })
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdClusterLinkCreated | null>) => {
+      return data.metadata;
+    });
+};
+
+export const updateClusterLink = async (
+  link: string,
+  body: string,
+): Promise<void> => {
+  await fetch(`${ROOT_PATH}/1.0/cluster/links/${encodeURIComponent(link)}`, {
+    method: "PUT",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body,
+  }).then(handleResponse);
+};
+
+export const deleteClusterLink = async (link: string): Promise<void> => {
+  await fetch(`${ROOT_PATH}/1.0/cluster/links/${encodeURIComponent(link)}`, {
+    method: "DELETE",
+  }).then(handleResponse);
+};
--- a/src/pages/cluster/ClusterLinkAddresses.tsx
+++ b/src/pages/cluster/ClusterLinkAddresses.tsx
@@ -1,0 +1,38 @@
+import type { FC } from "react";
+import ExpandableList from "components/ExpandableList";
+import type { LxdClusterLink } from "types/cluster";
+import { Icon } from "@canonical/react-components";
+
+interface Props {
+  clusterLink: LxdClusterLink;
+}
+
+const ClusterLinkAddresses: FC<Props> = ({ clusterLink }) => {
+  const addresses = clusterLink.config["volatile.addresses"];
+  if (!addresses || addresses === "") {
+    return null;
+  }
+
+  return (
+    <ExpandableList
+      items={
+        addresses.split(",").map((address) => {
+          return (
+            <div key={address}>
+              <a
+                href={`https://${address}`}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {address}
+                <Icon className="external-link-icon" name="external-link" />
+              </a>
+            </div>
+          );
+        }) ?? []
+      }
+    />
+  );
+};
+
+export default ClusterLinkAddresses;
--- a/src/pages/cluster/ClusterLinkForm.tsx
+++ b/src/pages/cluster/ClusterLinkForm.tsx
@@ -1,0 +1,115 @@
+import { Form, Input, Label, RadioInput } from "@canonical/react-components";
+import type { FC } from "react";
+import GroupSelection from "pages/permissions/panels/GroupSelection";
+import { useAuthGroups } from "context/useAuthGroups";
+import type { FormikProps } from "formik/dist/types";
+
+export interface ClusterLinkFormValues {
+  name: string;
+  description?: string;
+  token?: string;
+  tokenType?: "generate" | "consume";
+  authGroups: string[];
+  isCreating: boolean;
+  initialAuthGroups?: string[];
+}
+
+interface Props {
+  formik: FormikProps<ClusterLinkFormValues>;
+}
+
+const ClusterLinkForm: FC<Props> = ({ formik }) => {
+  const { data: authGroups = [] } = useAuthGroups();
+
+  const selectedGroups = new Set(formik.values.authGroups);
+  const initial = new Set(formik.values.initialAuthGroups);
+  const removed = [...initial].filter((g) => !selectedGroups.has(g));
+  const added = [...selectedGroups].filter((g) => !initial.has(g));
+  const modifiedGroups = new Set([...removed, ...added]);
+
+  return (
+    <Form onSubmit={formik.handleSubmit}>
+      {/* hidden submit to enable enter key in inputs */}
+      <Input type="submit" hidden value="Hidden input" />
+      {formik.values.isCreating && (
+        <Input
+          {...formik.getFieldProps("name")}
+          type="text"
+          label="Name"
+          placeholder="Enter name"
+          required
+          autoFocus
+          error={formik.touched.name ? formik.errors.name : null}
+        />
+      )}
+      <Input
+        {...formik.getFieldProps("description")}
+        type="text"
+        label="Description"
+        placeholder="Enter description"
+      />
+      {formik.values.isCreating && (
+        <>
+          <div className="u-sv1">
+            <RadioInput
+              inline
+              labelClassName="margin-right--large"
+              label="Generate token"
+              checked={formik.values.tokenType === "generate"}
+              onChange={() => {
+                formik.setFieldValue("tokenType", "generate");
+              }}
+            />
+            <RadioInput
+              inline
+              label="Consume token"
+              checked={formik.values.tokenType === "consume"}
+              onChange={() => {
+                formik.setFieldValue("tokenType", "consume");
+              }}
+            />
+          </div>
+          {formik.values.tokenType === "consume" && (
+            <Input
+              {...formik.getFieldProps("token")}
+              type="text"
+              label="Token"
+              placeholder="Enter token"
+            />
+          )}
+        </>
+      )}
+      <Label className="u-sv-2">Auth groups</Label>
+      <p className="u-text--muted u-sv-1">
+        Control access for incoming requests through this cluster link.
+      </p>
+      <GroupSelection
+        groups={authGroups}
+        modifiedGroups={modifiedGroups}
+        parentItemName="cluster link"
+        selectedGroups={selectedGroups}
+        setSelectedGroups={(val, isUnselectAll) => {
+          if (isUnselectAll) {
+            formik.setFieldValue("authGroups", []);
+          } else {
+            formik.setFieldValue("authGroups", val);
+          }
+        }}
+        toggleGroup={(group) => {
+          const currentGroups = formik.values.authGroups;
+          if (currentGroups.includes(group)) {
+            formik.setFieldValue(
+              "authGroups",
+              currentGroups.filter((g) => g !== group),
+            );
+          } else {
+            formik.setFieldValue("authGroups", [...currentGroups, group]);
+          }
+        }}
+        scrollDependencies={[formik]}
+      />
+    </Form>
+  );
+};
+
+export default ClusterLinkForm;
--- a/src/pages/cluster/ClusterLinkForm.tsx
+++ b/src/pages/cluster/ClusterLinkForm.tsx
@@ -26,6 +26,7 @@ const ClusterLinkForm: FC<Props> = ({ formik }) => {
   const removed = [...initial].filter((g) => !selectedGroups.has(g));
   const added = [...selectedGroups].filter((g) => !initial.has(g));
   const modifiedGroups = new Set([...removed, ...added]);
+  const hasName = formik.values.name !== "";
 
   return (
     <Form onSubmit={formik.handleSubmit}>
@@ -47,10 +48,23 @@ const ClusterLinkForm: FC<Props> = ({ formik }) => {
         type="text"
         label="Description"
         placeholder="Enter description"
+        disabled={!hasName}
+        title={
+          hasName
+            ? undefined
+            : "Please enter a name before adding a description"
+        }
       />
       {formik.values.isCreating && (
         <>
-          <div className="u-sv1">
+          <div
+            className="u-sv1"
+            title={
+              hasName
+                ? undefined
+                : "Please enter a name before selecting token options"
+            }
+          >
             <RadioInput
               inline
               labelClassName="margin-right--large"
@@ -59,6 +73,7 @@ const ClusterLinkForm: FC<Props> = ({ formik }) => {
               onChange={() => {
                 formik.setFieldValue("tokenType", "generate");
               }}
+              disabled={!hasName}
             />
             <RadioInput
               inline
@@ -67,6 +82,7 @@ const ClusterLinkForm: FC<Props> = ({ formik }) => {
               onChange={() => {
                 formik.setFieldValue("tokenType", "consume");
               }}
+              disabled={!hasName}
             />
           </div>
           {formik.values.tokenType === "consume" && (
@@ -80,7 +96,7 @@ const ClusterLinkForm: FC<Props> = ({ formik }) => {
         </>
       )}
       <Label className="u-sv-2">Auth groups</Label>
-      <p className="u-text--muted u-sv-1">
+      <p className="u-text--muted u-sv-1 p-text--small">
         Control access for incoming requests through this cluster link.
       </p>
       <GroupSelection
@@ -107,6 +123,7 @@ const ClusterLinkForm: FC<Props> = ({ formik }) => {
           }
         }}
         scrollDependencies={[formik]}
+        disabled={!hasName}
       />
     </Form>
   );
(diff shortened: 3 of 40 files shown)


src/api/cluster-links.tsx
```diff
@@ -0,0 +1,89 @@
...
+import type {
+  LxdClusterLink,
+  LxdClusterLinkCreated,
+  LxdClusterLinkState,
+} from "types/cluster";
+import type { LxdApiResponse } from "types/apiResponse";
+import { ROOT_PATH } from "util/rootPath";
+import { addEntitlements } from "util/entitlements/api";
+
+const clusterLinkEntitlements = ["can_edit", "can_delete"];
```
kimanhou, line 11:
```suggestion
const CLUSTER_LINK_ENTITLEMENTS = ["can_edit", "can_delete"];
```

If this constant is just for this file, then the cluster link prefix might not be needed but I don't have a strong opinion on this. However, let's use all capital for constants.

src/pages/cluster/actions/CreateClusterLinkBtn.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Button
+      appearance="positive"
+      className="u-no-margin--bottom"
+      hasIcon
+      onClick={panelParams.openCreateClusterLink}
+      title={disableReason}
+      disabled={disableReason !== undefined}
+    >
+      <Icon name="plus" light />
```
kimanhou, line 22: Sometimes we the `+` icon with `light`, sometimes without. Any reason for that ?

src/pages/cluster/actions/EditClusterLinksBtn.tsx
```diff
@@ -0,0 +1,37 @@
...
+const EditClusterLinkBtn: FC<Props> = ({ clusterLink }) => {
+  const { canEditClusterLink } = useClusterLinkEntitlements();
+  const panel = usePanelParams();
+
+  const canEdit = canEditClusterLink(clusterLink);
+
+  return (
+    <Button
+      appearance={"base"}
+      className={"u-no-margin--bottom"}
```
kimanhou:
```suggestion
      appearance="base"
      className="u-no-margin--bottom"
```

No need for curl brackets around constants

src/pages/cluster/actions/DeleteClusterLinkBtn.tsx
```diff
@@ -0,0 +1,85 @@
...
+              type="cluster-link"
+              value={clusterLink.name}
+              to={`${ROOT_PATH}/ui/cluster/links`}
+            />{" "}
+            deleted.
+          </>,
+        );
+      })
+      .catch((e) => {
+        setLoading(false);
```
kimanhou: I know that most our components are built this way, but should `setLoading(false);` be in `finally` ?

src/pages/cluster/actions/EditClusterLinksBtn.tsx
```diff
@@ -0,0 +1,37 @@
...
+import type { LxdClusterLink } from "types/cluster";
+import { useClusterLinkEntitlements } from "util/entitlements/cluster-links";
+
+interface Props {
+  clusterLink: LxdClusterLink;
+}
+
+const EditClusterLinkBtn: FC<Props> = ({ clusterLink }) => {
+  const { canEditClusterLink } = useClusterLinkEntitlements();
+  const panel = usePanelParams();
```
kimanhou:
```suggestion
  const panelParams = usePanelParams();
```

In the rest of the codebase, we call it `panelParams`

src/pages/cluster/modals/CreateClusterLinkModal.tsx
```diff
@@ -0,0 +1,130 @@
...
+  token: string;
+  linkName: string;
+}
+
+const CreateClusterLinkModal: FC<Props> = ({ onClose, token, linkName }) => {
+  const [isConfirmed, setConfirmed] = useState(false);
+  const [howToUseActiveTab, setHowToUseActiveTab] = useState("ui-tab");
+
+  return (
+    <Modal
```
kimanhou, line 26: Do you want to match exactly the beautiful `CreateIdentityModal` ? If so, here are the differences:

- tooltipMessage="Copy token"
- name of the cluster link in the title
- onClose not called in Modal's `close` (but I'm not sure what are the consequences)

src/pages/cluster/panels/EditClusterLinkPanel.tsx
```diff
@@ -0,0 +1,129 @@
...
+      isCreating: false,
+      initialAuthGroups: identity?.groups ?? [],
+    },
+    enableReinitialize: true,
+    onSubmit: (values) => {
+      const payload = {
+        ...clusterLink,
+        name: clusterLink?.name,
+        description: values.description,
+      };
```
kimanhou, line 56:
```suggestion
      const payload = {
        ...clusterLink,
        description: values.description,
      };
```

The name should already be included in the spreading, no ?

src/pages/cluster/panels/CreateClusterLinkPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+            title={error.title}
+            severity="negative"
+            onDismiss={() => {
+              setError(null);
+            }}
+          >
+            {error.message}
+          </Notification>
+        )}
+      </Row>
```
kimanhou, line 130: Any reason to use a `Notification` in a `Row` here and a `NotificationRow` in `EditClusterLinkPanel` ?

src/pages/cluster/ClusterLinkAddresses.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface Props {
+  clusterLink: LxdClusterLink;
+}
+
+const ClusterLinkAddresses: FC<Props> = ({ clusterLink }) => {
+  return (
+    <ExpandableList
+      items={
+        clusterLink.config["volatile.addresses"]?.split(",").map((address) => {
```
kimanhou:
```suggestion
  const addresses = clusterLink?.config["volatile.addresses"]?.split(",")
    .map((a) => a.trim())
    .filter(Boolean);

  if (addresses.length === 0) {
    return <span className="u-text--muted">No addresses available</span>;
  }
```

Just to be safe ...
If `clusterLink.config["volatile.addresses"]` is an empty string, then the `split` returns `[""]` and we might end up with an empty tag.

src/pages/cluster/ClusterLinkForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+      />
+      {formik.values.isCreating && (
+        <>
+          <div className="u-sv1">
+            <RadioInput
+              inline
+              labelClassName="margin-right"
+              label="Generate token"
+              checked={formik.values.tokenType === "generate"}
+              onClick={() => {
```
kimanhou:
```suggestion
              onChange={() => {
```

src/pages/cluster/ClusterLinkStatus.tsx
```diff
@@ -0,0 +1,53 @@
...
+      {
+        Reachable: "status-succeeded-small",
+        Unreachable: "status-failed-small",
+        Pending: "status-queued-small",
+      }[status] ?? ""
+    );
+  };
+
+  const getStatus = (): StatusCaption => {
+    if (identity?.type.includes("(pending)")) {
```
kimanhou:
```suggestion
    if (identity?.type?.toLocaleLowerCase().includes("(pending)")) {
```

Should this be case insensitive ?

src/pages/cluster/ClusterLinkStatus.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const getIconNameForStatus = (status: StatusCaption) => {
+    return (
+      {
+        Reachable: "status-succeeded-small",
+        Unreachable: "status-failed-small",
+        Pending: "status-queued-small",
+      }[status] ?? ""
+    );
+  };
```
kimanhou: Should this function be in a helper, as it's being recreated at every render ?

src/pages/cluster/ClusterLinkStatus.tsx
```diff
@@ -0,0 +1,53 @@
...
+import type { LxdIdentity } from "types/permissions";
+import type { LxdClusterLink } from "types/cluster";
+import { useClusterLinkState } from "context/useClusterLinks";
+
+interface Props {
+  link: LxdClusterLink;
+  identity?: LxdIdentity;
+}
+
+type StatusCaption = "Pending" | "Reachable" | "Unreachable";
```
kimanhou: Can this be moved to `types/cluster.d.ts` ?

src/pages/cluster/ClusterLinkStatus.tsx
```diff
@@ -0,0 +1,53 @@
...
+  if (isLoading) {
+    return null;
+  }
+
+  const status = getStatus();
+
+  return (
+    <>
+      <Icon name={getIconNameForStatus(status)} className="status-icon" />
+      {status}
```
kimanhou:
```suggestion
      const STATUS_ICONS: Record<StatusCaption, string> = {
  Reachable: "status-succeeded-small",
  Unreachable: "status-failed-small",
  Pending: "status-queued-small",
};

<Icon
        name={STATUS_ICONS[status]}
        className="status-icon"
      />
```

Just a suggestion

src/pages/cluster/CreateClusterLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+  });
+  const [token, setToken] = useState("");
+  const [linkName, setLinkName] = useState("");
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <CreateClusterLinkModal
+            onClose={closePortal}
```
kimanhou:
```suggestion
            onClose={() => {
              closePortal();
              setToken("");
              setLinkName("");
            }}
```

Should we clean up on close ?

src/pages/cluster/ClusterLinkList.tsx
```diff
@@ -0,0 +1,228 @@
...
+                <TablePagination
+                  data={sortedRows}
+                  id="pagination"
+                  itemName="cluster link"
+                  className="u-no-margin--top"
+                  aria-label="Table pagination control"
+                >
+                  <MainTable
+                    id="cluster-link-table"
+                    className="cluster-link-table"
```
kimanhou:
```suggestion
                    id="cluster-link-table-id"
                    className="cluster-link-table-classname"
```
Nitpick: I don't like when `id` and `className` are the same because then I am never sure if the CSS is correct.

src/sass/_cluster_list.scss
```diff
@@ -99,3 +99,9 @@
     }
   }
 }
+
+.cluster-link-table {
+  .status-header {
+    padding-left: 1.8rem;
```
kimanhou, line 105: This is the 3rd time that this magic number appears in our CSS. Time to store it in a variable with a meaningful name ?

src/pages/cluster/panels/CreateClusterLinkPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+import ResourceLink from "components/ResourceLink";
+import type { ClusterLinkFormValues } from "pages/cluster/ClusterLinkForm";
+import ClusterLinkForm from "pages/cluster/ClusterLinkForm";
+import { ROOT_PATH } from "util/rootPath";
+
+interface Props {
+  onSuccess: (identityName: string, token: string) => void;
+}
+
+const CreateClusterLinkPanel: FC<Props> = ({ onSuccess }) => {
```
kimanhou, line 30: Similar to `CreateGroupPanel`, should all fields be disabled until a valid name is entered ?

src/pages/cluster/ClusterLinkForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <Input
+        {...formik.getFieldProps("description")}
+        type="text"
+        label="Description"
+        placeholder="Enter description"
+      />
+      {formik.values.isCreating && (
+        <>
+          <div className="u-sv1">
+            <RadioInput
```
kimanhou, line 68: When Generate token radio button is selected and you click on it, it looks weird: it looks like a double outline, that ends up being truncated on the left.

Image

src/api/cluster-links.tsx
edlerd: It is more a consistency between the API files that we use camel case for this constant.

src/pages/cluster/actions/CreateClusterLinkBtn.tsx
edlerd: Within a button with a positive appearance, it needs to be the light version of the icon to be white and aligned with the text color.

src/pages/cluster/actions/DeleteClusterLinkBtn.tsx
edlerd: I think for delete it doesn't make a big difference, because the item is removed and not rendered on success. I'll move it over to finally just in case ;)

src/pages/cluster/modals/CreateClusterLinkModal.tsx
edlerd: all good observations :+1:

src/pages/cluster/panels/CreateClusterLinkPanel.tsx
edlerd: We should use a dedicated `error` state in both panels. This is the better solution. Otherwise we show the error in the panel and in the main page.

src/pages/cluster/ClusterLinkAddresses.tsx
edlerd: good catch. Adding an early exit for that case.

src/pages/cluster/ClusterLinkStatus.tsx
edlerd: I'll move it out to the higher scope.

src/pages/cluster/ClusterLinkList.tsx
edlerd: I'll keep the short class name and append the ID for the id.

src/sass/_cluster_list.scss
edlerd: I am not sure there is a semantic overlap to extract a variable for it.

src/pages/cluster/actions/DeleteClusterLinkBtn.tsx
```diff
@@ -0,0 +1,85 @@
...
+      .finally(() => {
+        setLoading(false);
+        queryClient.invalidateQueries({
+          queryKey: [queryKeys.cluster, queryKeys.links],
+        });
+      });
+  };
+
+  return (
+    <ConfirmationButton
```
kimanhou, line 59: This should have the className `has-icon`. Currently, the icon is not centered in the button, and it shows on hover.

Image

src/pages/cluster/ClusterLinkStatus.tsx
```diff
@@ -0,0 +1,35 @@
...
+const STATUS_ICONS: Record<StatusCaption, string> = {
+  Reachable: "status-succeeded-small",
+  Unreachable: "status-failed-small",
+  Pending: "status-queued-small",
+};
+
+const ClusterLinkStatus: FC<Props> = ({ link, identity }) => {
+  const { data: state, isLoading } = useClusterLinkState(link.name);
+  if (isLoading) {
+    return null;
```
kimanhou: I am wondering if returning `null` is the UX. Would a spinner be better ?
It currently looks empty then all of a sudden a status appears.

https://github.com/user-attachments/assets/ca88aebc-693a-4c51-8967-0a9ff4afd9d0

src/pages/cluster/panels/CreateClusterLinkPanel.tsx
edlerd: Here we can't move to another section to get lost. And when un-focusing the name an error pops up. I think that is okay as is.

src/pages/cluster/ClusterLinkForm.tsx
edlerd: Good catch!

src/pages/cluster/ClusterLinkForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+              {...formik.getFieldProps("token")}
+              type="text"
+              label="Token"
+              placeholder="Enter token"
+            />
+          )}
+        </>
+      )}
+      <Label className="u-sv-2">Auth groups</Label>
+      <p className="u-text--muted u-sv-1">
```
kimanhou:
```suggestion
      <p className="u-text--muted p-text--small u-sv-1">
```
Should this be in small size, as it resembles a helptext ?

src/pages/cluster/modals/CreateClusterLinkModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+      className="create-cluster-link"
+      title={
+        <>
+          Cluster link{" "}
+          <ResourceLabel type="cluster-link" value={linkName} bold truncate />{" "}
```
kimanhou, line 31: The icon looks weird for some reason: too small and maybe too low ? It looks better when an identity is created, I don't have any idea how to improve it. Wdyt ?

Image

kimanhou (commented):
I tried doing QA but I get an error `Invalid trust token: illegal base64 data at input byte 476`. Here is what I did:

- create 2 clusters
- on one of the cluster, create a cluster link and generate a token
- on the other cluster, create a cluster link with the same name and consume the token

Is it the correct way to create a cluster link ?

Image

src/pages/cluster/ClusterLinkForm.tsx
edlerd: Yes, good idea!

src/pages/cluster/modals/CreateClusterLinkModal.tsx
edlerd: True, I added some custom CSS to improve the styling here and also on identity creation.

edlerd:
> I tried doing QA but I get an error `Invalid trust token: illegal base64 data at input byte 476`. Here is what I did:
> 
> * create 2 clusters
> * on one of the cluster, create a cluster link and generate a token
> * on the other cluster, create a cluster link with the same name and consume the token
> 
> Is it the correct way to create a cluster link ?

This is the right way, yes. The links can have different names, but that should not lead to your error. Did you maybe copy the token with spaces or incomplete?

edlerd:
@kimanhou the easiest setup I had to test it:
1. Have 1 container with latest edge lxd as a one node clustered LXD.
2. Use the IP of that container as a backend for dotrun in `.env.local` as `LXD_UI_BACKEND_IP`
3. Create the cluster link with the UI
4. Copy the container and start the copy on your host LXD UI.
5. Stop dotrun and switch your backend to the copy in `.env`, then restart dotrun.
6. Now you can use the token generated in 3 to create a cluster link on the copy.
7. The link should show immediately as status "Reachable" and type "Bidirectional".

kimanhou:
Yaaaaaaay

image

kimanhou:
I think I know why it wasn't working for me before: it's probably because I was in a double nested VM. The error was not very clear, but it's a comment for backend.

When creating the cluster link, it takes a while. During the wait, "Create cluster link" button is correctly showing a spinner but you can still interact with the side panel. You can click on "Cancel", which closes the side panel, but doesn't cancel the request. I am wondering if the whole side panel, including the Cancel button, should be disabled while we wait. But if the request hangs forever then there is no choice other than refreshing the whole page, which is a terrible UX too. Wdyt ?

Other than that, all good for me ! Great work !

edlerd:
> When creating the cluster link, it takes a while. During the wait, "Create cluster link" button is correctly showing a spinner but you can still interact with the side panel. You can click on "Cancel", which closes the side panel, but doesn't cancel the request. I am wondering if the whole side panel, including the Cancel button, should be disabled while we wait. But if the request hangs forever then there is no choice other than refreshing the whole page, which is a terrible UX too. Wdyt ?

You are raising a fair point with the wait time. I think the best would be to have the `POST` endpoint for cluster links be async and return immediately to unblock the user in the UI. Then we can inform the user the creation was started and track the operation and report on it finishing with either error or success. I'll ping this to @kadinsayani and @tomponline as a suggestion.

Though suggest we merge as is and do the operation as a follow up.

edlerd:
> You are raising a fair point with the wait time. I think the best would be to have the `POST` endpoint for cluster links be async and return immediately to unblock the user in the UI. Then we can inform the user the creation was started and track the operation and report on it finishing with either error or success. I'll ping this to @kadinsayani and @tomponline as a suggestion.
> 
> Though suggest we merge as is and do the operation as a follow up.

Ah wait, that doesn't work with the operation, because we do need to report on the token to the user. I think disabling the cancel or side panel closing could work here.

tomponline:
> Ah wait, that doesn't work with the operation, because we do need to report on the token to the user. I think disabling the cancel or side panel closing could work here.

You can use metadata to return tokens to the end user from an operation - so an operation is a possibility indeed.

edlerd:
> > Ah wait, that doesn't work with the operation, because we do need to report on the token to the user. I think disabling the cancel or side panel closing could work here.
> 
> You can use metadata to return tokens to the end user from an operation - so an operation is a possibility indeed.

Thanks. I was typing faster than thinking. Due to the nature of the interaction, the user has to wait for the token to continue their workflow. Therefore, I think this is a better UX to keep the user busy waiting here. And without operation (as is), is actually correct.

kimanhou:
So just disabling the Cancel button while we wait ? Or we merge it as it is ?

edlerd:
> So just disabling the Cancel button while we wait ? Or we merge it as it is ?

I think I want to do small adjustments to the form:

1. As suggested, disable cancel, or if we keep it, make it cancel the request.
2. I think your previous suggestion of forcing name input by disabling other inputs before appears to be better. On manual testing I fell into the layout shift because of the "Name is required" error popping up and shifting other fields down.

edlerd:
@kimanhou ready for review from my side.

kimanhou:
Should we do something for extra long names ?
image

Otherwise all good !

edlerd:
> Should we do something for extra long names ? 

I think the layout is ok, not perfect, but a user might want to see the whole details here if those names are auto-generated by some system with multiple pieces of information in them, like data center, region, etc.

edlerd:
I guess we could clamp at a single line and only reveal detail on hover though. Let me adjust it.

edlerd:
Good catch @kimanhou 
I added truncation for long cluster link names in the overview table and the edit panel. Please have another review.

kimanhou (approved):
Let's goooooooo

Issue #1508, opened 2025-09-18
[Feature Request] Auto generate SSH key during instance created


sanskritkm:
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

it speed up time to have each instances ssh key or we have option to insert ourself. This make every instance created has unique ssh by default because it has option auto generate.

<img width="641" height="535" alt="Image" src="https://github.com/user-attachments/assets/31e111aa-d57d-4ea5-ac05-689918a60953" />

### Describe the solution you'd like

Option to one-click to auto generate ssh key insert into newly create instance and auto download private key into local pc.

PR #1572, opened 2025-10-24
feat support ssh key generation in UI

omarelkashef:
## Done

- Feat: Support SSH key pair generation and download in the UI. Addresses #1508 

## QA

1. Run the LXD-UI:
    - Go to an instance Configuration tab.
    - Click on "+ New SSH key" button.
    - Click on "Generate key" option.
    - Click on "Download private key" button.
    - Click on "Add key" button.
    - Click on "Save 1 change" button.
    - Make sure your new public key is added to the instance and your private key is downloaded on your device.
    - SSH into your instance using your new key pair and ensure you can log in.

## Screenshots

image
image
image
image

### DIFF
--- a/src/components/forms/SshKeyAddModal.tsx
+++ b/src/components/forms/SshKeyAddModal.tsx
@@ -2,6 +2,7 @@ import type { FC, FormEvent, ReactNode } from "react";
 import { useState } from "react";
 import { Button, Form, Input, Label, Modal } from "@canonical/react-components";
 import AutoExpandingTextArea from "components/AutoExpandingTextArea";
+import SshKeyGenerateAndDownload from "./SshKeyGenerateAndDownload";
 
 interface Props {
   initialName: string;
@@ -9,7 +10,12 @@ interface Props {
   onClose: () => void;
 }
 
-type SourceType = "clipboard" | "fileUpload" | "github" | "launchpad";
+type SourceType =
+  | "clipboard"
+  | "fileUpload"
+  | "generate"
+  | "github"
+  | "launchpad";
 
 const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
   const [name, setName] = useState(initialName);
@@ -20,6 +26,7 @@ const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
     launchpad: "",
     fileUpload: "",
     clipboard: "",
+    generate: "",
   });
 
   const getFingerprint = () => {
@@ -32,6 +39,8 @@ const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
         return source.fileUpload;
       case "clipboard":
         return source.clipboard;
+      case "generate":
+        return source.generate;
     }
   };
 
@@ -79,6 +88,10 @@ const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
     document.getElementById("ssh-file-upload")?.click();
   };
 
+  const onSSHKeyGenerate = (key: string) => {
+    setSource({ ...source, generate: key });
+  };
+
   return (
     <Modal close={onClose} title="Add SSH key" className="ssh-key-add-modal">
       <Form onSubmit={handleAdd}>
@@ -112,6 +125,7 @@ const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
             aria-label="Source"
           >
             {segmentButton("clipboard", "Clipboard")}
+            {segmentButton("generate", "Generate key")}
             {segmentButton("fileUpload", "File upload")}
             {segmentButton("github", "Github")}
             {segmentButton("launchpad", "Launchpad")}
@@ -122,7 +136,7 @@ const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
           {segmentContent(
             "clipboard",
             <AutoExpandingTextArea
-              placeholder="ssh-rsa ..."
+              placeholder="ecdsa-sha2-nistp384 ..."
               help="Paste the contents of the public key file (.pub)"
               className="u-break-all"
               autoFocus
@@ -131,6 +145,14 @@ const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
               onChange={(e) => {
                 setSource({ ...source, clipboard: e.target.value });
               }}
+            />,
+          )}
+
+          {segmentContent(
+            "generate",
+            <SshKeyGenerateAndDownload
+              keyName={name}
+              setSSHPublicKey={onSSHKeyGenerate}
             />,
           )}
 
--- a/src/components/forms/SshKeyGenerateAndDownload.tsx
+++ b/src/components/forms/SshKeyGenerateAndDownload.tsx
@@ -1,0 +1,100 @@
+import { useEffect, useState, type FC } from "react";
+import { Button } from "@canonical/react-components";
+
+interface Props {
+  keyName: string;
+  setSSHPublicKey: (value: string) => void;
+}
+
+const SshKeyGenerateAndDownload: FC<Props> = ({ keyName, setSSHPublicKey }) => {
+  const [privateKey, setPrivateKey] = useState<CryptoKey>();
+  const [publicKey, setPublicKey] = useState<CryptoKey>();
+  const [isDownloaded, setIsDownloaded] = useState(false);
+  const generateSSHKey = async () => {
+    setSSHPublicKey("");
+    const key = await window.crypto.subtle.generateKey(
+      {
+        name: "ECDSA",
+        namedCurve: "P-384",
+      },
+      true,
+      ["sign", "verify"],
+    );
+    setPrivateKey(key.privateKey);
+    setPublicKey(key.publicKey);
+  };
+
+  useEffect(() => {
+    generateSSHKey();
+  }, []);
+
+  const exportEcdsaPublicKeyAsOpenSSH = (publicKey: ArrayBuffer) => {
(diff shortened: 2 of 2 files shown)


src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <Button type="button" onClick={generateAndDownloadSSHKey}>
+        Generate and download key pair
+      </Button>
+      <AutoExpandingTextArea
+        placeholder="Your auto-generated public key"
+        className="u-break-all"
+        readOnly={true}
+        cols={50}
+        value={source.generate}
+      />
```
edlerd: I think we shouldn't use the placeholder, but put the public key as the content and set it disabled.

Actually, I am not sure if we should display this at all. We might just display some indicator, that the key was generated and is ready to be used..

src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    const privateKeyPEM = convertToPEMFormat(privateKeyData, "EC PRIVATE KEY");
+    const publicKeyOpenSsh = exportEcdsaPublicKeyAsOpenSSH(publicKeyDataRaw);
+
+    setSSHKey(publicKeyOpenSsh);
+    downloadPEMFile(privateKeyPEM, "lxd_ecdsa.key");
+  };
+
+  return (
+    <div>
```
edlerd:
```suggestion
    <>
```
I think we can keep the DOM a bit more slim.

src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const publicKeyDataRaw = await window.crypto.subtle.exportKey(
+      "raw",
+      keyPair.publicKey,
+    );
+
+    const privateKeyPEM = convertToPEMFormat(privateKeyData, "EC PRIVATE KEY");
+    const publicKeyOpenSsh = exportEcdsaPublicKeyAsOpenSSH(publicKeyDataRaw);
+
+    setSSHKey(publicKeyOpenSsh);
+    downloadPEMFile(privateKeyPEM, "lxd_ecdsa.key");
```
edlerd: I think the file name could also include the `profile-key-1` or whatever the current name of the key is in the modal. Also, could include the server identifier, such as using `location.hostname` like in `src/pages/login/CertificateGenerateBtn.tsx`.

Maybe something like `lxd-${location.hostname}-${name}`.

src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  return (
+    <div>
+      <Button type="button" onClick={generateAndDownloadSSHKey}>
+        Generate and download key pair
+      </Button>
```
edlerd: The interaction here is a bit complicated. We could directly generate and download when this section opens. Though the download might be unexpected on clicking the "Generate" tab. Maybe just renaming the label of this button to "Download key pair" makes the interaction simpler?

edlerd (commented):
I am surprised how simple the key generation is with the crypto api! some concerns on the interaction and representation of the new functionality. Mostly questions on the UX.

QA works well, I tested the keys successfully and the new feature works as expected.

src/components/forms/SshKeyGenerateAndDownload.tsx
omarelkashef: I do not know if "download key pair" is descriptive enough. For one, we are only downloading the private key and also that skips the generation part.

edlerd: Maybe even shorter: "download key"? Maybe consult with @piperdeck or check how public clouds are solving this for inspiration.

edlerd: Reflecting on this again, some new observation: I think the UX for me was a little odd because I clicked on the "Generate" tab and then there was again a button with "Generate and download key pair". So I think the UX would be better if we not repeat two clicks on generate. So long thought process: Maybe it is ok to trigger the generation on a button that is only labeled "Download key" in the background. Or we generate the key after the tab is activated and download button does not regenerate. The later might be the cleanest solution. Wdyt?

kimanhou: The issue here is that "Clipboard," "File upload," "Github," and "Launchpad" are all sources for a key that already exists. "Generate Key" is an action that creates the key. I agree with @edlerd, we should generate the key as soon as the tab is active. Then, the tab's content should be a read-only text area, pre-filled with the generated public key (like it's already the case after clicking on the button). It could be nice to add a label "Your new public key". And then underneath a button "Download private key".

src/components/forms/SshKeyGenerateAndDownload.tsx
omarelkashef: It might also be a good idea to include the instance name to avoid conflict between ssh keys of different instances.

edlerd: We might not know the name if this is during creation and relying on the auto naming. Or if this is a profile it could be the profile name, but same issue during creation.

omarelkashef: That is a good point!

src/components/forms/SshKeyAddModal.tsx
```diff
@@ -112,6 +125,7 @@ const SshKeyAddModal: FC<Props> = ({ initialName, onSelect, onClose }) => {
             aria-label="Source"
           >
             {segmentButton("clipboard", "Clipboard")}
+            {segmentButton("generate", "Generate Key")}
```
edlerd: Please use lowercase `Generate key` for the second word.

src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,107 @@
...
+      const publicKeyDataRaw = await window.crypto.subtle.exportKey(
+        "raw",
+        key.publicKey,
+      );
+      const publicKeyOpenSsh = exportEcdsaPublicKeyAsOpenSSH(publicKeyDataRaw);
+      setSSHPublicKey(publicKeyOpenSsh);
+    };
+
+    generateSSHKey();
+  }, []);
```
edlerd, line 29: I think we should define the funciton outside the use effect, then this becomes a bit easier to read as `useEffect(generateSSHKey, []);`

edlerd (commented):
Thanks for the iteration, this looks much better interaction wise! Some change requests:
- Please block the `Add key` until the user downloaded the private key
- If the user downloaded the key and switches to another mode (like `Clipboard` and back to `Generate key`, I think you re-generate the key. Which is fine, but we should similar to the first point block the `Add key` button ot ensure the user downloaded the latest key.

omarelkashef:
> Thanks for the iteration, this looks much better interaction wise! Some change requests:
> 
> * Please block the `Add key` until the user downloaded the private key
> * If the user downloaded the key and switches to another mode (like `Clipboard` and back to `Generate key`, I think you re-generate the key. Which is fine, but we should similar to the first point block the `Add key` button ot ensure the user downloaded the latest key.

Very good point. It should be addressed now.

src/components/forms/SshKeyAddModal.tsx
```diff
@@ -2,14 +2,20 @@ import type { FC, FormEvent, ReactNode } from "react";
...
   onClose: () => void;
 }
 
-type SourceType = "clipboard" | "fileUpload" | "github" | "launchpad";
+type SourceType =
+  | "generate"
+  | "clipboard"
+  | "fileUpload"
+  | "github"
+  | "launchpad";
```
edlerd: Let's keep this sorted alphabetically please.

src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,108 @@
...
+    <>
+      {isDownloaded ? (
+        <p className="u-text--muted">
+          Your private key was downloaded successfully.
+        </p>
+      ) : (
+        <p className="u-text--muted">
+          Your key pair was generated successfully.
+        </p>
+      )}
```
edlerd:
```suggestion
        <p className="u-text--muted">
      {isDownloaded ? "Your private key was downloaded successfully." : "Your key pair was generated successfully."}
        </p>
```
Nitpick: a tiny bit less verbose.

src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <Button
+        disabled={!privateKey || isDownloaded}
```
edlerd:
```suggestion
        disabled={!privateKey}
```
I think we shouldn't block the second download. Maybe the first one failed or a user wants to save it in another location.

src/components/forms/SshKeyGenerateAndDownload.tsx
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  const downloadPrivateKey = async () => {
+    if (!privateKey || !publicKey) return;
+    const privateKeyData = await window.crypto.subtle.exportKey(
+      "pkcs8",
+      privateKey,
+    );
+    const privateKeyPEM = convertToPEMFormat(privateKeyData, "EC PRIVATE KEY");
+    downloadPEMFile(privateKeyPEM, `lxd-${location.hostname}-${keyName}`);
```
edlerd, line 77: I think we should have a file extension, maybe `.pub`?

edlerd: Edit: As this is the private key I think it is correct as is without extension.

PR #1577, opened 2025-10-28
fix(images) ensure local images in instance image selector have a unique key

### LINKED ISSUE #1576: duplicate entries in image list during instance creation
Image

It states I have 3 ubuntu 22.04 images in my local cache

but here I only have 2:

Image

Also would it make sense to clean up the list if there are multiple images of the same OS version? I think having one Ubuntu 22.04 container image in the list is plenty.

The only difference is that one of them is 4 days newer. I don't know what the third image is.

edlerd:
## Done

- fix(images) ensure local images in instance image selector have a unique key, could have multiple versions of the same image in cache.

Fixes #1576

## QA

1. Run the LXD-UI:
    - start image creation, open image selector. search for a locally cached image that has two different versions. Ensure selector behaves normally.

### DIFF
--- a/src/pages/images/ImageSelector.tsx
+++ b/src/pages/images/ImageSelector.tsx
@@ -244,7 +244,13 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
       };
 
       return {
-        key: itemType + item.os + displayRelease + displayVariant + item.server,
+        key:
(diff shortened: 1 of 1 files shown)


edlerd:
@Kxiru see linked issue for a way to reproduce the original issue.

PR #1587, opened 2025-10-30
feat(instance) save terminal connection defaults as instance user key

### LINKED ISSUE #1582: Change terminal connection settings default
Heyhey,

Trying out the NixOS image, the default terminal connection settings that the UI uses do not work because `su` doesn't exist in this image.

Image

Hitting Reconnect and specifying `/bin/sh` as command does work.

Can I request a feature that allows us to change the default terminal connect command or settings for an instance?

I also have a VyOS instance where the `vyos` user is the "root" user. It would be great if I request a terminal for this instance that I could use this feature to make it so that I login as `vyos` instead of root :)

edlerd:
## Done

- feat(instance) save terminal connection defaults as instnace user key

Fixes #1582

## QA

1. Run the LXD-UI:
    - open instance terminal and modify the reconnect settings, save them as default, ensure on reload the settings are restored

## Screenshots

image

### DIFF
--- a/src/pages/instances/InstanceTerminal.tsx
+++ b/src/pages/instances/InstanceTerminal.tsx
@@ -60,6 +60,17 @@ const defaultPayload: TerminalConnectPayload = {
   group: 0,
 };
 
+export const UI_TERMINAL_DEFAULT_PAYLOAD = "user.ui_terminal_default_payload";
+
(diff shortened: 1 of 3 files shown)


Kxiru (approved):
LGTM! Thanks a bunch!

Pre-reload:

image


Post reload:

image

Issue #1567, opened 2025-10-22
Feature: allow configuring user.* on server settings


vosdev:
I just wanted to set a `user.cluster_name` setting on my clusters and noticed that the UI does not offer this option. 

It was easily done through the cli but I figured that maybe you want this feature in the UI :)

`lxc config set user.cluster_name prod`
`lxc config set user.cluster_name lab`

edit: My new custom user tag doesn't show up in settings in the UI either.

<img width="960" height="584" alt="Image" src="https://github.com/user-attachments/assets/33e2669e-ed13-4b23-b73c-e79193a5a0de" />

edlerd:
Thanks for the request, this should be easy to add support for custom user settings (display and edit).

PR #1589, opened 2025-11-03
feat: support user defined settings

omarelkashef:
## Done

- feat: support adding, editing and deleting user defined settings
Addresses #1567 
 

## QA

1. Run the LXD-UI:
    - Go to Settings page.
    - At the bottom of the page, click on "Add key" button
    - Add a new key and value, and click the "Save" button.
    - Ensure the new key/value are added with the key being saved as user.{key}.
    - Click on the newly added value.
    - Change the value and click "Save".
    - Ensure the value is reflected.
    - Click on the newly added value.
    - Click the "Delete user setting" button.
    - Ensure that key/value are deleted.
    - Make sure the key input does not accept duplicate keys or empty values.

## Screenshots

image
image
image
image
image

### DIFF
--- a/src/pages/settings/SettingForm.tsx
+++ b/src/pages/settings/SettingForm.tsx
@@ -22,22 +22,18 @@ import ClusteredSettingFormInput from "./ClusteredSettingFormInput";
 import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 import { useIsClustered } from "context/useIsClustered";
 
-export const getConfigId = (key: string) => {
-  return key.replace(".", "___");
-};
-
 interface Props {
   configField: ConfigField;
+  onDelete: (key: string) => void;
   value?: string;
   clusteredValue?: ClusterSpecificValues;
-  isLast?: boolean;
 }
 
 const SettingForm: FC<Props> = ({
   configField,
+  onDelete,
   value,
   clusteredValue,
-  isLast,
 }) => {
   const { isRestricted } = useAuth();
   const [isEditMode, setEditMode] = useState(false);
@@ -104,7 +100,7 @@ const SettingForm: FC<Props> = ({
   };
 
   useEffect(() => {
-    if (isEditMode && isLast) {
+    if (isEditMode) {
       editRef.current?.scrollIntoView({
         behavior: "smooth",
         block: "nearest",
@@ -172,6 +168,7 @@ const SettingForm: FC<Props> = ({
                   initialValue={value ?? ""}
                   configField={configField}
                   onSubmit={onSubmit}
+                  onDelete={onDelete}
                   onCancel={onCancel}
                 />
               )}
--- a/src/pages/settings/SettingFormInput.tsx
+++ b/src/pages/settings/SettingFormInput.tsx
@@ -2,13 +2,13 @@ import type { FC } from "react";
 import { useState } from "react";
 import { Button, Form, Icon, Input } from "@canonical/react-components";
 import type { ConfigField } from "types/config";
-import { getConfigId } from "./SettingForm";
 import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
 
 interface Props {
   initialValue: string;
   configField: ConfigField;
   onSubmit: (newValue: string | boolean) => void;
+  onDelete: (key: string) => void;
   onCancel: () => void;
 }
 
@@ -16,6 +16,7 @@ const SettingFormInput: FC<Props> = ({
   initialValue,
   configField,
   onSubmit,
+  onDelete,
   onCancel,
 }) => {
   const [value, setValue] = useState(initialValue);
@@ -29,7 +30,8 @@ const SettingFormInput: FC<Props> = ({
     }
   };
 
-  const canBeReset = String(configField.default) !== String(value);
+  const canBeReset =
+    !configField.isUserDefined && String(configField.default) !== String(value);
 
   const resetToDefault = () => {
     setValue(configField.default);
@@ -46,7 +48,7 @@ const SettingFormInput: FC<Props> = ({
       <Input type="submit" hidden value="Hidden input" />
       <Input
         aria-label={configField.key}
-        id={getConfigId(configField.key)}
+        id={configField.key}
         wrapperClassName="input-wrapper"
         type={getInputType()}
         value={configField.type === "bool" ? undefined : String(value)}
@@ -61,6 +63,20 @@ const SettingFormInput: FC<Props> = ({
       <Button appearance="positive" type="submit">
         Save
       </Button>
+      {configField.isUserDefined && (
+        <Button
+          appearance="base"
+          hasIcon
+          className="delete-button"
+          type="button"
+          onClick={() => {
+            onDelete(configField.key);
+          }}
+        >
+          <Icon name="delete"></Icon>
+          <span>Delete</span>
+        </Button>
+      )}
       {canBeReset && (
         <Button
           className="reset-button"
--- a/src/pages/settings/Settings.tsx
+++ b/src/pages/settings/Settings.tsx
@@ -1,5 +1,5 @@
 import type { FC } from "react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import {
   MainTable,
   Notification,
@@ -9,34 +9,104 @@ import {
   useNotify,
   Spinner,
   CustomLayout,
+  useToastNotification,
 } from "@canonical/react-components";
-import SettingForm from "./SettingForm";
 import NotificationRow from "components/NotificationRow";
 import HelpLink from "components/HelpLink";
 import { queryKeys } from "util/queryKeys";
-import { fetchConfigOptions } from "api/server";
+import { fetchConfigOptions, updateSettings } from "api/server";
 import { useQuery } from "@tanstack/react-query";
-import type { ConfigField } from "types/config";
-import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
 import { toConfigFields } from "util/config";
 import PageHeader from "components/PageHeader";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
 import { useServerEntitlements } from "util/entitlements/server";
-import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 import { useClusteredSettings } from "context/useSettings";
-import type { LXDSettingOnClusterMember } from "types/server";
 import { useProjects } from "context/useProjects";
-import { getDefaultProject } from "util/loginProject";
+import {
+  type UserSetting,
+  getConfigFieldClusteredValue,
+  getUserSettings,
+} from "util/settings";
+import {
+  getSettingRow,
+  getUserSettingInputRow,
+  getAddSettingButton,
+} from "pages/settings/SettingsRow";
+import ResourceLabel from "components/ResourceLabel";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import type { ConfigField } from "types/config";
 
 const Settings: FC = () => {
   const [query, setQuery] = useState("");
+
+  const [userSettings, setUserSettings] = useState<UserSetting[]>([]);
+
   const notify = useNotify();
+  const toastNotify = useToastNotification();
+
   const {
     hasMetadataConfiguration,
     settings,
     isSettingsLoading,
     settingsError,
   } = useSupportedFeatures();
+
+  const { data: projects = [] } = useProjects();
+
+  useEffect(() => {
+    if (userSettings.length === 0 && !isSettingsLoading) {
+      setUserSettings(getUserSettings(settings?.config ?? {}, projects));
+    }
+  }, [settings, projects, isSettingsLoading]);
+
+  const saveUserSetting = (index: number) => {
+    const key = `user.${userSettings[index].key}`;
+    const value = userSettings[index].value;
+
+    const settingLabel = <ResourceLabel bold type="setting" value={key} />;
+    updateSettings({ [key]: value })
+      .then(() => {
+        setUserSettings((prev) => {
+          const copy = [...prev];
+          copy[index].isSaved = true;
+          copy[index].key = key;
+          return copy;
+        });
+        toastNotify.success(<>Setting {settingLabel} added</>);
+      })
+      .catch((e) => {
+        notify.failure(`Setting add failed`, e, settingLabel);
+      });
+  };
+
+  const deleteUserSetting = (userKey: string) => {
+    const index = userSettings.findIndex((s) => s.key === userKey);
+
+    const settingLabel = <ResourceLabel bold type="setting" value={userKey} />;
+    updateSettings({ [userKey]: "" })
+      .then(() => {
+        setUserSettings((prev) => {
+          const copy = [...prev];
+          copy.splice(index, 1);
+          return copy;
+        });
+        toastNotify.success(<>Setting {settingLabel} deleted</>);
+      })
+      .catch((e) => {
+        notify.failure(`Setting delete failed`, e, settingLabel);
+      });
+  };
+
+  const matchesQuery = (configField: ConfigField) => {
+    if (!query) {
+      return true;
+    }
+    return (
+      configField.key.toLowerCase().includes(query.toLowerCase()) ||
+      configField.shortdesc?.toLowerCase().includes(query.toLowerCase())
+    );
+  };
+
   const { canEditServerConfiguration } = useServerEntitlements();
 
   const { data: configOptions, isLoading: isConfigOptionsLoading } = useQuery({
@@ -45,8 +115,6 @@ const Settings: FC = () => {
   });
   const { data: clusteredSettings = [], error: clusterError } =
     useClusteredSettings();
-
-  const { data: projects = [] } = useProjects();
 
   if (clusterError) {
     notify.failure("Loading clustered settings failed", clusterError);
@@ -60,35 +128,6 @@ const Settings: FC = () => {
     notify.failure("Loading settings failed", settingsError);
   }
 
-  const getValue = (configField: ConfigField): string | undefined => {
-    for (const [key, value] of Object.entries(settings?.config ?? {})) {
-      if (key === configField.key) {
-        return value;
-      }
-    }
-    if (configField.type === "bool") {
-      return configField.default === "true" ? "true" : "false";
-    }
-    if (configField.default === "-") {
-      return undefined;
-    }
-    return configField.default;
-  };
-
-  const getClusteredValue = (
-    clusteredSettings: LXDSettingOnClusterMember[],
-    configField: ConfigField,
-  ): ClusterSpecificValues => {
-    const settingPerClusterMember: ClusterSpecificValues = {};
-
-    clusteredSettings?.forEach((item) => {
-      settingPerClusterMember[item.memberName] =
-        item.config?.[configField.key] ?? configField.default ?? "";
-    });
-
-    return settingPerClusterMember;
-  };
-
   const headers = [
     { content: "Group", className: "group" },
     { content: "Key", className: "key" },
@@ -97,108 +136,43 @@ const Settings: FC = () => {
 
   const configFields = toConfigFields(configOptions?.configs?.server ?? {});
 
-  configFields.push({
-    key: "user.ui_grafana_base_url",
-    category: "user",
-    default: "",
-    longdesc:
-      "e.g. https://example.org/dashboard?project={project}&name={instance}\n or https://192.0.2.1:3000/d/bGY-LSB7k/lxd?orgId=1",
-    shortdesc:
-      "LXD will replace `{instance}` and `{project}` with project and instance names for deep-linking to individual grafana pages.\nSee {ref}`grafana` for more information.",
-    type: "string",
-  });
-
-  configFields.push({
-    key: "user.ui_login_project",
-    category: "user",
-    default: getDefaultProject(projects),
-    shortdesc: "Project to display on login.",
-    type: "string",
-  });
-
-  configFields.push({
-    key: "user.ui_theme",
-    category: "user",
-    default: "",
-    shortdesc:
-      "Set UI to dark theme, light theme, or to match the system theme.",
-    type: "string",
-  });
-
-  configFields.push({
-    key: "user.ui_title",
-    category: "user",
-    default: "",
-    shortdesc: "Title for the LXD-UI web page. Shows the hostname when unset.",
-    type: "string",
-  });
-
   let lastCategory = "";
-  const rows = configFields
-    .filter((configField) => {
-      if (!query) {
-        return true;
-      }
-      return (
-        configField.key.toLowerCase().includes(query.toLowerCase()) ||
-        configField.shortdesc?.toLowerCase().includes(query.toLowerCase())
+  const rows: MainTableRow[] = configFields
+    .filter(matchesQuery)
+    .map((configField) => {
+      const clusteredValue = getConfigFieldClusteredValue(
+        clusteredSettings,
+        configField,
       );
-    })
-    .map((configField, index, { length }) => {
-      const isDefault = !Object.keys(settings?.config ?? {}).some(
-        (key) => key === configField.key,
-      );
-      const value = getValue(configField);
-
-      const clusteredValue = getClusteredValue(clusteredSettings, configField);
-
       const isNewCategory = lastCategory !== configField.category;
       lastCategory = configField.category;
-
-      return {
-        key: configField.key,
-        columns: [
-          {
-            content: isNewCategory && (
-              <h2 className="p-heading--5">{configField.category}</h2>
-            ),
-            role: "rowheader",
-            className: "group",
-            "aria-label": "Group",
-          },
-          {
-            content: (
-              <div className="key-cell">
-                {isDefault ? (
-                  configField.key
-                ) : (
-                  <strong>{configField.key}</strong>
-                )}
-                <p className="p-text--small u-text--muted u-no-margin--bottom">
-                  <ConfigFieldDescription description={configField.shortdesc} />
-                </p>
-              </div>
-            ),
-            role: "cell",
-            className: "key",
-            "aria-label": "Key",
-          },
-          {
-            content: (
-              <SettingForm
-                configField={configField}
-                value={value}
-                clusteredValue={clusteredValue}
-                isLast={index === length - 1}
-              />
-            ),
-            role: "cell",
-            "aria-label": "Value",
-            className: "u-vertical-align-middle",
-          },
-        ],
-      };
+      return getSettingRow(
+        configField,
+        isNewCategory,
+        clusteredValue,
+        deleteUserSetting,
+        settings,
+      );
     });
+
+  userSettings.filter(matchesQuery).forEach((setting, index) => {
+    const isNewCategory = index === 0;
+
+    const row = setting.isSaved
+      ? getSettingRow(setting, isNewCategory, {}, deleteUserSetting, settings)
+      : getUserSettingInputRow(
+          userSettings,
+          index,
+          setUserSettings,
+          saveUserSetting,
+        );
+
+    rows.push(row);
+  });
+
+  if (!query) {
+    rows.push(getAddSettingButton(setUserSettings));
+  }
 
   return (
     <>
--- a/src/pages/settings/SettingsRow.tsx
+++ b/src/pages/settings/SettingsRow.tsx
@@ -1,0 +1,225 @@
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
+import SettingForm from "pages/settings/SettingForm";
+import type { ConfigField } from "types/config";
+import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
+import { Input, Button, Form } from "@canonical/react-components";
+import { generateUUID } from "util/helpers";
+import { getConfigFieldValue, type UserSetting } from "util/settings";
+import type { LxdSettings } from "types/server";
+
+type SetUserSettings = (prev: (prev: UserSetting[]) => UserSetting[]) => void;
+
+export const getSettingRow = (
+  configField: ConfigField,
+  isNewCategory: boolean,
+  clusteredValue: ClusterSpecificValues,
+  deleteUserSetting: (key: string) => void,
+  settings?: LxdSettings,
+): MainTableRow => {
+  const isDefault = !Object.keys(settings?.config ?? {}).some(
+    (key) => key === configField.key,
+  );
+  const value = getConfigFieldValue(configField, settings);
+  return {
+    key: configField.key,
+    columns: [
+      {
+        content: isNewCategory && (
+          <h2 className="p-heading--5">{configField.category}</h2>
+        ),
+        role: "cell",
+        className: "group",
+        "aria-label": "Group",
+      },
+      {
+        content: (
+          <div className="key-cell">
+            {isDefault ? configField.key : <strong>{configField.key}</strong>}
+            <p className="p-text--small u-text--muted u-no-margin--bottom">
+              <ConfigFieldDescription description={configField.shortdesc} />
+            </p>
+          </div>
+        ),
+        role: "rowheader",
+        className: "key",
+        "aria-label": "Key",
+      },
+      {
+        content: (
+          <SettingForm
+            configField={configField}
+            value={value}
+            clusteredValue={clusteredValue}
+            onDelete={deleteUserSetting}
+          />
+        ),
+        role: "cell",
+        "aria-label": "Value",
+        className: "u-vertical-align-middle",
+      },
+    ],
+  };
+};
+
+export const getUserSettingInputRow = (
+  userSettings: UserSetting[],
+  index: number,
+  setUserSettings: SetUserSettings,
+  saveUserSetting: (index: number) => void,
+): MainTableRow => {
+  const userSetting = userSettings[index];
+  const isKeyDuplicate = userSettings.some(
+    (setting) => `user.${userSetting.key}` === setting.key,
+  );
+  const isFormDisabled =
+    isKeyDuplicate || !userSetting.value || !userSetting.key;
+
+  return {
+    key: userSetting.id,
+    columns: [
+      {
+        content: false,
+        role: "cell",
+        className: "group",
+        "aria-label": "Group",
+      },
+      {
+        content: (
+          <>
+            <Input
+              aria-label="new user key"
+              id={`new-user-defined-key-${index}`}
+              placeholder="User key"
+              type="text"
+              value={userSetting.key}
+              autoFocus
+              error={
+                isKeyDuplicate && <>Setting with this name already exists</>
+              }
+              onChange={(e) => {
+                setUserSettings((prev) => {
+                  const copy = [...prev];
+                  copy[index] = {
+                    ...copy[index],
+                    key: e.target.value,
+                  };
+                  return copy;
+                });
+              }}
+              help={
+                <>
+                  Key will be saved as <code>{"user.{your-key}"}</code>. Enter
+                  only the part after user.
+                </>
+              }
+            />
+          </>
+        ),
+        role: "rowheader",
+        className: "key",
+        "aria-label": "key",
+      },
+      {
+        content: (
+          <Form
+            onSubmit={(e) => {
+              e.preventDefault();
+              saveUserSetting(index);
+            }}
+          >
+            <Input
+              type="submit"
+              hidden
+              value="Hidden input"
+              disabled={isFormDisabled}
+            />
+            <Input
+              aria-label="new user value"
+              id={`new-user-defined-value-${index}`}
+              placeholder="Value"
+              type="text"
+              value={userSetting.value}
+              onChange={(e) => {
+                setUserSettings((prev) => {
+                  const copy = [...prev];
+                  copy[index] = {
+                    ...copy[index],
+                    value: e.target.value,
+                  };
+                  return copy;
+                });
+              }}
+            />
+            <Button
+              type="button"
+              appearance="base"
+              className="button"
+              onClick={() => {
+                setUserSettings((prev) => {
+                  const copy = [...prev];
+                  copy.splice(index, 1);
+                  return copy;
+                });
+              }}
+            >
+              Cancel
+            </Button>
+            <Button
+              type="submit"
+              disabled={isFormDisabled}
+              appearance="positive"
+            >
+              Save
+            </Button>
+          </Form>
+        ),
+        role: "cell",
+        className: "u-vertical-align-middle",
+        "aria-label": "Value",
+      },
+    ],
+  };
+};
+
+export const getAddSettingButton = (setNewUserSettings: SetUserSettings) => {
+  return {
+    key: "add-user-config-button",
+    columns: [
+      {
+        content: false,
+        role: "cell",
+        className: "group",
+        "aria-label": "Group",
+      },
+      {
+        content: (
+          <Button
+            type="button"
+            onClick={() => {
+              setNewUserSettings((prev) => {
+                return [
+                  ...prev,
+                  {
+                    key: "",
+                    value: "",
+                    default: "",
+                    category: "user",
+                    type: "string",
+                    isSaved: false,
+                    isUserDefined: true,
+                    id: generateUUID(),
+                  },
+                ];
+              });
+            }}
+          >
+            Add user setting
+          </Button>
+        ),
+        role: "rowheader",
+        className: "key",
+      },
+    ],
+  };
+};
--- a/src/sass/_settings_page.scss
+++ b/src/sass/_settings_page.scss
@@ -38,9 +38,9 @@
     }
   }
 
-  .reset-button {
-    padding-left: 0.5rem;
-    padding-right: 0.5rem;
+  .reset-button,
+  .delete-button {
+    padding: 0 $sph--small;
   }
 
   .cluster-specific-input {
--- a/src/util/settings.tsx
+++ b/src/util/settings.tsx
@@ -1,4 +1,16 @@
 import type { LxdSettings } from "types/server";
+import type { ConfigField, LxdConfigPair } from "types/config";
+import type { LXDSettingOnClusterMember } from "types/server";
+import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
+
+import { getDefaultProject } from "util/loginProject";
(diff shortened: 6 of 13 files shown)


Kxiru (commented):
Hiya!
Here are some initial comments.

1) I just thought I would show you this custom-key area of the system. It can be found by creating a graphic instance (search up desktop in the image selector and go with ubuntu noble for example), running the instance, heading to the Terminal tab, and clicking reconnect. I suggest this because your current design approach could perhaps mimic this further for consistency.

image

2) Creation / edit modes should not be visible via default, and so the current approach is inconsistent with the system design. (See empty state below). I would recommend having a secondary button somewhere (Primary seems a bit too bold but that could be a design choice), and maybe calling a modal of sorts. @edlerd , what do you think? If not a modal, then at the very least, we can add in your current implementation, but perhaps with both inputs (key and value), visible at once rather than one -> button click input -> and then the other.
image


3) Try testing out your implementation in different window sizes, as it isn't entirely responsive right now.
image

4) Additionally, I would move the delete button to be adjacent to the edit icon, again for consistency, as is the case in the action columns of entity rows in tables, or most input configs. Additionally, for the delete button, I would suggest using a ConfirmationButton with appearance = "base". (See DeleteStorageVolumeBtn.tsx).

As seen, here:
image

edlerd:
> * After "Add key" the UX suggests something was saved, but at that point all changes are in the local state. I think the `Add key` could give us two inputs (one for name and another for value). That way we only have one Save button which aligns with when things get actually stored.

A bit more detail, I think this could work similarly to the "Add variable" pattern we have for reconnecting an instance terminal.
image

edlerd:
> Thank you @Kxiru and @edlerd for the thorough feedback. So, we decided that it will be a modal? If so, what should the button that opens the modal look like? I am assuming it will still be part of the config table under the user group, but what column should it be in?

As discussed, I don't think we need a modal, it was more an example of the pattern that is visible in the modal context. For the settings, I think we can keep the interaction on the main page, no need to adopt a modal.

src/pages/settings/SettingForm.tsx
```diff
@@ -86,10 +94,19 @@ const SettingForm: FC<Props> = ({
       });
   };
 
+  const onSubmit = (newValue: string | boolean | ClusterSpecificValues) => {
+    updateConfig(newValue);
+  };
```
edlerd: This callback seems redundant, can we directly use `updateConfig`?

src/pages/settings/SettingForm.tsx
```diff
@@ -86,10 +94,19 @@ const SettingForm: FC<Props> = ({
...
+
   const onCancel = () => {
     setEditMode(false);
   };
 
+  const onDelete = () => {
+    setEditMode(false);
+    updateConfig("");
+  };
```
edlerd: The `setEditMode(false)` happens as a side effect of `updateConfig`, I think we can remove this callback.

src/pages/settings/SettingFormInput.tsx
```diff
@@ -10,13 +10,15 @@ interface Props {
   configField: ConfigField;
   onSubmit: (newValue: string | boolean) => void;
   onCancel: () => void;
+  onDelete: () => void;
```
edlerd: I think we can remove the additional callback for `onDelete` here and use `onSubmit("")` instead to trigger deletion.

src/pages/settings/SettingFormInput.tsx
```diff
@@ -61,15 +63,27 @@ const SettingFormInput: FC<Props> = ({
...
           onClick={resetToDefault}
           hasIcon
         >
           <Icon name="restart" className="flip-horizontally" />
-          <span>Reset to default</span>
+          <span>
+            {" "}
+            {configField.isUserDefined ? "Delete" : "Reset to default"}
+          </span>
         </Button>
```
edlerd: This seems redundant, we check for `!configField.isUserDefined` outside and then have a conditional on `configField.isUserDefined` inside.

src/pages/settings/SettingFormInput.tsx
```diff
@@ -61,15 +63,27 @@ const SettingFormInput: FC<Props> = ({
...
       </Button>
-      {canBeReset && (
+      {configField.isUserDefined && (
         <Button
-          className="reset-button"
+          appearance="base"
+          className="reset-or-delete-button"
+          onClick={onDelete}
+        >
+          Delete key
```
edlerd:
```suggestion
          Delete user setting
```
Suggestion for a bit better copy. I am open to further suggestions, as I am not 100% sure this is the right way to label it to make it precisely understood.

src/pages/settings/Settings.tsx
```diff
@@ -58,6 +72,38 @@ const Settings: FC = () => {
     notify.failure("Loading settings failed", settingsError);
   }
 
+  const isKeyDuplicate = (key: string) => {
+    return configFields.some((cf) => `user.${key}` === cf.key);
+  };
```
edlerd: This should trigger a warning on the key input, currently it is unclear why the save button is disabled. We can use the `error` prop on the key input to show a warning like "Setting with this name already exists".

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,146 @@ const Settings: FC = () => {
...
+                    const copy = [...prev];
+                    copy.splice(index, 1);
+                    return copy;
+                  });
+                }}
+              >
+                Cancel
+              </Button>
+            </>
+          ),
```
edlerd: The order here should switch, so the cancel button is on the left, as that is the established pattern in the app.

tests/server.spec.ts
```diff
@@ -56,5 +56,22 @@ test("only user server setting available for lxd v5.0/edge", async ({
...
+  expect(allSettingRows.length).toEqual(4); //3 + 1 for "Add key button"
+});
+
+test("add user defined setting", async ({ page, lxdVersion }) => {
+  test.skip(
+    lxdVersion !== "5.0-edge",
+    `this test is specific to lxd v5.0/edge, current lxd snap channel is ${lxdVersion}`,
+  );
+
+  const key = "key";
```
edlerd: Better add some randomization to the key. See how we create other entities like instances or profiles, there are generic helpers available for suffixing random strings. The reason is so that this has a chance to succeed if it fails in between and is re-run.

tests/server.spec.ts
```diff
@@ -56,5 +56,22 @@ test("only user server setting available for lxd v5.0/edge", async ({
...
+  const value = "value";
+
+  await visitServerSettings(page);
+  await page.getByRole("button", { name: "Add key", exact: true }).click();
+  await page.getByPlaceholder("Custom key").fill(key);
+  await page.getByPlaceholder("Value").fill(value);
+  await page.getByRole("button", { name: "Save", exact: true }).click();
+  await page.getByText(`Setting user.${key}: ${value} added`).click();
```
edlerd: Please, can you also test deletion?

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,146 @@ const Settings: FC = () => {
...
+                  setUserDefinedConfigs((prev) => {
+                    const copy = [...prev];
+                    copy[index] = {
+                      ...copy[index],
+                      value: e.target.value,
+                    };
+                    return copy;
+                  });
+                }}
+              />
```
edlerd: I think we can add some help text on the key input to explain this is too be prefixed with user. later on and only the part after user. to be entered. Try to find a precise and concise explanation.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,146 @@ const Settings: FC = () => {
...
+          className: "group",
+          "aria-label": "Group",
+        },
+        {
+          content: (
+            <>
+              <Input
+                aria-label={`new-user-defined-key-${index}`}
+                id={`new-user-defined-key-${index}`}
+                placeholder="Custom key"
```
edlerd: Maybe "User key"?

edlerd (commented):
Thanks for the iteration! The interaction is nearly perfect, just one concern on the button order and some ideas around adding help text or adjusting copy.

Some small comments on the code as well below.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+          content: false,
+          role: "rowheader",
+          className: "group",
+          "aria-label": "Group",
+        },
+        {
+          content: (
+            <>
+              <Input
+                aria-label={`new-user-defined-key-${index}`}
```
edlerd: this is for humans and should be a nicer label.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+                id={`new-user-defined-key-${index}`}
+                placeholder="User key"
+                type="text"
+                value={userDefinedConfigs[index].key}
```
edlerd: I think we can rely on the value that `userDefinedConfigs.forEach` gives us instead of using `userDefinedConfigs[index]` here.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+                    const copy = [...prev];
+                    copy[index] = {
+                      ...copy[index],
+                      key: e.target.value,
+                    };
+                    return copy;
+                  });
+                }}
+              />
+              {index == userDefinedConfigs.length - 1 && (
```
edlerd: I think this should be visible on all keys, not just on the last one.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+                <p className="p-text--small u-text--muted u-no-margin--bottom">
+                  <ConfigFieldDescription
+                    description={
+                      "Key will be saved as user.`{your-key}`. Enter only the part after user."
+                    }
+                  />
+                </p>
```
edlerd: The spacing here is off and this is a bit too complicated. You can set the help property on `Input` instead to keep this simple.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+                    category: "user",
+                    type: "string",
+                    isUserDefined: true,
+                  },
+                ];
+              });
+            }}
+          >
+            Add key
+          </Button>
```
edlerd: I think the on click should focus the key input, so a user can immediately start typing the key.

tests/server.spec.ts
```diff
@@ -56,5 +60,31 @@ test("only user server setting available for lxd v5.0/edge", async ({
...
+  );
+
+  const key = randomSettingKey();
+  const value = randomSettingValue();
+
+  await visitServerSettings(page);
+  await addUserDefinedSetting(page, key, value);
+
+  await page.getByText(`Setting user.${key}: ${value} added`).click();
+  const keyElement = page.locator("td div strong", { hasText: `user.${key}` });
```
edlerd: I think the `td div strong` is a bit too close on the implementation, is there a way to define it on a more semantic level?

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+        ),
+        role: "cell",
+        className: "key",
+        "aria-label": "key",
+      },
+      {
+        content: false,
+        role: "cell",
+        className: "key",
+        "aria-label": "value",
```
edlerd:
```suggestion
```
I don't think we should have a label on an empty column

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+          </Button>
+        ),
+        role: "cell",
+        className: "key",
+        "aria-label": "key",
+      },
+      {
+        content: false,
+        role: "cell",
+        className: "key",
```
edlerd:
```suggestion
```
I don't think we need the class here.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +250,159 @@ const Settings: FC = () => {
...
+                ];
+              });
+            }}
+          >
+            Add key
+          </Button>
+        ),
+        role: "cell",
+        className: "key",
+        "aria-label": "key",
```
edlerd: this label seems off, maybe can be removed?

src/pages/settings/Settings.tsx
```diff
@@ -58,6 +72,38 @@ const Settings: FC = () => {
...
+        queryClient.invalidateQueries({
+          queryKey: [queryKeys.settings, queryKeys.cluster],
+        });
+      });
+
+    setUserDefinedConfigs((prev) => {
+      const copy = [...prev];
+      copy.splice(index, 1);
+      return copy;
+    });
```
edlerd: I think this should only happen in the `then` branch above and not in the error case.

src/pages/settings/Settings.tsx
```diff
@@ -58,6 +72,38 @@ const Settings: FC = () => {
...
+  const isKeyDuplicate = (key: string) => {
+    return configFields.some((cf) => `user.${key}` === cf.key);
+  };
+
+  const addUserDefinedConfig = (index: number) => {
+    const key = userDefinedConfigs[index].key;
+    const value = userDefinedConfigs[index].value;
+
+    const mutation = updateSettings({ [`user.${key}`]: value });
+    mutation
```
edlerd:
```suggestion
    updateSettings({ [`user.${key}`]: value })
```
Slightly simpler -- I don't think we need to assign this if we only call it once directly after.

src/pages/settings/Settings.tsx
```diff
@@ -58,6 +72,38 @@ const Settings: FC = () => {
...
+      .then(() => {
+        toastNotify.success(<>{`Setting user.${key}: ${value} added`}</>);
```
edlerd:
```suggestion
        toastNotify.success(<>{`Setting user.${key} added`}</>);
```
I think we only need to mention the setting here, but we should probably use the resource label component with the icon here, so the update message matches the other notifications:

image

src/pages/settings/Settings.tsx
```diff
@@ -58,6 +72,38 @@ const Settings: FC = () => {
...
+      })
+      .catch((e) => {
+        notify.failure(`Setting user.${key}: ${value} add failed`, e, key);
```
edlerd:
```suggestion
        notify.failure(`Setting user.${key} update failed`, e, key);
```
I think we should use the resource label component with the icon here as well.

src/pages/settings/Settings.tsx
```diff
@@ -58,6 +72,38 @@ const Settings: FC = () => {
     notify.failure("Loading settings failed", settingsError);
   }
 
+  const isKeyDuplicate = (key: string) => {
+    return configFields.some((cf) => `user.${key}` === cf.key);
+  };
+
+  const addUserDefinedConfig = (index: number) => {
```
edlerd: Inside this function, we repeat the `user.${key}` concatenation a few times. Should we assign it to a const to reuse?

src/pages/settings/Settings.tsx
edlerd: Also the `ConfigFieldDescription` is meant to enrich help strings that come from the metadata api with links to docs. As we produce this help string in the UI we don't need it here.

tests/server.spec.ts
omarelkashef: I also had the same concern. I was using the key alone to look up the element but it showed up in multiple elements.

edlerd: I think it matches the notification. You can use the `dismissNotification` helper to close the notification, and then the selector by text only should be unique.

src/pages/settings/Settings.tsx
omarelkashef: className and aria-label are required in the column, so we can make them empty strings.

src/pages/settings/Settings.tsx
omarelkashef: After addressing the earlier comments, it is only repeated twice.

src/pages/settings/Settings.tsx
edlerd: I think there is an inferred type in place, and typescript needs a hint that the rows object is an array of `MainTableRow`. In it, both fields are optional.

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
edlerd, line 1: This seems unrelated changes, please revert them.

src/pages/settings/Settings.tsx
```diff
@@ -26,11 +30,23 @@ import { useServerEntitlements } from "util/entitlements/server";
 import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 import { useClusteredSettings } from "context/useSettings";
 import type { LXDSettingOnClusterMember } from "types/server";
+import ResourceLabel from "components/ResourceLabel";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+
+type userDefinedConfig = ConfigField & { value: string };
```
edlerd:
```suggestion
type UserDefinedConfig = ConfigField & { value: string };
```
Types should start with uppercase letters

src/pages/settings/Settings.tsx
```diff
@@ -123,8 +171,23 @@ const Settings: FC = () => {
     type: "string",
   });
 
+  Object.entries(settings?.config ?? {})
+    .filter(
+      ([key, _]) =>
+        key.startsWith("user.") && !configFields.some((cf) => cf.key === key),
+    )
```
edlerd: Why do we filter and what is this doing?

I think this is adding the definition of user defined fields, but filtering out settings that are already defined. Might be worth explaining why this is necessary in a comment.

src/util/formChangeCount.tsx
edlerd, line 1: Please revert those unrelated changes.

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +253,145 @@ const Settings: FC = () => {
...
+            }}
+          >
+            Add key
+          </Button>
+        ),
+        role: "cell",
+        className: "key",
+      },
+    ],
+  });
```
edlerd: Can we move this together with the `addUserDefinedConfig` function into a dedicated component to separate the concerns?

src/pages/settings/Settings.tsx
```diff
@@ -26,11 +30,23 @@ import { useServerEntitlements } from "util/entitlements/server";
...
+type userDefinedConfig = ConfigField & { value: string };
 
 const Settings: FC = () => {
   const docBaseLink = useDocs();
   const [query, setQuery] = useState("");
+
+  const [userDefinedConfigs, setUserDefinedConfigs] = useState<
+    userDefinedConfig[]
+  >([]);
+
```
edlerd: Discovered one QA issue: On adding a new setting where other user settings exist, the new setting will "vanish" into the list alphabetically and move around. This is slightly unexpected, and ideally the new value stays in line. On reload, it can be sorted with other keys alphabetically.



https://github.com/user-attachments/assets/affa1c75-525a-45ee-a41e-951248c90102

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +253,145 @@ const Settings: FC = () => {
...
+                  setUserDefinedConfigs((prev) => {
+                    const copy = [...prev];
+                    copy[index] = {
+                      ...copy[index],
+                      value: e.target.value,
+                    };
+                    return copy;
+                  });
+                }}
+              />
```
edlerd: We should support submitting the form input with the enter key.

src/pages/settings/Settings.tsx
edlerd: Second thought: Maybe we can completely avoid adding the user rows to the configFields array. Instead, a custom function can produce the rows for the user defined configs in the correct order. Then in this Settings file we simply call that custom function to append the rows, keeping the changes and extension of the component minimal and focused on the high level of settings.

src/pages/settings/SettingFormInput.tsx
```diff
@@ -61,9 +62,21 @@ const SettingFormInput: FC<Props> = ({
...
+            onSubmit("");
+          }}
+        >
+          Delete user setting
+        </Button>
+      )}
       {canBeReset && (
         <Button
-          className="reset-button"
+          className="reset-or-delete-button"
```
kimanhou: I think this change affects other `Reset to default` buttons (in `SettingFormCheckbox` and `ClusteredSettingFormInput`). On hover, the buttons look different.

src/sass/_settings_page.scss
```diff
@@ -38,7 +38,7 @@
     }
   }
 
-  .reset-button {
+  .reset-or-delete-button {
     padding-left: 0.5rem;
```
kimanhou: I know it's not your code but could you please use Vanilla's variables for horizontal spacing ? You can also use the shorthand syntax as padding left and right are the same here.

image

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +253,145 @@ const Settings: FC = () => {
...
+                  setUserDefinedConfigs((prev) => {
+                    const copy = [...prev];
+                    copy[index] = {
+                      ...copy[index],
+                      key: e.target.value,
+                    };
+                    return copy;
+                  });
+                }}
+                help="Key will be saved as user.{your-key}. Enter only the part after user."
```
kimanhou: Can `user.{your-key}` and `user.` be inside `<code>` tags, similar to {instance} and {project} in the helptext of `user.ui_grafana_base_url` ?

image

src/pages/settings/Settings.tsx
```diff
@@ -190,6 +253,145 @@ const Settings: FC = () => {
...
+          role: "cell",
+          className: "u-vertical-align-middle",
+          "aria-label": "Value",
+        },
+      ],
+    });
+  });
+
+  rows.push({
+    key: "new-user-defined-config",
```
kimanhou: What if the user multiple settings ? Will the key be the same for all of them ?
If you need random keys, you can use React's `useId`.

kimanhou (commented):
Thank you for adding this, it looks sleek ! Please rebase on top of main, feel free to ping me if you need help as conflicts are with my work on login project.

QA looks good, except the issues already mentioned by David. Some comments below:

Should we allow non alphanumeric characters ? I am currently able to create a user setting with `key=]`
image


Could you add a garbage bin icon next to `Delete user setting` to match the design of `Reset to default`. Also would "Delete" be enough ?
Screenshot from 2025-11-17 09-39-16

src/pages/settings/Settings.tsx
omarelkashef: This seems like it would create duplicate code of adding rows, once for user defined configs and for other configs.

edlerd: Good point! I wonder if we could extract the common pattern. Then compose user defined and standard configs from it. A principle similar to the `getConfigurationRow` pattern for the configuration tables. Though much simpler as I expect the settings to be very equal between the user and standard config.

src/pages/settings/Settings.tsx
edlerd: In other places we add this form element to allow submission with enter: `<Input type="submit" hidden value="Hidden input" />`. And a wrapping `<form>`, but I think that form element is already in place here.

src/pages/settings/Settings.tsx
omarelkashef: Yeah I think this would be much cleaner.

src/pages/settings/Settings.tsx
omarelkashef: If we want the user defined configs to be in the same table as the rest of the configs, we need to add them to the rows variable. I am wondering how we can do that if we extract it to a dedicated component.

edlerd: I think we can keep the current `const rows = configFields. [...] .map [...]` as is and after that do something like `rows.push(getCustomUserConfigs())`. Where `getCustomUserConfigs` refers to an imported function that lives in a dedicated file bundling all the custom user config logic.

src/pages/settings/Settings.tsx
omarelkashef: I believe the keys are jumping because of how saving the SettingForm works. We issue a get request after every patch, so the keys are sorted after every addition.

edlerd: Yes, because the patch is invalidating the get query cache and `useQuery` is fetching the data again leading to a rerender.

I think keeping the custom user values in a local state that persists the order of them can solve this. We only initialize that local state on first render. So their order is not determined by the get response contents.

omarelkashef:
> Thank you for adding this, it looks sleek ! Please rebase on top of main, feel free to ping me if you need help as conflicts are with my work on login project.
> 
> QA looks good, except the issues already mentioned by David. Some comments below:
> 
> Should we allow non alphanumeric characters ? I am currently able to create a user setting with `key=]` image
> 
> Could you add a garbage bin icon next to `Delete user setting` to match the design of `Reset to default`. Also would "Delete" be enough ? Screenshot from 2025-11-17 09-39-16

These are purely user configs and I believe we only display them and do no processing, so it should be fine letting user enter non-alphanumeric characters. Do you think there are downsides to it?

src/pages/settings/Settings.tsx
omarelkashef: The key is kind of deceiving I agree. This is the "add key" button which exists only once in the page. I have changed the key to avoid confusion.

kimanhou:
It looks like there is a circular dependency:
- `SettingForm` imports `SettingFormInput`, `SettingFormCheckbox`, `SettingFormPassword` and `ClusteredSettingFormInput`
- `SettingFormInput`, `SettingFormCheckbox`, `SettingFormPassword` and `ClusteredSettingFormInput` import `SettingForm`

To resolve, move `getConfigId` from `SettingsForm` to `util/settings.tsx`

src/pages/settings/Settings.tsx
```diff
@@ -9,34 +9,104 @@ import {
...
+import ResourceLabel from "components/ResourceLabel";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
 
 const Settings: FC = () => {
   const [query, setQuery] = useState("");
+
+  const [existingUserConfigs, setExistingUserConfigs] = useState<
+    UserDefinedConfig[]
+  >([]);
+  const [newUserConfigs, setNewUserConfigs] = useState<UserDefinedConfig[]>([]);
```
edlerd: I think this should be one state `userConfigs` for both new and existing. The entries can have a property to mark them as new or existing. Currently, the entries can jump around on save as they are moved between the two arrays.

edlerd (commented):
Good progress, one observation on the new states below.

src/pages/settings/Settings.tsx
omarelkashef: The new state  is for uncommitted configs and is rendered as input fields, so if it is one state, when a config moves from being new to existing, we would have existing configs in between inputs. Currently, all the existing fields come first, and then the inputs for the new fields

edlerd: I think the code will simplify with a single state. Because the complexity from moving entries between the state arrays vanishes.

See below for an example for when the rows jump around with the split state arrays.

https://github.com/user-attachments/assets/22fe525c-eeb3-4921-a20f-53a5ca5f60bc

src/pages/settings/SettingFormInput.tsx
```diff
@@ -61,9 +64,23 @@ const SettingFormInput: FC<Props> = ({
...
+          }}
+        >
+          <Icon name="delete"></Icon>
+          <span>Delete</span>
+        </Button>
+      )}
       {canBeReset && (
         <Button
-          className="reset-button"
+          className="reset--button"
```
edlerd:
```suggestion
          className="reset-button"
```
I think this change should be reverted to match the scss.

src/pages/settings/Settings.tsx
```diff
@@ -9,34 +9,95 @@ import {
...
+  getAddConfigButton,
+} from "util/settings";
+import ResourceLabel from "components/ResourceLabel";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
 
 const Settings: FC = () => {
   const [query, setQuery] = useState("");
+
+  const [userConfigs, setUserConfigs] = useState<UserConfigField[]>([]);
+  //const [newUserConfigs, setNewUserConfigs] = useState<UserDefinedConfig[]>([]);
```
edlerd:
```suggestion
```

src/sass/_settings_page.scss
```diff
@@ -38,9 +38,9 @@
     }
   }
 
-  .reset-button {
-    padding-left: 0.5rem;
-    padding-right: 0.5rem;
+  .reset-button,
+  .delete-button {
+    padding: 0 $sph--small;
```
edlerd, line 43: Good idea to use the variable and to simplify this.

tests/server.spec.ts
```diff
@@ -57,6 +62,32 @@ test("only user server setting available for lxd v5.0/edge", async ({
...
-  expect(allSettingRows.length).toEqual(4);
+  // only 4 user settings available (user.ui_title, user.ui_grafana_base_url, user.ui_theme, user.ui_current_project) + 1 "Add key" button.
+  expect(allSettingRows.length).toEqual(5);
+});
+
+test("add and delete user defined setting", async ({ page, lxdVersion }) => {
+  test.skip(
+    lxdVersion !== "5.0-edge",
+    `this test is specific to lxd v5.0/edge, current lxd snap channel is ${lxdVersion}`,
+  );
```
edlerd: This seems wrong, why only run this test for the ancient 5.0 version?

src/pages/settings/Settings.tsx
```diff
@@ -9,34 +9,95 @@ import {
...
     isSettingsLoading,
     settingsError,
   } = useSupportedFeatures();
+
+  const { data: projects = [] } = useProjects();
+
+  useEffect(() => {
+    if (userConfigs.length === 0) {
+      setUserConfigs(getUserConfigs(settings?.config ?? {}, projects));
+    }
```
edlerd: This needs to make sure `settings` are fully loaded before initializing. I think we should put `userConfigs`, `settings` and `projects` into the dependency list of the use effect.

src/util/settings.tsx
```diff
@@ -17,3 +39,316 @@ export const isClusteredServer = (
...
+                    key: e.target.value,
+                  };
+                  return copy;
+                });
+              }}
+              help={
+                <p>
+                  Key will be saved as user.<code>{"{your-key}"}</code>. Enter
+                  only the part after user.
+                </p>
```
edlerd:
```suggestion
                <>
                  Key will be saved as <code>user.{"{your-key}"}</code>. Enter
                  only the part after user.
                </>
```
This should be a fragment to avoid nesting p inside of another p.

Also think we should keep the `user.` in the code block as well.

src/util/settings.tsx
```diff
@@ -17,3 +39,316 @@ export const isClusteredServer = (
...
+            <Input
+              aria-label="new user key"
+              id={`new-user-defined-key-${index}`}
+              placeholder="User key"
+              type="text"
+              value={userConfig.key}
+              autoFocus
+              error={
+                isKeyDuplicate(userConfig.key) && (
+                  <p>Setting with this name already exists</p>
```
edlerd:
```suggestion
                  <>Setting with this name already exists</>
```
Avoid nesting paragraphs.

src/util/settings.tsx
```diff
@@ -17,3 +39,316 @@ export const isClusteredServer = (
...
+  const isKeyDuplicate = (key: string) => {
+    return userConfigs.some((cf) => `user.${key}` === cf.key);
+  };
+  const isFormDisabled = () => {
+    return (
+      isKeyDuplicate(userConfig.key) || !userConfig.value || !userConfig.key
+    );
+  };
+  return {
+    key: `new-config-${index}`,
```
edlerd: I think this is an unstable key, because we can add / remove the user defined keys. So the same index can relate to different rows in the lifecycle of the table. Can we use `generateUUID` from the helpers on adding a new row and store it on the `userConfig` state entry instead? Then the key will be stable also on adding/removing rows.

src/util/settings.tsx
```diff
@@ -1,4 +1,24 @@
 import type { LxdSettings } from "types/server";
+import type { ConfigField, LxdConfigPair } from "types/config";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import type { LXDSettingOnClusterMember } from "types/server";
+import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
+import SettingForm from "pages/settings/SettingForm";
+import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
```
edlerd: We still get errors in CI due to circular dependencies. I think this util file is the cause.

src/pages/settings/ConfigurationRow.tsx
edlerd, line 1: The name clashes with the existing `src/components/ConfigurationRow.tsx`. Can we call this one `SettingsRow.tsx` and use `Setting` instead of `ConfigField` in the function names below to avoid confusion?

src/pages/settings/Settings.tsx
```diff
@@ -9,34 +9,96 @@ import {
...
     hasMetadataConfiguration,
     settings,
     isSettingsLoading,
     settingsError,
   } = useSupportedFeatures();
+
+  const { data: projects = [] } = useProjects();
+
+  useEffect(() => {
+    if (userConfigs.length === 0) {
```
edlerd:
```suggestion
    if (userConfigs.length === 0 && !isSettingsLoading) {
```
I think we have to guard this against the settings being loaded slowly.

src/pages/settings/Settings.tsx
```diff
@@ -9,34 +9,96 @@ import {
...
+      setUserConfigs(getUserConfigs(settings?.config ?? {}, projects));
+    }
+  }, [settings, projects]);
+
+  const addUserSetting = (index: number) => {
```
edlerd:
```suggestion
  const saveUserSetting = (index: number) => {
```
Maybe a bit more precise? Add can be confused with the non-saved addition.

src/pages/settings/SettingForm.tsx
```diff
@@ -172,6 +170,10 @@ const SettingForm: FC<Props> = ({
                   initialValue={value ?? ""}
                   configField={configField}
                   onSubmit={onSubmit}
+                  onDelete={(key: string) => {
+                    onDelete(key);
+                    setEditMode(false);
+                  }}
```
edlerd:
```suggestion
                  onDelete={onDelete}
```
A bit simpler, the field vanishes so we don't need to disable edit mode. On failure, it is preferred to stay in edit mode, so the user can retry.

src/pages/settings/ConfigurationRow.tsx
```diff
@@ -0,0 +1,234 @@
...
+        role: "cell",
+        "aria-label": "Value",
+        className: "u-vertical-align-middle",
+      },
+    ],
+  };
+};
+
+export const getUserConfigFieldInputRow = (
+  userConfig: UserConfigField,
```
edlerd:
```suggestion
```
We don't need to pass it in, but can look it up inside as `const userConfig = userConfigs[index]` to simplify the interface.

src/pages/settings/ConfigurationRow.tsx
```diff
@@ -0,0 +1,234 @@
...
+) => void;
+
+export const getConfigFieldRow = (
+  configField: ConfigField,
+  isLast: boolean,
+  isNewCategory: boolean,
+  clusteredValue: ClusterSpecificValues,
+  deleteUserSetting: (key: string) => void,
+  settings?: LxdSettings,
+  value?: string,
```
edlerd:
```suggestion
```
We should do the call to `getConfigFieldValue` inside this function to keep the interface small and clean.

src/pages/settings/ConfigurationRow.tsx
```diff
@@ -0,0 +1,234 @@
...
+import type { UserConfigField } from "util/settings";
+import type { LxdSettings } from "types/server";
+
+type SetUserConfigs = (
+  prev: (prev: UserConfigField[]) => UserConfigField[],
+) => void;
+
+export const getConfigFieldRow = (
+  configField: ConfigField,
+  isLast: boolean,
```
edlerd:
```suggestion
```
I think we should remove this here and also downstream. It seems only used in a `useEffect` as a guard. We should remove the guard and always scroll into view, independent of it being the last element.

src/pages/settings/ConfigurationRow.tsx
```diff
@@ -0,0 +1,234 @@
...
+    return userConfigs.some((cf) => `user.${key}` === cf.key);
+  };
+  const isFormDisabled = () => {
+    return (
+      isKeyDuplicate(userConfig.key) || !userConfig.value || !userConfig.key
+    );
+  };
+
+  return {
+    key: userConfig.id || `new-user-config-${index}`,
```
edlerd:
```suggestion
    key: userConfig.id,
```
I think we don't need the fallback here.

src/pages/settings/ConfigurationRow.tsx
```diff
@@ -0,0 +1,234 @@
...
+    columns: [
+      {
+        content: false,
+        role: "rowheader",
```
edlerd:
```suggestion
        role: "cell",
```
I think generally, for this and the other 3 occasions in this file, the second column should be the row header. The first column is often empty and representative for the row.

src/pages/settings/ConfigurationRow.tsx
```diff
@@ -0,0 +1,234 @@
...
+                    type: "string",
+                    isSaved: false,
+                    isUserDefined: true,
+                    id: generateUUID(),
+                  },
+                ];
+              });
+            }}
+          >
+            Add key
```
edlerd:
```suggestion
            Add user setting
```
Maybe a bit clearer to understand?

src/pages/settings/SettingForm.tsx
```diff
@@ -172,6 +168,9 @@ const SettingForm: FC<Props> = ({
                   initialValue={value ?? ""}
                   configField={configField}
                   onSubmit={onSubmit}
+                  onDelete={(key: string) => {
+                    onDelete(key);
+                  }}
```
edlerd:
```suggestion
                  onDelete={onDelete}
```
Slightly simpler

src/pages/settings/Settings.tsx
```diff
@@ -9,34 +9,95 @@ import {
...
     settingsError,
   } = useSupportedFeatures();
+
+  const { data: projects = [] } = useProjects();
+
+  useEffect(() => {
+    if (userSettings.length === 0 && !isSettingsLoading) {
+      setUserSettings(getUserSettings(settings?.config ?? {}, projects));
+    }
+  }, [settings, projects]);
```
edlerd:
```suggestion
  }, [settings, projects, isSettingsLoading]);
```
This seems to be missing from dependencies

src/pages/settings/Settings.tsx
```diff
@@ -9,34 +9,95 @@ import {
...
+
+  const saveUserSetting = (index: number) => {
+    const key = userSettings[index].key;
```
edlerd: I'd suggest prefixing with `user.` here and then the three usages below simplify.

src/pages/settings/Settings.tsx
```diff
@@ -144,62 +138,39 @@ const Settings: FC = () => {
...
+      return getSettingRow(
+        configField,
+        isNewCategory,
+        clusteredValue,
+        deleteUserSetting,
+        settings,
+      );
     });
 
+  userSettings.forEach((s, index) => {
```
edlerd:
```suggestion
  userSettings.forEach((setting, index) => {
```
Let's make the code below easier to read.

src/pages/settings/SettingsRow.tsx
```diff
@@ -0,0 +1,230 @@
...
+): MainTableRow => {
+  const userSetting = userSettings[index];
+  const isKeyDuplicate = (key: string) => {
+    return userSettings.some((cf) => `user.${key}` === cf.key);
+  };
+  const isFormDisabled = () => {
+    return (
+      isKeyDuplicate(userSetting.key) || !userSetting.value || !userSetting.key
+    );
+  };
```
edlerd:
```suggestion
  const isKeyDuplicate = userSettings.some(
    (setting) => `user.${userSetting.key}` === setting.key,
  );
  const isFormDisabled =
    isKeyDuplicate || !userSetting.value || !userSetting.key;
```

This simplifies as we don't need the callbacks here, but can directly evaluate on render.

src/pages/settings/Settings.tsx
edlerd: We should also filter based on the query. Maybe extract the filter function as below and reuse it here and above:

```
  const matchesQuery = (configField: ConfigField) => {
    if (!query) {
      return true;
    }
    return (
      configField.key.toLowerCase().includes(query.toLowerCase()) ||
      configField.shortdesc?.toLowerCase().includes(query.toLowerCase())
    );
  };
  ```

File: src/pages/settings/Settings.tsx [hunk 59]
```diff
@@ -97,109 +136,48 @@ const Settings: FC = () => {
...
+            userSettings,
+            index,
+            setUserSettings,
+            saveUserSetting,
+          );
+
+      rows.push(row);
     });
 
+  rows.push(getAddSettingButton(setUserSettings));
```
edlerd: We should hide the button when a search is active. Otherwise, it seems like the button is not adding new rows.

src/pages/settings/Settings.tsx
```diff
@@ -97,109 +136,48 @@ const Settings: FC = () => {
...
-            role: "cell",
-            "aria-label": "Value",
-            className: "u-vertical-align-middle",
-          },
-        ],
-      };
+  userSettings
+    .filter((setting) => {
+      return matchesQuery(setting);
+    })
```
edlerd:
```suggestion
    .filter(matchesQuery)
```
A bit simpler

src/pages/settings/Settings.tsx
```diff
@@ -97,109 +136,48 @@ const Settings: FC = () => {
...
+  const rows: MainTableRow[] = configFields
     .filter((configField) => {
-      if (!query) {
-        return true;
-      }
-      return (
-        configField.key.toLowerCase().includes(query.toLowerCase()) ||
-        configField.shortdesc?.toLowerCase().includes(query.toLowerCase())
-      );
+      return matchesQuery(configField);
```
edlerd:
```suggestion
    .filter(matchesQuery)
```
A bit simpler.

edlerd (commented):
One last bit, then this should be good to merge.

PR #1590, opened 2025-11-03
feat(active project): Add active project in user settings, store in localStorage

### LINKED ISSUE #1564: [Idea] Ability to set default project
Hello,

LXD behavior when login it set the default project.

Image
 
It make confusing if we have instance running in another project while we also can't set another project to default-project either.

Image

Idea:
1. Able to set project-defualt when login to admin
2. Or LXD startup with All projects by default it will show all instances.

Image

but i think let user configure in LXD setting options is good.

kimanhou:
## Done

- Add a new user setting `user.ui_current_project`. Edit field is a dropdown list of projects, including All projects.
- Save `user.ui_current_project` in local storage.
- On login, if `user.ui_current_project` is correctly set, the current project is automatically selected.

Collateral improvements:
- On CertificateAdd page, add a Copy command button and display command on several lines.
- In LoggedInUserNotification, fix console error (p cannot be a descendant of p)

Fixes #1564

## QA

1. Run the LXD-UI:

## Screenshots

Screenshot from 2025-11-10 11-12-22

Screenshot from 2025-11-10 11-12-34

Screenshot from 2025-11-10 11-12-41

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,7 @@ import {
   Spinner,
 } from "@canonical/react-components";
 import { setFavicon } from "util/favicon";
+import { ALL_PROJECTS } from "util/loginProject";
 
 const CertificateAdd = lazy(async () => import("pages/login/CertificateAdd"));
 const CertificateGenerate = lazy(
@@ -172,7 +173,7 @@ const App: FC = () => {
             element={
               <Navigate
                 to={
-                  hasNoProjects
+                  hasNoProjects || defaultProject === ALL_PROJECTS
                     ? "/ui/all-projects/instances"
                     : `/ui/project/${encodeURIComponent(defaultProject)}/instances`
                 }
--- a/src/context/auth.tsx
+++ b/src/context/auth.tsx
@@ -6,6 +6,7 @@ import { fetchCertificates } from "api/certificates";
 import { fetchProjects } from "api/projects";
 import { fetchCurrentIdentity } from "api/auth-identities";
 import { useSupportedFeatures } from "./useSupportedFeatures";
+import { getLoginProject } from "util/loginProject";
 
 interface ContextProps {
   isAuthenticated: boolean;
@@ -81,10 +82,7 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
     enabled: settings?.auth === "trusted" && isFineGrained() !== null,
   });
 
-  const defaultProject =
-    projects.length < 1 || projects.find((p) => p.name === "default")
-      ? "default"
-      : projects[0].name;
+  const defaultProject = getLoginProject(projects);
 
   const isTls = settings?.auth_user_method === "tls";
 
--- a/src/pages/settings/Settings.tsx
+++ b/src/pages/settings/Settings.tsx
@@ -26,6 +26,8 @@ import { useServerEntitlements } from "util/entitlements/server";
 import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 import { useClusteredSettings } from "context/useSettings";
 import type { LXDSettingOnClusterMember } from "types/server";
+import { useProjects } from "context/useProjects";
+import { getDefaultProject } from "util/loginProject";
 
 const Settings: FC = () => {
   const docBaseLink = useDocs();
@@ -45,6 +47,8 @@ const Settings: FC = () => {
   });
   const { data: clusteredSettings = [], error: clusterError } =
     useClusteredSettings();
+
+  const { data: projects = [] } = useProjects();
 
   if (clusterError) {
     notify.failure("Loading clustered settings failed", clusterError);
@@ -96,14 +100,6 @@ const Settings: FC = () => {
   const configFields = toConfigFields(configOptions?.configs?.server ?? {});
 
   configFields.push({
-    key: "user.ui_title",
-    category: "user",
-    default: "",
-    shortdesc: "Title for the LXD-UI web page. Shows the hostname when unset.",
-    type: "string",
-  });
-
-  configFields.push({
     key: "user.ui_grafana_base_url",
     category: "user",
     default: "",
@@ -115,11 +111,27 @@ const Settings: FC = () => {
   });
 
   configFields.push({
+    key: "user.ui_login_project",
+    category: "user",
+    default: getDefaultProject(projects),
+    shortdesc: "Project to display on login.",
+    type: "string",
+  });
+
+  configFields.push({
     key: "user.ui_theme",
     category: "user",
     default: "",
     shortdesc:
       "Set UI to dark theme, light theme, or to match the system theme.",
+    type: "string",
+  });
+
+  configFields.push({
+    key: "user.ui_title",
+    category: "user",
+    default: "",
+    shortdesc: "Title for the LXD-UI web page. Shows the hostname when unset.",
     type: "string",
   });
 
--- a/src/pages/settings/LoginProjectSelect.tsx
+++ b/src/pages/settings/LoginProjectSelect.tsx
@@ -1,0 +1,129 @@
+import type { FC } from "react";
+import { useEffect, useState } from "react";
+import {
+  Button,
+  Form,
+  Icon,
+  Select,
+  useToastNotification,
+} from "@canonical/react-components";
+import { useProjects } from "context/useProjects";
+import {
+  ALL_PROJECTS,
+  getDefaultProject,
+  getLoginProject,
+  loadLoginProject,
+  saveLoginProject,
+} from "util/loginProject";
+import ResourceLabel from "components/ResourceLabel";
+import type { ConfigField } from "types/config";
+
+interface Props {
+  configField: ConfigField;
+}
+
+const LoginProjectSelect: FC<Props> = ({ configField }) => {
+  const [isEditMode, setEditMode] = useState(false);
+  const { data: projects = [] } = useProjects();
+  const [value, setValue] = useState<string>(loadLoginProject() || "");
+  const toastNotify = useToastNotification();
+
+  const projectOptions = [
+    {
+      label: "All projects",
+      value: ALL_PROJECTS,
+    },
+    ...projects.map((project) => ({
+      label: project.name,
+      value: project.name,
+    })),
+  ];
+
+  const canBeReset = String(configField.default) !== String(value);
+
+  const handleSave = () => {
+    saveLoginProject(value);
+
+    const settingLabel = (
+      <ResourceLabel bold type="setting" value={configField.key} />
+    );
+    toastNotify.success(<>Setting {settingLabel} updated.</>);
+    setEditMode(false);
+  };
+
+  const resetToDefault = () => {
+    setValue(configField.default || "");
+  };
+
+  const onCancel = () => {
+    setValue(loadLoginProject() || "");
+    setEditMode(false);
+  };
+
+  useEffect(() => {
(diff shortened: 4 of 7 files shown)


src/context/auth.tsx
```diff
@@ -81,10 +82,24 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
...
+      }
+      if (savedProject && projects.find((p) => p.name === savedProject)) {
+        return savedProject; // Return saved project if it exists in the list
+      }
+    }
+    // Fall back to original logic
+    return projects.length < 1 || projects.find((p) => p.name === "default")
       ? "default"
       : projects[0].name;
+  })();
```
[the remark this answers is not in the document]
kimanhou: I think it's fine, as the list of projects should not be long. Adding memoization would make the code more complicated for small value.

src/App.tsx
```diff
@@ -172,7 +172,7 @@ const App: FC = () => {
             element={
               <Navigate
                 to={
-                  hasNoProjects
+                  hasNoProjects || defaultProject === "all"
```
edlerd: This is problematic as `all` is a valid project name. Maybe use a special string with a space, I think spaces are invalid characters for project names. Then define it in a constant to refer to.

src/pages/settings/Settings.tsx
```diff
@@ -123,6 +128,15 @@ const Settings: FC = () => {
     type: "string",
   });
 
+  configFields.push({
+    key: "user.ui_current_project",
+    category: "user",
+    default: "",
+    shortdesc:
+      "Set project to current project on login. Type `__all_projects__` for all projects",
```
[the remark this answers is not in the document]
kimanhou: We decided to remove the 2nd sentence, as the label All projects is very clear.

kimanhou:
There is an issue in react-components library, which make the tests to fail.

src/pages/login/CertificateAdd.tsx
edlerd, line 1: The changes to this file seem unrelated to the PR title. Also the layout seems a step back, maybe undo or extract this into a dedicated PR.

src/pages/login/CertificateAddForm.tsx
edlerd, line 1: Unrelated changes, please extract to a dedicated PR.

src/pages/permissions/panels/LoggedInUserNotification.tsx
edlerd, line 1: Unrelated changes, please extract to a dedicated PR.

src/pages/settings/Settings.tsx
```diff
@@ -123,6 +123,14 @@ const Settings: FC = () => {
     type: "string",
   });
 
+  configFields.push({
+    key: "user.ui_current_project",
+    category: "user",
+    default: "",
+    shortdesc: "Set the project to display on login.",
```
edlerd:
```suggestion
    shortdesc: "Project to display on login.",
```
A bit more clear?

src/pages/settings/CurrentProjectSelect.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const configField = {
+    key: "user.ui_current_project",
+    default: "",
+  };
+
+  const projectOptions = [
+    {
+      label: "Select a project...",
+      value: "",
+    },
```
edlerd:
```suggestion
```
I think the default value should be the value from the `defaultProject` function in `auth.tsx`: either the `default` project or the first project the user has access to.

We probably need to export the `defaultProject` function from auth.tsx so we can use it here to calculate the default.

And then we don't need the "Select a project..." option.

src/pages/settings/CurrentProjectSelect.tsx
```diff
@@ -0,0 +1,146 @@
...
+              hasIcon
+              disabled={!canEditServerConfiguration()}
+              title={
+                canEditServerConfiguration()
+                  ? ""
+                  : "You do not have permission to edit server configuration"
+              }
+            >
+              <div className="readmode-value u-truncate">
+                {value === ALL_PROJECTS ? "All projects" : value || "-"}
```
edlerd: Here we should display the calculated default from the `defaultProject` function.

src/context/auth.tsx
```diff
@@ -81,10 +83,19 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
...
+      return ALL_PROJECTS;
+    }
+    if (savedProject && projects.find((p) => p.name === savedProject)) {
+      return savedProject;
+    }
+
+    return projects.length < 1 || projects.find((p) => p.name === "default")
       ? "default"
       : projects[0].name;
+  })();
```
edlerd: The logic for calculating the default depending on available projects per user (the old scope of this function) should probably be extracted into a function we can then use to also calculate the default for the new setting.

edlerd (commented):
Functionality seems all right, thanks for the iteration. Some ideas on a better presentation of the default value. Please extract the unrelated changes to a dedicated PR.

src/context/auth.tsx
```diff
@@ -6,6 +6,8 @@ import { fetchCertificates } from "api/certificates";
 import { fetchProjects } from "api/projects";
 import { fetchCurrentIdentity } from "api/auth-identities";
 import { useSupportedFeatures } from "./useSupportedFeatures";
+import { ALL_PROJECTS } from "util/projects";
+import { loadCurrentProject } from "pages/settings/CurrentProjectSelect";
```
edlerd: I suspect this line is causing the CI failures as it creates a circular dependency from `auth.tsx` to `CurrentProjectSelect.tsx` back to `auth.tsx`.
To resolve it best move the function (and maybe also the ALL_PROJECTS constant) into a new file under utils, maybe `utils/currentProject.tsx`.

src/pages/settings/CurrentProjectSelect.tsx
kimanhou: I am not sure I understood correctly but I refactored the code.

src/context/auth.tsx
kimanhou: Yes !

src/util/projects.tsx
```diff
@@ -59,3 +60,17 @@ export const isProjectWithProfiles = (project?: LxdProject): boolean =>
...
+  if (savedProject === ALL_PROJECTS) {
+    return ALL_PROJECTS;
+  }
+  if (savedProject && projects.find((p) => p.name === savedProject)) {
+    return savedProject;
+  }
+
+  return projects.length < 1 || projects.find((p) => p.name === "default")
+    ? "default"
+    : projects[0].name;
```
edlerd: We could extract this bit into its own dedicated function to calculate the default without user override. Though not sure how to name the current `getDefaultProject` and the new extracted function. But the extracted function will be helpful in another part (see other comment).

src/pages/settings/Settings.tsx
```diff
@@ -123,6 +123,14 @@ const Settings: FC = () => {
     type: "string",
   });
 
+  configFields.push({
+    key: "user.ui_current_project",
+    category: "user",
+    default: "",
```
edlerd, line 133: Here we should call the new extracted function, so we display the correct value instead of "-".

src/pages/settings/CurrentProjectSelect.tsx
```diff
@@ -0,0 +1,144 @@
...
+              onClick={() => {
+                setEditMode(true);
+              }}
+              hasIcon
+              disabled={!canEditServerConfiguration()}
+              title={
+                canEditServerConfiguration()
+                  ? ""
+                  : "You do not have permission to edit server configuration"
+              }
```
edlerd:
```suggestion
```
This is only local storage, so not depending on permissions.

src/pages/settings/CurrentProjectSelect.tsx
```diff
@@ -0,0 +1,144 @@
...
+          )}
+        </Form>
+      )}
+      {!isEditMode && (
+        <>
+          {isRestricted ? (
+            <span>
+              {value === ALL_PROJECTS ? "All projects" : (value ?? "-")}
+            </span>
+          ) : (
```
edlerd:
```suggestion
```
I think this setting is never restricted, so we can simplify here.

src/pages/settings/CurrentProjectSelect.tsx
```diff
@@ -0,0 +1,144 @@
...
+              hasIcon
+              disabled={!canEditServerConfiguration()}
+              title={
+                canEditServerConfiguration()
+                  ? ""
+                  : "You do not have permission to edit server configuration"
+              }
+            >
+              <div className="readmode-value u-truncate">
+                {value === ALL_PROJECTS ? "All projects" : value || "-"}
```
edlerd: Instead of falling back to "-" I think we should fall back to the new extracted function.

edlerd (commented):
This seems nearly perfect, just some more simplification and some improvement idea to present the correct default to the user.

Also, I think we might want to sort the user keys alphabetically. Currently, they are ordered a bit randomly Wdyt?

src/util/projects.tsx
kimanhou: What about `getDefaultProject` (unchanged) and `getFallbackProject` ?

```
const getFallbackProject = (projects: LxdProject[]) => {
  return projects.length < 1 || projects.find((p) => p.name === "default")
    ? "default"
    : projects[0].name;
};

export const getDefaultProject = (projects: LxdProject[]) => {
  const savedProject = loadCurrentProject();
  if (savedProject === ALL_PROJECTS) {
    return ALL_PROJECTS;
  }
  if (savedProject && projects.find((p) => p.name === savedProject)) {
    return savedProject;
  }

  return getFallbackProject(projects);
};
```

src/pages/settings/CurrentProjectSelect.tsx
```diff
@@ -0,0 +1,128 @@
...
+const CurrentProjectSelect: FC = () => {
+  const [isEditMode, setEditMode] = useState(false);
+  const { data: projects = [] } = useProjects();
+  const [value, setValue] = useState<string>(loadCurrentProject() || "");
+  const toastNotify = useToastNotification();
+
+  const configField = {
+    key: "user.ui_current_project",
+    default: "",
+  };
```
edlerd: We shouldn't redefine this here, better pass it in from the parent.
A side effect of redefining it here is after hitting "reset to default" we are currently showing the wrong project in the dropdown.

src/util/projects.tsx
```diff
@@ -59,3 +60,24 @@ export const isProjectWithProfiles = (project?: LxdProject): boolean =>
...
+export const getFallbackProject = (projects: LxdProject[]) => {
+  const project =
+    projects.length < 1 || projects.find((p) => p.name === "default")
+      ? "default"
+      : projects[0].name;
+
+  return project;
+};
+
+export const getDefaultProject = (projects: LxdProject[]) => {
```
edlerd: Shall we call these `getLoginProject` and the upper one `getRawLoginProject` and move them into the `loginProject.ts` file?

edlerd (dismissed):
QA and code LGTM, thanks for the adjustments and driving this to perfection. One last idea for adjusting the function naming and placement, good to merge as is as well.

src/util/projects.tsx
kimanhou: What about `getLoginProject` and `getDefaultProject` ?

edlerd: Yes, that is better indeed.

PR #1591, opened 2025-11-03
refactor(ui) extract appProviders to simplify Root

edlerd:
Encapsulate all global context providers in appProviders.tsx to eliminate deep nesting in Root.tsx and improve code organization.

## Changes
- Create src/context/appProviders.tsx to compose all global providers
- Simplify Root.tsx from 9 levels of provider nesting to single AppProviders component
- Inline CombinedNotificationProvider and move useLocation hook into appProviders
- Remove unused CombinedNotificationProvider.tsx


kimanhou (approved):
That's way simpler, thank you !

PR #1600, opened 2025-11-11
feat(force delete project): list content

kimanhou:
## Done

- Enable `Delete project` button for non empty projects
- List content in the confirmation modal
- Make the user type project name to confirm deletion both for empty and non empty projects
- Keep behavior the same on LXD versions that don't support force deletion (force delete project was introduced in 6.6)
- Add tests for old and new process

## QA

1. Run the LXD-UI:
    - Running against latest edge, delete empty project: you should have to type the project name
    - Running against latest edge, delete non empty project: force delete should work
    - Repeat those steps while running against an version that doesn't support force delete (5.21 for example). You should be able to delete an empty project. The delete project button should be disabled for non empty project, with a tooltip listing its content.

## Screenshots

modal-wip
modal-wip-2
</br></br>

When deleting an empty project:
empty

### DIFF
--- a/src/api/projects.tsx
+++ b/src/api/projects.tsx
@@ -85,8 +85,17 @@ export const renameProject = async (
     });
 };
 
-export const deleteProject = async (project: LxdProject): Promise<void> => {
-  await fetch(`/1.0/projects/${encodeURIComponent(project.name)}`, {
+export const deleteProject = async (
+  project: LxdProject,
+  force?: boolean,
+): Promise<void> => {
+  const params = new URLSearchParams();
+  if (force) {
+    params.set("force", "1");
+  }
+  const url = `/1.0/projects/${encodeURIComponent(project.name)}?${params.toString()}`;
+
+  await fetch(url, {
     method: "DELETE",
   }).then(handleResponse);
 };
--- a/src/pages/projects/actions/DeleteProjectBtn.tsx
+++ b/src/pages/projects/actions/DeleteProjectBtn.tsx
@@ -8,25 +8,45 @@ import { useQueryClient } from "@tanstack/react-query";
 import { isProjectEmpty } from "util/projects";
 import { useIsScreenBelow } from "context/useIsScreenBelow";
 import {
-  ConfirmationButton,
+  Button,
   Icon,
   Tooltip,
   useNotify,
+  usePortal,
   useToastNotification,
 } from "@canonical/react-components";
-import classnames from "classnames";
-import { filterUsedByType } from "util/usedBy";
-import type { ResourceType } from "util/resourceDetails";
 import ResourceLabel from "components/ResourceLabel";
 import { useProjectEntitlements } from "util/entitlements/projects";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
+import DeleteProjectModal from "./DeleteProjectModal";
+import { filterUsedByType } from "util/usedBy";
+import type { ResourceType } from "util/resourceDetails";
 
-interface Props {
-  project: LxdProject;
-}
+const generateTooltipMessage = (
+  project: LxdProject,
+  canDelete: boolean,
+  isDefault: boolean,
+  isEmpty: boolean,
+  hasForceDelete: boolean,
+): ReactNode => {
+  if (!canDelete) {
+    return "You do not have permission to delete this project";
+  }
 
-const generateProjectUsedByTooltip = (project: LxdProject) => {
-  const resourceLabelAndLink: Record<string, { label: string; link: string }> =
-    {
+  if (isDefault) {
+    return "The default project cannot be deleted";
+  }
+
+  if (isEmpty) {
+    return "Delete project";
+  }
+
+  if (!hasForceDelete) {
+    // Non-empty project without force delete support - show detailed tooltip
+    const resourceLabelAndLink: Record<
+      string,
+      { label: string; link: string }
+    > = {
       instance: {
         label: "Instances",
         link: `/ui/project/${encodeURIComponent(project.name)}/instances`,
@@ -45,39 +65,51 @@ const generateProjectUsedByTooltip = (project: LxdProject) => {
       },
     };
 
-  const resourceTypes = Object.keys(resourceLabelAndLink);
-  const usedByItems: ReactNode[] = [];
-  for (const resourceType of resourceTypes) {
-    const usedBy = filterUsedByType(
-      resourceType as ResourceType,
-      project.used_by?.filter(
-        // the default profile is not blocking project deletion and can't be removed itself
-        (item) => !item.startsWith("/1.0/profiles/default"),
-      ),
-    );
+    const resourceTypes = Object.keys(resourceLabelAndLink);
+    const usedByItems: ReactNode[] = [];
 
-    if (usedBy.length > 0) {
-      const label = resourceLabelAndLink[resourceType].label;
-      const link = resourceLabelAndLink[resourceType].link;
-      usedByItems.push(
-        <li
-          key={resourceType}
-          className="p-list__item is-dark u-no-margin--bottom"
-        >
-          {<Link to={link}>{label}</Link>} ({usedBy.length})
-        </li>,
+    for (const resourceType of resourceTypes) {
+      const usedBy = filterUsedByType(
+        resourceType as ResourceType,
+        project.used_by?.filter(
+          (item) => !item.startsWith("/1.0/profiles/default"),
+        ),
       );
+
+      if (usedBy.length > 0) {
+        const label = resourceLabelAndLink[resourceType].label;
+        const link = resourceLabelAndLink[resourceType].link;
+        usedByItems.push(
+          <li
+            key={resourceType}
+            className="p-list__item is-dark u-no-margin--bottom"
+          >
+            <Link to={link}>{label}</Link> ({usedBy.length})
+          </li>,
+        );
+      }
     }
+
+    return (
+      <>
+        Cannot delete non-empty project.
+        <br />
+        <br />
+        Project is used by:
+        <ul className="p-list u-no-margin--bottom">{usedByItems}</ul>
+        <br />
+        Remove all resources first, or upgrade to LXD 6.6 or newer to use force
+        deletion.
+      </>
+    );
   }
 
-  return (
-    <>
-      Non-empty project cannot be deleted.
-      <p className="u-no-margin--bottom">Project is used by:</p>
-      <ul className="p-list u-no-margin--bottom">{usedByItems}</ul>
-    </>
-  );
+  return "Delete project and all its resources";
 };
+
+interface Props {
+  project: LxdProject;
+}
 
 const DeleteProjectBtn: FC<Props> = ({ project }) => {
   const isSmallScreen = useIsScreenBelow();
@@ -87,25 +119,21 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
   const [isLoading, setLoading] = useState(false);
   const navigate = useNavigate();
   const { canDeleteProject } = useProjectEntitlements();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const { hasProjectForceDelete } = useSupportedFeatures();
 
   const isDefaultProject = project.name === "default";
   const isEmpty = isProjectEmpty(project);
-  const getHoverText = () => {
-    if (!canDeleteProject(project)) {
-      return "You do not have permission to delete this project";
-    }
-    if (isDefaultProject) {
-      return "The default project cannot be deleted";
-    }
-    if (!isEmpty) {
-      return "";
-    }
-    return "Delete project";
+
+  const handleClosePortal = () => {
+    notify.clear();
+    closePortal();
   };
 
   const handleDelete = () => {
     setLoading(true);
-    deleteProject(project)
+    const force = !isEmpty && hasProjectForceDelete;
+    deleteProject(project, force)
       .then(() => {
         navigate(`/ui/project/default/instances`);
         toastNotify.success(
@@ -114,10 +142,22 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
             deleted.
           </>,
         );
+        handleClosePortal();
       })
       .catch((e) => {
         setLoading(false);
         notify.failure("Project deletion failed", e);
+
+        // Scroll to top of modal to show error notification
+        const modalDialog = document.querySelector(
+          ".delete-project-dialog .p-modal__dialog",
+        );
+        if (modalDialog) {
+          modalDialog.scrollTo({
+            top: 0,
+            behavior: "smooth",
+          });
+        }
       })
       .finally(() => {
         queryClient.invalidateQueries({
@@ -127,41 +167,43 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
   };
 
   return (
-    <ConfirmationButton
-      onHoverText={getHoverText()}
-      className={classnames("u-no-margin--bottom", {
-        "has-icon": !isSmallScreen,
-      })}
-      loading={isLoading}
-      disabled={
-        !canDeleteProject(project) || isDefaultProject || !isEmpty || isLoading
-      }
-      confirmationModalProps={{
-        title: "Confirm delete",
-        confirmButtonLabel: "Delete",
-        onConfirm: handleDelete,
-        children: (
-          <p>
-            This will permanently delete project{" "}
-            <ResourceLabel type="project" value={project.name} bold />.<br />
-            This action cannot be undone, and can result in data loss.
-          </p>
-        ),
-      }}
-      shiftClickEnabled
-      showShiftClickHint
-    >
-      <Tooltip
-        message={
-          !isEmpty && !isDefaultProject
-            ? generateProjectUsedByTooltip(project)
-            : ""
+    <>
+      <Button
+        onClick={openPortal}
+        hasIcon={!isSmallScreen}
+        disabled={
+          !canDeleteProject(project) ||
+          isDefaultProject ||
+          isLoading ||
+          (!isEmpty && !hasProjectForceDelete)
         }
+        className="u-no-margin--bottom"
       >
-        {!isSmallScreen && <Icon name="delete" />}
-        <span>Delete project</span>
-      </Tooltip>
-    </ConfirmationButton>
+        <Tooltip
+          message={generateTooltipMessage(
+            project,
+            canDeleteProject(project),
+            isDefaultProject,
+            isEmpty,
+            hasProjectForceDelete,
+          )}
+        >
+          {!isSmallScreen && <Icon name="delete" />}
+          <span>Delete project</span>
+        </Tooltip>
+      </Button>
+
+      {isOpen && (
+        <Portal>
+          <DeleteProjectModal
+            project={project}
+            handleDelete={handleDelete}
+            isLoading={isLoading}
+            closePortal={handleClosePortal}
+          />
+        </Portal>
+      )}
+    </>
   );
 };
 
--- a/src/pages/projects/actions/DeleteProjectModal.tsx
+++ b/src/pages/projects/actions/DeleteProjectModal.tsx
@@ -1,0 +1,86 @@
+import ResourceLabel from "components/ResourceLabel";
+import type { ChangeEvent, FC, KeyboardEvent } from "react";
+import { useState } from "react";
+import type { LxdProject } from "types/project";
+import { isProjectEmpty } from "util/projects";
+import ProjectUsedBy from "pages/projects/actions/ProjectUsedBy";
+import {
+  ActionButton,
+  Input,
+  Modal,
+  NotificationConsumer,
+} from "@canonical/react-components";
+
+interface Props {
+  project: LxdProject;
+  handleDelete: () => void;
+  isLoading: boolean;
+  closePortal: () => void;
+}
+
+const DeleteProjectModal: FC<Props> = ({
+  project,
+  handleDelete,
+  isLoading,
+  closePortal,
+}) => {
+  const [disableConfirm, setDisableConfirm] = useState(true);
+
+  const isEmpty = isProjectEmpty(project);
+
+  const handleConfirmInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setDisableConfirm(e.target.value !== project.name);
+  };
+
+  const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Enter" && !disableConfirm && !isLoading) {
+      handleDelete();
+    }
+  };
+
+  return (
+    <Modal
+      title="Confirm delete"
+      className="delete-project-dialog"
+      close={closePortal}
+      buttonRow={[
+        <span className="u-float-left confirm-input" key="confirm-input">
+          <Input
+            name="confirm-delete-project-input"
+            type="text"
+            onChange={handleConfirmInputChange}
+            onKeyDown={handleKeyDown}
+            placeholder={project.name}
+            className="u-no-margin--bottom"
+          />
+        </span>,
+        <ActionButton
+          key="confirm-action-button"
+          appearance="negative"
+          className="u-no-margin--bottom"
+          onClick={handleDelete}
+          loading={isLoading}
+          disabled={disableConfirm}
+        >
+          Permanently delete {project.name}
+        </ActionButton>,
+      ]}
+    >
+      <NotificationConsumer />
+      <p>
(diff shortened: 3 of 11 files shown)


src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -126,41 +72,63 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
...
-      disabled={
-        !canDeleteProject(project) || isDefaultProject || !isEmpty || isLoading
-      }
+      disabled={!canDeleteProject(project) || isDefaultProject || isLoading}
       confirmationModalProps={{
-        title: "Confirm delete",
+        className: "delete-project-dialog",
+        title: (
+          <>
+            Delete project <b>{project.name}</b>
```
edlerd: I think we don't need the project name in the header, maybe keep this as "Confirm delete" as it was? Otherwise we repeat the project name a bit too often in this modal.

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -126,41 +72,63 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
...
+          <div>
             This will permanently delete project{" "}
             <ResourceLabel type="project" value={project.name} bold />.<br />
             This action cannot be undone, and can result in data loss.
-          </p>
+            <br />
+            <br />
+            {!isEmpty && (
+              <>
+                By deleting project <b>{project.name}</b>, you will also delete:
```
edlerd: this should either reuse `ResourceLabel` for the project name, or maybe we can avoid repeating the project name in the body text and shorten this to "You will also delete:".

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -126,41 +72,63 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
...
-          !isEmpty && !isDefaultProject
-            ? generateProjectUsedByTooltip(project)
-            : ""
-        }
-      >
-        {!isSmallScreen && <Icon name="delete" />}
-        <span>Delete project</span>
-      </Tooltip>
+      {!isSmallScreen && <Icon name="delete" />}
+      <span>Delete project</span>
```
edlerd: Maybe make this more clear as "Permanently delete project ${foo} and 23 contents". Or think of a better caption.

src/pages/projects/actions/ProjectUsedByList.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }, [instances, profiles, images, volumes]);
+
+  return (
+    <>
+      <MainTable
+        headers={[{ content: "type" }, { content: "name" }]}
+        rows={tableRows}
+        className="p-main-table delete-project-table"
+      />
+    </>
```
edlerd: We don't need the fragment, but my main concern is: It might be simpler to rely on a raw table instead of using the main table. See how the used by list on a storage pool is built, like in [this example](https://lxd-ui-1600.demos.haus/ui/project/pr1/storage/pool/default). Or check `StorageUsedBy.tsx`, which might be more react like and does the collapsing if more than 5 of one entity type are present.

edlerd: Actually I think you can reuse most of the components there, like the `UsedByItem` and `ExpandableList` as well as the `filterUsedByType` helpers should be applicable here just as they are.

src/pages/projects/actions/DeleteProjectBtn.tsx
kimanhou: I was following Juju's design. I do like having the name of the project in the title, however it would mean we have to align all similar modals ("Confirm delete" is the title of 16 modals), which would be too much trouble. But something to discuss on Pragma maybe ?

src/pages/projects/actions/DeleteProjectBtn.tsx
kimanhou: I replaced it with "The following items will also be deleted:"

src/pages/projects/actions/DeleteProjectBtn.tsx
kimanhou: In a tooltip, right ? Not the label of the button ?

edlerd: I was thinking the button label, but on second thought it would be way too long. Hrm, not sure. Maybe shorter is better. I'd avoid a tooltip, we already have enough space in the modal body.

src/pages/projects/actions/DeleteProjectBtn.tsx
edlerd: I am not too concerned of keeping all modals aligned. More that the project name is repeated too often in the same view.

src/pages/projects/actions/DeleteProjectBtn.tsx
kimanhou: How flexible are you on the current location of this button ?
I find it weird to have it at the same place as the "Create" buttons from other screens and it looks so lonely and sad in that corner.
Yes, it's an action, but it's not a "quick action" in the way that it's not used as often as "Start an instance" for example.
I like what Github does in the Settings of a repository: a danger zone at the very end that handles deletion and other permanent actions. What do you think ?

image

edlerd: I think in the scope of a modal, the modal footer is the right place for actions. Even destructive ones. This seems like an expected pattern for confirmation.

kimanhou: I'm talking about the configuration screen of the project, not the modal.
An idea would also be to add the danger zone in the menu, underneath "Restrictions".

image

edlerd: That would break with the pattern on other pages like instances, profiles, networks. Actions are always on the top right aligned. So I wouldn't break the pattern for project details.

kimanhou:
Should we make the user type the name of the project when the project is empty ?

src/api/projects.tsx
```diff
@@ -85,8 +85,13 @@ export const renameProject = async (
     });
 };
 
-export const deleteProject = async (project: LxdProject): Promise<void> => {
-  await fetch(`/1.0/projects/${encodeURIComponent(project.name)}`, {
+export const deleteProject = async (
+  project: LxdProject,
+  force?: boolean,
+): Promise<void> => {
+  const url = `/1.0/projects/${encodeURIComponent(project.name)}${force ? "?force=true" : ""}`;
```
edlerd: Shall we use `params` and the below suggestion to keep it consistent with other endpoints?

```suggestion
  const params = new URLSearchParams()
  if (force) {
    params.set("force", "1");
  }
  const url = `/1.0/projects/${encodeURIComponent(project.name)}?${params.toString()}`;
```

src/pages/projects/actions/ProjectUsedByList.tsx
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  return (
+    <table className="p-main-table delete-project-table">
+      <thead>
+        <tr role="row">
+          <th role="columnheader">type</th>
+          <th role="columnheader">name</th>
+        </tr>
+      </thead>
```
edlerd: We probably don't need the headings, wdyt?

src/sass/_project_configuration.scss
```diff
@@ -0,0 +1,44 @@
+.delete-project-dialog {
+  .p-modal__dialog {
+    left: 50%;
+    position: absolute;
+    right: 50%;
+    transform: translateX(-50%);
+    width: 50%;
```
edlerd: This seems too much customization. I think what you probably want to do is something like this:

```
  .p-modal__dialog {
    @include large {
      width: 35rem;
    }
  }
```

src/sass/_project_configuration.scss
```diff
@@ -0,0 +1,44 @@
...
+
+    .delete-project-modal-notification-row {
+      padding: 0 2px;
+    }
```
edlerd:
```suggestion
```
Use `<NotificationConsumer />` directly, without the `NotificationRow` wrapper and then the placement should be fine out of the box.

src/sass/_project_configuration.scss
```diff
@@ -0,0 +1,44 @@
...
+          th:first-child {
+            width: 16rem;
+          }
+        }
+      }
+
+      .icon {
+        margin-right: 10px;
+      }
+    }
```
edlerd, line 23: This should be responsive, on very narrow screens the chip's contents are fully cut.

src/sass/_project_configuration.scss
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+@media screen and (max-width: $breakpoint-large) {
+  .delete-project-dialog {
+    .p-modal__dialog {
+      width: 80%;
+    }
+  }
+}
```
edlerd:
```suggestion
```
Probably not needed, see above comment.

tests/projects.spec.ts
```diff
@@ -172,3 +178,110 @@ test("retain custom project selection on browsing pages for all projects", async
 
   expect(page.getByText("Project" + project)).toBeVisible();
 });
+
+test("project deletion with instances - force delete supported", async ({
+  page,
+  lxdVersion,
+}) => {
+  // projects_force_delete was introduced in 6.6
```
edlerd:
```suggestion
```
I think the message below explains this already.

tests/projects.spec.ts
```diff
@@ -172,3 +178,110 @@ test("retain custom project selection on browsing pages for all projects", async
...
+    "LXD versions 6.6 and newer support projects_force_delete API extension",
+  );
+
+  const project = randomProjectName();
+  const instance = randomInstanceName();
+
+  await createProject(page, project);
+  await createInstance(page, instance, "container", project);
+  await page.getByRole("link", { name: "Configuration" }).click();
+  await page.waitForTimeout(2000); // Wait for the form state to be fully loaded
```
edlerd: Instead of a timeout, better wait for an element on the screen to be visible that indicates we moved to the project list. Such as the text `Project configuration`. This has the advantages that a) we don't slow down quick test runs and b) we don't break the test if navigation takes longer than 2 seconds.

tests/projects.spec.ts
```diff
@@ -172,3 +178,110 @@ test("retain custom project selection on browsing pages for all projects", async
...
+  ).toBeVisible();
+  await expect(page.getByText("Instances (1)")).toBeVisible();
+  await expect(page.getByText(instance)).toBeVisible();
+
+  await page.getByPlaceholder(project).fill(project);
+  const confirmDeleteButton = page
+    .getByRole("dialog", { name: "Confirm delete" })
+    .getByRole("button", { name: `Permanently delete ${project}` });
+
+  await expect(confirmDeleteButton).toBeEnabled();
```
edlerd:
```suggestion
```
I think we don't need this, the click action below would fail if it were not enabled. But maybe we should expect it to be disabled before typing the project name?

tests/projects.spec.ts
```diff
@@ -172,3 +178,110 @@ test("retain custom project selection on browsing pages for all projects", async
...
+  await confirmDeleteButton.click();
+
+  await page.waitForSelector(`text=Project ${project} deleted.`);
+  await expect(page.getByRole("button", { name: "default" })).toBeVisible();
```
edlerd:
```suggestion
```
Not sure why or what this button is about in relation to the test?

src/pages/projects/actions/ProjectUsedBy.tsx
```diff
@@ -0,0 +1,144 @@
...
+                    item={item}
+                    activeProject={project.name}
+                    type="instance"
+                    to={`/ui/project/${encodeURIComponent(item.project)}/instance/${encodeURIComponent(item.name)}`}
+                  />
+                ))}
+              />
+            </td>
+          </tr>
+        )}
```
edlerd, line 57: Those blocks are also frequently used in `NetworkAclUsedBy` and `StorageUsedBy`. I also had a new use case in the open placement groups PR. Shall we extract them as a reusable component?

src/pages/projects/actions/ProjectUsedByList.tsx
```diff
@@ -0,0 +1,144 @@
...
+const INSTANCES = "Instances";
+const PROFILES = "Profiles";
+const IMAGES = "Images";
+const CUSTOM_VOLUMES = "Custom volumes";
+
+interface Props {
+  project: LxdProject;
+}
+
+const ProjectUsedByList: FC<Props> = ({ project }) => {
```
edlerd:
```suggestion
const ProjectUsedBy FC<Props> = ({ project }) => {
```
Better keep the naming aligned with other components.

src/pages/projects/actions/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+          className="u-no-margin--bottom"
+          onClick={handleDelete}
+          loading={isLoading}
+          disabled={disableConfirm}
+        >
+          {`Permanently delete ${project.name}`}
+        </ActionButton>,
+      ]}
+    >
+      <div>
```
edlerd:
```suggestion
```
Probably can keep the DOM a bit simpler?

src/pages/projects/actions/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <NotificationRow className="delete-project-modal-notification-row" />
```
edlerd: As suggested above, I think better use `<NotificationConsumer />` directly here.

src/pages/projects/actions/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+        This will permanently delete project{" "}
+        <ResourceLabel type="project" value={project.name} bold />.<br />
+        This action cannot be undone, and can result in data loss.
+        <br />
+        <br />
```
edlerd:
```suggestion
        <p>
          This will permanently delete project{" "}
          <ResourceLabel type="project" value={project.name} bold />.<br />
          This action cannot be undone, and can result in data loss.
        </p>
```
I think this gives better spacing and semantics.

src/pages/projects/actions/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+        </span>,
+        <ActionButton
+          key="confirm-action-button"
+          appearance="negative"
+          className="u-no-margin--bottom"
+          onClick={handleDelete}
+          loading={isLoading}
+          disabled={disableConfirm}
+        >
+          {`Permanently delete ${project.name}`}
```
edlerd:
```suggestion
          Permanently delete {project.name}
```
A bit simpler

src/pages/projects/actions/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+  closePortal: () => void;
+}
+
+const DeleteProjectModal: FC<Props> = ({
+  project,
+  handleDelete,
+  isLoading,
+  closePortal,
+}) => {
+  const [confirmInput, setConfirmInput] = useState("");
```
edlerd:
```suggestion
```
I think we can rely on the Input to keep its state and don't need to control it from here.

src/pages/projects/actions/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [disableConfirm, setDisableConfirm] = useState(true);
+
+  const isEmpty = isProjectEmpty(project);
+
+  const handleConfirmInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    if (e.target.value === project.name) {
+      setDisableConfirm(false);
+    } else {
+      setDisableConfirm(true);
+    }
```
edlerd:
```suggestion
    setDisableConfirm(e.target.value !== project.name);
```
A bit simpler?

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -45,40 +66,52 @@ const generateProjectUsedByTooltip = (project: LxdProject) => {
...
+            <Link to={link}>{label}</Link> ({usedBy.length})
+          </li>,
+        );
+      }
     }
+
+    return (
+      <>
+        Cannot delete non-empty project. This LXD version does not support force
+        deletion.
```
edlerd:
```suggestion
        Cannot delete non-empty project.
```
I think it is enough to mention the new feature below.

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -87,25 +120,23 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
   const [isLoading, setLoading] = useState(false);
   const navigate = useNavigate();
   const { canDeleteProject } = useProjectEntitlements();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const { hasProjectForceDelete } = useSupportedFeatures();
+  const { data: freshProject } = useProject(project.name);
+  const currentProject = freshProject || project; // Fallback to prop if query hasn't loaded yet
```
edlerd:
```suggestion
```
This shouldn't be needed. We receive the project as a prop already?

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -127,41 +170,45 @@ const DeleteProjectBtn: FC<Props> = ({ project }) => {
...
-      <Tooltip
-        message={
-          !isEmpty && !isDefaultProject
-            ? generateProjectUsedByTooltip(project)
-            : ""
+    <>
+      <Button
+        onClick={(e) => {
+          openPortal(e);
+        }}
```
edlerd:
```suggestion
        onClick={openPortal}
```
A bit simpler

src/pages/projects/actions/DeleteProjectBtn.tsx
```diff
@@ -45,40 +66,52 @@ const generateProjectUsedByTooltip = (project: LxdProject) => {
...
+    return (
+      <>
+        Cannot delete non-empty project. This LXD version does not support force
+        deletion.
+        <br />
+        <br />
+        Project is used by:
+        <ul className="p-list u-no-margin--bottom">{usedByItems}</ul>
+        <br />
+        Remove all resources first, or upgrade LXD to use force deletion.
```
edlerd:
```suggestion
        Remove all resources first, or upgrade to LXD 6.6 or newer to use force deletion.
```
A bit more specific.

src/pages/projects/actions/ProjectUsedByList.tsx
kimanhou: I like the heading but don't have a strong opinion on it. I removed it.

tests/projects.spec.ts
kimanhou: Sorry, missed that

src/pages/projects/actions/ProjectUsedBy.tsx
kimanhou: YES !

src/pages/projects/actions/DeleteProjectModal.tsx
kimanhou: What about the following sentence "The following items will also be deleted:" ? Should it also be in a paragraph ?

edlerd: No, I think it is fine to keep it outside so that sentence has some space above and is closer to the content it refers to below.

src/pages/projects/actions/DeleteProjectModal.tsx
kimanhou: I copied the logic from DeleteGroupModal. We might want to change stuff in there too, like extra curly brackets.

src/pages/projects/actions/DeleteProjectBtn.tsx
kimanhou: It doesn't get updated properly on older versions after you delete content. For example, you go to the configuration of a non empty project, the delete button is disabled because there is an instance. Delete the instance and come back to the configuration page, the delete button is still disabled. After page refresh, the delete button is enabled.

src/pages/projects/actions/ProjectUsedBy.tsx
kimanhou: Let me do this in a separate PR, and then refactor the other places.

src/pages/projects/actions/DeleteProjectBtn.tsx
edlerd: I see, seems like our cache invalidation rules are not accounting for the project entity on deletion. The right solution should be to invalidate the project query cache on deletion of instance, profile, volume or alike to ensure the used by field gets updated.

I am surprised this fixes it, because we do the same `useProject` call in the `ProjectProvider`, so this call in here should also hit the cache.

Wdyt?

kimanhou: Actually after testing, it turns we don't need to refresh the project. It works now.

src/pages/projects/actions/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+          This action cannot be undone, and can result in data loss.
+        </p>
+        {!isEmpty && (
+          <>
+            The following items will also be deleted:
+            <ProjectUsedBy project={project} />
+          </>
+        )}
+        <p>To continue, please type the project name below.</p>
+      </>
```
edlerd:
```suggestion
      <NotificationConsumer />
      <p>
        This will permanently delete project{" "}
        <ResourceLabel type="project" value={project.name} bold />.<br />
        This action cannot be undone, and can result in data loss.
      </p>
      {!isEmpty && (
        <>
          The following items will also be deleted:
          <ProjectUsedBy project={project} />
        </>
      )}
      <p>To continue, please type the project name below.</p>
```
A bit simpler without the fragment

tests/projects.spec.ts
```diff
@@ -172,3 +178,111 @@ test("retain custom project selection on browsing pages for all projects", async
...
+  await confirmDeleteButton.click();
+
+  await page.waitForSelector(`text=Project ${project} deleted.`);
+});
+
+test("project deletion with instances - force delete not supported", async ({
+  page,
+  lxdVersion,
+}) => {
+  // projects_force_delete was introduced in 6.6
```
edlerd:
```suggestion
```
Probably clear from the message below.

tests/projects.spec.ts
```diff
@@ -172,3 +178,111 @@ test("retain custom project selection on browsing pages for all projects", async
...
+      "Remove all resources first, or upgrade LXD to use force deletion.",
+    ),
+  ).toBeVisible();
+
+  await deleteInstance(page, instance, project);
+  await deleteAllImages(page, project);
+
+  await page.getByRole("link", { name: "Configuration" }).click();
+  await page.waitForLoadState("networkidle");
+  await page.waitForTimeout(2000);
```
edlerd:
```suggestion
  await page.waitForSelector("text=Project configuration");
```
Similar to the other test.

src/sass/_project_configuration.scss
```diff
@@ -0,0 +1,39 @@
...
+
+            @include mobile-and-tablet {
+              width: 9rem;
+            }
+          }
+        }
+      }
+
+      .icon {
+        margin-right: 10px;
```
edlerd:
```suggestion
        margin-right: $sph--large;
```
Better use a variable here.

tests/projects.spec.ts
edlerd: This seems still open.

kimanhou: Mmh I removed one occurrence of this line, but where ?...

tests/projects.spec.ts
```diff
@@ -172,3 +178,109 @@ test("retain custom project selection on browsing pages for all projects", async
...
+
+  await deleteInstance(page, instance, project);
+  await deleteAllImages(page, project);
+
+  await page.getByRole("link", { name: "Configuration" }).click();
+  await page.waitForLoadState("networkidle");
+  await page.waitForSelector("text=Project configuration");
+
+  // Now the delete button should be enabled because the project is empty
+  await expect(deleteButton).toBeEnabled();
```
edlerd, line 273: I think the test is failing here because the `BulkDeleteImageBtn` is not clearing the project cache as it should be. That's why the "fresh project" was needed. I assume if cache invalidation is added to the bulk image deletion component, this will make the older test suites pass.

edlerd (dismissed):
Code LGTM, but tests need 1 more fix.

edlerd (approved):

PR #1601, opened 2025-11-11
Login with TLS improvements

kimanhou:
## Done

- On CertificateAdd page, add a Copy button at the end of the line (similar to Anbox).
- In LoggedInUserNotification, fix console error (p cannot be a descendant of p)

## QA

1. Run the LXD-UI:
    - Log out and log back in with TLS, skip to step 2: there should be a `Copy command` button and the command should be entirely visible, on multiple lines if needed.

## Screenshots

image


src/pages/login/CertificateAdd.tsx
```diff
@@ -72,10 +75,15 @@ const CertificateAdd: FC = () => {
               className="u-no-margin--bottom"
               blocks={[
                 {
-                  code: `if ! lxc auth group show admins ; then lxc auth group create admins ; lxc auth group permission add admins server admin ; fi ; lxc auth identity create tls/lxd-ui --group admins`,
+                  code: identityTrustTokenCommand,
+                  wrapLines: true,
```
edlerd:
```suggestion
```
I think not wrapping is better here.

src/pages/login/CertificateAdd.tsx
```diff
@@ -72,10 +75,15 @@ const CertificateAdd: FC = () => {
...
                 },
               ]}
             />
+            <CopyTextBtn
+              label="Copy command"
+              text={identityTrustTokenCommand}
+            />
```
edlerd: I think the button placement is a bit off. I like how GitHub does this, such as on our repos README:

image

Maybe we can right align this copy button with the command and show it on the same line as the command, like in the screen above? Some other app might have solved it. I suspect Anbox has such a component already. Worth reaching out to other teams. It might be worth upstreaming the pattern so we can reuse it here. Though that will turn this task into a bit of a bigger story.

src/pages/login/CertificateAdd.tsx
kimanhou: But then you can't see the whole command line.

src/pages/login/CertificateAdd.tsx
kimanhou: I reused the CopyTextBtn from CreateIdentityModal
image

But I agree, Github's design looks better. I asked @anusha-c18

src/pages/login/CertificateAdd.tsx
edlerd: Yes, that is true. This is meant a one-liner and I think we should show it as such. I wouldn't expect users to fully understand the code just from here, the "What is this doing" link below is meant to explain it in detail.

src/pages/login/CertificateAdd.tsx
edlerd: Right, I think in the modal the button placement is naturally in the footer. The issue I saw here is about placing the button in the page, and it seemed not well-balanced. Hence, the idea with in lining the button with the code line.

kimanhou: Thanks to @lorumic, it now looks super duper awesome !

image

src/components/CopyToClipboard.tsx
edlerd, line 1: Can we unify the `CopyTextBtn` and `CopyToClipboard` components? They seem to be solving the same problem.

I think we should use the same approach in the `CreateIdentityModal`: Hide the token, show only one line and truncate it. Then inline we can display the copy button and the modal footer does only need the close button.

kimanhou: Would the user understand they really need to copy this token somewhere ?

image

edlerd: I think this identity creation process is something we should revisit. Even before, I think this was unclear how to use the token just from the modal. I think this is a feature we should test and improve on. For the scope of this PR that is about improving the copy logic, I think it is fine.

edlerd: With test I mean quick user interaction test, I think the steps in this modal should give better guidance in what the token is and how it is to be used. Then it will also be clear to copy it when the modal is open.

kimanhou:
JIRA ticket created to improve identity creation flow: [WD-31254](https://warthogs.atlassian.net/browse/WD-31254)

[WD-31254]: https://warthogs.atlassian.net/browse/WD-31254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1604, opened 2025-11-12
feat(network) add network type help text to clarify network types

edlerd:
## Done

- feat(network) add network type help text to clarify network types

## QA

1. Run the LXD-UI:
    - start network creation and explore the network type selector

## Screenshots

image

### DIFF
--- a/src/pages/networks/forms/NetworkTypeSelector.tsx
+++ b/src/pages/networks/forms/NetworkTypeSelector.tsx
@@ -1,5 +1,5 @@
 import type { FC } from "react";
-import { Select } from "@canonical/react-components";
+import { CustomSelect } from "@canonical/react-components";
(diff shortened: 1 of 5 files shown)


edlerd:
> The clarifying text can be more accessible on smaller screens, maybe it would wrap.

Good catch! Adjusted the mobile representation of the selector to display all the text.

edlerd:
> The text looks good on smaller screen now. However, the select is much larger vertically than the options

Good catch! This is a side effect of the `white-space: break-spaces` to let the text wrap. I did find a middle ground by only applying it to the explanation under a specific width -- when the explanation won't fit into a single line with the type. That way it looks good in desktop and in smaller screens we ensure the full description is visible, but the height might be slightly too big.

omarelkashef (approved):
Looks great now!

PR #1605, opened 2025-11-13
refactor(DocLink): create 2 components to handle links to documentation

kimanhou:
## Done

- Create DocLink component
- Use DocLink in existing HelpLink component

## QA

1. Run the LXD-UI:
    - Make sure all documentation links still work

## Screenshots

No visual changes

### DIFF
--- a/src/components/DocLink.tsx
+++ b/src/components/DocLink.tsx
@@ -1,0 +1,38 @@
+import { Icon } from "@canonical/react-components";
+import { useDocs } from "context/useDocs";
+import type { FC, ReactNode } from "react";
+
+interface Props {
+  docPath: string;
+  children: ReactNode;
+  hasExternalIcon?: boolean;
+  className?: string;
+  title?: string;
+}
+
+const DocLink: FC<Props> = ({
+  docPath,
+  children,
+  hasExternalIcon = false,
+  className,
+  title,
+}) => {
+  const docBaseLink = useDocs();
+
+  return (
+    <a
+      className={className}
+      href={`${docBaseLink}${docPath}`}
+      target="_blank"
+      rel="noopener noreferrer"
+      title={title}
+    >
+      {children}
+      {hasExternalIcon && (
+        <Icon className="external-link-icon" name="external-link" />
+      )}
+    </a>
+  );
+};
+
+export default DocLink;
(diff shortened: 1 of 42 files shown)


src/components/DocLink.tsx
```diff
@@ -0,0 +1,27 @@
+import { useDocs } from "context/useDocs";
+import type { FC, ReactNode } from "react";
+
+interface Props {
+  docPath: string;
+  content: ReactNode;
+  className?: string;
+  title?: string;
+}
```
edlerd, line 11:
```suggestion
interface Props {
  docPath: string;
  children: ReactNode;
  hasIcon?: boolean;
  className?: string;
  title?: string;
}
```
I think we should use the more native `children` to pass the content of the link.
Also, I think we can add a `hasIcon` prop with a default of `false` to also cover the scope of `ExternalDocLink` and remove that 2nd component. Then instead of `ExternalDocLink` we can do `<DocLink withIcon docPath={...`

edlerd (commented):
Thanks for streamlining our doc links with new components. I think we can combine two of them in one, wdyt?

@roosterfish noted in another comment that there should be a validation of the linked documentation page. With this new structure it should be easy to write a test that a) collects all links to doc pages by usage of the `DocLink` component. And b) checks the latest/edge lxd version that all those links do actually exist in the documentation packaged within the snap. So we identify broken links through doc structure changes in the future. Though this should be a follow-up task in the future as it massively increases scope. Please can you file a ticket for this?

kimanhou:
JIRA ticket created to validate documentation links: [WD-31255](https://warthogs.atlassian.net/browse/WD-31255) 

[WD-31255]: https://warthogs.atlassian.net/browse/WD-31255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/components/DocLink.tsx
kimanhou: Yes ! I just renamed `hasIcon` into `hasExternalIcon` as `hasIcon` in `Button` for example makes you think you can add any icon.

Issue #1549, opened 2025-10-16
[Suggestion] Set static ip


sanskritkm:
Hello,

It would be better if we can setup static ip in UI:

<img width="677" height="211" alt="Image" src="https://github.com/user-attachments/assets/6ff37ae8-759c-4adf-8d71-53e7261df85a" />

PR #1610, opened 2025-11-14
feat(NIC) Support NIC static IP

omarelkashef:
## Done

- Support NIC static IP address on networks that allow it. #1549 

## QA

1. Run the LXD-UI:
    - Go Instances/Profiles > Configuration > Devices > Network
    - Attach a Network and enter IPv4 and/or IPv6 address
    - Ensure IP inputs are disabled if the network has ipv4 or ipv6 disabled or if the network does not allow static ip addresses.
    - Save Changes and ensure the new NIC has the correct IP addresses.
## Screenshots

image
image

### DIFF
--- a/src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
+++ b/src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
@@ -7,6 +7,7 @@ import NetworkSelector from "pages/projects/forms/NetworkSelector";
 import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
 import { supportsNicDeviceAcls } from "util/networks";
 import { isNoneDevice } from "util/devices";
+import { NetworkDeviceIPAddress } from "./NetworkDeviceIPAddress";
 
 interface Props {
   readOnly: boolean;
@@ -34,7 +35,6 @@ const NetworkDeviceContent: FC<Props> = ({
       </span>
     );
   }
-
   if (readOnly) {
     return (
       <>
@@ -50,6 +50,26 @@ const NetworkDeviceContent: FC<Props> = ({
           device={device}
           readOnly
         />
+
+        {formik.values.entityType === "instance" && (
+          <>
+            <NetworkDeviceIPAddress
+              formik={formik}
+              index={index}
+              network={network}
+              device={device}
+              family="IPv4"
+            />
+
+            <NetworkDeviceIPAddress
+              formik={formik}
+              index={index}
+              network={network}
+              device={device}
+              family="IPv6"
+            />
+          </>
+        )}
       </>
     );
   }
@@ -82,6 +102,24 @@ const NetworkDeviceContent: FC<Props> = ({
         formik={formik}
         canSelectManualAcls={supportsNicDeviceAcls(network)}
       />
+      {formik.values.entityType === "instance" && (
+        <>
+          <NetworkDeviceIPAddress
+            formik={formik}
+            index={index}
+            network={network}
+            device={device}
+            family="IPv4"
+          />
+          <NetworkDeviceIPAddress
+            formik={formik}
+            index={index}
+            network={network}
+            device={device}
+            family="IPv6"
+          />
+        </>
+      )}
     </>
   );
 };
--- a/src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
+++ b/src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
@@ -1,0 +1,77 @@
+import { type FC } from "react";
+import type { InstanceAndProfileFormikProps } from "../instanceAndProfileFormValues";
+import type { EditInstanceFormValues } from "pages/instances/EditInstance";
+import { Input } from "@canonical/react-components";
+import { type LxdNetwork } from "types/network";
+import { type LxdNicDevice } from "types/device";
+import { typesWithNicStaticIPSupport } from "util/networks";
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  index: number;
+  network?: LxdNetwork;
+  device: LxdNicDevice;
+  family: "IPv4" | "IPv6";
+}
+
+export const NetworkDeviceIPAddress: FC<Props> = ({
+  formik,
+  index,
+  network,
+  device,
+  family,
+}) => {
+  if (!network || !network.config) {
+    return null;
+  }
+
+  const readOnly = (formik.values as EditInstanceFormValues).readOnly;
+  const addressString = family === "IPv4" ? "ipv4.address" : "ipv6.address";
+  const networkIP = network.config[addressString];
+  const deviceIP = device[addressString];
+  const isStaticIPAllowed =
+    network.managed && typesWithNicStaticIPSupport.includes(network.type);
+  const canSelectIP = isStaticIPAllowed && networkIP && networkIP !== "none";
+
+  if (readOnly) {
+    return (
+      networkIP !== "none" && (
+        <div className="ip-content">
+          <>
+            <div>{family}</div>
+            <div className="mono-font">
+              <b>{deviceIP || "dynamic"}</b>
+            </div>
+          </>
+        </div>
+      )
+    );
+  }
+
+  return (
+    <div className="ip-content-edit">
+      <Input
+        id={`${addressString}-${index}`}
+        name={`${addressString}-${index}`}
+        label={`${family} address reservation`}
+        type="text"
+        placeholder={`Enter ${family} address`}
+        onChange={(e) => {
+          formik.setFieldValue(
+            `devices.${index}["${addressString}"]`,
+            e.target.value,
+          );
+        }}
+        value={deviceIP}
+        disabled={!canSelectIP}
+        help={
+          <>
+            {canSelectIP
+              ? `Choose an ${family} within the subnet range ${networkIP}, or leave it empty to be dynamically set.`
+              : `Static ${family} is not available for this network`}
+          </>
+        }
+      />
+    </div>
+  );
+};
--- a/src/util/formDevices.spec.ts
+++ b/src/util/formDevices.spec.ts
@@ -24,9 +24,13 @@ const deviceYaml =
   "    type: nic\n" +
   "  eth1:\n" +
   "    ipv4.address: 10.76.171.21\n" +
-  "    name: eth0\n" +
   "    network: mybr\n" +
   "    type: nic\n" +
+  "  eth2:\n" +
+  "    ipv4.address: 10.76.172.21\n" +
+  "    network: mybr1\n" +
+  "    type: nic\n" +
+  "    hostname: myhost\n" +
   "  grafananat:\n" +
   "    connect: tcp:10.76.171.21:3000\n" +
   "    listen: tcp:192.168.0.90:3000\n" +
@@ -53,7 +57,7 @@ describe("parseDevices and formDeviceToPayload", () => {
       Object.values(formDevices).filter((item) => item.type === deviceType);
 
     expect(matchFormDeviceType("disk").length).toBe(1);
-    expect(matchFormDeviceType("nic").length).toBe(1);
+    expect(matchFormDeviceType("nic").length).toBe(2);
     expect(matchFormDeviceType("custom-nic").length).toBe(1);
     expect(matchFormDeviceType("proxy").length).toBe(2);
     expect(matchFormDeviceType("gpu").length).toBe(1);
--- a/src/util/networks.tsx
+++ b/src/util/networks.tsx
@@ -24,6 +24,7 @@ export const typesWithLeases = [bridgeType, ovnType];
 export const typesWithParent = [physicalType, sriovType, macvlanType];
 export const typesWithStatistics = [bridgeType, ovnType, physicalType];
 export const typesWithNicDeviceAcls = [ovnType];
+export const typesWithNicStaticIPSupport = [bridgeType, ovnType];
 export const typesWithLocalPeerings = [ovnType];
 
 export const getIpAddresses = (
--- a/src/sass/_forms.scss
+++ b/src/sass/_forms.scss
@@ -92,6 +92,18 @@
       margin-top: $spv--small;
     }
 
+    .ip-content {
+      margin-top: $spv--small;
+    }
+
+    .ip-content-edit {
+      margin-top: $sph--x-large;
+
+      .p-form__label {
+        margin-bottom: $spv--small;
+      }
+    }
+
     .image-change-link {
       align-self: end;
       margin-bottom: 6px;
--- a/src/types/device.d.ts
+++ b/src/types/device.d.ts
@@ -24,6 +24,8 @@ export interface LxdNicDevice {
   parent?: string;
   network: string;
   type: "nic";
+  "ipv4.address"?: string;
+  "ipv6.address"?: string;
   "security.acls"?: string;
 }
 
--- a/src/util/formDevices.tsx
+++ b/src/util/formDevices.tsx
@@ -113,7 +113,15 @@ export const parseDevices = (devices: LxdDevices): FormDevice[] => {
     const isCustomNetwork =
       item.type === "nic" &&
       Object.keys(item).some(
-        (key) => !["type", "name", "network", "security.acls"].includes(key),
+        (key) =>
+          ![
+            "type",
+            "name",
+            "network",
+            "security.acls",
+            "ipv4.address",
+            "ipv6.address",
+          ].includes(key),
       );
 
     if (isCustomNetwork) {
@@ -136,6 +144,8 @@ export const parseDevices = (devices: LxdDevices): FormDevice[] => {
       case "nic":
         return {
           name: key,
+          "ipv4.address": item["ipv4.address"],
+          "ipv6.address": item["ipv6.address"],
           network: item.network,
           type: "nic",
           "security.acls": item["security.acls"],


src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,145 @@
...
+            </p>
+          </>
+        )}
+      </div>
+    );
+  }
+
+  return (
+    <>
+      <label
```
edlerd: Better use the label property of the Input component. That should be simpler and the spacing should also be better

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,145 @@
...
+}
+
+export const NetworkDeviceIPAddress: FC<Props> = ({
+  formik,
+  index,
+  network,
+  device,
+  readOnly,
+}) => {
+  const [ipv4Address, setipv4Address] = useState(device["ipv4.address"] || "");
```
edlerd: Better rely on formik for state handling. Can we remove the state completely from this component so it is purely representation and stateless?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,145 @@
...
+            `devices.${index}["ipv4.address"]`,
+            e.target.value,
+          );
+        }}
+        value={ipv4Address}
+        disabled={!canSelectIPv4()}
+        help={
+          <>
+            {canSelectIPv4()
+              ? `Use CIDR notation in the range ${getNetworkIPv4()}, or leave it empty to use a dynamic IPv4 address.`
```
edlerd: Nice 👍

src/util/networks.tsx
```diff
@@ -23,6 +23,7 @@ export const typesWithForwards = [bridgeType, ovnType];
 export const typesWithParent = [physicalType, sriovType, macvlanType];
 export const typesWithStatistics = [bridgeType, ovnType, physicalType];
 export const typesWithNicDeviceAcls = [ovnType];
+export const typesWithStaticIP = [bridgeType, ovnType];
```
edlerd: Maybe refine the name a bit more to make it clearer this is not about the network config, but about using the network as a nic device.

src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
```diff
@@ -82,6 +89,13 @@ const NetworkDeviceContent: FC<Props> = ({
         formik={formik}
         canSelectManualAcls={supportsNicDeviceAcls(network)}
       />
+      <NetworkDeviceIPAddress
+        formik={formik}
```
edlerd: I suspect this will leak it nic device configuration for profiles as well. We should add a condition to only have it show for instances.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  index: number;
+  network?: LxdNetwork;
+  device: LxdNicDevice;
+  readOnly: boolean;
+}
+
+export const NetworkDeviceIPAddress: FC<Props> = ({
```
edlerd, line 17:
```suggestion
export const NetworkDeviceIPAddresses: FC<Props> = ({
```
Should this be plural?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,145 @@
...
+      config["ipv6.address"] &&
+      config["ipv6.address"] != "none"
+    );
+  };
+
+  if (readOnly) {
+    return (
+      <div className="ip-content">
+        {device["ipv4.address"] && (
+          <>
```
edlerd: Shall we remove the condition here and render "dynamic" or "none" in case it is not set?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
edlerd: Second thought: this component could receive a prop for ipv4 / ipv6 and we call it twice with each mode. That should reduce most of the internal duplication as I assume both modes are mostly the same.

edlerd:
> I'm trying to add a network but I get this error on save: Cannot specify "ipv6.address" when DHCP or "ipv6.dhcp.stateful" are disabled (unless using security.ipv6_filtering) on network "lxdbr0"

Good catch!

@omarelkashef and I discussed this error last week. For the first iteration (this PR) we decided to build the static IP address form for instances. There will be a follow-up to avoid users running into the error you also encountered. Plan is to avoid users running into the error and give users direction how to unblock by changing network config. @omarelkashef can you please create a follow up task or have you already created one so we keep track of the plan?

src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
```diff
@@ -82,6 +104,22 @@ const NetworkDeviceContent: FC<Props> = ({
...
+        family="IPv4"
+      />
+      <NetworkDeviceIPAddress
+        formik={formik}
+        index={index}
+        network={network}
+        readOnly={readOnly}
+        device={device}
+        family="IPv6"
+      />
```
edlerd: I think this needs the `formik.values.entityType == "instance"` guard as well as the readmode above to avoid leaking this into profile configuration edit.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,93 @@
...
+  readOnly,
+  family,
+}) => {
+  if (!network || !network.config) {
+    return null;
+  }
+
+  const getIPAddressString = () => {
+    return family === "IPv4" ? "ipv4.address" : "ipv6.address";
+  };
```
edlerd:
```suggestion
  const addressString = family === "IPv4" ? "ipv4.address" : "ipv6.address";
```
Consider using directly evaluated `const`, it seems shorter and simpler than defining callbacks.

edlerd (commented):
Code looks much better, thanks for the iteration! Two small further ideas below.

src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
omarelkashef: So, setting static IP on nics in profiles is not allowed?

edlerd: Technically, it is allowed.

My train of thought was this: Profiles are meant as a configuration template applied to multiple instances. With a static IP on a NIC device, you can effectively only apply that profile to a single instance. So probably we should hide the static IP inputs on profiles.

src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
```diff
@@ -50,6 +50,28 @@ const NetworkDeviceContent: FC<Props> = ({
           device={device}
           readOnly
         />
+
+        {formik.values.entityType == "instance" && (
```
edlerd:
```suggestion
        {formik.values.entityType === "instance" && (
```
Missed this before, please always use the type safe comparisons in JS.

src/components/forms/NetworkDevicesForm/NetworkDeviceContent.tsx
```diff
@@ -82,6 +104,26 @@ const NetworkDeviceContent: FC<Props> = ({
         formik={formik}
         canSelectManualAcls={supportsNicDeviceAcls(network)}
       />
+      {formik.values.entityType == "instance" && (
```
edlerd:
```suggestion
      {formik.values.entityType === "instance" && (
```
Here as well.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,89 @@
...
+}) => {
+  if (!network || !network.config) {
+    return null;
+  }
+
+  const addressString = family === "IPv4" ? "ipv4.address" : "ipv6.address";
+
+  const getNetworkIP = () => {
+    return network.config[addressString];
+  };
```
edlerd: Can this and the callbacks below also be directly evaluated `const` strings to simplify?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,76 @@
...
+  if (!network || !network.config) {
+    return null;
+  }
+
+  const addressString = family === "IPv4" ? "ipv4.address" : "ipv6.address";
+  const networkIP = network.config[addressString];
+  const deviceIP = device[addressString];
+  const isStaticIPAllowed =
+    network.managed && typesWithNicStaticIPSupport.includes(network.type);
+  const canSelectIP = isStaticIPAllowed && networkIP && deviceIP != "none";
```
edlerd:
```suggestion
  const canSelectIP = isStaticIPAllowed && networkIP && networkIP != "none";
```
Shouldn't this look at the network ip?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,76 @@
...
+        aria-label={`${family} address`}
+        type="text"
+        placeholder={`Enter ${family} address`}
+        onChange={(e) => {
+          formik.setFieldValue(
+            `devices.${index}["${addressString}"]`,
+            e.target.value,
+          );
+        }}
+        value={(formik.values.devices[index] as LxdNicDevice)[addressString]}
```
edlerd:
```suggestion
        value={deviceIP}
```
A bit simpler, I suspect it is the same.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,76 @@
...
+      </div>
+    );
+  }
+
+  return (
+    <>
+      <Input
+        id={`${addressString}-${index}`}
+        name={`${addressString}-${index}`}
+        aria-label={`${family} address`}
```
edlerd:
```suggestion
```
We don't need the aria label, but should set a label instead.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,76 @@
...
+        placeholder={`Enter ${family} address`}
+        onChange={(e) => {
+          formik.setFieldValue(
+            `devices.${index}["${addressString}"]`,
+            e.target.value,
+          );
+        }}
+        value={(formik.values.devices[index] as LxdNicDevice)[addressString]}
+        disabled={!canSelectIP}
+        label={<label className={canSelectIP ? "" : "u-text--muted"}></label>}
```
edlerd:
```suggestion
        label={`${family} address`}
```
Let's use the label here.

edlerd: We don't need the custom muted class here, the disabled prop above takes care of muting the label for us.

kimanhou:
Following up on your demo, the placeholders should be examples of valid input, here "valid" (as valid as we can guess at this point) IP addresses. Similar to when you fill up a form with first name and last name, the placeholder for first name is usually "John" and the placeholder for last name is "Smith".

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
edlerd: We might want to add an info icon next to the label with a tooltip attached. In the tooltip, we can clarify this is a "DHCP reservation" and not an actual assignment. Layout can be similar to what we do for the SSH key with info icon and tooltip:

Screenshot from 2025-11-20 17-34-41

edlerd: Or second thought maybe label this as "IPv4 address reservation"?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
omarelkashef: Oh yes, I must have changed it by mistake will addressing feedback.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { type LxdNetwork } from "types/network";
+import { type LxdNicDevice } from "types/device";
+import { typesWithNicStaticIPSupport } from "util/networks";
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  index: number;
+  network?: LxdNetwork;
+  device: LxdNicDevice;
+  readOnly: boolean;
```
edlerd:
```suggestion
```
I'd suggest keeping the interface smaller, and inside this we can read it as `const readOnly = formik.values.readonly`.

src/util/formDevices.spec.ts
```diff
@@ -24,7 +24,6 @@ const deviceYaml =
   "    type: nic\n" +
   "  eth1:\n" +
   "    ipv4.address: 10.76.171.21\n" +
-  "    name: eth0\n" +
```
edlerd, line 27: Instead of altering this one, maybe add a new device entry that contains an ipv4 and ipv6 address. With this removed, we don't test the custom nic branch of the function any more.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const canSelectIP = isStaticIPAllowed && networkIP && networkIP != "none";
+
+  if (readOnly) {
+    return (
+      <div className="ip-content">
+        <>
+          <div>{family}</div>
+          <p className="u-no-margin--top u-no-margin--bottom">
+            {deviceIP || "Dynamic"}
+          </p>
```
edlerd:
```suggestion
          <div class="mono-font">
            <b>{deviceIP || "Dynamic"}</b>
          </div>
```
If we don't need the spacing of a paragraph, better use a `div` here. Seems also to be the better choice semantically. 

Also we should use the mono font in bold to clarify this is the value.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+    );
+  }
+
+  return (
+    <>
+      <Input
+        id={`${addressString}-${index}`}
+        name={`${addressString}-${index}`}
+        label={`${family} address reservation`}
+        type="text"
```
edlerd, line 57:
```suggestion
        type="text"
        class="ip-content-edit"
```
This needs some spacing below and above. I'd suggest a new class `ip-content-edit` with spacing above and below of `$sph--x-large`.

edlerd:
Also, the labels for the address inputs are a bit too close to the inputs, I suspect this is a global override in the configuration tables that we should remove:

image

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+  if (!network || !network.config) {
+    return null;
+  }
+
+  const addressString = family === "IPv4" ? "ipv4.address" : "ipv6.address";
+  const networkIP = network.config[addressString];
+  const deviceIP = device[addressString];
+  const isStaticIPAllowed =
+    network.managed && typesWithNicStaticIPSupport.includes(network.type);
+  const canSelectIP = isStaticIPAllowed && networkIP && networkIP != "none";
```
kimanhou:
```suggestion
  const canSelectIP = isStaticIPAllowed && networkIP && networkIP !== "none";
```
Please avoid single equals

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return (
+    <>
+      <Input
+        id={`${addressString}-${index}`}
+        name={`${addressString}-${index}`}
+        label={`${family} address reservation`}
+        type="text"
+        placeholder={`Enter ${family} address`}
```
kimanhou: I would really prefer the placeholder to be an IP address.
```suggestion
        const placeholder = networkIP?.split("/")[0] || `Enter ${family} address`;
        placeholder={placeholder}
```

image

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+            `devices.${index}["${addressString}"]`,
+            e.target.value,
+          );
+        }}
+        value={deviceIP}
+        disabled={!canSelectIP}
+        help={
+          <>
+            {canSelectIP
+              ? `Use an ${family} in the range ${networkIP}, or leave it empty to be dynamically set.`
```
kimanhou:
```suggestion
              ? `Choose an ${family} within the subnet range ${networkIP}, or leave it empty to be dynamically set.`
```

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+      </div>
+    );
+  }
+
+  return (
+    <>
+      <Input
+        id={`${addressString}-${index}`}
+        name={`${addressString}-${index}`}
+        label={`${family} address reservation`}
```
kimanhou, line 56:
```suggestion
        label={`${family} static address reservation`}
```
Clearer but then very long. Wdyt ?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
kimanhou, line 58:
```suggestion
        placeholder={`${networkIp.split('/')[0].slice(0, -1)}[1-254]`}
```

FYI MAAS displays an IP address as a placeholder and pre-fills the beginning of the input value. We cannot do the prefill as the input is optional but I like the clarity it brings 
image

It seems like a small and worthwhile addition. What do you guys think ?

edlerd: Does MAAS have an ip address component? We could use it here and also for the network forwards.

This seems much simpler to use.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
omarelkashef: Yes, I agree the label is getting kinda long.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
omarelkashef: After discussion with @kimanhou , the suggested code would only work with a /24 mask, so we will leave the placeholder as is and provide a more helpful one that works with all networks in a future PR.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,76 @@
...
+        type="text"
+        className="ip-content-edit"
```
edlerd: This is applying the class to the input, we'd need to have it on a wrapper around the label, input and help text to add the spacing above the label and below help text.

Current:

image

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
kimanhou: @edlerd Here is MAAS code, it's quite complex but very complete.

https://github.com/canonical/maas-ui/blob/main/src/app/base/components/PrefixedIpInput/PrefixedIpInput.tsx
https://github.com/canonical/maas-ui/blob/main/src/app/utils/subnetIpRange.ts

edlerd: Can you create a follow-up task to see if we can upstream or adopt this component for NIC device IP and network forwards IP inputs?

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
```diff
@@ -0,0 +1,75 @@
...
+    network.managed && typesWithNicStaticIPSupport.includes(network.type);
+  const canSelectIP = isStaticIPAllowed && networkIP && networkIP !== "none";
+
+  if (readOnly) {
+    return (
+      <div className="ip-content">
+        <>
+          <div>{family}</div>
+          <div className="mono-font">
+            <b>{deviceIP || "Dynamic"}</b>
```
edlerd: This should also rely on `networkIP !== "none"`, currently it falls back to "dynamic" even when the network is set to none for the family at hand.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
edlerd: I also think better to keep as is, than with the additional static keyword. It seems more verbose and not adding information.

src/components/forms/NetworkDevicesForm/NetworkDeviceIPAddress.tsx
edlerd: Adding to the above: I think "dynamic" in lowercase would look better on this one.

edlerd (approved):
Code and QA LGTM, thanks for the good work.

Please rebase on main to remove the merge commit and then force push before merge to main.

PR #1618, opened 2025-11-18
fix(project creation): handle emoji in project name

kimanhou:
## Done

- Decode project name in URL

## QA

1. Run the LXD-UI:
    - Create a project with an emoji in the name, for example 🐧-Penguin-Pen. Make sure that creation works and you can access the project
    - Create a project with no emoji in the name, for example boring-name. Make sure everything still works fine.
    - Access a URL with an emoji in the project name but the project doesn't exist, for example 🐧-Penguin-Peeeeen. Make sure the error message is displayed correctly.

## Screenshots

project-list-and-url
project-not-found


kimanhou:
I fixed the project deletion. However instances can't be started because back end doesn't support it. So is it worth it ?

image

kimanhou:
I think it's better to keep rules consistent. I can live with naming my project Turtle-Tunnel instead of 🐢-Turtle-Tunnel

PR #1623, opened 2025-11-20
Refactor usedBy and DeleteGroupModal

kimanhou:
## Done

- Refactor `DeleteGroupModal` to be similar to `DeleteProjectModal`
- Add keys in `StoragePoolSize` to fix console error
- Create `UsedByRow` component and re-use in `ProjectUsedBy`, `NetworkAclUsedBy` and `StorageUsedBy`. 

Slight changes in UI of Used By: 
- if there is no entities, the whole row is not displayed
- in `StorageUsedBy`, if an image is not part of the current project, the alias is displayed (instead of the fingerprint)

## QA

1. Run the LXD-UI:
    - Go to Permissions > Group. Delete a group byt typing confirm-delete-group, make sure the delete button is correctly displayed, is enabled and you can delete the group.
    - Delete non empty project: make sure the list of resources look the same as before and all links work
    - Go to Networking > ACLs. Select an ACL that is used: make sure the list looks the same as before (without rows with count = 0, see screenshot) and all links work
    - Go to Storage > Pools. Select a pool that is used: make sure the list looks the same as before (without rows with count = 0, see screenshot) and all links work

## Screenshots

**ProjectUsedBy**
image

**NetworkAclUsedBy**
BEFORE
image

AFTER
image

**StorageUsedBy**
BEFORE
image

AFTER
image

### DIFF
--- a/src/components/ErrorPage.tsx
+++ b/src/components/ErrorPage.tsx
@@ -57,7 +57,7 @@ If the problem continues, ensure your connection to the LXD server is active or 
       <Notification severity="negative" title="Error">
         Something has gone wrong. If this issue persists,{" "}
         <a href={url} rel="noopener noreferrer" target="_blank">
-          please raise an issue on Github.
+          please raise an issue on GitHub.
         </a>
       </Notification>
       <CodeSnippet
--- a/src/components/UsedByItem.tsx
+++ b/src/components/UsedByItem.tsx
@@ -23,7 +23,7 @@ const UsedByItem: FC<Props> = ({
 }) => {
   const isImageQueryEnabled = type === "image";
   const { data: images = [] } = useImagesInProject(
-    activeProject,
+    item.project || activeProject,
     isImageQueryEnabled,
   );
 
--- a/src/components/UsedByRow.tsx
+++ b/src/components/UsedByRow.tsx
@@ -1,0 +1,55 @@
+import { type FC } from "react";
+import UsedByItem from "./UsedByItem";
+import ExpandableList from "./ExpandableList";
+import { filterUsedByType, getLinkTarget } from "util/usedBy";
+import { pluralize } from "util/instanceBulkActions";
+import type { ResourceIconType } from "components/ResourceIcon";
+import { capitalizeFirstLetter } from "util/helpers";
+import ResourceIcon from "components/ResourceIcon";
+import { useCurrentProject } from "context/useCurrentProject";
+import type { ResourceType } from "util/resourceDetails";
+
+interface Props {
+  entityType: ResourceType;
+  usedBy: string[] | undefined;
+}
+
+const UsedByRow: FC<Props> = ({ entityType, usedBy }) => {
+  if (!usedBy) return null;
+
+  const { project } = useCurrentProject();
+
+  const filteredResources = filterUsedByType(entityType, usedBy);
+
+  const entityNameSingular =
+    entityType === "volume"
+      ? "Custom volume"
+      : capitalizeFirstLetter(entityType.replaceAll("-", " "));
+  const entityName = pluralize(entityNameSingular, filteredResources.length);
+
+  return (
+    <tr className="used-by-row">
+      <th className="u-text--muted used-by-row-header">
+        <ResourceIcon type={entityType as ResourceIconType} className="icon" />
+        {entityName} ({filteredResources.length})
+      </th>
+      <td>
+        <ExpandableList
+          items={filteredResources.map((item) => {
+            return (
+              <UsedByItem
+                key={`${item.name}-${item.project}-${item.target ?? ""}`}
+                item={item}
+                activeProject={project?.name || "default"}
+                type={entityType as ResourceIconType}
+                to={getLinkTarget(item, entityType)}
+              />
+            );
+          })}
+        />
+      </td>
+    </tr>
+  );
+};
+
+export default UsedByRow;
--- a/src/pages/networks/NetworkAclUsedBy.tsx
+++ b/src/pages/networks/NetworkAclUsedBy.tsx
@@ -1,18 +1,11 @@
 import type { FC } from "react";
-import type { LxdUsedBy } from "util/usedBy";
-import { filterUsedByType } from "util/usedBy";
-import ExpandableList from "components/ExpandableList";
-import UsedByItem from "components/UsedByItem";
 import type { LxdNetworkAcl } from "types/network";
 import { useParams } from "react-router-dom";
+import UsedByRow from "components/UsedByRow";
 
 interface Props {
   networkAcl: LxdNetworkAcl;
 }
-
-const INSTANCES = "Instances";
-const PROFILES = "Profiles";
-const NETWORKS = "Networks";
 
 const NetworkAclUsedBy: FC<Props> = ({ networkAcl }) => {
   const { project } = useParams<{ project: string }>();
@@ -21,65 +14,12 @@ const NetworkAclUsedBy: FC<Props> = ({ networkAcl }) => {
     return null;
   }
 
-  const data: Record<string, LxdUsedBy[]> = {
-    [INSTANCES]: filterUsedByType("instance", networkAcl.used_by),
-    [PROFILES]: filterUsedByType("profile", networkAcl.used_by),
-    [NETWORKS]: filterUsedByType("network", networkAcl.used_by),
-  };
-
   return (
-    <table>
+    <table className="network-acl-used-by-table">
       <tbody>
-        <tr>
-          <th className="u-text--muted">
-            Instances ({data[INSTANCES].length})
-          </th>
-          <td>
-            <ExpandableList
-              items={data[INSTANCES].map((item) => (
-                <UsedByItem
-                  key={`${item.name}-${item.project}`}
-                  item={item}
-                  activeProject={project}
-                  type="instance"
-                  to={`/ui/project/${encodeURIComponent(item.project)}/instance/${encodeURIComponent(item.name)}`}
-                />
-              ))}
-            />
-          </td>
-        </tr>
-        <tr>
-          <th className="u-text--muted">Profiles ({data[PROFILES].length})</th>
-          <td>
-            <ExpandableList
-              items={data[PROFILES].map((item) => (
-                <UsedByItem
-                  key={`${item.name}-${item.project}`}
-                  item={item}
-                  activeProject={project}
-                  type="profile"
-                  to={`/ui/project/${encodeURIComponent(item.project)}/profile/${encodeURIComponent(item.name)}`}
-                />
-              ))}
-            />
-          </td>
-        </tr>
-        <tr>
-          <th className="u-text--muted">Networks ({data[NETWORKS].length})</th>
-          <td>
-            <ExpandableList
-              items={data[NETWORKS].map((item) => (
-                <UsedByItem
-                  key={`${item.name}-${item.project}`}
-                  item={item}
-                  activeProject={project}
-                  type="network"
-                  to={`/ui/project/${encodeURIComponent(item.project)}/network/${encodeURIComponent(item.name)}`}
-                />
-              ))}
-            />
-          </td>
-        </tr>
+        <UsedByRow entityType="instance" usedBy={networkAcl.used_by} />
+        <UsedByRow entityType="profile" usedBy={networkAcl.used_by} />
+        <UsedByRow entityType="network" usedBy={networkAcl.used_by} />
       </tbody>
     </table>
   );
--- a/src/pages/projects/actions/ProjectUsedBy.tsx
+++ b/src/pages/projects/actions/ProjectUsedBy.tsx
@@ -1,135 +1,24 @@
 import { type FC } from "react";
 import type { LxdProject } from "types/project";
-import type { LxdUsedBy } from "util/usedBy";
-import { filterUsedByType } from "util/usedBy";
-import { Icon } from "@canonical/react-components";
-import { useImagesInProject } from "context/useImages";
-import ExpandableList from "components/ExpandableList";
-import UsedByItem from "components/UsedByItem";
-
-const INSTANCES = "Instances";
-const PROFILES = "Profiles";
-const IMAGES = "Images";
-const CUSTOM_VOLUMES = "Custom volumes";
+import UsedByRow from "components/UsedByRow";
 
 interface Props {
   project: LxdProject;
 }
 
 const ProjectUsedBy: FC<Props> = ({ project }) => {
-  const { data: imagesFromProject = [] } = useImagesInProject(project.name);
-  const data: Record<string, LxdUsedBy[]> = {
-    [INSTANCES]: filterUsedByType("instance", project.used_by),
-    [PROFILES]: filterUsedByType(
-      "profile",
-      project.used_by?.filter(
-        // the default profile is not blocking project deletion and can't be removed itself
-        (item) => !item.startsWith("/1.0/profiles/default"),
-      ),
-    ),
-    [IMAGES]: filterUsedByType("image", project.used_by),
-    [CUSTOM_VOLUMES]: filterUsedByType("volume", project.used_by),
-  };
+  // the default profile is not blocking project deletion and can't be removed itself
+  const usedBy = project.used_by?.filter(
+    (t) => !t.startsWith("/1.0/profiles/default"),
+  );
 
   return (
     <table className="p-main-table delete-project-table">
       <tbody>
-        {data[INSTANCES].length > 0 && (
-          <tr>
-            <th>
-              <Icon name="pods" className="icon" />
-              Instances ({data[INSTANCES].length})
-            </th>
-            <td>
-              <ExpandableList
-                items={data[INSTANCES].map((item) => (
-                  <UsedByItem
-                    key={`${item.name}-${item.project}`}
-                    item={item}
-                    activeProject={project.name}
-                    type="instance"
-                    to={`/ui/project/${encodeURIComponent(item.project)}/instance/${encodeURIComponent(item.name)}`}
-                  />
-                ))}
-              />
-            </td>
-          </tr>
-        )}
-
-        {data[PROFILES].length > 0 && (
-          <tr>
-            <th>
-              <Icon name="repository" className="icon" />
-              Profiles ({data[PROFILES].length})
-            </th>
-            <td>
-              <ExpandableList
-                items={data[PROFILES].map((item) => (
-                  <UsedByItem
-                    key={`${item.name}-${item.project}`}
-                    item={item}
-                    activeProject={project.name}
-                    type="profile"
-                    to={`/ui/project/${encodeURIComponent(item.project)}/profile/${encodeURIComponent(item.name)}`}
-                  />
-                ))}
-              />
-            </td>
-          </tr>
-        )}
-
-        {data[IMAGES].length > 0 && (
-          <tr>
-            <th>
-              <Icon name="image" className="icon" />
-              Images ({data[IMAGES].length})
-            </th>
-            <td>
-              <ExpandableList
-                items={data[IMAGES].map((image) => {
-                  const imageFromProject = imagesFromProject.find(
-                    (img) => img.fingerprint === image.name,
-                  );
-                  return {
-                    ...image,
-                    name:
-                      imageFromProject?.properties?.description || image.name,
-                  };
-                }).map((item) => (
-                  <UsedByItem
-                    key={`${item.name}-${item.project}`}
-                    item={item}
-                    activeProject={project.name}
-                    type="image"
-                    to={`/ui/project/${encodeURIComponent(item.project)}/images`}
-                  />
-                ))}
-              />
-            </td>
-          </tr>
-        )}
-
-        {data[CUSTOM_VOLUMES].length > 0 && (
-          <tr>
-            <th>
-              <Icon name="switcher-dashboard" className="icon" />
-              Volumes ({data[CUSTOM_VOLUMES].length})
-            </th>
-            <td>
-              <ExpandableList
-                items={data[CUSTOM_VOLUMES].map((item) => (
-                  <UsedByItem
-                    key={`${item.name}-${item.project}-${item.target}`}
-                    item={item}
-                    activeProject={project.name}
-                    type="volume"
-                    to={`/ui/project/${encodeURIComponent(item.project)}/storage/volumes`}
-                  />
-                ))}
-              />
-            </td>
-          </tr>
-        )}
+        <UsedByRow entityType="instance" usedBy={usedBy} />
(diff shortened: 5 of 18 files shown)


src/pages/networks/NetworkAclUsedBy.tsx
```diff
@@ -30,56 +29,24 @@ const NetworkAclUsedBy: FC<Props> = ({ networkAcl }) => {
...
-                  type="network"
-                  to={`/ui/project/${encodeURIComponent(item.project)}/network/${encodeURIComponent(item.name)}`}
-                />
-              ))}
-            />
-          </td>
-        </tr>
+        <UsedByRow
+          entityType="instance"
+          usedBy={data[INSTANCES]}
```
edlerd:
```suggestion
          entityType={INSTANCES}
          usedBy={networkAcl.used_by}
```
Shall we push the `filterUsedByType` call into the `UsedByRow` component as well?

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { linkForVolumeDetail } from "util/storageVolume";
+import type { LxdStorageVolume } from "types/storage";
+import { getStorageBucketURL } from "util/storageBucket";
+
+interface Props {
+  entityType: string;
+  usedBy: LxdUsedBy[];
+  projectName: string;
+  iconName?: string;
+  isMuted?: boolean;
```
edlerd:
```suggestion
```
Shall we make this always muted for consistency? Not sure if it looks nice in all contexts though.

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { capitalizeFirstLetter } from "util/helpers";
+import { linkForVolumeDetail } from "util/storageVolume";
+import type { LxdStorageVolume } from "types/storage";
+import { getStorageBucketURL } from "util/storageBucket";
+
+interface Props {
+  entityType: string;
+  usedBy: LxdUsedBy[];
+  projectName: string;
+  iconName?: string;
```
edlerd:
```suggestion
```
Shall we always show the icon in the header? Then we could use the `ResourceIcon` component with the `entityType` in here and can make the interface here simpler.

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,95 @@
...
+  isMuted?: boolean;
+  storageName?: string;
+  entityName?: string;
```
edlerd:
```suggestion
```
I think this is only used for volume to display it as "Custom volumes". Maybe the special case can live in this component?

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,95 @@
...
+import type { ResourceIconType } from "components/ResourceIcon";
+import { capitalizeFirstLetter } from "util/helpers";
+import { linkForVolumeDetail } from "util/storageVolume";
+import type { LxdStorageVolume } from "types/storage";
+import { getStorageBucketURL } from "util/storageBucket";
+
+interface Props {
+  entityType: string;
+  usedBy: LxdUsedBy[];
+  projectName: string;
```
edlerd:
```suggestion
```
Shall we read the current project from the url instead of passing this in?

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,95 @@
...
+  iconName?: string;
+  isMuted?: boolean;
+  storageName?: string;
```
edlerd:
```suggestion
```
I think this is available on the `UsedByItem`s as the `.pool` property. Can we use that instead and remove this prop?

src/components/UsedByRow.tsx
kimanhou: There are 2 places using entityType="volume":
- in `StorageUsedBy`, the displayed label is "Custom volumes"
- in `DeleteProjectModal`, the displayed label is "Volumes"

We can standardize those 2 occurrences: either display "Custom volumes" for both or "Volumes" for both ?
If we want to keep the different names, then I didn't find a better solution to override it in one of the cases.

Wdyt ?

edlerd: I think "Custom volumes" is best for both.

src/components/UsedByRow.tsx
kimanhou: I thought about it, because for now only `ProjectDeleteModal` is not muted.

Current - not muted
image

Muted
image

I think it looks good muted as well and it makes the UI consistent and the code simpler. Wdyt ?

src/components/UsedByRow.tsx
kimanhou: I thought about it, I like the icons but added them only in `ProjectDeleteModal`, so it would change the UI of `NetworkAclUsedBy` and `StorageUsedBy`. If you're happy with this change, then yes we can always show the icon in the header.

src/components/UsedByRow.tsx
edlerd: Yeah, I think that is okay to use muted there as well.

src/components/UsedByRow.tsx
edlerd: I suspect both places would be okay with an icon, but not sure without having it seen in context.

kimanhou: `ProjectDeleteModal` - with icon and muted text
image

`NetworkAclUsedBy`
image

`StorageUsedBy`
image

edlerd: I think the icons on the heading work really well.

kimanhou: By the way, why does `able-fawn` appears twice in `networkAcl.used_by` in `NetworkAclUsedBy` ? This behavior dates from before my changes.

image

edlerd: That looks like a bug. Do you see it appear twice in the API response for the network ACL? If yes, please report it as an issue in the LXD repository.

The used by API seems to not be heavily used by others than us, as we already found and fixed several issues with it in the past.

src/pages/networks/NetworkAclUsedBy.tsx
kimanhou: We could, but then how would we filter the default profile in `DeleteProjectModal` ?

edlerd: We can filter it from the list before passing it on to the `UsedByItem`. Something like `const usedByUrls = project.used_by.filter((url) ...)` and then `<UsedByItem usedBy={usedByUrls} ...`

src/components/UsedByRow.tsx
edlerd: Yes, this seems like a bug in the API, it repeats the instance on an ACL used by list if the instance has multiple ACLs applied.

image

kimanhou: I created a [Github issue](https://github.com/canonical/lxd/issues/17011) on LXD repository.

Is "`status-queued-small`" the right icon for pool, volume and bucket ? This is currently the case in `ResourceIcon` but this icon seems to be used as a status in `ClusterMemberStatus` and `InstanceStatusIcon`.

edlerd: It is for now. We should request a proper storage icon from the design team to use here. I think a cylinder shape would work well for it. We use the queued small icon (a black dot) as we don't have a better suiting icon here to use at the moment.

kimanhou: Should we use "`switcher-dashboard`" ? The use of `ResourceIcon` replaced the icon for Volumes in `DeleteProjectModal`.
image

edlerd: I would avoid switching to an intermediary icon and then going for the best choice. Would suggest keeping as is and requesting the right ones from design and updating it later to the right one.

edlerd: Created [this ticket](https://warthogs.atlassian.net/browse/WD-31461) to track the new icons.

kimanhou: So you're okay with the icon of Custom volume to be a dot in `ProjectDeleteModal` in the meantime ?

edlerd: Yes, please.

src/components/UsedByRow.tsx
kimanhou: Looks like the .pool property is undefined here, am I missing something ? 🤔

image

I think it's because `extractResourceDetailsFromUrl` inside `filterUsedByType` expects `storage-volume` instead of `volume`. But then I'm scared of the unexpected consequences if I change `filterUsedByType`.

I can also update `resourceDetails`.

I have 2 very different URLs for volumes used by a project in ProjectDeleteModal:
https://localhost:8407/ui/project/Duck-Dock/storage/pool/local/member/latest-edge-cluster/volumes/custom/poney
https://localhost:8407/ui/project/Duck-Dock/instance/concrete-mule

Before those changes, both chips would link to https://localhost:8407/ui/project/Duck-Dock/storage/volumes

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+                project: item.project,
+                pool: item.pool,
+                type: "custom",
+                location: item.target ?? "",
+              } as LxdStorageVolume);
+            } else if (entityType === "bucket" && item.pool) {
+              to = getStorageBucketURL(item.name, item.pool, item.project);
+            } else if (entityType === "image") {
+              to = `/ui/project/${encodeURIComponent(item.project)}/images`;
+            }
```
edlerd: Can you extract a `getLinkTarget` function with early exits on the type check?

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+  if (lxdUsedBy.length === 0) return null;
+
+  let entityName = pluralize(
+    capitalizeFirstLetter(entityType.replaceAll("-", " ")),
+    lxdUsedBy.length,
+  );
+
+  if (entityType === "volume") {
+    entityName = pluralize("Custom volume", lxdUsedBy.length);
+  }
```
edlerd:
```suggestion
  const entityNameSingular =
    entityType === "volume"
      ? "Custom volume"
      : capitalizeFirstLetter(entityType.replaceAll("-", " "));
  const entityName = pluralize(entityNameSingular, lxdUsedBy.length);
```
A bit simpler?

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+  entityType: ResourceType;
+  usedBy: string[] | undefined;
+}
+
+const UsedByRow: FC<Props> = ({ entityType, usedBy }) => {
+  if (!usedBy || usedBy.length === 0) return null;
+
+  const { project } = useCurrentProject();
+
+  const lxdUsedBy = filterUsedByType(entityType, usedBy);
```
edlerd:
```suggestion
  const usedByItems = filterUsedByType(entityType, usedBy);
```
Maybe a better name? Not sure on this one though, yours might be better as well.

src/components/UsedByItem.tsx
```diff
@@ -12,6 +12,7 @@ interface Props {
   type: ResourceIconType;
   to: string;
   projectLinkDetailPage?: string;
+  itemProject?: string;
```
edlerd:
```suggestion
```
I don't think we need this here, this is the same as `item.project`.

src/pages/projects/actions/ProjectUsedBy.tsx
```diff
@@ -1,135 +1,23 @@
...
-        // the default profile is not blocking project deletion and can't be removed itself
-        (item) => !item.startsWith("/1.0/profiles/default"),
-      ),
-    ),
-    [IMAGES]: filterUsedByType("image", project.used_by),
-    [CUSTOM_VOLUMES]: filterUsedByType("volume", project.used_by),
-  };
+  const usedBy = project.used_by?.filter(
+    (t) => !t.startsWith("/1.0/profiles/default"),
+  );
```
edlerd, line 13:
```suggestion
  // the default profile is not blocking project deletion and can't be removed itself
  const usedBy = project.used_by?.filter(
    (t) => !t.startsWith("/1.0/profiles/default"),
  );
```
Shall we keep the comment?

src/pages/storage/StoragePoolSize.tsx
edlerd, line 1: Was this throwing any errors? I think this was okay before. The only case where `resourceList` has more than one element is if `poolResource.memberName` is defined. So maybe revert the changes here?

src/components/UsedByRow.tsx
edlerd: Depending on the storage pool: 

- Some pools are cluster member specific (like directory). For those we need to put the cluster member into the url.
- Other pools have a driver that is cluster wide (like ceph). For those there is no cluster member in the url.

See also this function building the url:

image

src/util/resourceDetails.tsx
```diff
@@ -57,7 +57,7 @@ export const extractResourceDetailsFromUrl = (
     resourceDetail.name = "server";
   }
 
-  if (resourceType === "storage_volume") {
+  if (resourceType === "storage_volume" || resourceType === "volume") {
```
edlerd, line 62: I see why this is needed, but think we should add a comment explaining why there are two values here.
```suggestion
  // permission selectors fetch their options from the metadata api, which uses "storage_volume"
  // ui internal resourceType from the ResourceIconType is slightly simpler "volume"
  if (resourceType === "storage_volume" || resourceType === "volume") {
```

src/pages/networks/NetworkAclUsedBy.tsx
edlerd, line 1: The used by list on this page uses a bit too much space for the sections. I think we should give the table to a max width

image

src/components/UsedByRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+    capitalizeFirstLetter(entityType.replaceAll("-", " ")),
+    lxdUsedBy.length,
+  );
+
+  if (entityType === "volume") {
+    entityName = pluralize("Custom volume", lxdUsedBy.length);
+  }
+
+  return (
+    <tr>
```
edlerd: I think we should display a thin line between the rows if there are more than one, something like this:

```
.used-by-row:not(:first-child) {
    border-top: $border-thin;
}
```
```suggestion
    <tr classname="used-by-row">
```

I think the lines improve especially very busy pages like pools:

image

edlerd (commented):
Thanks for making the used by lists consistent and easily reusable! Some more ideas below, then this should be good to merge from my side.

src/components/ErrorPage.tsx
```diff
@@ -57,7 +57,7 @@ If the problem continues, ensure your connection to the LXD server is active or
       <Notification severity="negative" title="Error">
         Something has gone wrong. If this issue persists,{" "}
         <a href={url} rel="noopener noreferrer" target="_blank">
-          please raise an issue on GitHub.
+          please raise an issue on Github.
```
edlerd: I think this was the right spelling.

src/components/UsedByRow.tsx
kimanhou: `filteredResources` ?

kimanhou:
In storage pool overview, the total count of used by list is displayed next to the header. It's not the case in other screens. Should we keep it ?
image

src/pages/storage/StoragePoolSize.tsx
kimanhou: Yes, the usual warning in console about React keys duplicated or missing.

image

Non clustered pools don't have a member name so the key would be empty.
If there are multiple undefined pool resources in `resourceList` then we need a way to differentiate them.

src/pages/networks/NetworkAclUsedBy.tsx
kimanhou: I applied the same rule as in `ProjectDeleteModal`: `width: 14rem` on the header and `width: 9rem` on small screens.

image

src/components/UsedByRow.tsx
kimanhou: It looks like the borders are already there by default.

Screenshot of ACL used by
image

I updated` _storage_detail_overview` CSS file to not target `used-by-row`.

src/components/UsedByRow.tsx
kimanhou: I think my `concrete-mule` volume was creating from an instance. But in `UsedByRow` (now `getLinkTarget`), I always set `type="custom"` if the entityType is volume. How do I know if the volume was created from an instance at that point ?

In `ProjectDeleteModal`, should I check if an instance has the same name as my volume, and then filter it out if so ?

src/components/UsedByRow.tsx
edlerd: Yeah seems better as well.

edlerd:
> In storage pool overview, the total count of used by list is displayed next to the header. It's not the case in other screens. Should we keep it ? 

Good catch, yes, I think we can remove it there.

src/pages/storage/StorageUsedBy.tsx
```diff
@@ -1,168 +1,21 @@
...
-  const data: Record<string, LxdUsedBy[]> = {
-    [INSTANCES]: filterUsedByType("instance", storage.used_by),
-    [PROFILES]: filterUsedByType("profile", storage.used_by),
-    [IMAGES]: filterUsedByType("image", storage.used_by),
-    [SNAPSHOTS]: filterUsedByType("snapshot", storage.used_by),
-    [CUSTOM_VOLUMES]: filterUsedByType("volume", storage.used_by),
-    [BUCKETS]: filterUsedByType("bucket", storage.used_by),
-  };
-
+const StorageUsedBy: FC<Props> = ({ storage }) => {
```
edlerd, line 9: There is an edge case here, when there are no usages. I think we should render a muted "none" or "-" next to used by.

image

src/components/UsedByRow.tsx
edlerd: This seems like an API bug. I wouldn't filter it, but report it as a bug instead to LXD.

The issue is `GET 1.0/storage-pools/:name?recursion=1` returns instances, and **not** the root volumes for instances. On the other hand `GET 1.0/projects/:name` returns both, the instance names and the volumes that are the root volumes for the instances. I think the second one should be omitted in the API response.

src/pages/storage/StorageUsedBy.tsx
edlerd: This is actually a problem on others used by section as well. I think there is value in showing all possible types as it communicates "This thing can be used by those other things", even if there is no usage in place. Also, it goes with the suggestion to never hide empty objects in user interfaces, so a user can build a mental modal more easily. They don't have to learn that the object is sometimes there and sometimes not, and what the effective condition for that is.

src/util/usedBy.tsx
```diff
@@ -92,3 +95,45 @@ export const getProfileInstances = (
...
+
+  if (entityType === "bucket" && resource.pool) {
+    return getStorageBucketURL(resource.name, resource.pool, resource.project);
+  }
+
+  if (entityType === "image") {
+    return `/ui/project/${encodedProject}/images`;
+  }
+
+  return `/ui/project/${encodedProject}/${entityType}/${encodedName}`;
```
edlerd: Can use `linkForInstanceDetail`

src/util/usedBy.tsx
```diff
@@ -92,3 +95,45 @@ export const getProfileInstances = (
...
   });
 };
+
+export const getLinkTarget = (resource: LxdUsedBy, entityType: string) => {
+  const encodedProject = encodeURIComponent(resource.project);
+  const encodedName = encodeURIComponent(resource.name);
+
+  if (entityType === "snapshot") {
+    if (resource.instance) {
+      return `/ui/project/${encodedProject}/instance/${encodeURIComponent(resource.instance)}/snapshots`;
```
edlerd: This can also use `linkForInstanceDetail` helper.

src/util/usedBy.tsx
```diff
@@ -92,3 +95,45 @@ export const getProfileInstances = (
     return project === instance.project;
   });
 };
+
+export const getLinkTarget = (resource: LxdUsedBy, entityType: string) => {
+  const encodedProject = encodeURIComponent(resource.project);
+  const encodedName = encodeURIComponent(resource.name);
```
edlerd:
```suggestion
```
With the helpers in place, we probably don't need these two and can build `encodedProject` in place for images.

src/util/usedBy.tsx
kimanhou: By the way, in src/api/instances.tsx on line 250, are the parameters in the wrong order ?
```
href: linkForInstanceDetail(project, name),
```

src/util/usedBy.tsx
kimanhou: But what if the entityType is not instance ?

src/util/usedBy.tsx
edlerd: > By the way, in src/api/instances.tsx on line 250, are the parameters in the wrong order ?
> 
> ```
> href: linkForInstanceDetail(project, name),
> ```

The order is mixed up, yes. Please, can you correct it?

src/util/usedBy.tsx
edlerd: Oh good point, I was missing this is generic. I think there are `instance`, `profile` and `network`. I wonder if we should have helpers for profile and network as well. Though this is scope creeping, so maybe keep as a follow up?

src/pages/storage/StorageUsedBy.tsx
kimanhou: Mmh I can't find a good UI

image

image

image

image

edlerd: I was first thinking last one of your screens and then considered the first one to be right. Wdyt?

src/util/usedBy.tsx
kimanhou: I think this is fine for now

src/components/UsedByRow.tsx
kimanhou: [Github issue](https://github.com/canonical/lxd/issues/17046) created on LXD repository

src/pages/storage/StorageUsedBy.tsx
kimanhou: First screenshot makes the most sense to me. Let's go with that. That will affect all use of `UsedByRow`.

src/components/UsedByRow.tsx
kimanhou: In the meantime, in `ProjectDeleteModal`, should I force the link of the chips to be the list of volumes (like it initially was before this refactoring): `https://localhost:8407/ui/project/{projectName}/storage/volumes`

Or do we accept to have a chip with a wrong link ?

edlerd: I think we should rely on the backend fix and not have a custom fix in place, so keep it as is on our end as that is the right way.

kimanhou:
It looks like the failing test is not related to my changes.

PR #1624, opened 2025-11-20
feat: [WD-30925] Network local peering CRUD

Kxiru:
## Done
LP = "Local Peering"

- Added Create LP functionality
- Added Delete LP functionality
- Added permissions (entitlements)

To do:
- [x] Add Update functionality
- [x] Use Project and Network Selector in Create Panel instead of Inputs
- [x] Diagnose why the page list does not refresh upon successful CRUD operations.
- [x] Prevent peering with self
- [x] Add text input for networks

## QA

1. Run the LXD-UI:
    - Pull the PR locally as the Demo server is not OVN-compatible (currently only OVN networks support Local Peering)
    - (Create and) Navigate to an OVN network, view the Local Peerings Tab and verify you can access it.
    - Attempt to create a new peering by clicking "Create Local Peering"
    - Attempt CRUD operations.

## Screenshots

image
image
image
image
image


src/api/network-local-peering.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return fetch(
+    `/1.0/networks/${encodeURIComponent(network)}/peers?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdNetworkPeer[]>) => {
+      return data.metadata;
+    });
+};
+
+export const createLocalPeering = async (
```
edlerd:
```suggestion
export const createNetworkPeer = async (
```
I would consider this file as the layer of transition between UI wording and API wording. So keeping the function names in here and the file name close to the api routes. Wdyt?

src/context/useLocalPeerings.tsx
```diff
@@ -0,0 +1,21 @@
+import { queryKeys } from "util/queryKeys";
+import type { UseQueryResult } from "@tanstack/react-query";
+import { useQuery } from "@tanstack/react-query";
+import type { LxdNetwork, LxdNetworkPeer } from "types/network";
+import { fetchNetworkPeers } from "api/network-local-peering";
+
+export const useLocalPeerings = (
```
edlerd, line 7: Good idea to introduce this, also I think for this layer it is ok to use the UI wording as is.

src/pages/networks/NetworkForwards.tsx
edlerd, line 1: Please revert these unrelated changes

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const queryClient = useQueryClient();
+  const { canEditNetwork } = useNetworkEntitlements();
+  const { project } = useCurrentProject();
+  const projectName = project?.name || "";
+  const toastNotify = useToastNotification();
+
+  const onFinish = () => {
+    toastNotify.success(
+      <>
+        Local peering {localPeering} deleted for network{" "}
```
edlerd: Please use a `ResourceLabel` for the peering as well. We need to introduce a new type for it in the resource label and resource icon.

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,92 @@
...
+  };
+
+  return (
+    <ConfirmationButton
+      loading={isLoading}
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: (
+          <p>
+            This will permanently delete local peering {localPeering}.
```
edlerd: This should also use the resource label, and I'd repeat the network name here, like in the notification above.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+      <NetworkSelector
+        id="targetNetwork"
+        name="targetNetwork"
+        label="Target network"
+        value={formik.values.targetNetwork}
+        setValue={(value) => {
+          formik.setFieldValue("targetNetwork", value);
+        }}
+        disabled={isChecked || isEditing || !!LPEditRestriction}
+        managedNetworks={managedNetworks}
```
edlerd: These should be the networks from the target project that was chosen.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+    >
+      {/* hidden submit to enable enter key in inputs */}
+      <Input type="submit" hidden value="Hidden input" />
+      <Input
+        {...getFormProps("name")}
+        type="text"
+        label="Name"
+        required
+        autoFocus
+        disabled={isEditing || !!LPEditRestriction}
```
edlerd: I think the name should be editable.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+        title={LPEditRestriction}
+      />
+      <AutoExpandingTextArea
+        {...getFormProps("description")}
+        label="Description"
+        disabled={!!LPEditRestriction}
+        title={LPEditRestriction}
+      />
+      <SelectableProjectSelector
```
edlerd: We also need a fallback to manually enter the target project if the user has no access to it.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,144 @@
...
+          controllerState,
+        ),
+      )
+      .required("Local peering name is required"),
+  });
+
+  const handleSuccess = (peerName: string) => {
+    toastNotify.success(
+      <>
+        Local peering <ResourceLabel type={"peering"} value={peerName} />{" "}
```
edlerd: lets make the label bold, or better use the resource link here.

src/pages/networks/panels/EditLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,170 @@
...
+    data: localPeer,
+    error,
+    isLoading,
+  } = useLocalPeering(network, project, localPeering ?? "");
+
+  const handleSuccess = () => {
+    toastNotify.success(
+      <>
+        Local peering{" "}
+        <ResourceLabel type="peering" value={localPeering ?? ""} /> updated.
```
edlerd: better use the resource link here and also mention the network as a resource link.

src/pages/networks/panels/EditLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  if (error) {
+    notify.failure("Loading local peering failed", error);
+  }
+
+  if (isLoading) {
+    return <Spinner className="u-loader" text="Loading..." isMainComponent />;
+  } else if (!localPeer) {
+    return <>Loading local peering failed</>;
+  }
```
edlerd: `!localPeer` is already handled above.

src/pages/networks/panels/EditLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const changeCount =
+    formik.values.description !== localPeer?.description ? 1 : 0;
+
+  return (
+    <>
+      <SidePanel>
+        <SidePanel.Header>
+          <SidePanel.HeaderTitle>
+            Edit local peering {localPeering}
```
edlerd, line 131: This needs truncation for when the name overflows.

src/pages/networks/forms/SelectableProjectSelector.tsx
```diff
@@ -0,0 +1,86 @@
...
+import type { LxdProject } from "types/project";
+
+interface Props {
+  value: string;
+  setValue: (value: string) => void;
+  projects: LxdProject[];
+  hasNoneOption?: boolean;
+}
+
+const SelectableProjectSelector: FC<
```
edlerd: This is a bit of an odd naming. I see we already have `ProjectSelection` for the navigation. Maybe we can find better names for this one and potentially rename the other one? Or keep the other one and find a better name for this one only.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+        managedNetworks={managedNetworks}
+        required={!isChecked}
+        help={
+          "Peering is limited to intra-cluster OVN networks; non-OVN or cross-cluster network peering is not supported."
+        }
+      />
+
+      <CheckboxInput
+        id={`local-peering-network-checkbox`}
+        label="I don't have access to the target network"
```
edlerd: I wonder if this is the right pattern and wording to use here. Better copy might be "Manually enter target network". Can you reach out to design to get a UX input on this, please?

Also use the same pattern for manually entering the project (see other comment).

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const queryClient = useQueryClient();
+  const { canEditNetwork } = useNetworkEntitlements();
+  const { project } = useCurrentProject();
+  const projectName = project?.name || "";
+  const toastNotify = useToastNotification();
+
+  const onFinish = () => {
+    toastNotify.success(
+      <>
+        Local peering <ResourceLabel type={"peering"} value={localPeering} />{" "}
```
edlerd: better make the label bold

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+        deleted for network{" "}
+        <ResourceLabel bold type="network" value={network.name} />.
```
edlerd: This should be a resource link. The idea is to always use the resource link. Except after deletion, when the link target has become invalid.

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <ConfirmationButton
+      loading={isLoading}
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: (
+          <p>
+            This will permanently delete local peering{" "}
+            <ResourceLabel type={"peering"} value={localPeering} /> .
```
edlerd: this can be a resource link

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <br /> for network{" "}
+            <ResourceLabel bold type="network" value={network.name} />. This
```
edlerd: this can be a resource link

src/pages/networks/NetworkPeers.tsx
```diff
@@ -27,29 +30,22 @@ const NetworkPeers: FC<Props> = ({ network, project }) => {
...
-    queryKey: [
-      queryKeys.projects,
-      project,
-      queryKeys.networks,
-      network.name,
-      queryKeys.peers,
-    ],
-    queryFn: async () => fetchNetworkPeers(network.name, project),
-  });
+  } = useLocalPeerings(network, project);
```
edlerd, line 34: The order of the api response is random, meaning if you have multiple peers and switch tabs, the order changes when coming back to the tab. Best apply a stable sorting to it before passing it to rendering. Seems best to introduce this even in the api layer. We do something similar, search for `ensureStableSorting` to see examples of it.

src/api/network-local-peering.tsx
```diff
@@ -0,0 +1,96 @@
...
+    .then((data: LxdApiResponse<LxdNetworkPeer[]>) => {
+      return data.metadata;
+    });
+};
+
+export const fetchNetworkPeer = async (
+  network: string,
+  project: string,
+  localPeering: string,
+): Promise<LxdNetworkPeer[]> => {
```
edlerd:
```suggestion
): Promise<LxdNetworkPeer> => {
```
This shouldn't be an array, right?

src/pages/networks/actions/EditLocalPeerBtn.tsx
```diff
@@ -0,0 +1,35 @@
...
+      key={`${localPeering}-edit`}
+      appearance="base"
+      hasIcon
+      onClick={() => {
+        panelParams.openEditLocalPeering(localPeering);
+      }}
+      title={
+        canEditNetwork(network)
+          ? "Edit local peer"
+          : "You do not have permission to edit this local peer"
```
edlerd:
```suggestion
          ? "Edit local peering"
          : "You do not have permission to edit this local peering"
```
A bit more consistent naming?

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
+
+const NetworkLocalPeeringForm: FC<Props> = ({ formik, network, isEditing }) => {
+  const { canEditNetwork } = useNetworkEntitlements();
+  const { project } = useCurrentProject();
+  const { data: projects = [] } = useProjects();
+  const { data: networks = [] } = useNetworks(project?.name || "default");
+  const [isChecked, setIsChecked] = useState(false);
+  const managedNetworks = networks
+    .filter((network_) => network_.type === ovnType)
```
edlerd:
```suggestion
    .filter((item) => item.type === ovnType)
```
Maybe better to avoid confusing the names. Same on the line below.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+      name: id,
+      onBlur: formik.handleBlur,
+      onChange: formik.handleChange,
+      value: formik.values[id] ?? "",
+      error: formik.touched[id] ? (formik.errors[id] as ReactNode) : null,
+      placeholder: `Enter ${id.replaceAll("-", " ")}`,
+    };
+  };
+
+  const LPEditRestriction = canEditNetwork(network)
```
edlerd:
```suggestion
  const editRestriction = canEditNetwork(network)
```
Please don't use custom abbreviations. I think in this component we can keep it generic though as the context clarifies.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
omarelkashef: This can also use typesWithLocalPeerings type.

edlerd: > This can also use typesWithLocalPeerings type.

Good point, that is better than redefining the types here.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
Kxiru: I've changed the useNetworks() to be the targetProject, but I think we clarified in discussion that networks can be peered irrespective of the project. I.e; cross-project. How does this change affect the implementation?

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
Kxiru: ResourceLink for the peering? Should it just link back to the peering page?

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
edlerd: I think by selecting a project first and a network second, that is exactly what we get -- peering cross projects.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
edlerd: Second thought: When manually entering the project, we should not try to load the networks from the project, but the network should directly be a text input. Because if a user can't see the project, they also can't read the networks in the project.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
edlerd: Yes, I think so. We do link to the list page where there is no detail page in other entities, like for images.

src/pages/networks/NetworkPeers.tsx
```diff
@@ -27,29 +31,25 @@ const NetworkPeers: FC<Props> = ({ network, project }) => {
...
   const headers = [
     { content: "Name", sortKey: "name" },
     { content: "Description", sortKey: "description" },
-    { content: "Peer project", sortKey: "project" },
-    { content: "Peer network", sortKey: "peerNetwork" },
-    { content: "Status", sortKey: "status" },
+    { content: "Target project", sortKey: "targetProject" },
+    { content: "Target network", sortKey: "targetNetwork" },
+    {
+      content: <span className="status-header">Status</span>,
```
edlerd, line 49: The class doesn't yet align the header with the content text, I think some scss needs to be applied as well for `.status-header`.

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,103 @@
...
+      loading={isLoading}
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: (
+          <p>
+            This will permanently delete local peering{" "}
+            <ResourceLink
+              type={"peering"}
+              value={localPeering}
+              to={`${networkURL}/local-peering`}
```
edlerd:
```suggestion
              to={`${networkURL}/local-peerings`}
```
Using the correct url here

src/pages/networks/actions/EditLocalPeerBtn.tsx
```diff
@@ -0,0 +1,35 @@
...
+  network: LxdNetwork;
+  localPeering: string;
+}
+
+const EditLocalPeerBtn: FC<Props> = ({ network, localPeering }) => {
+  const panelParams = usePanelParams();
+  const { canEditNetwork } = useNetworkEntitlements();
+
+  return (
+    <Button
```
edlerd, line 20: I think this is missing the disabled prop for when the permission is missing.

src/pages/networks/NetworkPeers.tsx
Kxiru: I see this in the codebase, but ensure stable sorting works because it takes an iterable to different entities (such as a locations array or used_by array) and sorts that, thus returning a "sorted" list of entities, however LxdNetworkPeer does not have any iterable parameters at all and therefore cannot be iterated on.

Kxiru: Do you have any other suggestions? Perhaps if we cannot sort in the API layer we can sort later? I thought we had a defaultSort option in our table somewhere?

edlerd: I meant to sort the entries on the api layer similar to how the other function is doing it. Not reusing it. We could sort solely on the name in `fetchPeers` before returning the data object.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { createNetworkPeer } from "api/network-local-peering";
+import NetworkLocalPeeringForm from "../forms/NetworkLocalPeeringForm";
+import type { LocalPeeringFormValues } from "../forms/NetworkLocalPeeringForm";
+import { testDuplicateLocalPeeringName } from "util/networks";
+
+interface Props {
+  network: LxdNetwork;
+}
+
+const CreateLocalPeeringPanel: FC<Props> = ({ network }) => {
```
edlerd, line 28: This blocks submit in a valid state:

image

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,158 @@
...
+  customTargetNetwork?: string;
+}
+
+interface Props {
+  formik: FormikProps<LocalPeeringFormValues>;
+  network: LxdNetwork;
+  isEditing?: boolean;
+}
+
+const NetworkLocalPeeringForm: FC<Props> = ({ formik, network, isEditing }) => {
```
edlerd, line 29: I think the manual option on network shouldn't show up as type ovn. Or maybeit should?

image

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
edlerd: Also the help text "Peering is limited to intra-cluster OVN networks; non-OVN or cross-cluster network peering is not supported." is missing in that path above.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
Kxiru: So the awkard thing about this, is to add an "other" option to the network list, it neets to be LxdNetwork type :sweat: . Meaning at minimum, it needs a type (and there are strict types here) and config (empty).

I feel like showing the Ovn type is okay for now as Ovn is a valid network type for peering, but so are Bridge networks. We can either not show the type at all for all networks, or some up with a separate way to implement the "other" feature. Checkboxes would circumvent this but we already went down that alley...

edlerd: I think the extra option is the right way to do it and wouldn't go for the checkbox option. We could weaken the type, use empty string or keep ovn as is to satisfy the network selector requirements.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
Kxiru: I've come to a new working version (pushed), but it only validates up to name and project input, for some reason, attempting to validate the network is being difficult and creating issues. Will continue to look into this.

src/pages/networks/NetworkPeers.tsx
Kxiru: image
image

I'm not sure I follow, please could you clarify?

edlerd: The "Status" text in the header should be aligned with the "Pending" and "Created" text below. Like on the instance list. This class was added, but the effects of the class (as in the instance list) are not applied to the local peerings table.

src/pages/networks/actions/EditLocalPeerBtn.tsx
```diff
@@ -0,0 +1,35 @@
...
+      key={`${localPeering}-edit`}
+      appearance="base"
+      hasIcon
+      onClick={() => {
+        panelParams.openEditLocalPeering(localPeering);
+      }}
+      title={editRestriction}
+      disabled={!!editRestriction}
```
edlerd: This is always disabled now.

src/pages/networks/actions/CreateNetworkPeeringBtn.tsx
edlerd, line 1: Creation fails with a valid state:

image

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+    { name: projectOtherLabel, config: {}, description: "" },
+  ];
+
+  const managedNetworks = networks
+    .filter((item) => typesWithLocalPeerings.includes(item.type))
+    .filter((item) => item.name !== network.name);
+
+  const networkList = [
+    ...managedNetworks,
+    { name: networkOtherLabel, type: "", config: {} } as LxdNetwork,
```
edlerd: On selection this leaves `()` as part of the selected label:

image

Kxiru: Yes sorry, this was an experiment I left behind. It also shows how no type is a bit awkward. I'll add Ovn back in for now.

edlerd: Sounds good. I just realized after submitting the review this might still be wip.

src/pages/networks/actions/CreateNetworkPeeringBtn.tsx
Kxiru: Thanks for this, fixing this one was tricky but I think I did it.

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const DeleteLocalPeerBtn: FC<Props> = ({ network, localPeering }) => {
+  const notify = useNotify();
+  const [isLoading, setLoading] = useState(false);
+  const queryClient = useQueryClient();
+  const { canEditNetwork } = useNetworkEntitlements();
+  const { project } = useCurrentProject();
+  const projectName = project?.name || "";
+  const toastNotify = useToastNotification();
+  const networkURL = `/ui/project/${encodeURIComponent(project?.name ?? "")}/network/${encodeURIComponent(network.name)}`;
```
edlerd:
```suggestion
  const networkURL = `/ui/project/${encodeURIComponent(projectName)}/network/${encodeURIComponent(network.name)}`;
```
A bit simpler

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const onFinish = () => {
```
edlerd:
```suggestion
  const onSuccess = () => {
```
A bit more precise

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,103 @@
...
+          <p>
+            This will permanently delete local peering{" "}
+            <ResourceLink
+              type={"peering"}
+              value={localPeering}
+              to={`${networkURL}/local-peerings`}
+            />{" "}
+            .
+            <br /> for network{" "}
+            <ResourceLink type="network" value={network.name} to={networkURL} />
```
edlerd:
```suggestion
            This will permanently delete the local peering{" "}
            <ResourceLink
              type={"peering"}
              value={localPeering}
              to={`${networkURL}/local-peerings`}
            />{" "}for network{" "}
            <ResourceLink type="network" value={network.name} to={networkURL} />.<br />
```
The message here is bamboozled.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const networkOtherLabel = "Manually enter network";
+
+  const localPeeringSchema = Yup.object().shape({
+    name: Yup.string()
+      .test(
+        ...testDuplicateLocalPeeringName(
+          panelParams.project,
+          network.name,
+          controllerState,
+        ),
```
edlerd, line 74: This doesn't work, I can submit the form with a duplicate name.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,157 @@
...
+      )
+      .required("Local peering name is required"),
+    targetProject: Yup.string().required("Target project is required"),
```
edlerd, line 44: I think we should also validate that either targetNetwork or customTargetNetwork are set. also if targetProject is set to other, we should expect the customTargetProject to be set.

src/pages/networks/NetworkPeers.tsx
edlerd, line 1: In the empty state the create button should move down under the text, like on the instance list or other list pages.

image

src/pages/networks/panels/EditLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,174 @@
...
+    );
+    closePanel();
+  };
+
+  const formik = useFormik<LocalPeeringFormValues>({
+    initialValues: {
+      name: localPeer?.name ?? "",
+      description: localPeer?.description,
+      targetProject: localPeer?.target_project ?? "",
+      targetNetwork: localPeer?.target_network ?? "",
```
edlerd, line 65: This needs to check if the user has access to the items and initialize those values to "other" if they don't have access. Currently those values are not visible on edit (see below). Or on edit we render both of these as plain inputs as they are disabled anyways. Then we don't rely on the values being present in the selectors.

image

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { createNetworkPeer } from "api/network-local-peering";
+import NetworkLocalPeeringForm from "../forms/NetworkLocalPeeringForm";
+import type { LocalPeeringFormValues } from "../forms/NetworkLocalPeeringForm";
+import { testDuplicateLocalPeeringName } from "util/networks";
+
+interface Props {
+  network: LxdNetwork;
+}
+
+const CreateLocalPeeringPanel: FC<Props> = ({ network }) => {
```
edlerd, line 28: When the user selects both project and network from the dropdowns they likely have permission on the target network and we should add a checkbox to create mutual peering that is selected by default. If it is checked, the peering entry from the target network to the source should be created as well.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,157 @@
...
+      </>,
+    );
+    closePanel();
+  };
+  const formik = useFormik<LocalPeeringFormValues>({
+    initialValues: {
+      name: "",
+      description: "",
+      targetProject: "",
+      targetNetwork: "",
```
edlerd, line 112: The default project should be the current project. That way we immediately show the target network selector, which is a bit more intuitive than the current way of hiding the network input initially.

src/util/networks.tsx
```diff
@@ -215,3 +218,26 @@ export const supportsNicDeviceAcls = (network?: LxdNetwork) => {
...
+    "deduplicate",
+    "A local peering with this name already exists",
+    async (value?: string) => {
+      return (
+        (excludeName && value === excludeName) ||
+        checkDuplicateName(
+          value,
+          project,
+          controllerState,
+          `networks/${encodeURIComponent(networkName)}/local-peerings`,
```
edlerd:
```suggestion
          `networks/${encodeURIComponent(networkName)}/peers`,
```
This should fix the deduplication.

src/pages/networks/NetworkPeers.tsx
Kxiru: Should I also apply this change to Network Forwards?

edlerd: Yes, please! Good catch!

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
Kxiru: So I can't yet figure out why, but I semi-successfully implemented this and it only works in some cases:

image
image

As you can see in the screenshots, we have a mutual pairing between a default network and anotherovnnetwork that has been created with the same peer name and project, but it stuck in pending state, not "created". Below this one, we have a successfully created mutual peering. I'm not sure if this is some kind of API nuance, a race condition or something else, but I'll push this for you to have a look.

edlerd: I can imagine that you can only have one peering between two networks. So as there is already a `Created` peering, the second one might fail internally when creating it between the same two networks.

Kxiru: I think you may be right, only one peering is permitted between two networks irrespective of the target project provided.
image

src/pages/networks/NetworkForwards.tsx
```diff
@@ -179,23 +179,27 @@ const NetworkForwards: FC<Props> = ({ network, project }) => {
...
+            <Button
+              appearance="positive"
+              className="u-float-right u-no-margin--bottom"
+              disabled
+              title="You do not have permission to create network forwards for this network"
+            >
+              <span>Create forward</span>
+            </Button>
+          )}
+        </>
```
edlerd: Please extract this into a reusable component, so we can also have the permission check below in the empty state without duplicating it.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const queryClient = useQueryClient();
+  const controllerState = useState<AbortController | null>(null);
+  const closePanel = () => {
+    panelParams.clear();
+    notify.clear();
+  };
+  const networkURL = `/ui/project/${encodeURIComponent(project?.name ?? "")}/network/${encodeURIComponent(network.name)}`;
+  const projectOtherLabel = "Manually enter project";
+  const networkOtherLabel = "Manually enter network";
+  const [createMutualPeering, setCreateMutualPeering] = useState(false);
```
edlerd: I think this should also use the formik state, and default to `true`.

Moving this to formik also simplifies the interfaces as we already share the `formik` object and don't need to add the setter for this local state.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+        network.name,
+        panelParams.project,
+        JSON.stringify(localPeeringPayload),
+      )
+        .then(() => {
+          if (createMutualPeering) {
+            const mutualPeeringPayload = {
+              name: values.name,
+              description: values.description,
+              target_project: values.targetProject,
```
edlerd:
```suggestion
              target_project: panelParams.project,
```
This should be the current project

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+              target_network: network.name,
+            };
+
+            createNetworkPeer(
+              values.targetNetwork,
+              panelParams.project,
```
edlerd:
```suggestion
              values.targetProject,
```
This should be the target project.

src/types/network.d.ts
```diff
@@ -4,7 +4,8 @@ export type LxdNetworkType =
   | "ovn"
   | "physical"
   | "macvlan"
-  | "sriov";
+  | "sriov"
+  | "";
```
edlerd: This seems wrong, empty string is not a valid network type.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
Kxiru: I am having trouble with this.

edlerd: Let's have a chat on this in sync to find a solution for it.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+            formik.setFieldValue("targetProject", value, false);
+            if (value !== projectOtherLabel) {
+              formik.setFieldValue("customTargetProject", "", false);
+              formik.setFieldValue("targetNetwork", "", false);
+              formik.validateForm();
+            }
+            if (value === projectOtherLabel) {
+              formik.setFieldValue("targetNetwork", networkOtherLabel, false);
+            }
+            formik.validateForm();
```
edlerd:
```suggestion
            formik.setFieldValue("targetProject", value, false);
            formik.setFieldValue("customTargetProject", "", false);
            if (value === projectOtherLabel) {
              formik.setFieldValue("targetNetwork", networkOtherLabel, false);
            } else {
              formik.setFieldValue("targetNetwork", "", false);
            }
            formik.setFieldValue("customTargetNetwork", "", false);
            formik.setFieldValue("createMutualPeering", false, false);
            setTimeout(() => {
              formik.validateForm();
            }, 100);
```
I think this simplifies, as we need to reset `customTargetProject`,  `targetNetwork`, `customTargetNetwork` and `createMutualPeering` whenever the project changes.

Also we must give some time for the state to propagate before revalidation.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+      />
+      <AutoExpandingTextArea
+        {...getFormProps("description")}
+        label="Description"
+        disabled={!!editRestriction}
+        title={editRestriction}
+      />
+      {isEditing ? (
+        <Input
+          id="customTargetProject"
```
edlerd:
```suggestion
          id="targetProject"
```
Small nitpick

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+      className={"local-peering-create-form"}
+    >
+      {/* hidden submit to enable enter key in inputs */}
+      <Input type="submit" hidden value="Hidden input" />
+      <Input
+        {...getFormProps("name")}
+        type="text"
+        label="Name"
+        required
+        autoFocus
```
edlerd:
```suggestion
        autoFocus={!isEditing}
```
On edit the description field should receive the focus.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+        disabled={isEditing || !!editRestriction}
+        title={editRestriction}
+      />
+      <AutoExpandingTextArea
+        {...getFormProps("description")}
+        label="Description"
```
edlerd, line 95:
```suggestion
        label="Description"
        autoFocus={isEditing}
```

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+            type="checkbox"
+            label="Create mutual peering"
+            checked={formik.values.createMutualPeering}
+            onChange={(e) => {
+              formik.setFieldValue("createMutualPeering", e.target.checked);
+            }}
+            disabled={!!editRestriction}
+            title={editRestriction}
+          />
+        )}
```
edlerd: I think we should always display this input, but disable the field with a help text if either network or project are set to  `other`.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <NetworkSelector
+            id="targetNetwork"
+            name="targetNetwork"
+            label="Target network"
+            value={formik.values.targetNetwork}
+            setValue={(value) => {
+              formik.setFieldValue("targetNetwork", value, false);
+              if (value !== networkOtherLabel) {
+                formik.setFieldValue("customTargetNetwork", "", false);
+              }
```
edlerd:
```suggestion
              if (value === networkOtherLabel) {
                formik.setFieldValue("createMutualPeering", false, true);
              } else {
                formik.setFieldValue("customTargetNetwork", "", false);
                formik.setFieldValue("createMutualPeering", true, false);
              }
            setTimeout(() => {
              formik.validateForm();
            }, 100);
```
I think this field should drive also the `createMutualPeering`. Wdyt?

And again some time before triggering validation for the state to propagate first.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+          title={editRestriction}
+          help={
+            "Peering is limited to intra-cluster OVN networks; non-OVN or cross-cluster network peering is not supported."
+          }
+        />
+      )}
+
+      {formik.values.targetNetwork !== networkOtherLabel &&
+        formik.values.targetProject !== projectOtherLabel &&
+        !isEditing && (
```
edlerd:
```suggestion
      {!isEditing && (
```
I think we should always display this input, but disable the field with a help text if either network or project are set to other.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,212 @@
...
+        network.name,
+        panelParams.project,
+        JSON.stringify(localPeeringPayload),
+      )
+        .then(() => {
+          if (formik.values.createMutualPeering) {
+            const mutualPeeringPayload = {
+              name: values.name,
+              description: values.description,
+              panelParams: panelParams.project,
```
edlerd:
```suggestion
              target_project: panelParams.project,
```
This is the wrong field name.

src/pages/networks/LocalPeeringStatusIcon.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface Props {
+  localPeer: LxdNetworkPeer;
+}
+
+const LocalPeeringStatusIcon: FC<Props> = ({ localPeer }) => {
+  const getIconNameForStatus = (status: string) => {
+    return (
+      {
+        Pending: "status-waiting-small",
+        Created: "status-succeeded-small",
```
edlerd, line 14: We also need to handle the `Errored` status here.

src/pages/networks/NetworkPeers.tsx
edlerd, line 1: `peer.target_network` can be empty string, and we should avoid rendering a chip in that case.

I got the state below by creating a mutual peering and deleting the other edge of it.

image

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+            value={formik.values.targetNetwork}
+            setValue={(value) => {
+              formik.setFieldValue("targetNetwork", value, false);
+              if (value !== networkOtherLabel) {
+                formik.setFieldValue("customTargetNetwork", "", false);
+              }
+              formik.validateForm();
+            }}
+            disabled={isEditing || !!editRestriction}
+            managedNetworks={networkList}
```
edlerd:
```suggestion
            networkList={networkList}
```
I think the prop on the `NetworkSelector` has a bit of an odd name. Shall we call it `networkList`? Because the selector has no opinion of which networks it displays and if they are managed or not.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+        )
+      )}
+
+      {(formik.values.targetNetwork === networkOtherLabel ||
+        formik.values.targetProject === projectOtherLabel) && (
+        <Input
+          id="customTargetNetwork"
+          name="customTargetNetwork"
+          type="text"
+          label="Target network"
```
edlerd: I think we should avoid the duplicated label, If selected a project and the manual network is selected

image

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,212 @@
...
+                queryKey: [
+                  queryKeys.projects,
+                  panelParams.project,
+                  queryKeys.networks,
+                  network.name,
+                  queryKeys.peers,
+                ],
+              });
+              handleSuccess(values.name);
+            });
```
edlerd: We need a catch on this one as well to handle the error case.

tests/permissions-read-only.spec.ts
```diff
@@ -277,7 +277,7 @@ test.describe("Given a user with Viewer Server permissions...", () => {
     await page.getByRole("link", { name: "Forwards" }).click();
     await expect(
       page.getByRole("button", { name: "Create forward" }),
-    ).toBeDisabled();
+    ).toBeVisible();
```
edlerd: This seems currently broken, and we shouldn't make the test broken as well, but fix the issue :)

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return (
+    <ConfirmationButton
+      loading={isLoading}
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: (
+          <p>
+            This will permanently delete the local peering{" "}
+            <ResourceLink
+              type={"peering"}
```
edlerd:
```suggestion
              type="peering"
```
Can simplify.

src/pages/networks/actions/CreateNetworkForwardBtn.tsx
edlerd, line 1: This file should be in the actions folder

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
Kxiru: Sure thing, do you have any ideas on the help text copy? I feel th elabel is pretty self-explanatory.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+          checked={formik.values.createMutualPeering}
+          onChange={(e) => {
+            formik.setFieldValue("createMutualPeering", e.target.checked);
+          }}
+          disabled={
+            !!editRestriction ||
+            formik.values.targetNetwork === networkOtherLabel ||
+            formik.values.targetProject === projectOtherLabel
+          }
+          title={editRestriction}
```
edlerd:
```suggestion
```
This is not displayed on edit.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+          id="createMutualPeering"
+          name="createMutualPeering"
+          type="checkbox"
+          label="Create mutual peering"
+          checked={formik.values.createMutualPeering}
+          onChange={(e) => {
+            formik.setFieldValue("createMutualPeering", e.target.checked);
+          }}
+          disabled={
+            !!editRestriction ||
```
edlerd:
```suggestion
```
This is not displayed on edit.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
edlerd: I thought we should clarify the disabling of the input. Something like "Unavailable when using manual project" and "Unavailable when using manual network". Depending on which one matches first. Wdyt?

src/pages/networks/actions/DeleteLocalPeerBtn.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { project } = useCurrentProject();
+  const projectName = project?.name || "";
+  const toastNotify = useToastNotification();
+  const networkURL = `/ui/project/${encodeURIComponent(projectName)}/network/${encodeURIComponent(network.name)}`;
+
+  const onSuccess = () => {
+    toastNotify.success(
+      <>
+        Local peering{" "}
+        <ResourceLabel type={"peering"} value={localPeering} bold /> deleted for
```
edlerd:
```suggestion
        <ResourceLabel type="peering" value={localPeering} bold /> deleted for
```
slightly simpler

src/pages/networks/NetworkPeers.tsx
```diff
@@ -109,39 +121,55 @@ const NetworkPeers: FC<Props> = ({ network, project }) => {
...
+            title="No local peerings found"
+          >
+            <p>There are no local peerings in this network and project.</p>
+            <p>
+              <DocLink docPath={`/howto/network_ovn_peers`}>
+                Learn more about local peering
+              </DocLink>
+            </p>
+            <CreateNetworkPeeringBtn
+              className={"empty-state-button"}
```
edlerd:
```suggestion
              className="empty-state-button"
```

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+  formik: FormikProps<LocalPeeringFormValues>;
+  network: LxdNetwork;
+  isEditing?: boolean;
+}
+
+const NetworkLocalPeeringForm: FC<Props> = ({ formik, network, isEditing }) => {
+  const { canEditNetwork } = useNetworkEntitlements();
+  const { data: projects = [] } = useProjects();
+  const { data: networks = [] } = useNetworks(
+    formik.values.targetProject || "default",
```
edlerd, line 35: This creates an invalid request with the project "Manually enter project" if that option is selected. I think we should use the third argument of `useNetwork` to disable the query if it matches that constant.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+      {/* hidden submit to enable enter key in inputs */}
+      <Input type="submit" hidden value="Hidden input" />
+      <Input
+        {...getFormProps("name")}
+        type="text"
+        label="Name"
+        required
+        autoFocus={!isEditing}
+        disabled={isEditing || !!editRestriction}
+        title={editRestriction}
```
edlerd: On edit we should display a title value that clarifies why it is disabled. Just like the target project below. Something like "Name cannot be changed"

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+        autoFocus={isEditing}
+      />
+      {isEditing ? (
+        <Input
+          id="targetProject"
+          type="text"
+          label="Target project"
+          value={formik.values.targetProject}
+          disabled={isEditing}
+          title={"Target project cannot be changed"}
```
edlerd:
```suggestion
          title="Target project cannot be changed"
```

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+        title={editRestriction}
+        autoFocus={isEditing}
+      />
+      {isEditing ? (
+        <Input
+          id="targetProject"
+          type="text"
+          label="Target project"
+          value={formik.values.targetProject}
+          disabled={isEditing}
```
edlerd:
```suggestion
          disabled
```
This is always true, a bit simpler.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+        <Input
+          id="customTargetProject"
+          name="customTargetProject"
+          type="text"
+          placeholder="Enter target project name"
+          value={formik.values.customTargetProject}
+          onChange={formik.handleChange}
+          onBlur={formik.handleBlur}
+          disabled={isEditing || !!editRestriction}
+          title={editRestriction}
```
edlerd:
```suggestion
```
A bit simpler. We should never hit those code branch as edit is handled above.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+        />
+      )}
+
+      {isEditing ? (
+        <Input
+          id="targetNetwork"
+          type="text"
+          label="Target network"
+          value={formik.values.targetNetwork}
+          disabled={isEditing}
```
edlerd:
```suggestion
          disabled
```

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+          title={"Target network cannot be changed"}
```
edlerd:
```suggestion
          title="Target network cannot be changed"
```

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+                formik.setFieldValue("createMutualPeering", false, true);
+              } else {
+                formik.setFieldValue("customTargetNetwork", "", false);
+                formik.setFieldValue("createMutualPeering", true, false);
+              }
+              setTimeout(() => {
+                formik.validateForm();
+              }, 100);
+            }}
+            disabled={isEditing || !!editRestriction}
```
edlerd:
```suggestion
```

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+            (formik.values.targetNetwork !== networkOtherLabel ||
+              formik.values.targetProject === projectOtherLabel) &&
+            "Target network"
+          }
+          placeholder="Enter target network name"
+          value={formik.values.customTargetNetwork}
+          onChange={formik.handleChange}
+          onBlur={formik.handleBlur}
+          disabled={isEditing || !!editRestriction}
+          title={editRestriction}
```
edlerd:
```suggestion
```

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
```diff
@@ -0,0 +1,234 @@
...
+          help={
+            "Peering is limited to intra-cluster OVN networks; non-OVN or cross-cluster network peering is not supported."
+          }
```
edlerd:
```suggestion
          help="Peering is limited to intra-cluster OVN networks; non-OVN or cross-cluster network peering is not supported."
          required
```
A bit simpler.

This should be required if displayed.

src/pages/networks/actions/CreateNetworkPeeringBtn.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <Button
+      appearance="positive"
+      hasIcon={!isSmallScreen}
+      onClick={() => {
+        panelParams.openCreateLocalPeering();
+      }}
+      className={classnames(
+        "p-button--positive u-no-margin--bottom",
```
edlerd: On small screens, this is positioned a bit too close to the table below

image

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
Kxiru: I think I have addressed this, but please let me know if I should change the approach.

src/pages/networks/actions/CreateNetworkPeeringBtn.tsx
Kxiru: Will address this next.

src/pages/networks/forms/NetworkLocalPeeringForm.tsx
edlerd: It seems like you set the third argument for useNetworks to false, but it should be dynamic depending on the project selected.

src/pages/networks/NetworkPeers.tsx
```diff
@@ -67,38 +67,50 @@ const NetworkPeers: FC<Props> = ({ network, project }) => {
...
         },
         {
-          content: peer.status,
+          content: <LocalPeeringStatusIcon localPeer={peer} />,
           role: "cell",
           "aria-label": "Status",
         },
+        {
+          content: (
+            <LocalPeeringActions network={network} localPeering={peer.name} />
```
edlerd: How about calling it consistently, instead of `localPeer`, `localPeering` and `peer` in here. We should have a consistent wording in all components.

edlerd (commented):
Nearly perfect :)

I think only three things remain for this to merge: consistent naming, the spacing on smaller screens and the query deactivation.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -0,0 +1,217 @@
...
+                    network.name,
+                    queryKeys.peers,
+                  ],
+                });
+                handleSuccess(values.name);
+              })
+              .catch((e) => {
+                formik.setSubmitting(false);
+                notify.failure(`Mutual local peering creation failed`, e);
+              });
```
edlerd, line 162: We should invalidate the query in this branch as the first edge of the peering was created successfully. Maybe put the invalidation into a function to simplify.

PR #1632, opened 2025-11-25
feat(DeleteInstanceBtn): force stop and delete running or frozen instance

kimanhou:
## Done

- Delete a running instance
- Delete a frozen instance

Note: bulk delete has been left untouched, meaning that you cannot select a running instance and delete it from the instance list screen. Is it acceptable ?

## QA

1. Run the LXD-UI:
    - Delete a running instance: 
        - go to a running instance page
        - make sure the delete button is enabled
        - click on delete button, a confirmation modal will open 
        - in the modal, make sure the delete button is disabled
        - tick the checkbox, it should enable the delete button
        - click on delete
        - delete should work and the it should navigate to the instance list screen
    - Delete a frozen instance: to freeze an instance, run `lxc pause {instanceName}`. Repeat above steps, the modal should display "The instance is currently frozen" instead of "The instance is currently running"

## Screenshots

image

### DIFF
--- a/src/pages/instances/actions/DeleteInstanceBtn.tsx
+++ b/src/pages/instances/actions/DeleteInstanceBtn.tsx
@@ -1,6 +1,6 @@
 import type { FC } from "react";
 import { useState } from "react";
-import { deleteInstance } from "api/instances";
+import { deleteInstance, stopInstance } from "api/instances";
 import type { LxdInstance } from "types/instance";
 import { useNavigate } from "react-router-dom";
 import { deletableStatuses } from "util/instanceDelete";
@@ -17,6 +17,9 @@ import { useInstanceLoading } from "context/instanceLoading";
 import ResourceLabel from "components/ResourceLabel";
 import InstanceLinkChip from "../InstanceLinkChip";
 import { useInstanceEntitlements } from "util/entitlements/instances";
+import { isInstanceFrozen, isInstanceRunning } from "util/instanceStatus";
+import { Notification } from "@canonical/react-components";
+import ConfirmationForce from "components/ConfirmationForce";
 
 interface Props {
   instance: LxdInstance;
@@ -38,9 +41,12 @@ const DeleteInstanceBtn: FC<Props> = ({
   const [isLoading, setLoading] = useState(false);
   const navigate = useNavigate();
   const { canDeleteInstance } = useInstanceEntitlements();
+  const [isForce, setForce] = useState(false);
 
-  const handleDelete = () => {
-    setLoading(true);
+  const isRunningOrFrozen =
+    isInstanceRunning(instance) || isInstanceFrozen(instance);
+
+  const doDelete = () => {
     const instanceLink = <InstanceLinkChip instance={instance} />;
 
     deleteInstance(instance)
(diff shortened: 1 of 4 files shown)


src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -106,22 +148,45 @@ const DeleteInstanceBtn: FC<Props> = ({
...
+            {isRunningOrFrozen && (
+              <>
+                <br />
+                <br />
+                <Notification
+                  severity="caution"
+                  title={`The instance is currently ${isInstanceFrozen(instance) ? "frozen" : "running"}.`}
+                >
+                  It will be force stopped before deletion. Tick the box below
+                  to proceed.
```
edlerd:
```suggestion
                  Confirm to force delete it to continue.
```
I don't think we have to expose the full internals to the user here. A bit more concise?

src/util/instanceDelete.tsx
```diff
@@ -1,3 +1,10 @@
 import type { LxdInstanceStatus } from "types/instance";
 
-export const deletableStatuses: LxdInstanceStatus[] = ["Error", "Stopped"];
+export const deletableStatuses: LxdInstanceStatus[] = [
+  "Error",
+  "Stopped",
+  "Running",
+  "Frozen",
+];
```
edlerd, line 8: Do we still need this? I think with the force flag, we always allow deletion of single instances.

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -106,22 +148,45 @@ const DeleteInstanceBtn: FC<Props> = ({
...
+                </Notification>
+              </>
+            )}
+          </div>
         ),
         onConfirm: handleDelete,
         confirmButtonLabel: "Delete",
+        confirmExtra: isRunningOrFrozen ? (
+          <ConfirmationForce
+            label="Force stop and delete"
```
edlerd:
```suggestion
            label="Force delete"
```
I think this should be consistent with the CLI, where I assume it is just a force flag without notion of stop.

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -83,6 +89,42 @@ const DeleteInstanceBtn: FC<Props> = ({
...
+    if (isRunningOrFrozen) {
+      const instanceLink = <InstanceLinkChip instance={instance} />;
+
+      stopInstance(instance, true)
+        .then((operation) => {
+          eventQueue.set(
+            operation.metadata.id,
+            () => {
+              doDelete();
+            },
```
edlerd:
```suggestion
            doDelete,
```
A bit simpler.

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -106,22 +148,45 @@ const DeleteInstanceBtn: FC<Props> = ({
...
             <ResourceLabel type={instance.type} value={instance.name} bold />.
             <br />
             This action cannot be undone, and can result in data loss.
-          </p>
+            {isRunningOrFrozen && (
+              <>
+                <br />
+                <br />
+                <Notification
+                  severity="caution"
```
edlerd, line 161:
```suggestion
                  severity="caution"
                  borderless
```
I think this might look better without border, wdyt?

src/util/instanceDelete.tsx
kimanhou: Here are all the instance statuses: 
```
   "Error"
  | "Freezing"
  | "Frozen"
  | "Restarting"
  | "Running"
  | "Starting"
  | "Stopped"
  | "Stopping"
  | "Migrating"
```

For now, we don't use the force flag so would it work if the instance is migrating ?

src/pages/instances/actions/DeleteInstanceBtn.tsx
kimanhou: I replaced it by `Confirm to force delete it.`

src/pages/instances/actions/DeleteInstanceBtn.tsx
kimanhou: I prefer with the border, especially because we don't use borderless notification anywhere.

src/util/instanceDelete.tsx
edlerd: Without the flag, I suspect a deletion would fail during migration. The other *ing statuses are not coming from the API, but are UI internal. So let's keep it as is for now, seems correct as is.

PR #1634, opened 2025-11-25
feat: [WD-31107] Add profile copy button and modal

Kxiru:
## Done

- Added responsive profile copy button
- Added modal with name input.

## QA

1. Run the LXD-UI:
    - Create a new profile with some configurations and copy (duplicate) it.
    - Verify all configurations have carried over.

## Screenshots

image
image

### DIFF
--- a/src/pages/profiles/actions/CopyProfileBtn.tsx
+++ b/src/pages/profiles/actions/CopyProfileBtn.tsx
@@ -1,0 +1,39 @@
+import type { FC } from "react";
+import type { LxdProfile } from "types/profile";
+import { Button, Icon, usePortal } from "@canonical/react-components";
+import classnames from "classnames";
+import CopyProfileForm from "../forms/CopyProfileForm";
+
+interface Props {
+  profile: LxdProfile;
+  className?: string;
+}
+
+const CopyProfileBtn: FC<Props> = ({ profile, className }) => {
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const handleClose = () => {
+    closePortal();
+  };
+
+  return (
+    <>
+      {isOpen && (
+        <Portal>
+          <CopyProfileForm close={handleClose} profile={profile} />
+        </Portal>
+      )}
+      <Button
+        appearance="default"
+        aria-label="Copy profile"
+        className={classnames("u-no-margin--bottom has-icon", className)}
+        onClick={openPortal}
+        title={"Copy profile"}
+      >
+        <Icon name="canvas" />
+        <span>Copy</span>
+      </Button>
+    </>
+  );
+};
+
+export default CopyProfileBtn;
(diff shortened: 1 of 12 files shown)


omarelkashef:
image
I am getting an "Only alphanumeric and hyphen characters are allowed" error even though I only clicked on the copy button and did not change the name. It would also be nice if I get an error that this project name already exists in the input before I submit.

src/pages/profiles/actions/CopyProfileBtn.tsx
```diff
@@ -0,0 +1,39 @@
+import type { FC } from "react";
+import type { LxdProfile } from "types/profile";
+import { Button, Icon, usePortal } from "@canonical/react-components";
+import classnames from "classnames";
+import CopyProfileForm from "../forms/CopyProfileForm";
+
+interface Props {
+  profile: LxdProfile;
+  classname?: string;
```
omarelkashef: I think the convention is className.

src/pages/profiles/actions/DeleteProfileBtn.tsx
```diff
@@ -19,9 +19,10 @@ import { useProfileEntitlements } from "util/entitlements/profiles";
 interface Props {
   profile: LxdProfile;
   project: string;
+  classname?: string;
```
omarelkashef: Here also className

src/pages/profiles/actions/ProfileDetailActions.tsx
```diff
@@ -0,0 +1,60 @@
...
+      project={project}
+      classname={classname}
+    />,
+  ];
+
+  return (
+    <>
+      {isSmallScreen ? (
+        <ContextualMenu
+          closeOnOutsideClick={false}
```
omarelkashef, line 37: Is there a reason closeOnOutsideClick is false?

Kxiru:
> image I am getting an "Only alphanumeric and hyphen characters are allowed" error even though I only clicked on the copy button and did not change the name.

Ah, thank you. This is probably because your initial profile had a space in it which has transferred to the new input. I will find out whether spaces can be valid (they should be since the previous profile just had one), so on second thought, I'll just take out that line of validation.

src/pages/profiles/actions/ProfileDetailActions.tsx
Kxiru: Yes, when the page is in small screen mode, this action menu becomes a dropdown, and being able to close the dropdown on clicking outside of it is an industry and LXD standard.

omarelkashef: Yes, that is what I assumed but when it is set to false, the dropdown does not close when you click outside.

edlerd: You have to set this to false. Otherwise interacting with the copy modal will close the dropdown and then the copy modal will unmount. So you wouldn't be able to interact with the modal on small screens.

Though there should be a callback, that if the copy modal is cancelled or finished, the contextual menu closes. See the copy modal on instance page in small screens. It receives an on close callback that triggers the contextual menu to close as well, when the modal closes.

omarelkashef: Oh I understand! and I agree with the callback idea.

omarelkashef (approved):
Thank you for the iteration! The QA and code look good now.

PR #1637, opened 2025-11-25
feat(ClusterMemberList): add Memory, remove failure domain

kimanhou:
## Done

- Remove failure domain from cluster member list
- Add memory column with a Meter in cluster member list
- Add Meter in cluster member detail hardware page

## QA

1. Run the LXD-UI:
    - See cluster member list

## Screenshots

image
</br></br>
ClusterMemberDetailMemory
image

### DIFF
--- a/src/pages/cluster/ClusterMemberDetailMemory.tsx
+++ b/src/pages/cluster/ClusterMemberDetailMemory.tsx
@@ -2,20 +2,28 @@ import type { FC } from "react";
 import type { LxdResources } from "types/resources";
 import { humanFileSize } from "util/helpers";
 import Meter from "components/Meter";
-import type { LxdClusterMemberState } from "types/cluster";
+import type { LxdClusterMemberState, LxdClusterMember } from "types/cluster";
+import ClusterMemberMemoryUsage from "pages/cluster/ClusterMemberMemoryUsage";
 
 interface Props {
   resources: LxdResources;
   state?: LxdClusterMemberState;
+  member: LxdClusterMember;
 }
 
-const ClusterMemberDetailMemory: FC<Props> = ({ resources, state }) => {
+const ClusterMemberDetailMemory: FC<Props> = ({ resources, state, member }) => {
   const totalSwap = state?.sysinfo.free_swap ?? 0;
   const freeSwap = state?.sysinfo.free_swap ?? 0;
 
   return (
     <table>
       <tbody>
+        <tr>
+          <th className="u-text--muted">Overview</th>
+          <td>
+            <ClusterMemberMemoryUsage member={member} />
+          </td>
+        </tr>
         <tr>
           <th className="u-text--muted">Total</th>
           <td>{humanFileSize(resources?.memory?.total ?? 0)}</td>
--- a/src/pages/cluster/ClusterMemberList.tsx
+++ b/src/pages/cluster/ClusterMemberList.tsx
@@ -17,6 +17,7 @@ import ClusterMemberActions from "pages/cluster/ClusterMemberActions";
 import { useClusterMembers } from "context/useClusterMembers";
 import usePanelParams from "util/usePanelParams";
 import ClusterMemberStatus from "pages/cluster/ClusterMemberStatus";
+import ClusterMemberMemoryUsage from "pages/cluster/ClusterMemberMemoryUsage";
 import { useMemberLoading } from "context/memberLoading";
 import { useServerEntitlements } from "util/entitlements/server";
 
@@ -44,9 +45,8 @@ const ClusterMemberList: FC = () => {
     },
     { content: "Roles", sortKey: "roles", className: "roles" },
     {
-      content: "Failure domain",
-      className: "failure-domain",
-      sortKey: "failureDomain",
+      content: "Memory",
+      className: "memory",
     },
     {
       content: "Description",
@@ -111,10 +111,10 @@ const ClusterMemberList: FC = () => {
           className: "roles",
         },
         {
-          content: member.failure_domain,
+          content: <ClusterMemberMemoryUsage member={member} />,
           role: "cell",
-          "aria-label": "Failure domain",
-          className: "failure-domain",
+          "aria-label": "Memory",
+          className: "memory",
         },
         {
           content: member.description,
@@ -144,7 +144,6 @@ const ClusterMemberList: FC = () => {
       sortData: {
         name: member.server_name.toLowerCase(),
         status: member.status.toLowerCase(),
-        failureDomain: member.failure_domain.toLowerCase(),
         roles: member.roles,
         description: member.description?.toLowerCase(),
         groups: groupCount,
--- a/src/pages/cluster/ClusterMemberMemoryUsage.tsx
+++ b/src/pages/cluster/ClusterMemberMemoryUsage.tsx
@@ -1,0 +1,55 @@
+import type { FC } from "react";
+import { useQuery } from "@tanstack/react-query";
+import Meter from "components/Meter";
+import { fetchClusterMemberState } from "api/cluster-members";
+import { queryKeys } from "util/queryKeys";
+import { humanFileSize } from "util/helpers";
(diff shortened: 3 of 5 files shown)


src/pages/cluster/ClusterMemberList.tsx
```diff
@@ -144,7 +145,7 @@ const ClusterMemberList: FC = () => {
       sortData: {
         name: member.server_name.toLowerCase(),
         status: member.status.toLowerCase(),
-        failureDomain: member.failure_domain.toLowerCase(),
+        memoryUsage: 0,
```
edlerd: Better remove the sort key on the header and here, so we can't sort by this column.

src/pages/cluster/ClusterMemberMemoryUsage.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const { data: state } = useQuery({
+    queryKey: [
+      queryKeys.cluster,
+      queryKeys.members,
+      member.server_name,
+      queryKeys.state,
+    ],
+    queryFn: async () => fetchClusterMemberState(member.server_name),
+  });
```
edlerd: This is rather expensive to fetch for each member the resources *and* state. Maybe we avoid this call and remove the hover text to keep the overhead smaller?

src/pages/cluster/ClusterMemberMemoryUsage.tsx
```diff
@@ -0,0 +1,76 @@
...
+const ClusterMemberMemoryUsage: FC<Props> = ({ member }) => {
+  const { data: resources } = useQuery({
+    queryKey: [
+      queryKeys.cluster,
+      queryKeys.members,
+      member.server_name,
+      queryKeys.resources,
+    ],
+    queryFn: async () => fetchResources(member.server_name),
+  });
```
edlerd: This is rather expensive to fetch for each member the resources and state. Maybe we avoid this call and rely solely on the state?

src/pages/cluster/ClusterMemberMemoryUsage.tsx
edlerd, line 22: I think we can use `state.sysinfo.total_ram`, `state.sysinfo.free_ram` as `percentage` on the `Meter`.

For the `secondaryPercentage` on the `Meter` rely on  `state.sysinfo.shared_ram + state.sysinfo.shared.buffered`.

Similar to how the `InstanceUsageMemory` is doing it.

src/pages/cluster/ClusterMemberMemoryUsage.tsx
```diff
@@ -0,0 +1,76 @@
...
+      hoverText += `shared: ${humanFileSize(state.sysinfo.shared_ram)}\n`;
+    }
+    if (state.sysinfo.buffered_ram > 0) {
+      hoverText += `buffered: ${humanFileSize(state.sysinfo.buffered_ram)}\n`;
+    }
+  } else {
+    hoverText += `free: ${humanFileSize(free)}\n`;
+  }
+
+  hoverText += `total: ${humanFileSize(total)}`;
```
edlerd: I'd slightly change this to align with `InstanceUsageMemory`. Remove the total and reduce it to 3 lines:

free: ...
used: ..
cached: (buffered + shared)

src/pages/cluster/ClusterMemberMemoryUsage.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { fetchClusterMemberState } from "api/cluster-members";
+import { queryKeys } from "util/queryKeys";
+import { humanFileSize } from "util/helpers";
+import type { LxdClusterMember } from "types/cluster";
+
+interface Props {
+  member: LxdClusterMember;
+}
+
+const ClusterMemberMemoryUsage: FC<Props> = ({ member }) => {
```
edlerd, line 13: I think we should show this in the `ClusterMemberDetailMemory` as well.

edlerd (commented):
Thanks for adding this so quickly, some thoughts on it below.

src/pages/cluster/ClusterMemberMemoryUsage.tsx
kimanhou: What's cached ?

edlerd: Cached is the sum of buffered and shared memory :D

I was trying to align this with the instance memory reporting. Cache means memory that is "not really used" by programs. Cached memory will be freed by the OS as it gets memory pressure from running programs. Interestingly, instances report directly cached memory, but here LXD api reports it split between two values.

src/pages/cluster/ClusterMemberList.tsx
```diff
@@ -44,9 +45,9 @@ const ClusterMemberList: FC = () => {
     },
     { content: "Roles", sortKey: "roles", className: "roles" },
     {
-      content: "Failure domain",
-      className: "failure-domain",
-      sortKey: "failureDomain",
+      content: "Memory",
+      className: "memory",
+      sortKey: "memory",
```
edlerd:
```suggestion
```
Disable sorting by this column.

src/pages/cluster/ClusterMemberMemoryUsage.tsx
```diff
@@ -0,0 +1,67 @@
...
+    return <span className="u-text--muted">No data</span>;
+  }
+
+  const memoryText = `${humanFileSize(usedMemory)} of ${humanFileSize(totalMemory)}`;
+
+  let hoverText = `free: ${humanFileSize(freeMemory)}\n`;
+  hoverText += `used: ${humanFileSize(usedMemory)}\n`;
+  if (sharedMemory > 0) {
+    hoverText += `cached: ${humanFileSize(bufferedMemory + sharedMemory)}\n`;
+  }
```
edlerd:
```suggestion
    hoverText += `cached: ${humanFileSize(bufferedMemory + sharedMemory)}\n`;
```
I think we can drop the conditional here.

src/pages/cluster/ClusterMemberMemoryUsage.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const memoryPercentage =
+    totalMemory > 0 ? (usedMemory / totalMemory) * 100 : 0;
+
+  const secondaryMemory = sharedMemory + bufferedMemory;
+  const secondaryPercentage =
+    totalMemory > 0 ? (secondaryMemory / totalMemory) * 100 : 0;
+
+  if (totalMemory === 0) {
+    return <span className="u-text--muted">No data</span>;
+  }
```
edlerd: Move this a bit up and we can remove the conditions above checking `totalMemory > 0 ? `

src/pages/cluster/ClusterMemberDetailMemory.tsx
```diff
@@ -2,82 +2,91 @@ import type { FC } from "react";
...
-          </td>
-        </tr>
-        {resources?.memory.nodes?.map((node, index) => (
-          <tr key={index}>
-            <th className="u-text--muted">NUMA Node {node.numa_node}</th>
+    <>
+      <ClusterMemberMemoryUsage
+        member={member}
+        className="u-sv1 memory-usage"
+      />
```
edlerd: I think this can be in the table, with a label of "Overview". Like this, it can take a lot of width:

image

src/pages/cluster/ClusterMemberMemoryUsage.tsx
```diff
@@ -0,0 +1,67 @@
...
+      queryKeys.members,
+      member.server_name,
+      queryKeys.state,
+    ],
+    queryFn: async () => fetchClusterMemberState(member.server_name),
+  });
+
+  if (!state?.sysinfo) {
+    return <span className="u-text--muted">-</span>;
+  }
```
edlerd: This should be combined with the `totalMemory === 0` check below. Maybe we don't actually need it as that check below catches the same situation.

PR #1640, opened 2025-11-26
Link nic static IP page from network lease and IPAM pages

omarelkashef:
## Done

- Link nic static IP page from network lease and IPAM pages

## QA

1. Run the LXD-UI:
    - Go to Networks page
    - Choose a network that supports static ip (bridge or OVN)
    - Go to Leases page
    - Click on the IP address for an instance
    - Ensure you are navigated to the  Configuration -> Networks tab of that instance.

## Screenshots

image
image
image

### DIFF
--- a/src/components/InstanceIpEdit.tsx
+++ b/src/components/InstanceIpEdit.tsx
@@ -1,0 +1,32 @@
+import type { FC } from "react";
+import { Icon } from "@canonical/react-components";
+import { Link } from "react-router-dom";
+
+interface Props {
+  address: string;
+  instanceName: string;
+  projectName: string;
+}
+
+export const InstanceIpEdit: FC<Props> = ({
+  address,
+  instanceName,
+  projectName,
+}) => {
+  return (
+    <div className="instance-ip-edit-content">
+      <div className="u-truncate instance-ip-edit-address" title={address}>
+        {address}
+      </div>
+      <Link
+        to={`/ui/project/${encodeURIComponent(projectName)}/instance/${encodeURIComponent(instanceName)}/configuration/network`}
+        rel="noopener noreferrer"
+        className="p-button-base instance-ip-edit-link is-dense"
+        type="button"
+        title={`Edit network configuration for instance ${instanceName}`}
+      >
+        <Icon name="edit" />
+      </Link>
+    </div>
+  );
+};
--- a/src/pages/networks/NetworkIPAM.tsx
+++ b/src/pages/networks/NetworkIPAM.tsx
@@ -22,6 +22,10 @@ import { filterUsedByType } from "util/usedBy";
 import UsedByItem from "components/UsedByItem";
 import ResourceLink from "components/ResourceLink";
 import DocLink from "components/DocLink";
+import { InstanceIpEdit } from "components/InstanceIpEdit";
+import { typesWithNicStaticIPSupport } from "util/networks";
+import { useNetworks } from "context/useNetworks";
+import type { LxdNetworkAllocation } from "types/network";
 
 const NetworkIPAM: FC = () => {
   const notify = useNotify();
@@ -36,6 +40,12 @@ const NetworkIPAM: FC = () => {
     queryFn: async () => fetchNetworkAllocations(project ?? "default"),
   });
 
+  const { data: networks } = useNetworks(project ?? "");
+
+  const getNetworkTypeFromAllocation = (allocation: LxdNetworkAllocation) => {
+    return networks?.find((n) => n.name === allocation.network)?.type;
+  };
+
   useEffect(() => {
     if (error) {
       notify.failure("Loading network allocations failed", error);
@@ -49,14 +59,17 @@ const NetworkIPAM: FC = () => {
   const headers = [
     { content: "Type", sortKey: "type", className: "type" },
     { content: "Used by", sortKey: "usedBy", className: "usedBy" },
-    { content: "Address", sortKey: "address", className: "address" },
     { content: "Network", sortKey: "network", className: "network" },
     { content: "NAT", sortKey: "nat", className: "nat" },
     { content: "MAC address", sortKey: "hwaddress", className: "hwaddr" },
+    { content: "Address", sortKey: "address", className: "address" },
   ];
 
   const rows = allocations.map((allocation) => {
-    const usedByItems = filterUsedByType(allocation.type, [allocation.used_by]);
+    const usedByItemArray = filterUsedByType(allocation.type, [
+      allocation.used_by,
+    ]);
+    const usedByItem = usedByItemArray.length === 1 ? usedByItemArray[0] : null;
 
     const getUsedByUrl = (item: LxdUsedBy) => {
       if (allocation.type === "instance") {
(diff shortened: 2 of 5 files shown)


src/pages/networks/NetworkLeases.tsx
```diff
@@ -72,9 +74,20 @@ const NetworkLeases: FC<Props> = ({ network, project }) => {
           "aria-label": "Hostname",
         },
         {
-          content: lease.address,
+          content:
+            typesWithNicStaticIPSupport.includes(network.type) &&
+            ["static", "dynamic"].includes(lease.type) ? (
+              <Link
+                title={`Edit ${lease.hostname} network`}
+                to={`/ui/project/${encodeURIComponent(project)}/instance/${encodeURIComponent(lease.hostname)}/configuration/network`}
```
edlerd:
```suggestion
                to={`/ui/project/${encodeURIComponent(lease.project)}/instance/${encodeURIComponent(lease.hostname)}/configuration/network`}
```
The `lease.project` could differ from current project.

src/pages/networks/NetworkLeases.tsx
```diff
@@ -72,9 +74,20 @@ const NetworkLeases: FC<Props> = ({ network, project }) => {
...
+              >
+                {lease.address}
```
edlerd: We shouldn't link on the ip address, better use the pencil icon next to the ip. Also might be worth extracting the pencil button and link building into a dedicated component to keep this component small.

src/pages/networks/NetworkIPAM.tsx
```diff
@@ -56,7 +66,9 @@ const NetworkIPAM: FC = () => {
   ];
 
   const rows = allocations.map((allocation) => {
-    const usedByItems = filterUsedByType(allocation.type, [allocation.used_by]);
+    const usedByItem = filterUsedByType(allocation.type, [
+      allocation.used_by,
+    ])[0];
```
edlerd: I think this can break when the used by entry returned by `filterUsedByType`. Then the `[0]` access will be a null pointer. Maybe make this a bit more defensive by also handling that the returned array from `filterUsedByType` can be empty?

src/pages/networks/NetworkIPAM.tsx
```diff
@@ -81,24 +93,30 @@ const NetworkIPAM: FC = () => {
...
+          content:
+            allocation.type === "instance" &&
+            typesWithNicStaticIPSupport.includes(
+              getNetworkTypeFromAllocation(allocation) ?? "",
+            ) ? (
+              <NetworkLink
+                address={allocation.addresses}
+                instanceName={usedByItem.name}
+                projectName={project}
+              ></NetworkLink>
```
edlerd:
```suggestion
              <NetworkLink
                address={allocation.addresses}
                instanceName={usedByItem.name}
                projectName={project}
              />
```
A bit simpler

src/pages/networks/NetworkLeases.tsx
```diff
@@ -72,9 +74,19 @@ const NetworkLeases: FC<Props> = ({ network, project }) => {
...
         {
-          content: lease.address,
+          content:
+            typesWithNicStaticIPSupport.includes(network.type) &&
+            ["static", "dynamic"].includes(lease.type) ? (
+              <NetworkLink
+                address={lease.address}
+                instanceName={lease.hostname}
+                projectName={project ?? "default"}
+              ></NetworkLink>
```
edlerd:
```suggestion
              <NetworkLink
                address={lease.address}
                instanceName={lease.hostname}
                projectName={lease.project ?? "default"}
              />
```
A bit simpler and this should be the lease project.

src/components/NetworkLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+      `/ui/project/${encodeURIComponent(projectName)}/instance/${encodeURIComponent(instanceName)}/configuration/network`,
+    );
+  };
+
+  return (
+    <div className="network-link">
+      <div className="u-truncate network-link-address" title={address}>
+        {address}
+      </div>
+      <Button
```
edlerd: Better use the `Link` component, so it can also be opened in a new tab. You can apply the `p-button-base` class to style it like a button.

src/components/NetworkLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Button, Icon } from "@canonical/react-components";
+import { useNavigate } from "react-router-dom";
+
+interface Props {
+  address: string;
+  instanceName: string;
+  projectName: string;
+}
+
+export const NetworkLink: FC<Props> = ({
```
edlerd: I am not 100% sure about the layout. Maybe the pencil icon should be left aligned before the address? Or maybe not pushed to the right-hand side of the column, but appear right next to the IP if the column fits the whole IP? Can you reach out to design to get an informed opinion on this? Personally not sure, might be also be good as is.

Another thought: Maybe moving the IP address column all the way to the right will solve the design challenge here. Because usually we have the actions on the right-hand side of the row and having the IP as last column would be aligning with the other pages.


image

edlerd (commented):
Nice progress :+1:

Some further thoughts and comments below. This should be good to merge very soon.

src/components/NetworkLink.tsx
edlerd: Please also apply the `is-dense` class, so the rows have equal height between the ones with and the ones without the button.

src/components/NetworkLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+  projectName,
+}) => {
+  return (
+    <div className="network-link-content">
+      <div className="u-truncate network-link-address" title={address}>
+        {address}
+      </div>
+      <Link
+        to={`/ui/project/${encodeURIComponent(projectName)}/instance/${encodeURIComponent(instanceName)}/configuration/network`}
+        target="_blank"
```
edlerd:
```suggestion
```
Sorry, this was a misunderstanding. I meant we should render this as a link, so the user has the option to also open it in another tab. Default for the link should still be same tab though.

src/components/InstanceIpEdit.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <div className="network-link-content">
+      <div className="u-truncate network-link-address" title={address}>
+        {address}
+      </div>
+      <Link
+        to={`/ui/project/${encodeURIComponent(projectName)}/instance/${encodeURIComponent(instanceName)}/configuration/network`}
+        target="_blank"
+        rel="noopener noreferrer"
+        className="p-button-base network-link is-dense"
+        type="button"
```
edlerd, line 25:
```suggestion
        type="button"
        title={`Edit network configuration for instance ${instanceName}`}
```
Let's add the title back to give a bit more guidance on what this does.

src/pages/networks/NetworkIPAM.tsx
```diff
@@ -123,14 +125,32 @@ const NetworkIPAM: FC = () => {
...
+          content:
+            allocation.type === "instance" &&
+            usedByItem &&
+            typesWithNicStaticIPSupport.includes(
+              getNetworkTypeFromAllocation(allocation) ?? "",
+            ) ? (
+              <NetworkLink
+                address={allocation.addresses}
+                instanceName={usedByItem.name}
+                projectName={project}
```
edlerd:
```suggestion
                projectName={usedByItem.project}
```
This can be different than the active project.

src/components/NetworkLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Icon } from "@canonical/react-components";
+import { Link } from "react-router-dom";
+
+interface Props {
+  address: string;
+  instanceName: string;
+  projectName: string;
+}
+
+export const NetworkLink: FC<Props> = ({
```
edlerd: The name is slightly imprecise, this is not a network link. Maybe `InstanceAddressEdit` or `InstanceIpEdit`?

src/sass/_instance_ip_edit.scss
edlerd, line 1: Let's align this with a better name for the component.

src/pages/networks/NetworkIPAM.tsx
omarelkashef: But we want to navigate to the instance (usedByItem) so we will need its project not the active project (if they are different). Or am I misunderstanding something?

edlerd: Yes, we should use the `usedByItem.project` as that is the project of the instance. The `project` variable refers to the current project, which might differ from the instance project.

omarelkashef: Oh I misread the suggestion and thought you suggested the opposite.

PR #1661, opened 2025-12-02
feat(device network configuration): Move edit mode in side panel

kimanhou:
## Done

- NIC device configuration: move edit and create in a side panel
- Show blue dot on modified but unsaved rows

## QA

1. Run the LXD-UI:
    - Inherited
    - Local

## Screenshots

image


src/components/forms/NetworkDevicesForm/edit/EditNetworkDevicePanel.tsx
```diff
@@ -0,0 +1,288 @@
...
+        <Button appearance="base" onClick={handleCancel} type="button">
+          Cancel
+        </Button>
+        <ActionButton
+          appearance="positive"
+          loading={formik.isSubmitting}
+          disabled={
+            isCreatingOverride
+              ? !formik.isValid
+              : !formik.isValid || !formik.dirty
```
edlerd:
```suggestion
!formik.isValid
```
I think we can simplify this. Currently, there is an edge case where I apply a new profile that inherits a nic, then create an override without changes -- I can't save it at first, because it is not dirty. I think we shouldn't block it.

src/components/forms/NetworkDevicesForm/edit/EditNetworkDevicePanel.tsx
```diff
@@ -0,0 +1,288 @@
...
+          }
+          onClick={() => {
+            formik.handleSubmit();
+          }}
+          type="button"
+        >
+          {isCreatingOverride ? "Create override" : "Apply changes"}
```
edlerd:
```suggestion
          "Apply changes"
```
I think in the mental model of the user this is the same action, and we should maybe present it equally as an edit nic device with "Apply changes". Wdyt?

src/components/forms/NetworkDevicesForm/edit/EditNetworkDevicePanel.tsx
```diff
@@ -0,0 +1,288 @@
...
+    formik.resetForm();
+    notify.clear();
+    onClose();
+  };
+
+  return (
+    <SidePanel>
+      <SidePanel.Header>
+        <SidePanel.HeaderTitle>
+          {isCreatingOverride ? "Create override" : "Edit network device"}
```
edlerd:
```suggestion
          "Edit network device"
```
I think for the easier UX point of view, this can be equal. Wdyt?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
edlerd, line 1: There is a lot of duplication between create and edit network device. I wonder if they should use the same component, and we remove one of them. So this could become a `NetworkDevicePanel`. Wdyt?

src/pages/instances/EditInstance.tsx
```diff
@@ -311,6 +321,92 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
+            const deviceIndex = formik.values.devices.length - 1;
+            const focusField = `devices.${deviceIndex}.name`;
+            setTimeout(() => {
+              const element = document.querySelector(
+                `[name="${focusField}"]`,
+              ) as HTMLInputElement;
+              if (element) {
+                element.focus();
+              }
+            }, 100);
```
edlerd: Not sure what this is doing, it seems to be broken?

src/pages/profiles/EditProfile.tsx
```diff
@@ -305,6 +313,81 @@ const EditProfile: FC<Props> = ({ profile }) => {
...
+            }, 100);
+          }}
+        />
+      )}
+
+      {panelParams.panel === panels.editNetworkDevice &&
+        panelParams.deviceName &&
+        (() => {
+          const urlParams = new URLSearchParams(window.location.search);
+          const createOverride = urlParams.get("create-override") === "true";
```
edlerd: Do we need this explicitly in the url? I think this can be implicit based on the device name. Wdyt?

src/pages/profiles/CreateProfile.tsx
```diff
@@ -299,6 +303,76 @@ const CreateProfile: FC = () => {
...
+            <EditNetworkDevicePanel
+              device={device as FormNetworkDevice}
+              deviceIndex={deviceIndex}
+              project={project}
+              formik={formik}
+              onClose={panelParams.clear}
+              onSave={panelParams.clear}
+            />
+          ) : null;
+        })()}
```
edlerd: This should be using the same component between edit/create instance and profile, not duplicating the logic for all four cases.

src/components/forms/NetworkDevicesForm/edit/CreateNetworkDevicePanel.tsx
```diff
@@ -0,0 +1,244 @@
...
+        network: values.network,
+        "security.acls":
+          userSelectedAcls.length > 0 ? userSelectedAcls.join(",") : undefined,
+        "ipv4.address": values.ipv4 || undefined,
+        "ipv6.address": values.ipv6 || undefined,
+      };
+
+      const newDevices = [...parentFormik.values.devices, newDevice];
+
+      await parentFormik.setFieldValue("devices", newDevices);
```
edlerd: I think this component shouldn't handle the parentFormik updates, we can do that in the caller, wdyt?

src/components/forms/NetworkDevicesForm/edit/EditNetworkDevicePanel.tsx
kimanhou: I think it's nice for the user to know what they are doing.
This is what I have for now:
```
  const getTitle = () => {
    if (mode === "create") return "Create network device";
    if (isCreatingOverride) return "Create override";
    if (isEditingOverride) return "Edit override";
    return "Edit network device";
  };
  ```

File: src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
Comment (kimanhou) [reply src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx:1]: Yes, I started last night !

File: src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx [hunk 8]
```diff
@@ -0,0 +1,337 @@
...
+                } else {
+                  void formik.setFieldValue("acls", newNetworkAcls.join(","));
+                }
+
+                if (!supportsStaticIPs(selectedNetwork)) {
+                  void formik.setFieldValue("ipv4", "");
+                  void formik.setFieldValue("ipv6", "");
+                }
+              }}
+              managedNetworks={managedNetworks}
```
edlerd: The prop name was changed in another pr to `networkList`. Please adjust to fix the side panel.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -29,20 +30,38 @@ export const getInheritedNetworkRow = ({
...
+        )}
+        <b
+          className={classnames({
+            "u-text--muted": isDetached,
+            "u-text--line-through": isDetached,
+          })}
+        >
+          {device.key}
+        </b>
+      </div>
```
edlerd: I think this can be a component receiving the `hasChanges` and `key` as a prop.

I think we should remove the empty `network-device-spacer`, and reserve the space with CSS. An empty element for layout seems semantically wrong.

src/components/forms/NetworkDevicesForm/InheritedNetworkRow.tsx
```diff
@@ -29,20 +30,38 @@ export const getInheritedNetworkRow = ({
...
-        })}
-      >
-        {device.key}
-      </b>
+      <div className="network-device-name">
+        {hasChanges ? (
+          <span
+            className="network-device-change-indicator"
+            title="This inherited network has been modified"
+          ></span>
```
edlerd:
```suggestion
          <span
            className="network-device-change-indicator"
            title="This inherited network has been modified"
          />
```
This can be self closing.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -0,0 +1,54 @@
...
+  network?: LxdNetwork;
+  family: "IPv4" | "IPv6";
+}
+
+const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
+  if (!network || !network.config) {
+    return null;
+  }
+
+  const addressString = family === "IPv4" ? "ipv4.address" : "ipv6.address";
```
edlerd:
```suggestion
  const networkFieldName = family === "IPv4" ? "ipv4.address" : "ipv6.address";
```
Better name?

edlerd: Also we should declare a `const foo = network.config?.[addressString]` to simplify below.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+  const inheritedNetworkNames = inheritedNetworks.map((item) => item.key);
+
+  const isOverride = isCreatingOverride || isEditingOverride;
+
+  const getInvalidNames = () => {
+    if (isCreatingLocal) return existingDeviceNames;
+    if (isCreatingOverride)
+      return inheritedNetworkNames.filter(
+        (t) => t !== props.inheritedDevice?.key,
+      );
```
edlerd: If we create an override, there is only one valid name which is the inherited device name.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+    return existingDeviceNames.filter((name) => name !== props.device?.name);
```
edlerd: I think this generalizes with `isCreatingLocal` case, as in that case `props.device?.name` would be empty and we would get exactly the same whole list of `existingDeviceNames`. Meaning we can probably remove the special case for it above.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+        name: deviceName,
+        network: defaultNetworkName,
+        acls: defaultAcls,
+        ipv4: "",
+        ipv6: "",
+      };
+    } else if (isCreatingOverride) {
+      return {
+        name: props.inheritedDevice?.key || "",
+        network: defaultNetworkName,
```
edlerd: I think this should be the inherited device network name.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+        acls: defaultAcls,
```
edlerd: This should be the inherited device acls.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+        ...props.networkDeviceNames,
+        ...inheritedNetworkNames,
+      ];
+
+      const deviceName = deduplicateName("eth", 1, allNetworkNames);
+
+      return {
+        name: deviceName,
+        network: defaultNetworkName,
+        acls: defaultAcls,
```
edlerd, line 130: This should be an empty list.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+        name: props.inheritedDevice?.key || "",
+        network: defaultNetworkName,
+        acls: defaultAcls,
+        ipv4: "",
+        ipv6: "",
+      };
+    } else {
+      if (!props.device || isNoneDevice(props.device)) {
+        return { name: "", network: "", acls: "", ipv4: "", ipv6: "" };
+      }
```
edlerd: What is this special case about? I think this can be removed.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+  const selectedAcls = Array.from(
+    new Set(networkAcls.concat(userSelectedAcls)),
+  );
+
+  const getAclHelperText = (): ReactNode => {
+    if (!supportsNicDeviceAcls(selectedNetwork))
+      return "Network must be of type OVN to customize ACLs.";
+    return networkAcls.length > 0
+      ? "Some ACLs are inherited from the network. They cannot be deselected here."
+      : undefined;
```
edlerd:
```suggestion
    if (supportsNicDeviceAcls(selectedNetwork)) {
    return networkAcls.length > 0
      ? "Some ACLs are inherited from the network. They cannot be deselected here."
      : undefined;
    } 
    return "Network must be of type OVN to customize ACLs.";
```
Please use positive conditions and always use brackets to avoid bugs.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+        >
+          <Form onSubmit={formik.handleSubmit} stacked>
+            <Input
+              label="Device name"
+              required
+              id="name"
+              name="name"
+              value={formik.values.name}
+              onChange={formik.handleChange}
+              onBlur={formik.handleBlur}
```
edlerd:
```suggestion
              {...formik.getFieldProps("name")}
```
I think all of these can be simplified by the formik helper.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,337 @@
...
+        <Button appearance="base" onClick={handleCancel} type="button">
+          Cancel
+        </Button>
+        <ActionButton
+          appearance="positive"
+          loading={formik.isSubmitting}
+          disabled={!formik.isValid}
+          onClick={() => {
+            formik.submitForm();
+          }}
```
edlerd:
```suggestion
            onClick={() => void formik.submitForm()}
```
A bit shorter?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanelContainer.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const inheritedDevice = inheritedNetworks.find((d) => d.key === deviceName);
+
+  const isCreateOverride = inheritedDevice && !device;
+  const isEditOverride = inheritedDevice && device;
+  const editMode = isCreateOverride
+    ? "create-override"
+    : isEditOverride
+      ? "edit-override"
+      : "edit-local";
```
edlerd: This should better be a callback with early exits. That should be easier to read.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceContent.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+  return (
+    <>
+      <div>Network</div>
+      <ResourceLink
+        type="network"
+        value={device.network}
+        to={`/ui/project/${encodeURIComponent(project ?? "")}/network/${encodeURIComponent(device.network)}`}
+      />
+      <ReadOnlyAclsList project={project} network={network} device={device} />
```
edlerd:
```suggestion
      <NetworkDeviceAclListRead project={project} network={network} device={device} />
```
A bit more consistent naming?

src/pages/instances/CreateInstance.tsx
```diff
@@ -578,6 +583,22 @@ const CreateInstance: FC = () => {
...
+            isEditNetworkDevice && panelParams.deviceName
+              ? panelParams.deviceName
+              : undefined
+          }
+          project={project}
+          formik={formik}
+          onClose={panelParams.clear}
+          onSave={panelParams.clear}
+        />
+      )}
```
edlerd, line 588: I think this should stay visible when clicking to another part of the form (say go from networks to Main configuration). currently it unmounts due to the url losing the panel parameters.

src/pages/profiles/CreateProfile.tsx
```diff
@@ -299,6 +304,22 @@ const CreateProfile: FC = () => {
...
+
+      {(panelParams.panel === panels.editNetworkDevice ||
+        panelParams.panel === panels.createNetworkDevice) && (
+        <NetworkDevicePanelContainer
+          mode={isEditNetworkDevice ? "edit" : "create"}
+          deviceName={
+            isEditNetworkDevice && panelParams.deviceName
+              ? panelParams.deviceName
+              : undefined
+          }
```
edlerd: I think this logic should move into the `NetworkDevicePanelContainer`, where we can do `const panelParams = usePanelParams();` to read it. No need to pass it in and duplicate this bit.

src/sass/_forms.scss
edlerd, line 1: I think there is plenty SCSS for the network device edit mode, that can be removed.

src/pages/instances/CreateInstance.tsx
kimanhou: Mmh I'm not convinced the side panel should stay open when you go to another part of the form. If the user is modifying something, then that something should be visible. And then what happens when the user clicks on `Apply changes` ?
All panels are linked to the URL ?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: It's a type check. I refactored it to throw an error.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: What if the selected network (here `defaultNetworkName`) has ACLs on it ? Shouldn't we display them and prevent the user to deselect them here ?

src/pages/instances/CreateInstance.tsx
edlerd: Huh, I am not sure on the right behaviour. On one hand it seems unexpected that the panel closes when navigating away. On the other hand submitting the panel on another page would not show the changes applied with the blue dot. Maybe current status quo of it closing is ok after all.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
edlerd: I confused how the `NetworkAclSelector` works. Currently, indeed, the `selectedAcls` must include the inherited ACLs. But then again, we already pass in the `inheritedAcls`. So we could infer that they are selected inside. I was expecting the `formik.values.acls` to contain only the local ACLs on the network device and not the ones from the network. Shall we change it?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
edlerd: Ah right, I think an error is better than silently handling it.

src/pages/profiles/CreateProfile.tsx
kimanhou: Yes ! And I can also remove `mode` here

src/sass/_forms.scss
kimanhou: Euh really ? I didn't find anything as the rules also apply for read mode.
However `cluster-group-form` is not found anywhere in the code. Should I remove it in this PR ?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: I don't want to change the logic of the `NetworkAclSelector`. In read mode, the ACLs are displayed as a flat list without any difference on which level they belong to, so I would like to keep a flat list of all ACLs everywhere, or at least where we can.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+    mode,
+    project,
+    onClose,
+    onSave,
+    existingDeviceNames,
+    inheritedNetworks,
+    shouldDisplayIpAddresses,
+  } = props;
+  const notify = useNotify();
+  const networkAclSelectorId = useId();
```
edlerd: I think we don't need this as there is only ever one acl input visible now we can use a static id.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+  const validationSchema = Yup.object({
+    name: isCreatingOverride
+      ? Yup.string().required("Device name is required")
+      : Yup.string()
+          .required("Device name is required")
+          .matches(/^[A-Za-z0-9/\-:_.]+$/, {
+            message:
+              "Name can only contain alphanumeric, forward slash, hyphen, colon, underscore and full stop characters",
+          })
```
edlerd: This validation is also relevant on creation.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+export interface NetworkDeviceFormValues {
+  name: string;
+  network: string;
+  acls?: string;
+  ipv4?: string;
+  ipv6?: string;
+}
+
+interface CreateProps {
+  mode: "create";
```
edlerd:
```suggestion
  mode: "create-local";
```
A bit more in line with the modes below?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+  inheritedDevice?: InheritedNetwork;
+  project: string;
+  onClose: () => void;
+  onSave: (updatedDevice: FormNetworkDevice) => void;
+  existingDeviceNames: string[];
+  inheritedNetworks: InheritedNetwork[];
+  shouldDisplayIpAddresses: boolean;
+}
+
+type Props = CreateProps | EditProps;
```
edlerd: I think it might be simpler to follow if we combine the two into one, wdyt?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+  const managedNetworks = networks.filter((network) => network.managed);
+
+  const isCreatingOverride = mode === "create-override";
+  const isCreatingLocal = mode === "create";
+  const isEditingOverride = mode === "edit-override";
+
+  const isOverride = isCreatingOverride || isEditingOverride;
+
+  const getInvalidNames = () => {
+    if (isCreatingOverride) {
```
edlerd:
```suggestion
    if (isCreatingOverride || isEditingOverride) {
```
Should we add a comment like `// Can't rename override devices, so the input is disabled without validation.` below?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+const NetworkDevicePanel: FC<Props> = (props) => {
+  const {
+    mode,
+    project,
+    onClose,
+    onSave,
+    existingDeviceNames,
+    inheritedNetworks,
+    shouldDisplayIpAddresses,
+  } = props;
```
edlerd:
```suggestion
const NetworkDevicePanel: FC<Props> = ({
    mode,
    project,
    onClose,
    onSave,
    existingDeviceNames,
    inheritedNetworks,
    shouldDisplayIpAddresses,
  }) => {
```
We can destruct them right above.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+          ),
+    network: Yup.string().required("Network is required"),
+  });
+
+  const getInitialValues = (): NetworkDeviceFormValues => {
+    const defaultNetworkName = managedNetworks[0]?.name ?? "";
+    const defaultAcls = getNetworkAcls(managedNetworks[0]).join(",");
+
+    if (isCreatingLocal) {
+      const { networkDeviceNames } = props;
```
edlerd:
```suggestion
```
I think we should destructure this in the component header on line 65.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+        name: deviceName,
+        network: defaultNetworkName,
+        acls: defaultAcls,
+        ipv4: "",
+        ipv6: "",
+      };
+    }
+
+    if (isCreatingOverride) {
+      const { inheritedDevice } = props;
```
edlerd:
```suggestion
```

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+          getNetworkAcls(network),
+          getDeviceAcls(inheritedDevice.network),
+        ).join(","),
+        ipv4: "",
+        ipv6: "",
+      };
+    }
+
+    // Edit modes: edit-override or edit-local
+    const { device } = props;
```
edlerd:
```suggestion
```

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+    },
+  });
+
+  const selectedNetwork = managedNetworks.find(
+    (n) => n.name === formik.values.network,
+  );
+  const networkAcls = getNetworkAcls(selectedNetwork);
+  const userSelectedAcls = formik.values.acls
+    ? formik.values.acls.split(",").filter(Boolean)
+    : [];
```
edlerd: this is confusing, the same name as line 190, but semantically different.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+            <NetworkSelector
+              value={formik.values.network}
+              setValue={(value) => {
+                void formik.setFieldValue("network", value);
+
+                const selectedNetwork = managedNetworks.find(
+                  (n) => n.name === value,
+                );
+                const newNetworkAcls = getNetworkAcls(selectedNetwork);
```
edlerd: I think this can move into the else branch, to define it closer to its usage.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+                if (!supportsNicDeviceAcls(selectedNetwork)) {
+                  void formik.setFieldValue("acls", "");
+                } else {
+                  void formik.setFieldValue("acls", newNetworkAcls.join(","));
+                }
+
+                if (!supportsStaticIPs(selectedNetwork)) {
+                  void formik.setFieldValue("ipv4", "");
+                  void formik.setFieldValue("ipv6", "");
+                }
```
edlerd:
```suggestion
                  void formik.setFieldValue("ipv4", "");
                  void formik.setFieldValue("ipv6", "");
```
I think we should drop the condition here, and we should always reset the addresses, otherwise we might set them on a network that doesn't support it.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,351 @@
...
+            />
+
+            <label
+              className={classnames({
+                "u-text--muted": !supportsNicDeviceAcls(selectedNetwork),
+              })}
+              htmlFor={networkAclSelectorId}
+            >
+              ACLs
+            </label>
```
edlerd: This should move into the `NetworkAclSelector` to also disable on the same conditions as the input. We currently duplicate the disable logic, and it is out of sync. Internally, the network acl selector also disables when `hasAcls` is `false`.

We can give the `NetworkAclSelector` a `label` prop, which we set to `ACLs` from here.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanelContainer.tsx
```diff
@@ -0,0 +1,109 @@
...
+    | LxdNicDevice
+    | LxdNoneDevice;
+
+  const inheritedDevice = inheritedNetworks.find((d) => d.key === deviceName);
+
+  const getMode = ():
+    | "create"
+    | "create-override"
+    | "edit-override"
+    | "edit-local" => {
```
edlerd: Best create a union type for this to reuse it here.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanelContainer.tsx
```diff
@@ -0,0 +1,109 @@
...
+          formik.setFieldValue("devices", newDevices);
+          onSave();
+        }}
+        existingDeviceNames={existingDeviceNames}
+        inheritedNetworks={inheritedNetworks}
+        networkDeviceNames={networkDeviceNames}
+        shouldDisplayIpAddresses={shouldDisplayIpAddresses}
+      />
+    );
+  }
```
edlerd: Shall we unify this with the general case below? Seems a bit simpler and less duplication.

src/pages/instances/CreateInstance.tsx
```diff
@@ -578,6 +581,16 @@ const CreateInstance: FC = () => {
...
         )}
       </FormFooterLayout>
+
+      {(panelParams.panel === panels.editNetworkDevice ||
+        panelParams.panel === panels.createNetworkDevice) && (
+        <NetworkDevicePanelContainer
+          project={project}
+          formik={formik}
+          onClose={panelParams.clear}
+          onSave={panelParams.clear}
```
edlerd: Can we move those two calls inside and remove them here and in the other usages of the `NetworkDevicePanelContainer`.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanelContainer.tsx
```diff
@@ -0,0 +1,109 @@
...
+import usePanelParams, { panels } from "util/usePanelParams";
+
+interface Props {
+  project: string;
+  formik: InstanceAndProfileFormikProps;
+  onClose: () => void;
+  onSave: () => void;
+}
+
+const NetworkDevicePanelContainer: FC<Props> = (props) => {
```
edlerd: The naming of this one is a bit strange, maybe this should be included in the `NetworkDevicePanel` and we remove this component?

src/components/forms/NetworkDevicesForm/read/NetworkDeviceName.tsx
```diff
@@ -0,0 +1,39 @@
...
+}) => {
+  const LabelTag = isInherited ? "b" : "span";
+
+  return (
+    <div className="network-device-name">
+      {hasChanges && (
+        <span
+          className="network-device-change-indicator"
+          title="This device has unsaved changes"
+        />
```
edlerd: Can we reuse the `<Icon name="status-in-progress-small"` and not recreate the blue dot in css?

src/components/forms/NetworkDevicesForm/read/NetworkDeviceName.tsx
```diff
@@ -0,0 +1,39 @@
...
+  isInherited?: boolean;
+}
+
+const NetworkDeviceName: FC<Props> = ({
+  name,
+  hasChanges,
+  isDetached,
+  isInherited,
+}) => {
+  const LabelTag = isInherited ? "b" : "span";
```
edlerd: This seems like an oversight in the previous implementation. I think we should always use bold font for the device name here.

src/sass/_forms.scss
edlerd: Ah yes, please remove the `cluster-group-form`.

I think the `ip-content-edit` class and usage can be removed now. But you are right, there seems to be not much more unused CSS now -- I thought there would be more.

kimanhou: Deleted `cluster-group-form` and removed the div with class `ip-content-edit`

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: When you say creation, do you mean creation of an override or creation of a local device ?
This is currently applied for all cases except creation of an override because when creating an override, the name comes from the inherited network and can't be changed.

Actually maybe we should not have any validation when creating an override. 🤔 Wdyt ?

edlerd: Oh right, I think I misread the code here. Ignore this comment, this seems good as is.

edlerd: Yes, maybe the condition should be base on "is this an override" and skip name validation completely in that case.

src/pages/instances/CreateInstance.tsx
kimanhou: `onClose` yes.
`onSave` has `ensureEditMode(formik);` in EditInstance and EditProfile and not in CreateInstance and CreateProfile. But it looks like it has no effect if we add that to the latter.

edlerd: I think we should only move the `panelParams.clear` in and keep the callbacks, but optional. Then in the edit cases, we do the `ensureEditMode` with the callback. Here on creation, we can not pass the callback.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: I'm not sure I understood everything in this comment but I did something.

src/pages/networks/forms/NetworkAclSelector.tsx
```diff
@@ -47,25 +50,37 @@ const NetworkAclSelector: FC<Props> = ({
...
   return (
-    <MultiSelect
-      items={toOptionList(
-        availableAcls.map((acl) => acl.name),
-        inheritedAcls,
+    <>
+      {label && (
+        <label
+          className={classnames({
+            "u-text--muted": !canSelectManualAcls,
```
edlerd:
```suggestion
            "u-text--muted": !hasAcls || !canSelectManualAcls,
```
Shouldn't this be the same condition as the disabled below?

Then we should probably extract it into a const to reuse here and below.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -0,0 +1,53 @@
...
+      onChange={(e) => {
+        void formik.setFieldValue(formikFieldName, e.target.value);
+      }}
+      onBlur={() => {
+        void formik.setFieldTouched(formikFieldName, true);
+      }}
+      disabled={
+        !network?.managed ||
+        !typesWithNicStaticIPSupport.includes(network.type) ||
+        ipAddressConfigValue === "none"
```
edlerd: Better extract a `const isDisabled` for this value and reuse it in the `help` condition below.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+
+const NetworkDevicePanel: FC<Props> = ({
+  project,
+  formik: parentFormik,
+  onSave,
+}) => {
+  const notify = useNotify();
+  const panelParams = usePanelParams();
+  const { data: profiles = [] } = useProfiles(project);
+
```
edlerd:
```suggestion
```

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+      // Can't rename override devices, so the input is disabled without validation.
+      return [];
+    }
+
+    if (isCreatingLocal(mode)) {
+      return existingDeviceNames;
+    }
+
+    return existingDeviceNames.filter((name) => name !== device?.name);
+  };
```
edlerd: I think this whole function can be removed, as we call it only without override. The last line `existingDeviceNames.filter((name) => name !== device?.name)` generalizes to the remaining two lines. So we can inline that directly. Wdyt?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+        .notOneOf(getInvalidNames(), "A device with this name already exists"),
+    }),
+    network: Yup.string().required("Network is required"),
+  });
+
+  const getInitialValues = (): NetworkDeviceFormValues => {
+    const defaultNetworkName = managedNetworks[0]?.name ?? "";
+    const defaultAcls = getNetworkAcls(managedNetworks[0]).join(",");
+
+    if (isCreatingLocal(mode) && networkDeviceNames) {
```
edlerd:
```suggestion
    if (isCreatingLocal(mode)) {
```
Why this 2nd condition? I think it can empty array so falsy, and we'd still want to hit the conditional branch.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+        acls: defaultAcls,
+        ipv4: "",
+        ipv6: "",
+      };
+    }
+
+    if (isCreatingOverride(mode)) {
+      if (!inheritedDevice) {
+        throw new Error("Cannot create override: inheritedDevice is required");
+      }
```
edlerd: I think this can throw when `useProfiles` is very slow, while other queries satisfy quickly. Maybe we can handle the loading mode more gracefully?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+    validationSchema,
+    onSubmit: (values) => {
+      const allSelectedAcls = values.acls
+        ? values.acls.split(",").filter(Boolean)
+        : [];
+      const userSelectedAcls = allSelectedAcls.filter(
+        (acl) =>
+          !getNetworkAcls(
+            managedNetworks.find((n) => n.name === values.network),
+          ).includes(acl),
```
edlerd: We should extract the result of `getNetworkAcls` into a `const` outside the loop. To avoid calling it on every iteration.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+        );
+        let updatedDevices: FormNetworkDevice[];
+
+        if (existingDeviceIndex !== -1) {
+          updatedDevices = [...devices] as FormNetworkDevice[];
+          updatedDevices[existingDeviceIndex] = deviceData;
+        } else {
+          updatedDevices = (devices as FormNetworkDevice[])
+            .filter((d) => d.name !== originalDeviceName)
+            .concat(deviceData);
```
edlerd: I think this moves the device to the back of the list on rename, we should have a stable ordering.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        parentFormik.setFieldValue("devices", updatedDevices);
+      }
+
+      if (onSave) {
+        onSave();
+      }
```
edlerd:
```suggestion
        onSave?.();
```
A bit more concise

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,375 @@
...
+  });
+
+  const selectedNetwork = managedNetworks.find(
+    (n) => n.name === formik.values.network,
+  );
+  const networkAcls = getNetworkAcls(selectedNetwork);
+  const allSelectedAcls = formik.values.acls
+    ? formik.values.acls.split(",").filter(Boolean)
+    : [];
+  const selectedAcls = Array.from(new Set(networkAcls.concat(allSelectedAcls)));
```
edlerd, line 222: Similar calculation is done for the initial values and in the network selector on change handler. I wonder if we can extract some common functionality here.

src/sass/_forms.scss
```diff
@@ -546,3 +537,15 @@ input[type="time"]::-webkit-calendar-picker-indicator {
...
 }
+
+.network-device-name {
+  align-items: center;
+  display: flex;
+  gap: $sph--small;
+
+  .network-device-label {
+    &:not(.has-changes) {
+      margin-left: calc(16px + $sph--small);
```
edlerd: Where does the magic 16 come from? If it is the icon size, this changes on very large screens. I wonder if there is a css variable we could use here instead?

src/types/device.d.ts
```diff
@@ -1,3 +1,9 @@
+export type NetworkDeviceMode =
+  | "create-local"
+  | "create-override"
+  | "edit-override"
+  | "edit-local";
+
```
edlerd: This file is for types relating to the LXD api, I'd suggest keeping this type in `NetworkDevicePanel`, together with the `isCreatingOverride` and other helpers.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: I created a function that gets the ACLs from the network name

```
const getNetworkAclsByNetworkName = (networkName: string) => {
    const network = managedNetworks.find((n) => n.name === networkName);
    return getNetworkAcls(network);
  };
```

src/sass/_forms.scss
kimanhou: Yes, it's the icon size. I put `1rem` instead because I couldn't find the icon size.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: It's a type check. I refactored it.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+      return "edit-override";
+    }
+    return "edit-local";
+  };
+
+  const mode = getMode();
+
+  const isCreatingOverride = mode === "create-override";
+  const isCreatingLocal = mode === "create-local";
+  const isEditingOverride = mode === "edit-override";
```
edlerd: It seems we only use the mode here to evaluate the booleans. Maybe we can remove the mode and do the evaluation directly. Wdyt?

```suggestion
  const isCreatingLocal = panelParams?.panel === panels.createNetworkDevice;
  const isEditingLocal = !isCreatingLocal && !inheritedDevice;
  const isCreatingOverride = inheritedDevice && !device;
  const isEditingOverride = inheritedDevice && device;
```

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+      <SidePanel>
+        <SidePanel.Header>
+          <SidePanel.HeaderTitle>{getTitle()}</SidePanel.HeaderTitle>
+        </SidePanel.Header>
+        <SidePanel.Content className="u-no-padding">
+          <Spinner className="u-loader" text="Loading..." />
+        </SidePanel.Content>
+      </SidePanel>
+    );
+  }
```
edlerd: This will cause an error on loading, because formik below is a hook that will be conditionally rendered.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+  const { data: networks = [] } = useNetworks(project);
+
+  const managedNetworks = networks.filter((network) => network.managed);
+
+  const existingDeviceNames = getExistingDeviceNames(
+    parentFormik.values,
+    profiles,
+  );
+  const inheritedNetworks = getInheritedNetworks(parentFormik.values, profiles);
+  const shouldDisplayIpAddresses =
```
edlerd:
```suggestion
  const isInstance =
```
Maybe a bit clearer?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+        ipv6: "",
+      };
+    }
+
+    // Edit modes: edit-override or edit-local
+    if (!device || isNoneDevice(device)) {
+      throw new Error(
+        "Cannot edit network device: device is missing or is of type 'none'",
+      );
+    }
```
edlerd:
```suggestion
    if (!device || isNoneDevice(device)) {
      throw new Error(
        "Cannot edit network device: device is missing or is of type 'none'",
      );
    }
```
You might want to raise this to very early in the component to guard the types also for other use cases.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+    if (isCreatingOverride) {
+      if (!inheritedDevice) {
+        return {
+          name: deviceName || "eth0",
+          network: defaultNetworkName,
+          acls: defaultAcls,
+          ipv4: "",
+          ipv6: "",
+        };
+      }
```
edlerd:
```suggestion
```
This combination should be impossible.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+      ipv6: device["ipv6.address"] || "",
+    };
+  };
+
+  const formik = useFormik<NetworkDeviceFormValues>({
+    initialValues: getInitialValues(),
+    validationSchema,
+    onSubmit: (values) => {
+      const allSelectedAcls = values.acls
+        ? values.acls.split(",").filter(Boolean)
```
edlerd:
```suggestion
        ? values.acls.split(",")
```
The filter seems redundant. ACLs can't be empty.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+        );
+        let updatedDevices: FormNetworkDevice[];
+
+        if (existingDeviceIndex !== -1) {
+          updatedDevices = [...devices] as FormNetworkDevice[];
+          updatedDevices[existingDeviceIndex] = deviceData;
+        } else {
+          const originalIndex = devices.findIndex(
+            (d) => d.name === originalDeviceName,
+          );
```
edlerd: This duplicates `existingDeviceIndex`.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+          if (originalIndex !== -1) {
+            updatedDevices = [...devices] as FormNetworkDevice[];
+            updatedDevices[originalIndex] = deviceData;
+          } else {
```
edlerd: This case is already handled above.
```suggestion
```

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+      panelParams.clear();
+    },
+  });
+
+  const selectedNetwork = managedNetworks.find(
+    (n) => n.name === formik.values.network,
+  );
+  const networkAcls = getNetworkAclsByNetworkName(formik.values.network);
+  const allSelectedAcls = formik.values.acls
+    ? formik.values.acls.split(",").filter(Boolean)
```
edlerd:
```suggestion
    ? formik.values.acls.split(",")
```
This is duplicating the logic above.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,392 @@
...
+              value={formik.values.network}
+              setValue={(value) => {
+                void formik.setFieldValue("network", value);
+
+                if (!supportsNicDeviceAcls(selectedNetwork)) {
+                  void formik.setFieldValue("acls", "");
+                } else {
+                  const newNetworkAcls = getNetworkAclsByNetworkName(value);
+                  void formik.setFieldValue("acls", newNetworkAcls.join(","));
+                }
```
edlerd:
```suggestion
                  const newNetworkAcls = getNetworkAclsByNetworkName(value);
                  void formik.setFieldValue("acls", newNetworkAcls.join(","));
```
If we make the functions handle the empty case, then we can set acls to empty string and this generalizes.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: `isEditingLocal` is not needed

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: `getNetworkAclsByNetworkName` already handles the case where the network name is not part of managedNetworks and the case where the managedNetwork doesn't have ACLs. It won't check the type of the network to know if the network supports IP addresses.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,350 @@
...
+        acls: combineAcls(
+          getNetworkAclsByNetworkName(inheritedDevice.network?.name || ""),
+          getDeviceAcls(inheritedDevice.network),
+        ).join(","),
+        ipv4: "",
+        ipv6: "",
+      };
+    }
+
+    const nicDevice = device as LxdNicDevice;
```
edlerd:
```suggestion
```
The casting here seems invalid, device can also be undefined. I think above we should define device as `LxdNicDevice | undefined`. This cast currently masks an error (see below).

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,350 @@
...
+    const deviceNetworkName = nicDevice.network || defaultNetworkName;
```
edlerd: There is now again the problem with the devices loading slowly. On the demo server, I could surface an error when reloading on [this url](https://lxd-ui-1661.demos.haus/ui/project/default/instance/DemoInstance/configuration/network?panel=edit-network-device&device-name=eth-1).

One idea is to set `enableReinitialize: true,` on formik and safeguard in the get initial values.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,350 @@
...
+          (d) => d.name === originalDeviceName,
+        );
+        let updatedDevices: FormNetworkDevice[];
+
+        if (existingDeviceIndex !== -1) {
+          updatedDevices = [...devices] as FormNetworkDevice[];
+          updatedDevices[existingDeviceIndex] = deviceData;
+        } else {
+          updatedDevices = [...devices, deviceData] as FormNetworkDevice[];
+        }
```
edlerd:
```suggestion
        let updatedDevices: FormDevice[];
        if (existingDeviceIndex !== -1) {
          updatedDevices = [...devices];
          updatedDevices[existingDeviceIndex] = deviceData;
        } else {
          updatedDevices = [...devices, deviceData];
        }
```
Let's use the right type for updated devices, then the other casts can be removed.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
edlerd:
```suggestion
    const deviceNetworkName = nicDevice.network;
```
The fallback shouldn't be necessary, this is the edit case. Then we can remove the variable and inline it below.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,350 @@
...
+      <SidePanel>
+        <SidePanel.Header>
+          <SidePanel.HeaderTitle>{getTitle()}</SidePanel.HeaderTitle>
+        </SidePanel.Header>
+        <SidePanel.Content className="u-no-padding">
+          <Spinner className="u-loader" text="Loading..." />
+        </SidePanel.Content>
+      </SidePanel>
+    );
+  }
```
edlerd:
```suggestion
```
Probably better to render the form right away, but replace the values as they get loaded.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: TypeScript still wants to handle the case where `nicDevice` (renamed `device`) is undefined.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,334 @@
...
+  });
+
+  if (
+    (isEditingLocal || isEditingOverride) &&
+    (!device || isNoneDevice(device))
+  ) {
+    throw new Error(
+      "Cannot edit network device: device is missing or is of type 'none'",
+    );
+  }
```
edlerd:
```suggestion
  if (isNoneDevice(device)) {
    throw new Error(
      "Cannot edit network device of type 'none'",
    );
  }
```
I think this is the only case that can happen here based on definition of the variables.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: [this url](https://lxd-ui-1661.demos.haus/ui/project/default/instance/DemoInstance/configuration/network?panel=edit-network-device&device-name=eth-1) works now

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+        parentFormik.setFieldValue("devices", updatedDevices);
+      }
+
+      onSave?.();
+      panelParams.clear();
+    },
+  });
+
+  if ((isEditingLocal || isEditingOverride) && !device) {
```
edlerd: this is not a possible combination, shouldn't this only check for is none device?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,327 @@
...
+        let updatedDevices: FormDevice[];
+        if (existingDeviceIndex !== -1) {
+          updatedDevices = [...devices];
+          updatedDevices[existingDeviceIndex] = deviceData;
+        } else {
+          updatedDevices = [...devices, deviceData];
+        }
+
+        parentFormik.setFieldValue("devices", updatedDevices);
+      }
```
edlerd:
```suggestion
        const originalDeviceName = deviceName;
        const { devices } = parentFormik.values;
        const existingDeviceIndex = devices.findIndex(
          (d) => d.name === originalDeviceName,
        );
        let updatedDevices: FormDevice[];
        if (existingDeviceIndex !== -1) {
          updatedDevices = [...devices];
          updatedDevices[existingDeviceIndex] = deviceData;
        } else {
          updatedDevices = [...devices, deviceData];
        }

        parentFormik.setFieldValue("devices", updatedDevices);
```
I think the creation case generalizes into the other ones.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,327 @@
...
+    profiles,
+  );
+  const inheritedNetworks = getInheritedNetworks(parentFormik.values, profiles);
+  const isInstance = parentFormik.values.entityType === "instance";
+  const deviceName = panelParams?.deviceName;
+  const device = parentFormik.values.devices.find(
+    (d) => d.name === deviceName,
+  ) as LxdNicDevice | undefined;
+
+  const inheritedDevice = inheritedNetworks.find((d) => d.key === deviceName);
```
edlerd:
```suggestion
  const isInstance = parentFormik.values.entityType === "instance";

  const deviceName = panelParams?.deviceName;
  const device = parentFormik.values.devices.find(
    (d) => d.name === deviceName,
  ) as LxdNicDevice | undefined;

  const inheritedNetworks = getInheritedNetworks(parentFormik.values, profiles);
  const inheritedDevice = inheritedNetworks.find((d) => d.key === deviceName);
```
A bit nicer grouping.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -0,0 +1,327 @@
...
+  const validationSchema = Yup.object({
+    ...(!isOverride && {
+      name: Yup.string()
+        .required("Device name is required")
+        .matches(/^[A-Za-z0-9/\-:_.]+$/, {
+          message:
+            "Name can only contain alphanumeric, forward slash, hyphen, colon, underscore and full stop characters",
+        })
+        .notOneOf(
+          existingDeviceNames.filter((name) => name !== device?.name),
```
edlerd: I think we might introduce a `const invalidDeviceNames` to reuse here. I suspect this re-evaluates the filter on each validation.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
kimanhou: We should never have a none device in the panel, since the Edit button is not displayed if the inherited network is detached. I refactored the cast on lines 66-68 to always get a nic device.

edlerd (approved):
Code and QA LGTM, thanks for improving this and the many great iterations to make it very clear and easy to maintain in the future :+1:

PR #1663, opened 2025-12-02
feat: [WD-30458] ACL Default Action

Kxiru:
## Done

- Add Default ACL Actions to Network configuration
- Config values successfully update

- [x] Fix: Once updated, the config needs to show the values.

## QA

1. Run the LXD-UI:
    - Navigate to Network > Any Bridge/OVN network > Click on edit ACL
    - Select an ACL if none is already selected.
    - View the defualt ACL action component enable,
    - Set default actions for both directions and then for single directions.
    - Ensure in the Inspector tools network tab that requests are being fulfilled successfully.

## Screenshots

image
image
image

### DIFF
--- a/src/pages/networks/forms/NetworkAcls.tsx
+++ b/src/pages/networks/forms/NetworkAcls.tsx
@@ -7,6 +7,8 @@ import FormEditButton from "components/FormEditButton";
 import { ensureEditMode } from "util/instanceEdit";
 import NetworkAclSelector from "pages/networks/forms/NetworkAclSelector";
 import { Label } from "@canonical/react-components";
+import NetworkDefaultACLSelector from "./NetworkDefaultACLSelector";
+import NetworkDefaultACLRead from "./NetworkDefaultACLRead";
 
 interface Props {
   project: string;
@@ -15,6 +17,10 @@ interface Props {
 
 const NetworkAcls: FC<Props> = ({ formik, project }) => {
   const networlAclSelectorId = useId();
+  const defaultEgressIngress = {
+    Egress: formik.values.security_acls_default_egress ?? "",
+    Ingress: formik.values.security_acls_default_ingress ?? "",
+  };
 
   return (
     <div className="general-field">
@@ -67,6 +73,19 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
               formik.setFieldValue("security_acls", selectedItems);
             }}
             id={networlAclSelectorId}
+          />
+        )}
+        {formik.values.readOnly ? (
+          formik.values.security_acls.length !== 0 && (
+            <NetworkDefaultACLRead values={defaultEgressIngress} />
+          )
+        ) : (
+          <NetworkDefaultACLSelector
+            onChange={(fieldValue, value) => {
+              formik.setFieldValue(fieldValue, value);
+            }}
+            values={defaultEgressIngress}
+            disabled={formik.values.security_acls.length === 0}
           />
         )}
       </div>
--- a/src/pages/networks/forms/NetworkDefaultACLRead.tsx
+++ b/src/pages/networks/forms/NetworkDefaultACLRead.tsx
@@ -1,0 +1,26 @@
+import type { FC } from "react";
+import { Icon } from "@canonical/react-components";
+import { conjugateACLAction } from "util/helpers";
+import type { Direction } from "./NetworkDefaultACLSelector";
+
+const NetworkDefaultACLRead: FC<{
+  values: Record<Direction, string>;
+}> = ({ values }) => {
+  const egressAction = values.Egress;
+  const ingressAction = values.Ingress;
+
+  return (
+    <div className="u-sv1">
+      When no ACL rule matches:
+      <br />
+      <Icon name="arrow-left" className="network-default-acl-icon" />
+      Egress traffic will be{" "}
+      <code>{conjugateACLAction(egressAction || "reject")}</code>
+      <br />
+      <Icon name="arrow-right" className="network-default-acl-icon" />
+      Ingress traffic will be{" "}
+      <code>{conjugateACLAction(ingressAction || "reject")}</code>
+    </div>
+  );
+};
+export default NetworkDefaultACLRead;
--- a/src/pages/networks/forms/NetworkDefaultACLSelector.tsx
+++ b/src/pages/networks/forms/NetworkDefaultACLSelector.tsx
@@ -1,0 +1,67 @@
+import type { FC } from "react";
+import { Fragment } from "react";
+import { Icon, Select } from "@canonical/react-components";
+import { conjugateACLAction } from "util/helpers";
+
+export type Direction = "Ingress" | "Egress";
+
+interface Props {
+  onChange: (fieldValue: string, value: string) => void;
+  values?: Record<Direction, string>;
+  disabled?: boolean;
+}
+
+const NetworkDefaultACLSelector: FC<Props> = ({
+  values,
+  onChange,
+  disabled,
+}) => {
+  const DIRECTIONS: Direction[] = ["Egress", "Ingress"];
+  const ACTIONS = ["allow", "drop", "reject"];
+  const FIELD_BY_DIRECTION: Record<Direction, string> = {
+    Egress: "security_acls_default_egress",
+    Ingress: "security_acls_default_ingress",
+  };
+
+  const options = [
+    { label: "Select option", value: "" },
+    ...ACTIONS.map((value) => ({
+      label: conjugateACLAction(value),
+      value: value,
+    })),
+  ];
+
+  return (
+    <div>
+      <div className="u-sv2">When no ACL rule matches:</div>
+      <div className="network-default-acl-selector u-sv1">
+        {DIRECTIONS.map((direction) => {
+          return (
+            <Fragment key={direction}>
+              <div>
+                <Icon
+                  name={direction === "Egress" ? "arrow-left" : "arrow-right"}
+                  className="network-default-acl-icon"
+                />
+                {direction} traffic will be{" "}
+              </div>
+              <div>
+                <Select
+                  className="u-no-margin--bottom"
+                  options={options}
+                  onChange={(e) => {
+                    onChange(FIELD_BY_DIRECTION[direction], e.target.value);
+                  }}
+                  value={values?.[direction]}
+                  disabled={disabled}
+                />
+              </div>
+            </Fragment>
+          );
+        })}
+      </div>
+    </div>
+  );
+};
+
+export default NetworkDefaultACLSelector;
--- a/src/pages/networks/forms/NetworkForm.tsx
+++ b/src/pages/networks/forms/NetworkForm.tsx
@@ -100,6 +100,8 @@ export interface NetworkFormValues {
   entityType: "network";
   bareNetwork?: LxdNetwork;
   editRestriction?: string;
+  security_acls_default_egress?: string;
+  security_acls_default_ingress?: string;
 }
 
 export const toNetwork = (values: NetworkFormValues): Partial<LxdNetwork> => {
@@ -178,6 +180,10 @@ export const toNetwork = (values: NetworkFormValues): Partial<LxdNetwork> => {
         values.security_acls.length > 0
           ? values.security_acls.join(",")
           : undefined,
+      [getNetworkKey("security_acls_default_egress")]:
+        values.security_acls_default_egress,
+      [getNetworkKey("security_acls_default_ingress")]:
+        values.security_acls_default_ingress,
       [getNetworkKey("vlan")]: values.vlan ? values.vlan.toString() : undefined,
     },
   };
@@ -234,7 +240,7 @@ const NetworkForm: FC<Props> = ({
     return filteredRows;
   };
 
-  const isManageed = formik.values.bareNetwork?.managed ?? true;
+  const isManaged = formik.values.bareNetwork?.managed ?? true;
 
   // see https://documentation.ubuntu.com/lxd/en/latest/reference/networks/
   // for details on available sections per type
@@ -246,19 +252,19 @@ const NetworkForm: FC<Props> = ({
   const isPhysical = formik.values.networkType === physicalType;
   const isSriov = formik.values.networkType === sriovType;
 
-  if (isManageed && !isPhysical && !isMacvlan && !isSriov) {
+  if (isManaged && !isPhysical && !isMacvlan && !isSriov) {
     availableSections.push(BRIDGE);
   }
-  if (isManageed && hasIpV4 && !isMacvlan && !isSriov) {
+  if (isManaged && hasIpV4 && !isMacvlan && !isSriov) {
     availableSections.push(IPV4);
   }
-  if (isManageed && hasIpV6 && !isMacvlan && !isSriov) {
+  if (isManaged && hasIpV6 && !isMacvlan && !isSriov) {
     availableSections.push(IPV6);
   }
-  if (isManageed && !isMacvlan && !isSriov) {
+  if (isManaged && !isMacvlan && !isSriov) {
     availableSections.push(DNS);
   }
-  if (isManageed && isPhysical) {
+  if (isManaged && isPhysical) {
     availableSections.push(OVN);
   }
 
--- a/src/sass/_network_default_acl_selector.scss
+++ b/src/sass/_network_default_acl_selector.scss
@@ -1,0 +1,15 @@
+.network-default-acl-selector {
+  align-items: center;
+  display: grid;
+  gap: $sp-medium;
+  grid-template-columns: fit-content(50%) 1fr;
+  margin-left: $sph--small;
+  margin-right: $sph--small;
+  padding-bottom: $spv--large;
+}
+
+.network-default-acl-icon {
+  margin-left: $sph--small;
+  margin-right: $sph--small;
+  opacity: 0.66;
+}
--- a/src/types/network.d.ts
+++ b/src/types/network.d.ts
@@ -62,6 +62,8 @@ export interface LxdNetworkConfig {
   vlan?: string;
   [key: `user.${string}`]: string;
   [key: `volatile.${string}`]: string;
+  "security.acls.default.egress.action"?: string;
+  "security.acls.default.ingress.action"?: string;
 }
 
 export interface LxdNetwork {
--- a/src/util/networks.tsx
+++ b/src/util/networks.tsx
@@ -113,6 +113,8 @@ export const networkFormFieldToPayloadName: Record<
   parent: "parent",
   security_acls: "security.acls",
   vlan: "vlan",
+  security_acls_default_egress: "security.acls.default.egress.action",
+  security_acls_default_ingress: "security.acls.default.ingress.action",
 };
 
 export const getHandledNetworkConfigKeys = () => {
--- a/src/sass/styles.scss
+++ b/src/sass/styles.scss
@@ -8,6 +8,8 @@
 @include vf-p-icon-add-canvas;
 @include vf-p-icon-add-logical-volume;
 @include vf-p-icon-applications;
+@include vf-p-icon-arrow-left;
+@include vf-p-icon-arrow-right;
 @include vf-p-icon-begin-downloading;
 @include vf-p-icon-book;
 @include vf-p-icon-canvas;
@@ -104,6 +106,7 @@ $border-thin: 1px solid $colors--theme--border-low-contrast !default;
 @import "meter";
 @import "migrate_instance";
 @import "modified_actions";
+@import "network_default_acl_selector";
 @import "network_form";
 @import "network_forwards_form";
 @import "network_ipam";
--- a/src/util/helpers.tsx
+++ b/src/util/helpers.tsx
@@ -435,3 +435,7 @@ export const fileToSanitisedName = (
   const newName = truncateEntityName(sanitisedFileName, suffix);
   return newName;
 };
+
+export const conjugateACLAction = (action: string) => {
+  return action === "drop" ? "dropped" : action + "ed";
+};
--- a/src/util/networkForm.tsx
+++ b/src/util/networkForm.tsx
@@ -82,5 +82,9 @@ export const toNetworkFormValues = (
     entityType: "network",
     bareNetwork: network,
     editRestriction,
+    security_acls_default_egress:
+      network.config[getNetworkKey("security_acls_default_egress")],
+    security_acls_default_ingress:
+      network.config[getNetworkKey("security_acls_default_ingress")],
   };
 };


.github/workflows/screenshot_automation.yaml
edlerd, line 1: What happened here, bad merge? Please revert.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +70,24 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
             id={networlAclSelectorId}
           />
         )}
+        {!formik.values.readOnly && (
```
edlerd: I think we should also display this config in read-only mode, not only in edit mode.

src/util/networks.tsx
```diff
@@ -110,6 +110,8 @@ export const networkFormFieldToPayloadName: Record<
   parent: "parent",
   security_acls: "security.acls",
   vlan: "vlan",
+  security_acls_default_egress: "security.acls.default.egress.action",
+  security_acls_default_ingress: "security.acls.default.ingress.action",
```
edlerd, line 117: This is the correct way to have the keys persist on save. I also see the new fields in the API response after saving.

What is missing to read the changes is to add the two new fields also to the `toNetworkFormValues` function in `networkForm.tsx`.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const setDirectionValue = (direction: Direction, value: string) => {
+    onChange(FIELD_BY_DIRECTION[direction], value);
+  };
+
+  useEffect(() => {
+    if (!isSpecific && !values) {
+      setValueForAllDirections(firstValue ?? "");
+    }
+  }, [isSpecific, firstValue]);
```
edlerd: Why is this needed?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+  trafficDirection: string;
+  values: string[];
+}
+
+interface Props {
+  id: string;
+  className?: string;
+  isReadOnly: boolean;
+  onChange: (fieldValue: string, value: string) => void;
+  values?: Record<string, string>;
```
edlerd: I think this is a bit too complicated structure, as we have exactly two well-defined values here. Should we factor this out to be `ingressValue` and `egressValue`?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+  canToggleSpecific?: boolean;
```
edlerd:
```suggestion
```
I think this will always be true, I don't see a use case for disabling this, so maybe can be removed?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+  isDefaultSpecific?: boolean;
```
edlerd: I think we can calculate this internally based on the incoming values. No need to expose it in the interface.
```suggestion
```

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+export interface DefaultACLSelectOption {
+  trafficDirection: string;
+  values: string[];
+}
+
+interface Props {
+  id: string;
+  className?: string;
+  isReadOnly: boolean;
```
edlerd: I think we should start using it (instead of hiding, see other comment). But maybe the read only case can be a bit lighter on the design, especially without using form elements.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+  onChange: (fieldValue: string, value: string) => void;
+  values?: Record<string, string>;
+  canToggleSpecific?: boolean;
+  isDefaultSpecific?: boolean;
+  disabled?: boolean;
+  helpText?: string | ReactNode;
```
edlerd:
```suggestion
```
I don't see a use case for help text right now, but this might change. If we don't need it, I think we should remove it from the interface.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+      )}
+      {isSpecific && (
+        <div className="cluster-specific-input">
+          {DIRECTIONS.map((direction) => {
+            const activeValue = values?.[direction];
+
+            return (
+              <Fragment key={direction}>
+                <div className="cluster-specific-member">{direction}</div>
+                <div className="cluster-specific-value">
```
edlerd: This is a bit unclear. As this is not a cluster related component. Should we rename those classes or factor out the common pattern into a pure component for the layout?

This can also be a follow-up to clean up, as I think if we factored out the pattern it would touch a lot of other files and places.

src/pages/networks/forms/NetworkForm.tsx
```diff
@@ -234,7 +240,7 @@ const NetworkForm: FC<Props> = ({
     return filteredRows;
   };
 
-  const isManageed = formik.values.bareNetwork?.managed ?? true;
+  const isManaged = formik.values.bareNetwork?.managed ?? true;
```
edlerd, line 243: good catch!

edlerd (commented):
Good start, the interaction is pretty clear for editing and the components look nicely structured.

src/pages/networks/forms/NetworkAcls.tsx
Kxiru: Could you give me some more insight into what you expect here, design wise? As displaying the Input fields feels like a break in the design:
image

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: Remnants of the original CMS inplementation that I thought were still functional, but I've just tested without them and everything looks okay so I'll remove them.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: I think the current implementation makes defining the activeVale in line 80 a lot simpler. Also do you mean factoring it out within this component or within the parent component? Please explain.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: Let's iterate on the design in the earlier comment.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: Happy to refactor the scss naming now. I was just focused on getting it working and visible first :)

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+  trafficDirection: string;
+  values: string[];
+}
+
+interface Props {
+  id: string;
+  className?: string;
+  isReadOnly: boolean;
+  onChange: (fieldValue: string, value: string) => void;
+  values?: Record<string, string>;
```
edlerd:
```suggestion
  values?: Record<Direction, string>;
```
I think we can use the stricter type here?

src/types/network.d.ts
```diff
@@ -75,6 +77,8 @@ export interface LxdNetwork {
   used_by?: string[];
   etag?: string;
   access_entitlements?: string[];
+  security_acls_default_egress?: string;
+  security_acls_default_ingress?: string;
```
edlerd:
```suggestion
```
This seems wrong, the fields are only nested in the config part of this structure.

src/sass/_settings_page.scss
edlerd, line 1: Changes to this file seem wrong, as everything in this file is name spaced under the `settings` class that doesn't apply to the changes in this PR. Should they be reverted?

src/pages/networks/forms/NetworkAcls.tsx
edlerd: 1. It should only be visible if an ACL is assigned, I think the settings has no effect if no ACL applies. Please verify with the docs
2. I could imagine two sentences below the ACL chips: `When no ACL rule matches, <bold>egress</bold> traffic will be <code>dropped</code>` and `When no ACL rule matches, <bold>ingress</bold> traffic will be <code>rejected</code>`. Or if both are the same, a single sentence `When no ACL rule matches, <bold>egress</bold> and <bold>ingress</blod>traffic will be <code>dropped</code>`.
3. The read-only mode should reflect the LXD default if no specific setting is available on the network and the network has ACLs applied. Please consult with the docs what the default is.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+        disabled={isReadOnly || disabled}
+      />
+      {isSpecific && (
+        <div className="direction-specific-input">
+          {DIRECTIONS.map((direction) => {
+            const activeValue = values?.[direction];
+
+            return (
+              <Fragment key={direction}>
+                <div className="direction-specific-member">{direction}</div>
```
edlerd: The class name seems wrong, there is no member.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +70,24 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
...
+          <NetworkDefaultACLSelector
+            key={"NetworkDefaultACLSelector"}
+            id={""}
```
edlerd:
```suggestion
            id={""}
```
We probably don't need a key and should have a valid id.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,42 @@
+import type { FormikProps } from "formik";
+import type { FC } from "react";
+import type { NetworkFormValues } from "./NetworkForm";
+
+const NetworkDefaultACLReadOnly: FC<{
```
edlerd: This needs a default. I think the default is drop, but please double check with documentation.

image

edlerd: This should hide when no ACL is applied

image

edlerd: I think we should always have this tabular layout, even if the egress and ingress match. It seems much easier to read. Maybe we should even drop the checkbox for the `Bi-directional behavior when no ACL rule matches`. And make it always two dropdowns. That seems like the simpler interaction to me.

image

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,42 @@
...
+  formik: FormikProps<NetworkFormValues>;
+}> = ({ formik }) => {
+  const egress = formik.values.security_acls_default_egress;
+  const ingress = formik.values.security_acls_default_ingress;
+
+  if (egress === "" && ingress === "") {
```
kimanhou: This could be refactored into a getText(egress, ingress) function that returns the string (or ReactNode to include bold). Then the return should only takes care of the rendering. It would avoid duplication and be easier to maintain.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+      />
+      {isSpecific && (
+        <div className="direction-specific-input">
+          {DIRECTIONS.map((direction) => {
+            const activeValue = values?.[direction as Direction];
+
+            return (
+              <Fragment key={direction}>
+                <div className="direction-specific-key">{direction}</div>
+                <div className="direction-specific-value">
```
kimanhou: Is the classname used ? If not, then do we need to `div` ?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    disabled={isReadOnly || disabled}
+                  />
+                </div>
+              </Fragment>
+            );
+          })}
+        </div>
+      )}
+      {!isSpecific && (
+        <div className="direction-specific-value-wrapper">
```
kimanhou: Is the classname used ? If not, then do we need to `div` ?

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
Kxiru:
~~~The default is Rejected, sorry, it seems my latest push wasn't successful.~~~

Oh, the push was successful but I think the logic is bamboozled, will address.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
Kxiru: I've changed this logic due to other changes, so this is no longer applicable, thanks!

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: "Is the className used?" I don't understand the question.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
Kxiru: Hmm,
Firstly, this wasn't what I had in mind when I suggested a tabular layout, this is the textual layout to me, by tabular I meant the device table in the instance detail page for example. I'm not a huge fan of this design above.
In any case, where do we go form here? You now just want two selector inputs instead of one selector for both and the checkbox?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
kimanhou: When you search for `direction-specific-value` in the codebase, it does not appear. So is it needed or can it be removed ?

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
edlerd: > In any case, where do we go form here? You now just want two selector inputs instead of one selector for both and the checkbox?

We should probably discuss in person to find the best solution. Personally, I think the version in the screen above is easiest to parse and understand for a user.
Then my thinking was that the edit mode without checkbox and always having it in sync with the read mode with two selectors might be easier to use and understand too.

Kxiru: Discussed in meeting.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,51 @@
...
+      </div>
+    );
+  } else if (egress === ingress) {
+    return (
+      <div className="">
+        When no ACL rule matches, <b>egress</b> and <b>ingress</b> traffic will
+        be <code>{conjugateAction(egress ?? "")}</code>
+      </div>
+    );
+  } else {
```
edlerd: I think we should remove these cases and always show the split view with icons. Then the user doesn't have to think when going into edit mode to understand a very different representation of the information. Currently, the representation changes. With removing this, the representation becomes symmetrical. So a user learns this once and can reuse that on edit.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return (
+      <div className="u-sv1">
+        When no ACL rule matches
+        <br />
+        <Icon
+          name="arrow-up"
+          style={{ marginRight: "0.5rem", marginLeft: "0.5rem", opacity: 0.66 }}
```
edlerd: This was a quick sketch. It should be refactored to use a class and SCSS variables.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,51 @@
+import type { FormikProps } from "formik";
+import type { FC } from "react";
+import type { NetworkFormValues } from "./NetworkForm";
+import { Icon } from "@canonical/react-components";
+
+const NetworkDefaultACLReadOnly: FC<{
+  formik: FormikProps<NetworkFormValues>;
+}> = ({ formik }) => {
+  const egress = formik.values.security_acls_default_egress;
+  const ingress = formik.values.security_acls_default_ingress;
```
edlerd:
```suggestion
  const egressAction = formik.values.security_acls_default_egress;
  const ingressAction = formik.values.security_acls_default_ingress;
```
A bit clearer?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const ACTIONS = ["Allow", "Drop", "Reject"];
+  const FIELD_BY_DIRECTION: Record<Direction, string> = {
+    Egress: "security_acls_default_egress",
+    Ingress: "security_acls_default_ingress",
+  };
+
+  const allOptions = [
+    { label: "Select option", value: "" },
+    ...ACTIONS.map((value) => ({
+      label: value,
```
edlerd: I think the label should be conjugated. Maybe reuse the same function here as in read mode?

image

src/sass/_direction_specific_input.scss
```diff
@@ -0,0 +1,11 @@
+.direction-specific-input {
+  align-items: center;
+  display: grid;
+  gap: $sp-medium;
+  grid-template-columns: fit-content(50%) 1fr;
+  padding-bottom: $spv--large;
+
+  .direction-specific-key {
+    overflow: hidden;
+  }
```
edlerd:
```suggestion
```
I don't think we need this?
Maybe we can remove both of the `direction-specific-key` and `direction-specific-value` classes completely?

src/sass/_direction_specific_input.scss
```diff
@@ -0,0 +1,11 @@
+.direction-specific-input {
```
edlerd: Naming of the class and file should match the component name a bit closer.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,51 @@
+import type { FormikProps } from "formik";
+import type { FC } from "react";
+import type { NetworkFormValues } from "./NetworkForm";
+import { Icon } from "@canonical/react-components";
+
+const NetworkDefaultACLReadOnly: FC<{
```
edlerd: The default case is broken. We should state the correct default if the keys are missing.

image

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
Kxiru: Ah yes, I forgot.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +71,27 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
             id={networlAclSelectorId}
           />
         )}
+        {formik.values.readOnly ? (
+          formik.values.security_acls.length !== 0 && (
+            <NetworkDefaultACLReadOnly formik={formik} />
+          )
+        ) : (
+          <NetworkDefaultACLSelector
+            id={"NetworkDefaultACLSelector"}
```
edlerd:
```suggestion
```
We probably don't need the id prop.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +71,27 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
...
+            isReadOnly={false}
```
edlerd:
```suggestion
```
We probably can remove this as well.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +71,27 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
...
+              formik.setFieldValue(fieldValue, value);
+            }}
+            values={{
+              Egress: formik.values.security_acls_default_egress ?? "",
+              Ingress: formik.values.security_acls_default_ingress ?? "",
+            }}
+            disabled={
+              !!formik.values.editRestriction ||
+              formik.values.security_acls.length === 0
+            }
```
edlerd:
```suggestion
            disabled={formik.values.security_acls.length === 0}
```
We probably don't need the first line here, as this will only be rendered in edit mode enabled, which is blocked for users with restrictions.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,27 @@
...
+  formik: FormikProps<NetworkFormValues>;
+}> = ({ formik }) => {
+  const egressAction = formik.values.security_acls_default_egress;
+  const ingressAction = formik.values.security_acls_default_ingress;
+
+  return (
+    <div className="u-sv1">
+      When no ACL rule matches
+      <br />
+      <Icon name="arrow-up" className="network-default-acl-readonly" />
```
edlerd:
```suggestion
      <Icon name="arrow-up" className="network-default-acl-icon" />
```
I'd reuse this on edit, wdyt?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export interface DefaultACLSelectOption {
+  trafficDirection: string;
+  values: string[];
+}
+
+interface Props {
+  id: string;
+  className?: string;
+  isReadOnly: boolean;
```
edlerd:
```suggestion
```
I think we can make the interface a bit smaller here.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <div className="network-default-acl-selector u-sv1">
+        {DIRECTIONS.map((direction) => {
+          const activeValue = values?.[direction as Direction];
+
+          return (
+            <Fragment key={direction}>
+              <div>
+                <Icon
+                  name="arrow-up"
+                  className="network-default-acl-selector"
```
edlerd: Should this be the icon class name?

src/sass/_network_default_acl_selector.scss
```diff
@@ -0,0 +1,7 @@
+.network-default-acl-selector {
```
edlerd, line 1: please adjust the file name as well.

src/sass/_network_form.scss
```diff
@@ -127,3 +127,14 @@
...
+.network-default-acl-readonly {
+  margin-left: "0.5rem";
+  margin-right: "0.5rem";
+  opacity: 0.66;
+}
+
+.network-default-acl-selector {
+  margin-left: "0.5rem";
+  margin-right: "0.5rem";
+}
```
edlerd: This can move into the same file as the .network-default-acl-selector

src/sass/_network_form.scss
```diff
@@ -127,3 +127,14 @@
     }
   }
 }
+
+.network-default-acl-readonly {
+  margin-left: "0.5rem";
+  margin-right: "0.5rem";
```
edlerd: css is broken here, also please use variables from vanilla

edlerd (commented):
Good progress, some more refinement, and this should be good to merge.

Sorry for the back and forth with the checkbox. I thought it would be easier, but then remembered the "Don't make me think" mantra for good UX and switched back to two selectors.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <div className="u-sv2">When no ACL rule matches</div>
+      <div className="network-default-acl-selector u-sv1">
+        {DIRECTIONS.map((direction) => {
+          const activeValue = values?.[direction as Direction];
+
+          return (
+            <Fragment key={direction}>
+              <div>
+                <Icon
+                  name="arrow-up"
```
kimanhou: The icon should be dynamic according to the direction.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,27 @@
+import type { FormikProps } from "formik";
+import type { FC } from "react";
+import type { NetworkFormValues } from "./NetworkForm";
+import { Icon } from "@canonical/react-components";
+import { conjugateACLAction } from "util/helpers";
+
+const NetworkDefaultACLReadOnly: FC<{
```
kimanhou: In Network device configuration, I used the prefix "Read" in file names, instead of "ReadOnly".

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
kimanhou: Should the text be muted when the selector is disabled ?

src/sass/_network_form.scss
```diff
@@ -127,3 +127,14 @@
...
+  opacity: 0.66;
+}
+
+.network-default-acl-selector {
+  margin-left: "0.5rem";
```
kimanhou: You don't need double quotes here. And you can use Vanilla variables.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,84 @@
...
+                  className="network-default-acl-selector"
+                />
+                {direction} traffic will be{" "}
+              </div>
+              <div>
+                <Select
+                  id={`${id}-${direction}`}
```
kimanhou: What is the id for ?

edlerd: Some react form elements must have an id for accessibility to work properly. Like for the label to reference the input internally. Select might be one of those.

kimanhou: I know but then wouldn't the `id` be on the label too ? Or is the label included in the `Select` component ?
If we remove the `id` props from here, then it would just be `direction` ?

edlerd: Yes, the label is added by the Select component. But I just checked, and the component also [generates an ID if none is passed in](https://github.com/canonical/react-components/blob/main/src/components/Select/Select.tsx#L117-L118). So we can remove the ID in here.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: This file seems outdated as I have since removed these from the file.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: @kimanhou , the text should be muted when no ACL is selected.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
Kxiru: Happy to use Read as a suffix rather than a prefix, but otherwise the codebase convention specifies the entity first within the name, and then the page/action.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
Kxiru: Thanks for this reminder!

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <div className="u-sv1">
+      When no ACL rule matches
+      <br />
+      <Icon name="arrow-up" className="network-default-acl-icon" />
+      Egress traffic will be{" "}
+      <code>{conjugateACLAction(egressAction ?? "reject")}</code>
+      <br />
+      <Icon name="arrow-down" className="network-default-acl-icon" />
+      Ingress traffic will be{" "}
+      <code>{conjugateACLAction(ingressAction ?? "reject")}</code>
```
edlerd:
```suggestion
      When no ACL rule matches,
      <br />
      <Icon name="arrow-up" className="network-default-acl-icon" />
      egress traffic will be{" "}
      <code>{conjugateACLAction(egressAction ?? "reject")}</code>
      <br />
      <Icon name="arrow-down" className="network-default-acl-icon" />
      ingress traffic will be{" "}
      <code>{conjugateACLAction(ingressAction ?? "reject")}</code>
```
Maybe this is a way to make the line above more connected to the two lines below? Or we could bring the `:` back. Any solution should be applied here and in the edit form.
Not sure what is best though, wdyt?

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Icon, Select } from "@canonical/react-components";
+import { conjugateACLAction } from "util/helpers";
+
+type Direction = "Ingress" | "Egress";
+
+export interface DefaultACLSelectOption {
+  trafficDirection: string;
+  values: string[];
+}
+
```
edlerd:
```suggestion
```
This seems unused.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+  disabled?: boolean;
+}
+
+const NetworkDefaultACLSelector: FC<Props> = ({
+  values,
+  onChange,
+  disabled,
+}) => {
+  const DIRECTIONS = ["Egress", "Ingress"];
+  const ACTIONS = ["Allow", "Drop", "Reject"];
```
edlerd:
```suggestion
  const ACTIONS = ["allow", "drop", "reject"];
```
Let's save the lower case function calls below.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+    { label: "Select option", value: "" },
+    ...ACTIONS.map((value) => ({
+      label: conjugateACLAction(value.toLowerCase()),
+      value: value.toLowerCase(),
+    })),
+  ];
+
+  const setDirectionValue = (direction: Direction, value: string) => {
+    onChange(FIELD_BY_DIRECTION[direction], value);
+  };
```
edlerd:
```suggestion
```
I think we should inline this as it is a one line function.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+  values?: Record<Direction, string>;
+  disabled?: boolean;
+}
+
+const NetworkDefaultACLSelector: FC<Props> = ({
+  values,
+  onChange,
+  disabled,
+}) => {
+  const DIRECTIONS = ["Egress", "Ingress"];
```
edlerd:
```suggestion
  const DIRECTIONS: Direction[] = ["Egress", "Ingress"];
```
If we type this correctly, we can save some casts below.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+              <div>
+                <Icon
+                  name={direction === "Egress" ? "arrow-up" : "arrow-down"}
+                  className="network-default-acl-icon"
+                />
+                {direction} traffic will be{" "}
+              </div>
+              <div>
+                <Select
+                  id={`NetworkDefaultACLSelector-${direction}`}
```
edlerd:
```suggestion
```

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,78 @@
...
+  disabled,
+}) => {
+  const DIRECTIONS = ["Egress", "Ingress"];
+  const ACTIONS = ["Allow", "Drop", "Reject"];
+  const FIELD_BY_DIRECTION: Record<Direction, string> = {
+    Egress: "security_acls_default_egress",
+    Ingress: "security_acls_default_ingress",
+  };
+
+  const allOptions = [
```
edlerd:
```suggestion
  const options = [
```

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
edlerd: I think this is about renaming file and component from `NetworkDefaultACLReadOnly` to `NetworkDefaultACLRead`.

src/pages/networks/forms/NetworkDefaultACLRead.tsx
```diff
@@ -0,0 +1,27 @@
+import type { FormikProps } from "formik";
+import type { FC } from "react";
+import type { NetworkFormValues } from "./NetworkForm";
+import { Icon } from "@canonical/react-components";
+import { conjugateACLAction } from "util/helpers";
+
+const NetworkDefaultACLReadOnly: FC<{
+  formik: FormikProps<NetworkFormValues>;
```
edlerd: Can we reuse the same `values?: Record<Direction, string>` interface as for the edit mode? Because we will want to reuse this on the instance nic devices and there we'll have a different formik type. The interface for edit mode will be possible to adapt to instance and profile nic devices.

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
kimanhou: Yes ⬆️
Sorry, I confused prefix and suffix in my original comment

src/pages/networks/forms/NetworkDefaultACLRead.tsx
kimanhou: > Because we will want to reuse this on the instance nic devices

We will ?! 😱

src/pages/networks/forms/NetworkDefaultACLReadOnly.tsx
Kxiru: I'm not a fan of the design, wdyt?

Old version:
image

Your version:
image

Maybe our best bet is with a colon:
image

edlerd: Sure, can go with the last one with colon.

src/pages/networks/forms/NetworkDefaultACLRead.tsx
Kxiru: I'm not sure what you are asking me to do, here @edlerd

edlerd: The idea was not to receive the formik object, but to unpack the values in the caller already. Reusing the interface for values as a record as you defined it in edit mode.

This will unlock reusing this component in instance and profile network config pages that have a different formik type.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,69 @@
...
+      value: value,
+    })),
+  ];
+
+  return (
+    <div>
+      <div className="u-sv2">When no ACL rule matches:</div>
+      <div className="network-default-acl-selector u-sv1">
+        {DIRECTIONS.map((direction) => {
+          const activeValue = values?.[direction];
```
edlerd:
```suggestion
```
I think we can inline this now it is so simple without the cast.

src/pages/networks/forms/NetworkDefaultACLRead.tsx
edlerd: Can we change this to make it reusable? Instead of receiving the formik object (which is not compatible with the formik object used for profiles or instances), pass in the `values?: Record<Direction, string>`, just like in edit mode.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +71,27 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
             id={networlAclSelectorId}
           />
         )}
+        {formik.values.readOnly ? (
+          formik.values.security_acls.length !== 0 && (
+            <NetworkDefaultACLRead
+              values={{
+                Egress: formik.values.security_acls_default_egress ?? "",
+                Ingress: formik.values.security_acls_default_ingress ?? "",
```
edlerd: This breaks the default fallback inside
image

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <div>
+      <div className="u-sv2">When no ACL rule matches:</div>
+      <div className="network-default-acl-selector u-sv1">
+        {DIRECTIONS.map((direction) => {
+          return (
+            <Fragment key={direction}>
+              <div>
+                <Icon
+                  name={direction === "Egress" ? "arrow-up" : "arrow-down"}
```
omarelkashef: Is the decision to use arrow up and down for Egress and Ingress, respectively, because we are dealing with VMs and containers? Usually in cloud providers,  they are depicted as left and right, respectively.

edlerd: Good point, maybe better to use this one for ingress:
image

and that one for egress:

image

Wdyt?

Kxiru: Discussed in meeting.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +71,27 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
...
+          )
+        ) : (
+          <NetworkDefaultACLSelector
+            onChange={(fieldValue, value) => {
+              formik.setFieldValue(fieldValue, value);
+            }}
+            values={{
+              Egress: formik.values.security_acls_default_egress ?? "",
+              Ingress: formik.values.security_acls_default_ingress ?? "",
+            }}
```
edlerd:
```suggestion
const defaultEgressIngress = {
              Egress: formik.values.security_acls_default_egress ?? "",
              Ingress: formik.values.security_acls_default_ingress ?? "",
            }
```
Let's extract a variable not to repat it and reuse it here and above.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
edlerd: As discussed, lets use the upload/download icons here and in read mode.

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +75,27 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
             id={networlAclSelectorId}
           />
         )}
+        {formik.values.readOnly ? (
+          formik.values.security_acls.length !== 0 && (
+            <NetworkDefaultACLRead
+              values={{
+                Egress: defaultEgressIngress.Egress,
+                Ingress: defaultEgressIngress.Ingress,
+              }}
```
edlerd:
```suggestion
              values={defaultEgressIngress}
```
The idea was to simplify this :)

src/pages/networks/forms/NetworkAcls.tsx
```diff
@@ -69,6 +75,27 @@ const NetworkAcls: FC<Props> = ({ formik, project }) => {
...
+          )
+        ) : (
+          <NetworkDefaultACLSelector
+            onChange={(fieldValue, value) => {
+              formik.setFieldValue(fieldValue, value);
+            }}
+            values={{
+              Egress: defaultEgressIngress.Egress,
+              Ingress: defaultEgressIngress.Ingress,
+            }}
```
edlerd:
```suggestion
            values={defaultEgressIngress}
```

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
edlerd: Did we discuss in standup to use the right/left arrows?

Kxiru: image
You requested this in Mattermost. I can switch this back to Up/Down arrows if you so wish.

kimanhou: ~~I thought we agreed on download/upload~~
Just read the end of the thread of Mattermost

Kxiru: @kimanhou ~~I tested this and presented it above (Please expand resolved conversations), but it didn't look nice.~~ I also put this in Mattermost I believe. Also the icons aren't the same size by design for some reason :skull: .

Edit: Apologies, it is just in Mattermost.
image

edlerd: @Kxiru so did we try and share the left / right arrows? I thought you mentioned in standup you thanked Omar for the suggestion of right left and said that you tried it, and it was better. So I was confused why the PR use up/down still.

edlerd: On some further testing, I realize those are rotated 90 degrees. I think this looks better, yes!

Can we use left for egress and right for ingress without rotation? As vanilla exports them, we can use the correctly named ones :)

src/sass/_network_default_acl_selector.scss
```diff
@@ -0,0 +1,16 @@
...
+  margin-left: $sph--small;
+  margin-right: $sph--small;
+  padding-bottom: $spv--large;
+}
+
+.network-default-acl-icon {
+  margin-left: $sph--small;
+  margin-right: $sph--small;
+  opacity: 0.66;
+  width: $sp-x-large;
```
edlerd:
```suggestion
```
I think we should rely on the default size of the icon. I think the spacing is slightly better without stretching the icons. Why expose a width on it?

edlerd (dismissed):
LGTM, thanks for driving this to perfection.
One nitpick on the styling below.

src/sass/_network_default_acl_selector.scss
Kxiru: Done

PR #1664, opened 2025-12-03
feat: [WD-31442] Rich instance tooltips on chips

omarelkashef:
## Done

- Add rich tooltips on chips for instances

## QA

1. Run the LXD-UI:
    - Go to a page with an instance chip, e.g. Networks -> IPAM, Networks -> network -> Configuration
    - Hover over instance chip
    - Ensure tooltip is displayed correctly

## Screenshots

image
image


src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const InstanceRichTooltipMessage: FC<Props> = ({
+  instanceName,
+  projectName,
+}) => {
+  const instanceLoading = useInstanceLoading();
+  const { data: instance } = useInstance(instanceName, projectName);
+
+  if (!instance) {
+    return <></>;
+  }
```
edlerd: We should have some form of loading indicator instead. And we should reserve the right amount of space, so the tooltip can position itself correctly. Currently with a cold cache it behaves a bit wired and positions wrongly.

https://github.com/user-attachments/assets/3a66a8e0-3e6e-4015-8335-8f76fcc52c44

src/components/InstanceRichTooltip.tsx
```diff
@@ -0,0 +1,36 @@
+import type { FC } from "react";
+import { Tooltip } from "@canonical/react-components";
+import { InstanceRichTooltipMessage } from "./InstanceRichTooltipMessage";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+
+interface Props {
+  children: React.ReactNode;
```
edlerd: I think the interface can be simplified. It seems we always want a `ResourceLink` with the instance as a child. So we should not expect that from the caller, as we can build it here internally.

```suggestion
```

src/pages/networks/NetworkTopology.tsx
```diff
@@ -51,7 +54,12 @@ const NetworkTopology: FC<Props> = ({ formik, project, isServerClustered }) => {
     );
   });
 
-  const instances = filterUsedByType("instance", network.used_by);
+  const instancesNames = filterUsedByType("instance", network.used_by).map(
+    (instance) => instance.name,
+  );
+  const { data: instances = [] } = useInstances(project) ?? [];
+  instances.filter((item) => instancesNames.includes(item.name));
+
```
edlerd: What does this do?

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -67,7 +67,7 @@ const UploadInstanceBackupFileForm: FC<Props> = ({
     const message = (
       <>
         Created instance{" "}
-        <ResourceLink type="instance" value={instanceName} to={instanceUrl} />.
+        <ResourceLink type="instance" value={instanceName} to={instanceUrl} />.{" "}
```
edlerd: Why add a space here?

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -109,7 +109,7 @@ const UploadInstanceBackupFileForm: FC<Props> = ({
         toastNotify.info(
           <>
             Upload completed. Now creating instance{" "}
-            <ResourceLabel bold type="instance" value={values.name} />.
+            <ResourceLabel bold type="instance" value={values.name} />
```
edlerd: Positive notifications should end with a dot. Why remove it?

src/pages/networks/NetworkTopology.tsx
omarelkashef: This is left by mistake from the old implementation where we passed the instance to the tooltip component. It should be removed.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <th className="u-text--muted">IPv4</th>
+          <td>
+            {instance ? <InstanceIps instance={instance} family="inet" /> : "-"}
+          </td>
+        </tr>
+        <tr>
+          <th className="u-text--muted">IPv6</th>
+          <td>
+            {instance ? (
+              <InstanceIps instance={instance} family="inet6" />
```
edlerd: Those lists can be too long, so the tooltip won't fit the screen any more. I think we can use `InstanceListIps` which has max 3 lines here.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,93 @@
...
+              "-"
+            )}
+          </td>
+        </tr>
+        <tr>
+          <th className="u-text--muted">MAC addresses</th>
+          <td>
+            {instance ? <InstanceMACAddresses instance={instance} /> : "-"}
+          </td>
+        </tr>
```
edlerd: This has the same problem of growing long if many interfaces are assigned. Shall we remove the mac here or limit to 2 items as well?

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <td>{instance ? <InstanceStatusIcon instance={instance} /> : "-"}</td>
+        </tr>
+        <tr>
+          <th className="u-text--muted">Architecture</th>
+          <td>{instance ? instance.architecture : "-"}</td>
+        </tr>
+        <tr>
+          <th className="u-text--muted">Description</th>
+          <td>
+            {instance && instance.description ? instance.description : "-"}
```
edlerd: This needs some truncation not to overflow.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  return (
+    <table className="u-table-layout--auto u-no-margin--bottom">
+      <tbody>
+        <tr>
+          <th className="u-text--muted">Name</th>
+          <td>
+            {instance ? <InstanceLink instance={instance} /> : <Spinner />}
+          </td>
```
edlerd: This needs truncation as well

src/sass/_instance_rich_tooltip.scss
```diff
@@ -0,0 +1,7 @@
+body.is-dark .instance-rich-tooltip-table .p-icon--status-queued-small {
+  @include vf-icon-status-small($colors--light-theme--icon);
+}
+
+.instance-rich-tooltip-table .p-icon--status-queued-small {
+  @include vf-icon-status-small($colors--dark-theme--icon);
+}
```
edlerd: I think we need overrides for all the instance states here, no?

omarelkashef: Other status icons are colored, not black and white so they do not have the same problem as this one and do not need overriding. For example, https://github.com/canonical/vanilla-framework/blob/2929c627e3fad89e8366c93a804051416734f30f/scss/_patterns_icons.scss#L1362-L1374

src/components/InstanceRichTooltipMessage.tsx
omarelkashef: Yes we should limit it and it might be a good idea to create a similar component to `InstanceListIps`  or maybe generalize `InstanceListIps`  to `InstanceListAddresses` and it would take input addresses instead of ips and the internal implementation would not change.

src/sass/_instance_rich_tooltip.scss
edlerd: I think the stopped icon (like in the screenshot in the description) should be a white circle if on a dark background.

omarelkashef: The stopped status uses p-icon--status-queued-small which is overridden above. the screenshot should be updated to reflect this.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const macAddresses = instance ? getInstanceMacAddresses(instance) : [];
+
+  return (
+    <table className="u-table-layout--auto u-no-margin--bottom instance-rich-tooltip-table">
+      <tbody>
+        <tr>
+          <th className="u-text--muted">Name</th>
+          <td className="u-truncate">
+            {instance ? <InstanceLink instance={instance} /> : <Spinner />}
+          </td>
```
edlerd: We should add a title with the instance name, so if it is truncated, we get the tooltip on hover with the full name.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </tr>
+        <tr>
+          <th className="u-text--muted">Architecture</th>
+          <td>{instance ? instance.architecture : "-"}</td>
+        </tr>
+        <tr>
+          <th className="u-text--muted">Description</th>
+          <td className="u-truncate">
+            {instance && instance.description ? instance.description : "-"}
+          </td>
```
edlerd: Please add a title with the description (same as above).

src/pages/instances/InstanceListAddresses.tsx
```diff
@@ -0,0 +1,23 @@
+import type { FC } from "react";
+
+interface Props {
+  addresses: string[];
+}
+
+const InstanceListAddresses: FC<Props> = ({ addresses }) => {
```
edlerd: Good idea to rename and reuse this!

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,113 @@
...
+import type { IpFamily, LxdInstance } from "types/instance";
+import InstanceListAddresses from "pages/instances/InstanceListAddresses";
+import { getInstanceMacAddresses } from "util/instances";
+
+interface Props {
+  instanceName: string;
+  projectName: string;
+}
+
+export const InstanceRichTooltipMessage: FC<Props> = ({
```
edlerd: This is sometimes too big to fit on the screen:

image

Maybe we should remove architecture, mac address or even the dates?
Or find a layout that is not so high?
We could also be responsive in the sense that on larger height there are more properties displayed.

src/components/InstanceRichTooltip.tsx
```diff
@@ -0,0 +1,43 @@
...
+  projectName,
+}) => {
+  const isSmallScreen = useIsScreenBelow();
+  const instanceUrl = `/ui/project/${encodeURIComponent(projectName)}/instance/${encodeURIComponent(instanceName)}`;
+  const resourceLink = (
+    <ResourceLink
+      type="instance"
+      value={instanceName}
+      to={instanceUrl}
+      hasTitle={false}
```
edlerd:
```suggestion
      hasTitle={!isSmallScreen}
```
We should have the title on small screens, right?

src/components/InstanceRichTooltip.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { InstanceRichTooltipMessage } from "./InstanceRichTooltipMessage";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import ResourceLink from "./ResourceLink";
+
+interface Props {
+  instanceName: string;
+  projectName: string;
+}
+
+export const InstanceRichTooltip: FC<Props> = ({
```
edlerd: We can apply this new component to some more places. Like search for `type="instance` to find several `<ResourceLinks` that can be changed.

Also, the `InstanceLinkChip` should probably use this rich tooltip or maybe call this component?

omarelkashef: All usages of `<ResourceLinks type="instance"` outside this PR as well as `InstanceLinkChip` are for notifications which I think should not have tooltips. What do you think?

edlerd: Yes, we previously discussed not having the tooltip when the chip is in a notification.

Experiencing the feature now on the demo server, I had second thoughts on it. Especially when browsing the notification centre, it seems valuable to have the tooltip available. Also, to have things consistent, it seems better to have the rich chip everywhere.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }
+  const ipv4Addresses = instance ? getAddresses(instance, "inet") : [];
+  const ipv6Addresses = instance ? getAddresses(instance, "inet6") : [];
+  const macAddresses = instance ? getInstanceMacAddresses(instance) : [];
+
+  return (
+    <table className="u-table-layout--auto u-no-margin--bottom instance-rich-tooltip-table">
+      <tbody>
+        <tr>
+          <th className="u-text--muted">Name</th>
```
edlerd:
```suggestion
          <th className="u-text--muted">Instance name</th>
```
Do we have enough space for the longer version?
I think the tooltip should repeat the type of the entity.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,122 @@
...
+          <td title={instance ? instance.name : "-"} className="u-truncate">
```
edlerd:
```suggestion
          <td title={instance?.name} className="u-truncate">
```
I think when there is no content, we don't need the title. Its main use is for the overflow case for when the content is truncated and only half visible.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,122 @@
...
+}) => {
+  const getAddresses = (instance: LxdInstance, family: IpFamily) => {
+    const addresses = getIpAddresses(instance, family);
+    const sortedAddresses =
+      family === "inet6" ? sortIpv6Addresses(addresses) : addresses;
+    return sortedAddresses.map((item) => item.address);
+  };
+
+  const { data: instance, isLoading } = useInstance(instanceName, projectName);
+  const isLargeScreen = !useIsScreenBelow(largeScreenBreakpoint);
```
edlerd: I think this should react to screen height, not width. Even on a large window, I get this edge case below.

image

src/components/InstanceRichTooltipMessage.tsx
edlerd: If "instance name" is too long, we could shorten it to "Instance". Maybe that is better as more concise in any case.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <tbody>
+        <tr>
+          <th className="u-text--muted">Name</th>
+          <td title={instance ? instance.name : "-"} className="u-truncate">
+            {instance ? <InstanceLink instance={instance} /> : <Spinner />}
+          </td>
+        </tr>
+        <tr>
+          <th className="u-text--muted">Status</th>
+          <td>{instance ? <InstanceStatusIcon instance={instance} /> : "-"}</td>
```
edlerd:
```suggestion
          <td>
            {instance ? (
              <>
                <span className="margin-right">
                  <InstanceStatusIcon instance={instance} />
                </span>
                <InstanceStateActions
                  instance={instance}
                  className="u-no-margin--bottom"
                />
              </>
            ) : (
              "-"
            )}
          </td>
```
See https://github.com/canonical/lxd-ui/pull/1665 which this suggestion depends on. With that PR we should be able to show the instance state actions within the tooltip.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,133 @@
...
+            {instance ? (
+              <>
+                <InstanceStatusIcon instance={instance} />
+                <InstanceStateActions instance={instance} />
+              </>
+            ) : (
+              "-"
+            )}
+          </td>
+        </tr>
```
kimanhou: Should we have a reusable component RichTooltipRow with props: {label, value, tdClassName, tdTitle} ? Then we would have
```
<RichTooltipRow label="Status" value={...} />
<RichTooltipRow label="Description" value={...} />
```

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const ipv6Addresses = instance ? getAddresses(instance, "inet6") : [];
+  const macAddresses = instance ? getInstanceMacAddresses(instance) : [];
+
+  return (
+    <table className="u-table-layout--auto u-no-margin--bottom instance-rich-tooltip-table">
+      <tbody>
+        <tr>
+          <th className="u-text--muted">Instance</th>
+          <td title={instance?.name} className="u-truncate">
+            {instance ? <InstanceLink instance={instance} /> : <Spinner />}
```
kimanhou: Would it make sense to have an early return in case instance doesn't exist and not having ternary conditions for each row ? We would then have two clean return instead of one complicated return. What do you think ?

src/components/ResourceLink.tsx
```diff
@@ -23,7 +31,7 @@ const ResourceLink: FC<Props> = ({ type, value, to, disabled, className }) => {
         className,
       )}
       to={to}
-      title={value}
+      title={hasTitle ? value : ""}
```
kimanhou: `title` can be undefined so no need for empty string
```suggestion
      title={hasTitle ? value : undefined}
```

src/components/UsedByItem.tsx
```diff
@@ -68,7 +69,15 @@ const UsedByItem: FC<Props> = ({
           /{" "}
         </>
       )}
-      <ResourceLink type={type} value={label} to={to} />
+      {type === "instance" && (
```
kimanhou, line 72: We might want to use a switch in the future as there will be a different rich tooltip for each entity type.

src/components/InstanceRichTooltipMessage.tsx
omarelkashef: and value would be a React node? If so, that does not simplify the code much.

src/components/UsedByItem.tsx
omarelkashef: If I am not mistaken, we decided to limit only to instances @edlerd. But if not, then yes it would be good to add a switch.

omarelkashef: If I am not mistaken, we decided to limit only to instances @edlerd. But if not, then yes it would be good to add a switch.

src/components/InstanceRichTooltipMessage.tsx
omarelkashef: There is an early return for when the instance is not found. The ternary operators here are for the loading state. But now that I think about, we should change the condition to isLoading to make the code more readable and accurate.

src/components/InstanceRichTooltipMessage.tsx
edlerd: Maybe a better interface is to use children? Something like
```
<RichTooltipRow title="Status">{value}</RichTooltipRow/>

File: src/components/UsedByItem.tsx (hunk 24 above)
Comment (edlerd) [reply src/components/UsedByItem.tsx:72]: I think other types would be interesting to explore too as a next step. But as long as there is only one case I'd keep the switch as a refactor for the future with the 2nd case present.

File: src/components/InstanceRichTooltipMessage.tsx (hunk 21 above)
Comment (kimanhou) [reply src/components/InstanceRichTooltipMessage.tsx]: Yes the value would be a React node. It would save repeating `<tr><th className="u-text--muted">` and `<td>`.
If there is no more ternary condition in the value, that should looke way cleaner too.

File: src/components/InstanceRichTooltipMessage.tsx (hunk 22 above)
Comment (omarelkashef) [reply src/components/InstanceRichTooltipMessage.tsx]: I realized now why the condition was based on instance and not isLoading; it is because some of the components take instance as input and do not accept undefined, so using instance as condition with an early return if isLoading is false and instance is undefined achieves both goals: ensures instance is loaded and avoids ts errors.

File: src/sass/_instance_rich_tooltip.scss [hunk 25]
```diff
@@ -0,0 +1,29 @@
...
+
+.instance-rich-tooltip-table {
+  margin: $sp-small 0 0;
+  min-height: 400px;
+  min-width: 350px;
```
edlerd: Can you make the width of the tooltip stable between loading and after loading? Currently, there is some flickering once it is loaded due to width change.


https://github.com/user-attachments/assets/fdfeda07-2c8d-41d1-99b6-c521c8c9470c

src/sass/_instance_rich_tooltip.scss
```diff
@@ -0,0 +1,29 @@
...
+
+  td {
+    max-width: 300px;
+  }
+
+  th {
+    padding: $sp-small 0 0;
+  }
```
edlerd: If there is a single line of text, the rows in the tooltip should have the same spacing in terms of height.

Only the description column in the video below is symmetrical in terms of above and below spacing. The same should be applied to the status, type and other one line text rows.

https://github.com/user-attachments/assets/f5dcbe00-b82c-4314-bffb-ccc6a85dd749

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,135 @@
...
+          value={instance ? <InstanceLink instance={instance} /> : <Spinner />}
+          valueTitle={instanceName}
+        />
+        <InstanceRichTooltipRow
+          title="Status"
+          value={
+            instance ? (
+              <>
+                <InstanceStatusIcon instance={instance} />
+                <InstanceStateActions instance={instance} />
```
edlerd: Can we make sure the actions are always right aligned? Maybe give it a flex wrapper and make the first item with the status icon flex grow to push the state actions to the right? Because the status icon width can be variable depending on the status text.

src/components/InstanceRichTooltip.tsx
edlerd: This is still open.

src/components/InstanceRichTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+  instanceName,
+  projectName,
+  instance,
+}) => {
+  const showTooltip = !useIsScreenBelow(mediumScreenBreakpoint, "height");
+
+  const instanceUrl = `/ui/project/${encodeURIComponent(projectName)}/instance/${encodeURIComponent(instanceName)}`;
+  const resourceLink = (
+    <ResourceLink
+      type={instance ? instance.type : "instance"}
```
kimanhou: Can `instance.type` be different than `"instance"` ?

src/sass/_instance_rich_tooltip.scss
```diff
@@ -0,0 +1,35 @@
...
+  td {
+    max-width: 300px;
+  }
+
+  .actions-list {
+    margin: 0;
+  }
+
+  .status-row {
+    button {
```
kimanhou: Let's avoid targeting tag in CSS. Is this for `InstanceStateActions` ? Are the buttons in `status-row-content` ? Do they have the same classnames, like `p-button--base has-icon is-dense` ?

src/components/InstanceRichTooltip.tsx
omarelkashef: When it comes from`<InstanceLinkChip>`, it can be VM, container, or instance

src/sass/_instance_rich_tooltip.scss
omarelkashef: They do not have any descriptive class names that we can target. Do you think targeting something like p-button--base  is better than button? Both are generic.

edlerd: We should best define a new class and add it to the button to target.

edlerd: But do we even need this override or can we stay with the default as this is now the first row?

src/pages/instances/InstanceLinkChip.tsx
```diff
@@ -13,10 +12,10 @@ interface Props {
 
 const InstanceLinkChip: FC<Props> = ({ instance }) => {
```
edlerd: This component has a handful of use cases like the one below, where using the instance object does not have all properties. We should not cast it, but use the `InstanceRichTooltip` instead.

We can keep this component but should change the type of the receiving `instance` to be `LxdInstance` and reject with type error if the caller is building the instance object on the fly. Meaning, we should replace the partial type of the receiving prop to the proper one.

The naming between `InstanceRichTooltip` and `InstanceLinkChip` should be aligned. So it is clear one is the lazy loading one and the other receives the full object. Maybe `InstanceChip` and `InstanceChipLazy`?


image

src/components/InstanceRichTooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+import ResourceLink from "./ResourceLink";
+import type { LxdInstance } from "types/instance";
+
+interface Props {
+  instanceName: string;
+  projectName: string;
+  instance?: LxdInstance;
+}
+
+export const InstanceRichTooltip: FC<Props> = ({
```
edlerd: This and other `InstanceRich...` components should live in `pages/instances`.

src/components/InstanceRichTooltipRow.tsx
```diff
@@ -0,0 +1,23 @@
+import type { FC } from "react";
+interface Props {
+  title: string;
+  value: React.ReactNode;
+  valueTitle?: string;
+  className?: string;
+}
+
+export const InstanceRichTooltipRow: FC<Props> = ({
```
edlerd: We should keep this component in the `src/components` folder, but make it generic. I suggest a more generic name of `RichTooltipRow` as it is not instance specific and can be reused once we come up with other rich tooltips.

src/components/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,158 @@
...
+  }
+  const ipv4Addresses = instanceData ? getAddresses(instanceData, "inet") : [];
+  const ipv6Addresses = instanceData ? getAddresses(instanceData, "inet6") : [];
+  const macAddresses = instanceData
+    ? getInstanceMacAddresses(instanceData)
+    : [];
+
+  return (
+    <table className="u-table-layout--auto u-no-margin--bottom instance-rich-tooltip-table">
+      <tbody>
```
edlerd: I think we should split this component into a `RichTooltipMessage` that stays under the `src/components` which has the `rich-tooltip-table` class. It can be reused by other rich tooltips. The Instance specific functionality should move into the `pages/instances` tree and call this component as well as the `RichTooltipRow`.

src/components/RichTooltipRow.tsx
```diff
@@ -0,0 +1,23 @@
...
+  valueTitle,
+  className,
+}) => {
+  return (
+    <tr className={className}>
+      <th className="u-text--muted">{title}</th>
+      <td title={valueTitle} className="u-truncate">
+        {value}
+      </td>
+    </tr>
```
edlerd, line 21: The distribution of space between the columns shifts when loading. Can we give those fixed widths so there is less flicker?

https://github.com/user-attachments/assets/37cb02b9-7cca-401d-afe1-b882059b5193

src/sass/_instance_rich_tooltip.scss
```diff
@@ -0,0 +1,35 @@
+body.is-dark .instance-rich-tooltip-table .p-icon--status-queued-small {
```
edlerd: Classes and filename should be renamed to make them reusable for other entities and not instance specific.

edlerd (commented):
Visual and code looks pretty nice. Some thoughts on improving structure and more alignment in naming below.

src/pages/instances/InstanceLinkChip.tsx
omarelkashef: I agree with having this component only accept proper LxdInstance. But do we not need a name that shows  the component has a tooltip? Is InstanceChip clear enough?

edlerd: We can start calling it `InstanceRichChip` and `InstanceRichChipLazy`. Wdyt?

src/util/instances.tsx
```diff
@@ -18,18 +18,15 @@ export const instanceLinkFromOperation = (args: {
   project?: string;
   instanceType: InstanceIconType;
 }): ReactNode | undefined => {
-  const { operation, project, instanceType } = args;
+  const { operation, project } = args;
```
edlerd, line 20: Let's also remove the `instanceType` from the function parameters and stop passing it in from the two usages of it.

src/pages/instances/actions/StopInstanceBtn.tsx
```diff
@@ -101,7 +101,7 @@ const StopInstanceBtn: FC<Props> = ({ instance }) => {
           ? "Stop"
           : "You do not have permission to stop this instance",
       }}
-      className="has-icon is-dense"
+      className="has-icon is-dense instance-action-button"
```
edlerd:
```suggestion
      className="has-icon is-dense"
```
I think we don't need this any more.

src/sass/instance_rich_tooltip.scss
```diff
@@ -0,0 +1,17 @@
+.instance-rich-tooltip-table {
+  .actions-list {
+    margin: 0;
+  }
+
+  .instance-action-button {
+    margin-bottom: 0;
+  }
+
```
edlerd:
```suggestion
```
can be removed I guess

src/sass/rich_tooltip.scss
```diff
@@ -0,0 +1,22 @@
...
+  margin: 0;
+  min-height: 400px;
+  width: 425px;
+
+  td {
+    max-width: 300px;
+    width: 300px;
+  }
+
+  th {
```
edlerd:
```suggestion
  .rich-tooltip-value {
    max-width: 300px;
    width: 300px;
  }

  .rich-tooltip-heading {
```
Don't target elements without a class. Better introduce a custom class for it. And for naming best use something close to the [BEM / Block Element Modifier](https://en.bem.info/methodology/quick-start/) methodology.

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -17,6 +17,7 @@ const StartInstanceBtn: FC<Props> = ({ instance }) => {
     <Button
       appearance="base"
       hasIcon
+      className="instance-action-button"
```
edlerd:
```suggestion
```

src/pages/instances/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      title: "Description",
+      value: instanceData?.description || "-",
+      valueTitle: instanceData?.description || "-",
+    },
+    {
+      title: "Type",
+      value: instanceData
+        ? instanceCreationTypes.filter(
+            (item) => item.value === instanceData.type,
+          )[0].label
```
edlerd: Shall we introduce a helper for this and remove the other similar usages? Something like `getType(instance)`. We should also use the "better" implementation with find:
```
instanceCreationTypes.find(
                    (item) => item.value === instance.type,
                  )?.label
                  
```

src/pages/instances/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,150 @@
...
+    ? instance.name
+    : (props as InstanceRichChipLookupProps).instanceName;
+
+  const isQueryEnabled = !instance;
+  const { data: fetchedInstance, isLoading } = useInstance(
+    instanceName,
+    projectName,
+    isQueryEnabled,
+  );
+  const instanceData = instance || fetchedInstance;
```
edlerd:
```suggestion
  const hasResolvedInstance = (
    props: InstanceRichChipProps,
  ): props is InstanceRichChipResolvedProps => {
    return "instance" in props;
  };

  const hasPreloaded = hasResolvedInstance(props);
  const instanceName = hasPreloaded ? props.instance.name : props.instanceName;
  const projectName = hasPreloaded ? props.instance.project : props.projectName;
  const hasQuery = !hasPreloaded;
  const { data, isLoading } = useInstance(instanceName, projectName, hasQuery);
  const instance = hasPreloaded ? props.instance : data;
```
I'd suggest a helper function to narrow down the type in favour of non-guarded casting.

Also better name the "final" object instance, so the code below gets a bit easier to read.

src/pages/instances/InstanceRichChip.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const instance = (props as InstanceRichChipResolvedProps).instance;
+
+  const projectName =
+    (instance
+      ? instance.project
+      : (props as InstanceRichChipLookupProps).projectName) || "default";
+
+  const instanceName = instance
+    ? instance.name
+    : (props as InstanceRichChipLookupProps).instanceName;
```
edlerd: We probably should export the helper `const hasResolvedInstance` to be reused here.

src/util/instances.tsx
```diff
@@ -4,32 +4,28 @@ import type { ReactNode } from "react";
...
   }
   return (
-    <InstanceLinkChip
-      instance={{
-        name: linkText,
-        project: project || "default",
-        type: instanceType,
-      }}
+    <InstanceRichChip
+      instanceName="linkText"
```
```suggestion
      instanceName={linkText}
```
this should be the variable.

src/util/instances.tsx
```diff
@@ -100,3 +96,24 @@ export const getImageLink = (instance: LxdInstance) => {
...
+      key.split(".").length === 3 &&
+      value
+    ) {
+      hwaddrs.push(value);
+    }
+  }
+  return hwaddrs;
+};
+
+export const getInstanceType = (instance: LxdInstance) => {
```
edlerd, line 116: There is one more occasion, in `InstanceList.tsx` calling `instanceCreationTypes.find`, that can use this function.

src/util/instances.tsx
```diff
@@ -4,32 +4,28 @@ import type { ReactNode } from "react";
...
 export const CLUSTER_GROUP_PREFIX = "@";
 
 export const instanceLinkFromOperation = (args: {
   operation?: LxdOperationResponse;
   project?: string;
-  instanceType: InstanceIconType;
 }): ReactNode | undefined => {
-  const { operation, project, instanceType } = args;
+  const { operation, project } = args;
   const linkText = getInstanceName(operation?.metadata);
```
```suggestion
  const instanceName = getInstanceName(operation?.metadata);
```

src/pages/instances/InstanceRichTooltipMessage.tsx
```diff
@@ -0,0 +1,115 @@
...
+import InstanceStateActions from "pages/instances/actions/InstanceStateActions";
+import { type TooltipRow } from "../../components/RichTooltipRow";
+import { RichTooltipTable } from "../../components/RichTooltipTable";
+
+interface Props {
+  instanceName: string;
+  projectName: string;
+}
+
+export const InstanceRichTooltipMessage: FC<Props> = ({
```
edlerd:
```suggestion
export const InstanceRichTooltip: FC<Props> = ({
```
I think this can be a bit shorter and still be concise. Should rename the file as well.

src/pages/instances/InstanceRichChip.tsx
```diff
@@ -0,0 +1,45 @@
...
+import ResourceLink from "../../components/ResourceLink";
+import { InstanceRichTooltipMessage } from "./InstanceRichTooltipMessage";
+
+interface Props {
+  instanceName: string;
+  projectName: string;
+}
+
+export const InstanceRichChip: FC<Props> = ({ instanceName, projectName }) => {
+  const showTooltip = !useIsScreenBelow(mediumScreenBreakpoint, "height");
```
edlerd: There is this edge case where the tooltip is only half visible

image

src/context/appProviders.tsx
```diff
@@ -45,27 +45,27 @@ const AppProviders: FC<AppProvidersProps> = ({ children }) => {
...
-    <ToastNotificationProvider>
-      <NotificationProvider state={notificationState} pathname={pathname}>
-        <QueryClientProvider client={queryClient}>
-          <AuthProvider>
-            <ProjectProvider>
-              <OperationsProvider>
-                <InstanceLoadingProvider>
-                  <MemberLoadingProvider>
-                    <EventQueueProvider>
+    <QueryClientProvider client={queryClient}>
```
edlerd, line 48:
```suggestion
    <QueryClientProvider client={queryClient}>
      <AuthProvider>
        <ProjectProvider>
          <OperationsProvider>
            <InstanceLoadingProvider>
              <MemberLoadingProvider>
                <EventQueueProvider>
                  <ModalProvider>
                    <ToastNotificationProvider>
                      <NotificationProvider
                        state={notificationState}
                        pathname={pathname}
                      >
                        <MetricHistoryProvider>
                          {children}
                        </MetricHistoryProvider>
                      </NotificationProvider>
                    </ToastNotificationProvider>
                  </ModalProvider>
                </EventQueueProvider>
              </MemberLoadingProvider>
            </InstanceLoadingProvider>
          </OperationsProvider>
        </ProjectProvider>
      </AuthProvider>
    </QueryClientProvider>
```

The order here is fairly important. We should make sure the tooltip from a notification can also access the instance loading state and trigger a modal.

I think keeping the old order and just moving the notification providers down in the hierarchy might be the safest way.

src/pages/instances/InstanceRichChip.tsx
edlerd: Instead of hiding right away, I think we should define another breakpoint where we hide the addresses but show the smaller tooltip.

src/pages/instances/InstanceList.tsx
```diff
@@ -459,15 +458,7 @@ const InstanceList: FC = () => {
...
-            content: (
-              <>
-                {
-                  instanceCreationTypes.find(
-                    (item) => item.value === instance.type,
-                  )?.label
-                }
-              </>
-            ),
+            content: <>{getInstanceType(instance)}</>,
```
```suggestion
            content: getInstanceType(instance)
```

src/pages/instances/InstanceRichChip.tsx
```diff
@@ -0,0 +1,45 @@
...
+import ResourceLink from "../../components/ResourceLink";
+import { InstanceRichTooltip } from "./InstanceRichTooltip";
+
+interface Props {
+  instanceName: string;
+  projectName: string;
+}
+
+export const InstanceRichChip: FC<Props> = ({ instanceName, projectName }) => {
+  const showTooltip = !useIsScreenBelow(mediumScreenBreakpoint, "height");
```
edlerd:
```suggestion
  const showTooltip = !useIsScreenBelow(500, "height");
```
The smallest version of the tooltip is only 220px height, so should fit a more narrow screen as well from the middle of the screen. We are hiding it a bit too early now.

src/pages/instances/InstanceRichTooltip.tsx
```diff
@@ -0,0 +1,121 @@
...
+        value: ipv4Addresses.length ? (
+          <InstanceListAddresses addresses={ipv4Addresses} numberToShow={1} />
+        ) : (
+          "-"
+        ),
+      },
+      {
+        title: "IPV6",
+        value: ipv6Addresses.length ? (
+          <InstanceListAddresses addresses={ipv6Addresses} numberToShow={1} />
```
edlerd, line 90: There is a bit too much space for the ipv6 row:

image

src/pages/instances/InstanceRichTooltip.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const getAddresses = (instance: LxdInstance, family: IpFamily) => {
+    const addresses = getIpAddresses(instance, family);
+    const sortedAddresses =
+      family === "inet6" ? sortIpv6Addresses(addresses) : addresses;
+    return sortedAddresses.map((item) => item.address);
+  };
+
+  const { data: instance, isLoading } = useInstance(instanceName, projectName);
+  const showAddresses = !useIsScreenBelow(900, "height");
+  const isLargeScreen = !useIsScreenBelow(largeScreenBreakpoint, "height");
```
edlerd: How do the breakpoints come up? I think these should be half the height with a bit of buffer compared to the max possible height of the tooltip. Where max height is considering all three addresses spill into a 2nd line due to overflow.

src/sass/rich_tooltip.scss
```diff
@@ -0,0 +1,26 @@
...
+  @include vf-icon-status-small($colors--dark-theme--icon);
+}
+
+.rich-tooltip-table {
+  margin: 0;
+  min-height: 400px;
+  width: 430px;
+
+  @media (height <= 900px) {
+    min-height: 200px;
```
edlerd: The min height here relates to the breakpoints in `InstanceRichTooltip`? Maybe this should be in the `instance-rich-tooltip-table` class and not here? Also there should be a comment on the scss and on the breakpoint to hint these have to be in sync.

omarelkashef: I think all tooltips should collapse rows when screen is small, so this concerns the generic component.

edlerd: Sounds ok, shall we define those constants in `RichTooltipTable.tsx` with a comment hinting at the SCSS. And a comment in here hinting at the constants. Then the instance implementation can reference the constants. And other future rich chips can reuse the breakpoints.

omarelkashef: Yes, good idea!

edlerd (commented):
QA LGTM. Code also mostly LGTM, only the breakpoints need some tweaking. I think we need a third case for min-height <900, <1100 and >1100. And we should use constants with comments referencing each other to ease future maintenance.

src/sass/rich_tooltip.scss
```diff
@@ -0,0 +1,33 @@
+$medium-tooltip-breakpoint: 900px;
+$large-tooltip-breakpoint: 1100px;
```
edlerd, line 3: Can we add a comment here as well so that anyone changing gets reminded to also check the JS variables?

edlerd (dismissed):
QA and code LGTM.
Thanks for the iterations, good work :)

src/pages/instances/forms/UploadExternalFormatFileForm.tsx
```diff
@@ -110,7 +110,10 @@ const UploadExternalFormatFileForm: FC<Props> = ({
     const message = (
       <>
         Created instance{" "}
-        <ResourceLink type="instance" value={instanceName} to={instanceUrl} />.
+        <InstanceRichChip
+          instanceName={instanceName}
+          projectName={project?.name ?? ""}
+        />
```
edlerd, line 116: This is dropping the . and breaking tests.
```suggestion
        <InstanceRichChip
          instanceName={instanceName}
          projectName={project?.name ?? ""}
        />.
```

src/pages/instances/forms/UploadInstanceBackupFileForm.tsx
```diff
@@ -67,7 +67,10 @@ const UploadInstanceBackupFileForm: FC<Props> = ({
     const message = (
       <>
         Created instance{" "}
-        <ResourceLink type="instance" value={instanceName} to={instanceUrl} />.
+        <InstanceRichChip
+          instanceName={instanceName}
+          projectName={project?.name ?? ""}
+        />
```
edlerd, line 73:
```suggestion
        <InstanceRichChip
          instanceName={instanceName}
          projectName={project?.name ?? ""}
        />.
```

PR #1668, opened 2025-12-08
feat(Error screens): Harmonize not found screens

kimanhou:
## Done

- Design a screen for when an object is not found
- Use it for instance, profile, storage bucket, storage pool, storage volume, network ACL and network
- Replace ProjectNotFound
- Don't retry when 404 error. The other errors should still retry 3 times.

## QA

1. Run the LXD-UI:
    - Go to a URL with an instance that doesn't exist, e.g. `/ui/project/default/instance/i-dont-exist `
    - Go to a URL with a profile that doesn't exist, e.g. `/ui/project/default/profile/i-dont-exist `
    - Go to a URL with a storage bucket that doesn't exist, e.g. `/ui/project/default/storage/pool/default/bucket/i-dont-exist `
    - Go to a URL with a storage pool that doesn't exist, e.g. `/ui/project/default/storage/pool/i-dont-exist`
    - Go to a URL with a storage volume that doesn't exist, e.g. `/ui/project/default/storage/pool/Ponderous-Panda/volumes/custom/i-dont-exist (`Ponderous-Panda` is a pool that exists !)
    - Go to a URL with a network ACL that doesn't exist, e.g. `/ui/project/default/network-acl/i-dont-exist`
    - Go to a URL with a network that doesn't exist, e.g. `/ui/project/default/network/i-dont-exist`
    - Go to a URL with a project that doesn't exist, e.g. `/ui/project/i-dont-exist/instances` 

## Screenshots

**Instance**
image

**Profile**
new-profile-with-error

**Storage bucket**
new-storage-bucket-with-error

**Storage pool**
new-storage-pool-with-error

**Storage volume**
new-storage-volume

**Network ACL**
new-network-acl

**Network**
new-network

**Project**
image


src/components/NotFound.tsx
```diff
@@ -0,0 +1,39 @@
...
+import ResourceIcon from "components/ResourceIcon";
+
+interface Props {
+  entityType: string;
+  errorMessage?: string;
+}
+
+const NotFound: FC<Props> = ({ entityType, errorMessage }) => {
+  const url = location.pathname;
+  const entityName = url.split(`/${entityType}/`)[1];
```
omarelkashef: This might not be enough if the user enters a url that has  segments after the entity such as /ui/project/default/instance/i-donot-exist/configuration. Something like const entityName = (url.split(`/${entityType}/`)[1]).split('/')[0]; should work.

src/components/NotFound.tsx
```diff
@@ -0,0 +1,39 @@
...
+const NotFound: FC<Props> = ({ entityType, errorMessage }) => {
+  const url = location.pathname;
+  const entityName = url.split(`/${entityType}/`)[1].split("/")[0];
+  const entityLabel = entityType.replace(/-/g, " ");
+
+  return (
+    <Row className="empty-state u-no-margin--left">
+      <Col size={4} className="u-align--right col-4 col-medium-2 col-small-1">
+        <ResourceIcon
+          type={entityType as ResourceIconType}
```
edlerd:
```suggestion
          type={entityType}
```
We should type the prop above correctly and avoid this cast.

src/components/NotFound.tsx
edlerd: I think we should better pass in the name as a parameter. Or not parrot it back to the user at all and just mention the type.

src/components/NotFound.tsx
edlerd: This one seems to be unresolved.

src/components/NotFound.tsx
```diff
@@ -0,0 +1,38 @@
...
+      </Col>
+      <Col size={8} className="u-align--left col-8 col-medium-4 col-small-3">
+        <p className="p-heading--4 u-no-margin--bottom">
+          {capitalizeFirstLetter(entityLabel)} not found
+        </p>
+        <p>
+          The {entityLabel} {entityName && <code>{entityName}</code>} is missing
+          or you do not have the <code>viewer</code> permission for it.
+        </p>
+        {errorMessage && <p className="u-text--muted">Error: {errorMessage}</p>}
```
edlerd: The muted font seems a bit too light for the error message. Should we use notification with severity negative instead?

src/components/NotFound.tsx
```diff
@@ -0,0 +1,38 @@
...
+import type { ResourceIconType } from "components/ResourceIcon";
+import ResourceIcon from "components/ResourceIcon";
+
+interface Props {
+  entityType: string;
+  entityName?: string;
+  errorMessage?: string;
+}
+
+const NotFound: FC<Props> = ({ entityType, entityName, errorMessage }) => {
```
edlerd, line 15: There is also the `ProjectNotFound` component, we can probably remove it or call this one from it.

src/context/appProviders.tsx
```diff
@@ -14,8 +14,25 @@ import OperationsProvider from "context/operationsProvider";
 import { MetricHistoryProvider } from "context/metricHistory";
 import { MemberLoadingProvider } from "context/memberLoading";
 import { ModalProvider } from "context/useModal";
+import { LxdApiError } from "util/helpers";
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({
```
edlerd, line 19: This is great, thanks for adding it!

Please, can you confirm that other errors like 5xx or alike are still handled correctly with 3 retries and back off?

I am not aware of a way to trigger any other error than 404 though, so maybe switch the 404 in here for a 403 locally to check if it triggers the retries on 404 then.

edlerd (commented):
good progress on this one :+1:

src/components/NotFound.tsx
kimanhou: Sorry, it got lost in a force push

src/components/NotFound.tsx
kimanhou: Does it look weird ? I think it looks weird. Let me think about it
image

src/components/NotFound.tsx
kimanhou: Good idea !
image

src/context/appProviders.tsx
kimanhou: I confirm
image

src/components/NotFound.tsx
kimanhou: Code snippet block ?

image

edlerd: I think notification is the right component. Maybe `borderless`?

kimanhou: Then it just looks like an icon with a text

image

edlerd: Can you ask in ~design about it? This seems more in line with design guidelines. The heading should probably have the same distance to the first paragraph than above the notification text.

Also, the indentation of the text could be the same between heading, paragraph and notification. But then we have two icons to handle. So not sure about the best way to arrange it.

kimanhou: Done. Let's see what feedback we receive
https://chat.canonical.com/canonical/pl/m1mgzcamfprx7y8tyj7aka9zfe

src/components/NotFound.tsx
```diff
@@ -0,0 +1,39 @@
...
+        </p>
+        <p>
+          The {entityLabel} {entityName && <code>{entityName}</code>} is missing
+          or you do not have the <code>viewer</code> permission for it.
+        </p>
+        {errorMessage && (
+          <Notification severity="negative" borderless>
+            <code className="no-background">{errorMessage}</code>
+          </Notification>
+        )}
```
edlerd: I wonder if the messaging here is right. This also is invoked when there are other errors like 5xx from the server. Shall we make the copy a bit more generic like "Instance foo could not be loaded" as title and below "The instance is missing or you do not have the viewer permission on it". Below that, the error message from api. Wdyt?

src/sass/_not_found.scss
```diff
@@ -0,0 +1,30 @@
+.not-found {
```
edlerd, line 1: Shall we adjust the spacing on very small screens a bit?

Screenshot from 2025-12-11 14-38-52

kimanhou: Sure, what do you think ?

image

src/components/NotFound.tsx
kimanhou: Sure !

image

src/sass/_not_found.scss
edlerd: I think left aligning the heading might be easier to read. Padding now looks better.

PR #1669, opened 2025-12-09
fix: [WD-31443] total memory in cluster

### LINKED ISSUE #1625: "Total memory" incorrect in a cluster
Hey again!

Image

It says I have 24GiB total memory, but this is a 4 node cluster and they each have 24GiB memory. I think perhaps the wording should be different for a cluster.

LXD 6.5

omarelkashef:
## Done

- Fixes available memory in cluster

Fixes #1625 
## QA

1. Run the LXD-UI:
    - Go to Create an instance page and then to Resource Limits tab
    - Create an override for Memory Limit
    - Ensure available memory is reported correctly as:
           (i) a value if you selected a cluster member as target
           (ii) a value if you have not selected a cluster member as target and all the cluster members have equal limits
           (iii) a range if you have not selected a cluster member as target and the cluster members do not have equal limits
     - Go to one instance -> Configuration -> Resource Limits
      - Create an override for Memory Limit
      - Ensure available memory is reported correctly as a value

## Screenshots

image

image

### DIFF
--- a/src/api/server.tsx
+++ b/src/api/server.tsx
@@ -110,6 +110,24 @@ export const fetchResources = async (
     });
 };
 
+export const fetchClusterMembersResources = async (
+  clusterMembers: LxdClusterMember[],
+): Promise<LxdResources[]> => {
+  const results = await Promise.allSettled(
+    clusterMembers.map(async (member) => fetchResources(member.server_name)),
+  );
+  const data: LxdResources[] = [];
+  for (let i = 0; i < clusterMembers.length; i++) {
+    const member = clusterMembers[i];
+    const result = results[i];
+    if (result.status === "rejected") {
+      throw constructMemberError(result, member.server_name);
+    }
+    data.push(result.value);
+  }
+  return data;
+};
+
 export const fetchConfigOptions = async (
   hasMetadataConfiguration: boolean,
 ): Promise<LxdMetadata | null> => {
--- a/src/components/forms/DiskSizeSelector.tsx
+++ b/src/components/forms/DiskSizeSelector.tsx
@@ -76,12 +76,11 @@ const DiskSizeSelector: FC<Props> = ({
           title={disabledReason}
         />
       </div>
-      {(help || helpTotal) && (
-        <p className="p-form-help-text">
-          {help}
-          {helpTotal}
-        </p>
-      )}
+      <p className="p-form-help-text">
+        {help}
+        {help && helpTotal && <br />}
+        {helpTotal}
+      </p>
     </div>
   );
 };
--- a/src/components/forms/MemoryLimitAvailable.tsx
+++ b/src/components/forms/MemoryLimitAvailable.tsx
@@ -1,59 +1,138 @@
 import type { FC } from "react";
 import { useNotify, Spinner } from "@canonical/react-components";
-import { useQuery } from "@tanstack/react-query";
-import { fetchResources } from "api/server";
-import { queryKeys } from "util/queryKeys";
 import { humanFileSize } from "util/helpers";
 import { limitToBytes } from "util/limits";
 import type { LxdProject } from "types/project";
-import { useServerEntitlements } from "util/entitlements/server";
+import { Icon } from "@canonical/react-components";
+import type { InstanceAndProfileFormikProps } from "./instanceAndProfileFormValues";
+import type { CreateInstanceFormValues } from "pages/instances/CreateInstance";
+import type { EditInstanceFormValues } from "pages/instances/EditInstance";
+import { CLUSTER_GROUP_PREFIX } from "util/instances";
+import ResourceLink from "components/ResourceLink";
+import { useIsClustered } from "context/useIsClustered";
+import { useResources } from "context/useResources";
 
+interface MemoryLimit {
+  min: number;
+  max: number;
+  source: "project" | "cluster-member" | "";
+}
 interface Props {
+  formik?: InstanceAndProfileFormikProps;
   project?: LxdProject;
 }
 
-const MemoryLimitAvailable: FC<Props> = ({ project }) => {
+const MemoryLimitAvailable: FC<Props> = ({ formik, project }) => {
   const notify = useNotify();
-  const { canViewResources } = useServerEntitlements();
+  const isClustered = useIsClustered();
+  const values = formik?.values;
+  const projectResourceMemory = project?.config["limits.memory"]
+    ? limitToBytes(project?.config["limits.memory"])
+    : undefined;
+  const { target } = values as CreateInstanceFormValues;
+  const { location } = values as EditInstanceFormValues;
+  const isClusterGroupTarget = target?.startsWith(CLUSTER_GROUP_PREFIX);
+  const validTarget = target && !isClusterGroupTarget ? target : undefined;
+  const clusterMember = isClustered ? (validTarget ?? location) : undefined;
 
-  const {
-    data: resources,
-    error,
-    isLoading,
-  } = useQuery({
-    queryKey: [queryKeys.resources],
-    queryFn: async () => fetchResources(),
-    enabled: canViewResources(),
-  });
+  const { data: resources, error, isLoading } = useResources(clusterMember);
 
   if (isLoading) {
     return <Spinner className="u-loader" text="Loading resources..." />;
   }
-
   if (error) {
     notify.failure("Loading resources failed", error);
+    return null;
+  }
+  if (!resources) {
+    return null;
   }
 
-  const getAvailableMemory = () => {
-    if (!project?.config["limits.memory"]) {
-      return resources?.memory.total;
+  const getSourceQualifier = (source: string) => {
+    if (source === "project") {
+      return (
+        <>
+          {"in project "}
+          <ResourceLink
+            type="project"
+            value={project?.name || "default"}
+            to={`/ui/projects/${encodeURIComponent(project?.name || "default")}`}
+          />
+        </>
+      );
     }
-    if (!resources?.memory.total) {
-      return limitToBytes(project.config["limits.memory"]);
+
+    if (source === "cluster-member") {
+      return (
+        <>
+          {"on cluster member "}
+          <ResourceLink
+            type="cluster-member"
+            value={clusterMember ?? ""}
+            to={`/ui/cluster/member/${encodeURIComponent(clusterMember ?? "")}`}
+          />
+        </>
+      );
     }
-    return Math.min(
-      resources.memory.total,
-      Number(limitToBytes(project.config["limits.memory"])),
-    );
+    return null;
   };
 
-  const maxMemory = getAvailableMemory();
+  const getAvailableMemory = (): MemoryLimit => {
+    const resourceArray = Array.isArray(resources) ? resources : [resources];
+    const projectLimit = Number(projectResourceMemory);
+    const resourcesMemories = resourceArray.map((r) => r.memory.total);
+    const minResourceMemory = Math.min(...resourcesMemories);
+    const maxResourceMemory = Math.max(...resourcesMemories);
 
-  return maxMemory ? (
+    if (resourceArray.length === 0 || projectLimit < minResourceMemory) {
+      return {
+        min: projectLimit,
+        max: projectLimit,
+        source: projectLimit ? "project" : "",
+      };
+    }
+
+    if (!projectResourceMemory || maxResourceMemory < projectLimit) {
+      return {
+        min: minResourceMemory,
+        max: maxResourceMemory,
+        source: clusterMember ? "cluster-member" : "",
+      };
+    }
+
+    return {
+      min: minResourceMemory,
+      max: projectLimit,
+      source: "",
+    };
+  };
+
+  const { min: minMemory, max: maxMemory, source } = getAvailableMemory();
+  const showHelpIcon = minMemory !== maxMemory;
+  const helpIconText = `The available memory depends on the cluster member that the instance will be created on.\
+  To determine the available memory exactly, go to the "Main configuration" section, and from\
+  the "Target" dropdown, select the "Cluster Member" option, then choose a member.`;
+
+  return (
     <>
-      Total memory: <b>{humanFileSize(maxMemory)}</b>
+      Total memory:{" "}
+      <b>
+        {humanFileSize(minMemory)}
+        {minMemory !== maxMemory && ` - ${humanFileSize(maxMemory)}`}
+        {showHelpIcon && (
+          <>
+            {" "}
+            <Icon
+              name="information"
+              className="help-link-icon"
+              title={helpIconText}
+            />
+          </>
+        )}
+      </b>
+      <br />
+      {getSourceQualifier(source)}
     </>
-  ) : null;
+  );
 };
-
 export default MemoryLimitAvailable;
--- a/src/components/forms/MemoryLimitSelector.tsx
+++ b/src/components/forms/MemoryLimitSelector.tsx
@@ -4,13 +4,19 @@ import type { MemoryLimit } from "types/limits";
 import { BYTES_UNITS, MEM_LIMIT_TYPE } from "types/limits";
 import MemoryLimitAvailable from "components/forms/MemoryLimitAvailable";
 import { useCurrentProject } from "context/useCurrentProject";
+import type { InstanceAndProfileFormikProps } from "./instanceAndProfileFormValues";
 
 interface Props {
+  formik: InstanceAndProfileFormikProps;
   memoryLimit?: MemoryLimit;
   setMemoryLimit: (memoryLimit: MemoryLimit) => void;
 }
 
-const MemoryLimitSelector: FC<Props> = ({ memoryLimit, setMemoryLimit }) => {
+const MemoryLimitSelector: FC<Props> = ({
+  formik,
+  memoryLimit,
+  setMemoryLimit,
+}) => {
   const { project } = useCurrentProject();
 
   if (!memoryLimit) {
@@ -58,7 +64,7 @@ const MemoryLimitSelector: FC<Props> = ({ memoryLimit, setMemoryLimit }) => {
             setMemoryLimit({ ...memoryLimit, value: +e.target.value });
           }}
           value={`${memoryLimit.value ? memoryLimit.value : ""}`}
-          help={<MemoryLimitAvailable project={project} />}
+          help={<MemoryLimitAvailable project={project} formik={formik} />}
         />
       )}
       {memoryLimit.selectedType === MEM_LIMIT_TYPE.FIXED && (
@@ -74,7 +80,7 @@ const MemoryLimitSelector: FC<Props> = ({ memoryLimit, setMemoryLimit }) => {
               setMemoryLimit({ ...memoryLimit, value: +e.target.value });
             }}
             value={`${memoryLimit.value ? memoryLimit.value : ""}`}
-            help={<MemoryLimitAvailable project={project} />}
+            help={<MemoryLimitAvailable project={project} formik={formik} />}
           />
           <Select
             id="memUnitSelect"
--- a/src/context/useResources.tsx
+++ b/src/context/useResources.tsx
@@ -1,0 +1,28 @@
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { useServerEntitlements } from "util/entitlements/server";
+import { useIsClustered } from "./useIsClustered";
+import { fetchResources, fetchClusterMembersResources } from "api/server";
+import { useClusterMembers } from "./useClusterMembers";
+
+export const useResources = (currentServerInCluster?: string) => {
+  const isClustered = useIsClustered();
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const { canViewResources } = useServerEntitlements();
+  const singleNodeQuery = useQuery({
+    queryKey: [queryKeys.resources],
+    queryFn: async () => fetchResources(),
+    enabled: canViewResources() && !isClustered,
+  });
+  const filteredClusterMembers = clusterMembers.filter(
+    (member) =>
+      !currentServerInCluster || member.server_name === currentServerInCluster,
+  );
+  const clusterQuery = useQuery({
+    queryKey: [queryKeys.resources, filteredClusterMembers],
+    queryFn: async () => fetchClusterMembersResources(filteredClusterMembers),
+    enabled: canViewResources() && isClustered,
+  });
+
+  return isClustered ? clusterQuery : singleNodeQuery;
+};
--- a/src/components/forms/ResourceLimitsForm.tsx
+++ b/src/components/forms/ResourceLimitsForm.tsx
@@ -76,6 +76,7 @@ const ResourceLimitsForm: FC<Props> = ({ formik }) => {
             memoryLimitToPayload(val as MemoryLimit | undefined) ?? "",
           children: (
             <MemoryLimitSelector
+              formik={formik}
               memoryLimit={formik.values.limits_memory}
               setMemoryLimit={(memoryLimit) =>
                 void formik.setFieldValue("limits_memory", memoryLimit)
--- a/src/pages/instances/forms/InstanceTargetSelect.tsx
+++ b/src/pages/instances/forms/InstanceTargetSelect.tsx
@@ -20,6 +20,7 @@ import { useClusterMembers } from "context/useClusterMembers";
 import type { SelectRef } from "@canonical/react-components/dist/components/CustomSelect/CustomSelect";
 import PlacementGroupSelect from "pages/instances/forms/PlacementGroupSelect";
 import { PLACEMENT_GROUP_POLICY_COMPACT } from "pages/placement-groups/PlacementGroupForm";
+import { CLUSTER_GROUP_PREFIX } from "util/instances";
 
 export const TARGET = {
   AUTO: "auto",
@@ -27,8 +28,6 @@ export const TARGET = {
   CLUSTER_MEMBER: "clusterMember",
   PLACEMENT_GROUP: "placementGroup",
 };
-
-export const CLUSTER_GROUP_PREFIX = "@";
 
 interface Props {
   formik: FormikProps<CreateInstanceFormValues>;
--- a/src/pages/projects/ProjectMemoryLimit.tsx
+++ b/src/pages/projects/ProjectMemoryLimit.tsx
@@ -1,0 +1,30 @@
+import { useResources } from "context/useResources";
+import type { FC } from "react";
+import { humanFileSize } from "util/helpers";
+import { useNotify } from "@canonical/react-components";
+import { Spinner } from "@canonical/react-components";
+
+export const ProjectMemoryLimit: FC = () => {
+  const notify = useNotify();
+
+  const { data: resources = [], error, isLoading } = useResources();
+  if (isLoading) {
+    return <Spinner className="u-loader" text="Loading resources..." />;
+  }
+  if (error) {
+    notify.failure("Loading resources failed", error);
+    return null;
+  }
+  const resourceArray = Array.isArray(resources) ? resources : [resources];
+
+  const sumMemoryLimit = resourceArray.reduce(
+    (sum, resource) => sum + resource.memory.total,
+    0,
+  );
+
+  return (
+    <>
+      Total memory: <b>{humanFileSize(sumMemoryLimit)}</b>
+    </>
+  );
+};
--- a/src/pages/projects/forms/ProjectResourceLimitsForm.tsx
+++ b/src/pages/projects/forms/ProjectResourceLimitsForm.tsx
@@ -8,7 +8,7 @@ import DiskSizeSelector from "components/forms/DiskSizeSelector";
 import { getProjectKey } from "util/projectConfigFields";
 import type { LxdConfigPair } from "types/config";
 import CpuLimitInput from "components/forms/CpuLimitInput";
-import MemoryLimitAvailable from "components/forms/MemoryLimitAvailable";
+import { ProjectMemoryLimit } from "../ProjectMemoryLimit";
 
 export interface ProjectResourceLimitsFormValues {
   limits_instances?: number;
@@ -109,7 +109,7 @@ const ProjectResourceLimitsForm: FC<Props> = ({ formik }) => {
               setMemoryLimit={(val?: string) =>
                 void formik.setFieldValue("limits_memory", val)
               }
-              helpTotal={<MemoryLimitAvailable />}
+              helpTotal={<ProjectMemoryLimit />}
             />
           ),
         }),
--- a/src/util/instances.tsx
+++ b/src/util/instances.tsx
@@ -11,6 +11,8 @@ import { useImagesInProject } from "context/useImages";
 import ResourceLabel from "components/ResourceLabel";
 import ResourceLink from "components/ResourceLink";
 
+export const CLUSTER_GROUP_PREFIX = "@";
+
 export const instanceLinkFromOperation = (args: {
   operation?: LxdOperationResponse;
   project?: string;


src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -33,27 +63,80 @@ const MemoryLimitAvailable: FC<Props> = ({ project }) => {
...
+  }
+  const maxMemory = getAvailableMemory(resources[0]);
+  return (
     <>
-      Total memory: <b>{humanFileSize(maxMemory)}</b>
+      {maxMemory && (
+        <>
+          Total memory:{" "}
+          <b>
+            {humanFileSize(maxMemory)} {!currentMember && "on current member"}
```
edlerd: I think we should mention the member here as a chip.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -33,27 +63,80 @@ const MemoryLimitAvailable: FC<Props> = ({ project }) => {
...
+    return (
+      <>
+        Total memory:{" "}
+        <b>
+          {humanFileSize(minMemory)} - {humanFileSize(maxMemory)}
+        </b>{" "}
+        <Icon
+          name="information"
+          className="help-link-icon"
+          title="The available memory depends on the member that the instance will be created on. You can also specify the member to determine available memory exactly."
```
edlerd: We should give exact hints how to "also specify the member".

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -33,27 +63,80 @@ const MemoryLimitAvailable: FC<Props> = ({ project }) => {
...
   };
 
-  const maxMemory = getAvailableMemory();
+  const getClusterAvailableMemoryRange = (resources: LxdResources[]) => {
+    const sortedResources = resources
+      .map((r) => r.memory.total)
+      .sort((a, b) => a - b);
+
+    if (!project?.config["limits.memory"]) {
+      return [sortedResources[0], sortedResources[sortedResources.length - 1]];
```
edlerd: Better introduce `const minResource = sortedResources[0]` and `maxResource` to reuse it here and below. Same for the `const projectResource = project.config["limits.memory"]`.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,30 +1,60 @@
...
-  const {
-    data: resources,
-    error,
-    isLoading,
-  } = useQuery({
-    queryKey: [queryKeys.resources],
-    queryFn: async () => fetchResources(),
-    enabled: canViewResources(),
+  const queries = useQueries({
+    queries: ((!isProfile && values ? clusterMembers : []) ?? [])
```
edlerd: Why the `!isProfile`?

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,30 +1,60 @@
...
+import { useClusterMembers } from "context/useClusterMembers";
+import { CLUSTER_GROUP_PREFIX } from "util/instances";
 
 interface Props {
+  formik?: InstanceAndProfileFormikProps;
   project?: LxdProject;
 }
 
-const MemoryLimitAvailable: FC<Props> = ({ project }) => {
+const MemoryLimitAvailable: FC<Props> = ({ formik, project }) => {
```
edlerd, line 25: The changes here break for the unclustered case, where there are no cluster members. We should activate the queries based on `const isClustered = useIsClustered();`.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
-    data: resources,
-    error,
-    isLoading,
-  } = useQuery({
-    queryKey: [queryKeys.resources],
-    queryFn: async () => fetchResources(),
-    enabled: canViewResources(),
+  const clusterMemberNames = clusterMembers.map((member) => member.server_name);
+  const queries = useQueries({
+    queries: (isClustered ? clusterMemberNames : [currentServer])
```
edlerd: The logic here seems off. Just because the environment is clustered shouldn't have an influence on whether the form at hand is targeting an instance for a specific cluster member.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
+      .filter(
+        (member) => !isClustered || !currentServer || member === currentServer,
+      )
+      .map((member) => ({
+        queryKey: [queryKeys.resources],
+        queryFn: async () => fetchResources(isClustered ? member : ""),
+        enabled: canViewResources(),
```
edlerd:
```suggestion
        enabled: canViewResources() && isClustered,
```
Here we can use the clustered variable to disable the query if not on a cluster.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
-      return resources?.memory.total;
+  const getAvailableMemory = (resource: LxdResources) => {
+    if (!projectResourceMemory) {
+      return resource?.memory.total;
     }
-    if (!resources?.memory.total) {
-      return limitToBytes(project.config["limits.memory"]);
+    if (!resource?.memory.total) {
+      console.log(limitToBytes(projectResourceMemory));
+
```
edlerd:
```suggestion
```

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
 }
 
-const MemoryLimitAvailable: FC<Props> = ({ project }) => {
+const MemoryLimitAvailable: FC<Props> = ({ formik, project }) => {
   const notify = useNotify();
   const { canViewResources } = useServerEntitlements();
+  const isClustered = useIsClustered();
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const values = formik?.values;
+  const projectResourceMemory = project?.config["limits.memory"];
```
edlerd: I think this should directly do the conversion with `Number(limitToBytes` to simplify below code and avoid repetition.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
   }
-
   if (error) {
     notify.failure("Loading resources failed", error);
   }
-
-  const getAvailableMemory = () => {
-    if (!project?.config["limits.memory"]) {
-      return resources?.memory.total;
+  const getAvailableMemory = (resource: LxdResources) => {
```
edlerd: I think we should evolve the name a bit to make clear this function is for a targeted cluster member / unclustered case.

src/components/forms/DiskSizeSelector.tsx
```diff
@@ -78,8 +78,7 @@ const DiskSizeSelector: FC<Props> = ({
       </div>
       {(help || helpTotal) && (
         <p className="p-form-help-text">
-          {help}
-          {helpTotal}
+          {help} {helpTotal}
```
edlerd: Why did this change? Should it be reverted?

src/components/forms/MemoryLimitAvailable.tsx
edlerd: We usually push this logic down into the api layer, see `fetchPoolFromClusterMembers` and its usage in `usePoolFromClusterMembers` for an example.

Then we can have two queries here, one for the clustered and one for the unclustered case that enable / disable based on the context.

All the batching and awaiting the parallel promises for the cluster case is done in the lower layer. This has the advantage, that all the clustered fetching and enabling logic is isolated into a single function and this component stays a bit more close to the rendering and calculation of values for display.

src/components/forms/MemoryLimitAvailable.tsx
omarelkashef: When the environment is unclustered, we use the current server, and when it is, we use the cluster members which are later filtered depending on whether the user has selected a target cluster member

src/components/forms/MemoryLimitAvailable.tsx
omarelkashef: We still need the query for the current server when unclustered.

src/components/forms/DiskSizeSelector.tsx
omarelkashef: The two texts did not have a space between.

edlerd: As both are optional, we render potentially just one with an empty space. Maybe we should only render the space when both are set?

src/components/forms/DiskSizeSelector.tsx
```diff
@@ -77,10 +77,11 @@ const DiskSizeSelector: FC<Props> = ({
...
       {(help || helpTotal) && (
-        <p className="p-form-help-text">
-          {help}
-          {helpTotal}
-        </p>
+        <>
+          {help && <p className="p-form-help-text">{help}</p>}
+          {help && helpTotal && " "}
+          {helpTotal && <p className="p-form-help-text">{helpTotal}</p>}
+        </>
```
edlerd:
```suggestion
        <p className="p-form-help-text">
          {help}
          {help && helpTotal && " "}
          {helpTotal}
        </p>
```
A bit simpler.

If we went with the two paragraphs, we wouldn't need the extra space. But I think we want it in a single line, so should keep it within the same paragraph.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,154 @@
...
-const MemoryLimitAvailable: FC<Props> = ({ project }) => {
+const MemoryLimitAvailable: FC<Props> = ({ formik, project }) => {
   const notify = useNotify();
-  const { canViewResources } = useServerEntitlements();
+  const isClustered = useIsClustered();
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const values = formik?.values;
+  const projectResourceMemory = limitToBytes(project?.config["limits.memory"]);
+  const createTarget =
+    (values as CreateInstanceFormValues)?.target ?? undefined;
```
edlerd: This should probably move into the `currentServer` block below. And we should use `currentServer` instead of `createTarget` in the render far below.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,154 @@
...
+    if (createTarget && !createTarget.startsWith(CLUSTER_GROUP_PREFIX)) {
+      return createTarget;
+    }
+    const location = (values as EditInstanceFormValues)?.location ?? undefined;
+    return location || undefined;
+  })();
+
+  if (currentServer) {
+    clusterMembers.filter((member) => member.server_name === currentServer);
+  }
```
edlerd: This is a no-op we filter and don't use the result?
```suggestion
```

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,154 @@
...
+
+  const {
+    data: clusteredResources,
+    error: clusteredError,
+    isLoading: isClusteredLoading,
+  } = useClusterMemberResources(clusterMembers);
```
edlerd: When `currentServer` is set, we should limit this to only load that server's resources, instead of all. I think that is what you tried to do with the filter above. We can inline the filter here maybe?

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,154 @@
...
     }
-    return Math.min(
-      resources.memory.total,
-      Number(limitToBytes(project.config["limits.memory"])),
-    );
+    return Math.min(resource.memory.total, Number(projectResourceMemory));
   };
 
-  const maxMemory = getAvailableMemory();
+  const getClusterAvailableMemoryRange = (resources: LxdResources[]) => {
```
edlerd: I think we should generalize the two functions into one. Instead of `getSingleResourceAvailableMemory` we can pass `[unClusteredResources]` to the `getClusterAvailableMemoryRange` for the same result.

Also, the render function generalizes between clustered and unclustered similarly. So we can reduce the duplication.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,154 @@
...
+              <>
+                <br />
+                {" on "}
+                <ResourceLink
+                  type="cluster-member"
+                  value={createTarget}
+                  to={`/ui/cluster/member/${encodeURIComponent(createTarget)}`}
+                />
+              </>
+            )}
```
edlerd: We never have a target in the unclustered case. We only want to show the member when `currentServer` is set.

I would change the layout slightly to say "on cluster member (foo)" where (foo) is the chip. this might fit into a single line with "Total memory: xx".

There is also another case, where the limit comes from the project, we might want to highlight that source as well. To say something like "Total memory: xx (from project configuration)".

So instead of returning an array with `[min, max]` we also need the information if there is a source of the min / max and what it is. It can be empty, cluster member name or the project limit.

Also the qualifier "from project..." or "on cluster member (foo)" should not be bold, only the actual limit.

src/api/server.tsx
```diff
@@ -110,6 +110,33 @@ export const fetchResources = async (
...
+      .then((results) => {
+        const data: LxdResources[] = [];
+        for (let i = 0; i < clusterMembers.length; i++) {
+          const member = clusterMembers[i];
+          const result = results[i];
+          if (result.status === "fulfilled") {
+            const promise = result;
+            data.push(promise.value);
+          } else if (result.status === "rejected") {
+            reject(constructMemberError(result, member.server_name));
```
edlerd: This means we surface an error if any of the cluster members is unhealthy. I am not 100% sure this is the right thing, we might want a more graceful handling. Say you have a 50 member cluster and one node is down, then would you want the error to show or see the limits just for the remaining 49 members? I think the current implementation is fine, as the error should surface. Wdyt?

omarelkashef: If onne cluster member is down, the reporting might be inaccurate if we ignore that error and calculate available memory of the healthy nodes. So, I think we should not report memory limit if any node is down.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
+              className="help-link-icon"
+              title={`The available memory depends on the member that the instance will be created on. To determine available memory exactly, Select the "Cluster Member" option from the "Target" dropdown and choose a member.`}
+            />
+          </>
+        )}
+      </b>
+      {resourceLink && (
+        <>
+          {" "}
+          on {source} {resourceLink}
```
edlerd: This should probably be "in project (foo)" and "on cluster member (bar)". But not "on project (foo)" as now..

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
     }
-    if (!resources?.memory.total) {
-      return limitToBytes(project.config["limits.memory"]);
+    const projectLimit = Number(projectResourceMemory);
+
+    if (sortedResourcesMemories.length === 0) {
+      return {
+        range: [projectLimit, projectLimit],
+        source: "project",
+      };
```
edlerd, line 92:
```suggestion
      return {
        min: projectLimit
        max:, projectLimit,
        source: "project",
      };
```
Better define a type for this and not use an anonymous array and a named field for source.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,153 @@
...
-    queryKey: [queryKeys.resources],
-    queryFn: async () => fetchResources(),
-    enabled: canViewResources(),
-  });
-
-  if (isLoading) {
+    data: unClusteredResource,
+    error: unClusteredError,
+    isLoading: isUnClusteredLoading,
+  } = useUnClusterMemberResources();
```
edlerd: This is pretty nice already. I wonder if can combine the data loding to pass in the currentServer into a `useResources` function and it internally loads the cluster members if server is clustered and picks the right query between clustered and unclustered cases. That should simplify the rendering logic in here as we can always refer to the same resources array for both cases. So we remove the conditions between clusteredResources and unClusteredResource

src/api/server.tsx
```diff
@@ -110,6 +110,33 @@ export const fetchResources = async (
...
+      }),
+    )
+      .then((results) => {
+        const data: LxdResources[] = [];
+        for (let i = 0; i < clusterMembers.length; i++) {
+          const member = clusterMembers[i];
+          const result = results[i];
+          if (result.status === "fulfilled") {
+            const promise = result;
+            data.push(promise.value);
```
edlerd:
```suggestion
            data.push(result.value);
```
Why the extra variable?

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,143 @@
...
+    const sortedResourcesMemories = resources
+      .map((r) => r.memory.total)
+      .sort((a, b) => a - b);
+    const minResourceMemory = sortedResourcesMemories[0];
+    const maxResourceMemory =
+      sortedResourcesMemories[sortedResourcesMemories.length - 1];
+    if (!projectResourceMemory) {
+      return {
+        range: [minResourceMemory, maxResourceMemory],
+        source: isClustered && currentServer ? "cluster-member" : "",
```
edlerd:
```suggestion
        source: currentServer ? "cluster-member" : "",
```
I think current server is only set if we are clustered, so we don't need the explicit check here.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,143 @@
...
+  formik?: InstanceAndProfileFormikProps;
   project?: LxdProject;
 }
 
-const MemoryLimitAvailable: FC<Props> = ({ project }) => {
+const MemoryLimitAvailable: FC<Props> = ({ formik, project }) => {
   const notify = useNotify();
-  const { canViewResources } = useServerEntitlements();
+  const isClustered = useIsClustered();
+  const { data: clusterMembers = [] } = useClusterMembers();
```
edlerd:
```suggestion
```
We probably don't need this here. See other comments below.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,143 @@
...
-    error,
-    isLoading,
-  } = useQuery({
-    queryKey: [queryKeys.resources],
-    queryFn: async () => fetchResources(),
-    enabled: canViewResources(),
-  });
+    error: error,
+    isLoading: isLoading,
+  } = useResources(clusterMembers, currentServer);
```
edlerd:
```suggestion
  } = useResources(currentServer);
```
We can push the loading of cluster members into the hook, as we don't need them here.

src/components/forms/MemoryLimitAvailable.tsx
edlerd: This seems to still be open.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,143 @@
...
+  let resourceLink = null;
+  if (source === "project") {
+    resourceLink = (
+      <ResourceLink
+        type="project"
+        value={project?.name || "default"}
+        to={`/ui/projects/${encodeURIComponent(project?.name || "default")}`}
+      />
+    );
+  } else if (source === "cluster-member" && currentServer) {
```
edlerd:
```suggestion
  } else if (source === "cluster-member") {
```
I think current server is always defined in this case.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,143 @@
...
+    resourceLink = (
+      <ResourceLink
+        type="cluster-member"
+        value={currentServer}
+        to={`/ui/cluster/member/${encodeURIComponent(currentServer)}`}
+      />
+    );
+  }
+  const showHelpIcon = isClustered && minMemory !== maxMemory;
```
edlerd:
```suggestion
  const showHelpIcon = minMemory !== maxMemory;
```
This is implicitly true. I think we don't have to check for it here.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,143 @@
...
+  let sourceText = "";
+  if (source === "project") {
+    sourceText = "in project";
+  } else if (source === "cluster-member") {
+    sourceText = "on cluster member";
+  }
```
edlerd: This is repeating the check on source we already have on the definition of the resource link above. I think we should define one function `getSourceQualifier()` that returns the source text and the resource link as a fragment. Then we need to branch on the source only once inside that function. The usage below also gets simpler.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,143 @@
...
+      <b>
+        {humanFileSize(minMemory)}
+        {minMemory !== maxMemory && ` - ${humanFileSize(maxMemory)}`}
+        {showHelpIcon && (
+          <>
+            {" "}
+            <Icon
+              name="information"
+              className="help-link-icon"
+              title={`The available memory depends on the member that the instance will be created on. To determine available memory exactly, Select the "Cluster Member" option from the "Target" dropdown and choose a member.`}
```
edlerd: I think we should mention the "Main configuration" as the user has to switch there first to reach the target selector.

src/context/useResources.tsx
```diff
@@ -0,0 +1,31 @@
...
+    enabled: canViewResources() && !isClustered,
+  });
+  const clusteredMembersResourcesQuery = useQuery({
+    queryKey: [queryKeys.resources, clusterMembers],
+    queryFn: async () =>
+      fetchClusterMembersResources(
+        clusterMembers.filter(
+          (member) => !currentServer || member.server_name === currentServer,
+        ),
+      ),
```
edlerd: We need to do the filtering of cluster members before and use the filtered list in the query key. Otherwise, there can be cache key collisions with different values for current server

src/components/forms/MemoryLimitAvailable.tsx
omarelkashef: A MemoryLimit interface was defined and used in that file.

edlerd: Oh I see, what I meant before though is to change the interface to be a flat object without anonymous array. See suggestion above.

src/context/useResources.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useServerEntitlements } from "util/entitlements/server";
+import { useIsClustered } from "./useIsClustered";
+import { fetchResources, fetchClusterMembersResources } from "api/server";
+import { useClusterMembers } from "./useClusterMembers";
+
+export const useResources = (currentServerInCluster?: string) => {
+  const isClustered = useIsClustered();
+  const { data: clusterMembers = [] } = useClusterMembers();
+  const { canViewResources } = useServerEntitlements();
+  const unClusteredMembersResourcesQuery = useQuery({
```
edlerd:
```suggestion
  const singleNodeQuery = useQuery({
```

src/context/useResources.tsx
```diff
@@ -0,0 +1,30 @@
...
+    queryKey: [queryKeys.resources],
+    queryFn: async () => fetchResources(),
+    enabled: canViewResources() && !isClustered,
+  });
+  const filteredClusterMembers = clusterMembers.filter(
+    (member) =>
+      !currentServerInCluster || member.server_name === currentServerInCluster,
+  );
+  const clusteredMembersResourcesQuery = useQuery({
```
edlerd:
```suggestion
  const clusterQuery = useQuery({
```

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
 }
 
-const MemoryLimitAvailable: FC<Props> = ({ project }) => {
+const MemoryLimitAvailable: FC<Props> = ({ formik, project }) => {
   const notify = useNotify();
-  const { canViewResources } = useServerEntitlements();
+  const isClustered = useIsClustered();
+  const values = formik?.values;
+  const projectResourceMemory = limitToBytes(project?.config["limits.memory"]);
+  const currentServerInCluster = (() => {
```
edlerd:
```suggestion
  const clusterMember = (() => {
```

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
     data: resources,
-    error,
-    isLoading,
-  } = useQuery({
-    queryKey: [queryKeys.resources],
-    queryFn: async () => fetchResources(),
-    enabled: canViewResources(),
-  });
+    error: error,
+    isLoading: isLoading,
```
edlerd:
```suggestion
    error,
    isLoading,
```

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
+  } = useResources(currentServerInCluster);
 
   if (isLoading) {
     return <Spinner className="u-loader" text="Loading resources..." />;
   }
-
   if (error) {
     notify.failure("Loading resources failed", error);
+    return null;
```
edlerd, line 45: Why change the early return here?

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
   }
 
-  const getAvailableMemory = () => {
-    if (!project?.config["limits.memory"]) {
-      return resources?.memory.total;
+  const getSourceQualifier = () => {
```
edlerd:
```suggestion
  const getSourceQualifier = (source: string) => {
```
I guess without passing it in, this will always return empty.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
+            to={`/ui/cluster/member/${encodeURIComponent(currentServerInCluster ?? "")}`}
+          />
+        </>
+      );
+    }
+    return null;
+  };
+
+  const getAvailableMemory = (
+    resources: LxdResources[] | LxdResources,
```
edlerd:
```suggestion
```
Here, I think we can rely on the outside scope. Resources comes from the hook above.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
+  ): MemoryLimit => {
+    resources = Array.isArray(resources) ? resources : [resources];
```
edlerd:
```suggestion
    const resourceArray = Array.isArray(resources) ? resources : [resources];
```
I think this is always a reference, and we should treat `resources` immutable.

src/api/server.tsx
```diff
@@ -110,6 +110,32 @@ export const fetchResources = async (
...
+          } else if (result.status === "rejected") {
+            reject(constructMemberError(result, member.server_name));
+          }
+        }
+        resolve(data);
+      })
+      .catch(reject);
+  });
+};
+
```
kimanhou, line 130:
```suggestion
export const fetchClusterMembersResources = async (
  clusterMembers: LxdClusterMember[],
): Promise<LxdResources[]> => {
  const results = await Promise.allSettled(
    clusterMembers.map(async (member) => fetchResources(member.server_name)),
  );

  const data: LxdResources[] = [];

  for (let i = 0; i < clusterMembers.length; i++) {
    const member = clusterMembers[i];
    const result = results[i];

    if (result.status === "rejected") {
      throw constructMemberError(result, member.server_name);
    }

    data.push(result.value);
  }

  return data;
};
```

I know that the other functions are written like this, but should we use `async`/`await` ?

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
-  const maxMemory = getAvailableMemory();
+  const {
+    min: minMemory,
+    max: maxMemory,
+    source,
+  } = getAvailableMemory(resources ?? []);
+  const showHelpIcon = minMemory !== maxMemory;
+  const helpIconText = `The available memory depends on the cluster member that the instance will be created on.\
+  To determine the available memory exactly, go to the "Main configuration" section,\
+  select the "Cluster Member" option from the "Target" dropdown, and choose a member.`;
```
edlerd:
```suggestion
  from the "Target" dropdown, select the "Cluster Member" option and choose a member.`;
```
Better use the order the user has to execute / think in.

src/components/forms/DiskSizeSelector.tsx
```diff
@@ -76,12 +76,11 @@ const DiskSizeSelector: FC<Props> = ({
...
-        <p className="p-form-help-text">
-          {help}
-          {helpTotal}
-        </p>
-      )}
+      <p className="p-form-help-text">
+        {help}
+        {help && helpTotal && " "}
+        {helpTotal}
+      </p>
```
kimanhou, line 83:
```suggestion
      <p className="p-form-help-text">
        {[help, helpTotal].filter(Boolean).join(' ')}
      </p>
```

More elegant but harder to read. Wdyt ?

src/components/forms/MemoryLimitSelector.tsx
```diff
@@ -4,13 +4,19 @@ import type { MemoryLimit } from "types/limits";
 import { BYTES_UNITS, MEM_LIMIT_TYPE } from "types/limits";
 import MemoryLimitAvailable from "components/forms/MemoryLimitAvailable";
 import { useCurrentProject } from "context/useCurrentProject";
+import type { InstanceAndProfileFormikProps } from "./instanceAndProfileFormValues";
 
 interface Props {
+  formik: InstanceAndProfileFormikProps;
```
edlerd, line 10: There is another use case in `ProjectResourceLimitsForm`, there we should report to total sum of memory available. Should this be a different mode of this component or wdyt?

image

src/components/forms/DiskSizeSelector.tsx
```diff
@@ -76,12 +76,11 @@ const DiskSizeSelector: FC<Props> = ({
...
       </div>
-      {(help || helpTotal) && (
-        <p className="p-form-help-text">
-          {help}
-          {helpTotal}
-        </p>
-      )}
+      <p className="p-form-help-text">
+        {help}
+        {help && helpTotal && " "}
```
edlerd:
```suggestion
        {help && helpTotal && <br />}
```
I think a new line would fit better here.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
+      return undefined;
+    }
+    const createTarget =
+      (values as CreateInstanceFormValues)?.target ?? undefined;
+    if (createTarget && !createTarget.startsWith(CLUSTER_GROUP_PREFIX)) {
+      return createTarget;
+    }
+    const location = (values as EditInstanceFormValues)?.location ?? undefined;
+    return location || undefined;
+  })();
```
kimanhou: I'm not a fan of immediately invoking a function in a variable.

Either create a function then call it:
```suggestion
    const getCurrentServer = (
    values:
      | CreateInstanceFormValues
      | EditInstanceFormValues
      | CreateProfileFormValues
      | EditProfileFormValues
      | undefined,
    isClustered: boolean,
  ): string | undefined => {
    if (!isClustered || !values) return undefined;

    const target = "target" in values ? values.target : undefined;
    const location = "location" in values ? values.location : undefined;

    if (target && !target.startsWith(CLUSTER_GROUP_PREFIX)) {
      return target;
    }

    return location;
  };

  const currentServerInCluster = getCurrentServer(values, isClustered);
```

Or keep things inline:
```suggestion
  const { target } = values as CreateInstanceFormValues;
  const { location } = values as EditInstanceFormValues;
  const isTargetGroup = target?.startsWith(CLUSTER_GROUP_PREFIX);
  const validTarget = target && !isTargetGroup ? target : undefined;
  const currentServerInCluster = isClustered
    ? (validTarget ?? location)
    : undefined;
```

Those suggestions need refining, especially with types but you get the idea.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
+        <>
+          {"in project "}
+          <ResourceLink
+            type="project"
+            value={project?.name || "default"}
+            to={`/ui/projects/${encodeURIComponent(project?.name || "default")}`}
+          />
+        </>
+      );
+    } else if (source === "cluster-member") {
```
kimanhou:
```suggestion
    } if (source === "cluster-member") {
```

You don't need `else` after a `return`

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,149 @@
...
     }
-    return Math.min(
-      resources.memory.total,
-      Number(limitToBytes(project.config["limits.memory"])),
-    );
+    return {
+      min: Math.min(minResourceMemory, projectLimit),
+      max: Math.min(maxResourceMemory, projectLimit),
+      source: "",
+    };
```
kimanhou: Can `resources` be empty ? If so, I think it crashes.
Something like this should fix it:
```
  const getAvailableMemory = (
    resources: LxdResources[] | LxdResources,
  ): MemoryLimit => {
    const list = Array.isArray(resources) ? resources : [resources];
    const projectLimit = projectResourceMemory
      ? Number(projectResourceMemory)
      : null;

    if (list.length === 0) {
      return {
        min: projectLimit ?? 0,
        max: projectLimit ?? 0,
        source: "project",
      };
    }

    const memories = list.map((r) => r.memory.total);
    const hwMin = Math.min(...memories);
    const hwMax = Math.max(...memories);

    if (projectLimit === null) {
      return {
        min: hwMin,
        max: hwMax,
        source: currentServerInCluster ? "cluster-member" : "",
      };
    }

    return {
      min: Math.min(hwMin, projectLimit),
      max: Math.min(hwMax, projectLimit),
      source: "",
    };
  };
```

src/components/forms/MemoryLimitSelector.tsx
edlerd: You built the nice abstraction to load resources for both cluster and single node backends.

I think we should introduce a new `ProjectMemoryLimit` component with the different semantic of summing the memory limits per member. Should be very easy to build now with the reusable `useResources` function. In the project case we have no current server, we have no project limit so it is a simple sum over the returned array or the limit from the single node.

src/api/server.tsx
omarelkashef: I initially wanted to match the convention, but your suggestion is cleaner. I think we should do it.

src/components/forms/DiskSizeSelector.tsx
omarelkashef: @edlerd suggested to use a `<br />` between the two texts.

src/components/forms/MemoryLimitAvailable.tsx
omarelkashef: I prefer the second option.

src/components/forms/DiskSizeSelector.tsx
kimanhou: Yeah, I saw afterwards. That breaks my elegant solution ! 🫩

src/api/server.tsx
kimanhou: You would then have to change the other functions in this file at least, which could be a big change. Are you okay with that ? If not, do it in a separate PR. Or create a JIRA ticket to do it later (and not necessarily by you)

src/components/forms/MemoryLimitSelector.tsx
omarelkashef: Yes, that is an easy component.

src/components/forms/MemoryLimitAvailable.tsx
edlerd: This is still open.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,134 @@
...
-    if (!resources?.memory.total) {
-      return limitToBytes(project.config["limits.memory"]);
+    return null;
+  };
+
+  const getAvailableMemory = (): MemoryLimit => {
+    const resourceArray = Array.isArray(resources) ? resources : [resources];
+    const projectLimit = Number(projectResourceMemory);
+
+    if (resourceArray.length === 0) {
```
edlerd:
```suggestion
    if (resourceArray.length === 0 || projectLimit < minResourceMemory ) {
```
If the project limit is the lower bound, we should report it as such.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,134 @@
...
+      };
     }
-    return Math.min(
-      resources.memory.total,
-      Number(limitToBytes(project.config["limits.memory"])),
-    );
+    const resourcesMemories = resourceArray.map((r) => r.memory.total);
+    const minResourceMemory = Math.min(...resourcesMemories);
+    const maxResourceMemory = Math.max(...resourcesMemories);
+    if (!projectResourceMemory) {
```
edlerd:
```suggestion
    if (!projectResourceMemory || maxResourceMemory < projectResourceMemory) {
```
If the member resources are the limiting factor, report them as such.

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,134 @@
...
+            {" "}
+            <Icon
+              name="information"
+              className="help-link-icon"
+              title={helpIconText}
+            />
+          </>
+        )}
+      </b>{" "}
+      {getSourceQualifier(source)}
```
edlerd, line 134: The spacing is subobtimal here:

image


I think we should put the qualifier in its own line:

image

src/components/forms/MemoryLimitAvailable.tsx
```diff
@@ -1,59 +1,134 @@
...
+        min: minResourceMemory,
+        max: maxResourceMemory,
+        source: clusterMember ? "cluster-member" : "",
+      };
+    }
+
+    return {
+      min: Math.min(minResourceMemory, projectLimit),
+      max: Math.min(maxResourceMemory, projectLimit),
+      source: "",
```
edlerd:
```suggestion
      min: minResourceMemory,
      max: projectLimit,
      source: "",
```
I think all other cases are handled above now. 

The remaining case is a wired edge. The project limit is between two cluster member resources. I think it is OK to not report where the limit comes from in that case, though.

src/components/forms/MemoryLimitAvailable.tsx
omarelkashef: Which early return is changed?  This if statement is still there
``` 
 if (error) {
    return null;
  }```

edlerd: I was curious why adding the early return in that condition.

omarelkashef: If we have an error in loading the resources, the component is not useful anymore. We should not be displaying anything in that case.

edlerd: I see, seems correct. Thanks :)

src/components/forms/MemoryLimitAvailable.tsx
omarelkashef: Yes, there are a couple of edge cases that would make the code more complicated and it is not worth it to report the limit in that case.

edlerd: Sounds good to me.
I think we should simplify by removing the Math.min / Math.max cases though, to make clear which edge case remains here.

PR #1673, opened 2025-12-09
feat(CloudInitForm): allow editor in full screen

### LINKED ISSUE #1641: Larger cloud-init yaml editor
I would like to request a larger yaml editor for the cloud-init page.

This is as large as it gets:

Image

At the moment I simply copy all text, paste it in VS Code and make my modifications there, and then paste it back. Editing inside this small screen is very difficult for bigger cloud-init setups

Perhaps a button that makes the editor of your choice wider? Or make use of the whitespace on the right :)

For reference, here is my cloud-init config:

```yaml
## template: jinja
#cloud-config

{% set distro  = v1.distro | lower %}
{% set ver     = v1.distro_version %}
{% set major   = ver.split('.')[0] %}
{% set suite   = 'ubuntu' ~ ver if distro == 'ubuntu'
               else 'debian' ~ major if distro == 'debian'
               else '' %}

{% if distro in ['ubuntu','debian'] %}
apt:
  sources:
    openvox8:
      # e.g. ubuntu24.04 or debian13
      source: "deb https://apt.voxpupuli.org {{ suite }} openvox8"
      keyserver: "hkps://keyserver.ubuntu.com"
      keyid: "0x5FB999C2D62FF3D9"
package_update: true
{% endif %}

{% if distro in ['redhat','centos','rocky','almalinux','oracle','ol','fedora'] %}
yum_repos:
  openvox8:
    name: OpenVox 8
    baseurl: "https://yum.voxpupuli.org/openvox8/{% if distro == 'fedora' %}fedora{% else %}el{% endif %}/$releasever/$basearch"
    enabled: true
    gpgcheck: true
    gpgkey: https://yum.voxpupuli.org/GPG-KEY-openvox.pub
    skip_if_unavailable: true
package_update: true
{% endif %}


timezone: Europe/Amsterdam
package_upgrade: false
packages:
    - openssh-server
    - curl
    - openvox-agent
users:
  - name: ansible
    uid: 600
    no_user_group: true
    primary_group: nogroup
    groups: sudo
    shell: /bin/bash
    sudo: "ALL=(ALL) NOPASSWD:ALL"
    ssh_authorized_keys:
      - ssh-ed25519 AAAAC3NzaC1lZDI1NTEredacted/mAHA ansible@example.com
      - ssh-ed25519 AAAAC3NzaC1lZDredactedCCukV Semaphore
bootcmd:
  - |
    short=$(hostname)
    search=$(awk '/^search/{print $2; exit}' /etc/resolv.conf)

    if [ -z "$search" ]; then
      exit 0
    fi

    fqdn="${short}.${search}"

    if grep -q '^127\.0\.1\.1' /etc/hosts; then
      sed -i "s/^127\.0\.1\.1.*/127.0.1.1 ${fqdn} ${short}/" /etc/hosts
    else
      sed -i "2i 127.0.1.1 ${fqdn} ${short}" /etc/hosts

    fi

runcmd:
  - /opt/puppetlabs/bin/puppet config set server openvox.infra.example.com
  - /opt/puppetlabs/bin/puppet config set environment production --section agent
  - /opt/puppetlabs/bin/puppet config set certname $(hostname -f) --section agent

```

kimanhou:
## Done

- In instance/profile configuration > cloud init, add an expand icon for each editor
- When click on the expand icon, a big modal opens.
- The user can edit in the modal,and the changes are reflected in the editor when the modal gets closed. The changes are however not saved.
- The user can view inherited config in the modal, but not edit it there.
- When click on the chips in the modal title, the modal closes and the url changes to the one provided by the chip.
- In both inline editor and the modal, a read only message was added.

Fixes #1641 

## QA

1. Run the LXD-UI:
    - Go to a profile or instance > Configuration > Cloud Init
    - There should be expand icon for each editor
    - Make sure you can't edit an inherited config
    - Make sure you can edit when onverriding a config

## Screenshots

Read only
image

Edit
image

### DIFF
--- a/src/components/forms/CloudInitExpandButton.tsx
+++ b/src/components/forms/CloudInitExpandButton.tsx
@@ -1,0 +1,141 @@
+import type { FC, MouseEvent, ReactNode } from "react";
+import { Button, Icon } from "@canonical/react-components";
+import { useModal } from "context/useModal";
+import ResourceLink from "components/ResourceLink";
+import type { CloudInitKey } from "components/forms/CloudInitForm";
+import YamlModal from "components/forms/YamlModal";
+import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import { ensureEditMode } from "util/instanceEdit";
+import classnames from "classnames";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import { useLocation } from "react-router-dom";
+import ResourceLabel from "components/ResourceLabel";
+
+const LABELS: Record<CloudInitKey, string> = {
+  cloud_init_network_config: "cloud-init network config",
+  cloud_init_user_data: "cloud-init user data",
+  cloud_init_vendor_data: "cloud-init vendor data",
+};
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  project: string;
+  name: CloudInitKey;
+  initialValue: string;
+  source?: ReactNode;
+  onApplyChanges?: () => void;
+  isReadOnly?: boolean;
+  className?: string;
+}
+
+const CloudInitExpandButton: FC<Props> = ({
+  formik,
+  project,
+  name,
+  initialValue,
+  source,
+  onApplyChanges,
+  isReadOnly,
+  className,
+}) => {
+  const { showModal, hideModal } = useModal();
+  const isSmallScreen = useIsScreenBelow();
+  const location = useLocation();
+
+  const getModalTitle = () => {
+    const label = LABELS[name];
+    const entityName = formik.values.name || "";
+    const entityType = formik.values.entityType;
+
+    const isCreating = location.pathname.includes("/create");
+    const isNew = isCreating || !entityName;
+    const isInherited = source && source !== entityName;
+
+    let entityIdentifier = null;
+
+    if (!isNew && entityName) {
+      entityIdentifier = (
+        <ResourceLink
+          type={entityType}
+          value={entityName}
+          to={`/ui/project/${encodeURIComponent(project)}/${entityType}/${encodeURIComponent(entityName)}`}
+        />
+      );
+    } else if (isNew && entityName) {
+      entityIdentifier = <ResourceLabel type={entityType} value={entityName} />;
+    }
+
+    return (
+      <div
+        className="u-no-margin"
+        onClick={(event) => {
+          const target = event.target as HTMLElement;
+          const link = target.closest("a");
+
+          if (link) {
+            const openInNewTab =
+              event.ctrlKey || event.metaKey || event.shiftKey;
+
+            if (!openInNewTab) {
+              hideModal();
+            }
+          }
+        }}
+      >
+        <h4 className="p-heading--4 u-no-margin--bottom">
+          {isReadOnly ? "View" : "Edit"} {label}
+        </h4>
+        <p className="p-heading--6 u-text--muted u-no-margin--top u-no-margin--bottom font-weight-normal">
+          For {isNew && "new "}
+          {entityType} {entityIdentifier}
+          {isSmallScreen && <br />}
+          {isInherited && (
+            <>
+              {" inherited from "}
+              {source}
+            </>
+          )}
+        </p>
+      </div>
+    );
+  };
+
+  const handleApply = (newValue: string) => {
+    ensureEditMode(formik);
+    formik.setFieldValue(name, newValue);
+    onApplyChanges?.();
+  };
+
+  const handleClick = (e: MouseEvent<HTMLElement>) => {
+    showModal(
+      <YamlModal
+        title={getModalTitle()}
+        onClose={hideModal}
+        applyChanges={handleApply}
+        readOnly={isReadOnly || !!formik.values.editRestriction}
+        readOnlyMessage={
+          isReadOnly
+            ? "Read only editor for inherited value. Create an override to modify."
+            : formik.values.editRestriction
+        }
+        initialValue={formik.values[name] || initialValue}
+      />,
+      e,
+    );
+  };
+
+  return (
+    <Button
+      onClick={handleClick}
+      type="button"
+      appearance="base"
+      title={isReadOnly ? "Expand view" : "Expand editor"}
+      hasIcon
+      className={classnames("u-no-margin--bottom", className)}
+    >
+      <Icon name="fullscreen" className="full-screen-icon" />
+    </Button>
+  );
+};
+
+export default CloudInitExpandButton;
--- a/src/components/forms/CloudInitForm.tsx
+++ b/src/components/forms/CloudInitForm.tsx
@@ -1,4 +1,5 @@
 import type { FC } from "react";
+import { useState } from "react";
 import { Button, Icon, Tooltip } from "@canonical/react-components";
 import CloudInitConfig from "components/forms/CloudInitConfig";
 import type {
@@ -11,12 +12,16 @@ import { getInstanceField } from "util/instanceConfigFields";
 import { ensureEditMode } from "util/instanceEdit";
 import classnames from "classnames";
 import { getConfigRowMetadata } from "util/configInheritance";
+import CloudInitExpandButton from "components/forms/CloudInitExpandButton";
+import ResourceLink from "components/ResourceLink";
 
 export interface CloudInitFormValues {
   cloud_init_network_config?: string;
   cloud_init_user_data?: string;
   cloud_init_vendor_data?: string;
 }
+
+export type CloudInitKey = keyof CloudInitFormValues;
 
 export const cloudInitPayload = (values: InstanceAndProfileFormValues) => {
   return {
@@ -29,12 +34,53 @@ export const cloudInitPayload = (values: InstanceAndProfileFormValues) => {
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
+  project: string;
 }
 
-const CloudInitForm: FC<Props> = ({ formik }) => {
-  const getCloudInitRow = (label: string, name: string, value?: string) => {
+const CloudInitForm: FC<Props> = ({ formik, project }) => {
+  const getCloudInitRow = (
+    label: string,
+    name: CloudInitKey,
+    value: string | undefined,
+  ) => {
+    const [rowVersion, setRowVersion] = useState(0);
     const metadata = getConfigRowMetadata(formik.values, name);
     const isOverridden = value !== undefined;
+
+    const forceUpdate = () => {
+      setRowVersion((prev) => prev + 1);
+    };
+
+    const metadataSource = metadata?.source || "";
+    const isInherited = metadataSource && metadataSource !== formik.values.name;
+
+    const profileNameFromSource = (source: string) => {
+      if (!source || !source.includes(" profile")) {
+        return null;
+      }
+
+      return source.split(" profile")[0];
+    };
+
+    const profileName = isInherited
+      ? profileNameFromSource(metadataSource)
+      : null;
+
+    const source = profileName ? (
+      <>
+        profile{" "}
+        <ResourceLink
+          type="profile"
+          value={profileName}
+          className={classnames({
+            "u-text--line-through": isOverridden,
+          })}
+          to={`/ui/project/${encodeURIComponent(project)}/profile/${encodeURIComponent(profileName)}`}
+        />
+      </>
+    ) : (
+      metadataSource
+    );
 
     return getConfigurationRowBase({
       configuration: <strong>{label}</strong>,
@@ -54,8 +100,20 @@ const CloudInitForm: FC<Props> = ({ formik }) => {
             </b>
           </div>
           {metadata && (
-            <div className="p-text--small u-text--muted">
-              From: {metadata.source}
+            <div className="u-d-flex u-margin-top--small">
+              <span className="p-text--small u-text--muted">From {source}</span>
+
+              {metadata.value && (
+                <CloudInitExpandButton
+                  formik={formik}
+                  project={project}
+                  name={name}
+                  source={source}
+                  initialValue={metadata.value}
+                  isReadOnly
+                  className="u-no-margin--top u-margin-left--small"
+                />
+              )}
             </div>
           )}
         </div>
@@ -63,6 +121,7 @@ const CloudInitForm: FC<Props> = ({ formik }) => {
       override: isOverridden ? (
         <>
           <CloudInitConfig
+            key={`cloud-init-override-${name}-${rowVersion}`}
             config={value ?? ""}
             setConfig={(config) => {
               ensureEditMode(formik);
@@ -79,10 +138,18 @@ const CloudInitForm: FC<Props> = ({ formik }) => {
             title={formik.values.editRestriction ?? "Clear override"}
             disabled={!!formik.values.editRestriction}
             hasIcon
-            className="u-no-margin--bottom"
+            className="u-no-margin--bottom u-no-margin--right"
           >
             <Icon name="close" className="clear-configuration-icon" />
           </Button>
+          <CloudInitExpandButton
+            formik={formik}
+            project={project}
+            name={name}
+            initialValue={value ?? ""}
+            onApplyChanges={forceUpdate}
+            isReadOnly={!!formik.values.editRestriction}
+          />
         </>
       ) : (
         <Button
--- a/src/components/forms/YamlForm.tsx
+++ b/src/components/forms/YamlForm.tsx
@@ -6,6 +6,7 @@ import { editor } from "monaco-editor/esm/vs/editor/editor.api";
 import IStandaloneCodeEditor = editor.IStandaloneCodeEditor;
 import classnames from "classnames";
 import { useListener } from "@canonical/react-components";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
 
 export interface YamlFormValues {
   yaml?: string;
@@ -30,6 +31,7 @@ const YamlForm: FC<Props> = ({
 }) => {
   const [editor, setEditor] = useState<IStandaloneCodeEditor | null>(null);
   const containerRef = useRef<HTMLDivElement>(null);
+  const isSmallScreen = useIsScreenBelow();
 
   loader.config({ paths: { vs: "/ui/monaco-editor/min/vs" } });
 
@@ -70,12 +72,15 @@ const YamlForm: FC<Props> = ({
             scrollBeyondLastLine: false,
             wordWrap: "on",
             wrappingStrategy: "advanced",
-            minimap: {
-              enabled: false,
-            },
+            minimap: { enabled: false },
             overviewRulerLanes: 0,
             readOnly: readOnly,
+            scrollbar: {
+              vertical: "auto",
+              alwaysConsumeMouseWheel: false,
+            },
             readOnlyMessage: { value: readOnlyMessage ?? "" },
+            lineNumbersMinChars: isSmallScreen ? 2 : 5,
           }}
           onMount={(editor: IStandaloneCodeEditor) => {
             setEditor(editor);
--- a/src/components/forms/YamlModal.tsx
+++ b/src/components/forms/YamlModal.tsx
@@ -1,0 +1,56 @@
+import type { ReactNode } from "react";
+import { useState, useEffect, type FC } from "react";
+import { Modal } from "@canonical/react-components";
+import YamlForm from "components/forms/YamlForm";
+
+interface Props {
+  title: ReactNode;
+  onClose: () => void;
+  applyChanges: (newValue: string) => void;
+  readOnly?: boolean;
+  readOnlyMessage?: string;
+  initialValue: string;
+}
+
+const YamlModal: FC<Props> = ({
+  title,
+  onClose,
+  applyChanges,
+  readOnly,
+  readOnlyMessage,
+  initialValue,
+}) => {
+  const [value, setValue] = useState<string>(initialValue ?? "");
+
+  useEffect(() => {
+    setValue(initialValue ?? "");
+  }, [initialValue]);
+
+  const handleClose = () => {
+    if (!readOnly) {
+      applyChanges(value);
+    }
+    onClose();
+  };
+
+  return (
+    <Modal
+      close={handleClose}
+      title={title}
+      className="cloud-init-full-editor-modal"
+      closeOnOutsideClick={false}
+    >
+      <div className="cloud-init-modal-content">
+        <YamlForm
+          yaml={value}
+          setYaml={setValue}
+          readOnly={readOnly}
+          readOnlyMessage={readOnlyMessage}
+          autoResize
+        />
+      </div>
+    </Modal>
+  );
+};
+
+export default YamlModal;
--- a/src/context/useModal.tsx
+++ b/src/context/useModal.tsx
@@ -1,9 +1,9 @@
-import type { ReactNode } from "react";
+import type { ReactNode, SyntheticEvent } from "react";
 import { createContext, useContext, useState, useCallback } from "react";
 import { usePortal } from "@canonical/react-components";
 
 export interface ModalType {
-  showModal: (content: ReactNode) => void;
+  showModal: (content: ReactNode, event?: SyntheticEvent<HTMLElement>) => void;
   hideModal: () => void;
 }
 
@@ -14,7 +14,9 @@ const ModalContext = createContext<ModalType>({
 
 export const ModalProvider = ({ children }: { children: ReactNode }) => {
   const [content, setContent] = useState<ReactNode | null>(null);
-  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal({
+    programmaticallyOpen: true,
+  });
 
   const showModal = useCallback((content: ReactNode) => {
     setContent(content);
--- a/src/sass/_cloud_init_form.scss
+++ b/src/sass/_cloud_init_form.scss
@@ -1,0 +1,58 @@
+.cloud-init {
+  .configuration-table {
+    .configuration {
+      width: 9rem;
+    }
+
+    .inherited {
+      width: auto;
+    }
+
+    .override {
+      width: auto;
+    }
+
+    .override-form {
+      align-items: flex-end;
+      flex-direction: column;
+
+      > :first-child {
+        width: 100%;
+      }
+    }
+  }
+
+  .cloud-init-config {
+    .code-editor-wrapper {
+      max-height: 35dvh;
+      min-height: 7.5rem;
+    }
+  }
+}
+
+.cloud-init-full-editor-modal {
+  .p-modal__dialog {
+    display: flex;
+    flex-direction: column;
+    height: 80dvh;
+
+    @include large {
+      width: 80dvw;
+    }
+
+    .p-modal__header + div {
+      flex: 1;
+
+      .cloud-init-modal-content {
+        display: flex;
+        flex-direction: column;
+        height: 100%;
+        padding-right: 2px;
+
+        .code-editor-wrapper {
+          min-height: 100%;
+        }
+      }
+    }
+  }
+}
--- a/src/components/forms/CloudInitConfig.tsx
+++ b/src/components/forms/CloudInitConfig.tsx
@@ -14,6 +14,7 @@ const CloudInitConfig: FC<Props> = ({ config, setConfig }) => {
         setYaml={setConfig}
         autoResize={true}
         readOnly={!setConfig}
+        readOnlyMessage="Read only editor for inherited value. Create an override to modify."
       />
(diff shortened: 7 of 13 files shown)


src/components/forms/CloudInitForm.tsx
```diff
@@ -48,21 +112,41 @@ const CloudInitForm: FC<Props> = ({ formik }) => {
...
+              <span className="p-text--small u-text--muted">
+                From: {metadata.source}
+              </span>
+
+              {metadata.value && (
+                <Button
+                  onClick={() => {
+                    setModalName(name);
+                    setIsReadOnlyModal(true);
+                    openPortal();
```
edlerd: Maybe check the `const { showModal, hideModal } = useModal();` helper, it might simplify things here as we don't need multiple states and avoid the portal.

src/components/forms/CloudInitForm.tsx
```diff
@@ -32,8 +36,68 @@ interface Props {
...
+  );
+
+  const getCurrentModalMetadata = () => {
+    if (modalName === "cloud_init_network_config") return networkMeta;
+    if (modalName === "cloud_init_user_data") return userMeta;
+    if (modalName === "cloud_init_vendor_data") return vendorMeta;
+    return null;
+  };
+
+  const getModalTitle = (name: CloudInitKey | null, isReadOnly = false) => {
```
edlerd: Maybe mention the instance name and that this is cloud init data?

Also we might want to clarify this is inherited or local in the title -- though it shouldn't be too long either.

edlerd (commented):
This is pretty nice to use! Some thoughts on the code and details below.

src/components/forms/CloudInitForm.tsx
edlerd: We might create a `CloudInitExpandButton` to encapsulate the modal in it.

src/components/forms/CloudInitForm.tsx
kimanhou: What do you think ?
Screenshot from 2025-12-10 14-18-17
Screenshot from 2025-12-10 14-18-23

edlerd: Looks great!
Maybe better "For instance (cloud)", where (cloud) is the chip. Or "For profile (foo)". I always try to use the chip with the entity type as I wouldn't expect a user to know the icons by heart, so adding a qualifier seems good opportunity for them to learn the icons. Wdyt?

kimanhou: Sure ! 
Screenshot from 2025-12-10 15-31-45
Screenshot from 2025-12-10 15-31-59

src/components/forms/CloudInitForm.tsx
kimanhou: I created `CloudInitExpandButton` and put the modal in it.

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <h4 className="p-heading--4 u-no-margin--bottom">
+          {isReadOnly ? "View" : "Edit"} {label}
+        </h4>
+        <p className="p-heading--6 u-text--muted u-no-margin--top u-no-margin--bottom font-weight-normal">
+          For {entityType}{" "}
+          <ResourceLink
+            type={entityType}
+            value={entityName}
+            to={`/ui/project/${encodeURIComponent(project)}/${entityType}/${encodeURIComponent(entityName)}`}
+          />
```
edlerd: entity name will be empty during creation, so we should not have the chip, but only the type maybe?

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+          {isSmallScreen && <br />}
+          {isInherited && <> inherited from {source}</>}
```
edlerd:
```suggestion
          {isInherited && <>inherited from {source}</>}
```
We don't need the space, do we?

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onApplyChanges?: () => void;
+  isReadOnly?: boolean;
+  className?: string;
+}
+
+const CloudInitExpandButton: FC<Props> = ({
+  formik,
+  project,
+  name,
+  metadata,
```
edlerd:
```suggestion
  source,
```
Shall we unpack the metadata object in the caller?

src/components/forms/CloudInitYamlEditorModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface Props {
+  title: ReactNode;
+  onClose: () => void;
+  applyChanges: (newValue: string) => void;
+  readOnly?: boolean;
+  readOnlyMessage?: string;
+  initialValue: string;
+}
+
+const CloudInitYamlEditorModal: FC<Props> = ({
```
edlerd:
```suggestion
const YamlModal: FC<Props> = ({
```
This seems to be a generic component. Shall we rename component, filename and css classes to hint at that?

src/components/forms/CloudInitYamlEditorModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  initialValue,
+}) => {
+  const [value, setValue] = useState<string>(initialValue ?? "");
+
+  return (
+    <Modal
+      close={onClose}
+      title={title}
+      className="cloud-init-full-editor-modal"
+      closeOnOutsideClick={readOnly}
```
edlerd:
```suggestion
      closeOnOutsideClick={false}
```
I don't think this should be connected to the read only-ness.

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +34,39 @@ export const cloudInitPayload = (values: InstanceAndProfileFormValues) => {
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
+  project: string;
 }
 
-const CloudInitForm: FC<Props> = ({ formik }) => {
-  const getCloudInitRow = (label: string, name: string, value?: string) => {
-    const metadata = getConfigRowMetadata(formik.values, name);
```
edlerd, line 36: Can we keep the `metadata = getConfigRowMetadata` inside the `getCloudInitRow` function? Seems too complex to pull it out and define multiple variables for it just to pass it back in again.

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+      />,
+    );
+  };
+
+  return (
+    <Button
+      onClick={handleClick}
+      type="button"
+      appearance="base"
+      title={isReadOnly ? "View full content" : "Expand editor"}
```
edlerd:
```suggestion
      title={isReadOnly ? "Expand view" : "Expand editor"}
```
Maybe more clear?

src/components/forms/CloudInitExpandButton.tsx
edlerd: Also, this shows up on the editor where it is not inherited. I think it should be hidden in that case.

image

src/components/forms/CloudInitExpandButton.tsx
kimanhou: Good catch !

image

src/components/forms/CloudInitExpandButton.tsx
kimanhou: We do need the space, otherwise it looks like this
image

What does "inherited from LXD" mean ? I mimicked the "From LXD" underneath the editor in the configuration screen.
image

edlerd: Yes, when viewing the inherited column, this is correct. From LXD means it is the default value.

I was hinting at the 3rd column with an override. There we should hide the inherited from completely.

src/components/forms/CloudInitExpandButton.tsx
kimanhou: Yes

Should we have a chip for the profile ?
image

edlerd: Sure, I think we should use the chip.

src/components/forms/CloudInitForm.tsx
kimanhou: I put it out because otherwise there was not the same number of hooks called for read and edit modes. We can put it back now.

src/components/forms/CloudInitExpandButton.tsx
kimanhou: Is it functionnally correct to pass `source=""`  (empty string) for the 3rd column with an override then ?

src/components/forms/CloudInitExpandButton.tsx
kimanhou: Cool

image

However, when we click on the chip, it goes to the right page but underneath the modal. This is unexpected but what should be the correct behavior ?
Ctrl + click and mouse wheel click work and open a new tab.

kimanhou: Also the config can only be inherited from a profile, right ? No other entity type ?

edlerd: > However, when we click on the chip, it goes to the right page but underneath the modal. This is unexpected but what should be the correct behavior ?

This is the downside of the use modal hook. Not sure what is best to do here.

edlerd: > Also the config can only be inherited from a profile, right ? No other entity type ?

Yes, I think so, only profile or default LXD.

src/components/forms/CloudInitForm.tsx
```diff
@@ -48,21 +69,36 @@ const CloudInitForm: FC<Props> = ({ formik }) => {
...
+                config={(metadata.value as string) || ""}
               />
             </b>
           </div>
           {metadata && (
-            <div className="p-text--small u-text--muted">
-              From: {metadata.source}
+            <div className="u-d-flex u-margin-top--small">
+              <span className="p-text--small u-text--muted">
+                From: {metadata.source}
```
edlerd: Can we align this to be "From profile (foo)" with a chip and link?

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +33,29 @@ export const cloudInitPayload = (values: InstanceAndProfileFormValues) => {
...
   formik: InstanceAndProfileFormikProps;
+  project: string;
 }
 
-const CloudInitForm: FC<Props> = ({ formik }) => {
-  const getCloudInitRow = (label: string, name: string, value?: string) => {
+const CloudInitForm: FC<Props> = ({ formik, project }) => {
+  const [fieldVersions, setFieldVersions] = useState<Record<string, number>>(
+    {},
+  );
```
edlerd: We can move this into the get row function, and then we don't have to keep track of which state is for which row:
```suggestion
  const getCloudInitRow = (
    label: string,
    name: CloudInitKey,
    value: string | undefined,
  ) => {
    const [rowVersion, setRowVersion] = useState(0);

    const metadata = getConfigRowMetadata(formik.values, name);
    const isOverridden = value !== undefined;
    const fieldVersion = rowVersion;

    const forceUpdate = () => {
      setRowVersion((prev) => prev + 1);
    };
```

src/components/forms/CloudInitExpandButton.tsx
edlerd: Maybe a callback from the chip that allows for closing the modal?
Something like below. Or go back to mounting the modal locally so it gets unmounted automatically.
```
  const handleClick = (event) => {
    // If link is opening in a new tab/window, don't close modal
    const openInNewTab =
      event.ctrlKey || event.metaKey || event.shiftKey || event.button === 1;

    if (!openInNewTab) {
      event.preventDefault(); // Prevent default link behavior
      setIsOpen(false);       // Close the modal
    }
    // Otherwise, let the link open in new tab
  };

```

src/components/forms/YamlModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const YamlModal: FC<Props> = ({
+  title,
+  onClose,
+  applyChanges,
+  readOnly,
+  readOnlyMessage,
+  initialValue,
+}) => {
+  const [value, setValue] = useState<string>(initialValue ?? "");
```
edlerd, line 23: Do we need this intermediate state? Or should we update the formik state directly? I think updating formik state directly might be simpler. We wouldn't need the "Apply changes" or "Cancel" buttons and could give more space to the editor.
```suggestion
```

src/context/useModal.tsx
```diff
@@ -16,10 +16,13 @@ export const ModalProvider = ({ children }: { children: ReactNode }) => {
...
   const { openPortal, closePortal, isOpen, Portal } = usePortal();
 
-  const showModal = useCallback((content: ReactNode) => {
-    setContent(content);
-    openPortal();
-  }, []);
+  const showModal = useCallback(
+    (content: ReactNode, event?: SyntheticEvent<HTMLElement>) => {
+      setContent(content);
+      openPortal(event);
```
edlerd: Can we set `programmaticallyOpen: true` instead of relaying the event here? Seems simpler:
```
  const { openPortal, closePortal, isOpen, Portal } = usePortal({
    programmaticallyOpen: true,
  });
```

src/components/forms/CloudInitExpandButton.tsx
edlerd: I think you can make source optional and avoid setting it in that column.

src/components/forms/CloudInitForm.tsx
```diff
@@ -48,21 +69,36 @@ const CloudInitForm: FC<Props> = ({ formik }) => {
...
           {metadata && (
-            <div className="p-text--small u-text--muted">
-              From: {metadata.source}
+            <div className="u-d-flex u-margin-top--small">
+              <span className="p-text--small u-text--muted">
+                From: {metadata.source}
+              </span>
+
+              {metadata.value && (
+                <CloudInitExpandButton
```
edlerd, line 107: The second line could be improved on creation. Maybe change it to "For new instance,"?

image

src/components/forms/CloudInitForm.tsx
kimanhou: Should it be "From profile (foo)" or "From (foo) profile" ?
FYI `metadata.source = "foo profile"` so it's currently "from (foo) profile" in the modal title.

edlerd: I think "From profile (foo)" is more natural to read.

src/components/forms/CloudInitForm.tsx
kimanhou: Sure. There is an issue on profile creation, since we have the name of the profile, a chip is displayed. I need to fix it to "For new profile".

kimanhou: Fixed. I also display the name in bold on instance or profile creation when the name exists.

src/components/forms/YamlModal.tsx
kimanhou: I don't manage to make it work without this intermediate state. When I try to update the `formik` directly, the modal content is not re-rendering because the modal portal is created once and doesn't automatically react to external state changes.

But I removed the "Cancel" and "Apply changes" buttons and changes are now applied when the modal is closed.

src/components/forms/CloudInitExpandButton.tsx
kimanhou: I added a callback to a wrapper around the chip

src/components/forms/YamlModal.tsx
edlerd: sounds good, thanks.

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  const handleClick = (e: MouseEvent<HTMLElement>) => {
+    showModal(
+      <YamlModal
+        title={getModalTitle()}
+        onClose={hideModal}
+        applyChanges={handleApply}
+        readOnly={isReadOnly || !!formik.values.editRestriction}
+        readOnlyMessage={formik.values.editRestriction}
```
edlerd: Shall we add a message when this is read-only mode? Currently, it pops up an empty help bubble.

Same problem on the inline editor, it would be nice to have a message stating it is inherited and read only.

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +35,49 @@ export const cloudInitPayload = (values: InstanceAndProfileFormValues) => {
...
+
+  const getCloudInitRow = (
+    label: string,
+    name: CloudInitKey,
+    value: string | undefined,
+  ) => {
+    const [rowVersion, setRowVersion] = useState(0);
     const metadata = getConfigRowMetadata(formik.values, name);
     const isOverridden = value !== undefined;
+    const fieldVersion = rowVersion;
```
edlerd:
```suggestion
```
Can directly use the other variable.

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +35,49 @@ export const cloudInitPayload = (values: InstanceAndProfileFormValues) => {
...
+    const metadataSource = metadata?.source || "";
+    const isInherited = metadataSource && metadataSource !== formik.values.name;
+    let profile;
+
+    if (isInherited && profiles) {
+      const profileNameFromSource = metadataSource.includes(" profile")
+        ? metadataSource.split(" profile")[0]
+        : null;
+      profile = profiles.find((t) => t.name === profileNameFromSource);
+    }
```
edlerd: Why load the profiles and fetch from them? Shall we directly use the `profileNameFromSource`?

src/components/forms/CloudInitForm.tsx
```diff
@@ -48,21 +91,34 @@ const CloudInitForm: FC<Props> = ({ formik }) => {
           <div className="mono-font">
             <b>
               <CloudInitConfig
-                key={`cloud-init-${name}-${metadata.value}`}
-                config={metadata.value as string}
+                key={`cloud-init-inherited-${name}-${metadata.value ?? ""}`}
+                config={(metadata.value as string) || ""}
```
edlerd: We don't have to change this, can it be reverted?

src/components/forms/CloudInitForm.tsx
```diff
@@ -29,12 +35,49 @@ export const cloudInitPayload = (values: InstanceAndProfileFormValues) => {
...
+    }
+
+    const source = profile ? (
+      <>
+        profile{" "}
+        <ResourceLink
+          type="profile"
+          value={profile.name}
+          to={`/ui/project/${encodeURIComponent(project)}/profile/${encodeURIComponent(profile.name)}`}
+        />
```
edlerd, line 79: We should add the `u-text--line-through` class name when an override exists.

src/sass/_cloud_init_form.scss
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+  }
+
+  .p-modal__footer {
+    align-items: center;
+    display: flex;
+    gap: $sph--large;
+    justify-content: flex-end;
+  }
```
edlerd:
```suggestion
```
I think we don't need this any more.

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+        >
+          <ResourceLink
+            type={entityType}
+            value={entityName}
+            to={`/ui/project/${encodeURIComponent(project)}/${entityType}/${encodeURIComponent(entityName)}`}
+          />
+        </span>
+      );
+    } else if (isNew && entityName) {
+      entityIdentifier = <strong>{entityName}</strong>;
```
edlerd: Can we use the `<ResourceLabel` with the icon here?

edlerd (commented):
Nearly perfect, thanks for the changes!

src/components/forms/CloudInitExpandButton.tsx
kimanhou: On the inline editor, the read only message takes so much space. Is it acceptable since it only stays for a few seconds ?

image

edlerd: I think it is better than the empty message, yes. Maybe change the caption. Something like "Read only editor for inherited value. Create an override to modify."

src/components/forms/CloudInitForm.tsx
kimanhou: I wanted to make sure that the profile already exists before displaying the chip, but now I'm using the URL to know if we are creating an instance or a profile so I guess we can indeed use `profileNameFromSource`

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { useLocation } from "react-router-dom";
+import ResourceLabel from "components/ResourceLabel";
+
+const LABELS: Record<CloudInitKey, string> = {
+  cloud_init_network_config: "cloud-init network config",
+  cloud_init_user_data: "cloud-init user data",
+  cloud_init_vendor_data: "cloud-init vendor data",
+};
+
+const wrapSourceWithClickHandler = (
```
edlerd: What happened here -- this seems like duplicate logic?

We could extract this into a dedicated component

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+  };
+
+  const handleClick = (e: MouseEvent<HTMLElement>) => {
+    showModal(
+      <YamlModal
+        title={getModalTitle()}
+        onClose={hideModal}
+        applyChanges={handleApply}
+        readOnly={isReadOnly || !!formik.values.editRestriction}
+        readOnlyMessage="Read only editor for inherited value. Create an override to modify."
```
edlerd: this message should switch to `formik.values.editRestriction` when `isReadOnly` is false.

src/components/forms/CloudInitExpandButton.tsx
kimanhou: When an instance inherits from a profile, the instance's chip had the wrapper, but not the profile chip.
I need the profile chip under the inline editor and in the modal title so I create it once in `CloudInitForm` then add the wrapper only in the modal. I can't add the wrapper directly in `CloudInitForm` because it doesn't have access to `hideModal` there.

edlerd: I don't like that we have two solutions to the same problem. Can we define a reusable `<LinkInterceptor callback={hideModalOnSameTabClick}>{source}</...` component? 
We can then define a `const hideModalOnSameTabClick` that handles the check about new tab and triggers close modal. The wrapper can be used in both instances. For source passed in and the link defined locally.

src/components/forms/CloudInitForm.tsx
edlerd: This is still open.

kimanhou: Sorry, missed htat

src/components/forms/CloudInitExpandButton.tsx
kimanhou: I did something, please have a look and tell me if it's what you meant.

src/components/forms/CloudInitExpandButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+import ResourceLabel from "components/ResourceLabel";
+import LinkInterceptor from "components/LinkInterceptor";
+
+const LABELS: Record<CloudInitKey, string> = {
+  cloud_init_network_config: "cloud-init network config",
+  cloud_init_user_data: "cloud-init user data",
+  cloud_init_vendor_data: "cloud-init vendor data",
+};
+
+const hideModalOnSameTabClick =
```
edlerd:
```suggestion
const onSameTabClick =
```
Maybe easier to read?

src/components/LinkInterceptor.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  if (children.type === ResourceLink) {
+    const linkProps = children.props as React.ComponentProps<
+      typeof ResourceLink
+    >;
+    return <ResourceLink {...linkProps} onClick={callback} />;
+  }
+
+  return <>{children}</>;
```
edlerd:
```suggestion
  return <span onClick={callback}>{children}</span>;
```
Your previous solution of wrapping the link in an element to get the click should work just as fine. then we can revert the changes to `ResourceLink` component. Maybe the name of this component should then be `ClickInterceptor`.

kimanhou: I know but I thought adding the onClick on the ResourceLink directly was a bit cleaner and would simplify the DOM. Wdyt ?

edlerd: Yes, cleaner, I was afraid the code is fairly complex, very specific, and as such harder to maintain.

kimanhou: Also your solution adds the callback to "profile {CHIP}" and not just the chip, so it redirects when you click on the string "profile" as well.

edlerd: Yeah, it is not ideal either.

edlerd: Maybe easier to define an onclick handler for the modal body. Then check if the click is towards a link and close the modal. Then the whole logic is contained in the modal and we don't need additional components or interfaces.

kimanhou: Oooh nice idea !

src/components/ResourceLink.tsx
```diff
@@ -10,9 +10,17 @@ interface Props {
   to: string;
   disabled?: boolean;
   className?: string;
+  onClick?: (e?: MouseEvent<HTMLElement>) => void;
```
edlerd:
```suggestion
```
Shall we revert the changes here?

src/components/forms/YamlModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+interface Props {
+  title: ReactNode;
+  onClose: () => void;
+  applyChanges: (newValue: string) => void;
+  readOnly?: boolean;
+  readOnlyMessage?: string;
+  initialValue: string;
+}
+
+const YamlModal: FC<Props> = ({
```
edlerd, line 15: There are some edge cases, where when I resize the window (increasing and the reducing height) the modal has a scrollbar. This makes it pretty hard to interact with the multi line editor as content grows. I also sometimes get a tiny scrollbar directly after opening on very wide screens.

image

kimanhou: I tried something

edlerd: The phenomenon hints at some miscalculated available height in one of the helpers such as the update max height function. Michele logged a theoretical bug in it into our backlog a while ago. That never surfaced in the past. But this might be the similar case here?

src/sass/_cloud_init_form.scss
```diff
@@ -0,0 +1,58 @@
...
+      flex: 1;
+
+      .cloud-init-modal-content {
+        display: flex;
+        flex-direction: column;
+        height: 100%;
+        padding-right: 2px;
+
+        .code-editor-wrapper {
+          height: 100% !important;
```
edlerd:
```suggestion
          min-height: 100%;
```
This is the problem why the editor is not responsive. Changing it to min height removes any scrollbars in the modal from my tests.

src/components/forms/YamlForm.tsx
```diff
@@ -25,27 +27,40 @@ const YamlForm: FC<Props> = ({
...
+
+    // Calculate target height: clamp to max available, add 2px buffer to prevent scrollbar flickering
+    const targetHeight = Math.min(contentHeight, maxAvailableHeight) + 2;
+
+    containerRef.current.style.height = `${targetHeight}px`;
+
+    // Reflow again to fit the editor into the new container size
     editor.layout();
   };
+
```
edlerd: I think the changes here are not needed, can we restore the old version?

src/components/forms/YamlForm.tsx
```diff
@@ -25,27 +27,40 @@ const YamlForm: FC<Props> = ({
   setYaml,
   children,
   autoResize = false,
+  maxHeightBuffer = 0,
```
edlerd:
```suggestion
```
this should not be needed

src/components/forms/YamlForm.tsx
```diff
@@ -16,6 +17,7 @@ interface Props {
   setYaml?: (text: string) => void;
   children?: ReactNode;
   autoResize?: boolean;
+  maxHeightBuffer?: number;
```
edlerd:
```suggestion
```

src/components/forms/YamlModal.tsx
edlerd: I looked into this in detail and see other comments for a possible solution.

edlerd (approved):
LGTM, good work and thanks for the iteration to make it perfect :)

PR #1677, opened 2025-12-12
feat(nic device configuration): Revamp read mode

kimanhou:
## Done

- Instance/profile nic device configuration: revamp read mode
- Apply similar changes to disk device for consistency
- Apply "From profile" changes and blue dot to gpu, proxy and other devices for consistency
- Fixes on cloud init configuration: truncate profile chip, add margin left

## QA

1. Run the LXD-UI:
    - This screen is used in: CreateInstance, EditInstance, CreateProfile, EditProfile
    - Go to any of the above and make sure the information are displayed correctly.
    - On an inherited device, you should be abled to create an override, clear the override, detach and reattach the device. Make sure those changes are reflected correctly.
    - On custom device, you should be able to attache a new device, edit an existing device, detach an existing device.

## Screenshots

Network device - Inherited
https://github.com/user-attachments/assets/970b2bd6-d2be-4d8f-8212-bdb19221640a

Network device - Custom
https://github.com/user-attachments/assets/d78f12f4-c5ac-4364-a0e8-3d32aa870661

Disk device
https://github.com/user-attachments/assets/1c66864e-8326-460b-b04c-23f9a37e4f06


edlerd:
Good start, for custom network much more readable than before!

I think we should remove the border around the nic devices and use the button without border. That is a bit more minimalistic and in line with other pages. Wdyt?

src/components/forms/NetworkDevicesForm/read/NetworkDeviceFormInherited.tsx
```diff
@@ -0,0 +1,193 @@
...
+import NetworkDeviceCard from "components/forms/NetworkDevicesForm/read/NetworkDeviceCard";
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  inheritedNetworkDevices: InheritedNetwork[];
+  project: string;
+  managedNetworks: LxdNetwork[];
+}
+
+const NetworkDeviceFormInherited: FC<Props> = ({
```
edlerd, line 21: I think those nested tables are way too busy, especially if many nic devices are inherited.

image


You had an idea to hide the inherited values once overridden. I think that might make this easier to display. We could use the same style as for the "Custom network devices" section, with some more buttons. Maybe change the qualifier for those to "from profile (foo) with local override". We only need a way to display a detached inherited device, but that should be easy too with muted text and strike-through values. Wdyt?

kimanhou: If we don't need to display the inherited values once overriden, it can simplify the code a lot.

image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceIPAddressRead.tsx
```diff
@@ -22,15 +18,8 @@ export const NetworkDeviceIPAddressRead: FC<Props> = ({
...
-      <div className="ip-content">
-        <>
-          <div>{family}</div>
-          <div className="mono-font">
-            <b>{deviceIP || "dynamic"}</b>
-          </div>
-        </>
-      </div>
-    )
+    networkIP !== "none" && <b className="mono-font">{deviceIP || "dynamic"}</b>
```
edlerd:
```suggestion
    <b className="mono-font">{deviceIP || "dynamic"}</b>
```
Is this guard needed? I think the check is now at a higher level?

src/components/forms/NetworkDevicesForm/read/NetworkDeviceCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+  hasChanges: boolean;
+  rows: MainTableRow[];
+  isDetached?: boolean;
+  isInherited?: boolean;
+  sourceProfile?: ReactNode;
+  actions?: ReactNode;
+  className?: string;
+}
+
+const NetworkDeviceCard: FC<Props> = ({
```
edlerd:
```suggestion
const NetworkDevice: FC<Props> = ({
```
The old name seems to be more concise and precise.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceActionButtons.tsx
```diff
@@ -60,56 +60,27 @@ const NetworkDeviceActionButtons: FC<Props> = ({
...
-          disabled={isDisabled}
-        >
-          <Icon name="disconnect" />
-          <span>Detach</span>
-        </Button>
-      )}
-
-      {hasNicOverride && (
-        <>
+    <div className="network-device-actions p-segmented-control">
```
edlerd, line 183: I think those buttons are way too many and the positioning needs more alignment.

image

1. How about moving the edit button next to the values and only show the small pencil icon? They will all open the same side panel, but focus on the field the user clicked on.
2. I think we should only show one button on the right of the device name row. Those are Detach or Reattach or Clear. But we never show more than one button at a time and the buttons all have the same alignment vertically.

See below from disk devices for where the ideas are coming from:

image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceIPAddressRead.tsx
kimanhou: I removed it at the higher level, so this one should be needed 🤔

src/components/forms/NetworkDevicesForm/read/NetworkDeviceActionButtons.tsx
kimanhou: 1. I am not a fan of one Edit button for each field that can be editable. That's a lot of pencil icons. It makes sense in disk since the editting happens inline. But for network device, I would really like to keep one Edit button to open the panel.
2. It means that we lose the ability to Detach in one click. You would have to Clear then Detach. 
If we want one "thing" next to the device name to avoid clutter, we can put all the buttons in a dropdown menu called Actions or a kebab menu with 3 dots. Wdyt ?

edlerd: Maybe it will look better having the buttons below each other instead of in line?

kimanhou: image

edlerd: I think collapsing is making the interaction harder and hides the options the user has. We have enough space to show all actions below each other, no?

kimanhou: Sorry, didn't see your previous comment (before my screenshot).
Yes, the actions could be on top of each other. I just need to make sure they never overlap with the configuration table underneath.

kimanhou: I am not convinved. Either the dropdown menu or inline look better to me.

image

kimanhou: What about 3 dots instead of Actions ? It hides the buttons but the UI is cleaner.

image

kimanhou: I put back the action buttons inline and aligned them. But then because of the source profile, they end up 6 km away from the device name.

Screenshot from 2025-12-16 17-09-50


Also should we align them with the ones in Custom section ?
image

edlerd: Yes, inline is probably the best for the actions as it saves us one click compared to the dropdown. And we have enough space to lie the buttons out. Yes I agree we should align everything between custom and inherited to have a consistent flow.

The "from profile" should align with the values below. This qualifier might span into multiple lines if too long. Such as the "with local override" could be in a new line.
I see you currently instantiate one table per device. I think the better approach is to have one table for all  inherited network devices and a second one for custom network devices. The distribution is:
First column: name and labels (Network, ACLs, etc)
Second column: The qualifier (from profile...) and the values (lxdbr)
Third column: action buttons.
This is the same structure that have on the disk devices, gpu devices and for the other devices. So all device sections are guaranteed to have the same spacing. In this structure, the buttons move closer to the device names and are nicely aligned between all sections.

kimanhou: I put the device name outside of the configuration table because I didn't manage to get the indentation right. I figured it out in the end, I just needed to add a wrapper in the configuration cell.

Screenshot from 2025-12-17 10-39-15
Screenshot from 2025-12-17 10-39-30

edlerd: Yeah I think the buttons align better now and the structure of content makes it also easier to read.

edlerd: There is still some indentation between the "Inherited network devices" heading and the device names. I think the device names should be on the same vertical line as the headings.

kimanhou: It's for the blue dot when the device has changes. I don't want the device names to be misaligned when one device has changes and the other one doesn't.

edlerd: Maybe better to give the blue dot a negative margin to keep the text aligned. Wdyt?

kimanhou: The configuration table has `overflow: hidden` a bit everywhere by default so I have to fight against that. But I think it's doable

kimanhou: Actually I couldn't make the blue dot appear. So I added a `margin-left` on the heading.

image

edlerd: > added a margin-left on the heading

We shouldn't change it, so that between device types and other pages the positions are aligned.

kimanhou: I can't have the header and the device name aligned with the blue dot on the left.

kimanhou: Unless I set `overflow: visible` on the gridcell and content-details. I'm not sure what are the consequences of this.

edlerd: When there are many nic devices, the network section does not scroll any more now.

kimanhou: Leia sent me this from Landscape

image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceFormInherited.tsx
```diff
@@ -0,0 +1,103 @@
...
+        : ({ ...device.network, name: device.key } as LxdNicDevice);
+
+    const deviceModified =
+      overrideDevice && isDeviceModified(formik, device.key);
+    const initialOverride = formik.initialValues.devices.find(
+      (t) => t.name === device.key,
+    );
+    const hasOverrideBeenRemoved = initialOverride && !overrideDevice;
+
+    return getNetworkDeviceRows({
```
edlerd, line 51: I think the visibility rules for the rows can be better:

- We might want to hide ACL if none are assigned.
- On inherited devices the `IPv4` and `IPV6` are sometimes hidden where they should say dynamic.

image

- If either v4 or v6 is set to none, the row should probably be hidden. currently it renders empty.

image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceFormInherited.tsx
```diff
@@ -0,0 +1,103 @@
...
+            </>
+          )}
+          {isDetached && (
+            <>
+              <br />
+              (detached)
+            </>
+          )}
+        </>
+      ),
```
edlerd, line 79: I think moving this into its own row is a good idea.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceFormInherited.tsx
```diff
@@ -0,0 +1,103 @@
...
+      actions: (
+        <NetworkDeviceActionButtons
+          formik={formik}
+          device={overrideDevice}
+          inheritedDevice={device}
+        />
+      ),
```
edlerd, line 86: I think the actions in the 3rd column, such as on the disk devices, might look a bit better and cleaner. It helps identify the available actions quickly and helps identity that "this is a device heading". They can be left aligned in the third column so they are not too far from the device name.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceFormInherited.tsx
kimanhou: > We might want to hide ACL if none are assigned.

I thought you prefer to always display things, so that the user can have a mental map of what is there. But okay, I prefer hiding things where they are empty to avoid clutter.

> On inherited devices the IPv4 and IPV6 are sometimes hidden where they should say dynamic.

It's because they are hidden in `EditProfile` and `CreateProfile`, so I assume they can't be inherited. Should I just display `dynamic` for all inherited network devices ?

edlerd: > I thought you prefer to always display things, so that the user can have a mental map of what is there. But okay, I prefer hiding things where they are empty to avoid clutter.

True, maybe keep it in then. I am not sure what is best here.

> It's because they are hidden in `EditProfile` and `CreateProfile`, so I assume they can't be inherited. Should I just display `dynamic` for all inherited network devices ?

I think on profiles, we should hide them in read mode (also for custom devices on the profile). For inherited, I think they should react to whatever the profile sets. Like the profile could define an ipv4 and in that case we should show it, but otherwise dynamic. We don't allow in the UI to set a ip on profiles, but LXD would allow it via cli or yaml editor.

kimanhou: > True, maybe keep it in then. I am not sure what is best here.

I removed them.

> I think on profiles, we should hide them in read mode (also for custom devices on the profile). 

IP addresses are always hidden on profiles.
```
showIpAddresses: formik.values.entityType === "instance"
```

> LXD would allow it via cli or yaml editor.

Ah, I didn't know that. I put back the IP addresses for inherited, so now they are shown for both inherited and custom for instances only.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  rows.push(
+    getConfigurationRowBase({
+      className: "device-first-row",
+      configuration: (
+        <NetworkDeviceName
+          name={device.name || ""}
+          hasChanges={hasChanges}
+          isDetached={isDetached}
+        />
```
edlerd: Can we show the "has changes" icon after the name?

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -0,0 +1,219 @@
...
+    }),
+  );
+
+  if (sourceProfile) {
+    rows.push(
+      getConfigurationRowBase({
+        className: "no-border-top",
+        configuration: (
+          <div className="u-text--muted">
+            <Label>From</Label>
```
edlerd: This and other qualifier should not be labels as this is only read mode.

src/sass/_network_device_form.scss
```diff
@@ -0,0 +1,28 @@
...
+    .content-details {
+      .inherited-network-device-configuration-table,
+      .custom-network-device-configuration-table {
+        thead {
+          display: none;
+        }
+
+        .inherited {
+          width: unset;
+        }
```
edlerd:
```suggestion
```
Should we rely on the default here? Seems like it brings the action buttons a bit closer to the instance name on large screens.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -0,0 +1,219 @@
...
+        />
+      ),
+      inherited: null,
+      override: actions,
+    }),
+  );
+
+  if (sourceProfile) {
+    rows.push(
+      getConfigurationRowBase({
```
edlerd, line 75: We should move the "from (source)" to a dedicated row also on other devices (like storage volumes, gpus, proxy and other) to have this consistently presented.

kimanhou: Should we keep the string "profile" or would the chip be enough ?
image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
kimanhou: Yes
image

src/components/forms/DeviceName.tsx
```diff
@@ -6,31 +6,25 @@ interface Props {
...
         })}
       >
         {name}
       </b>
+      {hasChanges && (
+        <Icon
+          name="status-in-progress-small"
+          title="This device has unsaved changes"
+        />
+      )}
```
edlerd, line 27: Should we roll it out to other devices types?

src/components/forms/GPUDeviceForm.tsx
edlerd, line 1: On local fields with direct input, like the highlighted one below, we should keep them as labels for the form elements. Just inherited or read only qualifiers, should omit being labels:

image

src/components/forms/DiskDeviceForm.tsx
edlerd, line 1: Can we indent the Pool and Size qualifiers for root storage? And move the "from default profile" into a dedicated row (if there is inheritance in play).

image

edlerd (commented):
Design is getting better and better :)

src/components/forms/DiskDeviceForm.tsx
kimanhou: I also removed the configuration table headers and added "Edit" and "Clear" in the button.

image

image

edlerd: Thanks! the Edit / Clear buttons seem a bit too low aligned comparing to the "Root storage" line.

src/components/forms/GPUDeviceForm.tsx
kimanhou: Mmh I didn't touch this. GPUDeviceForm uses `getInheritedDeviceRow`, which I did modify but only to always put the label as muted text. I left the choice between `Label` or `p` or `div` as it was.

Which local fields have direct input ? How can you visually tell if it's a label or not ?

kimanhou: It looks like Identify by was always a `div` and not a `label`. I changed it to be muted tho, maybe I can revert that back.

image

src/components/forms/DeviceName.tsx
kimanhou: I tried but it doesn't work well. Needs some more work on `isDeviceModified` tomorrow.

src/components/forms/GPUDeviceForm.tsx
edlerd: Oh, you are right! I thought it was fallout from other changes. Then ignore my comment.

src/components/forms/DiskDeviceForm.tsx
edlerd: Shall we remove one of the two "Root storage" sections? Either the heading or the device name. If we remove the device name, we can move the Edit/Clear buttons into the row with the "From profile".

src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -96,16 +118,39 @@ const DiskDeviceFormRoot: FC<Props> = ({ formik, pools, profiles }) => {
...
+                      value={getProfileFromSource(inheritSource)}
+                      to={`/ui/project/${encodeURIComponent(project)}/profile/${encodeURIComponent(getProfileFromSource(inheritSource))}`}
+                      className={classnames({
+                        "u-text--line-through": hasRootStorage,
+                      })}
+                    />
+                  ),
+                  readOnly: true,
+                  noMonoFont: true,
+                }),
```
edlerd: Shall we extract a helper for this, something like `const getInheritedSourceRow = (project, profile, hasOverride) => ...`?

src/components/forms/DiskDeviceForm.tsx
kimanhou: But then where do we put the blue dot ?

edlerd: We could put it after the heading?

kimanhou: I also put the buttons with the heading

image

edlerd: Looks great!

Does the button have some custom styling applied? It still sits a bit too low in my opinion when checking the screenshot:

image

kimanhou: Yes, otherwise the button sits too high
image

I apply a padding-top, but I was missing 1.5px. Better ?
image

edlerd: Looks perfect, thanks :)

src/components/forms/DeviceName.tsx
kimanhou: It works now

kimanhou: Shold we add "with local override" and "(detached)" to other device types too ?

src/components/forms/DiskDeviceForm.tsx
```diff
@@ -53,10 +53,16 @@ const DiskDeviceForm: FC<Props> = ({ formik, project }) => {
       <ScrollableForm>
         {/* hidden submit to enable enter key in inputs */}
         <Input type="submit" hidden value="Hidden input" />
-        <DiskDeviceFormRoot formik={formik} pools={pools} profiles={profiles} />
+        <DiskDeviceFormRoot
```
edlerd, line 56: Can we add the "with local override" and "detached" copy to the "From profile" rows? This applies to root disk and inherited disk devices, also gpu, proxy and other?

image

src/components/forms/DeviceName.tsx
edlerd: Yes, please.

Why is it "(detached)" instead of "detached"? Should we change it?

kimanhou: I cannot find a good way to display "detached". Which option do you prefer ?

1.
Screenshot from 2026-01-05 09-21-26

2.
Screenshot from 2026-01-05 09-21-41

3.
Screenshot from 2026-01-05 09-22-04

4.
image

5.
image

src/components/forms/DiskDeviceForm.tsx
kimanhou: Root disk can only be overridden, not detached.
Inherited disk, gpu, proxy and others can only be detached, not overridden. 
Am I correct ?

Should we match the display of "with local override" with "detached" ?
image

edlerd: > Root disk can only be overridden, not detached.
> Inherited disk, gpu, proxy and others can only be detached, not overridden.
> Am I correct ?

Yes, I think so.

> Should we match the display of "with local override" with "detached" ?

Yes, I think that is more clear and consistent with the copy "with local override".

src/components/forms/DeviceName.tsx
edlerd: I lean towards 4. or 5. Shall it be the same styling for the "with local override"?

edlerd: In the detached vie, the values for IPv4 and IPv6 still show as if they are applied. Maybe they can be hidden or strike through as well?

kimanhou: > I lean towards 4. or 5. Shall it be the same styling for the "with local override"?

I chose 5 and applied the same styling for both.

> In the detached vie, the values for IPv4 and IPv6 still show as if they are applied. Maybe they can be hidden or strike through as well?

Indeed, thank you for catching this. It's fixed now.
image

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -61,86 +49,33 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
+          inheritedNetworkDevices={inheritedNetworks}
+          project={project}
+          managedNetworks={managedNetworks}
+        />
+        <NetworkDeviceFormCustom
+          formik={formik}
+          inheritedNetworkDevices={inheritedNetworks}
+          project={project}
+          managedNetworks={managedNetworks}
+        />
```
edlerd, line 66: Detaching network devices, leave behind a `none` device, for a device that is not inherited. We should remove the device and only leave a `none` device for inherited devices.


https://github.com/user-attachments/assets/2cc3edb6-d885-440f-b09e-c18fc8677b02

src/components/forms/NetworkDevicesForm/NetworkDevicesForm.tsx
```diff
@@ -61,86 +49,33 @@ const NetworkDevicesForm: FC<Props> = ({ formik, project }) => {
...
+        <Button
+          onClick={panelParams.openCreateNetworkDevice}
+          type="button"
+          hasIcon
+          disabled={!!formik.values.editRestriction}
+          title={formik.values.editRestriction}
+        >
+          <Icon name="plus" />
+          <span>Attach network</span>
+        </Button>
```
edlerd, line 76: The name generation and/or validation for new networks is hitting some edge case in this video below. It should create a free name and validation should pass in that case.

https://github.com/user-attachments/assets/bd565a41-19a3-4ee7-85ff-681b9eff2aff

src/components/forms/DiskDeviceFormCustom.tsx
edlerd, line 1: There is an older bug in here, where when a custom volume is edited and the mount point matches `/`, then the field above changes type. Not sure what triggers it, maybe you can find a fix?


https://github.com/user-attachments/assets/319b3dff-2879-43f5-b0dd-1465515aab3a

src/components/forms/InheritedDeviceRow.tsx
```diff
@@ -17,6 +18,8 @@ interface Props {
   isDeactivated?: boolean;
   className?: string;
   disabledReason?: string;
+  noMonoFont?: boolean;
```
edlerd:
```suggestion
  monoFont?: boolean;
```
Let's keep the name positive to make it easier to read.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -0,0 +1,211 @@
...
+    getConfigurationRowBase({
+      className: "device-first-row",
+      configuration: (
+        <Label>
+          <DeviceName
+            name={device.name || ""}
+            hasChanges={hasChanges}
+            isDetached={isDetached}
+          />
+        </Label>
```
edlerd:
```suggestion
          <DeviceName
            name={device.name || ""}
            hasChanges={hasChanges}
            isDetached={isDetached}
          />
```
There is no form element referenced, so I think this shouldn't be a label.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -0,0 +1,211 @@
...
+        inherited: (
+          <div>
+            <ResourceLink
+              type="network"
+              value={device.network}
+              to={getUrl({
+                project,
+                entityType: "network",
+                name: device.network,
+              })}
```
edlerd:
```suggestion
              to={`/ui/project/${encodeURIComponent(project || "default")}/network/${encodeURIComponent(device.network)}`}
```
I think it is better to inline this value. It allows you to find all links in the code by searching for `/network/` and the helper is only used once per entity type.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -0,0 +1,211 @@
...
+                items={acls.map((acl) => (
+                  <ResourceLink
+                    key={acl}
+                    type="network-acl"
+                    value={acl}
+                    to={getUrl({
+                      project,
+                      entityType: "network-acl",
+                      name: acl,
+                    })}
```
edlerd: Here as well, probably easier to maintain if inlined than with the get url function.

src/sass/_rename_device.scss
```diff
@@ -3,5 +3,6 @@
 
   > :first-child {
     margin-right: $sph--small;
+    max-width: 20rem;
```
edlerd, line 6: Why is this needed? I think the parent width should constrain the input width already.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceFormInherited.tsx
```diff
@@ -0,0 +1,101 @@
...
+            </>
+          )}
+        </>
+      ),
+      actions: (
+        <NetworkDeviceActionButtons
+          formik={formik}
+          device={overrideDevice}
+          inheritedDevice={device}
+        />
```
edlerd, line 85: There are some edge cases with responsiveness where the buttons get truncated. In other pages the table starts to scroll horizontally at a breakpoint where the action buttons are fully readable. Let's do the same here. But we need some adjustment as we have a longer button row.


https://github.com/user-attachments/assets/9f94b62b-4074-49e2-adf1-69bc7fcd2cdb

src/sass/_rename_device.scss
kimanhou: It's the crop the name when it's really long in read mode. Otherwise it overflows on top of the Detach button.

Without it
Screenshot from 2026-01-07 17-34-40

With it
Screenshot from 2026-01-07 17-35-08

src/components/forms/DiskDeviceFormCustom.tsx
edlerd: Thanks for fixing the drive by!

src/components/forms/NetworkDevicesForm/read/NetworkDeviceFormInherited.tsx
edlerd: Thanks for the fix, responsiveness is working well on the custom networks. However for the inherited devices it seems the breakpoints and not yet reserving enough space

image

kimanhou: Is it better now ? I updated the `min-width` for both inherited and custom, to keep the buttons vertically aligned among all rows.

edlerd: yes, thanks :)

edlerd (approved):
Very nice, good work!

PR #1683, opened 2025-12-16
feat: [WD-31000] Migrate storage volume to another cluster

Kxiru:
## Done

- [x] Restructure files for the new nested modal system
- [x] Recreate old Pool migration flow
- [x] Create new cluster member migration flow
- [x] Fix any tests
- [x] Migrate storage volumes across projects

Outstanding questions / issues
- ~~Upon successful migration, the cluster member page does not update the pool shown. Query key issue? Should the UI navigate to the new storage volume link, as the old link is now ineffective. Refreshing the page at this stage show an error (expected).~~ [FIXED]
- ~~Cluster member migration at the API level doesn't seem to be working. Is there a specific way the API should be structured?~~ [FIXED]

## QA

1. Run the LXD-UI:
    - Navigate to Storage Volume > Any Custom Storage Volume > Ensure that the two migration flows, work.
    - A clustered environment is required.

## Screenshots (may be outdated)

image
image
image

### DIFF
--- a/src/api/storage-volumes.tsx
+++ b/src/api/storage-volumes.tsx
@@ -271,12 +271,13 @@ export const deleteStorageVolumeBulk = async (
 
 export const migrateStorageVolume = async (
   volume: LxdStorageVolume,
-  targetPool: string,
-  targetProject: string,
+  project: string,
+  targetPool?: string,
   target?: string,
+  targetProject?: string,
 ): Promise<LxdOperationResponse> => {
   const params = new URLSearchParams();
-  params.set("project", targetProject);
+  params.set("project", project);
   addTarget(params, target);
 
   return fetch(
@@ -287,8 +288,9 @@ export const migrateStorageVolume = async (
         "Content-Type": "application/json",
       },
       body: JSON.stringify({
+        project: targetProject,
         name: volume.name,
-        pool: targetPool,
+        pool: targetPool ?? volume.pool,
       }),
     },
   )
--- a/src/pages/storage/MigrateVolumeBtn.tsx
+++ b/src/pages/storage/MigrateVolumeBtn.tsx
@@ -1,177 +1,51 @@
 import type { FC } from "react";
-import { useState } from "react";
-import {
-  ActionButton,
-  Icon,
-  usePortal,
-  useToastNotification,
-} from "@canonical/react-components";
-import { useQueryClient } from "@tanstack/react-query";
-import { queryKeys } from "util/queryKeys";
-import { useEventQueue } from "context/eventQueue";
+import { ActionButton, Icon, usePortal } from "@canonical/react-components";
 import type { LxdStorageVolume } from "types/storage";
 import MigrateVolumeModal from "./MigrateVolumeModal";
-import { useNavigate } from "react-router-dom";
-import ResourceLabel from "components/ResourceLabel";
-import ResourceLink from "components/ResourceLink";
-import { migrateStorageVolume } from "api/storage-volumes";
 import { useStorageVolumeEntitlements } from "util/entitlements/storage-volumes";
-import { hasLocation } from "util/storageVolume";
-import VolumeLinkChip from "pages/storage/VolumeLinkChip";
 import classNames from "classnames";
 
 interface Props {
   volume: LxdStorageVolume;
-  project: string;
   classname?: string;
   onClose?: () => void;
 }
 
-const MigrateVolumeBtn: FC<Props> = ({
-  volume,
-  project,
-  classname,
-  onClose,
-}) => {
-  const eventQueue = useEventQueue();
-  const toastNotify = useToastNotification();
+const MigrateVolumeBtn: FC<Props> = ({ volume, classname, onClose }) => {
   const { openPortal, closePortal, isOpen, Portal } = usePortal();
-  const queryClient = useQueryClient();
-  const navigate = useNavigate();
-  const [isVolumeLoading, setVolumeLoading] = useState(false);
   const { canEditVolume } = useStorageVolumeEntitlements();
-
-  const handleSuccess = (
-    newTarget: string,
-    storageVolume: LxdStorageVolume,
-  ) => {
-    const memberPath = hasLocation(volume)
-      ? `/member/${encodeURIComponent(volume.location)}`
-      : "";
-    const oldVolumeUrl = `/ui/project/${encodeURIComponent(storageVolume.project)}/storage/pool/${encodeURIComponent(storageVolume.pool)}${memberPath}/volumes/${encodeURIComponent(storageVolume.type)}/${encodeURIComponent(storageVolume.name)}`;
-    const newVolumeUrl = `/ui/project/${encodeURIComponent(storageVolume.project)}/storage/pool/${encodeURIComponent(newTarget)}${memberPath}/volumes/${encodeURIComponent(storageVolume.type)}/${encodeURIComponent(storageVolume.name)}`;
-
-    const volumeLink = (
-      <ResourceLink
-        type="volume"
-        value={storageVolume.name}
-        to={newVolumeUrl}
-      />
-    );
-    const poolLink = (
-      <ResourceLink
-        type="pool"
-        value={newTarget}
-        to={`/ui/project/${encodeURIComponent(storageVolume.project)}/storage/pool/${encodeURIComponent(newTarget)}`}
-      />
-    );
-    toastNotify.success(
-      <>
-        Volume {volumeLink} successfully migrated to pool {poolLink}
-      </>,
-    );
-
-    if (window.location.pathname.startsWith(oldVolumeUrl)) {
-      navigate(newVolumeUrl);
-    }
-  };
-
-  const notifyFailure = (
-    e: unknown,
-    volumeName: string,
-    targetPool: string,
-  ) => {
-    setVolumeLoading(false);
-    toastNotify.failure(
-      `Migration failed for volume ${volumeName} to pool ${targetPool}`,
-      e,
-      <VolumeLinkChip volume={volume} />,
-    );
-  };
-
-  const handleFailure = (
-    msg: string,
-    storageVolume: string,
-    targetPool: string,
-  ) => {
-    notifyFailure(new Error(msg), storageVolume, targetPool);
-  };
-
-  const handleFinish = () => {
-    queryClient.invalidateQueries({
-      queryKey: [queryKeys.storage, volume.name],
-    });
-    setVolumeLoading(false);
-  };
-
-  const handleMigrate = (targetPool: string) => {
-    setVolumeLoading(true);
-    migrateStorageVolume(volume, targetPool, volume.project, volume.location)
-      .then((operation) => {
-        eventQueue.set(
-          operation.metadata.id,
-          () => {
-            handleSuccess(targetPool, volume);
-          },
-          (err) => {
-            handleFailure(err, volume.name, targetPool);
-          },
-          handleFinish,
-        );
-        const volumeLabel = (
-          <ResourceLabel bold type="volume" value={volume.name} />
-        );
-        const poolLink = (
-          <ResourceLink
-            type="pool"
-            value={targetPool}
-            to={`/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(targetPool)}`}
-          />
-        );
-        toastNotify.info(
-          <>
-            Migration started for volume {volumeLabel} to pool {poolLink}
-          </>,
-        );
-        queryClient.invalidateQueries({
-          queryKey: [queryKeys.storage, volume.name, project],
-        });
-      })
-      .catch((e) => {
-        notifyFailure(e, volume.name, targetPool);
-      })
-      .finally(() => {
-        handleClose();
-      });
-  };
+  const isEmpty = volume.used_by?.length === 0;
 
   const handleClose = () => {
     closePortal();
     onClose?.();
   };
 
+  const getTitle = () => {
+    if (!canEditVolume(volume)) {
+      return "You do not have permission to migrate this volume";
+    }
+
+    if (isEmpty) {
+      return "Migrate volume";
+    }
+
+    return "Volume is in use";
+  };
+
   return (
     <>
       {isOpen && (
         <Portal>
-          <MigrateVolumeModal
-            close={handleClose}
-            migrate={handleMigrate}
-            storageVolume={volume}
-          />
+          <MigrateVolumeModal close={handleClose} volume={volume} />
         </Portal>
       )}
       <ActionButton
         onClick={openPortal}
         type="button"
         className={classNames("u-no-margin--bottom has-icon", classname)}
-        loading={isVolumeLoading}
-        disabled={!canEditVolume(volume) || isVolumeLoading}
-        title={
-          canEditVolume(volume)
-            ? "Migrate volume"
-            : "You do not have permission to migrate this volume"
-        }
+        disabled={!canEditVolume(volume) || !isEmpty}
+        title={getTitle()}
       >
         <Icon name="machines" />
         <span>Migrate</span>
--- a/src/pages/storage/MigrateVolumeModal.tsx
+++ b/src/pages/storage/MigrateVolumeModal.tsx
@@ -1,17 +1,36 @@
-import type { FC, KeyboardEvent } from "react";
+import type { FC, KeyboardEvent, ReactNode } from "react";
 import { useState } from "react";
-import { ActionButton, Button, Modal } from "@canonical/react-components";
+import { Modal } from "@canonical/react-components";
 import type { LxdStorageVolume } from "types/storage";
-import StoragePoolSelectTable from "./StoragePoolSelectTable";
+import FormLink from "components/FormLink";
+import { useIsClustered } from "context/useIsClustered";
+import {
+  useStorageVolumeMigration,
+  type VolumeMigrationType,
+} from "util/storageVolumeMigration";
+import StorageVolumePoolMigration from "./StorageVolumePoolMigration";
+import BackLink from "components/BackLink";
+import StorageVolumeClusterMemberMigration from "./StorageVolumeClusterMemberMigration";
+import StorageVolumeProjectMigration from "./StorageVolumeProjectMigration";
+import { isClusterLocalDriver } from "util/storagePool";
+import { useStoragePool } from "context/useStoragePools";
 
 interface Props {
   close: () => void;
-  migrate: (target: string) => void;
-  storageVolume: LxdStorageVolume;
+  volume: LxdStorageVolume;
 }
 
-const MigrateVolumeModal: FC<Props> = ({ close, migrate, storageVolume }) => {
-  const [selectedPool, setSelectedPool] = useState("");
+const MigrateVolumeModal: FC<Props> = ({ close, volume }) => {
+  const [type, setType] = useState<VolumeMigrationType>("");
+  const [target, setTarget] = useState("");
+  const isClustered = useIsClustered();
+  const { data: pool } = useStoragePool(volume.pool);
+  const { handleMigrate, handleMemberMigrate } = useStorageVolumeMigration({
+    close,
+    volume,
+    type,
+    target,
+  });
 
   const handleEscKey = (e: KeyboardEvent<HTMLElement>) => {
     if (e.key === "Escape") {
@@ -19,67 +38,121 @@ const MigrateVolumeModal: FC<Props> = ({ close, migrate, storageVolume }) => {
     }
   };
 
-  const handleCancel = () => {
-    if (selectedPool) {
-      setSelectedPool("");
+  const handleGoBack = () => {
+    // if target is set, we are on the confirmation stage
+    if (target) {
+      setTarget("");
       return;
     }
-
-    close();
+    // if type is set, we are on migration target selection stage
+    if (type) {
+      setType("");
+      return;
+    }
   };
 
-  const modalTitle = selectedPool
-    ? "Confirm migration"
-    : `Choose storage pool for volume ${storageVolume.name}`;
-
-  const summary = (
-    <div className="migrate-instance-summary">
-      <p>
-        This will migrate volume <strong>{storageVolume.name}</strong> to
-        storage pool <b>{selectedPool}</b>.
-      </p>
-    </div>
-  );
+  const renderTitle = (): ReactNode => {
+    if (!type) {
+      return "Choose migration method";
+    }
+    if (target) {
+      return (
+        <BackLink
+          title="Confirm migration"
+          linkText={`Choose ${type}`}
+          onClick={handleGoBack}
+        />
+      );
+    }
+    return (
+      <BackLink
+        title={
+          <>
+            Choose {type} for volume <strong>{volume.name}</strong>
+          </>
+        }
+        linkText="Choose migration method"
+        onClick={handleGoBack}
+      />
+    );
+  };
 
   return (
     <Modal
       close={close}
       className="migrate-instance-modal"
-      title={modalTitle}
-      buttonRow={
-        <div id="migrate-instance-actions">
-          <Button
-            className="u-no-margin--bottom"
-            type="button"
-            aria-label="cancel migrate"
-            appearance="base"
-            onClick={handleCancel}
-          >
-            Cancel
-          </Button>
-          <ActionButton
-            appearance="positive"
-            className="u-no-margin--bottom"
-            onClick={() => {
-              migrate(selectedPool);
-            }}
-            disabled={!selectedPool}
-          >
-            Migrate
-          </ActionButton>
-        </div>
-      }
       onKeyDown={handleEscKey}
     >
-      {selectedPool ? (
-        summary
-      ) : (
-        <StoragePoolSelectTable
-          onSelect={setSelectedPool}
-          disablePool={{
-            name: storageVolume.pool,
-            reason: "Volume is already in this pool",
-          }}
+      <header className="p-modal__header">
+        <h2
+          className="p-modal__title"
+          key={type ? (target ? "confirm" : "select") : "start"}
+          id="migrate-title"
+        >
+          {renderTitle()}
+        </h2>
+        <button
+          className="p-modal__close"
+          aria-label="Close active modal"
+          onClick={close}
+        >
+          Close
+        </button>
+      </header>
+      {!type && (
+        <div className="choose-migration-type">
+          {isClustered && isClusterLocalDriver(pool?.driver || "") && (
+            <FormLink
+              icon="cluster-host"
+              title="Migrate volume to a different cluster member"
+              onClick={() => {
+                setType("cluster member");
+              }}
+            />
+          )}
+          <FormLink
+            icon="switcher-dashboard"
+            title={`Move volume to a different storage pool`}
+            onClick={() => {
+              setType("pool");
+            }}
+          />
+          <FormLink
+            icon="folder"
+            title="Move volume to a different project"
+            onClick={() => {
+              setType("project");
+            }}
+          />
+        </div>
+      )}
+
+      {type === "cluster member" && (
+        <StorageVolumeClusterMemberMigration
+          volume={volume}
+          onSelect={setTarget}
+          targetMember={target}
+          onCancel={handleGoBack}
+          migrate={handleMemberMigrate}
+        />
+      )}
+
+      {type === "pool" && (
+        <StorageVolumePoolMigration
+          volume={volume}
+          onSelect={setTarget}
+          onCancel={handleGoBack}
+          migrate={handleMigrate}
+        />
+      )}
+
+      {type === "project" && (
+        <StorageVolumeProjectMigration
+          volume={volume}
+          onSelect={setTarget}
+          targetProject={target}
+          onCancel={handleGoBack}
+          migrate={handleMigrate}
         />
       )}
     </Modal>
--- a/src/util/storageVolumeMigration.tsx
+++ b/src/util/storageVolumeMigration.tsx
@@ -1,0 +1,272 @@
+import { useToastNotification } from "@canonical/react-components";
+import { useQueryClient } from "@tanstack/react-query";
+import {
+  copyStorageVolume,
+  deleteStorageVolume,
+  migrateStorageVolume,
+} from "api/storage-volumes";
+import ResourceLink from "components/ResourceLink";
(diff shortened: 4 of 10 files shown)


src/pages/storage/MigrateVolumeModal.tsx
```diff
@@ -1,85 +1,147 @@
...
+              icon="cluster-host"
+              title="Migrate volume to a different cluster member"
+              onClick={() => {
+                setType("cluster member");
+              }}
+            />
+          )}
+          <FormLink
+            icon="folder"
+            title={`Choose storage pool for volume ${volume.name}`}
```
omarelkashef: I think we should also bold the volume name here like in the other places..

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const handleSuccess = () => {
+    let successMessage: ReactNode = "";
+
+    const memberPath = hasLocation(volume)
+      ? `/member/${encodeURIComponent(volume.location)}`
+      : "";
+    const oldVolumeUrl = `/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(volume.pool)}${memberPath}/volumes/${encodeURIComponent(volume.type)}/${encodeURIComponent(volume.name)}`;
+    const newVolumeUrl = `/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(target)}${memberPath}/volumes/${encodeURIComponent(volume.type)}/${encodeURIComponent(volume.name)}`;
+
+    const volumeLink = (
```
omarelkashef: If volumeLink and poolLink are only used in one place, we can define them in the blocks they are used.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,182 @@
...
+        navigate(newUrl);
+      }
+    }
+
+    toastNotify.success(successMessage);
+  };
+
+  const notifyFailure = (e: unknown) => {
+    let failureMessage = "";
+    if (type === "cluster member") {
```
omarelkashef, line 110: Should we standardize the messages? For example `Migration failed for volume ${volume.name} to cluster member `${target}`` and similarly for the project case.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,182 @@
...
+      );
+
+      const oldUrl = window.location.pathname;
+      const newUrl = oldUrl.replace(
+        `/project/${volume.project}/instance/${volume.name}`,
+        `/project/${target}/instance/${volume.name}`,
+      );
+      if (oldUrl !== newUrl) {
+        navigate(newUrl);
+      }
```
edlerd: This does not update when migrating to a different project. We should consider the new project in `newUrl`. 

Better use the `linkForVolumeDetail` helper to build volume URLs. This relates to all the places in this PR that handle a volume URL.

Volume URLs are more complicated than instance URLs. There are different cases such as volumes that need the pool and cluster member in the URL or in other cases just the pool.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,182 @@
...
+    let successMessage: ReactNode = "";
+
+    const memberPath = hasLocation(volume)
+      ? `/member/${encodeURIComponent(volume.location)}`
+      : "";
+    const oldVolumeUrl = `/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(volume.pool)}${memberPath}/volumes/${encodeURIComponent(volume.type)}/${encodeURIComponent(volume.name)}`;
+    const newVolumeUrl = `/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(target)}${memberPath}/volumes/${encodeURIComponent(volume.type)}/${encodeURIComponent(volume.name)}`;
+
+    const volumeLink = (
+      <ResourceLink type="volume" value={volume.name} to={newVolumeUrl} />
```
omarelkashef: What is the difference between `ResourceLink` here and `VolumeLinkChip` in other places? Do things break if we use `VolumeLinkChip` here?

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,182 @@
...
+      );
+      if (oldVolumeUrl !== newVolumeUrl) {
+        navigate(newVolumeUrl);
+      }
+    }
+
+    if (type === "project") {
+      successMessage = (
+        <>
+          Instance <VolumeLinkChip volume={{ ...volume, project: target }} />
```
edlerd:
```suggestion
          Volume <VolumeLinkChip volume={{ ...volume, project: target }} />
```

src/pages/storage/MigrateVolumeModal.tsx
```diff
@@ -1,85 +1,147 @@
...
+            <FormLink
+              icon="cluster-host"
+              title="Migrate volume to a different cluster member"
+              onClick={() => {
+                setType("cluster member");
+              }}
+            />
+          )}
+          <FormLink
+            icon="folder"
```
edlerd: We should use a different icon, see instance migration.

src/pages/instances/InstanceStoragePoolMigration.tsx
```diff
@@ -40,7 +40,7 @@ const InstanceStoragePoolMigration: FC<Props> = ({
           }}
         />
       )}
-      <footer id="migrate-instance-actions" className="p-modal__footer">
+      <footer id="migrate-volume-actions" className="p-modal__footer">
```
edlerd: This seems like a wrong rename, should this be reverted?

src/api/storage-volumes.tsx
```diff
@@ -287,8 +288,10 @@ export const migrateStorageVolume = async (
         "Content-Type": "application/json",
       },
       body: JSON.stringify({
+        project: targetProject,
+        migration: Boolean(target),
         name: volume.name,
-        pool: targetPool,
+        pool: targetPool ?? volume.pool,
       }),
```
edlerd, line 294: Somehow, this doesn't work for cluster member migration. I am not sure what the correct payload is to trigger it correctly. Hope we get some hints on how to initiate this.

edlerd: As discussed, the cluster member migration should do a copy of the volume to the new member and then delete the old volume.

src/pages/storage/MigrateVolumeModal.tsx
```diff
@@ -1,85 +1,147 @@
...
+        <div className="choose-migration-type">
+          {isClustered && (
+            <FormLink
+              icon="cluster-host"
+              title="Migrate volume to a different cluster member"
+              onClick={() => {
+                setType("cluster member");
+              }}
+            />
+          )}
```
edlerd: We need another guard here, adding to`isClustered`. Some storage pools are remote (not local) and on those the volume is not relating to a cluster member. So for those this option should be hidden. We can use `isClusterLocalDriver` on the pool driver as a guard.

src/util/storageVolumeMigration.tsx
Kxiru: They are used in multiple places which is why they have been extracted :)

src/pages/storage/MigrateVolumeModal.tsx
Kxiru: Within the migration modal, the convention is to keep these unemboldened (See instance migration)

src/util/storageVolumeMigration.tsx
Kxiru: I'm not sure what you mean? Could you provide examples for each case, as they are already case specific.

src/util/storageVolumeMigration.tsx
Kxiru: VolumeLinkChip is a dedicated component that is built upon ResourceLinks. Technically replacing it wouldn't "break" things, persay, but they are both one line imports, and both components have already been imported previously, so interchanging wouldn't be any more or less efficient.

src/util/storageVolumeMigration.tsx
Kxiru: @omarelkashef , suggested switching out ResourceLink for VolumeLinkChip, which has the minor addition of using the linkForVolumeDetail function you are suggesting. I'll replac this where possible.

src/pages/instances/InstanceStoragePoolMigration.tsx
Kxiru: I figured this was valid as it is a footer ID rather than a classname. Wdyt?

src/pages/storage/MigrateVolumeModal.tsx
Kxiru: Adding this makes this option only appear for instances, excluding custom storage volumes that are CMS. Was this your intention?

src/pages/instances/InstanceStoragePoolMigration.tsx
edlerd: There are several references to the old name: `migrate-instance-actions`.
This is not a volume migration component, so I was wondering why to rename it in this PR.

I suspect the change is breaking responsiveness?

src/pages/storage/MigrateVolumeModal.tsx
edlerd: When the storage pool is base on a ceph driver (or another driver that is cluster wide): then there is no migration to another cluster member possible. So we should not show the option in that case.

src/pages/instances/InstanceStoragePoolMigration.tsx
Kxiru: As yes, you are correct, thank you!

src/pages/storage/MigrateVolumeModal.tsx
Kxiru: I'll add it so you can take a look and see if that is what you wanted :)

Kxiru: Oh wait, I had already added it, not sure why it's not showing up in git. I'll push again :+1:

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,256 @@
...
+              volume.pool,
+              queryKeys.volumes,
+              volume.project,
+              volume.location,
+            ],
+          });
+          queryClient.invalidateQueries({
+            predicate: (query) => query.queryKey[0] === queryKeys.volumes,
+          });
+        });
```
edlerd: The sequence of operations needs to be corrected here:
Do not yet call `handleSuccess` in the event queue after the copy operation finishes in line 190. In line 190 we should start the delete operation.
Then, only after the delete finishes (currently line 216) we should call `handleSuccess`.

Currently, we risk deleting the volume before the copy finished, and we issue a success message too early only after the copy is finished though delete is still pending.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,255 @@
...
+      .then((operation) => {
+        eventQueue.set(
+          operation.metadata.id,
+          () => {
+            toastNotify.info(
+              <>
+                {capitalizeFirstLetter(type)} migration started for{" "}
+                <VolumeLinkChip volume={volume} />.
+              </>,
+            );
```
edlerd: Let's move this bit outside the event queue, to happen after we submitted the copy volume request (to line 187 above). 

Currently, this is visible only after the copy operation finishes.

src/pages/storage/MigrateVolumeModal.tsx
```diff
@@ -1,85 +1,148 @@
...
+      onClick={handleGoBack}
+      linkText={target ? `Choose ${type}` : "Choose migration method"}
+    />
   );
 
   return (
     <Modal
       close={close}
       className="migrate-instance-modal"
-      title={modalTitle}
```
kimanhou, line 48: Why stop using the title prop ?

```suggestion
      title={renderTitle()}
      
  const renderTitle = (): ReactNode => {
    // STEP 1: Initial method selection
    if (!type) {
      return "Choose migration method";
    }

    // STEP 3: Final confirmation (target is selected)
    if (target) {
      return (
        <BackLink
          title="Confirm migration"
          linkText={`Choose ${type}`}
          onClick={handleGoBack}
        />
      );
    }

    // STEP 2: Selecting the specific target (pool, project, or member)
    return (
      <BackLink
        title={
          <>
            Choose {type} for volume <strong>{volume.name}</strong>
          </>
        }
        linkText="Choose migration method"
        onClick={handleGoBack}
      />
    );
  };
```

You could also modify `MigrateInstanceModal` from I guess you copied the code from, for consistency.

Kxiru: Sorry I don't understand what your code suggestion is for, is there a problem with my implementation? Your suggestion seems more verbose but is completing the same thing. Are you suggesting I use comments to explain the code?

kimanhou: No, the comments are there for you to understand the code, you don't need to include them if you don't want to. 
My suggestion is about readability and keeping the JSX clean. Yes, it's completing the same thing. No problem with the logic itself!

When the logic for a prop (like `title`) involves multiple `if` statements and complex JSX, pulling it into a `renderTitle` function makes the main component body easier to scan. It also avoids having a 'mega-expression' inside the `return` statement.

src/pages/storage/MigrateVolumeModal.tsx
```diff
@@ -1,85 +1,148 @@
...
+              icon="cluster-host"
+              title="Migrate volume to a different cluster member"
+              onClick={() => {
+                setType("cluster member");
+              }}
+            />
+          )}
+          <FormLink
+            icon="switcher-dashboard"
+            title={`Choose storage pool for volume ${volume.name}`}
```
edlerd:
```suggestion
            title="Move volume to a different storage pool"
```
I think this is easier to understand and in line with instance migration.

src/pages/storage/MigrateVolumeModal.tsx
```diff
@@ -1,85 +1,148 @@
...
+          className="p-modal__close"
+          aria-label="Close active modal"
+          onClick={close}
+        >
+          Close
+        </button>
+      </header>
+      {!type && (
+        <div className="choose-migration-type">
+          {isClustered && isClusterLocalDriver(volume.pool) && (
```
edlerd: We should be passing the driver here, not the pool name.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const navigate = useNavigate();
+
+  const handleSuccess = () => {
+    let successMessage: ReactNode = "";
+
+    const memberPath = hasLocation(volume)
+      ? `/member/${encodeURIComponent(volume.location)}`
+      : "";
+    const oldMigratePoolUrl = `/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(volume.pool)}${memberPath}/volumes/${encodeURIComponent(volume.type)}/${encodeURIComponent(volume.name)}`;
+    const newMigratePoolUrl = `/ui/project/${encodeURIComponent(type === "project" ? target : volume.project)}/storage/pool/${encodeURIComponent(type === "pool" ? target : volume.pool)}${memberPath}/volumes/${encodeURIComponent(volume.type)}/${encodeURIComponent(volume.name)}`;
```
edlerd: Shall we use the `linkForVolumeDetail` helper here for constructing the two URLs?

We can construct a volume object to pass on the fly, like in `src/components/UsedByItem.tsx:61`.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,250 @@
...
+      .finally(() => {
+        close();
+      });
+  };
+
+  const handleMemberMigrate = () => {
+    //Copies volume to a new cluster member and deletes the old volume.
+    const targetMember = type === "cluster member" ? target : undefined;
+    const oldMember = volume.location;
+    copyStorageVolume(volume, volume.pool, volume.project, targetMember)
```
edlerd: The first argument is not the plain volume, but a construction from it. See `src/pages/storage/forms/CopyVolumeForm.tsx:118` for an example of the correct payload to be used here. Maybe extract that payload generation into a reusable function.

Currently this is working, but not returning an operation. So we don't delete the old volume on the previous cluster member.

src/pages/storage/MigrateVolumeModal.tsx
Kxiru: Thank you!

src/util/storageVolumeMigration.tsx
Kxiru: Sure

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,277 @@
...
+      content_type: volume.content_type,
+      source: {
+        name: volume.name,
+        type: "copy",
+        pool: volume.pool,
+        volume_only: false,
+        project: volume.project !== volume.project ? volume.project : undefined,
+        location: volume.location,
+      },
+    };
```
edlerd, line 200: Let's put building of this object into a reusable function, so we don't duplicate it from the other caller of `copyStorageVolume`.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,277 @@
...
+  const eventQueue = useEventQueue();
+  const queryClient = useQueryClient();
+  const navigate = useNavigate();
+
+  const handleSuccess = () => {
+    let successMessage: ReactNode = "";
+
+    const memberPath = hasLocation(volume)
+      ? `/member/${encodeURIComponent(volume.location)}`
+      : "";
```
edlerd:
```suggestion
```
Unused?

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,277 @@
...
+    const oldMigratePoolUrl = linkForVolumeDetail({
+      name: volume.name,
+      project: volume.project,
+      pool: volume.pool,
+      type: volume.type,
+    } as LxdStorageVolume);
```
edlerd:
```suggestion
    const oldMigratePoolUrl = linkForVolumeDetail(volume);
```
We can simply pass the volume here

src/util/storageVolumeMigration.tsx
edlerd, line 1: This is broken. When using any of the three migration options, there is no success message and the url is not updating.

Kxiru: Sorry, my dotrun issues were not allowing me to test it. Will be fixing when I'm successfully back in.

edlerd: On the demo server, the project or pool migration still fails. Also, CI is showing some errors relating to this.

Kxiru: CI issues are to do with testing I believe (but will look into them again).
Odd, all migration options worked fine for me locally. Will look into this)

edlerd: On the demo server this happens: 
image

I figured, though, this is happening with a volume that is in use by a profile. So it is not broken. But LXD backend giving an unclear error message. Maybe we should disable the migrate button if the volume is in use and give a tooltip message such as "Volume is in use".

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -156,17 +32,16 @@ const MigrateVolumeBtn: FC<Props> = ({
...
           />
         </Portal>
       )}
       <ActionButton
         onClick={openPortal}
         type="button"
         className={classNames("u-no-margin--bottom has-icon", classname)}
-        loading={isVolumeLoading}
-        disabled={!canEditVolume(volume) || isVolumeLoading}
+        disabled={!canEditVolume(volume)}
```
edlerd: When the volume has usages, any migration of it will fail. So we should disable the button and show a tooltip with a message like "Volume is in use". See also [this comment for more details on the error](https://github.com/canonical/lxd-ui/pull/1683#discussion_r2687035931).

tests/storage.spec.ts
```diff
@@ -69,7 +69,7 @@ test("storage volume create, edit and remove", async ({ page }) => {
   await assertTextVisible(page, "size2GiB");
 });
 
-test("storage volume migrate", async ({ page }) => {
+test("storage volume migrate pool", async ({ page }) => {
```
edlerd, line 72: This test is failing in CI.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,272 @@
...
+        type="pool"
+        value={target}
+        to={`/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(target)}`}
+      />
+    );
+
+    if (type === "cluster member") {
+      successMessage = (
+        <>
+          Volume <VolumeLinkChip volume={volume} /> successfully migrated to
```
edlerd: The link should go to the new volume. It currently goes to the old url of the volume before migration.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,272 @@
...
+            to={`/ui/cluster/member/${encodeURIComponent(target)}`}
+          />
+        </>
+      );
+    }
+
+    if (type === "pool") {
+      successMessage = (
+        <>
+          Volume {volumeLink} successfully migrated to pool {poolLink}
```
edlerd: The `volumeLink` should go to the new volume. It currently goes to the volume on the old pool.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,272 @@
...
+      );
+      if (oldMigratePoolUrl !== newMigratePoolUrl) {
+        navigate(newMigratePoolUrl);
+      }
+    }
+
+    if (type === "project") {
+      successMessage = (
+        <>
+          Volume <VolumeLinkChip volume={{ ...volume, project: target }} />
```
edlerd, line 90: This one is correct :)

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,272 @@
...
+              eventQueue.set(
+                operation.metadata.id,
+                () => {
+                  handleSuccess();
+                },
+                (err) => {
+                  handleFailure(err);
+                },
+                handleFinish,
+              );
```
edlerd, line 232: `deleteStorageVolume` does not return an operation. It is a sync endpoint, so inside the `then` block, we can assume it has already succeeded.

Currently the success message is not getting visible because of that.

edlerd (commented):
Good progress. I did a thorough QA and code review. Most issues are resolved. Some observations are below.

src/util/storageVolumeMigration.tsx
edlerd: Oh wait, this is more complicated than I thought. `deleteStorageVolume` does return an operation with newer LXD backends. I was testing it with an older version. Hence I was running into the error. I suspect this is fine as is. Will test further.

edlerd: this works with a recent backend as expected.

src/util/storageVolumeMigration.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const handleSuccess = () => {
+    let successMessage: ReactNode = "";
+
+    const oldMigratePoolUrl = linkForVolumeDetail(volume);
+    const newMigratePoolUrl = linkForVolumeDetail({
+      name: volume.name,
+      project: type === "project" ? target : volume.project,
+      pool: type === "pool" ? target : volume.pool,
+      type: volume.type,
+      location: volume.location,
```
edlerd: On cluster member migration, this should update to the new member.

src/util/storageVolumeMigration.tsx
Kxiru: Is there an existing flag for volumes that are in use?

edlerd: Yes, the `volume.used_by` list.

src/util/storageVolumeMigration.tsx
Kxiru: Alrighty, I'll keep this as is then.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -156,21 +32,22 @@ const MigrateVolumeBtn: FC<Props> = ({
...
-        loading={isVolumeLoading}
-        disabled={!canEditVolume(volume) || isVolumeLoading}
+        disabled={!canEditVolume(volume) || volume.used_by?.length != 0}
         title={
           canEditVolume(volume)
             ? "Migrate volume"
-            : "You do not have permission to migrate this volume"
+            : volume.used_by?.length != 0
+              ? "Volume is in use"
+              : "You do not have permission to migrate this volume"
```
edlerd: This is the wrong condition. If I can edit the volume it will never show me that the volume is in use.

src/pages/storage/forms/CopyVolumeForm.tsx
```diff
@@ -115,24 +116,7 @@ const CopyVolumeForm: FC<Props> = ({ volume, close }) => {
...
-          pool: volume.pool,
-          volume_only: !values.copySnapshots,
-          // logic from the lxc source code.
-          // We should not set source.project if target project is the same as the source project
-          project:
-            values.project !== volume.project ? volume.project : undefined,
-          location: volume.location,
-        },
-      };
+      const payload = createStorageVolumePayload(volume, volume.project);
```
edlerd: The change is neglecting the `values.project`, `values.name` and `values.copySnapshots`.

tests/storage.spec.ts
edlerd, line 72: This test is still failing in CI.

edlerd (commented):
Thanks for the changes. Some further issues below.

src/pages/storage/forms/CopyVolumeForm.tsx
```diff
@@ -115,24 +116,21 @@ const CopyVolumeForm: FC<Props> = ({ volume, close }) => {
...
+            name: values.name,
+            pool: volume.pool,
+            type: "copy",
+            volume_only: !values.copySnapshots,
+            project:
+              values.project !== volume.project ? volume.project : undefined,
+          },
         },
-      };
+        volume.project,
```
edlerd: There are some translation errors, like `values.name` should change the root name field, not the source.name field.

Also I'd suggest using spread more so that we keep this small, somethign like:

```
      const payload = createStorageVolumePayload(
        {
          ...volume,
          name: values.name,
          source: {
            ...volume.source
            volume_only: !values.copySnapshots,
            project:
              values.project !== volume.project ? volume.project : undefined,
          },
        },
      };
```
Beware, the order of the fields matter, so we override the name on the root field instead of using it from the spread.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -156,20 +33,21 @@ const MigrateVolumeBtn: FC<Props> = ({
...
         className={classNames("u-no-margin--bottom has-icon", classname)}
-        loading={isVolumeLoading}
-        disabled={!canEditVolume(volume) || isVolumeLoading}
+        disabled={!canEditVolume(volume) || isEmpty}
         title={
           canEditVolume(volume)
-            ? "Migrate volume"
+            ? isEmpty
+              ? "Volume is in use"
+              : "Migrate volume"
```
edlerd: Let's turn this into a function with early exits. Nesting the shorthand makes it fairly difficult to read.

src/pages/storage/forms/CopyVolumeForm.tsx
Kxiru: I used this approach earlier, but had errors presumably because VScode wanted me to explicitly restate some attributes such as the location. I'll push this anyway but I'm not sure how to get rid of the error without the implementation I suggested previously.
image

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -29,149 +14,46 @@ interface Props {
...
   };
 
+  const getTitle = () => {
+    if (!canEditVolume(volume)) {
+      return "You do not have permission to migrate this volume";
+    }
+
+    if (isEmpty) {
+      return "Migrate volume";
+    } else return "Volume is in use";
```
edlerd:
```suggestion
    }
    
    return "Volume is in use";
```
Slightly simpler as we don't need the `else` here.

src/pages/storage/forms/CopyVolumeForm.tsx
```diff
@@ -115,24 +116,19 @@ const CopyVolumeForm: FC<Props> = ({ volume, close }) => {
...
+          source: {
+            ...volume.source,
+            volume_only: !values.copySnapshots,
+            project:
+              values.project !== volume.project ? volume.project : undefined,
+          },
         },
-      };
+        volume.project
+      );
```
edlerd: You introduced the new `createStorageVolumePayload` function as suggested to avoid duplicating the payload object for `copyStorageVolume`. The data being generated by `createStorageVolumePayload` does not match what was here before. Please ensure this is unchanged. There are several problems causing test failures and breaking the functionality to copy a storage volume
1. The function hardocodes `volume_only: false` but we need a way to control it from here.
2. We don't have a `source` field to be set here, but we should pass the volume and need some way to control the mutable fields like `source.volume_only` and `source.project` in the returned object.

src/pages/storage/forms/CopyVolumeForm.tsx
edlerd: Continuing this discussion in https://github.com/canonical/lxd-ui/pull/1683#discussion_r2703875524

src/pages/storage/forms/CopyVolumeForm.tsx
Kxiru: I'm confused, on lines 125 and 126, you can control the source fields.

Also, all of the mutable fields do already match what was here before, with the only difference being that unchanged fields have been absorbed by the spread operator. I'm not sure what you would like me to change here.
image

edlerd: The problem is with the `createStorageVolumePayload`, which does not look at the `source.*` fields.

Kxiru: So it should be something like this?

```

export const createStorageVolumePayload = (
  volume: LxdStorageVolume,
  project: string,
): Partial<LxdStorageVolume> => {
  return {
    name: volume.name,
    type: "custom",
    config: volume.config,
    description: volume.description,
    content_type: volume.content_type,
    source: {
      ...volume.source
    },
  };
};
```

"Source" has an error in this case.
image

Should I just unpack them?

PR #1684, opened 2025-12-18
fix: [WD-32286] total cpu limit in cluster

### LINKED ISSUE #1625: "Total memory" incorrect in a cluster

omarelkashef:
## Done

- Fixes total cpu limit in cluster

Fixes #1625 

## QA

1. Run the LXD-UI:
    - Go to Create an instance page and then to Resource Limits tab
    - Create an override for CPU Limit
    - Ensure CPU limit is reported correctly as:
        (i) a value if you are in a non-cluster environment
        (ii) a value if you selected a cluster member as target
        (iii) a value if you have not selected a cluster member as target and all the cluster members have equal limits
        (iv) a range if you have not selected a cluster member as target and the cluster members do not have equal limits
    - Go to one instance -> Configuration -> Resource Limits
    - Create an override for Exposed CPU Limit
    - Ensure available memory is reported correctly as a value
    - Go to a project configuration page -> Resource Limit
    - Create an override for Exposed CPU Limit
    - Ensure CPU limit is reported correctly as the limit of the server in non-cluster environment and the sum of all nodes' limits in clustered environment

## Screenshots

image

image

### DIFF
--- a/src/components/forms/CpuLimitAvailable.tsx
+++ b/src/components/forms/CpuLimitAvailable.tsx
@@ -1,9 +1,8 @@
 import type { FC } from "react";
-import { humanFileSize } from "util/helpers";
-import { Icon } from "@canonical/react-components";
 import type { InstanceAndProfileFormikProps } from "./instanceAndProfileFormValues";
 import type { CreateInstanceFormValues } from "pages/instances/CreateInstance";
 import type { EditInstanceFormValues } from "pages/instances/EditInstance";
+import { Icon } from "@canonical/react-components";
 import { useResourceLimit } from "context/useResourceLimit";
 import { ResourceLimitIcon } from "components/ResourceLimitIcon";
 
@@ -11,33 +10,26 @@ interface Props {
   formik: InstanceAndProfileFormikProps;
 }
 
-const MemoryLimitAvailable: FC<Props> = ({ formik }) => {
+export const CpuLimitAvailable: FC<Props> = ({ formik }) => {
   const values = formik?.values as
     | CreateInstanceFormValues
     | EditInstanceFormValues;
 
-  const resourceLimit = useResourceLimit("memory", values);
+  const resourceLimit = useResourceLimit("cpu", values);
   if (!resourceLimit) {
     return null;
   }
-  const {
-    min: minMemory,
-    max: maxMemory,
-    sourceType,
-    sourceName,
-  } = resourceLimit;
-
-  const showHelpIcon = minMemory !== maxMemory;
-  const helpIconText = `The available memory depends on the cluster member that the instance will be created on.\
-  To determine the available memory exactly, go to the "Main configuration" section, and from\
-  the "Target" dropdown, select the "Cluster Member" option, then choose a member.`;
-
+  const { min: minCpu, max: maxCpu, sourceType, sourceName } = resourceLimit;
+  const showHelpIcon = minCpu !== maxCpu;
+  const helpIconText = `The total number of CPU cores depends on the cluster member that the instance will be created on.\
+      To determine the number of cores exactly, go to the "Main configuration" section, and from\
+      the "Target" dropdown, select the "Cluster Member" option, then choose a member.`;
   return (
     <>
-      Total memory:{" "}
+      Total number of CPU cores:{" "}
       <b>
-        {humanFileSize(minMemory)}
-        {minMemory !== maxMemory && ` - ${humanFileSize(maxMemory)}`}
+        {minCpu}
+        {minCpu !== maxCpu && ` - ${maxCpu}`}
         {showHelpIcon && (
           <>
             {" "}
@@ -50,10 +42,9 @@ const MemoryLimitAvailable: FC<Props> = ({ formik }) => {
         )}
       </b>
       <br />
-      {sourceName && sourceType && (
+      {sourceType && sourceName && (
         <ResourceLimitIcon source={sourceName} sourceType={sourceType} />
       )}
     </>
   );
 };
-export default MemoryLimitAvailable;
--- a/src/context/useResourceLimit.tsx
+++ b/src/context/useResourceLimit.tsx
@@ -1,0 +1,57 @@
+import { useNotify } from "@canonical/react-components";
+import { useResources } from "context/useResources";
+import type { CreateInstanceFormValues } from "pages/instances/CreateInstance";
+import type { EditInstanceFormValues } from "pages/instances/EditInstance";
+import { CLUSTER_GROUP_PREFIX } from "util/instances";
+import { getResourceLimit } from "util/resourceLimits";
+import { useIsClustered } from "./useIsClustered";
+import { useCurrentProject } from "./useCurrentProject";
+import { limitToBytes } from "util/limits";
+import type { LxdResources } from "types/resources";
+
+export const ensureArray = (resources: LxdResources | LxdResources[]) =>
+  Array.isArray(resources) ? resources : [resources];
+
+export const useResourceLimit = (
+  mode: "cpu" | "memory",
+  formValues: CreateInstanceFormValues | EditInstanceFormValues,
+) => {
+  const notify = useNotify();
+  const isClustered = useIsClustered();
+  const { target } = formValues as CreateInstanceFormValues;
+  const { location } = formValues as EditInstanceFormValues;
+  const isClusterGroupTarget = target?.startsWith(CLUSTER_GROUP_PREFIX);
+  const validTarget = target && !isClusterGroupTarget ? target : undefined;
+  const clusterMember = isClustered ? (validTarget ?? location) : undefined;
+  const { data: resources, error, isLoading } = useResources(clusterMember);
+  const { project } = useCurrentProject();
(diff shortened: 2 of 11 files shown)


src/util/resourceLimits.tsx
```diff
@@ -0,0 +1,74 @@
...
+export type ResourceLimitSource = "project" | "cluster-member";
+export interface ResourceLimit {
+  min: number;
+  max: number;
+  source?: ResourceLimitSource;
+}
+
+export const getResourceLimit = (
+  resources: number[],
+  isClusterMemberDefined: boolean,
```
edlerd:
```suggestion
  clusterMember: string | undefined,
```
I think passing in the member and checking on it below is bit simpler.

src/components/forms/CpuLimitAvailable.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const resourceArray = Array.isArray(resources) ? resources : [resources];
+  const {
+    min: minCpu,
+    max: maxCpu,
+    source,
+  } = getResourceLimit(
+    resourceArray.map((r) => r.cpu.total),
+    !!clusterMember,
+    projectResourceCpu,
+  );
```
edlerd: This is duplicating from `MemoryLimitAvailable`. Maybe instead of extracting the `getResourceLimit` we can define a hook like below. It can fetch via `useResources` internally, do all the calculations from `getResourceLimit` and the ones above. Instead of handling loading and error state, we return null if the max returned is not defined. Wdyt?
```
const {
    min: minCpu,
    max: maxCpu,
    source,
  } = useResourceLimit(projectResourceCpu)
```

edlerd (commented):
Works nicely, two ideas to improve a bit on the code structure.

src/components/forms/CpuLimitAvailable.tsx
omarelkashef: We will still need to handle the loading state in the component to display a spinner and handle the null returned in the case of error, so we will end up with roughly the same amount of duplication. Is it worth it?

edlerd: I was thinking to remove both. Render null if loading or on error case. The available resources will not appear in those cases. Wdyt?

omarelkashef: Ok, I updated the components to use useResourceLimit hook,

omarelkashef: I will refactor some more code.

src/util/resourceLimits.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  return {
+    min: minResource,
+    max: projectLimit,
+    sourceType: undefined,
+    sourceName: undefined,
+  };
+};
+
+export const getResourceLimitIcon = (
```
edlerd: Please transform this into a regular react component in a dedicated file.

To be called as `<ResourceLimitIcon source={... />` instead of a plain JS function.

src/pages/projects/ProjectCpuLimitInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+type Props = InputProps;
+
+export const ProjectCpuLimitInput: FC<Props> = ({ help, ...props }) => {
+  const notify = useNotify();
+
+  const { data: resources = [], error, isLoading } = useResources();
+  let sumCpuLimit = null;
+  if (isLoading) {
+    sumCpuLimit = <Spinner className="u-loader" text="Loading resources..." />;
```
edlerd: This is unexpected. The variable with name `sumCpuLimit` should not hold a spinner. We could just set it to null here as well in my opinion.

src/pages/projects/ProjectCpuLimitInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <>
+      Total number of CPU cores: <b>{sumCpuLimit}</b>
+    </>
+  ) : (
+    ""
+  );
+
+  const helpText = (
+    <>
+      {help} {help && sumCpuLimitText && <br />}
```
edlerd:
```suggestion
      {help}
      {help && sumCpuLimitText && <br />}
```
Slightly clearer formatting.

src/pages/projects/ProjectCpuLimitInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+  let sumCpuLimit = null;
+  if (isLoading) {
+    sumCpuLimit = <Spinner className="u-loader" text="Loading resources..." />;
+  }
+  if (error) {
+    notify.failure("Loading resources failed", error);
+    sumCpuLimit = null;
+  }
+  if (resources) {
+    const resourceArray = Array.isArray(resources) ? resources : [resources];
```
edlerd: How about a new `ensureArray = (foo) => Array.isArray(foo) ? foo : [foo]` helper to be reused here and in other places?

src/context/useResourceLimit.tsx
```diff
@@ -0,0 +1,41 @@
...
+import type { CreateInstanceFormValues } from "pages/instances/CreateInstance";
+import type { EditInstanceFormValues } from "pages/instances/EditInstance";
+import type { LxdResources } from "types/resources";
+import { CLUSTER_GROUP_PREFIX } from "util/instances";
+import { getResourceLimit } from "util/resourceLimits";
+import { useIsClustered } from "./useIsClustered";
+import type { LxdProject } from "types/project";
+
+export const useResourceLimit = (
+  resourceExtractor: (resources: LxdResources) => number,
```
edlerd:
```suggestion
  mode: "cpu" | "memory",
```
This extractor pattern makes it fairly hard to reason about what the code is doing as we jump in and out of the local function. Maybe better introduce a more simple mode field, that we handle locally in this function.

src/context/useResourceLimit.tsx
```diff
@@ -0,0 +1,41 @@
...
+  formValues: CreateInstanceFormValues | EditInstanceFormValues,
+  project?: LxdProject,
+  projectResourceLimit?: number,
```
edlerd:
```suggestion
```
I think with the `mode` in place, we can move this into this function as well.

src/context/useResourceLimit.tsx
```diff
@@ -0,0 +1,41 @@
...
+import type { LxdResources } from "types/resources";
+import { CLUSTER_GROUP_PREFIX } from "util/instances";
+import { getResourceLimit } from "util/resourceLimits";
+import { useIsClustered } from "./useIsClustered";
+import type { LxdProject } from "types/project";
+
+export const useResourceLimit = (
+  resourceExtractor: (resources: LxdResources) => number,
+  formValues: CreateInstanceFormValues | EditInstanceFormValues,
+  project?: LxdProject,
```
edlerd:
```suggestion
```
If you follow this project variable upwards, we can pull this into the function as `  const { project } = useCurrentProject();`. And then remove the prop drilling through many of the layers as no other usages of `project` remain upstream.

edlerd (commented):
Good generalization with the new `useResourceLimit` function! Some ideas simplify the structure below. The goal is to increase maintainability and increase readability.

src/context/useResourceLimit.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const clusterMember = isClustered ? (validTarget ?? location) : undefined;
+  const { data: resources, error, isLoading } = useResources(clusterMember);
+  const { project } = useCurrentProject();
+  const resourceLimitName = mode === "memory" ? "limits.memory" : "limits.cpu";
+  const projectResourceLimit = project?.config[resourceLimitName]
+    ? limitToBytes(project?.config[resourceLimitName])
+    : undefined;
+
+  if (error) {
+    notify.failure("Loading resources failed", error);
```
edlerd, line 34:
```suggestion
    notify.failure("Loading resources failed", error);
    return null;
```
Shall we return here, not to present a wrongly calculate limit based on project configuration?

src/pages/projects/ProjectCpuLimitInput.tsx
```diff
@@ -0,0 +1,42 @@
...
+      0,
+    );
+  }
+  const sumCpuLimitText = sumCpuLimit ? (
+    <>
+      Total number of CPU cores: <b>{sumCpuLimit}</b>
+    </>
+  ) : (
+    ""
+  );
```
edlerd:
```suggestion
  const limitSum = ensureArray(resources).reduce(
    (sum, resource) => sum + resource.cpu.total,
    0,
  );
  const hasLimits = limitSum > 0;
  const limitText = hasLimits ? (
    <>
      Total number of CPU cores: <b>{limitSum}</b>
    </>
  ) : null;
```
A bit simpler?

edlerd (approved):
QA LGTM, just two nitpicks on code.

src/pages/projects/ProjectCpuLimitInput.tsx
omarelkashef: It is indeed simpler.

PR #1692, opened 2026-01-05
feat(profile): Create rich tooltip for profile

kimanhou:
## Done

- Create rich tooltip for profiles
- Use it
- Harmonize side panel and profile overview
- Harmonize profile side panel and instance side panel

## QA

1. Run the LXD-UI:
    - Hover on a profile chip and make sure the rich tooltip is displayed correctly
    - Profile chips can be found in:
        {instance_name} > Configuration > Cloud init
        {instance_name} > Configuration > Devices > Network
        {instance_name} > Configuration > Main configuration > SSH keys
        Instance list > click on the row of an instance of open the instance details side panel > Profiles
        {instance_name} > Overview > Profiles
        Success toast notification after a profile is created
        Success toast notification after a profile is updated
        Success toast notification after the name of a profile is updated from the profile details page's header
        Modal when a profile is about to be deleted
        Configuration (of a project) > Project details > Default instance storage pool's help text
        Configuration (of a project) > Project details > Default instance network help text

## Screenshots
Rich tooltip
image

Profile side panel
profile-side-panel

Profile overview (I only changed the name of the sections "Limits" to "Resource limits" and "Usage" to "Used by")
image

Instance side panel
image


src/pages/profiles/ProfileRichChip.tsx
```diff
@@ -0,0 +1,54 @@
...
+  resourceLinkDisabled?: boolean;
+}
+
+const ProfileRichChip: FC<Props> = ({
+  profileName,
+  projectName,
+  resourceLinkClassName,
+  resourceLinkDisabled,
+}) => {
+  const showTooltip = !useIsScreenBelow(SMALL_TOOLTIP_BREAKPOINT, "height");
```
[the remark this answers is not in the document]
kimanhou: Will keep it as it is for consistency with `InstanceRichChip`

omarelkashef (commented):
The height of the tooltip is not correct. The min-height should be overriden for the profile tooltips (maybe 200px) because it only has 3 rows.

image

omarelkashef:
There are also rows for Devices and Used by in the Profile side panel. Would they be helpful in the tooltips while avoiding nested tooltips?

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { data: profile, isLoading: isProfileLoading } = useProfile(
+    profileName,
+    projectName,
+  );
+  const { project, isLoading: isProjectLoading } = useCurrentProject();
+  const featuresProfiles = isProjectWithProfiles(project);
+
+  const isLoading = isProfileLoading || isProjectLoading;
+
+  if (!profile && !isLoading) {
```
omarelkashef, line 40: How about displaying a message that this profile could not be found and a similar text in other tooltips instead of returning empty tags as it does not look the nicest.

image

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,53 @@
...
+      valueTitle: profileDescription,
+    },
+    {
+      title: "Defined in",
+      value: `${featuresProfiles ? "Current" : "Default"} project`,
+    },
+  ];
+
+  return (
+    <RichTooltipTable rows={rows} className="profile-rich-tooltip-table" />
```
omarelkashef, line 130: I do not think this class is being used currently. You can override the min-height with it.

src/sass/rich_tooltip.scss
```diff
@@ -15,14 +15,14 @@ body.is-dark .rich-tooltip-table .p-icon--status-queued-small {
...
 
-  @media (height <= $medium-tooltip-breakpoint) {
-    min-height: 200px;
-  }
-
   @media (height <= $large-tooltip-breakpoint) {
     min-height: 400px;
   }
 
+  @media (height <= $medium-tooltip-breakpoint) {
```
omarelkashef, line 51: Thank you for this fix!

omarelkashef (commented):
Great work! Just a few comments

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: It is now, I just pushed some changes.

kimanhou:
The rich tooltip has now the following info:
Networks, Devices (a summary), Resource limits, Storage, Cloud init, Used by (count)

Compared to the rich tooltip, the side panel is missing the following info:
Devices, Resource limits, Cloud init. 
Should I add them ?

Compared to the rich tooltip, the profile overview is missing the following info:
Storage.
Should I add it ?

There is also a discrepancy in the naming:
- in overview "Limits" vs in tooltip "Resource limits". Should I harmonize this and if so which one to use ?
- in overview "Usage" vs in side panel "Used by". Should I harmonize this and if so which one to use ?
- in overview, there is a section for Networks and a section for Devices (+ a section for Storage if I add it) vs in side panel there is a section for Devices and Networks are part of this section. How should I harmonize this ?

edlerd:
> There is a discrepancy in the naming:
> 
> * in overview "Limits" vs in tooltip "Resource limits". I chose "Resource limits".

Do you think "Limits" is confusing? I slightly prefer it as it is more concise.

> * in overview "Usage" vs in side panel "Used by". I chose "Used by", as "Usage" can be confused with how much memory has been used in a storage.

Agreed, good call!

> * in overview, there is a section for Networks and a section for Devices vs in side panel there is a section for Devices and Networks are part of this section. In the side panel, I put a subsection called "Devices" inside "Devices", which is not ideal. Any idea ?

Split up one section "Networks" and below another section "Devices". Yes, networks are also devices, but I think this is obvious to understand. And splitting is in sync with the instance overview. Generally speaking, I'd try to reuse the structure between profiles and instances where possible as this make the interface learnable and gives users a nice "oh I know how this works" moment.

src/components/RichTooltipRow.tsx
```diff
@@ -1,21 +1,30 @@
 import type { FC } from "react";
+import classnames from "classnames";
+
 export interface TooltipRow {
   title: string;
   value: React.ReactNode;
   valueTitle?: string;
   className?: string;
+  noTruncate?: boolean;
```
edlerd:
```suggestion
  truncate?: boolean;
```
Better keep it positive, it is easier to read. With a default value you can keep the current behaviour without breaking existing usages.

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,126 @@
...
+      ),
+      noTruncate: true,
+    },
+  ];
+
+  if (isAboveMedium) {
+    rows.push(
+      {
+        title: "Storage",
+        value: <ProfileStorageList profile={profile} project={projectName} />,
```
edlerd: The chips have a different shade between storage and network.

image

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,126 @@
...
+        noTruncate: true,
+        className: "storage-list",
+      },
+      {
+        title: "Resource limits",
+        value:
+          `CPU: ${profile.config["limits.cpu"] || "-"}, ` +
+          `Memory: ${profile.config["limits.memory"] || "-"}`,
```
edlerd: Maybe use an inline or [middot list here](https://canonical.github.io/react-components/?path=/story/components-list--middot)?

src/sass/_profile_rich_tooltip.scss
edlerd, line 1: The tooltip opens to the top for me on first try (while the profile is loading). If I close the tooltip and open it again, then it opens correctly to the bottom. I think there might be a wrong or no min-height applied, maybe?

image

src/sass/_profile_rich_tooltip.scss
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  .p-icon--exposed {
+    @include vf-icon-exposed($colors--light-theme--icon);
+  }
+
+  > span.p-chip__value {
+    color: $colors--light-theme--text-default;
+  }
+}
```
edlerd: This should not be necessary, but the theme should automatically flip in the tooltip. We had a similar issue first with the start/stop/etc. actions in the instance tooltip. Maybe @omarelkashef can give a hint how we fixed it there to work in both light and dark themes.

omarelkashef: Theme inversion worked out of the box. I did not need to change anything except for some icons that depended on the body theme. You can look at rich_tooltip.scss for how it can be done.

kimanhou: Is this how the chip is supposed to look on a light background ?
image

edlerd: I think there is something wrong with the chip backgrounds. This is how it looks on the instance list with a light theme:

image

kimanhou: I have been investigating since yesterday and the only conclusion I can reach is that it's the same background color.

kimanhou: Aaaaaaaaaah the background of the tooltip is darker than on the main theme !

kimanhou: I added some CSS in rich_tooltip to make all chips inside rich tooltips have a darker background
Screenshot from 2026-01-08 11-36-14
image

src/pages/instances/InstanceOverview.tsx
```diff
@@ -125,7 +125,7 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
       </Row>
       <Row className="usage">
         <Col size={3}>
-          <h2 className="p-heading--5">Usage</h2>
+          <h2 className="p-heading--5">Used by</h2>
```
edlerd: This should be reverted. Maybe use an entirely different and more specific term if you find one here. It relates to memory/CPU/disk usage of an instance

src/pages/profiles/ProfileDetailOverview.tsx
```diff
@@ -111,7 +107,7 @@ const ProfileDetailOverview: FC<Props> = ({ profile }) => {
       </Row>
       <Row className="section">
         <Col size={3}>
-          <h2 className="p-heading--5">Limits</h2>
+          <h2 className="p-heading--5">Resource limits</h2>
```
edlerd, line 110: Good idea to align this with the configuration screen.

src/pages/instances/InstanceOverview.tsx
edlerd: Maybe "Resource usage"?

src/pages/profiles/ProfileDetailPanelContent.tsx
```diff
@@ -52,11 +54,37 @@ const ProfileDetailPanelContent: FC<Props> = ({ profile, project }) => {
...
         <tr className="u-no-border list-wrapper">
-          <th className="u-text--muted last-of-section">Storage</th>
+          <th className="u-text--muted">Devices</th>
           <td>
+            <ProfileDevicesSummaryList
+              profile={profile}
+              project={project.name}
+            />
+          </td>
+        </tr>
```
edlerd, line 73: This looks a bit odd here in the panel with the heading repeated below. Maybe remove the heading?

Also, we tend to use "disk" instead of "storage" in other places where referencing storage devices for profiles or instances.

Also it would be nice to have more alignment between the instance side panel/tooltip content and the profile ones.

image

src/components/TruncatedList.tsx
```diff
@@ -0,0 +1,25 @@
+import type { FC, ReactNode } from "react";
+
+interface Props {
+  items: ReactNode[];
+  numberToShow?: number;
+}
+
+const TruncatedList: FC<Props> = ({ items, numberToShow = 2 }) => {
```
edlerd, line 8: Good idea to extract this as a reusable component. We might upstream this later on.

src/sass/_profile_rich_tooltip.scss
edlerd: The contrast for the chip background for the light background appears to be too low. For the dark background, the contrast seems too high.

src/pages/instances/InstanceOverview.tsx
kimanhou: Oh I got the wrong thing. `Usage` is fine

src/pages/profiles/ProfileDetailPanelContent.tsx
kimanhou: Instance side panel is indeed very different:
    - it has a dedicated section for `Networks`
    - it has a row for `Root storage pool`
    - it doesn't have anything for devices (other than nic devices)

I tried harmonizing the side panels of instance and profile. Wdyt ?

Profile side panel
image

    
Instance side panel
image

edlerd: Looks better aligned, thanks. I'd keep the "Root storage pool" label also for profiles.

edlerd: Can we omit the "1 root disk" from the devices list? As it appears right below anyhow.

src/pages/profiles/ProfileDetailPanelContent.tsx
```diff
@@ -40,23 +45,49 @@ const ProfileDetailPanelContent: FC<Props> = ({ profile, project }) => {
...
+          <td>
+            <DevicesSummaryList
+              devices={Object.values(profile.devices).filter(
+                (device) => !isNicDevice(device),
+              )}
+            />
+          </td>
+        </tr>
+        <tr>
+          <th className="u-text--muted">Default storage pool</th>
```
edlerd:
```suggestion
          <th className="u-text--muted">Root storage pool</th>
```
Better align it with instance panel?

src/pages/profiles/ProfileDetailPanelContent.tsx
```diff
@@ -40,23 +45,49 @@ const ProfileDetailPanelContent: FC<Props> = ({ profile, project }) => {
...
           <td>
-            <ProfileStorageList profile={profile} project={project.name} />
+            <ResourceLink
+              type="pool"
+              value={defaultStoragePool}
+              to={`/ui/project/${encodeURIComponent(project.name)}/storage/pool/${encodeURIComponent(defaultStoragePool)}`}
+            />
```
edlerd: The storage pool can be empty:

image

src/pages/profiles/ProfileRichChip.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+interface Props {
+  profileName: string;
+  projectName: string;
+  resourceLinkClassName?: string;
+  resourceLinkDisabled?: boolean;
+  positionElementClassName?: string;
+}
+
+const ProfileRichChip: FC<Props> = ({
```
edlerd, line 15: There are some new usages in the instance/configuration/disk and maybe other sections that can use the profile chip with tooltip.

src/pages/profiles/ProfileRichChip.tsx
```diff
@@ -0,0 +1,57 @@
...
+import ResourceLink from "components/ResourceLink";
+import ProfileRichTooltip from "pages/profiles/ProfileRichTooltip";
+import { SMALL_TOOLTIP_BREAKPOINT } from "components/RichTooltipTable";
+
+interface Props {
+  profileName: string;
+  projectName: string;
+  resourceLinkClassName?: string;
+  resourceLinkDisabled?: boolean;
+  positionElementClassName?: string;
```
edlerd:
```suggestion
  className?: string;
  disabled?: boolean;
```
Slightly more concise naming.

Can we do without the `positionElementClassName`?

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,137 @@
...
+    },
+  ];
+
+  if (isAboveMedium) {
+    rows.push(
+      {
+        title: "Storage",
+        value: profile ? (
+          <ProfileDiskDeviceList profile={profile} project={projectName} />
+        ) : (
```
edlerd:
```suggestion
        title: "Root storage",
        value: profile ? (
          <ProfileDiskDeviceList profile={profile} project={projectName} />
        ) : (
```
I think we should only show the root pool here, and any volumes will appear under the devices section. That way we can ensure this line does not overflow a single line.

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,137 @@
...
+      ) : (
+        "-"
+      ),
+      truncate: false,
+      className: "networks-list",
+    },
+    {
+      title: "Devices",
+      value: profile ? (
+        <DevicesSummaryList devices={Object.values(profile.devices)} />
```
edlerd: Let's ignore the root storage in this list, not to show the info more than once:

image

src/pages/profiles/ProfileDetailPanelContent.tsx
edlerd: Added [another comment about this](https://github.com/canonical/lxd-ui/pull/1692#discussion_r2676319031), resolving this one.

src/pages/profiles/ProfileRichChip.tsx
kimanhou: Not really, they are not on the same element

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: I called it `Root storage pool` tho, to be consistent with the side panels. 
image

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: I'm a bit lost in your comments so let's centralize the discussion here.

I can omit "1 root disk" in `DevicesSummaryList`. However in the tooltip, the row "Root storage pool" only appears on medium screens and bigger, so on small screens there will only be Networks and Devices. Is it misleading ?

edlerd: You are raising a good point. Let's switch the order in the tooltip and side panels to be:

storage
networks
devices

Wdyt?

src/pages/profiles/ProfileRichChip.tsx
edlerd: Sorry, my question was a bit vague. Why do we need the `positionElementClassName`? I see it is used to set a 100% width on the tooltip. But why is that even necessary? I haven't seen this in the other tooltips, like the one for instances.

kimanhou: Ah it's to nicely truncate the chip in Inherited network devices

Without it
image


With it
image

src/pages/profiles/ProfileRichChip.tsx
kimanhou: Good point, I added it in `getInheritedSourceRow` which is used for disk, root disk, gpu, proxy and other devices.

src/sass/_profile_rich_tooltip.scss
kimanhou: I cheated by using a brightness filter in CSS on hover. It's not ideal because the whole chip is affected, not just the background. Otherwise I would have to hardcode a new color. Any idea ?

src/pages/profiles/ProfileRichChip.tsx
edlerd: I think we can do this instead of the `profile-rich-tooltip-position-element`. This should be generic, so don't need the fix on all other chips as well:
```
  .inherited {
    [...]

    span:has(.p-chip) {
      max-width: 100%;
    }
  }
```

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,143 @@
...
+      title: "Root storage pool",
+      value: profile ? (
+        <ResourceLink
+          type="pool"
+          value={defaultStoragePool}
+          to={`/ui/project/${encodeURIComponent(projectName)}/storage/pool/${encodeURIComponent(defaultStoragePool)}`}
+        />
+      ) : (
+        "-"
+      ),
```
edlerd:
```suggestion
      title: "Root storage",
      value: defaultStoragePool ? (
        <ResourceLink
          type="pool"
          value={defaultStoragePool}
          to={`/ui/project/${encodeURIComponent(projectName)}/storage/pool/${encodeURIComponent(defaultStoragePool)}`}
        />
      ) : (
        "-"
      ),
```
The other title is slightly too long.

Also, I think we check for the wrong thing here when the pool is empty.

image

src/sass/_profile_rich_tooltip.scss
edlerd: I think right now it is okay, LGTM.

src/pages/profiles/ProfileRichChip.tsx
kimanhou: Oh yeah that works well, thank you !

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,98 @@
...
+  if (devicesOtherThanNic.length === 0) {
+    return "-";
+  }
+
+  const deviceCounts: Record<string, number> = {};
+
+  // Find the primary root disk (same logic as getDefaultStoragePool/getRootPool)
+  const primaryRootDisk = devices
+    .filter((device) => !isNicDevice(device) && isDiskDevice(device))
+    .find((device) => isRootDisk(device as FormDevice));
```
edlerd:
```suggestion
    .find((device) => isRootDisk(device as FormDevice));
```
Isn't this simpler and the same?

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  devicesOtherThanNic.forEach((device) => {
+    if (isDiskDevice(device)) {
+      if (isRootDisk(device as unknown as FormDevice)) {
+        if (device === primaryRootDisk) {
+          return;
+        } else {
+          deviceCounts.rootDisk = (deviceCounts.rootDisk || 0) + 1;
+        }
```
edlerd:
```suggestion
      if (isRootDisk(device as unknown as FormDevice)) {
        return;
```
There is only one root disk that we filter out. I think tis simplifies.

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,98 @@
...
+    "unix-hotplug",
+    "infiniband",
+  ];
+
+  const deviceSummary = deviceTypeOrder
+    .filter((type) => deviceCounts[type] > 0)
+    .map((type) => {
+      const count = deviceCounts[type];
+      const label = deviceTypeLabels[type] || type;
+      return `${count} ${label}${count > 1 ? "s" : ""}`;
```
edlerd: better use the `pluralize` helper. It seems we might have some special cases like proxy > proxies though.

src/pages/profiles/ProfileRichChip.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import ResourceLink from "components/ResourceLink";
+import ProfileRichTooltip from "pages/profiles/ProfileRichTooltip";
+import { SMALL_TOOLTIP_BREAKPOINT } from "components/RichTooltipTable";
+
+interface Props {
+  profileName: string;
+  projectName: string;
+  resourceLinkClassName?: string;
+  resourceLinkDisabled?: boolean;
```
edlerd:
```suggestion
  className?: string;
  disabled?: boolean;
```
Can we simplify those names?

src/pages/profiles/ProfileDetailPanelContent.tsx
```diff
@@ -40,23 +44,49 @@ const ProfileDetailPanelContent: FC<Props> = ({ profile, project }) => {
           <th className="u-text--muted last-of-section">Defined in</th>
           <td>{featuresProfiles ? "Current" : "Default"} project</td>
         </tr>
-        <tr className="u-no-border">
-          <th colSpan={2}>
-            <h3 className="p-muted-heading p-heading--5">Devices</h3>
-          </th>
+        <tr>
+          <th className="u-text--muted">Root storage pool</th>
```
edlerd:
```suggestion
          <th className="u-text--muted">Root storage</th>
```
Shall we shorten it here as well?

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,98 @@
...
+    disk: "disk",
+    gpu: "GPU",
+    proxy: "proxy",
+    usb: "USB",
+    pci: "PCI",
+    tpm: "TPM",
+    "unix-block": "unix-block",
+    "unix-char": "unix-char",
+    "unix-hotplug": "unix-hotplug",
+    infiniband: "infiniband",
```
edlerd: I think they should all start with a capital letter. Wdyt?

src/components/DevicesSummaryList.tsx
edlerd: Actually, we might not need this `primaryRootDisk` at all.

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,143 @@
...
+      title: "Description",
+      value: profileDescription,
+      valueTitle: profileDescription,
+    },
+    {
+      title: "Defined in",
+      value: `${featuresProfiles ? "Current" : "Default"} project`,
+    },
+    {
+      title: "Root storage",
```
edlerd, line 73: Somehow the network chips seem to be aligned, but the storage pool chip does not:

image

src/pages/profiles/ProfileDetailPanelContent.tsx
kimanhou: Yes, but also in `InstanceDetailPanelContent`.

src/util/instanceBulkActions.tsx
```diff
@@ -69,17 +70,20 @@ export const instanceActionLabel = (action: LxdInstanceAction): string => {
...
-    return item.toLowerCase().replace("identity", "identities");
-  }
+  const exceptions: Record<string, string> = {
+    identity: "identities",
+    proxy: "proxies",
+  };
+
+  const pluralForm = exceptions[itemLower] || `${itemLower}s`;
+
+  const isCapitalized = item.charAt(0) === item.charAt(0).toUpperCase();
```
edlerd:
```suggestion
  const isCapitalized = item.charAt(0) !== itemLower.charAt(0);
```
Is this simpler?

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,90 @@
...
+    } else if (isProxyDevice(device)) {
+      deviceCounts.proxy = (deviceCounts.proxy || 0) + 1;
+    } else if (isOtherDevice(device)) {
+      deviceCounts[device.type] = (deviceCounts[device.type] || 0) + 1;
+    }
+  });
+
+  // Convert counts to readable strings
+  const deviceTypeLabels: Record<string, string> = {
+    rootDisk: "Root disk",
```
edlerd:
```suggestion
```
We removed it :)

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,90 @@
...
+    disk: "Disk",
+    gpu: "GPU",
+    proxy: "Proxy",
+    usb: "USB",
+    pci: "PCI",
+    tpm: "TPM",
+    "unix-block": "unix-block",
+    "unix-char": "unix-char",
+    "unix-hotplug": "unix-hotplug",
+    infiniband: "infiniband",
```
edlerd:
```suggestion
    other: "Other",
```
I wonder if we should combine those to avoid too much overflow in the lists when we have a large variety. Wdyt?

If we do, we should simplify above and the `deviceTypeOrder` below as well.

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,143 @@
...
+        title: "Limits",
+        value: (
+          <List
+            items={[
+              `CPU: ${profile?.config["limits.cpu"] || "-"}`,
+              `Memory: ${profile?.config["limits.memory"] || "-"}`,
+            ]}
+            middot
+            className="u-no-margin"
+          />
```
edlerd: The empty state for those looks still a bit weird and difficult to read. Maybe split them out as dedicated "CPU limit" and "Memory limit" rows?

image

src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -135,51 +166,15 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
...
-            <td colSpan={2}>
-              <h3 className="p-muted-heading p-heading--5">Networks</h3>
-              <p>
-                No networks found.
-                <br />
-                <Link
-                  to={`/ui/project/${encodeURIComponent(instance.project)}/instance/${encodeURIComponent(instance.name)}/configuration/networks`}
-                >
-                  Configure instance networks
-                </Link>
```
edlerd, line 178: Can we keep this link when there is no network attached?

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,90 @@
...
+  ];
+
+  const deviceSummary = deviceTypeOrder
+    .filter((type) => deviceCounts[type] > 0)
+    .map((type) => {
+      const count = deviceCounts[type];
+      const label = deviceTypeLabels[type] || type;
+      return `${count} ${pluralize(label, count)}`;
+    })
+    .join(", ");
```
edlerd: Let's use an inline list here as well instead of comma-separated values.

src/sass/rich_tooltip.scss
```diff
@@ -10,25 +10,60 @@ body.is-dark .rich-tooltip-table .p-icon--status-queued-small {
...
+    .resource-link.p-chip {
+      background-color: $colors--dark-theme--background-active;
+
+      &:hover {
+        filter: brightness(0.9);
+      }
+
+      > span.p-chip__value {
+        color: $colors--dark-theme--text-default;
+      }
```
edlerd, line 39:
```suggestion
```
This seems wrong. Why do we change the colors under a specific height breakpoint?

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,143 @@
...
+      },
+    );
+  }
+
+  if (isAboveLarge) {
+    rows.push(
+      {
+        title: "Has cloud init",
+        value: profile ? (hasCloudInit(profile) ? "Yes" : "No") : "-",
+      },
```
edlerd: Shall we make this more generic? It could be a list of the below sections, similar to the devices list. With counts per section. We'd need a good name for the row though.

Resource limits
Security policies
Snapshots
Migration
Boot
Cloud init

edlerd (commented):
Thanks for the iteration. This works nicely. Some ideas to potentially polish it even more are below.

src/util/instanceBulkActions.tsx
kimanhou: Bof, I prefer the positive equals

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: I really like having CPU and Memory on the same row, it makes sense and it saves vertical space.

For the empty state, I suggest having a "No limits" string:
image

If CPU limit is set but not Memory, this is how it looks:
image

Wdyt ?

src/pages/instances/InstanceDetailPanelContent.tsx
kimanhou: Just the link or the string "No networks found." too ?

kimanhou: Just the link
image

Link + text
image

src/pages/profiles/ProfileRichTooltip.tsx
edlerd: Seems ok; maybe simply render `-` if no limits apply. Similar to other rows.

src/pages/instances/InstanceDetailPanelContent.tsx
edlerd: I think both. It seems a misconfiguration and good to guide the user how to fix it.

src/components/DevicesSummaryList.tsx
kimanhou: With a middot too right ?
image

edlerd: Yes, that's what I think looks best and consistent.

src/sass/rich_tooltip.scss
kimanhou: ??
There is no specific height breakpoint applied on this. This is directly inside `.rich-tooltip-row-content`, which is inside `.rich-tooltip-table`

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: `Configuration` ?

I am not sure how would the count be calculated for each of the rows.
Otherwise I can list what configuration has been set:
image

src/components/DevicesSummaryList.tsx
kimanhou: Yes good idea

src/sass/rich_tooltip.scss
edlerd: Oh I see, then all good :+1: 

I tend to keep the media query based styles at the end and misread the diff.

src/pages/profiles/ProfileRichTooltip.tsx
edlerd: I think you can group the profiles' `configuration` field keys by prefix. Like the example below, would have an entry for "Security policies" and "Cloud init". But maybe just listing without actually providing the counts is better.

image

edlerd: I agree, "Configuration" seems a good label.

src/sass/rich_tooltip.scss
kimanhou: I moved the media query to the end of the block to avoid confusion.

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: Okay so we would remove the `Limits` row (CPU and Memory) completely ?
And replace it with the `Configuration` row, right ?

I'm also thinking of `Configured sections` as the label of this row ?

src/sass/rich_tooltip.scss
edlerd: Thanks.

src/pages/profiles/ProfileRichTooltip.tsx
edlerd: > Okay so we would remove the `Limits` row (CPU and Memory) completely ? And replace it with the `Configuration` row, right ?

We can also continue to highlight those in a dedicated row as they seem important. No strong opinion from my side.

> I'm also thinking of `Configured sections` as the label of this row ?

Yeah, a bit more precise and seems easier to understand.

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,81 @@
...
+    volume: "Volume",
+    disk: "Disk",
+    gpu: "GPU",
+    proxy: "Proxy",
+    other: "Other",
+  };
+
+  // Define order to match menu: root disk, volumes, other disks, GPU, proxy, then others
+  const deviceTypeOrder = [
+    "rootDisk",
```
edlerd:
```suggestion
```

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,169 @@
...
+      hasSecurityPolicies && "Security policies",
+      hasSnapshots && "Snapshots",
+      hasMigration && "Migration",
+      hasBoot && "Boot",
+      hasCloudInitConfig && "Cloud init",
+    ].filter(Boolean);
+
+    const configSummary =
+      configSections.length > 0 ? (
+        <List items={configSections} middot className="u-no-margin" />
```
edlerd: Can we put this into a component, such as `ProfileConfigurationSections`? We might also add it to the profile overview and side panel.

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,169 @@
...
+      ) : (
+        "-"
+      );
+    rows.push(
+      {
+        title: "Devices",
+        value: profile ? (
+          <DevicesSummaryList devices={Object.values(profile.devices)} />
```
edlerd, line 103: This and the config sections need to handle the overflow with long lists

image

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: > We can also continue to highlight those in a dedicated row as they seem important. No strong opinion from my side.

I don't want redundant information. Either we keep a dedicated row for resource limits and remove it from the `Configuration` row, either we remove the resource limits row and include this info in the `Configuration` row.

1.
image

OR 2.
image


I don't like ⬇️
image



> I'm also thinking of Configured sections as the label of this row ?

Actually `Configured sections` is too long, it doesn't fit. I'll keep `Configuration`.

edlerd: > I don't want redundant information. Either we keep a dedicated row for resource limits and remove it from the Configuration row, either we remove the resource limits row and include this info in the Configuration row.

I see, Personally leaning to 2. but no strong preference from my side.

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: What should we do then ? Wrap ? Or truncate (display `policies...`) ?
Ideally we should figure out how much can fit then display `+ n more`. I let copilot do something but it definitely smells fishy

edlerd: I think truncate with title should work fine. Wdyt?

kimanhou: Like this ?
image


This would be way simpler

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: > Can we put this into a component, such as ProfileConfigurationSections?

Yes

> We might also add it to the profile overview and side panel.

I put it in the Profile side panel. In the overview, I'm not sure as it already has resource limits and cloud init. I haven't touched the profile overview so far, except for changing labels. I think it can stay as it is, wdyt ?

src/pages/profiles/ProfileRichTooltip.tsx
edlerd: Yeah, I think this is better and simpler :+1: 

Though I realize we can't use the list as a title on the content, because it is not text only. so we'd need to build the list and the title as a string in parallel.

kimanhou: Oooh I just understood what you meant with title. Like this ?
image
Screenshot from 2026-01-14 11-26-41

edlerd: Yes, that was the idea. Though to verify -- is that title visible if you hover the configuration content with your mouse?

kimanhou: Yes ! I can't figure out how to take a screenshot but it does ! The title is fully visible on mouse hover.

edlerd: Perfect!

On Ubuntu, you can press the "print screen" button to start the screenshot app and capture context menus or open tooltips.

kimanhou: I don't have a print screen button on my keyboard and when I press it on the laptop directly, the title disappears 😬

src/components/DevicesSummaryList.tsx
```diff
@@ -0,0 +1,80 @@
...
+  maxWidth?: number;
+}
+
+const DevicesSummaryList: FC<Props> = ({ devices }) => {
+  const devicesOtherThanNic = devices.filter((device) => !isNicDevice(device));
+
+  if (devicesOtherThanNic.length === 0) {
+    return "-";
+  }
+
```
edlerd:
```suggestion
```
We probably don't need this here, and can rely on the code below to handle the case just fine.

src/pages/profiles/ProfileRichTooltip.tsx
```diff
@@ -0,0 +1,146 @@
...
+      </>
+    );
+  }
+
+  const profileDescription = profile ? profile.description || "-" : "-";
+  const usageCount = getProfileInstances(
+    projectName,
+    projectName === "default",
+    profile ? profile.used_by : [],
+  ).length;
```
edlerd, line 54:
```suggestion
  const usageCount = profile? profile.used_by.length : 0;
```
Can this be much simpler and include any usage?
Same for the `ProfileDetailPanelContent`.

src/util/instanceBulkActions.tsx
```diff
@@ -69,17 +70,34 @@ export const instanceActionLabel = (action: LxdInstanceAction): string => {
...
+    } else {
+      return pluralForm;
+    }
   }
 
-  return `${item}s`;
+  if (item === item.toUpperCase()) {
+    return `${item}S`;
+  } else if (item.charAt(0) === item.charAt(0).toUpperCase()) {
+    return `${item}s`;
```
edlerd:
```suggestion
```
We don't need this, do we?

src/pages/profiles/ProfileDetailPanelContent.tsx
```diff
@@ -40,21 +44,49 @@ const ProfileDetailPanelContent: FC<Props> = ({ profile, project }) => {
           <th className="u-text--muted last-of-section">Defined in</th>
           <td>{featuresProfiles ? "Current" : "Default"} project</td>
         </tr>
-        <tr className="u-no-border">
-          <th colSpan={2}>
-            <h3 className="p-muted-heading p-heading--5">Devices</h3>
-          </th>
+        <tr>
```
edlerd, line 48: Spacing in here is a bit inconsistent:
- "Defined in" and "Configuration" row should have the same height as description, root storage or others. Both are a bit too high at the moment.
- The "Limits" spacing and style need adjustments. The cpu and memory don't have borders and a bit more spacing than others. Shall we use the same as in the tooltip? So standard row like description or root storage with label "Limits" and value both cpu and memory in one value cell next to it with a middot. We could extract it as a component for reuse.

image

src/pages/profiles/ProfileRichTooltip.tsx
kimanhou: I'm not sure it's functionally the same thing. I copied it from `ProfileDetailPanelContent`, where only the instances seem to be counted. There is also a "[Go to instances](https://localhost:8407/ui/project/default/instances?profile=profile-with-a-super-duper-long-ass-name)" link next to it. I would prefer to leave this untouched.

image

edlerd: Ah right, that is why we filter by instances. Let's keep it as is :+1:

src/pages/profiles/ProfileDetailPanelContent.tsx
kimanhou: > "Defined in" and "Configuration" row should have the same height as description, root storage or others. Both are a bit too high at the moment.

Yes, sorry, forgot to remove a className after copy pasting.


> The "Limits" spacing and style need adjustments. The cpu and memory don't have borders and a bit more spacing than others. Shall we use the same as in the tooltip? So standard row like description or root storage with label "Limits" and value both cpu and memory in one value cell next to it with a middot. We could extract it as a component for reuse.

Yes

src/util/instanceBulkActions.tsx
```diff
@@ -69,17 +70,32 @@ export const instanceActionLabel = (action: LxdInstanceAction): string => {
...
+    }
   }
 
-  return `${item}s`;
+  if (item === item.toUpperCase()) {
+    return `${item}S`;
+  } else {
+    return `${item}s`;
+  }
 };
```
edlerd, line 106: Can you add a test for this? Especially the special cases. It seems the GPU to GPUs case is broken.

PR #1695, opened 2026-01-06
chore(deps) revert to not using swc version vitejs plugin react

edlerd:
## Done

- chore(deps) revert to not using swc version vitejs plugin react as it is incompatible with s390x architecture build


Kxiru (approved):
As long as the failing test is CI related, this is good to go! Please verify.

PR #1696, opened 2026-01-06
feature:[WD-31497] revamp nic static ip

omarelkashef:
## Done

- Uses a prefixed input for nic static ip inputs.

## QA

1. Run the LXD-UI:
    - Go to an instance (or create a new one)-> Configuration -> Network
    - Edit an existing network or attach a new one
    - Add static ipv4 and/or ipv6
    - Click Apply changes, then click Save

## Screenshots

image

### DIFF
--- a/src/components/PrefixedInput.tsx
+++ b/src/components/PrefixedInput.tsx
@@ -1,0 +1,54 @@
+import type { ReactElement } from "react";
+import { useLayoutEffect, useRef } from "react";
+
+import type { InputProps } from "@canonical/react-components";
+import { Input } from "@canonical/react-components";
+import classNames from "classnames";
+
+export type PrefixedInputProps = Omit<InputProps, "type"> & {
+  immutableText: string;
+};
+
+const PrefixedInput = ({
+  immutableText,
+  ...props
+}: PrefixedInputProps): ReactElement => {
+  const prefixTextRef = useRef<HTMLDivElement>(null);
+  const inputWrapperRef = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
+    const prefixElement = prefixTextRef.current;
+    const inputElement = inputWrapperRef.current?.querySelector("input");
+
+    if (prefixElement && inputElement) {
+      // Adjust the left padding of the input to be the same width as the immutable text.
+      // This displays the user input and the unchangeable text together as one combined string.
+      const prefixWidth = prefixElement.getBoundingClientRect().width;
+      inputElement.style.paddingLeft = `${prefixWidth}px`;
+    }
+  }, [immutableText, props.label]);
+  return (
+    <div
+      className={classNames("prefixed-input", {
+        "prefixed-input--with-label": !!props.label,
+      })}
+    >
+      <div className="prefixed-input__text" ref={prefixTextRef}>
+        {immutableText}
+      </div>
+      <div ref={inputWrapperRef}>
+        <Input
+          className={classNames("prefixed-input__input", props.className)}
+          type="text"
+          wrapperClassName={classNames(
+            "prefixed-input__wrapper",
+            props.wrapperClassName,
+          )}
+          {...props}
+        />
+      </div>
+    </div>
+  );
+};
+
+export default PrefixedInput;
--- a/src/components/PrefixedIpInput.tsx
+++ b/src/components/PrefixedIpInput.tsx
@@ -1,0 +1,105 @@
+import { type ClipboardEvent, type ReactElement } from "react";
+import PrefixedInput from "./PrefixedInput";
+import type { PrefixedInputProps } from "./PrefixedInput";
+import { getImmutableAndEditable, isIPv4 } from "util/subnetIpRange";
+
+type Props = Omit<
+  PrefixedInputProps,
+  "immutableText" | "maxLength" | "placeholder" | "name"
+> & {
+  cidr: string;
+  ip: string;
+  name: string;
+  onIpChange: (ip: string) => void;
+};
+
+const PrefixedIpInput = ({
+  cidr,
+  help,
+  onIpChange,
+  ip,
+  name,
+  ...props
+}: Props): ReactElement => {
+  const [networkAddress] = cidr.split("/");
+  const isIPV4 = isIPv4(networkAddress);
+
+  const [immutable, editable] = getImmutableAndEditable(cidr);
+
+  const inputValue = isIPV4
+    ? ip.split(".").slice(immutable.split(".").length).join(".")
+    : ip.replace(immutable, "");
+
+  const getIPv4MaxLength = () => {
+    const immutableOctetsLength = immutable.split(".").length;
+    const lengths = [15, 11, 7, 3]; // Corresponding to 0-3 immutable octets
+    return lengths[immutableOctetsLength];
+  };
+
+  const maxLength = isIPV4 ? getIPv4MaxLength() : editable.length;
+  const placeholder = props.disabled ? "" : editable;
+
+  const setIp = (editableValue: string) => {
+    const fullIp = editableValue
+      ? isIPV4
+        ? `${immutable}.${editableValue}`
+        : `${immutable}${editableValue}`
+      : "";
+    onIpChange(fullIp);
+  };
+
+  const handlePaste = (e: ClipboardEvent<HTMLInputElement>) => {
+    e.preventDefault();
+    const pastedText = e.clipboardData.getData("text");
+    if (isIPV4) {
+      const octets = pastedText.split(".");
+      const trimmed = octets.slice(0 - editable.split(".").length);
+      const ip = trimmed.join(".");
+      setIp(ip);
+    } else {
+      const ip = pastedText.replace(immutable, "");
+      setIp(ip);
+    }
+  };
+  return (
+    <PrefixedInput
+      help={
+        help ? (
+          help
+        ) : (
+          <>
+            {isIPV4 ? (
+              <>
+                The available range in this subnet is{" "}
+                <code>
+                  {immutable}.{editable}
+                </code>
+              </>
+            ) : (
+              <>
+                The available IPV6 address range is{" "}
+                <code>
+                  {immutable}
+                  {editable}
+                </code>
+              </>
+            )}
+            . If left empty, the address will be dynamically assigned.
+          </>
+        )
+      }
+      immutableText={isIPV4 ? `${immutable}.` : immutable}
+      maxLength={maxLength}
+      name={name}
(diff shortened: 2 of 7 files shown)


src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [networkAddress] = cidr.split("/");
+  const ipv6Prefix = networkAddress.substring(
+    0,
+    networkAddress.lastIndexOf(":"),
+  );
+  const subnetIsIpv4 = isIPv4(networkAddress);
+
+  const [startIp, endIp] = getIpRangeFromCidr(cidr);
+  const [immutable, editable] = getImmutableAndEditableOctets(startIp, endIp);
+  const [fullIp, setFullIp] = useState(ip);
```
edlerd:
```suggestion
```
Better avoid local state and rely on upstream caller to handle it. Otherwise, we risk going out of sync with the state this component keeps locally and the upstream state.

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onIpChange,
+  ip,
+  ...props
+}: Props): ReactElement => {
+  const [networkAddress] = cidr.split("/");
+  const ipv6Prefix = networkAddress.substring(
+    0,
+    networkAddress.lastIndexOf(":"),
+  );
+  const subnetIsIpv4 = isIPv4(networkAddress);
```
edlerd:
```suggestion
  const isV4 = isIPv4(networkAddress);
```

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const [startIp, endIp] = getIpRangeFromCidr(cidr);
+  const [immutable, editable] = getImmutableAndEditableOctets(startIp, endIp);
```
edlerd: I find it slightly confusing that `immutable` and `ipv6Prefix` are the same thing but have so different names.

Same for `editable` and `getIPv6Placeholder`. Can we align those a bit better?

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -24,25 +24,30 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
...
         isEnabled
-          ? `Choose an ${family} address within the subnet range ${ipAddressConfigValue || ""}, or leave empty for dynamic assignment`
+          ? family === "IPv6" &&
+            ipAddressConfigValue && (
+              <>
+                The available IPV6 address range is
+                <code>{ipAddressConfigValue}</code>. If left empty, the address
+                will be dynamically assigned.
+              </>
+            )
```
edlerd: Can we move this into the `PrefixedIpInput` as the other if family is ipv6?

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,117 @@
...
+          help
+        ) : subnetIsIpv4 ? (
+          <>
+            The available range in this subnet is{" "}
+            <code>
+              {immutable}.{editable}
+            </code>
+            . If left empty, the address will be dynamically assigned.
+          </>
+        ) : null
```
edlerd:
```suggestion
        ) : 
```
Instead of `null`, we should have the ipv6 case handled here. I think all info is available.

src/util/subnetIpRange.tsx
edlerd, line 1: Where does this come from? Is this missing the original licence? Can we also add tests for it?

src/util/subnetIpRange.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+
+/**
+ * Checks if an IPv4 address is valid for the given subnet.
+ *
+ * @param ip The IPv4 address to check, as a string
+ * @param cidr The subnet's CIDR notation e.g. 192.168.0.0/24
+ * @returns True if the IP is in the subnet, false otherwise
+ */
+export const isIpInSubnet = (ip: string, cidr: string): boolean => {
```
edlerd, line 74: I think we can directly use it for validation.

src/util/subnetIpRange.tsx
omarelkashef: They are copied over from https://github.com/canonical/maas-ui/blob/main/src/app/utils/subnetIpRange.ts

edlerd: Can we also pull the test file https://github.com/canonical/maas-ui/blob/main/src/app/utils/subnetIpRange.test.ts ?

src/util/subnetIpRange.tsx
omarelkashef: Do you mean for the follow up PR about validation? I agree with that.

src/util/subnetIpRange.tsx
omarelkashef: I added the tests from MAAS.

src/util/networks.spec.ts
edlerd, line 1: the new tests should be in a dedicated file for subnetIpRange.spec.ts

src/components/PrefixedInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const prefixTextRef = useRef<HTMLDivElement>(null);
+  const inputWrapperRef = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
+    const prefixElement = prefixTextRef.current;
+    const inputElement = inputWrapperRef.current?.querySelector("input");
+
+    if (prefixElement && inputElement) {
+      // Adjust the left padding of the input to be the same width as the immutable octets.
+      // This displays the user input and the unchangeable text together as one IP address.
```
edlerd:
```suggestion
      // Adjust the left padding of the input to be the same width as the immutable text.
      // This displays the user input and the unchangeable text together as one combined string.
```
This is a generic component and it seems the comment is misaligned.

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+    networkAddress.lastIndexOf(":"),
+  );
+  const ipv6PlaceholderColons = 7 - (immutableIPV6.match(/:/g) || []).length; // 7 is the maximum number of colons in an IPv6 address
+  const editableIPV6 = `${"0000:".repeat(ipv6PlaceholderColons)}0000`;
+
+  const [startIp, endIp] = getIpRangeFromCidr(cidr);
+  const [immutableIPV4, editableIPV4] = getImmutableAndEditableOctets(
+    startIp,
+    endIp,
+  );
```
edlerd:
```suggestion
  const [immutable, editable] = getImmutableAndEditable(cidr, isIPV4);
```
Let's extract most of the logic into a util function as below. We should add tests for the new cases. Extracting it also simplifies below as we deal only with one set of `immutable, editable` instead of two.

```
const getImmutableAndEditable = (cidr: string, isIPV4: boolean) => {
  if (isIPV4) {
    const [startIp, endIp] = getIpRangeFromCidr(cidr);
    return getImmutableAndEditableOctets(startIp, endIp);
  }

  const [networkAddress] = cidr.split("/");
  const immutableIPV6 = networkAddress.substring(
    0,
    networkAddress.lastIndexOf(":"),
  );
  const ipv6PlaceholderColons = 7 - (immutableIPV6.match(/:/g) || []).length; // 7 is the maximum number of colons in an IPv6 address
  const editableIPV6 = `${"0000:".repeat(ipv6PlaceholderColons)}0000`;

  return [immutableIPV6, editableIPV6];
};
```

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+const isIPv4 = (ip: string) => {
+  const ipv4Regex =
+    /^(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}$/;
+
+  return ipv4Regex.test(ip);
+};
+
+const PrefixedIpInput = ({
+  cidr,
```
edlerd, line 17: Do we need a fallback to a plain input when the cidr is empty?

The placeholder appears to be wrong in this case. Or we can set the placeholder to empty?

image

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+  name: string;
+  onIpChange: (ip: string) => void;
+};
+
+const isIPv4 = (ip: string) => {
+  const ipv4Regex =
+    /^(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}$/;
+
+  return ipv4Regex.test(ip);
+};
```
edlerd: Shall we move this to the util as well and add some test coverage?

edlerd (commented):
Some comments on the structure, but should be good to merge soon.

src/util/subnetIpRange.tsx
```diff
@@ -0,0 +1,134 @@
...
+};
+
+/**
+ * Get the immutable and editable parts of an IPv4 or IPv6 subnet.
+ *
+ * @param cidr The CIDR notation of the subnet
+ * @param isIPV4 Whether the subnet is IPv4 (true) or IPv6 (false)
+ * @returns The immutable and editable  as two strings in a list
+ */
+export const getImmutableAndEditable = (cidr: string, isIPV4: boolean) => {
```
edlerd: Please add some tests with ipv4 and ipv6 values.

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,104 @@
...
+  PrefixedInputProps,
+  "immutableText" | "maxLength" | "name" | "placeholder"
+> & {
+  cidr: string;
+  ip: string;
+  name: string;
+  onIpChange: (ip: string) => void;
+};
+
+const PrefixedIpInput = ({
```
edlerd, line 16: With v4 disabled, we get the wrong placeholder text

image

src/components/PrefixedIpInput.tsx
```diff
@@ -0,0 +1,103 @@
...
+                </code>
+              </>
+            )}
+            . If left empty, the address will be dynamically assigned.
+          </>
+        )
+      }
+      immutableText={isIPV4 ? `${immutable}.` : immutable}
+      maxLength={maxLength}
+      name={props.name}
```
edlerd:
```suggestion
      name={name}
```
Can we keep the name prop in the destructuring of the component?

PR #1697, opened 2026-01-06
feat(CreateIdentityModal): improve identity creation flow

kimanhou:
## Done

- improve CreateIdentityModal

## QA

1. Run the LXD-UI:
    - Go to Permissions > Identities
    - Click on "Create TLS Identity" (top right corner of the page)
    - Input a name, like "foo"
    - Press Enter, or click on "Create identity" (bottom right corner of the page)
    - Make sure the modal gives clear guidance in what the token is and how it is to be used
    - Make sure you can't close the modal unless the checkbox is ticked
    - Make sure the "Done" button is enabled after you tick the "I have copied the token" checkbox
    - Make sure the "Done" button is enabled after you click on the "Copy token" button
    - Make sure the CLI and UI tabs are correctly displayed and that you can switch from one to the other

## Screenshots

Before
before

After
Screenshot from 2026-01-09 15-00-41

### DIFF
--- a/src/pages/permissions/CreateIdentityModal.tsx
+++ b/src/pages/permissions/CreateIdentityModal.tsx
@@ -1,7 +1,16 @@
 import type { FC } from "react";
+import { useState } from "react";
 import ResourceLabel from "components/ResourceLabel";
-import { Modal } from "@canonical/react-components";
+import {
+  Accordion,
+  ActionButton,
+  List,
+  Modal,
+  Notification,
+  Tabs,
+} from "@canonical/react-components";
 import CodeSnippetWithCopyButton from "components/CodeSnippetWithCopyButton";
+import ConfirmationCheckbox from "components/ConfirmationCheckbox";
 
 interface Props {
   onClose: () => void;
@@ -10,25 +19,123 @@ interface Props {
 }
 
 const CreateIdentityModal: FC<Props> = ({ onClose, token, identityName }) => {
+  const [isConfirmed, setConfirmed] = useState(false);
+  const [howToUseActiveTab, setHowToUseActiveTab] = useState("cli-tab");
+
   return (
     <Modal
-      close={onClose}
       className="create-tls-identity"
-      title="Identity created"
+      title={
+        <>
+          Identity{" "}
+          <ResourceLabel
+            type="certificate"
+            value={identityName}
+            bold
+            truncate
+          />{" "}
+          created successfully
+        </>
+      }
+      buttonRow={[
+        <span className="u-float-left confirm-input" key="confirm-input">
+          <ConfirmationCheckbox
+            label="I have copied the token"
+            confirmed={[isConfirmed, setConfirmed]}
+          />
+        </span>,
+        <ActionButton
+          key="confirm-action-button"
+          appearance="positive"
+          className="u-no-margin--bottom"
+          onClick={onClose}
+          disabled={!isConfirmed}
+        >
+          Done
+        </ActionButton>,
+      ]}
+      closeOnOutsideClick={false}
     >
       {token && (
         <>
-          <p>
-            The identity trust token below can be used to log in with the newly
-            created identity{" "}
-            <ResourceLabel type="certificate" value={identityName} /> .{" "}
-            <b>
-              Once this modal is closed, the identity trust token can&rsquo;t be
-              generated again.
-            </b>
-          </p>
+          <CodeSnippetWithCopyButton
+            code={token}
+            title="Your identity trust token"
+            tooltipMessage="Copy token"
+            onCopyButtonClick={() => {
+              setConfirmed(true);
+            }}
+          />
+          <Notification
+            severity="caution"
+            title="Copy the identity trust token"
+          >
+            You will need the token to authenticate your client. <br />
+            Make sure to copy it now as you will not be able to see this again.
+          </Notification>
+          <Accordion
+            sections={[
+              {
+                title: "How to use it?",
+                content: (
+                  <>
+                    <Tabs
+                      links={[
+                        {
+                          label: "CLI",
+                          active: howToUseActiveTab === "cli-tab",
+                          onClick: () => {
+                            setHowToUseActiveTab("cli-tab");
+                          },
+                        },
+                        {
+                          label: "UI",
+                          active: howToUseActiveTab === "ui-tab",
+                          onClick: () => {
+                            setHowToUseActiveTab("ui-tab");
+                          },
+                        },
+                      ]}
+                    />
+                    {howToUseActiveTab === "cli-tab" && (
+                      <>
+                        For use with the LXC command-line tool, run:
+                        <CodeSnippetWithCopyButton
+                          code={`lxc remote add ${location.hostname} ${token}`}
+                          tooltipMessage="Copy command"
+                          className="u-no-margin--bottom"
+                        />
+                        <span className="u-text--muted p-text--small u-sv3">
+                          You can replace <code>{location.hostname}</code> with
+                          a nickname for this server.
+                        </span>
+                      </>
+                    )}
 
-          <CodeSnippetWithCopyButton code={token} />
+                    {howToUseActiveTab === "ui-tab" && (
+                      <List
+                        className="u-no-margin--bottom"
+                        items={[
+                          <>
+                            Open an unauthenticated browser on{" "}
+                            <code>{location.origin}</code>.
+                          </>,
+                          <>
+                            Select <b>Setup TLS login</b> and follow the
+                            instructions to configure the browser certificate.
+                          </>,
+                          <>
+                            Use this identity trust token to add the new browser
+                            certificate to this server&apos;s trust store.
+                          </>,
+                        ]}
+                      />
+                    )}
+                  </>
+                ),
+              },
+            ]}
+          />
         </>
       )}
     </Modal>
--- a/src/sass/styles.scss
+++ b/src/sass/styles.scss
@@ -403,6 +403,12 @@ body {
     @include large {
       width: 40rem;
     }
+
+    .p-modal__header {
+      .resource-label {
+        max-width: 14rem;
+      }
+    }
   }
 }
 
@@ -430,4 +436,8 @@ body {
 
 .u-margin-left--small {
   margin-left: $sph--small;
+}
+
+.mono-font {
+  font-family: unquote($font-monospace);
 }
--- a/src/components/CodeSnippetWithCopyButton.tsx
+++ b/src/components/CodeSnippetWithCopyButton.tsx
@@ -1,29 +1,44 @@
 import type { FC } from "react";
-import {
-  CodeSnippet,
-  CodeSnippetBlockAppearance,
-} from "@canonical/react-components";
+import type { ValueOf } from "@canonical/react-components";
+import { CodeSnippetBlockAppearance } from "@canonical/react-components";
+import { CodeSnippet } from "@canonical/react-components";
 import CopyToClipboard from "components/CopyToClipboard";
 import classnames from "classnames";
 
 interface Props {
   code: string;
   className?: string;
+  appearance?: ValueOf<typeof CodeSnippetBlockAppearance>;
+  title?: string;
+  tooltipMessage?: string;
+  onCopyButtonClick?: () => void;
 }
 
-const CodeSnippetWithCopyButton: FC<Props> = ({ code, className }) => {
+const CodeSnippetWithCopyButton: FC<Props> = ({
+  code,
+  className,
+  appearance = CodeSnippetBlockAppearance.LINUX_PROMPT,
+  title,
+  tooltipMessage,
+  onCopyButtonClick,
+}) => {
   return (
     <CodeSnippet
       className={classnames("code-snippet-with-copy-button-wrapper", className)}
       blocks={[
         {
-          appearance: CodeSnippetBlockAppearance.LINUX_PROMPT,
+          appearance,
+          title,
           code: (
             <div className="command-wrapper">
               <span className="command u-truncate" title={code}>
                 {code}
               </span>
-              <CopyToClipboard value={code} />
+              <CopyToClipboard
+                value={code}
+                tooltipMessage={tooltipMessage}
+                onCopyButtonClick={onCopyButtonClick}
+              />
             </div>
           ),
         },
--- a/src/components/ConfirmationCheckbox.tsx
+++ b/src/components/ConfirmationCheckbox.tsx
@@ -3,11 +3,11 @@ import { CheckboxInput } from "@canonical/react-components";
 
 interface Props {
   label: string;
-  force: [boolean, Dispatch<SetStateAction<boolean>>];
+  confirmed: [boolean, Dispatch<SetStateAction<boolean>>];
 }
 
-const ConfirmationForce: FC<Props> = ({ label, force }) => {
-  const [isForce, setForce] = force;
+const ConfirmationCheckbox: FC<Props> = ({ label, confirmed }) => {
+  const [isConfirmed, setConfirmed] = confirmed;
 
   return (
     <span className="u-float-left">
@@ -15,13 +15,13 @@ const ConfirmationForce: FC<Props> = ({ label, force }) => {
         inline
         label={label}
         tabIndex={-1}
-        defaultChecked={isForce}
+        checked={isConfirmed}
         onClick={() => {
-          setForce((prev) => !prev);
+          setConfirmed((prev) => !prev);
         }}
       />
     </span>
   );
 };
 
-export default ConfirmationForce;
+export default ConfirmationCheckbox;
--- a/src/components/CopyToClipboard.tsx
+++ b/src/components/CopyToClipboard.tsx
@@ -5,15 +5,23 @@ import { Button, Icon, Tooltip } from "@canonical/react-components";
 interface Props {
   value: string;
   children?: ReactNode;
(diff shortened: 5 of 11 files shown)


src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +12,60 @@ interface Props {
...
+
   return (
     <Modal
-      close={onClose}
       className="create-tls-identity"
-      title="Identity created"
+      title={
+        <>
+          Identity{" "}
+          <ResourceLabel type="certificate" value={identityName} bold /> created
```
edlerd:
```suggestion
           created
```
The title with the name creates an overflow problem with very long identity names. Better keep the user defined values out of the title.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +12,60 @@ interface Props {
 }
 
 const CreateIdentityModal: FC<Props> = ({ onClose, token, identityName }) => {
+  const [isConfirmed, setConfirmed] = useState(false);
```
edlerd, line 22: Good idea to keep track of this, can we set confirmed to true when the user copies the token with the button as well?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +12,60 @@ interface Props {
...
+          <Notification
+            severity="caution"
+            title="Copy the identity trust token"
+          >
+            You will need the token to authenticate your client. <br />
+            Make sure to copy it now as you will not be able to see this again.
+          </Notification>
+          To use this identity with the LXC command-line tool, run:
+          <CodeSnippetWithCopyButton
+            code={`lxc remote add <remote-name> ${token}`}
```
edlerd:
```suggestion
            code={`lxc remote add ${location.hostname} ${token}`}
```
can we use the hostname here? That should be a better default.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +12,60 @@ interface Props {
...
-              generated again.
-            </b>
-          </p>
-
-          <CodeSnippetWithCopyButton code={token} />
+          <CodeSnippetWithCopyButton
+            code={token}
+            title="Your identity trust token"
+            tooltipMessage="Copy token"
+          />
```
edlerd, line 68:
```suggestion
```
Shall we remove this one and keep just the command? It seems a bit excessive to display it twice in different code blocks.

edlerd (commented):
Good ideas on expanding the steps to give the user direction on how to use the token. Some thoughts on streamlining the contents a bit below.

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: Should I include the name somewhere else in the modal ?

edlerd: I think in the modal body, yes we can include it with truncation applied.

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: Okay

https://github.com/user-attachments/assets/51abecc3-7e19-4d89-8e16-509ce4e43266

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: Okay

image

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: Actually I set a max-width to the ResourceLabel. It works well for all screen sizes. What do you think ?
image

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: What if the user wants to keep the token somewhere else ? Or not use the command right away ? 
If we keep one, it's the token, not the command.

kimanhou: What about putting this section in an accordion ?

https://github.com/user-attachments/assets/beecae22-17c6-4c0b-baa1-1b827640e40b

edlerd: Accordion looks better :+1: 

Another thought on this:
- Users can use the token on the CLI -- this case you already explain well in the modal behind the accordion.
- Users can use the token in the UI to register their certificate. We should explain or link to docs on how to use the token in the UI as well.

src/pages/permissions/CreateIdentityModal.tsx
edlerd: perfect, thanks.

src/pages/permissions/CreateIdentityModal.tsx
edlerd: seems all right with truncation in the header :+1:

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: I did something but very unsure about the explanation.
image

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,181 @@ interface Props {
...
+                          <br />
+                          1. Open your browser and navigate to the address of
+                          the server you want to add (e.g.,{" "}
+                          <code>https://&lt;server-ip&gt;:8443</code>).
+                          <br />
+                          2. Follow the instructions to set up the browser
+                          certificate.
+                          <br />
+                          3. Run the command line displayed in the UI and paste
+                          the token the corresponding box.
```
edlerd: The token is for this server, so we can tell them which url to open, as it is the current domain and protocol / path.

2. Is correct, but then the 3rd step we can use this token as the identity trust token and don't need to use the command line.

kimanhou: Better ?

image

edlerd: Yes, thanks. I'll comment small adjustments below.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+                          position:
+                            howToUseActiveTab === "ui-tab"
+                              ? "relative"
+                              : "absolute",
+                        }}
+                      >
+                        <>
+                          To use this identity in the UI:
+                          <br />
+                          1. Open your browser and navigate to{" "}
```
edlerd:
```suggestion
                          1. Open an unauthenticated browser on {" "}
```

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+                          <code>{location.origin}</code>.
+                          <br />
+                          2. Follow the instructions to set up the browser
+                          certificate.
```
edlerd:
```suggestion
                          2. Select `Setup TLS login` and follow the instructions to configure the browser certificate.
```

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+                          <br />
+                          3. Use this token as the identity trust token in the
+                          corresponding input field.
```
edlerd:
```suggestion
                          3. Use this identity trust token to add the new browser certificate to this server’s trust store.
```

edlerd (commented):
Some thoughts on the ui steps. Wdyt?

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: I am not sure how to display `Setup TLS login`. Wdyt ?

1.
image

2.
image

3.
image

edlerd: I think 2. is fine.

src/components/ConfirmationForce.tsx
```diff
@@ -1,14 +1,21 @@
...
 import { CheckboxInput } from "@canonical/react-components";
 
 interface Props {
   label: string;
   force: [boolean, Dispatch<SetStateAction<boolean>>];
+  canBeSetExternally?: boolean;
 }
 
-const ConfirmationForce: FC<Props> = ({ label, force }) => {
+const ConfirmationForce: FC<Props> = ({ label, force, canBeSetExternally }) => {
```
edlerd: The name should be adjusted to the new use case. Maybe `ModalConfirmation`?

src/pages/login/CertificateAdd.tsx
```diff
@@ -72,7 +73,10 @@ const CertificateAdd: FC = () => {
               LXD is running:
             </p>
 
-            <CodeSnippetWithCopyButton code={identityTrustTokenCommand} />
+            <CodeSnippetWithCopyButton
+              code={identityTrustTokenCommand}
+              appearance={CodeSnippetBlockAppearance.LINUX_PROMPT}
+            />
```
edlerd: Better set the Linux prompt as the default, so we don't need to change this file. Wdyt?

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: Awesome ! It's already implemented then

image

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+  useLayoutEffect(() => {
+    // Measure both tab contents to get the maximum height
+    const measureHeight = () => {
+      if (cliContentRef.current && uiContentRef.current) {
+        const cliHeight = cliContentRef.current.offsetHeight;
+        const uiHeight = uiContentRef.current.offsetHeight;
+        const maxHeight = Math.max(cliHeight, uiHeight);
+        setTabContentHeight(maxHeight);
+      }
+    };
```
edlerd:
```suggestion
```
I think it is ok if the height changes between the different sections. Wdyt?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+            severity="caution"
+            title="Copy the identity trust token"
+          >
+            You will need the token to authenticate your client. <br />
+            Make sure to copy it now as you will not be able to see this again.
+          </Notification>
+          <Accordion
+            sections={[
+              {
+                title: "How to use this token?",
```
edlerd:
```suggestion
                title: "How to use this identity trust token?",
```
Shall we be more precise here? Though a bit more verbose.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+                              ? "visible"
+                              : "hidden",
+                          position:
+                            howToUseActiveTab === "cli-tab"
+                              ? "relative"
+                              : "absolute",
+                        }}
+                      >
+                        <>
+                          To use this identity with the LXC command-line tool,
```
edlerd:
```suggestion
                          For use with the LXC command-line tool,
```
This is not an identity, maybe this copy works?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+                              ? "visible"
+                              : "hidden",
+                          position:
+                            howToUseActiveTab === "ui-tab"
+                              ? "relative"
+                              : "absolute",
+                        }}
+                      >
+                        <>
+                          To use this identity in the UI:
```
edlerd:
```suggestion
```
We probably don't need an intro sentence. Wdyt?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +18,180 @@ interface Props {
...
+                          <br />
+                          1. Open an unauthenticated browser on{" "}
+                          <code>{location.origin}</code>.
+                          <br />
+                          2. Select <b>Setup TLS login</b> and follow the
+                          instructions to configure the browser certificate.
+                          <br />
+                          3. Use this identity trust token to add the new
+                          browser certificate to this server&apos;s trust store.
```
edlerd: Can we use a [list](https://canonical.github.io/react-components/?path=/story/components-list--default) without numbers here? I think it might provide a nicer formatting and be semantically correct.

src/sass/styles.scss
```diff
@@ -422,3 +437,7 @@ body {
     opacity: 0.6;
   }
 }
+
+.mono-font {
+  font-family: unquote($font-monospace);
+}
```
edlerd, line 443: good idea moving this here.

src/sass/styles.scss
```diff
@@ -402,6 +402,21 @@ body {
...
+    }
+
+    .tab-content-container {
+      overflow: hidden;
+      position: relative;
+
+      .tab-content-item {
+        width: 100%;
+      }
+    }
```
edlerd: I think we should conditionally render this in the component. Not rely on applied classes or CSS for displaying or hiding this content.
```suggestion
```

src/pages/permissions/CreateIdentityModal.tsx
edlerd:
```suggestion
                        <List
                          className="u-no-margin--bottom"
                          items={[
                            <>
                              Open an unauthenticated browser on{" "}
                              <code>{location.origin}</code>.
                            </>,
                            <>
                              Select <b>Setup TLS login</b> and follow the
                              instructions to configure the browser certificate.
                            </>,
                            <>
                              Use this identity trust token to add the new
                              browser certificate to this server&apos;s trust
                              store.
                            </>,
                          ]}
                        />
```
Can we use a [list](https://canonical.github.io/react-components/?path=/story/components-list--default) without numbers here? I think it might provide a nicer formatting and be semantically correct.

edlerd (commented):
Some more thoughts on the presentation copy and structure after a thorough review.

src/pages/permissions/CreateIdentityModal.tsx
kimanhou: It bothered me but yes I agree it might be overkill

src/components/ConfirmationForce.tsx
```diff
@@ -1,14 +1,21 @@
...
 import { CheckboxInput } from "@canonical/react-components";
 
 interface Props {
   label: string;
   force: [boolean, Dispatch<SetStateAction<boolean>>];
+  modalConfirmation?: boolean;
 }
 
-const ConfirmationForce: FC<Props> = ({ label, force }) => {
+const ConfirmationForce: FC<Props> = ({ label, force, modalConfirmation }) => {
```
edlerd:
```suggestion
const ModalConfirmation: FC<Props> = ({ label, force }) => {
```
I think there was a misunderstanding. I meant renaming the component as we use it not only for a force flag on the modal. Actually, already one other use case exists from `InstanceSnapshotActions` that should have renamed the component.

src/components/ConfirmationForce.tsx
```diff
@@ -1,14 +1,21 @@
-import type { Dispatch, FC, SetStateAction } from "react";
+import { useEffect, type Dispatch, type FC, type SetStateAction } from "react";
 import { CheckboxInput } from "@canonical/react-components";
 
 interface Props {
   label: string;
   force: [boolean, Dispatch<SetStateAction<boolean>>];
+  modalConfirmation?: boolean;
```
edlerd:
```suggestion
```
I don't understand why we need this. Can't we handle the state in the caller already?

src/components/ConfirmationForce.tsx
edlerd: We might want to rename other things inside the component. Then it might be an interesting candidate to upstream into react components, next to the shared Modal.

kimanhou: I refactored the component and renamed it `ConfirmationCheckbox` as it's only used in `confirmExtra` prop of `ConfirmationButton`.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +19,124 @@ interface Props {
...
+                      <>
+                        For use with the LXC command-line tool, run:
+                        <CodeSnippetWithCopyButton
+                          code={`lxc remote add ${location.hostname} ${token}`}
+                          tooltipMessage="Copy command"
+                          className="u-no-margin--bottom"
+                        />
+                        <span className="u-text--muted p-text--small u-sv3">
+                          You can replace{" "}
+                          <span className="mono-font">{location.hostname}</span>{" "}
```
edlerd:
```suggestion
                          <code>{location.hostname}</code>{" "}
```
I think `code` works better here.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -10,25 +19,124 @@ interface Props {
...
+            severity="caution"
+            title="Copy the identity trust token"
+          >
+            You will need the token to authenticate your client. <br />
+            Make sure to copy it now as you will not be able to see this again.
+          </Notification>
+          <Accordion
+            sections={[
+              {
+                title: "How to use this identity trust token?",
```
edlerd:
```suggestion
                title: "How to use it?",
```
Second thought, maybe more concise is better.

edlerd (approved):
LGTM, just tiny ideas.

PR #1698, opened 2026-01-07
feat: [WD-32447] Rich tooltips for pools

Kxiru:
## Done

- Created and added a Rich tooltip for pools in multiple places in the codebase.

Notes:
- ~~Please critique on whether the chip content is appropriate, should be spliced, or needds amendments.~~
- Includes special behaviour for cluster member specific pools so that only space usage on the primary member node is shown.

## QA

1. Run the LXD-UI:

## Screenshots

image

### DIFF
--- a/src/components/DeviceDetails.tsx
+++ b/src/components/DeviceDetails.tsx
@@ -4,23 +4,25 @@ import type { FormDevice } from "util/formDevices";
 import type { LxdDeviceValue } from "types/device";
 import ResourceLink from "components/ResourceLink";
 import { isHostDiskDevice } from "util/devices";
+import StoragePoolRichChip from "pages/storage/StoragePoolRichChip";
 import { ROOT_PATH } from "util/rootPath";
 
 interface Props {
   device: LxdDeviceValue;
   project: string;
+  location?: string;
 }
 
-const DeviceDetails: FC<Props> = ({ device, project }) => {
+const DeviceDetails: FC<Props> = ({ device, project, location }) => {
   if (device.type === "disk") {
     if (isRootDisk(device as FormDevice)) {
       return (
         <>
           Pool{" "}
-          <ResourceLink
-            type="pool"
-            value={device.pool || ""}
-            to={`${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
+          <StoragePoolRichChip
+            poolName={device.pool ?? ""}
+            projectName={project}
+            location={location}
           />
         </>
       );
@@ -39,10 +41,10 @@ const DeviceDetails: FC<Props> = ({ device, project }) => {
           to={`${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}/volumes/custom/${encodeURIComponent(device.source ?? "")}`}
         />{" "}
         on pool{" "}
-        <ResourceLink
-          type="pool"
-          value={device.pool || ""}
-          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
+        <StoragePoolRichChip
+          poolName={device.pool ?? ""}
+          projectName={project}
+          location={location}
         />
       </>
     );
--- a/src/components/DeviceListTable.tsx
+++ b/src/components/DeviceListTable.tsx
@@ -16,9 +16,10 @@ import type { LxdDeviceValue, LxdDevices } from "types/device";
 interface Props {
   configBaseURL: string;
   devices: LxdDevices;
+  location?: string;
 }
 
-const DeviceListTable: FC<Props> = ({ configBaseURL, devices }) => {
+const DeviceListTable: FC<Props> = ({ configBaseURL, devices, location }) => {
   const { project } = useParams<{ project: string }>();
 
   const byTypeAndName = (
@@ -88,7 +89,11 @@ const DeviceListTable: FC<Props> = ({ configBaseURL, devices }) => {
         },
         {
           content: (
-            <DeviceDetails device={device} project={project as string} />
+            <DeviceDetails
+              device={device}
+              project={project as string}
+              location={location}
+            />
           ),
           role: "cell",
           "aria-label": "Details",
--- a/src/context/useClusterMembers.tsx
+++ b/src/context/useClusterMembers.tsx
@@ -21,5 +21,6 @@ export const useClusterMember = (
   return useQuery({
     queryKey: [queryKeys.cluster, queryKeys.members, name],
     queryFn: async () => fetchClusterMember(name),
+    enabled: name !== "" && name !== "none",
   });
 };
--- a/src/pages/instances/InstanceDetailPanelContent.tsx
+++ b/src/pages/instances/InstanceDetailPanelContent.tsx
@@ -20,6 +20,7 @@ import type { LxdDevices } from "types/device";
 import NetworkRichChip from "pages/networks/NetworkRichChip";
 import ExpandableList from "components/ExpandableList";
 import ClusterMemberRichChip from "pages/cluster/ClusterMemberRichChip";
+import StoragePoolRichChip from "pages/storage/StoragePoolRichChip";
 
 const RECENT_SNAPSHOT_LIMIT = 5;
 
@@ -110,10 +111,10 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
         <tr>
           <th className="u-text--muted">Root storage</th>
           <td>
-            <ResourceLink
-              type="pool"
-              value={rootPool}
-              to={`${ROOT_PATH}/ui/project/${encodeURIComponent(instance.project)}/storage/pool/${encodeURIComponent(rootPool)}`}
+            <StoragePoolRichChip
+              poolName={rootPool}
+              projectName={instance.project}
+              location={instance.location}
             />
           </td>
         </tr>
--- a/src/pages/storage/StorageBuckets.tsx
+++ b/src/pages/storage/StorageBuckets.tsx
@@ -34,7 +34,6 @@ import StorageBucketActions from "./actions/StorageBucketActions";
 import CreateStorageBucketBtn from "./actions/CreateStorageBucketBtn";
 import SelectableMainTable from "components/SelectableMainTable";
 import SelectedTableNotification from "components/SelectedTableNotification";
-import ResourceLink from "components/ResourceLink";
 import usePanelParams, { panels } from "util/usePanelParams";
 import StorageBucketBulkDelete from "./actions/StorageBucketBulkDelete";
 import type { LxdStorageBucket } from "types/storage";
@@ -43,7 +42,7 @@ import EditStorageBucketPanel from "./panels/EditStorageBucketPanel";
 import StorageBucketLink from "./StorageBucketLink";
 import StorageBucketKeyCount from "./StorageBucketKeyCount";
 import DocLink from "components/DocLink";
-import { ROOT_PATH } from "util/rootPath";
+import StoragePoolRichChip from "./StoragePoolRichChip";
 
 const StorageBuckets: FC = () => {
   const notify = useNotify();
@@ -146,10 +145,10 @@ const StorageBuckets: FC = () => {
         },
         {
           content: (
-            <ResourceLink
-              type="pool"
-              value={bucket.pool}
-              to={`${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(bucket.pool)}`}
+            <StoragePoolRichChip
+              poolName={bucket.pool}
+              projectName={project}
+              location={bucket.location}
             />
           ),
           role: "cell",
--- a/src/pages/storage/StoragePoolHeader.tsx
+++ b/src/pages/storage/StoragePoolHeader.tsx
@@ -9,8 +9,8 @@ import * as Yup from "yup";
 import { testDuplicateStoragePoolName } from "util/storagePool";
 import { useFormik } from "formik";
 import { renameStoragePool } from "api/storage-pools";
-import ResourceLink from "components/ResourceLink";
 import { ROOT_PATH } from "util/rootPath";
+import StoragePoolRichChip from "./StoragePoolRichChip";
 
 interface Props {
   name: string;
@@ -49,7 +49,10 @@ const StoragePoolHeader: FC<Props> = ({ name, pool, project }) => {
           toastNotify.success(
             <>
               Storage pool <strong>{name}</strong> renamed to{" "}
-              <ResourceLink type="pool" value={values.name} to={url} />.
+              <StoragePoolRichChip
+                poolName={values.name}
+                projectName={project}
+              />
             </>,
           );
           formik.setFieldValue("isRenaming", false);
--- a/src/pages/storage/StoragePoolRichChip.tsx
+++ b/src/pages/storage/StoragePoolRichChip.tsx
@@ -1,0 +1,59 @@
+import type { FC } from "react";
+import { Tooltip } from "@canonical/react-components";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import ResourceLink from "components/ResourceLink";
+import { SMALL_TOOLTIP_BREAKPOINT } from "components/RichTooltipTable";
+import StoragePoolRichTooltip from "./StoragePoolRichTooltip";
+import { ROOT_PATH } from "util/rootPath";
+
+interface Props {
+  poolName: string;
+  projectName: string;
+  className?: string;
+  disabled?: boolean;
+  location?: string;
+}
+
+const StoragePoolRichChip: FC<Props> = ({
+  poolName,
+  projectName,
+  className,
+  disabled,
+  location,
+}) => {
+  const showTooltip = !useIsScreenBelow(SMALL_TOOLTIP_BREAKPOINT, "height");
+
+  const url = `${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/storage/pool/${encodeURIComponent(poolName)}`;
+
+  const resourceLink = (
+    <ResourceLink
+      type="pool"
+      value={poolName}
+      to={url}
+      hasTitle={!showTooltip}
+      className={className}
+      disabled={disabled}
+    />
+  );
+
+  if (!showTooltip) {
+    return <>{resourceLink}</>;
+  }
+
+  return (
+    <Tooltip
+      zIndex={1000}
+      message={
+        <StoragePoolRichTooltip
+          poolName={poolName}
+          url={url}
+          location={location}
+        />
+      }
+    >
+      {resourceLink}
+    </Tooltip>
+  );
+};
+
+export default StoragePoolRichChip;
--- a/src/pages/storage/StoragePoolRichTooltip.tsx
+++ b/src/pages/storage/StoragePoolRichTooltip.tsx
@@ -1,0 +1,105 @@
+import { type FC } from "react";
+import { Spinner } from "@canonical/react-components";
+import { type TooltipRow } from "components/RichTooltipRow";
+import {
+  MEDIUM_TOOLTIP_BREAKPOINT,
+  RichTooltipTable,
+} from "components/RichTooltipTable";
+import ResourceLabel from "components/ResourceLabel";
+import { Link } from "react-router-dom";
+import ItemName from "components/ItemName";
+import { useStoragePool } from "context/useStoragePools";
+import { useClusterMember } from "context/useClusterMembers";
+import StoragePoolSize from "./StoragePoolSize";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+
+interface Props {
+  poolName: string;
+  url: string;
+  location?: string;
+}
+
+const StoragePoolRichTooltip: FC<Props> = ({ poolName, url, location }) => {
+  const { data: pool, isLoading: isPoolLoading } = useStoragePool(poolName);
+  const { data: member, isLoading: isMemberLoading } = useClusterMember(
+    location || "none",
+  );
+
+  if (!pool && !isPoolLoading) {
+    return (
+      <>
+        Storage pool <ResourceLabel type="pool" value={poolName} bold /> not
+        found
+      </>
+    );
+  }
+
+  const showDriverDetails = !useIsScreenBelow(
+    MEDIUM_TOOLTIP_BREAKPOINT,
+    "height",
+  );
+
+  const rows: TooltipRow[] = [
+    {
+      title: "Storage pool",
+      value: pool ? (
+        <Link
+          to={url}
+          onClick={(e) => {
+            e.stopPropagation();
+          }}
+        >
+          <ItemName item={{ name: poolName }} />
+        </Link>
+      ) : (
+        <Spinner />
+      ),
+      valueTitle: poolName,
+    },
+    {
+      title: "Description",
+      value: pool?.description || "-",
+      valueTitle: pool?.description || "-",
+    },
+    {
+      title: "Driver",
+      value: pool?.driver || "-",
+    },
+    {
+      title: "Status",
+      value: pool?.status || "-",
+    },
+  ];
+
+  if (showDriverDetails) {
+    rows.push(
+      {
+        title: "Size",
+        value: pool ? (
+          isMemberLoading ? (
+            "-"
+          ) : (
+            <StoragePoolSize
+              pool={pool}
+              member={member ?? undefined}
+              hasMeterBar
+              forceSingleLine
+            />
+          )
+        ) : (
+          "-"
+        ),
+      },
+      {
+        title: "Used by",
+        value: pool ? pool.used_by?.length : "-",
+      },
+    );
+  }
+
+  return (
+    <RichTooltipTable rows={rows} className="storage-pool-rich-tooltip-table" />
+  );
+};
+
+export default StoragePoolRichTooltip;
--- a/src/pages/storage/StoragePoolSize.tsx
+++ b/src/pages/storage/StoragePoolSize.tsx
@@ -1,4 +1,4 @@
-import { type FC } from "react";
+import type { FC } from "react";
 import type { LxdStoragePool } from "types/storage";
 import { humanFileSize } from "util/helpers";
 import Meter from "components/Meter";
@@ -8,15 +8,24 @@ import { queryKeys } from "util/queryKeys";
 import { fetchStoragePoolResources } from "api/storage-pools";
 import { isClusterLocalDriver } from "util/storagePool";
 import { useIsClustered } from "context/useIsClustered";
+import type { LxdClusterMember } from "types/cluster";
 
 interface Props {
   pool: LxdStoragePool;
   hasMeterBar?: boolean;
+  member?: LxdClusterMember;
+  forceSingleLine?: boolean;
 }
 
-const StoragePoolSize: FC<Props> = ({ pool, hasMeterBar }) => {
+const StoragePoolSize: FC<Props> = ({
+  pool,
+  hasMeterBar,
+  member,
+  forceSingleLine,
+}) => {
+  // When a single member is provided, the resource usage is shown for that member only
   const { data: clusteredPoolResources = [] } =
-    useClusteredStoragePoolResources(pool.name);
+    useClusteredStoragePoolResources(pool.name, member);
   const isClustered = useIsClustered();
   const hasMemberSpecificSize =
     isClusterLocalDriver(pool.driver) && isClustered;
@@ -26,11 +35,12 @@ const StoragePoolSize: FC<Props> = ({ pool, hasMeterBar }) => {
     queryFn: async () => fetchStoragePoolResources(pool.name),
     enabled: !hasMemberSpecificSize,
   });
+
   const resourceList = hasMemberSpecificSize
     ? clusteredPoolResources
     : [poolResources];
 
-  if (!hasMeterBar && hasMemberSpecificSize) {
+  if (hasMemberSpecificSize && forceSingleLine && !member) {
     return "Cluster member dependent";
   }
 
--- a/src/pages/storage/StorageVolumeOverview.tsx
+++ b/src/pages/storage/StorageVolumeOverview.tsx
@@ -5,12 +5,11 @@ import StorageUsedBy from "pages/storage/StorageUsedBy";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import type { LxdStorageVolume } from "types/storage";
 import { isoTimeToString } from "util/helpers";
-import { ROOT_PATH } from "util/rootPath";
 import StorageVolumeSize from "pages/storage/StorageVolumeSize";
 import { renderContentType, renderVolumeType } from "util/storageVolume";
 import { useSettings } from "context/useSettings";
-import ResourceLink from "components/ResourceLink";
 import ClusterMemberRichChip from "pages/cluster/ClusterMemberRichChip";
+import StoragePoolRichChip from "./StoragePoolRichChip";
 
 interface Props {
   volume: LxdStorageVolume;
@@ -63,10 +62,10 @@ const StorageVolumeOverview: FC<Props> = ({ volume }) => {
               <tr>
                 <th className="u-text--muted">Pool</th>
                 <td>
-                  <ResourceLink
-                    type="pool"
-                    value={volume.pool}
-                    to={`${ROOT_PATH}/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(volume.pool)}`}
+                  <StoragePoolRichChip
+                    poolName={volume.pool}
+                    projectName={volume.project}
+                    location={volume.location}
                   />
                 </td>
               </tr>
--- a/src/pages/storage/StorageVolumes.tsx
+++ b/src/pages/storage/StorageVolumes.tsx
@@ -62,6 +62,7 @@ import DocLink from "components/DocLink";
 import ClusterMemberRichChip from "pages/cluster/ClusterMemberRichChip";
 import { ROOT_PATH } from "util/rootPath";
 import ProjectRichChip from "pages/projects/ProjectRichChip";
+import StoragePoolRichChip from "./StoragePoolRichChip";
 
 const StorageVolumes: FC = () => {
   const notify = useNotify();
@@ -251,10 +252,10 @@ const StorageVolumes: FC = () => {
         },
         {
           content: (
-            <ResourceLink
-              type="pool"
-              value={volume.pool}
-              to={`${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(volume.pool)}`}
+            <StoragePoolRichChip
+              poolName={volume.pool}
+              projectName={volume.project}
+              location={volume.location}
             />
           ),
           role: "cell",
--- a/src/util/instanceMigration.tsx
+++ b/src/util/instanceMigration.tsx
@@ -7,12 +7,12 @@ import type { LxdInstance } from "types/instance";
 import type { ReactNode } from "react";
 import { capitalizeFirstLetter } from "util/helpers";
 import { ROOT_PATH } from "util/rootPath";
-import ResourceLink from "components/ResourceLink";
 import { InstanceRichChip } from "pages/instances/InstanceRichChip";
 import { useNavigate } from "react-router-dom";
 import { useToastNotification } from "@canonical/react-components";
 import ClusterMemberRichChip from "pages/cluster/ClusterMemberRichChip";
 import ProjectRichChip from "pages/projects/ProjectRichChip";
+import StoragePoolRichChip from "pages/storage/StoragePoolRichChip";
 
 export type MigrationType =
   | "cluster member"
@@ -64,10 +64,10 @@ export const useInstanceMigration = ({
             projectName={instance.project}
           />{" "}
           root storage successfully moved to pool{" "}
-          <ResourceLink
-            type="pool"
-            value={target}
-            to={`${ROOT_PATH}/ui/project/${encodeURIComponent(instance.project)}/storage/pool/${encodeURIComponent(target)}`}
+          <StoragePoolRichChip
+            poolName={target}
+            projectName={instance.project}
+            location={instance.location}
           />
         </>
       );
--- a/src/context/useStoragePools.tsx
+++ b/src/context/useStoragePools.tsx
@@ -14,6 +14,7 @@ import {
   fetchStoragePools,
 } from "api/storage-pools";
 import { useClusterMembers } from "./useClusterMembers";
+import type { LxdClusterMember } from "types/cluster";
 
 export const useStoragePool = (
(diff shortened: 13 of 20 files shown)


src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,53 @@
...
+  projectName,
+  resourceLinkClassName,
+  resourceLinkDisabled,
+  url,
+}) => {
+  const showTooltip = !useIsScreenBelow(MEDIUM_TOOLTIP_BREAKPOINT, "height");
+
+  url =
+    url ??
+    `/ui/project/${encodeURIComponent(projectName)}/storage/pool/${encodeURIComponent(poolName)}`;
```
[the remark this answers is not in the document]
Kxiru: I have already removed redundant calls to the url prop within both StorageVolumes.tsx and StorageBuckets.tsx as url is resolved within this component and only areas with different / specific url path still maintain this call. If I have missed any elsewhere, please let me know.

src/pages/profiles/ProfileStorageList.tsx
```diff
@@ -17,11 +17,11 @@ const ProfileStorageList: FC<Props> = ({ profile, project }) => {
...
             .map((device) => (
-              <ResourceLink
+              <StoragePoolRichChip
                 key={device.path}
-                type="pool"
-                value={device.pool || ""}
-                to={`/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
+                poolName={device.pool || ""}
+                projectName={project}
+                url={`/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
```
omarelkashef: What is the difference between this url and the url constructed in the StoragePoolRichtooltip component?

src/pages/storage/StoragePoolHeader.tsx
```diff
@@ -48,7 +48,11 @@ const StoragePoolHeader: FC<Props> = ({ name, pool, project }) => {
           toastNotify.success(
             <>
               Storage pool <strong>{name}</strong> renamed to{" "}
-              <ResourceLink type="pool" value={values.name} to={url} />.
+              <StoragePoolRichChip
+                poolName={values.name}
+                projectName={project}
+                url={url}
```
omarelkashef: Same comment about url here.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,71 @@
...
+  if (!pool && !isPoolLoading) {
+    return (
+      <>
+        Storage pool <ResourceLabel type="pool" value={poolName} bold /> not
+        found
+      </>
+    );
+  }
+
+  if (!pool || isPoolLoading) {
```
omarelkashef: As discussed, this should be removed and we can add a spinner to the first row of the tooltip instead.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,71 @@
...
+    return <Spinner />;
+  }
+
+  const poolDescription = pool.description || "-";
+
+  const rows: TooltipRow[] = [
```
omarelkashef, line 42: The values of rows will need to have to be  "-" when the pool is not loaded yer after you move the if block above.

src/sass/_storage_pool_rich_tooltip.scss
```diff
@@ -0,0 +1,7 @@
+.storage-pool-rich-tooltip-table {
+  width: 550px;
```
omarelkashef: This width is very large

image

src/sass/_storage_pool_rich_tooltip.scss
```diff
@@ -0,0 +1,7 @@
...
+
+  .rich-tooltip-row-title {
```
omarelkashef: This makes the tooltip incosistent with the other tooltips. Is it necessary?

src/pages/profiles/ProfileStorageList.tsx
Kxiru: The clause that gets evaluated at the end has a ?? clause that needs to be evaluated.

src/pages/storage/StoragePoolHeader.tsx
Kxiru: None, here but the value was predefined for other purposes so it felt neither here nor there to include it.

src/sass/_storage_pool_rich_tooltip.scss
Kxiru: Yes, it was changed to accomodate the expendableList which we decided against earlier today.

src/sass/_storage_pool_rich_tooltip.scss
Kxiru: As above.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <ItemName item={{ name: poolName }} />
+        </Link>
+      ) : (
+        <Spinner />
+      ),
+      valueTitle: poolName,
+    },
+    {
+      title: "Description",
+      value: pool ? pool.description : "-",
```
omarelkashef: You can make this pool?.description || "-" so that we can still have "-" instead of an empty string when the pool does not have a description.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,83 @@
...
+      valueTitle: pool ? pool.description : "-",
+    },
+    {
+      title: "Status",
+      value: pool ? pool.status : "-",
+      valueTitle: pool ? pool.status : "-",
```
omarelkashef: If the status is never too long, and therefore never truncated, we do not need a valueTitle. This applies to other fields where the set of values are limited as well.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,83 @@
+import { type FC } from "react";
```
omarelkashef, line 1: If we can not find other rows for the tooltip, maybe having an exception and overriding the min-height is our best choice because the spacing does not look right. What do you think?

src/pages/profiles/ProfileStorageList.tsx
omarelkashef: Can this be done in the component?

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,83 @@
...
+      value: pool ? pool.status : "-",
+      valueTitle: pool ? pool.status : "-",
+    },
+    {
+      title: "Driver",
+      value: pool ? pool.driver : "-",
+      valueTitle: pool ? pool.driver : "-",
+    },
+    {
+      title: "Memory",
```
edlerd:
```suggestion
      title: "Size",
```
We use `Size` on the pool detail and overview page.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,83 @@
...
+      value: member ? (
+        <ClusterMemberMemoryUsage member={member} />
```
edlerd: This should be the `StoragePoolSize` in an unclustered environment.

For the clustered environment, we should use something like `StoragePoolClusterMember`, but only the bits for one cluster member that is passed in. If no member is passed in and there are multiple members, we show the fallback.

I think all this logic should go into a new component, something like `StoragePoolRichTooltipSize`.

src/pages/profiles/ProfileStorageList.tsx
Kxiru: Not this time. What usually goes in that clause is the poolName, and as you can see above they differ because one uses ?? and the other uses pipe symbols.

src/pages/storage/StoragePoolRichTooltip.tsx
Kxiru: I agree but @edlerd wishes for there to be consistency in the sizing across all the tooltips. @edlerd , what do you think of the current implementation? Would you add more rows or are you happy with it as is?

Kxiru:
> Can this be done in the component?

No, as the URL needs to access the device in the ..map and filter.

src/pages/storage/StoragePoolRichTooltip.tsx
Kxiru: How is StoragePoolClusterMember relevant to the size, here?
So I've implemented what it is I think you wanted to work with, but it seems like location isn't being seen in StoragePoolRichTooltipSize as the size field always defaults to StoragePoolSize.

Is it a safeassumption that if there is no cluster member, the user is not in a clustered environment?

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,21 @@
...
+  pool: LxdStoragePool;
+  member?: LxdClusterMember;
+  location?: string;
+}
+
+const StoragePoolRichTooltipSize: FC<Props> = ({ pool, member, location }) => {
+  return member && location ? (
+    <ClusterMemberMemoryUsage member={member} />
+  ) : (
+    <StoragePoolSize pool={pool} />
```
edlerd: We want the graphical version with a meter bar here, not the plain text one.

image

image

src/pages/storage/StoragePoolRichTooltip.tsx
edlerd: Ah right, `StoragePoolClusterMember` is the wrong component. Continuing the discussion in [another comment](https://github.com/canonical/lxd-ui/pull/1698#discussion_r2675049289).

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface Props {
+  pool: LxdStoragePool;
+  member?: LxdClusterMember;
+  location?: string;
+}
+
+const StoragePoolRichTooltipSize: FC<Props> = ({ pool, member, location }) => {
+  return member && location ? (
+    <ClusterMemberMemoryUsage member={member} />
```
edlerd: We want the storage pool size for a cluster member, but only if the pool is local and the environment is clustered and if the pool is local. This is the wrong component, pulling the cluster member memory. It should be the storage pool usage for the given cluster member instead.

If no cluster member is given here, we check if there is only one cluster member. If there is only one, we show the pool usage for the pool on that member

If the pool is not having a `isClusterLocalDriver` then `StoragePoolSize` is already doing the right thing.

If no cluster member is given here and there is more than one cluster member and `isClusterLocalDriver` for that pool driver, then we should show a fallback.

src/pages/storage/StoragePoolRichTooltip.tsx
edlerd: We discussed earlier this week having an exception for the pool tooltip due to a lack of things to show in it.

src/pages/storage/StoragePoolRichTooltipSize.tsx
Kxiru: I'm not sure I understand all of the different outcomes. For example, you said, "If no cluster member is given here and there is more than one cluster member ". How can we have no cluster member but also more than one?

Currently, this component accepts **one** cluster member, that either exists or doesn't. Will reach out to clarify the details.

edlerd: `ClusterMemberMemoryUsage` is the wrong component. We want the storage pool usage, not the memory/RAM.

> I'm not sure I understand all of the different outcomes. For example, you said, "If no cluster member is given here and there is more than one cluster member ". How can we have no cluster member but also more than one?

Depending on where this is used, we might have a cluster member defined or none defined, because the context does not define one. Then if we are in a cluster with more than one member, it becomes ambiguous which member's size to display here. That is the problem I referred to.

src/components/DeviceDetails.tsx
```diff
@@ -4,22 +4,23 @@ import type { FormDevice } from "util/formDevices";
...
       return (
         <>
           Pool{" "}
-          <ResourceLink
-            type="pool"
-            value={device.pool || ""}
-            to={`/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
+          <StoragePoolRichChip
+            poolName={device.pool ?? ""}
+            projectName={project}
```
edlerd, line 24: We should forward `location` here.

src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -106,10 +107,9 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
         <tr>
           <th className="u-text--muted">Root storage</th>
           <td>
-            <ResourceLink
-              type="pool"
-              value={rootPool}
-              to={`/ui/project/${encodeURIComponent(instance.project)}/storage/pool/${encodeURIComponent(rootPool)}`}
+            <StoragePoolRichChip
+              poolName={rootPool}
+              projectName={instance.project}
```
edlerd, line 116: We should forward `location` here from `instance.location`.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -59,10 +60,9 @@ const MigrateVolumeBtn: FC<Props> = ({
       />
     );
     const poolLink = (
-      <ResourceLink
-        type="pool"
-        value={newTarget}
-        to={`/ui/project/${encodeURIComponent(storageVolume.project)}/storage/pool/${encodeURIComponent(newTarget)}`}
+      <StoragePoolRichChip
+        poolName={newTarget}
+        projectName={storageVolume.project}
```
edlerd: We should forward `location` here from `volume.location`.

src/pages/storage/MigrateVolumeBtn.tsx
```diff
@@ -122,10 +122,9 @@ const MigrateVolumeBtn: FC<Props> = ({
           <ResourceLabel bold type="volume" value={volume.name} />
         );
         const poolLink = (
-          <ResourceLink
-            type="pool"
-            value={targetPool}
-            to={`/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(targetPool)}`}
+          <StoragePoolRichChip
+            poolName={targetPool}
+            projectName={volume.project}
```
edlerd: We should forward `location` here from `volume.location`.

src/pages/storage/StorageBuckets.tsx
```diff
@@ -145,11 +145,7 @@ const StorageBuckets: FC = () => {
         },
         {
           content: (
-            <ResourceLink
-              type="pool"
-              value={bucket.pool}
-              to={`/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(bucket.pool)}`}
-            />
+            <StoragePoolRichChip poolName={bucket.pool} projectName={project} />
```
edlerd: We should forward `location` here from `bucket.location`.

src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,61 @@
...
+  resourceLinkDisabled?: boolean;
+  url?: string;
+  location?: string;
+}
+
+const StoragePoolRichChip: FC<Props> = ({
+  poolName,
+  projectName,
+  resourceLinkClassName,
+  resourceLinkDisabled,
```
edlerd:
```suggestion
  className,
  disabled,
```
A bit simpler naming.

src/pages/storage/StorageVolumeOverview.tsx
```diff
@@ -65,10 +66,9 @@ const StorageVolumeOverview: FC<Props> = ({ volume }) => {
               <tr>
                 <th className="u-text--muted">Pool</th>
                 <td>
-                  <ResourceLink
-                    type="pool"
-                    value={volume.pool}
-                    to={`/ui/project/${encodeURIComponent(volume.project)}/storage/pool/${encodeURIComponent(volume.pool)}`}
+                  <StoragePoolRichChip
+                    poolName={volume.pool}
+                    projectName={volume.project}
```
edlerd, line 67: We should forward `location` here from `volume.location`.

src/pages/storage/StorageVolumes.tsx
```diff
@@ -248,10 +249,9 @@ const StorageVolumes: FC = () => {
         },
         {
           content: (
-            <ResourceLink
-              type="pool"
-              value={volume.pool}
-              to={`/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(volume.pool)}`}
+            <StoragePoolRichChip
+              poolName={volume.pool}
+              projectName={volume.project}
```
edlerd, line 257: We should forward `location` here from `volume.location`.

src/util/instanceMigration.tsx
```diff
@@ -65,10 +66,10 @@ export const useInstanceMigration = ({
...
           />{" "}
           root storage successfully moved to pool{" "}
-          <ResourceLink
-            type="pool"
-            value={target}
-            to={`/ui/project/${encodeURIComponent(instance.project)}/storage/pool/${encodeURIComponent(target)}`}
+          <StoragePoolRichChip
+            poolName={target}
+            projectName={instance.project}
+            url={`/ui/project/${encodeURIComponent(instance.project)}/storage/pool/${encodeURIComponent(target)}`}
```
edlerd: We should forward `location` here from `instance.location`.

edlerd (commented):
Some minor bits on naming and some cases where we can forward the location to the storage pool rich chip.

src/pages/storage/StoragePoolRichTooltipSize.tsx
edlerd: `ClusterMemberMemoryUsage` should not be used here. It relates to ram usage, not storage pool usage.
```suggestion
```

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+import ItemName from "components/ItemName";
+import { useStoragePool } from "context/useStoragePools";
+import type { LxdClusterMember } from "types/cluster";
+import StoragePoolRichTooltipSize from "./StoragePoolRichTooltipSize";
+import type { LxdStoragePool } from "types/storage";
+
+interface Props {
+  poolName: string;
+  url: string;
+  member?: LxdClusterMember;
```
edlerd:
```suggestion
```
We don't need the member here.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+  location?: string;
+}
+
+const StoragePoolRichTooltip: FC<Props> = ({
+  poolName,
+  url,
+  member,
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+    {
+      title: "Driver",
+      value: pool ? pool.driver : "-",
+    },
+    {
+      title: "Size",
+      value: (
+        <StoragePoolRichTooltipSize
+          pool={pool ?? ({} as LxdStoragePool)}
+          member={member}
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,89 @@
...
+    },
+    {
+      title: "Driver",
+      value: pool ? pool.driver : "-",
+    },
+    {
+      title: "Size",
+      value: (
+        <StoragePoolRichTooltipSize
+          pool={pool ?? ({} as LxdStoragePool)}
```
edlerd:
```suggestion
pool={pool}
```
a bit simpler. We should guard this component against an empty pool, not cast an empty object to a pool type.

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,21 @@
+import { type FC } from "react";
+import ClusterMemberMemoryUsage from "pages/cluster/ClusterMemberMemoryUsage";
+import type { LxdClusterMember } from "types/cluster";
+import StoragePoolSize from "./StoragePoolSize";
+import type { LxdStoragePool } from "types/storage";
+
+interface Props {
+  pool: LxdStoragePool;
+  member?: LxdClusterMember;
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,21 @@
...
+  location?: string;
+}
+
+const StoragePoolRichTooltipSize: FC<Props> = ({ pool, member, location }) => {
```
edlerd:
```suggestion
const StoragePoolRichTooltipSize: FC<Props> = ({ pool, location }) => {
```

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return member && location ? (
+    <ClusterMemberMemoryUsage member={member} />
+  ) : (
+    <StoragePoolSize pool={pool} hasMeterBar/>
```
edlerd: We need to find a way to display only the graph for the member that is passed in as `location` value.

See the screen below, the instance lives on member helped-polliwog, so we should only show the size of the pool on that member. Not the size from all three members in the cluster.

image

edlerd: See also this comment with a lot more context and explanation: https://github.com/canonical/lxd-ui/pull/1698#discussion_r2675049289

src/pages/storage/StoragePoolRichTooltip.tsx
Kxiru: Is that not what this is doing? Saying if there is no pool, send an empty object, the same way we would do (string ? string : "" ) How else would this be written?

Kxiru: We already have a guard on line 22 that isn't being picked up so how else do we guard it?

edlerd: `pool={pool ?? ({} as LxdStoragePool)}` is not a guard, but pretending the empty object has the right type. Guard is something like `pool && <StoragePoolRichTooltipSize pool=... />`

src/pages/profiles/ProfileStorageList.tsx
edlerd, line 1: This file is unused and should be removed.

src/components/DeviceListTable.tsx
edlerd, line 1: This component has a usage in profiles, where the location is unknown. This needs a fallback, as we have multiple members and as such we should show a fallback that fits into one line instead of reporting multiple sizes. Something like "Size | depending on location".

image

src/pages/storage/StoragePoolRichChip.tsx
edlerd, line 1: There is another use case in the profile list side panel, where we should apply the chip:

image

src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,61 @@
...
+  url?: string;
+  location?: string;
+}
+
+const StoragePoolRichChip: FC<Props> = ({
+  poolName,
+  projectName,
+  className,
+  disabled,
+  url,
```
edlerd:
```suggestion
```
We don't need the URL as a prop. We can rely on generating it internally.

src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,61 @@
...
+import ResourceLink from "components/ResourceLink";
+import { MEDIUM_TOOLTIP_BREAKPOINT } from "components/RichTooltipTable";
+import StoragePoolRichTooltip from "./StoragePoolRichTooltip";
+
+interface Props {
+  poolName: string;
+  projectName: string;
+  className?: string;
+  disabled?: boolean;
+  url?: string;
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const StoragePoolRichTooltipSize: FC<Props> = ({ pool, location }) => {
+  const { data: member } = useClusterMember(location);
+
+  return member ? (
+    <StoragePoolSize pool={pool} member={member} hasMeterBar />
+  ) : (
+    <StoragePoolSize pool={pool} hasMeterBar />
+  );
```
edlerd:
```suggestion
  const { data: member } = useClusterMember(location);

  return <StoragePoolSize pool={pool} member={member ?? undefined} hasMeterBar />;
```
I think we can now remove this component, as it is trivial.

src/components/DeviceListTable.tsx
edlerd: Be aware this component is also used in the storage pool list, where we don't want the fallback. And we already have the fallback with `!hasMeterBar`. Probably best to introduce a new prop to let the parent decide if the fallback should be used or all sizes should be listed.

src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { MEDIUM_TOOLTIP_BREAKPOINT } from "components/RichTooltipTable";
+import StoragePoolRichTooltip from "./StoragePoolRichTooltip";
+
+interface Props {
+  poolName: string;
+  projectName: string;
+  className?: string;
+  disabled?: boolean;
+  location?: string;
+  forceSingleLine?: boolean;
```
edlerd:
```suggestion
```
I wouldn't add this here to this level.

src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const StoragePoolRichChip: FC<Props> = ({
+  poolName,
+  projectName,
+  className,
+  disabled,
+  location,
+  forceSingleLine,
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  return (
+    <Tooltip
+      zIndex={1000}
+      message={
+        <StoragePoolRichTooltip
+          poolName={poolName}
+          url={url}
+          location={location}
+          forceSingleLine={forceSingleLine}
```
edlerd:
```suggestion
          forceSingleLine
```
We can set it at this level, as there is no tooltip where we don't want this.

src/components/DeviceDetails.tsx
```diff
@@ -4,22 +4,25 @@ import type { FormDevice } from "util/formDevices";
...
+            location={location}
+            forceSingleLine
```
edlerd:
```suggestion
```

src/components/DeviceDetails.tsx
```diff
@@ -38,10 +41,11 @@ const DeviceDetails: FC<Props> = ({ device, project }) => {
...
         on pool{" "}
-        <ResourceLink
-          type="pool"
-          value={device.pool || ""}
-          to={`/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
+        <StoragePoolRichChip
+          poolName={device.pool ?? ""}
+          projectName={project}
+          location={location}
+          forceSingleLine
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const StoragePoolRichTooltipSize: FC<Props> = ({
+  pool,
+  location,
+  forceSingleLine,
+}) => {
+  const { data: member, isLoading } = useClusterMember(location);
+
+  return isLoading ? (
+    "-"
```
edlerd:
```suggestion
    ""
```
We have different loading states and initially this is empty. So lets keep it empty until loading finished to avoid flicker.

src/pages/storage/StoragePoolRichTooltipSize.tsx
```diff
@@ -0,0 +1,31 @@
...
+import type { LxdStoragePool } from "types/storage";
+import { useClusterMember } from "context/useClusterMembers";
+
+interface Props {
+  pool: LxdStoragePool;
+  location: string;
+  forceSingleLine?: boolean;
+}
+
+const StoragePoolRichTooltipSize: FC<Props> = ({
```
edlerd: I don't think we need this component anymore. We can include this directly in `StoragePoolRichTooltip`. Wdyt?

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -8,15 +8,24 @@ import { queryKeys } from "util/queryKeys";
...
 
 interface Props {
   pool: LxdStoragePool;
   hasMeterBar?: boolean;
+  member?: LxdClusterMember;
+  forceSingleLine?: boolean;
 }
 
-const StoragePoolSize: FC<Props> = ({ pool, hasMeterBar }) => {
+const StoragePoolSize: FC<Props> = ({
```
edlerd, line 20: With the suggestion below, we have to revisit all usages where we either not set `hasMeterBar` or set it explicitly to false. In those we have to set `forceSingleLine`. But luckily there is only a handful or so, so the review and adjustment of it should be quick.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -26,11 +35,12 @@ const StoragePoolSize: FC<Props> = ({ pool, hasMeterBar }) => {
     queryFn: async () => fetchStoragePoolResources(pool.name),
     enabled: !hasMemberSpecificSize,
   });
+
   const resourceList = hasMemberSpecificSize
     ? clusteredPoolResources
     : [poolResources];
 
-  if (!hasMeterBar && hasMemberSpecificSize) {
+  if ((!hasMeterBar && hasMemberSpecificSize) || forceSingleLine) {
```
edlerd: I would suggest to make this a bit simpler. That way the component is more predictable and easier to reason about. This helps long term maintenance.
```suggestion
  if (hasMemberSpecificSize && forceSingleLine) {
```

src/pages/storage/StoragePoolSize.tsx
Kxiru: What is the reasoning for this?
You are suggesting:
- We revisit StoragePoolSize instances where MeterBar is NOT set,
- and proceed to mark these areas as false rather than default.

edlerd: I want to entangle the usage of hasmeterbar and forcesinglelline. The meter variable should not influence the decision of single line to make this complexity vanish.

But we need to make sure all existing useages stay as they were. So have to do a small survey of all usages and adjust to the new straight forward logic.

src/pages/storage/StoragePoolRichTooltipSize.tsx
Kxiru: It was nice to separate concerns but the logic can equally be moved, so no big deal.

src/pages/storage/StoragePoolSize.tsx
Kxiru: Okay, I believe there is only one occurence of StoragePoolSize where hasMeterBar is **not** set and we need to add ForceSingleLine. Pushing now!

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,96 @@
...
+      value:
+        pool &&
+        (isMemberLoading ? (
+          ""
+        ) : (
+          <StoragePoolSize
+            pool={pool}
+            member={member ?? undefined}
+            hasMeterBar
+            forceSingleLine={forceSingleLine}
```
edlerd:
```suggestion
            forceSingleLine
```
We can always enable it here.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,96 @@
...
+import ItemName from "components/ItemName";
+import { useStoragePool } from "context/useStoragePools";
+import { useClusterMember } from "context/useClusterMembers";
+import StoragePoolSize from "./StoragePoolSize";
+
+interface Props {
+  poolName: string;
+  url: string;
+  location?: string;
+  forceSingleLine?: boolean;
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+const StoragePoolRichTooltip: FC<Props> = ({
+  poolName,
+  url,
+  location,
+  forceSingleLine,
```
edlerd:
```suggestion
```

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,96 @@
...
+}) => {
+  const { data: pool, isLoading: isPoolLoading } = useStoragePool(poolName);
+  const { data: member, isLoading: isMemberLoading } = useClusterMember(
+    location || "",
+  );
```
edlerd, line 26: I suspect this will generate a request that causes an error when the location is empty. we should probalby do `enabled: name != "",` inside of the `useClusterMember` hook.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -26,11 +35,12 @@ const StoragePoolSize: FC<Props> = ({ pool, hasMeterBar }) => {
     queryFn: async () => fetchStoragePoolResources(pool.name),
     enabled: !hasMemberSpecificSize,
   });
+
   const resourceList = hasMemberSpecificSize
     ? clusteredPoolResources
     : [poolResources];
 
-  if (!hasMeterBar && hasMemberSpecificSize) {
+  if (hasMemberSpecificSize || forceSingleLine) {
```
```suggestion
  if (hasMemberSpecificSize && forceSingleLine) {
```
We only want the fallback if both of these are set.

src/pages/storage/StoragePoolRichChip.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+const StoragePoolRichChip: FC<Props> = ({
+  poolName,
+  projectName,
+  className,
+  disabled,
+  location,
+}) => {
+  const showTooltip = !useIsScreenBelow(MEDIUM_TOOLTIP_BREAKPOINT, "height");
```
edlerd: This should be the small breakpoints like in other chips. Currently the tooltip is hidden on reasonably large laptop sizes.

Also we should adjust the tooltip contents depending on various breakpoints. Such like in other existing chips.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -26,11 +35,12 @@ const StoragePoolSize: FC<Props> = ({ pool, hasMeterBar }) => {
     queryFn: async () => fetchStoragePoolResources(pool.name),
     enabled: !hasMemberSpecificSize,
   });
+
   const resourceList = hasMemberSpecificSize
     ? clusteredPoolResources
     : [poolResources];
 
-  if (!hasMeterBar && hasMemberSpecificSize) {
+  if (hasMemberSpecificSize && forceSingleLine) {
```
```suggestion
  if (hasMemberSpecificSize && forceSingleLine && !member) {
```
If the member is specified, we don't need to show the fallback here.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,109 @@
...
+            ""
+          ) : (
+            <StoragePoolSize
+              pool={pool}
+              member={member ?? undefined}
+              hasMeterBar
+              forceSingleLine
+            />
+          )),
+        valueTitle: "Size",
```
omarelkashef: I do not think there is a point to have a valueTitle if it is the same as title.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,109 @@
...
+    },
+  ];
+
+  if (showDriverDetails) {
+    rows.push(
+      {
+        title: "Size",
+        value:
+          pool &&
+          (isMemberLoading ? (
```
omarelkashef: Can we have the empty value as "-" to be consistent with other rows?

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,109 @@
...
+      value: pool ? pool.driver : "-",
+    },
+  ];
+
+  if (showDriverDetails) {
+    rows.push(
+      {
+        title: "Size",
+        value:
+          pool &&
```
omarelkashef: If pool is still loading, we will have an empty row.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,109 @@
...
+        title: "Status",
+        value: pool ? pool.status : "-",
+      },
+    );
+  }
+
+  if (showUsedBy) {
+    rows.push({
+      title: "Used by",
+      value: pool ? pool.used_by?.length : "-",
```
omarelkashef: nitpick: `pool?.used_by || "-"` to be consistent with the rest.

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,109 @@
...
+import StoragePoolSize from "./StoragePoolSize";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+
+interface Props {
+  poolName: string;
+  url: string;
+  location?: string;
+}
+
+const StoragePoolRichTooltip: FC<Props> = ({ poolName, url, location }) => {
```
omarelkashef, line 22: There is too much whitespace in the tooltip. If we can not find more fields to display, we can abandon the `LARGE_TOOLTIP_BREAKPOINT` and have one breakpoint instead of two and update the min-height for large screens accordingly in the css.

Image

Image

src/pages/storage/StoragePoolRichTooltip.tsx
```diff
@@ -0,0 +1,105 @@
...
+  if (!pool && !isPoolLoading) {
+    return (
+      <>
+        Storage pool <ResourceLabel type="pool" value={poolName} bold /> not
+        found
+      </>
+    );
+  }
+
+  const showDriverDetails = !useIsScreenBelow(
```
omarelkashef, line 37: It works well now, we only need to change the min-height to something around 270-300 px.

min-height: 300px
Image

current min-height: 400px

Image

src/context/useClusterMembers.tsx
```diff
@@ -21,5 +21,6 @@ export const useClusterMember = (
   return useQuery({
     queryKey: [queryKeys.cluster, queryKeys.members, name],
     queryFn: async () => fetchClusterMember(name),
+    enabled: !(name === "" || name === "none"),
```
edlerd:
```suggestion
    enabled: name !== "" && name !== "none",
```
Maybe a bit easier to read?

edlerd (approved):
LGTM, good stuff :)

PR #1701, opened 2026-01-08
feature: [WD-32446] Network rich chip

omarelkashef:
## Done

- Rich tooltip for networks

Note: src/components/TruncatedList.tsx is only included temporarily and should be removed once this branch is rebased on #1692 branch

## QA

1. Run the LXD-UI:
    - Network rich chips exist in the following pages:
      - Instance -> Overview 
      - Instance -> Configuration -> Network
      - Instances -> Side panel
      - Create network confirmation modal
      - Edit network confirmation modal
      - Delete network confirmation modal
      - Network -> Configuration
      - Network IPAM
      - Network -> Local peering
      -  Create local peer confirmation modal
      - Delete local peer confirmation modal
      - Profile -> Overview

## Screenshots

image


src/components/TruncatedList.tsx
edlerd, line 1: This seems to duplicate `InstanceListAddresses`. Can we adjust the use cases of `InstanceListAddresses` and use this one instead?

src/context/useNetworks.tsx
```diff
@@ -93,3 +94,27 @@ export const useNetworkFromClusterMembers = (
       (enabled ?? true) && isFineGrained !== null && clusterMembers.length > 0,
   });
 };
+
+export const useNetworkState = (
```
edlerd, line 98: Good idea to introduce it. Let's use it from `NetworkStatistics` as well.

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,198 @@
...
+import { useNetworkState } from "context/useNetworks";
+import { humanFileSize } from "util/helpers";
+
+interface Props {
+  networkName: string;
+  projectName: string;
+  clusterMemberName?: string;
+}
+
+const NetworkRichTooltip: FC<Props> = ({
```
edlerd, line 27: Let's use the property order [according to the spec](https://docs.google.com/document/d/128gz6ibtPq3I0M9yaWxBPRLVFsub5wX8aUDkxVENcag/edit?tab=t.0). Some properties are sitting too high, like the statistics.

src/pages/networks/forms/NetworkAddresses.tsx
```diff
@@ -19,19 +17,12 @@ const NetworkAddresses: FC<Props> = ({ formik, project }) => {
...
-      member,
-      queryKeys.state,
-    ],
-    retry: 0, // physical managed networks can sometimes 404, show error right away and don't retry
-    queryFn: async () => fetchNetworkState(networkName, project, member),
-  });
+  const { data: networkState, error } = useNetworkState(
+    networkName,
+    project,
+    true,
```
edlerd: Can we flip the order around so the enabled is the `true` element and can be omitted here?

src/pages/networks/NetworkRichChip.tsx
```diff
@@ -0,0 +1,59 @@
...
+import ResourceLink from "components/ResourceLink";
+import NetworkRichTooltip from "./NetworkRichTooltip";
+import { SMALL_TOOLTIP_BREAKPOINT } from "components/RichTooltipTable";
+
+interface Props {
+  networkName: string;
+  projectName: string;
+  clusterMemberName?: string;
+  resourceLinkClassName?: string;
+  resourceLinkDisabled?: boolean;
```
edlerd:
```suggestion
  clusterMember?: string;
  className?: string;
  disabled?: boolean;
```
I think we can use a bit more concise names, as their meaning is clear in the shorter form. `clusterMemberName` should also be shortened in the `NetworkRichTooltip` component.

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const rows: TooltipRow[] = [];
+
+    if (network.managed) {
+      rows.push({ title: "IPV4", value: ipv4 }, { title: "IPV6", value: ipv6 });
+    } else {
+      rows.push(
+        { title: "Parent", value: parent },
+        { title: "VLAN ID", value: vlanid },
+      );
+    }
```
edlerd: `network.managed` alone is the wrong condition. It has to be managed, and the OVN or bridge driver

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,199 @@
...
+              {counters?.packets_sent ?? 0} packets)
+            </div>
+          </>
+        ),
+      },
+      {
+        title: "Used by",
+        value: networkUsedbyCount,
+      },
+      getTypeSpecificRow(),
```
edlerd:
```suggestion
      getTypeSpecificRow(),
      {
        title: "Statistics",
        value: (
          <>
            <div className="general-field-label">
              RX: {humanFileSize(counters?.bytes_received ?? 0)} (
              {counters?.packets_received ?? 0} packets)
            </div>
            <div className="general-field-label">
              TX: {humanFileSize(counters?.bytes_sent ?? 0)} (
              {counters?.packets_sent ?? 0} packets)
            </div>
          </>
        ),
      },
      {
        title: "Used by",
        value: networkUsedbyCount,
      },
```
Let's match the order from the spec.

src/components/TruncatedList.tsx
```diff
@@ -0,0 +1,25 @@
+import type { FC, ReactNode } from "react";
+
+interface Props {
+  items: ReactNode[];
+  numberToShow?: number;
+}
+
+const TruncatedList: FC<Props> = ({ items, numberToShow = 2 }) => {
```
edlerd, line 8: I just realized @kimanhou is extracting the same component in another pr. Heads up as I hope there will not be conflicts between the two prs.

edlerd (commented):
Good progress :) 
Some comments below.

src/components/TruncatedList.tsx
omarelkashef: Yes, I realize that. I only added it to test it here. Once #1692 is merged, I will rebase and resolve conflict which should be easy.

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,202 @@
...
+      <>
+        Network <ResourceLabel type="network" value={networkName} bold /> not
+        found
+      </>
+    );
+  }
+
+  const networkUrl = clusterMember
+    ? `/ui/project/${encodeURIComponent(projectName)}/member/${encodeURIComponent(clusterMember)}/network/${encodeURIComponent(networkName)}`
+    : `/ui/project/${encodeURIComponent(projectName)}/network/${encodeURIComponent(networkName)}`;
```
edlerd: I think we shouldn't repeat this here. We could have a `getNetworkUrl` helper. Or maybe pass it in from the chip component?

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const rows = getSmallScreenRows();
+
+  if (isAboveMedium) {
+    rows.push(...getMediumScreenRows());
+  }
+
+  if (isAboveLarge) {
+    rows.push(...getLargeScreenRows());
+  }
```
edlerd, line 199: I, personally, prefer the more declarative way in `InstanceRichTooltip` with defining `const rows` and pushing to it within conditions. But this gets fairly complicated with the various cases to juggle about networks, as the fields differ by driver. Not really sure what is best here. Maybe keeping it as is is the best option.

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const networkUrl = clusterMember
+    ? `/ui/project/${encodeURIComponent(projectName)}/member/${encodeURIComponent(clusterMember)}/network/${encodeURIComponent(networkName)}`
+    : `/ui/project/${encodeURIComponent(projectName)}/network/${encodeURIComponent(networkName)}`;
+
+  const networkDescription = network?.description || "-";
+  const networkAcls = network?.config["security.acls"]?.split(",");
+  const ipv4 = network?.config["ipv4.address"] || "-";
+  const ipv6 = network?.config["ipv6.address"] || "-";
+  const networkUsedbyCount = network?.used_by?.length || 0;
+  const parent = network?.config.parent || "-";
```
edlerd: I just realized this is much more complicated. In a clustered environment, we have a parent per cluster member. It can be the same parent for all cluster members if their interfaces have the same name. It can also be different parents for some members.

To do this correctly, we'd have to fetch the config from each member and see if it is the same for all, if yes, we can display it here, otherwise we'd need a fallback to fit the single line.

edlerd: See `useNetworksFromClusterMembers` and `NetworkParentSelector` for how the clustered parents can be used.

src/pages/networks/NetworkRichTooltip.tsx
omarelkashef: I agree with you and prefer the declarative approach. However, in this case, the declarative approach does not look very clean due to the multiple conditionals. So, I think the way it is now is better.

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,218 @@
...
+    error: networkError,
+  } = useNetwork(networkName, projectName, clusterMember);
+  const { data: networkState } = useNetworkState(
+    networkName,
+    projectName,
+    clusterMember,
+    !isNetworkLoading || !networkError,
+  );
+  const { data: networksFromClusterMembers } =
+    useNetworksFromClusterMembers(projectName);
```
edlerd: We need to fetch the single network here that we pass in as `networkName`. Handling logic below should also adjust accordingly.

omarelkashef: Is it enough to filter fetched networks by name inside getParent()?

src/pages/networks/NetworkRichTooltip.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+      if (parents.length === 0) return "-";
+
+      const uniqueParents = [...new Set(parents)];
+      return uniqueParents.length === 1
+        ? uniqueParents[0]
+        : `${uniqueParents.length} different parents`;
+    }
+    return network?.config.parent || "-";
+  };
```
edlerd: Can we push this with the `useNetworkFromClusterMembers` into its own component? That way we don't trigger the api call on smaller screens where the information will not be displayed.

omarelkashef: We can pass isAboveMedium in place of the enabled parameter for `useNetworkFromClusterMembers` instead. We can not make the parent into its own component because all the rows are added to an array which are later rendered in the tooltip.

edlerd: I think the `rows` array can hold a reference to a dedicated component. And if the conditional part is not hit, so the component is not in the array, then it won't be rendered. It seems to isolate the individual parts a bit better and make this NetworkRichTooltip component a bit simpler.

edlerd: This will be similar to the ACLs row using the `TruncatedList` component.

src/pages/instances/InstanceDetailPanelContent.tsx
edlerd, line 1: This component was changed due to the profile-rich chip PR. Please adapt to the upstream change. Remove adding the networking section back and replace the network links with the network chip.

src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -113,33 +115,6 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
...
-                <br />
-                <Link
-                  to={`/ui/project/${encodeURIComponent(instance.project)}/instance/${encodeURIComponent(instance.name)}/configuration/networks`}
-                >
-                  Configure instance networks
-                </Link>
-              </p>
-            )}
-          </td>
-        </tr>
```
edlerd, line 142: Let's keep the networks up here and not move them to the bottom. This was a conscious choice to change them up here. This way the position is aligned between the profile side panel, profile rich chip and the instance side panel in here.

src/pages/instances/InstanceDetailPanelContent.tsx
```diff
@@ -113,33 +115,42 @@ const InstanceDetailPanelContent: FC<Props> = ({ instance }) => {
...
+              </h3>
+            </th>
+            <td>
+              <List
+                className="list"
+                items={networkDevices.map((item) => (
+                  <NetworkRichChip
+                    key={item.network}
+                    networkName={item.network}
+                    projectName={instance.project}
```
edlerd, line 128: There was a misunderstanding here. We don't want to change anything here, but the network chip. As is it is harming the flow of the side panel

image

edlerd: Especially the "Networks" shouldn't be a link or styled differently than "Root storage" in the row above.

PR #1708, opened 2026-01-09
feat(build): handle relative URL

### LINKED ISSUE #1379: Use relative URLs or add path prefix configuration
As is right now the ui basically requires to own the entire domain resource namespace from / (or at least (/ui/), preventing usage with other services.

Please either use relative URLs, allowing serving from any directory (though requires matching logic from end), or add configurable path prefix option.

Thank you

kimanhou:
## Done

- Support for dynamic path prefixes

Fixes #1379

## QA

1. Run the LXD-UI:
    - Checkout this branch
    - Build locally
    - Run playwright tests locally
    - Run locally


index.html
```diff
@@ -6,9 +6,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
     <title>LXD UI</title>
-    <link rel="shortcut icon" href="/assets/img/favicon-32x32.png" type="image/x-icon">
-    <link rel="manifest" href="/manifest.json" id="manifest">
-    <script src="/ui/assets/js/manifest.js"></script>
+    <link rel="shortcut icon" href="%BASE_URL%assets/img/favicon-32x32.png" type="image/x-icon">
+    <link rel="manifest" href="%BASE_URL%manifest.json" id="manifest">
+    <script src="%BASE_URL%assets/js/manifest.js"></script>
```
edlerd: This file is packaged at LXD build time. I don't think we should change it with a template here. Maybe we can update it on first load? This file is packaged at LXD build time. I don't think we should change it with a template here. Perhaps we can update it on first load?

src/App.tsx
```diff
@@ -108,7 +117,7 @@ const PermissionIdpGroups = lazy(
   async () => import("pages/permissions/PermissionIdpGroups"),
 );
 
-const HOME_REDIRECT_PATHS = ["/", "/ui", "/ui/project"];
+const HOME_REDIRECT_PATHS = ["/", "/project"];
```
edlerd: Why remove the `/ui` from this list? I think we should remove the `/` prefix from the original three.

src/util/helpers.tsx
```diff
@@ -286,19 +286,22 @@ export const isDimensionBelow = (
     : window.innerHeight < dimension;
 
 export const logout = (hasOidc?: boolean, hasCertificate?: boolean): void => {
-  if (window.location.href.includes("/ui/login")) {
+  const base = import.meta.env.BASE_URL;
```
edlerd: Where does the `import.meta.env.BASE_URL` come from? It seems like a vite specific variable that is not available in the compiled production environment.

edlerd (commented):
There might be a misunderstanding here. We do have to keep the ui prefix in the urls, but we might have additional path as prefix.

index.html
```diff
@@ -6,9 +6,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
     <title>LXD UI</title>
-    <link rel="shortcut icon" href="/assets/img/favicon-32x32.png" type="image/x-icon">
-    <link rel="manifest" href="/manifest.json" id="manifest">
-    <script src="/ui/assets/js/manifest.js"></script>
+    <link rel="shortcut icon" href="assets/img/favicon-32x32.png" type="image/x-icon">
+    <link rel="manifest" href="manifest.json" id="manifest">
+    <script src="assets/js/manifest.js"></script>
```
edlerd: Second thought: This must be an absolute path and it should be the correct one. Because a user can directly navigate to a subfolder such as `/ui/project/default/instances`. Using relative paths would lead to the wrong path when accessing nested folders in the ui.

I wonder if this problem makes the whole task infeasible or if there is a solution to this.

src/util/helpers.tsx
kimanhou: I think Vite actually handles this through static replacement during the build process. Vite finds every instance of `import.meta.env.BASE_URL` and replaces it with the string defined in our vite.config.ts (which defaults to `VITE_APP_BASE` from .env, or `./` if that is missing).

index.html
kimanhou: I think Vite handles this by rewriting absolute paths in index.html during the build process. It intercepts absolute-looking paths and automatically prepends `VITE_APP_BASE` or `./`.

edlerd: The problem I experience:

1. At build time, Vite cannot know under which path the UI will be served. Such as it could be served under `https://foo.bar/baz/bam/ui` So this has to be dynamic.
2. At run time the user can access a UI detail page, such as `/ui/project/default`, so this has to be a static url starting from the root path `/`. It cannot be a relative path.

A possible solution is to make the paths here static and assume the ui is served under the root path `/`. So this is the same as it was in the past.
Add a new a dynamic bit of JS to the index.html. It identifies if the UI is served under a different path than the root `/`. If it is not the root but a nested path, then the urls for the shortcut icon, manifest.json and manifest.js will be adjusted. Such as from `/ui/manifest.json` to `/baz/bam/ui/manifest.json`.

edlerd: Generally I think it is simpler to not configure `VITE_APP_BASE`, and have the `/ui` everywhere. But we need to add a dynamic bit that identifies the base path at runtime and changes `/ui` to `/baz/bam/ui`. Or whatever the currently used prefix is. We can't solve the problem at build time.

playwright.config.ts
```diff
@@ -42,7 +42,7 @@ const config: PlaywrightTestConfig<TestOptions> = {
   use: {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
     actionTimeout: 0,
-    baseURL: "https://localhost:8407/",
+    baseURL: "https://localhost:8407/ui/",
```
edlerd: I don't think we should change the test config here.

src/App.tsx
```diff
@@ -108,7 +123,7 @@ const PermissionIdpGroups = lazy(
   async () => import("pages/permissions/PermissionIdpGroups"),
 );
 
-const HOME_REDIRECT_PATHS = ["/", "/ui", "/ui/project"];
+const HOME_REDIRECT_PATHS = ["/", "/ui", "/project"];
```
edlerd: Shouldn't this stay as it was?

src/App.tsx
```diff
@@ -158,8 +172,13 @@ const App: FC = () => {
...
 
-  if (!isAuthenticated && !window.location.href.includes("/ui/login")) {
-    return <Navigate to="/ui/login" replace={true} />;
+  const isLoginPage =
+    window.location.pathname.includes(`${UI_BASE_PATH}login`) ||
+    window.location.pathname.includes("certificate-generate") ||
+    window.location.pathname.includes("certificate-add");
+
+  if (!isAuthenticated && !isLoginPage) {
+    return <Navigate to="/login" replace={true} />;
```
edlerd: I don't think this needs to change. It should work as it was.

src/App.tsx
```diff
@@ -20,6 +19,22 @@ import {
...
+}
+
+const NavigationShell: FC = () => (
+  <>
+    <Navigation />
+    <main className="l-main">
+      <Outlet />
+    </main>
+  </>
+);
```
edlerd: Why are we moving the navigation component here?

src/App.tsx
```diff
@@ -169,386 +188,381 @@ const App: FC = () => {
...
+              }
+            />
+            <Route
+              path="instances/create"
+              element={
+                <ProtectedRoute
+                  outlet={<ProjectLoader outlet={<CreateInstance />} />}
+                />
+              }
+            />
```
edlerd: Why are we changing all of these to be relative or nested? I think the previous more explicit way was a bit easier to understand. Though maybe we need to include the `UI_BASE_PATH` varibale in the route definitions?

src/api/auth-groups.tsx
```diff
@@ -16,7 +17,7 @@ export const fetchGroups = async (
   params.set("recursion", "1");
   addEntitlements(params, isFineGrained, groupEntitlements);
 
-  return fetch(`/1.0/auth/groups?${params.toString()}`)
+  return fetch(`${API_ROOT}/auth/groups?${params.toString()}`)
```
edlerd:
```suggestion
  return fetch(`${ROOT_PATH}/1.0/auth/groups?${params.toString()}`)
```
Shall we keep the /1.0 outside of the variable? In my opinion that has 3 advantages:
1. We can reuse the variable `ROOT_PATH` for the api, for the ui links and for route definitions. No need to distinguish between `API_ROOT` and `UI_ROOT`.
2. The diff becomes smaller and the code easier to maintain as the visible 1.0 in the path is easy to connect to lxd api. Especially for people that are not familiar with ui code base and come from the golang lxd team.
3. If the ROOT_PATH is not set or empty due to an error and users don't use a prefix, this works just fine. So it seems a bit more robust

src/components/DeviceDetails.tsx
```diff
@@ -19,7 +19,7 @@ const DeviceDetails: FC<Props> = ({ device, project }) => {
           <ResourceLink
             type="pool"
             value={device.pool || ""}
-            to={`/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
+            to={`/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
```
edlerd:
```suggestion
            to={`${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/storage/pool/${encodeURIComponent(device.pool ?? "")}`}
```
I think we should keep the ui prefix here and inject the ROOT_PATH on the fly. See other comment for full reasoning.

src/index.tsx
```diff
@@ -2,13 +2,16 @@ import { createRoot } from "react-dom/client";
...
 
 const rootElement = document.getElementById("app");
 
 if (!rootElement) throw new Error("Failed to find the root element");
 
 const root = createRoot(rootElement);
 
-const router = createBrowserRouter([{ path: "*", Component: Root }]);
+const router = createBrowserRouter([{ path: "*", Component: Root }], {
+  basename: UI_BASE_PATH,
```
edlerd: Considering the other comments, maybe this should stay as it was?

src/util/usedBy.tsx
```diff
@@ -56,7 +56,7 @@ export const filterUsedByType = (
           return path.includes("/forwards/");
         }
 
-        return path.startsWith(`/1.0/${type}`);
+        return path.includes(`/1.0/${type}`);
```
edlerd: I think this relates to api responses from LXD, so it can stay as it was.

src/util/resourceDetails.tsx
```diff
@@ -33,15 +41,15 @@ export const extractResourceDetailsFromUrl = (
...
   const urlSegments = url.pathname.split("/");
-  const name = decodeURIComponent(urlSegments[urlSegments.length - 1]);
+  const rawName = decodeURIComponent(urlSegments[urlSegments.length - 1]);
   const resourceName =
-    (identityNamesLookup ?? {})[name] ||
-    (imageLookup ?? {})[name]?.name ||
-    name;
+    (identityNamesLookup ?? {})[rawName] ||
+    (imageLookup ?? {})[rawName]?.name ||
+    rawName;
```
edlerd: This also refers to urls coming from LXD api responses, so probably no change needed.

src/util/resourceDetails.tsx
```diff
@@ -60,28 +68,35 @@ export const extractResourceDetailsFromUrl = (
...
+    if (path.includes("/storage-pools/")) {
+      resourceDetail.pool = getSegmentAfter(urlSegments, "storage-pools");
+
+      if (urlSegments.includes("custom")) {
+        resourceDetail.volume = getSegmentAfter(urlSegments, "custom");
+      } else if (urlSegments.includes("virtual-machine")) {
+        resourceDetail.volume = getSegmentAfter(urlSegments, "virtual-machine");
+      } else if (urlSegments.includes("container")) {
+        resourceDetail.volume = getSegmentAfter(urlSegments, "container");
+      }
```
edlerd: This also refers to urls coming from LXD api responses, so probably no change needed.

src/util/projects.tsx
```diff
@@ -21,10 +21,16 @@ export const projectSubpages = [
 ];
 
 export const getSubpageFromUrl = (url: string): string | undefined => {
-  const parts = url.split("/");
+  const parts = url.split("?")[0].split("/");
```
edlerd: I think it would be simpler to remove the prefix from the url and keep the rest of the logic as is.

src/util/operations.tsx
edlerd, line 1: I think all funcitons in this file refer to urls coming from LXD api responses, so probably no change needed.

src/context/useCurrentProject.tsx
edlerd, line 1: The simpler approach here might be as well to cleanup the url by removing any possible prefix and keeping the logic as is.

src/App.tsx
kimanhou: Because the Navigation component needs to know what project we are on and this now uses URL params (`useParams`), so it needs to be inside `RouterProvider`.

src/App.tsx
kimanhou: Yeah I added it because I was stuck in an infinite loop but it works with the old version now. I'll undo this change.

src/App.tsx
kimanhou: I thought it was cleaner and easier to maintain. If the prefix changes, we don't have to update dozens of explicit strings. But I don't have a strong opinion on this.

src/api/server.tsx
```diff
@@ -10,12 +10,13 @@ import type { LxdResources } from "types/resources";
...
 import type { ClusterSpecificValues } from "components/ClusterSpecificSelect";
 import { addTarget } from "util/target";
+import { ROOT_PATH } from "util/helpers";
 
 export const fetchSettings = async (target?: string): Promise<LxdSettings> => {
   const params = new URLSearchParams();
   addTarget(params, target);
 
-  return fetch(`/1.0?${params.toString()}`)
+  return fetch(`${ROOT_PATH}?${params.toString()}`)
```
edlerd: here we need the 1.0.

src/api/server.tsx
```diff
@@ -59,7 +60,7 @@ export const updateSettings = async (
   const params = new URLSearchParams();
   addTarget(params, target);
 
-  await fetch(`/1.0?${params.toString()}`, {
+  await fetch(`${ROOT_PATH}?${params.toString()}`, {
     method: "PATCH",
     headers: {
       "Content-Type": "application/json",
```
edlerd, line 66: here we need the 1.0.

src/api/server.tsx
```diff
@@ -149,7 +150,8 @@ export const fetchDocObjects = async (
     return [];
   }
 
-  return fetch("/documentation/objects.inv.txt")
+  const prefix = ROOT_PATH.replace("/1.0", "");
+  return fetch(`${prefix}/documentation/objects.inv.txt`)
```
edlerd:
```suggestion
  return fetch(`${ROOT_PATH}/documentation/objects.inv.txt`)
```
I think this simplifies.

src/api/images.tsx
```diff
@@ -76,7 +77,7 @@ export const deleteImageBulk = async (
         const item: BulkOperationItem = {
           name: name,
           type: "image",
-          href: `/ui/project/${encodeURIComponent(project)}/images`,
+          href: `/project/${encodeURIComponent(project)}/images`,
```
edlerd:
```suggestion
          href: `${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/images`,
```
I think we have to preserve ui here and insert the root path as well.

src/util/helpers.tsx
```diff
@@ -285,19 +285,24 @@ export const isDimensionBelow = (
...
+
+  if (window.location.pathname.includes(`${cleanBase}login`)) {
     return;
   }
-  void fetch("/oidc/logout").then(() => {
+
+  void fetch(`${cleanBase}oidc/logout`).then(() => {
     if (hasOidc) {
-      window.location.href = "/ui/login/";
+      window.location.href = `${cleanBase}login/`;
```
edlerd:
```suggestion
      window.location.href = `${ROOT_PATH}/ui/login/`;
```
Can we use the `ROOT_PATH` for the ui urls as well? Then most changes would be minimal when calculating links or target urls.

tests/docs-screenshots-clustered.spec.ts
```diff
@@ -15,7 +15,7 @@ test("Clustered storage pools", async ({ page }) => {
   //Clustered storage pool creation
   const poolName = "clustered-pool";
   await page.setViewportSize({ width: 1440, height: 1000 });
-  await gotoURL(page, "/ui/");
+  await gotoURL(page, "/");
```
edlerd: I think none of the test should need changes. As the feature add the possibility to run the ui with a prefixed path, the ci setup won't change, so under test the project should be unchanged.

src/App.tsx
edlerd: I see. With the `ROOT_URL` available and the tactic of removing the `ROOT_URL` prefix from the current url when figuring out the current project: I guess we don't have to change the navigation much any more. Wdyt?

src/App.tsx
edlerd: Right, I think there are arguments for both options.

src/api/server.tsx
kimanhou: THANK YOU ! I have been looking for those for 1 hour yesterday !

src/api/server.tsx
kimanhou: THANK YOU ! I have been looking for those for 1 hour yesterday !

src/util/helpers.tsx
kimanhou: I think we can use `ROOT_PATH` and `UI_BASE_PATH` in the `logout` function.

```
export const logout = (hasOidc?: boolean, hasCertificate?: boolean): void => {
  if (window.location.pathname.includes(`${UI_BASE_PATH}login`)) {
    return;
  }

  void fetch(`${ROOT_PATH}/oidc/logout`.replace(/\/+/g, "/")).then(() => {
    if (hasOidc) {
      window.location.href = `${UI_BASE_PATH}login/`;
      return;
    }

    if (hasCertificate) {
      window.location.href = `${UI_BASE_PATH}login/certificate-add`;
      return;
    }

    window.location.href = `${UI_BASE_PATH}login/certificate-generate`;
  });
};
```

kimanhou: Or did you not want to have the `UI_BASE_PATH` variable exist at all ?

src/App.tsx
```diff
@@ -140,7 +140,7 @@ const App: FC = () => {
       const logoutAction = [
         {
           label: "Logout",
-          onClick: () => (window.location.href = "/oidc/logout"),
+          onClick: () => (window.location.href = `${ROOT_PATH}/ui/oidc/logout`),
```
```suggestion
          onClick: () => (window.location.href = `${ROOT_PATH}/oidc/logout`),
```
Those paths with `/oidc/*` go to the LXD api directly.

src/App.tsx
```diff
@@ -177,378 +180,373 @@ const App: FC = () => {
...
-          path="/ui/project/:project/network/:network/member/:memberName/forwards/:forwardAddress/edit"
-          element={
-            <ProtectedRoute
-              outlet={<ProjectLoader outlet={<EditNetworkForward />} />}
-            />
-          }
-        />
-        <Route
-          path="/ui/project/:project/network-acls"
+          path={`${ROOT_PATH}/ui/project/:project/network-acls`}
```
edlerd, line 357: This route got lost.

vite.config.ts
edlerd, line 1: Do we need the changes to this file? I think it can stay as it was. Wdyt?

src/util/helpers.tsx
edlerd: Yeah, I think we don't need the `UI_BASE_PATH` and can simplify on the single `ROOT_PATH` variable.

src/App.tsx
edlerd: This is still open. There shouldn't be a `/ui` here.

src/pages/cluster/ClusterGroupList.tsx
```diff
@@ -78,13 +79,15 @@ const ClusterGroupList: FC = () => {
           content:
             usedBy.length > 0
               ? usedBy.map((resource) => {
-                  if (resource.startsWith("/1.0/projects/")) {
-                    const project = resource.replace("/1.0/projects/", "");
+                  const projectMatch = resource.match(/\/1.0\/projects\/(.+)$/);
+
+                  if (projectMatch) {
+                    const project = projectMatch[1];
```
edlerd: This can probably stay as it was.

src/pages/instances/InstanceDetailHeader.tsx
```diff
@@ -140,8 +141,8 @@ const InstanceDetailHeader: FC<Props> = ({
           <Link
             to={
               canViewProject
-                ? `/ui/project/${encodeURIComponent(project)}/instances`
-                : "/ui/all-projects/instances"
+                ? `${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/instances`
+                : "/all-projects/instances"
```
edlerd: The all projects path seems broken.

src/pages/instances/InstanceGraphicConsole.tsx
```diff
@@ -76,7 +76,11 @@ const InstanceGraphicConsole: FC<Props> = ({
...
 
-    const operationUrl = result.operation.split("?")[0];
+    let operationUrl = result.operation.split("?")[0];
+    if (!operationUrl.startsWith(ROOT_PATH)) {
+      operationUrl = `${ROOT_PATH}/${operationUrl}`;
+    }
+
     const protocol = location.protocol === "https:" ? "wss" : "ws";
     const dataUrl = `${protocol}://${location.host}${operationUrl}/websocket?secret=${result.metadata.metadata.fds["0"]}`;
     const controlUrl = `${protocol}://${location.host}${operationUrl}/websocket?secret=${result.metadata.metadata.fds.control}`;
```
edlerd: We don't need the conditional here, the path comes from the api without root path. Can we inject the root path in lines 85/86 instead?

src/pages/instances/InstanceTerminal.tsx
```diff
@@ -127,7 +127,12 @@ const InstanceTerminal: FC<Props> = ({ instance, refreshInstance }) => {
...
-    const operationUrl = result.operation.split("?")[0];
+    let operationUrl = result.operation.split("?")[0];
+
+    if (!operationUrl.startsWith(ROOT_PATH)) {
+      operationUrl = `${ROOT_PATH}/${operationUrl}`;
+    }
+
     const protocol = location.protocol === "https:" ? "wss" : "ws";
     const dataUrl = `${protocol}://${location.host}${operationUrl}/websocket?secret=${result.metadata.metadata.fds["0"]}`;
     const controlUrl = `${protocol}://${location.host}${operationUrl}/websocket?secret=${result.metadata.metadata.fds.control}`;
```
edlerd: Same as the graphic console file. Inject the root path in lines 137/138 directly without checking.

src/pages/instances/InstanceTextConsole.tsx
```diff
@@ -87,7 +87,12 @@ const InstanceTextConsole: FC<Props> = ({
...
-    const operationUrl = result.operation.split("?")[0];
+    let operationUrl = result.operation.split("?")[0];
+
+    if (!operationUrl.startsWith(ROOT_PATH)) {
+      operationUrl = `${ROOT_PATH}${operationUrl}`;
+    }
+
     const protocol = location.protocol === "https:" ? "wss" : "ws";
     const dataUrl = `${protocol}://${location.host}${operationUrl}/websocket?secret=${result.metadata.metadata.fds["0"]}`;
     const controlUrl = `${protocol}://${location.host}${operationUrl}/websocket?secret=${result.metadata.metadata.fds.control}`;
```
edlerd: Same as the graphic console file.

src/pages/login/Login.tsx
```diff
@@ -29,7 +30,10 @@ const Login: FC = () => {
             <p className="u-sv1">Choose your login method</p>
             <div className="auth-container">
               {hasOidc && (
-                <a className="p-button--positive has-icon" href="/oidc/login">
+                <a
+                  className="p-button--positive has-icon"
+                  href={`${ROOT_PATH}/ui/oidc/login`}
```
edlerd:
```suggestion
                  href={`${ROOT_PATH}/oidc/login`}
```
The OIDC endpoints are handled by the Golang application. No ui prefix needed.

src/context/useCurrentProject.tsx
```diff
@@ -29,21 +30,35 @@ interface ProviderProps {
...
+      setLastProject(project);
+    }
+  }, [project, lastProject]);
+
+  const activeProject = isAllProjects ? "" : project || lastProject;
+  const enabled = activeProject.length > 0;
+  const { data, isLoading: isProjectLoading } = useProject(
+    activeProject,
+    enabled,
+  );
```
edlerd: I think the change here should be much smaller, only something like this:
```
  const location = useLocation();
  const url = location.pathname.replace(ROOT_PATH, "");
```

src/context/useCurrentProject.tsx
```diff
@@ -29,21 +30,35 @@ interface ProviderProps {
...
 
   return (
     <ProjectContext.Provider
       value={{
         canViewProject:
           isProjectLoading ||
           isSettingsLoading ||
-          project === "" ||
+          activeProject === "" ||
```
edlerd: Probably shouldn't rename this?

src/util/usedBy.tsx
edlerd, line 1: This file is operating on urls from the api, so no root path needed in here.

src/pages/projects/ProjectLoader.tsx
```diff
@@ -14,9 +15,7 @@ const ProjectLoader = ({ outlet }: Props) => {
   }
 
   if (!project) {
-    const url = location.pathname;
-    const hasProjectInUrl = url.startsWith("/ui/project/");
-    const project = hasProjectInUrl ? url.split("/")[3] : "default";
+    const { project = "default" } = useParams<{ project: string }>();
```
edlerd: We should keep this as is but add URL cleaning. Like `const url = location.pathname.replace(ROOT_URL, "");`

src/pages/projects/NavigationProjectSelector.tsx
```diff
@@ -8,7 +8,7 @@ import {
 } from "@canonical/react-components";
 import { useNavigate } from "react-router-dom";
 import NavigationProjectSelectorList from "pages/projects/NavigationProjectSelectorList";
-import { defaultFirst } from "util/helpers";
+import { defaultFirst, ROOT_PATH } from "util/helpers";
```
edlerd: This and few other files use an invalid import.

src/pages/networks/NetworkTopology.tsx
```diff
@@ -49,7 +50,7 @@ const NetworkTopology: FC<Props> = ({ formik, project, isServerClustered }) => {
       downStreamCandidate.config.network === formik.values.name ||
       downStreamCandidate.config.parent === formik.values.name ||
       network.used_by?.includes(
-        `/1.0/networks/${encodeURIComponent(downStreamCandidate.name)}`,
+        `${ROOT_PATH}/1.0/networks/${encodeURIComponent(downStreamCandidate.name)}`,
```
edlerd: this can stay as it was.

index.html
```diff
@@ -1,20 +1,45 @@
...
+        const uiIndex = cleanPath.indexOf('/ui/');
+        let base = '';
+
+        if (uiIndex >= 0) {
+          base = cleanPath.substring(0, uiIndex + 4);
+        } else {
+          base = cleanPath + 'ui/';
+        }
+
+        window.LXD_UI_BASE_PATH = base;
```
edlerd: I think we can simplify the logic here and in util/rootPath to directly calculate the value of ROOT_PATH here and store it in window.ROOT_PATH. Then the util will simply return window.ROOT_PATH.

index.html
```diff
@@ -1,20 +1,45 @@
...
+
+        if (!cleanPath.startsWith(base)) {
+          window.location.href = base;
+          return;
+        }
+
+        document.write('<base href="' + base + '" />');
```
edlerd: Does this trigger a reload of the link, scripts and module imports below? Seems like a nicer solution then.

index.html
```diff
@@ -1,20 +1,45 @@
...
   </head>
   <body>
 
     <div id="app"></div>
-    <script type="module" src="/src/index.tsx"></script>
+    <script type="module" src="src/index.tsx"></script>
 
   </body>
-</html>
+</html>
```
edlerd:
```suggestion
</html>

```

vite.config.ts
kimanhou: Mmh if I undo the changes in `vite.config.ts` then the application doesn't run on http://localhost:8444/foo/bar/ui/project/default/instances anymore

src/util/usedBy.tsx
```diff
@@ -57,7 +58,7 @@ export const filterUsedByType = (
           return path.includes("/forwards/");
         }
 
-        return path.startsWith(`/1.0/${type}`);
+        return path.startsWith(`${ROOT_PATH}/1.0/${type}`);
```
edlerd: This should be reverted as it relates to api generated urls.

vite.config.ts
```diff
@@ -46,9 +51,4 @@ export default defineConfig({
     outDir: "./build/ui",
     minify: "esbuild",
   },
-  experimental: {
-    renderBuiltUrl(filename: string) {
-      return "/ui/" + filename;
-    },
-  },
```
edlerd, line 53: This needs to stay for the production build.

src/util/projects.tsx
```diff
@@ -21,10 +22,13 @@ export const projectSubpages = [
 ];
 
 export const getSubpageFromUrl = (url: string): string | undefined => {
-  const parts = url.split("/");
+  const urlWithoutQuery = url.split("?")[0];
+  const normalizedPath = urlWithoutQuery.replace(`${ROOT_PATH}/ui/`, "");
```
edlerd:
```suggestion
  const normalizedPath = urlWithoutQuery.replace(ROOT_PATH, "");
```
Maybe a bit simpler, and the indexes below don't need to change with this.

src/context/useCurrentProject.tsx
```diff
@@ -29,21 +30,35 @@ interface ProviderProps {
...
   const { isLoading: isSettingsLoading } = useSettings();
-  const location = useLocation();
-  const url = location.pathname;
-  const project = url.startsWith("/ui/project/") ? url.split("/")[3] : "";
-  const isAllProjects = url.startsWith("/ui/all-projects/");
+  const { pathname } = useLocation();
 
-  const enabled = project.length > 0 && !isAllProjects;
-  const { data, isLoading: isProjectLoading } = useProject(project, enabled);
+  const normalizedPath = pathname.replace(`${ROOT_PATH}/ui/`, "");
```
edlerd:
```suggestion
  const normalizedPath = pathname.replace(ROOT_PATH, "");
```
This seems a bit lighter and allows the rest of the code below to stay unchanged.

src/components/Navigation.tsx
```diff
@@ -64,43 +64,21 @@ const initialiseOpenNavMenus = (location: Location) => {
...
-    isLoading,
-  } = useCurrentProject();
+  const { project, isAllProjects } = useCurrentProject();
+
+  const projectName = isAllProjects ? ALL_PROJECTS : project?.name || "default";
   const isSmallScreen = useIsScreenBelow();
-  const [projectName, setProjectName] = useState(
-    initializeProjectName(isAllProjectsFromUrl, isLoading, project),
-  );
-  const isAllProjects = projectName === ALL_PROJECTS;
```
edlerd, line 96: I didn't test it yet, but suspect this is breaking the selection of the current project when a user lands on the page. As in be on a custom project or the all projects page and refresh the page.

src/components/Navigation.tsx
```diff
@@ -109,19 +87,6 @@ const Navigation: FC = () => {
...
-      setProjectName(ALL_PROJECTS);
-      setOpenNavMenus([]);
-      return;
-    }
-
-    if (project && project.name !== projectName) {
-      setProjectName(project.name);
-    }
-  }, [project?.name, isAllProjectsFromUrl, projectName]);
-
```
edlerd, line 124: This is handling an edge case where a user has no view access on any project, but they have access to a single or multiple instances directly. In that case we land them on the all projects page, as the default project instance list won't be readable for them.

src/context/useCurrentProject.tsx
kimanhou:
```
const { pathname } = useLocation();
const url = pathname.replace(`${ROOT_PATH}/ui/`, "");
```

src/context/useCurrentProject.tsx
```diff
@@ -29,21 +30,35 @@ interface ProviderProps {
...
+  const parts = normalizedPath.split("/");
+  const project = parts[0] === "project" ? parts[1] : "";
+  const isAllProjects = parts[0] === "all-projects";
+
+  const [lastProject, setLastProject] = useState("default");
```
edlerd: We can't fall back to default here as a user might only have access to some other project, but not the default one. Or no project at all.

I think it is best to keep most of this logic as is, not to break any of those edge cases.

src/context/useCurrentProject.tsx
edlerd: I wouldn't remove the `/ui` to keep the current behavior. It avoids having to adjust any existing logic. Which is always a bit of a risk to break things.

index.html
kimanhou: Yes I think 🤔

src/context/useCurrentProject.tsx
kimanhou: You don't seem to have the latest version 🤔

src/vite-env.d.ts
edlerd, line 1: Can we remove this file?

src/components/Navigation.tsx
```diff
@@ -580,32 +581,32 @@ const Navigation: FC = () => {
...
+                                  to={`${ROOT_PATH}/ui/permissions/groups`}
                                   title="Groups"
                                   onClick={softToggleMenu}
                                   className="accordion-nav-secondary"
                                 >
                                   Groups
                                 </NavLink>
                               </SideNavigationItem>,
-                              <SideNavigationItem key="/ui/permissions/idp-groups">
+                              <SideNavigationItem key="/permissions/idp-groups">
```
edlerd: We should keep the ui prefix here and in some more occurances above.

vite.config.ts
edlerd: I see a couple of issues with the changes:
1. When opening https://localhost:8407 in a normal dev server, we are missing the redirect to /ui
2. Loading the monaco editor in a normal dev server fails. Such as with a normal `dotrun` dev server without a prefix, on a YAML configuration page for an instance.
3. When using haproxy with a prefix, the index.js and index.scss are loaded from the /ui path and fail to load. But this might need changes in the index.html file in the project root.

edlerd: Point 3. can be resolved by removing from here the `experimental` bits, that you initially removed, then on my request restored. I think you were right to remove it considering the other changes.

index.html
```diff
@@ -4,17 +4,31 @@
...
+    <link rel="manifest" href="manifest.json" id="manifest-link">
+    <script src="assets/js/manifest.js" id="manifest-script"></script>
+    
     <script>const global = globalThis;</script>
   </head>
   <body>
 
     <div id="app"></div>
-    <script type="module" src="/src/index.tsx"></script>
+    <script type="module" src="src/index.tsx"></script>
```
edlerd: This gets turned into a relative link to the js and css file on built. That is okay with the base href that is set dynamically above.

However, on first load, the base href is not set, leading to two errors from both files failing to load. Let's prevent those errors somehow. This will also be a problem in standard environment without a non-prefixed root. As soon as we load a nested path like /ui/project/default/instance/foo we will face the same errors due to the initially missing base.

vite.config.ts
kimanhou: 1. I added the following in index.html:
```
if (path === "/" || path === "") {
    window.location.replace("/ui/");
    return;
}
```

2. I changed vite.config.ts to this:
```
"/ui/monaco-editor": {
  target: `http://localhost:${process.env.VITE_PORT || 3000}`,
  rewrite: (path) =>
    path.replace(/^\/ui\/monaco-editor/, "/node_modules/monaco-editor"),
  secure: false,
},
```
It works but I get this error when running locally:
```
Failed to load source map for /home/ubuntu/lxd-ui/node_modules/monaco-editor/min/vs/loader.js.
Error: An error occurred while trying to read the map file at ../../min-maps/vs/loader.js.map
Error: ENOENT: no such file or directory, open '/home/ubuntu/lxd-ui/node_modules/monaco-editor/min-maps/vs/loader.js.map'
    at async open (node:internal/fs/promises:641:25)
    at async Object.readFile (node:internal/fs/promises:1245:14)
    at async extractSourcemapFromFile (file:///home/ubuntu/lxd-ui/node_modules/vite/dist/node/chunks/config.js:8050:65)
    at async loadAndTransform (file:///home/ubuntu/lxd-ui/node_modules/vite/dist/node/chunks/config.js:22640:22)
```

3. I'm stuck on this, I can't find a solution. Do you have any idea ?

src/vite-env.d.ts
kimanhou: I thought TypeScript needed it but I removed it and nothing breaks so I guess we can remove it.

vite.config.ts
```diff
@@ -30,14 +31,19 @@ export default defineConfig({
...
-        rewrite: (path) => path.replace(/^\/ui/, ""),
+      "^/ui/(assets|manifest.json)": {
+        target: "https://localhost:8443/",
+        secure: false,
+        changeOrigin: true,
+      },
+      "/1.0": {
+        target: "https://localhost:8443/",
         secure: false,
       },
```
edlerd: The `/1.0` should not reach the dev server, but be caught by the [haproxy-dev](https://github.com/canonical/lxd-ui/blob/main/haproxy-dev.cfg#L15) that sits in front. I think we can remove this bit.

vite.config.ts
```diff
@@ -30,14 +31,19 @@ export default defineConfig({
...
     hmr: process.env.CI ? false : undefined,
     proxy: {
-      "/ui/assets": {
-        target: "https://localhost:8407/",
-        rewrite: (path) => path.replace(/^\/ui/, ""),
+      "^/ui/(assets|manifest.json)": {
+        target: "https://localhost:8443/",
+        secure: false,
+        changeOrigin: true,
+      },
```
edlerd, line 39: We should not serve assets from 8443 which is the snap install of lxd. Assets should come from the local checkouts public directory. Otherwise, newly added or modified assets will not be available during development.

I suspect we can load those similarly as the `target: http://localhost:${process.env.VITE_PORT || 3000},` below for the monaco editor with a fitting rewrite path.

src/util/rootPath.ts
```diff
@@ -0,0 +1,7 @@
+declare global {
+  interface Window {
+    ROOT_PATH: string;
+  }
+}
+
+export const ROOT_PATH = window.ROOT_PATH || "";
```
edlerd:
```suggestion
export const ROOT_PATH = window?.ROOT_PATH || "";
```
This should fix the tests.

vite.config.ts
edlerd: I see the error in the log about the Monaco source maps. In the main branch dev server, I see the same error, so probably not an issue to resolve here. Also, I think we don't need the source maps, so it's most likely ok to keep as is.

index.html
edlerd: This is a remaining problem.

index.html
edlerd, line 1: There is an issue with the lxd default redirect in Golang. When accessing http://localhost:8444/foo/bar Golang redirects to http://localhost:8444/ui instead of http://localhost:8444/foo/bar/ui/

The same happens when accessing http://localhost:8444/foo/bar/ui

So a user will have to enter http://localhost:8444/foo/bar/ui/ to reach the ui.

I don't think there is a nice solution here because Golang does not even know the prefix is in use. The redirect proxy config should probably be extended to catch those incomplete URLs and do the redirect. I wonder if we should add to the official docs to give guidelines on this.

vite.config.ts
edlerd: I think 3. is resolved now.

vite.config.ts
kimanhou: Like this ?

```
proxy: {
  "^/ui/(assets|manifest.json)": {
    target: `http://localhost:${process.env.VITE_PORT || 3000}`,
    rewrite: (path) => path.replace(/^\/ui/, ""),
    secure: false,
    changeOrigin: true,
  },
  "/ui/monaco-editor": {
    target: `http://localhost:${process.env.VITE_PORT || 3000}`,
    rewrite: (path) =>
      path.replace(/^\/ui\/monaco-editor/, "/node_modules/monaco-editor"),
    secure: false,
  },
},
 ```

File: index.html (hunk 47 above)
Comment (kimanhou) [reply index.html]: Yes, I can't find a solution for now. Still working on this

File: index.html
Comment (kimanhou) [reply index.html:1]: I let Copilot create a separate script to fix the generated paths after building.

File: package.json [hunk 51]
```diff
@@ -113,4 +114,4 @@
     "path-to-regexp": "8.3.0",
     "wrap-ansi": "9.0.2"
   }
-}
+}
```
edlerd:
```suggestion
}

```

index.html
```diff
@@ -1,20 +1,38 @@
...
+    <script type="module" src="/src/index.tsx"></script>
   </head>
   <body>
-
     <div id="app"></div>
-    <script type="module" src="/src/index.tsx"></script>
-
   </body>
-</html>
+</html>
```
edlerd:
```suggestion
</html>

```

fix-asset-paths.js
edlerd, line 1: maybe keep this in the root folder?

fix-asset-paths.js
```diff
@@ -0,0 +1,60 @@
...
+      const uiIndex = path.indexOf("/ui/");
+      const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
+      const link = document.createElement('link');
+      link.rel = 'stylesheet';
+      link.crossOrigin = 'anonymous';
+      link.href = rootPath + '/ui/assets/${filename}';
+      document.head.appendChild(link);
+    })();
+  </script>`;
+});
```
edlerd: This fixes the issue, but the code is very repetitive with what we already have in index.html. Also the output html is very repetitive:

```
<!doctype html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <title>LXD UI</title>
    <script>
      (function() {
        const path = window.location.pathname;
        
        if (path.endsWith('/ui')) {
            window.location.replace(path + '/');
            return;
        }
        if (path === "/" || path === "") {
            window.location.replace("/ui/");
            return;
        }

        const uiIndex = path.indexOf("/ui/");
        const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
        window.ROOT_PATH = rootPath;
        
        const basePath = rootPath + "/ui/";
        const baseTag = document.createElement('base');
        baseTag.href = window.location.origin + basePath;
        document.head.appendChild(baseTag);
      })();
    </script>
    
    <link rel="shortcut icon" href="assets/img/favicon-32x32.png" id="favicon">
    <script>const global = globalThis;</script>
    <script>
    (function() {
      const path = window.location.pathname;
      const uiIndex = path.indexOf("/ui/");
      const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
      const script = document.createElement('script');
      script.type = 'module';
      script.crossOrigin = 'anonymous';
      script.src = rootPath + '/ui/assets/index-BeHSjJeC.js';
      document.head.appendChild(script);
    })();
  </script>
    <script>
    (function() {
      const path = window.location.pathname;
      const uiIndex = path.indexOf("/ui/");
      const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
      const link = document.createElement('link');
      link.rel = 'stylesheet';
      link.crossOrigin = 'anonymous';
      link.href = rootPath + '/ui/assets/index-BfdIcvH7.css';
      document.head.appendChild(link);
    })();
  </script>
  </head>
  <body>
    <div id="app"></div>
  </body>
</html>
```

scripts/fix-asset-paths.js
```diff
@@ -0,0 +1,60 @@
...
+// Replace link tags
+html = html.replace(linkPattern, (match, filename) => {
+  return `<script>
+    (function() {
+      const path = window.location.pathname;
+      const uiIndex = path.indexOf("/ui/");
+      const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
+      const link = document.createElement('link');
+      link.rel = 'stylesheet';
+      link.crossOrigin = 'anonymous';
```
edlerd: Current values is empty string: `<link rel="stylesheet" crossorigin="" href="/ui/assets/index-B0v_Rfjz.css">`

scripts/fix-asset-paths.js
```diff
@@ -0,0 +1,60 @@
...
+// Replace script tags
+html = html.replace(scriptPattern, (match, filename) => {
+  return `<script>
+    (function() {
+      const path = window.location.pathname;
+      const uiIndex = path.indexOf("/ui/");
+      const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
+      const script = document.createElement('script');
+      script.type = 'module';
+      script.crossOrigin = 'anonymous';
```
edlerd:
```suggestion
      script.crossOrigin = '';
```

src/App.tsx
edlerd, line 1: One open issue is when I browse to http://localhost:8444/foo/bar/ui/ then I get page not found. Though when I click through the navigation and to different projects it all works. So I think the initial redirect is broken. I think we need to adjust the `HOME_REDIRECT_PATHS` in here to be prefixed with the `ROOT_PATH` to fix this.

image

scripts/fix-asset-paths.js
kimanhou: Current values of `crossorigin` ?

edlerd: Yes, here setting it to anonymous, but currently the value is empty string.

kimanhou: I believe it's fixed now. Can you confirm ?

index.html
```diff
@@ -1,20 +1,38 @@
...
+        const path = window.location.pathname;
+        
+        if (path.endsWith('/ui')) {
+            window.location.replace(path + '/');
+            return;
+        }
+        if (path === "/" || path === "") {
+            window.location.replace("/ui/");
+            return;
+        }
```
edlerd:
```suggestion
```
I assume this is only needed for the dev server, but putting it here it ends up also in the production build.

Alternatively we can do a redirect in the haproxy-dev.cfg file and match the paths just like here. Then this workaround won't leak into the production build.

haproxy-dev.cfg
```diff
@@ -9,6 +9,16 @@ defaults
...
   bind 0.0.0.0:8407 ssl verify optional crt keys/lxd-ui.pem ca-file keys/lxd-ui.crt
+
+  acl needs_trailing_slash path_reg ^.*/ui$
+  http-request redirect code 301 location %[path]/ if needs_trailing_slash
+
+  acl is_root path /
+  acl is_not_app path_beg /1.0 /oidc /documentation
+  acl is_not_ui   path_reg ^.*/ui/.*
+  http-request redirect code 301 location /ui/ if is_root
+  http-request redirect code 301 location %[path]/ui/ if !is_root !is_not_app !is_not_ui !needs_trailing_slash
```
edlerd: This doesn't work yet. With `dotrun` I get an empty page, and some paths like `https://localhost:8407/@react-refresh` get redirected to the invalid `https://localhost:8407/@react-refresh/ui/` path.

fix-asset-paths.js
edlerd, line 1: This is much better than before, but producing a bit of duplication. Can we turn this:
```
<!doctype html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <title>LXD UI</title>
    <script>
      (function() {
        const path = window.location.pathname;

        const uiIndex = path.indexOf("/ui/");
        const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
        window.ROOT_PATH = rootPath;
        
        const basePath = rootPath + "/ui/";
        const baseTag = document.createElement('base');
        baseTag.href = window.location.origin + basePath;
        document.head.appendChild(baseTag);
      })();
    </script>
    
    <link rel="shortcut icon" href="assets/img/favicon-32x32.png" id="favicon">
    <script>const global = globalThis;</script>
    
    
  <script>
    (function() {
      
      const script_index_B0HOiwNF_js = document.createElement('script');
      script_index_B0HOiwNF_js.type = 'module';
      script_index_B0HOiwNF_js.crossOrigin = '';
      script_index_B0HOiwNF_js.src = window.ROOT_PATH + '/ui/assets/index-B0HOiwNF.js';
      document.head.appendChild(script_index_B0HOiwNF_js);
      
      const link_index_C5oBt4SZ_css = document.createElement('link');
      link_index_C5oBt4SZ_css.rel = 'stylesheet';
      link_index_C5oBt4SZ_css.crossOrigin = '';
      link_index_C5oBt4SZ_css.href = window.ROOT_PATH + '/ui/assets/index-C5oBt4SZ.css';
      document.head.appendChild(link_index_C5oBt4SZ_css);
    })();
  </script>
  </head>
  <body>
    <div id="app"></div>
  </body>
</html>
```
into the slightly condensed that:
```
<!doctype html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <title>LXD UI</title>
    <script>
      (function() {
        const path = window.location.pathname;

        const uiIndex = path.indexOf("/ui/");
        const rootPath = uiIndex > -1 ? path.substring(0, uiIndex) : "";
        window.ROOT_PATH = rootPath;
        
        const basePath = rootPath + "/ui/";
        const baseTag = document.createElement('base');
        baseTag.href = window.location.origin + basePath;
        document.head.appendChild(baseTag);
      
        const script_index_B0HOiwNF_js = document.createElement('script');
        script_index_B0HOiwNF_js.type = 'module';
        script_index_B0HOiwNF_js.crossOrigin = '';
        script_index_B0HOiwNF_js.src = window.ROOT_PATH + '/ui/assets/index-B0HOiwNF.js';
        document.head.appendChild(script_index_B0HOiwNF_js);
        
        const link_index_C5oBt4SZ_css = document.createElement('link');
        link_index_C5oBt4SZ_css.rel = 'stylesheet';
        link_index_C5oBt4SZ_css.crossOrigin = '';
        link_index_C5oBt4SZ_css.href = window.ROOT_PATH + '/ui/assets/index-C5oBt4SZ.css';
        document.head.appendChild(link_index_C5oBt4SZ_css);
      })();
    </script>
    <link rel="shortcut icon" href="assets/img/favicon-32x32.png" id="favicon">
    <script>const global = globalThis;</script>
  </head>
  <body>
    <div id="app"></div>
  </body>
</html>
```

Wdyt?

PR #1714, opened 2026-01-15
fix(ConfirmationCheckbox)

kimanhou:
## Done

- Fix ConfirmationCheckbox

## QA

1. Run the LXD-UI:
    - Make sure the checkbox works in: `DeleteInstanceBtn`, `InstanceBulkActions` (restart and stop), `RestartInstanceBtn`, `StopInstanceBtn`, `InstanceSnapshotActions` (restore), `CreateIdentityModal`
    - in `CreateIdentityModal`, the checkbox also should get checked when you click on the copy token button.

## Screenshots
image

### DIFF
--- a/src/components/ConfirmationCheckbox.tsx
+++ b/src/components/ConfirmationCheckbox.tsx
@@ -12,11 +12,12 @@ const ConfirmationCheckbox: FC<Props> = ({ label, confirmed }) => {
   return (
     <span className="u-float-left">
       <CheckboxInput
+        key={`confirmation-checkbox-${isConfirmed}`}
         inline
         label={label}
         tabIndex={-1}
-        checked={isConfirmed}
-        onClick={() => {
+        defaultChecked={isConfirmed}
+        onChange={() => {
           setConfirmed((prev) => !prev);
         }}
       />


src/components/ConfirmationCheckbox.tsx
```diff
@@ -15,7 +15,7 @@ const ConfirmationCheckbox: FC<Props> = ({ label, confirmed }) => {
         inline
         label={label}
         tabIndex={-1}
-        checked={isConfirmed}
+        defaultChecked={isConfirmed}
```
edlerd, line 19: This fixes all the instance related usages. But the identity creation has an edge case with copying the code. It leads to the checkbox and the internal state being out of sync.


https://github.com/user-attachments/assets/bff1fd3e-0ae7-42b1-a448-158decf3c10b

kimanhou: Aaaah now I remember why I changed it in the first place. I can't find a better solution than having a double state: one internal and one external to the component. Which I think you asked me to simplify in the last PR 🤔

edlerd: Shadow state is not a good pattern, I think the `checked` instead of `defaultChecked` seems like the right solution.

So maybe we can fix the instance usages so they rerender correctly? I suspect they don't rerender because of the combined `  confirmed: [boolean, Dispatch<SetStateAction<boolean>>];` prop. Can you try splitting that into two props? Perhaps it resolves the issue there.

kimanhou: Splitting `combined` into two props didn't work. But adding a key that includes the value of  `isConfirmed` did ! The key forces re-rendering.

edlerd: Nice :+1:

src/components/ConfirmationCheckbox.tsx
```diff
@@ -15,8 +15,9 @@ const ConfirmationCheckbox: FC<Props> = ({ label, confirmed }) => {
         inline
         label={label}
         tabIndex={-1}
-        checked={isConfirmed}
-        onClick={() => {
+        defaultChecked={isConfirmed}
+        key={`confirmation-checkbox-${isConfirmed}`}
```
edlerd: Nitpick: Please, can you move the key up to be the first prop in the list?

PR #1716, opened 2026-01-16
feat(instance) highlight active instance and profile configuration sections

### LINKED ISSUE #1599: Highlight project/instance modifications
Hellooo,

Would it be a nice feature to highlight the modifications made to a project or instance? I find myself going to the YAML view every time to find out what a profile changes. There are so many pages i don't want to click all of them to find out.

for example my default project that has network and a disk configured a simple `font-weight: bold`

Image

edlerd:
## Done

- feat(instance) highlight active instance and profile configuration sections

Fixes #1599

## QA

1. Run the LXD-UI:
    - Open an instance configuration and ensure all local configuration sections in the menu with overrides are bold and those without are not bold
    - Repeat for profile configuration and both instance and profile creation.

## Screenshots

image

### DIFF
--- a/src/pages/instances/forms/InstanceFormMenu.tsx
+++ b/src/pages/instances/forms/InstanceFormMenu.tsx
@@ -4,6 +4,15 @@ import MenuItem from "components/forms/FormMenuItem";
 import { Button, useListener, useNotify } from "@canonical/react-components";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
+import type { InstanceAndProfileFormikProps } from "components/forms/instanceAndProfileFormValues";
+import { hasPrefixValue } from "util/formFields";
+import {
+  isDiskDevice,
+  isGPUDevice,
+  isNicDevice,
+  isOtherDevice,
+  isProxyDevice,
+} from "util/devices";
 
 export const MAIN_CONFIGURATION = "Main configuration";
 export const DISK_DEVICES = "Disk";
@@ -25,6 +34,7 @@ interface Props {
   setActive: (val: string) => void;
   hasDiskError: boolean;
   hasNetworkError: boolean;
+  formik: InstanceAndProfileFormikProps;
 }
 
 const InstanceFormMenu: FC<Props> = ({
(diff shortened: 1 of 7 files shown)


src/pages/instances/forms/InstanceFormMenu.tsx
```diff
@@ -82,25 +90,75 @@ const InstanceFormMenu: FC<Props> = ({
...
+                isBold={hasDeviceOfType(formik, "proxy")}
               />
-              <MenuItem label={GPU_DEVICES} {...menuItemProps} />
-              <MenuItem label={PROXY_DEVICES} {...menuItemProps} />
               {hasMetadataConfiguration && (
-                <MenuItem label={OTHER_DEVICES} {...menuItemProps} />
+                <MenuItem
+                  label={OTHER_DEVICES}
+                  {...menuItemProps}
+                  isBold={formik.values.devices.some(
```
omarelkashef: Image

The Other tab is bolded for some instances even when there are no overrides. I suspect it is because they have a device of type none. There is a `isOtherDevice` constant in utils/devices.tsx that we can use instead to mitigate this case.

src/util/formFields.tsx
```diff
@@ -41,3 +42,21 @@ export const optionRenderer = (
 export const focusField = (name: string) => {
   setTimeout(() => document.getElementById(name)?.focus(), 100);
 };
+
+export const hasPrefixValue = (
+  formik: InstanceAndProfileFormikProps,
+  keyPrefix: string,
+  ignoreKey?: string,
```
omarelkashef, line 49: Is it possible that we need to ignore more than one key?

src/pages/instances/forms/InstanceFormMenu.tsx
edlerd: Good hint! We can indeed get completely rid of the new custom `hasDeviceOfType` function and rely on the existing helpers.

src/util/formFields.tsx
edlerd: It is possible in the future. But I'd keep this simple, as today we only have need for a single exception.

edlerd:
> How about we use a dot with bolding like we do in the network device tab?

The blue dot is for pending unsaved changes. I would keep it separate from here, as we signal "this section has customization.".

I think the suggestion to use bold to signal customization is an easy and learnable way to signal it.

PR #1718, opened 2026-01-19
Add circular deps check

omarelkashef:
This PR is to stay open as to provide reference to circular dependency issues that should be fixed. Afterwards, it can merged.
## Done

-  add circular deps check to pre-commit hook and github CI

## QA

1. Run the LXD-UI:
    - Make sure circular dependancy check is run when new git commits are created and when commits are pushed to github

### DIFF
--- a/.dependency-cruiser.cjs
+++ b/.dependency-cruiser.cjs
@@ -1,0 +1,198 @@
+/** @type {import('dependency-cruiser').IConfiguration} */
+module.exports = {
+  options: { doNotFollow: { path: ["node_modules"] } },
+  forbidden: [
+    {
+      name: "no-circular",
+      severity: "error",
+      comment:
+        "This dependency is part of a circular relationship. You might want to revise " +
+        "your solution (i.e. use dependency inversion, make sure the modules have a single responsibility) ",
+      from: {},
+      to: {
+        circular: true,
+      },
+    },
+    {
+      name: "no-orphans",
+      comment:
+        "This is an orphan module - it's likely not used (anymore?). Either use it or " +
+        "remove it. If it's logical this module is an orphan (i.e. it's a config file), " +
+        "add an exception for it in your dependency-cruiser configuration. By default " +
+        "this rule does not scrutinize dot-files (e.g. .eslintrc.js), TypeScript declaration " +
+        "files (.d.ts), tsconfig.json and some of the babel and webpack configs.",
+      severity: "warn",
+      from: {
+        orphan: true,
+        pathNot: [
+          "(^|/)[.][^/]+[.](?:js|cjs|mjs|ts|cts|mts|json)$", // dot files
+          "[.]d[.]ts$", // TypeScript declaration files
+          "(^|/)tsconfig[.]json$", // TypeScript config
+          "(^|/)(?:babel|webpack)[.]config[.](?:js|cjs|mjs|ts|cts|mts|json)$", // other configs
+        ],
+      },
+      to: {},
+    },
+    {
+      name: "no-deprecated-core",
+      comment:
+        "A module depends on a node core module that has been deprecated. Find an alternative - these are " +
+        "bound to exist - node doesn't deprecate lightly.",
+      severity: "warn",
+      from: {},
+      to: {
+        dependencyTypes: ["core"],
+        path: [
+          "^v8/tools/codemap$",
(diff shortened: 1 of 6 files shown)


.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 npx lint-staged
+npm run check-circular-deps
```
edlerd, line 5: I wouldn't add it here, as it takes several seconds to complete without a max depth.

package.json
```diff
@@ -5,6 +5,7 @@
   "license": "LGPL-3.0-only",
   "type": "module",
   "scripts": {
+    "check-circular-deps": "npx depcruise src",
```
edlerd: Can you add a section about this to the contribution.md file?

package.json
```diff
@@ -70,8 +71,9 @@
     "babel-plugin-istanbul": "7.0.1",
     "concurrently": "9.2.1",
     "convert-source-map": "2.0.0",
-    "eslint": "9.39.2",
+    "dependency-cruiser": "^17.3.6",
```
edlerd, line 74: There should be an update to the yarn.lock file with this new dependency in place.

.github/workflows/pr.yaml
```diff
@@ -39,6 +38,9 @@ jobs:
 
       - name: Lint JS
         run: yarn lint-js
+      
+      -name: Check circular dependencies
+       run: yarn check-circular-deps
```
edlerd:
```suggestion
      - name: Check circular dependencies
        run: yarn check-circular-deps
```

.github/workflows/pr.yaml
```diff
@@ -174,28 +176,28 @@ jobs:
...
+      # NOTE: there is no need to install playwright dependencies since we only need to merge reports here
+      - name: Merge into HTML Report
+        run: npx playwright merge-reports --config=merge.playwright.config.ts ./blob-reports
+
+      - uses: actions/upload-artifact@v6
+        if: always()
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 14
```
edlerd: Can we do these and other formatting changes in a dedicated commit and PR?

.dependency-cruiser.cjs
```diff
@@ -0,0 +1,342 @@
...
+    // default because it's the safe option. It comes at a performance penalty, though
+    // As in practice only commonjs ('cjs') and ecmascript modules ('es6')
+    // are in wide use, you can limit the moduleSystems to those.
+    // moduleSystems: ['cjs', 'es6'],
+
+    // false: don't look at JSDoc imports (the default)
+    // true: detect dependencies in JSDoc-style import statements.
+    // Implies parser: 'tsc', which a.o. means the typescript compiler will need
+    // to be installed in the same spot you run dependency-cruiser from.
+    // detectJSDocImports: true,
```
edlerd: Shall we clean up those large blocks of commented config and explanations? Those tend to be difficult to maintain on upgrades. It seems better to rely on up to date documentation.

edlerd (commented):
Thanks for adding the check. Some comments on it are below.

.husky/pre-commit
edlerd: Actually maybe this is ok, it seems to take only 2-3 seconds.

CONTRIBUTING.md
```diff
@@ -134,6 +134,9 @@ For CSS formatting, we use [Stylelint](https://stylelint.io/). You can check you
 
     yarn lint-scss
 
+For circular dependency checks, we use [dependency-cruiser](https://www.npmjs.com/package/dependency-cruiser). You can run it with
+    yarn check-circular-deps
+
```
edlerd, line 140: This doesn't render nicely in markdown. I think missing a newline to enable formatting.
```suggestion
For circular dependency checks, we use [dependency-cruiser](https://www.npmjs.com/package/dependency-cruiser). You can run it with

    yarn check-circular-deps

```
image

package.json
edlerd: Reconsidering the previous discussion. As this is changing yarn.lock, I think we should merge it rather soon to avoid conflicts in that file which is very likely to update in between.

Let's put back the max depth of 1 here and merge it sooner.

omarelkashef: Good point!

PR #1733, opened 2026-01-23
test(local peering): add e2e tests

kimanhou:
---

### Pending issue:
I don't know how to create the other part of a peering in case of manual entry. Please help

---

## Done

- Add e2e tests to local peering:
    - Mutual peering
    - Non mutual peering
    - Manual entry

## QA

1. Run the LXD-UI:
    - Tests should pass locally and on CI
    - To run the tests locally: `npx playwright test tests/networks-clustered.spec.ts --project="chromium:lxd-latest-edge:clustered" --no-deps`

## Screenshots

image

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -25,7 +25,7 @@ jobs:
       uses: tim-actions/dco@master
       with:
         commits: ${{ steps.get-pr-commits.outputs.commits }}
-        
+
   lint:
     runs-on: ubuntu-latest
     steps:
@@ -196,11 +196,11 @@ jobs:
         path: blob-reports
         pattern: blob-report-*
         merge-multiple: true
-    
+
     # NOTE: there is no need to install playwright dependencies since we only need to merge reports here
     - name: Merge into HTML Report
       run: npx playwright merge-reports --config=merge.playwright.config.ts ./blob-reports
-    
+
     - uses: actions/upload-artifact@v6
       if: always()
       with:
--- a/tests/helpers/network.ts
+++ b/tests/helpers/network.ts
@@ -4,18 +4,33 @@ import { activateAllTableOverrides } from "./configuration";
 import { gotoURL } from "./navigate";
 import { expect } from "../fixtures/lxd-test";
 import { isServerClustered } from "./cluster";
+import { execSync } from "child_process";
 
 export const randomNetworkName = (): string => {
   return `test-${randomNameSuffix()}`;
+};
+
+export const ensureOvnNorthboundConnection = () => {
+  try {
+    execSync(
+      'sudo lxc config set network.ovn.northbound_connection "ssl:127.0.0.1:6641"',
+    );
+  } catch (error) {
+    console.log("Failed to set OVN northbound connection:", error);
+  }
 };
 
 export const createNetwork = async (
   page: Page,
   network: string,
   type = "bridge",
-  hasMemberSpecificParents?: boolean,
-  networkACL?: string,
+  options: {
+    hasMemberSpecificParents?: boolean;
+    networkACL?: string;
+    uplink?: string;
+  } = {},
 ) => {
+  const { hasMemberSpecificParents, networkACL, uplink } = options;
   await gotoURL(page, "/ui/");
   await page.getByRole("button", { name: "Networking" }).click();
   await page.getByRole("link", { name: "Networks", exact: true }).click();
@@ -33,7 +48,11 @@ export const createNetwork = async (
     await page.getByLabel("Parent").selectOption({ index: 1 });
   }
   if (type === "ovn") {
-    await page.getByLabel("Uplink").selectOption({ index: 1 });
+    if (uplink) {
+      await page.getByLabel("Uplink").selectOption(uplink);
+    } else {
+      await page.getByLabel("Uplink").selectOption({ index: 1 });
+    }
   }
   if (networkACL) {
     await page.getByLabel("Select ACLs").click();
@@ -44,7 +63,19 @@ export const createNetwork = async (
 
 export const deleteNetwork = async (page: Page, network: string) => {
   await visitNetwork(page, network);
-  await page.getByRole("button", { name: "Delete network" }).click();
+
+  const deleteButton = page.getByRole("button", { name: "Delete network" });
+
+  // Check if the delete button is enabled
+  const isEnabled = await deleteButton.isEnabled();
+  if (!isEnabled) {
+    console.log(
+      `Network ${network} cannot be deleted - it is currently in use`,
+    );
+    return;
+  }
+
+  await deleteButton.click();
   await page
     .getByRole("dialog", { name: "Confirm delete" })
     .getByRole("button", { name: "Delete" })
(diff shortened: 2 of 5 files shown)


tests/networks-clustered.spec.ts
```diff
@@ -44,7 +53,130 @@ test.describe("Network operations in a clustered environment with a single node"
...
+  test.beforeEach(async ({ lxdVersion, page }, testInfo) => {
+    skipIfNotSupported(lxdVersion);
+    skipIfNotClustered(testInfo.project.name);
+
+    await createOvnUplink(page, UPLINK_NAME);
+  });
+
+  test.afterEach(async ({ page }) => {
+    await deleteNetwork(page, UPLINK_NAME);
+  });
```
edlerd:
```suggestion
```
We probably don't need this with the after-all hook below and the `createOvnUplink` being idempotent.

tests/networks-clustered.spec.ts
```diff
@@ -44,7 +53,130 @@ test.describe("Network operations in a clustered environment with a single node"
...
+
+  test.afterAll(async ({ page }) => {
+    await deleteNetwork(page, UPLINK_NAME);
+  });
+
+  test("Create mutual peering between two OVN networks", async ({ page }) => {
+    const networkA = randomNetworkName();
+    const networkB = randomNetworkName();
+    const peeringName = "peer-mutual";
```
edlerd:
```suggestion
    const peering = "peer-mutual";
```
In tests, we tend to treat the names as the thing.

tests/helpers/network.ts
```diff
@@ -9,13 +9,21 @@ export const randomNetworkName = (): string => {
...
+  networkACL,
+  uplink,
+}: {
+  page: Page;
+  network: string;
+  type?: string;
+  hasMemberSpecificParents?: boolean;
+  networkACL?: string;
+  uplink?: string;
+}) => {
```
edlerd: I am not a fan of the named imports. They make all the usages verbose. I understand the problem of adding too many optional parameters here. Maybe only put the last three into an `options` object that will be named so we can mix and match the last three parameters. Wdyt?

tests/networks-clustered.spec.ts
```diff
@@ -44,7 +53,130 @@ test.describe("Network operations in a clustered environment with a single node"
...
+      uplink: UPLINK_NAME,
+    });
+
+    await createNetworkLocalPeering(
+      page,
+      networkA,
+      networkB,
+      peeringName,
+      true,
+    );
```
edlerd:
```suggestion
    await createNetworkLocalPeering(
      page,
      peeringName,
      networkA,
      networkB,
      true,
    );
```
This was an oversight when we added this function earlier this week. The name should be the 2nd parameter. Can you please adjust it?

tests/networks-clustered.spec.ts
```diff
@@ -44,7 +53,130 @@ test.describe("Network operations in a clustered environment with a single node"
...
+      .click();
+
+    await expect(page.getByLabel("Name")).toBeDisabled();
+    await expect(page.getByLabel("Target project")).toBeDisabled();
+    await expect(page.getByLabel("Target network")).toBeDisabled();
+
+    await page.getByLabel("Description").fill("Updated via Playwright");
+    await page.getByRole("button", { name: "Save changes" }).click();
+
+    await expect(page.getByText("Local peering updated")).toBeVisible();
```
edlerd: Shall we include this in the first test? Otherwise, we create and delete networks A and B multiple times in a row. Also, I think we miss the cleanup of them here, but with merging the functions, this should be resolved.

tests/helpers/network.ts
```diff
@@ -211,3 +223,33 @@ export const deleteLocalPeerings = async (
...
+  await page.getByLabel("submenu").getByText("bridge").first().click();
+  await page.getByLabel("Name", { exact: true }).fill(network);
+
+  const ovnRow = page.locator('tr[name*="ipv4_ovn_ranges"]');
+  await ovnRow.getByRole("button", { name: "Create override" }).click();
+  await page.getByLabel("IPv4 OVN ranges").fill("10.42.42.100-10.42.42.150");
+
+  const dhcpRow = page.locator('tr[name*="ipv4_dhcp_ranges"]');
+  await dhcpRow.getByRole("button", { name: "Create override" }).click();
+  await page.getByLabel("IPv4 DHCP ranges").fill("10.42.42.50-10.42.42.99");
```
edlerd, line 304: These are magic numbers, and I suspect them to cause problems down the line with infrastructure changes on the CI. I think we can make this more stable by: 
1. Create the bridge with automatic ip assignment.
2. Open the bridge detail and read the automatically assigned IP range of it. Say it is 10.25.102.1/24
3. Set DHCP ranges to 10.25.102.10-10.25.102.20
4. Set OVN ranges to 10.25.102.30-10.25.102.40

Or, much simpler, assign a custom ipv4 address here on creation by setting ipv4 address to `10.42.42.1/24`.

tests/networks-clustered.spec.ts
kimanhou: It wouldn't have happened if the parameter was an objet ! 😉

kimanhou: Wait the name is not the 2nd parameter

```
export const createNetworkLocalPeering = async (
  page: Page,
  network: string,
  targetNetwork: string,
  localPeeringName: string,
  toggleMutualPeering?: boolean,
) => {
```

edlerd: Yes, the function was built with wrong argument order. Can you move name to second place so it is aligned with other helpers and not a surprise ordering?

tests/networks-clustered.spec.ts
```diff
@@ -44,7 +50,145 @@ test.describe("Network operations in a clustered environment with a single node"
...
+  });
+
+  test("Create mutual peering between two OVN networks", async ({ page }) => {
+    const networkA = randomNetworkName();
+    const networkB = randomNetworkName();
+    const peering = "peer-mutual";
+
+    await createNetwork(page, networkA, "ovn", {
+      uplink: UPLINK_NAME,
+    });
```
edlerd: I think this fails because of missing server configuration. In my local microcloud I see the `network.ovn.northbound_connection` is set to `ssl:10.2.246.10:6641,ssl:10.2.246.20:6641,ssl:10.2.246.30:6641`, which are the ip addresses of the cluster members and the ovn manage process port. I assume you might have to set this server setting to `ssl:127.0.0.1:6641` in order to allow ovn network creation. Best reach out to @simondeziel as he knows the setup ovn ci helper best and can advise what needs to be done as a prerequisite to create the ovn network here.

image

simondeziel: Indeed, `lxc config set network.ovn.northbound_connection "ssl:127.0.0.1:6641"` should make it work.

kimanhou: Thank you both, all tests pass now ! ✅

.github/workflows/pr.yaml
```diff
@@ -143,6 +143,7 @@ jobs:
           sudo lxc config set core.https_address "[::]:8443"
           sudo lxc config trust add keys/lxd-ui.crt
           sudo lxc config set cluster.https_address "127.0.0.1"
+          sudo lxc config set network.ovn.northbound_connection "ssl:127.0.0.1:6641"
```
edlerd: This will only work for PR checks. But we also need it for the `coverage.yaml` tests. And we need it for when our test suite runs in the [lxd](https://github.com/canonical/lxd/blob/467163bc424f2b91956692a33496154e8ea35408/.github/workflows/tests.yml#L1015) and [lxd-ci](https://github.com/canonical/lxd-ci/blob/main/.github/workflows/tests.yml#L145-L146) repositories.

A better approach than adding the line here, is to add it to something more central of the tests. Maybe the `createOvnUplink` function can have a side effect call to `ensureOvnNorthboundConnection` that sets this via the `execSync` helper. Wdyt?

tests/networks-clustered.spec.ts
```diff
@@ -44,7 +50,145 @@ test.describe("Network operations in a clustered environment with a single node"
...
+    ).toBeVisible();
+
+    await expect(
+      page.getByRole("row").filter({ hasText: peering }).getByText("Pending"),
+    ).toBeVisible();
+
+    await deleteLocalPeerings(page, networkA, peering);
+    await deleteNetwork(page, networkA);
+  });
+});
```
edlerd: Shall we move this into a dedicated `network-local-peering-clustered.spec.ts` file?

PR #1742, opened 2026-01-27
feat(authentication): first user access flow

kimanhou:
## Done

Suggested flow: 

1. User runs lxd init |	CLI outputs https://<ip>:8443/?token=xyz...
2. User opens link |	UI detects token in URL, saves to Auth state, redirects to Dashboard.
3. The Warning	|        Status bar appears: ⚠️ Temporary session. Expires in 23 hours 59 minutes 10 seconds.
4. User clicks "Setup permanent access" |	UI navigates to Authentication setup page.
5. Selection	 |               User chooses OIDC or TLS to "lock in" access.
6. Clean up	 |               On successful login with TLS, UI calls logout endpoint to remove now unnecessary bearer token.

## QA

1. Run the LXD-UI:
    - Backend should be latest edge. Run `lxd init --ui-temporary-access-link`
    - Click on the link generated by backend. It should take you to LXD UI. Ensure you have access to everything.
    - Note the beautiful warning at the bottom of the page with a count down.
    - Manually logout. Ensure you can log back in with same link.
    - Ensure you can not log in with an expired or invalid token, you should see an error notification with action: logout. Click on logout and make sure the bearer token cookie is removed from your browser.
    - Log back in with the bearer token. Click on `Setup permanent access` > `Setup TLS login`. If not already done, import certificate in your browser, otherwise go to step 2. Make sure the name is pre-filled and `admins` group is pre-checked. Click on `Create identity`, you should be automatically logged in with TLS (you can check Permissions > Identities). Ensure the bearer token was automatically removed.

## Screenshots

Warning in status bar at the bottom of the screen with countdown every second
image

Authentication setup screen (after clicking on `Setup permanent access`)
image

Updated Login page to match Authentication setup page
Screenshot from 2026-02-20 10-11-36

CertificateGenerate with bearer token
image

Create TLS identity screen with bearer token
image


src/components/BearerTokenWarning.tsx
```diff
@@ -0,0 +1,44 @@
...
+        Temporary session.{" "}
+        {secondsLeft && secondsLeft > 0
+          ? `Expires in ${formatSeconds(secondsLeft)}.`
+          : ""}{" "}
+        <a href="/ui/login/authentication-setup" className="u-text--link">
+          Set up permanent access
+        </a>
+      </span>
+    </div>
+  );
```
edlerd, line 28: Let's align the language here with the spec / cli. Best discuss directly with @MusicDin it might be already OK as is.

src/components/BearerTokenWarning.tsx
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useState } from "react";
+import { Icon } from "@canonical/react-components";
+import { useSettings } from "context/useSettings";
+import { formatSeconds } from "util/seconds";
+
+const BearerTokenWarning: React.FC = () => {
```
edlerd, line 8: Please ensure this is responsive all the way down to a mobile screen size.

src/pages/login/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,47 @@
...
+            <DocLink
+              className="p-button--positive has-icon"
+              docPath="/howto/oidc"
+            >
+              <Icon name="security" light />
+              <span>Set up SSO login</span>
+            </DocLink>
+            <Link
+              className="has-icon p-button"
+              to="/ui/login/certificate-generate"
```
edlerd: I think these pages will immediately redirect the user back into the logged in pages, if the user is authenticated. We should have an exception for the bearer token authentication to not redirect.

src/types/server.d.ts
```diff
@@ -1,6 +1,6 @@
 import type { LxdConfigPair } from "./config";
 
-type LXDAuthMethods = "tls" | "oidc" | "unix";
+type LXDAuthMethods = "tls" | "oidc" | "unix" | "bearer";
```
edlerd: Shall we define constants for those values and use them in all places instead of the string values?

src/context/auth.tsx
```diff
@@ -117,6 +117,7 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
         hasNoProjects: projects.length === 0 && !isProjectsLoading,
         isFineGrained: isFineGrained(),
         serverEntitlements,
+        authMethod,
```
edlerd, line 124: We now expose it twice, here and also in `useLoggedInUser` shall we unify those to one hook?

src/pages/login/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <>
+      <CustomLayout contentClassName="authentication-setup u-flex-column">
+        <Icon name="pods" className="lxd-icon" />
+        <h1 className="p-heading--4 u-sv-2">Upgrade to permanent access</h1>
+
+        <>
+          <p className="u-sv1">
+            Your current bearer token session is temporary and will expire soon.
+            Choose a permanent method below.
```
edlerd: We might want to give a recommendation or some more explanation here which option the user should use here like a sneak preview of the documentation.

src/components/BearerTokenWarning.tsx
edlerd: One option might be to only show the icon on mobile and expose details on click with a tooltip.

src/components/BearerTokenWarning.tsx
kimanhou: Under discussion with @MusicDin

src/components/BearerTokenWarning.tsx
kimanhou: Wide screen
Screenshot from 2026-01-30 10-16-15


Medium screen
image

Small screen
image

Mobile
image

Needs tweaking for particular sizes when there is an operation but that's the idea
image

src/pages/login/AuthenticationSetup.tsx
kimanhou: I added some text but not sure it looks good. Happy to hear your suggestions

src/components/BearerTokenWarning.tsx
kimanhou: Validated by @MusicDin

src/components/BearerTokenWarning.tsx
edlerd: I'd avoid the centered text. Can we use left align?

kimanhou: Wuuut it took me ages to have that perfectly centered ! 😭

kimanhou: Oh wait you're talking about the text in the main screen `Your current bearer token session blablabla...` or the warning in the footer `Temporary access. Set up permanent access` ?

edlerd: Yes, exactly. 

Maybe we can reuse the empty state layout here and don't need to invent a new layout, similar to this one, but without the main heading:

image

src/pages/login/CertificateAdd.tsx
```diff
@@ -30,7 +32,7 @@ const CertificateAdd: FC = () => {
     return <Spinner className="u-loader" text="Loading..." isMainComponent />;
   }
 
-  if (isAuthenticated) {
+  if (isAuthenticated && isPermanentAuthMethod(authMethod as AuthMethod)) {
```
edlerd: Can useAuth already return the correct type instead of generic string?
```suggestion
  if (isAuthenticated && isPermanentAuthMethod(authMethod)) {
```

src/types/server.d.ts
```diff
@@ -1,6 +1,7 @@
 import type { LxdConfigPair } from "./config";
+import { type AuthMethod } from "util/authentication";
 
-type LXDAuthMethods = "tls" | "oidc" | "unix";
+type LXDAuthMethods = AuthMethod;
```
edlerd:
```suggestion
```
I don't think we need the alias here.

src/util/authentication.ts
```diff
@@ -0,0 +1,14 @@
+// Authentication method constants
+export const AUTH_METHOD = {
+  TLS: "tls",
+  OIDC: "oidc",
+  UNIX: "unix",
+  BEARER: "bearer",
+} as const;
+
+// Type derived from the constants
+export type AuthMethod = (typeof AUTH_METHOD)[keyof typeof AUTH_METHOD];
```
edlerd:
```suggestion
export type LXDAuthMethod = (typeof AUTH_METHOD)[keyof typeof AUTH_METHOD];
```
Maybe clearer and in line with the other naming?

src/pages/login/CertificateGenerate.tsx
```diff
@@ -23,7 +25,7 @@ const CertificateGenerate: FC = () => {
     return <Spinner className="u-loader" text="Loading..." isMainComponent />;
   }
 
-  if (isAuthenticated) {
+  if (isAuthenticated && isPermanentAuthMethod(authMethod as AuthMethod)) {
```
edlerd:
```suggestion
  if (isAuthenticated && isPermanentAuthMethod(authMethod)) {
```

src/App.tsx
```diff
@@ -139,35 +142,54 @@ const App: FC = () => {
...
+    const isTokenInvalid = authError.message.includes(
+      "token signature is invalid",
+    );
+    const isTokenExpired = authError.message.includes("token is expired");
 
-      notify.failure(title, authError, null, logoutAction);
+    if (!isSetupPath && (isTokenInvalid || isTokenExpired)) {
+      return (
+        <Navigate
+          to={`/ui/login/authentication-setup?reason=${isTokenInvalid ? "invalid" : "expired"}`}
```
```suggestion
          to="/ui/login/authentication-setup"
```
We can figure out the reason in the target component. No need to encode the error in the URL. As the context changes, it is better to react to the new context than to encode it in the URL.

src/pages/login/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  return (
+    <CustomLayout contentClassName="authentication-setup u-flex-column">
+      <h1 className="p-heading--2 u-sv-1">Set up permanent access</h1>
+      {reason === "expired" && (
+        <Notification severity="caution">Your token has expired</Notification>
+      )}
+      {reason === "invalid" && (
+        <Notification severity="caution">Your token is invalid</Notification>
+      )}
```
edlerd: I think in both cases, we need to hit an endpoint that cleans up the invalid token cookie. @MusicDin what did we decide here to use as an endpoint?

src/sass/styles.scss
```diff
@@ -452,3 +453,7 @@ body {
 .mono-font {
   font-family: unquote($font-monospace);
 }
+
+.text-align--left {
+  text-align: left;
+}
```
```suggestion
```
Seems unused.

src/pages/login/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,80 @@
...
+  CustomLayout,
+  Notification,
+  Card,
+} from "@canonical/react-components";
+import { Link, Navigate } from "react-router-dom";
+import { useAuth } from "context/auth";
+import type { AuthMethod } from "util/authentication";
+import { isPermanentAuthMethod } from "util/authentication";
+
+const AuthenticationSetup: FC = () => {
```
edlerd, line 16: Let's make this responsive.

Image

src/pages/login/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,80 @@
...
+            <li>No external identity provider required</li>
+          </ul>
+          <Link
+            className="p-button auth-setup-link"
+            to="/ui/login/certificate-generate"
+          >
+            <span>Set up TLS login</span>
+          </Link>
+        </Card>
+      </div>
```
edlerd: I think this should be the same as `Login.tsx`. What we come up with here should also work and be used on the login page.

Let's extract this into a component to reuse here and over there. Or maybe we don't even need a dedicated page and can reuse the existing login page.

src/sass/_authentication_setup.scss
```diff
@@ -0,0 +1,67 @@
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
edlerd, line 13: This is a lot of custom SCSS. It would be good to reuse existing patterns here more closely.

src/pages/login/AuthenticationSetup.tsx
kimanhou: If the browser has an expired cookie, it will automatically delete it.
If the cookie is not expired but somehow the token inside it is expired, the backend answers with a 403 error `token is expired`.
If it's invalid, then the backend answers with a 403 error `token signature is invalid`.

In case the backend answers with a 403 error, maybe the UI should clean the `lxd_session` cookie, since it's useless. Then the user will have to re-generate a token or set up a permanent access.

@MusicDin am I missing something ?

edlerd: > In case the backend answers with a 403 error, maybe the UI should clean the lxd_session cookie, since it's useless. Then the user will have to re-generate a token or set up a permanent access.

The UI has no access to the HTTP-only cookie. I think we do need an endpoint to remove it from the browser. Or do we rely on it becoming out of date and being cleaned up by the browser?

src/pages/login/AuthenticationSetup.tsx
kimanhou: What happened to the icon of SSO login ?! 👀
I fixed it.

image
image

src/pages/login/AuthenticationSetup.tsx
tomponline: There is a logout endpoint @MusicDin has added

src/pages/login/AuthenticationSetup.tsx
kimanhou: Good point, we actually do not need a new component. The Login page already does what we need.
However, as it is, I don't think it's a nice experience for the user:
image

I tried adding the notification but it just doesn't work:
image

Where should I go from here: rewriting the whole Login page with my new design (it inevitably results in a lot of custom CSS) ? Or trying to enhance the existing Login page with a more conventional approach ?

edlerd: I think we can keep both pages: "login" and the "new set up permanent access" one.

I'd extract the bits here with the two cards into a component and then we can replace the two buttons on the login page with that component.

src/pages/login/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+          <Card className="auth-setup-card u-flex-column">
+            <Icon
+              name="certificate"
+              className="auth-setup-icon u-hide--small"
+            />
+            <h2 className="p-heading--4">TLS</h2>
+
+            <ul>
+              <li>Quick setup with a simple browser certificate</li>
```
edlerd:
```suggestion
              <li>Quick setup with a browser certificate</li>
```
The word "simple" should be avoided in documentation. It can be understood as condescending by the user.

src/pages/login/AuthenticationSetup.tsx
kimanhou: Oh right okay
image

kimanhou:
Any preference for the icon in the Navigation ?

image

Profile like OIDC and UNIX ?
image

Or lock locked like TLS ?
image

Or private key (new) ?
image

tomponline:
> Any preference for the icon in the Navigation ?
> image
> 
> Profile like OIDC and UNIX ? image
> 
> Or lock locked like TLS ? image
> 
> Or private key (new) ? image

what does the icon indicate - is it the bearer auth method?

It seems to be the auth method.

Do you know why a different icon is used for different auth methods today?

src/App.tsx
```diff
@@ -145,10 +148,19 @@ const App: FC = () => {
...
+            if (isBearerAuthError(authError)) {
+              fetch(`${ROOT_PATH}/bearer/logout`.replace(/\/+/g, "/")).then(
+                () => {
+                  window.location.href = `${ROOT_PATH}/ui/login`;
+                },
+              );
+            } else {
+              window.location.href = `${ROOT_PATH}/oidc/logout`;
+            }
+          },
```
edlerd, line 157: can we reuse the `logout function here`?

src/util/helpers.tsx
```diff
@@ -285,7 +285,18 @@ export const isDimensionBelow = (
...
 
-export const logout = (hasOidc?: boolean, hasCertificate?: boolean): void => {
+export const logout = (
+  hasOidc?: boolean,
+  hasCertificate?: boolean,
+  isBearer?: boolean,
+): void => {
+  if (isBearer) {
+    void fetch(`${ROOT_PATH}/bearer/logout`.replace(/\/+/g, "/")).then(() => {
+      window.location.href = `${ROOT_PATH}/ui/login/`;
```
edlerd: I think this can be a bit more tailored to the context. Such as checking if OIDC is available, and if not, going to the right certificate page. The same thing the lines below are doing after the OIDC logout already.

Maybe extract that into a helper to call it here and in the lines below as well.

src/App.tsx
```diff
@@ -167,7 +179,7 @@ const App: FC = () => {
     !isAuthenticated &&
     !window.location.href.includes(`${ROOT_PATH}/ui/login`)
   ) {
-    return <Navigate to={`${ROOT_PATH}/ui/login`} replace={true} />;
+    return <Navigate to="/ui/login" replace={true} />;
```
edlerd: Should this change be reverted? I think we should have the root path here, no?

src/util/authentication.ts
```diff
@@ -0,0 +1,16 @@
...
+  TLS: "tls",
+  OIDC: "oidc",
+  UNIX: "unix",
+  BEARER: "bearer",
+} as const;
+
+// Type derived from the constants
+export type LXDAuthMethod = (typeof AUTH_METHOD)[keyof typeof AUTH_METHOD];
+
+export const isPermanentAuthMethod = (
```
edlerd:
```suggestion
export const isPermanent = (
```
Clear enough and shorter. wdyt?

src/context/useAuthCountDown.tsx
```diff
@@ -0,0 +1,36 @@
+import { useState, useEffect } from "react";
+
+export const useAuthCountdown = (expiresAt: string | null | undefined) => {
```
edlerd:
```suggestion
export const useSecondsLeft = (expiresAt: string | null | undefined) => {
```
It seems like this is not tailored towards authentication. Suggesting a more generic name here.
We might add a comment with an example for the format we expect for expires at.

src/context/operationsProvider.tsx
```diff
@@ -74,7 +80,7 @@ const OperationsProvider: FC<Props> = ({ children }) => {
   const ctx = {
     operations,
     runningOperations: running,
-    error,
+    error: error,
```
edlerd: This can stay as it was.

src/context/operationsProvider.tsx
```diff
@@ -30,7 +30,9 @@ const OperationsContext = createContext<OperationsContextType>({
 });
 
 const OperationsProvider: FC<Props> = ({ children }) => {
-  const { isAuthenticated } = useAuth();
+  const { isAuthenticated, authError, isAuthLoading } = useAuth();
+
+  const readyToFetch = !isAuthLoading && !authError && isAuthenticated;
```
edlerd: This is pausing the operation loading on initial landing to only start after the auth queries have finished. So it delays initial render of the application on landing. Can we not block here?

src/util/authentication.ts
edlerd: Do we even need this helper? I suspect we can directly rely on `authTokenExpiresAt` to be set to a value bigger than zero.

src/context/auth.tsx
```diff
@@ -117,6 +121,8 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
...
+        authTokenExpiresAt: currentIdentity?.expires_at ?? null,
```
edlerd:
```suggestion
        authExpiresAt: currentIdentity?.expires_at ?? null,
```
A bit shorter?

src/components/Navigation.tsx
```diff
@@ -708,20 +711,21 @@ const Navigation: FC = () => {
...
                             className="p-side-navigation__icon is-dark"
                             name="profile"
                           />
-                        ) : authMethod == "unix" ? (
+                        ) : authMethod == AUTH_METHOD.BEARER ? (
                           <Icon
                             className="p-side-navigation__icon is-dark"
-                            name="profile"
+                            name="private-key"
                           />
```
edlerd: This was already difficult to read. How about a `authIcon(authMethod)` function that returns the icon name with early exits? Then we can remove the nested conditionals here?

src/components/BearerTokenWarning.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Icon name="warning" />
+      <span className="u-margin-left--small u-hide--small">
+        Temporary session.
+      </span>
+      {secondsLeft && secondsLeft > 0 ? (
+        <span className="u-margin-left--small u-hide--medium u-hide--small">
+          Expires in {formatSeconds(secondsLeft)}.
+        </span>
+      ) : (
+        ""
```
edlerd:
```suggestion
        "Expired"
```
Assuming this is the right interpretation. It would be good if we could generalize this with `getExpiryMessage` in the other component to not have two places do similar logic.

src/components/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <Row>
+        <h1
+          className={classnames("p-heading--4", {
+            "u-sv3": !getExpiryMessage(),
+          })}
+        >
+          Set up permanent access
+        </h1>
+        {getExpiryMessage() && (
+          <Notification severity="caution">{getExpiryMessage()}</Notification>
```
edlerd: We should evaluate `getExpiryMessage` only once.

src/components/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,59 @@
...
+import AuthenticationOptions from "components/AuthenticationOptions";
+import { ROOT_PATH } from "util/rootPath";
+import classnames from "classnames";
+
+const AuthenticationSetup: FC = () => {
+  const { isAuthLoading, isAuthenticated, authMethod, authTokenExpiresAt } =
+    useAuth();
+  const secondsLeft = useAuthCountdown(authTokenExpiresAt);
+  const getExpiryMessage = () => {
+    if (secondsLeft === null) return "Your token expires soon.";
```
edlerd: Is this the right interpretation, or has the token already expired in this case?

src/util/authentication.ts
kimanhou: Technically we could but I find it a bit shaky, especially since `authTokenExpiresAt` keeps changing name and place. `authMethod` is more reliable.

src/context/operationsProvider.tsx
kimanhou: I added it because otherwise failed calls kept piling up indefinitely when the token is invalid.

src/components/AuthenticationSetup.tsx
kimanhou: `secondsLeft === null` means you are not logged in using TLS/bearer token. It should never happen in `AuthenticationSetup` unless there is a bug

src/context/operationsProvider.tsx
edlerd: We already deactivate the operation fetch when `isAuthenticated` is not truthy. So I am wondering why those operation fetches would be triggered with an invalid token. Wdyt?

src/util/helpers.tsx
edlerd: Wdyt about putting those lines from below in a helper function and triggering them here as well instead:

```
    if (hasOidc) {
      window.location.href = `${ROOT_PATH}/ui/login/`;
      return;
    }

    if (hasCertificate) {
      window.location.href = `${ROOT_PATH}/ui/login/certificate-add`;
      return;
    }

    window.location.href = `${ROOT_PATH}/ui/login/certificate-generate`;
  });
```

src/context/operationsProvider.tsx
kimanhou: Can `refetchOperations` be called somewhere else ? It looks like Events.tsx calls it in `onmessage` (line 180) and in `refetchOperations` (line 142)

src/util/helpers.tsx
kimanhou: Sorry I missed that. Fixed now

src/context/operationsProvider.tsx
edlerd: I see what you mean, yes. But I think then the change in this component can be minimal: The `debouncedRefetch` can have an early exit if `!isAuthenticated` instead of looking at the loading and error state of the auth call.

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,200 @@
...
+    createFineGrainedTlsIdentity(
+      values.name,
+      Array.from(desiredState.groupsAdded),
+      false,
+    )
+      .then(() => {
+        logout(false, false, true);
+      })
+      .catch((e) => {
+        notify.failure("TLS Identity failed to be created", e);
```
edlerd: In the error case, the button stays in a loading state

Image

src/components/AuthenticationTlsStepper.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const navigate = useNavigate();
+
+  return (
+    <Stepper
+      variant={isHorizontal ? "horizontal" : "vertical"}
+      steps={[
+        <Step
+          key="Step 1"
+          handleClick={() => {
+            navigate(`${ROOT_PATH}/ui/login/certificate-generate`);
```
edlerd, line 24: I can't go back from step 2 to step 1 with the stepper. The content stays with step 2. Also any other link in the side navigation won't work from step 2.

Image

When I go to "Set up permanent access" > "Set up TLS" > "Step 2" > "step 1" and then use the browser back button, the app crashes with an error

Image

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,200 @@
...
+    <CustomLayout mainClassName="certificates">
+      <AuthenticationTlsStepper
+        isHorizontal
+        step2Name="Create TLS identity"
+        step2Selected={location.pathname.includes(
+          "create-tls-identity-with-bearer-token",
+        )}
+      />
+      {notify.notification ? (
+        <NotificationRow />
```
edlerd: Shouldn't use the row here, but the `NotificationConsumer` directly. Because the row is adding some left/right spacing

Image

src/sass/_authentication_setup.scss
```diff
@@ -0,0 +1,25 @@
+@import "./breakpoints";
+
+.p-panel:has(> .p-panel__content.authentication-setup:first-child) {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
```
edlerd:
```suggestion
```
Better keep this standard left aligned?

src/components/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <Navigate to={`${ROOT_PATH}/ui/permissions/identities`} replace={true} />
+    );
+  }
+
+  if (!isAuthenticated) {
+    return <Navigate to={`${ROOT_PATH}/ui/login`} replace={true} />;
+  }
+
+  return (
+    <CustomLayout contentClassName="authentication-setup u-flex-column">
```
edlerd: I think left aligning the content and ensuring the notification and cards have the same width makes the layout a bit smoother. See quick sketchup below. Wdyt?

Image

src/pages/login/Login.tsx
```diff
@@ -27,34 +24,8 @@ const Login: FC = () => {
...
-              <Link
-                className="has-icon p-button"
-                to={`${ROOT_PATH}/ui/login/certificate-generate`}
-              >
-                <Icon name="certificate" />
-                <span>Set up TLS login</span>
-              </Link>
-            </div>
+            <p className="u-sv1 p-login-method">Choose your login method</p>
+            <AuthenticationOptions />
```
edlerd: The text in the TLS card should not be wrapping within the list elements:

Image

edlerd: Can you ask a designer for a review of this page? It is fairly important to get it right, and the heading and card setup seems a bit off to my eye. Though I can't point out what would improve it.

src/components/AuthenticationTlsStepper.tsx
kimanhou: At some point within the last hour, I had an infinite loop because of existing name checking. It should be fixed now 😬

src/sass/_authentication_setup.scss
kimanhou: You're sure ?
image

edlerd: Sure? No. But I tend to think we should use the same left alignment for the setup pages here as we do for all the other pages in favor of the current center alignment.

src/components/AuthenticationSetup.tsx
kimanhou: Wait they don't have the same width at the moment ? They are supposed to, they both have `flex: 1`
Okay for left aligning

image

src/pages/login/Login.tsx
kimanhou: > The text in the TLS card should not be wrapping within the list elements:

Yeah, I can't figure out why the browser does that. It just decides on its own that it's the best way to display the list content

kimanhou: I am posting on the Design channel. Why is the icon different in light mode and dark mode ?
Screenshot from 2026-02-19 18-21-20
Screenshot from 2026-02-19 18-21-07

src/sass/_authentication_setup.scss
kimanhou: I agree it looks more consistent. I used the `empty-state` layout. 
Should we keep it horizontally centered or left aligned ?

It's current horizontally centered. Here is how it looks on wide screens:
image

The TLS login pages are now left aligned:
image
image

edlerd: > I agree it looks more consistent. I used the `empty-state` layout. Should we keep it horizontally centered or left aligned ?

No strong opinions on this question from my side. Seems ok as is.

> The TLS login pages are now left aligned:

Can you give the content a bit of left padding? I think applying a `row` around all the content should do the right thing there. Can compare it to other list or detail pages for a reference.

src/pages/login/Login.tsx
edlerd: Additional thought on this. We might replace "Login / Choose your login method" with "Choose your login method" as the header with no subtitle. ~And maybe we don't need the container logo either in the heading.~

edlerd: Edit: I think we should keep the icon. It helps the user identify what they are logging in to. But we should fix it in dark mode. It might be a bug in vanilla -- it should stay the same icon but with different colors.

kimanhou: It's weird, the icon is pods with a className, this classname is overridden in styles.scss only for light mode.

```
<Icon name="pods" className="lxd-icon" />
```
```
.lxd-icon {
  background-image: url("data:image/svg+xml,%0A%3Csvg width='38' height='38' viewBox='0 0 38 38' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M28.5311 15.7998V5.71128L18.6554 0L8.77977 5.71128V15.7998L0 20.8683V32.2887L9.87566 38L18.6554 32.9083L27.4352 37.9768L37.3108 32.2655V20.8451L28.5311 15.7998ZM18.6554 2.57113L26.3182 6.99684V15.7998L18.6554 20.2487L10.9694 15.823V7.02002L18.6554 2.57113ZM9.87566 35.431L2.21286 31.0053V22.1538L9.85037 17.7513L17.5595 22.2002V31.0032L9.87355 35.4289L9.87566 35.431ZM35.098 31.0053L27.4352 35.431L19.7724 31.0053V22.2023L27.4816 17.7534L35.1191 22.156V31.0074H35.0959L35.098 31.0053Z' fill='%23666666'/%3E%3Cpath d='M28.1978 31.9072L32.4085 29.4099V24.4594L28.1978 26.9357V31.9072Z' fill='%23666666'/%3E%3Cpath d='M27.4119 20.7966L23.2012 23.2223L27.4372 25.627L31.648 23.1507L27.4119 20.7966Z' fill='%23666666'/%3E%3Cpath d='M22.4635 29.4816L26.6995 31.9558V26.9357L22.415 24.5311L22.4635 29.4816Z' fill='%23666666'/%3E%3Cpath d='M9.85235 20.7966L5.6416 23.2223L9.87553 25.627L14.0884 23.1507L9.85235 20.7966Z' fill='%23666666'/%3E%3Cpath d='M4.90371 29.4816L9.13764 31.9558V26.9357L4.87842 24.5311L4.90371 29.4816Z' fill='%23666666'/%3E%3Cpath d='M10.6387 31.9072L14.8726 29.4099V24.4594L10.6387 26.9357V31.9072Z' fill='%23666666'/%3E%3Cpath d='M22.8455 7.97052L18.6326 5.63965L14.4219 8.06536L18.6326 10.4447L22.8455 7.97052Z' fill='%23666666'/%3E%3Cpath d='M17.8946 16.7755V11.7534L13.6606 9.35083L13.6838 14.3245L17.8946 16.7755Z' fill='%23666666'/%3E%3Cpath d='M19.3955 16.7503L23.6294 14.2529V9.2561L19.3955 11.7535V16.7503Z' fill='%23666666'/%3E%3C/svg%3E%0A");
  height: 5rem;
  opacity: 1;
  width: 5rem;
}
```

I can hardcode something for dark mode but I feel like I am missing something. Wdyt ?

image

edlerd: I see. The override is likely from a time when the icon was not available in vanilla and we didn't have dark mode. Let's remove the `background-image` from our SCSS and switch to the right icon name: `cluster-host` instead of `pods` for the login page.

src/sass/_authentication_setup.scss
kimanhou: > No strong opinions on this question from my side. Seems ok as is.

OKay, I left it as is.

> Can you give the content a bit of left padding?

Done

src/pages/login/Login.tsx
kimanhou: I toned down the `opacity` to 0.5.
image
image

edlerd: Oh wait, this is a different logo. Sorry for the confusion. I think we might need the override, as this should be the LXD logo with the containers inside the hexagons like in the top left corner.

kimanhou: Yeah I prefer the "real" logo too

image

image

src/components/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return <Navigate to={`${ROOT_PATH}/ui/login`} replace={true} />;
+  }
+
+  return (
+    <CustomLayout contentClassName="authentication-setup empty-state">
+      <Row>
+        <h1
+          className={classnames("p-heading--4", {
+            "u-sv3": !expiryMessage,
+          })}
```
edlerd:
```suggestion
          className="p-heading--4"
```
The missing `expiryMessage` seems like an "impossible" edge case we should ignore to keep this simple. Wdyt?

src/components/AuthenticationTlsStepper.tsx
```diff
@@ -0,0 +1,55 @@
...
+const AuthenticationTlsStepper: FC<Props> = ({
+  isHorizontal,
+  step2Selected,
+  step2Name = "Identity trust token",
+}) => {
+  const navigate = useNavigate();
+
+  return (
+    <Stepper
+      variant={isHorizontal ? "horizontal" : "vertical"}
```
edlerd: Why turn this into a boolean? We can pass in the variant as a string and preserve the interface from the stepper component internally.

src/pages/login/CertificateAdd.tsx
```diff
@@ -3,37 +3,44 @@ import {
...
 
+  if (isAuthenticated && isBearerToken) {
+    return (
+      <Navigate
+        to={`${ROOT_PATH}/ui/login/create-tls-identity-with-bearer-token`}
+        replace={true}
+      />
+    );
+  }
+
```
edlerd, line 25:
```suggestion
```
I think we should have the same URL for `ui/login/create-tls-identity-with-bearer-token` and `ui/login/certificate-add`. In this component here, we can move everything into a component, and then this component does something like `return isBearer ? <CertifcateAddBearer /> : <CertificateAddToken />`. Maybe there is even more logic for both to share inside this component.

This simplifies various use cases, as we don't need to know which URL to link to, so things over there don't need logic about the two cases. 

Wdyt?

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,229 @@
...
+      }
+      return `${baseName}-${counter}`;
+    },
+    [identities],
+  );
+
+  const initialName = useMemo(
+    () => findAvailableName("lxd-ui"),
+    [findAvailableName],
+  );
```
edlerd:
```suggestion
```
Can we directly call `findAvailableName("lxd-ui")` where this is used?

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  const { desiredState, save: saveToPanelHistory } =
+    useEditHistory<GroupEditHistory>({
+      initialState: {
+        groupsAdded: new Set(),
+      },
+    });
+
+  if (error) {
+    notify.failure("Loading panel details failed", error);
```
edlerd:
```suggestion
    notify.failure("Loading details failed", error);
```

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,229 @@
...
+  useEffect(() => {
+    if (groups.length > 0 && desiredState.groupsAdded.size === 0) {
+      const adminsGroup = groups.find((group) => group.name === "admins");
+      if (adminsGroup) {
+        saveToPanelHistory({
+          groupsAdded: new Set(["admins"]),
+        });
+      }
+    }
+  }, [groups, desiredState.groupsAdded.size, saveToPanelHistory]);
```
edlerd:
```suggestion
```
This doesn't allow unselecting the group. Can we do this in the initialvalues?

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,229 @@
...
+    () => findAvailableName("lxd-ui"),
+    [findAvailableName],
+  );
+
+  const { desiredState, save: saveToPanelHistory } =
+    useEditHistory<GroupEditHistory>({
+      initialState: {
+        groupsAdded: new Set(),
+      },
+    });
```
edlerd:
```suggestion
```
We probably don't need the history feature here. We don't show the undo feature anyway. Do you think we can remove it and keep the state solely in formik?

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -118,7 +119,7 @@ const CreateIdentityModal: FC<Props> = ({ onClose, token, identityName }) => {
                         items={[
                           <>
                             Open an unauthenticated browser on{" "}
-                            <code>{location.origin}</code>.
+                            <Link to={location.origin}>{location.origin}</Link>.
```
edlerd, line 122: Thanks for the quick fix on this one :)

src/App.tsx
```diff
@@ -560,6 +571,14 @@ const App: FC = () => {
           path={`${ROOT_PATH}/ui/login/certificate-add`}
           element={<CertificateAdd />}
         />
+        <Route
+          path={`${ROOT_PATH}/ui/login/create-tls-identity-with-bearer-token`}
+          element={<CreateTlsIdentityWithBearerToken />}
+        />
+        <Route
+          path={`${ROOT_PATH}/ui/authentication-setup`}
```
edlerd, line 572:
```suggestion
          path={`${ROOT_PATH}/ui/login/authentication-setup`}
```
Shall we keep this under the login folder?

kimanhou: It was originally in the login folder but then App.tsx got really messy in case the bearer token is invalid. The main reason is that the login folder is for pages where the user is not authenticated while the authentication setup is.

src/pages/login/CertificateAdd.tsx
kimanhou: Good idea, that simplifies a lot of things

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  const formik = useFormik<TLSIdentityFormValues>({
+    initialValues: {
+      name: findAvailableName("lxd-ui"),
+      groups: initialGroups,
+    },
+    validationSchema: groupSchema,
+    onSubmit: handleSubmit,
+    enableReinitialize: identities.length > 0 || groups.length > 0,
```
edlerd:
```suggestion
    enableReinitialize: true,
```
A bit simpler?

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  // Revalidate when identities change to check for name conflicts
+  useEffect(() => {
+    if (formik.values.name && identities.length > 0) {
+      void formik.validateForm();
+    }
+  }, [identities, formik.validateForm, formik.values.name]);
```
edlerd:
```suggestion
```
I think we don't need this, because the re-initialization will set the new name anyhow.

src/util/helpers.tsx
```diff
@@ -285,23 +285,41 @@ export const isDimensionBelow = (
...
 
-    if (hasCertificate) {
-      window.location.href = `${ROOT_PATH}/ui/login/certificate-add`;
+export const logout = (
+  hasOidc?: boolean,
+  hasCertificate?: boolean,
+  isBearer?: boolean,
+): void => {
+  if (isBearer) {
+    void fetch(`${ROOT_PATH}/bearer/logout`.replace(/\/+/g, "/")).then(() => {
```
edlerd:
```suggestion
    void fetch(`${ROOT_PATH}/bearer/logout`).then(() => {
```
Why the replace?

src/util/helpers.tsx
```diff
@@ -285,23 +285,41 @@ export const isDimensionBelow = (
...
-    }
+    });
+  }
+
+  if (window.location.pathname.includes(`${ROOT_PATH}/ui/login`)) {
+    return;
+  }
 
-    window.location.href = `${ROOT_PATH}/ui/login/certificate-generate`;
+  void fetch(`${ROOT_PATH}/oidc/logout`.replace(/\/+/g, "/")).then(() => {
```
edlerd:
```suggestion
  void fetch(`${ROOT_PATH}/oidc/logout`).then(() => {
```
Why the replace?

src/util/helpers.tsx
```diff
@@ -285,23 +285,41 @@ export const isDimensionBelow = (
...
-    if (hasOidc) {
-      window.location.href = `${ROOT_PATH}/ui/login/`;
-      return;
-    }
+  if (hasCertificate) {
+    window.location.href = `${ROOT_PATH}/ui/login/certificate-add`;
+    return;
+  }
+
+  window.location.href = `${ROOT_PATH}/ui/login/certificate-generate`;
```
edlerd:
```suggestion
    window.location.href = `${ROOT_PATH}/ui/login/`;
```
Let's send the users to the login method choice here.

src/util/helpers.tsx
```diff
@@ -285,23 +285,41 @@ export const isDimensionBelow = (
...
+  hasCertificate?: boolean,
+  isBearer?: boolean,
+): void => {
+  if (isBearer) {
+    void fetch(`${ROOT_PATH}/bearer/logout`.replace(/\/+/g, "/")).then(() => {
+      redirectAfterLogout(hasOidc, hasCertificate);
       return;
-    }
+    });
+  }
```
edlerd:
```suggestion
 export const logoutBearerToken = () => {
    void fetch(`${ROOT_PATH}/bearer/logout`).then(() => {
      window.location.href = `${ROOT_PATH}/ui/login/`
return;
    });
  }
```
Let's move this to an explicit bearer logout function. To decouple the logic from the OIDC logout. It seems there is no overlap.

Also: Why the replace?

src/App.tsx
```diff
@@ -145,10 +148,15 @@ const App: FC = () => {
       const logoutAction = [
         {
           label: "Logout",
-          onClick: () => (window.location.href = `${ROOT_PATH}/oidc/logout`),
+          onClick: () => {
+            if (isBearerAuthError(authError)) {
+              logout(false, false, true);
+            } else {
+              window.location.href = `${ROOT_PATH}/oidc/logout`;
```
```suggestion
              logout(hasOidc, hasCertificate);;
```
Shall we use the logout function?

src/sass/_authentication_options.scss
```diff
@@ -0,0 +1,60 @@
...
+  justify-content: center;
+
+  @include mobile-and-tablet {
+    grid-template-columns: 1fr;
+  }
+
+  .auth-option-card {
+    background-color: var(--vf-color-background-active);
+    border: none;
+    border-radius: 4px;
```
edlerd:
```suggestion
```

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  const handleSubmit = (values: TLSIdentityFormValues) => {
+    createFineGrainedTlsIdentity(values.name, values.groups ?? [], false)
+      .then(() => {
+        logout(false, false, true);
+      })
+      .catch((e) => {
+        formik.setSubmitting(false);
+        notify.failure("TLS Identity failed to be created", e);
```
edlerd:
```suggestion
        notify.failure("Identity creation failed", e);
```
A bit simpler message

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,186 @@
...
+      });
+  };
+
+  const groupSchema = Yup.object().shape({
+    name: Yup.string()
+      .required("Identity name is required")
+      .test("unique-name", "Identity name already exists", function (value) {
```
edlerd:
```suggestion
      .test("unique-name", "An identity with this name already exists", function (value) {
```

src/util/helpers.tsx
kimanhou: I'm not sure, it was already there. I just copy pasted the same for bearer token

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
kimanhou: Shall we also change it in `CreateTLSIdentityPanel` ?

edlerd: Yes, please.

src/util/helpers.tsx
edlerd: It seems this was added with the relative URL support in January. Do you remember a need for it? I missed it back then in the review. If there is no need, it is best to remove it to keep things simple.

kimanhou: I think it was to avoid double slashes in case `ROOT_PATH` was empty or ended with a slash like `/foo/bar/`

edlerd: Then with the current approach, it is not needed anymore, right?

src/pages/login/AuthenticationSetup.tsx
edlerd, line 1: This should be in `src/pages` as it is not a reusable component, but a page.

src/pages/login/AuthenticationSetup.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  return (
+    <CustomLayout contentClassName="authentication-setup empty-state">
+      <Row>
+        <h1 className="p-heading--4 u-sv3">Set up permanent access</h1>
+        {expiryMessage && (
+          <Notification severity="caution">{expiryMessage}</Notification>
+        )}
+        <AuthenticationOptions />
```
edlerd, line 43: There are some edge cases for responsiveness where those are not aligned. See below:

Image

Image

Image

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -0,0 +1,186 @@
...
+  if (isAuthenticated && isPermanent(authMethod)) {
+    return <Navigate to={`${ROOT_PATH}/ui`} replace={true} />;
+  }
+
+  if (!isAuthenticated || !isBearerToken) {
+    return <Navigate to={`${ROOT_PATH}/ui/login`} replace={true} />;
+  }
+
+  return (
+    <CustomLayout mainClassName="certificates-with-bearer-token">
```
edlerd, line 116: This has too much bottom spacing with the button margin bottom and the panel having bottom spacing as well:

Image

src/util/helpers.tsx
```diff
@@ -285,23 +285,28 @@ export const isDimensionBelow = (
     ? window.innerWidth < dimension
     : window.innerHeight < dimension;
 
-export const logout = (hasOidc?: boolean, hasCertificate?: boolean): void => {
-  if (window.location.pathname.includes(`${ROOT_PATH}/ui/login`)) {
+const redirectAfterLogout = (hasCertificate?: boolean): void => {
+  if (hasCertificate) {
+    window.location.href = `${ROOT_PATH}/ui/login/certificate-add`;
     return;
   }
```
edlerd:
```suggestion
```
I am having second thoughts on this from testing. I think after logout we should always show the login page with the choice. Putting the user into the certificate flow might not be the right thing. That means we can probably remove the `redirectAfterLogout` and inline the redirect directly. Wdyt?

src/util/seconds.tsx
```diff
@@ -27,3 +27,24 @@ export const formatSeconds = (seconds: number): string => {
...
+    return shortFormat ? "Expired" : "Your token has expired";
+  }
+
+  const timeString = formatSeconds(secondsLeft);
+
+  if (shortFormat) {
+    return `Expires in ${timeString}`;
+  }
+
+  return `Your token will expire in ${timeString}`;
```
edlerd:
```suggestion
  return `Temporary access expires in ${timeString}`;
```
Shall we reuse the language from the cli?

src/util/seconds.tsx
```diff
@@ -27,3 +27,24 @@ export const formatSeconds = (seconds: number): string => {
...
+export const getExpiryMessage = (
+  secondsLeft: number | null,
+  shortFormat?: boolean,
+): string | null => {
+  if (secondsLeft === null) {
+    return null;
+  }
+
+  if (secondsLeft <= 0) {
+    return shortFormat ? "Expired" : "Your token has expired";
```
edlerd:
```suggestion
    return shortFormat ? "Expired" : "Temporary access has expired";
```

src/components/BearerTokenWarning.tsx
```diff
@@ -0,0 +1,34 @@
...
+const BearerTokenWarning: React.FC = () => {
+  const { authExpiresAt } = useAuth();
+  const secondsLeft = useSecondsLeft(authExpiresAt);
+  const expiryMessage = getExpiryMessage(secondsLeft, true);
+
+  return (
+    <div>
+      <Icon name="warning" />
+      <span className="u-margin-left--small u-hide--small">
+        Temporary session.
```
edlerd:
```suggestion
        Temporary access
```
Shall we make this a sentence such as `Temporary access expires in 23:55:29.`?

Also would shorten the time format to be much easier to gasp.

MusicDin: > Temporary access expires in 23:55:29.

I like this one.

Just note that the bearer token can also be issued with longer expiry (for example 10 years). This is something we neither recommend nor advertise, but is technically possible.

src/util/seconds.tsx
MusicDin: Maybe: Temporary access **token** expires?

edlerd: The user clicked a link and got access. I'd expect them to not know anything about tokens here.

MusicDin: Agree.

> The user clicked a link and got access.

True, but a token can be explicitly generated and the created identity is of type `bearer`. In any case, `Temporary access expires` seems fine.

edlerd: Yeah, I understand the concrete implementation details. The challenge is often that as a developer, we have a very detailed system model in our mind. For a good UX, we need to translate to the users mental model, so the message aligns nicely and is easy to understand.

Edit: The user's mental model is usually much simpler than the system model and built step by step while they are using the tool at hand.

MusicDin: Thanks for the explanation, makes sense 🚀

src/pages/login/AuthenticationSetup.tsx
kimanhou: Oh right, it can live inside `src/pages/login` but I would prefer the URL not to be under `login/`

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
kimanhou: I put back `u-no-margin--bottom` on the buttons

src/components/BearerTokenWarning.tsx
kimanhou: Long periods of time
image
image
image

Under 24 hours
image
image

edlerd: Nice! Do we need the leading zeros for the long periods?

kimanhou: I removed the padding for year, month and day
image

PR #1749, opened 2026-02-02
feat(delete instance): Use force flag

kimanhou:
## Done

- Delete instance: use force flag

## QA

1. Run the LXD-UI:
    - Have a running instance. Delete it, make sure in the Network tab that the force flag is sent to back end.
    - If you are connected to the latest version of back end, the deletion should work.

## Screenshots
image
image

### DIFF
--- a/src/pages/instances/actions/DeleteInstanceBtn.tsx
+++ b/src/pages/instances/actions/DeleteInstanceBtn.tsx
@@ -21,6 +21,7 @@ import { Notification } from "@canonical/react-components";
 import { InstanceRichChip } from "../InstanceRichChip";
 import ConfirmationCheckbox from "components/ConfirmationCheckbox";
 import { ROOT_PATH } from "util/rootPath";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
 
 interface Props {
   instance: LxdInstance;
@@ -42,6 +43,7 @@ const DeleteInstanceBtn: FC<Props> = ({
   const [isLoading, setLoading] = useState(false);
   const navigate = useNavigate();
   const { canDeleteInstance } = useInstanceEntitlements();
+  const { hasInstanceForceDelete } = useSupportedFeatures();
   const [isForce, setForce] = useState(false);
 
   const isRunningOrFrozen =
@@ -55,7 +57,7 @@ const DeleteInstanceBtn: FC<Props> = ({
       />
     );
 
-    deleteInstance(instance)
+    deleteInstance(instance, hasInstanceForceDelete && isForce)
       .then((operation) => {
         eventQueue.set(
           operation.metadata.id,
@@ -102,7 +104,8 @@ const DeleteInstanceBtn: FC<Props> = ({
 
     setLoading(true);
 
-    if (isRunningOrFrozen) {
+    // If backend doesn't support force delete, stop the instance first
+    if (isRunningOrFrozen && !hasInstanceForceDelete) {
       const instanceLink = (
         <InstanceRichChip
           instanceName={instance.name}
--- a/tests/instances.spec.ts
+++ b/tests/instances.spec.ts
@@ -410,3 +410,30 @@ test("'Other' tab is removed when config/creating Instances, on LXD Version 5.0"
   await editInstance(page, instance);
   await expect(page.getByText("Other", { exact: true })).not.toBeVisible();
 });
+
+test("instance deletion (with and without force delete support)", async ({
+  page,
+}) => {
+  await visitAndStartInstance(page, instance);
+
+  await page.getByRole("button", { name: "Delete" }).click();
+
+  const dialog = page.getByRole("dialog", { name: "Confirm delete" });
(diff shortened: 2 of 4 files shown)


src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -1,6 +1,6 @@
 import type { FC } from "react";
 import { useState } from "react";
-import { deleteInstance, stopInstance } from "api/instances";
+import { deleteInstance } from "api/instances";
```
edlerd: I think we have to keep the old implementation as a fallback for older backend versions. Can we react on the api extension for the force deletion here? Or is there no extension available?

edlerd: I am surprised this is not caught by CI failures. Can you please add an instance force delete test?

kimanhou: Yes, there is ! `instance_force_delete`

edlerd: Ok, then I think the UI should react to that api extension.

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -54,7 +56,7 @@ const DeleteInstanceBtn: FC<Props> = ({
       />
     );
 
-    deleteInstance(instance)
+    deleteInstance(instance, isForce)
```
edlerd: Can we change this so we only set here `isForce` if `hasInstanceForceDelete` and keep it undefined or false otherwise?

That makes it a bit more explicit. Currently this is implied by the API ignoring the force flag in older versions.

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -101,7 +103,7 @@ const DeleteInstanceBtn: FC<Props> = ({
 
     setLoading(true);
 
-    if (isRunningOrFrozen) {
+    if (isRunningOrFrozen && !hasInstanceForceDelete) {
```
edlerd, line 108: Can we add a comment here that we force stop the instance if the API does not support the force flag?

tests/instances.spec.ts
```diff
@@ -410,3 +432,90 @@ test("'Other' tab is removed when config/creating Instances, on LXD Version 5.0"
   await editInstance(page, instance);
   await expect(page.getByText("Other", { exact: true })).not.toBeVisible();
 });
+
+test("instance force delete", async ({ page, lxdVersion }) => {
+  test.skip(
+    lxdVersion === "5.0-edge",
+    "instance_force_delete API extension is not supported in LXD 5.0",
+  );
+
```
edlerd:
```suggestion
```
I think this is not true. Our workaround with stopping and then deleting should work on 5.0 just fine.

tests/instances.spec.ts
```diff
@@ -410,3 +432,90 @@ test("'Other' tab is removed when config/creating Instances, on LXD Version 5.0"
...
+  expect(stopData?.action).toBe("stop");
+
+  await expect(page.getByText(`Instance ${instance} deleted.`)).toBeVisible();
+
+  await page
+    .getByRole("link", { name: "Instances", exact: true })
+    .first()
+    .click();
+  await expect(page.getByText(instance)).not.toBeVisible();
+});
```
edlerd:
```suggestion
```
We probably don't need this and can rely on the other test. Wdyt?

tests/instances.spec.ts
```diff
@@ -53,11 +53,33 @@ test.beforeAll(async ({ browser, browserName }) => {
...
+    );
   }
-  await deleteProfile(page, profile);
+
+  try {
+    await deleteProfile(page, profile);
+  } catch (e) {
+    console.log(`Profile ${profile} cleanup failed. ${getErrorMessage(e)}`);
+  }
+
```
edlerd: Let's keep the shutdown functions here a bit more strict. Why allow failures in here?

tests/instances.spec.ts
```diff
@@ -410,3 +410,59 @@ test("'Other' tab is removed when config/creating Instances, on LXD Version 5.0"
...
+    const delReq = await deleteRequest;
+    const url = new URL(delReq.url());
+    console.log("force", url.searchParams.get("force"));
+    console.log("url", url.toString());
+    expect(url.searchParams.get("force")).toBe("1");
+  } else {
+    const [stopReq, _] = await Promise.all([stopRequest, deleteRequest]);
+    const stopData = stopReq.postDataJSON() as { action: string } | null;
+    expect(stopData?.action).toBe("stop");
+  }
```
edlerd:
```suggestion
```
I am not sure if we need this level of detail here. Maybe enough to only test the user facing, higher level, here. Wdyt?

edlerd (dismissed):
LGTM, one question on simplifying the test.

PR #1758, opened 2026-02-04
feat(delete instance): Disable delete button when instance is protected

kimanhou:
## Done

- Disable delete button when instance is protected

## QA

1. Run the LXD-UI:
    - Delete an instance that is not protected from deletion: the Delete button should be enabled
    - Go to the instance's security policies and set "Protect deletion" to `Yes`. Save change. The Delete button should now be disabled. When hovered, the button's title should display a meaningful message: `Instance is protected from deletion. Update security policies to allow deletion.`

## Screenshots

image

image

### DIFF
--- a/src/pages/instances/actions/DeleteInstanceBtn.tsx
+++ b/src/pages/instances/actions/DeleteInstanceBtn.tsx
@@ -134,17 +134,24 @@ const DeleteInstanceBtn: FC<Props> = ({
   };
 
   const isDeletableStatus = deletableStatuses.includes(instance.status);
+  const isProtectedFromDelete =
+    instance.expanded_config["security.protection.delete"] === "true";
   const isDisabled =
     isLoading ||
     !isDeletableStatus ||
     instanceLoading.getType(instance) === "Migrating" ||
-    !canDeleteInstance(instance);
+    !canDeleteInstance(instance) ||
+    isProtectedFromDelete;
+
   const getHoverText = () => {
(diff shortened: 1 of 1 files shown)


src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -134,18 +134,25 @@ const DeleteInstanceBtn: FC<Props> = ({
...
+
   const getHoverText = () => {
     if (!canDeleteInstance(instance)) {
       return "You do not have permission to delete this instance";
     }
     if (!isDeletableStatus) {
       return "Stop the instance to delete it";
     }
+    if (isProtectedFromDelete) {
+      return "Instance is protected from deletion. Update security policies to allow deletion.";
```
edlerd:
```suggestion
      return "Update security policies to remove the delete protection.";
```
Maybe a bit more concise?

src/pages/instances/actions/DeleteInstanceBtn.tsx
```diff
@@ -134,18 +134,25 @@ const DeleteInstanceBtn: FC<Props> = ({
   };
 
   const isDeletableStatus = deletableStatuses.includes(instance.status);
+  const isProtectedFromDelete =
+    instance.config["security.protection.delete"] === "true";
```
edlerd: Let's look at the `expanded_config` instead, as this can also be inherited from a profile. The expanded considers also inherited values.

PR #1759, opened 2026-02-04
feat(create auth group): disable all fields as long as name is invalid

kimanhou:
## Done

- Create auth group: disable all fields in the side panel as long as name is invalid

## QA

1. Run the LXD-UI:
    - Go to Permissions > Groups > Create auth group
    - All fields should be disabled
    - Type a valid name
    - All fields should be enabled

## Screenshots

Screenshot from 2026-02-04 11-25-39

### DIFF
--- a/src/pages/permissions/forms/GroupForm.tsx
+++ b/src/pages/permissions/forms/GroupForm.tsx
@@ -52,6 +52,22 @@ const GroupForm: FC<Props> = ({
       ? ""
       : "You do not have permission to modify this group";
 
+  const isNameInvalid = !formik.values.name || !!formik.errors.name;
+  const isFieldDisabled = !!groupEditRestriction || isNameInvalid;
+
+  const getDisableReason = () => {
+    if (groupEditRestriction) {
+      return groupEditRestriction;
+    }
+
+    if (isNameInvalid) {
+      return "Enter a valid group name first";
+    }
+
+    return undefined;
+  };
+  const disableReason = getDisableReason();
+
   return (
     <Form onSubmit={formik.handleSubmit}>
       {/* hidden submit to enable enter key in inputs */}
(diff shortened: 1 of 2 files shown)


src/pages/permissions/forms/GroupForm.tsx
```diff
@@ -68,8 +70,11 @@ const GroupForm: FC<Props> = ({
...
-        disabled={!!groupEditRestriction}
-        title={groupEditRestriction}
+        disabled={!!groupEditRestriction || isNameInvalid}
+        title={
+          groupEditRestriction ||
+          (isNameInvalid ? "Enter a valid group name first" : "")
+        }
       />
       <FormLink
         title={(isEditing ? "Edit " : "Add ") + pluralize("identity", 2)}
```
edlerd, line 91: We should alter the title here and below if the name is invalid.

kimanhou: What do you suggest ? 🤔

kimanhou: Wait, are you talking about changing "Add identites" and "Add permissions" ? Or the on hover title ?

edlerd: I confused the prop here. The title is indeed correct. But we should add a hover title. Like the description has a hover title when it is disabled. Those two other buttons for identities/permissions should also have a hover title.

kimanhou: Understood

src/pages/permissions/forms/GroupForm.tsx
```diff
@@ -52,6 +52,12 @@ const GroupForm: FC<Props> = ({
       ? ""
       : "You do not have permission to modify this group";
 
+  const isNameInvalid = !formik.values.name || !!formik.errors.name;
+  const isFieldDisabled = !!groupEditRestriction || isNameInvalid;
+  const hoverText =
+    groupEditRestriction ||
+    (isNameInvalid ? "Enter a valid group name first" : undefined);
```
edlerd: We tend to call this `disableReason` in other places. Better keep the naming consistent.

The calculation of the hover text with nested booleans is fairly complex and difficult to read. We can use a function with early exits to make the calculation more straightforward. Wdyt?

edlerd (approved):
LGTM, thanks :)

PR #1760, opened 2026-02-04
chore: upgrade circular deps check depth to 3 and fix errors WD-33022

omarelkashef:
## Done

- upgrade circular deps check depth to 3 and fix errors including moving some types to dedicated files

## QA

1. Run the LXD-UI:
    - Run `npm run check-circular-deps` and make sure no errors are reported.

## Screenshots


src/components/forms/InstanceSnapshotsForm.tsx
```diff
@@ -14,13 +12,6 @@ import { useCurrentProject } from "context/useCurrentProject";
 import { isSnapshotsDisabled } from "util/snapshots";
 import SnapshotDisabledWarningLink from "components/SnapshotDisabledWarningLink";
 
-export interface SnapshotFormValues {
-  snapshots_pattern?: string;
-  snapshots_expiry?: string;
-  snapshots_schedule?: string;
-  snapshots_schedule_stopped?: string;
-}
-
```
edlerd, line 23: We have no connection between this form and the `SnapshotFormValues` type apart from the name similarity. Can we make this more explicit?

I think we could do something like `export const snapshotsPayload = (values: SnapshotFormValues)` to keep the connection here and in the other similar instance/profile related forms.

src/types/formValues.d.ts
edlerd, line 1: This file is pulling any form type into a single file. It seems right to keep the instance- and profile-related ones together. Because they refer to and build the same formik object.

For the others, I'd suggest a dedicated file for each individual form. Maybe structure them as `src/types/forms/instancesProfiles.d.s`, `src/types/forms/storageBuckets.d.ts`, , `src/types/forms/storageBucketKeys.d.ts`, etc.

Also, I'd suggest moving the `SnapshotFormValues` from util/snapshots.tsx into `src/types/forms/snapshots.d.ts`. The name clashes with the instance related one, but keeping them in dedicated files clarifies the collision.

src/types/forms/networkAcl.d.ts
```diff
@@ -0,0 +1,28 @@
+import type { LxdNetworkAcl } from "types/network";
+
+export interface AclRuleFormValues {
```
edlerd:
```suggestion
export interface NetworkAclRuleFormValues {
```
Can we make the naming more consistent? We could also drop the network prefix on the interface below alternatively.

src/components/forms/BootForm.tsx
```diff
@@ -1,23 +1,13 @@
...
 
-export interface BootFormValues {
-  boot_autostart?: string;
-  boot_autostart_delay?: string;
-  boot_autostart_priority?: string;
-  boot_host_shutdown_timeout?: string;
-  boot_stop_priority?: string;
-}
-
 export const bootPayload = (values: InstanceAndProfileFormValues) => {
```
edlerd: Can we use the `BootFormValues` in the `bootPayload` signature as a type for the values? That way we keep the connection between the type and the source of the type.

The same is true for several other forms that are part of bigger forms, such as the migrationPayload, resourceLimitsPayload, sshKeyPayload and several others.

src/types/forms/group.d.ts
```diff
@@ -0,0 +1,4 @@
+export interface GroupFormValues {
```
edlerd:
```suggestion
export interface PermissionGroupFormValues {
```
Let's rephrase this to disambiguate it from the cluster group and idp group concepts.

src/types/device.d.ts
```diff
@@ -190,3 +190,55 @@ export type LxdDeviceValue =
...
+  | FormDiskDevice
+  | FormNetworkDevice
+  | UnknownDevice
+  | NoneDevice
+  | CustomNetworkDevice
+  | IsoVolumeDevice
+  | LxdGPUDevice
+  | LxdProxyDevice
+  | LxdOtherDevice
+  | EmptyDevice;
```
edlerd: Let's keep these in a dedicated formDevice file. Because the other types in this file relate to API types used in response/requests to/from LXD. Those marked here by this comment are only used in the UI to keep form state.

src/pages/instances/forms/InstanceCreateDetailsForm.tsx
```diff
@@ -17,31 +17,16 @@ import {
...
 import InstanceTargetSelect from "pages/instances/forms/InstanceTargetSelect";
 import UseCustomIsoBtn from "pages/images/actions/UseCustomIsoBtn";
 import AutoExpandingTextArea from "components/AutoExpandingTextArea";
 import ScrollableForm from "components/ScrollableForm";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
 import UploadInstanceFileBtn from "../actions/UploadInstanceFileBtn";
-import type { InstanceIconType } from "components/ResourceIcon";
 import SshKeyForm from "components/forms/SshKeyForm";
 
-export interface InstanceDetailsFormValues {
```
edlerd, line 31: this needs to be referenced below.

src/pages/profiles/forms/ProfileDetailsForm.tsx
```diff
@@ -10,15 +10,6 @@ import { useIsClustered } from "context/useIsClustered";
 import PlacementGroupSelect from "pages/instances/forms/PlacementGroupSelect";
 import { getInstanceField } from "util/instanceConfigFields";
 
-export interface ProfileDetailsFormValues {
```
edlerd, line 13: this needs to be referenced below.

src/types/forms/permissionGroup.d.ts
edlerd, line 1: Can we rename this file to "permissionGroup.d.ts"?

edlerd (approved):
LGTM, thanks for fixing and cleaning our types.

PR #1765, opened 2026-02-04
fix tests to cleanup images from previous test run failures

edlerd:
## Done

- fix tests to cleanup images from previous test run failures


Kxiru (approved):
LGTM, but couldn't this be a function that runs in every CI run regardless of whether CI fails for not? Eg; if this function runs and fails, it may delete the previous Imgs, but not those for this function itself.

Just a nitpick, really. Can merge as is, too.

PR #1767, opened 2026-02-04
fix(confirmation) ensure force flags are in sync with ui selections

### LINKED ISSUE #1746: force restart/poweroff ignored on 6/edge
on my production LXD 6.6, clicking Restart with the force checkmarks a VM immediately powers off and turns on again.

I just upgraded to 6/edge on my laptop and clicking Restart with the force checkmark now results in the VM trying to do a graceful shutdown. Force checkmark or not, the behavior is the same.

On 6.6 I sometimes forgot to use the force checkmark on a restart. I always "fixed" this by issueing a Stop with the force checkmark. This also no longer seems to work. I just issued the force stop 3x on the VM attempting the graceful shutdown and it simply stayed running. 

force power off seems to work by itself, but when combined with the restart it no longer works.

I do not know if this behavioral change is UI or LXD related. Also not sure if you want bug reports for 6/edge :)

https://github.com/user-attachments/assets/cdf82ff9-1e06-42ab-ba8e-97c644d58a29

edlerd:
## Done

- fix(confirmation) ensure force flags are in sync with ui selections.

Fixes #1746

## QA

1. Run the LXD-UI:
  - Go to an instance detail page and start the instance.
  - Refresh the instance detail page
  - Click stop > enable force flag > confirm stop
  - Investigate in the network tab that the request is sent with `force: true`

## Screenshots

image

### DIFF
--- a/src/pages/instances/actions/RestartInstanceBtn.tsx
+++ b/src/pages/instances/actions/RestartInstanceBtn.tsx
@@ -1,4 +1,5 @@
 import type { FC } from "react";
+import { useEffect, useRef } from "react";
 import { useState } from "react";
 import type { LxdInstance } from "types/instance";
 import { useQueryClient } from "@tanstack/react-query";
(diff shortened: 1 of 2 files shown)


kimanhou:
**QA**

- ✅ I can reproduce the bug for `StopInstanceBtn` and `RestartInstanceBtn`. With your fix, the request is sent with `force: true`, as shown in the UI.
- ❌ However I can't reproduce the bug for `DeleteInstanceBtn` and `InstaceBulkAction`. The difference I see is that stop and restart use `ConfirmationButton`, while delete and bulk action use `MountedConfirmationButton` but I can't really explain it.
- 😬 For restoring a snapshot, I don't know how to get the snapshot to have `stateful=true` so it always sends `stateful: false`

PR #1772, opened 2026-02-05
feat: [WD-32423] Add ACL default action for Instance NICs

Kxiru:
## Done
Add ACL default action for Instance NICs
~~Currently NON-FUNCTIONAL~~

- [x] Show ACL default action selector in Instance NIC side panel
- [x] Be able to edit default action selector (DAS)
- [x] Be able to display accurate values in DAS
- [x] Show read-only DAS accurately
- [x] Disable DAS for non-OVN networks (but still fetch data on bridge networks)

## QA

1. Run the LXD-UI:
    - Create an OVN network X
    - Create an ACL for OVN network rejecting everything and assign to OVN network
    - Create an instance and add custom network X to it.
    - Edit the custom network to have ACL default values that are NOT the same as those assigned to the network, such as Allow.
    - Save, and verify that the changes are reflected in the Read-Only and the YAML configuration.

## Screenshots

image

### DIFF
--- a/src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
+++ b/src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
@@ -32,8 +32,12 @@ import {
   supportsNicDeviceAcls,
   getNetworkAcls,
   combineAcls,
+  ovnType,
 } from "util/networks";
 import usePanelParams, { panels } from "util/usePanelParams";
+import NetworkDefaultACLSelector, {
+  type Direction,
+} from "pages/networks/forms/NetworkDefaultACLSelector";
 import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
 import type { NetworkDeviceFormValues } from "types/forms/networkDevice";
 
@@ -161,6 +165,10 @@ const NetworkDevicePanel: FC<Props> = ({
       acls: combinedAcls,
       ipv4: device?.["ipv4.address"] || "",
       ipv6: device?.["ipv6.address"] || "",
+      security_acls_default_egress_action:
+        device?.["security.acls.default.egress.action"] || "",
+      security_acls_default_ingress_action:
+        device?.["security.acls.default.ingress.action"] || "",
     };
   };
 
@@ -185,6 +193,10 @@ const NetworkDevicePanel: FC<Props> = ({
           userSelectedAcls.length > 0 ? userSelectedAcls.join(",") : undefined,
         "ipv4.address": values.ipv4 || undefined,
         "ipv6.address": values.ipv6 || undefined,
+        "security.acls.default.egress.action":
+          values.security_acls_default_egress_action || undefined,
+        "security.acls.default.ingress.action":
+          values.security_acls_default_ingress_action || undefined,
       };
 
       const originalDeviceName = deviceName;
@@ -265,6 +277,25 @@ const NetworkDevicePanel: FC<Props> = ({
     );
   }
 
+  const getDefaultEgressIngress = () => {
+    // Return NIC device default egress/ingress actions if set, fallback to network defaults if not, or rejected if neither are set
+    return {
+      Egress:
+        formik.values.security_acls_default_egress_action ??
+        selectedNetwork?.config["security.acls.default.egress.action"] ??
+        "",
+      Ingress:
+        formik.values.security_acls_default_ingress_action ??
+        selectedNetwork?.config["security.acls.default.ingress.action"] ??
+        "",
+    };
+  };
+
+  const directionField: Record<Direction, string> = {
+    Egress: "security_acls_default_egress_action",
+    Ingress: "security_acls_default_ingress_action",
+  };
+
   return (
     <SidePanel>
       <SidePanel.Header>
@@ -306,18 +337,6 @@ const NetworkDevicePanel: FC<Props> = ({
               error={formik.touched.network ? formik.errors.network : undefined}
             />
 
-            <NetworkAclSelector
-              project={project}
-              selectedAcls={selectedAcls}
-              setSelectedAcls={(selectedItems) => {
-                void formik.setFieldValue("acls", selectedItems.join(","));
-              }}
-              inheritedAcls={networkAcls}
-              canSelectManualAcls={supportsNicDeviceAcls(selectedNetwork)}
-              help={getAclHelperText()}
-              label="ACLs"
-            />
-
             {isInstance && selectedNetwork && (
               <>
                 <NetworkDeviceIPAddressEdit
@@ -332,6 +351,30 @@ const NetworkDevicePanel: FC<Props> = ({
                 />
               </>
             )}
+
+            <NetworkAclSelector
+              project={project}
+              selectedAcls={selectedAcls}
+              setSelectedAcls={(selectedItems) => {
+                void formik.setFieldValue("acls", selectedItems.join(","));
+              }}
+              inheritedAcls={networkAcls}
+              canSelectManualAcls={supportsNicDeviceAcls(selectedNetwork)}
+              help={getAclHelperText()}
+              label="ACLs"
+            />
+
+            <NetworkDefaultACLSelector
+              onChange={(fieldValue, value) => {
+                void formik.setFieldValue(fieldValue, value);
+              }}
+              values={getDefaultEgressIngress()}
+              disabled={
+                formik.values.acls?.length === 0 ||
+                selectedNetwork?.type !== ovnType
+              }
+              directionField={directionField}
+            />
           </Form>
         </ScrollableContainer>
       </SidePanel.Content>
--- a/src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
+++ b/src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
@@ -13,6 +13,7 @@ import { combineAcls, getNetworkAcls } from "util/networks";
 import { getDeviceAcls } from "util/devices";
 import NetworkRichChip from "pages/networks/NetworkRichChip";
 import { ROOT_PATH } from "util/rootPath";
+import NetworkDefaultACLRead from "pages/networks/forms/NetworkDefaultACLRead";
 
 const getNetworkDeviceIpAddress = ({
   network,
@@ -86,9 +87,9 @@ export const getNetworkDeviceRows = ({
   if (device.type === "custom-nic") {
     rows.push(
       getConfigurationRowBase({
-        className: "no-border-top",
+        className: "no-border-top device-last-row",
         configuration: (
-          <span
+          <div
             className={classnames({
               "u-text--line-through": isDetached,
             })}
@@ -97,7 +98,7 @@ export const getNetworkDeviceRows = ({
             <Tooltip message="A custom network can be viewed and edited only from the YAML configuration">
               <Icon name="information" />
             </Tooltip>
-          </span>
+          </div>
         ),
         inherited: null,
         override: null,
@@ -129,12 +130,45 @@ export const getNetworkDeviceRows = ({
     );
 
     const acls = combineAcls(getNetworkAcls(network), getDeviceAcls(device));
+
+    if (showIpAddresses) {
+      const families = ["IPv4", "IPv6"] as const;
+      const activeIps = families
+        .map((family) => ({
+          family,
+          ip: getNetworkDeviceIpAddress({ network, device, family }),
+        }))
+        .filter((item) => !!item.ip);
+
+      activeIps.forEach(({ family, ip }, index) => {
+        rows.push(
+          getConfigurationRowBase({
+            className: classnames("no-border-top", {
+              "device-last-row":
+                acls.length === 0 && index === activeIps.length - 1,
+            }),
+            configuration: <div className="u-text--muted">{family}</div>,
+            inherited: (
+              <div>
+                <b
+                  className={classnames("mono-font", {
+                    "u-text--line-through": isDetached,
+                  })}
+                >
+                  {ip}
+                </b>
+              </div>
+            ),
+            override: null,
+          }),
+        );
+      });
+    }
+
     if (acls.length > 0) {
       rows.push(
         getConfigurationRowBase({
-          className: classnames("no-border-top", {
-            "device-last-row": !showIpAddresses,
-          }),
+          className: "no-border-top",
           configuration: <div className="u-text--muted">ACLs</div>,
           inherited: (
             <div>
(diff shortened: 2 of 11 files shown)


src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -153,6 +154,26 @@ export const getNetworkDeviceRows = ({
...
+          inherited: (
+            <div>
+              <NetworkDefaultACLRead values={defaultEgressIngress} />
+            </div>
+          ),
+          override: null,
+        }),
+      );
+    }
+
```
edlerd: The layout is a bit bamboozled here.

- We should remove the "Default actions" label.
- We should make the "Egress traffic is: rejected" fit into a single line.
- Maybe this should be a bit closer to the ACL chip.
- Please consult with design to find a best fit here.

image

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -302,6 +324,18 @@ const NetworkDevicePanel: FC<Props> = ({
...
+                void formik.setFieldValue(fieldValue, value);
+              }}
+              values={defaultEgressIngress}
+              disabled={
+                formik.values.acls?.length === 0 ||
+                selectedNetwork?.type !== ovnType
+              }
+              directionField={directionField}
+            />
+
```
edlerd: Let's switch the order around. I think a user is more likely to want to customize ip addresses than the ACLs:
- Show the ipv4 and ipv6 after description, then the ACL and ACL defaults after that.
- Also in read mode, lets bring the ipv4/ipv6 up and list the ACL last.


image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -153,6 +154,26 @@ export const getNetworkDeviceRows = ({
...
+        Ingress: device["security.acls.default.ingress.action"] ?? "",
+      };
+
+      rows.push(
+        getConfigurationRowBase({
+          className: "no-border-top",
+          configuration: <div className="u-text--muted">Default actions</div>,
+          inherited: (
+            <div>
+              <NetworkDefaultACLRead values={defaultEgressIngress} />
```
edlerd: There is some logic missing here. When no defaults are selected on the NIC, the defaults from the network are applied.

If the network has defaults as `egress: allowed` and `ingress:dropped` and the NIC has no selections/override for the defaults, we should show the network defaults. Currently we fall back to always showing `rejected` in this case. The network defaults must be considered as a fallback before falling back to the lxd default of `rejected`.

So we have three layers of precedence:
1. nic default, then
2. network default, then
3. lxd default.

And I think this 3 layered default hierarchy needs to be considered for each ingress and egress individually.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
Kxiru: Can't we increase the inherited column a little? There's so much white space to the left of the actions column.

edlerd: Yes, that could be an option. Or reduce the first columns width.

src/sass/_forms.scss
```diff
@@ -213,7 +213,7 @@
     }
 
     .inherited {
-      width: 11rem;
+      width: 13rem;
```
edlerd: We can't do this globally as it affects other pages and breaks the layout there. We should do this only for the network form.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -185,6 +156,68 @@ export const getNetworkDeviceRows = ({
...
+          configuration: <div className="u-text--muted"></div>,
+          inherited: (
+            <div>
+              <NetworkDefaultACLRead values={getDefaultEgressIngress()} />
+            </div>
+          ),
+          override: null,
+        }),
+      );
+    }
```
edlerd, line 223: The spacing here needs refinement.
- between ipv6 and acl should be less, it should match the space between ipv4 and ipv6
- I think the "When no ACL rule matches:" should be closer to the ACL chip. Not 100% sure on this one though.

image

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
edlerd, line 1: This is now broken, setting ACL defaults where the network already has defaults applied does not work.

[Screencast from 2026-02-06 18-22-19.webm](https://github.com/user-attachments/assets/dfb21006-a701-4893-b76f-5a5fc110049f)

Kxiru: This is the fallback logic requested. If a new value is not set, it defaults back to the network's default as you suggested.

Kxiru: I think there is a small lag. For me the lag happens locally and it takes a few seconds to change the input, but it does change the value. Otherwise, I cannot recreate your issue. Can we sync on this?

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
Kxiru: I have addressed the IPV6 vs ACL spacing (though this spacing existed before this PR and was based on a conditional which I have now removed so the spacing never applies:

` "device-last-row": index === activeIps.length - 1,`

As for the "When No ACL matches", this logic is in a new configuration row, and therefore the default styling is applied. I could put the Read component inside the ACL row beneath the ACL list? But I do believe some visual separation is required. Let me know what you think.

image

src/sass/_forms.scss
```diff
@@ -220,6 +220,12 @@
     .p-form-help-text > * {
       margin-bottom: 0;
     }
+
+    .network-device-form {
+      .inherited {
+        width: 13rem;
+      }
+    }
```
edlerd: This has no effect, and the previous bug with the ACL defaults set to rejected, breaking into two lines, is now surfacing again on the demo server

Image

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
edlerd: This still reproduces for me exactly like in the video. Please try the following locally:
- OVN network with acl defaults applied
- add the network to an instance, try setting different acl defaults on the instance

The change should be applied to the form. Can you try reproducing it locally?

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
edlerd: Removing the `device-last-row` is messing up the spacing. We need to have it on the last row of the device to fix it.

1. Keep `device-last-row` on the IP if no ACLs are displayed.
2. Add the `device-last-row` to the ACL defaults, don't show it on IPs if ACLs are displayed.
3.  Add the `device-last-row` to the custom network. (This is a bug introduced recently and it would be nice to fix it as a drive by in this PR)

edlerd: Regarding the ACL defaults: We can give that cell a class and set padding-top to 0 to fix the spacing between ACL and ACL defaults.

Kxiru: Could you clarify what you mean when you ask for the row to be applied to the custom network? Empty state, or the literal custom network line? Eg; ovnnetwork in the below screenshot:

image

This ss also have stylings 1 and 2 you requested.

edlerd: See below from [demo server](https://lxd-ui-307.demos.haus/ui/project/default/instance/DemoInstance-copy/configuration/network), eth0 is a custom network and has too few bottom padding.

To reproduce it, got to yaml editor and add a custom configuration to the nic device. Like `user.foo: bar` in the demo example.

image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -190,6 +163,67 @@ export const getNetworkDeviceRows = ({
...
+          return {
+            Egress: device["security.acls.default.egress.action"],
+            Ingress: device["security.acls.default.ingress.action"],
+          };
+        }
+
+        return {
+          Egress: network.config["security.acls.default.egress.action"] ?? "",
+          Ingress: network.config["security.acls.default.ingress.action"] ?? "",
+        };
```
edlerd: This combined logic is wrong. Both fields can be overridden independently. See example from the screen below. In the read mode it should say "egress is allowed".

Also the spacing issue is visible here. The read mode defaults should fit into a single line.

Image

Kxiru: Yes, I needed to add the logic to the read mode as well, thanks for catching that. This has now been rectified.

As for the design issue though, looking into that...

Kxiru: Addressed!

src/pages/networks/forms/NetworkAcls.tsx
edlerd, line 1: Not related to your changes in here, but a small design fix: Can you please change the `NetworkAclSelector.tsx` to not mute the label, when it is disabled: `{label && <label htmlFor={id}>{label}</label>}`. We should not mute the label for disabled fields.

Image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -190,6 +163,63 @@ export const getNetworkDeviceRows = ({
...
+          Ingress:
+            device["security.acls.default.ingress.action"] ??
+            network.config["security.acls.default.ingress.action"] ??
+            "",
+        };
+      };
+
+      rows.push(
+        getConfigurationRowBase({
+          className: "no-border-top device-last-row",
```
edlerd:
```suggestion
          className: "no-border-top device-last-row acl-defaults",
```
Let's give this an additional class to optimize the spacing. This makes the connection to the acls more clear with custom css, adding to `_network_device_form.scss`:
```
.network-device-form {

  [...]
  .acl-defaults .inherited {
    padding-top: 0;
  }
}
```

Image

edlerd (commented):
QA and code LGTM, good work 👍 

Two tiny suggestions to optimize the design, and then this is good to merge from my side.

src/sass/_forms.scss
```diff
@@ -261,6 +265,10 @@
     .device-name {
       margin-left: 0 !important;
     }
+
+    .inherited {
+      width: 13rem;
+    }
```
edlerd: We need to scope those changes to only apply to the network page. We can use the parent `.network-device-form` for limiting the scope.

As is, this change alters the spacing for all other devices. Include the disks, and then the different sections in it are misaligned.

Image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -190,6 +163,63 @@ export const getNetworkDeviceRows = ({
         );
       });
     }
+
+    if (acls.length > 0) {
+      rows.push(
+        getConfigurationRowBase({
+          className: classnames("no-border-top", {
+            "device-last-row": !showIpAddresses,
+          }),
```
edlerd:
```suggestion
          className: "no-border-top",
```
This is never the last row, there will always be the acl defaults below.

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -171,7 +144,7 @@ export const getNetworkDeviceRows = ({
         rows.push(
           getConfigurationRowBase({
             className: classnames("no-border-top", {
-              "device-last-row": index === activeIps.length - 1,
+              "device-last-row": !acls && index === activeIps.length - 1,
```
edlerd:
```suggestion
              "device-last-row": acls.length === 0 && index === activeIps.length - 1,
```
otherwise the spacing is a little too narrow.

src/sass/_network_device_form.scss
```diff
@@ -1,4 +1,15 @@
 .network-device-form {
+  .acl-defaults .inherited {
+    padding-top: 0;
+  }
+
+  .custom-devices,
+  .inherited-devices {
+    .inherited {
+      width: 13rem;
```
edlerd:
```suggestion
      width: 12rem;
```
Otherwise the custom network looks wired with the icon falling to the next line.

Image

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
edlerd, line 94: Let's replace the custom device with the below, so it applies the right spacing below
```suggestion
  if (device.type === "custom-nic") {
    rows.push(
      getConfigurationRowBase({
        className: "no-border-top device-last-row",
        configuration: (
          <div
            className={classnames({
              "u-text--line-through": isDetached,
            })}
          >
            custom network{" "}
            <Tooltip message="A custom network can be viewed and edited only from the YAML configuration">
              <Icon name="information" />
            </Tooltip>
          </div>
        ),
        inherited: null,
        override: null,
      }),
    );
    return rows;
  }
```

src/components/forms/NetworkDevicesForm/read/NetworkDeviceRows.tsx
```diff
@@ -190,6 +164,61 @@ export const getNetworkDeviceRows = ({
...
+          Ingress:
+            device["security.acls.default.ingress.action"] ??
+            network.config["security.acls.default.ingress.action"] ??
+            "",
+        };
+      };
+
+      rows.push(
+        getConfigurationRowBase({
+          className: "no-border-top device-last-row acl-defaults",
```
edlerd:
```suggestion
          className: classnames("no-border-top device-last-row acl-defaults", {
            "u-text--line-through": isDetached,
          }),
```
When this is inherited but detached, we should strike it out.

Image

Kxiru: Okay, new addition, noted... In your screenshot, why is only Egress detached? If a NIC is detached at the very least shouldn't both ingress and egress be detached?
I'll assume this is a testing screenshot of when you were experimenting, as the logic you have provided with scour both when detached :+1:

PR #1785, opened 2026-02-11
fix(rich-chips) show rich chips on the right

edlerd:
## Done

- fix(rich-chips) show rich chips on the right

## QA

1. Run the LXD-UI:
    - check rich tooltip positions

Unblocked with https://github.com/canonical/react-components/pull/1314

### DIFF
--- a/src/pages/cluster/ClusterMemberRichTooltip.tsx
+++ b/src/pages/cluster/ClusterMemberRichTooltip.tsx
@@ -1,16 +1,11 @@
 import { type FC } from "react";
 import { Spinner } from "@canonical/react-components";
 import { type TooltipRow } from "components/RichTooltipRow";
-import {
-  LARGE_TOOLTIP_BREAKPOINT,
-  MEDIUM_TOOLTIP_BREAKPOINT,
-  RichTooltipTable,
-} from "components/RichTooltipTable";
+import { RichTooltipTable } from "components/RichTooltipTable";
 import { useClusterMember } from "context/useClusterMembers";
 import { Link } from "react-router-dom";
 import ItemName from "components/ItemName";
 import ClusterMemberStatus from "./ClusterMemberStatus";
-import { useIsScreenBelow } from "context/useIsScreenBelow";
 import ClusterMemberMemoryUsage from "./ClusterMemberMemoryUsage";
 import { useQuery } from "@tanstack/react-query";
 import { fetchClusterMemberState } from "api/cluster-members";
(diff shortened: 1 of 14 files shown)


src/pages/cluster/ClusterMemberRichTooltip.tsx
```diff
@@ -86,39 +77,29 @@ const ClusterMemberRichTooltip: FC<Props> = ({ clusterMember }) => {
...
-      },
-    );
-  }
-
-  if (isAboveLarge) {
-    rows.push({
+    {
+      title: "URL",
+      value: member ? member.url : "-",
+      valueTitle: member ? member.url : "URL",
```
omarelkashef: If there is no url, there is no need for a value title.

src/pages/cluster/ClusterMemberRichTooltip.tsx
```diff
@@ -86,39 +77,29 @@ const ClusterMemberRichTooltip: FC<Props> = ({ clusterMember }) => {
...
-        value: member ? <ClusterMemberMemoryUsage member={member} /> : "-",
-      },
-    );
-  }
-
-  if (isAboveLarge) {
-    rows.push({
+    {
+      title: "URL",
+      value: member ? member.url : "-",
```
omarelkashef: Is it possible to have an empty url? In that case we will have an empty value which we can avoid by using `member?.url || "-"`

src/pages/cluster/ClusterMemberRichTooltip.tsx
edlerd: Good catch, this was an overlook on previous reviews when the rich tooltip for cluster members was built.

omarelkashef:
> > I like the scrollbar. It looks good and avoids the complicated breakpoints logic. However, when it is the first time for the tooltip to render, it can be truncated by the right side of the screen. Maybe we can give a min-width and breakpoints to guarantee that enough horizontal space is reserved.
> 
> I don't understand exactly what you mean. The `rich-tooltip-table` has a width and reserves space for the scrollbar due to the new `scrollbar-gutter: stable` property. So it should always render similarly -- first time or the nth time it is opened, or rendered.

My assumption is that the data is not ready at first so the tooltip assumes it can fit on the right, but later the data is cached and the tooltips knows how much space it will take so it is rendered on the left.
[Screencast from 2026-02-18 14-36-31.webm](https://github.com/user-attachments/assets/25a1403c-b7d9-4760-bf34-40cf6e5a5a2a)

edlerd:
> My assumption is that the data is not ready at first so the tooltip assumes it can fit on the right, but later the data is cached and the tooltips knows how much space it will take so it is rendered on the left.
> [Screencast from 2026-02-18 14-36-31.webm](https://github.com/user-attachments/assets/25a1403c-b7d9-4760-bf34-40cf6e5a5a2a)

I was able to reproduce it, but think this is an edge case with the chrome mobile view. See the screen below -- on first render the `window.outerWidth` is correct, where the `window.innerWidth` is much higher. Which seems like a broken edge case:

image

When using the non-mobile view in chrome with the similar window size, both values get calculated correctly and the tooltip gets positioned correctly also on first render:

image

PR #1793, opened 2026-02-16
fix(instance) rely on instance name in operations new entity_url field

edlerd:
## Done

- fix(instance) rely on instance name in operations new entity_url field

## QA

1. Run the LXD-UI:
    - rely on ci to pass

### DIFF
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -192,7 +192,7 @@ const InstanceList: FC = () => {
       const isCreating = operation.description === "Creating instance";
       const isProcessing = processingNames.includes(name);
       const isMigrating = migrationNames.includes(name);
-      if (!isCreating || isProcessing || isMigrating) {
+      if (!isCreating || isProcessing || isMigrating || !name) {
         return false;
       }
       const isInInstanceList = instances.some((item) => item.name === name);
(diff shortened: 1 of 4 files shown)


kimanhou:
Should we not display the chip if the name is missing to avoid what's happening in snapshot creation notification ?

```
import { Link } from "react-router-dom";
import type { ResourceIconType } from "./ResourceIcon";
import ResourceIcon from "./ResourceIcon";

interface Props {
  type: ResourceIconType;
  value: string;
  to: string;
  disabled?: boolean;
  hasTitle?: boolean;
  className?: string;
}

const ResourceLink: FC<Props> = ({
  type,
  value,
  to,
  disabled,
  hasTitle = true,
  className,
}) => {
  return value ? (
    <Link
      className={classnames(
        "p-chip is-inline is-dense resource-link",
        {
          "p-chip--disabled": disabled,
        },
        className,
      )}
      to={to}
      title={hasTitle ? value : undefined}
    >
      <ResourceIcon type={type} />
      <span className="p-chip__value">{value}</span>
    </Link>
  ) : null;
};

export default ResourceLink;
```

image

edlerd:
> Both are due to the notification not having the name of the snapshot. Should we merge this PR as fix is on backend side ?

Yes, the fix is already in and I manually validated with a custom built LXD backend. Just waiting for the LXD edge build then it should be fixed. Merging it right away.

PR #1797, opened 2026-02-17
feat: add validation for nic static ip and pull upstream components W…

omarelkashef:
## Done

- Pull PrefixedIpInput from react-components
- Adds validation to NIC static ip input

## QA

1. Run the LXD-UI:
    - Go to {instance_name} -> Configuration -> Network
    - Attach/edit device
    - Make sure input is disabled/enabled according to the following rules with appropriate help text:
         - Network type must be supported - Only bridge and ovn network types support static IPs
         - Network must be managed 
         - Network must be in "Created" state
         - Bridge Network Rules
            -  IPv4:
               - DHCP is enabled
               - Network must have IP configured - Cannot set static IP when network's ipv4.address is "none" or undefined
           - IPv6:
             - DHCP and stateful DHCP
             - Network must have IP configured - Cannot set static IP when network's ipv6.address is "none" or undefined
         - OVN Network Rules
            - IPv4:
               - DHCP must be enabled
            - IPv6:
               - DHCP and stateful DHCP are enabled
               - Nic static IPv4 address must be set first
       -Make sure when input is enabled, it rejects invalid IPs

## Screenshots

image
image


src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -1,45 +1,55 @@
...
+  const { isAllowed: isStaticIpAllowed, reason } =
+    getIsNicStaticIpAllowedAndReason(formik.values, network, family, device);
 
-  const isEnabled =
-    network?.managed &&
-    typesWithNicStaticIPSupport.includes(network.type) &&
-    ipAddressConfigValue !== "none" &&
-    ipAddressConfigValue !== undefined;
   return (
     <PrefixedIpInput
```
edlerd, line 53: Can we change the placeholder/prefix to not show anything, or a "-" in case the field is disabled?

See the ipv6 in the below screen with a very busy placeholder, even though it is disabled. We might fall back to a "normal" text input for the disabled case to get the "-" placeholder rendered.

Image

src/util/devices.tsx
```diff
@@ -11,7 +11,33 @@ import type {
...
+import type { LxdConfigOptionsKeys } from "types/config";
+import { toConfigFields } from "./config";
+import { getLxdDefault, getConfigOptions } from "./configInheritance";
+
+interface NicStaticIpSupport {
+  isAllowed: boolean;
+  reason: string;
+}
+
+const nicTypeToOptionKey: Record<string, LxdConfigOptionsKeys> = {
```
edlerd: There is a bug on this page: https://lxd-ui-1797.demos.haus/ui/project/default/instance/DemoInstance/configuration/network?panel=edit-network-device&device-name=eth-6

Image

src/util/devices.tsx
```diff
@@ -169,3 +195,231 @@ export const getProfileFromSource = (source: string) => {
...
+  networkDHCPStateful: boolean,
+  ipFiltering: boolean,
+  networkCIDR: string | undefined,
+  family: "IPv4" | "IPv6",
+): NicStaticIpSupport => {
+  if (family === "IPv4") {
+    if (!networkDHCP && !ipFiltering) {
+      return {
+        isAllowed: false,
+        reason: `Cannot specify "ipv4.address" when DHCP is disabled (unless using security.ipv4_filtering).`,
```
```suggestion
        reason: `Cannot specify "ipv4.address" when DHCP is disabled.`,
```
Let's not mention unsupported fields in the UI.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
omarelkashef: Good point! It looks cluttered for no reason now.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -16,30 +17,61 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
...
+      : ["ipv6_dhcp", "ipv6_dhcp_stateful"],
+  );
   const networkFieldName = family === "IPv4" ? "ipv4.address" : "ipv6.address";
-  const ipAddressConfigValue = network.config?.[networkFieldName];
-  const formikFieldName = family.toLowerCase() as keyof NetworkDeviceFormValues;
+  const formFieldName = family.toLowerCase() as keyof NetworkDeviceFormValues;
+  const networkCIDR = network.config?.[networkFieldName];
+
+  const nicIP =
+    formik.values[family.toLowerCase() as keyof NetworkDeviceFormValues];
```
edlerd:
```suggestion
    formik.values[formFieldName];
```

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -16,30 +17,61 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
...
 
-  const isEnabled =
-    network?.managed &&
-    typesWithNicStaticIPSupport.includes(network.type) &&
-    ipAddressConfigValue !== "none" &&
-    ipAddressConfigValue !== undefined;
+  const { isAllowed: isStaticIpAllowed, reason } =
+    getIsNicStaticIpAllowedAndReason(
```
edlerd:
```suggestion
  const disableReason =
    getIsNicStaticIpAllowedAndReason(
```
Let's simplify the interface here so the function only returns a string or an empty string. The `isStaticIpAllowed` we can evaluate here by comparing the `disableReason` to an empty string locally.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -16,30 +17,61 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
     return null;
   }
 
+  const [dhcpDefault, dhcpStatefulDefault] = getNetworkMetadata(
+    network.type,
+    family === "IPv4"
+      ? ["ipv4_dhcp", "ipv6_dhcp_stateful"]
+      : ["ipv6_dhcp", "ipv6_dhcp_stateful"],
+  );
```
edlerd:
```suggestion
```
We seem to only need this inside of `getIsNicStaticIpAllowedAndReason`. And all the inputs for this call are already passed to the function. So let's move this function call inside `getIsNicStaticIpAllowedAndReason`.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -16,30 +17,61 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
...
+      dhcpStatefulDefault.value,
+    );
+
+  if (!isStaticIpAllowed) {
+    return (
+      <>
+        <label
+          htmlFor={networkFieldName}
+          className="u-text--muted"
+        >{`${family} address reservation`}</label>
```
edlerd:
```suggestion
```
Let's use the `label` and `labelClassName` props of the `Input` component below.

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+        "Cannot specify a static IP when the network CIDR is set to 'none' or is undefined.",
+    };
+  }
+
+  if (!networkDHCP || (family === "IPv6" && !networkDHCPStateful)) {
+    return {
+      isAllowed: false,
+      reason:
+        family === "IPv4"
+          ? `Cannot specify "ipv4.address" when DHCP is disabled.`
```
```suggestion
          ? `IPv4 DHCP is disabled on the network.`
```
Is this clear enough?

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+      isAllowed: false,
+      reason: "Static IPs can only be set on networks in the created state.",
+    };
+  }
+
+  if (networkCIDR === "none" || networkCIDR === undefined) {
+    return {
+      isAllowed: false,
+      reason:
+        "Cannot specify a static IP when the network CIDR is set to 'none' or is undefined.",
```
```suggestion
        `${family} is disabled on the network.`
```

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+    return {
+      isAllowed: false,
+      reason: "Static IPs are only supported for managed networks.",
+    };
+  }
+
+  if (network.status !== "Created") {
+    return {
+      isAllowed: false,
+      reason: "Static IPs can only be set on networks in the created state.",
```
```suggestion
      reason: "Network is not in status <code>created</code>.",
```

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+    return {
+      isAllowed: false,
+      reason: `Static IPs are only supported for ${typesWithNicStaticIPSupport.join(", ")} networks.`,
+    };
+  }
+
+  if (!network.managed) {
+    return {
+      isAllowed: false,
+      reason: "Static IPs are only supported for managed networks.",
```
```suggestion
      reason: "Network is not managed.",
```

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+  const networkDHCPStateful = getNetworkDHCPStatefulOrDefault(
+    network,
+    family,
+    dhcpStatefulDefault,
+  );
+
+  if (!typesWithNicStaticIPSupport.includes(network.type)) {
+    return {
+      isAllowed: false,
+      reason: `Static IPs are only supported for ${typesWithNicStaticIPSupport.join(", ")} networks.`,
```
```suggestion
      reason: `Network must of type ${typesWithNicStaticIPSupport.join(" or ")}.`,
```

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+          : `Cannot specify "ipv6.address" when DHCP or DHCP Stateful are disabled.`,
+    };
+  }
+
+  if (network.type === "ovn" && family === "IPv6" && !isStaticIpv4) {
+    return {
+      isAllowed: false,
+      reason: `Cannot specify "ipv6.address" when "ipv4.address" is not set.`,
+    };
+  }
```
edlerd, line 234: This seems odd. Can't OVN networking be IPv6 only?

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+  network: LxdNetwork,
+  family: "IPv4" | "IPv6",
+  defaultValue?: string,
+): boolean => {
+  const defaultVal = defaultValue === "true" ? "true" : "false"; // any non-true value should be treated as false
+  const result =
+    family === "IPv4"
+      ? (network.config["ipv4.dhcp"] ?? defaultVal)
+      : (network.config["ipv6.dhcp"] ?? defaultVal);
+  return result === "true";
```
```suggestion
  const result =
    family === "IPv4"
      ? (network.config["ipv4.dhcp"] ?? defaultValue)
      : (network.config["ipv6.dhcp"] ?? defaultValue);
  return result === "true";
```
Do we need the indirection here?

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
...
+    };
+  }
+
+  if (!networkDHCP || (family === "IPv6" && !networkDHCPStateful)) {
+    return {
+      isAllowed: false,
+      reason:
+        family === "IPv4"
+          ? `Cannot specify "ipv4.address" when DHCP is disabled.`
+          : `Cannot specify "ipv6.address" when DHCP or DHCP Stateful are disabled.`,
```
```suggestion
          : `IPv6 DHCP or IPv6 DHCP stateful are disabled on the network.`,
```

src/util/devices.tsx
```diff
@@ -169,3 +184,96 @@ export const getProfileFromSource = (source: string) => {
 
   return source.split(" profile")[0];
 };
+
+export const getIsNicStaticIpAllowedAndReason = (
```
```suggestion
export const getStaticIpDisableReason = (
```

src/util/configInheritance.tsx
```diff
@@ -158,16 +158,29 @@ const getNetworkRowMetadata = (
   values: NetworkFormValues,
   name: string,
 ): ConfigRowMetadata => {
+  return getNetworkMetadata(values.networkType, [name])[0];
+};
+
+export const getNetworkMetadata = (
+  networkType: string,
+  name: string[],
```
edlerd:
```suggestion
  name: string,
```
I would keep this simple and call it twice for ipv6 dhcp and ipv6 dhcp stateful. It seems much simpler and currently we call it for ipv4 and ipv6 dhcp stateful even though we only need the former.

src/util/devices.tsx
omarelkashef: According to https://github.com/canonical/lxd/blob/main/lxd/device/nic_ovn.go#L186-L189 , static ipv4 has to be set first.

src/util/devices.tsx
omarelkashef: Are you suggesting to change the return type of the function to a react node so that <code></code> is rendered correctly?

src/util/devices.tsx
edlerd: I see, then let's improve a bit on the message and use the labels visible in this screen. Something like "Enter an IPv4 address reservation to enable this field.". Wdyt?

src/util/devices.tsx
edlerd: Ah, I didn't think of it, but maybe that is the right thing to be able to use the `<code>` for highlighting. Wdyt?

src/util/devices.tsx
omarelkashef: All the other reasons are statement and not directions. For the sake of consistency, we can have something like "Static IPv6 is allowed only if static IPv4 is set as well" or "Static IPv4 must be set first to enable static IPv6".

src/util/devices.tsx
omarelkashef: Yes, good idea!

src/util/devices.tsx
edlerd: Right, I like the slight variation of your suggestion: "IPv4 address reservation must be set to enable this field.". Wdyt?

omarelkashef: Yes, that reads nicely.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -17,29 +18,54 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
...
-    ipAddressConfigValue !== undefined;
+  if (disableReason) {
+    return (
+      <Input
+        id={networkFieldName}
+        label={`${family} address reservation`}
+        labelClassName="u-text--muted"
+        type="text"
+        disabled
+        help={disableReason}
```
edlerd, line 47:
```suggestion
        title={disableReason}
```
We tend to use the title and not help text for this in other places. Wdyt?

src/util/devices.spec.ts
```diff
@@ -148,3 +158,345 @@ describe("getIndex", () => {
...
+
+    it("should reject networks not in Created state", () => {
+      const network = mockNetwork("test", "bridge", true, "Pending");
+      const formValues = mockNetworkFormValues();
+      const result = getNicStaticIpDisableReason(formValues, network, "IPv4");
+      const props = result?.props as React.PropsWithChildren;
+      if (Array.isArray(props.children)) {
+        console.log(props.children[1]);
+      }
+
```
edlerd:
```suggestion
```
Leftover from debugging?

src/util/devices.spec.ts
```diff
@@ -148,3 +158,345 @@ describe("getIndex", () => {
...
+          }
+          return "";
+        })
+        .join("");
+    }
+    if (React.isValidElement(children as object)) {
+      return extractReasonText(children as React.ReactElement);
+    }
+    return "";
+  };
```
edlerd: This seems like a useful helper for other tests. Do you want to raise this up and export it?

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
omarelkashef: I am not sure it will work. We changed disableReason to be a reactnode and title expects a string.

edlerd: Ah right, then lets keep it as is.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -17,29 +18,54 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
...
-    network?.managed &&
-    typesWithNicStaticIPSupport.includes(network.type) &&
-    ipAddressConfigValue !== "none" &&
-    ipAddressConfigValue !== undefined;
+  if (disableReason) {
+    return (
+      <Input
+        id={networkFieldName}
+        label={`${family} address reservation`}
+        labelClassName="u-text--muted"
```
edlerd:
```suggestion
```
We should never mute the label. This idea crept in with the ACL config right above, but we'll remove the muting there in another PR. Let's not mute the label here either.

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
```diff
@@ -17,29 +18,54 @@ const NetworkDeviceIPAddressEdit: FC<Props> = ({ formik, network, family }) => {
...
+        type="text"
+        disabled
+        help={disableReason}
+        value="-"
+      />
+    );
+  }
   return (
     <PrefixedIpInput
```
edlerd, line 53: The help text for IPv6 seems wrong, as it is not mentioning a range. I am not sure how to resolve it either. Do we even need help text here, or should we set it to empty?

Image

src/util/devices.tsx
```diff
@@ -169,3 +179,83 @@ export const getProfileFromSource = (source: string) => {
 
   return source.split(" profile")[0];
 };
+
+export const getNicStaticIpDisableReason = (
```
```suggestion
export const getNicIpDisableReason = (
```
A bit simpler?

src/components/forms/NetworkDevicesForm/edit/NetworkDeviceIPAddressEdit.tsx
omarelkashef: I think we should remove it completely. There is not intuitive way to display ipv6 range like for ipv4 as it is not a range of integers.

edlerd: Sounds good to me.

PR #1810, opened 2026-02-20
chore(ci) fix ovn setup

edlerd:


.github/workflows/coverage.yaml
```diff
@@ -32,6 +32,11 @@ jobs:
           sudo chmod 0777 ../lxd-ui
           dotrun install
 
+      - name: OVN dependencies
+        run: |
+          sudo snap refresh snapd --channel latest/beta || sudo snap install snapd --channel latest/beta
+          sudo snap refresh core26 --channel latest/edge || sudo snap install core26 --channel latest/edge
```
simondeziel, line 38: Is installing `core26` "manually" required? It certainly doesn't hurt but I'd expect it to be done automatically when `microovn` is installed.

edlerd: Without this, was [failing before](https://github.com/canonical/lxd-ui/actions/runs/22230631747/job/64308955671).

image

simondeziel: Ah, probably due to needing to pull from the edge channel, thanks! I'll incorporate that in the LXD PR too.

PR #1815, opened 2026-02-23
chore: fix screenshot workflow WD-33136

omarelkashef:
## Done

- fix screenshot workflow

## QA

1. Run the LXD-UI:
    - Run docs screenshot workflow and ensure everything passes.
Link to successful job run: https://github.com/omarelkashef/lxd-ui/actions/runs/22487790754/job/65141669207

### DIFF
--- a/tests/docs-screenshots.spec.ts
+++ b/tests/docs-screenshots.spec.ts
@@ -11,14 +11,16 @@ import {
   createNetworkLocalPeering,
   deleteLocalPeerings,
   deleteNetwork,
+  prepareCreateNetwork,
   submitCreateNetwork,
   visitNetwork,
+  createOvnUplink,
+  makeNetworkOvnUplink,
 } from "./helpers/network";
 import { createVolume, deleteVolume } from "./helpers/storageVolume";
 import { setOption } from "./helpers/configuration";
 import { getClipPosition } from "./helpers/doc-screenshots";
 import { openInstancePanel } from "./helpers/instancePanel";
-import { randomNameSuffix } from "./helpers/name";
 import { deleteNetworkAcl } from "./helpers/network-acls";
 
 test.beforeEach(() => {
@@ -58,7 +60,7 @@ test("instances", async ({ page }) => {
     clip: getClipPosition(240, 0, 1280, 675),
   });
 
-  await page.getByRole("button", { name: "Create override" }).click();
+  await page.getByTitle("Create override").click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/instances/create_instance_form.png",
     clip: getClipPosition(240, 0, 1360, 520),
@@ -72,8 +74,7 @@ test("instances", async ({ page }) => {
 
   await page.setViewportSize({ width: 1440, height: 500 });
   await page.getByRole("button", { name: "Attach custom volume" }).click();
-  await page.waitForSelector("text=Select");
-  await page.waitForTimeout(1000);
+  await page.getByRole("button", { name: "Select" }).first().waitFor();
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_attach_to_instance_3.png",
     clip: getClipPosition(85, 20, 1350, 470),
@@ -94,8 +95,11 @@ test("instances", async ({ page }) => {
     clip: getClipPosition(490, 70, 1245, 340),
   });
 
+  await page
+    .getByRole("button", { name: "Apply changes", exact: true })
+    .click();
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForSelector(`text=Created instance ${instance}.`);
+  await page.getByText(`Created instance ${instance}.`).click();
   await visitInstance(page, instance);
 
   await page.screenshot({
@@ -134,13 +138,7 @@ test("networks", async ({ page }) => {
   const networkACL = network + "-ACL";
   await page.setViewportSize({ width: 1440, height: 800 });
 
-  await gotoURL(page, "/ui/");
-  await page.getByText("Networking").click();
-  await page.getByText("Networks").click();
-  await page.getByText("Create network").click();
-  await page.getByPlaceholder("Enter name").fill(network);
-  await page.getByRole("button", { name: "Type" }).click();
-  await page.getByLabel("sub").getByText("bridge").first().click();
+  await prepareCreateNetwork(page, network, "bridge");
   await page.screenshot({
     path: "tests/screenshots/doc/images/networks/network_create.png",
     clip: getClipPosition(240, 0, 1420, 750),
@@ -154,7 +152,7 @@ test("networks", async ({ page }) => {
   });
 
   await page.getByRole("link", { name: "Leases" }).click();
-  await page.waitForSelector(`text=Hostname`);
+  await page.getByText("Hostname").waitFor();
   await page.screenshot({
     path: "tests/screenshots/doc/images/networks/network_view_leases.png",
     clip: getClipPosition(240, 0, 850, 250),
@@ -163,9 +161,7 @@ test("networks", async ({ page }) => {
   //Network ACLs
   await page.getByText("ACLs").click();
   await page.getByText("Create ACL").click();
-  await page.waitForSelector('[placeholder="Enter name"]', {
-    state: "visible",
-  });
+  await page.getByPlaceholder("Enter name").click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/networks/network_ACL_create.png",
     clip: getClipPosition(240, 0, 880, 360),
@@ -175,7 +171,7 @@ test("networks", async ({ page }) => {
   await page.getByRole("button", { name: "Create", exact: true }).click();
   await page.getByTestId("notification-close-button").click();
   await page.getByRole("link", { name: networkACL }).click();
-  await page.waitForSelector(`text=Click the ACL name`);
+  await page.getByText("Click the ACL name").waitFor();
   await page.screenshot({
     path: "tests/screenshots/doc/images/networks/network_ACLs.png",
     clip: getClipPosition(240, 0, 850, 590),
@@ -204,10 +200,11 @@ test("networks", async ({ page }) => {
 });
 
 test("network peering", async ({ page }) => {
-  const network = "OVNNetwork";
+  const uplinkNetwork = "BridgeNetwork1";
+  const network1 = "OVNNetwork";
   const network2 = "OVNNetwork2";
   const network3 = "OVNNetwork3";
-  const networkACL = network + "-ACL";
+  const networkACL = network1 + "-ACL";
   await page.setViewportSize({ width: 1440, height: 800 });
   await gotoURL(page, "/ui/");
 
@@ -224,17 +221,20 @@ test("network peering", async ({ page }) => {
   await page.getByTestId("notification-close-button").click();
 
   // Create network
-  await createNetwork(page, network, "ovn", {
+  await createOvnUplink(page, uplinkNetwork);
+  await createNetwork(page, network1, "ovn", {
     hasMemberSpecificParents: false,
+    uplink: uplinkNetwork,
     networkACL,
+    ipv6: "none",
   });
   await createNetwork(page, network2, "ovn");
   await createNetwork(page, network3, "ovn");
-  await createNetworkLocalPeering(page, "LocalPeering1", network, network2);
+  await createNetworkLocalPeering(page, "LocalPeering1", network1, network2);
   await createNetworkLocalPeering(
     page,
     "LocalPeering2",
-    network,
+    network1,
     network3,
     true,
   );
@@ -262,10 +262,12 @@ test("network peering", async ({ page }) => {
     clip: getClipPosition(240, 0, 1440, 800),
   });
 
+  await deleteLocalPeerings(page, network1, "LocalPeering1");
+  await deleteLocalPeerings(page, network1, "LocalPeering2");
   await deleteNetwork(page, network3);
-  await deleteLocalPeerings(page, network2, "LocalPeering1");
   await deleteNetwork(page, network2);
-  await deleteNetwork(page, network);
+  await deleteNetwork(page, network1);
+  await deleteNetwork(page, uplinkNetwork);
   await deleteNetworkAcl(page, networkACL);
 });
 
@@ -301,7 +303,6 @@ test("storage volumes", async ({ page }) => {
 
   await page.getByRole("link", { name: "Volumes", exact: true }).click();
   await page.getByText("Create volume").click();
-
   await page.getByRole("button", { name: "Upload volume file" }).click();
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_import.png",
@@ -311,7 +312,7 @@ test("storage volumes", async ({ page }) => {
 
   await page.getByPlaceholder("Enter name").fill(volumeName);
   await page.getByLabel("Storage pool", { exact: true }).click();
-  await page.getByLabel("sub").getByText("pool1").click();
+  await page.getByLabel("sub").getByText(poolName).click();
 
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_create.png",
@@ -323,7 +324,7 @@ test("storage volumes", async ({ page }) => {
   await page.getByRole("button", { name: "Create", exact: true }).click();
   await page.getByTestId("notification-close-button").click();
   await page.getByRole("link", { name: volumeName }).click();
-  await page.waitForSelector(`text=Overview`);
+  await page.getByText("Overview").waitFor();
   await page.screenshot({
     path: "tests/screenshots/doc/images/storage/storage_volumes_overview.png",
     clip: getClipPosition(240, 0, 1420, 550),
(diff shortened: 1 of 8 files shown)


tests/helpers/search.ts
```diff
@@ -0,0 +1,8 @@
+import type { Page } from "@playwright/test";
+
+export const searchEntityListPage = async (page: Page, name: string) => {
+  await page.getByPlaceholder("Search and filter").fill(name.toLowerCase());
+  await page.getByPlaceholder("Search and filter").press("Enter");
+  await page.getByPlaceholder("Add filter").press("Escape");
+  await page.getByRole("link", { name }).first().click();
+};
```
edlerd, line 7: Good idea to extract this. Maybe remove the last line and then we can reuse it in multiple other places. Wdyt?
```suggestion
export const searchEntityListPage = async (page: Page, name: string) => {
  await page.getByPlaceholder("Search and filter").fill(name.toLowerCase());
  await page.getByPlaceholder("Search and filter").press("Enter");
  await page.getByPlaceholder("Add filter").press("Escape");
};
```

tests/helpers/search.ts
edlerd, line 1: Can we call this file `tests/helpers/search.ts`?

tests/helpers/instances.ts
edlerd, line 1: I'd expect all functions in the helpers folder to be heavily used by other tests and to be reliable. I think changes to most of the helper files might not be necessary. We can refactor the helpers or add functionality, but we have to keep the behavior as is so as not to break other tests.

tests/helpers/network.ts
```diff
@@ -86,11 +128,8 @@ export const deleteNetwork = async (page: Page, network: string) => {
 };
 
 export const visitNetwork = async (page: Page, network: string) => {
-  await gotoURL(page, "/ui/");
-  await page.waitForLoadState("networkidle");
-  await page.getByRole("button", { name: "Networking" }).click();
-  await page.getByRole("link", { name: "Networks", exact: true }).click();
-  await page.getByRole("link", { name: network }).first().click();
+  const link = await getNetworkLink(page, network);
+  await link.click();
```
edlerd, line 108: This is changing this helper's behavior and likely breaking other tests.

tests/helpers/storagePool.ts
```diff
@@ -15,11 +14,11 @@ export const createPool = async (page: Page, pool: string, driver = "dir") => {
...
   await page.getByPlaceholder("Enter name").fill(pool);
   await page.getByLabel("Driver").selectOption(driver);
-  if (driver === cephObject) {
-    await page.getByLabel("Rados gateway endpoint").fill("http://localhost");
-  }
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForSelector(`text=Storage pool ${pool} created.`);
+  await page
+    .getByText(`Storage pool ${pool} created.`)
+    .waitFor({ timeout: 2000 });
```
edlerd: I think the default timeout should be sufficient here. Why reduce it? Or is this for debugging only?

tests/docs-screenshots.spec.ts
```diff
@@ -134,13 +136,7 @@ test("networks", async ({ page }) => {
...
   await page.setViewportSize({ width: 1440, height: 800 });
 
-  await gotoURL(page, "/ui/");
-  await page.getByText("Networking").click();
-  await page.getByText("Networks").click();
-  await page.getByText("Create network").click();
-  await page.getByPlaceholder("Enter name").fill(network);
-  await page.getByRole("button", { name: "Type" }).click();
-  await page.getByLabel("sub").getByText("bridge").first().click();
+  await prepareCreateNetwork(page, network, "bridge");
```
edlerd, line 141: Good idea to refactor and reuse the existing helper here. That ensures this won't break again as we rely on existing helper infrastructure also used by other tests. As such, it is maintained with future changes.

tests/docs-screenshots.spec.ts
```diff
@@ -221,12 +215,17 @@ test("network peering", async ({ page }) => {
     .click();
   await page.getByText("Add rule").click();
   await page.getByRole("button", { name: "Create", exact: true }).click();
+  await page
+    .getByTestId("notification-close-button")
+    .waitFor({ timeout: 1000 });
   await page.getByTestId("notification-close-button").click();
```
edlerd:
```suggestion
await page.getByTestId("notification-close-button").click();
```
I think the last line waits for the ID anyway with a much longer timeout.

tests/docs-screenshots.spec.ts
```diff
@@ -311,7 +312,7 @@ test("storage volumes", async ({ page }) => {
 
   await page.getByPlaceholder("Enter name").fill(volumeName);
   await page.getByLabel("Storage pool", { exact: true }).click();
-  await page.getByLabel("sub").getByText("pool1").click();
+  await page.getByLabel("submenu").getByText(poolName).click();
```
edlerd:
```suggestion
  await page.getByLabel("sub").getByText(poolName).click();
```
This was changed very recently with react-components update.

edlerd (commented):
Some thoughts on the open PR below.

tests/helpers/search.ts
omarelkashef: Are there uses to search and not click on the result to see the detail page? Maybe assert that an entity is deleted?

tests/helpers/instances.ts
omarelkashef: I will revert changes but keep the move from waitForSelector to locators as the use of waitForSelector is discouraged. https://playwright.dev/docs/api/class-page#page-wait-for-selector

tests/helpers/search.ts
edlerd: Yes, for instance the "Bulk start, pause, unpause and stop instances" which interacts with the instance row by the actions in the list and not by going to the detail page.

tests/helpers/instances.ts
edlerd: Yes, sounds good to me :+1:

tests/helpers/search.ts
omarelkashef: Ok sounds good!

tests/helpers/storageVolume.ts
```diff
@@ -38,11 +39,11 @@ export const deleteVolume = async (
...
 ) => {
   await visitVolume(page, volume, project);
+  await page.waitForTimeout(1000);
   await page.getByRole("button", { name: "Delete" }).click();
-  await page
-    .getByRole("dialog", { name: "Confirm delete" })
-    .getByRole("button", { name: "Delete" })
-    .click();
+  const dialog = page.getByRole("dialog", { name: "Confirm delete" });
+  await dialog.waitFor({ timeout: 2000 });
```
edlerd: Those timeouts should be avoided. The previous flow should work fine without timeouts. Why change it?

tests/helpers/network.ts
```diff
@@ -58,12 +60,52 @@ export const createNetwork = async (
...
+  page: Page,
+  network: string,
+  type = "bridge",
+  options: {
+    hasMemberSpecificParents?: boolean;
+    networkACL?: string;
+    uplink?: string;
+    ipv4?: string;
+    ipv6?: string;
+  } = {},
```
edlerd: Can we introduce a local type for this? To have a single source of it and avoid repeating the types in two places?

tests/helpers/network.ts
```diff
@@ -266,45 +320,45 @@ export const createOvnUplink = async (page: Page, network: string) => {
...
   const customRadio = page.getByRole("radio", { name: "Custom" }).first();
   await customRadio.check({ force: true });
   await expect(customRadio).toBeChecked();
 
   await expect(page.getByLabel("IPv4 address")).toBeEnabled();
-  await page.getByLabel("IPv4 address").fill("10.42.42.1/24");
+  await page.getByLabel("IPv4 address").fill(CIDR);
+
+  await page.getByRole("button", { name: "Save" }).click();
+  await visitNetwork(page, network);
```
edlerd: Why do we save it here and continue configuration as edit?

edlerd (commented):
Thanks for fixing it. Do you have a link to a local run of the screenshot tests on your branch?

Some observations on the helper changes are below.

tests/helpers/network.ts
omarelkashef: Good idea, yes.

tests/helpers/network.ts
omarelkashef: Because network subnet needs to be set and saved first before ovn and dhcp ranges are set to be within the subnet.

tests/helpers/storageVolume.ts
omarelkashef: Waiting for the dialog seems to be necessary. 

```
Error: locator.click: Test timeout of 120000ms exceeded.
    Call log:
      - waiting for getByRole('dialog', { name: 'Confirm delete' }).getByRole('button', { name: 'Delete' })
        - locator resolved to <button class="u-no-margin--bottom p-action-button p-button--negative">Delete</button>
      - attempting click action
        - waiting for element to be visible, enabled and stable
        - element is not stable
      - retrying click action
        - waiting for element to be visible, enabled and stable
      - element was detached from the DOM, retrying


       at helpers/storageVolume.ts:44

      42 |   await page.getByRole("button", { name: "Delete" }).click();
      43 |   const dialog = page.getByRole("dialog", { name: "Confirm delete" });
    > 44 |   await dialog.getByRole("button", { name: "Delete" }).click();
         |                                                        ^
      45 |   await expect(
      46 |     page.getByText(`Storage volume ${volume} deleted.`),
      47 |   ).toBeVisible();
```

tests/helpers/network.ts
```diff
@@ -80,17 +118,13 @@ export const deleteNetwork = async (page: Page, network: string) => {
     .getByRole("dialog", { name: "Confirm delete" })
     .getByRole("button", { name: "Delete" })
     .click();
-  await page.waitForTimeout(2500);
```
edlerd, line 83: We should restore this, it is to reduce flaky behavior on testing the network deletion, which might take a moment to complete depending on how busy the CI runner is.

tests/helpers/network.ts
edlerd: That seems wrong. Before, we created the network with OVN and DHCP ranges right away in one go. See the current state of the main branch that is successful in CI.

tests/helpers/storageVolume.ts
```diff
@@ -56,11 +57,8 @@ export const visitVolume = async (
   await page.getByRole("button", { name: "Storage" }).click();
   await page.getByRole("link", { name: "Volumes" }).click();
   await expect(page.getByText("Create volume")).toBeVisible();
-  await page.getByPlaceholder("Search and filter").fill(volume);
-  await page.getByPlaceholder("Search and filter").press("Enter");
-  await page.getByPlaceholder("Add filter").press("Escape");
+  await searchEntityListPage(page, volume);
   await page.getByRole("link", { name: volume }).first().click();
-  await expect(page.getByText(`Storage volumes${volume}`)).toBeVisible();
```
edlerd, line 63: We should keep this to ensure we have finished loading the storage volume detail page.

tests/helpers/storageVolume.ts
edlerd: The error is due to another change introduced in this PR. See my other comment on this file.

tests/docs-screenshots.spec.ts
```diff
@@ -204,10 +199,11 @@ test("networks", async ({ page }) => {
...
 test("network peering", async ({ page }) => {
-  const network = "OVNNetwork";
-  const network2 = "OVNNetwork2";
-  const network3 = "OVNNetwork3";
-  const networkACL = network + "-ACL";
+  const uplinkNetwork = "BridgeNetwork-1";
+  const network1 = "OVNNetwork-1";
+  const network2 = "OVNNetwork-2";
+  const network3 = "OVNNetwork-3";
+  const networkACL = network1 + "-ACL";
```
edlerd, line 207: Can we keep the names? They might be referenced in the docs as they were before.

tests/docs-screenshots.spec.ts
```diff
@@ -323,7 +324,7 @@ test("storage volumes", async ({ page }) => {
   await page.getByRole("button", { name: "Create", exact: true }).click();
   await page.getByTestId("notification-close-button").click();
   await page.getByRole("link", { name: volumeName }).click();
-  await page.waitForSelector(`text=Overview`);
+  await page.getByText("Overview").waitFor({ timeout: 120000 });
```
edlerd:
```suggestion
  await page.getByText("Overview").waitFor();
```
This is the same timeout set in playwright.config.ts, no?

tests/docs-screenshots.spec.ts
```diff
@@ -353,9 +353,9 @@ test("storage volumes", async ({ page }) => {
 });
 
 test("storage volume snapshots", async ({ page }) => {
-  const poolName = "pool1";
-  const volumeName = "CustomVol1";
-  const snapshot = "snapshot1";
+  const poolName = "pool-1";
+  const volumeName = "CustomVol-1";
+  const snapshot = "snapshot-1";
```
edlerd: Can we keep the names?

tests/docs-screenshots.spec.ts
```diff
@@ -381,11 +381,13 @@ test("storage volume snapshots", async ({ page }) => {
...
-    `text=Snapshot ${snapshot} created for volume ${volumeName}.`,
-  );
+  await expect(
+    page.getByText(`Snapshot ${snapshot} created for volume ${volumeName}.`),
+  ).toBeVisible();
   await page.reload();
-  await page.waitForSelector("text=Showing 1 out of 1 snapshot");
+  await page
+    .getByText("Showing 1 out of 1 snapshot")
+    .waitFor({ timeout: 120000 });
```
edlerd:
```suggestion
    .waitFor();
```

tests/docs-screenshots.spec.ts
```diff
@@ -403,20 +405,24 @@ test("storage volume snapshots", async ({ page }) => {
 
 test("LXD - Tutorial folder", async ({ page }) => {
   await page.setViewportSize({ width: 1440, height: 800 });
-
-  const instance = "Instance1";
+  const projectButton = page
+    .getByTitle("Select project (default)", { exact: true })
+    .getByRole("button");
+  const instance = "harmless-sailfish";
```
edlerd: Please keep the instance name as it was.

tests/docs-screenshots.spec.ts
```diff
@@ -492,14 +497,15 @@ test("LXD - Tutorial folder", async ({ page }) => {
...
+  await page.getByRole("button", { name: "Cancel", exact: true }).click();
   await page.getByTestId("tab-link-Terminal").click();
   await page.waitForTimeout(2000);
   await page.keyboard.type("exit");
 
   await page.keyboard.press("Enter");
-  await page.waitForSelector(`text=The connection was closed abnormally`);
+  await page
+    .getByText(`The connection was closed abnormally`)
+    .waitFor({ timeout: 120000 });
```
edlerd:
```suggestion
    .waitFor();
```

tests/helpers/network.ts
omarelkashef: At first, this function only supported new networks with an early return if the network already exists. Now, it supports existing networks. For existing networks, you can not change IP and set  ovn and dhcp ranges at the same time. Now, in order to not complicate the code, if the network does not exist, we create it with basic config so that we can treat all networks as existing networks.

tests/docs-screenshots.spec.ts
omarelkashef: Yes, I will remove the timeout.

tests/helpers/network.ts
edlerd: I think that is too much scope for a single function. The name `createOvnUplink` is now misleading, because we start using the function to edit existing networks as well.

Can we extract the bits we need on updating existing OVN networks? Like a function that only sets the ranges, but no saving or anything like that. Then we keep `createOvnUplink` as a create method with fewer changes. In it we can call the extracted method. And in the case where we want to edit, we can call the same extracted function.

tests/docs-screenshots.spec.ts
```diff
@@ -461,9 +464,9 @@ test("LXD - Tutorial folder", async ({ page }) => {
 
   await page.getByRole("button", { name: "Create and start" }).click();
   await page.getByTestId("notification-close-button").click();
-
+  await page.waitForTimeout(5000);
```
edlerd: I am not sure why this is needed. Can we wait for the message "Instance :name has been started" instead here?

tests/docs-screenshots.spec.ts
```diff
@@ -544,11 +548,12 @@ test("LXD - Tutorial - Graphical consoles", async ({ page }) => {
...
     .click();
   await page.getByTestId("notification-close-button").click();
+  await page.getByRole("link", { name: vminstance }).first().waitFor();
 
   await visitInstance(page, vminstance);
-  await page.waitForSelector("text=General");
+  await page.getByText("General").click();
   await page.getByTestId("tab-link-Console").click();
-  await page.waitForTimeout(45000);
+  await page.waitForTimeout(10000);
```
edlerd: This might not be enough for the CI to finish booting the VM. Resulting in a wrong screenshot of the VM booting instead of the desktop being shown. Maybe restore the old value here?

tests/helpers/network.ts
omarelkashef: I split it into `createOvnUplink` and `makeNetworkOvnUplink` to separate concerns.

PR #1816, opened 2026-02-23
feat(instance config): Add custom ISOs in devices > disk

kimanhou:
## Done

- Manage ISOs from instance configuration > Devices > Disk

## QA

1. Run the LXD-UI:
    - In an instance configuration > Devices > Disk, click on `Attach disk device` > `Attach ISO`. Select existing ISO volume. Make sure the newly added ISO appears in the `Custom disks devices` section with a blue dot. Save changes. 
    - Make sure the button `Attach ISO` is disabled when the instance has an ISO volume attached.
    - Detach ISO, save changes.
    - Create a profile with a custom ISO. Create an instance with this profile. Make sure the inherited custom ISO appears in the `Inherited disk devices` list. Make sure you can detach and re-attach the inherited ISO.
    - From the console tab, attach a custom ISO. Make sure it appears in the configuration tab under the `Custom disks devices` section. Make sure you can detach the ISO from the configuration tab and console tab.

## Screenshots

image
image
inherited-detached
inherited

### DIFF
--- a/src/components/forms/DiskDeviceFormCustom.tsx
+++ b/src/components/forms/DiskDeviceFormCustom.tsx
@@ -3,27 +3,40 @@ import { Button, Icon, Input, Label } from "@canonical/react-components";
 import type { InstanceAndProfileFormikProps } from "../../types/forms/instanceAndProfileFormProps";
 import type { EditInstanceFormValues } from "types/forms/instanceAndProfile";
 import CustomVolumeSelectBtn from "pages/storage/CustomVolumeSelectBtn";
-import type { FormDevice, FormDiskDevice } from "types/formDevice";
+import type {
+  CustomDiskDevice,
+  FormDevice,
+  FormDiskDevice,
+  IsoVolumeDevice,
+} from "types/formDevice";
 import {
   deduplicateName,
   isFormDiskDevice,
   removeDevice,
 } from "util/formDevices";
-import RenameDeviceInput from "./RenameDeviceInput";
+import DeviceName from "components/forms/DeviceName";
+import RenameDeviceInput from "components/forms/RenameDeviceInput";
 import ConfigurationTable from "components/ConfigurationTable";
 import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
 import { getConfigurationRowBase } from "components/ConfigurationRow";
 import DetachDiskDeviceBtn from "pages/instances/actions/DetachDiskDeviceBtn";
 import classnames from "classnames";
 import { isDiskDeviceMountPointMissing } from "util/instanceValidation";
-import { isRootDisk } from "util/devices";
+import {
+  getExistingDeviceNames,
+  isIsoDiskDevice,
+  ISO_VOLUME_NAME,
+  ISO_VOLUME_PROFILE_NAME,
+  ISO_VOLUME_TYPE,
+  isRootDisk,
+} from "util/devices";
 import { ensureEditMode, isInstanceCreation } from "util/instanceEdit";
-import { getExistingDeviceNames, isVolumeDevice } from "util/devices";
-import type { LxdProfile } from "types/profile";
+import { isVolumeDevice } from "util/devices";
 import { focusField } from "util/formFields";
 import AttachDiskDeviceBtn from "pages/storage/AttachDiskDeviceBtn";
-import type { LxdDiskDevice } from "types/device";
+import type { LxdProfile } from "types/profile";
 import type { LxdStorageVolume } from "types/storage";
+import StoragePoolRichChip from "pages/storage/StoragePoolRichChip";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
@@ -34,12 +47,23 @@ interface Props {
 const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
   const readOnly = (formik.values as EditInstanceFormValues).readOnly;
   const existingDeviceNames = getExistingDeviceNames(formik.values, profiles);
+  const isProfile = formik.values.entityType === "profile";
 
-  const addDiskDevice = (device: LxdDiskDevice) => {
+  const getInitialDeviceName = (
+    deviceType: string,
+    isProfile: boolean,
+  ): string => {
+    if (deviceType === ISO_VOLUME_TYPE) {
+      return isProfile ? ISO_VOLUME_PROFILE_NAME : ISO_VOLUME_NAME;
+    }
+    return deduplicateName("disk-device", 1, existingDeviceNames);
+  };
+
+  const addDiskDevice = (device: CustomDiskDevice) => {
     const copy = [...formik.values.devices];
     const newDevice: FormDevice = {
       ...device,
-      name: deduplicateName("disk-device", 1, existingDeviceNames),
+      name: getInitialDeviceName(device.type, isProfile),
     };
 
     copy.push(newDevice);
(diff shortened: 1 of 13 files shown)


src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -84,6 +101,19 @@ const AttachDiskDeviceModal: FC<Props> = ({
...
+              onClick={() => {
+                setType("iso");
+              }}
+              disabled={hasIsoDeviceAttached}
+              onHoverText={
+                hasIsoDeviceAttached
+                  ? "Only one ISO volume can be attached to an instance at a time"
+                  : undefined
+              }
+            />
```
edlerd: This should be hidden on containers, only VMs can have the iso volumes attached.

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -108,6 +138,9 @@ const AttachDiskDeviceModal: FC<Props> = ({
           title={modalTitle}
         />
       )}
+      {type === "iso" && (
+        <CustomIsoModal onClose={close} onSelect={handleSelectIso} />
```
edlerd: the footer in the modal should be "Back" instead of cancel to keep the modal open and return to the selection. Also the bottom spacing of the buttons is a bit too large. The "Mount host path" option has correct spacing. I think "Attach custom volume" also has slightly wrong bottom spacing.

kimanhou: I fixed the bottom spacing, it was due to a fixed `min-height`.
I changed the button to be `Back` if the modal is triggered from the configuration tab, it stays `Cancel` from the console tab. In the UploadCustomIso, I changed it to be always `Back`.

src/pages/images/CustomIsoSelector.tsx
```diff
@@ -127,7 +129,7 @@ const CustomIsoSelector: FC<Props> = ({
           className="u-no-margin--bottom"
           onClick={onCancel}
         >
-          Cancel
+          {cancelButtonText}
```
edlerd, line 132: Thanks for adding the back link here. We also tend to show a back link in the header, such as in the screen below. Can we add it here for the case of coming from the disk configuration as well?

Image

edlerd: By the way, slightly related: I realized in user testing people try to click on the arrow in the header and expect it to go back as well. But maybe something to do in a dedicated task and PR as it seems out of scope for this one.

kimanhou: I created a [JIRA ticket](https://warthogs.atlassian.net/browse/WD-34152) for it.

edlerd: Thanks, PR for it is over there at https://github.com/canonical/lxd-ui/pull/1822

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -91,160 +105,229 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
   let customDiskDeviceCount = 0;
   for (let index = 0; index < formik.values.devices.length; index++) {
     const item = formik.values.devices[index];
-    if (!isFormDiskDevice(item) || isRootDisk(item)) {
+    if (
+      (!isFormDiskDevice(item) && !isIsoDiskDevice(item)) ||
+      isRootDisk(item)
```
edlerd: I think this is correct, but I have to think long about why. Can we make this more obvious or simpler?

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -91,160 +105,229 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
-      }),
-    );
+    if (isIsoDiskDevice(item)) {
+      const hasBeenAdded = !formik.initialValues.devices.some(
+        (t) => t.name === item.name,
+      );
+      const getSource = (item: FormDevice) => {
+        const isoDevice = item as IsoVolumeDevice;
+        const diskDevice = item as FormDiskDevice;
+        return isoDevice.bare?.source ?? diskDevice.source ?? "";
```
edlerd: Why do we need the fallback here? Shouldn't this always be one of the two cases?

src/components/forms/DiskDeviceFormInherited.tsx
edlerd, line 1: We hide the "attach iso" on profiles. Do we want to support inheritance?
If we do, there are some problems to solve. Such as an inherited iso and a local iso. We can solve this by allowing two isos: one from the inheritance and one from local. The profile isos should then have a different name as the local ones.

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -84,6 +105,21 @@ const AttachDiskDeviceModal: FC<Props> = ({
                 setType("host path");
               }}
             />
+            {isVirtualMachine && (
```
edlerd: We might also want to allow profiles.

src/util/formDevices.tsx
```diff
@@ -66,8 +66,8 @@ export const parseDevices = (devices: LxdDevices): FormDevice[] => {
     if (key === "iso-volume") {
       return {
         name: key,
-        bare: item as unknown,
-        type: "unknown",
+        bare: item as LxdIsoDevice,
+        type: "iso-volume",
```
edlerd: Maybe this is the missing bit to generalize the cases as suggested in the [other comment](https://github.com/canonical/lxd-ui/pull/1816/changes#r2845059799). We should probably populate the source and pool field from the bare device.

Something like the sketch below (I didn't verify the bare fields are correct though):

```suggestion
        bare: item as LxdIsoDevice,
        type: "iso-volume",
        source: bare.source,
        pool: bare.pool,
```

src/pages/images/CustomIsoSelector.tsx
```diff
@@ -11,13 +11,15 @@ interface Props {
...
   onCancel: () => void;
+  cancelButtonText?: string;
 }
 
 const CustomIsoSelector: FC<Props> = ({
   primaryImage,
   onSelect,
   onUpload,
   onCancel,
+  cancelButtonText = "Cancel",
```
edlerd, line 22: Can we also add the back link in the header like the "Choose disk type" below? Also would be nice to have a "< Select custom ISO" back link on the upload iso step.

Image

edlerd (commented):
Good progress! Did a thorough review and QA, and some refinement ideas are below.

src/pages/images/CustomIsoSelector.tsx
kimanhou: Thank you, I confirm it works now

kimanhou: I added the back link for ISO. It also appears in the Upload custom iso modal, even when accessed thru the console tab.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -90,42 +95,69 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
-    } else {
       rows.push(
         getInheritedDeviceRow({
-          label: "Pool / volume",
+          label: "Pool",
           inheritValue: (
-            <>
-              {item.disk.pool} / {item.disk.source}
-            </>
+            <StoragePoolRichChip
```
edlerd, line 113: The pool should display strike-through when detached.

Image

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -34,12 +44,23 @@ interface Props {
...
   const readOnly = (formik.values as EditInstanceFormValues).readOnly;
   const existingDeviceNames = getExistingDeviceNames(formik.values, profiles);
+  const isProfile = formik.values.entityType === "profile";
+
+  const getInitialDeviceName = (
+    deviceType: string,
+    isProfile: boolean,
+  ): string => {
+    if (deviceType === "iso-volume") {
+      return isProfile ? "iso-volume-profile" : "iso-volume";
```
edlerd: Can we introduce a constant for the `iso-volume` string when it is used as a way to find the device?
Then we can reference it here, in the `AttachIsoBtn`, the `AttachDiskDeviceModal` among others.

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -47,18 +52,48 @@ const AttachDiskDeviceModal: FC<Props> = ({
...
-        onClick={handleGoBack}
-        linkText="Choose disk type"
-      />
-    );
+  const handleSelectIso = (image: RemoteImage) => {
+    const isoDevice = remoteImageToIsoDevice(image);
+
+    if (isoDevice.type === "iso-volume" && "bare" in isoDevice) {
+      onFinish(isoDevice);
+    }
```
edlerd:
```suggestion
    const isoDevice = remoteImageToIsoDevice(image);
    onFinish(isoDevice);
```
Do we need the extra check here?

src/pages/storage/AttachDiskDeviceModal.tsx
```diff
@@ -47,18 +52,48 @@ const AttachDiskDeviceModal: FC<Props> = ({
...
+      default:
+        return CHOOSE_DISK_TYPE_TITLE;
+    }
+  };
+
+  const modalTitle = getModalTitle();
+
+  const hasIsoDeviceAttached = formik.values.devices.some((device) =>
+    isIsoDiskDevice(device),
+  );
```
edlerd:
```suggestion
  const hasIsoDeviceAttached = formik.values.devices.some(isIsoDiskDevice);
```
Can use the shorthand here.

src/pages/storage/AttachDiskDeviceModal.tsx
edlerd: We need to change the return type of the helper:
```
remoteImageToIsoDevice = (image: RemoteImage): IsoVolumeDevice
```

src/components/forms/DiskDeviceFormCustom.tsx
kimanhou: I did something. Do you find it clearer ?

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -34,12 +44,23 @@ interface Props {
...
+    }
+    return deduplicateName("disk-device", 1, existingDeviceNames);
+  };
 
-  const addDiskDevice = (device: LxdDiskDevice) => {
+  const addDiskDevice = (device: LxdDiskDevice | IsoVolumeDevice) => {
```
edlerd: Shall we use an alias type for this one? something like 
```
export type CustomDiskDevice = LxdDiskDevice | IsoVolumeDevice;
```

This might go nicely with the `    const isCustomDisk = isFormDiskDevice(item) || isIsoDiskDevice(item);` you introduced in `DiskDeviceFormCustom`.

src/util/devices.tsx
```diff
@@ -41,6 +41,27 @@ export const isHostDiskDevice = (device: LxdDiskDevice): boolean => {
...
+  if (device.type === "disk") {
+    if (device["boot.priority"] !== undefined) {
+      return true;
+    }
+
+    const deviceWithBare = device as FormDevice & { bare?: LxdDiskDevice };
+    if (deviceWithBare.bare?.["boot.priority"] !== undefined) {
+      return true;
+    }
+  }
```
```suggestion
```
Why do we need the extra boot priority checks here?

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -91,160 +112,221 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
   let customDiskDeviceCount = 0;
   for (let index = 0; index < formik.values.devices.length; index++) {
     const item = formik.values.devices[index];
-    if (!isFormDiskDevice(item) || isRootDisk(item)) {
+    const isCustomDisk = isFormDiskDevice(item) || isIsoDiskDevice(item);
+    const isExcludedRoot = isRootDisk(item);
+
+    if (!isCustomDisk || isExcludedRoot) {
```
edlerd:
```suggestion
    const isCustomDisk = isFormDiskDevice(item) || isIsoDiskDevice(item);

    if (!isCustomDisk || isRootDisk(item)) {
```
I think introducing the `isCustomDisk` is enough to clarify this bit.

src/util/devices.tsx
kimanhou: Because when the device is inherited, it looks like this:
```
{
    "boot.priority": "10",
    "pool": "playwright-pool-7ee8f",
    "source": "playwright-iso-ote54",
    "type": "disk"
}
```

kimanhou: It looks like type="iso-volume" is not a valid device type in back end.

When creating a profile, here is how the YAML config looks:
image

And when attaching an ISO volume from the console tab, here is how the YAML config looks:
image

src/components/forms/DiskDeviceFormCustom.tsx
kimanhou: In `AttachIsoBtn`, the string `iso-volume` refers to the device **name**, not the **type**. I can introduce 2 different constants.
There is also the type of `ResourceLink`, but that's another story.

edlerd: I was referring to the usage as name only. The device type should be ok to keep as string.

kimanhou: Ah ... Now we have both, can we keep both ? I kinda like having a constant for device type too.

src/util/devices.tsx
edlerd: The `boot.priority` is also a valid key on other disk devices like volumes or mount paths. Maybe change the `parseDevices` function to identify both possible names, the `ISO_VOLUME_NAME` and the `ISO_VOLUME_PROFILE_NAME`. Then the form device we receive here should always have the `device.type === "iso-volume"`.

src/components/forms/DiskDeviceFormCustom.tsx
edlerd: But now that we have both it is also ok to keep them.

src/util/formDevices.tsx
```diff
@@ -63,11 +67,14 @@ export const parseDevices = (devices: LxdDevices): FormDevice[] => {
       };
     }
 
-    if (key === "iso-volume") {
+    if (key === ISO_VOLUME_NAME) {
```
edlerd: We might need to allow matching for the ISO_VOLUME_PROFILE_NAME as well.

src/util/devices.tsx
```diff
@@ -41,6 +44,15 @@ export const isHostDiskDevice = (device: LxdDiskDevice): boolean => {
   );
 };
 
+export const isIsoDiskDevice = (
+  device: LxdDiskDevice | FormDevice,
+): boolean => {
+  return (
+    device.type === "iso-volume" ||
+    (device.type === "disk" && "boot.priority" in device)
```
edlerd: type disk with boot priority seems like an unsafe heuristic here that can have false positives.

src/types/formDevice.d.ts
```diff
@@ -58,3 +60,5 @@ export type FormDevice =
   | LxdProxyDevice
   | LxdOtherDevice
   | EmptyDevice;
+
+export type CustomDiskDevice = LxdDiskDevice | IsoVolumeDevice;
```
edlerd, line 64: Let's move this to `types/formDevice.d.ts`

src/util/devices.tsx
kimanhou: I created two functions to determine if a device is ISO: one for inherited and one for local.
```
export const isIsoDiskDevice = (
  device: LxdDiskDevice | FormDevice,
): boolean => {
  return device.type === ISO_VOLUME_TYPE;
};

export const isInheritedIsoDiskDevice = (
  inheritedDevice: InheritedDiskDevice,
): boolean => {
  return (
    isDiskDevice(inheritedDevice.disk) &&
    (inheritedDevice.key === ISO_VOLUME_NAME ||
      inheritedDevice.key === ISO_VOLUME_PROFILE_NAME)
  );
};
```

Would you prefer to combine them into one function ?

edlerd: Yeah, I think one should be sufficient.

src/types/formDevice.d.ts
kimanhou: Isn't it already ?
image

edlerd: Ah yes, sorry mixed it up.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -90,42 +96,72 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
+              poolName={item.disk.pool ?? ""}
+              projectName={project}
+              className={classnames({
+                "u-text--line-through": isNoneDevice,
```
edlerd, line 117: Something is off with identifying the none device. Before saving:

Image

after saving the none device:

Image

Issue #1770, opened 2026-02-05
Empty icon on menu collapse


omarelkashef:
When menu is collapsed while a subsection is open, the subsection does not close and the menu selection is  empty.

<img width="1896" height="951" alt="Image" src="https://github.com/user-attachments/assets/7d624411-6092-4bf6-9154-1281357d36df" />

PR #1818, opened 2026-02-23
fix: Address empty icon on menu collapse

Kxiru:
## Done

- [x] On navigation collapse, sub-sections should no longer have space where their icons would be (navigation becomes compressed).
- [x] With menu closed, we should highlight the matching parent item in the collapsed menu.
- [x] On uncollapsing the navigation, we can restore the sub-menu open state from before and highlight the active sub-page again.

Issues: One teeeny tiny issue: If you select a submenu, close the Navigation (parent menu should be highlighted), and then refresh the page of the submenu _and close the Navigation again_, the parent menu is no longer highlighted as I believe context has been lost. By this thinking, I also assume that if you navigate to a submenu via URL rather than via navigation, this may also be the case. Investigating a URL based approach may be a related task.

Fixes:
- https://github.com/canonical/lxd-ui/issues/1770

## QA

1. Run the LXD-UI:
    - Find an entity that has subpages (Eg; Networks, Storage, clustering or permissions)
    - Navigate to one of these subpages (Eg; ACLs, Volumes, Groups)
    - Minimise the main navigation and ensure that:
      - There is no space where the subsection would have been (menu is compressed)
      - The parent entity icon is highlighted.

## Screenshots

New behaviour:
image
image

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -300,6 +300,7 @@ const Navigation: FC = () => {
                           Instances
                         </NavLink>
                       </SideNavigationItem>
+
                       <SideNavigationItem>
                         <NavLink
                           to={`${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/profiles`}
@@ -317,7 +318,9 @@ const Navigation: FC = () => {
 
                       <SideNavigationItem>
                         <NavAccordion
-                          baseUrl={`${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/network`}
+                          baseUrls={[
+                            `${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/network`,
+                          ]}
                           title={getNavTitle("networking")}
                           disabled={isAllProjects}
                           iconName="exposed"
@@ -325,7 +328,10 @@ const Navigation: FC = () => {
                           onOpen={() => {
                             toggleAccordionNav("networking");
                           }}
-                          open={openNavMenus.includes("networking")}
+                          open={
+                            openNavMenus.includes("networking") &&
+                            !menuCollapsed
+                          }
                         >
                           {[
                             <SideNavigationItem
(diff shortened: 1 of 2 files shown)


src/components/Navigation.tsx
edlerd, line 46: We can calculate the active super page from the definition here. I think with small refactoring we can calculate the super page directly without introducing new complexity.

src/components/Navigation.tsx
```diff
@@ -104,6 +104,7 @@ const Navigation: FC = () => {
   const [openNavMenus, setOpenNavMenus] = useState<AccordionNavMenu[]>(() =>
     initialiseOpenNavMenus(location),
   );
+  const [currentSuperpage, setCurrentSuperpage] = useState("");
```
edlerd:
```suggestion
```
I think we don't need extra state for this, but can calculate this on the fly during render. That way we also avoid new book keeping on navigation. Additionally I think the current approach falls short when on a detail page like network detail and reloading. Because on reload we loose the state of the super page.

Kxiru: I used a useState variable to capture the current open parent navigation because only one parent should be open/enabled at one time and it would be the most efficient way of passing a single boolean variable to the function that determines whether to highlight or not (getHighlightOnClose).

Without using useState, we will still need a way to store the current value AND unset it, **otherwise multiple superpages will be highlighted at the same time** (using useState conveniently overrides the old value). I assume you wanted me to pass these into the highlightOnClose attribute that I made, which only takes a boolean and isn't parent specific, so all determining logic needs to be done in Navigation.tsx. 

Perhaps you could explain the fix you envisioned, because certainly, a URL based approach would address the issue of the state being forgotten on reload.

Kxiru: Also, the useState logic also accounts for the non-parent super pages such as Images & Configuration, which serve to also override the old superpage. These non-parent pages have their own highlight logic, however if they aren't integrated into this approach, then navigating to a non-parent page from a parent-page keeps the parent-page highlighted.

edlerd: We should remove the state. There are two problems:
1. It introduces accidental complexity and additional maintenance burden
2. It is broken on refresh of the page as the initial value of the new state is empty

I quickly tested my suggestion above, and it seems to work. See this quick sketchup to changing the permission nav entry relying solely on the URL and not on the new state:

image

It works nicely on smaller layouts that collapse the menu, also works on reload:

image

It needs some refactoring. We don't want to duplicate the `location.pathname.includes("/permissions/")` logic already in place for the `initialiseOpenNavMenus` function. So need to rearrange to make the logic work for both cases.

Kxiru: Thanks for working with me on this!

PR #1825, opened 2026-02-24
fix(login) simpler oidc login screen

edlerd:
## Done

- fix(login) simpler oidc login screen

When OIDC is configured, use a simpler login screen with fewer explanations:

image

### DIFF
--- a/src/sass/_login.scss
+++ b/src/sass/_login.scss
@@ -11,6 +11,15 @@
     .p-login-method {
       max-width: unset;
     }
+
+    .auth-container {
+      a,
+      button {
+        display: block;
+        margin-bottom: $spv--small;
+        width: 100%;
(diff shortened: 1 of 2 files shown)


src/sass/_login.scss
```diff
@@ -11,6 +11,15 @@
     .p-login-method {
       max-width: unset;
     }
+
+    .auth-container {
+      a,
+      button {
+        display: block;
+        margin-bottom: 0.5rem;
```
kimanhou:
```suggestion
        margin-bottom: $spv--small;
```
Can we use a variable here ?

kimanhou (commented):
I'm afraid we will have to change this when we improve oidc flow but this is good to go with 6.7

PR #1828, opened 2026-02-25
feat(YAML config): Add expanded when editing an instance

### LINKED ISSUE #1456: [feature]  add `--expanded` option to yaml config view
currently, the view of raw yaml configuration on UI **does not show expanded configuration.**
i understand this may be due to the edition possibility restriction (does not make sense to edit inherited attributes there), but i'm sure we can figure something out from the UX point of view.

i spotted this need on instances but maybe it applies also to other resources.

Image

kimanhou:
## Done

- When editing an instance, add expanded YAML configuration

Fixes #1456

## QA

1. Run the LXD-UI:
    - Use an instance that inherits from a profile, preferably a profile with a device.
    - Go to the instance configuration, switch to YAML configuration: 2 radio buttons should appear (`Instance only` and `Expanded`)
    - Check `Expanded`. Make sure the YAML is read only with a hover text and `Cancel` and `Save` buttons are not displayed. Make sure you can see the inherited profile configuration.

## Screenshots

image

Medium screen with line break
image

Annotated expanded YAML
image

### DIFF
--- a/src/components/forms/YamlConfirmation.tsx
+++ b/src/components/forms/YamlConfirmation.tsx
@@ -16,8 +16,7 @@ const YamlConfirmation: FC<Props> = ({ onConfirm, close }) => {
       title="Confirm"
     >
       <p>
-        Switching back to guided forms will discard all changes in the YAML
-        editor.
+        This will discard all changes in the YAML editor.
         <br />
         Are you sure you want to proceed?
       </p>
--- a/src/components/forms/YamlTypeSelector.tsx
+++ b/src/components/forms/YamlTypeSelector.tsx
@@ -1,0 +1,80 @@
+import type { FC, FormEvent } from "react";
+import { useState } from "react";
+import { RadioInput, usePortal } from "@canonical/react-components";
+import YamlConfirmation from "components/forms/YamlConfirmation";
+import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
+
+export type YamlType = "local" | "expandInheritedValues";
+
+interface Props {
+  yamlType: YamlType;
+  setYamlType: (yamlType: YamlType) => void;
+  formik: InstanceAndProfileFormikProps;
+}
+
+const YamlTypeSelector: FC<Props> = ({ yamlType, setYamlType, formik }) => {
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const [pendingYamlType, setPendingYamlType] = useState<YamlType | null>(null);
+
+  const performSwitch = (newYamlType: YamlType) => {
+    void formik.setFieldValue("yaml", undefined);
+
+    if (newYamlType === "expandInheritedValues") {
+      void formik.setFieldValue("readOnly", true);
+    }
+
+    setYamlType(newYamlType);
+  };
+
+  const handleChange = (
+    newYamlType: YamlType,
+    e: FormEvent<HTMLInputElement>,
+  ) => {
+    if (formik.values.yaml) {
+      setPendingYamlType(newYamlType);
+      openPortal(e);
+      return;
+    }
+
+    performSwitch(newYamlType);
+  };
+
+  const handleConfirm = () => {
+    if (pendingYamlType) {
+      performSwitch(pendingYamlType);
+    }
+    setPendingYamlType(null);
+    closePortal();
+  };
+
+  const handleCancel = () => {
+    setPendingYamlType(null);
+    closePortal();
+  };
+
+  return (
+    <div className="yaml-type-radio-wrapper">
+      {isOpen && (
+        <Portal>
+          <YamlConfirmation onConfirm={handleConfirm} close={handleCancel} />
+        </Portal>
+      )}
+      <RadioInput
+        label="Local"
+        checked={yamlType === "local"}
+        onChange={(e) => {
+          handleChange("local", e);
+        }}
+      />
+      <RadioInput
+        label="Expand inherited values"
+        checked={yamlType === "expandInheritedValues"}
+        onChange={(e) => {
+          handleChange("expandInheritedValues", e);
+        }}
+      />
+    </div>
+  );
+};
+
+export default YamlTypeSelector;
--- a/src/pages/instances/EditInstance.tsx
+++ b/src/pages/instances/EditInstance.tsx
@@ -12,7 +12,11 @@ import { useFormik } from "formik";
 import { updateInstance } from "api/instances";
 import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
-import { objectToYaml, yamlToObject } from "util/yaml";
+import {
+  expandInheritedValuesYaml,
+  objectToYaml,
+  yamlToObject,
+} from "util/yaml";
 import { useNavigate, useParams } from "react-router-dom";
 import type { LxdInstance } from "types/instance";
 import SecurityPoliciesForm from "components/forms/SecurityPoliciesForm";
@@ -48,7 +52,6 @@ import {
 import { slugify } from "util/slugify";
 import { useEventQueue } from "context/eventQueue";
 import { hasDiskError, hasNetworkError } from "util/instanceValidation";
-import FormFooterLayout from "components/forms/FormFooterLayout";
 import MigrationForm from "components/forms/MigrationForm";
 import GPUDeviceForm from "components/forms/GPUDeviceForm";
 import OtherDeviceForm from "components/forms/OtherDeviceForm";
@@ -65,6 +68,9 @@ import NetworkDevicePanel from "components/forms/NetworkDevicesForm/edit/Network
 import { InstanceRichChip } from "./InstanceRichChip";
 import { ROOT_PATH } from "util/rootPath";
 import type { EditInstanceFormValues } from "types/forms/instanceAndProfile";
+import type { YamlType } from "components/forms/YamlTypeSelector";
+import YamlTypeSelector from "components/forms/YamlTypeSelector";
+import FormFooterLayout from "components/forms/FormFooterLayout";
 
 interface Props {
   instance: LxdInstance;
@@ -83,6 +89,7 @@ const EditInstance: FC<Props> = ({ instance }) => {
   const navigate = useNavigate();
   const [version, setVersion] = useState(0);
   const { canEditInstance } = useInstanceEntitlements();
+  const [yamlType, setYamlType] = useState<YamlType>("local");
 
   if (!project) {
     return <>Missing project</>;
@@ -174,13 +181,20 @@ const EditInstance: FC<Props> = ({ instance }) => {
       "expanded_devices",
       "etag",
     ]);
+
     const bareInstance = Object.fromEntries(
-      Object.entries(instance).filter((e) => !exclude.has(e[0])),
+      Object.entries(instance).filter(([key]) => !exclude.has(key)),
     );
-    return objectToYaml(bareInstance);
+
+    if (yamlType === "local") {
+      return objectToYaml(bareInstance);
+    }
+
+    return expandInheritedValuesYaml(instance, profiles);
   };
 
   const readOnly = formik.values.readOnly;
+  const isYamlSection = section === slugify(YAML_CONFIGURATION);
 
   return (
     <div className="edit-instance">
@@ -256,16 +270,23 @@ const EditInstance: FC<Props> = ({ instance }) => {
               />
             )}
 
-            {section === slugify(YAML_CONFIGURATION) && (
+            {isYamlSection && (
               <YamlForm
-                key={`yaml-form-${version}`}
+                key={`yaml-form-${yamlType}-${version}`}
                 yaml={getYaml()}
                 setYaml={(yaml) => {
                   ensureEditMode(formik);
                   formik.setFieldValue("yaml", yaml);
                 }}
-                readOnly={!!formik.values.editRestriction}
-                readOnlyMessage={formik.values.editRestriction}
+                readOnly={
+                  yamlType === "expandInheritedValues" ||
+                  !!formik.values.editRestriction
+                }
+                readOnlyMessage={
+                  yamlType === "expandInheritedValues"
+                    ? "Expanded configuration is read-only"
+                    : formik.values.editRestriction
+                }
               >
                 <YamlNotification entity="instance" docPath="/instances" />
               </YamlForm>
@@ -273,14 +294,21 @@ const EditInstance: FC<Props> = ({ instance }) => {
           </Col>
         </Row>
       </Form>
-      <FormFooterLayout>
+      <FormFooterLayout isAlignRight={false}>
         <YamlSwitch
           formik={formik}
           section={section}
           setSection={updateSection}
         />
+        {isYamlSection && (
+          <YamlTypeSelector
+            yamlType={yamlType}
+            setYamlType={setYamlType}
+            formik={formik}
+          />
+        )}
         {readOnly ? null : (
-          <>
+          <div className="cancel-save-container">
             <Button
               appearance="base"
               onClick={() => {
@@ -295,10 +323,10 @@ const EditInstance: FC<Props> = ({ instance }) => {
             <FormSubmitBtn
               formik={formik}
               baseUrl={baseUrl}
-              isYaml={section === slugify(YAML_CONFIGURATION)}
+              isYaml={isYamlSection}
               disabled={hasDiskError(formik) || hasNetworkError(formik)}
             />
-          </>
+          </div>
         )}
       </FormFooterLayout>
 
--- a/src/util/yaml.spec.ts
+++ b/src/util/yaml.spec.ts
@@ -1,0 +1,125 @@
+import { expandInheritedValuesYaml } from "./yaml";
+import type { LxdInstance } from "types/instance";
+import type { LxdProfile } from "types/profile";
+
+describe("expandInheritedValuesYaml", () => {
+  const profiles: LxdProfile[] = [
+    {
+      name: "default",
+      description: "Default LXD profile",
+      config: { "boot.autostart": "true" },
+      devices: {
+        root: { path: "/", pool: "default", type: "disk" },
+      },
+    },
+    {
+      name: "cloud-init",
+      description: "Cloud-init configuration",
+      config: {
+        "cloud-init.user-data": "#cloud-config\npackage_update: true\n",
+      },
+      devices: {
+        eth0: {
+          nictype: "bridged",
+          parent: "lxdbr0",
+          type: "nic",
+          network: "lxdbr0",
+        },
+      },
+    },
+  ];
+
+  const instance: LxdInstance = {
+    name: "test-vm",
+    status: "Running",
+    profiles: ["default", "cloud-init"],
+    config: {
+      "user.meta": "local-value",
+    },
+    expanded_config: {
+      "boot.autostart": "true",
+      "user.meta": "local-value",
+      "cloud-init.user-data": "#cloud-config\npackage_update: true\n",
+    },
+    devices: {
+      root: { path: "/", pool: "default", type: "disk" },
+    },
+    expanded_devices: {
+      root: { path: "/", pool: "default", type: "disk" },
+      eth0: {
+        nictype: "bridged",
+        parent: "lxdbr0",
+        type: "nic",
+        network: "lxdbr0",
+      },
+    },
+    architecture: "x86_64",
+    created_at: "2024-01-01T00:00:00Z",
+    description: "A test VM",
+    ephemeral: false,
+    last_used_at: "2024-01-02T00:00:00Z",
+    location: "local",
+    project: "default",
+    stateful: false,
+    type: "virtual-machine",
+    snapshots: null,
+  };
+
+  it("should generate an annotated expanded YAML", () => {
+    const expectedYaml = `name: test-vm
+status: Running
+profiles:
+  - default
+  - cloud-init
+architecture: x86_64
+created_at: '2024-01-01T00:00:00Z'
+description: A test VM
+ephemeral: false
+last_used_at: '2024-01-02T00:00:00Z'
+location: local
+project: default
+stateful: false
+type: virtual-machine
+config:
+  boot.autostart: 'true' # inherited from profile: default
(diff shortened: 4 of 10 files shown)


src/components/forms/YamlExpandedSelector.tsx
```diff
@@ -0,0 +1,36 @@
...
+  setIsExpanded: (expanded: boolean) => void;
+}
+
+const YamlExpandedSelector: FC<Props> = ({ isExpanded, setIsExpanded }) => {
+  const isMediumScreen = useIsScreenBelow(mediumScreenBreakpoint);
+
+  return (
+    <div className="yaml-expanded-radio-wrapper">
+      <RadioInput
+        label={isMediumScreen ? "Local" : "Instance only"}
```
kimanhou: Any suggestion for a better text here ?

vosdev: what about:

[ ] Local  [x] Inherited

kimanhou: That could work too actually. I put Expanded to be aligned with the CLI `lxc config show INSTANCE --expanded`

edlerd: Maybe we can have both and use "Expand inherited values"?

kimanhou: Like this ? "Expand inherited values" is too long
image

src/pages/instances/EditInstance.tsx
```diff
@@ -258,49 +269,61 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
+        <footer className="edit-instance-footer bottom-btns">
+          <YamlSwitch
+            formik={formik}
+            section={section}
+            setSection={updateSection}
+            extraControls={
+              <YamlExpandedSelector
+                isExpanded={isYamlExpanded}
+                setIsExpanded={setIsYamlExpanded}
+              />
```
edlerd: Can we make this a sibling to the `YamlSwitch`, so we don't need any changes to the other component?

It seems not necessary to push it into the `YamlSwitch`.

src/pages/instances/EditInstance.tsx
```diff
@@ -83,6 +83,7 @@ const EditInstance: FC<Props> = ({ instance }) => {
   const navigate = useNavigate();
   const [version, setVersion] = useState(0);
   const { canEditInstance } = useInstanceEntitlements();
+  const [isYamlExpanded, setIsYamlExpanded] = useState(false);
```
edlerd: I confused this initially with "is the yaml editor open". Maybe a slightly different naming can help here. With second option "expandInheritedValues":
```suggestion
  const [yamlType, yamlType] = useState("local");
```

src/pages/instances/EditInstance.tsx
```diff
@@ -158,6 +159,10 @@ const EditInstance: FC<Props> = ({ instance }) => {
       formik.setFieldValue("yaml", undefined);
     }
 
+    if (newSection !== YAML_CONFIGURATION) {
+      setIsYamlExpanded(false);
+    }
+
```
edlerd:
```suggestion
```
I think we can keep the current state here as is. A user might switch back and forth, and forcing them always into the local view might not be intended.

src/pages/instances/EditInstance.tsx
```diff
@@ -258,49 +269,61 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
                 setYaml={(yaml) => {
                   ensureEditMode(formik);
                   formik.setFieldValue("yaml", yaml);
                 }}
-                readOnly={!!formik.values.editRestriction}
-                readOnlyMessage={formik.values.editRestriction}
+                readOnly={isYamlExpanded || !!formik.values.editRestriction}
+                readOnlyMessage={
+                  isYamlExpanded
+                    ? "Expanded configuration is read-only"
```
edlerd, line 287: I wanted to suggest this and am happy to see it already in place. 👍

src/pages/instances/EditInstance.tsx
```diff
@@ -258,49 +269,61 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
-              formik={formik}
-              baseUrl={baseUrl}
-              isYaml={section === slugify(YAML_CONFIGURATION)}
-              disabled={hasDiskError(formik) || hasNetworkError(formik)}
-            />
-          </>
-        )}
-      </FormFooterLayout>
+      <div className="p-bottom-controls form-footer" id="form-footer">
+        <footer className="edit-instance-footer bottom-btns">
```
edlerd: I wonder if we should keep the `FormFooterLayout` here and override styling as you already do by targeting `.edit-instance .bottom-btns` and its descendants. With the current approach, it becomes a rather bespoke component. Wdyt?

src/components/forms/YamlExpandedSelector.tsx
```diff
@@ -0,0 +1,36 @@
...
+  mediumScreenBreakpoint,
+  useIsScreenBelow,
+} from "context/useIsScreenBelow";
+
+interface Props {
+  isExpanded: boolean;
+  setIsExpanded: (expanded: boolean) => void;
+}
+
+const YamlExpandedSelector: FC<Props> = ({ isExpanded, setIsExpanded }) => {
```
edlerd:
```suggestion
const YamlTypeSelector: FC<Props> = ({ isExpanded, setIsExpanded }) => {
```
Pending the other naming suggestion -- maybe clearer?

src/util/yaml.tsx
```diff
@@ -7,3 +9,90 @@ export const yamlToObject = (yamlString: string): object => {
...
+        device &&
+        (device as unknown as Record<string, unknown>)[key] !== undefined
+      );
+    }
+    return p?.config?.[key] !== undefined;
+  });
+  return source || "profile";
+};
+
+export const getAnnotatedYaml = (
```
edlerd:
```suggestion
export const expandInheritedValuesYaml = (
```
Shall we align this with the visible option?

edlerd: Sometimes the comments are on a new line, see [this instance](https://lxd-ui-1828.demos.haus/ui/project/default/instance/DemoInstance/configuration/yaml-configuration). I think the comment should always be on the same line.

Image

src/util/yaml.tsx
```diff
@@ -7,3 +9,90 @@ export const yamlToObject = (yamlString: string): object => {
...
+      // Fully inherited device block
+      const source = reversedProfiles.find(
+        (p) => profiles.find((all) => all.name === p)?.devices?.[name],
+      );
+      sections.push(
+        `  # --- inherited from profile: ${source || "profile"} ---`,
+      );
+      sections.push(`  ${name}:`);
+      sections.push(objectToYaml(config).replace(/^/gm, "    ").trimEnd());
+      sections.push(`  # --- end of inherited ---`);
```
edlerd: I think instead of beginning and end comments, better comment directly on the start of the device:
```
  disk-device-1: # inherited from profile: disk
    path: /foo
    pool: pool1
    source: vol1
    type: disk
```

src/util/yaml.tsx
```diff
@@ -7,3 +9,90 @@ export const yamlToObject = (yamlString: string): object => {
...
+    "snapshots",
+    "state",
+    "etag",
+  ];
+  const metadata = Object.fromEntries(
+    Object.entries(instance).filter(([key]) => !exclude.includes(key)),
+  );
+  sections.push(objectToYaml(metadata).trim());
+
+  sections.push("config:");
```
edlerd: This generates two config sections where there should be only one:

https://github.com/user-attachments/assets/951003a0-5725-4b6c-871f-a7695d052155

Same for the devices if local and inherited devices are present.

src/components/forms/YamlExpandedSelector.tsx
edlerd: It looks okay in the screenshot. No?

src/util/yaml.tsx
kimanhou: Is it because the network config in cloud-init profile is an empty line ? What to do in that case ?

image

kimanhou: I did something
image

src/pages/instances/EditInstance.tsx
```diff
@@ -256,31 +271,41 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
+                }
               >
                 <YamlNotification entity="instance" docPath="/instances" />
               </YamlForm>
             )}
           </Col>
         </Row>
       </Form>
-      <FormFooterLayout>
+      <FormFooterLayout isAlignRight={false}>
```
edlerd, line 297:
```suggestion
      <FormFooterLayout isAlignRight>
```

src/pages/instances/EditInstance.tsx
```diff
@@ -256,31 +271,41 @@ const EditInstance: FC<Props> = ({ instance }) => {
...
           formik={formik}
           section={section}
           setSection={updateSection}
         />
-        {readOnly ? null : (
-          <>
+        {isYamlSection && (
+          <YamlTypeSelector yamlType={yamlType} setYamlType={setYamlType} />
+        )}
+        {readOnly || yamlType === "expandInheritedValues" ? null : (
```
edlerd:
```suggestion
        {readOnly ? null : (
```
When we go to the `expandInheritedValues`, we should also cancel the current YAML editing. That simplifies the case here, but pending changes will be discarded. That pending changes will be discarded is not necessarily expected. So I think we should also show the warning if any pending changes are present, like when going into YAML edit, starting to change, and leaving YAML mode.

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,30 @@
+import type { FC } from "react";
+import { RadioInput } from "@canonical/react-components";
+
+interface Props {
+  yamlType: "local" | "expandInheritedValues";
+  setYamlType: (yamlType: "local" | "expandInheritedValues") => void;
```
edlerd:
```suggestion
export type YamlType = "local" | "expandInheritedValues";

interface Props {
  yamlType: YamlType;
  setYamlType: (yamlType: YamlType) => void;
```
Let's introduce a type for it, we can also use it on the state in `EditInstance`.

src/pages/instances/EditInstance.tsx
edlerd: There is also a bug here currently. If we switch to `expandInheritedValues` and then leave the yaml view, the save button will be hidden.

src/util/yaml.tsx
```diff
@@ -7,3 +9,132 @@ export const yamlToObject = (yamlString: string): object => {
...
+              const source = getSourceProfile(
+                reversedProfiles,
+                profiles,
+                propKey,
+                name,
+              );
+              const lines = propYaml.split("\n");
+              lines[0] = `${lines[0]} # inherited from profile: ${source}`;
+              propYaml = lines.join("\n");
+            }
```
edlerd: I doubt that is possible. Either the device is inherited, or it is locally overridden or locally defined. But single properties on a device should never come from another source as the device itself.

src/util/yaml.tsx
```diff
@@ -7,3 +9,132 @@ export const yamlToObject = (yamlString: string): object => {
...
+    if (deviceName) {
+      const device = p?.devices?.[deviceName];
+      return (
+        device &&
+        (device as unknown as Record<string, unknown>)[key] !== undefined
+      );
+    }
+    return p?.config?.[key] !== undefined;
+  });
+  return source || "profile";
```
edlerd: I find this difficult to read and suspect a for loop over `allProfiles` with an early exit on match might be simpler. Also, I think we wouldn't need to reverse the profiles with that approach. Wdyt?

edlerd (commented):
This is so useful to discover misconfigurations! Excellent work. Some edges are a bit too sharp, but I think with minor changes it should be good to merge.

src/pages/instances/EditInstance.tsx
kimanhou: Do you want to align the YamlSwitch on the right ?
image

edlerd: Sorry, this was a mix-up on my side. I thought removing the boolean would simplify it, but it actually sets it to false, so it is needed. In my mind this was `isAlignLeft` and then the shorthand would work. 

Let's keep it as is.

src/components/forms/YamlConfirmation.tsx
```diff
@@ -16,8 +21,8 @@ const YamlConfirmation: FC<Props> = ({ onConfirm, close }) => {
       title="Confirm"
     >
       <p>
-        Switching back to guided forms will discard all changes in the YAML
-        editor.
+        Switching {isExpanded ? "back to guided forms" : "to expanded view"}{" "}
+        will discard all changes in the YAML editor.
```
edlerd:
```suggestion
        This will discard all changes in the YAML editor.
```
Maybe simplify the message and keep the interface as it was. Wdyt?

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { RadioInput, usePortal } from "@canonical/react-components";
+import YamlConfirmation from "components/forms/YamlConfirmation";
+import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
+
+export type YamlType = "local" | "expandInheritedValues";
+
+interface Props {
+  yamlType: YamlType;
+  setYamlType: (yamlType: YamlType) => void;
+  formik?: InstanceAndProfileFormikProps;
```
edlerd:
```suggestion
  formik: InstanceAndProfileFormikProps;
```
Formik should be mandatory here, no?

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const handleConfirmYamlTypeChange = () => {
+    if (formik && pendingYamlType) {
+      void formik.setFieldValue("yaml", undefined);
+      if (pendingYamlType === "expandInheritedValues") {
+        void formik.setFieldValue("readOnly", true);
+      }
+      setYamlType(pendingYamlType);
+    }
+    setPendingYamlType(null);
+    closePortal();
```
edlerd:
```suggestion
    void formik.setFieldValue("yaml", undefined);
    setYamlType(pendingYamlType as YamlType);
    setPendingYamlType(null);
    closePortal();
```
A bit more straight forward?

Also the read only seems to be unused?

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    if (formik) {
+      if (newYamlType === "expandInheritedValues") {
+        void formik.setFieldValue("yaml", undefined);
+        void formik.setFieldValue("readOnly", true);
+      }
+    }
+
+    setYamlType(newYamlType);
```
edlerd:
```suggestion
    void formik.setFieldValue("yaml", undefined);
    setYamlType(newYamlType);
```
A bit simpler? This seems to have a large overlap with `handleConfirmYamlTypeChange` so maybe we can extract the general function and reuse it.

Also read only seems to be unused?

src/util/yaml.tsx
```diff
@@ -7,3 +9,119 @@ export const yamlToObject = (yamlString: string): object => {
...
+  for (const pName of profileNames) {
+    const profile = allProfiles.find((profile) => profile.name === pName);
+    if (!profile) continue;
+
+    if (deviceName) {
+      const device = profile.devices?.[deviceName];
+      if (
+        device &&
+        (device as unknown as Record<string, unknown>)[key] !== undefined
+      ) {
```
edlerd:
```suggestion
      if (device) {
```
I think we don't care about the `type` key on the device here. If the device is found, we found the source of it.

src/util/yaml.tsx
```diff
@@ -7,3 +9,119 @@ export const yamlToObject = (yamlString: string): object => {
...
+  if (Object.keys(expandedConfig).length > 0) {
+    sections.push("config:");
+    Object.entries(expandedConfig).forEach(([key, value]) => {
+      const isLocal = Object.prototype.hasOwnProperty.call(localConfig, key);
+
+      if (!isLocal && (value === "" || value === null)) {
+        const source = getSourceProfile(reversedProfiles, profiles, key);
+        sections.push(`  ${key}: "" # inherited from profile: ${source}`);
+        return;
+      }
```
edlerd:
```suggestion
```
Does this case exist? I think the value is either set or unset, but never empty.

src/components/forms/YamlTypeSelector.tsx
kimanhou: The read only is necessary for hiding the `Cancel` and `Save` buttons when switching to expanded mode.

edlerd: Ah, I see, then let's keep it in :+1:

src/util/yaml.tsx
```diff
@@ -7,3 +9,109 @@ export const yamlToObject = (yamlString: string): object => {
...
+    } else {
+      if (profile.config?.[key] !== undefined) {
+        return pName;
+      }
+    }
+  }
+  return "profile";
+};
+
+export const expandInheritedValuesYaml = (
```
edlerd, line 37: Can we add a unit sec test for this with the main cases?

I think that will help maintain this in the long term. Especially if we discover edge cases to add to the test suite, the newly discovered case, and making sure we don't break something accidentally.

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,83 @@
...
+  yamlType: YamlType;
+  setYamlType: (yamlType: YamlType) => void;
+  formik: InstanceAndProfileFormikProps;
+}
+
+const YamlTypeSelector: FC<Props> = ({ yamlType, setYamlType, formik }) => {
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const [pendingYamlType, setPendingYamlType] = useState<YamlType | null>(null);
+
+  const performYamlTypeSwitch = (newYamlType: YamlType) => {
```
edlerd:
```suggestion
  const performSwitch = (newYamlType: YamlType) => {
```

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,83 @@
...
+    void formik.setFieldValue("yaml", undefined);
+
+    if (newYamlType === "expandInheritedValues") {
+      void formik.setFieldValue("readOnly", true);
+    }
+
+    setYamlType(newYamlType);
+  };
+
+  const handleYamlTypeChange = (
```
edlerd:
```suggestion
  const handleChange = (
```

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (formik.values.yaml) {
+      setPendingYamlType(newYamlType);
+      openPortal(e);
+      return;
+    }
+
+    performYamlTypeSwitch(newYamlType);
+  };
+
+  const handleConfirmYamlTypeChange = () => {
```
edlerd:
```suggestion
  const handleConfirm = () => {
```

src/components/forms/YamlTypeSelector.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (pendingYamlType) {
+      performYamlTypeSwitch(pendingYamlType);
+    }
+    setPendingYamlType(null);
+    closePortal();
+  };
+
+  const handleCancelYamlTypeChange = () => {
```
edlerd:
```suggestion
  const handleCancel = () => {
```

src/util/yaml.tsx
```diff
@@ -7,3 +9,109 @@ export const yamlToObject = (yamlString: string): object => {
...
   return dumpYaml(obj, { lineWidth: -1 });
 };
+
+const getSourceProfile = (
+  profileNames: string[],
+  allProfiles: LxdProfile[],
+  key: string,
+  deviceName?: string,
+): string => {
+  for (const pName of profileNames) {
```
edlerd:
```suggestion
  for (const profileName of profileNames) {
```

src/util/yaml.tsx
kimanhou: I added some tests but the formatting can be a bit different

src/util/yaml.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    );
+  });
+
+  it("should clean up excessive newlines while keeping padding for blocks", () => {
+    const result = expandInheritedValuesYaml(mockInstance, mockProfiles);
+    // Ensure we don't have triple newlines from the join/padding logic
+    expect(result).not.toMatch(/\n{3,}/);
+    // Ensure padding exists after the cloud-init block
+    expect(result).toMatch(/package_update: true\n\s*\n\s*/);
+  });
```
edlerd: The cases are all using the same data and not really covering the cases. Like the one below, does not truly ensure there is no inheritance comment on the line:
```
    // Local config should not have a comment
    expect(result).toContain("user.meta: local-value");
```
How about a slightly different approach with having the expected YAML output here in the test file? Then a single test case can match the output of the function against that.

src/util/yaml.spec.ts
```diff
@@ -0,0 +1,130 @@
+import { expandInheritedValuesYaml } from "./yaml";
+import type { LxdInstance } from "types/instance";
+import type { LxdProfile } from "types/profile";
+
+describe("expandInheritedValuesYaml", () => {
+  const mockProfiles: LxdProfile[] = [
```
edlerd:
```suggestion
  const profiles: LxdProfile[] = [
```

src/util/yaml.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+          nictype: "bridged",
+          parent: "lxdbr0",
+          type: "nic",
+          network: "lxdbr0",
+        },
+      },
+    },
+  ];
+
+  const mockInstance: LxdInstance = {
```
edlerd:
```suggestion
  const instance: LxdInstance = {
```

edlerd (dismissed):
LGTM, thanks for adding the tests. One idea for the test, though good to go as is from my side :)

PR #1831, opened 2026-02-26
test(ACLs): e2e tests on applying ACLs on network and instance/profile

kimanhou:
## Done

- Add e2e tests on applying ACLs to a network
- Add e2e tests on applying ACLs to a nic device on a profile

## QA

1. Run the LXD-UI:
    - Rely on CI passing

## Screenshots

image

### DIFF
--- a/src/pages/networks/forms/NetworkDefaultACLSelector.tsx
+++ b/src/pages/networks/forms/NetworkDefaultACLSelector.tsx
@@ -1,6 +1,6 @@
 import type { FC } from "react";
 import { Fragment } from "react";
-import { Icon, Select } from "@canonical/react-components";
+import { Icon, Label, Select } from "@canonical/react-components";
 import { conjugateACLAction } from "util/helpers";
 
 export type Direction = "Ingress" | "Egress";
@@ -29,6 +29,9 @@ const NetworkDefaultACLSelector: FC<Props> = ({
     })),
   ];
 
+  const getId = (direction: Direction) =>
+    `${direction.toLowerCase()}-default-action`;
+
   return (
     <div>
       <div className="u-sv2">When no ACL rule matches:</div>
@@ -36,13 +39,13 @@ const NetworkDefaultACLSelector: FC<Props> = ({
         {DIRECTIONS.map((direction) => {
           return (
             <Fragment key={direction}>
-              <div>
+              <Label forId={getId(direction)}>
                 <Icon
                   name={direction === "Egress" ? "arrow-left" : "arrow-right"}
                   className="network-default-acl-icon"
                 />
-                {direction} traffic will be{" "}
-              </div>
+                {direction} traffic is{" "}
+              </Label>
               <div>
                 <Select
                   className="u-no-margin--bottom"
@@ -52,6 +55,7 @@ const NetworkDefaultACLSelector: FC<Props> = ({
                   }}
                   value={values?.[direction]}
                   disabled={disabled}
+                  id={getId(direction)}
                 />
               </div>
             </Fragment>
--- a/tests/network-acls.spec.ts
+++ b/tests/network-acls.spec.ts
@@ -1,11 +1,30 @@
-import { test } from "./fixtures/lxd-test";
+import { test, expect } from "./fixtures/lxd-test";
+import { skipIfNotClustered, skipIfNotSupported } from "./helpers/cluster";
 import {
+  createNetwork,
+  deleteNetwork,
+  getNetworkLink,
+  randomNetworkName,
+  visitNetwork,
+} from "./helpers/network";
+import {
+  attachNetworkAclsToNetwork,
   createNetworkAcl,
   deleteNetworkAcl,
+  detachNetworkAclsFromNetwork,
   editNetworkAcl,
   randomNetworkAclName,
   renameNetworkAcl,
+  selectNetworkAcls,
+  setAclDefaults,
 } from "./helpers/network-acls";
+import {
+  deleteProfile,
+  finishProfileCreation,
+  randomProfileName,
+  startProfileCreation,
+  visitProfile,
+} from "./helpers/profile";
 
 test("create a network acl", async ({ page }) => {
   const networkAcl = randomNetworkAclName();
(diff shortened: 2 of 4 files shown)


tests/network-acls.spec.ts
```diff
@@ -15,3 +33,201 @@ test("create a network acl", async ({ page }) => {
...
+
+  test("apply 2 ACLs to a network", async ({ page }) => {
+    await attachNetworkAclsToNetwork(page, [acl, acl2], network);
+
+    const defaultActionContainer = page.locator(
+      ".network-default-acl-selector",
+    );
+    const egressSelect = defaultActionContainer.locator("select").nth(0);
+    await expect(egressSelect).toBeEnabled();
+    await egressSelect.selectOption("allow");
```
edlerd:
```suggestion
    await page.getByLabel("Egress traffic").selectOption("allow");
```
I think there is an accessibility bug here, as the label is neither rendered as a label nor linked to the input. Can you please fix it as a drive-by and simplify the selection test here?

edlerd: Also, please align the label from `Egress traffic will be` to `Egress traffic is` to match the updated read mode. Both were aligned before, but the edit mode label changed recently.

tests/network-acls.spec.ts
```diff
@@ -15,3 +33,201 @@ test("create a network acl", async ({ page }) => {
...
+
+    const defaultActionContainer = page.locator(
+      ".network-default-acl-selector",
+    );
+    const egressSelect = defaultActionContainer.locator("select").nth(0);
+    await expect(egressSelect).toBeEnabled();
+    await egressSelect.selectOption("reject");
+    const ingressSelect = defaultActionContainer.locator("select").nth(1);
+    await expect(ingressSelect).toBeEnabled();
+    await ingressSelect.selectOption("allow");
```
edlerd: we should put this in a helper to reuse from above. Something like `setAclDefaults("reject", "allow")`

tests/network-acls.spec.ts
```diff
@@ -15,3 +33,201 @@ test("create a network acl", async ({ page }) => {
...
+
+    const profile = randomProfileName();
+    await startProfileCreation(page, profile);
+    await page.getByText("Network", { exact: true }).click();
+    await page.getByRole("button", { name: "Attach network" }).click();
+    await page.locator("#network").click();
+    await expect(page.getByText("NameTypeACLs")).toBeVisible();
+    await page.getByRole("listbox").getByText(network, { exact: true }).click();
+
+    const aclList = page.locator("#network-acl-selector");
```
edlerd: Can we use `getByLabel("ACLs")` here instead, so we cover accessibility and not the implementation details of css?

tests/network-acls.spec.ts
```diff
@@ -15,3 +33,201 @@ test("create a network acl", async ({ page }) => {
...
+    await page.getByRole("button", { name: "Save" }).click();
+    await page.waitForTimeout(5000);
+
+    const profile = randomProfileName();
+    await startProfileCreation(page, profile);
+    await page.getByText("Network", { exact: true }).click();
+    await page.getByRole("button", { name: "Attach network" }).click();
+    await page.locator("#network").click();
+    await expect(page.getByText("NameTypeACLs")).toBeVisible();
+    await page.getByRole("listbox").getByText(network, { exact: true }).click();
```
edlerd: Let's use the accessibility labels here. This is untested, but we use something similar elsewhere for the custom select:
```
  await page.getByRole("button", { name: "Network" }).click();
  await page.getByLabel("sub").getByText(network).first().click();
```

tests/network-acls.spec.ts
```diff
@@ -15,3 +33,201 @@ test("create a network acl", async ({ page }) => {
...
+      page.getByText("Network must be of type OVN to customize ACLs."),
+    ).not.toBeVisible();
+    await expect(egressSelect).toBeDisabled();
+    await expect(ingressSelect).toBeDisabled();
+
+    await page.getByLabel("ACLs").click();
+    await selectNetworkAcls(page, [acl, acl2]);
+
+    await expect(egressSelect).toBeEnabled();
+    await expect(ingressSelect).toBeEnabled();
```
edlerd:
```suggestion
```
The two lines below imply those to be enabled.

tests/network-acls.spec.ts
```diff
@@ -15,3 +33,201 @@ test("create a network acl", async ({ page }) => {
...
+      page.getByText("Network must be of type OVN to customize ACLs."),
+    ).toBeVisible();
+
+    const defaultActionContainer = page.locator(
+      ".network-default-acl-selector",
+    );
+    const egressSelect = defaultActionContainer.locator("select").nth(0);
+    await expect(egressSelect).toBeDisabled();
+    const ingressSelect = defaultActionContainer.locator("select").nth(1);
+    await expect(ingressSelect).toBeDisabled();
```
edlerd: This seems to repeat things already tested in the case above.

tests/network-acls.spec.ts
kimanhou: I had to pass the `label` to the `MultiSelect` inside `NetworkAclSelector` for it to work.

tests/network-acls.spec.ts
kimanhou: Not exactly because in `attach network with ACLs to a profile`, we attach a bridge network with ACLs.
In `attach network to a profile then add ACLs to the nic device`, we attach a bridge network with no ACLs first, then an OVN network with no ACLs then add ACLs.
I want to make sure that when changing from a bridge network to an ovn network, ACLs selector and default actions go from disabled to enabled.

src/pages/networks/forms/NetworkDefaultACLSelector.tsx
```diff
@@ -36,13 +36,13 @@ const NetworkDefaultACLSelector: FC<Props> = ({
         {DIRECTIONS.map((direction) => {
           return (
             <Fragment key={direction}>
-              <div>
+              <Label forId={`${direction.toLowerCase()}-default-action`}>
```
edlerd:
```suggestion
              <Label forId={labelId}>
```
Please assign this to a const above to have a single source for it.

tests/network-acls.spec.ts
edlerd: Huh, I am confused why that was needed. There is already a label with the value and linking to the right ID. Testing it locally: I removed the label forwarding to MultiSelect in NetworkAclSelector. It also worked to click on the "ACLs" text on the left to open the dropdown with ACLs on the right. So playwright should find it as the label is properly configured. I think with forwarding there are two labels for the ACL selector, which seems not right.

tests/network-acls.spec.ts
edlerd: I see, good idea :+1:

tests/network-acls.spec.ts
kimanhou: When I remove the `label` forwarding to `MultiSelect` in `NetworkAclSelector` and run the test locally, I get this error:
```
Error: expect(locator).toContainText(expected) failed

Locator: getByLabel('ACLs')
Expected substring: "test-ymthph, test-t1jwe"
Error: element(s) not found
```
I think it's because getByLabel also looks at the `aria-label` and the `MultiSelect` component sets it to the placeholder, which is "-" when disabled.
```
aria-label={label || placeholder || "Select items"}
```

```
  const getPlaceholder = () => {
    if (!canSelectManualAcls) {
      return "-";
    }
    if (!hasAcls) {
      return "No ACLs available";
    }
    if (hasAcls) {
      return "Select ACLs";
    }
  };
```
I do confirm that clicking on ACLs opens the ACLs dropdown list.

PR #1838, opened 2026-03-02
fix(TLS login): Update CLI to create TLS identity

kimanhou:
## Done

- TLS login > Certificate add : update CLI so that it doesn't throw an error if `admins` group does not exist

## QA

1. Run the LXD-UI:
    - Log out of LXD UI
    - Delete `admins` group: `lxc auth group delete admins`
    - From Login page, click on TLS login, then skip to step 2.
    - Make sure you understand you understand what the command does.
    - Copy paste the command line in a terminal
    - Make sure no errors is thrown

## Screenshots
image

image

### DIFF
--- a/src/pages/login/CertificateAddToken.tsx
+++ b/src/pages/login/CertificateAddToken.tsx
@@ -21,7 +21,7 @@ const CertificateAddToken: FC = () => {
   const { isAuthenticated, isAuthLoading, authMethod } = useAuth();
   const notify = useNotify();
   const identityTrustTokenCommand =
-    "if ! lxc auth group show admins ; then lxc auth group create admins ; lxc auth group permission add admins server admin ; fi ; lxc auth identity create tls/lxd-ui --group admins";
+    "if ! lxc auth group show admins >/dev/null 2>&1; then lxc auth group create admins && lxc auth group permission add admins server admin; fi; lxc auth identity create tls/lxd-ui --group admins";
 
   if (isAuthLoading) {
     return <Spinner className="u-loader" text="Loading..." isMainComponent />;
@@ -62,12 +62,14 @@ const CertificateAddToken: FC = () => {
                       </p>
                       <div>
                         First, the command checks to see if there is an auth
-                        group <code>admins</code>.
+                        group <code>admins</code>. The{" "}
+                        <code>{`>/dev/null 2>&1`}</code> part ensures that if
+                        the group is missing, no error is shown.
                       </div>
                       <CodeSnippet
                         blocks={[
                           {
-                            code: `if ! lxc auth group show admins`,
+                            code: `if ! lxc auth group show admins >/dev/null 2>&1;`,
                             wrapLines: true,
                           },
                         ]}


src/pages/login/CertificateAddToken.tsx
```diff
@@ -21,7 +21,7 @@ const CertificateAddToken: FC = () => {
   const { isAuthenticated, isAuthLoading, authMethod } = useAuth();
   const notify = useNotify();
   const identityTrustTokenCommand =
-    "if ! lxc auth group show admins ; then lxc auth group create admins ; lxc auth group permission add admins server admin ; fi ; lxc auth identity create tls/lxd-ui --group admins";
```
edlerd, line 24: Let's keep this as is, but only suppress the error in the first check:
`if ! lxc auth group show admins >/dev/null 2>&1`

And adjust the explanation for it. The rest can stay as is. I think it is a bit simpler to follow that way. Wdyt?

kimanhou: I'm confused. Is what you're suggesting this ?
```
if ! lxc auth group show admins >/dev/null 2>&1; then lxc auth group permission add admins server admin; fi; lxc auth identity create tls/lxd-ui --group admins
```
Then it throws an error:
```
Error: AuthGroup not found
Error: Failed to create pending TLS identity: One or more groups were not found: "admins"
```

edlerd: We might need a subshell: `if ! $(lxc auth group show admins >/dev/null 2>&1) ; then`

kimanhou:
```
if ! $(lxc auth group show admins >/dev/null 2>&1) ; then lxc auth group permission add admins server admin; fi; lxc auth identity create tls/lxd-ui --group admins
Error: AuthGroup not found
Error: Failed to create pending TLS identity: One or more groups were not found: "admins"
```
Using $(...) tells the shell to run the command and then treat the output as a command itself. Since `lxc auth group show admins` produces no text output (it’s redirected to /dev/null), the shell tries to execute an "empty" command, which causes confusion.

edlerd: Hm, that is odd, it works for me. Is there somethign shell specific maybe?
```
$ if ! lxc auth group show non-existing-group >/dev/null 2>&1 ; then echo "exists not " ; fi
exists not 

$ if ! lxc auth group show admins >/dev/null 2>&1 ; then echo "exists not " ; fi

$ lxc auth group ls
+-------------------------+-------------------------+
|          NAME           |       DESCRIPTION       |
+-------------------------+-------------------------+
| read-only               |                         |
+-------------------------+-------------------------+

```

edlerd: You need two commands inside the conditional. One to create the group and a second one to assign server admin permissions to it.

kimanhou: Like this ?
```
if ! lxc auth group show admins >/dev/null 2>&1; then 
    lxc auth group create admins
    lxc auth group permission add admins server admin
fi
lxc auth identity create tls/lxd-ui --group admins
```

But then if `admins` group already exists, we don't ensure it has admin permission. Is that okay ?

edlerd: > But then if `admins` group already exists, we don't ensure it has admin permission. Is that okay ?

Yes, I think we always neglected that case, if a user messed with the admins group. I think in the future the admins group will be locked anyway, so it always exists and has fixed permissions. So eventually we might simplify the steps here in future versions.

kimanhou: Okay then

PR #1844, opened 2026-03-03
feat(InstanceList): display instance that are being uploaded from a file

kimanhou:
## Done

- Display uploaded instance from file in instance list

## QA

1. Run the LXD-UI:
    - If you don't have a `.tar` file ready, export an instance.
    - On instance list page, click on Create instance, then Upload instance file. Select the file and click Upload and create
    - Make sure the uploaded instance appears in the instance list.

## Screenshots

image

### DIFF
--- a/src/components/InstanceCreationProgress.tsx
+++ b/src/components/InstanceCreationProgress.tsx
@@ -1,0 +1,37 @@
+import type { FC } from "react";
+import type { LxdOperation } from "types/operation";
+import { getInstanceName, isRestoringBackup } from "util/operations";
+
+interface Props {
+  operation: LxdOperation;
+}
+
+const InstanceCreationProgress: FC<Props> = ({ operation }) => {
+  const metadata = operation.metadata ?? {};
+
+  if (isRestoringBackup(operation)) {
+    return `Restoring backup for ${getInstanceName(operation)}`;
+  }
+
+  if (metadata.format_progress_progress) {
+    return <div>{metadata.format_progress_progress}</div>;
+  }
+
+  const EXCLUDED_KEYS = ["entity_url", "fs"];
+  const firstValidEntry = Object.entries(metadata).find(
+    ([key]) => !EXCLUDED_KEYS.includes(key),
+  );
+
+  if (firstValidEntry) {
+    const [key, value] = firstValidEntry;
+    return (
+      <div>
+        {key}: {value}
+      </div>
+    );
+  }
+
+  return null;
+};
+
+export default InstanceCreationProgress;
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -56,7 +56,11 @@ import {
   TYPE,
   USER_HIDEABLE_COLUMNS,
 } from "util/instanceTable";
-import { getInstanceName } from "util/operations";
+import {
+  getInstanceName,
+  isCreatingInstance,
+  isRestoringBackup,
+} from "util/operations";
 import NotificationRow from "components/NotificationRow";
 import SelectedTableNotification from "components/SelectedTableNotification";
 import HelpLink from "components/HelpLink";
@@ -81,6 +85,7 @@ import TruncatedList from "components/TruncatedList";
 import ClusterMemberRichChip from "pages/cluster/ClusterMemberRichChip";
 import { ROOT_PATH } from "util/rootPath";
 import ProjectRichChip from "pages/projects/ProjectRichChip";
+import InstanceCreationProgress from "components/InstanceCreationProgress";
 
 const loadHidden = () => {
   const saved = localStorage.getItem("instanceListHiddenColumns");
@@ -189,10 +194,11 @@ const InstanceList: FC = () => {
     .concat(operationList?.success ?? [])
     .filter((operation) => {
       const name = getInstanceName(operation);
-      const isCreating = operation.description === "Creating instance";
+      const isCreatingOrRestoring =
+        isCreatingInstance(operation) || isRestoringBackup(operation);
       const isProcessing = processingNames.includes(name);
       const isMigrating = migrationNames.includes(name);
-      if (!isCreating || isProcessing || isMigrating || !name) {
+      if (!isCreatingOrRestoring || isProcessing || isMigrating || !name) {
         return false;
       }
       const isInInstanceList = instances.some((item) => item.name === name);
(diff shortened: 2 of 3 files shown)


src/pages/instances/InstanceList.tsx
```diff
@@ -376,13 +387,16 @@ const InstanceList: FC = () => {
...
                     <i key={JSON.stringify(operation.metadata ?? {})}>
-                      {Object.entries(operation.metadata ?? {})
-                        .slice(0, 1)
-                        .map(([key, value], index) => (
-                          <div key={index}>
-                            {key}: {value}
-                          </div>
-                        ))}
+                      {isRestoringBackup(operation) &&
+                        `Restoring backup for ${getInstanceName(operation)}`}
```
edlerd: In case of uploading a VMDK file, we should surface the nested `metadata.format_progress_progress` content from the operation:

Image

src/pages/instances/InstanceList.tsx
edlerd, line 379: In case of uploading a LXD export, the instance name is not available, just the name of the instance on export. So this wrongly reports `fitting-lamb`, though the instance import name was set to `f23itting-lamb-2026-03-03T16-31-38-tar-import23`.

Image

src/util/operations.tsx
```diff
@@ -109,3 +109,14 @@ export const getFingerprint = (result: LxdEvent | LxdOperation): string => {
...
+
+export const isRestoringBackup = (operation: LxdOperation): boolean => {
+  return operation.description === "Restoring backup";
+};
+
+export const isCreatingInstance = (operation: LxdOperation): boolean => {
+  return (
+    operation.description === "Creating instance" ||
+    isRestoringBackup(operation)
+  );
```
edlerd:
```suggestion
  return (
    operation.description === "Creating instance"
  );
```
I would keep this more aligned with the operation and combine the two cases in the use case. Wdyt?

src/pages/instances/InstanceList.tsx
kimanhou: Yes, that's why I had to update the instance list. Otherwise `fitting-lamb` had the `Setting up` status.
Since the operation doesn't include `f23itting-lamb-2026-03-03T16-31-38-tar-import23`, what should we display ?

src/pages/instances/InstanceList.tsx
kimanhou: Should we display all `metadata` then ?

image

edlerd: I would filter out the entity URL that is not helpful. The URL hides the interesting progress report that comes afterward. We could have some logic to figure out if `format_progress_progress` is set and show its content, then fall back to other filtered metadata. Wdyt?

src/pages/instances/InstanceList.tsx
edlerd: Can you ask if we can extend the operation metadata to include the name in the API? Or we could have a placeholder for the importing name. I think we should not show a wrong name.

kimanhou: I opened a [Github issue](https://github.com/canonical/lxd/issues/17809) on backend. What should we do in the meantime ?

kimanhou: A PR has already been raised to send the target name in the operation. I think we can wait until that's merged.

edlerd: Thanks for raising it and getting it fixed :)

I think we can then progress with this one, assuming the correct name will appear as soon as the other PR is merged.

src/pages/instances/InstanceList.tsx
edlerd: Maybe extract this logic into a dedicated `InstanceCreationProgress` component.

src/pages/instances/InstanceList.tsx
```diff
@@ -202,7 +205,11 @@ const InstanceList: FC = () => {
         return false;
       }
 
-      creationNames.push(name);
+      // TODO: remove if when backend returns the target name
+      if (!isInInstanceList) {
+        creationNames.push(name);
+      }
+
```
edlerd: I don't understand why we filter here. Can we keep it as is?

src/pages/instances/InstanceList.tsx
```diff
@@ -189,7 +190,9 @@ const InstanceList: FC = () => {
     .concat(operationList?.success ?? [])
     .filter((operation) => {
       const name = getInstanceName(operation);
-      const isCreating = operation.description === "Creating instance";
+      const isCreating =
+        operation.description === "Creating instance" ||
+        isRestoringBackup(operation);
```
edlerd: Let's have a helper for both or for none of the two. I am fine either way.

edlerd: Can we find a better name for `isCreating`? It seems unexpected to include both `Creating instance` and `Restoring backup`.

src/components/InstanceCreationProgress.tsx
```diff
@@ -0,0 +1,36 @@
...
+  if (isRestoringBackup(operation)) {
+    return `Restoring backup for ${getInstanceName(operation)}`;
+  }
+
+  if (metadata.format_progress_progress) {
+    return <div>{metadata.format_progress_progress}</div>;
+  }
+
+  const firstValidEntry = Object.entries(metadata).find(
+    ([key]) => !key.includes("url") && key !== "entity_url",
```
edlerd: Are there other urls than the `entity_url` we have to match for here? I think we should also filter out the `fs` key.

src/pages/instances/InstanceList.tsx
kimanhou: We can but until Mark's PR is merged, it will look like this when restoring a backup of iso-instance:
image

src/components/InstanceCreationProgress.tsx
kimanhou: We used to not filter anything. I added the filtering out of `fs`

src/pages/instances/InstanceList.tsx
edlerd: I think it is ok to be shortly broken until the fix lands in the API.

src/components/InstanceCreationProgress.tsx
edlerd: Thanks :)

I was wondering why the extra `!key.includes("url")`?

src/pages/instances/InstanceList.tsx
kimanhou: The PR is merged in backend ! All good

src/components/InstanceCreationProgress.tsx
kimanhou: A copy paste gone wrong ... Better now with the array

PR #1856, opened 2026-03-05
fix(footer layout): of Create ACLs, Create network and Create forwards

kimanhou:
## Done

- Create ACLs, Create network, Create network forwards: the footer was too long

## QA

1. Run the LXD-UI:
    - Make sure that the buttons are horizontally aligned with the rest of the page to the right

## Screenshots

Create ACL before
image

Create ACL after
image

Create network before
image

Create network after
image

Create network forwards before
image

Create network forwards after
image

### DIFF
--- a/src/sass/_forms.scss
+++ b/src/sass/_forms.scss
@@ -338,6 +338,8 @@
 }
 
 .create-instance,
+.create-network,
+.create-network-acl,
 .create-profile,
 .create-project,
 .edit-project,
@@ -479,8 +481,6 @@
 }
 
 .create-instance,
-.create-network,
-.create-network-acl,
 .create-profile,
 .create-project,
 .edit-project,


src/sass/_forms.scss
```diff
@@ -488,6 +488,7 @@
 .storage-volume-form {
   .form-footer {
     margin-top: 1rem;
+    max-width: 70.5rem;
```
edlerd: I suspect the problem was introduced in https://github.com/canonical/lxd-ui/commit/c63b99afdc6a9829018704a8eb9222530bc37001 and partially fixed with https://github.com/canonical/lxd-ui/commit/a80d7bccbbad1f5ec8dfc15ac064655f2869bb55

Shall we restore the original version before the first change?

kimanhou: I don't know what exactly triggered the bug. Putting back the original version of `FormFooterLayout` doesn't fix the issue. Which original version are you thinking of ?

```
import type { FC, PropsWithChildren } from "react";

const FormFooterLayout: FC<PropsWithChildren> = ({ children }) => {
  return (
    <div className="p-bottom-controls form-footer" id="form-footer">
      <footer className="u-align--right bottom-btns">{children}</footer>
    </div>
  );
};

export default FormFooterLayout;
```

edlerd: Sorry I was a bit too unspecific. I meant the changes to `src/sass/_forms.scss`.

kimanhou: Aaaaaah I remember the chaos that lint made me create that day ! It should be fixed now.

PR #1858, opened 2026-03-06
feat(Create network forwards): Use PrefixedIpInput

kimanhou:
## Done

- Use PrefixedIpInput in CreateNetworkForwards

## QA

1. Run the LXD-UI:
    - Create a network forwards.
    - Edit a network forwards: you should not be able to change the target address family

## Screenshots

image

### DIFF
--- a/src/pages/networks/forms/NetworkForwardForm.tsx
+++ b/src/pages/networks/forms/NetworkForwardForm.tsx
@@ -8,6 +8,7 @@ import {
   Input,
   Label,
   Notification,
+  PrefixedIpInput,
   RadioInput,
   Row,
   useListener,
@@ -17,7 +18,13 @@ import type { FormikProps } from "formik/dist/types";
 import * as Yup from "yup";
 import type { LxdNetwork, LxdNetworkForward } from "types/network";
 import { updateMaxHeight } from "util/updateMaxHeight";
-import { isTypeOvn, testValidIp, testValidPort } from "util/networks";
+import {
+  isTypeOvn,
+  testValidIp,
+  testValidPort,
+  getIpAddressFamily,
+  getCidr,
+} from "util/networks";
 import NotificationRow from "components/NotificationRow";
 import NetworkForwardFormPorts from "pages/networks/forms/NetworkForwardFormPorts";
 import ScrollableForm from "components/ScrollableForm";
@@ -50,6 +57,11 @@ export const NetworkForwardSchema = Yup.object().shape({
   listenAddress: Yup.string()
     .test("valid-ip", "Invalid IP address", testValidIp)
     .required("Listen address is required"),
+  defaultTargetAddress: Yup.string().test(
+    "valid-ip",
+    "Invalid IP address",
+    (value) => !value || testValidIp(value),
+  ),
   ports: Yup.array().of(
     Yup.object().shape({
       listenPort: Yup.string()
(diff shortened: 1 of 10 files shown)


src/types/forms/network.d.ts
```diff
@@ -59,3 +59,5 @@ export interface NetworkFormValues {
   security_acls_default_egress?: string;
   security_acls_default_ingress?: string;
 }
+
+export type IpAddressFamily = "IPv4" | "IPv6";
```
edlerd, line 63: Good idea to extract this!

src/pages/networks/forms/NetworkForwardForm.tsx
```diff
@@ -188,22 +200,67 @@ const NetworkForwardForm: FC<Props> = ({ formik, isEdit, network }) => {
...
-                  <br />
-                  Must be from the network <b>{network?.name}</b>.
-                </>
-              }
-              placeholder="Enter IP address"
-              stacked
-            />
+            <Row>
+              <Col size={4}>
+                <Label>Target address family</Label>
```
edlerd: Can we add some help text to this one, similar to the existing "Listen address can't be changed after creation." to clarify why the address is disabled?

src/pages/networks/forms/NetworkForwardFormPorts.tsx
```diff
@@ -97,24 +107,32 @@ const NetworkForwardFormPorts: FC<Props> = ({ formik, network }) => {
                 />
               </td>
               <td className="target-address">
-                <Input
-                  {...formik.getFieldProps(`ports.${index}.targetAddress`)}
+                <PrefixedIpInput
```
edlerd, line 108: On large screens, the editable part of the input is not aligned: 

Image

This might be a bug in the upstream component. Or we could have a local workaround?

src/pages/networks/forms/NetworkForwardForm.tsx
```diff
@@ -74,6 +81,8 @@ interface Props {
 
 const NetworkForwardForm: FC<Props> = ({ formik, isEdit, network }) => {
   const notify = useNotify();
+  const [targetAddressFamily, setTargetAddressFamily] =
+    useState<IpAddressFamily>("IPv4");
```
edlerd: Let's figure out the default from the listen address.

Actually, we might not need the "Target address family" input. Because the family is determined by the listen address family.

So maybe disable all other inputs until the "Listen address" has a valid value. And then we can calculate the correct family from the listen address value. So we can remove the family input and this state.

edlerd (commented):
Some thoughts on this below.

src/pages/networks/forms/NetworkForwardForm.tsx
kimanhou: Only on Edit right ?

image

edlerd: Yes, that was the idea. Though considering the other comment -- we might not need those radios after all.

src/pages/networks/forms/NetworkForwardFormPorts.tsx
kimanhou: Does this have to do with this in `_network_device_panel.scss` ?
@omarelkashef 

```
@include extra-large {
  .prefixed-input__input {
    padding-top: 0.35rem !important;
  }
}
```

src/pages/networks/forms/NetworkForwardForm.tsx
kimanhou: Nope, I removed them

src/pages/networks/forms/NetworkForwardForm.tsx
```diff
@@ -188,22 +211,41 @@ const NetworkForwardForm: FC<Props> = ({ formik, isEdit, network }) => {
...
+                  }}
+                  onBlur={() => {
+                    void formik.setFieldTouched("defaultTargetAddress", true);
+                  }}
+                  error={
+                    formik.touched.defaultTargetAddress
+                      ? formik.errors.defaultTargetAddress
+                      : undefined
+                  }
+                  disabled={isEdit || !isListenAddressValid}
```
edlerd:
```suggestion
                  disabled={!isListenAddressValid}
```
I think this is allowed to be changed on edit.

kimanhou: Indeed

src/pages/networks/forms/NetworkForwardFormPorts.tsx
kimanhou: I fixed it by scoping Omar's fix so the network device side panel but I can't explain why.

src/sass/_network_device_panel.scss
```diff
@@ -1,5 +1,7 @@
 @include extra-large {
-  .prefixed-input__input {
-    padding-top: 0.35rem !important;
+  .network-device-panel-content {
```
edlerd: Namespacing fixes one, but breaks the other case

Image

PR #1860, opened 2026-03-06
fix: make makeNetworkOvnUplink in tests/helpers/network.ts idempotent

omarelkashef:
WD-34596

## Done

- make makeNetworkOvnUplink in tests/helpers/network.ts idempotent

## QA

1. Run the LXD-UI:
    - Run CI tests and make sure that calling makeNetworkOvnUplink twice with the same parameters in the same test does not cause problems.

### DIFF
--- a/tests/helpers/network.ts
+++ b/tests/helpers/network.ts
@@ -336,20 +336,33 @@ export const makeNetworkOvnUplink = async (
   } = options;
 
   await visitNetwork(page, network);
-  await page
-    .locator(".ip-address", { hasText: "IPv4 address" })
-    .getByRole("button")
-    .click();
+  let hasChanges = false;
+  const ipv4Container = page.locator(".ip-address", {
+    hasText: "IPv4 address",
+  });
 
-  await setNetworkIP(page, CIDR, "IPv4");
+  if (!(await ipv4Container.getByText(CIDR).isVisible())) {
+    hasChanges = true;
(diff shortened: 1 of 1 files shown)


tests/helpers/network.ts
```diff
@@ -336,20 +336,33 @@ export const makeNetworkOvnUplink = async (
...
+  const dhcpRangesRow = page.locator('tr[name*="ipv4_dhcp_ranges"]');
+  if (!(await dhcpRangesRow.getByText(dhcpIpv4Range).isVisible())) {
+    hasChanges = true;
+    await setNetworkDHCPRanges(page, dhcpIpv4Range);
+  }
 
-  await page.getByRole("button", { name: "Save" }).click();
+  if (hasChanges) {
+    await page.getByRole("button", { name: "Save" }).click();
+  }
```
edlerd, line 365: Maybe click cancel in the else case to ensure we land on the instance list after this?

Not relevant now, as all usages continue with navigation after calling it, but might be relevant in the future if we reuse this function.

omarelkashef: Good idea, but we can navigate to  /ui instead because if all network config matches, there will be no cancel button.

edlerd: Oh right, this is edit, not create. Then we should stay on the page, and all is good as is. Because after saving, we would be on the same page.

PR #1861, opened 2026-03-07
feat(instances): support selective state recursion

### LINKED ISSUE #1857: Optimize instance state queries by fetching only required fields
Now that https://github.com/canonical/lxd/pull/17378 and https://github.com/canonical/lxd/pull/17679 are merged to `main` on LXD we can  make use of new query parameters to optimize instance state queries in lxd-ui.  This was also added to https://github.com/canonical/pylxd/issues/706  recently. Happy to help contribute this feature and send in a pull request. But we can discuss this first before making any code changes.

gajeshbhat:
## Done

- Support the `instances_state_selective_recursion` API extension in instance queries
- Request only `state.network` when selective recursion is supported
- Keep fine-grained permissions working by fetching `access_entitlements` separately and merging them client-side
- Fall back to regular `recursion=2` when the extension is unavailable
- Add targeted tests for selective recursion, fallback behavior, and separate entitlement fetching

Fixes #1857

## QA

- I tested locally against two Multipass-backed LXD servers:
  - `6/stable` for the optimized path (`instances_state_selective_recursion`)
  - `5.21/stable` for the fallback path

- Provision the two local LXD backends:
````bash mode=EXCERPT
multipass launch 24.04 --name lxd-67 --cpus 2 --memory 4G --disk 20G
multipass exec lxd-67 -- sudo snap install lxd --channel=6/stable
multipass exec lxd-67 -- sudo lxd init --auto --network-address='[::]' --trust-password test123
multipass launch 24.04 --name lxd-521 --cpus 2 --memory 4G --disk 20G
multipass exec lxd-521 -- sudo snap install lxd --channel=5.21/stable
multipass exec lxd-521 -- sudo lxd init --auto --network-address='[::]' --trust-password test123
````

- Add both remotes, create a test instance on each, and confirm the extension difference:
````bash mode=EXCERPT
lxc remote add lxd67 "$(multipass info lxd-67 | awk '/IPv4/{print $2; exit}')" --password test123 --accept-certificate
lxc remote add lxd521 "$(multipass info lxd-521 | awk '/IPv4/{print $2; exit}')" --password test123 --accept-certificate
lxc launch ubuntu:24.04 lxd67:qa-c1
lxc launch ubuntu:24.04 lxd521:qa-c1
lxc query lxd67:/1.0 | jq -r '.api_extensions[]' | grep instances_state_selective_recursion
lxc query lxd521:/1.0 | jq -r '.api_extensions[]' | grep instances_state_selective_recursion || true
````

- Generate the local `dotrun` cert once, then trust it on both backends:
````bash mode=EXCERPT
dotrun
# stop it after keys/lxd-ui.crt has been generated
multipass transfer keys/lxd-ui.crt lxd-67:/home/ubuntu/lxd-ui.crt
multipass transfer keys/lxd-ui.crt lxd-521:/home/ubuntu/lxd-ui.crt
multipass exec lxd-67 -- lxc auth identity create tls/dotrun-cert /home/ubuntu/lxd-ui.crt --group admins
multipass exec lxd-521 -- lxc auth identity create tls/dotrun-cert /home/ubuntu/lxd-ui.crt --group admins
````

- Point `dotrun` at each backend in separate runs and compare behavior:
````bash mode=EXCERPT
echo "LXD_UI_BACKEND_IP=$(multipass info lxd-67 | awk '/IPv4/{print $2; exit}')" > .env.local
dotrun
echo "LXD_UI_BACKEND_IP=$(multipass info lxd-521 | awk '/IPv4/{print $2; exit}')" > .env.local
dotrun
````

- On the `6.7` run:
  - Open `https://localhost:8407/ui/` and go to the Instances page
  - Confirm instances load without errors
  - In DevTools Network, verify the main instances request uses `recursion=2;fields=state.network`
  - In fine-grained mode, verify a second request is made with `with-access-entitlements`
  - Confirm instance actions and IP/network data still render correctly

- On the `5.21` run:
  - Open the same page after restarting `dotrun`
  - Confirm instances still load without errors
  - Verify the UI falls back to plain `recursion=2`

- To compare the optimization:
  - Load the same Instances page against both backends
  - Compare the `/1.0/instances` request URL, transferred size, and total time in DevTools
  - On `6.7`, the main payload should be smaller because only `state.network` is requested
  - In fine-grained mode, the `6.7` path should show the optimized main request plus a separate entitlement request

## Screenshots
Not applicable for this change; verification is primarily through instance loading behavior and DevTools network requests.

## Small note
I intentionally kept this QA section focused on **local reproduction with `dotrun`** and not on any shared/demo deployment. But I think we should do a test deployment here and have others verify the change

### DIFF
--- a/src/api/instances.spec.ts
+++ b/src/api/instances.spec.ts
@@ -1,0 +1,32 @@
+import { addRecursion, instanceFields } from "./instances";
+
+describe("addRecursion", () => {
+  it("uses selective recursion when supported and not fine-grained", () => {
+    const params = new URLSearchParams();
+    addRecursion(params, true, false);
+    expect(params.get("recursion")).toBe(
+      `2;fields=${instanceFields.join(",")}`,
+    );
+  });
+
+  it("falls back to plain recursion when feature is not supported", () => {
+    const params = new URLSearchParams();
+    addRecursion(params, false, false);
+    expect(params.get("recursion")).toBe("2");
+  });
+
+  it("falls back to plain recursion when fine-grained permissions are enabled", () => {
+    // Server doesn't support selective recursion with entitlements
+    const params = new URLSearchParams();
+    addRecursion(params, true, true);
+    expect(params.get("recursion")).toBe("2");
+  });
+
+  it("uses selective recursion when isFineGrained is null", () => {
+    const params = new URLSearchParams();
+    addRecursion(params, true, null);
+    expect(params.get("recursion")).toBe(
+      `2;fields=${instanceFields.join(",")}`,
+    );
+  });
+});
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -28,14 +28,33 @@ export const instanceEntitlements = [
   "can_update_state",
 ];
 
+// Only state.disk and state.network are valid selective recursion fields.
+// Other state fields (cpu, memory, pid, processes, status) are always included.
+export const instanceFields = ["state.disk", "state.network"];
+
+export const addRecursion = (
+  params: URLSearchParams,
+  hasSelectiveRecursion: boolean,
+  isFineGrained: boolean | null,
+): void => {
+  // Server doesn't support selective recursion with entitlements,
+  // so only use selective recursion when not requesting entitlements
+  if (hasSelectiveRecursion && isFineGrained !== true) {
+    params.set("recursion", `2;fields=${instanceFields.join(",")}`);
+  } else {
+    params.set("recursion", "2");
+  }
+};
+
 export const fetchInstance = async (
   name: string,
   project: string,
   isFineGrained: boolean | null,
+  hasSelectiveRecursion: boolean,
 ): Promise<LxdInstance> => {
   const params = new URLSearchParams();
   params.set("project", project);
-  params.set("recursion", "2");
+  addRecursion(params, hasSelectiveRecursion, isFineGrained);
   addEntitlements(params, isFineGrained, instanceEntitlements);
 
   return fetch(
@@ -50,9 +69,10 @@ export const fetchInstance = async (
 export const fetchInstances = async (
   project: string | null,
   isFineGrained: boolean | null,
+  hasSelectiveRecursion: boolean,
 ): Promise<LxdInstance[]> => {
   const params = new URLSearchParams();
-  params.set("recursion", "2");
+  addRecursion(params, hasSelectiveRecursion, isFineGrained);
   if (project) {
     params.set("project", project);
   } else {
--- a/src/context/useInstances.tsx
+++ b/src/context/useInstances.tsx
@@ -3,15 +3,23 @@ import { queryKeys } from "util/queryKeys";
 import type { UseQueryResult } from "@tanstack/react-query";
 import { fetchInstance, fetchInstances } from "api/instances";
 import { useAuth } from "./auth";
+import { useSupportedFeatures } from "./useSupportedFeatures";
 import type { LxdInstance } from "types/instance";
 
 export const useInstances = (
   project: string | null,
 ): UseQueryResult<LxdInstance[]> => {
   const { isFineGrained } = useAuth();
+  const { hasInstanceStateSelectiveRecursion } = useSupportedFeatures();
+
   return useQuery({
     queryKey: [queryKeys.instances, project],
-    queryFn: async () => fetchInstances(project, isFineGrained),
+    queryFn: async () =>
+      fetchInstances(
+        project,
+        isFineGrained,
+        hasInstanceStateSelectiveRecursion,
+      ),
     enabled: isFineGrained !== null,
   });
 };
@@ -22,9 +30,17 @@ export const useInstance = (
   enabled?: boolean,
 ): UseQueryResult<LxdInstance> => {
   const { isFineGrained } = useAuth();
+  const { hasInstanceStateSelectiveRecursion } = useSupportedFeatures();
+
   return useQuery({
     queryKey: [queryKeys.instances, name, project],
-    queryFn: async () => fetchInstance(name, project, isFineGrained),
-    enabled: enabled && isFineGrained !== null,
+    queryFn: async () =>
+      fetchInstance(
+        name,
+        project,
+        isFineGrained,
+        hasInstanceStateSelectiveRecursion,
+      ),
+    enabled: (enabled ?? true) && isFineGrained !== null,
   });
 };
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: src/api/instances.tsx [hunk 1]
```diff
@@ -28,43 +28,141 @@ export const instanceEntitlements = [
   "can_update_state",
 ];
 
+export const instanceSelectiveRecursionFields = ["state.network"];
```
edlerd: This should be a more exhaustive list. Maybe a bit simpler name for it?
```suggestion
export const instanceFields = ["state.network"];
```

src/api/instances.tsx
```diff
@@ -28,43 +28,141 @@ export const instanceEntitlements = [
...
+
+  return instances.map((instance) => ({
+    ...instance,
+    access_entitlements: entitlementsByInstance.get(
+      getInstanceLookupKey(instance),
+    )?.access_entitlements,
+  }));
+};
+
 export const fetchInstance = async (
```
edlerd, line 49: There is a lot of accidental complexity introduced to this function.

In my opinion, the feature justifies a single extra call, like `addFields(params,...` right after we add the entitlements. No need to refactor or do multiple API calls. Either we use the fields or not. The safeguarding might be in here or in the new `addFields` function.

The same goes for the `fetchInstances` function below.

edlerd (commented):
Thanks for this contribution. It looks promising.

I think the changes can be a lot simpler. The code in the `api` folder should be trivial with close to zero complexity. So we might not need a local unit test for them if we keep them simple.

src/api/instances.tsx
gajeshbhat: Done! Renamed to `instanceFields`. Regarding a more exhaustive list - should I add more fields now, or is `state.network` sufficient for this initial PR? Happy to expand it if you have specific fields in mind.

src/api/instances.tsx
gajeshbhat: Agreed, I have removed all the complexity. It's now a single `addRecursion(params, hasSelectiveRecursion, isFineGrained)` call right after entitlements.

One thing I discovered during testing: the LXD server doesn't accept `recursion=2;fields=...` when `with-access-entitlements` is present - it returns "**Entitlements can only be requested when recursion is enabled"**. So selective recursion is only applied when fine-grained permissions are not enabled. This keeps the code simple while maintaining compatibility. Let me know what you think, we can change it if necessary.

src/api/instances.tsx
```diff
@@ -28,14 +28,31 @@ export const instanceEntitlements = [
   "can_update_state",
 ];
 
+export const instanceFields = ["state.network"];
```
edlerd: Is this really the exhaustive list of recursive fields to be fetched? I suspect there are several fields missing. See also [instance.d.ts](https://github.com/canonical/lxd-ui/blob/main/src/types/instance.d.ts) for all referenced fields.

src/api/instances.spec.ts
```diff
@@ -0,0 +1,79 @@
+import { fetchInstance, fetchInstances, instanceFields } from "./instances";
+
+describe("fetchInstance", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+    vi.unstubAllGlobals();
+  });
+
+  it("uses selective recursion fields when supported", async () => {
+    const fetchMock = vi.fn().mockResolvedValue({
```
edlerd: We tend not to test with mocks as they are pretty expensive to maintain. I think we can keep a test here, but would focus only on the `addRecursion` method. It might be valuable to introduce a unit test to ensure the specific behavior mentioned in the comment inside that function is covered.

edlerd (commented):
Thanks for the iteration. This looks way better already!

Two smaller comments below, then this should be good to merge code-wise from my side 👍

src/api/instances.spec.ts
gajeshbhat: Done! Refactored the tests to directly test `addRecursion` without any mocks. Please let me know if you have any questions. The tests now cover:
1. Uses selective recursion when supported and not fine-grained
2. Falls back to plain recursion when feature is not supported
3. Falls back to plain recursion when fine-grained permissions are enabled (the server limitation)
4. Uses selective recursion when `isFineGrained` is null

src/api/instances.tsx
gajeshbhat: Updated to include all state fields from LxdInstanceState: cpu, disk, memory, network, pid, processes, status. This matches the full interface in `instance.d.ts.` Let me know what you think.

src/api/instances.tsx
```diff
@@ -28,14 +28,39 @@ export const instanceEntitlements = [
...
 
+export const instanceFields = [
+  "state.cpu",
+  "state.disk",
+  "state.memory",
+  "state.network",
+  "state.pid",
+  "state.processes",
+  "state.status",
+];
```
edlerd: I think there was a misunderstanding here. This produces an error:
```
Invalid state field: "state.cpu" (valid fields: state.disk, state.network)
```

Does the `fields` parameter only allow omitting those two fields, and any others are always included in the result?

edlerd (commented):
Thanks for the iteration. CI issues below might be unrelated and are being worked on by the team. But there is one issue identified with the changes, see below.

src/api/instances.tsx
gajeshbhat: Thanks for catching that! Fixed and updated to only include the valid selective recursion fields: `["state.disk", "state.network"]`. Added a comment explaining that other state fields (cpu, memory, pid, processes, status) are always included in the response automatically.

gajeshbhat:
> > Thanks for the iteration. This looks way better already!
> > Two smaller comments below, then this should be good to merge code-wise from my side 👍
> 
> Thank you for the review @edlerd . I have squashed, signed and pushed the commits and updated changes based on the PR feedback. Please feel free to let me know if you have any questions. Can you kindly approve the workflow to test the latest changes on the CI ?

Thank you for the review @edlerd . Updated the code based on PR comments.

edlerd (approved):
Code LGTM, thanks for adding this and for the iterations perfecting it 👍 

CI is currently being worked on to be fixed in https://github.com/canonical/lxd-ui/pull/1946 After that other PR has landed, please rebase this PR so we can see CI pass on the changes proposed in this PR.

gajeshbhat:
> Code LGTM, thanks for adding this and for the iterations perfecting it 👍
> 
> CI is currently being worked on to be fixed in #1946 After that other PR has landed, please rebase this PR so we can see CI pass on the changes proposed in this PR.

Sounds good. Will wait for that PR to be merged and then rebase. Thanks for the update @edlerd . Cheers.

kimanhou:
@gajeshbhat CI is now fixed

gajeshbhat:
> @gajeshbhat CI is now fixed

Thank you @kimanhou . I have rebase and pushed up the changes. Can you approve the CI run ? Should be good to go now.

edlerd (approved):
LGTM, thanks for your contribution @gajeshbhat !

PR #1865, opened 2026-03-11
feat: show available pool size WD-34210

omarelkashef:
## Done

- Show available memory in a pool when configuring size
## QA

1. Run the LXD-UI:
    - Go to instance create/edit -> Configuration -> Disk
    - Edit the root disk device
    - Ensure the pool available size is reported correctly
    - Go to Storage -> Volume -> Create/edit
    - Ensure the pool available size is reported correctly
## Screenshots

Clustered
image
image

Unclustered

image

### DIFF
--- a/src/components/forms/DiskDeviceFormRoot.tsx
+++ b/src/components/forms/DiskDeviceFormRoot.tsx
@@ -22,6 +22,8 @@ import { focusField } from "util/formFields";
 import DiskSizeQuotaLimitation from "components/forms/DiskSizeQuotaLimitation";
 import { getProfileFromSource } from "util/devices";
 import { isDeviceModified } from "util/formChangeCount";
+import StoragePoolSizeAvailable from "./StoragePoolSizeAvailable";
+import { getInstanceClusterMember } from "util/instances";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
@@ -48,14 +50,18 @@ const DiskDeviceFormRoot: FC<Props> = ({
     formik.values.entityType === "instance" &&
     formik.values.instanceType === "virtual-machine";
   const defaultSize = isVirtualMachine ? "10GiB" : "unlimited";
-  const poolDriver = pools.find(
+
+  const rootStoragePool = pools.find(
     (item) => item.name === formRootDevice?.pool,
-  )?.driver;
+  );
+  const poolDriver = rootStoragePool?.driver;
 
   const [inheritValue, inheritSource] = getInheritedRootStorage(
     formik.values,
     profiles,
   );
+
+  const clusterMemberName = getInstanceClusterMember(formik);
 
   const addRootStorage = () => {
     const copy = [...formik.values.devices];
@@ -235,8 +241,15 @@ const DiskDeviceFormRoot: FC<Props> = ({
                 />
                 <p className="p-form-help-text">
                   <DiskSizeQuotaLimitation driver={poolDriver} />
-                  Size of root storage. If empty, root storage will{" "}
-                  {isVirtualMachine ? "be 10GiB." : "not have a size limit."}
+                  If empty, root storage will{" "}
+                  {isVirtualMachine
+                    ? "be limited at 10GiB."
+                    : "not have a size limit."}
+                  <br />
+                  <StoragePoolSizeAvailable
+                    pool={rootStoragePool}
+                    clusterMember={clusterMemberName}
+                  />
                 </p>
               </>
             ),
--- a/src/components/forms/StoragePoolSizeAvailable.tsx
+++ b/src/components/forms/StoragePoolSizeAvailable.tsx
@@ -1,0 +1,52 @@
+import type { FC } from "react";
+import { humanFileSize } from "util/helpers";
+import { Icon } from "@canonical/react-components";
+import { useStoragePoolResourceLimit } from "context/useStoragePoolResourceLimit";
+import type { LxdStoragePool } from "types/storage";
+import { ResourceLimitIcon } from "components/ResourceLimitIcon";
+import { isClusterLocalDriver } from "util/storagePool";
+
+interface Props {
+  pool?: LxdStoragePool;
+  clusterMember?: string;
+}
+
+const StoragePoolSizeAvailable: FC<Props> = ({ pool, clusterMember }) => {
+  const resourceLimit = useStoragePoolResourceLimit(pool, clusterMember);
+  if (!resourceLimit) {
+    return null;
+  }
+  const { min: minSize, max: maxSize, sourceName } = resourceLimit;
+
+  const showHelpIcon = minSize !== maxSize;
+  const helpIconText =
+    "The available space depends on the target cluster member.";
+
+  return (
+    <>
+      Available space:{" "}
+      <b>
+        {humanFileSize(minSize)}
+        {minSize !== maxSize && ` - ${humanFileSize(maxSize)}`}
+        {showHelpIcon && (
+          <>
+            {" "}
+            <Icon
+              name="information"
+              className="help-link-icon"
+              title={helpIconText}
+            />
+          </>
+        )}
+      </b>
+      {sourceName && isClusterLocalDriver(pool?.driver ?? "") && (
+        <>
+          {" "}
+          <ResourceLimitIcon source={sourceName} sourceType="cluster-member" />
+        </>
+      )}
+      {"."}
+    </>
+  );
+};
+export default StoragePoolSizeAvailable;
--- a/src/context/useResourceLimit.tsx
+++ b/src/context/useResourceLimit.tsx
@@ -1,28 +1,18 @@
 import { useNotify } from "@canonical/react-components";
 import { useResources } from "context/useResources";
-import type { CreateInstanceFormValues } from "types/forms/instanceAndProfile";
-import type { EditInstanceFormValues } from "types/forms/instanceAndProfile";
-import { CLUSTER_GROUP_PREFIX } from "util/instances";
+import { getInstanceClusterMember } from "util/instances";
 import { getResourceLimit } from "util/resourceLimits";
-import { useIsClustered } from "./useIsClustered";
 import { useCurrentProject } from "./useCurrentProject";
 import { limitToBytes } from "util/limits";
-import type { LxdResources } from "types/resources";
-
-export const ensureArray = (resources: LxdResources | LxdResources[]) =>
-  Array.isArray(resources) ? resources : [resources];
+import { ensureArray } from "util/helpers";
+import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
 
 export const useResourceLimit = (
   mode: "cpu" | "memory",
-  formValues: CreateInstanceFormValues | EditInstanceFormValues,
+  formik: InstanceAndProfileFormikProps,
 ) => {
   const notify = useNotify();
-  const isClustered = useIsClustered();
-  const { target } = formValues as CreateInstanceFormValues;
-  const { location } = formValues as EditInstanceFormValues;
-  const isClusterGroupTarget = target?.startsWith(CLUSTER_GROUP_PREFIX);
-  const validTarget = target && !isClusterGroupTarget ? target : undefined;
-  const clusterMember = isClustered ? (validTarget ?? location) : undefined;
+  const clusterMember = getInstanceClusterMember(formik);
   const { data: resources, error, isLoading } = useResources(clusterMember);
   const { project } = useCurrentProject();
   const resourceLimitName = mode === "memory" ? "limits.memory" : "limits.cpu";
--- a/src/context/useStoragePoolResourceLimit.tsx
+++ b/src/context/useStoragePoolResourceLimit.tsx
@@ -1,0 +1,34 @@
+import type { LxdStoragePool } from "types/storage";
+import { useClusterMember } from "./useClusterMembers";
+import { useStoragePoolResources } from "./useStoragePoolResources";
+import { useNotify } from "@canonical/react-components";
+import { getResourceLimit } from "util/resourceLimits";
+import { ensureArray } from "util/helpers";
+
+export const useStoragePoolResourceLimit = (
+  pool?: LxdStoragePool,
+  clusterMemberName?: string,
+) => {
+  const notify = useNotify();
+  const { data: clusterMember } = useClusterMember(clusterMemberName ?? "");
+  const {
+    data: resources,
+    isLoading,
+    error,
+  } = useStoragePoolResources(pool, clusterMember);
+
+  if (error) {
+    notify.failure("Loading resources failed", error);
+    return null;
+  }
+
+  if (isLoading || !resources) {
+    return null;
+  }
+
+  const resourceArray = ensureArray(resources);
+  const availableSpaces = resourceArray.map(
+    (r) => r.space.total - (r.space.used ?? 0),
+  );
+  return getResourceLimit(availableSpaces, clusterMemberName);
+};
--- a/src/context/useStoragePoolResources.tsx
+++ b/src/context/useStoragePoolResources.tsx
@@ -1,0 +1,36 @@
+import { useQuery } from "@tanstack/react-query";
+import { useClusteredStoragePoolResources } from "./useStoragePools";
+import { useIsClustered } from "./useIsClustered";
+import type { LxdStoragePool } from "types/storage";
+import { isClusterLocalDriver } from "util/storagePool";
+import type { LxdClusterMember } from "types/cluster";
+import { queryKeys } from "util/queryKeys";
+import { fetchStoragePoolResources } from "api/storage-pools";
+
+export const useStoragePoolResources = (
+  pool?: LxdStoragePool,
+  member?: LxdClusterMember,
+) => {
+  const isClustered = useIsClustered();
+  const hasMemberSpecificSize =
+    isClusterLocalDriver(pool?.driver ?? "") && isClustered;
+  const clusteredQuery = useClusteredStoragePoolResources(
+    pool?.name,
+    member,
+    hasMemberSpecificSize,
+  );
+  const nonlocalDriverOrUnclusteredQuery = useQuery({
+    queryKey: [
+      queryKeys.storage,
+      pool?.name,
+      queryKeys.resources,
+      member?.server_name,
+    ],
+    queryFn: async () => fetchStoragePoolResources(pool?.name ?? ""),
+    enabled: !!pool && !hasMemberSpecificSize,
+  });
+
+  return hasMemberSpecificSize
+    ? clusteredQuery
+    : nonlocalDriverOrUnclusteredQuery;
+};
--- a/src/context/useStoragePools.tsx
+++ b/src/context/useStoragePools.tsx
@@ -54,17 +54,22 @@ export const usePoolFromClusterMembers = (
 };
 
 export const useClusteredStoragePoolResources = (
-  pool: string,
+  pool?: string,
   member?: LxdClusterMember,
+  enabled = true,
 ): UseQueryResult<LxdStoragePoolResources[]> => {
   const { data: clusterMembers = [] } = useClusterMembers();
+  const members = member ? [member] : clusterMembers;
   return useQuery({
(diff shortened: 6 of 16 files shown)


src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -235,7 +238,11 @@ const DiskDeviceFormRoot: FC<Props> = ({
                 />
                 <p className="p-form-help-text">
                   <DiskSizeQuotaLimitation driver={poolDriver} />
-                  Size of root storage. If empty, root storage will{" "}
+                  <StoragePoolSizeAvailable
+                    pool={rootStoragePool}
+                    formik={formik}
+                  />{" "}
```
edlerd: Let's always put this at the end in a new line. It is not connected to the "if empty ..." sentence, so it is best to keep them on dedicated lines. Same for the other use cases for this new component on storage volumes.

src/pages/storage/forms/StorageVolumeFormMain.tsx
```diff
@@ -113,8 +113,12 @@ const StorageVolumeFormMain: FC<Props> = ({
               (
                 <>
                   <DiskSizeQuotaLimitation driver={poolDriver} />
+                  <StoragePoolSizeAvailable
+                    pool={pool}
+                    clusterMember={formik.values.clusterMember}
+                  />{" "}
```
edlerd: Here as well, let's move this into a dedicated line after the other text.

src/context/useStoragePools.tsx
```diff
@@ -54,17 +54,18 @@ export const usePoolFromClusterMembers = (
 };
 
 export const useClusteredStoragePoolResources = (
-  pool: string,
+  pool?: string,
   member?: LxdClusterMember,
+  enabled = true,
```
edlerd, line 59:
```suggestion
```
We don't need to expose it here. It is enabled if a pool is provided and disabled otherwise -- if I read the code correctly.

src/context/useStoragePoolResourcesLimit.tsx
```diff
@@ -0,0 +1,56 @@
...
+} from "types/forms/instanceAndProfile";
+import { useClusterMembers } from "./useClusterMembers";
+import { useStoragePoolResources } from "./useStoragePoolResources";
+import { useNotify } from "@canonical/react-components";
+import { getResourceLimit } from "util/resourceLimits";
+import { ensureArray } from "util/helpers";
+
+export const useStoragePoolResourcesLimit = (
+  pool?: LxdStoragePool,
+  formik?: InstanceAndProfileFormikProps,
```
edlerd:
```suggestion
```
I don't think we should leak the formik object into this component. Because we also use it for volumes. The interface then should not include instance or profile specifics.

src/context/useStoragePoolResourcesLimit.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { data: clusterMembers } = useClusterMembers();
+  const clusterMember = clusterMembers?.find(
+    (member) => member.server_name === clusterMemberName,
+  );
+
+  const {
+    data: resources,
+    isLoading,
+    error,
+  } = useStoragePoolResources(pool, clusterMember, isQueryEnabled);
```
edlerd: I think there is too much logic in here. As suggested in the other comment, I think we should move the formik logic out and then this becomes much more straight forward:
```
export const useStoragePoolResourcesLimit = (
  pool?: LxdStoragePool,
  targetMember?: string,
) => {
  const { data: clusterMember } = useClusterMember(targetMember ?? "");
  const { data: resources, isLoading } = useStoragePoolResources(
    pool,
    clusterMember,
  );

  if (isLoading || !resources) {
    return null;
  }

  const resourceArray = ensureArray(resources);
  const availableSpaces = resourceArray.map(
    (r) => r.space.total - (r.space.used ?? 0),
  );
  return getResourceLimit(availableSpaces, targetMember);
};
```

src/context/useStoragePoolResources.tsx
```diff
@@ -0,0 +1,33 @@
...
+import type { LxdStoragePool } from "types/storage";
+import { isClusterLocalDriver } from "util/storagePool";
+import type { LxdClusterMember } from "types/cluster";
+import { queryKeys } from "util/queryKeys";
+import { fetchStoragePoolResources } from "api/storage-pools";
+
+export const useStoragePoolResources = (
+  pool?: LxdStoragePool,
+  member?: LxdClusterMember,
+  enabled = true,
```
edlerd:
```suggestion
```
We don't need to expose it, but can make the decision internally based on the existence of `pool`.

src/context/useStoragePoolResources.tsx
```diff
@@ -0,0 +1,33 @@
...
+) => {
+  const isClustered = useIsClustered();
+  const clusteredQuery = useClusteredStoragePoolResources(
+    pool?.name,
+    member,
+    enabled,
```
edlerd: This should also consider `hasMemberSpecificSize` for enabling/disabling.

src/context/useStoragePoolResourcesLimit.tsx
omarelkashef: I was going to go this route because I did not like exposing `InstanceAndProfileFormikProps` in the interface, but that means we will have to do all the unpacking logic in `DiskDeviceFormRoot` and anywhere this component will be used.

edlerd: If I recall correctly, the `DiskDeviceFormRoot` is the only usage with this formik type. We could introduce another adapter component to isolate the unpacking, but maybe it is okay to have that complexity in the `DiskDeviceFormRoot`. Wdyt?

src/context/useStoragePoolResourceLimit.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  if (isLoading || !resources) {
+    return null;
+  }
+
+  const resourceArray = ensureArray(resources);
+  const availableSpaces = resourceArray.map(
+    (r) => r.space.total - (r.space.used ?? 0),
+  );
+  return getResourceLimit(availableSpaces, clusterMemberName);
```
edlerd, line 33: It is so nice, that we can reuse this logic here again. Good stuff :)

src/context/useStoragePoolResourcesLimit.tsx
omarelkashef: I moved it to `DiskDeviceFormRoot` and it does not look cluttered, so it would be good to keep  it there.

src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -235,7 +250,12 @@ const DiskDeviceFormRoot: FC<Props> = ({
                 />
                 <p className="p-form-help-text">
                   <DiskSizeQuotaLimitation driver={poolDriver} />
-                  Size of root storage. If empty, root storage will{" "}
+                  <StoragePoolSizeAvailable
+                    pool={rootStoragePool}
+                    clusterMember={clusterMemberName}
+                  />
+                  <br />
+                  If empty, root storage will{" "}
```
edlerd: There is an edge case here with VMs and low space left. This case is still valid, as the size is actually an upper limit. I think we should slightly change the message to `If empty, root storage will be limited at 10GiB.`. Wdyt?

Image

src/components/forms/StoragePoolSizeAvailable.tsx
```diff
@@ -0,0 +1,54 @@
...
+            />
+          </>
+        )}
+      </b>
+      {sourceName && (
+        <>
+          {" "}
+          <ResourceLimitIcon
+            source={sourceName}
+            sourceType={"cluster-member"}
```
edlerd:
```suggestion
            sourceType="cluster-member"
```

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -24,22 +21,12 @@ const StoragePoolSize: FC<Props> = ({
...
-    enabled: !hasMemberSpecificSize,
-  });
-
-  const resourceList = hasMemberSpecificSize
-    ? clusteredPoolResources
-    : [poolResources];
-
+  const { data: resources } = useStoragePoolResources(pool, member);
+  const resourceList = ensureArray(resources);
   if (hasMemberSpecificSize && forceSingleLine && !member) {
```
```suggestion
  if (hasMemberSpecificSize && forceSingleLine && !member && resourceList.length > 1) {
```
We were previously missing an edge case here.

src/context/useStoragePoolResourcesLimit.tsx
```diff
@@ -0,0 +1,35 @@
+import type { LxdStoragePool } from "types/storage";
+import { useClusterMember } from "./useClusterMembers";
+import { useStoragePoolResources } from "./useStoragePoolResources";
+import { useNotify } from "@canonical/react-components";
+import { getResourceLimit } from "util/resourceLimits";
+import { ensureArray } from "util/helpers";
+
+export const useStoragePoolResourcesLimit = (
```
edlerd:
```suggestion
export const useStoragePoolResourceLimit = (
```

src/context/useStoragePoolResources.tsx
```diff
@@ -0,0 +1,31 @@
...
+  pool?: LxdStoragePool,
+  member?: LxdClusterMember,
+) => {
+  const isClustered = useIsClustered();
+  const hasMemberSpecificSize =
+    isClusterLocalDriver(pool?.driver ?? "") && isClustered;
+  const clusteredQuery = useClusteredStoragePoolResources(
+    pool?.name,
+    member,
+    hasMemberSpecificSize,
```
edlerd, line 20:
```suggestion
    !!pool && hasMemberSpecificSize,
```
I think we need the pool here as well to enable it. Maybe introduce a variable to make the meaning of the 3rd parameter more clear.

src/context/useStoragePoolResources.tsx
```diff
@@ -0,0 +1,31 @@
...
+  );
+  const nonlocalDriverOrUnclusteredQuery = useQuery({
```
edlerd, line 22: Naming is a bit bulky. How about `unifiedQuery` for this one?

src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -235,7 +250,12 @@ const DiskDeviceFormRoot: FC<Props> = ({
                 />
                 <p className="p-form-help-text">
                   <DiskSizeQuotaLimitation driver={poolDriver} />
-                  Size of root storage. If empty, root storage will{" "}
+                  <StoragePoolSizeAvailable
+                    pool={rootStoragePool}
+                    clusterMember={clusterMemberName}
+                  />
```
edlerd, line 252: Let's swap this to be last after the other message. Same in the volume form. I think it is slightly nicer, as 1) this sentence might come in delayed on loading 2) it is consistent with the other limits and 3) the sentence is shorter, so it looks visually a bit cleaner.

src/components/forms/StoragePoolSizeAvailable.tsx
```diff
@@ -0,0 +1,54 @@
...
+  pool?: LxdStoragePool;
+  clusterMember?: string;
+}
+
+const StoragePoolSizeAvailable: FC<Props> = ({ pool, clusterMember }) => {
+  const resourceLimit = useStoragePoolResourcesLimit(pool, clusterMember);
+  if (!resourceLimit) {
+    return null;
+  }
+  const { min: minSize, max: maxSize, sourceName } = resourceLimit;
```
edlerd: There is a race condition with the loading here. Below happens on editing an instance, when I switch tabs and come back it is correctly displayed.

This might be related to the next comment.

Image

src/context/useStoragePoolResources.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { fetchStoragePoolResources } from "api/storage-pools";
+
+export const useStoragePoolResources = (
+  pool?: LxdStoragePool,
+  member?: LxdClusterMember,
+) => {
+  const isClustered = useIsClustered();
+  const hasMemberSpecificSize =
+    isClusterLocalDriver(pool?.driver ?? "") && isClustered;
+  const clusteredQuery = useClusteredStoragePoolResources(
```
edlerd, line 17: The `useClusteredStoragePoolResources` is slightly broken. The `member` is an input, but not considered in the cache key. I think the below problem of only updating and showing correct values after switching tabs back and forth (to trigger a refetch of the cached values) is due to that problem.

https://github.com/user-attachments/assets/b1f7206b-a729-4c7c-9fe2-5945d456579c

src/components/forms/DiskDeviceFormRoot.tsx
omarelkashef: I agree,  this is a more accurate statement!

src/context/useStoragePoolResources.tsx
omarelkashef: I understood your previous feedback to remove the pool from here because it is already checked internally in `useClusteredStoragePoolResources`

src/context/useStoragePoolResources.tsx
omarelkashef: Does `unifiedQuery` accurately explain it? `nonlocalDriverOrUnclusteredQuery` is a bit bulky, indeed, but it is only used internally and it explains a somewhat hard-to-explain query.

src/components/forms/DiskDeviceFormRoot.tsx
omarelkashef: Good idea, I agree.

src/context/useStoragePoolResources.tsx
edlerd: I thought back then we could completely remove the `enabled` parameter from `useClusteredStoragePoolResources`. As it is checked internally, everything seems fine to keep as is.

src/context/useStoragePoolResources.tsx
edlerd: Yeah, true. Probably best to keep this as is, agreed.

src/context/useStoragePoolResources.tsx
omarelkashef: Yes, good catch!

src/context/useStoragePools.tsx
```diff
@@ -54,17 +54,24 @@ export const usePoolFromClusterMembers = (
...
 ): UseQueryResult<LxdStoragePoolResources[]> => {
   const { data: clusterMembers = [] } = useClusterMembers();
   return useQuery({
-    queryKey: [queryKeys.storage, pool, queryKeys.cluster, queryKeys.resources],
+    queryKey: [
+      queryKeys.storage,
+      pool,
+      queryKeys.cluster,
+      queryKeys.resources,
+      member?.server_name,
```
edlerd: This should match the input to the function. It doesn't have to be a primitive and can be an array. Can we assign `member ? [member] : clusterMembers` to a variable and use it here and below?

src/components/forms/StoragePoolSizeAvailable.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useStoragePoolResourceLimit } from "context/useStoragePoolResourceLimit";
+import type { LxdStoragePool } from "types/storage";
+import { ResourceLimitIcon } from "components/ResourceLimitIcon";
+
+interface Props {
+  pool?: LxdStoragePool;
+  clusterMember?: string;
+}
+
+const StoragePoolSizeAvailable: FC<Props> = ({ pool, clusterMember }) => {
```
edlerd, line 14: Let's apply this also to the instance quick volume creation:

Image

src/components/forms/DiskDeviceFormRoot.tsx
```diff
@@ -235,8 +250,15 @@ const DiskDeviceFormRoot: FC<Props> = ({
...
-                  {isVirtualMachine ? "be 10GiB." : "not have a size limit."}
+                  If empty, root storage will{" "}
+                  {isVirtualMachine
+                    ? "be limited at 10GiB."
+                    : "not have a size limit."}
+                  <br />
+                  <StoragePoolSizeAvailable
+                    pool={rootStoragePool}
+                    clusterMember={clusterMemberName}
+                  />
```
edlerd, line 252: There is an edge case for when using a ceph pool and editing an instance, we should not say "on cluster member", because ceph is cluster wide.

Image

omarelkashef: I assume that applies to all non-local drivers, right?

src/context/useResourceLimit.tsx
```diff
@@ -1,28 +1,27 @@
...
-  const validTarget = target && !isClusterGroupTarget ? target : undefined;
-  const clusterMember = isClustered ? (validTarget ?? location) : undefined;
+  const instanceLocation = getInstanceLocation(formik);
+
+  const isClusterGroupTarget =
+    instanceLocation?.startsWith(CLUSTER_GROUP_PREFIX);
+  const clusterMember =
+    instanceLocation && instanceLocation !== "any" && !isClusterGroupTarget
+      ? instanceLocation
+      : undefined;
```
edlerd: This is repeated from `DiskDeviceFormRoot`, can we introduce a helper like `getInstanceClusterMember` that internally uses `getInstanceLocation` and reduces the duplications?

src/pages/storage/forms/StorageVolumeFormMain.tsx
```diff
@@ -114,8 +113,13 @@ const StorageVolumeFormMain: FC<Props> = ({
                 <>
                   <DiskSizeQuotaLimitation driver={poolDriver} />
                   {formik.values.volumeType === "custom"
-                    ? "Size of storage volume. If empty, volume will not have a size limit within its storage pool."
+                    ? "If empty, volume will not have a size limit within its storage pool."
                     : "Size is immutable for non-custom volumes."}
+                  <br />
+                  <StoragePoolSizeAvailable
+                    pool={pool}
+                    clusterMember={clusterMembers[0]?.server_name}
```
edlerd: I think this picks the wrong member if we create a storage volume manually. We might select another member as the first one there.

edlerd: From this screen: 
image

src/util/instances.tsx
```diff
@@ -118,3 +120,16 @@ export const getInstanceType = (instance: LxdInstance) => {
...
+  formik: InstanceAndProfileFormikProps,
+) => {
+  const instanceLocation = getInstanceLocation(formik);
+  const isClusterGroupTarget =
+    instanceLocation?.startsWith(CLUSTER_GROUP_PREFIX);
+  const clusterMemberName =
+    instanceLocation && instanceLocation !== "any" && !isClusterGroupTarget
+      ? instanceLocation
+      : undefined;
+  return clusterMemberName;
```
```suggestion
  const location = getInstanceLocation(formik);
  if (!location || location === "any") {
    return undefined;
  }

  const isGroup = location?.startsWith(CLUSTER_GROUP_PREFIX);
  if (isGroup) {
    return undefined;
  }

  return location;
```
Slightly simpler?

src/pages/storage/CustomVolumeCreateModal.tsx
```diff
@@ -129,7 +132,17 @@ const CustomVolumeCreateModal: FC<Props> = ({
...
+        <StorageVolumeFormMain
+          formik={formik}
+          poolError={poolError}
+          pools={pools}
+          settings={settings}
+          clusterMembers={
+            instanceLocation
+              ? ([{ server_name: instanceLocation }] as LxdClusterMember[])
+              : []
+          }
```
edlerd, line 144:
```suggestion
          clusterMembers={clusterMembers}
```
Let's call `  const { data: clusterMembers = [] } = useClusterMembers();` above and use it here directly. The current version is a bit hacky and might bite us later.

edlerd (commented):
QA LGTM, two nitpicks on code structure, then this should be good to merge 👍

src/pages/storage/CustomVolumeCreateModal.tsx
omarelkashef: Why make an extra api call? Do we need all the cluster members?

edlerd: True, that is not necessary here. Let's keep it as is then.

PR #1873, opened 2026-03-16
chore: update circular deps check depth to 5 WD-34344

omarelkashef:
## Done

- update circular deps check depth to 5 
## QA

1. Run the LXD-UI:
    - Check CI or run `npx depcruise src --max-depth 5` locally and make sure no errors are sufraced.


src/components/forms/BootForm.tsx
```diff
@@ -2,28 +2,11 @@ import type { FC } from "react";
...
-    [getInstanceField("boot_autostart_priority")]:
-      values.boot_autostart_priority?.toString(),
-    [getInstanceField("boot_host_shutdown_timeout")]:
-      values.boot_host_shutdown_timeout?.toString(),
-    [getInstanceField("boot_mode")]: values.boot_mode?.toString(),
-    [getInstanceField("boot_stop_priority")]:
-      values.boot_stop_priority?.toString(),
-  };
-};
-
```
edlerd, line 26: Can we revert this back into this file? We tend to keep building the payload next to the visual form elements. So it is easier to maintain.

src/pages/instances/InstanceTerminal.tsx
```diff
@@ -61,17 +62,6 @@ const defaultPayload: TerminalConnectPayload = {
...
-
-const getDefaultPayload = (instance: LxdInstance) => {
-  const userPayload = instance.config[UI_TERMINAL_DEFAULT_PAYLOAD];
-  if (userPayload) {
-    return JSON.parse(userPayload) as TerminalConnectPayload;
-  }
-
-  return defaultPayload;
-};
-
```
edlerd, line 74: This has only one use case in this file, better keep it here. Keeping the constant in the utils seems right though.

src/util/instanceImage.tsx
```diff
@@ -0,0 +1,29 @@
+import type { LxdInstance } from "types/instance";
+import { useImagesInProject } from "context/useImages";
+import ResourceLink from "components/ResourceLink";
+import ResourceLabel from "components/ResourceLabel";
+import { ROOT_PATH } from "util/rootPath";
+
+export const getImageLink = (instance: LxdInstance) => {
```
edlerd, line 7: I think this should be a react component such as `pages/instance/ImageLink.tsx`

src/components/forms/BootForm.tsx
omarelkashef: It creates the following circular dependency  

      src/components/ConfigurationRow.tsx → 
      src/util/instanceEdit.tsx →
      src/components/forms/BootForm.tsx →
      src/components/ConfigurationRow.tsx

it is not ideal that a util function imports a function from a component file (going frominstanceEdit.tsx to BootForm.tsx)

src/pages/instances/InstanceTerminal.tsx
omarelkashef: `UI_TERMINAL_DEFAULT_PAYLOAD` causes a circular dependency because it is imported in `src/pages/instances/TerminalPayloadForm.tsx`, and  `UI_TERMINAL_DEFAULT_PAYLOAD` is closely related to `getDefaultPayload`, so moving them together makes sense.

PR #1878, opened 2026-03-17
feat(Ceph): drop source

### LINKED ISSUE #1877: Drop `source` config key for Ceph RBD and CephFS drivers
LXD dropped support for the member specific config key `source` for both the Ceph RBD and CephFS drivers in https://github.com/canonical/lxd/pull/17853.

Instead of providing the pool name for each member of a cluster using `source`, a user either sets `ceph.osd.pool_name` for Ceph RBD or `cephfs.path` for CephFS (non member specific keys).
This should also be reflected in the UI which currently offers to set `source` for both Ceph RBD and CephFS.
The absence of the `storage_remote_drop_source` API extension can be used to figure whether or not LXD still supports setting `source` for both Ceph RBD and CephFS.

Also the pool creation window for Ceph RBD currently doesn't seem to allow setting `ceph.osd.pool_name` when not using the direct YAML input.

All of the other remote drivers never used `source`. The config key should only ever be used by the local drivers going forward.

kimanhou:
## Done

- Drop source config key for Ceph RBD and CephFS drivers

Fixes #1877 

## QA

1. Run the LXD-UI:
    - Create a storage pool, enter a name, select `driver=Ceph`
    - Make sure `Source` doesn't appear on the main configuration screen
    - Make sure `Ceph OSD pool name` appears on the Ceph specific screen
    - Save
    - Edit
    - Create a storage pool, enter a name, select `driver=CephFS`
    - Make sure `Source` doesn't appear on the main configuration screen
    - Make sure `Path` appears on the Ceph specific screen (it was already there)
    - Save
    - Edit

## Screenshots

image

image

### DIFF
--- a/src/pages/storage/CreateStoragePool.tsx
+++ b/src/pages/storage/CreateStoragePool.tsx
@@ -32,6 +32,7 @@ import type { LxdStoragePool } from "types/storage";
 import YamlSwitch from "components/forms/YamlSwitch";
 import StoragePoolRichChip from "./StoragePoolRichChip";
 import { ROOT_PATH } from "util/rootPath";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
 
 const CreateStoragePool: FC = () => {
   const navigate = useNavigate();
@@ -42,6 +43,7 @@ const CreateStoragePool: FC = () => {
   const [section, setSection] = useState(slugify(MAIN_CONFIGURATION));
   const controllerState = useState<AbortController | null>(null);
   const { data: clusterMembers = [] } = useClusterMembers();
+  const { hasRemoteDropSource } = useSupportedFeatures();
 
   if (!project) {
     return <>Missing project</>;
@@ -68,7 +70,7 @@ const CreateStoragePool: FC = () => {
     onSubmit: (values) => {
       const storagePool = values.yaml
         ? (yamlToObject(values.yaml) as LxdStoragePool)
-        : toStoragePool(values);
+        : toStoragePool(values, hasRemoteDropSource);
 
       const mutation =
         clusterMembers.length > 0
--- a/src/pages/storage/forms/StoragePoolForm.tsx
+++ b/src/pages/storage/forms/StoragePoolForm.tsx
@@ -27,15 +27,13 @@ import type { LxdStoragePool } from "types/storage";
 import {
   alletraDriver,
   btrfsDriver,
-  cephDriver,
-  cephFSDriver,
   cephObject,
   getSupportedStorageDrivers,
   powerFlex,
   pureStorage,
   zfsDriver,
 } from "util/storageOptions";
-import { getPoolKey } from "util/storagePool";
+import { getPoolKey, isCephDriver, isCephFSDriver } from "util/storagePool";
 import { slugify } from "util/slugify";
 import YamlForm from "components/forms/YamlForm";
 import { handleConfigKeys } from "util/storagePoolForm";
@@ -61,9 +59,8 @@ interface Props {
 
 export const toStoragePool = (
   values: StoragePoolFormValues,
+  hasRemoteDropSource?: boolean,
 ): LxdStoragePool => {
-  const isCephDriver = values.driver === cephDriver;
-  const isCephFSDriver = values.driver === cephFSDriver;
   const isPowerFlexDriver = values.driver === powerFlex;
   const isPureDriver = values.driver === pureStorage;
   const isZFSDriver = values.driver === zfsDriver;
@@ -72,7 +69,7 @@ export const toStoragePool = (
   const hasValidSize = values.size?.match(/^\d/);
 
   const getConfig = () => {
-    if (isCephDriver) {
+    if (isCephDriver(values)) {
       return {
         [getPoolKey("ceph_cluster_name")]: values.ceph_cluster_name,
         [getPoolKey("ceph_osd_pg_num")]: values.ceph_osd_pg_num?.toString(),
@@ -80,18 +77,22 @@ export const toStoragePool = (
         [getPoolKey("ceph_rbd_du")]: values.ceph_rbd_du,
         [getPoolKey("ceph_user_name")]: values.ceph_user_name,
         [getPoolKey("ceph_rbd_features")]: values.ceph_rbd_features,
-        source: values.source,
+        source: hasRemoteDropSource ? undefined : values.source,
+        [getPoolKey("ceph_osd_pool_name")]: hasRemoteDropSource
+          ? values.ceph_osd_pool_name
+          : undefined,
       };
     }
-    if (isCephFSDriver) {
+    if (isCephFSDriver(values)) {
       return {
         [getPoolKey("cephfs_cluster_name")]: values.cephfs_cluster_name,
         [getPoolKey("cephfs_create_missing")]: values.cephfs_create_missing,
         [getPoolKey("cephfs_fscache")]: values.cephfs_fscache,
         [getPoolKey("cephfs_osd_pg_num")]: values.cephfs_osd_pg_num?.toString(),
-        [getPoolKey("cephfs_path")]: values.cephfs_path,
-        [getPoolKey("cephfs_user_name")]: values.cephfs_user_name,
-        source: values.source,
+        source: hasRemoteDropSource ? undefined : values.source,
+        [getPoolKey("cephfs_path")]: hasRemoteDropSource
+          ? values.cephfs_path
+          : undefined,
       };
     }
     if (isCephObjectDriver) {
--- a/src/pages/storage/forms/StoragePoolFormCeph.tsx
+++ b/src/pages/storage/forms/StoragePoolFormCeph.tsx
@@ -5,15 +5,29 @@ import { getConfigurationRow } from "components/ConfigurationRow";
 import { Input, Select } from "@canonical/react-components";
 import { optionTrueFalse } from "util/instanceOptions";
 import ScrollableConfigurationTable from "components/forms/ScrollableConfigurationTable";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
 
 interface Props {
   formik: FormikProps<StoragePoolFormValues>;
 }
 
 const StoragePoolFormCeph: FC<Props> = ({ formik }) => {
+  const { hasRemoteDropSource } = useSupportedFeatures();
+
   return (
     <ScrollableConfigurationTable
       rows={[
+        ...(hasRemoteDropSource
+          ? [
+              getConfigurationRow({
+                formik,
+                label: "Ceph OSD pool name",
+                name: "ceph_osd_pool_name",
+                defaultValue: "",
+                children: <Input type="text" placeholder="Enter pool name" />,
+              }),
+            ]
+          : []),
         getConfigurationRow({
           formik,
           label: "Cluster name",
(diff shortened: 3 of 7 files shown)


src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -72,26 +69,26 @@ export const toStoragePool = (
...
+    if (isCephDriver(values)) {
       return {
         [getPoolKey("ceph_cluster_name")]: values.ceph_cluster_name,
         [getPoolKey("ceph_osd_pg_num")]: values.ceph_osd_pg_num?.toString(),
         [getPoolKey("ceph_rbd_clone_copy")]: values.ceph_rbd_clone_copy,
         [getPoolKey("ceph_rbd_du")]: values.ceph_rbd_du,
         [getPoolKey("ceph_user_name")]: values.ceph_user_name,
         [getPoolKey("ceph_rbd_features")]: values.ceph_rbd_features,
-        source: values.source,
+        [hasRemoteDropSource ? "ceph.osd.pool_name" : "source"]: values.source,
```
roosterfish: Not sure if this is the right place here, but `ceph.osd.pool_name` should always be a valid option for a Ceph RBD pool no matter if `storage_remote_drop_source` is present or not.

Same as the `[getPoolKey("cephfs_path")]: values.cephfs_path,` below.

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -72,26 +68,25 @@ export const toStoragePool = (
...
       return {
         [getPoolKey("ceph_cluster_name")]: values.ceph_cluster_name,
         [getPoolKey("ceph_osd_pg_num")]: values.ceph_osd_pg_num?.toString(),
         [getPoolKey("ceph_rbd_clone_copy")]: values.ceph_rbd_clone_copy,
         [getPoolKey("ceph_rbd_du")]: values.ceph_rbd_du,
         [getPoolKey("ceph_user_name")]: values.ceph_user_name,
         [getPoolKey("ceph_rbd_features")]: values.ceph_rbd_features,
-        source: values.source,
+        [getPoolKey("ceph_osd_pool_name")]:
+          values.ceph_osd_pool_name || values.source,
```
edlerd: Why the fallback to `values.source` here?

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -72,26 +68,25 @@ export const toStoragePool = (
...
+    if (isCephFSDriver(values)) {
       return {
         [getPoolKey("cephfs_cluster_name")]: values.cephfs_cluster_name,
         [getPoolKey("cephfs_create_missing")]: values.cephfs_create_missing,
         [getPoolKey("cephfs_fscache")]: values.cephfs_fscache,
         [getPoolKey("cephfs_osd_pg_num")]: values.cephfs_osd_pg_num?.toString(),
-        [getPoolKey("cephfs_path")]: values.cephfs_path,
-        [getPoolKey("cephfs_user_name")]: values.cephfs_user_name,
-        source: values.source,
+        [getPoolKey("cephfs_path")]: values.cephfs_path || values.source,
```
edlerd: Same as above: Why the fallback?

src/pages/storage/CreateStoragePool.tsx
```diff
@@ -76,7 +80,10 @@ const CreateStoragePool: FC = () => {
               createClusteredPool(
                 storagePool,
                 clusterMembers,
-                values.sourcePerClusterMember,
+                hasRemoteDropSource &&
+                  (isCephDriver(values) || isCephFSDriver(values))
+                  ? {}
+                  : values.sourcePerClusterMember,
```
edlerd: This shouldn't be necessary as the value is empty anyway with the new extension and the source field hidden.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -65,7 +69,9 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
     !isPureDriver &&
     !isPowerFlexDriver &&
     !isCephObjectDriver &&
-    !isAlletraDriver;
+    !isAlletraDriver &&
+    !isCephDriver(formik.values) &&
+    !isCephFSDriver(formik.values);
```
edlerd: This should consider the new api extension in combination with the two new checks.

src/pages/storage/EditStoragePool.tsx
```diff
@@ -101,7 +104,10 @@ const EditStoragePool: FC<Props> = ({ pool }) => {
               updateClusteredPool(
                 savedPool,
                 clusterMembers,
-                values.sourcePerClusterMember,
+                hasRemoteDropSource &&
+                  (isCephDriver(values) || isCephFSDriver(values))
+                  ? {}
+                  : values.sourcePerClusterMember,
```
edlerd: Here as well, shoulnd't be necessary

src/util/configInheritance.tsx
```diff
@@ -188,7 +188,14 @@ const getStoragePoolRowMetadata = (
...
 
-  const lxdDefault = getLxdDefault(configField);
+  let lxdDefault = getLxdDefault(configField);
+
+  if (
+    configKey === "ceph.osd.pool_name" &&
+    (!lxdDefault || lxdDefault === "-")
+  ) {
+    lxdDefault = values.name;
+  }
```
edlerd: We shouldn't set the LXD default value to a user input here. That seems like a misuse of the structure.

Consider restoring the original code here. I suspect this to be okay to be empty on creation. On edit, I guess LXD would have the value set anyway to be the value of the pool name.

src/pages/storage/forms/StoragePoolFormCeph.tsx
```diff
@@ -61,6 +61,13 @@ const StoragePoolFormCeph: FC<Props> = ({ formik }) => {
           defaultValue: "",
           children: <Input type="text" placeholder="Enter RBD features" />,
         }),
+        getConfigurationRow({
+          formik,
+          label: "Ceph OSD pool name",
+          name: "ceph_osd_pool_name",
+          defaultValue: "",
+          children: <Input type="text" placeholder="Enter pool name" />,
+        }),
```
edlerd: This should be the first property in the array.

kimanhou:
@roosterfish I opened a PR in LXD to update the description of Ceph OSD pool name, it will appear in help text in the UI. Is it clearer ?

https://github.com/canonical/lxd/pull/17922

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -60,12 +64,15 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
+  const isCephVariant =
+    isCephDriver(formik.values) || isCephFSDriver(formik.values);
 
   const hasSource =
     !isPureDriver &&
     !isPowerFlexDriver &&
     !isCephObjectDriver &&
-    !isAlletraDriver;
+    !isAlletraDriver &&
+    (isCephVariant ? !hasRemoteDropSource : !isCephVariant);
```
edlerd: What I don't understand is, if it is false, then true?

Did you mean:
```suggestion
    (!isCephVariant || !hasRemoteDropSource);
```

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -206,7 +213,7 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
               />
             ))}
           {hasSource &&
-            (isClusteredServer(settings) ? (
+            (isClusteredServer(settings) && !hasRemoteDropSource ? (
```
edlerd:
```suggestion
            (isClusteredServer(settings) ? (
```
I don't think the condition here is correct. Why always hide the clustered source with the new api extension?

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -72,26 +68,24 @@ export const toStoragePool = (
...
+    if (isCephDriver(values)) {
       return {
         [getPoolKey("ceph_cluster_name")]: values.ceph_cluster_name,
         [getPoolKey("ceph_osd_pg_num")]: values.ceph_osd_pg_num?.toString(),
         [getPoolKey("ceph_rbd_clone_copy")]: values.ceph_rbd_clone_copy,
         [getPoolKey("ceph_rbd_du")]: values.ceph_rbd_du,
         [getPoolKey("ceph_user_name")]: values.ceph_user_name,
         [getPoolKey("ceph_rbd_features")]: values.ceph_rbd_features,
-        source: values.source,
+        [getPoolKey("ceph_osd_pool_name")]: values.ceph_osd_pool_name,
```
edlerd: I think we need to consider the new api extension here as well. If it is not present, we keep the source. If present we do not fill source, but the new key.

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -72,26 +68,24 @@ export const toStoragePool = (
...
-    if (isCephFSDriver) {
+    if (isCephFSDriver(values)) {
       return {
         [getPoolKey("cephfs_cluster_name")]: values.cephfs_cluster_name,
         [getPoolKey("cephfs_create_missing")]: values.cephfs_create_missing,
         [getPoolKey("cephfs_fscache")]: values.cephfs_fscache,
         [getPoolKey("cephfs_osd_pg_num")]: values.cephfs_osd_pg_num?.toString(),
         [getPoolKey("cephfs_path")]: values.cephfs_path,
-        [getPoolKey("cephfs_user_name")]: values.cephfs_user_name,
-        source: values.source,
```
edlerd, line 94: I think we need to consider the api extension here as well.

src/pages/storage/forms/StoragePoolFormMain.tsx
kimanhou: Or `!(isCephVariant && hasRemoteDropSource)` ?

edlerd: > Or `!(isCephVariant && hasRemoteDropSource)` ?

Yes, we should choose the option that is easier to understand. It seems simpler without the ternary, in my opinion. From the other two options, I'm not sure.

kimanhou: We can make it similar to the other conditions in `hasSource`.

```
  const isCephVariant =
    isCephDriver(formik.values) || isCephFSDriver(formik.values);
  const isCephVariantWithoutSource = isCephVariant && hasRemoteDropSource;

  const hasSource =
    !isPureDriver &&
    !isPowerFlexDriver &&
    !isCephObjectDriver &&
    !isAlletraDriver &&
    !isCephVariantWithoutSource;
```

src/pages/storage/forms/StoragePoolFormCeph.tsx
```diff
@@ -14,6 +14,13 @@ const StoragePoolFormCeph: FC<Props> = ({ formik }) => {
   return (
     <ScrollableConfigurationTable
       rows={[
+        getConfigurationRow({
+          formik,
+          label: "Ceph OSD pool name",
+          name: "ceph_osd_pool_name",
+          defaultValue: "",
+          children: <Input type="text" placeholder="Enter pool name" />,
+        }),
```
edlerd: The presence of this has to be guarded with the feature flag.

PR #1880, opened 2026-03-17
feat: [WD-34759] Clarify ACL dropped vs rejected

Kxiru:
## Done

- Adds clarification to ACL rules
Outstanding questions: Considering a drive-by fix adding LXD priority as help text:
image

## QA

1. Run the LXD-UI:
    - Observe new Custom select for ACL rules clarifying options.

## Screenshots

image


src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -59,15 +60,61 @@ const NetworkAclRuleModal: FC<Props> = ({
...
+          value={formik.values.action}
+          onChange={(value) => {
+            formik.setFieldValue("action", value);
+          }}
+          header={
+            <div className="header">
+              <span className="name">Action</span>
+              <span className="policy">Description</span>
+            </div>
+          }
```
edlerd: Can we have no or an empty header? It seems not needed.

src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -59,15 +60,61 @@ const NetworkAclRuleModal: FC<Props> = ({
...
           options={[
-            { label: "Allow", value: "allow" },
-            { label: "Reject", value: "reject" },
-            { label: "Drop", value: "drop" },
+            {
+              label: (
+                <div className="label network-type-label">
+                  <span className="name">Allow</span>
+                  <span className="members">
+                    Accepts the traffic and permits it to reach its destination.
```
edlerd:
```suggestion
                    Accepts the traffic.
```
I think this is clear enough. Wdyt?

src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -59,15 +60,61 @@ const NetworkAclRuleModal: FC<Props> = ({
...
+              ),
+              text: "Reject",
+              value: "reject",
+            },
+            {
+              label: (
+                <div className="label network-type-label">
+                  <span className="name">Drop</span>
+                  <span className="members">
+                    Blocks the traffic without sending any response.
```
edlerd:
```suggestion
                    Blocks the traffic without response.
```

src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -59,15 +60,61 @@ const NetworkAclRuleModal: FC<Props> = ({
...
+              text: "Allow",
+              value: "allow",
+            },
+            {
+              label: (
+                <div className="label network-type-label">
+                  <span className="name">Reject</span>
+                  <span className="members">
+                    Blocks the traffic and notifies the sender that it was
+                    refused.{" "}
```
edlerd:
```suggestion
                    Blocks the traffic and notifies the sender about it.
```

src/pages/networks/forms/NetworkAclRuleModal.tsx
edlerd: Second thought: The original copy suggested by you might actually be clearer here.

src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -61,15 +60,52 @@ const NetworkAclRuleModal: FC<Props> = ({
...
+          value={formik.values.action}
+          onChange={(value) => {
+            formik.setFieldValue("action", value);
+          }}
           options={[
-            { label: "Allow", value: "allow" },
-            { label: "Reject", value: "reject" },
-            { label: "Drop", value: "drop" },
+            {
+              label: (
```
MichaelKapustey: What do you think about having an internal component named OptionLabel? Smth like:
```tsx
(props)=><div className="label network-type-label">
    <span className="name">{props.name}</span>
    <span className="member">{props.members}</span>
</div>
```
You may use it 3 times in this file and it will make the code easier to read.
However, the customization is limited. If you expect the internal structure to change differently per element its not worth it

Kxiru: The idea of using a function to render a these labels within this component can definitely save us some lines. I'd approach it a little different to how you did, but I think it's a great idea. Please hold that thought and I'll push something new.

src/pages/networks/forms/NetworkAclRuleModal.tsx
```diff
@@ -61,15 +60,52 @@ const NetworkAclRuleModal: FC<Props> = ({
       <Form onSubmit={formik.handleSubmit}>
         {/* hidden submit to enable enter key in inputs */}
         <Input type="submit" hidden value="Hidden input" />
-        <Select
+
+        <CustomSelect
           id="action"
           label="Action"
+          wrapperClassName="select-input"
+          dropdownClassName="instance-target-dropdown"
```
omarelkashef: This classname as well as `network-type-label`  in the options below add necessary css but the names are not accurate here.

src/pages/networks/forms/NetworkAclRuleModal.tsx
Kxiru: Addressed this!

PR #1883, opened 2026-03-19
feat(edit cluster member): Add roles

### LINKED ISSUE #1859: Modify cluster member roles
https://documentation.ubuntu.com/lxd/stable-5.21/explanation/clusters/#member-roles

Not a widely used feature i think, but I noticed the ability to modify cluster member roles via the UI is missing  

(`ovn-chassis` and `event-hub`)

Image

kimanhou:
## Done

- Add cluster member role in EditClusterMemberPanel

Fixes #1859

### 🛠 Prerequisites
*Ensure your environment matches the following before testing:*

- [ ] **Environment:** Use the demo server (after we enable clustering) or a [local development build](https://github.com/canonical/lxd-ui/blob/main/CONTRIBUTING.md#setting-up-for-development).
- [ ] **Clustering:** Must be enabled. (To enable clustering in the UI, navigate to `Server` > `Clustering` -> `Enable clustering`).
- [ ] **Data State:** Ensure at least one cluster member is present and editable.

### 📝 QA steps
| Step | Action | Expected result |
| :--- | :--- | :--- |
| 1 | Open the **Edit cluster member** side panel. | The `Roles` section is now visible. |
| 2 | Check role state. | Current member roles are correctly pre-selected. |
| 3 | Attempt to toggle `database-*` roles. | `voter`, `leader`, and `standby` are **read-only/disabled**. |
| 4 | Toggle `control-plane` or `ovn-chassis`. | The checkboxes are interactive and hold their state. |
| 5 | Click **Save changes**. | Panel closes and changes are persisted (verify by re-opening). |

### BEFORE
*Member roles are missing from the side panel.*
image

### AFTER
*Roles visible; read-only logic applied to database roles.*
image
image

### DIFF
--- a/src/context/useSupportedFeatures.tsx
+++ b/src/context/useSupportedFeatures.tsx
@@ -49,5 +49,6 @@ export const useSupportedFeatures = () => {
     hasInstanceBootMode: apiExtensions.has("instance_boot_mode"),
     hasProjectDeleteOperation: apiExtensions.has("project_delete_operation"),
     hasRemoteDropSource: apiExtensions.has("storage_remote_drop_source"),
+    hasClusteringControlPlane: apiExtensions.has("clustering_control_plane"),
   };
 };
--- a/src/pages/cluster/ClusterMemberOverview.tsx
+++ b/src/pages/cluster/ClusterMemberOverview.tsx
@@ -4,12 +4,17 @@ import ResourceLink from "components/ResourceLink";
 import type { LxdClusterMember } from "types/cluster";
 import ClusterMemberStatus from "pages/cluster/ClusterMemberStatus";
 import { ROOT_PATH } from "util/rootPath";
+import { getClusterMemberRolesList } from "util/clusterMember";
 
 interface Props {
   member: LxdClusterMember;
 }
 
 const ClusterMemberOverview: FC<Props> = ({ member }) => {
+  const { automaticRoles, customRoles } = getClusterMemberRolesList(
+    member.roles,
+  );
+
   return (
     <Row className="general">
       <Col size={3}>
@@ -37,12 +42,16 @@ const ClusterMemberOverview: FC<Props> = ({ member }) => {
               <td>{member.message}</td>
             </tr>
             <tr>
-              <th className="u-text--muted">Url</th>
+              <th className="u-text--muted">URL</th>
               <td>{member.url}</td>
             </tr>
             <tr>
-              <th className="u-text--muted">Roles</th>
-              <td>{member.roles.join(", ")}</td>
+              <th className="u-text--muted">Automatic roles</th>
+              <td>{automaticRoles}</td>
+            </tr>
+            <tr>
+              <th className="u-text--muted">Custom roles</th>
+              <td>{customRoles}</td>
             </tr>
             <tr>
               <th className="u-text--muted">Groups</th>
--- a/src/pages/cluster/ClusterMemberRichTooltip.tsx
+++ b/src/pages/cluster/ClusterMemberRichTooltip.tsx
@@ -13,6 +13,7 @@ import { queryKeys } from "util/queryKeys";
 import { formatSeconds } from "util/seconds";
 import ResourceLabel from "components/ResourceLabel";
 import { ROOT_PATH } from "util/rootPath";
+import { getClusterMemberRolesList } from "util/clusterMember";
 
 interface Props {
   clusterMember: string;
@@ -41,6 +42,10 @@ const ClusterMemberRichTooltip: FC<Props> = ({ clusterMember }) => {
       </>
     );
   }
+
+  const { automaticRoles, customRoles } = getClusterMemberRolesList(
+    member?.roles,
+  );
 
   const rows: TooltipRow[] = [
     {
@@ -81,9 +86,14 @@ const ClusterMemberRichTooltip: FC<Props> = ({ clusterMember }) => {
       valueTitle: member?.url || "",
     },
     {
-      title: "Roles",
-      value: member ? member.roles.join(", ") : "-",
-      valueTitle: member ? member.roles.join(", ") : "",
+      title: "Automatic roles",
+      value: automaticRoles,
+      valueTitle: automaticRoles,
+    },
+    {
+      title: "Custom roles",
+      value: customRoles,
+      valueTitle: customRoles,
     },
     {
       title: "Uptime",
--- a/src/pages/cluster/panels/ClusterMemberRolesSelector.tsx
+++ b/src/pages/cluster/panels/ClusterMemberRolesSelector.tsx
@@ -1,0 +1,67 @@
+import type { FC, ReactNode } from "react";
+import { MultiSelect } from "@canonical/react-components";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
+import { classifyClusterMemberRoles } from "util/clusterMember";
+
+interface Props {
+  setSelectedRoles: (roles: string[]) => void;
+  selectedRoles: string[];
+  id?: string;
+  help?: ReactNode;
+  label?: string;
+}
+
+const ClusterMemberRolesSelector: FC<Props> = ({
+  setSelectedRoles,
+  selectedRoles,
+  id = "cluster-member-roles-selector",
+  help,
+  label = "Custom roles",
+}) => {
+  const { hasClusteringControlPlane } = useSupportedFeatures();
+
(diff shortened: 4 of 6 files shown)


src/pages/cluster/panels/EditClusterMemberPanel.tsx
```diff
@@ -124,6 +128,20 @@ const EditClusterMemberPanel: FC<Props> = ({ onClose }) => {
...
+              setSelectedRoles={(roles) => {
+                formik.setFieldValue("roles", roles);
+              }}
+              label={
+                <HelpLink
+                  docPath="/explanation/clusters/#member-roles"
+                  title="Learn more about cluster member roles"
+                >
+                  Roles
+                </HelpLink>
```
edlerd: The icon gets slightly misaligned, see screenshot below. I think it could be better to display a normal link in the muted help text below the component. If we keep the icon next to the label, I think it should be outside the `<label>` tags, but I'm not 100% sure on this one.


Image

src/pages/cluster/panels/ClusterMemberRolesSelector.tsx
```diff
@@ -0,0 +1,70 @@
...
+  selectedRoles,
+  id = "cluster-member-roles-selector",
+  help,
+  label,
+}) => {
+  const ROLES = [
+    "database-voter",
+    "database-leader",
+    "database-standby",
+    "control-plane",
```
edlerd: Shall we use the api extension to display the `event-hub` here instead of `control-plane` for when the extension is not present?

src/pages/cluster/panels/EditClusterMemberPanel.tsx
kimanhou: Okay, let's find a good solution for this and I'll create a component that can be re-used in the application. I want to add more of these documentation links everywhere as raised in the user testing (if there is a relevant documentation about an input, then link it in the form)

src/pages/cluster/panels/ClusterMemberRolesSelector.tsx
kimanhou: I also asked in LXD private about `database` vs `database-voter`

src/pages/cluster/panels/EditClusterMemberPanel.tsx
kimanhou: image

src/context/useSupportedFeatures.tsx
```diff
@@ -48,5 +48,6 @@ export const useSupportedFeatures = () => {
     hasInstanceForceDelete: apiExtensions.has("instance_force_delete"),
     hasInstanceBootMode: apiExtensions.has("instance_boot_mode"),
     hasProjectDeleteOperation: apiExtensions.has("project_delete_operation"),
+    hasClusteringControlPlane: apiExtensions.has("clustering_control_plane"),
```
kimanhou, line 52: New api extension
In <=5.21, the member role is called `event-hub`
From 6.7, the member role is called `controle-plane`

src/pages/cluster/panels/ClusterMemberRolesSelector.tsx
```diff
@@ -0,0 +1,73 @@
...
+          setSelectedRoles(items.map((item) => item.value as string));
+        }}
+        showDropdownFooter={false}
+        id={id}
+        disabledItems={AUTOMATIC_ROLES.map((t) => {
+          return { label: t, value: t };
+        })}
+        help={help}
+        hasSelectedItemsFirst={false}
+        isSortedAlphabetically={false}
```
kimanhou: By default, the selected items will be placed first and the items will be sorted by alphabetical order.
I chose to keep the same order as the documentation and keep it fixed with the automatic roles first.

src/pages/cluster/ClusterMemberOverview.tsx
```diff
@@ -37,7 +37,7 @@ const ClusterMemberOverview: FC<Props> = ({ member }) => {
               <td>{member.message}</td>
             </tr>
             <tr>
-              <th className="u-text--muted">Url</th>
+              <th className="u-text--muted">URL</th>
```
kimanhou, line 45: Unrelated to this PR.
I noticed that the cluster member rich chip, the word URL was written in all capitals, so I changed it here for consistency.

src/pages/cluster/panels/EditClusterMemberPanel.tsx
```diff
@@ -124,6 +128,18 @@ const EditClusterMemberPanel: FC<Props> = ({ onClose }) => {
               label="Failure domain"
               placeholder="Enter failure domain"
             />
+            <ClusterMemberRolesSelector
```
edlerd, line 131: The selector looks slightly broken with no custom roles attached

Image

kimanhou: Fixed
image

src/pages/cluster/ClusterMemberRichTooltip.tsx
```diff
@@ -42,6 +43,15 @@ const ClusterMemberRichTooltip: FC<Props> = ({ clusterMember }) => {
...
   }
 
+  const clusterMemberAutomaticRoles =
+    member?.roles && member?.roles.length > 0
+      ? member?.roles.filter((t) => isClusterMemberRoleAutomatic(t)).join(", ")
+      : "-";
+  const clusterMemberConfiguredRoles =
+    member?.roles && member?.roles.length > 0
+      ? member?.roles.filter((t) => !isClusterMemberRoleAutomatic(t)).join(", ")
+      : "-";
```
edlerd: 1. Both might render empty where they should render a dash. In case there are only automatic or only custom roles assigned.

2. This is repeating logic from the overview component. Maybe a better approach is to do `const {automaticRoles, customRoles } = classifyClusterMemberRoles(member.roles);`. Wdyt?

3. I'd drop the `clusterMember` prefix, as the context here implies that. The same goes for the overview component.

src/pages/cluster/panels/ClusterMemberRolesSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+    hasClusteringControlPlane ? "control-plane" : "event-hub",
+    "ovn-chassis",
+  ];
+
+  const hasRoles = selectedRoles.length > 0;
+
+  return (
+    <>
+      {label && <label htmlFor={id}>{label}</label>}
+      <MultiSelect
```
edlerd, line 36: Once items are selected, we should display them, not the placeholder. Like in the ACL selector shown in this video

https://github.com/user-attachments/assets/06f62999-0a22-46c1-af73-a9132cacf15a

kimanhou: I could not see the video but it's okay, I know the expected behavior.
Sorry, I got confused with hasRoles and hasAcls, they are actually not the same thing at all.
Fixed now.


https://github.com/user-attachments/assets/6152f97f-a7b8-4660-ad52-3e83f03f1631

src/pages/cluster/panels/EditClusterMemberPanel.tsx
```diff
@@ -124,6 +129,20 @@ const EditClusterMemberPanel: FC<Props> = ({ onClose }) => {
...
+              selectedRoles={formik.values.roles.filter(
+                (role) => !isClusterMemberRoleAutomatic(role),
+              )}
+              setSelectedRoles={(roles) => {
+                formik.setFieldValue("roles", roles);
+              }}
```
edlerd: Do we store all roles in `formik.values.roles`? If yes, then the set roles will remove any non automatic roles. Will this lead to inconsistency?

Maybe better to filter the custom roles on initial state, and here simply forwarding the value
```suggestion
              selectedRoles={formik.values.roles}
              setSelectedRoles={(roles) => {
                formik.setFieldValue("roles", roles);
              }}
```

src/pages/cluster/panels/ClusterMemberRolesSelector.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const ROLES = [
+    hasClusteringControlPlane ? "control-plane" : "event-hub",
+    "ovn-chassis",
+  ];
+
+  return (
+    <>
+      {label && <label htmlFor={id}>{label}</label>}
+      <MultiSelect
+        label="Cluster member roles"
```
edlerd: We should only have one label for this field.

PR #1885, opened 2026-03-20
feat: filter out used networks in network selector WD-34762

omarelkashef:
## Done

- Filter out used networks from network selector

1. Run the LXD-UI:
    - Go to Instance
    - Create or edit instance
    - Go to Configuration -> Devices -> Networks
    - Attach or edit network
    - Ensure any network already attached to the instance does not show in the network selector unless dns.mode = "none" on the network or you are currently editing that network.
   NOTE: it is possible now to create a profile attached to a network twice. This will cause an error when starting an instance with that profile. This can be fixed in another PR.

### DIFF
--- a/src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
+++ b/src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
@@ -41,6 +41,8 @@ import NetworkDefaultACLSelector, {
 import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
 import type { NetworkDeviceFormValues } from "types/forms/networkDevice";
 import { useNetworkAcls } from "context/useNetworkAcls";
+import { getAttachedNetworkNames } from "util/devices";
+import type { LxdNetwork } from "types/network";
 
 interface Props {
   project: string;
@@ -82,11 +84,31 @@ const NetworkDevicePanel: FC<Props> = ({
   const isEditingOverride = inheritedDevice && device;
 
   const isOverride = Boolean(isCreatingOverride || isEditingOverride);
+  const isEditing = panelParams.panel === panels.editNetworkDevice;
 
-  const managedNetworks = networks.filter((network) => network.managed);
+  const attachedNetworks = getAttachedNetworkNames(
+    parentFormik.values,
+    inheritedNetworks,
+  );
+
+  const isAvailable = (network: LxdNetwork) => {
+    if (!network.managed) {
+      return false;
+    }
+    const isSelectedNetwork =
+      isEditing &&
+      (device?.network === network.name ||
+        inheritedDevice?.network?.network === network.name);
+    const isAttached = attachedNetworks.includes(network.name);
+    const isProfile = parentFormik.values.entityType === "profile";
+    const allowsDuplicateAttachment =
+      network.config["dns.mode"] === "none" || isProfile;
+    return isSelectedNetwork || !isAttached || allowsDuplicateAttachment;
+  };
+  const networkOptions = networks.filter(isAvailable);
 
   const getNetworkAclsByNetworkName = (networkName: string) => {
-    const network = managedNetworks.find((n) => n.name === networkName);
+    const network = networkOptions.find((n) => n.name === networkName);
     return getNetworkAcls(network);
   };
 
@@ -125,7 +147,7 @@ const NetworkDevicePanel: FC<Props> = ({
       });
 
   const getInitialValues = (): NetworkDeviceFormValues => {
-    const defaultNetworkName = managedNetworks[0]?.name ?? "";
+    const defaultNetworkName = networkOptions[0]?.name ?? "";
     const defaultAcls =
       getNetworkAclsByNetworkName(defaultNetworkName).join(",");
 
@@ -181,7 +203,7 @@ const NetworkDevicePanel: FC<Props> = ({
     onSubmit: (values) => {
       const allSelectedAcls = values.acls ? values.acls.split(",") : [];
       const networkAcls = getNetworkAcls(
-        managedNetworks.find((n) => n.name === values.network),
+        networkOptions.find((n) => n.name === values.network),
       );
       const userSelectedAcls = allSelectedAcls.filter(
         (acl) => !networkAcls.includes(acl),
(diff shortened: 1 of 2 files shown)


src/util/devices.tsx
```diff
@@ -173,6 +174,31 @@ export const getExistingDeviceNames = (
...
+      .map((item) => (item as FormNetworkDevice).network ?? ""),
+  );
+  // Returns networks from any of the profiles assigned to the instance
+  if (values.entityType === "instance") {
+    const appliedProfiles = getAppliedProfiles(values, profiles);
+    for (const profile of appliedProfiles) {
+      Object.values(profile.devices).map((item) => {
+        existingNetworksNames.push((item as LxdNicDevice).network);
+      });
+    }
```
edlerd: We need to be more careful here. Some of those inherited devices can be overridden by other profiles or by instance networks. Maybe better to use the `getInheritedNetworks` helper. Which we actually already call in `NetworkDevicePanel`, so perhaps we can pass those into this function instead of calculating them afresh.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -82,8 +85,23 @@ const NetworkDevicePanel: FC<Props> = ({
...
+      network.config["dns.mode"] === "none"
+    );
+  };
 
-  const managedNetworks = networks.filter((network) => network.managed);
+  const managedNetworks = networks.filter(
+    (network) =>
+      network.managed &&
+      (isEditingNetwork(network) || canAttachNetworkAgain(network)),
+  );
```
edlerd, line 92:
```suggestion
  const isAvailable = (network: LxdNetwork) => {
    if (network.managed) {
      return false;
    }

    const isSelectedNetwork = isEditing && device?.network === network.name;
    const isUnused = !existingNetworks.includes(network.name);
    const allowsDuplicateAttachment = network.config["dns.mode"] === "none";

    return isSelectedNetwork || isUnused || allowsDuplicateAttachment;
  };

  const networkOptions = networks.filter(isAvailable);
```
Suggesting to slightly restructure the filter logic here. Wdyt?

src/util/devices.tsx
```diff
@@ -173,6 +175,32 @@ export const getExistingDeviceNames = (
...
+        (device) => device.name === network.key,
+      );
+      if (!localOverride) {
+        existingNetworksNames.push(network.network?.network ?? "");
+      }
+    }
+  }
+  return existingNetworksNames;
+};
+
```
```suggestion
export const getConnectedNetworkNames = (
  values: InstanceAndProfileFormValues,
  inheritedNetworks: InheritedNetwork[],
): string[] => {
  const connectedNetworks = values.devices
    .filter((item) => isNicDevice(item))
    .map((item) => item.network ?? "");

  const localDevices = new Set(values.devices.map((item) => item.name));
  if (values.entityType === "instance") {
    for (const inheritedNetwork of inheritedNetworks) {
      // Find a local device that overrides the inherited device by key (device name)
      const hasLocalOverride = localDevices.has(inheritedNetwork.key);
      if (!hasLocalOverride) {
        connectedNetworks.push(inheritedNetwork.network?.network ?? "");
      }
    }
  }
  return connectedNetworks;
};
```
Suggesting slightly different structure here.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -82,11 +84,31 @@ const NetworkDevicePanel: FC<Props> = ({
...
+      isEditing &&
+      (device?.network === network.name ||
+        inheritedDevice?.network?.network === network.name);
+    const isUnused = !existingNetworks.includes(network.name);
+    const isProfile = parentFormik.values.entityType === "profile";
+    const allowsDuplicateAttachment =
+      network.config["dns.mode"] === "none" || isProfile;
+    return isSelectedNetwork || isUnused || allowsDuplicateAttachment;
+  };
+  const networkOptions = networks.filter(isAvailable);
```
edlerd, line 108: There is a new edge case when this list is empty on attaching a new network. I am not sure how to improve it, but maybe there is a way to make that less frustrating for the user. So they don't run directly into the error case on attaching a new network where none is available..

src/util/devices.tsx
edlerd: Another thought on naming. Maybe stick with `getAttachedNetworkNames`. And also change it to `isAttached` instead of `isUsed` in the other file. Then naming matches the UI labels.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -82,11 +84,31 @@ const NetworkDevicePanel: FC<Props> = ({
   const isEditingOverride = inheritedDevice && device;
 
   const isOverride = Boolean(isCreatingOverride || isEditingOverride);
+  const isEditing = panelParams.panel === panels.editNetworkDevice;
 
-  const managedNetworks = networks.filter((network) => network.managed);
+  const existingNetworks = getAttachedNetworkNames(
```
edlerd:
```suggestion
  const attachedNetworks = getAttachedNetworkNames(
```

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -221,14 +243,21 @@ const NetworkDevicePanel: FC<Props> = ({
     },
   });
 
+  const networkSelectorError =
+    networkOptions.length === 0
+      ? "No available networks to attach. Please create and manage networks in the Networks page before attaching them here."
```
edlerd:
```suggestion
      ? "No available networks found."
```
I think this message can be more brief. Wdyt?

src/util/devices.tsx
```diff
@@ -173,6 +175,32 @@ export const getExistingDeviceNames = (
...
+  values: InstanceAndProfileFormValues,
+  inheritedNetworks: InheritedNetwork[],
+): string[] => {
+  const connectedNetworks = values.devices
+    .filter((item) => isNicDevice(item) || isNoneDevice(item))
+    .map((item) => item as FormNetworkDevice);
+
+  const connectedNetworksNames = connectedNetworks.map(
+    (device) => device.network || "",
+  );
```
```suggestion
    const connectedNetworksNames = values.devices
    .filter((item) => isNicDevice(item))
    .map((item) => item.network);
```
A bit simpler?

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -221,14 +243,21 @@ const NetworkDevicePanel: FC<Props> = ({
...
+      : formik.touched.network
+        ? formik.errors.network
+        : undefined;
```
edlerd: It's better to avoid nesting the ternaries here. This can function with early exits to manage the different priorities of error messages.

src/util/devices.tsx
omarelkashef: Yes, this split was necessary in a previous iteration but no anymore. But I believe having the `isNoneDevice` is necessary so that we do not include overriden detached networks from profiles.

edlerd: The mapping would just produce empty strings on the none devices.  We consider them below on the inherited devices anyways, but that is independent from the code here.

src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
```diff
@@ -221,14 +243,20 @@ const NetworkDevicePanel: FC<Props> = ({
     },
   });
 
+  const networkSelectorError = () => {
+    if (networkOptions.length === 0) return "No available networks found.";
+    if (formik.touched.network) return formik.errors.network;
```
edlerd:
```suggestion
    if (networkOptions.length === 0) {
      return "No available networks found.";
    }
    if (formik.touched.network) {
      return formik.errors.network;
    }
```
Please no shorthand style, it tends to lead to bugs on refactor or due to oversight.

PR #1894, opened 2026-03-23
fix: screenshot automation workflow

omarelkashef:
## Done

- Fix screenshot automation workflow

## QA

1. Run the LXD-UI:
    - Ensure workflow passes.
    - Successful run: https://github.com/omarelkashef/lxd-ui/actions/runs/23442787149/job/68198134121

### DIFF
--- a/tests/docs-screenshots.spec.ts
+++ b/tests/docs-screenshots.spec.ts
@@ -732,10 +732,11 @@ test("LXD - UI Folder - Networks", async ({ page }) => {
   await page.getByText("IPv4 address", { exact: true }).click();
   await page.getByText("/24").getByRole("button").click();
   let networkSubnet = await page.inputValue("input#ipv4_address");
-  let listenAddress = networkSubnet.replace("1/24", "1");
-  let targetAddress = networkSubnet.replace("1/24", "3");
+  const listenOctets = "1";
+  let listenAddress = networkSubnet.replace("1/24", listenOctets);
   await page.getByRole("link", { name: "Forwards" }).click();
   await page.getByTitle("Create forward").click();
+  const targetAddress = "3";
   await page.getByLabel("Listen address").fill(listenAddress);
   await page.getByLabel("Default target address").fill(targetAddress);
   await page.screenshot({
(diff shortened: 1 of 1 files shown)


tests/docs-screenshots.spec.ts
```diff
@@ -732,10 +732,11 @@ test("LXD - UI Folder - Networks", async ({ page }) => {
   await page.getByText("IPv4 address", { exact: true }).click();
   await page.getByText("/24").getByRole("button").click();
   let networkSubnet = await page.inputValue("input#ipv4_address");
-  let listenAddress = networkSubnet.replace("1/24", "1");
-  let targetAddress = networkSubnet.replace("1/24", "3");
+  const listenOctets = "1";
+  const listenAddress = networkSubnet.replace("1/24", listenOctets);
```
kimanhou: Sorry I changed these `tests > helpers > network.ts > createNetworkForward` but didn't see this occurrence. We need to find a way to link this with the e2e tests. Is there a way `docs-screenshots` uses `createNetworkForward` from the helpers ?

omarelkashef: No problem. I created a ticket to do a more thorough refactoring of the screenshot workflow to use e2e helpers https://warthogs.atlassian.net/browse/WD-35326. I would leave it as it is for now because I suspect the refactoring will be tricky because we want to use helpers but still take screenshots of different pages.

tests/docs-screenshots.spec.ts
```diff
@@ -800,8 +801,6 @@ test("LXD - UI Folder - Networks", async ({ page }) => {
 
   await page.getByText("/24").getByRole("button").click();
   networkSubnet = await page.inputValue("input#ipv4_address");
-  listenAddress = networkSubnet.replace("1/24", "1");
```
edlerd, line 803: Don't we need to keep this here? Otherwise we create a forward with a listen address already in use by the other network.

omarelkashef: Yes, it should stay. Thank you for catching it.

PR #1895, opened 2026-03-23
feat: [WD-34755] Replace disabled input fields with muted text

Kxiru:
## Done

- Created a new read-only component to provide muted Name fields for entities whose names cannot be changed. This PR only covers Instances, however addition iterations should be made to cover all relevant entity types.

## QA

1. Run the LXD-UI:
    - Navigate to the configuration page of an Instance and verify that the instance name in the config section cannot be changed and presents as displayed within the screenshot.

## Screenshots

image

### DIFF
--- a/src/components/OutputField.tsx
+++ b/src/components/OutputField.tsx
@@ -1,0 +1,27 @@
+import type { FC } from "react";
+import { Field } from "@canonical/react-components";
+
+interface Props {
+  id: string;
+  label: string;
+  value: string;
+  help?: string;
+}
+
+const OutputField: FC<Props> = ({ id, label, value, help }) => {
+  return (
+    <Field
+      forId={id}
+      label={label}
+      help={help}
+      labelClassName="u-no-margin--bottom"
+      className="output-field"
+    >
+      <output id={id} className="mono-font u-sv2">
+        <b>{value}</b>
+      </output>
+    </Field>
+  );
+};
+
+export default OutputField;
--- a/src/pages/instances/forms/EditInstanceDetails.tsx
+++ b/src/pages/instances/forms/EditInstanceDetails.tsx
@@ -1,5 +1,5 @@
 import type { FC } from "react";
-import { Col, Input, Row, Select } from "@canonical/react-components";
+import { Col, Row, Select } from "@canonical/react-components";
 import ProfileSelector from "pages/profiles/ProfileSelector";
 import type { FormikProps } from "formik/dist/types";
 import type { EditInstanceFormValues } from "types/forms/instanceAndProfile";
@@ -10,6 +10,7 @@ import SshKeyForm from "components/forms/SshKeyForm";
 import { useIsClustered } from "context/useIsClustered";
 import PlacementGroupSelect from "pages/instances/forms/PlacementGroupSelect";
 import { getInstanceField } from "util/instanceConfigFields";
+import OutputField from "components/OutputField";
 
 export const instanceEditDetailPayload = (values: EditInstanceFormValues) => {
   return {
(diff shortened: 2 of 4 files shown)


src/components/MutedEntityName.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface Props {
+  value: string;
+  help?: string;
+}
+
+const MutedEntityName: FC<Props> = ({ value, help }) => {
+  return (
+    <div>
+      <Label>Name</Label>
+      <div className="mono-font  u-sv2">
```
edlerd:
```suggestion
      <output id={id} className="mono-font u-sv2">
```
If we use the output tag here, we can give it also an id and point the label at it.

src/components/MutedEntityName.tsx
```diff
@@ -0,0 +1,21 @@
+import type { FC } from "react";
+import { Label } from "@canonical/react-components";
+
+interface Props {
+  value: string;
+  help?: string;
+}
+
+const MutedEntityName: FC<Props> = ({ value, help }) => {
```
edlerd:
```suggestion
const OutputField: FC<Props> = ({ value, help }) => {
```
We also want to use this for other contexts, so naming should not relate to its usage as entity name.

src/components/MutedEntityName.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <div>
```
edlerd: 1. Spacing is not equal to form elements. Be aware spacing changes depending on screen dimensions on the forms. It might be best to reuse the `Field` component from react components for this instead of rolling our own composition of label, and help text.
2. We should expose the label as a prop for reuse.
```suggestion
    <Field label={label} help={help}>
```

edlerd: See 
Screenshot from 2026-03-24 07-20-24

src/components/OutputField.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface Props {
+  id: string;
+  label: string;
+  value: string;
+  help?: string;
+}
+
+const OutputField: FC<Props> = ({ id, label, value, help }) => {
+  return (
+    <Field label={label} help={help}>
```
edlerd: This is missing the `forId` property. Currently creates an orphaned label.

edlerd:
```suggestion
    <Field label={label} help={help} labelClassName="u-no-margin--bottom">
```
Let's reduce the spacing a bit.

src/components/OutputField.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Field } from "@canonical/react-components";
+
+interface Props {
+  id: string;
+  label: string;
+  value: string;
+  help?: string;
+}
+
+const OutputField: FC<Props> = ({ id, label, value, help }) => {
```
edlerd, line 11: This looks nearly perfect. Let's tweak the spacing for the help text a bit. The default margin-top is `-0.5 rem` through vanilla. Best to reduce it to `-0.65rem` in my opinion.

To implement, I'd give the field a class name, such as `output-field` and then have some custom SCSS target the `.output-field .p-form-help-text` with the increased negative margin.

edlerd: Screenshot from 2026-03-25 09-01-34

kimanhou: Agreed with David about the help text, it should sit closer to the label.

I also experimented with the spacing between `Name` and `DemoInstance`. With a margin-top of 1.75rem, it almost takes the same space as the Input, which leaves the rest of the screen untouched. However it looks weird:
image

I suggest having a tiny margin-top of 0.25 rem for `DemoInstance`.
image

edlerd: Wild idea that we could also give the output a 1rem left margin so it aligns with the inputs. Not sure if that is better though:

image

kimanhou: I bet Lyubo would like it. I don't.

Kxiru: Also not a fan of the left margin, m'afraid. Nothing else on the page is in alignment with it.
@kimanhou , @edlerd , please take a look at the latest changes reflecting your suggestions.

src/sass/_detail_page.scss
```diff
@@ -2,3 +2,13 @@
...
+
+.output-field {
+  .p-form-help-text {
+    margin-top: -0.65rem;
+  }
+
+  .p-form__control {
+    margin-top: 0.25rem;
+  }
+}
```
edlerd: This is best in a dedicated file.

PR #1899, opened 2026-03-25
fix(storage pool): rely on operation when api extension is present

kimanhou:
## Done

- fix(storage pool): rely on operation when api extension `storage_and_network_operations` is present
- Affected methods:  `createPool`, `createClusteredPool`, `updatePool`, `updateClusteredPool`, `renameStoragePool` and `deleteStoragePool`.

Follow up of https://github.com/canonical/lxd/pull/17955

## QA

1. Run the LXD-UI:
2. Perform the following QA steps and for each step, make sure the action is performed successfully and that 2 notifications (started and success) are displayed correctly:
    - Create pool
    - Create pool in a cluster
    - Update pool
    - Update clustered pool
    - Rename pool
    - Delete pool
_Note: QA cannot be performed on the demo server, as the demo server doesn't have clustering enabled. Reviewers have to check out the branch locally._

## Screenshots
image

### DIFF
--- a/src/api/operations.tsx
+++ b/src/api/operations.tsx
@@ -38,3 +38,44 @@ export const cancelOperation = async (id: string): Promise<void> => {
     method: "DELETE",
   }).then(handleResponse);
 };
+
+export const waitForOperation = async (
+  id: string,
+  member?: string,
+  maxTimeoutMs = 120000,
+): Promise<void> => {
+  const endpoint = `${ROOT_PATH}/1.0/operations/${encodeURIComponent(id)}`;
+  const startTime = Date.now();
+  const memberPrefix = member ? `Member: ${member} - ` : "";
+
+  while (true) {
+    try {
+      const response = (await fetch(endpoint).then(
+        handleResponse,
+      )) as LxdApiResponse<LxdOperation>;
+      const operation = response.metadata;
+
+      if (operation.status_code === 200) {
+        return;
+      }
+
+      if (operation.status_code >= 400) {
+        throw new Error(
+          `${operation.err || `Operation ${id} failed with status ${operation.status}`}`,
+        );
+      }
+    } catch (error) {
+      // Add memberPrefix to all other errors before re-throwing
+      if (error instanceof Error) {
+        throw new Error(`${memberPrefix}${error.message}`);
+      }
+      throw new Error(`${memberPrefix}Unknown error occurred`);
+    }
+
+    await new Promise((resolve) => setTimeout(resolve, 1000));
+
+    if (Date.now() - startTime > maxTimeoutMs) {
+      throw new Error(`${memberPrefix}Operation timed out.`);
+    }
+  }
+};
--- a/src/api/storage-pools.tsx
+++ b/src/api/storage-pools.tsx
@@ -2,7 +2,6 @@ import {
   constructMemberError,
   handleEtagResponse,
   handleResponse,
-  handleSettledResult,
 } from "util/helpers";
 import type {
   LxdStoragePool,
@@ -12,9 +11,11 @@ import type {
 import type { LxdApiResponse } from "types/apiResponse";
 import type { LxdClusterMember } from "types/cluster";
 import type { ClusterSpecificValues } from "types/cluster";
+import type { LxdOperationResponse } from "types/operation";
 import { addEntitlements } from "util/entitlements/api";
 import { addTarget } from "util/target";
 import { ROOT_PATH } from "util/rootPath";
+import { waitForOperation } from "api/operations";
 
 export const storagePoolEntitlements = ["can_edit", "can_delete"];
 
@@ -101,17 +102,21 @@ export const fetchClusteredStoragePoolResources = async (
 export const createPool = async (
   pool: Partial<LxdStoragePool>,
   target?: string,
-): Promise<void> => {
+): Promise<LxdOperationResponse> => {
   const params = new URLSearchParams();
   addTarget(params, target);
 
-  await fetch(`${ROOT_PATH}/1.0/storage-pools?${params.toString()}`, {
+  return fetch(`${ROOT_PATH}/1.0/storage-pools?${params.toString()}`, {
     method: "POST",
     headers: {
       "Content-Type": "application/json",
     },
     body: JSON.stringify(pool),
-  }).then(handleResponse);
+  })
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
 };
 
 const getClusterAndMemberPoolPayload = (pool: LxdStoragePool) => {
@@ -146,13 +151,15 @@ const getClusterAndMemberPoolPayload = (pool: LxdStoragePool) => {
 export const createClusteredPool = async (
   pool: LxdStoragePool,
   clusterMembers: LxdClusterMember[],
+  hasStorageAndNetworkOperations: boolean,
   sourcePerClusterMember?: ClusterSpecificValues,
   zfsPoolNamePerClusterMember?: ClusterSpecificValues,
   sizePerClusterMember?: ClusterSpecificValues,
-): Promise<void> => {
+): Promise<LxdOperationResponse> => {
   const { memberPoolPayload, clusterPoolPayload } =
     getClusterAndMemberPoolPayload(pool);
-  return Promise.allSettled(
+
+  const operations = await Promise.allSettled(
     clusterMembers.map(async (item) => {
       const clusteredMemberPool = {
         ...memberPoolPayload,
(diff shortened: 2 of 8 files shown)


src/api/storage-pools.tsx
edlerd, line 194: I think this is way harder with the extension. Because we have to wait until all operations of the cluster member pools are finished before we do the `createPool(clusterPoolPayload)` call.

So the `createPool(clusterPoolPayload)` call has to await all operations for each cluster member to finish.

kimanhou: Ah 😬

kimanhou: I implemented a new `waitForOperation` function that polls `/1.0/operations/{id}` endpoint and then the final API call once all operations are in success status.

src/pages/storage/StoragePoolHeader.tsx
```diff
@@ -43,18 +56,37 @@ const StoragePoolHeader: FC<Props> = ({ name, pool, project }) => {
...
-            <>
-              Storage pool <strong>{name}</strong> renamed to{" "}
-              <StoragePoolRichChip
-                poolName={values.name}
-                projectName={project}
-              />
-            </>,
-          );
+
+          if (hasStorageAndNetworkOperations) {
```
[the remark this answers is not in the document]
kimanhou: It looks like you cannot rename a storage pool from the UI so I am wondering why we even have a `renameStoragePool` method

image

src/api/operations.tsx
```diff
@@ -38,3 +38,48 @@ export const cancelOperation = async (id: string): Promise<void> => {
...
+): Promise<void> => {
+  const endpoint = `${ROOT_PATH}/1.0/operations/${encodeURIComponent(id)}`;
+  const startTime = Date.now();
+  let delay = 500;
+
+  while (true) {
+    if (Date.now() - startTime > maxTimeoutMs) {
+      const memberPrefix = member ? `[Member: ${member}] ` : "";
+      throw new Error(`${memberPrefix}Operation timed out.`);
+    }
```
edlerd: Let's move this down after the delay so the function flow is easier to read.

src/api/operations.tsx
```diff
@@ -38,3 +38,48 @@ export const cancelOperation = async (id: string): Promise<void> => {
...
+
+      if (operation.status_code >= 400) {
+        throw new Error(
+          operation.err ||
+            `Operation ${id} failed with status ${operation.status}`,
+        );
+      }
+    } catch (error) {
+      if (error instanceof LxdApiError && error.status === 404) {
+        const memberPrefix = member ? `[Member: ${member}] ` : "";
```
edlerd: 1. This prefix should always be added, not just for a 404. Let's apply it also outside of the conditional block.
2. I am not happy with the `[`  brackets. Can we leave them out and keep it as `Member: foo `?

src/api/operations.tsx
```diff
@@ -38,3 +38,48 @@ export const cancelOperation = async (id: string): Promise<void> => {
...
+        throw new Error(`${memberPrefix}Operation not found on server.`);
+      }
+      // Re-throw all other errors
+      throw error;
+    }
+
+    await new Promise((resolve) => setTimeout(resolve, delay));
+    delay = Math.min(delay * 1.5, 5000);
```
edlerd: Do we really need the backoff here? I think a constant delay of 1 second is better suited.

edlerd (commented):
Thanks for handling the complex clustered case. Some thoughts on it below.

I think we don't have a clustered storage test, would be good to add now with this changeset, or as a follow up in a dedicated task.

src/pages/storage/StoragePoolHeader.tsx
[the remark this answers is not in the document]
edlerd: We purely have this for reusing components. One could make an argument about consistency in how to rename any entity in the UI as well. Do you this this should be changed?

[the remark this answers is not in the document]
kimanhou: Oh how quaint ! I fixed the component so that it might work if one day the renaming of storage pool is implemented.

src/api/operations.tsx
```diff
@@ -38,3 +38,44 @@ export const cancelOperation = async (id: string): Promise<void> => {
...
+
+      if (operation.status_code >= 400) {
+        throw new Error(
+          `${memberPrefix}${operation.err || `Operation ${id} failed with status ${operation.status}`}`,
+        );
+      }
+    } catch (error) {
+      // Add memberPrefix to all other errors before re-throwing
+      if (error instanceof Error) {
+        throw new Error(`${memberPrefix}${error.message}`);
```
edlerd:
```suggestion
          operation.err || `Operation ${id} failed with status ${operation.status}`,
        );
      }
    } catch (error) {
      // Add memberPrefix to all other errors before re-throwing
      if (error instanceof Error) {
        throw new Error(`${memberPrefix}${error.message}`);
```
I think we are adding the prefix twice here.

src/api/storage-pools.tsx
```diff
@@ -149,10 +155,11 @@ export const createClusteredPool = async (
   sourcePerClusterMember?: ClusterSpecificValues,
   zfsPoolNamePerClusterMember?: ClusterSpecificValues,
   sizePerClusterMember?: ClusterSpecificValues,
-): Promise<void> => {
+): Promise<LxdOperationResponse> => {
   const { memberPoolPayload, clusterPoolPayload } =
     getClusterAndMemberPoolPayload(pool);
-  return Promise.allSettled(
+
+  const results = await Promise.allSettled(
```
edlerd:
```suggestion
  const operations = await Promise.allSettled(
```
I think renaming helps to reason about the following lines.

src/api/storage-pools.tsx
```diff
@@ -163,23 +170,41 @@ export const createClusteredPool = async (
...
+  const operationsWithMembers = results
+    .filter(
+      (
+        res,
+      ): res is PromiseFulfilledResult<{
+        operation: LxdOperationResponse;
+        member: string;
+      }> => res.status === "fulfilled",
+    )
+    .map((res) => res.value);
```
edlerd: Both `handleSettledResult` and the `results.filter` seem to be doing similar things. Maybe we can unify them and not rely on the helper function here?
```suggestion
  const pendingOperations = operations.map((res) => {
    if (res.status === "rejected") {
      throw res?.reason as Error;
    }
    return res.value;
  });
```

src/pages/storage/actions/DeleteStoragePoolBtn.tsx
```diff
@@ -35,23 +38,57 @@ const DeleteStoragePoolBtn: FC<Props> = ({
...
+        Storage pool <ResourceLabel bold type="pool" value={poolName} />{" "}
+        deleted.
+      </>,
+    );
+  };
+
+  const onSuccess = () => {
+    queryClient.invalidateQueries({
+      queryKey: [queryKeys.storage],
+    });
```
edlerd: We should invalidate the cache also in case of failure. Maybe move it to a dedicated helper to call it here, in the `catch` and also in the `eventQueue` failure handler.

src/pages/storage/CreateStoragePool.tsx
```diff
@@ -55,6 +58,27 @@ const CreateStoragePool: FC = () => {
...
+        <StoragePoolRichChip poolName={poolName} projectName={project} />{" "}
+        created.
+      </>,
+    );
+  };
+
+  const onSuccess = (storagePoolName: string) => {
+    queryClient.invalidateQueries({
+      queryKey: [queryKeys.storage],
+    });
```
edlerd: similar here about invalidation on failure.

src/pages/storage/EditStoragePool.tsx
```diff
@@ -108,39 +146,44 @@ const EditStoragePool: FC<Props> = ({ pool }) => {
...
-          if (pool.driver === cephDriver && values.ceph_rbd_du === "false") {
-            // Clear the storage volume sizes from the cache. The sizes are not available
-            // after disabling `ceph_rbd_du` and the volume state queries will fail. So we
-            // remove the queries to avoid serving the size from a stale cache.
-            queryClient.removeQueries({
-              predicate: (query) =>
-                query.queryKey[0] === queryKeys.storage &&
-                query.queryKey[1] === pool.name,
-            });
-          }
```
edlerd, line 143: The invalidation now needs to be triggered here, in the event queue success and in the event queue failure. Best move to a helper and call it from those three places.

src/api/storage-pools.tsx
```diff
@@ -220,32 +250,65 @@ export const updateClusteredPool = async (
...
```
edlerd: Similar to the above comment in the same file, I think this can be simplified.

edlerd (commented):
Thanks for the iteration on this one. It looks pretty good, with some small issues below. 

Can you please test this with a clustered backend to ensure it works well in the complex case? (Manual testing for now is fine, for adding it to CI, we should file a dedicated ticket to do in the future.).

src/api/storage-pools.tsx
kimanhou: Do we need to store it in a variable ? We just want to throw the first error we find, right ?

```
  const rejectedOperation = operations.find((res) => res.status === "rejected");
  if (rejectedOperation) {
    throw rejectedOperation.reason;
  }
```

edlerd: The `throw` in line 180 of the suggestion above does the same thing. It stops the map from continuing the iteration and bubbles up to the parent scope.

edlerd: So in case of throw, we don't need the `pendingOperations`, we only need it in the success case as then the function continues and we need the `pendingOperations` below.

edlerd (commented):
The changes need to be scoped with the API extension. Storage pool creation should stay compatible with the older LXD versions, that do not respond with the operation. But do the storage pool creation synchronously.

edlerd (approved):
LGTM

PR #1907, opened 2026-03-30
feat(ProfileSelector): Move helptext in Tooltip

kimanhou:
## Done

- When creating/editing an instance, in the Profiles section: move helptext from under the dropdown list to an icon next to the label

## QA

1. Run the LXD-UI:
    - Create an instance
    - Make sure the Profiles label has an info icon (similar to Target and SSH keys). When hover on the icon, a help text should appear: `Profiles store a set of configuration options, such as instance and device options.`
    - Select more than one profile, make sure the helptext is updated correctly to:
    ```
    Profiles store a set of configuration options, such as instance and device options.
    Each profile overrides the settings specified in previous profiles.
    ```

## Screenshots

Screenshot from 2026-03-27 13-16-16
Screenshot from 2026-03-27 13-16-26

### DIFF
--- a/src/pages/profiles/ProfileSelector.tsx
+++ b/src/pages/profiles/ProfileSelector.tsx
@@ -7,6 +7,7 @@ import {
   Select,
   useNotify,
   Spinner,
+  Tooltip,
 } from "@canonical/react-components";
 import { defaultFirst } from "util/helpers";
 import { useProfiles } from "context/useProfiles";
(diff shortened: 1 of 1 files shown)


src/pages/profiles/ProfileSelector.tsx
```diff
@@ -69,34 +70,39 @@ const ProfileSelector: FC<Props> = ({
     }
   };
 
-  const getHelp = (index: number) => {
-    const profileIntro =
+  const getHelp = () => {
+    const intro =
```
omarelkashef: I think it would be better to always have the full message. Users will more likely look at the helper text before adding a new profile and might not look again. That means they might miss the important detail that later profiles override earlier ones.

omarelkashef (commented):
Thank you for this PR. I left one comment below.

src/pages/profiles/ProfileSelector.tsx
kimanhou: I mimicked the current behavior. But I think you're right, it's different from the helptext as the update won't be visible to the user once they select more than one profile. That simplifies the code.

omarelkashef (approved):
Code and QA look good. Thank you!

PR #1919, opened 2026-04-03
feat: image registries list page

omarelkashef:
## Done

- Add a new page for listing image registries

## QA

1. Run the LXD-UI:
    - This needs to run against a version lxd with image registries support. For now, you can run against lxd of the following branch https://github.com/canonical/lxd/pull/17680
    - Go to Images -> Registries
    - Registry creation is currently not available in the UI. Instead, you can run `sudo lxc image registry create ubuntu1 --protocol simplestreams url=https://ubuntu.com public=true` to create a new public registry pointing to ubuntu.com 
   
    - Explore the page, especially the following components:
       - Search and filter 
       - Table layout
       - Rows selection

## Screenshots
image
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -95,6 +95,10 @@ const StoragePoolDetail = lazy(
 const CreateStorageVolume = lazy(
   async () => import("pages/storage/forms/CreateStorageVolume"),
 );
+const ImageRegistriesList = lazy(
+  async () => import("pages/images/ImageRegistriesList"),
+);
+
 const StorageVolumeDetail = lazy(
   async () => import("pages/storage/StorageVolumeDetail"),
 );
@@ -508,6 +512,10 @@ const App: FC = () => {
         <Route
           path={`${ROOT_PATH}/ui/project/:project/local-images`}
           element={<ProtectedRoute outlet={<LocalImageList />} />}
+        />
+        <Route
+          path={`${ROOT_PATH}/ui/image-registries`}
+          element={<ProtectedRoute outlet={<ImageRegistriesList />} />}
         />
         <Route
           path={`${ROOT_PATH}/ui/server`}
--- a/src/api/image-registries.tsx
+++ b/src/api/image-registries.tsx
@@ -1,0 +1,36 @@
+import { handleResponse } from "util/helpers";
+import type { LxdImageRegistry } from "types/image";
+import type { LxdApiResponse } from "types/apiResponse";
+import { addEntitlements } from "util/entitlements/api";
+import { ROOT_PATH } from "util/rootPath";
+
+const imageRegistryEntitlements = ["can_edit", "can_delete"];
+
+export const fetchImageRegistries = async (
+  isFineGrained: boolean | null,
+): Promise<LxdImageRegistry[]> => {
+  const params = new URLSearchParams();
+  params.set("recursion", "1");
+  addEntitlements(params, isFineGrained, imageRegistryEntitlements);
+  return fetch(`${ROOT_PATH}/1.0/image-registries?${params.toString()}`)
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdImageRegistry[]>) => {
+      return data.metadata;
+    });
+};
+
+export const fetchImageRegistry = async (
+  name: string,
+  isFineGrained: boolean | null,
+): Promise<LxdImageRegistry> => {
+  const params = new URLSearchParams();
+  params.set("recursion", "1");
+  addEntitlements(params, isFineGrained, imageRegistryEntitlements);
+  return fetch(
+    `${ROOT_PATH}/1.0/image-registries/${encodeURIComponent(name)}?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdImageRegistry>) => {
+      return data.metadata;
+    });
+};
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -44,6 +44,10 @@ const initialiseOpenNavMenus = (location: Location) => {
   const openCluster =
     location.pathname.includes("/cluster/") ||
     location.pathname.includes("/placement-groups");
+  const openImages =
+    location.pathname.includes("/image-registries") ||
+    location.pathname.includes("/local-images");
+
   const initialOpenMenus: AccordionNavMenu[] = [];
   if (openPermissions) {
     initialOpenMenus.push("permissions");
@@ -59,6 +63,10 @@ const initialiseOpenNavMenus = (location: Location) => {
 
   if (openCluster) {
     initialOpenMenus.push("clustering");
+  }
+
+  if (openImages) {
+    initialOpenMenus.push("images");
   }
 
   return initialOpenMenus;
@@ -97,7 +105,8 @@ const Navigation: FC = () => {
     initializeProjectName(isAllProjectsFromUrl, isLoading, project),
   );
   const isAllProjects = projectName === ALL_PROJECTS;
-  const { hasCustomVolumeIso, hasAccessManagement } = useSupportedFeatures();
+  const { hasCustomVolumeIso, hasAccessManagement, hasImageRegistries } =
+    useSupportedFeatures();
   const { loggedInUserName, loggedInUserID } = useLoggedInUser();
   const [scroll, setScroll] = useState(false);
   const location = useLocation();
@@ -459,10 +468,10 @@ const Navigation: FC = () => {
                       <SideNavigationItem>
                         <NavAccordion
                           baseUrls={[
-                            `${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/image`,
+                            `${ROOT_PATH}/ui/image-registries`,
+                            `${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/local-images`,
                           ]}
                           title={getNavTitle("image")}
-                          disabled={isAllProjects}
                           iconName="image"
                           label="Images"
                           onOpen={() => {
@@ -480,6 +489,15 @@ const Navigation: FC = () => {
                           >
                             Local images
                           </NavLink>
+                          {hasImageRegistries && (
+                            <NavLink
+                              to={`${ROOT_PATH}/ui/image-registries`}
+                              title={getNavTitle("image registries")}
+                              onClick={softToggleMenu}
+                            >
+                              Image registries
+                            </NavLink>
+                          )}
                         </NavAccordion>
                       </SideNavigationItem>
                       <SideNavigationItem>
--- a/src/context/useImageRegistries.tsx
+++ b/src/context/useImageRegistries.tsx
@@ -1,0 +1,22 @@
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { useAuth } from "./auth";
+import { fetchImageRegistries, fetchImageRegistry } from "api/image-registries";
+
+export const useImageRegistries = (isEnabled = true) => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.imageRegistries],
+    queryFn: async () => fetchImageRegistries(isFineGrained),
+    enabled: isEnabled && isFineGrained !== null,
+  });
+};
+
+export const useImageRegistry = (name: string, isEnabled = true) => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.imageRegistries, name],
+    queryFn: async () => fetchImageRegistry(name, isFineGrained),
+    enabled: isEnabled && isFineGrained !== null,
+  });
+};
--- a/src/context/useSupportedFeatures.tsx
+++ b/src/context/useSupportedFeatures.tsx
@@ -53,5 +53,6 @@ export const useSupportedFeatures = () => {
     hasStorageAndNetworkOperations: apiExtensions.has(
       "storage_and_network_operations",
     ),
+    hasImageRegistries: apiExtensions.has("image_registries"),
   };
 };
--- a/src/pages/images/ImageRegistriesList.tsx
+++ b/src/pages/images/ImageRegistriesList.tsx
@@ -1,0 +1,197 @@
+import { type FC } from "react";
+import {
+  Row,
+  ScrollableTable,
+  TablePagination,
+  useNotify,
+  CustomLayout,
+  Spinner,
+  MainTable,
+} from "@canonical/react-components";
+import useSortTableData from "util/useSortTableData";
+import { useImageRegistries } from "context/useImageRegistries";
+import HelpLink from "components/HelpLink";
+import NotificationRow from "components/NotificationRow";
+import PageHeader from "components/PageHeader";
+import ImageRegistriesSearchFilter, {
+  type ImageRegistryFilter,
+} from "./ImageRegistriesSearchFilter";
+import { useSearchParams } from "react-router-dom";
+import type { LxdImageRegistryProtocol } from "types/image";
+import { isRegistryPublic } from "util/image-registries";
+import { CreateImageRegistryButton } from "./actions/CreateImageRegistryButton";
+
+const ImageRegistriesList: FC = () => {
+  const notify = useNotify();
+  const [searchParams] = useSearchParams();
+  const { data: imageRegistries = [], error, isLoading } = useImageRegistries();
+
+  if (error) {
+    notify.failure("Loading image registries failed", error);
+  }
+
+  const headers = [
+    { content: "Name", sortKey: "name" },
+    { content: "Description", sortKey: "description" },
+    {
+      content: "Protocol",
+      sortKey: "protocol",
+    },
+    {
+      content: "Built-in",
+      sortKey: "builtin",
+    },
+    {
+      content: "Public",
+      sortKey: "public",
+    },
+  ];
+
+  const filters: ImageRegistryFilter = {
+    queries: searchParams.getAll("query").map((value) => value.toLowerCase()),
+    protocol: searchParams.getAll("protocol") as LxdImageRegistryProtocol[],
+    builtin: searchParams
+      .getAll("builtin")
+      .map((value) => value.toLowerCase() === "yes"),
+    public: searchParams
+      .getAll("public")
+      .map((value) => value.toLowerCase() === "yes"),
+  };
+
+  const filteredImageRegistries = imageRegistries.filter(
+    (item) =>
+      (!filters.queries.length ||
+        filters.queries.some(
+          (query) =>
+            (item?.description ?? "").toLowerCase().includes(query) ||
+            item.name.toLowerCase().includes(query),
+        )) &&
+      (!filters.protocol.length || filters.protocol.includes(item.protocol)) &&
+      (!filters.builtin.length || filters.builtin.includes(item.builtin)) &&
+      (!filters.public.length ||
+        filters.public.includes(isRegistryPublic(item))),
+  );
+
+  const rows = filteredImageRegistries.map((registry) => {
+    return {
+      key: registry.name,
+      name: registry.name,
+      columns: [
+        {
+          content: registry.name,
+          role: "rowheader",
+          "aria-label": "Name",
+          title: `Image registry ${registry.name}`,
+        },
+        {
+          content: (
+            <div className="u-truncate" title={registry.description}>
+              {registry.description}
+            </div>
+          ),
+          role: "cell",
+          "aria-label": "Description",
+        },
+        {
+          content: registry.protocol,
+          role: "cell",
+          "aria-label": "Protocol",
+        },
+        {
+          content: registry.builtin ? "Yes" : "No",
+          role: "cell",
+          "aria-label": "Built-in",
+        },
+        {
+          content: isRegistryPublic(registry) ? "Yes" : "No",
+          role: "cell",
+          "aria-label": "Public",
+        },
+      ],
+      sortData: {
+        name: registry.name.toLowerCase(),
+        description: registry.description.toLowerCase(),
+        protocol: registry.protocol.toLowerCase(),
+        builtin: registry.builtin,
+        public: isRegistryPublic(registry),
+      },
+    };
+  });
+  const { rows: sortedRows, updateSort } = useSortTableData({ rows });
+  if (isLoading) {
+    return <Spinner className="u-loader" text="Loading..." isMainComponent />;
+  }
+
+  const getTablePaginationDescription = () => {
+    // This is needed because TablePagination does not support the plural form of registry
+    if (rows.length === 0) {
+      return "Showing 0 image registries";
+    }
+    const defaultPaginationDescription =
+      rows.length > 1
+        ? `Showing all ${rows.length} image registries`
+        : `Showing 1 out of 1 image registry`;
+
+    return defaultPaginationDescription;
+  };
+
+  return (
+    <>
+      <CustomLayout
+        contentClassName="u-no-padding--bottom"
+        header={
+          <PageHeader>
+            <PageHeader.Left>
+              <PageHeader.Title>
+                <HelpLink
+                  docPath="/image-handling/"
+                  title="Learn more about image registries"
+                >
+                  Image registries
+                </HelpLink>
+              </PageHeader.Title>
+              {imageRegistries.length > 0 && (
+                <PageHeader.Search>
+                  <ImageRegistriesSearchFilter />
+                </PageHeader.Search>
+              )}
+            </PageHeader.Left>
+            <PageHeader.BaseActions>
+              <CreateImageRegistryButton />
+            </PageHeader.BaseActions>
+          </PageHeader>
+        }
+      >
+        <NotificationRow />
+        <Row>
+          <ScrollableTable
+            dependencies={[imageRegistries]}
+            tableId="image-registries-table"
+            belowIds={["status-bar"]}
+          >
+            <TablePagination
+              data={sortedRows}
+              id="pagination"
+              className="u-no-margin--top"
+              itemName="registry"
+              aria-label="Table pagination control"
+              description={getTablePaginationDescription()}
+            >
+              <MainTable
+                id="image-registries-table"
+                defaultSort="Name"
+                headers={headers}
+                rows={rows}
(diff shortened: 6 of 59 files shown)


src/api/images.tsx
```diff
@@ -12,6 +12,7 @@ import { addEntitlements } from "util/entitlements/api";
 import { ROOT_PATH } from "util/rootPath";
 
 const imageEntitlements = ["can_delete"];
+const imageRegistryEntitlements = ["can_edit", "can_delete"];
```
edlerd: Let's move the registry endpoints to a dedicated file please.

src/context/useImages.tsx
```diff
@@ -30,3 +33,25 @@ export const useImagesInAllProjects = (
     enabled: (enabled ?? true) && isFineGrained !== null,
   });
 };
+
+export const useImageRegistries = (enabled?: boolean) => {
```
edlerd: Let's move these into a new file `useImageRegistries.tsx`.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,315 @@
...
+  type ImageRegistryFilter,
+} from "./ImageRegistriesSearchFilter";
+import { useSearchParams } from "react-router-dom";
+import type { LxdImageRegistry, LxdImageRegistryProtocol } from "types/image";
+import usePanelParams, { panels } from "util/usePanelParams";
+import ImageRegistryDetailPanel from "./ImageRegistryDetailPanel";
+
+const isRegistryPublic = (registry: LxdImageRegistry): boolean => {
+  return registry.config?.public?.toLowerCase() === "true";
+};
```
edlerd: Let's move this to a dedicated util file.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,315 @@
...
+                  name="Delete"
+                  disabled={deletableImageRegistries.length === 0}
+                  hasIcon
+                  className="u-float-right u-no-margin--bottom"
+                  title="Delete image registries"
+                >
+                  <Icon name="delete" className="u-margin--right" />
+                  <span>Delete</span>
+                </Button>
+              )}
```
edlerd: This seems to do nothing. Let's build this without any bulk actions. I think most users will only have a handful of custom registries, or even none. So we don't need to support bulk actions from the start and can keep this a bit simpler.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,315 @@
...
+                name="Create registry"
+                disabled={!canCreateImageRegistries()}
+                hasIcon
+                appearance="positive"
+                className="u-float-right u-no-margin--bottom"
+                title="Create registry"
+              >
+                <Icon name="plus" className="u-margin--right" />
+                <span> Create registry</span>
+              </Button>
```
edlerd: Please extract into a `CreateImageRegistryBtn`.

src/sass/_image_registries.scss
```diff
@@ -0,0 +1,4 @@
+.u-row:hover,
+.u-row-selected {
+  background-color: $colors--theme--background-hover;
+}
```
edlerd: This should be namespaced for an `.image-registries` class.

src/types/image.d.ts
```diff
@@ -32,6 +33,21 @@ export interface LxdImage {
...
+  name: string;
+  description: string;
+  protocol: LxdImageRegistryProtocol;
+  builtin: boolean;
+  config?: {
+    url: string;
+    public: string;
+    cluster: string;
+    source_project: string;
+    [key: string]: string;
```
edlerd: Can we avoid the generic here and be more specific?
```suggestion
```

tests/helpers/images.ts
```diff
@@ -40,3 +40,13 @@ export const deleteImage = async (page: Page, imageIdentifier: string) => {
 
   await page.waitForSelector(`text=Image ${imageName} deleted.`);
 };
+
+export const visitImageRegistries = async (page: Page, project: string) => {
+  await gotoURL(page, `/ui/project/${project}/image-registries`);
+  await expect(page.getByTitle("Create image registry")).toBeDisabled();
```
edlerd: This seems like an unexpected side effect. Why expect the button to be disabled when visiting the page?

tests/image-registries.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { expect, test } from "./fixtures/lxd-test";
+import { visitImageRegistries } from "./helpers/images";
+import { gotoURL } from "./helpers/navigate";
+
+test("search for an image registry", async ({ page }) => {
+  await gotoURL(page, "/ui");
+  await page.getByRole("button", { name: "Images" }).click();
+  await page.getByRole("link", { name: "Registries", exact: true }).click();
+  await expect(page.getByTitle("Create registry")).toBeDisabled();
```
edlerd:
```suggestion
```

src/App.tsx
```diff
@@ -21,6 +21,7 @@ import {
 import { setFavicon } from "util/favicon";
 import { ALL_PROJECTS } from "util/loginProject";
 import { AUTH_METHOD } from "util/authentication";
+import { ImageRegistriesList } from "pages/images/ImageRegistriesList";
```
edlerd: This should be lazy loaded.

src/pages/images/ImageRegistryDetailPanel.tsx
```diff
@@ -0,0 +1,103 @@
...
+  Button,
+  Icon,
+  SidePanel,
+  useNotify,
+} from "@canonical/react-components";
+import usePanelParams from "util/usePanelParams";
+
+import { useImageRegistry } from "context/useImages";
+
+const ImageRegistryDetailPanel: FC = () => {
```
edlerd: I think we wanted a detail page, not a detail panel. Because we also need to list the images in the registry.

edlerd (commented):
Good start on the registries! Read through the changes and a few requests can be found below.

src/types/image.d.ts
omarelkashef: Lxd allows users to store key/value pairs in the registry. Some of them are used for configuration such as source_project, url, etc, but they can also be free-form user.x key/value pairs. Is the plan to support these user keys?

edlerd: The user keys exist on all entities. I wouldn't support them right away. For the others, maybe it's best to list the ones we are using in the UI as optional string fields.

src/pages/images/ImageRegistryDetailPanel.tsx
omarelkashef: The registry detail page has its own ticket which will come after this. Do we also want a detail panel or a detail page suffices?

edlerd: The detail page should be sufficient. No need for a side panel.

tests/helpers/images.ts
omarelkashef: This was left by mistake. At first, this PR was going to ignore the user permissions to create a registry, so the button would always be disabled and this the support for this permissions would be introduced in another PR.

src/components/Navigation.tsx
```diff
@@ -473,13 +479,22 @@ const Navigation: FC = () => {
...
                           >
                             Local images
                           </NavLink>
+                          {hasImageRegistry && (
+                            <NavLink
+                              to={`${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/image-registries`}
+                              title={getNavTitle("image registries")}
+                              onClick={softToggleMenu}
+                            >
+                              Registries
```
edlerd:
```suggestion
                              Image registries
```
Shall we use the full name here?

src/context/useImageRegistries.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useServerEntitlements } from "util/entitlements/server";
+
+export const useImageRegistries = (enabled?: boolean) => {
+  const { isFineGrained } = useAuth();
+  const { canViewImageRegistries } = useServerEntitlements();
+  return useQuery({
+    queryKey: [queryKeys.imageRegistries],
+    queryFn: async () => fetchImageRegistries(isFineGrained),
+    enabled:
+      canViewImageRegistries() && (enabled ?? true) && isFineGrained !== null,
```
edlerd:
```suggestion
  return useQuery({
    queryKey: [queryKeys.imageRegistries],
    queryFn: async () => fetchImageRegistries(isFineGrained),
    enabled:
      isEnabled && isFineGrained !== null,
```
1. No need to check the view entitlements. Those are implicit by the response.
2. Let's give the `enabled` a prefix and a default value.

src/context/useImageRegistries.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const useImageRegistry = (name: string, enabled?: boolean) => {
+  const { isFineGrained } = useAuth();
+  const { canViewImageRegistries } = useServerEntitlements();
+  return useQuery({
+    queryKey: [queryKeys.imageRegistries, name],
+    queryFn: async () => fetchImageRegistry(name, isFineGrained),
+    enabled:
+      canViewImageRegistries() && (enabled ?? true) && isFineGrained !== null,
```
edlerd:
```suggestion
  return useQuery({
    queryKey: [queryKeys.imageRegistries, name],
    queryFn: async () => fetchImageRegistry(name, isFineGrained),
    enabled:
      enabled && isFineGrained !== null,
```
Same as above.

src/context/useImages.tsx
```diff
@@ -6,6 +6,7 @@ import {
   fetchImagesInAllProjects,
   fetchLocalImagesInProject,
 } from "api/images";
+
```
edlerd:
```suggestion
```

src/context/useSupportedFeatures.tsx
```diff
@@ -53,5 +53,6 @@ export const useSupportedFeatures = () => {
     hasStorageAndNetworkOperations: apiExtensions.has(
       "storage_and_network_operations",
     ),
+    hasImageRegistry: apiExtensions.has("image_registries"),
```
edlerd:
```suggestion
    hasImageRegistries: apiExtensions.has("image_registries"),
```

src/pages/images/actions/CreateImageRegistryButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <Button
+      name="Create registry"
+      disabled={!canCreateImageRegistries()}
+      hasIcon
+      appearance="positive"
+      className="u-float-right u-no-margin--bottom"
+      title="Create registry"
+    >
+      <Icon name="plus" className="u-margin--right" />
+      <span> Create registry</span>
```
edlerd:
```suggestion
      <span>Create registry</span>
```

src/pages/images/actions/CreateImageRegistryButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const CreateImageRegistryButton: FC = () => {
+  const { canCreateImageRegistries } = useServerEntitlements();
+  return (
+    <Button
+      name="Create registry"
+      disabled={!canCreateImageRegistries()}
+      hasIcon
+      appearance="positive"
+      className="u-float-right u-no-margin--bottom"
+      title="Create registry"
```
edlerd: The title should be empty or have a description that permissions are missing if it is disabled. No need to put "Create registry" as that is already the label.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+import { isRegistryPublic } from "util/image-registries";
+import { CreateImageRegistryButton } from "./actions/CreateImageRegistryButton";
+
+const ImageRegistriesList: FC = () => {
+  const notify = useNotify();
+  const [selectedNames, setSelectedNames] = useState<string[]>([]);
+  const [searchParams] = useSearchParams();
+
+  const { canCreateImageRegistries } = useServerEntitlements();
+
```
edlerd:
```suggestion
```

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+  type ImageRegistryFilter,
+} from "./ImageRegistriesSearchFilter";
+import { useSearchParams } from "react-router-dom";
+import type { LxdImageRegistryProtocol } from "types/image";
+import { isRegistryPublic } from "util/image-registries";
+import { CreateImageRegistryButton } from "./actions/CreateImageRegistryButton";
+
+const ImageRegistriesList: FC = () => {
+  const notify = useNotify();
+  const [selectedNames, setSelectedNames] = useState<string[]>([]);
```
edlerd:
```suggestion
```

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+  useEffect(() => {
+    const validNames = new Set(imageRegistries?.map((image) => image.name));
+    const validSelections = selectedNames.filter((name) =>
+      validNames.has(name),
+    );
+    if (validSelections.length !== selectedNames.length) {
+      setSelectedNames(validSelections);
+    }
+  }, [imageRegistries]);
+
```
edlerd:
```suggestion
```

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+      (!filters.protocol.length || filters.protocol.includes(item.protocol)) &&
+      (!filters.builtin.length || filters.builtin.includes(item.builtin)) &&
+      (!filters.public.length ||
+        filters.public.includes(isRegistryPublic(item))),
+  );
+
+  const selectedImageRegistries = imageRegistries.filter((registry) =>
+    selectedNames.includes(registry.name),
+  );
+
```
edlerd:
```suggestion
```

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+        <SelectedTableNotification
+          totalCount={filteredImageRegistries.length ?? 0}
+          itemName="registry"
+          selectedNames={selectedNames}
+          setSelectedNames={setSelectedNames}
+          filteredNames={filteredImageRegistries.map((item) => item.name)}
+        />
+      );
+    }
+
```
edlerd:
```suggestion
```

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+    return {
+      key: registry.name,
+      name: registry.name,
+      columns: [
+        {
+          content: registry.name,
+          role: "rowheader",
+          "aria-label": "Name",
+          title: `Image registry ${registry.name}`,
+          className: "clickable-cell",
```
edlerd:
```suggestion
```
None of the other cells should be clickable either.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+  );
+
+  const selectedImageRegistries = imageRegistries.filter((registry) =>
+    selectedNames.includes(registry.name),
+  );
+
+  const rows = filteredImageRegistries.map((registry) => {
+    const actions = (
+      <List inline className="actions-list u-no-margin--bottom" items={[]} />
+    );
```
edlerd:
```suggestion
```
Let's limit the actions to the detail page.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+          role: "cell",
+          "aria-label": "Actions",
+        },
+      ],
+      sortData: {
+        name: registry.name.toLowerCase(),
+        description: registry.description.toLowerCase(),
+        protocol: registry.protocol.toLowerCase(),
+        builtin: registry.builtin,
+      },
```
edlerd, line 117: public is missing from the list.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+  return (
+    <>
+      <CustomLayout
+        contentClassName="u-no-padding--bottom"
+        header={
+          <PageHeader>
+            <PageHeader.Left>
+              <PageHeader.Title>
+                <HelpLink
+                  docPath="/image-handling/"
```
edlerd, line 147: Do we have a better entry page or maybe an anchor for this?

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+                  title="Learn more about image registries"
+                >
+                  Image registries
+                </HelpLink>
+              </PageHeader.Title>
+              {selectedNames.length === 0 && imageRegistries.length > 0 && (
```
edlerd:
```suggestion
              {imageRegistries.length > 0 && (
```

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+              title="No image registries found"
+            >
+              <p>
+                There are no image registries.
+                {canCreateImageRegistries()
+                  ? " Create your first image registry!"
+                  : ""}
+              </p>
+            </EmptyState>
+          )}
```
edlerd: Is this even possible? I think the built in registries will always be available, no?

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+          {imageRegistries.length > 0 && (
+            <ScrollableTable
+              dependencies={[imageRegistries]}
```
edlerd: Should this depend on the filtered list instead?

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,268 @@
...
+            >
+              <TablePagination
+                data={sortedRows}
+                id="pagination"
+                className="u-no-margin--top"
+                itemName="registry"
+                aria-label="Table pagination control"
+                description={getTablePaginationDescription()}
+              >
+                <SelectableMainTable
```
edlerd: Let's use a non-selectable table.

src/sass/_image_registries.scss
```diff
@@ -0,0 +1,6 @@
+.image-registries {
+  .u-row:hover,
+  .u-row-selected {
```
edlerd:
```suggestion
   {
```

src/sass/_image_registry_detail_panel.scss
edlerd, line 1: I think this file can be removed.

src/util/usePanelParams.tsx
```diff
@@ -129,7 +129,6 @@ const usePanelParams = (): PanelHelper => {
     target: params.get("target") ?? "",
     deviceName: params.get("device-name"),
     localPeering: params.get("local-peering"),
-
```
edlerd, line 132: Let's restore the empty line here.

src/api/image-registries.tsx
```diff
@@ -0,0 +1,35 @@
+import { handleResponse } from "util/helpers";
+import type { LxdImageRegistry } from "types/image";
+import type { LxdApiResponse } from "types/apiResponse";
+const imageRegistryEntitlements = ["can_edit", "can_delete"];
```
edlerd: Let's move this outside of the import block.

edlerd (commented):
Thanks for the iteration. Did a thorough code review; many smallish things below.

src/pages/images/ImageRegistriesList.tsx
omarelkashef: If it depended on the filtered list, then if the user chooses a combination of filters than returns 0 registries, the table (including the headers) will be gone. But now, the table still exists but it shows zero data which is neater.

src/pages/images/actions/CreateImageRegistryButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+      name="Create registry"
+      disabled={isDisabled}
+      hasIcon
+      appearance="positive"
+      className="u-float-right u-no-margin--bottom"
+      title={
+        isDisabled && "You don't have permissions to create image registries"
+      }
+    >
+      <Icon name="plus" className="u-margin--right" />
```
edlerd:
```suggestion
      <Icon name="plus" light className="u-margin--right" />
```

src/util/entitlements/server.tsx
```diff
@@ -67,10 +67,26 @@ export const useServerEntitlements = () => {
...
+      "can_create_image_registries",
+      serverEntitlements,
+    ) || hasEntitlement(isFineGrained, "admin", serverEntitlements);
+
+  const canViewImageRegistries = () =>
+    hasEntitlement(
+      isFineGrained,
+      "can_view_image_registries",
+      serverEntitlements,
+    ) || hasEntitlement(isFineGrained, "admin", serverEntitlements);
```
edlerd:
```suggestion
```
The view permission is implicit.

src/sass/_image_registries.scss
```diff
@@ -0,0 +1,5 @@
+.image-registries {
+  .u-row:hover {
+    background-color: $colors--theme--background-hover;
+  }
+}
```
edlerd:
```suggestion
```
We probably don't need this, at least on other tables we don't use it except on instances and profiles.

src/App.tsx
```diff
@@ -506,9 +510,13 @@ const App: FC = () => {
           element={<ProtectedRoute outlet={<StorageBucketDetail />} />}
         />
         <Route
-          path={`${ROOT_PATH}/ui/project/:project/local-images`}
+          path={`${ROOT_PATH}/ui/project/:project/images-local`}
```
edlerd: Let's keep the speaking URL. The new one is a bit weird, in my opinion. We can have two cases in the navigation to handle the initial open and a second entry on the images entry as a base URL to allow this.

edlerd (commented):
Code mostly LGTM, 52 of 61 files are ok 👍 
Waiting for upstream to resolve conflicts to do a QA with the branch.

src/components/Navigation.tsx
```diff
@@ -462,7 +471,6 @@ const Navigation: FC = () => {
                             `${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/image`,
```
edlerd: We need to extend this to include the `.../local-images` path. So when collapsing this on the local images page, the parent becomes active.

src/pages/images/actions/CreateImageRegistryButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+      name="Create registry"
+      disabled={isDisabled}
+      hasIcon
+      appearance="positive"
+      className="u-float-right u-no-margin--bottom"
+      title={
+        isDisabled && "You don't have permissions to create image registries"
+      }
+    >
+      <Icon name="plus" className="light u-margin--right" />
```
edlerd:
```suggestion
      <Icon name="plus" light className="u-margin--right" />
```
This should be a property, not a class name

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const [searchParams] = useSearchParams();
+  const { data: imageRegistries = [], error, isLoading } = useImageRegistries();
+
+  if (error) {
+    notify.failure("Loading image registries failed", error);
+  }
+
+  const headers = [
+    { content: "Name", sortKey: "name" },
+    { content: "Description" },
```
edlerd: sort key is missing for description.

src/App.tsx
```diff
@@ -509,6 +513,10 @@ const App: FC = () => {
           path={`${ROOT_PATH}/ui/project/:project/local-images`}
           element={<ProtectedRoute outlet={<LocalImageList />} />}
         />
+        <Route
+          path={`${ROOT_PATH}/ui/project/:project/image-registries`}
```
```suggestion
          path={`${ROOT_PATH}/ui/image-registries`}
```
I think the path should not be under a project here. Also for the detail page. Wdyt?

FYI @kimanhou

tests/helpers/images.ts
```diff
@@ -40,3 +40,13 @@ export const deleteImage = async (page: Page, imageIdentifier: string) => {
...
+
+export const visitImageRegistries = async (page: Page, project: string) => {
+  await gotoURL(page, `/ui/project/${project}/image-registries`);
+  await expect(page.getByTitle("Create image registry")).toBeVisible();
+};
+
+export const selectAllRegistries = async (page: Page) => {
+  await page.getByLabel("multiselect rows").first().click();
+  await page.getByRole("menuitem", { name: "Select all" }).click();
+};
```
edlerd:
```suggestion
```

edlerd (commented):
Did a full QA and some more code review. Small issues left, then this should be good to merge.

src/pages/images/ImageRegistriesList.tsx
omarelkashef: Is it common to sort by description?

src/App.tsx
omarelkashef: It makes sense conceptually, but it will break the consistency since all the sections around it have project prefix in the url and the ones the do not are all the way down in the nav menu and are separated by a horizontal line.
image

src/components/Navigation.tsx
omarelkashef: Yes, good catch!

src/App.tsx
edlerd: I think most users don't get that connection of "above the line is project specific" anyway. I am not 100% confident about this, though. Furthermore, I lean to remove the project here as it is technically more correct and supports the user to learn registries as a global (non-project-specific) entity.

omarelkashef: In that case, should we separate local images from registries as local images are project scoped but not registries?

edlerd: As discussed, from the main navigation I'd keep them together as not to grow the total number of items in the main nav.

src/pages/images/ImageRegistriesList.tsx
edlerd: We should support it if we can.

src/App.tsx
omarelkashef: Agreed!

edlerd (dismissed):
LGTM, good stuff 👍

edlerd (approved):

PR #1927, opened 2026-04-07
fix: adjust operations api usage WD-35708

omarelkashef:
## Done

- Fix the usage of operations API after https://github.com/canonical/lxd/pull/18028

## QA

1. Run the LXD-UI:
    - Make sure operations such as renaming instance, deleting instance, starting instance from snapshot, creating volume snapshot work, etc. correctly and the notifications are accurate

### DIFF
--- a/src/types/operation.d.ts
+++ b/src/types/operation.d.ts
@@ -20,6 +20,8 @@ export interface LxdOperation {
     instances_snapshots?: string[];
     storage_volume_snapshots?: string[];
   };
+  entity_url: string;
+  original_entity_url: string;
   status: LxdOperationStatus;
   status_code: number;
   updated_at: string;
--- a/src/util/operations.spec.ts
+++ b/src/util/operations.spec.ts
@@ -6,12 +6,16 @@ import {
 } from "./operations";
 import type { LxdOperation } from "types/operation";
 
-const craftOperation = (...url: string[]) => {
+const craftOperation = (
+  resourceUrls: string[] = [],
+  entityUrl?: string,
+  originalEntityUrl?: string,
+) => {
   const images: string[] = [];
   const instances: string[] = [];
   const instances_snapshots: string[] = [];
   const storage_volume_snapshots: string[] = [];
-  for (const u of url) {
+  for (const u of resourceUrls) {
     const segments = u.split("/");
     if (u.includes("snapshots") && u.includes("storage-pools")) {
       storage_volume_snapshots.push(u);
@@ -34,20 +38,30 @@ const craftOperation = (...url: string[]) => {
       instances.push(u);
     }
   }
-
   return {
+    metadata: {
+      entity_url: entityUrl ? entityUrl : undefined,
+      original_entity_url: originalEntityUrl ? originalEntityUrl : undefined,
+    },
     resources: {
       images,
       instances,
       instances_snapshots,
       storage_volume_snapshots,
     },
-  } as LxdOperation;
+  } as unknown as LxdOperation;
 };
 
 describe("getInstanceName", () => {
-  it("identifies instance name from an instance operation", () => {
-    const operation = craftOperation("/1.0/instances/testInstance1");
+  it("identifies instance name from an instance operation using resources field", () => {
+    const operation = craftOperation(["/1.0/instances/testInstance1"]);
+    const name = getInstanceName(operation);
+
+    expect(name).toBe("testInstance1");
+  });
+
+  it("identifies instance name from an instance operation using entity_url field", () => {
+    const operation = craftOperation([], "/1.0/instances/testInstance1");
     const name = getInstanceName(operation);
 
     expect(name).toBe("testInstance1");
@@ -55,6 +69,7 @@ describe("getInstanceName", () => {
 
   it("identifies instance name from an instance operation in a custom project", () => {
     const operation = craftOperation(
+      [],
       "/1.0/instances/testInstance2?project=project",
     );
     const name = getInstanceName(operation);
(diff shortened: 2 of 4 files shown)


src/pages/operations/OperationInstanceName.tsx
```diff
@@ -38,7 +38,13 @@ const OperationInstanceName: FC<Props> = ({ operation }) => {
 
   if (isLinkable && projectName) {
     return (
-      <InstanceRichChip instanceName={instanceName} projectName={projectName} />
+      <>
```
kimanhou: Is this needed ?

omarelkashef: It is necessary to add a space between the action and the instance name.

kimanhou: Okay but then it looks weird

with a space
Screenshot from 2026-04-07 15-45-04

without a space
Screenshot from 2026-04-07 15-46-12

Kxiru: I agree that no space looks nicer if the action column is always stacked like that, but if it gets long enough for the action and entity name to share a line, then the space is necessary. What is the likelihood of the latter?

kimanhou: In which case the action column is not stacked like that ?

omarelkashef: Ok, I agree. It looks fine with no space because the chip has padding.

src/util/operations.spec.ts
```diff
@@ -64,7 +31,6 @@ describe("getInstanceName", () => {
 
   it("identifies instance name from an instance creation operation with snapshot as source", () => {
     const operation = craftOperation(
-      "/1.0/instances/targetInstanceName",
```
kimanhou, line 67: How can `getInstanceName` find `targetInstanceName` when `targetInstanceName` doesn't appear on the operation ?

src/types/operation.d.ts
```diff
@@ -13,13 +13,7 @@ export interface LxdOperation {
...
   metadata?: Record<string, string>;
   may_cancel: boolean;
-  resources?: {
-    images?: string[];
-    instance?: string[];
-    instances?: string[];
-    instances_snapshots?: string[];
-    storage_volume_snapshots?: string[];
-  };
+  entity_url: string;
```
kimanhou, line 23: Wasn't `entity_url` in `metadata` ?

omarelkashef: Now it is its own field https://github.com/canonical/lxd/pull/18028/changes

kimanhou: Okay but then does it still exist in metadata ? 

```
if (operation?.metadata?.["entity_url"]) {
    candidates.push(operation.metadata["entity_url"]);
  }
```

omarelkashef: Yes, and they might be different. For example, in instance backup, metadata.entityURL is /1.0/instances/{instance}/backups/{backupName}?project={project} while entityURL is /1.0/instances/{instance}?project={project}

kimanhou: What you're saying is referring to latest edge, right ? 
Is it also the case for older versions ?

omarelkashef: Yes, but instead of `entity_url`, it was in `resources`.

src/pages/operations/OperationInstanceName.tsx
Kxiru: > In which case the action column is not stacked like that ?

I was asking the same question.

LGTM, @omarelkashef !

Kxiru (dismissed):
LGTM, thank you for addressing this so quickly! I understand this is waiting on another PR?

src/util/operations.tsx
```diff
@@ -5,27 +5,33 @@ import type { LxdOperationResponse } from "types/operation";
...
+  entities: string[] = [],
 ): string[] => {
-  const candidates = entities ?? [];
+  const candidates = new Set<string>();
+  if (operation?.entity_url) {
+    candidates.add(operation.entity_url);
+  }
+  for (const entity of entities) {
+    candidates.add(entity);
+  }
```
edlerd:
```suggestion
  const candidates = new Set<string>(entities);
  if (operation?.entity_url) {
    candidates.add(operation.entity_url);
  }
```
I think we can slightly simplify here.

edlerd (commented):
LGTM, thanks for the prompt fix here! Small suggestion below.

src/util/operations.tsx
omarelkashef: Yes, much simpler.

edlerd (dismissed):
LGTM

kimanhou:
I assume this failed test is linked to Din's changes on network forward API endpoints. Should be fixed next pulse.

image

I expect other tests to fail (related to storage buckets, network forwards and network peers)

omarelkashef:
> I assume this failed test is linked to Din's changes on network forward API endpoints. Should be fixed next pulse.
> 
> image
> I expect other tests to fail (related to storage buckets, network forwards and network peers)

It makes sense they related. We will fix the CI one test at a time.

kimanhou:
Create network forward will be fixed next pulse.
@omarelkashef Any idea about the other two tests ?
image

edlerd:
> Create network forward will be fixed next pulse. @omarelkashef Any idea about the other two tests ? 

LXD latest/edge is not yet including the operation API fixes from https://github.com/canonical/lxd/pull/18033

Let's re-run CI on this PR to ensure those are fixed once the changes are built into the edge channel.

omarelkashef:
> LXD latest/edge is not yet including the operation API fixes from [canonical/lxd#18033](https://github.com/canonical/lxd/pull/18033)

Changes have not landed in latest/edge yet, but we are rebasing an pushing to be ready.

src/util/operations.tsx
```diff
@@ -4,28 +4,44 @@ import type { LxdEvent } from "types/event";
...
   operation?: LxdOperation,
+  entities: string[] = [],
 ): string[] => {
-  const candidates = entities ?? [];
-  if (operation?.metadata?.["entity_url"]) {
-    candidates.push(operation.metadata["entity_url"]);
+  const candidates = new Set<string>(entities);
+  const entityUrlField = isRenamingOperation(operation)
+    ? "original_entity_url"
+    : "entity_url";
```
edlerd: We need to stay backward compatible here. I think this only works with the new schema.

src/util/operations.spec.ts
edlerd, line 1: Can we keep the tests to ensure backward compatibility and add new tests for the new schema?

src/util/operations.tsx
omarelkashef: In the old schema, `original_entity_url` is empty so `isRenamingOperation` returns false and we proceed as normal in `getOperationEntityUrls`

edlerd (approved):
LGTM 👍

PR #1928, opened 2026-04-07
WD-35391: Do not display "Same for all members" checkbox if the cluster has only one node

kimanhou:
## Done

- Do not display `Same for all cluster members` checkbox if the cluster has only one node.
- Affected screens: 
    - existing storage pool > Configuration
    - Settings > core.syslog_socket
    - Settings > cluster.https_address
    - Network configuration with Parent (type=macvlan)
    - Network configuration > External interfaces
    - Create storage pool: source
    - Create storage pool > ZFS

I updated the test `create a member-specific physical network`, which was failing because `Same for all cluster members` checkbox is not displayed anymore. I created a VM and added it to the cluster, however it's offline.
The good news is that the test doesn't fail on `await page.getByText("Same for all cluster members").click();` anymore.
The less good news is that the `Create` button is disabled because the new member is offline so the test fails anyway.

## QA

1. Run the LXD-UI:
    - Check all affected screens in a one-member cluster and ensure the `Same for all cluster members` checkbox is **not** displayed.
    - Check all affected screens in a 2-member cluster and ensure the `Same for all cluster members` checkbox is displayed.
    - Rely on CI passing

## Screenshots

Tests
Screenshot from 2026-04-30 15-14-33

Affected screens
Existing storage pool > Configuration with `Same for all cluster members` checkbox
Screenshot from 2026-04-01 09-59-12

Existing storage pool > Configuration without `Same for all cluster members` checkbox
Screenshot from 2026-04-01 09-58-39

Settings > core.syslog_socket with `Same for all cluster members` checkbox
Screenshot from 2026-04-01 09-59-43

Settings > core.syslog_socket without `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-03-16

Settings > cluster.https_address with `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-14-03

Settings > cluster.https_address without `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-14-22

Network configuration with Parent with `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-06-22

Network configuration with Parent without `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-06-41

Network configuration > External interfaces with `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-12-11

Network configuration > External interfaces without `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-12-31

Create storage pool: source with `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-09-26

Create storage pool: source without `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-10-01

Create storage pool > ZFS with `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-16-21

Create storage pool > ZFS without `Same for all cluster members` checkbox
Screenshot from 2026-04-01 10-16-48

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -31,7 +31,6 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
-
       - name: Install dependencies
         run: yarn install --immutable
 
@@ -93,7 +92,12 @@ jobs:
       matrix:
         lxd_channel: ["5.0/edge", "5.21/edge", "latest/edge"]
         browser: ["chromium", "firefox"]
-        deployment: ["unclustered", "clustered"]
+        deployment: ["unclustered", "clustered", "single-node-cluster"]
+        exclude:
+          - lxd_channel: "5.0/edge"
+            deployment: "clustered"
+          - lxd_channel: "5.0/edge"
+            deployment: "single-node-cluster"
     outputs:
       job_status: ${{job.status}}
     env:
@@ -115,24 +119,19 @@ jobs:
 
       - name: Install LXD-UI dependencies
         run: |
-          set -x
           sudo chmod 0777 ../lxd-ui
           dotrun install
+
+      - name: Generate SSL keys for HAProxy
+        run: |
+          mkdir -p keys
+          openssl req -x509 -newkey rsa:4096 -keyout keys/lxd-ui.key -out keys/lxd-ui.crt -days 365 -nodes -subj "/CN=localhost"
+          cat keys/lxd-ui.key keys/lxd-ui.crt > keys/lxd-ui.pem
 
       - name: OVN dependencies
         run: |
           sudo snap refresh snapd --channel latest/beta || sudo snap install snapd --channel latest/beta
           sudo snap refresh core26 --channel latest/edge || sudo snap install core26 --channel latest/edge
-
-      - name: Run LXD-UI
-        env:
-          ENVIRONMENT: devel
-          PORT: 8407
-          LXD_UI_BACKEND_IP: 172.17.0.1
-          VITE_PORT: 3000
-        run: |
-          dotrun &
-          curl --head --fail --retry-delay 2 --retry 100 --retry-connrefused --insecure https://localhost:8407
 
       - name: Setup Ceph
         if: ${{ matrix.lxd_channel != '5.0/edge' }}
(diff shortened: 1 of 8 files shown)


tests/enable-clustering.spec.ts
```diff
@@ -12,7 +12,7 @@ test("check enabling clustering", async ({ page, lxdVersion }, testInfo) => {
   await expect(page.getByText("This server is not clustered")).toBeVisible();
   await page.getByRole("button", { name: "Enable clustering" }).click();
   await page.getByLabel("Server name").fill("micro1");
-  await page.getByLabel("Cluster address").fill("127.0.0.1");
+  await page.getByLabel("Cluster address").fill("192.168.1.216");
```
kimanhou: I had to change this to make it work locally but it's obviously wrong. Please help 🙏

tests/helpers/cluster.ts
```diff
@@ -89,3 +100,106 @@ export const getFirstClusterMember = async (page: Page): Promise<string> => {
     .textContent();
   return firstCellContent?.split("http")[0] ?? "";
 };
+
+export const addMemberToCluster = async (
```
edlerd: As discussed, the best approach is very likely to boot two nested VMs in the workflow instead. Then install the LXD snap with the right channel in the VMs with `lxc exec vm1 snap install lxc --channel=foo`. Then configure both VMs as a cluster with CLI commands directly in the workflow file `pr.yaml` using `lxc exec ...`. This setup step must be exclusive to the `clustered` matrix tests. We then can update the `LXD_UI_BACKEND_IP` in the workflow to the IP of one of the two VMs before starting `dotrun` and have the suite run against a cluster with two members.

.github/workflows/pr.yaml
```diff
@@ -163,25 +261,14 @@ jobs:
...
-
       - name: Run Playwright tests
         env:
           BROWSER: ${{ matrix.browser }}
           DEPLOYMENT: ${{ matrix.deployment }}
-          LXD_CHANNEL: ${{ steps.lxd-env.outputs.LXD_CHANNEL }}
+          LXD_CHANNEL: ${{ matrix.lxd_channel }}
         run: |
-          npx playwright test --project "${BROWSER}:lxd-${LXD_CHANNEL}:${DEPLOYMENT}"
+          LXD_CH=$(echo "${LXD_CHANNEL}" | sed 's#/#-#g')
```
edlerd:
```suggestion
          PLAYWRIGHT_CHANNEL=$(echo "${LXD_CHANNEL}" | sed 's#/#-#g')
```
Maybe a better name?

.github/workflows/pr.yaml
```diff
@@ -238,9 +325,8 @@ jobs:
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         if: always()
         with:
-          name: playwright-report
-          path: playwright-report/
-          retention-days: 14
+          name: test-results-${{ matrix.browser }}-${{ matrix.deployment }}
+          path: test-results/
```
edlerd: I think this should be reverted, looks like this was changed by another PR while this PR was open.

tests/helpers/cluster.ts
```diff
@@ -1,14 +1,24 @@
...
 export const skipIfNotClustered = (projectName: string) => {
   test.skip(!projectName.includes(":clustered"));
 };
 
 export const getFirstClusterMember = async (page: Page): Promise<string> => {
   await gotoURL(page, "/ui/");
-  await page.getByRole("button", { name: "Clustering" }).click();
+  await page
+    .getByRole("button", { name: "Clustering" })
+    .click({ timeout: 5000 });
```
edlerd: This should not be needed. Can we revert it?

tests/helpers/projects.ts
```diff
@@ -17,7 +17,7 @@ const openProjectCreationForm = async (page: Page) => {
   await expect(page.getByText("Loading storage pools")).not.toBeVisible();
 
   await page.getByLabel("Default instance network", { exact: true }).click();
-  await page.getByText("No network", { exact: true }).click();
+  await page.getByTitle("No network", { exact: true }).click();
```
edlerd, line 21: Why this change?

tests/scripts/setup_test
```diff
@@ -1,59 +1,45 @@
...
-hasNeededAPIExtension() (
-    needed_extension="${1}"
-    lxc info | grep -qxFm1 -- "- ${needed_extension}"
-)
-
-# setup oidc configs
-if [ -f .env.local ]
-then
-  set -o allexport; source .env.local; set +o allexport
+EXEC=""
```
edlerd:
```suggestion
LXC_CMD="lxc"
if [ "$DEPLOYMENT" = "clustered" ]; then
  LXC_CMD="sudo lxc exec vm1 -- lxc"
fi
```
This seems like a less invasive and more straightforward approach, below we can then replace `lxc` with `$LXC_CMD` to run in the right environment: host or via vm.

tests/scripts/setup_test
```diff
@@ -1,59 +1,45 @@
...
+    EXEC="sudo lxc exec vm1 -- "
 fi
-lxc config set oidc.issuer=${LXD_OIDC_ISSUER}
-lxc config set oidc.client.id=${LXD_OIDC_CLIENT_ID}
-if hasNeededAPIExtension "oidc_client_secret"; then
-    lxc config set oidc.client.secret=${LXD_OIDC_CLIENT_SECRET}
-    echo "OIDC client secret is set."
+
+# Function to set config by trying every known naming convention
+smart_set() {
```
edlerd: I think we can avoid such indirections and keep directly calling `$LXC_CMD config set oidc.issuer=${LXD_OIDC_ISSUER}` .

tests/scripts/setup_test
```diff
@@ -1,59 +1,45 @@
...
+}
+
+# Apply OIDC settings
+smart_set "issuer" "${LXD_OIDC_ISSUER}"
+smart_set "client.id" "${LXD_OIDC_CLIENT_ID}"
+smart_set "audience" "${LXD_OIDC_AUDIENCE}"
+smart_set "groups.claim" "${LXD_OIDC_GROUPS_CLAIM}"
+
+# Logic for Secret (Only if the extension exists)
+if $EXEC lxc info | grep -q "oidc_client_secret"; then
```
edlerd: Let's keep the `hasNeededAPIExtension`, if it internally also uses `$LXC_CMD` it should "just work" as before.

tests/scripts/setup_test
```diff
@@ -1,59 +1,45 @@
...
-        (SELECT id from identities WHERE name='foo'),
-        2, 
-        5, 
-        'foo@foo.com', 
-        'foo', 
-        '{}'
-    );
-"
+# SQL - Ensure we don't fail if the user already exists
+$EXEC lxd sql global "INSERT OR REPLACE INTO identities (auth_method, type, identifier, name, metadata) VALUES (2, 5, 'foo@foo.com', 'foo', '{}');" || true
```
edlerd: Next to `LXC_CMD` we can also have a `LXD_CMD` to be used here.

tests/scripts/setup_test
```diff
@@ -1,59 +1,45 @@
...
-        5, 
-        'bar@bar.com', 
-        'bar', 
-        '{}'
-    );
-"
+# THE RESTART: Forces discovery before Playwright starts
+echo "Restarting LXD to force OIDC discovery..."
+$EXEC snap restart lxd
+sleep 20
```
edlerd: Where does this requirement come from? I am not sure if this is actually needed here.

playwright.config.ts
```diff
@@ -123,32 +111,48 @@ const config: PlaywrightTestConfig<TestOptions> = {
...
+        lxdVersion: "5.21-edge",
       },
-      testMatch: "*-clustered.spec.ts",
-      dependencies: ["enable-clustering-chrome"],
+      dependencies: ["login-chromium"],
+      testMatch: "enable-clustering.spec.ts",
     },
     {
-      name: "firefox:lxd-5.0-edge:clustered",
+      name: "firefox:lxd-5.21-edge:single-cluster",
```
edlerd:
```suggestion
      name: "firefox:lxd-5.21-edge:single-node-cluster",
```
Maybe a more understandable name? Though it is a bit more bulky.

We also need to adjust pr.yaml to have this dimension in the `matrix.deployment`. From the setup it should likely be treated similarly to the `Setup LXD (Unclustered)`. Maybe change that name to `Setup LXD (Unclustered on host)` and the other one to `LXD (Clustered in VMs)`.

edlerd (commented):
Great progress, the `5.0/edge` need to skip some steps if clustered. But the main thing is that the clustered tests with newer LXD are passing in a reasonable time.

Some more ideas to simplify and streamline are below.

tests/scripts/setup_test
```diff
@@ -1,59 +1,41 @@
...
-lxc auth group create login-admin
-lxc auth group permission add login-admin server admin
-# The name of the identity provider group should be the same as the role name assigned to the user
-lxc auth identity-provider-group create admin
-lxc auth identity-provider-group group add admin login-admin
+# Auth groups
+$LXC_CMD auth group create login-admin || true
+$LXC_CMD auth group permission add login-admin server admin || true
+$LXC_CMD auth identity-provider-group create admin || true
+$LXC_CMD auth identity-provider-group group add admin login-admin || true
```
edlerd: We shouldn't mask the errors, can we remove the `|| true`?

tests/scripts/setup_test
```diff
@@ -1,59 +1,41 @@
...
+    echo "Configuring clustered environment via vm1..."
+    LXC_CMD="sudo lxc exec vm1 -- lxc"
+    LXD_CMD="sudo lxc exec vm1 -- lxd"
+    SNAP_CMD="sudo lxc exec vm1 -- snap"
+fi
+
 # hasNeededAPIExtension: check if LXD supports the needed extension.
 hasNeededAPIExtension() (
     needed_extension="${1}"
     lxc info | grep -qxFm1 -- "- ${needed_extension}"
```
edlerd: This also needs to use `LXC_CMD`

tests/networks-clustered.spec.ts
edlerd, line 1: The import changes should probably be reverted.

.github/workflows/pr.yaml
edlerd, line 1: We need to add the new single-node-cluster dimension here, so we run the new playwright channel as well.

.github/workflows/pr.yaml
```diff
@@ -146,42 +145,168 @@ jobs:
         uses: canonical/setup-lxd@8c6a87bfb56aa48f3fb9b830baa18562d8bfd4ee # v1
         with:
           channel: ${{ matrix.lxd_channel }}
-          group: lxd
 
-      - name: Setup LXD
+      - name: Setup LXD (Clustered in VMs)
+        if: ${{ matrix.deployment == 'clustered' && matrix.lxd_channel != '5.0/edge' }}
```
edlerd:
```suggestion
        if: ${{ matrix.deployment == 'clustered' }}
```
Simplify with the exclusion list above.

.github/workflows/pr.yaml
```diff
@@ -93,7 +92,12 @@ jobs:
       matrix:
         lxd_channel: ["5.0/edge", "5.21/edge", "latest/edge"]
         browser: ["chromium", "firefox"]
-        deployment: ["unclustered", "clustered"]
+        deployment: ["unclustered", "clustered", "single-node-cluster"]
+        exclude:
+          - lxd_channel: "5.0/edge"
+            deployment: "clustered"
+          - lxd_channel: "5.0/edge"
+            deployment: "single-node-cluster"
```
edlerd, line 100: good idea to use the exclusion list here.

.github/workflows/pr.yaml
```diff
@@ -146,42 +145,168 @@ jobs:
...
-        with:
-          node-version: 24
-
-      - name: Install Playwright Browsers
-        run: npx playwright install --with-deps ${{ matrix.browser }}
 
-      - name: Set lxd channel env variable
-        id: lxd-env
+      - name: Setup LXD (Single-node cluster)
+        if: ${{ matrix.deployment == 'single-node-cluster' && matrix.lxd_channel != '5.0/edge' }}
```
edlerd:
```suggestion
        if: ${{ matrix.deployment == 'single-node-cluster' }}
```

tests/scripts/setup_test
```diff
@@ -1,59 +1,41 @@
...
     needed_extension="${1}"
-    lxc info | grep -qxFm1 -- "- ${needed_extension}"
+    $LXC_CMD info | grep -qxFm1 -- "- ${needed_extension}"
 )
 
-# setup oidc configs
-if [ -f .env.local ]
-then
-  set -o allexport; source .env.local; set +o allexport
-fi
```
edlerd, line 14: Let's keep reading the env file to allow setting the environment variables used below.

edlerd (commented):
Good progress, few comments on the changes below.

kimanhou:
I have no idea why 5.21 clustered fails on chromium but passes on firefox

image

image

edlerd:
> I have no idea why 5.21 clustered fails on chromium but passes on firefox

The two OVN networks created by the test to then link them are in error status.

image

This is the error on the instance page:

image

I suspect some timing issue with the OVN setup. I don't think it is due to chrome, more likely to fail randomly and now we have a flaky situation with OVN.

tests/helpers/network.ts
```diff
@@ -218,6 +218,49 @@ export const submitCreateNetwork = async (page: Page, network: string) => {
   await expect(networkLink).toBeVisible();
 };
 
+export const waitForNetworkReady = async (
```
edlerd: I am not sure if this is needed. Wasn't the previous solution waiting up to 5 seconds and then checking once stable enough? I think the OVN error with peering tests was unrelated, as the networks stayed in an error state back then. So this busy wait approach wouldn't have resolved it either.

edlerd (commented):
Great stuff, finally passing, and it looks clean 👍 

Two remaining concerns: one open question on the project's test helper and wondering about the new network busy-wait approach.

tests/helpers/projects.ts
kimanhou: Otherwise the tests failed because they would find 2 components matching the request

tests/helpers/network.ts
kimanhou: I removed it. CI is green again. I left the sleep and the check for OVN in pr.yaml but I'm not sure they are needed

```
          # Wait for OVN chassis to be fully registered
          echo "Waiting for OVN chassis registration..."
          sleep 15

          # Verify chassis are registered
          echo "Checking chassis status..."
          sudo lxc exec vm1 -- sudo ovn-sbctl show

          # Wait for ovn-controller to be ready on both nodes
          for vm in vm1 vm2; do
            echo "Checking ovn-controller on $vm..."
            sudo lxc exec $vm -- sudo systemctl is-active ovn-controller || true
            sudo lxc exec $vm -- sudo ovn-sbctl list chassis || true
          done
```

edlerd (approved):
LGTM, thank you so much for making our CI also run on a real clustered environment in VMs!

PR #1933, opened 2026-04-08
fix(permission-groups): select group before managing identities

kimanhou:
## Done

- fix(permission-groups): select group before clicking on `Manage identities` (previously the groups stay selected after panel closes)

## QA

1. Run the LXD-UI:
    - Rely on CI

## Screenshots
image

### DIFF
--- a/src/pages/permissions/PermissionGroups.tsx
+++ b/src/pages/permissions/PermissionGroups.tsx
@@ -56,10 +56,8 @@ const PermissionGroups: FC = () => {
   useEffect(() => {
     if (panelParams.group) {
       setSelectedGroupNames([panelParams.group]);
-    } else if (panelParams.panel === null) {
-      setSelectedGroupNames([]);
     }
-  }, [panelParams.group, panelParams.panel, groups]);
+  }, [panelParams.group, groups]);
 
   const headers = [
     { content: "Name", className: "name", sortKey: "name" },


tests/permission-groups.spec.ts
```diff
@@ -120,6 +120,7 @@ test("manage identities for many groups", async ({ page, lxdVersion }) => {
   );
   await page.getByRole("button", { name: "Confirm changes" }).click();
   await page.waitForSelector(`text=Updated identities for 2 groups`);
+  await selectGroupsToModify(page, [groupOne, groupTwo]);
```
edlerd: Let's ensure the groups stay selected instead. Why does the selection state get lost now?

edlerd (commented):
I think we shouldn't need to adjust the test here. See below.

tests/permission-groups.spec.ts
kimanhou: Because of this [PR](https://github.com/canonical/lxd-ui/pull/1909)

image

In which case should the group stay selected or not ?
Once a group was updated and the side panel is closed, it makes sense to unselect it, no ?

edlerd: We tend to keep the selection also on the instance list. Gmail also does it on bulk actions, so it seems expected to keep the state.

The user might have done "a lot of work" to choose the right items. They can easily unselect with a single click, but reselecting might be hard. There might be follow up things they want to do with the selection.

edlerd (approved):
LGTM, thanks for the fix.

PR #1940, opened 2026-04-10
feat(image registry): detail page

kimanhou:
## Done

- feat(image registry): detail page

## QA

1. Run the LXD-UI:
    - Create an image registry with the command line: `lxc image registry create poney --protocol simplestreams url=https://ubuntu.com public=true`
    - Go to the image registry detail page: https://localhost:8407/ui/image-registry/poney

## Screenshots

Detail page - Images
Screenshot from 2026-04-24 12-18-38

Detail page - Configuration
image


omarelkashef:
Should the URL include project ?
-> I think so for the sake of consistency with the rest of the registries pages, but we will need to inform the user that it is a global entity and not project-scoped. We already have a ticket for that.

What icon should be used ? I am using the same as images for now.
-> I am using the repository for now since it indicates a collection of things, but I am open to explore more or even ask for a new icon.

When renaming an image registry, should we check for duplicated names ? If so, how ?
-> Yes, I am using `checkDuplicateName` function which sends a GET request and if it gets a 404, it means that name is not duplicated.

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+    {
+      content: "Upload date",
+      sortKey: "uploaded_at",
+      className: "uploaded_at",
+    },
+    { content: "Size", sortKey: "size", className: "u-align--right size" },
+    { "aria-label": "Actions", className: "actions" },
+  ];
+
+  const rows = images.map((image) => {
```
edlerd, line 72: These are huge lists.

Let's add a search and filter component to the image list. By default images without an alias should be hidden. By default only the architecture that matches the lxd server should be shown.

On images with a non-supported architecture, we should also hide the play button.

src/components/Navigation.tsx
```diff
@@ -72,38 +71,23 @@ const initialiseOpenNavMenus = (location: Location) => {
...
+interface Props {
+  projectName: string;
+  setProjectName: (projectName: string) => void;
+}
 
-  return "default";
-};
-
-const Navigation: FC = () => {
+const Navigation: FC<Props> = ({ projectName, setProjectName }) => {
```
edlerd: We also need to consider the image registry detail page to open the images sub-menu and to mark it active on collapsing the images section (like on the local images or registry list page).

src/context/useImageRegistries.tsx
```diff
@@ -20,3 +24,11 @@ export const useImageRegistry = (name: string, isEnabled = true) => {
     enabled: isEnabled && isFineGrained !== null,
   });
 };
+
+export const useImageRegistryImages = (imageRegistryName: string) => {
+  return useQuery({
+    queryKey: [queryKeys.imageRegistries, imageRegistryName, "images"],
+    queryFn: async () => fetchImageRegistryImages(imageRegistryName),
```
edlerd: Maybe shorten this to `useRegistryImages` and `fetchRegistryImages`. Wdyt?

edlerd (commented):
Good start! Some high level thoughts on the draft PR below.

src/pages/images/ImagesTable.tsx
kimanhou: > Let's add a search and filter component to the image list. 

Done

> By default images without an alias should be hidden.

Done

> By default only the architecture that matches the lxd server should be shown.

Done

> On images with a non-supported architecture, we should also hide the play button.

I don't understand, when would an image with a non-supported architecture appear, since we are filtering them out of the list ?

edlerd: I thought the filter component defaults to only compatible image architectures.

The user can reveal other architectures available on the registry by changing the filter status. The filter chips should show all architectures, but only the compatible ones are active when landing on the page.

kimanhou: Should I reuse `ImageSelector` then ?

edlerd: Good question.
I would argue this use case is slightly different because it can give you an exhaustive list of images in a registry. The ones compatible but also the incompatible ones with the current cluster. So you are exploring the registry images, not selecting an image for instance creation.

kimanhou: Just pushed and updated the screenshot in the PR description.

edlerd: I think there was a misunderstanding. I was thinking of a search and filter component such as on the instance list or in the permission identities.

Also, this shouldn't be a selectable table.

edlerd: In the permission identities we already have the pre-set filters on landing to show relevant items. For Registry images I'd select all architectures that apply, not just a single one.

edlerd: nice :+1:

src/components/Navigation.tsx
edlerd: This is partially solved.

One bit missing: on the registry detail page, when collapsing the "images" side nav, the "images" side nav entry should become the active one.

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,159 @@
...
+    {
+      content: "Architecture",
+      sortKey: "architecture",
+      className: "architecture",
+    },
+    {
+      content: "Cached",
+      sortKey: "cached",
+      className: "cached",
+    },
```
edlerd:
```suggestion
```
I think this is not relevant here; it will always be false.

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,159 @@
...
+          content: imageName,
+          role: "rowheader",
+          "aria-label": "Name",
+        },
+        {
+          content: imageAlias,
+          role: "cell",
+          "aria-label": "Aliases",
+          className: "aliases",
+        },
```
edlerd, line 104: This column in the table is a bit messy. The aliases often come in multiples and tend to be very long. Maybe we should show one alias per line, truncated with a title, or use an expandable list here. Wdyt?

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,159 @@
...
+          "aria-label": "Name",
+        },
+        {
+          content: imageAlias,
+          role: "cell",
+          "aria-label": "Aliases",
+          className: "aliases",
+        },
+        {
+          content: image.architecture,
```
edlerd: In the image selector we have some translation to a nicer display variant of the architectures. We don't have that on the local images and neither here. I think it would be nice to use it here as well. And in the search and filter above the table. Basically wherever we surface an architecture to the user.

src/components/Navigation.tsx
kimanhou: It should be fixed now

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,159 @@
...
+    {
+      content: "Upload date",
+      sortKey: "uploaded_at",
+      className: "uploaded_at",
+    },
+    { content: "Size", sortKey: "size", className: "u-align--right size" },
+    { "aria-label": "Actions", className: "actions" },
+  ];
+
+  const rows = images.map((image) => {
```
edlerd, line 72: When landing on the page, the entries are ordered in a bit random way. Maybe can enable sort by name by default? Or we could apply the `byOSRelease` to the images list?

edlerd (commented):
Good progress, the search and filter on the registry image list work nicely.

src/components/Navigation.tsx
edlerd: yes, thanks :)

src/pages/images/ImageTable.tsx
edlerd: Ah, I think `byOSRelease` is incompatible here as we deal with another type. but something similar to it might be nice to apply.

edlerd: We sometimes apply ordering in the api layer. Maybe we can sort in `fetchRegistryImages`.

src/context/useImageRegistries.tsx
```diff
@@ -20,3 +24,11 @@ export const useImageRegistry = (name: string, isEnabled = true) => {
     enabled: isEnabled && isFineGrained !== null,
   });
 };
+
+export const useRegistryImages = (imageRegistryName: string) => {
+  return useQuery({
+    queryKey: [queryKeys.imageRegistries, imageRegistryName, "images"],
+    queryFn: async () => fetchRegistryImages(imageRegistryName),
+    enabled: !!imageRegistryName,
```
edlerd:
```suggestion
```
This seems not needed. Is this a guard for empty string?

src/pages/images/ImageRegistryDetail.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return (
+      <NotFound
+        entityType="image-registry"
+        entityName={name}
+        errorMessage={error?.message}
+      />
+    );
+  }
+
+  const tabs: (string | TabLink)[] = ["Overview", "Images"];
```
edlerd:
```suggestion
  const tabs: string[] = ["Overview", "Images"];
```
A bit simpler

src/pages/images/ImageRegistryDetailHeader.tsx
```diff
@@ -0,0 +1,120 @@
...
+      return "Invalid Image Registry: Cannot be renamed";
+    }
+
+    return undefined;
+  };
+
+  const onSuccess = (imageRegistryName: string) => {
+    const url = `${ROOT_PATH}/ui/image-registry/${encodeURIComponent(imageRegistryName)}`;
+    navigate(url);
+    console.log("Navigating to", url);
```
edlerd:
```suggestion
```

src/Root.tsx
```diff
@@ -1,22 +1,53 @@
...
+    if (project && !isLoading) {
+      return project.name;
+    }
+
+    return "default";
+  };
+
+  const [projectName, setProjectName] = useState(
+    initializeProjectName(isAllProjectsFromUrl, isLoading, project),
+  );
```
edlerd: I think instead of leaking this logic in here, we can extend the `ProjectProvider` to have a `projectName` state. Then we also can avoid the prop drilling. We can call `useCurrentProject` directly in the registry detail pages.

src/pages/images/ImagesSearchFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+} from "util/searchAndFilter";
+
+export interface ImagesSearchFilterType {
+  queries: string[];
+  arch: string[];
+}
+
+export const QUERY = "query";
+export const ARCHITECTURE = "arch";
+const QUERY_PARAMS = [QUERY, ARCHITECTURE];
```
edlerd: A further interesting filter could be type container/vm.

edlerd (commented):
Good progress on the filter and the registry image list!

Some more observations from a high level code review are below.

src/pages/images/ImagesTable.tsx
kimanhou: I don't see a translation in ImageSelector.

```
  const getOptionList: (
    mapper: (item: RemoteImage) => string,
    filter?: (item: RemoteImage) => boolean,
  ) => OptionHTMLAttributes<HTMLOptionElement>[] = (
    mapper,
    filter = () => true,
  ) => {
    const options = [...new Set(images.filter(filter).map(mapper))]
      .sort()
      .map((item: string) => {
        return {
          label: item,
          value: item,
        };
      });
    options.unshift({
      label: "Any",
      value: "",
    });
    return options;
  };
  ```
  
  The current ImageSelector shows `amd64` and `i386`. Are they nice display names ?

File: src/pages/images/ImageTable.tsx (hunk 7 above)
Comment (kimanhou) [reply src/pages/images/ImageTable.tsx:72]: We can use byOSRelease if we convert the LXD image into a remote image first, no ?

```
export const fetchRegistryImages = async (
  name: string,
  isFineGrained: boolean | null,
): Promise<LxdImage[]> => {
  params.set("recursion", "1");
  addEntitlements(params, isFineGrained, imageRegistryEntitlements);
  return fetch(
    `${ROOT_PATH}/1.0/image-registries/${encodeURIComponent(name)}/images?${params.toString()}`,
  )
    .then(handleResponse)
    .then((data: LxdApiResponse<LxdImage[]>) => {
      return data.metadata.sort((a, b) => {
        const aRemote = localLxdToRemoteImage(a);
        const bRemote = localLxdToRemoteImage(b);
        return byOSRelease(aRemote, bRemote);
      });
    });
};
```

Comment (edlerd) [reply src/pages/images/ImageTable.tsx:72]: > We can use byOSRelease if we convert the LXD image into a remote image first, no ?

It could be ok, though calling `localLxdToRemoteImage` inside the sort is very costly. We will convert each image multiple times for comparison.

File: src/pages/images/ImagesTable.tsx (hunk 6 above)
Comment (edlerd) [reply src/pages/images/ImagesTable.tsx]: Oh, I might mix things up. Assume there is a conversion layer, but maybe it is LXD doing that?

I think the `amd64` is the best way, and `i386` as well. Both are better than the strings with underscores we show currently on the detail page image list and filter. But up for discussion. Sounds like a good topic for the core team to discuss.

File: src/pages/images/ImageTable.tsx (hunk 7 above)
Comment (kimanhou) [reply src/pages/images/ImageTable.tsx:72]: Indeed

Comment (edlerd) [reply src/pages/images/ImageTable.tsx:72]: Can we adjust `byOSRelease` to be compatible with both types? I think it should be more or less trivial to make it compatible.

File: src/Root.tsx [hunk 13]
```diff
@@ -1,29 +1,36 @@
...
 import ErrorBoundary from "components/ErrorBoundary";
 import ErrorPage from "components/ErrorPage";
 import StatusBar from "components/StatusBar";
 import AppProviders from "context/appProviders";
+import { useCurrentProject } from "context/useCurrentProject";
+import App from "./App";
+import Events from "pages/instances/Events";
+
+const AppContent: FC = () => {
+  const { projectName, setProjectName } = useCurrentProject();
```
edlerd: I think we can keep `Root.tsx` unchanged. Inside the `Navigation` component, we already use the hook `useCurrentProject` and can receive the `projectName` and `setProjectName` directly from it.

src/context/useImageRegistries.tsx
```diff
@@ -20,3 +24,11 @@ export const useImageRegistry = (name: string, isEnabled = true) => {
     enabled: isEnabled && isFineGrained !== null,
   });
 };
+
+export const useRegistryImages = (imageRegistryName: string) => {
+  const { isFineGrained } = useAuth();
+  return useQuery({
+    queryKey: [queryKeys.imageRegistries, imageRegistryName, "images"],
+    queryFn: async () => fetchRegistryImages(imageRegistryName, isFineGrained),
```
edlerd, line 32:
```suggestion
    queryFn: async () => fetchRegistryImages(imageRegistryName, isFineGrained),
    enabled: isFineGrained !== null,
```
I think there is a race condition without this.

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -78,7 +79,13 @@ const ImageRegistriesList: FC = () => {
       name: registry.name,
       columns: [
         {
-          content: registry.name,
+          content: (
+            <Link
+              to={`${ROOT_PATH}/ui/image-registry/${encodeURIComponent(registry.name)}`}
+            >
+              {registry.name}
```
edlerd, line 91: Should we add truncation here? I suspect the registry name can be very long.

src/pages/images/ImagesSearchFilter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const chips = useMemo(
+    () => searchParamsToChips(searchParams, QUERY_PARAMS),
+    [searchParams],
+  );
+
+  const searchAndFilterData: SearchAndFilterData[] = [
+    {
+      id: 1,
+      heading: "Architecture",
+      chips: availableArchitectures.map((arch) => ({
```
edlerd, line 121: Can we also use the nicer display names here?

src/api/image-registries.tsx
```diff
@@ -47,6 +48,28 @@ export const fetchRegistryImages = async (
   )
     .then(handleResponse)
     .then((data: LxdApiResponse<LxdImage[]>) => {
-      return data.metadata;
+      return data.metadata.sort((a, b) => byOSRelease(a, b));
```
edlerd: This somehow gets bamboozled on the registry detail page images list. I get 25, then 26, then 24 for Ubuntu daily. For others as well, the order seems randomized.

Image

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,164 @@
...
+          className: "u-align--right size",
+        },
+        {
+          content: supportedArchitectures.includes(image.architecture) ? (
+            <CreateInstanceFromImageBtn
+              key="launch"
+              projectName={project}
+              image={localLxdToRemoteImage(image)}
+            />
+          ) : null,
```
edlerd: This should consider project restrictions. If the project is restricted and the registry is not allowed for the project, the button should disable.

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,164 @@
...
+        },
+        {
+          content: humanFileSize(image.size),
+          role: "cell",
+          "aria-label": "Size",
+          className: "u-align--right size",
+        },
+        {
+          content: supportedArchitectures.includes(image.architecture) ? (
+            <CreateInstanceFromImageBtn
```
edlerd: I think we shouldn't hide, but disable the button with a title that explains why it is disabled.

src/components/Navigation.tsx
```diff
@@ -72,38 +72,23 @@ const initialiseOpenNavMenus = (location: Location) => {
...
```
edlerd: There is an interaction currently broken from the navigation: switching project when
a) on the local images page
b) on the image registry list page
c) on any of the tabs of the image registry detail page

We should try to keep the user in the context they are in. So can stay on the registry list or detail page, when switching the project in the navigation.

edlerd (commented):
Thanks for the iteration! Previous issues seem all resolved. Some more observations and suggestions below.

src/pages/images/ImagesSearchFilter.tsx
kimanhou: Yes, but what about the URL ?

src/pages/images/actions/CreateInstanceFromImageBtn.tsx
```diff
@@ -28,18 +33,24 @@ const CreateInstanceFromImageBtn: FC<Props> = ({ image, projectName }) => {
     );
   };
 
+  const getTitle = () => {
+    if (!canCreateInstances(project)) {
+      return "You do not have permission to create instances";
+    }
+    if (disabled) {
+      return "You cannot use this image";
```
edlerd: Can we be more specific? Such as `This image is incompatible with your hardware.` or `The current project restricts usage of this image`. Or something similar.

src/Root.tsx
edlerd, line 1: There is no change to this file, right? Can we restore the original version?

src/util/images.tsx
```diff
@@ -73,25 +73,34 @@ export const byLtsFirst = (a: RemoteImage, b: RemoteImage): number => {
...
-    return 1;
-  }
-  return 0;
+    // LxdImage
+    const p = img.properties;
+    const isLTS = p?.description?.toLowerCase().includes("lts");
+    const releaseTitle = `${p?.version ?? ""}${isLTS ? " LTS" : ""}`;
+    const server = img.cached ? (img.update_source?.server ?? "") : LOCAL_IMAGE;
+
+    return `${capitalizeFirstLetter(p?.os ?? "")}${releaseTitle}${p?.release ?? ""}${img.type}${server}`;
```
edlerd: Instead of `capitalizeFirstLetter`, should we lowercase the whole string?

edlerd (commented):
Good progress :)

partial review notes below.

src/pages/images/ImagesSearchFilter.tsx
edlerd: Would be nice to also have the same value in the url. Is that making things too complicated?

kimanhou: I think I managed

src/components/Navigation.tsx
kimanhou: It should be resolved now.

src/api/image-registries.tsx
kimanhou: It should resolved now. The table was sorted by `uploaded_at`.

kimanhou:
Found a funny bug. If you call your project "image-registry", then it will be highlighted in the side navigation even when you're on the instances page. Fixed it by adding "ui/" in front of the URL checks.

image

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+          content: (
+            <CreateInstanceFromImageBtn
+              key="launch"
+              projectName={projectName}
+              image={localLxdToRemoteImage(image)}
+              disabledReason={
+                !supportedArchitectures.includes(image.architecture)
+                  ? "This image is incompatible with your hardware architecture"
+                  : !isImageRegistryAllowed()
+                    ? `The project ${projectName} is restricted from using images from the registry ${imageRegistryName}`
```
edlerd:
```suggestion
                    ? `Images from this registry are not allowed in this project`
```
More concise?

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+        },
+        {
+          content: (
+            <CreateInstanceFromImageBtn
+              key="launch"
+              projectName={projectName}
+              image={localLxdToRemoteImage(image)}
+              disabledReason={
+                !supportedArchitectures.includes(image.architecture)
+                  ? "This image is incompatible with your hardware architecture"
```
edlerd:
```suggestion
                  ? "Image is incompatible with your hardware architecture"
```

src/context/useCurrentProject.tsx
```diff
@@ -40,6 +45,50 @@ export const ProjectProvider: FC<ProviderProps> = ({ children }) => {
...
+    if (project || isAllProjects) {
+      const newProjectName = initializeProjectName(
+        isAllProjects,
+        isProjectLoading || isSettingsLoading,
+        data,
+        project,
+      );
+      setProjectName(newProjectName);
+    }
+  }, [project, isAllProjects, isProjectLoading, isSettingsLoading, data]);
```
edlerd:
```suggestion
  const isLoading = isProjectLoading || isSettingsLoading;
  const [projectName, setProjectName] = useState<string>(
    initializeProjectName(isAllProjects, isLoading, data, project),
  );

  useEffect(() => {
    if (project || isAllProjects) {
      const newProjectName = initializeProjectName(
        isAllProjects,
        isLoading,
        data,
        project,
      );
      setProjectName(newProjectName);
    }
  }, [project, isAllProjects, isLoading, data]);
```
Small simplification.

src/pages/images/actions/DeleteImageRegistryBtn.tsx
```diff
@@ -0,0 +1,131 @@
...
+          <p>
+            This will permanently delete image registry{" "}
+            <ResourceLabel
+              type="image-registry"
+              value={imageRegistry.name}
+              bold
+            />
+            .
+            <br />
+            This action cannot be undone, and can result in data loss.
```
edlerd:
```suggestion
```
I would remove this part of the warning here, as the registry is solely a link to another resource and not holding any data itself.

src/pages/images/actions/DeleteImageRegistryBtn.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const handleDelete = () => {
+    setLoading(true);
+    deleteImageRegistry(imageRegistry.name)
+      .then(() => {
+        onSuccess();
+      })
+      .catch((e) => {
+        onFailure(e);
+      });
```
edlerd:
```suggestion
  deleteImageRegistry(imageRegistry.name).then(onSuccess).catch(onFailure);
```
Let's use the shorthand here.

src/pages/images/actions/DeleteImageRegistryBtn.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { deleteImageRegistry } from "api/image-registries";
+
+interface Props {
+  imageRegistry: LxdImageRegistry;
+  shouldExpand?: boolean;
+}
+
+const DeleteImageRegistryDeleteBtn: FC<Props> = ({
+  imageRegistry,
+  shouldExpand = false,
```
edlerd:
```suggestion
```
This is always true, can we remove it?

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,172 @@
...
+      }
+    } else {
+      initialized.current = true;
+    }
+  }, [
+    availableArchitectures,
+    supportedArchitectures,
+    isImagesLoading,
+    setSearchParams,
+  ]);
```
edlerd: All of this seems to relate to the initialization of the filter. Can we move it into the filter component?

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  const filters: ImagesSearchFilterType = {
+    queries: searchParams.getAll(QUERY).map((q) => q.toLowerCase()),
+    arch: searchParams.getAll(ARCHITECTURE),
+    type: searchParams.getAll(TYPE),
+  };
+
+  const filteredImages = (images ?? []).filter((image) => {
+    if (!image.aliases || image.aliases.length === 0) return false;
```
edlerd: I wonder if we should do this only for built in registries. Because we know their content and want to avoid showing the duplicates. On custom registries the user might have or not have aliases in place.

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    if (!matchesType) return false;
+
+    if (filters.queries.length > 0) {
+      const description = (image.properties?.description ?? "").toLowerCase();
+      const matchesQuery = filters.queries.some(
+        (q) =>
+          description.includes(q) ||
+          image.aliases.some((alias) => alias.name.toLowerCase().includes(q)),
+      );
```
edlerd, line 84: Let's also match the query against the name.

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,172 @@
...
+const TYPE_DISPLAY_MAP = new Map<string, string>([
+  ["container", "Container"],
+  ["virtual-machine", "VM"],
+]);
+
+interface Props {
+  imageRegistry: LxdImageRegistry;
+}
+
+const ImageRegistryImages: FC<Props> = ({ imageRegistry }) => {
```
edlerd, line 24: I think there is an early exit in here somewhere causing the page to crash. I added a non matching filter query, such as "asdf".

Image

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,172 @@
...
+      </Col>
+    );
+  }
+
+  return (
+    <>
+      <div className="u-sv3">
+        <ImageRegistryImagesFilter
+          availableArchitectures={availableArchitectures}
+          availableTypes={availableTypes}
```
edlerd:
```suggestion
          images={images}
```
Probably better to keep those calculations local and expose the images to the filter.

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,172 @@
...
+import { useCurrentProject } from "context/useCurrentProject";
+import { useRegistryImages } from "context/useImageRegistries";
+import { useSettings } from "context/useSettings";
+import ImagesTable from "pages/images/ImagesTable";
+import type { ImagesSearchFilterType } from "pages/images/ImagesSearchFilter";
+import ImageRegistryImagesFilter, {
+  QUERY,
+  ARCHITECTURE,
+  TYPE,
+} from "pages/images/ImagesSearchFilter";
```
edlerd: Let's align the names here. I would also advice singular, seems more natural.
```suggestion
import ImageSearchFilter, {
  QUERY,
  ARCHITECTURE,
  TYPE,
} from "pages/images/ImageSearchFilter";
```

src/pages/images/ImageRegistryOverview.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <Col size={7}>
+          <table>
+            <tbody>
+              <tr>
+                <th className="u-text--muted">Name</th>
+                <td>{imageRegistry.name}</td>
+              </tr>
+              <tr>
+                <th className="u-text--muted">Description</th>
+                <td>{imageRegistry.description ?? "-"}</td>
```
edlerd: This doesn't work on the built-in registries, it renders empty.

Image

src/pages/images/ImageRegistryOverview.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+      <Row className="section">
+        <Col size={3}>
+          <h2 className="p-heading--5">Configuration</h2>
+        </Col>
+        <Col size={7}>
+          <table>
+            <tbody>
+              <tr>
+                <th className="u-text--muted">URL</th>
```
edlerd:
```suggestion
                <th className="u-text--muted">Server</th>
```
Better aligned with the creation form

src/pages/images/ImageRegistryOverview.tsx
```diff
@@ -0,0 +1,71 @@
...
+                <td>{imageRegistry.config?.url ?? "-"}</td>
+              </tr>
+              <tr>
+                <th className="u-text--muted">Public</th>
+                <td>{imageRegistry.config?.public ?? "-"}</td>
```
edlerd: This should use the helper function from @omarelkashef depending on what lands first, this should be resolved by him or you in here.

src/pages/images/ImageRegistryOverview.tsx
```diff
@@ -0,0 +1,71 @@
...
+              </tr>
+              <tr>
+                <th className="u-text--muted">Cluster</th>
+                <td>{imageRegistry.config?.cluster ?? "-"}</td>
```
edlerd: this is a cluster link, let's use the `ResourceLink` targeting the cluster link list.

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+import useSortTableData from "util/useSortTableData";
+
+interface Props {
+  images: LxdImage[];
+  projectName: string;
+  imageRegistryName: string;
+  supportedArchitectures?: string[];
+}
+
+const ImagesTable: FC<Props> = ({
```
edlerd:
```suggestion
const ImageTable: FC<Props> = ({
```

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+  supportedArchitectures?: string[];
+}
+
+const ImagesTable: FC<Props> = ({
+  images,
+  projectName,
+  supportedArchitectures = [],
+  imageRegistryName,
+}) => {
+  const { data: project } = useProject(projectName);
```
edlerd: We can call `useCurrentProject` here and remove it from the props.

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  const isImageRegistryAllowed = (): boolean => {
+    const isProjectRestricted =
+      project?.config["restricted"] === "true" &&
+      project?.config["features.images"] === "true";
```
edlerd: Is the `features.images` really relevant here? My understanding might be wrong, but I assume that is only about local images, not about registries or registry images.

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+      return true;
+    }
+
+    const allowedRegistries =
+      project?.config["restricted.registries"]?.split(",") ?? [];
+
+    // If no registries are specified, assume all are allowed
+    if (allowedRegistries.length === 0) {
+      return true;
+    }
```
edlerd:
```suggestion
```
The inverse is true: if the list is empty, no registry is allowed.

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+    // If no registries are specified, assume all are allowed
+    if (allowedRegistries.length === 0) {
+      return true;
+    }
+
+    return allowedRegistries.includes(imageRegistryName);
+  };
+
+  if (images.length === 0) {
```
edlerd: In this case we should also hide the search and filter component.

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+      className: "architecture",
+    },
+    { content: "Type", sortKey: "type", className: "type" },
+    {
+      content: "Upload date",
+      sortKey: "uploaded_at",
+      className: "uploaded_at",
+    },
+    { content: "Size", sortKey: "size", className: "u-align--right size" },
+    { "aria-label": "Actions", className: "actions" },
```
edlerd, line 69: Can we adjust the spacing of the table? Especially on large screens those two columns are too wide and letting other columns such as the name wrap.

Image

src/pages/images/ImagesTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+      columns: [
+        {
+          content: imageName,
+          role: "rowheader",
+          "aria-label": "Name",
+        },
+        {
+          content:
+            aliasItems.length > 0 ? (
+              <ExpandableList items={aliasItems} displayCount={3} />
```
edlerd:
```suggestion
              <ExpandableList items={aliasItems} displayCount={2} />
```
I think we can reduce it even more here. Good idea to make this a parameter on the component!

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+          className: "aliases",
+        },
+        {
+          content: getArchitectureDisplayName(image.architecture),
+          role: "cell",
+          "aria-label": "Architecture",
+          className: "architecture",
+        },
+        {
+          content: image.type == "virtual-machine" ? "VM" : "Container",
```
edlerd: This `type == "virtual-machine" ? "VM" : "Container"` conversion has spread into so many places. Should we extract it to a util function and replace the spread? Can also be a follow up task to clean up later.

src/pages/images/LocalImageList.tsx
```diff
@@ -140,13 +153,16 @@ const LocalImageList: FC = () => {
           "aria-label": "Name",
         },
         {
-          content: imageAlias,
+          content:
+            aliasItems.length > 0 ? (
+              <ExpandableList items={aliasItems} displayCount={3} />
```
edlerd:
```suggestion
              <ExpandableList items={aliasItems} displayCount={2} />
```
Thanks for adjusting it here as well 👍

src/sass/_images.scss
```diff
@@ -70,3 +70,7 @@
     }
   }
 }
+
+.images-list-search-box-container {
+  margin-top: 2.5rem;
+}
```
edlerd:
```suggestion
```
This seems to be unused.

src/util/architectures.tsx
```diff
@@ -51,3 +51,26 @@ export const getArchitectureAliases = (names: string[]): string[] => {
...
+    x86_64: "amd64",
+    i686: "i386",
+    aarch64: "arm64",
+  };
+
+  return (
+    Object.keys(mapping).find((key) => mapping[key] === displayName) ||
+    displayName
+  );
+};
```
edlerd, line 73:
```suggestion
export const getArchitectureDisplayName = (arch: string): string => {
  const nameKeys = Object.keys(ARCHITECTURE_NAMES);
  const nameKey = nameKeys.find((key) => ARCHITECTURE_NAMES[key] === arch);

  return nameKey && nameKey in ARCHITECTURE_ALIASES
    ? ARCHITECTURE_ALIASES[nameKey][0]
    : arch;
};

export const getArchitectureTechnicalName = (displayName: string): string => {
  const aliasKeys = Object.keys(ARCHITECTURE_ALIASES);
  const aliasKey = aliasKeys.find((key) =>
    ARCHITECTURE_ALIASES[key].includes(displayName),
  );

  return aliasKey && aliasKey in ARCHITECTURE_NAMES
    ? ARCHITECTURE_NAMES[aliasKey]
    : displayName;
};
```
We can reuse the definitions above for translating between display and technical name.

edlerd: Can you please extend the `architectures.spec.ts` with the two new functions?

src/util/images.tsx
```diff
@@ -73,25 +73,121 @@ export const byLtsFirst = (a: RemoteImage, b: RemoteImage): number => {
...
+        return img.title || img.aliases || `${img.os} ${img.release}`;
+      } else {
+        // LxdImage - use name or description or fingerprint
+        return img.name || img.properties?.description || img.fingerprint;
+      }
+    };
+    return getName(a).localeCompare(getName(b));
+  }
+
+  return result;
```
edlerd: This seems a bit too complicated comparing it to the old method with just 10 lines turning into 100. 

I think introducing the `getImageInfo` function is a good idea to handle the two types. In it I suggest concatenating the parts as we did before instead of returning a complex object. Then we simplify the sort logic to a single line as `return aInfo.localeCompare(bInfo);`. 
Also, can we add a `image.spec.ts` for this function?

src/util/projects.tsx
```diff
@@ -48,6 +48,23 @@ export const getSubpageFromUrl = (url: string): string | undefined => {
...
+  url: string,
+  projectName: string,
+): string => {
+  const urlWithoutQuery = url.split("?")[0];
+
+  if (
+    urlWithoutQuery.includes("/ui/image-registry/") ||
+    urlWithoutQuery.includes("/ui/image-registries")
+  ) {
+    return `${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/instances`;
```
edlerd, line 61: I think in this case we can return the same url as before, keeping the user on the same page. Wdyt?

edlerd (commented):
Excellent work, this is really in good shape!

Did a very thorough review and, of course, had a thousand comments on the 1k+ line PR ;) Most of them should be quick fixes or suggestions.

src/pages/images/ImageRegistryOverview.tsx
kimanhou: It looks like the description is always an empty string for built-in registries

image

edlerd: I think in this screen we should show the `-` instead of empty. Wdyt?

kimanhou: Ah yes

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,170 @@
...
+  ["virtual-machine", "VM"],
+]);
+
+interface Props {
+  imageRegistry: LxdImageRegistry;
+}
+
+const ImageRegistryImages: FC<Props> = ({ imageRegistry }) => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const { data: images, isLoading: isImagesLoading } = useRegistryImages(
```
edlerd: We need some error handling here in case of a configuration error, this currently masks an error fetching the images.

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,189 @@
...
+  supportedArchitectures?: string[];
+}
+
+const ImageTable: FC<Props> = ({
+  images,
+  supportedArchitectures = [],
+  imageRegistryName,
+}) => {
+  const { projectName } = useCurrentProject();
+  const { data: project } = useProject(projectName);
```
edlerd:
```suggestion
  const { project, projectName } = useCurrentProject();
```

src/pages/images/ImageRegistryImages.tsx
kimanhou: What you said in stand up worked. Thank you !

src/pages/images/ImageRegistryOverview.tsx
kimanhou: Mmh are you talking about this ?

```
  const isPublic = isSimpleStreams || formik.values.cluster === "";
    const config: LxdImageRegistryConfig = {
      public: isPublic ? "true" : "false",
    };
```

src/pages/images/ImageTable.tsx
kimanhou: > so many places

I found only two: here and in local images. Where else are you thinking of ?

edlerd: Those two and the instance list also have it. And we already have a renderVolumeType but that should probably stay independent.

I searched for "VM" and found 9 matches. But some of those are also translating back, and others are about the storage volumes. So maybe not too bad as I initially thought.

src/pages/images/ImageRegistryOverview.tsx
edlerd: I meant `isImageRegistryPublic`, which actually already landed as we extracted the registry rich chip into a dedicated PR to land first. So best use it here as well.

src/util/projects.tsx
kimanhou: I don't know how to stay on the same page but update the current project

edlerd: I see, maybe that is why all the global pages skip to the instance page.

Though I wonder, with the state in the project context, can we call `setProjectName` from the context provider to update it?

src/pages/images/ImageRegistryOverview.tsx
kimanhou: Oh that's neat ! Done

src/util/projects.tsx
kimanhou: As discussed, let's do this in a separate task. I created a [JIRA ticket](https://warthogs.atlassian.net/browse/WD-36262) (but I can't find it in our backlog, I must have missed a field)

edlerd: Thanks for filing, I changed the type from epic to task and added it to the next sprint. Probably the type prevented it from showing in the backlog.

kimanhou: Aaah thank you ! Epic was selected by default after next cycle planning

src/pages/images/ImageRegistryDetail.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  if (!name) {
+    return <>Missing name</>;
+  }
+
+  const { data: imageRegistry, error, isLoading } = useImageRegistry(name);
+
+  if (isLoading) {
+    return <Spinner className="u-loader" text="Loading..." isMainComponent />;
+  } else if (!imageRegistry) {
```
omarelkashef: No need for an else here.

src/pages/images/ImageRegistryDetail.tsx
```diff
@@ -0,0 +1,68 @@
...
+    >
+      <NotificationRow />
+      <Row>
+        <TabLinks
+          tabs={tabs}
+          activeTab={activeTab}
+          tabUrl={`${ROOT_PATH}/ui/image-registry/${encodeURIComponent(name)}`}
+        />
+
+        {!activeTab && (
```
omarelkashef, line 54: Should we make this a check for `activeTab === "images"`, otherwise fall back to `configuration` so if the user manually edits the url they always good a valid screen?

src/pages/images/ImageRegistryDetailHeader.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const RenameSchema = Yup.object().shape({
+    name: Yup.string()
+      .test(
+        "deduplicate",
+        "An image registry with this name already exists",
+        async (value) =>
+          checkDuplicateName(
+            value,
+            projectName,
```
omarelkashef: You do not need the project name here since registries are a global entity.

omarelkashef (commented):
Great work, one comment and a few nitpicks. QA works well! Thank you!

src/api/image-registries.tsx
```diff
@@ -47,7 +48,7 @@ export const fetchRegistryImages = async (
...
```
edlerd:
```suggestion
      return data.metadata.sort(byOSRelease);
```
A bit simpler

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,117 @@
...
+    type: searchParams.getAll(TYPE),
+  };
+
+  const filteredImages = (images ?? []).filter((image) => {
+    // For custom registries, don't filter images
+    if (!imageRegistry.builtin) {
+      return true;
+    }
+
+    if (!image.aliases || image.aliases.length === 0) return false;
```
edlerd: This is breaking the search and filter for custom registries.

We should not filter images without aliases on non-built in registries, but the manual search and filter should still apply.

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,117 @@
...
+  if (isImagesLoading) {
+    return (
+      <Col size={9}>
+        <Spinner text="Loading images..." />
+      </Col>
+    );
+  }
+
+  if (images && images.length === 0) {
+    return <Col size={9}>No images found</Col>;
```
edlerd: Why the `Col` with size 9 here? Can this be a normal div?

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  return (
+    <>
+      <div className="u-sv3">
+        <ImagesSearchFilter
+          images={images}
+          supportedArchitectures={supportedArchitectures}
+          isImagesLoading={isImagesLoading}
+        />
+      </div>
```
edlerd: We should hide this if there are no images available. Not if they are all filtered, though, only if the registry is empty or broken.

src/pages/images/ImagesSearchFilter.tsx
```diff
@@ -0,0 +1,180 @@
...
+  supportedArchitectures: string[];
+  isImagesLoading: boolean;
+}
+
+const ImagesSearchFilter: FC<Props> = ({
+  images,
+  supportedArchitectures,
+  isImagesLoading,
+}) => {
+  const [searchParams, setSearchParams] = useSearchParams();
```
edlerd, line 35: Something is applying a sort to the selected filter criteria. They should appear in the order they are selected in by the user instead.

https://github.com/user-attachments/assets/7f7f5cbf-8a46-4010-a713-b7bd7b0d7db6

src/pages/images/ImageTable.tsx
```diff
@@ -0,0 +1,184 @@
...
+            <CreateInstanceFromImageBtn
+              key="launch"
+              projectName={projectName}
+              image={localLxdToRemoteImage(image)}
+              disabledReason={
+                !supportedArchitectures.includes(image.architecture)
+                  ? "Image is incompatible with your hardware architecture"
+                  : !isImageRegistryAllowed()
+                    ? "Images from this registry are not allowed in this project"
+                    : undefined
```
edlerd: Can we extract this to a function with early exits?

This is currently not working for a project, that is restricted, but has no registries listed. The Create button stays active instead of getting disabled.

src/pages/instances/InstanceList.tsx
```diff
@@ -119,7 +120,7 @@ const InstanceList: FC = () => {
     ),
     types: searchParams
       .getAll("type")
-      .map((value) => (value === "VM" ? "virtual-machine" : "container")),
+      .map((value) => getTypeFromDisplayName(value)),
```
```suggestion
      .map(getTypeFromDisplayName),
```

tests/images.spec.ts
```diff
@@ -75,3 +77,211 @@ test("Export and Import an image", async ({ page }) => {
...
+      const ubuntu = createRemoteImage("Ubuntu", "22.04");
+      const alpine = createRemoteImage("Alpine", "3.18");
+      const centos = createRemoteImage("CentOS", "9");
+
+      const sorted = [ubuntu, centos, alpine].sort(byOSRelease);
+
+      expect(sorted.map((img) => img.os)).toEqual([
+        "Ubuntu",
+        "CentOS",
+        "Alpine",
```
edlerd: This should be in alphabetical, not reverse alphabetical, order. Also in other tests.

For ubuntu registries we could reverse the order, but I think it is also fine to show them in alphabetical order and not have special logic for it.

edlerd (commented):
Thanks for the iteration, a few remaining issues are highlighted below.

src/pages/images/actions/DeleteImageRegistryBtn.tsx
```diff
@@ -0,0 +1,118 @@
...
+    queryClient.invalidateQueries({
+      queryKey: [queryKeys.imageRegistries],
+    });
+  };
+
+  const onSuccess = () => {
+    invalidateCache();
+
+    // Only navigate to the image registries list if we are still on the deleted registry's detail page
+    const registryDetailPath = `${ROOT_PATH}/ui/image-registry/${encodeURIComponent(imageRegistry.name)}`;
```
omarelkashef, line 54: Why is this necessary? What other pages could we be on other than registry detail or listing pages? If we are already on the listing page, does calling `navigate` does an unnecessary rerender?

src/pages/images/ImageRegistryImages.tsx
edlerd: So images from build-in registries get filtered out by default if they have no alias. But images from custom registries do not get filtered out by default. The latter should still apply the search criteria from the search and filter component.

src/pages/images/ImageRegistryDetail.tsx
kimanhou: `imageRegistry` can be `undefined` tho

src/pages/images/ImageRegistryDetail.tsx
kimanhou: I'm sorry I don't understand. The check for `activeTab === "images`" is just below. If there is no active tab, then show the overview page of the image registry. What's `configuration` ?

src/pages/images/ImageRegistryImages.tsx
kimanhou: It's to add the same margins as the spinner, so that it doesn't flicker when it goes from `Loading images` to `No images found`

src/pages/images/actions/DeleteImageRegistryBtn.tsx
kimanhou: If you are still on the registry detail page and it got deleted, then the page is not valid anymore, so you have to navigate else where. If you clicked on Delete and left the page before the deletion was complete, then there is no need to redirect you somewhere. If you are already on the listing page, then `navigate` should not be called.

src/pages/images/ImageRegistryDetail.tsx
omarelkashef: I meant it can be 

```
if (isLoading) {
    return <Spinner className="u-loader" text="Loading..." isMainComponent />;
  }
if (!imageRegistry) {...
```
since the first branch has a return

src/pages/images/actions/DeleteImageRegistryBtn.tsx
omarelkashef: I see, that makes sense.

src/pages/images/ImageRegistryDetail.tsx
omarelkashef: Yes, you are right I got confused, sorry for confusing you too.

src/pages/images/ImageRegistryImages.tsx
edlerd: Can we use `Row` instead?

src/pages/images/ImageTable.tsx
kimanhou: The issue was in `useCurrentProject`. I hope it's good now.

src/pages/images/ImageRegistryDetail.tsx
kimanhou: Ah yes indeed !

src/pages/images/ImageRegistryDetail.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return (
+      <NotFound
+        entityType="image-registry"
+        entityName={name}
+        errorMessage={error?.message}
+      />
+    );
+  }
+
+  const tabs: string[] = ["Overview", "Images"];
```
edlerd: Shall we make the Images tab the default and push overview to 2nd place. Rename it to "Configuration".  I think the image list is more important here. Wdyt?

edlerd (commented):
Mostly LGTM. Most issues are resolved, and the layout/most of the code is also great now! Two issues above are open, and one idea for the detail page is below.

src/pages/images/ImageRegistryDetailHeader.tsx
kimanhou: I'm not sure how to get the list of all available names.

src/pages/images/ImageRegistryImages.tsx
kimanhou: Should be fixed now

src/pages/images/ImageRegistryDetail.tsx
kimanhou: Sure, done

src/pages/images/ImageRegistryDetailHeader.tsx
omarelkashef: Why do you need a list of names? `checkDuplicateName` makes a GET request to the endpoint with the entity name that you give it and if the response is 404, it means the name is available.

kimanhou: Does it work if `project=""` ?

omarelkashef: Yes, the project is only used as a param in this function and is not needed for an image registry endpoint. You can check out similar logic in the creation flow.

src/pages/images/ImageRegistryImages.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const filteredImages = (images ?? []).filter((image) => {
+    // For built-in registries, filter out images without aliases
+    if (
+      imageRegistry.builtin &&
+      (!image.aliases || image.aliases.length === 0)
+    ) {
+      return false;
+    }
+
+    const technicalArchFilters = filters.arch.map(getArchitectureTechnicalName);
```
edlerd: Let's move this out from the filter loop over images.

src/pages/images/ImagesSearchFilter.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const initialized = useRef(false);
+
+  const availableArchitectures = useMemo(() => {
+    return [
+      ...new Set(
+        (images ?? [])
+          .filter((image) => image.aliases && image.aliases.length > 0)
```
edlerd:
```suggestion
```
We should consider all architectures here, no?

src/pages/images/ImagesSearchFilter.tsx
```diff
@@ -0,0 +1,185 @@
...
+          .map((image) => image.architecture),
+      ),
+    ];
+  }, [images]);
+
+  const availableTypes = useMemo(() => {
+    const uniqueTypes = [
+      ...new Set(
+        (images ?? [])
+          .filter((image) => image.aliases && image.aliases.length > 0)
```
edlerd:
```suggestion
```

tests/images.spec.ts
```diff
@@ -75,3 +77,216 @@ test("Export and Import an image", async ({ page }) => {
...
+        "Ubuntu",
+        "22.04",
+        undefined,
+        "virtual-machine",
+        "images",
+      );
+
+      const sorted = [vm, container].sort(byOSRelease);
+
+      expect(sorted[0]).toBe(vm); // "virtual-machine" comes after "container"
```
edlerd:
```suggestion
      expect(sorted[1]).toBe(vm); // "virtual-machine" comes after "container"
```
This should be the other way round.

edlerd (commented):
Nearly perfect, QA and code are great, a few minor comments.

edlerd (approved):
LGTM, thanks for driving this big PR to perfection

PR #1941, opened 2026-04-10
Instance image selector to load images from registries

edlerd:
## Done

- use the image registries images when creating a new instance
- on browsing images, we present images from all configured registries. This replaces the previously hardcoded registries in the UI for ubuntu, ubuntu-minimal and lxd-images. Instead we rely on all registries. Including the old ones and the new ones: ubuntu-daily and ubuntu-minimal-daily.
- drive by fixes to remove unused code and fix the link to the moved local images page.

## QA

1. Run the LXD-UI:
    - run with a lxd backend that supports image registries, such as side loading lxd from https://github.com/canonical/lxd/pull/17680
    - go to create instance > browse images
    - ensure the image filtering and selection works
    - ensure architecture filtering and searching works.
    - ensure images for containers and vms can be used to launch new instances.
    - create an instance from a registry
    - create an instance from an instance upload (image export + import)
    - create an instance from a locally cached image (previously used) ensure the cache is used
    - create an image from an instance, then create an instance from that custom image. ensure the notification links to the correct page


src/util/imageRegistry.ts
```diff
@@ -0,0 +1,97 @@
...
+            server: registry.config?.url,
+            title: item.os + item.release_title + item.release + item.server,
+          };
+        })
+        .sort(byOSRelease);
+    }),
+  );
+
+  // if any image request failed, throw
+  imageRequests.forEach((res) => {
```
omarelkashef, line 60: Why throw an error if any request fails? Some servers might be down and we still want to show the user the rest.

src/pages/images/ImageSelector.tsx
```diff
@@ -231,13 +195,19 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
         if (!item.cached && item.created_at) {
           source = "Local";
         }
+        if (item.registryName) {
+          source = item.registryName
+            .split("-")
```
omarelkashef: Maybe we should do this transformation on built-in registries only. If the user adds a registry "my-registry" for example, they should see it "my-registry" instead of "My Registry".

src/pages/images/ImageSelector.tsx
```diff
@@ -231,13 +195,19 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+            .join(" ");
+        }
         if (item.server === canonicalServer) {
           source = "Ubuntu";
         }
         if (item.server === minimalServer) {
           source = "Ubuntu Minimal";
         }
-        if (item.server === imagesLxdServer) {
+        if (item.server === imagesLxdServer || item.registryName === "images") {
```
omarelkashef: Why do we need to check the `registryName` here? The `server` field is populated with the url already.

omarelkashef (commented):
Good PR. I have a couple of comments/questions.

src/util/imageRegistry.ts
```diff
@@ -0,0 +1,97 @@
...
+    ...imagesByRegistry["ubuntu-minimal"],
+    ...("images" in imagesByRegistry ? imagesByRegistry["images"] : []),
+    ...imagesByRegistry["ubuntu-daily"],
+    ...imagesByRegistry["ubuntu-minimal-daily"],
+  ];
+
+  // add images from custom registries
+  Object.entries(imagesByRegistry).forEach(([registry, images]) => {
+    if (!ubuntuRegistries.includes(registry) && registry !== "images") {
+      result.push(...images);
```
omarelkashef: If we do not want to show duplicate images, we should filter here by unique title because a user can have more than one image registry point to the same server.

src/util/imageRegistry.ts
edlerd: Yes, good point. Or a user might have configured a faulty registry. Will add detailed error reporting instead.

src/pages/images/ImageSelector.tsx
edlerd: I think it is okay to convert it to a more human-friendly format as well for custom registries. No?

src/pages/images/ImageSelector.tsx
edlerd: We can slightly adjust the constant to simplify this.

src/util/imageRegistry.ts
edlerd: We can pull `24.04 LTS` from different registries, and it is a different image. Such as ubuntu-daily, ubuntu-minimal or ubuntu as a source.

We combine some images into a single entry on the built-in registries because the same alias can be used for containers and VMs. For custom servers, we can't make that assumption, so I'd keep it as is and show all images from custom registries.

src/pages/images/ImageSelector.tsx
omarelkashef: The custom registries can be of any format. It will be weird and inconsistent if we only convert ones with "-" to a human-friendly format. For example, "my-registry" will be "My Registry" but "my_registry"  and "myRegistry" will stay the same which will confuse the user.

src/util/imageRegistry.ts
omarelkashef: My suggestion was regarding two custom image registries that point to the same server, but now I agree with you because two registries might point to the same server, but pull from different projects.

src/pages/images/ImageSelector.tsx
```diff
@@ -60,56 +58,50 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+  const { isFineGrained } = useAuth();
 
   const { data: settings, isLoading: isSettingsLoading } = useSettings();
 
-  const { data: canonicalImages = [], isLoading: isCiLoading } = useQuery({
-    queryKey: [queryKeys.images, canonicalServer],
-    queryFn: async () => loadImages(canonicalJson, canonicalServer),
-    retry: false, // avoid retry to ease experience in airgapped deployments
-  });
+  const { data: legacyImages = [], isLoading: isLegacyImageLoading } = useQuery(
```
omarelkashef: We can combine this with the other query into a `useImages` function that enables the correct query behind the scenes.

omarelkashef (commented):
QA looks good, but the selector takes too long to load. I am assuming this will be fixed by requesting the backend to gzip the json file.

src/pages/images/ImageSelector.tsx
```diff
@@ -458,20 +421,32 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
-              <SearchBox
-                autoFocus
-                className="search-image"
-                name="search-image"
-                type="text"
-                onChange={(value) => {
-                  setQuery(value.toLowerCase());
-                  setOs("");
-                  setRelease("");
+            {error && !hideError ? (
```
omarelkashef, line 425: Why do we hide the searchbox if there is an error?

omarelkashef (commented):
Looks very good, only one comment below then it will be ready to merge!

src/pages/images/ImageSelector.tsx
edlerd: I chose this approach for two reasons:
1. There is not enough space to show both, especially on smaller screens
2. When the error pops up, it would push down the content

omarelkashef: I see, it is not too bad since the user still has the filter and if they close the error notification, they get the searchbox again.

omarelkashef (dismissed):
Thank you for the iterations, looks great!

edlerd:
Updated the QA steps as there were some things broken previously: instances couldn't be created when using an image registry image in the selector, and local images didn't start either. Now everything should work even when using the sideloaded image registries LXD backend.

src/context/useImages.tsx
```diff
@@ -30,3 +35,24 @@ export const useImagesInAllProjects = (
     enabled: (enabled ?? true) && isFineGrained !== null,
   });
 };
+
+export const useRemoteImages = (): UseQueryResult<RemoteImagesResult> => {
+  const { isFineGrained } = useAuth();
+  const { hasImageRegistries } = useSupportedFeatures();
+  const { project: projectName } = useParams<{ project: string }>();
```
omarelkashef: Can we use `useCurrentProject()` here?

src/pages/instances/forms/InstanceCreateDetailsForm.tsx
```diff
@@ -55,14 +53,35 @@ export const instanceDetailPayload = (values: InstanceDetailsFormValues) => {
...
+    return {
       type: "none",
       certificate: "",
       allow_inconsistent: false,
     };
   }
 
-  return payload;
+  // legacy image from hardcoded remotes
+  const serverField = hasImageRegistries ? "image_registry" : "server";
```
omarelkashef: If `hasImageregistries` is true, we will only get here if `values.image?.registryName` is empty. Is that possible? If it is not, we do not need this ternary operator.

edlerd: The comment here is off. This also applies to custom images. I'll adjust it.

edlerd: Actually custom images are handled above by fingerprint. I think this is really only for legacy images and we don't need the switch here. will remove it.

kimanhou:
I did the QA, except this step: create an instance from a locally cached image (previously used) ensure the cache is used

For create an image from an instance, then create an instance from that custom image. ensure the notification links to the correct page, it's supposed to link to the local images page, right ?

There is a failing test about storage bucket keys. It seems unrelated to this PR but I don't understand why it fails (fails to delete a pool because it's in use.)

edlerd:
> I did the QA, except this step: create an instance from a locally cached image (previously used) ensure the cache is used

Thanks, :)

> For create an image from an instance, then create an instance from that custom image. ensure the notification links to the correct page, it's supposed to link to the local images page, right ?

Yes, exactly.

> There is a failing test about storage bucket keys. It seems unrelated to this PR but I don't understand why it fails (fails to delete a pool because it's in use.)

I will take a look.

edlerd:
Problem with the test is unrelated. The test is not repeatable due to the explicit naming of the ceph object pool in it. I'll send an extra PR for that and retry it here.

kimanhou:
QA create an instance from a locally cached image (previously used) ensure the cache is used

Done

omarelkashef (approved):
QA and code look good.

PR #1945, opened 2026-04-13
WD-35829 network forwards async

kimanhou:
## Done

- fix(network forwards): rely on operation when api extension hasStorageAndNetworkOperations is present

Follow up of https://github.com/canonical/lxd/pull/18072

## QA

1. Run the LXD-UI:
    - Create, update and delete a network fowards

### DIFF
--- a/src/api/network-forwards.tsx
+++ b/src/api/network-forwards.tsx
@@ -1,6 +1,7 @@
-import { handleRawResponse, handleResponse } from "util/helpers";
+import { handleResponse } from "util/helpers";
 import type { LxdNetwork, LxdNetworkForward } from "types/network";
 import type { LxdApiResponse } from "types/apiResponse";
+import type { LxdOperationResponse } from "types/operation";
 import { addTarget } from "util/target";
 import { ROOT_PATH } from "util/rootPath";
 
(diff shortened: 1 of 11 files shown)


src/api/network-forwards.tsx
```diff
@@ -49,7 +50,7 @@ export const createNetworkForward = async (
   network: string,
   forward: Partial<LxdNetworkForward>,
   project: string,
-): Promise<string> => {
+): Promise<{ operation: LxdOperationResponse; listenAddress: string }> => {
```
edlerd:
```suggestion
): Promise<{ operation?: LxdOperationResponse; listenAddress?: string }> => {
```
Should both be optional because the older version doesn't send the operation in the response?

src/api/storage-buckets.tsx
```diff
@@ -102,12 +102,12 @@ export const updateStorageBucket = async (
   pool: string,
   project: string,
   target?: string,
-): Promise<void> => {
+): Promise<LxdOperationResponse> => {
```
edlerd, line 105: Shall we close the storage bucket pr and do both of them in one?

edlerd (commented):
Thanks for bringing this up quickly! Two questions early on.

src/api/storage-buckets.tsx
kimanhou: I opened a third one ... Yeah we can do everything in one PR if you prefer

edlerd: I think one might be easier to see the CI green in one go. Wdyt?

PR #1946, opened 2026-04-13
WD-35831 network peers async

kimanhou:
## Done

- fix(network peers): rely on operation when api extension `hasStorageAndNetworkOperations` is present

Follow up of https://github.com/canonical/lxd/pull/18073

## QA

1. Run the LXD-UI:
    - Create, update and delete a local peering
    - Create, update and delete a mutual peering

### DIFF
--- a/src/api/storage-buckets.tsx
+++ b/src/api/storage-buckets.tsx
@@ -102,12 +102,12 @@ export const updateStorageBucket = async (
   pool: string,
   project: string,
   target?: string,
-): Promise<void> => {
+): Promise<LxdOperationResponse> => {
   const params = new URLSearchParams();
   params.set("project", project);
   addTarget(params, target);
 
-  await fetch(
+  return fetch(
     `${ROOT_PATH}/1.0/storage-pools/${encodeURIComponent(pool)}/buckets/${encodeURIComponent(bucket.name)}?${params.toString()}`,
     {
       method: "PUT",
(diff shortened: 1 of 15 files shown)


src/api/storage-buckets.tsx
```diff
@@ -116,23 +116,31 @@ export const updateStorageBucket = async (
...
     },
-  ).then(handleResponse);
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
 };
 
 export const deleteStorageBucketBulk = async (
```
edlerd, line 146: I think this function also needs to be handled specially. But we can also log this as a follow up fix?

src/api/storage-buckets.tsx
```diff
@@ -245,24 +253,32 @@ export const updateStorageBucketKey = async (
...
     },
-  ).then(handleResponse);
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
 };
 
 export const deleteStorageBucketKeyBulk = async (
```
edlerd, line 284: This bulk delete also needs to wait for the operation to finish. Again we can do it as a follow up if we log a ticket for it.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -142,7 +157,7 @@ const CreateLocalPeeringPanel: FC<Props> = ({ network }) => {
         panelParams.project,
         JSON.stringify(localPeeringPayload),
       )
-        .then(() => {
+        .then((operation) => {
```
edlerd, line 160: In case we are creating a mutual peering, we have to also track this operation and report on its success. Only if succeeds, we should continue creating the other edge. On failure we have to report the error.

edlerd (approved):
Good and clean handling of the operation responses! There are three edge cases that are not handling the async nature correctly. Two about bulk deleting storage buckets and bucket keys. A third about the mutual local peering creation.

Also fine to merge as is and do those three as a follow up via logging a ticket from my side.

kimanhou:
Okay I'll tackle those three cases tomorrow.

kimanhou:
I am merging this now to fix CI. 
Follow up JIRA ticket: https://warthogs.atlassian.net/browse/WD-36053

PR #1916, opened 2026-04-02
feat: [WD-34728] Storage driver custom select

Kxiru:
## Done

- Added a Custom select dropdown list for Storage drivers
- Descriptions open to copy adjustments.
  - [x] Copies verified and cross referenced with LXD Docs team
  - [x] Updated OutputField React Component use.
- Minor amendments to tests to suppor the new design and ensure uniformity across custom creation of pools w.r.t naming.

## QA

1. Run the LXD-UI:
    - On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @Kxiru or @edlerd for access.
    - With a local copy of this branch, [build and run as described in the docs](https://github.com/canonical/lxd-ui/blob/main/CONTRIBUTING.md#setting-up-for-development).
2. Perform the following QA steps:
    - Navigate to Storage Pools > Create
    - Observe that the dropdown in the Storage pool creation page displays the Custom select dropdown and descriptions.
    - Verify that previous functionality to save the pool still works.

## Screenshots

image

Read only output field usage:
image

Help text specification for the storage drivers that require storage appliances:
image

PR #1952, opened 2026-04-14
feat: use OutputField in forms WD-35457

omarelkashef:
## Done

- Use upstreamed OutputField in forms for readonly fields.

## QA

1. Run the LXD-UI:
    - Go to the following pages and ensure the outputfield is used correctly with label and help text if possible for readonly field
        - Instances -> {instance_name} -> Configuration
        - Profiles -> {profile_name} -> Configuration
        - Networking -> Networks -> {network_name}
        - Networking -> ACLs -> {acl_name}
        - Storage -> Pools -> {pool_name} -> Configuration
        - Storage -> Volumes -> {custom_volume_name} -> Configuration
        - Configuration

## Screenshots

image
image
image
image
image
image
image


src/pages/networks/forms/NetworkFormMain.tsx
```diff
@@ -61,6 +61,21 @@ const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
...
+          <div className="general-field">
+            <div className="general-field-label">
+              <Label forId="networkName">Name</Label>
+            </div>
+            <div className="general-field-content">
+              <OutputField
+                value={formik.values.name}
+                id="networkName"
+                label=""
+                help="Click the name in the header to rename the network."
```
edlerd: Should we say "To rename this network, remove all usages"? Otherwise, we send the user on a journey to find out that the network can't be renamed in the header either, while it is in use?

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -85,24 +86,32 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
...
-            required
-          />
+          {isEdit ? (
+            <OutputField
+              id="name"
+              label="Project name"
+              value={formik.values.name}
+              help={
+                !isDefaultProject
+                  ? "Click the name in the header to rename the project."
```
edlerd, line 96: Here as well, should we tell them the project needs to be empty for renaming? Also can we disable the rename in the header, when the project is not empty? Currently we only show the error after they try to rename it.

src/pages/storage/forms/StoragePoolFormMain.tsx
```diff
@@ -112,73 +160,80 @@ const StoragePoolFormMain: FC<Props> = ({ formik }) => {
...
+              label="Driver"
+              value={formik.values.driver}
+              help="Driver can't be changed"
+            />
+          ) : (
+            <Select
+              id="driver"
+              name="driver"
+              help={
+                formik.values.driver === zfsDriver
```
edlerd: This will create a conflict with the pending PR in https://github.com/canonical/lxd-ui/pull/1916 by @kimanhou

src/pages/storage/forms/StorageVolumeFormMain.tsx
```diff
@@ -34,80 +41,103 @@ const StorageVolumeFormMain: FC<Props> = ({
...
+      return null;
+    }
+
+    if (!formik.values.isCreating) {
+      return (
+        <OutputField
+          id="clusterMember"
+          label="Cluster member"
+          value={formik.values.clusterMember}
+          help="Cluster member is immutable after creation."
```
edlerd: This is not true anymore. You can migrate the volume in a cluster with the "Migrate" button, just like for instances.

src/pages/storage/forms/StorageVolumeFormMain.tsx
```diff
@@ -34,80 +41,103 @@ const StorageVolumeFormMain: FC<Props> = ({
...
+          return {
+            label: member.server_name,
+            value: member.server_name,
+          };
+        })}
+        disabled={formik.values.isClusterMemberLocked}
+        required
+      />
+    );
+  };
```
edlerd: Shall we move this to a StorageVolumeClusterMember component?

edlerd (commented):
Good PR, some ideas to improve UX and a few other suggestions.

src/pages/networks/forms/NetworkFormMain.tsx
omarelkashef: It will make the help text too noisy. The header is very close to this help text so the user does not have to go very far and it informs the user that this network cannot be renamed while in use.

src/pages/projects/forms/ProjectDetailsForm.tsx
omarelkashef: I agree, and we should make consistent with the network one, both in the header or both in the outputfield help text.

src/pages/projects/ProjectConfigurationHeader.tsx
```diff
@@ -107,6 +107,10 @@ const ProjectConfigurationHeader: FC<Props> = ({ project }) => {
       return "Cannot rename the default project";
     }
 
+    if (project.used_by && project.used_by.length > 0) {
+      return "Cannot rename a non-empty project. Please remove all instances, profiles, and networks using this project and try again.";
```
edlerd:
```suggestion
      return "Cannot rename a non-empty project.";
```
It would be complicated to give the right list here, there are also images and volumes, but the project can also share some entities with the default project, when feature isolation is for the project is not enabled. So maybe best to keep this as a short message?

src/pages/networks/forms/NetworkFormMain.tsx
```diff
@@ -61,6 +61,21 @@ const NetworkFormMain: FC<Props> = ({ formik, project, isClustered }) => {
...
+              />
+            </div>
+          </div>
+        )}
```
edlerd: I think the output field style doesn't fit to the network form layout, lets remove it here.
```suggestion
```

edlerd (commented):
Thanks for the iteration! Two observations below.

edlerd (commented):
Project rename test seems to be conflicted with the changes.

edlerd (dismissed):
LGTM, thanks.

edlerd (approved):

Kxiru:
Merging so I can amend OutputField in my PR: https://github.com/canonical/lxd-ui/pull/1916

PR #1953, opened 2026-04-15
fix(async operations): bulk delete storage bucket & key, mutual peering

kimanhou:
## Done

- fix bulk delete of storage buckets
- fix bulk delete of storage bucket keys
- fix mutual peering

Fixes https://github.com/canonical/lxd-ui/pull/1946

## QA

1. Run the LXD-UI:
    - Bulk delete storage buckets
    - Bulk delete storage bucket keys
    - Create mutual peering

## Screenshots

image

image

image

### DIFF
--- a/src/pages/networks/panels/CreateLocalPeeringPanel.tsx
+++ b/src/pages/networks/panels/CreateLocalPeeringPanel.tsx
@@ -124,6 +124,33 @@ const CreateLocalPeeringPanel: FC<Props> = ({ network }) => {
     );
   };
 
+  const createMutualPeering = (
+    network: string,
+    project: string,
+    payload: object,
+    localPeeringName: string,
+  ) => {
+    createNetworkPeer(network, project, JSON.stringify(payload))
+      .then((mutualOperation) => {
+        if (hasStorageAndNetworkOperations) {
+          eventQueue.set(
+            mutualOperation.metadata.id,
+            () => {
+              onSuccess(localPeeringName);
+            },
+            (msg) => {
+              onFailure(false, localPeeringName, new Error(msg));
+            },
+          );
+        } else {
+          onSuccess(localPeeringName);
+        }
+      })
+      .catch((e) => {
+        onFailure(false, localPeeringName, e);
+      });
+  };
+
   const formik = useFormik<LocalPeeringFormValues>({
     initialValues: {
       name: "",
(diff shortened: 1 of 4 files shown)


src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -124,6 +124,46 @@ const CreateLocalPeeringPanel: FC<Props> = ({ network }) => {
...
+    )
+      .then((mutualOperation) => {
+        if (hasStorageAndNetworkOperations) {
+          toastNotify.info(
+            <>
+              Creation of the second half of the mutual peering{" "}
+              <ResourceLabel bold type="peering" value={values.name} /> has
+              started.
+            </>,
+          );
```
edlerd:
```suggestion
```
I wouldn't notify on this. There is no additional user interaction, and this can be perceived as one operation from the user perspective. So this is just an internal system state change we don't need to expose.

src/pages/networks/panels/CreateLocalPeeringPanel.tsx
```diff
@@ -124,6 +124,46 @@ const CreateLocalPeeringPanel: FC<Props> = ({ network }) => {
...
+    values: LocalPeeringFormValues,
+    mutualPeeringPayload: object,
+  ) => {
+    createNetworkPeer(
+      values.targetNetwork === networkOtherLabel
+        ? values.customTargetNetwork || ""
+        : values.targetNetwork,
+      values.targetProject === projectOtherLabel
+        ? values.customTargetProject || ""
+        : values.targetProject,
```
edlerd: I find this confusing. Can we store this calculation already before using it for the `localPeeringPayload`. Somethign like `const targetProject = values.targetProject === projectOtherLabel ? values.customTargetProject : values.targetProject`. And then we can do the call as `createMutualPeering(targetNetwork, targetProject, name, mutualPeeringPayload)`. In here we can change the function header similar to this:
```
  const createMutualPeering = (
    network: string,
    project: string,
    payload: object,
  ) => {
```
So the function call flips the usage from target and source network/project.
The name of the peering we can fetch inside `createMutualPeering` as `payload.name`.

edlerd (commented):
Mostly LGTM, suggestions to simplify the local peering logic and UX below.

edlerd (approved):
LGTM

PR #1954, opened 2026-04-15
feat(images) registry configuration on restricted projects

edlerd:
## Done

- add registry configuration on restricted projects

Fixes WD-34378

## QA

1. Run the LXD-UI:
    - run against a LXD backend that has sideloaded the image registries feature from https://github.com/canonical/lxd/pull/17680
    - create a new project, enable restrictions and checkout image restrictions, ensure available image registries can be selected and are persisted correctly.
    - edit a restricted project, ensure both image restrictions save correctly.

## Screenshots

image

image

### DIFF
--- a/src/pages/images/ImageRegistrySelector.tsx
+++ b/src/pages/images/ImageRegistrySelector.tsx
@@ -1,0 +1,52 @@
+import type { FC, ReactNode } from "react";
+import { MultiSelect } from "@canonical/react-components";
+import type { FormikProps } from "formik/dist/types";
+import type { ProjectFormValues } from "types/forms/project";
+import { useImageRegistries } from "context/useImageRegistries";
+
+interface Props {
+  formik: FormikProps<ProjectFormValues>;
+  help?: ReactNode;
+}
+
+const ImageRegistrySelector: FC<Props> = ({ formik, help }) => {
+  const { data: registries = [] } = useImageRegistries();
+
+  const selectedValues =
+    formik.values.restricted_registries?.split(",").filter(Boolean) ?? [];
+
+  const setValues = (values: string[]) => {
+    formik.setFieldValue(
+      "restricted_registries",
+      values.filter(Boolean).join(","),
+    );
+  };
+
+  return (
+    <div className="restricted-image-registries">
+      <MultiSelect
+        id="restricted_registries"
+        placeholder="Select registries"
+        help={help}
+        items={registries.map((registry) => {
+          return { label: registry.name, value: registry.name };
+        })}
+        selectedItems={selectedValues.map((value) => {
+          return { value: value, label: value };
+        })}
+        onDeselectItem={(item) => {
+          setValues(selectedValues.filter((value) => value !== item.value));
+        }}
+        onSelectItem={(item) => {
+          setValues(selectedValues.concat([item.value as string]));
+        }}
+        onItemsUpdate={(items) => {
+          setValues(items.map((item) => item.value as string));
+        }}
+        variant="condensed"
+      />
+    </div>
+  );
+};
+
+export default ImageRegistrySelector;
--- a/src/pages/projects/forms/ImageRestrictionForm.tsx
+++ b/src/pages/projects/forms/ImageRestrictionForm.tsx
@@ -1,0 +1,62 @@
+import type { FC, ReactNode } from "react";
+import { getConfigurationRow } from "components/ConfigurationRow";
+import ScrollableConfigurationTable from "components/forms/ScrollableConfigurationTable";
+import type {
+  ImageRestrictionFormValues,
+  ProjectFormValues,
+} from "types/forms/project";
+import type { FormikProps } from "formik/dist/types";
+import { getProjectKey } from "util/projectConfigFields";
+import type { LxdConfigPair } from "types/config";
+import ImageRegistrySelector from "pages/images/ImageRegistrySelector";
+import ResourceLink from "components/ResourceLink";
+import { ROOT_PATH } from "util/rootPath";
+
+export const imageRestrictionPayload = (
+  values: ImageRestrictionFormValues,
+): LxdConfigPair => {
+  return {
+    [getProjectKey("restricted_registries")]: values.restricted_registries,
+  };
+};
+
+interface Props {
+  formik: FormikProps<ProjectFormValues>;
+}
+
+const ImageRestrictionForm: FC<Props> = ({ formik }) => {
+  return (
+    <ScrollableConfigurationTable
+      rows={[
+        getConfigurationRow({
+          formik,
+          name: "restricted_registries",
+          label: "Available image registries",
+          defaultValue: "",
+          children: <ImageRegistrySelector formik={formik} />,
+          readOnlyRenderer: (val): ReactNode => {
+            if (val === "block" || typeof val !== "string") {
+              return val;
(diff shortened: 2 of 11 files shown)


nmezhenskyi:
Hey! `restricted.images.publish` configuration option will now be redundant, since we've decided to restrict all public images to the `default` project only after some back and forth https://github.com/canonical/lxd/pull/18052.

So I've removed the `restricted.images.publish` config key from the Image Registries PR and from the spec.

The `restricted.registries` config option remains the same.

edlerd:
> Hey! `restricted.images.publish` configuration option will now be redundant, since we've decided to restrict all public images to the `default` project only after some back and forth [canonical/lxd#18052](https://github.com/canonical/lxd/pull/18052).
> 
> So I've removed the `restricted.images.publish` config key from the Image Registries PR and from the spec.
> 
> The `restricted.registries` config option remains the same.

Thanks, I removed the publish config, and now there is only one config in the UI for images.

kimanhou:
I did the QA. I now have a project with `ubuntu-daily` and `ubuntu-minimal-daily` **unchecked**.

Should they appear in the registry list ?
image

src/pages/images/ImageRegistrySelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const setValues = (values: string[]) => {
+    formik.setFieldValue(
+      "restricted_registries",
+      values.filter(Boolean).join(","),
+    );
+  };
+
+  return (
+    <div className="restricted-cluster-groups">
```
kimanhou: Should we update the className here ? I see that restricted-image-registries is already use in `ImageRestrictionForm` and in the CSS.

```suggestion
    <div className="restricted-image-registries">
```

edlerd:
> I did the QA. I now have a project with `ubuntu-daily` and `ubuntu-minimal-daily` **unchecked**.
> 
> Should they appear in the registry list ?

Commented on the detail PR on this topic. Let's display the registry, but disable the "Create instance" button if the registry is not allowed on the current project.

src/pages/images/ImageRegistrySelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <MultiSelect
```
kimanhou, line 27: Should we add a label here ?


```suggestion
      <MultiSelect
          id="restricted_registries"
          label="Restricted image registries"
```

edlerd: The label is outside the component in the table. It is connected by the same ID in the table label that this component has: `restricted_registries`.

src/pages/images/ImageRegistrySelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const setValues = (values: string[]) => {
+    formik.setFieldValue(
+      "restricted_registries",
+      values.filter(Boolean).join(","),
+    );
+  };
+
+  return (
+    <div className="restricted-image-registries">
+      <MultiSelect
```
omarelkashef, line 27: When all registries are selected, the drop down does not work.

https://github.com/user-attachments/assets/d84ce449-3a60-45b2-92a1-f8eb6f927885

omarelkashef, line 27: Does the MultiSelect let us change the label from "Select items" to "Select registries"

src/pages/projects/forms/ImageRestrictionForm.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <ScrollableConfigurationTable
+      rows={[
+        getConfigurationRow({
+          formik,
+          name: "restricted_registries",
+          label: "Available image registries",
+          defaultValue: "",
+          children: <ImageRegistrySelector formik={formik} />,
+          readOnlyRenderer: (val): ReactNode => {
+            if (val === "block" || typeof val !== "string") {
```
omarelkashef, line 38: What does "block" mean here and in other configuration rows?

omarelkashef (commented):
Thank you for the PR. QA has one issue and just a couple of comments on code.

src/pages/images/ImageRegistrySelector.tsx
kimanhou:
```suggestion
      <MultiSelect
          placeholder="Select registries"
```

edlerd: I think there is a max width missing for the selector. Taking a look.

src/pages/projects/forms/ImageRestrictionForm.tsx
edlerd: This comes from the metadata API, see [discussion on LXD](https://github.com/canonical/lxd/pull/17680#discussion_r3085027391) about this. Block means nothing is allowed. This should be changed within the API.

omarelkashef, line 38: When is `val` not a string?

src/pages/projects/forms/ProjectForm.tsx
```diff
@@ -77,6 +79,9 @@ const ProjectForm: FC<Props> = ({
             {section === slugify(NETWORKS) && (
               <NetworkRestrictionForm formik={formik} />
             )}
+            {section === slugify(IMAGES) && (
```
omarelkashef, line 82: We might want to put these behind api extensions as well because skipping the navigation and using the url directly will show the page and with clusters, 

Image



Image

omarelkashef (commented):
Thank you for fixing the QA issues. Only two things left and it will be ready from my side.

src/pages/projects/forms/ProjectForm.tsx
edlerd: I think that it is ok to let users reach those by entering the URL. There is no natural way of navigating to it, and the forms are still working. The properties configured here would not be relevant for the installation, so it seems to be ok.

src/pages/projects/forms/ProjectFormMenu.tsx
```diff
@@ -49,10 +55,13 @@ const ProjectFormMenu: FC<Props> = ({
               className="p-side-navigation__list"
               aria-expanded={isRestrictionsOpen ? "true" : "false"}
             >
-              <MenuItem label={CLUSTERS} {...menuItemProps} />
+              {isClustered && <MenuItem label={CLUSTERS} {...menuItemProps} />}
```
omarelkashef, line 58: Actually, this change breaks a lot of unclustered tests that used the clusters tab. To me, the change is correct and these parts of the tests are the ones that should be moved to clustered tests.

src/pages/projects/forms/ImageRestrictionForm.tsx
edlerd: You are right, `val` should always be a string in this configuration. This is a type guard. The component calling the `readOnlyRenderer` has other use cases that do sometimes call it with a React node instead of a string.

edlerd: I think the type is slightly off though. The `readOnlyRenderer` can be called with a string or several object types. But it is not called with react node. Anyhow, the non-string type guard here works also in that case.

omarelkashef (approved):
QA and code look good to me. Thank you for the iterations!

PR #1957, opened 2026-04-16
refactor: rename upload image to import image WD-35583

omarelkashef:
## Done

- Rename upload image to import image to be in line with LXD.

## QA

1. Run the LXD-UI:
    - Go to Images -> Local images.
    - Ensure the import image flow is named correctly.

## Screenshots
image
image
image


src/types/upload.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface UploadState {
```
Kxiru, line 1: Should this State also be renamed to ImportState, as well as it's file name? (upload.d.ts -> import.d.ts), for consistency with the new naming.

Good idea to put this in it's own file.

Kxiru: Hmm, I've just noticed that a lot of the old file names also use Upload. How do you feel about renaming them? 😅

src/pages/images/actions/forms/ImportImageForm.tsx
Kxiru, line 50: I think the tests are failing due to the images.spec.ts test, which you did rename to "Image imported", but I'm not sure you renamed the source file, UploadImageForm.tsx where the toastNotify declares this.

src/types/upload.d.ts
omarelkashef: Yes, UploadState comes from axios and is used in multiple places and it makes sense to stay upload. Only  images has the notion of "import".

Kxiru (approved):
LGTM! Thanks for the change and test fixes.

src/pages/images/LocalImageList.tsx
```diff
@@ -81,7 +81,7 @@ const LocalImageList: FC = () => {
     },
     { content: "Type", sortKey: "type", className: "type" },
     {
-      content: "Upload date",
+      content: "Import date",
```
edlerd, line 84: This I would keep as it is aligned with the API name as "uploaded".

src/pages/images/LocalImageList.tsx
```diff
@@ -166,7 +166,7 @@ const LocalImageList: FC = () => {
         {
           content: isoTimeToString(image.uploaded_at),
           role: "cell",
-          "aria-label": "Upload date",
+          "aria-label": "Import date",
```
edlerd, line 169: Here as well, I think we should keep "Upload" here.

edlerd (commented):
Thanks for the adjustment. Two comments on the list page.

src/pages/images/LocalImageList.tsx
edlerd: Actually, I am having second thoughts. Because this also lists images fetched from a remote server. This seems like the much more frequent use case, and for that one, import makes more sense.

omarelkashef: I was not sure at first, but I ended up choosing to make it import since it is consistent with the button and the action that the user is taking. The user does not know what the api calls it unless they have used the api or CLI directly.

PR #1965, opened 2026-04-21
chore(circular dependency): increase check depth to 7 WD-35860

omarelkashef:
## Done

- increase check depth to 7.

## QA

1. Run the LXD-UI:
    - Run `npx depcruise src --max-depth 7` and make sure no errors are reported.


src/components/forms/SecurityPoliciesForm.tsx
```diff
@@ -17,30 +14,10 @@ import {
...
-      values.security_idmap_size?.toString(),
-    [getInstanceField("security_idmap_isolated")]:
-      values.security_idmap_isolated,
-    [getInstanceField("security_devlxd")]: values.security_devlxd,
-    [getInstanceField("security_devlxd_images")]: values.security_devlxd_images,
-    [getInstanceField("security_secureboot")]: values.security_secureboot,
-    [getInstanceField("security_csm")]: values.security_csm,
-  };
-};
+export { securityPoliciesPayload } from "util/instanceAndProfilePayloads";
```
edlerd:
```suggestion
```
Can we avoid such indirections in imports?

src/components/forms/SshKeyForm.tsx
```diff
@@ -22,17 +20,7 @@ import { useParams } from "react-router-dom";
...
-  const result: Record<string, string | undefined> = {};
-
-  values.cloud_init_ssh_keys?.forEach((record) => {
-    result[`cloud-init.ssh-keys.${record.name}`] =
-      `${record.user}:${record.fingerprint}`;
-  });
-
-  return result;
-};
+export { sshKeyPayload } from "util/instanceAndProfilePayloads";
```
edlerd: same as above
```suggestion
```

src/types/forms/configInheritance.d.ts
edlerd, line 1: The name is a bit generic, maybe `inhericedDevies.d.ts` or `configInheritance.d.ts` to match the util file?

src/util/formDevices.spec.ts
edlerd, line 1: The spec file should match the util file under test. When moving the functions around, we need to keep the test files in sync.

src/util/instanceEdit.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { getInstanceEditValues, getInstancePayload } from "./instanceEdit";
+import { getInstanceEditValues } from "./instanceEdit";
+import { getInstancePayload } from "util/instanceAndProfilePayloads";
```
edlerd: Here as well, the tests and utility functions under test should keep the file naming connection.

edlerd (commented):
Thanks for improving on this. A few concerns on the proposed changes below.

src/util/formDevices.spec.ts
```diff
@@ -1,15 +1,18 @@
...
-import {
-  addNicDevice,
-  addNoneDevice,
-  formDeviceToPayload,
-  parseDevices,
-  removeNicDevice,
-} from "./formDevices";
-import { objectToYaml, yamlToObject } from "./yaml";
+import { addNicDevice, addNoneDevice, removeNicDevice } from "./formDevices";
+import { parseDevices } from "./formDevices";
```
edlerd:
```suggestion
import { addNicDevice, addNoneDevice, parseDevices, removeNicDevice } from "./formDevices";
```

This should be one import. It seems like we are missing a linter for this one. Can you please file a follow up to fix the linter to catch such duplicate import lines?

edlerd (dismissed):
LGTM, thanks.

One observation on the import structure, and it seems there is still one cyclic dependency in the linter errors that needs to be fixed.

src/util/formDevices.spec.ts
omarelkashef: Yes, I noticed this problem also and already created a ticket for it.

src/util/loginProject.ts
```diff
@@ -1,7 +1,7 @@
 import type { LxdProject } from "types/project";
+import { ALL_PROJECTS } from "util/projects";
 
-export const ALL_PROJECTS = "__all_projects__";
-const LOCAL_STORAGE_KEY = "lxdLoginProject";
+export const LOCAL_STORAGE_KEY = "lxdLoginProject";
```
edlerd, line 4: I don't think we need to export this.

tests/login-project.spec.ts
edlerd, line 1: I would rather have an end to end test, that creates a project, sets that project as the login project and then opens `/ui` and ensures we land in the right project.

That test should be shorter than the test proposed here, and it will cover the whole flow including integration, not just the util part.

src/util/instanceAndProfilePayloads.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+  "  prometheusnat:\n" +
+  "    connect: tcp:10.76.171.21:9090\n" +
+  "    listen: tcp:192.168.0.90:9090\n" +
+  "    nat: 'true'\n" +
+  "    type: proxy\n" +
+  "  foobar:\n" +
+  "    id: ababab\n" +
+  "    gputype: physical\n" +
+  "    type: gpu\n" +
+  "";
```
edlerd, line 209: This is duplicating the other test, can we reuse instead?

edlerd (commented):
Thanks for the adjustment, few concerns are remaining below.

src/util/instanceAndProfilePayloads.spec.ts
omarelkashef: We can reuse it, but this is more of an accidental duplication. The two instances are used in different tests and they do not have to be the same (they happen to be for now and the tests should run fine if they are not in the future). I think it would couple the tests if we reuse it. Where do you suggest we place the common yaml?

src/util/loginProject.ts
omarelkashef: It is needed for the unit tests in `util/loginProject.spec.ts`.

src/util/instanceAndProfilePayloads.spec.ts
edlerd: Yeah, true coupling them to a common constant seems unexpected. Let's keep as is.

edlerd (dismissed):
LGTM

edlerd (approved):
LGTM

PR #1966, opened 2026-04-22
feat: [WD-31002] Add Dell PowerStore storage driver

Kxiru:
## Done

- Draft PR to add the Dell PowerStore driver. Pending https://github.com/canonical/lxd/pull/18009 merge.

## QA

1. Run the LXD-UI:
    - Checkout the PR locally
    - Attempt to create a Powerstore Pool (this will fail)
    - Ensure the network request supports the appropriate payload for Powerstore: https://github.com/canonical/lxd/pull/18009

## Screenshots

image
image


src/pages/storage/forms/StorageDriverSelect.tsx
edlerd, line 1: There is a bug in this component leading to test failures. The list is very long now and test are hitting a condition where the last element in the list is behind the footer. We need to patch the dropdown to be above the footer instead:

Image

src/util/storageOptions.tsx
```diff
@@ -35,6 +37,7 @@ export const storageDriverDescriptions: { [key: string]: string } = {
   [dirDriver]: "Basic local directory (no native snapshots or quotas)",
   [lvmDriver]: "Logical volume-backed block storage with thin provisioning",
   [powerFlex]: "Dell PowerFlex software-defined block storage",
+  [powerStore]: "Dell PowerStore software-defined block storage",
```
edlerd: Did you receive this description or come up with it yourself? If the latter, please verify.

edlerd (commented):
Mostly LGTM, thanks for bringing this in. One observation and one issue that needs fixing below.

src/util/storageOptions.tsx
Kxiru: Verifyinggg~

src/pages/storage/forms/StorageDriverSelect.tsx
Kxiru: Is putting it over the notification bar what we want to do? Perhaps we can set a max height and scroll within it instead.

edlerd: It already adjusts drop-down height depending on screen dimensions and where the toggle is on the screen. It also starts scrolling if there is too few available space. Problem is that the bottom part of it is covered by the status bar.

edlerd: Actually you can see in the screenshot above that the options are scrollable.

Kxiru: Yes but the height of the options is the problem, so we can make the height a generally good height like 40vh so that it doesn't intersect on normal devices.
I've messed around with both options. The 40vh works nicely. ~~But to bring the dropdown above the footer using something like z-index, we would need to change the dropdown to be positioned.~~

Edit: It's not that my z-index method wasn't working, it's that it just wasnt high enough lol. All solved!

edlerd, line 1: Shall we remove the search from the custom select?

I think it might be beneficial to hide the search, as the options are fairly limited in amount. There is a prop to disable it. Or if we keep the search, we should make it match both the name and description.

src/util/storageOptions.tsx
```diff
@@ -54,8 +57,13 @@ export const isClusterWideSourceDriver = (driver: string): boolean => {
...
 ): CustomSelectOption[] => {
-  const serverSupportedStorageDrivers =
-    settings?.environment?.storage_supported_drivers || [];
+  // const serverSupportedStorageDrivers =
+  //   settings?.environment?.storage_supported_drivers || [];
+
+  const serverSupportedStorageDrivers = [
+    settings?.environment?.storage_supported_drivers || [],
+    { Name: powerStore, Version: "N/A", Remote: true },
+  ].flat();
```
edlerd: This should be reverted before merge.

edlerd (dismissed):
Perfect, thanks for fixing the z-index with the status bar! Code and QA LGTM as is. One comment on the search, and of course the removal of the hack to make it QA-able.

MusicDin (commented):
Thanks, already looking good.
Below are some comments from my side.

---

I see PowerStore twice in the menu where selecting storage pool driver:

Image

---

Once storage pool is created `source` appears in the storage pool configuration. Just opening editor and clicking save without doing any changes results in an error due to `source` being present.

Image

Image

---

It would be nice to be able to select `Mode` in the initial creation screen. Also, may I ask from where the values for Modes are sourced?

Image

---

Field `transport` was dropped. We will have all values in the `mode`, which will combine the transport in its name (for example `nvme-fc`).

Image


---

Otherwise, using created storage pool it was easy to create instances, snapshots, custom volumes (fs and block), restore from snapshots etc.

Thanks again 🚀

edlerd:
Thanks for the feedback @MusicDin  :)

> I see PowerStore twice in the menu where selecting storage pool driver:

This is for development purposes. We temporarily force the power store option, even if the server does not support it. You run with a server that supports powerstore, so you get it twice. This will be cleaned up before merge. See [this comment](https://github.com/canonical/lxd-ui/pull/1966#discussion_r3129243525).

@Kxiru the other points above seem like things that need to be iterated on.

MusicDin:
>> I see PowerStore twice in the menu where selecting storage pool driver:

> This is for development purposes. We temporarily force the power store option, even if the server does not support it. You run with a server that supports powerstore, so you get it twice. This will be cleaned up before merge. See https://github.com/canonical/lxd-ui/pull/1966#discussion_r3129243525.

Thanks for clarification 🚀

Kxiru:
@MusicDin 

> It would be nice to be able to select Mode in the initial creation screen. Also, may I ask from where the values for Modes are sourced?

This is defined client side in instanceOptions.tsx, lines 124-138. The values are as follows:

image

I _could_ move Mode to the initial creation screen, however for most other storage pools this is a secondary field and is usually put in the secondary menu. This selector seems to match the same values as previous drivers. Is Mode necessarily different or more important for Powerstore?

> Field transport was dropped. We will have all values in the mode, which will combine the transport in its name (for example nvme-fc).

Ah, this is important. please could you point me to which values should be within the mode? I will remove all transport related configs now.  I have now added the Fibre channel config option to the mode dropdown but it would be great to double check.

> Once storage pool is created source appears in the storage pool configuration. Just opening editor and clicking save without doing any changes results in an error due to source being present.

Looking into this now...

MusicDin:
> please could you point me to which values should be within the mode?

Currently just `iscsi`, but I will let you know once we can land support for more. 
We are planning to eventually support: `iscsi`, `nvme-tcp`, `nvme-fc`, `scsi-fc`.

Kxiru:
> Currently just `iscsi`, but I will let you know once we can land support for more. We are planning to eventually support: `iscsi`, `nvme-tcp`, `nvme-fc`, `scsi-fc`.

I will put these into the PR provisionally. @MusicDin please could you confirm that these will likely be the correct values?

```
export const optionPowerStoreMode = [
  {
    label: "Select option",
    value: "",
    disabled: true,
  },
  {
    label: "iSCSI over TCP",
    value: "iscsi",
  },
  {
    label: "iSCSI over FC",
    value: "iscsi-fc",
  },
  {
    label: "NVMe over TCP",
    value: "nvme",
  },
  {
    label: "NVMe over FC",
    value: "nvme-fc",
  },
];
```

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -119,6 +123,18 @@ export const toStoragePool = (
         [getPoolKey("powerflex_user_password")]: values.powerflex_user_password,
       };
     }
+    if (isPowerStoreDriver) {
+      return {
+        [getPoolKey("powerstore_gateway")]: values.powerstore_gateway,
+        [getPoolKey("powerstore_gateway_verify")]:
+          values.powerstore_gateway_verify,
+        [getPoolKey("powerstore_mode")]: values.powerstore_mode,
+        [getPoolKey("powerstore_target")]: values.powerstore_target,
```
edlerd:
```suggestion
```
Target should be removed here and in other places.

src/util/storageOptions.tsx
edlerd: Let's keep it open until resolved.

src/util/instanceOptions.tsx
```diff
@@ -137,6 +137,30 @@ export const optionIscsiNvme = [
...
+  {
+    label: "NVMe over TCP",
+    value: "nvme",
+  },
+  {
+    label: "NVMe over FC",
+    value: "nvme-fc",
+  },
+];
+
```
edlerd, line 151: @MusicDin is this the correct list of modes?

src/util/storageOptions.tsx
edlerd: Powerstore is not software defined, maybe `Dell PowerStore block and file storage`?

edlerd (commented):
Thanks for the iteration, some observations below.

src/util/storageOptions.tsx
Kxiru: I feel like you are reviewing an older version of the branch... I have already made all of these changes + the target changes... Will push again...

Kxiru: All done!

edlerd: This is still unchanged as `Dell PowerStore software-defined block storage`

src/util/storagePool.tsx
```diff
@@ -190,3 +214,8 @@ export const isCephDriver = (values: StoragePoolFormValues) => {
 export const isCephFSDriver = (values: StoragePoolFormValues) => {
   return values.driver === cephFSDriver;
 };
+
+export const hasSource = (driver: string): boolean => {
+  const driversWithSource = [btrfsDriver, dirDriver, lvmDriver, zfsDriver];
+  return driversWithSource.includes(driver);
+};
```
edlerd: This is a bit more complex, we need to check the api extension:
```
  const { hasRemoteDropSource } = useSupportedFeatures();
```
If this extension is not present, the list also includes `cephDriver` and `cephFSDriver`. Best pass this extension into the function as a parameter.

edlerd (commented):
Thanks for the updates and extracting the has source into a single source of truth. Two remaining concerns below.

src/util/instanceOptions.tsx
MusicDin: I suggest keeping only `iSCSI` for now, as it is the only one supported.
```
  {
    label: "SCSI over TCP",
    value: "iscsi",
  },
```

I will keep you updated before new modes become available.

This will be the first driver to support fiber channel and now we want to encode the transport into the names. As a result we will change the mode names (and either patch old names or leave them as aliases to new ones), but is not yet agreed.

So the actual **unconfirmed** list of modes:
- `scsi-tcp` or `iscsi` - SCSI over TCP
- `scsi-fc` or `fc` - SCSI over FC (not iSCSI, `i` stands for internet and uses IP/TCP)
- `nvme-tcp` - NVMe over TCP
- `nvme-fc`  - NVMe over FC

src/util/storageOptions.tsx
Kxiru: I must be going insane

src/util/instanceOptions.tsx
Kxiru: @edlerd , as the namings of the modes are unconfirmed, I will remove all but iscsi.

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -62,6 +69,7 @@ export const toStoragePool = (
   hasRemoteDropSource?: boolean,
```
edlerd:
```suggestion
  hasRemoteDropSource: boolean,
```
There is an existing bug that surfaces now: this should not be nullable. The bug is that some callers of `toStoragePool` do not pass a second argument at all. All callers should be adjusted to fetch the api extension and pass it into the function on calling it.

In the spec tests calling `toStoragePool`, we can provide an arbitrary `false` value to fix them.

edlerd (commented):
LGTM code and QA wise, one relevant bug is surfaced. Can you fix it please, then we can merge it.

src/pages/storage/forms/StoragePoolForm.tsx
edlerd: @Kxiru we can't set all of them to false, only in tests. In real components we have to use the actual value.

edlerd (changes requested):

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -184,7 +203,9 @@ export const toStoragePool = (
     config: {
       ...missingConfigFields,
       ...getConfig(),
-      source: values.driver !== btrfsDriver ? values.source : undefined,
+      source: hasSource(values.driver, hasRemoteDropSource ?? false)
```
edlerd:
```suggestion
      source: hasSource(values.driver, hasRemoteDropSource)
```
We don't need the fallback here now with the strict type.

src/pages/storage/forms/StoragePoolForm.tsx
```diff
@@ -205,7 +226,7 @@ const StoragePoolForm: FC<Props> = ({
   useListener(window, updateFormHeight, "resize", true);
 
   const getYaml = () => {
-    const payload = toStoragePool(formik.values);
+    const payload = toStoragePool(formik.values, false);
```
edlerd: We shouldn't hardcode the `false` value here, but use the dynamic api extension.

src/pages/storage/EditStoragePool.tsx
```diff
@@ -141,7 +141,7 @@ const EditStoragePool: FC<Props> = ({ pool }) => {
     onSubmit: (values) => {
       const savedPool = values.yaml
         ? (yamlToObject(values.yaml) as LxdStoragePool)
-        : toStoragePool(values);
+        : toStoragePool(values, false);
```
edlerd: We shouldn't hardcode the `false` value here, but use the dynamic api extension.

edlerd (changes requested):

src/pages/storage/EditStoragePool.tsx
Kxiru: Sorry David, I don't think my git is pushing properly, I have already made all these changes.
image

src/pages/storage/forms/StoragePoolForm.tsx
Kxiru: Made this change :+1:

edlerd (dismissed):
LGTM, thanks for adding :)

Kxiru:
@edlerd , do you mind re-approving this PR, please? I pushed a lint fix.

edlerd (approved):

PR #1968, opened 2026-04-22
Ux fixes

edlerd:
## Done

- fix(nav) ensure custom iso appear next to images in side navigation 
- fix(instance) disable boot mode configuration on containers 
- fix(forwards) positive appearance on network forward create button 

## QA

1. Run the LXD-UI:
    - check side navigation storage section has the custom ISOs as last entry
    - check instance configuration on a container, the boot > boot mode option should be disabled for override and marked as for vms only
    - check instance configuration on a container, the boot > boot mode option should be enabled, and a override should save properly and persist a reload
    - create a network forward networks > any bridge network > forwards > create. Ensure the form has a green/positive button to submit
    - browse a running instance console, ensure the border color matches the background. check the terminal for instances has the proper ubuntu theme.

## Screenshots

image

image

image

### DIFF
--- a/src/components/forms/BootForm.tsx
+++ b/src/components/forms/BootForm.tsx
@@ -6,6 +6,7 @@ import { getConfigurationRow } from "components/ConfigurationRow";
 import ScrollableConfigurationTable from "components/forms/ScrollableConfigurationTable";
 import { optionRenderer } from "util/formFields";
 import { useSupportedFeatures } from "context/useSupportedFeatures";
+import type { CreateInstanceFormValues } from "types/forms/instanceAndProfile";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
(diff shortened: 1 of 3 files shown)


src/components/forms/BootForm.tsx
```diff
@@ -55,10 +61,19 @@ const BootForm: FC<Props> = ({ formik }) => {
           ? [
               getConfigurationRow({
                 formik,
-                label: "Boot mode",
+                label: "Boot mode (VMs only)",
```
kimanhou, line 64: Should we display (VMs only) for a VM ?

kimanhou (commented):
Could you rebase please ?

image

Just a tiny thing, otherwise QA and code all good.

src/components/forms/BootForm.tsx
edlerd: We do this as well for other options in the "Security" section. I think it is valid information that this config is only for VMs. Independent of whether you are browsing the configuration for a container or a VM.

PR #1972, opened 2026-04-24
WD-34376 image registry edit

omarelkashef:
## Done

- Image registry edit panel

## QA

1. Run the LXD-UI:

## Screenshots

image


src/pages/images/ImageRegistriesList.tsx
```diff
@@ -49,6 +51,9 @@ const ImageRegistriesList: FC = () => {
       content: "Public",
       sortKey: "public",
     },
+    {
+      content: "Actions",
```
edlerd: This should be right aligned and without a visible heading. See other lists with similar setup.

edlerd: Actually, I am not sure if we need the action column here. It seems ok to only edit and delete on the detail page. Both seem like rare occasions, and it's better to do them on the detail page, where the user has full context. Both are expected to have low-frequency usages, and we want to avoid mixing them up, such as when deleting the wrong entry. Wdyt?

src/pages/images/ImageRegistryForm.tsx
```diff
@@ -1,36 +1,59 @@
...
+          id="Name"
+          label="Name"
+          value={formik.values.name}
+          help={<>Image registry can be renamed in the {link}.</>}
+        />
+      ) : (
+        <Input
+          {...formik.getFieldProps("name")}
+          type="text"
+          autoFocus={!isEdit}
```
edlerd:
```suggestion
          autoFocus
```
When this is rendered, it should focus.

src/pages/images/ImageRegistryForm.tsx
```diff
@@ -1,36 +1,59 @@
...
+  const isEdit = !formik.values.isCreating;
 
   const getFieldError = (fieldName: keyof ImageRegistryFormValues) => {
     return formik.touched[fieldName] ? formik.errors[fieldName] : undefined;
   };
 
+  const link = (
+    <Link to={`${ROOT_PATH}/ui/image-registry/${formik.values.name}`}>
+      {" "}
+      detail page
```
edlerd:
```suggestion
      detail page header
```
Better give clear instructions. Actually if we remove edit from the list view, we can simplify and say "Click the image registry name in the header to rename the registry".

src/pages/images/ImageRegistryForm.tsx
```diff
@@ -39,7 +62,16 @@ export const ImageRegistryForm: FC<Props> = ({ formik, isEdit = false }) => {
         placeholder="Enter description"
         error={getFieldError("description")}
       />
-      <ImageRegistryProtocolSelector formik={formik} />
+      {isEdit ? (
+        <OutputField
+          id="Protocol"
+          label="Protocol"
+          value={formik.values.protocol}
+          help="Protocol cannot be changed after creation."
```
edlerd:
```suggestion
          help="Protocol cannot be changed."
```
Simplify

src/pages/images/ImageRegistriesList.tsx
```diff
@@ -22,6 +22,8 @@ import { isImageRegistryPublic } from "util/image-registries";
 import { CreateImageRegistryButton } from "./actions/CreateImageRegistryButton";
 import usePanelParams, { panels } from "util/usePanelParams";
 import { CreateImageRegistryPanel } from "./panels/CreateImageRegistryPanel";
+import { ImageRegistryActions } from "./actions/forms/ImageRegistryActions";
+import { EditImageRegistryPanel } from "./panels/EditImageRegistryPanel";
 
 const ImageRegistriesList: FC = () => {
```
edlerd, line 28: I would also like to see the URL or the cluster+project in the list, depending on the chosen protocol. Though both can be fairly long, so I don't have a direct idea how to display them nicely.

Maybe switch the protocol and description column order, then make the description also contain those details in a dedicated line. Or perhaps you have a better idea. Wdyt?

edlerd (commented):
Thanks, this works! Some more ideas to refine the UX are below.

src/pages/images/ImageRegistriesList.tsx
omarelkashef: If we add the URL and source project, the user will have all the context they need, but I agree that it's rare that they need to take action on a registry, so it would be better to not clutter this table.

src/pages/images/actions/DeleteImageRegistryBtn.tsx
```diff
@@ -72,10 +72,10 @@ const DeleteImageRegistryDeleteBtn: FC<Props> = ({ imageRegistry }) => {
 
   const disabledReason = () => {
     if (imageRegistry.builtin) {
-      return "Built-in image registries cannot be deleted";
+      return "Built-in image registries cannot be deleted.";
     }
     if (!canDeleteImageRegistry(imageRegistry)) {
-      return "You do not have permission to delete this image registry";
+      return "You do not have permission to delete this image registry.";
```
edlerd: We don't usually end the disable reason title with a dot. Should this be reverted?

src/pages/images/actions/EditImageRegistryButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  imageRegistry: LxdImageRegistry;
+}
+
+const EditImageRegistryButton: FC<Props> = ({ imageRegistry }) => {
+  const { openEditImageRegistry } = usePanelParams();
+  const { canEditImageRegistry } = useImageRegistriesEntitlements();
+  const isSmallScreen = useIsScreenBelow();
+
+  const isDisabled =
+    imageRegistry.builtin || !canEditImageRegistry(imageRegistry);
```
edlerd:
```suggestion
    disabledReason() === undefined;
```
Better use a single source of truth for it instead of duplicating the logic here..

src/pages/images/panels/EditImageRegistryPanel.tsx
```diff
@@ -0,0 +1,143 @@
...
+    if (isSimpleStreams) {
+      config.url = formik.values.url;
+    } else {
+      config.cluster = formik.values.cluster;
+      config.source_project = formik.values.sourceProject;
+      config.url = formik.values.url;
+    }
+    payload.config = config;
+    return payload;
+  };
```
edlerd, line 59: We can simplify this by building the config first and then returning the payload object. Probably worth doing the same in the create panel.

```
  const getPayload = () => {
    const isSimpleStreams = formik.values.protocol === "simplestreams";
    const isPublic = isSimpleStreams || formik.values.cluster === "";
    const config: LxdImageRegistryConfig = {
      public: isPublic ? "true" : "false",
    };

    if (isSimpleStreams) {
      config.url = formik.values.url;
    } else {
      config.cluster = formik.values.cluster;
      config.source_project = formik.values.sourceProject;
      config.url = formik.values.url;
    }
    return {
      description: formik.values.description,
      config,
    } as Partial<LxdImageRegistry>;
  };
```

src/pages/images/ImageRegistryProtocol.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface Props {
+  imageRegistry: LxdImageRegistry;
+}
+
+export const ImageRegistryProtocol: FC<Props> = ({ imageRegistry }) => {
+  const isSimpleStreams = imageRegistry.protocol === "simplestreams";
+  const url = imageRegistry.config?.url ?? "";
+  const sourceProject = imageRegistry.config?.source_project ?? "";
+  const cluster = imageRegistry.config?.cluster ?? "";
+  const source = isSimpleStreams ? url : `${cluster}/${sourceProject}`;
```
edlerd: I am not sure if the `cluster/project` formatting is understandable. But I also have no better idea for it.

edlerd (commented):
Thanks for the iteration, mostly LGTM, some small suggestions below.

src/pages/images/panels/CreateImageRegistryPanel.tsx
edlerd, line 1: Found an edge case with the cluster selector involving long names on the cluster link. Can we add truncation and a title to it?

Image

edlerd (commented):
QA and code LGTM, found one related edge case on long cluster link names. Can you fix it as a drive by?

edlerd (approved):
LGTM, thanks :)

PR #1985, opened 2026-04-29
feat(test) add image registry tests WD-35861

omarelkashef:
## Done

- Add image registry tests.

## QA

1. Run the LXD-UI:
    - Switch to a version of lxd that supports image registries.
    - Update the conditon in `skipIfNotSupported` in tests/helpers/image-registries.ts to enable the tests for the lxd version you are on.
    - Run unclustered and clustered tests with `npx playwright test tests/image-registries.spec.ts --project chromium:lxd-latest-edge:unclustered` and `npx playwright test tests/image-registries-clustered.spec.ts --project chromium:lxd-latest-edge:clustered`, respectively after replacing `lxd-latest-edge` with your lxd version.
    - Ensure all tests pass.

### DIFF
--- a/src/pages/images/panels/CreateImageRegistryPanel.tsx
+++ b/src/pages/images/panels/CreateImageRegistryPanel.tsx
@@ -59,7 +59,7 @@ export const CreateImageRegistryPanel: FC = () => {
     } else {
       config.cluster = formik.values.cluster;
       config.source_project = formik.values.sourceProject;
-      config.url = formik.values.url;
+      config.url = "https://cloud-images.ubuntu.com/daily/";
     }
 
     return {
--- a/tests/helpers/image-registries.ts
+++ b/tests/helpers/image-registries.ts
@@ -2,6 +2,8 @@ import { expect, test, type LxdVersions } from "../fixtures/lxd-test";
 import type { Page } from "@playwright/test";
 import { gotoURL } from "./navigate";
 import { randomNameSuffix } from "./name";
+import { dismissNotification } from "./notification";
+import { searchEntityListPage } from "./search";
 
 export const randomImageRegistryName = () => {
   return `playwright-image-registry-${randomNameSuffix()}`;
@@ -19,4 +21,63 @@ export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
 export const visitImageRegistries = async (page: Page) => {
   await gotoURL(page, `/ui/image-registries`);
   await expect(page.getByTitle("Create registry")).toBeVisible();
+};
+
+export const visitImageRegistry = async (page: Page, name: string) => {
+  await visitImageRegistries(page);
+  await searchEntityListPage(page, name);
+  await page
+    .getByRole("row")
+    .filter({ hasText: name })
+    .getByRole("rowheader", { name: "Name" })
+    .getByRole("link")
+    .click();
+  await expect(
+    page.getByRole("button").filter({ hasText: "Edit registry" }),
+  ).toBeVisible();
+};
+
+export const createImageRegistry = async (
+  page: Page,
+  name: string,
+  protocol: "SimpleStreams" | "LXD",
+  config: { url?: string; cluster?: string; sourceProject?: string } = {},
+) => {
+  await visitImageRegistries(page);
+  await page.getByTitle("Create registry").click();
+  const sidePanel = page.getByLabel("Side panel");
+  await sidePanel.getByPlaceholder("Enter name").fill(name);
+  await sidePanel.getByRole("radio", { name: protocol }).check({ force: true });
+
+  if (config.url) {
+    await sidePanel.getByLabel("Server").fill(config.url);
+  }
+  if (config.cluster) {
+    await sidePanel
+      .getByLabel("Cluster")
+      .selectOption({ label: config.cluster });
+  }
+  if (config.sourceProject) {
+    await sidePanel.getByLabel("Source project").fill(config.sourceProject);
+  }
+  await sidePanel.getByRole("button", { name: "Create", exact: true }).click();
+  await dismissNotification(page, `Image registry ${name} created.`);
+};
+
+export const deleteImageRegistry = async (page: Page, name: string) => {
+  await visitImageRegistry(page, name);
+  await page.getByRole("button", { name: "Delete registry" }).click();
+  await page
+    .getByRole("dialog", { name: "Confirm delete", exact: true })
+    .getByRole("button", { name: "Delete registry", exact: true })
+    .click();
+  await dismissNotification(page, `Image registry ${name} deleted.`);
+};
+
+export const openImageRegistryEditPanel = async (page: Page, name: string) => {
+  await visitImageRegistry(page, name);
+  await page.getByRole("button").filter({ hasText: "Edit registry" }).click();
+  await page
+    .getByRole("heading", { name: "Edit image registry", exact: true })
+    .isVisible();
 };
--- a/tests/helpers/image-registries.ts
+++ b/tests/helpers/image-registries.ts
@@ -20,7 +20,7 @@ export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
 
 export const visitImageRegistries = async (page: Page) => {
   await gotoURL(page, `/ui/image-registries`);
-  await expect(page.getByTitle("Create registry")).toBeVisible();
+  await expect(page.getByText("Create registry")).toBeVisible();
 };
 
 export const visitImageRegistry = async (page: Page, name: string) => {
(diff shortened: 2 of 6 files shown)


src/pages/images/actions/CreateImageRegistryButton.tsx
```diff
@@ -18,7 +18,7 @@ export const CreateImageRegistryButton: FC = () => {
       title={
         isDisabled
           ? "You don't have permissions to create image registries"
-          : undefined
+          : "Create registry"
```
edlerd: The caption is always visible and not truncated. In that case we should not duplicate it here in the title.

src/pages/images/panels/CreateImageRegistryPanel.tsx
```diff
@@ -59,7 +59,7 @@ export const CreateImageRegistryPanel: FC = () => {
     } else {
       config.cluster = formik.values.cluster;
       config.source_project = formik.values.sourceProject;
-      config.url = formik.values.url;
+      config.url = "https://cloud-images.ubuntu.com/daily/";
```
edlerd, line 62:
```suggestion
```
I think this should be completely removed. Can you validate with Nikita if it has been fixed upstream to not be necessry?

tests/helpers/image-registries.ts
```diff
@@ -20,3 +22,62 @@ export const visitImageRegistries = async (page: Page) => {
...
+export const createImageRegistry = async (
+  page: Page,
+  name: string,
+  protocol: "SimpleStreams" | "LXD",
+  config: { url?: string; cluster?: string; sourceProject?: string } = {},
+) => {
+  await visitImageRegistries(page);
+  await page.getByTitle("Create registry").click();
+  const sidePanel = page.getByLabel("Side panel");
+  await sidePanel.getByPlaceholder("Enter name").fill(name);
```
edlerd: can we use get by label instead?

tests/helpers/projects.ts
```diff
@@ -18,7 +18,6 @@ const openProjectCreationForm = async (page: Page) => {
   await expect(page.getByText("Loading storage pools")).not.toBeVisible();
 
   await page.getByLabel("Default instance network", { exact: true }).click();
-  await page.getByText("No network", { exact: true }).click();
```
edlerd: Why this change? Seems unrelated and should probably be restored.

tests/image-registries.spec.ts
```diff
@@ -43,54 +54,172 @@ test("search for built-in image registries", async ({ page, lxdVersion }) => {
...
   );
+
+  await deleteImageRegistry(page, registryName);
+});
+
+test("view image registry detail page", async ({ page, lxdVersion }) => {
+  skipIfNotSupported(lxdVersion);
+
+  const builtinRegistryName = "ubuntu-daily";
+  const url = "https://cloud-images.ubuntu.com/daily/";
```
edlerd: Can we make this file level constants and reuse them between tests?

tests/image-registries.spec.ts
```diff
@@ -43,54 +54,172 @@ test("search for built-in image registries", async ({ page, lxdVersion }) => {
...
-  await expect(page.getByLabel("Cluster link")).not.toBeVisible();
-  await page.getByLabel("Server URL").fill(url);
+    .check({ force: true });
+  await expect(sidePanel.getByLabel("Source project")).not.toBeVisible();
+  await expect(sidePanel.getByLabel("Cluster")).not.toBeVisible();
+  await sidePanel.getByLabel("Server").fill(url);
 
-  await page.getByRole("button", { name: "Create" }).click();
+  await sidePanel.getByRole("button", { name: "Create", exact: true }).click();
+  await dismissNotification(page, `Image registry ${registryName} created.`);
```
edlerd: Can we use the create registry helper instead of duplicating the creation process?

tests/image-registries-clustered.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  await sidePanel.getByLabel("Description").fill("Playwright LXD registry");
+  await sidePanel.getByRole("radio", { name: "LXD" }).check({ force: true });
+
+  await expect(sidePanel.getByLabel("Server")).not.toBeVisible();
+  await sidePanel.getByLabel("Source project").fill(projectName);
+  await sidePanel.getByLabel("Cluster").click();
+  await page.getByTitle(clusterName).click();
+
+  await sidePanel.getByRole("button", { name: "Create", exact: true }).click();
+  await dismissNotification(page, `Image registry ${registryName} created.`);
```
edlerd: We should reuse the create registry helper here as well.

tests/image-registries.spec.ts
```diff
@@ -43,54 +54,172 @@ test("search for built-in image registries", async ({ page, lxdVersion }) => {
...
+
+  await expect(
+    page.getByRole("link", { name: "Images", exact: true }),
+  ).toBeVisible();
+  await page.getByTestId("tab-link-Configuration").click();
+
+  const panel = page.getByRole("tabpanel", { name: "Configuration" });
+  await expect(panel.locator("tr").filter({ hasText: "Name" })).toContainText(
+    builtinRegistryName,
+  );
```
edlerd: Let's extract this into a helper such as `validateRegistryRow(page, registry, "Name", "ubuntu")`

tests/image-registries.spec.ts
```diff
@@ -43,54 +54,172 @@ test("search for built-in image registries", async ({ page, lxdVersion }) => {
...
+  const registryName = randomImageRegistryName();
+  await createImageRegistry(page, registryName, "SimpleStreams", {
+    url: "https://cloud-images.ubuntu.com/releases/",
+  });
+  const row = page.getByRole("row").filter({ hasText: registryName });
+  await expect(row).toBeVisible();
+
+  await deleteImageRegistry(page, registryName);
+  await expect(row).not.toBeVisible();
+});
```
edlerd: We don't need three test cases for this. We can have one for create/edit/delete test.

tests/image-registries.spec.ts
```diff
@@ -43,54 +54,172 @@ test("search for built-in image registries", async ({ page, lxdVersion }) => {
...
+
+  await activateOverride(page, "Available image registries");
+
+  await page
+    .getByRole("row")
+    .filter({ hasText: "Available image registries" })
+    .getByLabel("Select registries")
+    .click();
+
+  await page.getByRole("button", { name: "Clear", exact: true }).click();
```
edlerd: This should not be needed, the list will be empty by default.
```suggestion
```

edlerd (commented):
Thanks for the extended tests! Some observations below.

tests/helpers/image-registries.ts
omarelkashef: The labels also match with the registries list cells.

tests/helpers/projects.ts
omarelkashef: The function is to open the project creation form. Choosing "No network" is an unexpected side effect, and "No network" is the default anyways so we are changing the functionality.

src/pages/images/panels/CreateImageRegistryPanel.tsx
omarelkashef: I will reach out to check on the progress, but we have a ticket to update the creation process anyways https://warthogs.atlassian.net/browse/WD-36203?atlOrigin=eyJpIjoiNjgyZDE1YjIxOWIyNGUzZjgyNTZiNjgwYTNmZGYxNjgiLCJwIjoiaiJ9

tests/helpers/projects.ts
edlerd: No network is only the default in some environments. I fear this might break other environments where the tests are run.

tests/helpers/image-registries.ts
edlerd: The scoped `sidePanel` together with the label `name` should be unique, no?

src/pages/images/panels/CreateImageRegistryPanel.tsx
edlerd: Please stay closely updated with Nikita on this. Also to ensure it gets fixed.

tests/helpers/image-registries.ts
omarelkashef: Oh I thought you meant to ditch the side panel and use getByLabel alone. You mean to use getByLabel for the name, right?

tests/image-registries.spec.ts
```diff
@@ -2,95 +2,145 @@ import { expect, test } from "./fixtures/lxd-test";
...
+  await openProjectConfiguration(page);
+
+  await page.getByText("Allow custom restrictions on a project level").click();
+
+  await page
+    .getByRole("navigation", { name: "Project form navigation" })
+    .getByText("Images")
+    .click();
+
+  // No image registry is allowed for use in the project by default.
```
edlerd, line 141: This is not setting the image registry restriction on the project anymore. Is that missing?

omarelkashef: It is allowing custom restrictions on the project level, but it is missing the check that no image registry is selected by default. We can not test a specific image registry because the image selector does show which registry is the source.

edlerd: This should fail the test suite in its current state. It creates a project, then enables custom restrictions, but never saves that change.

tests/image-registries.spec.ts
```diff
@@ -2,95 +2,148 @@ import { expect, test } from "./fixtures/lxd-test";
...
+  await assertReadMode(page, "Available image registries", "");
+
+  await visitCreateInstancePage(page, project);
+  await page.getByRole("button", { name: "Browse images" }).click();
+  await expect(page.getByText("No matching images found")).toBeVisible();
```
edlerd, line 146: Do we need to save the "Allow custom restrictions on a project level" change after clicking it in line 133?

edlerd (approved):
LGTM, thanks!

PR #1987, opened 2026-04-30
feat(navigation): stay on the same global page when switching project

kimanhou:
## Done

- Switching project when on a global page doesn’t trigger navigation but newly selected project is correctly set.
Global pages: Server, Cluster, Operations, Warnings, Permissions, Settings, Image registries (and all sub-pages)

## QA

1. Run the LXD-UI:
    - Go to a global page. Select another project: you should stay on the same page.
    - Go to Instances list: you should see the instances of the newly selected project (you can check the URL to be sure)

## Screenshots

https://github.com/user-attachments/assets/94148b25-1e28-41c1-8f4e-98669b7bade9

### DIFF
--- a/src/pages/projects/NavigationProjectSelector.tsx
+++ b/src/pages/projects/NavigationProjectSelector.tsx
@@ -6,12 +6,14 @@ import {
   Icon,
   SearchBox,
 } from "@canonical/react-components";
-import { useNavigate } from "react-router-dom";
-import NavigationProjectSelectorList from "pages/projects/NavigationProjectSelectorList";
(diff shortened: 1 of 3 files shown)


edlerd (commented):
There is some bug with this, when first landing on [operations](https://lxd-ui-1987.demos.haus/ui/operations) and switching to project `test` it stays on the default project.

I also have second thoughts about whether we should even do the change proposed here in the first place. Because the interaction will result in the user changing the project, but the UI will not reflect any change. This might be perceived as broken behavior.

PR #1989, opened 2026-04-30
fix: Image tab state in All Projects view

### LINKED ISSUE #1988: Images tab not disabled in All Projects View.
The Images tab remains highlighted in the All projects view, where it should be disabled like the other tabs even when "Images" is extended. The Local images subtab is also disabled so there is no value in having the Images tab enabled.

Image

Image

Kxiru:
## Done

- Adds disabled flag to Image Navigation component.

Fixes https://github.com/canonical/lxd-ui/issues/1988

## QA

1. Run the LXD-UI:
    - Open LXDUI (Or demo server)
    - In the project dropdown, select 'All projects'
    - Ensure that the Images tab is disabled in the navigation.
    - Go back to the project dropdown and navigate back to the default project (or any project)
    - Ensure the Images tab is enabled and accessible.

## Screenshots

image

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -462,6 +462,7 @@ const Navigation: FC = () => {
                           open={
                             openNavMenus.includes("images") && !menuCollapsed
                           }
+                          disabled={isAllProjects}
                         >
                           <NavLink
                             to={`${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/local-images`}
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -462,6 +462,7 @@ const Navigation: FC = () => {
                           open={
                             openNavMenus.includes("images") && !menuCollapsed
                           }
+                          disabled={isAllProjects && !hasImageRegistries}
                         >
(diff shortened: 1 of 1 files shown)


src/components/Navigation.tsx
```diff
@@ -462,6 +462,7 @@ const Navigation: FC = () => {
                           open={
                             openNavMenus.includes("images") && !menuCollapsed
                           }
+                          disabled={isAllProjects}
```
edlerd: This should only disable when is all projects *and* the image registries api extension is not present.

kimanhou: But what about local images ? I think, in addition to this, we should update `baseUrls`

src/components/Navigation.tsx
```diff
@@ -462,6 +462,7 @@ const Navigation: FC = () => {
                           open={
                             openNavMenus.includes("images") && !menuCollapsed
                           }
+                          disabled={isAllProjects || !hasImageRegistries}
```
edlerd: This is or, but should be and logic.

edlerd (approved):
LGTM, thanks

PR #2000, opened 2026-05-14
chore(test) fix coverage run, use enable clustering test

edlerd:
## Done

[Coverage tests](https://github.com/canonical/lxd-ui/actions/workflows/coverage.yaml) are currently failing because the clustering tests rely on a two node setup as a prerequisite.

This PR changes the clustered coverage test to only run the single node tests, which work on a single node, previously unclustered backend. 

Ideally we also run other clustered tests with coverage as a follow up.


omarelkashef (commented):
Does this not significantly lower the coverage?

Kxiru:
> Does this not significantly lower the coverage?

Yes, this now does not cover cluster links, network-peering, projects, docs screenshots, networks for the sake of the coverage tests, but what about in general?

edlerd:
> Does this not significantly lower the coverage?

As discussed, this reduces the coverage to only consider enable-clustering and unclustered e2e suites. We should also include the clustered e2e suites. But they require a two-node setup. So this PR needs a bit more work to set up two VMs as a backend for a "real" e2e flow with the clustered suite.

edlerd:
Update to setup two node cluster for backend tests, see [coverage workflow run on edlerd:main](https://github.com/edlerd/lxd-ui/actions/runs/26169970327/job/76984593971)

.github/workflows/coverage.yaml
```diff
@@ -84,7 +85,113 @@ jobs:
         run: |
           set -x
           sudo chmod -R 0777 ../lxd-ui
-          yarn test-coverage
+          yarn test-js-coverage
+          yarn test-e2e-coverage
+          # stop dotrun
```
omarelkashef, line 92: Can we convert the following steps into small single-purpose functions that can be used in other places later, especially in `workflow/pr.yaml`?

package.json
```diff
@@ -27,9 +27,9 @@
     "test-e2e-edge": "tests/scripts/setup_test && npx playwright test --project chromium:lxd-latest-edge firefox:lxd-latest-edge && tests/scripts/teardown_test",
     "test-e2e-5.21-edge": "tests/scripts/setup_test && npx playwright test --project chromium:lxd-5.21-edge firefox:lxd-5.21-edge && tests/scripts/teardown_test",
     "test-e2e-5.0-edge": "npx playwright test --project chromium:lxd-5.0-edge firefox:lxd-5.0-edge",
-    "test-coverage": "yarn test-js-coverage && yarn test-e2e-coverage && yarn test-report-coverage",
     "test-js-coverage": "vitest --run --coverage",
-    "test-e2e-coverage": "rm -rf coverage/playwright* ;  PW_TEST_HTML_REPORT_OPEN='never' npx playwright test --project coverage && npx playwright test --project coverage:clustered",
+    "test-e2e-coverage": "rm -rf coverage/playwright* ;  PW_TEST_HTML_REPORT_OPEN='never' npx playwright test --project coverage && npx playwright test --project coverage:cluster-enable",
```
omarelkashef: There is an inconsistency and coupling because  "test-e2e-coverage" removes the coverage folder but "test-e2e-cluster-coverage" does not which means they have to be called in that order. Maybe we can do the cleanup in the workflow where they are called.

omarelkashef (commented):
Code looks good, only two coments,but the coverage run fails.

package.json
edlerd: yes, good idea.

edlerd: Initially this was added to make the test-e2e-coverage re-runnable on local tests. But we can also compose a command and re-run the composition easily as well.

.github/workflows/coverage.yaml
edlerd: Yes, good idea. I'll do this next.

First, I am making sure the suite passes before making further adjustments to refactor it. See [this run](https://github.com/edlerd/lxd-ui/actions/runs/26275555468) -- hope it succeeds.

edlerd:
All comments resolved, previous run was successful, lets see [this one](https://github.com/edlerd/lxd-ui/actions/runs/26284535580) after refactoring.

edlerd:
First [step of coverage](https://github.com/edlerd/lxd-ui/actions/runs/26286982296) on the branch succeeded, @omarelkashef can you give this another read?

omarelkashef (approved):
Looks good now. Thank you for fixing coverage and refactoring the cluster setup script!

Issue #1827, opened 2026-02-24
`lxc file push/pull` support in ui


vosdev:
I would like to request the ability to download or upload a file to or from an instance in the web ui

cli equivalent of `lxc file push` and `lxc file pull`

edlerd:
Initial explorer landed, file upload is coming in https://github.com/canonical/lxd-ui/pull/2059

PR #2004, opened 2026-05-19
feat: Instance File Explorer Initial Implementation [WD-36294]

Kxiru:
## Done

- [x] Exploration of SFTP endpoint
- [x] Added new File Explorer Tab to Instance detail page
- [x] Early design concept
- [x] List files and directories with icons
- [x] Allow file explorer automatically on click tab (Not on "connect" button)
- [x] Display name, type and modification date.
- [x] Backwards navigation (URL)
- [x] Breadcrumb navigation
- [x] File Download - ~~**DO NOT TEST, CURRENTLY DOWNLOADS INFINITELY**~~

Errors to fix:
- ~~Clicking on a **file** navigates back to root as it tries to "load" the filepath. Ideally this functionality would attempt to download the file instead.~~

Relevant issue: https://github.com/canonical/lxd-ui/issues/1827

Fixes Lack of File Explorer Feature

## QA

1. Run the LXD-UI:
    - Navigate to an Instance detail page. Preferably an instance that is running.
    - Select the file explorer tab
    - Explore the additional features as defined in the "Done" list.

## Screenshots
image


src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+              className="u-truncate"
+              style={{ display: "flex", gap: "0.5rem", alignItems: "center" }}
+            >
+              <Icon name={getIcon()} />
+              <span>{fileName}</span>
+            </div>
+          ),
+          role: "rowheader",
+          "aria-label": "Name",
+          className: isDirectory ? "u-text--muted" : "",
```
edlerd: Let's use a blue link style for directories. Wdyt?

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,152 @@
...
+interface Props {
+  instance: LxdInstance;
+}
+
+const InstanceFileExplorer: FC<Props> = ({ instance }) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const [isConnected, setIsConnected] = useState(false);
+  const [tree, setTree] = useState(new Map<string, string[]>());
+  const [currentPath, setCurrentPath] = useState("/");
```
edlerd: Let's store the path in the url as a parameter. Then we should be able to use the back button in the browser to navigate up a directory, which seem better.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,152 @@
...
+    return (
+      <EmptyState
+        className="empty-state"
+        image={<Icon name="folder" className="empty-state-icon" />}
+        title="File explorer"
+      >
+        <p>Browse and manage files in this instance.</p>
+        <Button
+          appearance="positive"
+          onClick={handleConnect}
```
edlerd: Let's connect automatically, without user interaction.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+interface Props {
+  instance: LxdInstance;
+}
+
+const InstanceFileExplorer: FC<Props> = ({ instance }) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const [isConnected, setIsConnected] = useState(false);
+  const [tree, setTree] = useState(new Map<string, string[]>());
```
edlerd, line 24: If we stay with only ever showing a single directory, this can be a simple array. Or we might want to cache other directories then we should keep it as a `Map<string, string[]>`. But currently we don't make use of the cache. On navigation we always load the tree for the directory.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const [currentPath, setCurrentPath] = useState("/");
+
+  const loadInstancePath = async (path: string) => {
+    const fileExplorerItem = await fetchInstanceFile(
```
edlerd:
```suggestion
    const directory = await fetchInstanceFile(
```
I think this is only used to list directories.

src/pages/instances/InstanceDetail.tsx
```diff
@@ -23,6 +24,7 @@ const tabs: string[] = [
   "Terminal",
   "Console",
   "Logs",
+  "File Explorer",
```
edlerd: I think we had another place in mind, didn't we define it in the spec last week?

src/util/instanceFileExplorer.ts
edlerd, line 1: This should be in `api/instances.ts`, not a new file in utils.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  if (isLoading) {
+    return <Spinner className="u-loader" text="Loading file system..." />;
+  }
+
+  return (
+    <div className="p-panel">
+      <div className="p-panel__header">
+        <h2 className="p-panel__title">Path - {currentPath}</h2>
```
edlerd:
```suggestion
        <h2 className="p-panel__title">Path {currentPath}</h2>
```
I would also like a way to navigate up one directory. Maybe we can split the current path by `/` and make the nested directories clickable.

edlerd (commented):
Very nice! Some ideas to drive this further are below.

src/pages/instances/FileExplorerTable.tsx
Kxiru: I thought the new tabular format with the muting was nice, but here is what the blue link version would look like. Wdyt?
image

src/pages/instances/InstanceDetail.tsx
Kxiru: I thought we settled on File Explorer (2 words) or Filesystem (if we used 1 word). Am I missing something?

src/util/instanceFileExplorer.ts
Kxiru: Fair enough. I wanted to separate concerns.

src/pages/instances/InstanceDetail.tsx
edlerd: The caption is correct, I was thinking about the placement in the tab list. Did we say to put it on the end of the list?

src/pages/instances/FileExplorerTable.tsx
edlerd: I think without the blue, it is not immediately clear that those are links. Though this is a UX question, so there is no right or wrong.

src/pages/instances/InstanceDetail.tsx
omarelkashef: @edlerd and @Kxiru said it should be next to snapshots and I said it should be next to terminal and console, so we decided the overlap is between snapshots and terminal.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,152 @@
...
+    return true;
+  };
+
+  const handleConnect = async () => {
+    setIsLoading(true);
+    setError(null);
+    setIsConnected(false);
+
+    try {
+      console.log("Connecting to SFTP for instance:", instance.name);
```
omarelkashef: This is probably left from debugging.

src/pages/instances/InstanceFileExplorer.tsx
omarelkashef: Yes, we should do that!

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+              filePath,
+            );
+
+            metadataMap.set(fileName, {
+              name: fileName,
+              type: response.headers.get("x-lxd-type") ?? "unknown",
+              modified: response.headers.get("x-lxd-modified") ?? "-",
+            });
+          } catch (e) {
+            console.error(`Failed to fetch metadata for ${fileName}:`, e);
```
omarelkashef, line 52: This should also be removed.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+    new Map(),
+  );
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const fetchMetadata = async () => {
+      setIsLoading(true);
+      const metadataMap = new Map<string, FileMetadata>();
+
+      await Promise.all(
```
omarelkashef, line 36: Maybe we should use Promise.allSettled to display the metadata that we were able to load and show error if any failed.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+            });
+          }
+        }),
+      );
+
+      setFileMetadata(metadataMap);
+      setIsLoading(false);
+    };
+
+    void fetchMetadata();
```
omarelkashef, line 65: Is there a reason to define the function and call it directly? Could we define the function outside the useEffect() or just invoke the function body directly?

omarelkashef (commented):
This is really nice and well-thought! A couple of comments below

src/pages/instances/InstanceFileExplorer.tsx
omarelkashef: I think navigating back with a dedicated back button and clickable path is better. If you are deep into the filesystem, and you want to leave the file explorer, you would have to click your way back until the root and then one more time out of the file explorer.

src/pages/instances/InstanceDetail.tsx
Kxiru: Ah yes, the placement, yep, I'll move that :)

src/pages/instances/InstanceFileExplorer.tsx
Kxiru: Yep, I've changed this already locally, ty :)

src/pages/instances/FileExplorerTable.tsx
Kxiru: I think it is okay to leave this in. There are other areas in the codebase that use this type of error display.

src/pages/instances/FileExplorerTable.tsx
omarelkashef: Actually, I think the catch block handles this.

Kxiru: Nice idea, though I already handle errors inside the mapped async function, so every mapped promise already resolves successfully, since the catch prevents rejection. Basically a promise will never reject, because it either succeeds or is handled in the catch.

omarelkashef:
It would be a nice addition if we allow the user to navigate to a directory by typing the path. This might be out of scope, so we can add a ticket and plan it for the future.

src/pages/instances/FileExplorerTable.tsx
omarelkashef: Yes, I realized that after another look.

Kxiru:
> It would be a nice addition if we allow the user to navigate to a directory by typing the path. This might be out of scope, so we can add a ticket and plan it for the future.

:melting_face: 
Haha, perhaps in a future ticket. Like the listing one since I've already  started listing here, but in that task I will add features such as downloading files and backwards navigation.

src/pages/instances/FileExplorerTable.tsx
Kxiru: useEffect itself cannot be async, so when calling async funcs, this is the appropriate method.

Edit: You technically can immediately invoke it inline. But defining it outside the useEffect would basically create redundant code lines as you have to send in all the params anyway. Eg;

```
const fetchMetadata = async (
  files,
  currentPath,
  instanceName,
  projectName,
) => {
  ...
};

useEffect(() => {
  void fetchMetadata(
    files,
    currentPath,
    instanceName,
    projectName,
  );
}, [files, currentPath, instanceName, projectName]);
```

edlerd:
> It would be a nice addition if we allow the user to navigate to a directory by typing the path. This might be out of scope, so we can add a ticket and plan it for the future.

I think an easy way to satisfy this requirement is to move the current directory state to the url.

Kxiru:
> I think an easy way to satisfy this requirement is to move the current directory state to the url.

@edlerd , but won't we end up with really really long URLs? Even when going a few directories deep. Plus long instance names, etc.

edlerd:
> > I think an easy way to satisfy this requirement is to move the current directory state to the url.
> 
> really really long URLs? Even when going a few directories deep. Plus long instance names, etc.

Max url length is ~2000 characters. Instance name is capped at 64 and with the other parameters, the file explorer base url will be <300 characters, leaving >1000 chars for the path. There is an edge case where users could create very long nested folder names that would exceed that. Though I think normal usage should never exceed it. So in my opinion this should be fine.

Kxiru:
@edlerd , fair enough.
I think this PR is ready for review. I will take:
- Back button
- Breadcrumb backwards navigation
- Directory state in URL
Into the next ticket.

src/pages/instances/InstanceFileExplorer.tsx
Kxiru: @omarelkashef  You can leave the file explorer itself at any time. Though I am considering adding an alert when navigating away from the page, the same way you would for the terminal.

edlerd: > considering adding an alert when navigating away from the page

I wouldn't do that. There is no interaction the user does with results of their interaction that is lost when navigating away. With the path in the url, they can simply go back and land exactly where they left.

Kxiru: > There is no interaction the user does with results of their interaction that is lost when navigating away.

Yet!

edlerd: I am confused. Which interaction do you see coming that might have a work state? All I can think of is changing directory.

Or do you think of the text editor? For that, I agree we should block navigating with unsaved changes. But then limit it to that case, and after saving, clear that nav-blocker.

Kxiru: I suppose my suggestion was moreso for if we stayed with the connection via button route, and if we _didn't_ have backwards navigation. It feels like a "session" is being interrupted. But happy to go without.

src/pages/instances/FileExplorerTable.tsx
kimanhou: Does this fail silently in the UI ?

src/pages/instances/FileExplorerTable.tsx
kimanhou: You usually need to use the function in the `useEffect`'s `return` to clear it.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,149 @@
...
+      await loadInstancePath("/");
+
+      setTimeout(() => {
+        setIsConnected(true);
+        setIsLoading(false);
+      }, 1000);
+    } catch (e) {
+      setError(e instanceof Error ? e.message : "Failed to connect");
```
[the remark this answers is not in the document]
kimanhou: Was this addressed ?

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,166 @@
...
+          style: isDirectory ? { cursor: "pointer" } : undefined,
+        },
+        {
+          content: metadata?.type ?? "loading...",
+          role: "cell",
+          "aria-label": "Type",
+        },
+        {
+          content: metadata?.modified
+            ? isoTimeToString(metadata.modified)
```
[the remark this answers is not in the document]
kimanhou: Was this addressed ?

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,166 @@
...
+            console.error(`Failed to fetch metadata for ${fileName}:`, e);
+            metadataMap.set(fileName, {
+              name: fileName,
+              type: "unknown",
+              modified: "",
+            });
+          }
+        }),
+      );
+
```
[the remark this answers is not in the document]
kimanhou: Was this addressed ?

src/pages/instances/InstanceFileExplorer.tsx
Kxiru: Either way we need it though because FileExplorerTable needs the keyValue array (returning all files), no?

src/pages/instances/BreadcrumbPath.tsx
```diff
@@ -0,0 +1,58 @@
...
+                <Link
+                  to="#"
+                  onClick={(e) => {
+                    e.preventDefault();
+                    onNavigate(crumb.path);
+                  }}
+                >
+                  {crumb.label}
+                </Link>
+              )}
```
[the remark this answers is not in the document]
Kxiru: @edlerd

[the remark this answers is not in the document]
edlerd: We should build the full url here to file explorer with path and use that as `to`. Then we can remove the click handler and use the link normally.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,175 @@
...
+        {
+          content: (
+            <Link to="#" className="fe-item">
+              <Icon name={getIcon()} />
+              <span>{fileName}</span>
+            </Link>
+          ),
+          role: "rowheader",
+          "aria-label": "Name",
+          onClick: isDirectory
```
edlerd: We should remove this onclick handler, and use a proper path for the link above instead. The links `to` prop can point to the nested folder.

If this is a file, we need special handling to start the download.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+const InstanceFileExplorer: FC<Props> = ({ instance }) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+  const [tree, setTree] = useState(new Map<string, string[]>());
+  const [searchParams, setSearchParams] = useSearchParams();
+  const currentPath = searchParams.get("path") || "/";
+
+  useEffect(() => {
+    if (isInstanceRunning(instance) && !tree.has(currentPath)) {
```
edlerd: If this is a container, the file explorer is also available for stopped instances.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,141 @@
...
+  };
+
+  const handleConnect = async () => {
+    const hasLoadedPath = await loadInstancePath("/");
+    if (!hasLoadedPath) {
+      setError((prev) => prev ?? "Failed to load initial path");
+    }
+  };
+
+  if (!isInstanceRunning(instance)) {
```
edlerd: We should only do this for non-running VMs.

edlerd (commented):
Good progress, some things are still broken and need adjustments. See below.

src/pages/instances/InstanceFileExplorer.tsx
Kxiru: How can we differentiate between VMs and Containers, here?

src/pages/instances/FileExplorerTable.tsx
Kxiru: So some kind of function to construct the URL, for the "to" component. I'll look into it. As for the special download handling, yes, working on it :melting_face:

src/pages/instances/InstanceFileExplorer.tsx
edlerd: The instance object should have a field, I think it is called `type`.

src/api/instances.tsx
```diff
@@ -552,3 +555,72 @@ export const createInstanceBackup = async (
...
+    if (!response.ok) {
+      throw new Error(
+        `Failed to fetch instance file header: ${response.status} ${response.statusText}`,
+      );
+    }
+    return response;
+  });
+};
+
+export const fetchInstanceFile = async (
```
edlerd: We shouldn't use a fetch for files. Files can be a normal `Link` with the target path we build down here.
 We probably don't need the `fetchInstanceFile` file in the api folder at all.

src/util/instances.tsx
```diff
@@ -85,3 +85,34 @@ export const getInstanceClusterMember = (
...
+  projectName: string,
+) => {
+  const baseURL = `${ROOT_PATH}/ui/project/${projectName}/instance/${instanceName}/file-explorer?path=`;
+  if (path === "ROOT") {
+    return `${baseURL}/`;
+  }
+
+  return path === "/"
+    ? `${baseURL}/${fileName}`
+    : `${baseURL}${path}/${fileName}`;
```
edlerd: Let's use a url builder with params here instead of string concatenation. Similar to how we build the params in the api folder.

Also we have to URL encode the query string path value, we shouldn't have `/` in it without encoding the character. So the output should be `https://lxd-ui-2004.demos.haus/ui/project/default/instance/MB-test/file-explorer?path=%2Fetc%2Flbu` and not `https://lxd-ui-2004.demos.haus/ui/project/default/instance/MB-test/file-explorer?path=/etc/ibu`. The `URLSearchParams` builder might do the encoding already, but this needs to be tested. If not we need to manually encode it.

```
  params.set("path", path);

return ${baseURL}${params.toString}`;
```

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,134 @@
...
+  instance: LxdInstance;
+}
+
+const InstanceFileExplorer: FC<Props> = ({ instance }) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+  const [tree, setTree] = useState(new Map<string, string[]>());
+  const [searchParams, setSearchParams] = useSearchParams();
+  const currentPath = searchParams.get("path") || "/";
+  const isInstanceAvailable =
```
edlerd: Naming is a bit vague. Maybe `canServeFiles`?

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  const handleConnect = async () => {
+    const hasLoadedPath = await loadInstancePath("/");
+    if (!hasLoadedPath) {
+      setError((prev) => prev ?? "Failed to load initial path");
+    }
+  };
+
+  if (!isInstanceAvailable) {
```
edlerd: The loading state is broken. This scenario should work:
1) Open detail page for a stopped VM
2) Go to file explorer tabe
3) Start instance

Currently this shows a loader forever. It should automatically load the files as soon as the agent inside the VM is started.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,175 @@
...
+      key: fileName,
+      columns: [
+        {
+          content: isDirectory ? (
+            <FileExplorerDirectory
+              fileName={fileName}
+              currentPath={currentPath}
+              instanceName={instanceName}
+              projectName={projectName}
+              icon={getIcon()}
```
edlerd:
```suggestion
```
We can define the folder icon statically now.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const rows: MainTableRow[] = files.map((fileName) => {
+    const metadata = fileMetadata.get(fileName);
+    const isDirectory = metadata?.type === "directory";
+    const isFile = metadata?.type === "file";
+
+    const getIcon = () => {
+      if (isDirectory) return "folder";
+      if (isFile) return "file-blank";
+      return "pods";
+    };
```
edlerd: We probably don't need this anymore.

src/pages/instances/FileExplorerFile.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const link = generateLink(linkId, href, fileName);
+    buttonRef.current?.appendChild(link);
+    link.click();
+    link.remove();
+    window.URL.revokeObjectURL(href);
+  };
+  ``;
+
+  return (
+    <div className="fe-item" onClick={getFileContents} ref={buttonRef}>
```
edlerd: This should render a `Link` with `to` pointing to the `1.0/...` url directly.

edlerd: Or maybe better not link at all, but show a download icon?

src/pages/instances/FileExplorerFile.tsx
```diff
@@ -0,0 +1,40 @@
...
+    );
+    const href = URL.createObjectURL(fileContentsBlob);
+
+    const link = generateLink(linkId, href, fileName);
+    buttonRef.current?.appendChild(link);
+    link.click();
+    link.remove();
+    window.URL.revokeObjectURL(href);
+  };
+  ``;
```
edlerd: I think we don't need this, make it a normal link to the file.

src/pages/instances/FileExplorerDirectory.tsx
```diff
@@ -0,0 +1,24 @@
+import { Icon } from "@canonical/react-components";
+import type { FC } from "react";
+import { Link } from "react-router-dom";
+import { getFileDirectoryURL } from "util/instances";
+
+const FileExplorerDirectory: FC<{
+  fileName: string;
```
edlerd:
```suggestion
  dirName: string;
```

src/pages/instances/FileExplorerDirectory.tsx
```diff
@@ -0,0 +1,24 @@
...
+  currentPath: string;
```
edlerd:
```suggestion
  parentPath: string;
```

src/util/instances.tsx
```diff
@@ -85,3 +85,34 @@ export const getInstanceClusterMember = (
   }
   return location;
 };
+
+export const getFileDirectoryURL = (
```
```suggestion
export const getFileExplorerDirectoryURL = (
```

src/util/instances.tsx
```diff
@@ -85,3 +85,34 @@ export const getInstanceClusterMember = (
...
+  fileName: string,
+  path: string,
```
```suggestion
  fullPath: string,
```
I think it gets simpler if we expect the caller to do the concatenation of paths.

edlerd (commented):
Great progress, some observations below.

src/pages/instances/FileExplorerTable.tsx
Kxiru: We still need it to dictate when to use pods (symlink icons).
I considered moving it to FileExplorerFile, but in order to identify the type of the file there (symlink/file), we need to pass through the metadata variable (line 94) which I think it a bit much.

Edit: I'll replace the getIcon() function with an explicit folder icon for FileExplorerDirectory, and a ternary operator for FileExplorerFile.

src/pages/instances/FileExplorerFile.tsx
Kxiru: Did we decide to make it be a link and _not_ have a download icon in the action bar in the end?

src/pages/instances/InstanceFileExplorer.tsx
Kxiru: Will address this today.

Kxiru: I have added "canServeFiles" to the dependency array for the useEffect.

src/api/instances.tsx
```diff
@@ -552,3 +553,48 @@ export const createInstanceBackup = async (
...
+    .then((data: LxdFileExplorerItem) => {
+      return data;
+    });
+};
+
+export const fetchInstanceFileHeader = async (
+  name: string,
+  project: string,
+  path: string,
+): Promise<Response> => {
```
edlerd: Let's introduce a `LXDFileHeader` struct and unpack the headers inside this function to that struct. So we don't expose the raw response to usages of this function but structured information.

src/pages/instances/BreadcrumbPath.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return path || "/";
+  };
+
+  return (
+    <nav
+      className="p-breadcrumbs p-breadcrumbs--large"
+      aria-label="File Explorer Path"
+    >
+      <ol className="p-breadcrumbs__items breadcrumb-wrapper">
+        <li className="p-heading--4 breadcrumb-header">Directory:&nbsp; </li>
```
edlerd:
```suggestion
        <li className="p-heading--4 breadcrumb-header">Directory:&nbsp;</li>
```
We only need one space here.

src/pages/instances/BreadcrumbPath.tsx
```diff
@@ -0,0 +1,65 @@
...
+      path: "/" + segments.slice(0, index + 1).join("/"),
+    })),
+  ];
+
+  const getParentPath = (crumbLabel: string, path: string) => {
+    if (crumbLabel === "root") {
+      return "ROOT";
+    }
+    return path || "/";
+  };
```
edlerd:
```suggestion
```
I think we don't need this indirection with small adjustments to the other function.

src/util/instances.tsx
```diff
@@ -85,3 +85,31 @@ export const getInstanceClusterMember = (
...
+
+export const getFileExplorerDirectoryURL = (
+  fullPath: string,
+  instanceName: string,
+  projectName: string,
+) => {
+  const baseURL = `${ROOT_PATH}/ui/project/${projectName}/instance/${instanceName}/file-explorer?path=`;
+  if (fullPath === "ROOT") {
+    return `${baseURL}/`;
+  }
```
```suggestion
```
No need for a special case here.

src/util/instances.tsx
```diff
@@ -85,3 +85,31 @@ export const getInstanceClusterMember = (
...
+
+  return `${baseURL}${fullPath}`;
```
edlerd: 1) We have to url encode the fullpath.
2) We should use params object here, like in the api.

src/pages/instances/DownloadButton.tsx
```diff
@@ -42,7 +32,7 @@ const DownloadButton: FC<DownloadButtonProps> = ({
     const blob = new Blob([data], { type: "text/plain" });
     const href = URL.createObjectURL(blob);
 
-    const link = generateLink(linkId, href);
+    const link = generateLink(linkId, href, fileName);
```
edlerd: Let's keep the helper local, no need to extract it.

src/pages/instances/FileExplorerFile.tsx
```diff
@@ -0,0 +1,26 @@
...
+  currentPath: string;
+  instanceName: string;
+  projectName: string;
+  icon: string;
+}> = ({ fileName, currentPath, instanceName, projectName, icon }) => {
+  const params = new URLSearchParams();
+  params.set("project", projectName);
+  params.set("path", currentPath);
+
+  const fileURL = `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instanceName)}/files?${params.toString()}/${encodeURIComponent(fileName)}`;
```
edlerd:
```suggestion
  const fileURL = `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instanceName)}/files?${params.toString()}`;
```
We shouldn't string concatenate to the params but do it above when we do `params.set("path"...`.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,169 @@
...
+      key: fileName,
+      columns: [
+        {
+          content: isDirectory ? (
+            <FileExplorerDirectory
+              dirName={fileName}
+              parentPath={currentPath}
+              instanceName={instanceName}
+              projectName={projectName}
+              icon={"folder"}
```
edlerd:
```suggestion
```
We don't need this param, as it is static.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,169 @@
...
+            />
+          ) : (
+            <FileExplorerFile
+              fileName={fileName}
+              currentPath={currentPath}
+              instanceName={instanceName}
+              projectName={projectName}
+              icon={isFile ? "file-blank" : "pods"}
```
edlerd, line 107:
```suggestion
```
We can remove the "pods" and also use the file-blank icon for symlinks.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const InstanceFileExplorer: FC<Props> = ({ instance }) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+  const [tree, setTree] = useState(new Map<string, string[]>());
+  const [searchParams, setSearchParams] = useSearchParams();
+  const currentPath = searchParams.get("path") || "/";
+  const canServeFiles =
+    isInstanceRunning(instance) ||
+    (instance.type === "container" && instance.status === "Stopped");
```
edlerd:
```suggestion
    instance.type === "container";
```
I think containers serve files in any state.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,134 @@
...
+      );
+
+      if (!currentDirectory) return false;
+      setTree((prev) => {
+        const newMap = new Map(prev);
+        newMap.set(path, currentDirectory.metadata);
+        return newMap;
+      });
+
+      setSearchParams({ path });
```
edlerd:
```suggestion
```
I think we don't need this here.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,134 @@
...
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const handleConnect = async () => {
+    const hasLoadedPath = await loadInstancePath("/");
+    if (!hasLoadedPath) {
+      setError((prev) => prev ?? "Failed to load initial path");
+    }
```
edlerd:
```suggestion
    } catch (e) {
      setError(e instanceof Error ? e.message : "Failed to load path");
    } finally {
      setIsLoading(false);
    }
  };

  const handleConnect = async () => {
    await loadInstancePath("/");
```
I think we can simplify here, as we already handle the error inside, we don't have to expose the result state.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  if (!canServeFiles) {
+    return (
+      <EmptyState
+        className="empty-state"
+        image={<Icon name="containers" className="empty-state-icon" />}
+        title="Instance is not running"
+      >
+        <p>The instance must be running to browse files.</p>
```
edlerd:
```suggestion
        <p>Virtual machines must be running to browse files.</p>
```
A bit more specific?

src/sass/_file_explorer.scss
```diff
@@ -0,0 +1,11 @@
+.fe-item {
+  align-items: center;
+  cursor: pointer;
```
edlerd:
```suggestion
```
This is implicit through the link.

src/sass/_file_explorer.scss
```diff
@@ -0,0 +1,11 @@
...
+  display: flex;
+  gap: 0.5rem;
+}
+
+.breadcrumb-header {
+  display: inline-block;
+  margin-bottom: 0 !important;
```
edlerd:
```suggestion
```
I think the extra spacing here helps.

src/types/fileExplorer.ts
```diff
@@ -0,0 +1,10 @@
+export interface LxdFileExplorerItem {
+  operation: string;
+  metadata: string[];
+}
+
+export interface FileMetadata {
```
edlerd:
```suggestion
export interface LxdFileMetadata {
```
As per the other comment, let's convert to this type already in the api layer.

src/util/instances.tsx
```diff
@@ -85,3 +85,31 @@ export const getInstanceClusterMember = (
   }
   return location;
 };
+
+export const getFileExplorerDirectoryURL = (
```
edlerd, line 89: We should use this function also in the `FileExplorerFile`.

edlerd (commented):
The interactions with this are very nice.

Many ideas to refine the layout and code structure, then it should be good from my side.

src/pages/instances/FileExplorerTable.tsx
edlerd: Just read your other comment. We could also keep it as is. More a UX question of if we should distinguish symlinks and normal files.

Kxiru: I think we should distinguish symlinks and normal files. They are differentiated in the backend so they should be differentiated in the front end too.

src/sass/_file_explorer.scss
Kxiru: Hm, I'm not a fan but I'll takeit out

src/util/instances.tsx
Kxiru: That uses a different endpoint/URL.

src/types/fileExplorer.ts
```diff
@@ -0,0 +1,10 @@
+export interface LxdFileExplorerItem {
+  operation: string;
+  metadata: string[];
+}
+
+export interface LxdFileMetadata {
+  name: string;
```
edlerd:
```suggestion
```
We don't write the name here.

src/api/instances.tsx
```diff
@@ -552,3 +553,51 @@ export const createInstanceBackup = async (
...
+  ).then((response: Response) => {
+    if (!response.ok) {
+      throw new Error(
+        `Failed to fetch instance file header: ${response.status} ${response.statusText}`,
+      );
+    }
+    return {
+      type: response.headers.get("x-lxd-type") ?? "unknown",
+      modified: response.headers.get("x-lxd-modified") ?? "-",
+    } as LxdFileMetadata;
```
```suggestion
    };
```
With the correct type, we don't need the cast here.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+              instanceName,
+              projectName,
+              filePath,
+            );
+
+            metadataMap.set(fileName, {
+              name: fileName,
+              type: fileMetadata.type,
+              modified: fileMetadata.modified,
+            });
```
edlerd:
```suggestion
            metadataMap.set(fileName, fileMetadata);
```
This simplifies with the right types.

src/pages/instances/BreadcrumbPath.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Link } from "react-router-dom";
+import type { LxdInstance } from "types/instance";
+import { getFileExplorerDirectoryURL } from "util/instances";
+
+interface Props {
+  currentPath: string;
+  instance: LxdInstance;
+}
+
+const BreadcrumbPath: FC<Props> = ({ currentPath, instance }) => {
```
edlerd:
```suggestion
const FileExplorerPath: FC<Props> = ({ currentPath, instance }) => {
```
Let's tie the naming closer to its use case. This is not generic after all.

src/pages/instances/BreadcrumbPath.tsx
```diff
@@ -0,0 +1,57 @@
...
+  ];
+  return (
+    <nav
+      className="p-breadcrumbs p-breadcrumbs--large"
+      aria-label="File Explorer Path"
+    >
+      <ol className="p-breadcrumbs__items breadcrumb-wrapper">
+        <li className="p-heading--4 breadcrumb-header">Directory:&nbsp;</li>
+        {breadcrumbs.map((crumb, index) => {
+          const currentDirectory = index === breadcrumbs.length - 1;
```
edlerd:
```suggestion
          const isCurrentDirectory = index === breadcrumbs.length - 1;
```
Let's name this as a boolean.

src/pages/instances/BreadcrumbPath.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+          return (
+            <li key={crumb.path} className="p-heading--4 continuous-breadcrumb">
+              {currentDirectory ? (
+                <span>{crumb.label}</span>
+              ) : (
+                <Link
+                  to={getFileExplorerDirectoryURL(
+                    crumb.path || "/",
```
edlerd:
```suggestion
                    crumb.path,
```
I don't see why we need the fallback here.

src/pages/instances/BreadcrumbPath.tsx
```diff
@@ -0,0 +1,57 @@
...
+                    instance.name,
+                    instance.project,
```
edlerd:
```suggestion
                    instance,
```
This applies to various places; we should pass the instance around, not unpack the instance to individual name and project fields. That way the interface stays smaller and type safe.

src/pages/instances/FileExplorerDirectory.tsx
```diff
@@ -0,0 +1,26 @@
+import { Icon } from "@canonical/react-components";
+import type { FC } from "react";
+import { Link } from "react-router-dom";
+import { getFileExplorerDirectoryURL } from "util/instances";
+
+const FileExplorerDirectory: FC<{
+  dirName: string;
+  parentPath: string;
+  instanceName: string;
+  projectName: string;
```
edlerd:
```suggestion
  instance: LxdInstance;
```

src/pages/instances/FileExplorerDirectory.tsx
```diff
@@ -0,0 +1,26 @@
...
+}> = ({ dirName, parentPath, instanceName, projectName }) => {
+  const fullPath =
+    parentPath === "/" ? `/${dirName}` : `${parentPath}/${dirName}`;
+
+  return (
+    <Link
+      to={getFileExplorerDirectoryURL(fullPath, instanceName, projectName)}
```
edlerd:
```suggestion
      to={getFileExplorerDirectoryURL(fullPath, instance)}
```

src/pages/instances/FileExplorerFile.tsx
```diff
@@ -0,0 +1,29 @@
+import type { FC } from "react";
+import { Icon } from "@canonical/react-components";
+import { getFileExplorerFileURL } from "util/instances";
+
+const FileExplorerFile: FC<{
+  fileName: string;
+  currentPath: string;
+  instanceName: string;
+  projectName: string;
```
edlerd:
```suggestion
  instance: LXDInstance;
```

src/pages/instances/FileExplorerFile.tsx
```diff
@@ -0,0 +1,29 @@
...
+  icon: string;
+}> = ({ fileName, currentPath, instanceName, projectName, icon }) => {
+  return (
+    <a
+      href={getFileExplorerFileURL(
+        currentPath,
+        fileName,
+        instanceName,
+        projectName,
```
edlerd:
```suggestion
        instance,
```

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { isoTimeToString } from "util/helpers";
+import { fetchInstanceFileHeader } from "api/instances";
+import FileExplorerDirectory from "./FileExplorerDirectory";
+import FileExplorerFile from "./FileExplorerFile";
+
+interface Props {
+  files: string[];
+  currentPath: string;
+  instanceName: string;
+  projectName: string;
```
edlerd:
```suggestion
  instance: LXDInstance;
```

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+            metadataMap.set(fileName, {
+              name: fileName,
+              type: fileMetadata.type,
+              modified: fileMetadata.modified,
+            });
+          } catch (e) {
+            console.error(`Failed to fetch metadata for ${fileName}:`, e);
+            metadataMap.set(fileName, {
+              name: fileName,
```
edlerd:
```suggestion
```
Adjusting to the new type.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+    return {
+      key: fileName,
+      columns: [
+        {
+          content: isDirectory ? (
+            <FileExplorerDirectory
+              dirName={fileName}
+              parentPath={currentPath}
+              instanceName={instanceName}
+              projectName={projectName}
```
edlerd:
```suggestion
              instance={instance}
```

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+            />
+          ) : (
+            <FileExplorerFile
+              fileName={fileName}
+              currentPath={currentPath}
+              instanceName={instanceName}
+              projectName={projectName}
```
edlerd:
```suggestion
              instance={instance}
```

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -0,0 +1,126 @@
...
+    } catch (e) {
+      setError(e instanceof Error ? e.message : "Failed to load path");
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const handleConnect = async () => {
+    await loadInstancePath("/");
+  };
```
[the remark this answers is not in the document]
edlerd: This should be addressed.

src/sass/_file_explorer.scss
```diff
@@ -0,0 +1,9 @@
+.fe-item {
```
edlerd:
```suggestion
.file-explorer-item {
```
Let's use a clearer naming here.

edlerd (commented):
QA, interaction, and code structure LGTM; many small ideas to simplify the code.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+            ? isoTimeToString(metadata.modified)
+            : "-",
+          role: "cell",
+          "aria-label": "Modified",
+        },
+      ],
+      sortData: {
+        name: fileName.toLowerCase(),
+        type: metadata?.type ?? "unknown",
+        modified: metadata?.modified ?? "-",
```
edlerd: We are repeating the logic for `metadata?.type ?? "unknown"` here, and it is already out of sync with the actual display. Better assign both the type and modified values to a const each. Then we can use them for display and sorting.

src/util/instances.tsx
```diff
@@ -85,3 +85,29 @@ export const getInstanceClusterMember = (
...
+) => {
+  const params = new URLSearchParams();
+  params.set("project", instance.project);
+  params.set("path", fullPath);
+
+  return `${ROOT_PATH}/ui/project/${encodeURIComponent(instance.project)}/instance/${encodeURIComponent(instance.name)}/file-explorer?${params.toString()}`;
+};
+
+export const getFileExplorerFileURL = (
+  currentPath: string,
```
```suggestion
  parentPath: string,
```
Does this make more sense? Also in `FileExplorerFile` component, I think the prop should be called `parentPath` instead. In `FileExplorerTable` the name `currentPath` makes sense though.

edlerd (commented):
Great stuff, thanks for the iterations. Two likely last comments, then we can merge it :)

edlerd (approved):
LGTM, great work.

Kxiru:
> LGTM, great work.

Woohoo! Thanks a bunch :)

PR #2006, opened 2026-05-20
chore: catch duplicate imports in linter WD-36386

omarelkashef:
## Done

- Update linter to catch duplicate imports.
- Fix surfaced violations.

## QA

1. Run the LXD-UI:
    - Run `npx eslint` and make sure there are no more duplicate import violation.


eslint.config.mjs
```diff
@@ -10,65 +11,81 @@ import { FlatCompat } from "@eslint/eslintrc";
...
+export default defineConfig([
+  {
+    ignores: [
+      "src/lib/*",
+      "node_modules/*",
+      "eslint.config.mjs",
+      ".dependency-cruiser.cjs",
+      "fix-asset-paths.js",
+      "public/assets/js/manifest.js",
+      "babel.config.js",
```
edlerd: Can we keep those in and not ignore them?

edlerd (commented):
Good stuff, thanks for aligning! One question below.

eslint.config.mjs
omarelkashef: They cause an error because they are not ignored here and not included in tsconfig. If we want to lint these files, we need to add them tsconfig, otherwise we need to explicitly ignore them here.

Kxiru:
Great addition! We really needed this :melting_face: .

eslint.config.mjs
edlerd: I think we should lint them as well. Why didn't they fail/surface earlier though? I am a bit confused why with the changes of this PR we need to add them here.

omarelkashef: Yeah, we can lint them. The errors were not surfaced before because `lint-js-eslint` in `package.json` command only targeted `src` and `tests` directories. The errors that I mentioned in my previous comment are inline errors in the IDE

image

edlerd: I think that is a faulty setup of the IDE then. At least in pycharm there is no issue with the current config. Can we keep those included please?

.dependency-cruiser.cjs
```diff
@@ -1,6 +1,5 @@
 /** @type {import('dependency-cruiser').IConfiguration} */
 module.exports = {
-  options: { doNotFollow: { path: ["node_modules"] } },
```
edlerd, line 3: Why was this removed?

tsconfig.eslint.json
```diff
@@ -0,0 +1,19 @@
...
+    "tests/**/*.ts",
+    "babel.config.js",
+    "eslint.config.mjs",
+    "fix-asset-paths.js",
+    "merge.playwright.config.ts",
+    "playwright.config.ts",
+    "vite.config.ts",
+    "vitest.config.ts",
+    ".dependency-cruiser.cjs",
+    "public/assets/js/manifest.js"
```
edlerd: Is this only extending these two files? Maybe it's better not to lint those two files, and then we don't need this extra tsconfig file that duplicates the inclusion list from `tsconfig.json`.

.dependency-cruiser.cjs
omarelkashef: It is duplicated at line 170.

edlerd: ohh, good catch!

tsconfig.eslint.json
omarelkashef: It is adding  eslint.config.mjs and fix-asset-paths.js also

omarelkashef: It would be nice to lint those files too. The files in tsconfig do not change frequently so it is fairly safe to duplicate.

edlerd: Can we add those 4 files to the original `tsconfig.json`? Tested it locally and it seems to pass `dotrun lint-js`.

omarelkashef: It will pass yes, but adding configuration scripts to the production build in tsconfig is not best practice.

edlerd: I think as long as they are compatible, we should keep it all in one file.

src/pages/networks/NetworkTopology.tsx
```diff
@@ -133,7 +131,7 @@ const NetworkTopology: FC<Props> = ({ formik, project, isServerClustered }) => {
               return (
                 <div
                   key={networkUrl}
-                  className={classnames("downstream-item", {
+                  className={classNames("downstream-item", {
```
edlerd, line 134: This should always be `classNames` and there are several usages with `classnames` left. This can be a follow-up, though.

edlerd (approved):
LGTM, thanks :)

PR #2007, opened 2026-05-20
feat(load-balancers) add load balancer management WD-36668

edlerd:
## Done

- Add load balancer create / edit / delete
- Add load balancer pool create / edit / delete

Fixes WD-36668

## QA

1. Run the LXD-UI:
    - create a LXD container with lxd and microovn as backend
      - Create a LXD container with ubuntu 26.04, in it run:
      - sudo snap install lxd --channel=latest/edge
      - replace lxd with debug files build from this upstream branch https://github.com/canonical/lxd/pull/18179
        - lxc file push /home/david.edler@canonical.com/go/bin/lxd load-balancers/var/snap/lxd/common/lxd.debug --project=latest-lxd
        - lxc file push /home/david.edler@canonical.com/go/bin/lxc load-balancers/var/snap/lxd/common/lxc.debug --project=latest-lxd
        - reboot the container
      - sudo snap install microovn --channel=latest/edge
      - microovn init (create a new cluster)
      - lxd init (let it listen on the network)
      - set lxd config to `network.ovn.northbound_connection: ssl:127.0.0.1:6641`
    - connect the ui from this branch to the container as a backend
    - In the ui configure the lxdbr0 created by lxd init above to have:
      - IPv4 DHCP ranges: With about 20 ips, such as 10.115.100.170-10.115.100.190, needs to match the network mask
      - IPv4 OVN ranges: With about 50 ips, such as 10.115.100.100-10.115.100.150, needs to match the network mask
      - IPV4 routes: Set to a valid CIDR, such as 10.115.100.208/28
    - in the ui create an ovn network
      - use lxdbr0 as an uplink, other settings default
    - in the ui visit ovn network detail page
    - in the ui go to load balancers
    - in the ui create a load balancer and a load balancer pool. Use an ip from the OVN network pool. So if the network is 10.115.100.1/64, you could use the manual ip 10.115.100.222 for a load balancer. Automatic allocation will likely fail without further setup.

## Screenshots

image

image

image

image

image

image

image

### DIFF
--- a/src/api/load-balancers.tsx
+++ b/src/api/load-balancers.tsx
@@ -1,0 +1,110 @@
+import { handleResponse } from "util/helpers";
+import type { LxdApiResponse } from "types/apiResponse";
+import type { LxdOperationResponse } from "types/operation";
+import { ROOT_PATH } from "util/rootPath";
+import type { LxdLoadBalancer } from "types/loadBalancers";
+
+export const fetchLoadBalancers = async (
+  network: string,
+  project: string,
+): Promise<LxdLoadBalancer[]> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+  params.set("recursion", "1");
+
+  return fetch(
+    `${ROOT_PATH}/1.0/networks/${encodeURIComponent(network)}/load-balancers?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdLoadBalancer[]>) => {
+      return data.metadata.sort((a, b) =>
+        a.listen_address.localeCompare(b.listen_address),
+      );
+    });
+};
+
+export const fetchLoadBalancer = async (
+  network: string,
+  project: string,
+  listenAddress: string,
+): Promise<LxdLoadBalancer> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+  params.set("recursion", "1");
+
+  return fetch(
+    `${ROOT_PATH}/1.0/networks/${encodeURIComponent(network)}/load-balancers/${encodeURIComponent(listenAddress)}?${params.toString()}`,
+  )
+    .then(handleResponse)
+    .then((data: LxdApiResponse<LxdLoadBalancer>) => {
+      return data.metadata;
+    });
+};
+
+export const updateLoadBalancer = async (
+  network: string,
+  project: string,
+  loadBalancer: LxdLoadBalancer,
+): Promise<LxdOperationResponse> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+
+  return fetch(
+    `${ROOT_PATH}/1.0/networks/${encodeURIComponent(network)}/load-balancers/${encodeURIComponent(loadBalancer.listen_address)}?${params.toString()}`,
+    {
+      method: "PUT",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(loadBalancer),
+    },
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
+};
+
+export const createLoadBalancer = async (
+  network: string,
+  project: string,
+  loadBalancer: Partial<LxdLoadBalancer>,
+): Promise<LxdOperationResponse> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+
+  return fetch(
+    `${ROOT_PATH}/1.0/networks/${encodeURIComponent(network)}/load-balancers?${params.toString()}`,
+    {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(loadBalancer),
+    },
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
+};
+
+export const deleteLoadBalancer = async (
+  network: string,
+  project: string,
+  loadBalancer: LxdLoadBalancer,
+): Promise<LxdOperationResponse> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+
+  return fetch(
+    `${ROOT_PATH}/1.0/networks/${encodeURIComponent(network)}/load-balancers/${encodeURIComponent(loadBalancer.listen_address)}?${params.toString()}`,
+    {
+      method: "DELETE",
+    },
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
+};
--- a/src/context/useLoadBalancerPools.tsx
+++ b/src/context/useLoadBalancerPools.tsx
@@ -1,0 +1,64 @@
+import { useQuery, type UseQueryResult } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import {
+  fetchLoadBalancerPool,
+  fetchLoadBalancerPools,
+  fetchLoadBalancerPoolState,
+} from "api/load-balancer-pools";
+import type {
+  LxdLoadBalancerPool,
+  LxdLoadBalancerPoolStatus,
+} from "types/loadBalancers";
+
+export const useLoadBalancerPools = (
+  network: string,
+  project: string,
+): UseQueryResult<LxdLoadBalancerPool[]> => {
+  return useQuery({
+    queryKey: [
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network,
+      queryKeys.loadBalancerPools,
+    ],
+    queryFn: async () => fetchLoadBalancerPools(network, project),
+  });
+};
+
+export const useLoadBalancerPool = (
+  network: string,
+  project: string,
+  pool: string,
+): UseQueryResult<LxdLoadBalancerPool> => {
+  return useQuery({
+    queryKey: [
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network,
+      queryKeys.loadBalancerPools,
+      pool,
+    ],
+    queryFn: async () => fetchLoadBalancerPool(network, project, pool),
+  });
+};
+
+export const useLoadBalancerPoolState = (
+  network: string,
+  project: string,
+  pool: string,
+): UseQueryResult<LxdLoadBalancerPoolStatus> => {
+  return useQuery({
+    queryKey: [
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network,
+      queryKeys.loadBalancerPools,
+      pool,
+      queryKeys.state,
+    ],
+    queryFn: async () => fetchLoadBalancerPoolState(network, project, pool),
+  });
+};
--- a/src/context/useLoadBalancers.tsx
+++ b/src/context/useLoadBalancers.tsx
@@ -1,0 +1,38 @@
+import { useQuery, type UseQueryResult } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import type { LxdLoadBalancer } from "types/loadBalancers";
+import { fetchLoadBalancer, fetchLoadBalancers } from "api/load-balancers";
+
+export const useLoadBalancers = (
+  network: string,
+  project: string,
+): UseQueryResult<LxdLoadBalancer[]> => {
+  return useQuery({
+    queryKey: [
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network,
+      queryKeys.loadBalancers,
+    ],
+    queryFn: async () => fetchLoadBalancers(network, project),
+  });
+};
+
+export const useLoadBalancer = (
+  network: string,
+  project: string,
+  listenAddress: string,
+): UseQueryResult<LxdLoadBalancer> => {
+  return useQuery({
+    queryKey: [
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network,
+      queryKeys.loadBalancers,
+      listenAddress,
+    ],
+    queryFn: async () => fetchLoadBalancer(network, project, listenAddress),
+  });
+};
--- a/src/pages/networks/CreateLoadBalancer.tsx
+++ b/src/pages/networks/CreateLoadBalancer.tsx
@@ -1,0 +1,190 @@
+import { useEffect, type FC } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import {
+  ActionButton,
+  useNotify,
+  useToastNotification,
+} from "@canonical/react-components";
+import { useFormik } from "formik";
+import { Link, useNavigate, useParams } from "react-router-dom";
+import BaseLayout from "components/BaseLayout";
+import HelpLink from "components/HelpLink";
+import FormFooterLayout from "components/forms/FormFooterLayout";
+import { useEventQueue } from "context/eventQueue";
+import { useNetwork } from "context/useNetworks";
+import { ROOT_PATH } from "util/rootPath";
+import type { LoadBalancerFormValues } from "types/forms/loadBalancers";
+import LoadBalancerForm, {
+  LoadBalancerSchema,
+  toLoadBalancer,
+} from "pages/networks/forms/LoadBalancerForm";
+import { createLoadBalancer } from "api/load-balancers";
+import usePanelParams, { panels } from "util/usePanelParams";
+import CreateLoadBalancerPoolPanel from "pages/networks/panels/CreateLoadBalancerPoolPanel";
+import EditLoadBalancerPoolPanel from "pages/networks/panels/EditLoadBalancerPoolPanel";
+import { setLoadBalancerCreatedPool } from "util/loadBalancers";
+import { useLoadBalancerPools } from "context/useLoadBalancerPools";
+
+const CreateLoadBalancer: FC = () => {
+  const eventQueue = useEventQueue();
+  const navigate = useNavigate();
+  const notify = useNotify();
+  const panelParams = usePanelParams();
+  const toastNotify = useToastNotification();
+  const queryClient = useQueryClient();
+  const { network: networkName, project } = useParams<{
+    network: string;
+    project: string;
+  }>();
+
+  if (!networkName) {
+    return <>Missing Network</>;
+  }
+
+  if (!project) {
+    return <>Missing project</>;
+  }
+
+  const { data: network, error: networkError } = useNetwork(
+    networkName,
+    project,
+  );
+
+  const { data: pools = [] } = useLoadBalancerPools(networkName, project);
+
+  useEffect(() => {
+    if (networkError) {
+      notify.failure("Loading networks failed", networkError);
+    }
+  }, [networkError, notify]);
+
+  const getDefaultListenAddress = () => {
+    if (network?.config["ipv4.address"] !== "none") {
+      return "0.0.0.0";
+    }
+    if (network?.config["ipv6.address"] !== "none") {
+      return "::";
+    }
+    return "";
+  };
+
+  const invalidateCache = () => {
+    queryClient.invalidateQueries({
+      queryKey: [
+        queryKeys.projects,
+        project,
+        queryKeys.networks,
+        network?.name,
+        queryKeys.loadBalancers,
+      ],
+    });
+  };
+
+  const onSuccess = (listenAddress?: string) => {
+    invalidateCache();
+    toastNotify.success(`Load balancer ${listenAddress} created.`);
+  };
+
+  const onFailure = (e: unknown, listenAddress?: string) => {
+    invalidateCache();
+    formik.setSubmitting(false);
+    toastNotify.failure(`Creation of load balancer ${listenAddress} failed`, e);
+  };
+
+  const formik = useFormik<LoadBalancerFormValues>({
+    initialValues: {
+      listenAddress: getDefaultListenAddress(),
+      description: "",
+      ports: [
+        {
+          key: "initial",
+          protocol: "tcp",
+          listenPort: "",
+          targetPool: pools[0]?.name,
+        },
+      ],
+    },
+    enableReinitialize: true,
+    validationSchema: LoadBalancerSchema,
+    onSubmit: (values) => {
+      const loadBalancer = toLoadBalancer(values);
+      createLoadBalancer(networkName, project, loadBalancer)
+        .then((operation) => {
+          navigate(
+            `${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/network/${encodeURIComponent(networkName)}/load-balancers`,
+          );
+
+          if (operation?.metadata.id) {
+            toastNotify.info(
+              `Creation of load balancer ${formik.values.listenAddress} has started.`,
+            );
+            eventQueue.set(
+              operation.metadata.id,
+              () => {
+                onSuccess(formik.values.listenAddress);
+              },
+              (msg) => {
+                onFailure(new Error(msg), formik.values.listenAddress);
+              },
+            );
+          } else {
+            onSuccess(formik.values.listenAddress);
+          }
+        })
+        .catch((e) => {
+          onFailure(e, formik.values.listenAddress);
+        });
+    },
+  });
+
+  return (
+    <BaseLayout
+      title={
+        <HelpLink
+          docPath="/howto/network_load_balancers/"
+          title="Learn more about load balancers"
+        >
+          Create a load balancer
+        </HelpLink>
+      }
+      contentClassName="create-load-balancer"
+    >
+      {network && <LoadBalancerForm formik={formik} network={network} />}
+      <FormFooterLayout>
+        <Link
+          className="p-button--base"
+          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/network/${encodeURIComponent(networkName)}/load-balancers`}
+        >
+          Cancel
+        </Link>
+        <ActionButton
+          appearance="positive"
+          loading={formik.isSubmitting}
+          disabled={formik.isSubmitting}
+          onClick={() => void formik.submitForm()}
+        >
+          Create
+        </ActionButton>
+      </FormFooterLayout>
+
+      {panelParams.panel === panels.createLoadBalancerPool && network && (
+        <CreateLoadBalancerPoolPanel
+          network={network}
+          onCreate={(name) => {
+            setLoadBalancerCreatedPool(name, formik);
+          }}
+          onCancel={() => {
+            setLoadBalancerCreatedPool("", formik);
+          }}
+        />
+      )}
+
+      {panelParams.panel === panels.editLoadBalancerPool && network && (
+        <EditLoadBalancerPoolPanel network={network} />
+      )}
+    </BaseLayout>
+  );
+};
+
+export default CreateLoadBalancer;
--- a/src/pages/networks/EditLoadBalancer.tsx
+++ b/src/pages/networks/EditLoadBalancer.tsx
@@ -1,0 +1,201 @@
+import { useEffect, type FC } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import {
+  ActionButton,
+  Spinner,
+  useNotify,
+  useToastNotification,
+} from "@canonical/react-components";
+import { useFormik } from "formik";
+import { Link, useNavigate, useParams } from "react-router-dom";
+import BaseLayout from "components/BaseLayout";
+import HelpLink from "components/HelpLink";
+import FormFooterLayout from "components/forms/FormFooterLayout";
+import { useEventQueue } from "context/eventQueue";
+import { useNetwork } from "context/useNetworks";
+import { ROOT_PATH } from "util/rootPath";
+import { updateLoadBalancer } from "api/load-balancers";
+import LoadBalancerForm, {
+  LoadBalancerSchema,
+  toLoadBalancer,
+} from "pages/networks/forms/LoadBalancerForm";
+import type { LoadBalancerFormValues } from "types/forms/loadBalancers";
+import usePanelParams, { panels } from "util/usePanelParams";
+import CreateLoadBalancerPoolPanel from "pages/networks/panels/CreateLoadBalancerPoolPanel";
+import EditLoadBalancerPoolPanel from "pages/networks/panels/EditLoadBalancerPoolPanel";
+import { useLoadBalancer } from "context/useLoadBalancers";
+import { useCurrentProject } from "context/useCurrentProject";
+import { setLoadBalancerCreatedPool } from "util/loadBalancers";
+
+const EditLoadBalancer: FC = () => {
+  const navigate = useNavigate();
+  const notify = useNotify();
+  const panelParams = usePanelParams();
+  const toastNotify = useToastNotification();
+  const queryClient = useQueryClient();
+  const { projectName: project } = useCurrentProject();
+  const { network: networkName, listenAddress } = useParams<{
+    network: string;
+    listenAddress: string;
+  }>();
+
+  if (!networkName) {
+    return <>Missing Network</>;
+  }
+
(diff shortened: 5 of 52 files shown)


kimanhou:
I just finished the setup to test this PR.

Do we really need the little arrows on listen port ?
image

Same for target port
image

I am not sure what I did wrong but after creating a pool and a load balancer, I have this:
image

edlerd:
> I just finished the setup to test this PR.
> 
> Do we really need the little arrows on listen port ?

This is the browser rendering an input field for type number. I don't think we can hide those.


> I am not sure what I did wrong but after creating a pool and a load balancer, I have this: 

I forgot to mention in the QA steps: Best use an ip from the OVN network. So if the network has 10.115.100.1/64 you could use 10.115.100.222 as a manual IP. The automatic allocation (default) needs additional setup to work properly.

Edit: Updated also in QA steps above.

src/pages/networks/CreateLoadBalancer.tsx
```diff
@@ -0,0 +1,178 @@
...
+          Cancel
+        </Link>
+        <ActionButton
+          appearance="positive"
+          loading={formik.isSubmitting}
+          disabled={
+            !formik.isValid ||
+            formik.isSubmitting ||
+            !formik.values.listenAddress
+          }
```
kimanhou: When you first land on the page, after clicking on "Create load balancer" button from the empty state, the `Create` button is enabled, even tho nothing was inputted. I think it should be disabled because clicking on it results in an error.

Image

Image

Note: once you clik on "Add port", it's correctly disabled.

kimanhou:
You probably already thought about this but is there a way not to let the user start creating a load balancer if no load balancer pools exist yet ? By the time you realize you need to create a load balancer pool first, you are already 3 clicks down the road and emotionally invested so going on a side quest at that point is frustrating.

One thing that could be interesting to explore is that when you land on the load balancer tab for the first time, the empty state does not mention load balancer pools. Once you create a pool, then you have the two tabs "Load balancers" and "Pools". Would it make sense to show the concept of load balancer pools from the beginning ?

If a load balancer pool exists but no load balancers, the empty table in the load balancers tab looks a bit weird. Would it be better to have the same empty state design as when no load balancer pools exist ?
image

src/pages/networks/CreateLoadBalancer.tsx
edlerd: With a correct setup is is valid to submit the creation of the load balancer. It does not need to have ports and the auto assign should work to give you a load balancer with an ip.

edlerd:
> You probably already thought about this but is there a way not to let the user start creating a load balancer if no load balancer pools exist yet ? By the time you realize you need to create a load balancer pool first, you are already 3 clicks down the road and emotionally invested so going on a side quest at that point is frustrating.

Yes, I still had this on my to-do list but didn't have a good idea how to force the user into this flow of creating a pool first if none exist. Well, we could literally check if pools exist, and when a user starts creating a balancer, we show a "no pools exist, start by creating a pool" message that allows them to open the side panel for pool creation.

> One thing that could be interesting to explore is that when you land on the load balancer tab for the first time, the empty state does not mention load balancer pools. Once you create a pool, then you have the two tabs "Load balancers" and "Pools". Would it make sense to show the concept of load balancer pools from the beginning ?
> If a load balancer pool exists but no load balancers, the empty table in the load balancers tab looks a bit weird. Would it be better to have the same empty state design as when no load balancer pools exist ? 

Yes, both are good ideas. I changed it so we always show the balancer/pool selection and both tabs have independent empty states. That should make the concepts easier to gasp on first landing.

edlerd:
Adjusted balancer creation so it shows a notification to create a pool first, and we pre-select the first pool on balancer creation as a port.

src/pages/networks/actions/DeleteLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,112 @@
...
+            ?<br />
+          </p>
+        ),
+        onConfirm: handleDelete,
+      }}
+      className="u-no-margin--bottom has-icon"
+      loading={isLoading}
+      shiftClickEnabled
+      showShiftClickHint
+      disabled={!canEditNetwork(network) || isLoading || pool === ""}
```
kimanhou: This should also be disabled in the pool is referenced in a load balancer port. Currently you can click on the delete icon, then backend throws an error.

Image

src/pages/networks/CreateLoadBalancer.tsx
kimanhou: Is there a way to know if the creation of the load balancer will succeed or not, depending on the values inputted ?

kimanhou:
I really like your last changes. I think we are almost there. When you arrive on the load balancer tab, I suggest we check if a load balancer pool exists. If no pools exist, then show a warning message "no pools exist, start by creating a pool" with "creating a pool" either being a link to the pool tab (but that's an extra useless click) or opening the side panel directly. Finally we should disable the "Create load balancer button" at that point.

image

kimanhou:
>  it shows a notification to create a pool first

No, that is weird
image

> we pre-select the first pool on balancer creation as a port.

Yes, that is awesome. But does it make sense functionally ? I thought you said you could create a load balancer with no port in certain conditions ?

edlerd:
> > it shows a notification to create a pool first
 
> No, that is weird

Agree, the idea to disable creation with a not to create it on the list page should work smoother.

> > we pre-select the first pool on balancer creation as a port.

> Yes, that is awesome. But does it make sense functionally ? I thought you said you could create a load balancer with no port in certain conditions ?

It is possible, but not what the user wants. In that case you have a listener with no handlers, so any traffic directed to it will be lost.

src/pages/networks/CreateLoadBalancer.tsx
edlerd: I think we can assume the uplink is correctly configured for auto allocation.

What was missing is the IPV4 routes on the uplink, it must be a valid CIDR not overlapping with the other configs. I added an example to the QA steps above.

edlerd:
Updated to make the initial onboarding clearer.

kimanhou:
I really like what we have for the initial onboarding now.

However, when creating a load balancer, should we allow the user to edit the pool ? It took me a second to understand that "Edit pool" was for the pool and the delete icon is to delete the port.

image

edlerd:
> However, when creating a load balancer, should we allow the user to edit the pool ? It took me a second to understand that "Edit pool" was for the pool and the delete icon is to delete the port.

I think it is useful to quickly change the pool configuration when editing the ports.

Changed the layout and caption slightly. Also gave the target pool columns a slightly different background. Maybe it helps to distinguish.

kimanhou:
Huge improvement, thanks !

However, we might be able to improve the empty screen even further. For now, the most important call to action `Create load balancer pool` is a link, at the same level as `Learn more about load balancers`. It would be good to make `Create load balancer pool` pop. Perhaps having the documentation link first, then `Create load balancer pool` as the positive button and `Create load balancer` a secondary disabled button on the side of it ?  Something like this:

```
No load balancers found

There are neither load balancers nor pools in this network. You need to create a load balancer pool as a target for your load balancer first.

Learn more about load balancers

[ Create load balancer pool ] [ Create load balancer ]
```

image


On Create load balancer screen, `Instances` and `Target port` are not vertically aligned with the other headers.
The detach port icon has a border. Is this the only occurrence of an icon button with a border ?
image

edlerd:
Thanks for the QA review. @kimanhou I went for another iteration to resolve issues above.

src/api/load-balancers.tsx
```diff
@@ -0,0 +1,111 @@
...
+    },
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
+};
+
+export const deleteLoadBalancer = async (
+  network: LxdNetwork,
```
kimanhou: We don't need to pass the whole object here, just the `network.name`
Other delete functions only passes a string.

src/context/useLoadBalancerPools.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return useQuery({
+    queryKey: [
+      queryKeys.projects,
+      projectName,
+      queryKeys.networks,
+      networkName,
+      queryKeys.loadBalancerPools,
+    ],
+    queryFn: async () => fetchLoadBalancerPools(networkName, projectName),
+    enabled: networkName !== "" && projectName !== "",
```
kimanhou:
```suggestion
    enabled: Boolean(network && project)
```

Can this be simpler ?

src/context/useLoadBalancers.tsx
```diff
@@ -0,0 +1,46 @@
...
+      network,
+      queryKeys.loadBalancers,
+    ],
+    queryFn: async () => fetchLoadBalancers(network, project),
+    enabled: network !== "" && project !== "",
+  });
+};
+
+export const useLoadBalancer = (
+  network?: string,
```
kimanhou: Should we make the API call if network, project and listenAddress are empty ? Why are they optional here ?

src/pages/networks/actions/CreateLoadBalancerBtn.tsx
```diff
@@ -0,0 +1,64 @@
...
+  className,
+  appearance = "positive",
+  hasIcon = true,
+}) => {
+  const isSmallScreen = useIsScreenBelow();
+  const { canEditNetwork } = useNetworkEntitlements();
+  const panelParams = usePanelParams();
+  const navigate = useNavigate();
+  const { projectName: project } = useCurrentProject();
+  const { data: pools = [] } = useLoadBalancerPools(network.name, project);
```
kimanhou, line 28: Is there a difference between `project` coming from `useCurrentProject` and `panelParams.project` ? Can we use `panelParams.project` here and not called `useCurrentProject` ? Or not call `usePanelParams` and use `project` on line 50 ?

src/pages/networks/actions/CreateLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,39 @@
+import type { FC } from "react";
+import { Button, Icon } from "@canonical/react-components";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import { useNetworkEntitlements } from "util/entitlements/networks";
+import usePanelParams from "util/usePanelParams";
+import type { LxdNetwork } from "types/network";
+
+interface Props {
+  network?: LxdNetwork;
```
kimanhou: Shouldn't `network` be required ?

src/pages/networks/actions/DeleteLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,118 @@
...
+  useToastNotification,
+} from "@canonical/react-components";
+import { useNetworkEntitlements } from "util/entitlements/networks";
+import ResourceLabel from "components/ResourceLabel";
+import { deleteLoadBalancerPool } from "api/load-balancer-pools";
+import { useCurrentProject } from "context/useCurrentProject";
+import type { LxdLoadBalancerPool } from "types/loadBalancers";
+
+interface Props {
+  network?: LxdNetwork;
```
kimanhou: Same as in `CreateLoadBalancerPoolBtn`, shouldn't network be required ?
If not, it's best to put optional props at the end, so after `pool`.

src/pages/networks/actions/DeleteLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,118 @@
...
+    setLoading(false);
+    toastNotify.failure(
+      `Deletion of load balancer pool ${pool.name} failed`,
+      e,
+    );
+  };
+
+  const handleDelete = () => {
+    setLoading(true);
+    deleteLoadBalancerPool(network?.name ?? "", project?.name ?? "", pool.name)
```
kimanhou: It might not be a good idea to call the API with empty strings. Can we disable the button instead ?
Something like:
```
const isContextReady = Boolean(network?.name && project?.name);

disabled={!canEditNetwork(network) || isLoading || isPoolUsed || !isContextReady}
```

src/pages/networks/actions/EditLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,61 @@
+import type { FC } from "react";
+import { Button, Icon } from "@canonical/react-components";
+import usePanelParams from "util/usePanelParams";
+import type { LxdNetwork } from "types/network";
+import { useNetworkEntitlements } from "util/entitlements/networks";
+import { CREATE_POOL_VALUE } from "util/setLoadBalancerCreatedPool";
+
+interface Props {
+  network?: LxdNetwork;
```
kimanhou: Same as before

src/pages/networks/actions/EditLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <Button
+      onClick={() => {
+        panelParams.openEditLoadBalancerPool(pool ?? "");
+      }}
+      appearance="base"
+      hasIcon
+      className="u-no-margin--bottom"
+      type="button"
+      title={getTitle()}
+      disabled={
```
kimanhou, line 48: `const isPoolInvalid = !pool || pool === CREATE_POOL_VALUE;`

src/context/useLoadBalancers.tsx
edlerd: Yeah, we can be much stricter in catching the null cases early.

src/pages/networks/actions/CreateLoadBalancerPoolBtn.tsx
edlerd: yes, good catch.

src/pages/networks/actions/DeleteLoadBalancerPoolBtn.tsx
edlerd: we can make the requirement to the component more strict, so we don't have to deal with the null cases here.

src/pages/networks/actions/CreateLoadBalancerBtn.tsx
edlerd: Yes, we shouldn't use panel parameters to determine the project.

edlerd:
Slightly changed the empty state again to be more clear.

Resolved the nullable fields, which simplified plenty of things. Ready for another round @kimanhou

src/context/useLoadBalancerPools.tsx
```diff
@@ -0,0 +1,71 @@
...
+  LxdLoadBalancerPool,
+  LxdLoadBalancerPoolStatus,
+} from "types/loadBalancers";
+
+export const useLoadBalancerPools = (
+  network: string,
+  project: string,
+): UseQueryResult<LxdLoadBalancerPool[]> => {
+  const networkName = network ?? "";
+  const projectName = project ?? "";
```
kimanhou: This is now useless, no ?

src/context/useLoadBalancerPools.tsx
```diff
@@ -0,0 +1,71 @@
...
+    queryFn: async () => fetchLoadBalancerPool(network, project, pool),
+  });
+};
+
+export const useLoadBalancerPoolState = (
+  network: string,
+  project: string,
+  pool?: string,
+): UseQueryResult<LxdLoadBalancerPoolStatus> => {
+  const poolName = pool ?? "";
```
kimanhou: This is now useless, no ?

src/context/useLoadBalancerPools.tsx
```diff
@@ -0,0 +1,71 @@
...
+      queryKeys.projects,
+      project,
+      queryKeys.networks,
+      network,
+      queryKeys.loadBalancerPools,
+      poolName,
+      queryKeys.state,
+    ],
+    queryFn: async () => fetchLoadBalancerPoolState(network, project, poolName),
+    enabled: poolName !== "",
```
kimanhou:
```suggestion
    enabled: Boolean(pool),
```
For consistency

src/pages/networks/actions/CreateLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,46 @@
...
+}) => {
+  const isSmallScreen = useIsScreenBelow();
+  const { canEditNetwork } = useNetworkEntitlements();
+  const panelParams = usePanelParams();
+
+  return (
+    <Button
+      appearance={appearance}
+      className={className}
+      hasIcon={!isSmallScreen && hasIcon}
```
kimanhou: const showIcon = !isSmallScreen && hasIcon;
```suggestion
      hasIcon={showIcon}
```
and replace on line 40

src/pages/networks/actions/CreateLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,46 @@
...
+      onClick={() => {
+        panelParams.openCreateLoadBalancerPool();
+      }}
```
kimanhou:
```suggestion
      onClick={panelParams.openCreateLoadBalancerPool}
```

src/pages/networks/actions/DeleteLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,118 @@
...
+      confirmationModalProps={{
+        title: "Confirm delete",
+        confirmButtonAppearance: "negative",
+        confirmButtonLabel: "Delete",
+        children: (
+          <p>
+            Are you sure you want to delete the load balancer pool{" "}
+            <ResourceLabel type="load-balancer-pool" value={pool.name} bold />
+            ?<br />
+          </p>
```
kimanhou, line 101: Do we need the `<br />` ? I think it comes from `DeleteNetworkForwardBtn`.

src/pages/networks/actions/EditLoadBalancerPoolBtn.tsx
```diff
@@ -0,0 +1,58 @@
...
+const EditLoadBalancerPoolBtn: FC<Props> = ({
+  network,
+  pool,
+  hasCaption = true,
+}) => {
+  const panelParams = usePanelParams();
+  const { canEditNetwork } = useNetworkEntitlements();
+
+  const getTitle = () => {
+    if (pool === undefined || pool === "") {
```
kimanhou:
```suggestion
    if (!pool) {
```
Simpler ?

src/pages/networks/forms/LoadBalancerForm.tsx
```diff
@@ -0,0 +1,187 @@
...
+  useEffect(updateFormHeight, [notify.notification?.message]);
+  useListener(window, updateFormHeight, "resize", true);
+
+  const addPort = () => {
+    formik.setFieldValue("ports", [
+      ...formik.values.ports,
+      {
+        protocol: "tcp",
+      },
+    ]);
```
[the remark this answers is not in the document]
kimanhou: I kinda agree with Copilot. Isn't safer to initialize to 
```
{ protocol: "tcp", listenPort: "", targetPool: "" }
```
?

src/pages/networks/forms/LoadBalancerForm.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <Row className="p-form__group p-form-validation">
+          <NotificationRow />
+          <Notification
+            severity="information"
+            title="Network information"
+            titleElement="h2"
+          >
+            Name: {network?.name}
+            <br />
+            {network?.config["ipv4.address"] && (
```
kimanhou: Should we have variables for ipv4 and ipv6 for readability ?
```
const ipv4 = network?.config["ipv4.address"];
const ipv6 = network?.config["ipv6.address"];
```

They can also be used on lines 126 and 135

src/pages/networks/forms/LoadBalancerInstanceSelector.tsx
```diff
@@ -0,0 +1,58 @@
...
+            items.map((item) => {
+              return {
+                name: item.value as string,
+              };
+            }),
+          );
+        }}
+        showDropdownFooter={availableInstances.length > 5}
+        help={`Instances connected to this network that will receive load-balanced traffic`}
+        emptyMessage={`No instances connected to this network`}
```
kimanhou:
```suggestion
        help="Instances connected to this network that will receive load-balanced traffic"
        emptyMessage="No instances connected to this network"
```

Static strings

src/pages/networks/forms/LoadBalancerPoolSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+      id={id}
+      name={name}
+      help={help}
+      required
+      searchable="never"
+      value={value}
+      error={error}
+      onChange={(newValue) => {
+        setValue(newValue);
+      }}
```
kimanhou:
```suggestion
      onChange={setValue}
```
Simpler ?

src/pages/networks/forms/LoadBalancerPoolSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+      defaultToggleLabel={
```
kimanhou, line 35: Should the whole selector be disabled if no pool is available ?

src/pages/networks/forms/LoadBalancerPortsForm.tsx
```diff
@@ -0,0 +1,217 @@
...
+                );
+                panelParams.openCreateLoadBalancerPool();
+              }}
+            >
+              load balancer pool
+            </Button>
+          );
+
+          return (
+            <tr key={index}>
```
kimanhou: I'm surprised eslint doesn't pick this up. I don't have a good idea for a `key` here tho

src/pages/networks/panels/CreateLoadBalancerPoolPanel.tsx
```diff
@@ -0,0 +1,123 @@
...
+  );
+
+  const closePanel = () => {
+    panelParams.clear();
+    notify.clear();
+  };
+
+  const handleSubmit = (values: LoadBalancerPoolFormValues) => {
+    const pool = toLoadBalancerPool(values);
+    createLoadBalancerPool(network?.name ?? "", project, pool)
```
kimanhou:
```suggestion
    createLoadBalancerPool(network.name, project, pool)
```
`network` always exists, no ?

kimanhou (commented):
Mostly nitpicking on syntax

src/context/useLoadBalancerPools.tsx
edlerd: Yes, absolutely.

src/context/useLoadBalancerPools.tsx
edlerd: Actually yes, we can also make the pool non-nullable.

src/context/useLoadBalancerPools.tsx
edlerd: We can even remove it considering the other comment above :)

src/pages/networks/actions/DeleteLoadBalancerPoolBtn.tsx
edlerd: whoopsie, yes we can remove it.

src/pages/networks/panels/CreateLoadBalancerPoolPanel.tsx
edlerd: yes, missed it when removing the nullables previously

src/pages/networks/forms/LoadBalancerPortsForm.tsx
edlerd: We can generate a stable key on create/edit/add to use here.

edlerd:
Resolved all issues mentioned above @kimanhou

src/pages/networks/actions/DeleteLoadBalancerBtn.tsx
```diff
@@ -0,0 +1,135 @@
...
+        children: (
+          <p>
+            Are you sure you want to delete the load balancer with listen
+            address{" "}
+            <ResourceLabel
+              type="load-balancer"
+              value={loadBalancer.listen_address}
+              bold
+            />
+            ?<br />
```
kimanhou: Same as `DeleteLoadBalancerPoolBtn`. I think we don't need the extra `<br />`

src/pages/networks/CreateLoadBalancer.tsx
```diff
@@ -0,0 +1,193 @@
...
+          navigate(
+            `${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/network/${encodeURIComponent(networkName)}/load-balancers`,
+          );
+
+          if (operation?.metadata.id) {
+            toastNotify.info(
+              <>
+                Creation of load balancer {formik.values.listenAddress} has
+                started.
+              </>,
```
kimanhou:
```suggestion
              `Creation of load balancer ${formik.values.listenAddress} has
                started.`
```

Nitpicking: we don't need a `ReactNode` here. Most notifications use a `ReactNode` because they include a chip but I don't think we will ever have a chip for listenAddress.

src/pages/networks/EditLoadBalancer.tsx
```diff
@@ -0,0 +1,203 @@
...
+    validationSchema: LoadBalancerSchema,
+    onSubmit: (values) => {
+      const loadBalancer = toLoadBalancer(values);
+
+      updateLoadBalancer(networkName ?? "", project, loadBalancer)
+        .then((operation) => {
+          toastNotify.info(
+            <>
+              Update of load balancer {loadBalancer.listen_address} has started.
+            </>,
```
kimanhou:
```suggestion
            `Update of load balancer ${loadBalancer.listen_address} has started.`,
```

Same as above

src/pages/networks/LoadBalancerPoolInstances.tsx
```diff
@@ -0,0 +1,26 @@
...
+const LoadBalancerPoolInstances: FC<Props> = ({ pool }) => {
+  const { projectName: project } = useCurrentProject();
+
+  if (pool.instances.length === 0) {
+    return "-";
+  }
+
+  return pool.instances.map((instance) => {
+    return (
+      <div key={instance.name}>
```
kimanhou, line 19: Is this unique ? Or should the key include the project ? 
```
`${project}::${instance.name}`
```

src/pages/networks/LoadBalancerPoolInstanceStatus.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  const icon = getIcon(status);
+  return (
+    <>
+      <Icon name={icon} className="status-icon" />
+      {capitalizeFirstLetter(status)}
+    </>
+  );
+};
```
kimanhou:
```suggestion
 const normalizedStatus = status?.trim() ? status : "unknown";
  const icon = getStatusIcon(normalizedStatus);

  return (
    <>
      <Icon name={icon} className="status-icon" />
      {capitalizeFirstLetter(normalizedStatus)}
    </>
  );
};
};
```

This could be simplified. Also getIcon could be moved outside the component so that it's created only once.

src/pages/networks/LoadBalancerPoolInstanceStatuses.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <>
+        <Icon name="status-queued-small" className="status-icon" /> Unknown
+      </>
+    );
+  }
+
+  return pool.instances.map((instance) => {
+    const status = getInstanceStatus(instance.name);
+    return (
+      <div key={instance.name}>
```
kimanhou, line 41: Same as in `LoadBalancerPoolInstances`

src/pages/networks/LoadBalancerPoolInstanceStatuses.tsx
```diff
@@ -0,0 +1,48 @@
...
+    return (
+      state?.targets?.find((item) => item.name === instance)?.status ??
+      "unknown"
+    );
+  };
+
+  if (pool.instances.length === 0) {
+    return (
+      <>
+        <Icon name="status-queued-small" className="status-icon" /> Unknown
```
kimanhou: Is this intended ? If a pool doesn't have any instances, shouldn't the status be empty ("-"), instead of unknown ?

src/pages/networks/LoadBalancerTableHeading.tsx
```diff
@@ -0,0 +1,19 @@
...
+  children: React.ReactNode;
+}
+
+const LoadBalancerTableHeading: FC<Props> = ({ title, children }) => {
+  return (
+    <div className="load-balancer-table-heading">
+      <div className="left">
+        <h2 className="p-heading--4">{title}</h2>
+      </div>
+      <div className="right">{children}</div>
```
kimanhou: Is the className "right" used ?

src/pages/networks/LoadBalancerPortStatus.tsx
```diff
@@ -0,0 +1,83 @@
...
+      item.listen_port === port.listen_port
+    );
+  });
+
+  const onlineCount =
+    portTargets?.filter((target) => target.status === "online")?.length ?? 0;
+  const offlineCount =
+    portTargets?.filter((target) => target.status === "offline")?.length ?? 0;
+  const unknownCount =
+    portTargets?.filter((target) => target.status === "unknown")?.length ?? 0;
```
kimanhou:
```suggestion
let onlineCount = 0;
let offlineCount = 0;
let unknownCount = 0;

portTargets?.forEach((target) => {
  if (target.status === "online") onlineCount++;
  else if (target.status === "offline") offlineCount++;
  else if (target.status === "unknown") unknownCount++;
});
```

This could be optimized. Nitpikcing because there are not 1000 portTargets.

src/pages/networks/LoadBalancerPoolInstances.tsx
edlerd: The instances must be from the same project that the network is in. We don't allow load balancers in projects that have shared networks with the default project.

edlerd: So this should be ok as is.

src/pages/networks/LoadBalancerTableHeading.tsx
edlerd: good catch, will remove it.

src/pages/networks/LoadBalancerPortStatus.tsx
edlerd: This is easier to read :+1:

src/pages/networks/LoadBalancerPoolInstanceStatuses.tsx
edlerd: It can only be instances from one project, so the instance name will be unique.

src/util/getHealthCheckType.ts
```diff
@@ -0,0 +1,15 @@
+import type { LxdLoadBalancerPool } from "types/loadBalancers";
+
+export const getHealthCheckType = (pool?: LxdLoadBalancerPool) => {
```
kimanhou: What happens if pool is undefined ?

src/util/isLoadBalancerLegacy.ts
```diff
@@ -0,0 +1,5 @@
+import type { LxdLoadBalancer } from "types/loadBalancers";
+
+export const isLoadBalancerLegacy = (loadBalancer: LxdLoadBalancer) => {
```
kimanhou: Should we combine getHealthCheckType, isLoadBalancerLegacy and setLoadBalancerCreatedPool into the same file and call it `util/loadBalancers.ts` ?

src/util/getHealthCheckType.ts
edlerd: Better make this function type safe without nullable!

src/util/isLoadBalancerLegacy.ts
edlerd: yes, good idea.

kimanhou (approved):
Awesome work !

PR #2016, opened 2026-05-26
feat(replicators): create/edit side panel WD-36612

omarelkashef:
## Done

- Add replicators create/edit side panel

## QA

1. Run the LXD-UI:
    - If clustered, go to Clustering -> Replicators, otherwise go to Server -> Replicators
    - Click create replicator.
    - Click edit replicator button to open the edit side panel.

## Screenshots

image
image
image


src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+  },
+  {
+    label: "Yes",
+    value: "true",
+  },
+  {
+    label: "No",
+    value: "false",
+  },
+];
```
edlerd: Can we reuse this from util?

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <ClusterLinkSelector
+        formik={formik}
+        itemName="target project"
+        clusterLinkType="bidirectional"
+      />
+      <Select
+        {...formik.getFieldProps("snapshot")}
+        label="Snapshot"
+        options={optionYesNo}
+        required
```
edlerd:
```suggestion
```
This should have a default and then it does not need the required indicator.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+          id="project"
+          name="project"
+          label="Project"
+          value={formik.values.project || ""}
+          setValue={(value) => {
+            formik.setFieldValue("project", value, false);
+          }}
+          disabled={isEdit}
+          projects={projects}
+          required
```
edlerd, line 88:
```suggestion
```
If the project has a default, then we do not need a required indicator.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,163 @@
...
+      },
+    } as Partial<LxdReplicator>;
+  };
+
+  const formik = useFormik<ReplicatorFormValues>({
+    initialValues: {
+      isCreating: true,
+      name: "",
+      description: "",
+      project: currentProject?.name || "default",
```
edlerd: We should select a default project that the user has correct permissions on. Or maybe we shouldn't select a default at all?

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+        />
+      )}
+
+      <ClusterLinkSelector
```
edlerd: Should this be required?

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+      />
+      {isEdit ? (
+        <OutputField
+          id="Project"
+          label="Project"
+          value={formik.values.project}
+          help="The project of a replicator cannot be changed."
+        />
+      ) : (
+        <ProjectSelector
```
edlerd, line 76: I would add help text explaining this is the source project to be replicated.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+        formik={formik}
+        itemName="target project"
+        clusterLinkType="bidirectional"
+      />
+      <Select
+        {...formik.getFieldProps("snapshot")}
+        label="Snapshot"
+        options={optionYesNo}
+        required
+        help="Whether to take a snapshot of the instances before replication or not."
```
edlerd: Can we explain why I would want to enable this or not?

edlerd (commented):
Good start, some comments on the UX of the form below.

kimanhou:
I merged [replicator's detail page](https://github.com/canonical/lxd-ui/pull/2017). Could you please connect the "Edit" button in the detail page to your edit side panel please ? 🙏

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+          id="project"
+          name="project"
+          label="Project"
+          value={formik.values.project || ""}
+          setValue={(value) => {
+            formik.setFieldValue("project", value, false);
+          }}
+          disabled={isEdit}
+          projects={projects}
+          help="The source project to be replicated. This list only includes projects for which you have permission to create replicators."
```
edlerd:
```suggestion
          help="Source project. Only projects you can create replicators for are listed."
```
A bit more concise.

src/pages/images/ClusterLinkSelector.tsx
```diff
@@ -5,20 +5,34 @@ import type { FormikProps } from "formik";
...
 import { ROOT_PATH } from "util/rootPath";
+import type { ReplicatorFormValues } from "types/forms/replicator";
 
 interface Props {
-  formik: FormikProps<ImageRegistryFormValues>;
+  formik:
+    | FormikProps<ImageRegistryFormValues>
+    | FormikProps<ReplicatorFormValues>;
+  required?: boolean;
+  itemType: "image-registry" | "replicator";
```
edlerd: I think the item type is a fairly non-generic concept to introduce in the selector. It might be simpler to duplicate the component for the registry and replicator use case. Currently it is very hard to reason in this component about which code path is actually hit.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -27,23 +41,27 @@ export const ClusterLinkSelector: FC<Props> = ({ formik }) => {
...
+      Cluster containing the {itemName}. Create your first{" "}
       <Link to={`${ROOT_PATH}/ui/cluster/links`}>cluster link</Link>.
     </>
   ) : (
     <>
-      Cluster containing the images. Manage your{" "}
+      Cluster containing the {itemName}. Manage your{" "}
       <Link to={`${ROOT_PATH}/ui/cluster/links`}>cluster links</Link>.
     </>
   );
```
edlerd, line 52: I think the help text for replicators is not ideal. "Cluster containing the target project" is not according to the users mental model. Better say "Cluster to replicate into" or something similar.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+        options={optionYesNo}
+        help={
+          <>
+            Whether to take a snapshot of the instances before replication or
+            not. Enable snapshots to{" "}
+            <DocLink docPath="/reference/storage_drivers/#optimized-volume-refresh">
+              optimize instance refresh
+            </DocLink>
+            .
+          </>
```
edlerd:
```suggestion
            Enable snapshots to{" "}
            <DocLink docPath="/reference/storage_drivers/#optimized-volume-refresh">
              optimize instance refresh
            </DocLink>
            .
          </>
```
I think we should push the explanation into the option. It could be "Yes, take new snapshot before sync" and "No, Sync using an existing snapshot, or automatically create one if needed"

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+        }
+      />
+      <Input
+        {...formik.getFieldProps("schedule")}
```
edlerd: Can we make the `SnapshotScheduleInput` generic and reuse it here?

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+      <Input type="submit" hidden value="Hidden input" />
+      {isEdit ? (
+        <OutputField
+          id="Name"
+          label="Name"
+          value={formik.values.name}
+          help={
+            <>
+              Click the name in the header of the{" "}
+              <Link to={`/ui/cluster/replicators/${formik.values.name}`}>
```
edlerd: The url is wrong here.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+        label="Description"
+        placeholder="Enter description"
+        error={getFieldError("description")}
+      />
+      {isEdit ? (
+        <OutputField
+          id="Project"
+          label="Project"
+          value={formik.values.project}
+          help="The project of a replicator cannot be changed."
```
edlerd:
```suggestion
          help="Project cannot be changed."
```
A bit more concise. Or maybe we don't need this at all?

edlerd (commented):
Some more thoughts on the UX of the creation form.

src/pages/cluster/actions/CreateReplicatorBtn.tsx
```diff
@@ -1,13 +1,24 @@
...
 import { Button, Icon } from "@canonical/react-components";
+import usePanelParams from "util/usePanelParams";
+import { useProjectEntitlements } from "util/entitlements/projects";
+import { useProjects } from "context/useProjects";
 
 export const CreateReplicatorButton: FC = () => {
+  const { openCreateReplicator } = usePanelParams();
+  const { data: allProjects = [] } = useProjects();
+  const { canCreateReplicators } = useProjectEntitlements();
+  const isDisabled = !allProjects.some((project) =>
```
edlerd, line 11: We should add a disable reason in the button title if it applies.

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -14,12 +14,17 @@ import type { LxdReplicator } from "types/replicator";
 import { useReplicatorEntitlements } from "util/entitlements/replicators";
 import { queryKeys } from "util/queryKeys";
 import { ROOT_PATH } from "util/rootPath";
+import classNames from "classnames";
 
 interface Props {
   replicator: LxdReplicator;
+  shouldExpand?: boolean;
```
```suggestion
  hasLabel?: boolean;
```
Probably a clearer naming.

src/pages/cluster/actions/EditReplicatorBtn.tsx
```diff
@@ -3,15 +3,22 @@ import { Button, Icon } from "@canonical/react-components";
 import classnames from "classnames";
 import type { LxdReplicator } from "types/replicator";
 import { useReplicatorEntitlements } from "util/entitlements/replicators";
+import usePanelParams from "util/usePanelParams";
 
 interface Props {
   replicator: LxdReplicator;
   className?: string;
+  shouldExpand?: boolean;
```
edlerd:
```suggestion
  hasLabel?: boolean;
```
Probably a clearer naming.

src/pages/cluster/actions/ReplicatorDetailActions.tsx
```diff
@@ -24,13 +24,15 @@ const ReplicatorDetailActions: FC<Props> = ({ replicator }) => {
...
+      shouldExpand
     />,
     <EditReplicatorBtn
       key="edit"
       replicator={replicator}
       className={classname}
+      shouldExpand
     />,
-    <DeleteReplicatorBtn key="delete" replicator={replicator} />,
+    <DeleteReplicatorBtn key="delete" replicator={replicator} shouldExpand />,
```
edlerd: There is a bug with the borders between edit and delete. It should be lighter, such as between run and edit:

Image

src/pages/cluster/actions/EditReplicatorBtn.tsx
```diff
@@ -3,15 +3,22 @@ import { Button, Icon } from "@canonical/react-components";
...
   onClose?: () => void;
 }
 
-const EditReplicatorBtn: FC<Props> = ({ replicator, className, onClose }) => {
```
edlerd, line 13: We have to have onClose on this button, but also on the create and delete, as we shrink the buttons to a dropdown on the detail page and there it needs to call close.

src/pages/cluster/actions/ReplicatorDetailActions.tsx
edlerd: Also in the contextual menu style the button is rendered wrongly

Image

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+          );
+        },
+      )
+      .matches(/^[A-Za-z0-9/\-:_.]+$/, {
+        message:
+          "Name can only contain alphanumeric, forward slash, hyphen, colon, underscore and full stop characters.",
+      })
+      .required("Replicator name is required."),
+    project: Yup.string().required("Project is required."),
+    cluster: Yup.string().required("Cluster is required."),
```
edlerd: The selectors do not show the error. I think when you highlight them and leave, we should display an error if no selection was made.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+    return {
+      name: formik.values.name,
+      description: formik.values.description,
+      config: {
+        cluster: formik.values.cluster,
+        schedule: formik.values.schedule,
+        snapshot: formik.values.snapshot ? "true" : "false",
+      },
+    } as Partial<LxdReplicator>;
+  };
```
edlerd:
```suggestion
  const getPayload = (): Partial<LxdReplicator> => {
    return {
      name: formik.values.name,
      description: formik.values.description,
      config: {
        cluster: formik.values.cluster,
        schedule: formik.values.schedule,
        snapshot: formik.values.snapshot ? "true" : "false",
      },
    };
  };
```
Better rely on the partial type than casting to it.

src/pages/cluster/panels/EditReplicatorPanel.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const getPayload = () => {
+    return {
+      description: formik.values.description,
+      config: {
+        cluster: formik.values.cluster,
+        schedule: formik.values.schedule,
+        snapshot: formik.values.snapshot ? "true" : "false",
+      },
+    } as Partial<LxdReplicator>;
+  };
```
edlerd:
```suggestion
```
Let's share this with the creation form.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -0,0 +1,67 @@
...
+}) => {
+  const { data: links = [], error } = useClusterLinks();
+  const notify = useNotify();
+  const hasNoLinks = links.length === 0;
+
+  if (error) {
+    notify.failure("Loading cluster links failed", error);
+  }
+
+  const options = links.map((link) => ({
```
edlerd, line 27: I think we need to filter bidirectinal links.

src/types/replicator.d.ts
```diff
@@ -10,7 +10,7 @@ export interface LxdReplicator {
   project: string;
   config: {
     cluster: string;
-    snapshot: boolean;
+    snapshot: string;
```
edlerd:
```suggestion
    snapshot: "true" | "false";
```
We can be a bit more precise here.

src/util/options.ts
edlerd, line 1: Good idea to make this not instance-bound.

src/util/usePanelParams.tsx
```diff
@@ -120,6 +125,8 @@ const usePanelParams = (): PanelHelper => {
     newParams.delete("local-peering");
     newParams.delete("create-image-registry");
     newParams.delete("image-registry");
+    newParams.delete("create-replicator");
+    newParams.delete("replicator");
```
edlerd, line 130: edit-replicator is missing here.

edlerd (commented):
Good progress!
Some comments from a detailed code review are below.

src/pages/cluster/actions/EditReplicatorBtn.tsx
```diff
@@ -22,20 +30,27 @@ const EditReplicatorBtn: FC<Props> = ({ replicator, className, onClose }) => {
...
+      )}
       onClick={() => {
-        console.log(
-          `Edit replicator ${replicator.name} btn clicked. TODO: open side panel`,
-          onClose,
-        );
+        openEditReplicator(replicator.project, replicator.name);
+        if (onClose) {
+          onClose();
+        }
```
edlerd:
```suggestion
        onClose?.();
```

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const queryClient = useQueryClient();
+  const controllerState = useState<AbortController | null>(null);
+  const { project: currentProject } = useCurrentProject();
+
+  const closePanel = () => {
+    panelParams.clear();
+    notify.clear();
+  };
+
+  const schema = Yup.object().shape({
```
edlerd, line 79: Let's add validation for:
- projects `replica.cluster` config should be empty
- user has edit permissions on the project.
- cluster link status should be `Reachable`

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,116 @@
...
+          id="project"
+          name="project"
+          label="Project"
+          value={formik.values.project || ""}
+          setValue={(value) => {
+            formik.setFieldValue("project", value, false);
+          }}
+          disabled={isEdit}
+          projects={projects}
+          help="Source project. Only projects you can create replicators for are listed."
```
edlerd:
```suggestion
          help="Source project."
```
Let's show all the projects here and solve this edge case with validation.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,116 @@
...
+  formik: FormikProps<ReplicatorFormValues>;
+  isEdit?: boolean;
+}
+
+export const ReplicatorForm: FC<Props> = ({ formik, isEdit = false }) => {
+  const { data: allProjects = [] } = useProjects();
+  const { canCreateReplicators } = useProjectEntitlements();
+  const projects = allProjects.filter((project) =>
+    canCreateReplicators(project),
+  );
```
edlerd:
```suggestion
```
Let's not filter, but let the user select any project they can view instead.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,148 @@
...
+  });
+
+  const formik = useFormik<ReplicatorFormValues>({
+    initialValues: {
+      isCreating: true,
+      name: "",
+      description: "",
+      project: "",
+      cluster: "",
+      snapshot: "false",
```
edlerd, line 149: I am not sure if this is the right default. What is the best option here?

edlerd (commented):
Thanks for the iterations! Nearly perfect, few things are left.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
omarelkashef: This will only complicate the code and frustrate the user. I think filtering based on this criteria with a help text is better than showing invalid options and giving the user an error when they select them.

edlerd: All right, agreed.

edlerd: The other two validations should still be added. Wdyt?

omarelkashef: Yes, I will include them.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
omarelkashef: For which field? For project and cluster, we do not know which ones are valid. Do you want to make snapshots true by default?

edlerd: I was referring to `snapthot: false`. This might be the right default or not. Maybe discuss this further to find the best default.

omarelkashef: The snapshot default for LXD it also false.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+      .matches(/^[A-Za-z0-9/\-:_.]+$/, {
+        message:
+          "Name can only contain alphanumeric, forward slash, hyphen, colon, underscore and full stop characters.",
+      })
+      .required("Replicator name is required."),
+    project: Yup.string()
+      .required("Project is required.")
+      .test(
+        "No existing replica cluster",
+        "The selected project already has replica.cluster configured.",
```
edlerd:
```suggestion
        "Project already has replica.cluster configured. Change it in the [project configuration].",
```
Let's add a deep link to project config to make this more actionable? Maybe in a new tab, so the user doesn't loose the form state.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+          );
+        },
+      )
+      .matches(/^[A-Za-z0-9/\-:_.]+$/, {
+        message:
+          "Name can only contain alphanumeric, forward slash, hyphen, colon, underscore and full stop characters.",
+      })
+      .required("Replicator name is required."),
+    project: Yup.string()
+      .required("Project is required.")
```
edlerd, line 101: When selecting a valid project, this error briefly shows up for 1 second. This shouldn't be visible in that case.

https://github.com/user-attachments/assets/fd62ee05-9b87-40da-9256-8e07f5f2d4ab

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+      )
+      .test(
+        "Can edit and create replicator",
+        "You need permissions to edit and create replicator on this project.",
+        (value) => hasProjectReplicatorPermissions(value),
+      ),
+    cluster: Yup.string()
+      .test(
+        "Reachable cluster link",
+        "The selected cluster link must be Reachable.",
```
edlerd:
```suggestion
        "The selected cluster link must be Reachable. Make sure a matching link has been created on the target cluster using the generated token.",
```
A bit more actionable?

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const closePanel = () => {
+    panelParams.clear();
+    notify.clear();
+  };
+
+  const getSelectedProject = (projectName?: string) => {
+    if (!projectName || projects.length === 0) {
+      return undefined;
+    }
+
```
edlerd:
```suggestion
```
This doesn't change anything, does it?

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+    project: Yup.string()
+      .required("Project is required.")
+      .test(
+        "No existing replica cluster",
+        "The selected project already has replica.cluster configured.",
+        (value) => hasNoReplicaClusterConfig(value),
+      )
+      .test(
+        "Can edit and create replicator",
+        "You need permissions to edit and create replicator on this project.",
```
edlerd:
```suggestion
        "You need the `edit` and `can_create_replicators` permission on the project.",
```
A bit more clear? Currently sounds like "edit replicator permission".

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+        async (value, context) => {
+          try {
+            return await isSelectedClusterReachable(
+              value,
+              queryClient,
+              isFineGrained,
+            );
+          } catch {
+            return context.createError({
+              message: "The selected cluster link must be Reachable.",
```
edlerd: We shouldn't duplicate the message here. Shall we use a const?

src/pages/cluster/panels/EditReplicatorPanel.tsx
```diff
@@ -0,0 +1,155 @@
...
+              queryClient,
+              isFineGrained,
+            );
+          } catch {
+            return context.createError({
+              message: "The selected cluster link must be Reachable.",
+            });
+          }
+        },
+      )
```
edlerd: Can we make this reusable between create and edit? Similar to `testDuplicateStorageVolumeName`

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { ROOT_PATH } from "util/rootPath";
+import type { ReplicatorFormValues } from "types/forms/replicator";
+
+interface Props {
+  formik: FormikProps<ReplicatorFormValues>;
+  required?: boolean;
+}
+
+export const ReplicatorClusterLinkSelector: FC<
+  Props & Omit<CustomSelectProps, "onChange" | "options" | "value">
```
edlerd: Should this be the exhaustive list of props that are set here? Or shall we simplify and allow all props, without omit?

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from "@canonical/react-components";
+import { useClusterLinks } from "context/useClusterLinks";
+import type { FormikProps } from "formik";
+import { Link } from "react-router-dom";
+import { ROOT_PATH } from "util/rootPath";
+import type { ReplicatorFormValues } from "types/forms/replicator";
+
+interface Props {
+  formik: FormikProps<ReplicatorFormValues>;
+  required?: boolean;
```
edlerd:
```suggestion
```
We don't need to make this configurable.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+      queryFn: async () => fetchClusterLinkState(clusterName),
+    }));
+
+  const identities =
+    queryClient.getQueryData<LxdIdentity[]>(identityQueryKey) ??
+    (await queryClient.fetchQuery({
+      queryKey: identityQueryKey,
+      queryFn: async () => fetchIdentities(isFineGrained),
+    })) ??
+    [];
```
edlerd:
```suggestion
```
We don't need the identity here, as the question at hand is if the link is reachable. Check the implementation of `isClusterLinkReachable`, which calls `getClusterLinksStatus`.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+    queryKeys.state,
+  ];
+  const identityQueryKey = [queryKeys.identities];
+
+  const state =
+    queryClient.getQueryData<LxdClusterLinkState>(stateQueryKey) ??
+    (await queryClient.fetchQuery({
+      queryKey: stateQueryKey,
+      queryFn: async () => fetchClusterLinkState(clusterName),
+    }));
```
edlerd: I would keep this simple, and do an `await fetchClusterLinkState`. No need for the query client here.

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { fetchClusterLinkState } from "api/cluster-links";
+import { fetchIdentities } from "api/auth-identities";
+import { getLinkIdentity, isClusterLinkReachable } from "util/clusterLink";
+
+interface Props {
+  formik: FormikProps<ReplicatorFormValues>;
+  isEdit?: boolean;
+}
+
+export const isSelectedClusterReachable = async (
```
edlerd: Why does this function live here, and `isClusterLinkReachable` is in the utils folder? I think we can use a single function in the utils folder and don't need the indirection here.

src/types/project.d.ts
```diff
@@ -1,8 +1,15 @@
 import type { LxdConfigPair } from "types/config";
 
+interface LxdProjectReplicaConfig {
+  cluster: string;
+  mode: string;
+}
```
edlerd:
```suggestion
```
This seems unused

src/types/project.d.ts
```diff
@@ -1,8 +1,15 @@
...
 export interface LxdProject {
   name: string;
-  config: LxdConfigPair;
+  config: LxdConfigPair & {
+    "replica.mode"?: string;
+    "replica.cluster"?: string;
+  };
```
edlerd: Do we need to mention those here? I think `LxdConfigPair` is generic enough.

edlerd (commented):
Good progress on the validation! Some thoughts on refining the UX there and a few simplifications.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
omarelkashef: The project configuration page does not have the replica yet. We can have the text and add the link later.

edlerd: Let's add the link to the configuration, we can then switch the link target.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
omarelkashef: I cannot recreate this behavior. Maybe it was solved accidentally with some refactoring. Let me know if you still experience it.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
omarelkashef: Do you mean simplify to Props & CustomSelectProps ?

edlerd: Yes, that would work.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
edlerd: Sure thing, I'll test again :+1:

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
omarelkashef: I think we do not need the onChange and options from the consuming component because we set them here.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
edlerd: confirm it is fixed now.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,245 @@
...
+    );
+  };
+
+  const notReachableErrorMessage =
+    "The selected cluster link must be Reachable. Make sure a matching link has been created on the target cluster using the generated token.";
+
+  const getReplicaClusterConfigErrorMessage = (projectName?: string) => {
+    return (
+      <>
+        Project `replica.cluster` must match the selected cluster link. Change
```
edlerd:
```suggestion
        Projects <code>replica.cluster</code> configuration must match the selected cluster. Change
```
Looks a little nicer.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
edlerd: > I think we do not need the onChange and options from the consuming component because we set them here.

Yes, then this should be the exhaustive list, including name, label, error, etc.

src/pages/cluster/panels/EditReplicatorPanel.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const queryClient = useQueryClient();
+  const replicatorName = panelParams.replicator ?? "";
+  const { data: replicator } = useReplicator(
+    replicatorName,
+    panelParams.project,
+    !!replicatorName,
+  );
+
+  const notReachableErrorMessage =
+    "The selected cluster link must be Reachable. Make sure a matching link has been created on the target cluster using the generated token.";
```
edlerd: Can we define this next to `testReachableClusterLink` so we reuse between create and edit?

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,245 @@
...
+          }
+
+          return this.createError({
+            message: getReplicaClusterConfigErrorMessage(value),
+          });
+        },
+      })
+      .test(
+        "Can edit and create replicator",
+        "You need the `edit` and `can_create_replicators` permission on the project.",
```
edlerd:
```suggestion
        "You need the <code>edit</code> and <code>can_create_replicators</code> permission on the project.",
```
This is not mattermost ;)

edlerd (commented):
Nearly perfect, few tiny issues left.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -0,0 +1,239 @@
...
+  };
+
+  const getSelectedProject = (projectName?: string) => {
+    return projects.find((project) => project.name === projectName);
+  };
+
+  const getReplicaClusterConfigErrorMessage = (projectName?: string) => {
+    return (
+      <>
+        Project <code>replica.cluster</code> must match the selected cluster
```
edlerd: @kimanhou is changing the representation of this slightly in the run model PR. Maybe best to merge this occasion as is. Then align it in a next step within the run modal PR.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const options = links.map((link) => ({
+    value: link.name,
+    text: link.name,
+    label: (
+      <div className="cluster-link-label">
+        <span className="cluster-link-name u-truncate" title={link.name}>
+          {link.name}
+        </span>
+      </div>
```
edlerd, line 38: Shall we display the link status here as well? Or we could simplify this component to use a standard HTML select if we decide to only keep the link name in here.

edlerd (dismissed):
QA and code LGTM, smaller non-blocking ideas below.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
omarelkashef: I think we can show the status so that the user can avoid the Not reachable validation error.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -0,0 +1,80 @@
...
+    text: link.name,
+    label: (
+      <div className="cluster-link-label">
+        <span className="cluster-link-name u-truncate" title={link.name}>
+          {link.name}
+        </span>
+        <span className="cluster-link-description u-text--muted">
+          {isLinkStatesLoading
+            ? "loading state..."
+            : getClusterLinksStatus(undefined, linkStates[index])}
```
edlerd: The indexes are not reliable enough to pick the value here. Arrays have no guarantee to be in the same order.

src/api/cluster-links.tsx
```diff
@@ -53,6 +53,18 @@ export const fetchClusterLinkState = async (
     });
 };
 
+export const fetchClusterLinkStates = async (
+  links: string[],
+): Promise<LxdClusterLinkState[]> => {
+  const fetchPromises = links.map(async (link) =>
+    fetch(`${ROOT_PATH}/1.0/cluster/links/${encodeURIComponent(link)}/state`)
+      .then(handleResponse)
+      .then((data: LxdApiResponse<LxdClusterLinkState>) => data.metadata),
```
edlerd: This can be very slow. I think a better approach is to reuse the `ClusterLinkStatus` component in the custom selector. So we load the state per link and not all at once and wait for the last one before showing any status.

edlerd (approved):
LGTM, thanks for the iterations!

PR #2017, opened 2026-05-26
feat(replicators): add detail page

kimanhou:
## Questions

- Which icon for replicators ? Omar used `connected`, David suggested `change-version` and I used `revisions`.
I agree with David on `change-version`.

image

image

image

## Done

- Create detail page for replicator

## QA

1. Run the LXD-UI:
    - Use LXD latest/edge.
    - If clustered go to Clustering -> Replicators, otherwise go to Server -> Replicators.
    - If not already done, create a replicator with the CLI: `lxc replicator create my-replicator cluster={my-cluster}` (replace {my-cluster} with your cluster link)
    - Make sure `my-replicator` appears in the replicator list. Click on my-replicator name, you should get to the URL `/ui/project/:project/replicator/:my-replicator` with project being my-replicator's project.
    - Explore the detail page and make sure the info are correct. Note that the action buttons in the header Run and Edit do not do anything.
    - Click on Delete. Make sure the replicator is correctly deleted

## Screenshots

image

Success notification after rename
image

Delete confirmation modal
Screenshot from 2026-05-29 16-52-45

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -85,6 +85,9 @@ const ProjectConfig = lazy(
   async () => import("pages/projects/ProjectConfiguration"),
 );
 const ProtectedRoute = lazy(async () => import("components/ProtectedRoute"));
+const ReplicatorDetail = lazy(
+  async () => import("pages/cluster/ReplicatorDetail"),
+);
 const ReplicatorList = lazy(async () => import("pages/cluster/ReplicatorList"));
 const Server = lazy(async () => import("pages/cluster/Server"));
 const Settings = lazy(async () => import("pages/settings/Settings"));
@@ -576,6 +579,10 @@ const App: FC = () => {
         <Route
           path={`${ROOT_PATH}/ui/cluster/replicators`}
           element={<ProtectedRoute outlet={<ReplicatorList />} />}
+        />
+        <Route
+          path={`${ROOT_PATH}/ui/project/:project/replicator/:name`}
+          element={<ProtectedRoute outlet={<ReplicatorDetail />} />}
         />
         <Route
           path={`${ROOT_PATH}/ui/operations`}
--- a/src/pages/cluster/ReplicatorDetail.tsx
+++ b/src/pages/cluster/ReplicatorDetail.tsx
@@ -1,0 +1,162 @@
+import type { FC } from "react";
+import { useParams } from "react-router-dom";
+import { Col, CustomLayout, Row, Spinner } from "@canonical/react-components";
+import NotFound from "components/NotFound";
+import NotificationRow from "components/NotificationRow";
+import ResourceLink from "components/ResourceLink";
+import { useClusterLink } from "context/useClusterLinks";
+import { useReplicator } from "context/useReplicators";
+import ClusterLinkStatus from "pages/cluster/ClusterLinkStatus";
+import ReplicatorDetailHeader from "pages/cluster/ReplicatorDetailHeader";
+import ReplicatorRunTime from "pages/cluster/ReplicatorRunTime";
+import ReplicatorStatus from "pages/cluster/ReplicatorStatus";
+import ProjectRichChip from "pages/projects/ProjectRichChip";
+import { ROOT_PATH } from "util/rootPath";
+
+const ReplicatorDetail: FC = () => {
+  const { project, name } = useParams<{
+    project: string;
+    name: string;
+  }>();
+
+  if (!name) {
+    return <>Missing name</>;
+  }
+
+  const {
+    data: replicator,
+    error,
+    isLoading,
+  } = useReplicator(name, project || "default");
+  const { data: clusterLink } = useClusterLink(
+    replicator?.config?.cluster || "",
+  );
+
+  if (isLoading) {
+    return <Spinner className="u-loader" text="Loading..." isMainComponent />;
+  }
+
+  if (!replicator) {
+    return (
+      <NotFound
+        entityType="replicator"
+        entityName={name}
+        errorMessage={error?.message}
+      />
+    );
+  }
+
+  return (
+    <>
+      <CustomLayout
+        header={
+          <ReplicatorDetailHeader
+            replicator={replicator}
+            project={project || "default"}
+          />
+        }
+      >
+        <NotificationRow />
+        <Row className="replicator-detail">
+          <Row className="section">
+            <Col size={3}>
+              <h2 className="p-heading--5">General</h2>
+            </Col>
+            <Col size={7}>
+              <table>
+                <tbody>
+                  <tr>
+                    <th className="u-text--muted">Description</th>
+                    <td>{replicator.description || "-"}</td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Source project</th>
+                    <td>
+                      <ProjectRichChip projectName={replicator.project} />
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+            </Col>
+          </Row>
+
+          <Row className="section">
+            <Col size={3}>
+              <h2 className="p-heading--5">Target</h2>
+            </Col>
+            <Col size={7}>
+              <table>
+                <tbody>
+                  <tr>
+                    <th className="u-text--muted">Cluster</th>
+                    <td>
+                      {replicator.config?.cluster ? (
+                        <div className="u-flex">
+                          <ResourceLink
+                            type="cluster-link"
+                            value={replicator.config?.cluster}
+                            to={`${ROOT_PATH}/ui/cluster/links`}
+                          />
+                        </div>
+                      ) : (
+                        "-"
+                      )}
+                    </td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Cluster status</th>
+                    <td className="status-cell">
+                      {clusterLink ? (
+                        <ClusterLinkStatus link={clusterLink} />
+                      ) : (
+                        "-"
+                      )}
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+            </Col>
+          </Row>
+
+          <Row className="section">
+            <Col size={3}>
+              <h2 className="p-heading--5">Sync</h2>
+            </Col>
+            <Col size={7}>
+              <table>
+                <tbody>
+                  <tr>
+                    <th className="u-text--muted">Fresh snapshots</th>
+                    <td>
+                      {replicator.config?.snapshot
+                        ? "Sync using an existing snapshot, or automatically create one if needed"
+                        : "Take new snapshot before sync"}
+                    </td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Schedule</th>
+                    <td>{replicator.config?.schedule || "-"}</td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Last run at</th>
+                    <td>
+                      <ReplicatorRunTime replicator={replicator} />
+                    </td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Status</th>
+                    <td className="status-cell">
+                      <ReplicatorStatus replicator={replicator} />
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+            </Col>
+          </Row>
+        </Row>
+      </CustomLayout>
+    </>
+  );
+};
+
+export default ReplicatorDetail;
--- a/src/pages/cluster/ReplicatorList.tsx
+++ b/src/pages/cluster/ReplicatorList.tsx
@@ -1,4 +1,5 @@
 import { type FC } from "react";
+import { Link } from "react-router-dom";
 import {
   Row,
   ScrollableTable,
@@ -15,13 +16,13 @@ import HelpLink from "components/HelpLink";
 import NotificationRow from "components/NotificationRow";
 import { useReplicators } from "context/useReplicators";
 import useSortTableData from "util/useSortTableData";
-import { CreateReplicatorButton } from "./actions/CreateReplicatorBtn";
+import { CreateReplicatorButton } from "pages/cluster/actions/CreateReplicatorBtn";
+import ReplicatorRunTime from "pages/cluster/ReplicatorRunTime";
+import ReplicatorStatus from "pages/cluster/ReplicatorStatus";
 import { useDocs } from "context/useDocs";
-import ReplicatorStatus from "./ReplicatorStatus";
 import ProjectRichChip from "pages/projects/ProjectRichChip";
 import { ROOT_PATH } from "util/rootPath";
 import ResourceLink from "components/ResourceLink";
-import { ReplicatorRunTime } from "./ReplicatorRunTime";
 
 interface Props {
   variant?: "main" | "panel";
@@ -46,7 +47,11 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
       content: "Cluster",
       sortKey: "cluster",
     },
-    { content: "Status", sortKey: "status", className: "status-header" },
+    {
+      content: "Status",
+      sortKey: "status",
+      className: "status-header",
+    },
     {
       content: "Last run at",
       sortKey: "last_run_at",
@@ -60,7 +65,14 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
       name: rowKey,
       columns: [
         {
-          content: replicator.name,
+          content: (
+            <Link
+              to={`${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicator.name)}`}
+              className="u-truncate"
+            >
+              {replicator.name}
+            </Link>
+          ),
           role: "rowheader",
           "aria-label": "Name",
           title: `Replicator ${replicator.name}`,
@@ -68,7 +80,7 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
         {
           content: (
             <div className="u-truncate" title={replicator.description}>
-              {replicator.description}
+              {replicator.description || "-"}
             </div>
           ),
           role: "cell",
--- a/src/pages/cluster/ReplicatorStatus.tsx
+++ b/src/pages/cluster/ReplicatorStatus.tsx
@@ -1,6 +1,7 @@
 import type { FC } from "react";
 import { Icon } from "@canonical/react-components";
 import type { LxdReplicator, LxdReplicatorStatus } from "types/replicator";
+
 interface Props {
   replicator: LxdReplicator;
 }
--- a/src/pages/cluster/actions/DeleteReplicatorBtn.tsx
+++ b/src/pages/cluster/actions/DeleteReplicatorBtn.tsx
@@ -1,0 +1,121 @@
+import { useState, type FC } from "react";
+import { useNavigate } from "react-router-dom";
+import {
+  ConfirmationButton,
+  Icon,
+  Notification,
+  useNotify,
+  useToastNotification,
+} from "@canonical/react-components";
+import { useQueryClient } from "@tanstack/react-query";
+import { deleteReplicator } from "api/replicators";
+import ResourceLabel from "components/ResourceLabel";
+import type { LxdReplicator } from "types/replicator";
+import { useReplicatorEntitlements } from "util/entitlements/replicators";
+import { queryKeys } from "util/queryKeys";
+import { ROOT_PATH } from "util/rootPath";
+
+interface Props {
+  replicator: LxdReplicator;
+  project: string;
+}
+
+const DeleteReplicatorBtn: FC<Props> = ({ replicator, project }) => {
+  const queryClient = useQueryClient();
+  const navigate = useNavigate();
+  const notify = useNotify();
+  const toastNotify = useToastNotification();
+  const [isLoading, setLoading] = useState(false);
+  const { canDeleteReplicator } = useReplicatorEntitlements();
+  const isReplicatorRunning = replicator.last_run_status === "Running";
+
+  const disabledReason = () => {
+    if (!canDeleteReplicator(replicator)) {
+      return "You do not have permission to delete this replicator";
+    }
+
+    return undefined;
+  };
+
+  const notifySuccess = (replicatorName: string) => {
+    toastNotify.success(
+      <>
+        Replicator{" "}
+        <ResourceLabel bold type="replicator" value={replicatorName} /> deleted.
+      </>,
+    );
+  };
+
+  const invalidateCache = () => {
+    queryClient.invalidateQueries({
+      predicate: (query) => query.queryKey[2] === queryKeys.replicators,
+    });
+  };
+
+  const onSuccess = () => {
+    invalidateCache();
+
+    // Only navigate to the replicators list if we are still on the deleted replicator's detail page
+    const replicatorDetailPath = `${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicator.name)}`;
+    if (location.pathname.startsWith(replicatorDetailPath)) {
+      navigate(`${ROOT_PATH}/ui/cluster/replicators`);
+    }
+
+    notifySuccess(replicator.name);
+  };
+
+  const onFailure = (e: unknown) => {
+    invalidateCache();
+    setLoading(false);
+    notify.failure(`Deleting replicator ${replicator.name} failed`, e);
+  };
+
+  const handleDelete = () => {
+    setLoading(true);
+    deleteReplicator(replicator.name, project).then(onSuccess).catch(onFailure);
+  };
+
+  return (
+    <ConfirmationButton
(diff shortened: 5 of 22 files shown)


edlerd:
> * How do we get the project from a replicator's name ? `fetchReplicator` takes the replicator's name and the project as parameters.

In that case the project should be in the url for a replicator detail page. Why not `/ui/project/:project/replicator/:replicator`?

> * Which icon for replicators ? Omar used `connected` and I used `revisions`.

I would additionally throw `change-version` into the discussion.
image

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <>
+        Replicator{" "}
+        <ResourceLabel bold type="replicator" value={replicatorName} /> deleted.
+      </>,
+    );
+  };
+
+  const invalidateCache = () => {
+    queryClient.invalidateQueries({
+      queryKey: [queryKeys.replicators, project],
```
edlerd: This is the wrong query key. Let's use a predicate to invalidate any queries where the third entry is `queryKeys.replicators` instead:
```
    queryClient.invalidateQueries({
      predicate: (query) =>
        query.queryKey[2] === queryKeys.replicators
    });
```

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -0,0 +1,129 @@
...
+              <>
+                <br />
+                <br />
+                <Notification
+                  severity="caution"
+                  title="Replicator is currently running."
+                >
+                  This replicator is actively transferring data. Deleting it now
+                  may result in incomplete instance snapshots on the target
+                  cluster.
```
edlerd: Is this really the case? We should test this with an active replicator.

src/pages/cluster/actions/EditReplicatorBtn.tsx
```diff
@@ -0,0 +1,32 @@
...
+import classnames from "classnames";
+import type { LxdReplicator } from "types/replicator";
+
+interface Props {
+  replicator: LxdReplicator;
+  className?: string;
+  onClose?: () => void;
+}
+
+const EditReplicatorBtn: FC<Props> = ({ replicator, className, onClose }) => {
```
edlerd, line 13: Let's add the "can edit" permission check.

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -0,0 +1,129 @@
...
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: (
+          <p>
+            This will permanently delete replicator{" "}
+            <ResourceLabel type="replicator" value={replicator.name} bold />.
+            <br />
+            This action cannot be undone. <br />
+            It will immediately cancel all future automated refreshes to the
+            remote cluster link. <br />
```
edlerd: Copy is a bit confusing "immediately cancel all future".

How about:

```
This permanently deletes the replicator :foo.

Automated refreshes to the remote cluster will stop.

Existing replicated instances on the target cluster remain available, but they will no longer receive updates.
```

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+            </Col>
+          </Row>
+
+          <Row className="section">
+            <Col size={3}>
+              <h2 className="p-heading--5">Sync configuration</h2>
```
edlerd: We should add the status of the replicator here.

yarn.lock
edlerd, line 1: Can we undo the changes to the lock file?

src/App.tsx
```diff
@@ -577,6 +580,10 @@ const App: FC = () => {
           path={`${ROOT_PATH}/ui/cluster/replicators`}
           element={<ProtectedRoute outlet={<ReplicatorList />} />}
         />
+        <Route
+          path={`${ROOT_PATH}/ui/project/:project/replicator/:name`}
```
edlerd, line 584: We need to add the prefix of this path to the "server" and "clustering/replicators" sections in the side navigation, so they stay active on this url.

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { useReplicator } from "context/useReplicators";
+import ReplicatorDetailHeader from "pages/cluster/ReplicatorDetailHeader";
+import ProjectRichChip from "pages/projects/ProjectRichChip";
+import { ROOT_PATH } from "util/rootPath";
+import { timeToString } from "util/helpers";
+import { useIdentities } from "context/useIdentities";
+import ClusterLinkStatus from "pages/cluster/ClusterLinkStatus";
+import { useClusterLink } from "context/useClusterLinks";
+
+const ReplicatorDetail: FC = () => {
```
edlerd, line 16: Can we surface further information such as the detailed status of the last replicator run here? Currently the page is fairly empty. I think it was discussed to make such information visible, but we need to follow up if this is available already or planned to be surfaced later on.

edlerd (commented):
Good work.

Few ideas to enhance below, but this should be good to merge soon.

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: I forgot to add the status of the replicator. In my original design, I wanted to put it in the header like for instances. Wdyt ?

yarn.lock
kimanhou: Sorry, leftover of my experimentations with Storybook

src/App.tsx
kimanhou: If it's just to have the menu item highlighted in the side navigation, I can just add `activeUrlMatches={["/replicator/"]}` to the `NavLink` in Navigation. Wdyt ?

edlerd: Yes, that could work.

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: I just discovered the existence of `last_run_status` field in the frontend, used by Omar but it is not mentioned in the backend spec. I added the replicator's status in the Sync configuration section as well.

image

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: I added it in the Sync configuration section for now

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
kimanhou: Actually it should not.

src/pages/cluster/ReplicatorStatus.tsx
```diff
@@ -1,20 +1,26 @@
...
+  );
+
+  if (!replicatorState) {
+    return "-";
+  }
 
-const ReplicatorStatus: FC<Props> = ({ replicator }) => {
-  const status = replicator.last_run_status;
-  const iconName = STATUS_ICONS[status] ?? "";
+  const status = replicatorState.status;
```
edlerd: This is confusing. Why is there a replicator state status and a last run status? What exactly do they both mean and should we show both?

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -91,7 +99,7 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
           "aria-label": "Cluster",
         },
         {
-          content: <ReplicatorStatus replicator={replicator} />,
+          content: <ReplicatorLastRunStatus replicator={replicator} />,
```
edlerd: I think we need to disambiguate what type of status we show in the table.

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,172 @@
...
+  );
+
+  const getLinkIdentity = (linkName: string | undefined) => {
+    if (!linkName) return undefined;
+    return identities.find(
+      (identity) =>
+        identity.name === linkName &&
+        identity.type.startsWith("Cluster link certificate"),
+    );
+  };
```
edlerd: Let's push this into the `ClusterLinkStatus` component, not to have to copy this to every use case of the component.

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -0,0 +1,126 @@
...
+    <ConfirmationButton
+      onHoverText={disabledReason()}
+      appearance="default"
+      className="u-no-margin--bottom has-icon"
+      loading={isLoading}
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: (
+          <>
+            <p className="u-no-max-width">
```
```suggestion
            <p>
```
Too wide text is hard to read. We should trust the design system here and not override this.

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -0,0 +1,126 @@
...
+              This will permanently delete replicator{" "}
+              <ResourceLabel type="replicator" value={replicator.name} bold />.
+              <br />
+              Scheduled refreshes to the standby cluster will stop.
+              <br />
+              Existing replicated instances on the standby cluster remain
+              available, but they will no longer receive updates.
```
```suggestion
              Instances on the standby cluster remain
              available, but they will no longer receive updates.
```
Can we make this even shorter while still being informative?

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,172 @@
...
+import { useReplicator } from "context/useReplicators";
+import ClusterLinkStatus from "pages/cluster/ClusterLinkStatus";
+import ReplicatorDetailHeader from "pages/cluster/ReplicatorDetailHeader";
+import ReplicatorLastRunStatus from "pages/cluster/ReplicatorLastRunStatus";
+import ReplicatorRunTime from "pages/cluster/ReplicatorRunTime";
+import ReplicatorStatus from "pages/cluster/ReplicatorStatus";
+import ProjectRichChip from "pages/projects/ProjectRichChip";
+import { ROOT_PATH } from "util/rootPath";
+
+const ReplicatorDetail: FC = () => {
```
edlerd, line 16: There are too many statuses on this page now.

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,172 @@
...
+                    <td>
+                      <ReplicatorRunTime replicator={replicator} />
+                    </td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Last run status</th>
+                    <td>
+                      <ReplicatorLastRunStatus replicator={replicator} />
+                    </td>
+                  </tr>
```
edlerd, line 151: Can we combine this into a last run section?

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,172 @@
...
+        }
+      >
+        <NotificationRow />
+        <Row>
+          <Row className="section">
+            <Col size={3}>
+              <h2 className="p-heading--5">General</h2>
+            </Col>
+            <Col size={7}>
+              <table>
```
edlerd, line 66: On very large screens, this layout leaves too much space between labels and values

Image

edlerd (commented):
Some thoughts on spacing, copy, and code structure are below.

src/pages/cluster/ReplicatorStatus.tsx
kimanhou: The replicator state status is similar to a cluster link status, in a sense that there is a separate API call to an endpoint that only returns the state. It's the live status of the replicator.
The last run status (from my understanding, it's not in the backend spec) is the status of the last run, so it's more of a historical data. 

Omar shows the last run status in the replicator list.

src/pages/cluster/ReplicatorList.tsx
kimanhou: Should we show both statuses here ?

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: I moved it in the util function as it was used elsewhere in ClusterLinkList.

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: Should we keep repeating "Last run" on each row ?

It is repetitive
image

It can be confusing
image

kimanhou: Wild suggestion
image

edlerd: I think the first one is best. It also resolves slightly the confusion over the many statuses.

src/pages/cluster/ReplicatorDetail.tsx
edlerd: I think the only remaining confusion is the status next to the cluster link. Maybe add a "Target cluster" section and a "Target cluster" and "Target cluster status" lines in it?

edlerd: ~~And in the "Sync configuration" section, change the "Status" label to "Sync status". Actually maybe we should simply call it "Sync".~~

Let's rename the "Sync configuration" section to "Sync". In it keep only the line with "Status", but call it "Sync status" together with the "Schedule" and "Fresh snapshots".

src/pages/cluster/ReplicatorList.tsx
edlerd: Maybe yes, two columns "Sync" and "Last run", both could show the status. Wdyt?

src/pages/cluster/ReplicatorStatus.tsx
edlerd: I see. We should take care for a new user the type of status is easily learnable from the labels we show.

kimanhou: I sent the question to backend, see what they think.

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: image


What about not repeating "target" in the labels ?
image

src/pages/cluster/ReplicatorList.tsx
kimanhou: Mmh

image

src/pages/cluster/ReplicatorDetail.tsx
edlerd: I think this is better -- without repeating target, wdyt?

Maybe We should label it "Target" instead of "Target cluster".

And should we have the status always last in the block?

kimanhou: > And should we have the status always last in the block?

Last or first ? Also shoud last run status be even displayed if "Last run at" = Never ?! It doesn't make any sense.

Kadin made me remove all "target" from "target cluster" when I first wrote the spec.
image

edlerd: I think naming the section "Target" still makes sense and is consistent with terminology, as we call the fields "Cluster" and "Cluster status".

edlerd: > Last or first ? Also shoud last run status be even displayed if "Last run at" = Never ?! It doesn't make any sense.

Status last or first: depends, main point was to make it consistent. Wdyt?

Last run status: I think we should show it for learnability that this information is available here.

kimanhou: Last is best. I replaced last run status by "-" if the replicator has never run.

image


That change also affect the replicator list
image

edlerd: Can you apply the status header class in the list? Then the header should align with the status text and not with the icon.

kimanhou: Oops sorry, missed that

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <>
+            <p>
+              This will permanently delete replicator{" "}
+              <ResourceLabel type="replicator" value={replicator.name} bold />.
+              <br />
+              Scheduled refreshes to the standby cluster will stop.
+              <br />
+              Instances on the standby cluster remain available, but they will
+              no longer receive updates.
+            </p>
```
edlerd, line 96:
```suggestion
            <p>
              This will permanently delete replicator{" "}
              <ResourceLabel type="replicator" value={replicator.name} bold />.
            </p>
            <p>
              Scheduled refreshes to the standby cluster will stop.
            </p>
            <p>
              Instances on the standby cluster remain available, but they will
              no longer receive updates.
            </p>
```
I think dedicated paragraphs make this easier to read. Wdyt?

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,186 @@
...
+                <tbody>
+                  <tr>
+                    <th className="u-text--muted">Description</th>
+                    <td>{replicator.description || "-"}</td>
```
edlerd, line 70: Let's align all the contents to start on a single line. Currently this is spread out to three different lines:

Image

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -46,7 +48,16 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
...
       sortKey: "cluster",
     },
-    { content: "Status", sortKey: "status", className: "status-header" },
+    {
+      content: "Sync",
+      sortKey: "sync status-header",
+      className: "status-header",
+    },
+    {
+      content: "Last run",
```
edlerd:
```suggestion
      content: "Last run status",
```
Should we be clearer here?

Also should this come after "last run at"?

edlerd (commented):
Good progress, few observations left.

src/pages/cluster/ReplicatorList.tsx
kimanhou: Actually, turned out last_run_status and the replicator's state status from calling the other endpoint are the same thing. So I discarded my changes and went back to Omar's version. Sorry !

src/pages/cluster/ReplicatorStatus.tsx
kimanhou: It's the same thing 😭

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: 3 ? I can see only 2. It's because I had a margin left on the cluster status, due to previous design where it was inline with the cluster name. It's now fixed.

image

edlerd: Let's add some left margin to the non-status contents, so they align nicely with the status text and not with the icon.

src/pages/cluster/ReplicatorList.tsx
edlerd: Thanks. Can you please follow up on why there are two endpoints serving the same information? I think one should be removed from the API.

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -0,0 +1,126 @@
...
+            {isReplicatorRunning && (
+              <Notification
+                severity="caution"
+                title="Replicator is currently running."
+              >
+                This replicator is actively transferring data. Deleting it now
+                will not cancel ongoing instance refresh operations.
+              </Notification>
+            )}
```
edlerd, line 105: Can we test this condition to make sure what we say here is correct?

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  if (!name) {
+    return <>Missing name</>;
+  }
+
+  const {
+    data: replicator,
+    error,
+    isLoading,
+  } = useReplicator(name, project || "default", true);
```
edlerd:
```suggestion
  } = useReplicator(name, project || "default");
```
No need to provide the default here.

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,162 @@
...
+              <h2 className="p-heading--5">Sync</h2>
+            </Col>
+            <Col size={7}>
+              <table>
+                <tbody>
+                  <tr>
+                    <th className="u-text--muted">Fresh snapshots</th>
+                    <td>
+                      {replicator.config?.snapshot
+                        ? "Use existing snapshot"
```
edlerd:
```suggestion
                        ? "Use existing snapshot if available"
```
Is this correct and maybe a bit more precise?

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        : "Take new snapshot before sync"}
+                    </td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Schedule</th>
+                    <td>{replicator.config?.schedule || "-"}</td>
```
edlerd, line 138: What is the format of the schedule? Should we provide a nicer formatting for displaying it here?

edlerd (commented):
Thanks, this looks mostly perfect. Few remaining questions then this can be merged.

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: Like this ?

image

src/pages/cluster/ReplicatorList.tsx
kimanhou: Originally there was only the replicator's state status (and I supposed that's why Omar added the function in the API), then Kadin added the last_run_status on the object itself. Kadin still keeps the separate API call because

> The state endpoint may grow to also return status of child operations in the future.

For now, `fetchReplicatorState` is useless in the UI so I removed it. I also removed the hook `useReplicatorState` and the type `LxdReplicatorState` since they were unused or could be replaced by `replicator.last_run_status`.

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: It's a cron expression. I believe we don't have anything for snapshot schedule, do we ?

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: Sent a question to backend team

src/pages/cluster/ReplicatorDetail.tsx
edlerd: Yeah, that was the idea. It is a bit nicer, no? Wdyt?

src/pages/cluster/ReplicatorList.tsx
edlerd: Perfect, thanks.

src/pages/cluster/ReplicatorDetail.tsx
edlerd: Then it is okay to keep as is :+1:

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: `Use existing snapshot if exists` ?
Should we add a precision, like `otherwise take new snapshot before sync` ?

src/pages/cluster/ReplicatorDetail.tsx
kimanhou: Yeah sure

src/pages/cluster/ReplicatorDetail.tsx
edlerd: Yes, agreed. Maybe better `otherwise take snapshot before sync`

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -0,0 +1,162 @@
...
+              <h2 className="p-heading--5">Sync</h2>
+            </Col>
+            <Col size={7}>
+              <table>
+                <tbody>
+                  <tr>
+                    <th className="u-text--muted">Fresh snapshots</th>
+                    <td>
+                      {replicator.config?.snapshot
+                        ? "Use existing snapshot if exists, otherwise take snapshot before sync"
```
edlerd:
```suggestion
                        ? "Sync using an existing snapshot, or automatically create one if needed."
```
Second thoughts to make this easier to read.

edlerd (approved):
QA and Code LGTM, we can iterate on the open delete notification question before or after merging as well.

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
kimanhou: I tested and it's correct ! 🥳

edlerd: Great, thanks!

PR #2019, opened 2026-05-27
Dashboard init: feature flag & instances card

kimanhou:
## Done

- Introduction of feature flags
- Draft for Instances card in Overview page

## QA

1. Run the LXD-UI:
    - Enable dashboard: in your browser console, run `localStorage.setItem("lxdui_ff_OVERVIEW", "true");`. Then refresh, you should see an `Overview` item in the Navigation.
    - Click on Overview in the side navigation
    - Admire Instances card draft

## Screenshots

Dark mode
image

Light mode
image

Error
image

Loading
image

Empty state (no instances)
image

Mobile
image


Kxiru:
Can we add icons to the entity headers? :)
Perhaps the header title could be "<icon><Entity name/type>(<number>)" Eg "## Instances (10)". Wdyt?

vosdev:
Wooohoooo a dashboard!


Will this be an overview for the current selected project or will it contain data for all projects ?  (Because you're showing the amount of projects)

kimanhou:
Hi @vosdev ! It's gonna be on the cluster member level, so all projects, not on the project level. I hope you will like it !

edlerd:
> Hi @vosdev ! It's gonna be on the cluster member level, so all projects, not on the project level. I hope you will like it 

@kimanhou you meant "cluster", not "cluster member" level, right?

kimanhou:
@edlerd mmh no I meant cluster member, hence the first card called "Cluster member" (similar info as cluster manager)

edlerd:
> @edlerd mmh no I meant cluster member, hence the first card called "Cluster member" (similar info as cluster manager)

I think on the overview page we should show combined information for all cluster members. Not information about one specific cluster member, there could be dozens of members, so it would be better to show the aggregate of all members in the dashboard.

kimanhou:
Okay. Would that be a new card ? Or would that replace the "Cluster member" card ?

edlerd:
> Okay. Would that be a new card ? Or would that replace the "Cluster member" card ?

I think a replacement would be reasonable. We can't display dozens of cluster member cards.

kimanhou:
@sophiecarroll Demo server is up ! 

I need guidance on:
- colors
- how to call this: dashboard or overview ?
- is adding a new item in the navigation okay or are there now too many items ?

kimanhou:
About the count in parenthesis, we already use this concept in the delete project modal:

image

src/context/useFeatureFlags.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const FeatureFlagContext = createContext<FeatureFlagContextType | undefined>(
+  undefined,
+);
+
+export const FeatureFlagProvider: React.FC<{ children: ReactNode }> = ({
+  children,
+}) => {
+  const isFeatureEnabled = (flag: FeatureFlag): boolean => {
+    return localStorage.getItem(`${LOCAL_STORAGE_PREFIX}${flag}`) === "true";
```
omarelkashef, line 17: How/where will the feature flags be set if we need to release a feature but disable it.?

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,219 @@
...
+        />
+        <InstancesDashboardStatus
+          status="pending"
+          instances={pendingInstances}
+        />
+        <InstancesDashboardStatus status="error" instances={errorInstances} />
+      </div>
+
+      <div className="group-by-type-container">
+        <DoughnutChart
```
omarelkashef: I wonder if the doughnut should be for the statuses not the type. To me, a doughnut is more about ratio/percentage and I would care more about the percentage of active instances in my cluster (how much of it is healthy) than the ratio of containers to VMs. That would also reduce the vertical space of the instances card because they are only 4 different statuses but 2 types.

omarelkashef (commented):
Really nice progress. A few comment below.

src/pages/overview/InstancesCard.tsx
kimanhou: @sophiecarroll Wdyt ?

src/context/useFeatureFlags.tsx
kimanhou: All feature flags are disabled by default. Unless you set the right thing in your localStorage, you won't be able to see the feature.

omarelkashef: I see, makes sense.

sophiecarroll:
Action items after Kim Anh/Sophie meeting:

Call this overview not dashboard on the UI (already done)
- Make sure there is an Overview nav item (already done)
- Make the logo tag clickable and it leads to Overview section 
- Make Overview the automatic landing page–on final push remember to change url to just a slash

Make card titles clickable and they lead to their navigation items
- Instances --> instances nav item (already done)
- Cluster--> clustering nav item (we should align these two names, ongoing research into cluster or clustering better) OR server state: Server-->server nav item
- Move instances card to the left of cluster/server card for better anchoring
- Change title of Storage Pools card to Storage, and move No. of storage volumes inside this card (keep the existing storage pool info too) so there is just one Storage card
- Create a card titled Memory to the right of the Storage card (where storage volumes used to be). This displays the data found in server->hardware->memory (or cluster version). This is aligned with competitive analysis and data hierarchy.

Additions after original post:
- Change "top 3 most full pools..." text to: "Most utilized storage pools". We don't have to say top three because the user can either see there are three items, or if there are less than three pools, they won't see three items. We don't have to say the % part either we can just include the (x%) after the name of the pool. If there are more than three, we should have a "more" link as a fourth line that links to the storage nav item.

kimanhou:
@sophiecarroll 

✅ Make sure there is an Overview nav item (already done)
✅ Make the logo tag clickable and it leads to Overview section
🟧 Make Overview the automatic landing page–on final push remember to change url to just a slash -> [JIRA ticket](https://warthogs.atlassian.net/browse/WD-37378) created and planned for August
🟧 Make card titles clickable and they lead to their navigation items -> done for instances, will be done at the same time as each card is implemented
🟧 Cluster--> clustering nav item (we should align these two names, ongoing research into cluster or clustering better) OR server state: Server-->server nav item -> will be done when this card is implemented, [JIRA ticket](https://warthogs.atlassian.net/browse/WD-36624) updated
✅ Move instances card to the left of cluster/server card for better anchoring
✅ Change title of Storage Pools card to Storage, and move No. of storage volumes inside this card (keep the existing storage pool info too) so there is just one Storage card
✅ Create a card titled Memory to the right of the Storage card (where storage volumes used to be). This displays the data found in server->hardware->memory (or cluster version). This is aligned with competitive analysis and data hierarchy.
🟧 Change "top 3 most full pools..." text to: "Most utilized storage pools". We don't have to say top three because the user can either see there are three items, or if there are less than three pools, they won't see three items. We don't have to say the % part either we can just include the (x%) after the name of the pool. If there are more than three, we should have a "more" link as a fourth line that links to the storage nav item. -> will be done when this card is implemented.

src/pages/overview/InstancesCard.tsx
Kxiru: I agree that statuses would be more beneficial to view than types, and would be a better use of the doughtnut chart.

Kxiru (commented):
Love love love the progress on this, as well as the theming.
- I also think that Cluster should be renamed to "Clustering" as an overview.

kimanhou:
> I also think that Cluster should be renamed to "Clustering" as an overview.

@Kxiru Not related to this PR. You can update the [JIRA ticket](https://warthogs.atlassian.net/browse/WD-36624) or the spec.

sophiecarroll:
> Love love love the progress on this, as well as the theming.
> 
> * I also think that Cluster should be renamed to "Clustering" as an overview.

Agree with this @Kxiru and I will comment it on cluster ticket now

sophiecarroll:
Is it possible to add a green button to the Instances error state that says "create an issue on github," if this is what users see when they get an error in other parts of the application?

sophiecarroll:
Update to colors for this card: let's make sure the donut is true grey and not blue-grey as right now (not in Canonical design guidelines). Two shades of grey that are similar in hue/contrast and included in the design guidelines: make lighter part of donut #C5C5C5 and darker part #636363 (hex codes).

kimanhou:
> if this is what users see when they get an error in other parts of the application?

That is not what users see when they get an error in other parts of the application. Only the error is displayed.

Only in the `ErrorPage` and the `NotFound` page, there is a link, not a button, to create a Github issue. 
The `ErrorPage` appears when a page/component in the routed app throws while rendering. It does not appear for every failure, it does not handle ordinary handled API failures that are shown as notifications for example.

kimanhou:
> make lighter part of donut #C5C5C5 and darker part #636363 (hex codes).

In both dark and light modes ?

sophiecarroll:
Screenshot 2026-06-29 at 12 13 10 PM
Is there a way to control for overflow behavior? eg. when I have my window narrow can you make the donut text stack vertically instead of getting cut off?

kimanhou:
@sophiecarroll 

> Is there a way to control for overflow behavior? eg. when I have my window narrow can you make the donut text stack vertically instead of getting cut off?

I created a mobile version of the card. For tablet, the card remains the same.

src/components/Logo.tsx
```diff
@@ -20,15 +18,8 @@ const Logo: FC<Props> = ({ light }) => {
...
-  const getLogoLink = () => {
-    if (isLoading || !project) {
-      return `${ROOT_PATH}/ui/`;
-    }
-    return `${ROOT_PATH}/ui/project/${encodeURIComponent(project.name)}`;
-  };
-
   return (
-    <NavLink className="p-panel__logo" to={getLogoLink()}>
+    <NavLink className="p-panel__logo" to={`${ROOT_PATH}/ui/overview`}>
```
edlerd: This change needs to be feature flagged

src/components/Navigation.tsx
```diff
@@ -277,6 +279,21 @@ const Navigation: FC = () => {
...
+                      {isOverviewEnabled() && (
+                        <SideNavigationItem>
+                          <NavLink
+                            to={`${ROOT_PATH}/ui/overview`}
+                            title={`Overview (${projectName})`}
+                            onClick={softToggleMenu}
+                          >
+                            <Icon
+                              className="is-light p-side-navigation__icon"
+                              name="switcher-dashboard"
```
edlerd, line 291: Do we have opinions on the icon? Is this the one landscape UI is using as well?

src/context/useFeatureFlags.tsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  const isOverviewEnabled = () => isFeatureEnabled("OVERVIEW");
```
edlerd:
```suggestion
  const exposedFunctions = {
    isOverviewEnabled: () => isFeatureEnabled("OVERVIEW");
  };
```
It might be a bit simpler to extend if we inline the functions and avoid using constants. That reduces the places you need to touch for adding/removing a new feature flag to a single line (this one).

src/context/useFeatureFlags.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { createContext, type ReactNode, useContext } from "react";
+
+// To add a new feature flag, add it to the FeatureFlag type below
+// For example: export type FeatureFlag = "OVERVIEW" | "NEW_FEATURE";
+// Then add a corresponding function in the FeatureFlagProvider
+// For example: const isNewFeatureEnabled = () => isFeatureEnabled("NEW_FEATURE");
+export type FeatureFlag = "OVERVIEW";
+
```
edlerd:
```suggestion
```
We might not need this, see other comment.

src/sass/_overview.scss
```diff
@@ -0,0 +1,109 @@
...
+              border-color: #c18f02;
+            }
+
+            &.pending {
+              border-color: #6e7d8a;
+            }
+
+            &.error {
+              border-color: var(--vf-color-button-negative-default);
+            }
```
edlerd, line 55: We should use the themed colors, matching the `InstanceStatusIcon` components outputs. Those are currently not aligned. For instance "stopped" should be black in the light theme and white in the dark theme.

Pending is not a valid state AFAIK. I think we should have four states: Running, Stopped, Frozen, Error in that order. We should not add new states or redefine them here.

src/App.tsx
```diff
@@ -148,6 +150,7 @@ const App: FC = () => {
   } = useAuth();
   const notify = useNotify();
   const { data: settings } = useSettings();
+  const { isOverviewEnabled } = useFeatureFlags();
```
edlerd, line 155: Should the default page on login also be changed if this is enabled to take you to the overview?

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+const INITIAL_ACCUMULATOR: InstanceAccumulator = {
+  runningInstances: [],
+  stoppedInstances: [],
+  pendingInstances: [],
+  errorInstances: [],
+  containers: [],
+  virtualMachines: [],
+};
+
+const InstancesCard: FC = () => {
```
edlerd, line 21: It would be nice to also link to a filter view. A user is likely going to explore the instances in error status. Clicking on the count or label should allow you to explore which instances are behind the numbers given for each status.

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+import type { LxdInstance, LxdInstanceStatus } from "types/instance";
+import { capitalizeFirstLetter } from "util/helpers";
+import { ROOT_PATH } from "util/rootPath";
+
+interface InstancesOverviewStatusProps {
+  status: "running" | "stopped" | "pending" | "error";
+  instances: LxdInstance[];
+}
+
+const InstancesOverviewStatus: FC<InstancesOverviewStatusProps> = ({
```
edlerd: This should be in a dedicated file

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+    if (instances.length === 0) {
+      return INITIAL_ACCUMULATOR;
+    }
+
+    return instances.reduce<InstanceAccumulator>(
+      (accumulator, instance) => {
+        const { status, type } = instance;
+
+        if (status === "Running") {
+          accumulator.runningInstances.push(instance);
```
edlerd: Shall we collect the numbers here instead of an array with all instances? Seems like overkill to collect the whole objects in arrays, when all we need later on is the count.

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  const {
+    runningInstances,
+    stoppedInstances,
+    pendingInstances,
+    errorInstances,
+    containers,
+    virtualMachines,
+  } = useMemo(() => {
+    if (instances.length === 0) {
```
edlerd: I think this accumulation/counting logic should be in utils.

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+const PENDING_STATUSES = new Set<LxdInstanceStatus>([
+  "Restarting",
+  "Starting",
+  "Stopping",
+  "Migrating",
+]);
+const ERROR_STATUSES = new Set<LxdInstanceStatus>([
+  "Error",
+  "Freezing",
+  "Frozen",
```
edlerd: Those two are not error statuses, we should expose them as frozen instead.

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+  containers: LxdInstance[];
+  virtualMachines: LxdInstance[];
+}
+
+const PENDING_STATUSES = new Set<LxdInstanceStatus>([
+  "Restarting",
+  "Starting",
+  "Stopping",
+  "Migrating",
+]);
```
edlerd: I am not sure we should introduce a new pending status group just for the overview. Though it can be interesting, especially when this number is high as it indicates a certain load on the cluster where a lot of instances are in transition. Then again the backend doesn't give us these statuses, they are only tracked by the UI or figured out by the UI through matching running operations to instances.

I think we might not expose those, but instead the number of running operations should be a signal about heavy load.

edlerd (commented):
This looks great, especially the feature flag logic is sound and simple. Few thoughts and suggestions below.

src/components/Navigation.tsx
kimanhou: I like the icon ! Yes, it's the same as Landscape.

src/components/Logo.tsx
```diff
@@ -12,23 +13,27 @@ interface Props {
...
       return `${ROOT_PATH}/ui/`;
     }
     return `${ROOT_PATH}/ui/project/${encodeURIComponent(project.name)}`;
   };
 
+  const src = isMicroCloud
+    ? `${ROOT_PATH}/ui/assets/img/microCloud-logo.svg`
+    : `${ROOT_PATH}/ui/assets/img/lxd-logo.svg`;
+  const heading = isMicroCloud ? "MicroCloud" : "LXD";
+
```
edlerd: Why move this around? Can we keep this in the original position?

src/App.tsx
```diff
@@ -212,9 +215,11 @@ const App: FC = () => {
...
               <Navigate
                 to={
-                  hasNoProjects || defaultProject === ALL_PROJECTS
-                    ? `${ROOT_PATH}/ui/all-projects/instances`
-                    : `${ROOT_PATH}/ui/project/${encodeURIComponent(defaultProject)}/instances`
+                  isOverviewEnabled()
+                    ? `${ROOT_PATH}/ui/overview`
+                    : hasNoProjects || defaultProject === ALL_PROJECTS
+                      ? `${ROOT_PATH}/ui/all-projects/instances`
+                      : `${ROOT_PATH}/ui/project/${encodeURIComponent(defaultProject)}/instances`
```
edlerd: Let's not nest the ternaries and turn this into a function with early exits to keep it readable.

src/util/overviewInstances.ts
```diff
@@ -0,0 +1,55 @@
+import { type LxdInstanceStatus, type LxdInstance } from "types/instance";
+
+export interface InstanceOverviewCounts {
```
edlerd:
```suggestion
export interface InstanceDistribution {
```

src/util/overviewInstances.ts
```diff
@@ -0,0 +1,55 @@
...
+  runningCount: number;
+  stoppedCount: number;
+  frozenCount: number;
+  errorCount: number;
+  containerCount: number;
+  virtualMachineCount: number;
+}
+
+const ERROR_INSTANCE_STATUSES = new Set<LxdInstanceStatus>(["Error"]);
```
edlerd:
```suggestion
```
This can be a simple comparision instead.

src/util/overviewInstances.ts
```diff
@@ -0,0 +1,55 @@
...
+const INITIAL_INSTANCE_OVERVIEW_COUNTS: InstanceOverviewCounts = {
+  runningCount: 0,
+  stoppedCount: 0,
+  frozenCount: 0,
+  errorCount: 0,
+  containerCount: 0,
+  virtualMachineCount: 0,
+};
+
+export const getOverviewInstanceCounts = (
```
edlerd:
```suggestion
export const getInstanceDistribution = (
```
Let's refine the naming here and of the file as well. The concept at hand is not tied to the overview.

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,183 @@
...
+  } = useMemo<InstanceOverviewCounts>(
+    () => getOverviewInstanceCounts(instances),
+    [instances],
+  );
+
+  const cardClassName = "overview-card instances";
+  const cardTitle = (
+    <Link soft href={`${ROOT_PATH}/ui/all-projects/instances`}>
+      <Icon name="pods" /> Instances
+      {!isLoading && !error && instances.length > 0 && ` (${instances.length})`}
```
edlerd, line 41: Question for visual design: To me the icon looks slightly too big for the font.

src/pages/overview/InstancesOverviewStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <strong className="status-count">{count}</strong>
+    </>
+  );
+
+  return (
+    <div className={`group-by-status ${status}`}>
+      {href ? (
+        <Link className="status-link" href={href} soft>
+          {content}
+        </Link>
```
edlerd: The underline for the combined label and value over two lines is slightly odd. I wonder if we should only link the number instead.

edlerd (commented):
Great progress, few more thoughts on the changes.

src/pages/overview/InstancesCard.tsx
kimanhou: @sophiecarroll Wdyt ?

src/pages/overview/InstancesOverviewStatus.tsx
kimanhou: @sophiecarroll Wdyt ?

src/pages/overview/InstancesCard.tsx
sophiecarroll: I agree–I also think the bottom of the logo is slightlyyyy higher than the bottom of the text, so make logo size align with top and bottom of text

src/pages/overview/InstancesOverviewStatus.tsx
sophiecarroll: I think we should change the links to the system I proposed in Mattermost because of the debate about linking titles–a "see more" link on the lower righthand side that leads you to the instances nav item in this case, making the title just normal unlinked text

src/pages/overview/InstancesCard.tsx
kimanhou: image

src/pages/overview/InstancesOverviewStatus.tsx
kimanhou: We are talking about what happens when you hover over a status and a count:
image

sophiecarroll: I think @edlerd comment to only underline the number makes sense. So no underline on the word

src/pages/overview/InstancesOverviewStatus.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Link } from "@canonical/react-components";
+import { capitalizeFirstLetter } from "util/helpers";
+
+interface Props {
+  status: "running" | "stopped" | "frozen" | "error";
+  count: number;
+  href: string;
+}
+
+const InstancesOverviewStatus: FC<Props> = ({ status, count, href }) => {
```
edlerd: I think the spacing of the number and label is not correct. The number sits closer to the next label than to the one that it refers to. Either those margins should be equal, or the number should be close to the relating label and a bit more space to the next one.

Image

edlerd: I think without margin on the labels, would be best:

Image

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,187 @@
...
+          />
+          <InstancesOverviewStatus
+            status="error"
+            count={errorCount}
+            href={getStatusFilterHref("Error")}
+          />
+        </div>
+
+        <div className="group-by-type-container" onClick={handleDoughnutClick}>
+          <DoughnutChart
```
edlerd, line 89: We can treat this as a follow-up. Maybe only report it upstream. There is a bug in the component where the label stops following the mouse.

https://github.com/user-attachments/assets/61252ab4-7bdf-4a0c-b5b4-4d4bce87eb34

src/util/overviewInstances.ts
```diff
@@ -0,0 +1,53 @@
...
+  containerCount: number;
+  virtualMachineCount: number;
+}
+
+const FROZEN_INSTANCE_STATUSES = new Set<LxdInstanceStatus>([
+  "Freezing",
+  "Frozen",
+]);
+
+const INITIAL_INSTANCE_OVERVIEW_COUNTS: InstanceDistribution = {
```
edlerd:
```suggestion
const INITIAL_INSTANCE_DISTRIBUTION: InstanceDistribution = {
```

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,187 @@
...
+            {virtualMachineCount} {pluralize("VM", virtualMachineCount)}
+            <br />
+            <span
+              className="dot"
+              style={{
+                backgroundColor: containerColor,
+              }}
+            ></span>
+            {containerCount} {pluralize("container", containerCount)}
+          </span>
```
edlerd: This is not using the right semantic structure. Let's define a new `ChartLegend` component. It should use a list and list items within instead of nested span elements. We do have a similar implementation in [cluster manager](https://github.com/canonical/microcloud-cluster-manager/blob/main/ui/src/pages/clusters/metrics/ClusterDetailMemberGraph.tsx#L64-L87), which we should probably generalize to reuse here.

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  if (error) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <div className="error-message">
+          <Icon name="error" className="margin-right--large" /> Error while
+          loading instances: {error.message}
+        </div>
+      </Card>
```
edlerd, line 66: Why do we need an extra nested div element here? Can we only have the card instead?

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+    if (segmentIndex === 0) {
+      navigate(getTypeFilterHref("VM"));
+      return;
+    }
+
+    if (segmentIndex === 1) {
+      navigate(getTypeFilterHref("Container"));
+    }
+  };
```
edlerd: This is an indirect way of implementing the links. Let's extend the `DoughnutChart` `segments` to accept an optional `href` instead and handle the linking more directly inside. Wdyt?

edlerd (commented):
A few more concerns on the proposed changes. I like the refined style.

src/pages/overview/InstancesOverviewStatus.tsx
kimanhou: I forgot to remove a container in CSS when I fixed the mobile version yesterday. It should be back to its original state now:
image

src/pages/overview/InstancesCard.tsx
kimanhou: Something like this ?
https://github.com/canonical/react-components/pull/1393

src/pages/overview/InstancesCard.tsx
kimanhou: I can't reproduce it. Could you please create a Github issue in `react-components` ?

src/sass/_overview.scss
```diff
@@ -0,0 +1,92 @@
...
+            &.frozen {
+              border-color: var(--vf-color-link-default);
+            }
+
+            &.error {
+              border-color: var(--vf-color-button-negative-default);
+            }
+
+            p.status-label {
+              margin-bottom: $spv--small;
```
edlerd:
```suggestion
              margin-bottom: 0;
```
I think placing the value closer to the label helps readability. Wdyt?

src/pages/overview/InstancesCard.tsx
```diff
@@ -0,0 +1,181 @@
...
+    </>
+  );
+  const vmColor = "#C5C5C5";
+  const containerColor = "#636363";
+
+  const getStatusFilterHref = (status: LxdInstanceStatus) => {
+    const params = new URLSearchParams();
+    params.append("status", status);
+    return `${ROOT_PATH}/ui/all-projects/instances?${params.toString()}`;
+  };
```
edlerd: This can move into `InstancesOverviewStatus`.

edlerd (approved):
LGTM, thanks :)

PR #2024, opened 2026-06-01
chore(CI): fix playwright install browsers

kimanhou:
## Done

- Fix CI: CI was blocked at the step `Install Playwright browsers`, this step successfully downloaded the binaries but then got stuck at extraction time. All the tests would then time out after 6 hours and be automatically canceled.

Explanation by Gemini: The underlying culprit was a version incompatibility between Node.js 24 and the upstream file stream extraction utilities utilized internally by Playwright (specifically yauzl). When executing on the modern Ubuntu 24.04 kernel layout under Node 24, the extraction worker thread would enter an unhandled infinite synchronization loop rather than flushing the unzipped binary assets to disk.

As suggested by @edlerd, I upgraded Playwright.

## QA

1. Run the LXD-UI:
    - Trust CI passing

## Screenshots

image

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -101,7 +101,7 @@ jobs:
           curl --head --fail --retry-delay 2 --retry 100 --retry-connrefused --insecure https://localhost:8407
 
           yarn test-e2e-cluster-coverage
-          
+
           # combine coverage reports
           yarn test-report-coverage
           zip -r coverage/playwright-report/cobertura-coverage.zip coverage/playwright-report/cobertura-coverage.xml
@@ -151,7 +151,7 @@ jobs:
           run: |
             git add .
             git commit -m "workflow: update coverage report"
-            
+
             # In case of another action job pushing to gh-pages while we are rebasing for the current job
             while true; do
               git pull --rebase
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -222,10 +222,14 @@ jobs:
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 24
+          node-version: 22
 
       - name: Install Playwright Browsers
-        run: npx playwright install --with-deps ${{ matrix.browser }}
+        env:
+          DEBIAN_FRONTEND: noninteractive
```
edlerd: Is it this bit that is newly necessary? I guess we can undo the other changes then?

kimanhou: No, I hadn't it cleaned it up yet. I am still trying something (that's why it's still in draft)

edlerd: I think updating playwright might be the right solution. See also https://github.com/garrytan/gstack/issues/1703

edlerd (approved):
LGTM, thanks :)

PR #2027, opened 2026-06-01
feat(replicators): Run modal

kimanhou:
## Done

- Replicator: run modal

## QA

1. Run the LXD-UI:
    - Create 2 VMs with LXD latest edge installed. We are going to call them lxd1 (leader) and lxd2 (standby).
    - On lxd1 and lxd2, create auth group with admin rights. In the following commands, the auth group is called `link`:
    ```
    lxc auth group create link
    llxc auth group permission add link server admin
    ```

    - Create a cluster link between lxd1 and lxd2:
        - On lxd1, run `lxc cluster link create link-to-lxd2 --auth-group link`. This command returns a token, copy it.
        - On lxd2, run `lxc cluster link create link-to-lxd1 --token <token> --auth-group link`
    - Create projects for replication:
        - On lxd1, run `lxc project create polar-penguin -c replica.cluster=link-to-lxd2`
        - On lxd2, run `lxc project create polar-penguin -c replica.cluster=link-to-lxd1`
    - Setup projects for replication:
        - On **lxd2**, run `lxc project set polar-penguin replica.mode=standby`
        - On lxd1, run `lxc project set polar-penguin replica.mode=leader`
    - Create replicators:
        - On lxd1, run `lxc replicator create replicator-alligator cluster=link-to-lxd2 --project polar-penguin`
        - On lxd2, run `lxc replicator create repligator cluster=link-to-lxd1 --project polar-penguin`

You now have 2 replicators: `replicator-alligator` on lxd1 where the local project is leader and `repligator` on lxd2 where the local project is standby.

## Screenshots

Local project is in leader mode
image

Local project is in standby mode
image

Status on detail page is automatically updated
https://github.com/user-attachments/assets/fd33ff23-dd0a-472d-891e-708272ea19a0

Error handling when the operation fails
https://github.com/user-attachments/assets/d030fc76-3ec0-43ef-b24c-cd2c990246eb

### DIFF
--- a/src/Events.tsx
+++ b/src/Events.tsx
@@ -134,6 +134,20 @@ const Events: FC = () => {
     }
   };
 
+  const updateReplicators = (event: LxdEvent) => {
+    const isReplicatorOperation =
+      event.type === "operation" &&
+      event.metadata.description.toLowerCase().includes("replicator");
+
+    if (!isReplicatorOperation) {
+      return;
+    }
+
+    queryClient.invalidateQueries({
+      predicate: (query) => query.queryKey[2] === queryKeys.replicators,
+    });
+  };
+
   const reEvaluateOperations = () => {
     operations.forEach((operation) => {
       handleEvent(operation.id, operation.status, operation.err, operation);
@@ -206,6 +220,7 @@ const Events: FC = () => {
           });
         }
         updateClusterMemberLoading(event);
+        updateReplicators(event);
         // ensure open requests that reply with an operation and register
         // new handlers in the eventQueue are closed before handling the event
         setTimeout(() => {
--- a/src/api/replicators.tsx
+++ b/src/api/replicators.tsx
@@ -1,6 +1,7 @@
 import { handleResponse } from "util/helpers";
 import type { LxdReplicator } from "types/replicator";
 import type { LxdApiResponse } from "types/apiResponse";
+import type { LxdOperationResponse } from "types/operation";
 import { addEntitlements } from "util/entitlements/api";
 import { ROOT_PATH } from "util/rootPath";
 
@@ -100,4 +101,25 @@ export const updateReplicator = async (
       body: body,
     },
   ).then(handleResponse);
+};
+
+export const runReplicator = async (
+  name: string,
+  project: string,
+  action: "restore" | "start",
+): Promise<LxdOperationResponse> => {
+  return fetch(
+    `${ROOT_PATH}/1.0/replicators/${encodeURIComponent(name)}/state?project=${encodeURIComponent(project)}`,
+    {
+      method: "PUT",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ action }),
+    },
+  )
+    .then(handleResponse)
+    .then((data: LxdOperationResponse) => {
+      return data;
+    });
 };
--- a/src/context/replicatorLoading.tsx
+++ b/src/context/replicatorLoading.tsx
@@ -1,0 +1,201 @@
+import {
+  createContext,
+  useContext,
+  useState,
+  type FC,
+  type ReactNode,
+} from "react";
+import { useOperations } from "context/operationsProvider";
+import type { LxdOperation } from "types/operation";
+import type { LxdReplicator, LxdReplicatorStatus } from "types/replicator";
+import { getOperationEntityUrls } from "util/operations";
+
+interface ReplicatorLoadingType {
+  getStatus: (replicator: LxdReplicator) => LxdReplicatorStatus;
+  getLastRunError: (replicator: LxdReplicator) => string | undefined;
+  setRunning: (replicator: LxdReplicator) => void;
+  setFinish: (replicator: LxdReplicator) => void;
+  setLastRunError: (replicator: LxdReplicator, error: string) => void;
+  clearLastRunError: (replicator: LxdReplicator) => void;
+}
+
+interface Props {
+  children: ReactNode;
+}
+
+const ReplicatorRunningContext = createContext<ReplicatorLoadingType>({
+  getStatus: (replicator: LxdReplicator) => replicator.last_run_status,
+  getLastRunError: () => undefined,
+  setRunning: () => undefined,
+  setFinish: () => undefined,
+  setLastRunError: () => undefined,
+  clearLastRunError: () => undefined,
+});
+
+const getReplicatorKey = ({
+  name,
+  project,
+}: Pick<LxdReplicator, "name" | "project">) => `${project}/${name}`;
+
+const parseReplicatorKeyFromUrl = (url: string): string | null => {
+  try {
+    const parsed = new URL(url, "http://localhost");
+    const pathParts = parsed.pathname.split("/").filter(Boolean);
+    const replicatorIndex = pathParts.indexOf("replicators");
+
+    if (replicatorIndex === -1 || !pathParts[replicatorIndex + 1]) {
+      return null;
+    }
+
+    const name = decodeURIComponent(pathParts[replicatorIndex + 1]);
+    const project = decodeURIComponent(
+      parsed.searchParams.get("project") || "default",
+    );
+
+    return getReplicatorKey({ name, project });
+  } catch {
+    return null;
+  }
+};
+
+const getOperationReplicatorKeys = (operation: LxdOperation): string[] => {
+  return getOperationEntityUrls(operation, [])
+    .map(parseReplicatorKeyFromUrl)
+    .filter((key): key is string => Boolean(key));
+};
+
+const isRunningReplicatorOperation = (operation: LxdOperation): boolean => {
+  return operation.description === "Running replicator";
+};
+
+const isFailedReplicatorOperation = (operation: LxdOperation): boolean => {
+  return (
+    operation.description === "Running replicator" &&
+    operation.status === "Failure"
+  );
+};
+
+const getOperationTimestamp = (operation: LxdOperation): number => {
+  const timestamp = Date.parse(
+    operation.updated_at || operation.created_at || "",
+  );
+  return Number.isNaN(timestamp) ? 0 : timestamp;
+};
+
+export const ReplicatorRunningProvider: FC<Props> = ({ children }) => {
+  const { runningOperations, operations } = useOperations();
+  const [manualRunningKeys, setManualRunningKeys] = useState(new Set<string>());
+  const [manualLastRunErrors, setManualLastRunErrors] = useState(
+    new Map<string, string>(),
+  );
+
+  const setRunning = (replicator: LxdReplicator) => {
+    const key = getReplicatorKey(replicator);
+    setManualRunningKeys((prev) => {
+      if (prev.has(key)) {
+        return prev;
+      }
+      const next = new Set(prev);
+      next.add(key);
+      return next;
+    });
+  };
+
+  const setFinish = (replicator: LxdReplicator) => {
+    const key = getReplicatorKey(replicator);
+    setManualRunningKeys((prev) => {
+      if (!prev.has(key)) {
+        return prev;
+      }
+      const next = new Set(prev);
+      next.delete(key);
+      return next;
+    });
+  };
+
+  const setLastRunError = (replicator: LxdReplicator, error: string) => {
+    const key = getReplicatorKey(replicator);
+    setManualLastRunErrors((prev) => {
+      const next = new Map(prev);
+      next.set(key, error);
+      return next;
+    });
+  };
+
+  const clearLastRunError = (replicator: LxdReplicator) => {
+    const key = getReplicatorKey(replicator);
+    setManualLastRunErrors((prev) => {
+      const next = new Map(prev);
+      next.delete(key);
+      return next;
+    });
+  };
+
+  const getStatus = (replicator: LxdReplicator): LxdReplicatorStatus => {
+    const key = getReplicatorKey(replicator);
+    const hasRunningOperation = runningOperations
+      .filter(
+        (operation) =>
+          isRunningReplicatorOperation(operation) &&
+          operation.status === "Running",
+      )
+      .some((operation) => getOperationReplicatorKeys(operation).includes(key));
+
+    if (manualRunningKeys.has(key) || hasRunningOperation) {
+      return "Running";
+    }
+
+    return replicator.last_run_status;
+  };
+
+  const getLastRunError = (replicator: LxdReplicator): string | undefined => {
+    if (getStatus(replicator) !== "Failed") {
+      return undefined;
+    }
+
+    const key = getReplicatorKey(replicator);
+    let latestOperationError: string | undefined;
+    let latestOperationTimestamp = -1;
+
+    for (const operation of operations) {
+      if (
+        !isFailedReplicatorOperation(operation) ||
+        !operation.err ||
+        !getOperationReplicatorKeys(operation).includes(key)
+      ) {
+        continue;
+      }
+
+      const timestamp = getOperationTimestamp(operation);
+      if (timestamp >= latestOperationTimestamp) {
+        latestOperationTimestamp = timestamp;
+        latestOperationError = operation.err;
+      }
+    }
+
+    return (
+      replicator.last_run_error ||
+      latestOperationError ||
+      manualLastRunErrors.get(key)
+    );
+  };
+
+  return (
+    <ReplicatorRunningContext.Provider
+      value={{
+        getStatus,
+        getLastRunError,
+        setRunning,
+        setFinish,
+        setLastRunError,
+        clearLastRunError,
+      }}
+    >
+      {children}
+    </ReplicatorRunningContext.Provider>
+  );
+};
+
+export const useReplicatorLoading = (): ReplicatorLoadingType => {
+  return useContext(ReplicatorRunningContext);
+};
--- a/src/pages/cluster/ReplicatorDetail.tsx
+++ b/src/pages/cluster/ReplicatorDetail.tsx
@@ -1,18 +1,19 @@
-import type { FC } from "react";
+import { type FC } from "react";
 import { useParams } from "react-router-dom";
 import { Col, CustomLayout, Row, Spinner } from "@canonical/react-components";
 import NotFound from "components/NotFound";
 import NotificationRow from "components/NotificationRow";
 import { useClusterLink } from "context/useClusterLinks";
+import { useReplicatorLoading } from "context/replicatorLoading";
 import { useReplicator } from "context/useReplicators";
+import ClusterLinkRichChip from "pages/cluster/ClusterLinkRichChip";
 import ClusterLinkStatus from "pages/cluster/ClusterLinkStatus";
 import ReplicatorDetailHeader from "pages/cluster/ReplicatorDetailHeader";
 import ReplicatorRunTime from "pages/cluster/ReplicatorRunTime";
 import ReplicatorStatus from "pages/cluster/ReplicatorStatus";
+import EditReplicatorPanel from "pages/cluster/panels/EditReplicatorPanel";
 import ProjectRichChip from "pages/projects/ProjectRichChip";
-import ClusterLinkRichChip from "./ClusterLinkRichChip";
 import usePanelParams, { panels } from "util/usePanelParams";
-import { EditReplicatorPanel } from "./panels/EditReplicatorPanel";
 
 const ReplicatorDetail: FC = () => {
   const { project, name } = useParams<{
@@ -31,9 +32,14 @@ const ReplicatorDetail: FC = () => {
     error,
     isLoading,
   } = useReplicator(name, project || "default");
+  const replicatorRunning = useReplicatorLoading();
   const { data: clusterLink } = useClusterLink(
     replicator?.config?.cluster || "",
   );
+
+  const lastRunError = replicator
+    ? replicatorRunning.getLastRunError(replicator)
+    : undefined;
 
   if (isLoading) {
     return <Spinner className="u-loader" text="Loading..." isMainComponent />;
@@ -142,6 +148,10 @@ const ReplicatorDetail: FC = () => {
                     <td className="status-cell">
                       <ReplicatorStatus replicator={replicator} />
                     </td>
+                  </tr>
+                  <tr>
+                    <th className="u-text--muted">Last error</th>
+                    <td>{lastRunError ?? "-"}</td>
                   </tr>
                 </tbody>
               </table>
--- a/src/pages/cluster/actions/RunReplicatorBtn.tsx
+++ b/src/pages/cluster/actions/RunReplicatorBtn.tsx
@@ -1,7 +1,27 @@
-import type { FC } from "react";
-import { Button, Icon } from "@canonical/react-components";
+import { useState, type FC } from "react";
+import {
+  ConfirmationButton,
+  Icon,
+  Notification,
+  useNotify,
+  useToastNotification,
+} from "@canonical/react-components";
+import { useQueryClient } from "@tanstack/react-query";
+import { runReplicator } from "api/replicators";
 import classnames from "classnames";
+import ConfirmationCheckbox from "components/ConfirmationCheckbox";
+import ResourceLink from "components/ResourceLink";
+import { useReplicatorLoading } from "context/replicatorLoading";
+import { useProject } from "context/useProjects";
+import { useEventQueue } from "context/eventQueue";
+import RunReplicatorPreflightChecks from "pages/cluster/actions/RunReplicatorPreflightChecks";
+import ClusterLinkRichChip from "pages/cluster/ClusterLinkRichChip";
+import ProjectRichChip from "pages/projects/ProjectRichChip";
 import type { LxdReplicator } from "types/replicator";
+import { useReplicatorEntitlements } from "util/entitlements/replicators";
+import { isProjectReplicaModeStandby } from "util/projects";
+import { queryKeys } from "util/queryKeys";
+import { ROOT_PATH } from "util/rootPath";
 
 interface Props {
   replicator: LxdReplicator;
@@ -13,31 +33,220 @@ interface Props {
 const RunReplicatorBtn: FC<Props> = ({
   replicator,
   className,
-  onClose,
   hasLabel = false,
+  onClose,
 }) => {
-  return (
-    <Button
-      appearance={hasLabel ? "default" : "base"}
-      aria-label="Run replicator"
-      className={classnames(
-        "u-no-margin--bottom has-icon",
-        {
-          "is-dense": !hasLabel,
-        },
-        className,
-      )}
-      onClick={() => {
-        console.log(
-          `Run replicator ${replicator.name} btn clicked. TODO: open modal`,
-          onClose,
+  const queryClient = useQueryClient();
+  const eventQueue = useEventQueue();
+  const notify = useNotify();
+  const toastNotify = useToastNotification();
+  const replicatorLoading = useReplicatorLoading();
+  const [isLoading, setLoading] = useState(false);
+  const [canSubmit, setCanSubmit] = useState(false);
+  const { canEditReplicator } = useReplicatorEntitlements();
+  const { data: project, isLoading: isProjectLoading } = useProject(
+    replicator.project,
+  );
+  const [isOverwriteConfirmed, setIsOverwriteConfirmed] = useState(false);
+
+  const clusterLink = replicator.config.cluster;
+  const isRestore = isProjectReplicaModeStandby(project);
+  const buttonLabel = isRestore ? "Restore" : "Run";
+  const hasPermission = canEditReplicator(replicator);
+
+  const disabledReason = () => {
+    if (!hasPermission) {
+      return "You do not have permission to run this replicator";
+    }
+    if (isLoading) {
+      return "Replicator is starting...";
+    }
+    if (replicatorLoading.getStatus(replicator) === "Running") {
+      return "This replicator is currently running";
+    }
+    return undefined;
+  };
+
+  const invalidateCache = () => {
+    queryClient.invalidateQueries({
+      predicate: (query) => query.queryKey[2] === queryKeys.replicators,
+    });
+  };
+
+  const onReplicationStarted = () => {
+    setLoading(false);
+    replicatorLoading.clearLastRunError(replicator);
+    replicatorLoading.setRunning(replicator);
+    invalidateCache();
+
+    toastNotify.info(
+      <>
+        Replicator{" "}
+        <ResourceLink
+          type="replicator"
+          value={replicator.name}
+          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicator.name)}`}
+        />{" "}
+        started.
+      </>,
+    );
+  };
+
+  const onReplicatorStartFailure = (e: unknown) => {
+    setLoading(false);
+    replicatorLoading.setFinish(replicator);
+    invalidateCache();
+    notify.failure(`Running replicator ${replicator.name} failed`, e);
+  };
+
+  const onOperationSuccess = () => {
+    replicatorLoading.setFinish(replicator);
+    replicatorLoading.clearLastRunError(replicator);
+    toastNotify.success(
+      <>
+        Replicator{" "}
+        <ResourceLink
+          type="replicator"
+          value={replicator.name}
+          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicator.name)}`}
+        />{" "}
+        completed successfully.
+      </>,
+    );
+    invalidateCache();
+  };
+
+  const onOperationFailure = (msg: string) => {
+    replicatorLoading.setFinish(replicator);
+    replicatorLoading.setLastRunError(replicator, msg);
+    invalidateCache();
+    toastNotify.failure(
+      `Replicator${replicator.name} failed while running`,
+      new Error(msg),
+    );
+  };
+
+  const handleRun = () => {
+    setLoading(true);
+    runReplicator(
+      replicator.name,
+      replicator.project,
+      isRestore ? "restore" : "start",
+    )
+      .then((response) => {
+        onReplicationStarted();
+        eventQueue.set(
+          response.metadata.id,
+          onOperationSuccess,
+          onOperationFailure,
+          invalidateCache,
         );
+      })
+      .catch(onReplicatorStartFailure);
+  };
+
+  return (
+    <ConfirmationButton
+      onHoverText={disabledReason()}
+      appearance={hasLabel ? "default" : "base"}
+      className={classnames("u-no-margin has-icon", className, {
+        "is-dense": !hasLabel,
+      })}
+      loading={isLoading || isProjectLoading}
+      confirmationModalProps={{
+        className: "run-replicator-modal",
+        title: (
+          <>
+            {isRestore ? "Restore" : "Run"} replicator{" "}
+            <ResourceLink
+              type="replicator"
+              value={replicator.name}
+              to={`${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicator.name)}`}
+            />
+          </>
+        ),
+        children: (
+          <>
+            <RunReplicatorPreflightChecks
+              replicator={replicator}
+              onValidationChange={(isValid) => {
+                setCanSubmit(isValid);
+              }}
+              isRestore={isRestore}
+            />
+
+            <div
+              className={classnames("replicator-data-flow u-hide--small", {
+                "replicator-data-flow--restore": isRestore,
+              })}
+            >
+              <div className="replicator-data-flow__node replicator-data-flow__source">
+                <span className="node-heading u-text--muted">
+                  {isRestore ? "standby" : "leader"} project (local)
+                </span>
+                <ProjectRichChip projectName={replicator.project} />
+              </div>
+              <div className="replicator-data-flow__arrow-shaft">
+                <ClusterLinkRichChip clusterLink={clusterLink} />
+              </div>
+              <div className="replicator-data-flow__node replicator-data-flow__target">
+                <span className="node-heading u-text--muted">
+                  {isRestore ? "leader" : "standby"} project (remote)
+                </span>
+                <strong className="mono-font">{replicator.project}</strong>
+              </div>
+            </div>
+
+            {isRestore ? (
+              <p>
+                This will sync all instances from the{" "}
+                <ClusterLinkRichChip clusterLink={clusterLink} /> cluster back
+                to the <ProjectRichChip projectName={replicator.project} />{" "}
+                project.
+              </p>
+            ) : (
+              <p>
+                This will sync all instances from the source project{" "}
+                <ProjectRichChip projectName={replicator.project} /> to the
+                standby cluster{" "}
+                <ClusterLinkRichChip clusterLink={clusterLink} />.
+              </p>
+            )}
+
+            {isRestore && (
+              <Notification
+                severity="caution"
+                messageElement="div"
+                title="All local data will be overwritten by the remote version"
+              >
+                <p>
+                  All current instances in local project{" "}
+                  <ProjectRichChip projectName={replicator.project} /> will be
+                  permanently replaced by the version fetched from the remote
+                  cluster. This cannot be undone.
+                </p>
+              </Notification>
+            )}
+          </>
+        ),
+        onConfirm: handleRun,
+        confirmButtonLabel: buttonLabel,
+        confirmButtonDisabled:
+          !canSubmit || isLoading || (isRestore && !isOverwriteConfirmed),
+        close: onClose,
+        confirmExtra: isRestore ? (
+          <ConfirmationCheckbox
+            label="Overwrite local data"
+            confirmed={[isOverwriteConfirmed, setIsOverwriteConfirmed]}
+            disabled={!canSubmit || isLoading}
+          />
+        ) : undefined,
       }}
-      title="Run replicator"
+      disabled={Boolean(disabledReason())}
     >
       <Icon name="play" />
-      {hasLabel && <span>Run</span>}
-    </Button>
+      {hasLabel && <span>{buttonLabel}</span>}
+    </ConfirmationButton>
   );
 };
 
--- a/src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
+++ b/src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
@@ -1,0 +1,177 @@
+import { useEffect, type FC } from "react";
+import { Link } from "react-router";
+import { List } from "@canonical/react-components";
+import { useClusterLinkState } from "context/useClusterLinks";
+import { useIdentities } from "context/useIdentities";
+import { useInstances } from "context/useInstances";
+import { useProject } from "context/useProjects";
+import ClusterLinkRichChip from "pages/cluster/ClusterLinkRichChip";
+import RunReplicatorPreflightCheckRow, {
+  type PreflightCheck,
+} from "pages/cluster/actions/RunReplicatorPreflightCheckRow";
+import ProjectRichChip from "pages/projects/ProjectRichChip";
+import type { LxdReplicator } from "types/replicator";
+import { getClusterLinksStatus, getLinkIdentity } from "util/clusterLink";
+import { pluralize } from "util/helpers";
+
+interface Props {
+  replicator: LxdReplicator;
+  onValidationChange: (isValid: boolean) => void;
+  isRestore?: boolean;
+}
+
+const RunReplicatorPreflightChecks: FC<Props> = ({
+  replicator,
+  onValidationChange,
+  isRestore = false,
+}) => {
+  const { data: project, isLoading: isProjectLoading } = useProject(
+    replicator.project,
+  );
+  const { data: instances, isLoading: isInstancesLoading } = useInstances(
+    replicator.project,
+  );
+  const { data: identities = [] } = useIdentities();
+  const { data: clusterLinkState, isLoading: isClusterLinkStateLoading } =
+    useClusterLinkState(replicator.config?.cluster || "");
+  const identity = getLinkIdentity(identities, replicator.config?.cluster);
+
+  const checks: PreflightCheck[] = [];
+
+  const projectMode = project?.config?.["replica.mode"];
+  const projectModeValid =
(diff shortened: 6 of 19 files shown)


src/pages/cluster/actions/ReplicatorDetailActions.tsx
```diff
@@ -15,9 +15,7 @@ interface Props {
 
 const ReplicatorDetailActions: FC<Props> = ({ replicator }) => {
   const isSmallScreen = useIsScreenBelow(largeScreenBreakpoint);
-  const classname = isSmallScreen
-    ? "p-contextual-menu__link"
-    : "p-segmented-control__button";
+  const classname = isSmallScreen ? "p-contextual-menu__link" : undefined;
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -1,31 +1,176 @@
...
+        children: (
+          <>
+            <p>
+              The project <ProjectRichChip projectName={replicator.project} />{" "}
+              is in{" "}
+              <strong>{project?.config["replica.mode"] || "unknown"}</strong>{" "}
+              mode.
+              <br />
+              {/* TODO: use new API endoint to promote/demote a project
+              https://github.com/canonical/lxd/pull/18312
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -1,31 +1,176 @@
...
+          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicatorName)}`}
+        />{" "}
+        started.
+      </>,
+    );
+  };
+
+  const onSuccess = () => {
+    notifySuccess(replicator.name);
+  };
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -1,31 +1,176 @@
...
-      onClick={() => {
-        console.log(
-          `Run replicator ${replicator.name} btn clicked. TODO: open modal`,
-          onClose,
-        );
+      className={classnames("u-no-margin has-icon", className)}
+      loading={isLoading}
+      confirmationModalProps={{
+        className: "run-replicator-modal",
+        title: (
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -1,31 +1,176 @@
...
+import { ROOT_PATH } from "util/rootPath";
 
 interface Props {
   replicator: LxdReplicator;
   className?: string;
-  onClose?: () => void;
 }
 
-const RunReplicatorBtn: FC<Props> = ({ replicator, className, onClose }) => {
+const RunReplicatorBtn: FC<Props> = ({ replicator, className }) => {
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -1,31 +1,176 @@
...
+  const disabledReason = () => {
+    if (!canDeleteReplicator(replicator)) {
+      return "You do not have permission to run this replicator";
+    }
+
+    if (replicator.last_run_status === "Running") {
+      return "This replicator is currently running";
+    }
+
+    return undefined;
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -1,31 +1,176 @@
...
+  const [isLoading, setLoading] = useState(false);
+  const { canDeleteReplicator } = useReplicatorEntitlements();
+  const { data: project } = useProject(replicator.project);
+  const clusterLink = replicator.config?.cluster;
+  const isReverseFlow = isProjectReplicaModeStandby(project);
+
+  const disabledReason = () => {
+    if (!canDeleteReplicator(replicator)) {
+      return "You do not have permission to run this replicator";
+    }
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -1,31 +1,176 @@
...
 
-const RunReplicatorBtn: FC<Props> = ({ replicator, className, onClose }) => {
+const RunReplicatorBtn: FC<Props> = ({ replicator, className }) => {
+  const notify = useNotify();
+  const toastNotify = useToastNotification();
+  const [isLoading, setLoading] = useState(false);
+  const { canDeleteReplicator } = useReplicatorEntitlements();
+  const { data: project } = useProject(replicator.project);
+  const clusterLink = replicator.config?.cluster;
+  const isReverseFlow = isProjectReplicaModeStandby(project);
```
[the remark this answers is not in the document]
kimanhou: Addressed

src/pages/cluster/actions/EditReplicatorBtn.tsx
```diff
@@ -24,7 +24,7 @@ const EditReplicatorBtn: FC<Props> = ({ replicator, className, onClose }) => {
     <Button
       appearance="default"
       aria-label="Edit replicator"
-      className={classnames("u-no-margin--bottom has-icon", className)}
+      className={classnames("u-no-margin has-icon", className)}
       onClick={() => {
```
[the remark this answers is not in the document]
kimanhou: No, the button has an atuomatic margin-right so u-no-margin is needed

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -10,22 +24,169 @@ interface Props {
...
+        />{" "}
+        started.
+      </>,
+    );
+  };
+
+  const onSuccess = () => {
+    setLoading(false);
+    notifySuccess(replicator.name);
+  };
```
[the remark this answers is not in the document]
kimanhou: It's actually a good point. It might also be worth implementing a manual refresh button in the detail page.

src/api/replicators.tsx
```diff
@@ -67,3 +68,28 @@ export const deleteReplicator = async (
...
+  return fetch(
+    `${ROOT_PATH}/1.0/replicators/${encodeURIComponent(name)}/state?project=${encodeURIComponent(project)}`,
+    {
+      method: "PUT",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(body),
+    },
+  )
```
edlerd:
```suggestion
  action: "restore" | "start",
): Promise<LxdOperationResponse> => {
  return fetch(
    `${ROOT_PATH}/1.0/replicators/${encodeURIComponent(name)}/state?project=${encodeURIComponent(project)}`,
    {
      method: "PUT",
      headers: {
        "Content-Type": "application/json",
      },
      body: JSON.stringify({ action }),
    },
  )
```
Let's streamline the type and simplify this.

src/context/useReplicators.tsx
```diff
@@ -2,6 +2,36 @@ import { useQuery } from "@tanstack/react-query";
...
+  return {
+    ...newReplicator,
+    // Keep client-side error captured from operation events while API has no dedicated field yet.
+    last_run_error:
+      newReplicator.last_run_error ??
+      (newReplicator.last_run_status === "Failed"
+        ? oldReplicator.last_run_error
+        : undefined),
+  };
+};
```
edlerd: We should get this fixed upstream and remove the workaround here.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -10,22 +27,289 @@ interface Props {
...
+  const disabledReason = () => {
+    if (isProjectLoading) {
+      return "Loading project information...";
+    }
+
+    if (!project) {
+      return "Unable to load project information.";
+    }
+    if (!project.config["replica.mode"]) {
+      return "Project replica mode is not configured.";
```
edlerd: The copy should be more concrete on what to do for unblocking this.

Also I think on replica creation we should set the project replica mode automatically.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -10,22 +27,289 @@ interface Props {
...
+            ? {
+                ...currentReplicator,
+                last_run_status: "Running",
+                last_run_error: undefined,
+              }
+            : currentReplicator,
+        );
+      },
+    );
+  };
```
edlerd: I wouldn't mess with the API state here amending it with custom info. For instances we have a similar problem with the "starting" and "stopping" statuses. We established an extra global context, see `useInstanceLoading`. Maybe we should push this information here (and the last run error) into a dedicated `useReplicatorRunning` context. Wdyt?

edlerd (commented):
Tested the setup locally and it works fine. Some thoughts on the state handling and copy below.

src/pages/cluster/actions/RunReplicatorBtn.tsx
kimanhou: > Also I think on replica creation we should set the project replica mode automatically.

This info is for @omarelkashef 
That is possible for the local leader project. However, how can we achieve that for the remote standby project ?
What do we do if the replica already exists ? Do we overwrite the existing config ?

edlerd: Yes, I think we need a bit of custom validation for replica creation:

# Validation on replica creation / Project selector

Let the user select any project. This let's us remove parts of the current help text below the project selector to keep that more simple.

After selection of a project, do the following checks:

1) Do they have write permissions on the project? If not, show a validation error under the project selector. "You do not have permission to edit this project". This shortens the project selector help text we have right now.

2) Does the project already have `replica.mode` set to something other than empty or "leader"? > Show a validation error under the project selector:  "Project is already configured with replica mode :mode. See [project configuration]." Deep link to the new project config section for replication so the user can explore in the project config.

3) Does the project already have a `replica.cluster` set? > Show a validation error under the project selector: "Project is already setup with another replicator. See [project configuration]."  Deep link to the new project config section for replication so the user can explore in the project config.

# Submit replicator create/edit form

We should do a chained edit on submitting the create replica. After replica creation, we also set the two configurations on the selected project: `replica.mode` to `leader` and `replica.cluster` to the selected cluster link.

This might also be needed on edit, but I am not sure if the project of a replica can be changed. If not then this is only needed on creation.

edlerd: @kimanhou for the run modal here, I think we should do similar validations of the local project configuration. On failing validation, we should give direct steps how to fix and link to the project replica setting. So a user 1) quickly understands what the problem is and what is expected of them to fix it and 2) quickly gets to the place where they can fix it.

kimanhou: All of this can't be done in a title. Should we always enable the Run button, then the Modal would contain all validations, errors and links ?

edlerd: > All of this can't be done in a title. Should we always enable the Run button, then the Modal would contain all validations, errors and links ?

Yes, that sounds right.

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -27,10 +28,44 @@ const ReplicatorDetail: FC = () => {
...
+          operation.metadata?.original_entity_url ||
+          operation.metadata?.entity_url ||
+          "";
+
+        return (
+          operation.description === "Running replicator" &&
+          entityUrl.startsWith(replicatorPath) &&
+          entityUrl.includes(projectQuery)
+        );
+      });
```
edlerd: There seems to be no notion of "latest" in the filter logic.

Also I think we should match all of the url fields, not just the first that has a value in the lines 48 to 51.

edlerd: Also I think we want this on the detail page and on the list page. So maybe raising this into a context would be better. Similar to how `instanceLoading.getType(instance)` works, we can have a `const replicatorStatus = userReplicatorStatus()`. And then in here simplify to `replicatorStatus.getLastRunError(replicator)`.

So all the operation loading, saving the last error client side state and matching is pushed into the context and we keep this component focused on displayal. Also the api state will then be simpler and less bespoke. See other comments above.

kimanhou: Okay okay but what do you think of the preflight checks design ??

image

src/context/useReplicatorPreflight.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const cluster = replicator.config?.cluster;
+    list.push({
+      id: "target-cluster-set",
+      label: "Is cluster set on the replicator ?",
+      status: cluster ? "pass" : "fail",
+      message: !cluster
+        ? "Replicator cluster is missing. Please set it up to your cluster link name."
+        : undefined,
+    });
```
edlerd:
```suggestion
```
These two checks are always passing a replicator cannot exist without project or cluster.

src/context/useReplicatorPreflight.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const projectCluster = project?.config?.["replica.cluster"];
+    const replicatorCluster = replicator.config?.cluster;
+    const clusterMatches = projectCluster === replicatorCluster;
+    list.push({
+      id: "replica-cluster-match",
+      label:
+        "Is the replicator cluster the same as the project's replica cluster ?",
```
edlerd: Let's formulate these as statement. Maybe also include the project name here as a chip?
```suggestion
        "Replicator cluster matches the project :name replica cluster",
```

src/context/useReplicatorPreflight.tsx
```diff
@@ -0,0 +1,133 @@
...
+      status: isProjectLoading ? "loading" : clusterMatches ? "pass" : "fail",
+      message:
+        !isProjectLoading && !clusterMatches
+          ? `Project cluster '${projectCluster || "none"}' must match replicator cluster '${replicatorCluster || "none"}'.`
```
edlerd: We should deep link to the project configuration here, so the user can go in and fix it. Also would be good to make the copy more actionable.

src/context/useReplicatorPreflight.tsx
```diff
@@ -0,0 +1,133 @@
...
+      status: isInstancesLoading
+        ? "loading"
+        : allInstancesStopped
+          ? "pass"
+          : "fail",
+      message:
+        !isInstancesLoading && !allInstancesStopped
+          ? `${runningInstances.length} active instance(s) running in this project namespace. Stop them to prevent data drift during sync.`
+          : undefined,
+    });
```
edlerd: This is not a requirement to replicate. Only on restore they should be stopped.

src/components/PreflightCheckRow.tsx
```diff
@@ -0,0 +1,43 @@
+import type { FC } from "react";
+import { Icon, Spinner } from "@canonical/react-components";
+import classnames from "classnames";
+import type { PreflightCheck } from "context/useReplicatorPreflight";
+
+interface PreflightRowProps {
+  check: PreflightCheck;
+}
+
+export const PreflightCheckRow: FC<PreflightRowProps> = ({ check }) => {
```
edlerd: This is not a generic component, let's keep it in the same folder as the preflight one.

src/context/useReplicatorPreflight.tsx
```diff
@@ -0,0 +1,133 @@
...
+  replicator: LxdReplicator;
+  isProjectLoading: boolean;
+  isClusterLinkStateLoading: boolean;
+  isInstancesLoading: boolean;
+  project?: LxdProject;
+  clusterLinkState?: LxdClusterLinkState;
+  instances?: LxdInstance[];
+}
+
+export const useReplicatorPreflight = ({
```
edlerd: Can we move this directly into the `RunReplicatorPreflightChecks`? There will only ever be one use case for it so it might be simpler without a context as we don't need to move data around between components.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useProject } from "context/useProjects";
+import { useReplicatorPreflight } from "context/useReplicatorPreflight";
+import type { LxdReplicator } from "types/replicator";
+
+interface Props {
+  replicator: LxdReplicator;
+  onValidationChange: (isValid: boolean) => void;
+}
+
+const RunReplicatorPreflightChecks: FC<Props> = ({
```
edlerd, line 23: I think the UX on adding those is super nice.

src/pages/cluster/ReplicatorDetail.tsx
edlerd: > Okay okay but what do you think of the preflight checks design ??

The UX on loading and with the icons is super nice!

src/context/useReplicatorPreflight.tsx
kimanhou: I sent an AI agent on a speleo mission to dig out the conditions checked on backend side. Here is a summary:
- Action must be "restore" or "start" (always valid in the UI by design)
- Replicator must have cluster configured (_this check should always pass but it does exist on backend ! Shall I put it back in the frontend ?_)
- LXD must be able to load cluster link/cert and connect to target cluster (equivalent to "cluster link must be reachable" I assume ?)
- Target project must exist (we can't check that, can we ?)
- For normal run (start): source project must be leader, target project must be standby. (always valid in the UI by design)
- For restore: source project must be standby, target project must be leader. (always valid in the UI by design)
- For restore, all local project instances must be stopped (check implemented)
- Same replicator cannot run concurrently (button is disabled when replicator is running)
- Scheduler only considers replicators in leader projects (_I did not know that, is it worth mentioning it somewhere in the UI ?_)

TLDR; my preflight checks should be good. 2 comments who required your attention are in italic

src/context/replicatorRunning.tsx
```diff
@@ -0,0 +1,227 @@
...
+      ([key, value]) =>
+        (key === "entity_url" ||
+          key === "original_entity_url" ||
+          key.endsWith("_url")) &&
+        Boolean(value),
+    )
+    .map(([, value]) => value);
+
+  return [...new Set([...operationUrls, ...metadataUrls])];
+};
```
edlerd: I think this could live in `util/operations.tsx`. In fact we already have defined `getOperationEntityUrls` there.

src/context/useReplicators.tsx
```diff
@@ -2,6 +2,17 @@ import { useQuery } from "@tanstack/react-query";
...
+
+const POLLING_INTERVAL = 3000;
+
+const shouldPollReplicator = (replicator?: LxdReplicator) => {
+  if (!replicator) {
+    return false;
+  }
+
+  return replicator.last_run_status === "Running";
+};
```
edlerd: We don't need to do polling here. Instead we can register a function similar to `updateClusterMemberLoading` in `Events.tsx`, which clears the query keys for replicators once it observes a relevant operation.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -10,22 +29,214 @@ interface Props {
...
-          `Run replicator ${replicator.name} btn clicked. TODO: open modal`,
-          onClose,
-        );
+      className={classnames("u-no-margin has-icon", className)}
+      loading={isLoading || isProjectLoading}
+      confirmationModalProps={{
+        className: "run-replicator-modal",
+        title: (
+          <>
+            Run replicator{" "}
```
edlerd: I think the title should reflect when this is a restore.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const projectCluster = project?.config?.["replica.cluster"];
+  const replicatorCluster = replicator.config?.cluster;
+  const clusterMatches = projectCluster === replicatorCluster;
+  checks.push({
+    id: "replica-cluster-match",
+    label: (
+      <>
+        Replicator cluster matches the project{" "}
+        <ProjectRichChip projectName={project?.name || "default"} /> replica
+        cluster
```
edlerd:
```suggestion
        Projects{" "}
        <ProjectRichChip projectName={project?.name || "default"} /> <code>replica cluster</code> configuration matches replicator cluster
```
I wonder if the reverse is more natural way of thinking here.

Either way I'd add "Projects ... configuration".

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,139 @@
...
+        </>
+      ),
+      status: isInstancesLoading
+        ? "loading"
+        : allInstancesStopped
+          ? "pass"
+          : "fail",
+      message:
+        !isInstancesLoading && !allInstancesStopped
+          ? `${runningInstances.length} active instance(s) running in this project namespace. Stop them to prevent data drift during sync.`
```
edlerd:
```suggestion
          ? `${runningInstances.length} running instance in the project. Stop them to allow restore.`
```
Suggestion on the copy.

Please use pluralize on the "instance" word.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  useEffect(() => {
+    onValidationChange(allChecksPassed && !isAnyCheckLoading);
+  }, [allChecksPassed, isAnyCheckLoading, onValidationChange]);
+
+  return (
+    <>
+      <p>
+        <strong>Preflight checks:</strong>
+      </p>
```
edlerd:
```suggestion
```
We probably can do without this. Wdyt?

src/context/replicatorRunning.tsx
```diff
@@ -0,0 +1,227 @@
...
+      if (!prev.has(key)) {
+        return prev;
+      }
+      const next = new Set(prev);
+      next.delete(key);
+      return next;
+    });
+  };
+
+  const setLastRunError = (replicator: LxdReplicator, error?: string) => {
```
edlerd:
```suggestion
  const setLastRunError = (replicator: LxdReplicator, error: string) => {
```
Better define a second function to `clearLastRunError = (replicator: ...` to have a cleaner interface.

src/context/replicatorRunning.tsx
```diff
@@ -0,0 +1,227 @@
...
+      if (!isRunningReplicatorOperation(operation)) {
+        continue;
+      }
+
+      for (const key of getOperationReplicatorKeys(operation)) {
+        keys.add(key);
+      }
+    }
+    return keys;
+  }, [runningOperations]);
```
edlerd: We only need this inside of `getStatus`. So instead of precomputing, better make it an inline filter on the `runningOperations` inside `getStatus`. Wdyt?

src/context/replicatorRunning.tsx
```diff
@@ -0,0 +1,227 @@
...
+        if (!current || timestamp >= current.timestamp) {
+          latestErrors.set(key, { message: operation.err, timestamp });
+        }
+      }
+    }
+
+    return new Map(
+      [...latestErrors.entries()].map(([key, value]) => [key, value.message]),
+    );
+  }, [operations]);
```
edlerd: Similar to the above, we only need this inside of getLastRunError. So I'd inline this logic there and not precompute.

I think it will be simpler, as we can then filter operations by type "Running replicator" and match them against the replicator at hand.

src/context/replicatorLoading.tsx
```diff
@@ -0,0 +1,227 @@
...
+    .map(parseReplicatorKeyFromUrl)
+    .filter((key): key is string => Boolean(key));
+};
+
+const isRunningReplicatorOperation = (operation: LxdOperation): boolean => {
+  return (
+    operation.description === "Running replicator" &&
+    operation.status === "Running"
+  );
+};
```
edlerd, line 69:
```suggestion
const isRunningReplicatorOperation = (operation: LxdOperation): boolean => {
  return (
    operation.description === "Running replicator"
  );
};
```
Maybe we shouldn't mix the two concerns of operation status and operation type. Wdyt?

src/context/replicatorRunning.tsx
```diff
@@ -0,0 +1,227 @@
...
+        setFinish,
+        setLastRunError,
+      }}
+    >
+      {children}
+    </ReplicatorRunningContext.Provider>
+  );
+};
+
+export const useReplicatorRunning = (): ReplicatorRunningType => {
```
edlerd:
```suggestion
export const useReplicatorLoading = (): ReplicatorLoadingType => {
```
Should we align the name with `instanceLoading` and `memberLoading`?

edlerd (commented):
Good progress on this PR!

Few ideas to simplify the logic and some thoughts on the copy.

src/context/replicatorRunning.tsx
kimanhou: `getOperationEntityUrls` doesn't do the same thing, does it ? I renamed `replicatorRunning/getOperationEntityUrls` and moved it to `util/operations.tsx`

src/context/useReplicators.tsx
kimanhou: Okay but why does `Events.tsx` live inside `pages/instances` ?

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
kimanhou: I don't have an opinion on this but it should match the helptext, which is currently not the case. I'm removing the `code` tag because it looks terrible. I can replace it with a `strong` if you want to highlight this bit. 

image

image

src/context/replicatorRunning.tsx
edlerd: Yes, semantics are slightly different. I was wondering do we really need to read also the top level entity fields or is it enough to rely on the `metadata.*` sub-fields?

src/context/useReplicators.tsx
edlerd: Very good question, I think we should move it two levels up to `src/`.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
edlerd: I think bold is fine, even better actually. Style and copy/capitalizations of the configuration description should match the one in the project configuration screen. So a user quickly finds and understands what we are talking about and can make the connection easily between this message and the configuration screen.

src/context/replicatorRunning.tsx
kimanhou: It looks like metadata only is enough indeed.

src/context/replicatorLoading.tsx
```diff
@@ -0,0 +1,197 @@
...
+    if (replicatorIndex === -1 || !pathParts[replicatorIndex + 1]) {
+      return null;
+    }
+
+    const name = decodeURIComponent(pathParts[replicatorIndex + 1]);
+    const project = decodeURIComponent(
+      parsed.searchParams.get("project") || "default",
+    );
+
+    return `${project}/${name}`;
```
edlerd:
```suggestion
    return getReplicatorKey({ name, project });
```
Better rely on the funciton to build the key instead of creating a 2nd source of truth for it.

src/context/replicatorLoading.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const setFinish = (replicator: LxdReplicator) => {
+    const key = getReplicatorKey(replicator);
+    setManualRunningKeys((prev) => {
+      if (!prev.has(key)) {
+        return prev;
+      }
+      const next = new Set(prev);
+      next.delete(key);
+      return next;
+    });
```
edlerd, line 113:
```suggestion
    const key = getReplicatorKey(replicator);
      if (!manualRunningKeys.has(key)) {
        return;
      }
    setManualRunningKeys((prev) => {
      const next = new Set(prev);
      next.delete(key);
      return next;
    });
```
This can be an early exit to avoid rerenders due to state updates.

src/context/replicatorLoading.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const [manualLastRunErrors, setManualLastRunErrors] = useState(
+    new Map<string, string>(),
+  );
+
+  const setRunning = (replicator: LxdReplicator) => {
+    const key = getReplicatorKey(replicator);
+    setManualRunningKeys((prev) => {
+      if (prev.has(key)) {
+        return prev;
+      }
```
edlerd, line 97:
```suggestion
    const key = getReplicatorKey(replicator);
      if (manualRunningKeys.has(key)) {
        return;
      }
    setManualRunningKeys((prev) => {
```
Can be an early exit as well.

src/context/replicatorLoading.tsx
```diff
@@ -0,0 +1,197 @@
...
+      const next = new Map(prev);
+      next.delete(key);
+      return next;
+    });
+  };
+
+  const getStatus = (replicator: LxdReplicator): LxdReplicatorStatus => {
+    const key = getReplicatorKey(replicator);
+    const hasRunningOperation = runningOperations
+      .filter(isRunningReplicatorOperation)
```
edlerd: We changed the definition, so we now also need to filter the operation.status is running.

src/context/replicatorLoading.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const key = getReplicatorKey(replicator);
+    let latestOperationError: string | undefined;
+    let latestOperationTimestamp = -1;
+
+    for (const operation of operations) {
+      if (!isFailedReplicatorOperation(operation) || !operation.err) {
+        continue;
+      }
+
+      if (!getOperationReplicatorKeys(operation).includes(key)) {
```
edlerd: The `getOperationReplicatorKeys(operation)` should be done outside the for loop so we don't repeat it on every cycle.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -10,22 +29,214 @@ interface Props {
...
+  };
+
+  const notifySuccess = (replicatorName: string) => {
+    toastNotify.success(
+      <>
+        Replicator{" "}
+        <ResourceLink
+          type="replicator"
+          value={replicatorName}
+          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicatorName)}`}
```
edlerd:
```suggestion
  const notifySuccess = () => {
    toastNotify.success(
      <>
        Replicator{" "}
        <ResourceLink
          type="replicator"
          value={replicatorName}
          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(replicator.project)}/replicator/${encodeURIComponent(replicator.name)}`}
```
We can rely on the outer scope here.

edlerd: Should this be an info notification instead of success?

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -10,22 +29,214 @@ interface Props {
...
+      predicate: (query) => query.queryKey[2] === queryKeys.replicators,
+    });
+  };
+
+  const onSuccess = () => {
+    setLoading(false);
+    replicatorLoading.clearLastRunError(replicator);
+    replicatorLoading.setRunning(replicator);
+    invalidateCache();
+    notifySuccess(replicator.name);
```
edlerd:
```suggestion
```
Let's inline `notifySuccess` contents here. Also best to rename this to `onReplicationStarted` to better disambiguate from `onOperationSuccess`.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -10,22 +29,214 @@ interface Props {
...
+            )}
+
+            {isRestore && (
+              <Notification
+                severity="caution"
+                messageElement="div"
+                title="This operation will synchronize instances from the remote cluster to this project."
+              >
+                <p>
+                  All current local data will be overwritten by the remote
```
edlerd:
```suggestion
                  All local data will be overwritten by the remote
```

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,139 @@
...
+      </>
+    ),
+    status: isClusterLinkStateLoading
+      ? "loading"
+      : isReachable
+        ? "pass"
+        : "fail",
+    message:
+      !isClusterLinkStateLoading && !isReachable
+        ? `The cluster link connection state is '${clusterLinkStatus}'. It must be Reachable.`
```
edlerd:
```suggestion
        ? `Cluster link status is '${clusterLinkStatus}'.`
```
I think this can be concise as the copy above clarifies expectations.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,139 @@
...
+        <strong>replica cluster</strong> configuration matches replicator
+        cluster
+      </>
+    ),
+    status: isProjectLoading ? "loading" : clusterMatches ? "pass" : "fail",
+    message:
+      !isProjectLoading && !clusterMatches ? (
+        <>
+          Project replica cluster <strong>{projectCluster || "none"}</strong>{" "}
+          must match replicator cluster{" "}
```
edlerd:
```suggestion
          does not equal replicator cluster{" "}
```

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <strong>{replicatorCluster || "none"}</strong>.{" "}
+          <Link
+            to={`/ui/project/${encodeURIComponent(replicator.project)}/configuration`}
+          >
+            Update the project configuration
```
edlerd:
```suggestion
            See the project configuration
```

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,139 @@
...
+        ? "loading"
+        : allInstancesStopped
+          ? "pass"
+          : "fail",
+      message:
+        !isInstancesLoading && !allInstancesStopped
+          ? `${runningInstances.length} running ${pluralize(
+              "instance",
+              runningInstances.length,
+            )} in the project. Stop them to allow restore.`
```
edlerd:
```suggestion
            )} in the project.`
```
Maybe this is obvious?

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
edlerd: We could adjust this to the copy used on replica creation. When the status is not reachable, we show a validation error there.

src/util/operations.tsx
```diff
@@ -18,6 +18,23 @@ const getOperationEntityUrls = (
   return Array.from(candidates);
 };
 
+export const getReplicatorOperationUrls = (
```
edlerd: Can we use the function above? Not sure if we need the "ends in _url" semantic here.

src/Events.tsx
```diff
@@ -134,6 +134,20 @@ const Events: FC = () => {
     }
   };
 
+  const updateReplicatorLoading = (event: LxdEvent) => {
```
edlerd:
```suggestion
  const updateReplicators = (event: LxdEvent) => {
```
This better describes what the function is doing, no?

edlerd (commented):
Some quick thoughts on the changes after a thorough code review.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
kimanhou: NOT_REACHABLE_ERROR_MESSAGE ?

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
kimanhou: I would rather leave it to be sure. Nothing is too obvious

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
edlerd: Yeah, exactly that one.

src/context/replicatorLoading.tsx
kimanhou: I had to put this back inside the setState because it was breaking the update of the status in the detail page. It's usually safer to run this kinda check inside the setState

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
kimanhou: Mmh that doesn't work

src/context/replicatorLoading.tsx
```diff
@@ -0,0 +1,209 @@
...
+      if (!operationReplicatorKeys.includes(key)) {
+        continue;
+      }
+
+      const timestamp = getOperationTimestamp(operation);
+      if (timestamp >= latestOperationTimestamp) {
+        latestOperationTimestamp = timestamp;
+        latestOperationError = operation.err;
+      }
+    }
```
edlerd:
```suggestion
for (const operation of operations) {
  if (!isFailedReplicatorOperation(operation) || !operation.err) {
    continue;
  }

  if (!getOperationReplicatorKeys(operation).includes(key)) {
    continue;
  }

  const timestamp = getOperationTimestamp(operation);
  if (timestamp >= latestOperationTimestamp) {
    latestOperationTimestamp = timestamp;
    latestOperationError = operation.err;
  }
}
```
Can this be a bit more straight forward?

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </>
+    ),
+    status: isClusterLinkStateLoading
+      ? "loading"
+      : isReachable
+        ? "pass"
+        : "fail",
+    message:
+      !isClusterLinkStateLoading && !isReachable
+        ? `Cluster link status is '${clusterLinkStatus}' but must be 'Reachable' to run replicator. `
```
edlerd: Let's add some actionable copy here. Such as on the replicator creation. Or we might say something like "ensure the link is created in both ways" or "ensure the clusters can reach each other" or alike.

src/pages/cluster/ReplicatorDetail.tsx
```diff
@@ -143,6 +149,12 @@ const ReplicatorDetail: FC = () => {
                       <ReplicatorStatus replicator={replicator} />
                     </td>
                   </tr>
+                  {lastRunError && (
+                    <tr>
+                      <th className="u-text--muted">Last error</th>
+                      <td className="u-text--error">{lastRunError}</td>
+                    </tr>
+                  )}
```
edlerd: The column should always be here to make the location of it learnable.
```suggestion
                    <tr>
                      <th className="u-text--muted">Last error</th>
                      <td className="u-text--error">{lastRunError ?? "-"}</td>
                    </tr>
```

src/util/clusterLink.tsx
```diff
@@ -32,11 +32,22 @@ export const getLinkIdentity = (
...
+  identity?: LxdIdentity,
+  state?: LxdClusterLinkState,
+): { isReachable: boolean; status: StatusCaption } => {
+  const status = getClusterLinksStatus(identity, state);
+  return {
+    isReachable: status === "Reachable",
+    status,
+  };
+};
+
```
edlerd:
```suggestion
```
Better call `getClusterLinksStatus` directly and do the "is reachable" check in the callers.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -13,31 +32,213 @@ interface Props {
...
+      </>,
+    );
+    invalidateCache();
+  };
+
+  const onOperationFailure = (msg: string) => {
+    replicatorLoading.setFinish(replicator);
+    replicatorLoading.setLastRunError(replicator, msg);
+    invalidateCache();
+    notify.failure(
```
edlerd: This should be a toast notification as it is an async result.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -13,31 +32,213 @@ interface Props {
...
+              }}
+              isRestore={isRestore}
+            />
+
+            <p>
+              The project <ProjectRichChip projectName={replicator.project} />{" "}
+              is in{" "}
+              <strong>{project?.config["replica.mode"] || "unknown"}</strong>{" "}
+              mode.
+            </p>
```
edlerd: Let's make this part of the preflight checks. The spacing and placement seem a bit custom at the moment. Wdyt?

Maybe this can be the first check and start with "Project [foo] is in ..." so it matches the 2nd line.

Image

edlerd (commented):
Thanks for the iterations! Nearly perfect, a few comments left.

src/pages/cluster/actions/RunReplicatorBtn.tsx
kimanhou: image

edlerd: Yes, I think this is a much more stable layout, easier to scan and understand.

src/util/clusterLink.tsx
```diff
@@ -32,11 +32,15 @@ export const getLinkIdentity = (
   );
 };
 
-export const isClusterLinkReachable = async (
+const isClusterLinkReachable = async (
   cluster: string,
-): Promise<boolean> => {
+): Promise<{ isReachable: boolean; status: StatusCaption }> => {
```
edlerd:
```suggestion
): Promise<boolean> => {
```
Let's keep this simple.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  const allChecksPassed = checks.every((check) => check.status === "pass");
+  const isAnyCheckLoading = checks.some((check) => check.status === "loading");
+
+  useEffect(() => {
+    onValidationChange(allChecksPassed && !isAnyCheckLoading);
+  }, [allChecksPassed, isAnyCheckLoading, onValidationChange]);
+
+  return (
+    <List
```
edlerd, line 168: We should add no margin bottom helper class here, I think it is too much white space towards the graph below.

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -13,31 +32,206 @@ interface Props {
...
+            {isRestore ? (
+              <p>
+                This will sync all instances from the{" "}
+                <ClusterLinkRichChip clusterLink={clusterLink} /> cluster back
+                to the <ProjectRichChip projectName={replicator.project} />{" "}
+                project.
+              </p>
+            ) : (
+              <p>
+                This will sync all instances in the source project{" "}
```
edlerd:
```suggestion
                This will sync all instances from the source project{" "}
```

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -0,0 +1,176 @@
...
+        ? "loading"
+        : allInstancesStopped
+          ? "pass"
+          : "fail",
+      message:
+        !isInstancesLoading && !allInstancesStopped
+          ? `${runningInstances.length} running ${pluralize(
+              "instance",
+              runningInstances.length,
+            )} in the project. Stop them to allow restore.`
```
edlerd: We need pluralize here to distinguish "Stop it ..." and "Stop them...".

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -13,31 +32,206 @@ interface Props {
...
+              </p>
+            )}
+
+            {isRestore && (
+              <Notification
+                severity="caution"
+                messageElement="div"
+                title="This operation will synchronize instances from the remote cluster to this project."
+              >
+                <p>All local data will be overwritten by the remote version.</p>
```
edlerd: I wonder if we should have even more friction on this action such as a checkbox to confirm. Wdyt?

edlerd (commented):
Nearly perfect, thanks for the iterations! Few comments then can be merged.

src/pages/cluster/actions/RunReplicatorBtn.tsx
kimanhou: Just for the restore, right ? Yeah maybe. Where would you see the checkbox ? In the footer with a label "I understand the consequences" ?

edlerd: Idea was just for restore, yes.

Location and label is a good question. I think we can change the notification or maybe even remove it and have a checkbox with a label such as "Overwrite all local data with instances from the remote cluster". Wdyt?

kimanhou: I want to keep the checkbox in the footer for consistency. However this design doesn't allow for a long text so I suggest keeping the warning notification. Wdyt ? ⬇️

image

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -13,31 +33,220 @@ interface Props {
...
+        onConfirm: handleRun,
+        confirmButtonLabel: buttonLabel,
+        confirmButtonDisabled:
+          !canSubmit || isLoading || (isRestore && !isOverwriteConfirmed),
+        close: onClose,
+        confirmExtra: isRestore ? (
+          <ConfirmationCheckbox
+            label="Overwrite local data"
+            confirmed={[isOverwriteConfirmed, setIsOverwriteConfirmed]}
+            disabled={!canSubmit || isLoading}
```
edlerd, line 241:
```suggestion
```
I think we can enable this right away, no?

edlerd (approved):
LGTM, good stuff :)

src/pages/cluster/actions/RunReplicatorBtn.tsx
kimanhou: No, I think it's better to keep it disabled, that way the user has to wait for all the checks to pass before being able to click on it.

PR #2030, opened 2026-06-02
feat(FileExplorer): [WD-36643] Deletion of Files and Directories

Kxiru:
## Done

- [x] Delete Files
- [x] Delete directories
- [x] Confirmation modal
- [x] FE Actions column
- [x] Toast notifications

## QA

1. Run the LXD-UI:
    - Navigate to an instance VM or container. (If VM, ensure it is running)
    - In the instance terminal, `touch filename` in the root directory (or a directory of your choice)
    - Select File Explorer tab
    - Navigate to the directory within which you made the file
    - Attempt to delete the file using the Delete button in-row.
    - Confirm deletion works as expected.

## Screenshots

image
image
image

### DIFF
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -611,14 +611,10 @@ export const deleteInstanceFile = async (
   params.set("project", instance.project);
   params.set("path", path);
 
-  return fetch(
+  await fetch(
     `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instance.name)}/files?${params.toString()}`,
     {
       method: "DELETE",
     },
-  )
-    .then(handleResponse)
-    .then(() => {
-      return;
-    });
+  ).then(handleResponse);
 };
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -606,7 +606,7 @@ export const fetchInstanceFileHeader = async (
 export const deleteInstanceFile = async (
   instance: LxdInstance,
   path: string,
-): Promise<Response> => {
+): Promise<void> => {
   const params = new URLSearchParams();
   params.set("project", instance.project);
   params.set("path", path);
@@ -616,12 +616,9 @@ export const deleteInstanceFile = async (
     {
       method: "DELETE",
     },
-  ).then((response) => {
-    if (!response.ok) {
-      throw new Error(
-        `Failed to delete file: ${response.status} ${response.statusText}`,
-      );
-    }
-    return response;
-  });
+  )
+    .then(handleResponse)
+    .then(() => {
+      return;
+    });
 };
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -25,6 +25,7 @@ import type { LxdFileExplorerItem, LxdFileMetadata } from "types/fileExplorer";
 
 export const instanceEntitlements = [
   "can_access_console",
+  "can_access_files",
   "can_delete",
   "can_edit",
   "can_exec",
@@ -599,5 +600,28 @@ export const fetchInstanceFileHeader = async (
       type: response.headers.get("x-lxd-type") ?? "unknown",
       modified: response.headers.get("x-lxd-modified") ?? "-",
     };
+  });
+};
+
+export const deleteInstanceFile = async (
+  instance: LxdInstance,
+  path: string,
+): Promise<Response> => {
+  const params = new URLSearchParams();
+  params.set("project", instance.project);
+  params.set("path", path);
+
+  return fetch(
+    `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instance.name)}/files?${params.toString()}`,
+    {
+      method: "DELETE",
+    },
+  ).then((response) => {
+    if (!response.ok) {
+      throw new Error(
+        `Failed to delete file: ${response.status} ${response.statusText}`,
+      );
+    }
+    return response;
   });
 };
--- a/src/pages/instances/FileExplorerTable.tsx
+++ b/src/pages/instances/FileExplorerTable.tsx
@@ -14,15 +14,22 @@ import { isoTimeToString } from "util/helpers";
 import { fetchInstanceFileHeader } from "api/instances";
 import FileExplorerDirectory from "./FileExplorerDirectory";
 import FileExplorerFile from "./FileExplorerFile";
+import FileExplorerActions from "./actions/FileExplorerActions";
 import type { LxdInstance } from "types/instance";
 
 interface Props {
   files: string[];
   currentPath: string;
   instance: LxdInstance;
+  onDeleteSuccess: () => void;
 }
 
-const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
+const FileExplorerTable: FC<Props> = ({
+  files,
+  currentPath,
+  instance,
+  onDeleteSuccess,
+}) => {
   const [fileMetadata, setFileMetadata] = useState<
     Map<string, LxdFileMetadata>
   >(new Map());
@@ -78,6 +85,10 @@ const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
       content: "Modified",
       sortKey: "modified",
     },
+    {
+      "aria-label": "Actions",
+      className: "actions",
+    },
   ];
 
   const rows: MainTableRow[] = files.map((fileName) => {
@@ -88,6 +99,8 @@ const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
     const fileModified = metadata?.modified
       ? isoTimeToString(metadata.modified)
       : "-";
+    const filePath =
+      currentPath === "/" ? `/${fileName}` : `${currentPath}/${fileName}`;
 
     return {
       key: fileName,
@@ -120,7 +133,22 @@ const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
           role: "cell",
           "aria-label": "Modified",
         },
+        {
+          content: (
+            <FileExplorerActions
+              instance={instance}
+              fileName={fileName}
+              filePath={filePath}
+              fileType={fileType}
+              onDeleteSuccess={onDeleteSuccess}
+            />
+          ),
+          role: "cell",
+          className: "u-align--right",
+          "aria-label": "Actions",
+        },
       ],
+      className: "u-row",
       sortData: {
         name: fileName.toLowerCase(),
         type: fileType,
@@ -149,6 +177,7 @@ const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
         rows={sortedRows}
         onUpdateSort={updateSort}
         emptyStateMsg="No files or directories"
+        className="u-selectable-table-rows"
         sortable
       />
     </ScrollableTable>
--- a/src/pages/instances/FileExplorerTable.tsx
+++ b/src/pages/instances/FileExplorerTable.tsx
@@ -1,77 +1,20 @@
-import { useEffect, useState, type FC } from "react";
-import {
-  MainTable,
-  ScrollableTable,
-  Spinner,
-} from "@canonical/react-components";
+import { type FC } from "react";
+import { MainTable, ScrollableTable } from "@canonical/react-components";
 import type {
   MainTableHeader,
   MainTableRow,
 } from "@canonical/react-components/dist/components/MainTable/MainTable";
 import useSortTableData from "util/useSortTableData";
-import type { LxdFileMetadata } from "types/fileExplorer";
-import { isoTimeToString } from "util/helpers";
-import { fetchInstanceFileHeader } from "api/instances";
-import FileExplorerDirectory from "./FileExplorerDirectory";
-import FileExplorerFile from "./FileExplorerFile";
-import FileExplorerActions from "./actions/FileExplorerActions";
 import type { LxdInstance } from "types/instance";
+import getFileExplorerTableRow from "pages/instances/getFileExplorerTableRow";
 
 interface Props {
   files: string[];
   currentPath: string;
   instance: LxdInstance;
-  onDeleteSuccess: () => void;
 }
 
-const FileExplorerTable: FC<Props> = ({
-  files,
-  currentPath,
-  instance,
-  onDeleteSuccess,
-}) => {
-  const [fileMetadata, setFileMetadata] = useState<
-    Map<string, LxdFileMetadata>
-  >(new Map());
-  const [isLoading, setIsLoading] = useState(false);
-
-  useEffect(() => {
-    const fetchMetadata = async () => {
-      setIsLoading(true);
-      const metadataMap = new Map<string, LxdFileMetadata>();
-
-      await Promise.all(
-        files.map(async (fileName) => {
-          try {
-            const filePath =
-              currentPath === "/"
-                ? `/${fileName}`
-                : `${currentPath}/${fileName}`;
-
-            const fileMetadata = await fetchInstanceFileHeader(
-              instance.name,
-              instance.project,
-              filePath,
-            );
-
-            metadataMap.set(fileName, fileMetadata);
-          } catch (e) {
-            console.error(`Failed to fetch metadata for ${fileName}:`, e);
-            metadataMap.set(fileName, {
-              type: "unknown",
-              modified: "-",
-            });
-          }
-        }),
-      );
-
-      setFileMetadata(metadataMap);
-      setIsLoading(false);
-    };
-
-    void fetchMetadata();
-  }, [files, currentPath, instance.name, instance.project]);
-
+const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
   const headers: MainTableHeader[] = [
     {
       content: "Name",
@@ -91,80 +34,14 @@ const FileExplorerTable: FC<Props> = ({
     },
   ];
 
-  const rows: MainTableRow[] = files.map((fileName) => {
-    const metadata = fileMetadata.get(fileName);
-    const isDirectory = metadata?.type === "directory";
-    const isFile = metadata?.type === "file";
-    const fileType = metadata?.type ?? "unknown";
-    const fileModified = metadata?.modified
-      ? isoTimeToString(metadata.modified)
-      : "-";
-    const filePath =
-      currentPath === "/" ? `/${fileName}` : `${currentPath}/${fileName}`;
-
-    return {
-      key: fileName,
-      columns: [
-        {
-          content: isDirectory ? (
-            <FileExplorerDirectory
-              dirName={fileName}
-              parentPath={currentPath}
-              instance={instance}
-            />
-          ) : (
-            <FileExplorerFile
-              fileName={fileName}
-              parentPath={currentPath}
-              instance={instance}
-              icon={isFile ? "file-blank" : "pods"}
-            />
-          ),
-          role: "rowheader",
-          "aria-label": "Name",
-        },
-        {
-          content: fileType,
-          role: "cell",
-          "aria-label": "Type",
-        },
-        {
-          content: fileModified,
-          role: "cell",
-          "aria-label": "Modified",
-        },
-        {
-          content: (
-            <FileExplorerActions
-              instance={instance}
-              fileName={fileName}
-              filePath={filePath}
-              fileType={fileType}
-              onDeleteSuccess={onDeleteSuccess}
-            />
-          ),
-          role: "cell",
-          className: "u-align--right",
-          "aria-label": "Actions",
-        },
-      ],
-      className: "u-row",
-      sortData: {
-        name: fileName.toLowerCase(),
-        type: fileType,
-        modified: fileModified,
-      },
-    };
-  });
+  const rows: MainTableRow[] = files.map((fileName) =>
+    getFileExplorerTableRow(fileName, currentPath, instance),
+  );
 
   const { rows: sortedRows, updateSort } = useSortTableData({
     rows,
     defaultSort: "name",
   });
-
-  if (isLoading) {
-    return <Spinner text="Loading file metadata..." />;
-  }
 
   return (
     <ScrollableTable
--- a/src/pages/instances/InstanceFileExplorer.tsx
+++ b/src/pages/instances/InstanceFileExplorer.tsx
@@ -15,6 +15,7 @@ import { fetchInstanceDirectory } from "api/instances";
 import { useSearchParams } from "react-router-dom";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
+import StartInstanceBtn from "./actions/StartInstanceBtn";
 
 interface Props {
   instance: LxdInstance;
@@ -64,53 +65,48 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
         title="Instance is not running"
       >
         <p>Virtual machines must be running to browse files.</p>
-        <p>
-          <a
-            href={`https://documentation.ubuntu.com/lxd/en/latest/howto/instances_manage/`}
-            rel="noopener noreferrer"
-            target="_blank"
-          >
-            Learn more about managing instances
-            <Icon className="external-link-icon" name="external-link" />
-          </a>
-        </p>
+        <StartInstanceBtn
+          instance={instance}
+          appearance="positive"
+          dense={false}
+          iconClassname="is-light"
+          hasLabel
+          showBooting
+        />
       </EmptyState>
     );
-  }
-
-  if (error) {
-    return (
-      <>
-        <Notification
-          severity="negative"
-          title="Error connecting to file system"
-        >
-          {error.message}
-        </Notification>
-        <div className="p-panel">
-          <div className="p-panel__content">
-            <Button appearance="positive" onClick={invalidateCache}>
-              Retry connection
-            </Button>
-          </div>
-        </div>
-      </>
-    );
-  }
-
-  if (isLoading) {
-    return <Spinner className="u-loader" text="Loading file system..." />;
   }
 
   return (
     <Row className="general">
       <FileExplorerBreadcrumb currentPath={currentPath} instance={instance} />
-      <FileExplorerTable
-        key={directoryContent?.metadata.length}
-        files={directoryContent?.metadata ?? []}
-        currentPath={currentPath}
-        instance={instance}
-      />
+      {error ? (
+        <>
+          <Notification
+            severity="negative"
+            title="Error connecting to file system"
+            className="u-no-margin--bottom"
+          >
+            {error.message}
+          </Notification>
+          <div className="p-panel">
+            <div className="p-panel__content">
+              <Button appearance="positive" onClick={invalidateCache}>
+                Retry connection
+              </Button>
+            </div>
+          </div>
+        </>
+      ) : isLoading ? (
+        <Spinner className="u-loader" text="Loading file system..." />
+      ) : (
+        <FileExplorerTable
+          key={directoryContent?.metadata.length}
+          files={directoryContent?.metadata ?? []}
+          currentPath={currentPath}
+          instance={instance}
+        />
+      )}
     </Row>
   );
 };
--- a/src/pages/instances/InstanceFileExplorer.tsx
+++ b/src/pages/instances/InstanceFileExplorer.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState, type FC } from "react";
+import { type FC } from "react";
 import {
   Button,
   EmptyState,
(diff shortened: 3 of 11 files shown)


src/api/instances.tsx
```diff
@@ -601,3 +602,26 @@ export const fetchInstanceFileHeader = async (
...
+    `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instance.name)}/files?${params.toString()}`,
+    {
+      method: "DELETE",
+    },
+  ).then((response) => {
+    if (!response.ok) {
+      throw new Error(
+        `Failed to delete file: ${response.status} ${response.statusText}`,
+      );
+    }
```
edlerd: This is masking error messages. See example below, we should surface the detailed message from the api response in the UI:

Image

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+  fileType: string;
+  onDeleteSuccess: () => void;
+}
+
+const FileExplorerDeleteBtn: FC<Props> = ({
+  instance,
+  fileName,
+  filePath,
+  fileType,
+  onDeleteSuccess,
```
edlerd:
```suggestion
  onSuccess,
```
Can be a bit more concise.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { canAccessInstanceFiles } = useInstanceEntitlements();
+
+  const isDirectory = fileType === "directory";
+
+  const handleDelete = () => {
+    setLoading(true);
+    deleteInstanceFile(instance, filePath)
+      .then(() => {
+        toastNotify.success(
+          `${isDirectory ? "Directory" : "File"} ${fileName} deleted.`,
```
edlerd: We should mention the instance here as well with a rich chip.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+        );
+        onDeleteSuccess();
+      })
+      .catch((e) => {
+        toastNotify.failure(
+          `${isDirectory ? "Directory" : "File"} deletion failed`,
```
edlerd: Let's mention the file name here as well.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,95 @@
...
+      });
+  };
+
+  const hasPermission = canAccessInstanceFiles(instance);
+
+  const getHoverText = () => {
+    if (!hasPermission) {
+      return "You do not have permission to manage files on this instance";
+    }
+    return `Delete ${isDirectory ? "directory" : "file"}`;
```
edlerd:
```suggestion
    return `Delete ${directoryOrFile}`;
```
Let's introduce a `const directoryOrFile = isDirectory ? "directory" : "file"` to reuse and not repeat this three times in this file.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -120,7 +133,22 @@ const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
           role: "cell",
           "aria-label": "Modified",
         },
+        {
+          content: (
+            <FileExplorerActions
+              instance={instance}
+              fileName={fileName}
+              filePath={filePath}
+              fileType={fileType}
```
edlerd:
```suggestion
              filePath={filePath}
              fileType={fileType}
```
We can get the name from the path so I would keep the interface a bit smaller here.

src/pages/instances/FileExplorerTable.tsx
```diff
@@ -120,7 +133,22 @@ const FileExplorerTable: FC<Props> = ({ files, currentPath, instance }) => {
...
+              onDeleteSuccess={onDeleteSuccess}
```
edlerd:
```suggestion
```
Having to prop-drill this all the way up shows the downside of custom data loading in `InstanceFileExplorer`. I thought about it on reviewing the file explorer, but thought it is okay. Overseeing that we will have to invalidate queries later on as we add delete or edit.

Let's change the file explorer data loading to rely on `useQuery` instead. Then we can invalidate the cache in the delete button, and the code will become much less coupled.

edlerd: Something like this:

```
import { type FC } from "react";
import {
  Button,
  EmptyState,
  Icon,
  Notification,
  Row,
  Spinner,
import type { LxdInstance } from "types/instance";
import { isInstanceRunning } from "util/instanceStatus";
import FileExplorerTable from "./FileExplorerTable";
import FileExplorerBreadcrumb from "./FileExplorerBreadcrumb";
import { fetchInstanceDirectory } from "api/instances";
import { useSearchParams } from "react-router-dom";
import { useQuery } from "@tanstack/react-query";

interface Props {
  instance: LxdInstance;
}

const InstanceFileExplorer: FC<Props> = ({ instance }) => {
  const [searchParams] = useSearchParams();
  const currentPath = searchParams.get("path") || "/";
  const canServeFiles =
    isInstanceRunning(instance) || instance.type === "container";

  const {
    data: currentDirectory,
    isLoading,
    error,
    refetch,
  } = useQuery({
    queryKey: [
      queryKeys.instances,
      instance.name,
      instance.project,
      queryKeys.files,
      currentPath,
    ],
    queryFn: async () =>
      fetchInstanceDirectory(instance.name, instance.project, currentPath),
    enabled: canServeFiles,
  });

  if (!canServeFiles) {
    return (
      <EmptyState
        className="empty-state"
        image={<Icon name="containers" className="empty-state-icon" />}
        title="Instance is not running"
      >
        <p>Virtual machines must be running to browse files.</p>
        <p>
          <a
            href={`https://documentation.ubuntu.com/lxd/en/latest/howto/instances_manage/`}
            rel="noopener noreferrer"
            target="_blank"
          >
            Learn more about managing instances
            <Icon className="external-link-icon" name="external-link" />
          </a>
        </p>
      </EmptyState>
    );
  }

  if (error) {
    return (
      <>
        <Notification
          severity="negative"
          title="Error connecting to file system"
          onDismiss={() => {
            refetch();
          }}
        >
          {error.message}
        </Notification>
        <div className="p-panel">
          <div className="p-panel__content">
            <Button appearance="positive" onClick={refetch}>
              Retry connection
            </Button>
          </div>
        </div>
      </>
    );
  }

  if (isLoading) {
    return <Spinner className="u-loader" text="Loading file system..." />;
  }

  return (
    <Row className="general">
      <FileExplorerBreadcrumb currentPath={currentPath} instance={instance} />
      <FileExplorerTable
        files={currentDirectory?.metadata || []}
        currentPath={currentPath}
        instance={instance}
      />
    </Row>
  );
};

export default InstanceFileExplorer;

```

edlerd: We should also refactor the `FileExplorerTable` to use query on each file row instead of the custom built metadata fetching.

edlerd (commented):
Some thoughts to make the whole file explorer data fetching a bit more similar to the LXD UI standards.

src/api/instances.tsx
```diff
@@ -601,3 +603,30 @@ export const fetchInstanceFileHeader = async (
...
+  );
+
+  if (!response.ok) {
+    const errorData = (await response.json()) as ErrorResponse;
+
+    throw new Error(
+      errorData.error ??
+        `Failed to delete file: ${response.status} ${response.statusText}`,
+    );
+  }
```
edlerd: We shouldn't have bespoke handling here, but rely on the `handleResponse` helper instead.

src/api/instances.tsx
```diff
@@ -601,3 +603,30 @@ export const fetchInstanceFileHeader = async (
     };
   });
 };
+
+export const deleteInstanceFile = async (
+  instance: LxdInstance,
+  path: string,
+): Promise<Response> => {
```
```suggestion
): Promise<void> => {
```
This can have no return, the api response also has no content.

src/pages/instances/getFileExplorerTableRow.tsx
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  const isDirectory = metadata?.type === "directory";
+  const isFile = metadata?.type === "file";
+  const fileType = metadata?.type ?? "-";
+  const fileModified = metadata?.modified
+    ? isoTimeToString(metadata.modified)
+    : "-";
+  const filePath =
+    currentPath === "/" ? `/${fileName}` : `${currentPath}/${fileName}`;
```
edlerd:
```suggestion
```
Let's use `fullPath` which is the same as above.

.github/workflows/coverage.yaml
edlerd: Changes here and to several other files should vanish with rebase on main. We should remove them from the PR.

src/api/instances.tsx
Kxiru: I have adjusted this now, but I am not sure how you would like me to surface the api error. The above structure was taken from another part of the codebase.

src/api/instances.tsx
```diff
@@ -601,3 +602,23 @@ export const fetchInstanceFileHeader = async (
     };
   });
 };
+
+export const deleteInstanceFile = async (
+  instance: LxdInstance,
+  path: string,
+): Promise<LxdOperationResponse> => {
```
edlerd: This doesn't return an operation. We can simplify to use void here
```suggestion
): Promise<void> => {
```

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,118 @@
...
+        toastNotify.success(
+          <>
+            {directoryOrFile} {fileName} deleted from {instanceLink}
+          </>,
+        );
+        invalidateCache();
+      })
+      .catch((e) => {
+        toastNotify.failure(
+          `Deletion failed for ${directoryOrFile} ${fileName}`,
```
edlerd: directory should be lowercase

src/pages/instances/actions/FileExplorerActions.tsx
```diff
@@ -0,0 +1,37 @@
...
+  instance: LxdInstance;
+  fileName: string;
+  filePath: string;
+  fileType: string;
+}
+
+const FileExplorerActions: FC<Props> = ({
+  instance,
+  fileName,
+  filePath,
```
edlerd: Naming here is a bit ambiguous, is filePath the full path or the parent path?

This is also duplicating the data, how about we only pass in the fullPath and figure the file name by splitting in nested components?
```suggestion
  fullPath,
```

src/pages/instances/getFileExplorerTableRow.tsx
```diff
@@ -0,0 +1,92 @@
...
+import FileExplorerDirectory from "./FileExplorerDirectory";
+import FileExplorerFile from "./FileExplorerFile";
+import FileExplorerActions from "./actions/FileExplorerActions";
+import type { LxdInstance } from "types/instance";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+
+const getFileExplorerTableRow = (
+  fileName: string,
+  currentPath: string,
```
edlerd:
```suggestion
  parentPath: string,
```
I think in this context parent path makes more sense. There is no notion of current in here. Wdyt?

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { deleteInstanceFile } from "api/instances";
+import { useInstanceEntitlements } from "util/entitlements/instances";
+import { InstanceRichChip } from "../InstanceRichChip";
+import { queryKeys } from "util/queryKeys";
+import { useQueryClient } from "@tanstack/react-query";
+
+interface Props {
+  instance: LxdInstance;
+  fileName: string;
+  filePath: string;
```
edlerd:
```suggestion
  fullPath: string;
```
Maybe one is enough and we can split it in here?

edlerd (commented):
Thanks for the iteration. Nearly perfect, few comments left. Can you merge the commits into one or adjust the commit messages?

src/pages/instances/InstanceFileExplorer.tsx
edlerd, line 1: The loading state is a bit too invasive here. We should always render the breadcrumb but can use the spinner instead of the table when loading.

src/api/instances.tsx
```diff
@@ -601,3 +602,23 @@ export const fetchInstanceFileHeader = async (
...
+  return fetch(
+    `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instance.name)}/files?${params.toString()}`,
+    {
+      method: "DELETE",
+    },
+  )
+    .then(handleResponse)
+    .then(() => {
+      return;
+    });
```
```suggestion
    .then(handleResponse);
```
Slightly simpler.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -106,18 +98,19 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
...
   return (
     <Row className="general">
       <FileExplorerBreadcrumb currentPath={currentPath} instance={instance} />
-      <FileExplorerTable
-        files={(tree.get(currentPath) as string[]) ?? []}
-        currentPath={currentPath}
-        instance={instance}
-      />
+      {isLoading ? (
+        <Spinner className="u-loader" text="Loading file system..." />
```
edlerd, line 101: This loading behaviour is much nicer indeed!

Can we also make the error appear here instead of returning early without the breadcrumb. I was manually navigating to [this path](https://lxd-ui-2030.demos.haus/ui/project/default/instance/MB-test/file-explorer?project=default&path=%2Fdev%2F.lxd-mounts) on the demo server and had no breadcrumb to go back, but think it would have been nice.

edlerd (commented):
Nice, nearly perfect two tiny things.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <p>
+            This will permanently delete <b>{fileName}</b>.
+            <br />
+            This action cannot be undone, and can result in data loss.
+          </p>
+        ),
+        confirmButtonLabel: "Delete",
+        onConfirm: handleDelete,
+      }}
+      disabled={isLoading || !hasPermission}
```
omarelkashef, line 99: Should the disabled be applied to the button itself instead?

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,116 @@
...
+  fullPath: string;
+  fileType: string;
+}
+
+const FileExplorerDeleteBtn: FC<Props> = ({ instance, fullPath, fileType }) => {
+  const [isLoading, setLoading] = useState(false);
+  const queryClient = useQueryClient();
+  const toastNotify = useToastNotification();
+  const { canAccessInstanceFiles } = useInstanceEntitlements();
+  const fileName =
```
omarelkashef: This part is a bit confusing because the variable is called fileName but it can refer to a directory and the conditional logic is also not very simple.

omarelkashef (commented):
Really nice interaction with the filesystem! Two small comments below.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
Kxiru: Not sure what you mean here. I am using a ConfirmationButton component. Elsewhere in the codebase (Eg; DeleteImageBtn, DeleteNetworkBtn) also have the disabled params here. Am I misunderstanding something?

omarelkashef: Sorry,  my eyes were not working. I thought the `disabled` field was inside the `confirmationModalProps`.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
Kxiru: Yes, the naming is a bit crude. We need both to inform what the success notif displays. Perhaps I could rename to fileOrDirectoryName or fileDirectoryName, but I'm not sure what would be preferred.

Kxiru: This comment inspired me to review this file and so I have made some tiny refactoring amendments here.

src/pages/instances/InstanceFileExplorer.tsx
edlerd: The link should be to the local docs.

But I think we shouldn't have a link here at all. Better use the "Start instance" button here, such as on the terminal tab:

Image

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -83,41 +78,36 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
...
-
   return (
     <Row className="general">
       <FileExplorerBreadcrumb currentPath={currentPath} instance={instance} />
-      <FileExplorerTable
-        files={(tree.get(currentPath) as string[]) ?? []}
-        currentPath={currentPath}
-        instance={instance}
-      />
+      {error && (
```
edlerd: Let's hide the file explorer table in the error case. Wdyt?

Image

edlerd (commented):
Code LGTM, two small nitpicks to improve on the UX.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [isLoading, setLoading] = useState(false);
+  const queryClient = useQueryClient();
+  const toastNotify = useToastNotification();
+  const { canAccessInstanceFiles } = useInstanceEntitlements();
+
+  const isDirectory = fileType === "directory";
+  const pathType = isDirectory ? "Directory" : "File";
+  const pathName = isDirectory
+    ? fullPath.split("/").slice(-2, -1)[0]
+    : fullPath.split("/").slice(-1)[0];
```
edlerd: I think this is wrong; we don't need to distinguish between directory and file here. Directory copy is broken:

Image

edlerd (commented):
Discovered one bug below.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
omarelkashef: It is nice to have a guardrail that reminds the user that is a directory and everything inside will be deleted.

edlerd: > It is nice to have a guardrail that reminds the user that is a directory and everything inside will be deleted.

Yes, agreed. But this variable is used to display the directory name in the modal and that is showing empty or the parent, because the selection logic here is wrong.

omarelkashef: Oh yes

Kxiru: @edlerd It's because I changed it and reverse the labels in a recent push, thank you for the catch. Will address.

src/pages/instances/InstanceFileExplorer.tsx
Kxiru: Sounds good

Kxiru: Now I have a nested ternary... I'm not sure if there is another structure in the CB that I could take inspiration from but at least it works.

src/pages/instances/InstanceFileExplorer.tsx
Kxiru: Sure thing

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
Kxiru: Actually, I'm not sure what happened here. It used to work (see screenshot in PR message). I'll troubleshoot.

edlerd: I think this should be `const pathName = fullPath.split("/").slice(-1)[0];` as the name is not dependant on directory or file.

src/pages/instances/InstanceFileExplorer.tsx
edlerd: It is not the nicest thing to read indeed. One way would be to push things into a child component that then has early exits for error, loading and only if both are false renders the table.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
```diff
@@ -0,0 +1,108 @@
...
+      appearance="base"
+      className="has-icon is-dense"
+      loading={isLoading}
+      onHoverText={getHoverText()}
+      confirmationModalProps={{
+        title: "Confirm delete",
+        children: (
+          <p>
+            This will permanently delete {pathType.toLowerCase()}{" "}
+            <b>{pathName}</b> {isDirectory ? "and all of its contents." : "."}
```
edlerd: This seems wrong, when I tried deleting a directory with contents, it caused an error:

Image

Please test accordingly and adjust copy or API usage.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -69,55 +65,40 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
...
+            {error.message}
+          </Notification>
+          <div className="p-panel">
+            <div className="p-panel__content">
+              <Button appearance="positive" onClick={invalidateCache}>
+                Retry connection
+              </Button>
+            </div>
+          </div>
+        </>
```
edlerd, line 99: Spacing below the notification seems a bit too much. Can we add no margin bottom class?

Image

src/pages/instances/InstanceStartButton.tsx
```diff
@@ -0,0 +1,32 @@
+import { ActionButton } from "@canonical/react-components";
+import type { LxdInstance } from "types/instance";
+import { useInstanceEntitlements } from "util/entitlements/instances";
+import { useInstanceStart } from "util/instanceStart";
+import { isInstanceRunning } from "util/instanceStatus";
+
+const InstanceStartButton = ({ instance }: { instance: LxdInstance }) => {
```
edlerd: Please use the existing component `StartInstanceBtn`.

src/pages/instances/InstanceTerminal.tsx
```diff
@@ -327,19 +321,7 @@ const InstanceTerminal: FC<Props> = ({ instance, refreshInstance }) => {
...
-            disabled={isDisabled}
-            title={
-              canUpdateInstanceState(instance)
-                ? ""
-                : "You do not have permission to start this instance."
-            }
-          >
-            Start instance
-          </ActionButton>
+          <InstanceStartButton instance={instance} />
```
edlerd: Thanks for adjusting this use case to reuse the button here.

src/pages/instances/InstanceStartButton.tsx
Kxiru: This button is just an icon button, not a button button. So it is not applicable here.

src/pages/instances/actions/FileExplorerDeleteBtn.tsx
Kxiru: I tested before merge and I don't have this error locally?

edlerd: Reproducer:
1. In terminal run `mkdir bar` and then `touch bar/baz`
2. Open file explorer, go to "root" directory. Delete `bar` folder, error appears.

src/pages/instances/InstanceStartButton.tsx
edlerd: Can we use a similar approach as in the `DeleteStoragePoolBtn`, which reuses the same button logic but allows to change the presentation of the button via prop?

I don't think we should have both `InstanceStartButton` and `StartInstanceBtn` at the same time.

Kxiru: Okay, though they are used in quite different circumstances and are visually entirely different. It would effectively be a component with almost a single prop being passed to differentiate between whether it is used as an action bar subcomponent or a standalone button. I will use DeleteStoragePoolBtn as inspiration and look into it further.

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -4,34 +4,53 @@ import { Button, Icon } from "@canonical/react-components";
...
+  dense?: boolean;
+  iconClassname?: string;
 }
 
-const StartInstanceBtn: FC<Props> = ({ instance }) => {
+const StartInstanceBtn: FC<Props> = ({
+  instance,
+  hasLabel = false,
+  appearance = "base",
+  loading = false,
```
edlerd:
```suggestion
```
This seems unused. Can it be removed?

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -4,34 +4,53 @@ import { Button, Icon } from "@canonical/react-components";
...
+      appearance={appearance}
       hasIcon
-      dense={true}
-      disabled={isDisabled || !canUpdateInstanceState(instance)}
+      dense={dense}
+      disabled={isDisabled || !canUpdateInstanceState(instance) || isBooting}
       onClick={handleStart}
       type="button"
       aria-label={isLoading ? "Starting" : "Start"}
+      loading={loading}
```
edlerd:
```suggestion
```
We probably don't need this, as the icon already changes to a spinner below.

edlerd (commented):
Good stuff, two small comments left on the new combined start button.

edlerd:
@Kxiru can you also fix the sort by date issue as a drive by? Or log a ticket to fix it next.

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -29,9 +42,12 @@ const StartInstanceBtn: FC<Props> = ({ instance }) => {
       }
     >
       <Icon
-        className={classnames({ "u-animation--spin": isLoading })}
-        name={isLoading ? "spinner" : "play"}
+        className={classnames(iconClassname, {
+          "u-animation--spin": isLoading,
```
edlerd:
```suggestion
          "u-animation--spin": showSpinner,
```
We need a new concept for when to show the spinner.

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -4,21 +4,34 @@ import { Button, Icon } from "@canonical/react-components";
...
+  iconClassname?: string;
 }
 
-const StartInstanceBtn: FC<Props> = ({ instance }) => {
+const StartInstanceBtn: FC<Props> = ({
+  instance,
+  hasLabel = false,
+  appearance = "base",
+  dense = true,
+  iconClassname,
```
edlerd, line 23:
```suggestion
  iconClassname,
  showBooting = false,
```
We need to distinguish between the header, where booting should not cause a loading animation. In the terminal/file explorer tab during boot, we should display the loading indicator.

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -29,9 +42,12 @@ const StartInstanceBtn: FC<Props> = ({ instance }) => {
       }
     >
       <Icon
-        className={classnames({ "u-animation--spin": isLoading })}
-        name={isLoading ? "spinner" : "play"}
+        className={classnames(iconClassname, {
+          "u-animation--spin": isLoading,
+        })}
+        name={isLoading || isBooting ? "spinner" : "play"}
```
edlerd:
```suggestion
        name={showSpinner ? "spinner" : "play"}
```

src/pages/instances/actions/StartInstanceBtn.tsx
```diff
@@ -4,21 +4,34 @@ import { Button, Icon } from "@canonical/react-components";
...
+const StartInstanceBtn: FC<Props> = ({
+  instance,
+  hasLabel = false,
+  appearance = "base",
+  dense = true,
+  iconClassname,
+}) => {
   const { handleStart, isLoading, isDisabled } = useInstanceStart(instance);
+  const isRunning = isInstanceRunning(instance);
+  const isBooting = isRunning && (instance.state?.processes ?? 0) < 1;
```
edlerd, line 28:
```suggestion
  const isBooting = isRunning && (instance.state?.processes ?? 0) < 1;
  const showSpinner = showBooting ? isLoading || isBooting : isLoading;
```
This is to distinguish the loading state for header and in the tabs.

edlerd (commented):
The loading state needs a small refinement.

src/pages/instances/actions/StartInstanceBtn.tsx
Kxiru: Why shouldn't booting show a loading animation? It is still loading.

edlerd: Discussed in meet.

src/pages/instances/InstanceTerminal.tsx
```diff
@@ -327,19 +321,13 @@ const InstanceTerminal: FC<Props> = ({ instance, refreshInstance }) => {
...
-              canUpdateInstanceState(instance)
-                ? ""
-                : "You do not have permission to start this instance."
-            }
-          >
-            Start instance
-          </ActionButton>
+            dense={false}
+            iconClassname="is-light"
+            hasLabel
```
edlerd, line 329:
```suggestion
            hasLabel
            showSpinner
```
This tab also depends on the booting as we can't open the terminal before the agent is started within a running VM.

edlerd (dismissed):
LGTM, tiny fix still needed, then good to merge.

src/pages/instances/InstanceTerminal.tsx
Kxiru: Sure thing. I think you meant showBooting here. Pushing!

edlerd (approved):
LGTM, thanks for the iterations!

PR #2042, opened 2026-06-09
fix: standardize redirection links for cluster links WD-36883

omarelkashef:
## Done

- Standardize redirection links for cluster links depending on clustered vs unclustered environments.

## QA

1. Run the LXD-UI:
    - Ensure the links in the following places direct to "/ui/cluster/links" if clustered or  "/ui/server/cluster-links" otherwise
       - Clicking on a cluster link rich chip
       - Clicking on the name in the cluster link tooltip
       - Clicking on the help message under the cluster field in a replicator/image registry form when no cluster links exist (empty state)
       -  Clicking on the cluster link in the image registry tooltip

    - Ensure the links in the following places direct to "/ui/cluster/replicators" if clustered or  "/ui/server/replicators" otherwise
       - Clicking on the link in the replicator detail page header
       - After deleting a replicator

### DIFF
--- a/src/pages/cluster/ClusterLinkRichTooltip.tsx
+++ b/src/pages/cluster/ClusterLinkRichTooltip.tsx
@@ -7,10 +7,10 @@ import ItemName from "components/ItemName";
 import { useClusterLink } from "context/useClusterLinks";
 import ClusterLinkStatus from "./ClusterLinkStatus";
 import ResourceLabel from "components/ResourceLabel";
-import { ROOT_PATH } from "util/rootPath";
 import ClusterLinkAddresses from "./ClusterLinkAddresses";
-import { getLinkIdentity } from "util/clusterLink";
 import { useIdentities } from "context/useIdentities";
+import { getClusterLinkListUrl, getLinkIdentity } from "util/clusterLink";
+import { useIsClustered } from "context/useIsClustered";
 
 interface Props {
   clusterLink: string;
@@ -19,6 +19,7 @@ interface Props {
 const ClusterLinkRichTooltip: FC<Props> = ({ clusterLink }) => {
   const { data: link, isLoading: isLinkLoading } = useClusterLink(clusterLink);
   const { data: identities = [] } = useIdentities();
+  const isClustered = useIsClustered();
 
   if (!link && !isLinkLoading) {
     return (
@@ -37,7 +38,7 @@ const ClusterLinkRichTooltip: FC<Props> = ({ clusterLink }) => {
       title: "Cluster link",
       value: link ? (
         <Link
-          to={`${ROOT_PATH}/ui/cluster/links`}
+          to={getClusterLinkListUrl(isClustered)}
           onClick={(e) => {
             e.stopPropagation();
           }}
--- a/src/pages/cluster/ReplicatorClusterLinkSelector.tsx
+++ b/src/pages/cluster/ReplicatorClusterLinkSelector.tsx
@@ -3,9 +3,10 @@ import { CustomSelect, useNotify } from "@canonical/react-components";
 import { useClusterLinks } from "context/useClusterLinks";
 import type { FormikProps } from "formik";
 import { Link } from "react-router-dom";
-import { ROOT_PATH } from "util/rootPath";
 import type { ReplicatorFormValues } from "types/forms/replicator";
 import ClusterLinkStatus from "pages/cluster/ClusterLinkStatus";
+import { useIsClustered } from "context/useIsClustered";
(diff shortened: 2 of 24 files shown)


src/util/replicator.ts
```diff
@@ -15,3 +16,9 @@ export const getPayload = (
     },
   };
 };
+
+export const getReplicatorsPageLink = (isClustered = false): string => {
```
edlerd:
```suggestion
export const getReplicatorListUrl = (isClustered = false): string => {
```
We use this pattern of naming in other places.

src/util/clusterLink.tsx
```diff
@@ -76,3 +77,9 @@ export const testProjectReplicaCluster = (
 
   return replicaCluster === selectedCluster;
 };
+
+export const getClusterLinksPageLink = (isClustered = false): string => {
```
edlerd:
```suggestion
export const getClusterLinkListUrl = (isClustered = false): string => {
```

src/util/replicator.ts
edlerd: Actually we sometimes also use `linkFor` as a prefix, but only in two places. Can we adjust those two to also be `getVolumeDetailUrl` and `getInstanceDetailUrl`?

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
```diff
@@ -35,6 +37,8 @@ const DeleteReplicatorBtn: FC<Props> = ({
   const toastNotify = useToastNotification();
   const [isLoading, setLoading] = useState(false);
   const { canDeleteReplicator } = useReplicatorEntitlements();
+  const isClustered = useIsClustered();
+  const replicatorsPageLink = getReplicatorsPageLink(isClustered);
```
```suggestion
```
I think when there is a single usage in a file, we don't need the extra `const`. Wdyt?

edlerd (commented):
Thanks for addressing the issue here at hand. Few thoughts on it below.

src/pages/cluster/actions/DeleteReplicatorBtn.tsx
omarelkashef: Do you mean calling `getReplicatorListUrl(isClustered)` directly where it is needed? I do not think we have "nested" function calls like this `navigate(getReplicatorListUrl(isClustered))` anywhere in the codebase.

edlerd: Yeah true, that is harder to read and not clean style either. Let's keep as is where we call navigate with the result.

src/pages/cluster/ClusterLinkRichTooltip.tsx
```diff
@@ -37,7 +39,7 @@ const ClusterLinkRichTooltip: FC<Props> = ({ clusterLink }) => {
       title: "Cluster link",
       value: link ? (
         <Link
-          to={`${ROOT_PATH}/ui/cluster/links`}
+          to={clusterLinksPageLink}
```
edlerd:
```suggestion
          to={getClusterLinkListUrl(isClustered)}
```
I think we should inline it here and in the other places we use this.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -39,16 +40,14 @@ export const ReplicatorClusterLinkSelector: FC<Props> = ({
     ),
   }));
 
+  const helpLink = (
```
edlerd, line 43:
```suggestion
  const clusterLink = (
```

src/pages/images/ImageRegistryClusterLinkSelector.tsx
```diff
@@ -40,16 +42,14 @@ export const ImageRegistryClusterLinkSelector: FC<Props> = ({
     ),
   }));
 
+  const helpLink = (
```
edlerd, line 45:
```suggestion
  const clusterLink = (
```

edlerd (dismissed):
Thanks for the iterations, tiny non-blocking nitpick on naming, LGTM.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
omarelkashef: I am hesitant to call this variable `clusterLink` in a component about cluster links.

edlerd: `clusterLinkLink`? Probably best to keep as is then.

edlerd (approved):
LGTM, thanks.

PR #2043, opened 2026-06-09
feat(replicator): project configuration of replica mode & cluster

kimanhou:
## Done

- feat(replicator): add replica tab in project configuration

## QA

1. Run the LXD-UI:
    - Go to a project's configuration, there should be a new tab called "Replication"
    - Setup a mode and cluster.

## Screenshots

Project creation
image

Project edit
image

Project with no replicators
image


src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+          width: "100%",
+          height: "1px",
+          margin: "1rem 0",
+        }}
+      ></div>
+
+      <Row>
+        <Col size={12}>
+          Replicators for this project
+          <ReplicatorList
```
edlerd: I think we should hide the project column in it for this use case. Wdyt?

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+    [getProjectKey("replica.cluster")]:
+      replicaMode === "" ? "" : replicaCluster,
+  };
+};
+
+interface Props {
+  formik: FormikProps<ProjectFormValues>;
+}
+
+const ProjectReplicaForm: FC<Props> = ({ formik }) => {
```
edlerd: We should link to this section in the project configuration from the replicator pages. When a replicator page mentions a project, such as in the "configure project" pre-flight check links or from the project chip in the run replicator modal.

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -32,14 +32,15 @@ import { ROOT_PATH } from "util/rootPath";
 import usePanelParams, { panels } from "util/usePanelParams";
 
 interface Props {
-  variant?: "main" | "panel";
+  variant?: "main" | "panel" | "project-configuration";
+  project?: string;
 }
 
-const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
+const ReplicatorList: FC<Props> = ({ variant = "main", project }) => {
```
edlerd: The "Create replicator" button has too much right spacing on the project configuration. Also not sure if it should be a positive green button in the new project context.

edlerd (commented):
Good start, few comments on the interaction and representation

src/pages/projects/forms/ProjectFormMenu.tsx
```diff
@@ -6,6 +6,7 @@ import { useIsClustered } from "context/useIsClustered";
 
 export const PROJECT_DETAILS = "Project details";
 export const RESOURCE_LIMITS = "Resource limits";
+export const REPLICA = "Replica";
```
edlerd: Should this be "Replicator" to match the global page?

kimanhou: I don't know. I chose Replica to match the naming from backend but I am not sure it's easy to understand from the user perspective

kimanhou: "Replicator" is awkward because there is a section called "Replicators for this project" inside it. Also "replicator" is the name of another entity.
If we rename the tab "Replicator", then we have to rename "replica cluster" and "replica mode" into "replicator cluster" and "replicator mode" but that doesn't work.

We could rename the tab "Replication" and keep "replica cluster" and "replica mode". Wdyt ?

edlerd: Yes, I think "Replication" is the most common way that will be correctly understood.

edlerd: Maybe we should name the global nav entry also to "Replication". Wdyt?

kimanhou: You're talking about the "Replicators" in Clustering/Server section ?
image

edlerd: > You're talking about the "Replicators" in Clustering/Server section ?

Yes, that's what I was thinking about. Though now I'm not 100% convinced it needs to be aligned.

kimanhou: No I think it's fine as is.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+              value={formik.values.replica_mode || ""}
+              onChange={(value) => {
+                ensureEditMode(formik);
+                formik.setFieldValue("replica_mode", value);
+              }}
+              error={
+                formik.touched.replica_mode ? formik.errors.replica_mode : null
+              }
+              dropdownClassName="replica-mode-dropdown"
+            />
```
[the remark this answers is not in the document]
kimanhou: The replica cluster should not be ignored in the payload. Fixed

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+                formik.setFieldError("replica_cluster", undefined);
+                void formik.setFieldTouched("replica_cluster", true, false);
+                void formik
+                  .setFieldValue("replica_cluster", value, false)
+                  .then(() => {
+                    void formik.validateField("replica_cluster");
+                  });
+              }}
+              help={getClusterHelpText(formik.values.replica_mode)}
+            />
```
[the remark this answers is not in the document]
kimanhou: The user should be able to select a cluster, before/without a mode.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -1,76 +1,32 @@
...
-import ClusterLinkStatus from "pages/cluster/ClusterLinkStatus";
+import ClusterLinkSelector from "pages/cluster/ClusterLinkSelector";
 
-interface Props {
+interface ReplicatorClusterLinkSelectorProps {
   formik: FormikProps<ReplicatorFormValues>;
 }
 
-export const ReplicatorClusterLinkSelector: FC<Props> = ({
+const ReplicatorClusterLinkSelector: FC<ReplicatorClusterLinkSelectorProps> = ({
```
omarelkashef: Is there a reason for the indirection from ReplicatorClusterLinkSelector to ClusterLinkSelector?

src/pages/cluster/actions/EditReplicatorBtn.tsx
```diff
@@ -39,7 +39,7 @@ const EditReplicatorBtn: FC<Props> = ({
         openEditReplicator(replicator.project, replicator.name);
         onClose?.();
       }}
-      title={disabledReason()}
+      title={disabledReason() ?? "Edit"}
```
omarelkashef, line 43: Nitpick: if the button `hasLabel` is true, it will have "Edit" already. Same with other similar buttons as well.

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
```diff
@@ -82,7 +92,11 @@ const RunReplicatorPreflightChecks: FC<Props> = ({
     id: "replica-cluster-match",
     label: (
       <>
-        Project <ProjectRichChip projectName={project?.name || "default"} />{" "}
+        Project{" "}
+        <ProjectRichChip
+          projectName={project?.name || "default"}
+          urlSuffix="/configuration/replication"
+        />{" "}
         <strong>replica cluster</strong> configuration matches replicator
```
omarelkashef: should this line be negated to be  <strong>replica cluster</strong> configuration does not match replicator?

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,191 @@
...
+      ?.text ?? "-";
+
+  const getClusterHelpText = (mode?: string) => {
+    if (!mode) {
+      return null;
+    }
+
+    return (
+      <>
+        Cluster to replicate{" "}
```
omarelkashef: This is a bit complicated to read. Can we make it into a function with if statements instead?

omarelkashef (commented):
Thank you for this PR, the page looks nice! I have a few comments.

When I click the replicator edit button in the project configuration page, I get a toast notification that the project was updaed

https://github.com/user-attachments/assets/e7f3e685-5a40-4c3f-b57c-78c98e72797e


It would be nice if opening the replicator create form from the project config page can auto select the project, but it might need some refactoring

Image


If I remember correctly (remind if I am wrong) we are only supporting one replicator per project for now. If so, we should disable the create button if there is a replicator already. 

Image

edlerd:
> When I click the replicator edit button in the project configuration page, I get a toast notification that the project was updaed

That is because the button is inside the `<form>` block. We have to set `type="button"` on all buttons we use here to prevent the default `type="submit"` behaviour.

src/pages/cluster/actions/EditReplicatorBtn.tsx
kimanhou: When `hasLabel` is false then it doesn't have any title, which is why I added it.

Which other similar buttons are you talking about ?

omarelkashef: The replicator run and delete buttons also have a title even when the button name is shown.

kimanhou:
> That is because the button is inside the <form> block. We have to set type="button" on all buttons we use here to prevent the default type="submit" behaviour.

Yes, I know. I fixed it for RunReplicatorBtn and DeleteReplicatorBtn buttons.

kimanhou:
> If I remember correctly (remind if I am wrong) we are only supporting one replicator per project for now. If so, we should disable the create button if there is a replicator already.

For now, yes. But in the future there could be multiple replicators for one project. And nothing prevents the user to create multiple replicators for one project, as seen on the screenshots in the description of this PR. They can't be all run at the same time (without changing the config) but they can exist.

kimanhou:
> It would be nice if opening the replicator create form from the project config page can auto select the project, but it might need some refactoring

It can easily be done by putting the project in the URL

omarelkashef:
> > If I remember correctly (remind if I am wrong) we are only supporting one replicator per project for now. If so, we should disable the create button if there is a replicator already.
> 
> For now, yes. But in the future there could be multiple replicators for one project. And nothing prevents the user to create multiple replicators for one project, as seen on the screenshots in the description of this PR. They can't be all run at the same time (without changing the config) but they can exist.

Can they be run separately or once you run one replicator, you can not run the others?

kimanhou:
> Can they be run separately or once you run one replicator, you can not run the others?

You can run everything at the same time. Most of them will just fail. Only one can have the right config.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,204 @@
...
+          )}
+        </Col>
+      </Row>
+
+      <DividerLine />
+
+      <Row>
+        <Col size={12}>
+          Replicators for this project
+          <ReplicatorList
```
edlerd: The empty state is too narrow, the "There are no replicators for this project." line should not wrap.

Can the button here be also a standard one instead of green?

Image

edlerd: The icon in the create replicator button is not adjusting well:

Image

src/pages/cluster/ClusterLinkSelector.tsx
```diff
@@ -0,0 +1,92 @@
...
+      </div>
+    ),
+  }));
+
+  const helpText = (
+    <>
+      {help}
+      {hasNoLinks ? "Create your first" : "Manage your"}{" "}
+      <Link to={getClusterLinkListUrl(isClustered)}>
+        cluster {pluralize("link", links.length)}
```
edlerd: this should always say "manage your cluster links" independent of how many exist.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,204 @@
...
+                <FormEditButton
+                  toggleReadOnly={() => {
+                    ensureEditMode(formik);
+                  }}
+                  disableReason={formik.values.editRestriction}
+                />
+              </div>
+            </div>
+          ) : (
+            <ClusterLinkSelector
```
edlerd: Should we disable this one if replica mode is neither "leader" nor "standby"?

src/pages/projects/forms/ProjectFormMenu.tsx
```diff
@@ -41,6 +42,9 @@ const ProjectFormMenu: FC<Props> = ({
         <ul className="p-side-navigation__list">
           <MenuItem label={PROJECT_DETAILS} {...menuItemProps} />
           <MenuItem label={RESOURCE_LIMITS} {...menuItemProps} />
+          {hasReplicators && (
+            <MenuItem label={REPLICATION} {...menuItemProps} />
```
edlerd, line 45: Should this be 2nd instead of 3rd? It would make the items sorted alphabetically.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    return null;
+  };
+
+  return (
+    <ScrollableForm>
+      <Row>
+        <Col size={12}>
+          {formik.values.readOnly ? (
```
edlerd: I think we shouldn't have a read only mode here. We can directly display the form elements. On change we ensure edit mode. This is then similar to the project details, where the description and other options can be directly changed. We do not override the values here.

Also there is currently a glitch when going into edit mode on the replica cluster. That would vanish with the suggestion above.

edlerd (commented):
Few ideas on the copy and UX.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: I would rather not. The user can setup cluster and mode in whatever order they like.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: Funny the sentence doesn't wrap on my screen 🤔
image
I added classname` empty-state__full-width`. Let me know if it fixes the issue for you.

Yes, I updated the button to not be positive.

src/util/projectConfigFields.tsx
```diff
@@ -40,6 +40,8 @@ const projectConfigFormFieldsToPayload: Record<string, string> = {
   limits_cpu: "limits.cpu",
   limits_memory: "limits.memory",
   limits_processes: "limits.processes",
+  replica_mode: "replica.mode",
```
edlerd: I think this has been changed very recently to a dedicated endpoint. Testing the field with latest/edge results in a  `Invalid project configuration key "replica.mode"` error for me.

We should adjust the UX to allow for displaying the current status (leader or standby). Additionally, we should advertise the dedicated promote / demote endpoints. I am not sure what is the best way to surface this functionality. Maybe similar to cluster member evacuate / restore? Or we display the current status (none / leader / standby) and a single button next to it to (promote / demote)? See also [updated docs](https://canonical.com/lxd/docs/latest/howto/replicators_create/) on this.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+          />
+        </Col>
+      </Row>
+
+      <DividerLine />
+
+      <Row>
+        <Col size={12}>
+          Replicators for this project
+          <ReplicatorList
```
edlerd: In the list component, we continue showing the "Create replicator" and can indeed create multiple replicators on the same link. I think this should be prevented even by the API. Can you please reach out and clarify? We might want to disable the create button if a replicator already exists for the project at this point.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+            disabled={!!formik.values.editRestriction}
+          />
+        </Col>
+      </Row>
+
+      <DividerLine />
+
+      <Row>
+        <Col size={12}>
+          Replicators for this project
```
edlerd: These are just the outgoing replicators. On the standby cluster the list will be empty. I think we need to clarify that as it will not immediately be clear to new users of the feature.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+            disabled={!!formik.values.editRestriction}
+          />
+        </Col>
+      </Row>
+
+      <Row>
+        <Col size={12}>
+          <ClusterLinkSelector
+            name="replica_cluster"
+            label="Replica cluster"
```
edlerd, line 133: I think this should say "None" instead of "Select a cluster" if unset. Wdyt?

src/pages/cluster/ClusterLinkSelector.tsx
edlerd: Can we remove the pluralize here and always say "links"?

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+            }
+            dropdownClassName="replica-mode-dropdown"
+            disabled={!!formik.values.editRestriction}
+          />
+        </Col>
+      </Row>
+
+      <Row>
+        <Col size={12}>
+          <ClusterLinkSelector
```
edlerd, line 131: The empty message should not include the dot at the end:

Image

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  return payload;
+};
+
+interface Props {
+  formik: FormikProps<ProjectFormValues>;
+}
+
+const ProjectReplicaForm: FC<Props> = ({ formik }) => {
```
edlerd: I think we need to guide a bit better in case no cluster link exists. See current empty state. It has a link to "Mange your cluster links", but it is not immediately clear that is the next step to take. Maybe extend the empty state copy in the table to say "Create your first cluster link as a basis for your first replicator" (copy should be improved). Then disable the "Create replicator" and add a secondary button "Create cluster link". Wdyt?

Image

edlerd: Or maybe a better journey is to keep things as they are and move the "Cluster" to be the first element in the replicator form. Then the help text can say "Create your first cluster link" under the first element in the form (instead of "Manage your cluster links.":

Image

edlerd (commented):
Looks great, good iterations. A few ideas for UX are below.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: When using "Create replicator" and a "Replica cluster" is set on the project, then that cluster should be preselected in the create replicator form.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: Question sent

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: In the case of disaster recovery, the roles are switched: leader becomes standby and standby becomes leader. The list of replicators would then not be empty on the original leader/now standby project, even though the project will be standby. 

I will add a warning notification when the project is standby to clarify. But I need help with the copy. It's terrible at the moment:
image

src/pages/projects/CreateProject.tsx
```diff
@@ -127,11 +128,17 @@ const CreateProject: FC = () => {
         values.features_storage_buckets = undefined;
       }
 
+      if (!hasReplicators) {
+        values.replica_mode = undefined;
+        values.replica_cluster = undefined;
+      }
+
```
edlerd:
```suggestion
```
This shouldn't be necessary. We can't see the inputs for those anyhow.

src/pages/projects/EditProject.tsx
```diff
@@ -74,6 +74,11 @@ const EditProject: FC<Props> = ({ project }) => {
         values.features_storage_buckets = undefined;
       }
 
+      if (!hasReplicators) {
+        values.replica_mode = undefined;
+        values.replica_cluster = undefined;
+      }
+
```
edlerd:
```suggestion
```
I think this is not needed.

src/util/usePanelParams.tsx
```diff
@@ -20,6 +20,8 @@ export interface PanelHelper {
   localPeering: string | null;
   imageRegistry: string | null;
   replicator: string | null;
+  replicaProject: string | null;
```
edlerd, line 23:
```suggestion
```
Let's reuse the existing `project` constant, which refers to the `panelProject` query parameter.

src/util/usePanelParams.tsx
```diff
@@ -20,6 +20,8 @@ export interface PanelHelper {
...
+  replicaCluster: string | null;
```
edlerd:
```suggestion
  clusterLink: string | null;
```
Let's make this generic so we can reuse it and don't tie it to the use case.

src/util/usePanelParams.tsx
```diff
@@ -309,9 +318,15 @@ const usePanelParams = (): PanelHelper => {
...
-
-    openCreateReplicator: () => {
-      setPanelParams(panels.createReplicator);
+    openCreateReplicator: (replicaProject, replicaCluster) => {
+      const params: ParamMap = {};
+      if (replicaProject) {
+        params["panel-replica-project"] = replicaProject;
+      }
+      if (replicaCluster) {
+        params["panel-replica-cluster"] = replicaCluster;
```
edlerd:
```suggestion
        params["cluster-link"] = replicaCluster;
```
We only have the `panel-` prefix for the project param, because it clashes with a search/sort param for projects. I think the cluster link we can keep clean to have a nice URL.

src/pages/cluster/ClusterLinkSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return (
+    <CustomSelect
+      {...selectProps}
+      name={name}
+      label={label}
+      options={
+        hasNoLinks
+          ? [{ value: "", label: "No cluster links available" }]
+          : [{ value: "", label: "None" }, ...options]
```
edlerd: I think this needs to be distinguished. In the "Create replica" form, we should say "Select a cluster". In the project configuration, this can say "None".

src/pages/cluster/ClusterLinkSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+        </span>
+      </div>
+    ),
+  }));
+
+  const helpText = (
+    <>
+      {help}
+      Manage your{" "}
+      <Link to={getClusterLinkListUrl(isClustered)}>cluster links</Link>.
```
edlerd: This should adjust to be "Manage your cluster links." / "Create your first cluster link."

src/pages/cluster/actions/RunReplicatorPreflightChecks.tsx
edlerd: This is a check and if it passes the positive form is best. I think on failure it also works as the details will appear below of why it is failing.

src/components/OutputField.tsx
```diff
@@ -6,9 +6,18 @@ interface Props {
   label: string;
   value: string;
   help?: ReactNode;
+  suffix?: ReactNode;
+  isBold?: boolean;
 }
 
-const OutputField: FC<Props> = ({ id, label, value, help }) => {
+const OutputField: FC<Props> = ({
```
edlerd: This was already upstreamed, we should remove `OutputField` from our repository and adjust to the upstream variant.
I don't see a use case for the two new props.

src/components/DividerLine.tsx
```diff
@@ -0,0 +1,13 @@
+import type { FC } from "react";
+
+const DividerLine: FC = () => {
```
edlerd: I don't think we need this, we should use the `<hr>`, see also docs on the [Rule in vanilla](https://vanillaframework.io/docs/patterns/rule).

edlerd (commented):
Great progress! A few points from a deep dive into the code.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: Maybe the notification should always be there, independent of the replic mode of the project. Or we could always show it if the projects replicator list is empty. Because the mode can change.

Copy could be explaining only outgoing replicators are shown here and that any incoming replicators will be configured on the cluster referenced by replica cluster.

src/pages/projects/CreateProject.tsx
kimanhou: I had already removed the line for `replica_mode`. But don't we still need the line for `replica_cluster` ? Wouldn't backend complain if we send an unknown field on project ?

src/util/usePanelParams.tsx
kimanhou: I had originally done that but the behavior was weird because project defaults to `default`. If I opened the panel from a project configuration, the correct project would be pre-selected. But then, if I navigated to the replicator list and click on "Create replicator", then the panel would open with `default` pre-selected. After page refresh, the project in the panel would be "Select an option".

Line 157:
```
project: params.get("panel-project") ?? project?.name ?? "default",
```

src/pages/projects/CreateProject.tsx
edlerd: `values.replica_cluster` will already be `undefined` if `hasReplicators` is false. Because the user should have no way to set it, and even on edit it will be initialized to `undefined`. So no need to set it here or handle it in a special way in the payload building.

src/util/usePanelParams.tsx
edlerd: Oh, right. Good catch! Then let's keep it as is.

src/pages/cluster/ClusterLinkSelector.tsx
kimanhou: Didn't I just change that this morning ?

edlerd: When reviewing this file I got the impression we lost the "Create your first..." special case here.

src/components/OutputField.tsx
kimanhou: Oh yeah sorry, I made the changes when the page has a read mode. Undid them.

I remember Nkeiru upstreamed the component. Is there a ticket to remove this from our codebase and use the upstreamed version ?

src/components/DividerLine.tsx
kimanhou: Ooh I didn't know `<hr>` !

src/components/OutputField.tsx
edlerd: We already use the upstreamed one in most places. There is one usage left (apart from the new ones introduced in this PR). I think that usage is from a PR that was open while we removed the local version and that it was reintroduced accidentally. Do you mind removing it?

src/pages/cluster/ClusterLinkSelector.tsx
kimanhou: Yes I removed it this morning, I thought you always wanted "Manage your cluster links" no matter the number of cluster links.

Image

edlerd: Ah yes sorry, that was ambiguous. So I meant we should avoid "Manage your cluster link" if 1 exists, but always use the plural form.

When none exist, we should still say "Create your first", because there is none to manage :)

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: From Kadin:

> multiple replicators are allowed per project, only unique names within a project are enforced. There's currently no check preventing two replicators in the same project from targeting the same cluster link

edlerd: Thanks for the confirmation. Then all good as is :+1:

src/pages/cluster/ClusterLinkSelector.tsx
kimanhou: 🫠🫠🫠🫠🫠🫠🫠🫠🫠🫠

src/components/DividerLine.tsx
kimanhou: There was an issue with `<hr>`. Fixed in CSS now

src/util/projectConfigFields.tsx
kimanhou: > I think this has been changed very recently to a dedicated endpoint. Testing the field with latest/edge results in a Invalid project configuration key "replica.mode" error for me.

Fixed now. However this seems to work only in a clustered environment. Question sent to backend.

> We should adjust the UX to allow for displaying the current status (leader or standby).

For now, I left it as is. The current replica mode is displayed. When the user changes it, we send an API call to the separate endpoint in part of the update project process. This is not ideal because we always send a call to update the project, even when nothing on the project has changed except the replica mode. However I like it because everything happens under the hood. The user doesn't need to know it's a different endpoint.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: I don't want to clutter this screen when the project is leader.

```
            <Notification
              severity="caution"
              messageElement="div"
              title="Only outgoing replicators are shown"
            >
              <p>
                Because this project is in <strong>standby</strong> mode, it
                acts as a passive destination. Only outgoing replicators are
                shown here. Any incoming replicators are configured on the
                cluster referenced by <strong>replica cluster</strong>.
              </p>
            </Notification>
```

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: Would the order change always or just when we open the side panel from the project configuration ?

image

kimanhou: I implemented both suggestions. It surfaces an issue with naming. When creating a cluster link, we call it "cluster link". But when creating a replicator, we call it "cluster". Chaining the creation of cluster link and replicator makes this discrepancy more obvious than before.

src/pages/cluster/actions/CreateClusterLinkBtn.tsx
```diff
@@ -12,7 +16,8 @@ const CreateClusterLinkBtn: FC = () => {
 
   return (
     <Button
-      appearance="positive"
+      type="button"
+      appearance={isPositive ? "positive" : "default"}
```
edlerd: We must also swap the icon light property for non-positive buttons:

Image

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+              title="Only outgoing replicators are shown"
+            >
+              <p>
+                Because this project is in <strong>standby</strong> mode, it
+                acts as a passive destination. Only outgoing replicators are
+                shown here. Any incoming replicators are configured on the
+                cluster referenced by <strong>replica cluster</strong>.
+              </p>
+            </Notification>
+          )}
```
edlerd: Let's show this regardless of the replica mode. In failover scenarios this could be a leader with no replicas defined, but the other edge has the replicator.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+            <Notification
+              severity="caution"
+              messageElement="div"
+              title="Only outgoing replicators are shown"
+            >
+              <p>
+                Because this project is in <strong>standby</strong> mode, it
+                acts as a passive destination. Only outgoing replicators are
+                shown here. Any incoming replicators are configured on the
+                cluster referenced by <strong>replica cluster</strong>.
```
edlerd:
```suggestion
                cluster (link-name).
```
Let's do the work and look up the user. We should display a rich chip here if the replica.cluster is set. If not we can keep as is.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const getClusterHelpText = (mode?: string) => {
+    if (!mode) {
+      return null;
+    }
+
+    if (mode === "leader") {
+      return (
+        <>
+          Cluster to replicate to. The project will be replicated to this
+          cluster.
```
edlerd:
```suggestion
          Cluster to replicate this project to.
```

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <br />
+        </>
+      );
+    }
+
+    if (mode === "standby") {
+      return (
+        <>
+          Cluster to replicate from. The project will receive data from this
+          cluster.
```
edlerd:
```suggestion
          Cluster this project will receive data from.
```
Wdyt?

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+            }
+            dropdownClassName="replica-mode-dropdown"
+            disabled={!!formik.values.editRestriction}
+          />
+        </Col>
+      </Row>
+
+      <Row>
+        <Col size={12}>
+          <ClusterLinkSelector
```
edlerd, line 131: We should display that no links are available before opening the selector.

https://github.com/user-attachments/assets/eb5aa651-1856-4b17-a2b4-8ea5cd398a1d

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+    return null;
+  };
+
+  return (
+    <ScrollableForm className="project-replica-form">
+      <Row>
+        <Col size={12}>
+          <CustomSelect
+            name="replica_mode"
+            label="Replica mode"
```
edlerd, line 111: I think we should change the ux, so the replica mode switch is a dedicated button. (Promote to leader) or (Demote to standby) with the value on the left as output field. That makes it more clear what is happening, we get in line with the CLI and documentation and we simplify the nested api calls. Wdyt?

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+    {
+      value: "leader",
+      label: getModeLabel("Leader", "Active source project"),
+      text: "Leader",
+    },
+    {
+      value: "standby",
+      label: getModeLabel(
+        "Standby",
+        "Read-only failover target. Acts as a passive receiver for replicated data from the remote leader cluster.",
```
edlerd: Those labels are pretty neat. I think if we have an output field, those could be the help text below the value.

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -149,16 +154,118 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
...
+            </a>
+          </p>
+          <CreateReplicatorButton
+            className="u-no-margin--bottom"
+            project={isProjectConfiguration ? project : undefined}
+            cluster={isProjectConfiguration ? cluster : undefined}
+            isPositive={!isProjectConfiguration}
+            hasClusterLinks={hasClusterLinks}
+          />
+          {!hasClusterLinks && <CreateClusterLinkBtn isPositive={false} />}
```
edlerd:
```suggestion
          {<CreateClusterLinkBtn isPositive={false} />}
```
I think we should always show it, but maybe with a different appearance if a link exists.

src/pages/cluster/panels/CreateReplicatorPanel.tsx
```diff
@@ -144,11 +150,12 @@ const CreateReplicatorPanel: FC = () => {
       isCreating: true,
       name: "",
       description: "",
-      project: "",
-      cluster: "",
+      project: panelProject,
+      cluster: panelClusterLink,
```
edlerd, line 154: There is a problem with pre-setting these: They are not touched and the cluster can be failing validation. For instance because it is not active. In that case this form has a disabled create button and the user is left without a hint of what the problem is. Only after interacting with the cluster to make it touched the error appears.

https://github.com/user-attachments/assets/1ac05ab8-d454-43d3-953e-bc4b12d2d710

edlerd (commented):
I like the new create link and then replicator flow. A few more suggestions to refine the flow on this.

src/pages/cluster/actions/CreateClusterLinkBtn.tsx
kimanhou: Fixed
image

However, I noticed that in load balancers, you don't use an icon for the second button. Should we align on this ?

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: Okay but even when there is no replicators ?

image

src/pages/cluster/actions/CreateClusterLinkBtn.tsx
edlerd: No strong opinion; we can also hide the icon here I guess.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: You mean no "replica cluster" is defined, right? In that case, we probably can hide the notification, yes.

src/pages/cluster/actions/CreateClusterLinkBtn.tsx
kimanhou: I added the icon in load balancer, I prefer it.

image

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: No, I meant "no replicators". The notification says "outgoing replicators are shown" but there are no replicators, isn't it weird? I updated the copy to handle the case when there are no replicators.
image

I think it's still useful to have the notification when no replica cluster is defined, to tell the user what they are about to select.

edlerd: > No, I meant "no replicators". The notification says "outgoing replicators are shown" but there are no replicators, isn't it weird?

The notification is especially useful if there are no replicators. Because the incoming replicators are not shown and the list is likely empty on the standby.

>  I updated the copy to handle the case when there are no replicators.

Right it reads well in the screen above. I'd suggest to remove the "when present".

> I think it's still useful to have the notification when no replica cluster is defined, to tell the user what they are about to select.

yes, good point.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: Weird, I can't reproduce this behavior. 🤔 
And the code seems correct:
```
      options={
        hasNoLinks
          ? [{ value: "", label: "No cluster links available" }]
          : [{ value: "", label: emptyOptionLabel }, ...options]
      }
```

I tried opening two tabs: one with the project configuration and one where I create/delete a cluster link. But I could not reproduce the behavior you sent.

https://github.com/user-attachments/assets/359e20fb-2f8e-4ea1-87b6-4bd2dc51c8e7

I added a small safeguard.

src/pages/cluster/ReplicatorList.tsx
kimanhou: Really ? Isn't it weird without any explanation ? The user might not understand why this button is here.

image

The sentence "Replicators require at least one cluster link to function." would be even more confusing here.

Wdyt ?

edlerd: I thought it was odd that I create a link from here (because none exist) and then the button vanished.

I agree the case with having links and the "Create link" is also wired.

Btw. I think the caption should be "Create cluster link".

Maybe add a sentence to the empty state explaining relation of replicators and links. The docs has this nice explanation: "Replicators are LXD entities that periodically copy instances from one cluster to another across a cluster link."

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: It could be that I removed the links in another tab and came back here without reloading. Or is it because the replica mode was set to leader?

src/pages/cluster/ReplicatorList.tsx
kimanhou: The create cluster link button is also used in ClusterLinkList, I changed the caption there too.
image

A cluster link exists:
image


No cluster links exist:
image

edlerd: Thanks, this is great!

Tiny nitpick: The spacing in the "No cluster links exist" looks slightly off.

kimanhou: It's because there are two paragraphs:
- one for "Replicators periodically copy instances from one cluster to another across a cluster link. "
- one for "There are no replicators configured on this server. <br />To create a replicator, first create a cluster link."

If I put each sentence in its own paragraph, it looks like this :
image

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: > It could be that I removed the links in another tab and came back here without reloading. 

I tried that, but always got the correct "No cluster links available" selected. I can't get "Select one option" and "No cluster links available" in the same dropdown

> Or is it because the replica mode was set to leader?

YES indeed ! When a previously selected cluster link is deleted in another tab, the form can still hold the old value (the deleted link name). At that point, CustomSelect cannot match the current value to any available option, so it falls back to its default toggle label "Select an option", while the options list now only contains "No cluster links available".

The behavior looked different after interacting with replica mode because that path triggers extra Formik updates (touched validation), which can change render timing and make the stale-value state more visible. 

I added a `useEffect` in ClusterLinkSelector, that checks if the selected value still exists in the options list everytime the list changes. If not, it clears the value.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: I put them in the buttons' title

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: No replica mode selected
image

Project is leader
image

Project is standby
image

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+                hasIcon
+                appearance="base"
+                onClick={clearReplicaMode}
+                isLoading={isClearing}
+              >
+                <span>Clear replica mode</span>
+              </Button>
+            </>
+          )}
+        </Col>
```
edlerd, line 126: Thanks for adding those buttons. Let's put the buttons on the same line as the current value to the right side of it, between the label and help text.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+      setIsDemoting(false);
+    });
+  };
+
+  const clearReplicaMode = () => {
+    setIsClearing(true);
+    updateReplicaMode("").finally(() => {
+      setIsClearing(false);
+    });
+  };
```
edlerd: This doesn't work at the moment.

Image

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+          <p>Replicators for this project</p>
+          <Notification
+            severity="caution"
+            title="Only outgoing replicators are shown here."
+          >
+            Incoming replicators are configured on the cluster referenced by{" "}
+            {formik.values.replica_cluster ? (
+              <ClusterLinkRichChip
+                clusterLink={formik.values.replica_cluster}
+              />
```
edlerd:
```suggestion
            Incoming replicators are configured on the replica cluster{" "}
            {formik.values.replica_cluster ? (
              <ClusterLinkRichChip
                clusterLink={formik.values.replica_cluster}
              />
```
A bit better?

src/pages/projects/CreateProject.tsx
```diff
@@ -48,7 +48,7 @@ const CreateProject: FC = () => {
   const queryClient = useQueryClient();
   const controllerState = useState<AbortController | null>(null);
   const [section, setSection] = useState(slugify(PROJECT_DETAILS));
-  const { hasProjectsNetworksZones, hasStorageBuckets } =
+  const { hasProjectsNetworksZones, hasStorageBuckets, hasReplicators } =
```
edlerd, line 52: On creation we should hide the promote/demote buttons.

We should also hide the notification and replicator list. Maybe replace it with a message explaining this is where you can manage the replicators later on. Wdyt?

src/util/projectEdit.tsx
```diff
@@ -132,6 +140,12 @@ export const getProjectPayload = (
             ...imageRestrictionPayload(values),
           }
         : {}),
+
+      ...(values.replica_cluster !== undefined
+        ? {
+            "replica.cluster": values.replica_cluster,
+          }
+        : {}),
```
edlerd: This should come from the `ProjectReplicaForm` similar to how the other payloads are assembled with a function call. See `imageRestrictionPayload`.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+            ) : (
+              <strong>replica cluster</strong>
+            )}
+            .
+          </Notification>
+          <ReplicatorList
```
edlerd: This is a scrollable table and as such if the browser height is too low, there will not be any space to display the replicators:

Image

This project has a replicator, but the content of the table has a 0 height.

I think we should not make it a scrollable table in here as the table has too much content above.

src/api/projects.tsx
```diff
@@ -116,3 +116,29 @@ export const deleteProject = async (
...
+export const updateProjectReplicaMode = async (
+  projectName: string,
+  replicaMode: "leader" | "standby" | "",
+  force?: boolean,
+): Promise<LxdOperationResponse> => {
+  const params = new URLSearchParams();
+  if (force) {
+    params.set("force", "true");
+  }
+  const url = `${ROOT_PATH}/1.0/projects/${encodeURIComponent(projectName)}/state${force ? `?${params.toString()}` : ""}`;
```
edlerd:
```suggestion
  const url = `${ROOT_PATH}/1.0/projects/${encodeURIComponent(projectName)}/state?${params.toString()}`;
```
Let's simplify this.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: The empty state should probably be left aligned when displayed in here. Also I think in this case (links exist, but no replicator for this project) The "Create link" button should be a lesser priority as the "Create replicator" one. We can use the "base" appearance here.

Image

On the server/cluster > Replicators page the button appearance is fine, as the "Create replicator" is positive and the "Create link" is standard:

Image

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,273 @@
...
+                name="Clear replica mode"
+                hasIcon
+                appearance="base"
+                onClick={clearReplicaMode}
+                isLoading={isClearing}
+              >
+                <span>Clear replica mode</span>
+              </Button>
+            </>
+          )}
```
edlerd: I like the new experience around those buttons, with showing the relevant one paired with the "clear value" in lower appearance.

edlerd (commented):
Great progress, thanks for the updates!

A few comments, mostly small design refinements. Code mostly LGTM, just one bug left with clearing the replicator mode.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: As discussed, issue lies in the backend. Kadin opened a PR yesterday.
https://github.com/canonical/lxd/pull/18459

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: How do we feel about this ?

image

src/pages/projects/CreateProject.tsx
kimanhou: Project creation

image

What do we think ?

I'm also considering changing the severity of the "Only outgoing replicators are shown here" to information. Any objection ?

edlerd: Nice, I think it is clear.

I wonder if we should remove the last sentence, "Only outgoing..." as it seems not so relevant at this step of creating the project. Once a user returns to the section on a created project, the other notification will teach that bit.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: Nice, looks perfect :+1:

src/pages/projects/CreateProject.tsx
edlerd: Severity as "information" makes perfect sense, yes.

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: image

edlerd: Looks great, thanks!

One tiny remark: I think we should only have one positive button. Can we use the "default" appearance for "Create replicator" and "base" appearance for "Create cluster link" in the project context?

kimanhou: Oh yes sorry I misunderstood the first time.

image

src/pages/cluster/actions/CreateReplicatorBtn.tsx
```diff
@@ -4,28 +4,52 @@ import usePanelParams from "util/usePanelParams";
...
-  );
+  const isDisabled =
+    !hasClusterLinks ||
+    !allProjects.some((project) => canCreateReplicators(project));
+
+  const getTitle = () => {
+    if (!hasClusterLinks) {
+      return "You cannot create replicators because there are no cluster links.";
+    }
+    if (!allProjects.some((project) => canCreateReplicators(project))) {
```
edlerd, line 37: If a project is specified by props, we should check only that one.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+    "replica.cluster": values.replica_cluster,
+  };
+};
+
+interface Props {
+  formik: FormikProps<ProjectFormValues>;
+  isEdit: boolean;
+}
+
+const ProjectReplicaForm: FC<Props> = ({ formik, isEdit }) => {
```
edlerd, line 33: This error is confusing. I was doing a replica restore and think the message could be a bit more clear. What is "target project" refer to is not immediately obvious to a user that is not familiar with the internal system model. I think this directly comes out of the API response so might be an issue for @kadinsayani

Image

src/pages/cluster/actions/RunReplicatorBtn.tsx
```diff
@@ -209,7 +213,11 @@ const RunReplicatorBtn: FC<Props> = ({
                 This will sync {numberOfInstances}{" "}
                 {pluralize("instance", numberOfInstances)} from the{" "}
                 <ClusterLinkRichChip clusterLink={clusterLink} /> cluster back
-                to the <ProjectRichChip projectName={replicator.project} />{" "}
+                to the{" "}
+                <ProjectRichChip
+                  projectName={replicator.project}
+                  urlSuffix="/configuration/replication"
+                />{" "}
                 project.
```
edlerd: In case of restore, we don't know how many instances exist on the other cluster. So we should change the message here as the count now refers to the local project.

src/pages/cluster/ReplicatorClusterLinkSelector.tsx
```diff
@@ -1,75 +1,33 @@
...
-import { getClusterLinkListUrl } from "util/clusterLink";
+import ClusterLinkSelector from "pages/cluster/ClusterLinkSelector";
 
 interface Props {
   formik: FormikProps<ReplicatorFormValues>;
+  id?: string;
 }
 
-export const ReplicatorClusterLinkSelector: FC<Props> = ({
+const ReplicatorClusterLinkSelector: FC<Props> = ({
```
edlerd: Shall we remove this component and directly use `ClusterLinkSelector` in `ReplicatorForm`?

src/pages/cluster/ReplicatorForm.tsx
```diff
@@ -20,6 +20,14 @@ interface Props {
 export const ReplicatorForm: FC<Props> = ({ formik, isEdit = false }) => {
   const { data: projects = [] } = useProjects();
 
+  useEffect(() => {
+    if (isEdit) {
+      return;
+    }
+
+    focusField("replicatorClusterLink");
+  }, [isEdit]);
```
edlerd: I think we should always focus on the cluster, not only on creation. Also can we do this without a use effect? `CustomSelect` has a `takeFocus` prop that we can expose instead.

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -151,16 +156,117 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
...
+            </a>
+          </p>
+          <CreateReplicatorButton
+            className="u-no-margin--bottom"
+            project={isProjectConfiguration ? project : undefined}
+            cluster={isProjectConfiguration ? cluster : undefined}
+            hasClusterLinks={hasClusterLinks}
+            isPositive={false}
+          />
+          <CreateClusterLinkBtn isPositive={false} />
```
edlerd: This is a bit more complicated, on the global replicators page, we should have this one positive and the create link one standard button. Only in the project context those should be base and and standard.

It might be simpler to expose the button `appearance` directly instead of the boolean `isPositive` to solve this.

So on the project this is correct:

Image

On the global list there should be a positive button:

Image

src/pages/projects/CreateProject.tsx
```diff
@@ -136,6 +139,11 @@ const CreateProject: FC = () => {
             ...projectDetailRestrictionPayload(values),
             ...resourceLimitsPayload(values),
             ...restrictions,
+            ...(hasReplicators && replicaCluster
+              ? {
+                  "replica.cluster": replicaCluster,
+                }
+              : {}),
```
edlerd: Can we use the `replicaPayload` helper here?

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+                {getModeActionButtons(formik.values.replica_mode)}
+              </div>
+            </div>
+          </Field>
+        </Col>
+      </Row>
+
+      <Row>
+        <Col size={12}>
+          <ClusterLinkSelector
```
edlerd, line 131: When I open project creation and open the replication section, this field immediately gets focus. I think that is unexpected:

Image

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+          Cluster this project will receive data from.
+          <br />
+        </>
+      );
+    }
+
+    return null;
+  };
+
+  const updateReplicaMode = async (mode: "leader" | "standby" | "") => {
```
edlerd: Let's define a type for the replica mode, so we don't repeat the strings here.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+            }
+            onChange={(value) => {
+              ensureEditMode(formik);
+              formik.setFieldError("replica_cluster", undefined);
+              void formik.setFieldTouched("replica_cluster", true, false);
+              void formik
+                .setFieldValue("replica_cluster", value, false)
+                .then(() => {
+                  void formik.validateField("replica_cluster");
+                });
```
edlerd: I think there is no validation on this field, can this be simpler?

```suggestion
              void formik.setFieldValue("replica_cluster", value);
```

edlerd (commented):
Great stuff, few comments from doing very thorough QA and code review.

src/pages/cluster/actions/RunReplicatorBtn.tsx
edlerd: We can say "All instances" as we can't know the exact number.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: I think this is a side effect of the use effect I commented on in https://github.com/canonical/lxd-ui/pull/2043#discussion_r3429479155 It should vanish once that is addressed.

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -151,16 +156,119 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
...
+              Learn more about replicators
+              <Icon className="external-link-icon" name="external-link" />
+            </a>
+          </p>
+          <CreateReplicatorButton
+            className="u-no-margin--bottom"
+            project={isProjectConfiguration ? project : undefined}
+            cluster={isProjectConfiguration ? cluster : undefined}
+            hasClusterLinks={hasClusterLinks}
+            appearance={!isProjectConfiguration ? "positive" : "default"}
```
edlerd:
```suggestion
            appearance={isProjectConfiguration ? "default" : "positive"}
```

src/pages/projects/forms/ProjectReplicaForm.tsx
kadinsayani: Yes, this is an error returned by https://github.com/canonical/lxd/blob/2a7db370f66decceb8a4d65dc638708d00604043/lxd/api_replicators.go#L1808. What would make the error message more clear?

src/types/project.d.ts
```diff
@@ -1,10 +1,14 @@
...
+
 export interface LxdProject {
   name: string;
   config: LxdConfigPair;
   description: string;
   used_by?: string[];
   etag?: string;
   access_entitlements?: string[];
+  replica_mode?: ProjectReplicaMode;
+  replica_cluster?: string;
```
edlerd:
```suggestion
```
This field does not exist here. It is nested under `config` as `replica.cluster`.

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -151,16 +156,119 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
...
+            onUpdateSort={updateSort}
+          />
+        ) : (
+          <ScrollableTable
+            dependencies={[replicators, notify.notification]}
+            tableId="replicator-table-id"
+            belowIds={
+              isProjectConfiguration
+                ? ["form-footer", "status-bar"]
+                : ["status-bar"]
```
edlerd:
```suggestion
            belowIds={["status-bar"]
```
This will always be true due to the outer wrapper.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+          title={
+            isEdit
+              ? "Project will become the active replication source."
+              : "Project must be created before it can be promoted to leader."
+          }
+          className="u-no-margin--bottom"
+          disabled={!isEdit}
+        >
+          <span>{isSmallScreen ? "Promote" : "Promote to leader"}</span>
+        </Button>,
```
edlerd: Shall we push those buttons into their own component? We could also extract the `updateReplicaMode` into something a bit more generic, with callbacks, that can be hooked up with the dedicated  button components. Loading state could be contianed per button then as well and this component can become a bit simpler.

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -171,68 +279,24 @@ const ReplicatorList: FC<Props> = ({ variant = "main" }) => {
...
-        </Row>
-      </Element>
+        {tableContent}
+      </BaseElement>
+    );
+  };
 
+  return (
+    <>
+      {renderLayout()}
```
edlerd, line 278: This `renderLayout` and `tableContent` is not ideal, but I have no direct idea how to simplify this.

edlerd (commented):
Great progress, nearly perfect, probably last round of comments below.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: How about this copy: `Project on the replica cluster :clusterName must be in leader mode to run replicator restore` @kadinsayani ?

edlerd: Or we could avoid repeating the cluster name in the error message as well. Not sure if LXD usually does that or not.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: Meaning introducing a `PromoteProjectBtn`, `DemoteProjectBtn` and `ClearProjectReplicaModeBtn`.

kimanhou:
Test `cluster member evacuate and restore` is failing but I don't know why.

src/pages/projects/forms/ProjectReplicaForm.tsx
kadinsayani: I'd still like to keep the precision of naming both projects explicitly rather than just saying "must be in leader mode" with no subject, but I agree the message should tie the role to something concrete. How about:

`Target project (on the remote leader cluster) must be in leader mode to run replicator in restore mode`

`Source project (on this cluster) must be in standby mode to run replicator in restore mode`

edlerd: Right, I would slightly simplify:

`Project on the remote cluster must be in leader mode to run replicator in restore mode`

and

`Local project must be in standby mode to run replicator in restore mode`

Because both "target" and "source" have ambiguous meanings as we switch roles in a restore situation. So I'd avoid those terms completely.

src/pages/projects/actions/ClearProjectReplicaModeBtn.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <Button
+      type="button"
+      name="Clear replica mode"
+      hasIcon
+      appearance="base"
+      onClick={clearReplicaMode}
+      isLoading={isClearing}
+      className="u-no-margin--bottom"
+      disabled={!isEdit}
```
edlerd: Are we missing a permission check here?

src/util/projects.tsx
```diff
@@ -92,3 +94,24 @@ export const getInstancesUsedByProject = (project: LxdProject): string[] => {
 
   return project.used_by.filter((item) => item.startsWith("/1.0/instances/"));
 };
+
+export const updateReplicaMode = async (
+  project: string,
+  mode: ProjectReplicaMode,
+  onOperationIdMissing: () => void,
+  onSuccess: () => void,
+  onFailure: (e: unknown) => void,
```
edlerd:
```suggestion
  onSuccess: () => void,
  onFailure: (e: unknown) => void,
```
Let's simplify and have a single on failure callback. We can create the "missing id" error inside here for the other case.

src/pages/projects/actions/ClearProjectReplicaModeBtn.tsx
```diff
@@ -0,0 +1,70 @@
...
+        toastNotify.success(
+          <>
+            Replica mode cleared for project{" "}
+            <ProjectRichChip projectName={project} />.
+          </>,
+        );
+      },
+      (e: unknown) => {
+        notify.failure("Could not clear replica mode for project.", e as Error);
+      },
```
edlerd: Let's define `handleFailure` and `handleSuccess` handlers, so this becomes easier to read.

edlerd: Same for the other two buttons.

src/pages/projects/actions/ClearProjectReplicaModeBtn.tsx
edlerd: Same for the other two buttons.

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+            </Notification>
+          ) : (
+            <Notification
+              severity="information"
+              title="Outgoing replicators will be managed here."
+            >
+              Replicators can be created and managed once this project has been
+              successfully created.
+            </Notification>
+          )}
```
edlerd, line 167: Let's move this to be after the list. A user will look for this below the list most likely.

Image

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+            emptyOptionLabel="None"
+          />
+        </Col>
+      </Row>
+
+      <hr></hr>
+
+      <Row>
+        <Col size={12}>
+          <p>Replicators for this project</p>
```
edlerd:
```suggestion
          <h2 className="p-heading--4 u-no-margin--bottom">Replicators for this project</h2>
```
Let's use a heading for this. Maybe we should also remove the `hr` above.

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: However I think in the empty state it should be first.

I think the empty state is a bit confusing with the content and layout.

Image

Maybe hide the title of the empty state in the project context and bring it a bit closer to the notification, so the margin matches above and below the notification:

Image

kimanhou:
Test `cluster member evacuate and restore` is failing because there are 2 members in the cluster and evacuating one would break majority. How come this surfaces only in my PR ?

src/util/projects.tsx
kimanhou: Euh how ? We can't use `notify` in this function

edlerd:
> Test `cluster member evacuate and restore` is failing because there are 2 members in the cluster and evacuating one would break majority. How come this surfaces only in my PR ?

I saw it also on other PRs, a new check introduced by @kadinsayani few days ago: https://github.com/canonical/lxd/commit/a2d226cd9defa09955e8df35359332b233ffaa7e

src/util/projects.tsx
edlerd: Instead of the dedicated `onOperationIdMissing` we can call `onFailure(new Error("operation id missing"));`

src/pages/projects/forms/ProjectReplicaForm.tsx
kimanhou: Does the empty state need the replicator icon ?

image


By the way, have you noticed I changed the behavior when a project config holds a cluster link but the cluster link was deleted ? (see replica cluster in the above screenshot ⬆️)

kimanhou:
> I saw it also on other PRs, a new check introduced by @kadinsayani few days ago: https://github.com/canonical/lxd/commit/a2d226cd9defa09955e8df35359332b233ffaa7e

To fix this, we need to create another member so that the cluster has 3 members. Can you think of another solution ?

src/pages/projects/actions/PromoteProjectBtn.tsx
```diff
@@ -0,0 +1,76 @@
...
+import ProjectRichChip from "pages/projects/ProjectRichChip";
+import { useProjectEntitlements } from "util/entitlements/projects";
+import { updateReplicaMode } from "util/projects";
+
+interface Props {
+  project: string;
+  isEdit: boolean;
+}
+
+const PromoteProjectBtn: FC<Props> = ({ project, isEdit }: Props) => {
```
kimanhou, line 17: Should we disable Promote button if the project's cluster is errored ? Or do we just let the user run into a backend error ?

src/pages/projects/forms/ProjectReplicaForm.tsx
edlerd: > Does the empty state need the replicator icon ?

Without icon looks better in this context, good idea.

> By the way, have you noticed I changed the behavior when a project config holds a cluster link but the cluster link was deleted ? (see replica cluster in the above screenshot ⬆️)

Yes, very nice :)

src/pages/projects/actions/PromoteProjectBtn.tsx
edlerd: I think we don't need the additional cluster validation for the button.

kimanhou: Good, I tried, it's annoying

edlerd:
> > I saw it also on other PRs, a new check introduced by @kadinsayani few days ago: [canonical/lxd@a2d226c](https://github.com/canonical/lxd/commit/a2d226cd9defa09955e8df35359332b233ffaa7e)
> 
> To fix this, we need to create another member so that the cluster has 3 members. Can you think of another solution ?

Yes, that should be a viable solution. I just hope GitHub runners will be able to handle a 3rd one with their limited ressources.

src/sass/_project_configuration.scss
```diff
@@ -46,3 +46,31 @@
...
+
+    .p-form-help-text {
+      max-width: 15rem;
+    }
+  }
+
+  hr {
+    width: unset;
+  }
+
```
edlerd:
```suggestion
```
This can be removed?

src/sass/_project_configuration.scss
```diff
@@ -46,3 +46,31 @@
...
+  .empty-state.empty-state__full-width {
+    > .col-4 {
+      display: none;
+    }
+  }
```
edlerd:
```suggestion
```
I quite like the new slim empty state on the project configuration. I think we can simply show the "contents" of the empty state there. So avoid the empty state wrapper completely and have the empty state wrapper only in the non-project context. Then we can clean up the css override here and the many special cases within the empty state. Assign the "content" with its conditionals to a var and use it either with or without empty state wrapper.

edlerd (commented):
Great stuff, thanks!

edlerd (approved):
LGTM, good work!

src/pages/projects/forms/ProjectReplicaForm.tsx
kadinsayani: Works for me! I'll update the messages.

Thanks :)

kadinsayani: https://github.com/canonical/lxd/pull/18459

edlerd: Nice, thanks @kadinsayani

kadinsayani: Np! Thanks for flagging

PR #2045, opened 2026-06-10
fix(cluster-link) ensure entered token is only sent with token consume and not with token generate mode

edlerd:
## Done

- ensure entered token is only sent with token consume and not with token generate mode

Fixes WD-36882

## QA

1. Run the LXD-UI:
    - start cluster link creation, select "I have a token", enter token, then switch back to "generate token"
    - ensure the cluster link is generated.


omarelkashef (approved):
LGTM, thank you for the fix!

PR #2046, opened 2026-06-10
feat(instance creation): focus on newly added profile

### LINKED ISSUE #2023: Feedback about profiles in configuration
Hello 🎉101st issue !

I would like to give some feedback for the profiles part on the instance configuration.

Image

After adding a profile I get this:

Image

But it happens so quickly that I often don't catch what profile is added.

Perhaps it would be a nice idea to leave the dropdown menu empty, instead of automatically selecting the alphabetical first profile available. To make sure that no incorrect profiles get added to an instance.

Another thing I would like to mention is that I think the info description for profiles can be more clear regarding overrides. The bottom profile takes precedence/priority. I think someone would expect that the top profile would do that. I'm not suggesting you switch this around, because it should reflect on the yaml config. Just that the description could be better.

Also great job with the expanded yaml view, the way you inject the comments makes it very clear!

  # Metadata
    
  UI Version: 0.21
  Path: https://<ANONYMOUS_HOST>/ui/project/default/instance/wireguard/configuration

kimanhou:
## Done

- On instance creation, when adding a new profile: newly added profile is focused
- Improve message in tooltip

Fixes #2023

## QA

1. Run the LXD-UI:
    - Click on Create instance, select an image.
    - Click on "Add profile": the newly added profile should be focused.

## Screenshots

https://github.com/user-attachments/assets/3ae44a3d-b72a-4aba-8f97-94fa4abfaed7

### DIFF
--- a/src/pages/profiles/ProfileSelector.tsx
+++ b/src/pages/profiles/ProfileSelector.tsx
@@ -1,4 +1,4 @@
-import { useEffect, type FC } from "react";
+import { type FC } from "react";
 import {
   Button,
   Icon,
@@ -9,6 +9,7 @@ import {
   Tooltip,
 } from "@canonical/react-components";
 import { defaultFirst } from "util/helpers";
+import { focusField } from "util/formFields";
 import { useProfiles } from "context/useProfiles";
 
 interface Props {
@@ -34,14 +35,6 @@ const ProfileSelector: FC<Props> = ({
 
   const { data: profiles = [], error, isLoading } = useProfiles(project);
 
-  useEffect(() => {
-    const contentdetails = document.getElementById("content-details");
-
-    if (contentdetails) {
-      contentdetails.scrollTop = contentdetails.scrollHeight;
-    }
-  }, [selected]);
-
   if (isLoading) {
     return <Spinner className="u-loader" text="Loading profiles..." />;
   }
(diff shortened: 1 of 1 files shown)


src/pages/profiles/ProfileSelector.tsx
```diff
@@ -42,6 +43,31 @@ const ProfileSelector: FC<Props> = ({
     }
   }, [selected]);
 
+  useEffect(() => {
```
edlerd: This shouldn't be a use effect that reacts to the side effect of a state update. Instead, we should hook this up directly with the addProfile function. Actually, you can use the `focusField` helper, which uses a short timeout to ensure the new field is available.

src/pages/profiles/ProfileSelector.tsx
```diff
@@ -65,6 +91,7 @@ const ProfileSelector: FC<Props> = ({
   const addProfile = () => {
```
edlerd, line 58: There is another `useEffect` in this file that should be removed. It leads to unexpected scroll behaviour in some conditions such as when a profile is removed.

That use effect was introduced so that when adding a profile, we ensure the new input is visible. I assume the new focus logic already covers this. Both need further testing though.

edlerd (approved):
Thanks, this is much cleaner interaction and code wise.

Issue #2035, opened 2026-06-04
Inconsistent UX table behaviour


Kxiru:
There seems to be inconsistent behaviour across tables in the UI, where some tables have their rows highlighted on hover (eg volumes, instances, profiles) and others do not (Networks, pools, images).

<img width="1558" height="417" alt="Image" src="https://github.com/user-attachments/assets/06d06945-588f-4959-9dc8-d329271ee2d5" />

edlerd:
This was fixed in the linked PR.

PR #2047, opened 2026-06-10
feat: [WD-37003] Fixes inconsistent table behaviour

Kxiru:
## Done
- Consistent on-hover behaviour throughout the UI.
 
Fixes inconsistency in table behaviour: https://github.com/canonical/lxd-ui/issues/2035

## QA

1. Run the LXD-UI:
    - Find a table within the UI and ensure that a row is highlighted on hover.

## Screenshots

image

### DIFF
--- a/src/components/DeviceListTable.tsx
+++ b/src/components/DeviceListTable.tsx
@@ -70,6 +70,7 @@ const DeviceListTable: FC<Props> = ({ configBaseURL, devices, location }) => {
 
     return {
       key: devicename,
+      className: "u-row",
       columns: [
         {
           content: (
--- a/src/pages/cluster/ClusterGroupForm.tsx
+++ b/src/pages/cluster/ClusterGroupForm.tsx
@@ -108,6 +108,7 @@ const ClusterGroupForm: FC<Props> = ({ formik }) => {
             return {
               key: name,
               name: name,
+              className: "u-row",
               columns: [
                 {
                   content: name,
--- a/src/pages/networks/LoadBalancersTable.tsx
+++ b/src/pages/networks/LoadBalancersTable.tsx
@@ -34,6 +34,7 @@ const LoadBalancersTable: FC<Props> = ({ network, loadBalancers, project }) => {
   const rows = loadBalancers.map((loadBalancer) => {
     return {
       key: loadBalancer.listen_address,
+      className: "u-row",
       columns: [
         {
           content: (
--- a/src/pages/permissions/OidcConfigurationForm.tsx
+++ b/src/pages/permissions/OidcConfigurationForm.tsx
@@ -99,6 +99,7 @@ const OidcConfigurationForm: FC<Props> = ({ closeModal }: Props) => {
       <NotificationRow className="u-no-padding" />
       <MainTable
         id="settings-table"
+        className="oidc-configuration-table"
         headers={headers}
         rows={rows}
         emptyStateMsg="No data to display"
--- a/src/pages/settings/Settings.tsx
+++ b/src/pages/settings/Settings.tsx
@@ -233,7 +233,6 @@ const Settings: FC = () => {
             >
               <MainTable
                 id="settings-table"
-                className="settings-table"
                 headers={headers}
                 rows={rows}
                 emptyStateMsg="No data to display"
--- a/src/pages/settings/Settings.tsx
+++ b/src/pages/settings/Settings.tsx
@@ -233,6 +233,7 @@ const Settings: FC = () => {
             >
               <MainTable
                 id="settings-table"
+                className="settings-table"
                 headers={headers}
                 rows={rows}
                 emptyStateMsg="No data to display"
--- a/src/pages/storage/StoragePools.tsx
+++ b/src/pages/storage/StoragePools.tsx
@@ -60,6 +60,7 @@ const StoragePools: FC = () => {
   const rows = pools.map((pool) => {
     return {
       key: pool.name,
+      className: "u-row",
       columns: [
         {
           content: (
--- a/src/sass/styles.scss
+++ b/src/sass/styles.scss
@@ -546,7 +546,6 @@ body.is-dark .lxd-icon {
 .replicator-rich-tooltip-table,
 .replicator-table,
 .scrollable-table,
-.settings-table,
 .storage-buckets-table,
 .storage-pool-rich-tooltip-table,
 .storage-pool-table,
--- a/src/sass/styles.scss
+++ b/src/sass/styles.scss
@@ -500,3 +500,64 @@ body.is-dark .lxd-icon {
   display: flex;
   justify-content: flex-end;
 }
+
+.cluster-group-table,
+.cluster-link-rich-tooltip-table,
+.cluster-link-table,
+.cluster-member-table,
+.configuration-table,
+.confirm-table,
+.custom-iso-table,
+.device-table,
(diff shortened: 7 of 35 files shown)


src/sass/styles.scss
```diff
@@ -403,12 +403,14 @@ body.is-dark .lxd-icon {
 }
 
 .u-selectable-table-rows {
-  .u-row:hover {
+  .u-row:hover,
+  tbody tr:hover {
     background-color: $colors--theme--background-hover;
     cursor: pointer;
   }
```
edlerd, line 409: This is too broad. Because on ACLs and other tables that do not have an on-click handler on the table row, we shouldn't use the cursor pointer.

src/pages/storage/StoragePools.tsx
```diff
@@ -145,7 +145,7 @@ const StoragePools: FC = () => {
             headers={headers}
             rows={rows}
             sortable
-            className="storage-pool-table"
+            className="storage-pool-table u-selectable-table-rows"
```
edlerd: This seems like the wrong name to use here. Because the table rows are not selectable. Maybe a better approach is to keep a list of the tables that get the highlight on hover in CSS. Such as the `storage-pool-table`, but also all others. Then we don't need the extra CSS class.

edlerd (commented):
Good start, one concern and one suggestion below.

edlerd (changes requested):
There is no hover effect on the storage tables and neither on the network table now.

Kxiru:
> There is no hover effect on the storage tables and neither on the network table now.

Yes, this is far from finished, I'll convert it back to a draft.

edlerd:
@Kxiru is this ready for review or correctly labeled as a draft?

Kxiru:
> @Kxiru is this ready for review or correctly labeled as a draft?

Still a draft!

src/components/ConfigurationRow.tsx
```diff
@@ -232,7 +232,7 @@ export const getConfigurationRowBase = ({
   return {
     name,
     key: name,
-    className,
+    className: classnames("u-row", className),
```
edlerd: I think in the configuration tables the hover effect is not desired. Wdyt, shall we remove it here?

src/components/DeviceListTable.tsx
```diff
@@ -70,6 +70,7 @@ const DeviceListTable: FC<Props> = ({ configBaseURL, devices, location }) => {
 
     return {
       key: devicename,
+      className: "u-row",
```
edlerd, line 73: Good idea to add it here as well. Though there is the network table on the instance overview (directly above this one) still missing the hover effect.

src/pages/permissions/OidcConfigurationForm.tsx
```diff
@@ -99,6 +99,7 @@ const OidcConfigurationForm: FC<Props> = ({ closeModal }: Props) => {
       <NotificationRow className="u-no-padding" />
       <MainTable
         id="settings-table"
+        className="oidc-configuration-table"
```
edlerd, line 102: This one doesn't work yet.

src/pages/settings/Settings.tsx
```diff
@@ -233,6 +233,7 @@ const Settings: FC = () => {
             >
               <MainTable
                 id="settings-table"
+                className="settings-table"
```
edlerd: This one is missing the hover effect.

src/pages/cluster/ClusterGroupForm.tsx
```diff
@@ -108,6 +108,7 @@ const ClusterGroupForm: FC<Props> = ({ formik }) => {
             return {
               key: name,
               name: name,
+              className: "u-row",
```
edlerd, line 111: This is not working yet.

edlerd (changes requested):
Please check all the tables. Several are not to be configured correctly yet.

Also, I think the configuration tables might be better without. All the others seem fine.

src/components/ConfigurationRow.tsx
Kxiru: image
I actually like the configuration table hover. What don't you like about it? I'll try to remove the column header hover highlight.

edlerd: Let's get others to review to get more broad view on this.

edlerd: Let's not have the hover effect on the configuration tables:

Image

src/components/DeviceListTable.tsx
edlerd, line 73: This one works fine on the instance overview, but the networks table on the instance overview does not have the hover effect yet, but should have:

Image

src/pages/networks/LoadBalancersTable.tsx
```diff
@@ -34,6 +34,7 @@ const LoadBalancersTable: FC<Props> = ({ network, loadBalancers, project }) => {
   const rows = loadBalancers.map((loadBalancer) => {
     return {
       key: loadBalancer.listen_address,
+      className: "u-row",
```
edlerd, line 37: On the cluster groups page the effect is not working yet:

Image

edlerd (changes requested):
Few tables are in the wrong state still.

src/pages/networks/LoadBalancersTable.tsx
Kxiru: I have looked at the code but I have yet to verify this in the correct environment.

src/sass/styles.scss
```diff
@@ -500,3 +500,63 @@ body.is-dark .lxd-icon {
   display: flex;
   justify-content: flex-end;
 }
+
+.cluster-group-table,
```
omarelkashef, line 504: A non-blocking suggestion: if these classes are only used for the purpose of highlighting the hovered row, we can replace them with one generic class (e.g. or hoverable-table entity-table).

omarelkashef (commented):
The tables look better with the hover effect indeed. Only one non-blocking comment below

src/sass/styles.scss
Kxiru: I appreciate the granularity of seeing which specific tables are applied in styles.css. If a table were to be removed, both approaches would still require two changes (removing the css class, and removing the u-row class), so I do not see an increase in efficiency either. I'm happy to leave as is, but @kimanhou , @edlerd , @Sophie-32 wdyt? :)

edlerd: I think it is okay to maintain the list here.

src/components/ConfigurationRow.tsx
```diff
@@ -235,7 +235,7 @@ export const getConfigurationRowBase = ({
   return {
     name,
     key: name,
-    className,
+    className: className,
```
edlerd: this change can be reverted.

edlerd (dismissed):
LGTM, one nitpick below.

src/sass/styles.scss
Sophie-32: I'm not familiar with this item, am I the right Sophie?

Kxiru: @Sophie-32 No! Sorry, haha!

edlerd (approved):

src/sass/styles.scss
kimanhou: For future reference, the correct handle name for Sophie is @sophiecarroll

PR #2048, opened 2026-06-10
chore(clustered tests): setup another cluster and add cluster links tests

kimanhou:
## Done

- Extend the infrastructure of clustered tests to have a second one-node cluster. WD-37019
- Add tests for cluster links. WD-37020

## QA

1. Run the LXD-UI:
    - Trust CI

## Screenshots

image

### DIFF
--- a/tests/cluster-links-clustered.spec.ts
+++ b/tests/cluster-links-clustered.spec.ts
@@ -1,10 +1,12 @@
-import { test } from "./fixtures/lxd-test";
+import { test, expect } from "./fixtures/lxd-test";
 import {
   createClusterLink,
   deleteClusterLink,
   editClusterLink,
   randomLinkName,
   skipIfNotSupported,
+  createClusterLinkOnRemoteCluster,
+  deleteClusterLinkOnRemoteCluster,
 } from "./helpers/cluster-links";
 import { skipIfNotClustered } from "./helpers/cluster";
 
@@ -17,6 +19,64 @@ test("cluster link create edit delete", async ({
 
   const link = randomLinkName();
   await createClusterLink(page, link);
+  const row = page.getByRole("row").filter({ hasText: link });
+  await expect(row).toBeVisible();
+  await expect(row.getByRole("cell", { name: "Type" })).toHaveText(
+    "Bidirectional",
+  );
+  await expect(row.getByRole("cell", { name: "Status" })).toHaveText("Pending");
+  await expect(row.getByRole("cell", { name: "Addresses" })).toHaveText("-");
+  await expect(row.getByRole("cell", { name: "Description" })).toHaveText("");
+
   await editClusterLink(page, link);
+  await expect(row.getByRole("cell", { name: "Description" })).toHaveText(
+    "My link",
+  );
+
+  await deleteClusterLink(page, link);
+  await expect(row).toHaveCount(0);
+});
+
+test("cluster link table displays all links", async ({
+  page,
+  lxdVersion,
+}, testInfo) => {
+  skipIfNotSupported(lxdVersion);
+  skipIfNotClustered(testInfo.project.name);
+
+  const link1 = randomLinkName();
+  const link2 = randomLinkName();
+  await createClusterLink(page, link1);
+  await createClusterLink(page, link2);
+
+  const row1 = page.getByRole("row").filter({ hasText: link1 });
+  const row2 = page.getByRole("row").filter({ hasText: link2 });
+
+  await expect(row1).toBeVisible();
+  await expect(row2).toBeVisible();
+
+  await deleteClusterLink(page, link1);
+  await deleteClusterLink(page, link2);
+});
+
+test("consume token to create cluster link", async ({
+  page,
+  lxdVersion,
+}, testInfo) => {
+  skipIfNotSupported(lxdVersion);
+  skipIfNotClustered(testInfo.project.name);
+
+  const link = randomLinkName();
+
+  const token = createClusterLinkOnRemoteCluster(link);
+  await createClusterLink(page, link, token);
+
+  const row = page.getByRole("row").filter({ hasText: link });
+  await expect(row.getByRole("cell", { name: "Status" })).toHaveText(
+    "Reachable",
+  );
+  await expect(row.getByRole("cell", { name: "Auth groups" })).toHaveText("1");
+
+  deleteClusterLinkOnRemoteCluster(link);
   await deleteClusterLink(page, link);
 });
--- a/tests/helpers/cluster-links.ts
+++ b/tests/helpers/cluster-links.ts
@@ -3,6 +3,7 @@ import { randomNameSuffix } from "./name";
 import type { Page } from "@playwright/test";
 import { gotoURL } from "./navigate";
 import { dismissNotification } from "./notification";
+import { runCommand } from "./shell";
 
 export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
   test.skip(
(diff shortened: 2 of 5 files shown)


tests/scripts/start_vm_cluster.sh
```diff
@@ -62,9 +72,17 @@ cluster:
   cluster_token: ${TOKEN}
 EOF
 
+# Bootstrap VM3 as an unclustered standalone LXD.
+sudo lxc exec vm3 -- lxd init --auto
+sudo lxc exec vm3 -- lxc config set core.https_address :8443
+
+
```
edlerd:
```suggestion

```
Nitpick

tests/cluster-links-clustered.spec.ts
```diff
@@ -17,6 +18,77 @@ test("cluster link create edit delete", async ({
...
+  const link = randomLinkName();
+  const targetVM = process.env.LXD_UI_CLUSTER_LINK_TARGET_VM;
+  if (!targetVM) {
+    throw new Error("Missing required env var: LXD_UI_CLUSTER_LINK_TARGET_VM");
+  }
+  const authGroupName = `link-${link}`;
+  const generateTokenCommand = `
+    lxc auth group create ${authGroupName} &&
+    lxc auth group permission add ${authGroupName} server admin &&
+    lxc cluster link create ${link} --auth-group ${authGroupName}
```
edlerd:
```suggestion
    lxc cluster link create ${link} --auth-group admins
```
Let's rely on the default `admins` group.

tests/cluster-links-clustered.spec.ts
```diff
@@ -17,6 +18,77 @@ test("cluster link create edit delete", async ({
...
+  `;
+  const output = runCommand(
+    `lxc exec ${targetVM} -- sh -c '${generateTokenCommand}'`,
```
edlerd: `targetVM` is the ip, I think this doesn't work and we need the name `vm3` here instead.

tests/cluster-links-clustered.spec.ts
```diff
@@ -17,6 +18,77 @@ test("cluster link create edit delete", async ({
...
+  const row2 = page.getByRole("row").filter({ hasText: link2 });
+
+  await expect(row1).toBeVisible();
+  await expect(row2).toBeVisible();
+
+  await deleteClusterLink(page, link1);
+  await deleteClusterLink(page, link2);
+});
+
+test("consume token to create cluster link", async ({
```
edlerd, line 62: We should clean up the cluster link and delete it at the end.

tests/cluster-links-clustered.spec.ts
```diff
@@ -17,6 +18,77 @@ test("cluster link create edit delete", async ({
...
+    lxc cluster link create ${link} --auth-group ${authGroupName}
+  `;
+  const output = runCommand(
+    `lxc exec ${targetVM} -- sh -c '${generateTokenCommand}'`,
+  )
+    .toString()
+    .trim();
+
+  // Extract token from the output (it's on the last line)
+  const token = output.split("\n").pop() || "";
```
edlerd: Let's push the "create a cluster link token on the other vm" into a function, we can easily reuse it for other tests. I think we should also have a "Create all existing cluster links" cleanup at the beginning of that function so we make things re-tryable.

edlerd (commented):
Looks promising, few comments below.

tests/cluster-links-clustered.spec.ts
kimanhou: Nope, I don't save vm3's IP. `LXD_UI_CLUSTER_LINK_TARGET_VM` is the name `vm3`.
In start_vm_cluster.sh line 83:
```
LXD_UI_CLUSTER_LINK_TARGET_VM=vm3
```

edlerd: Ah right, I misread the line in the shell script.

tests/cluster-links-clustered.spec.ts
```diff
@@ -17,6 +19,67 @@ test("cluster link create edit delete", async ({
...
+}, testInfo) => {
+  skipIfNotSupported(lxdVersion);
+  skipIfNotClustered(testInfo.project.name);
+
+  const link = randomLinkName();
+  const targetVM = process.env.LXD_UI_CLUSTER_LINK_TARGET_VM;
+  if (!targetVM) {
+    throw new Error("Missing required env var: LXD_UI_CLUSTER_LINK_TARGET_VM");
+  }
+  const token = createClusterLinkonRemoteCluster(targetVM, link);
```
edlerd:
```suggestion
  const token = createClusterLinkOnRemoteCluster(targetVM, link);
```
Typo

tests/cluster-links-clustered.spec.ts
```diff
@@ -17,6 +19,67 @@ test("cluster link create edit delete", async ({
...
+  await createClusterLink(page, link, token);
+
+  const row = page.getByRole("row").filter({ hasText: link });
+  await expect(row.getByRole("cell", { name: "Status" })).toHaveText(
+    "Reachable",
+  );
+  await expect(row.getByRole("cell", { name: "Auth groups" })).toHaveText("1");
+
+  deleteClusterLinkonRemoteCluster(targetVM, link);
```
edlerd:
```suggestion
  deleteClusterLinkOnRemoteCluster(targetVM, link);
```

tests/server.spec.ts
```diff
@@ -87,3 +88,21 @@ test("add and delete user defined setting", async ({ page }) => {
...
+    serverPageTabs.push("Clustering");
+  }
+
+  if (lxdVersion !== "5.0-edge" && lxdVersion !== "5.21-edge") {
+    serverPageTabs.push("Replicators", "Cluster links");
+  }
+
+  for (const tab of serverPageTabs) {
+    await tabs.filter({ hasText: tab }).click();
+  }
```
edlerd, line 115: Maybe add a bit more validation here than just clicking through the tabs. Such as ensuring we are on the right tab with the correct heading. Perhaps remove the `serverPageTabs` concept as it slightly confuses. We can directly click the tabs guarded by the LXD-version checks.

edlerd (commented):
Good stuff, thanks for making it clean. A few tiny comments are left, then should be good to merge.

tests/cluster-links-clustered.spec.ts
```diff
@@ -17,6 +19,67 @@ test("cluster link create edit delete", async ({
...
+  lxdVersion,
+}, testInfo) => {
+  skipIfNotSupported(lxdVersion);
+  skipIfNotClustered(testInfo.project.name);
+
+  const link = randomLinkName();
+  const targetVM = process.env.LXD_UI_CLUSTER_LINK_TARGET_VM;
+  if (!targetVM) {
+    throw new Error("Missing required env var: LXD_UI_CLUSTER_LINK_TARGET_VM");
+  }
```
edlerd: Shall we push this into a new helper to be called by createClusterLinkOnRemoteCluster and deleteClusterLinkOnRemoteCluster?

Then any other tests that rely on existing links, can do simply two lines for setup:
```
  const token = createClusterLinkonRemoteCluster(link);
  await createClusterLink(page, link, token);
```

tests/cluster-links-clustered.spec.ts
omarelkashef: Thanks for the catch

tests/cluster-links-clustered.spec.ts
omarelkashef: These two functions already exist. Do you mean pushing the part about reading the env var into the helper function? That would mean we can only create a link to a specific node. Do we expect a need to create links to multiple nodes?

edlerd: Yes, reading the `targetVM` can be done inside those functions via extra helper. Then other tests that need cluster links don't need to worry about that aspect, and it becomes an implementation detail.

tests/helpers/cluster-links.ts
```diff
@@ -57,3 +75,27 @@ export const deleteClusterLink = async (page: Page, link: string) => {
...
+
+  // Extract token from the output (it's on the last line)
+  return output.split("\n").pop() || "";
+};
+
+export const deleteClusterLinkOnRemoteCluster = (link: string) => {
+  const targetVM = process.env.LXD_UI_CLUSTER_LINK_TARGET_VM;
+  if (!targetVM) {
+    throw new Error("Missing required env var: LXD_UI_CLUSTER_LINK_TARGET_VM");
+  }
```
edlerd, line 101: Can we extract this into a `getRemoteClusterVm` helper or similar? So we 1) don't have to repeat it and 2) it will become clear this and the function above refer to the same thing.

edlerd (commented):
Nearly perfect, tiny thing left.

We didn't add the "clear all cluster existing links if they exist" before creating one. But we can keep as is and see if CI is stable. Maybe we don't need it.

omarelkashef:
> Nearly perfect, tiny thing left.
> 
> We didn't add the "clear all cluster existing links if they exist" before creating one. But we can keep as is and see if CI is stable. Maybe we don't need it.

I agree as I am not sure it will be necessary.

edlerd (approved):
LGTM, thanks :)

PR #2055, opened 2026-06-12
feat(cluster link): disable delete button if cluster link is used

kimanhou:
## Done

- feat(cluster link): disable delete button if cluster link is used

## QA

1. Run the LXD-UI:
    - Go to cluster link list. Make sure the cluster links that are being used cannot be deleted.

## Screenshots

image

### DIFF
--- a/src/pages/cluster/actions/DeleteClusterLinkBtn.tsx
+++ b/src/pages/cluster/actions/DeleteClusterLinkBtn.tsx
@@ -49,12 +49,22 @@ const DeleteClusterLinkBtn: FC<Props> = ({ clusterLink }) => {
       });
   };
 
+  const disabledReason = () => {
+    if (clusterLink.used_by?.length) {
+      return `Cannot delete this cluster link as it is currently in use.`;
+    }
+    if (!canDelete) {
+      return "You do not have permission to delete this cluster link";
+    }
+    return undefined;
+  };
+
   return (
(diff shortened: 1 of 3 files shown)


src/pages/cluster/actions/DeleteClusterLinkBtn.tsx
```diff
@@ -49,12 +49,22 @@ const DeleteClusterLinkBtn: FC<Props> = ({ clusterLink }) => {
       });
   };
 
+  const disabledReason = () => {
+    if (clusterLink.used_by?.length) {
+      return `Cannot delete this cluster link as it is currently in use.`;
```
edlerd, line 54: This works correctly. The user is left with no guidance as to how to resolve this, though.

I could think of user stories where the target cluster has been removed or replaced. In that case, we should provide details as to what the link is used for. Solving that is probably too much scope for the current plan. Can you file a ticket to look into this later on?

edlerd (approved):
Code and QA LGTM, but there is another concern to be solved.

src/pages/cluster/actions/DeleteClusterLinkBtn.tsx
kimanhou: JIRA ticket created
https://warthogs.atlassian.net/browse/WD-37139

PR #2059, opened 2026-06-15
feat: [WD-36586] File Explorer Upload

Kxiru:
## Done

- Upload one or multiple files to the current directory.

Note: If you upload the same file twice, what should the behaviour be? (Duplicates)

## QA

1. Run the LXD-UI:
    - Navigate to a running VM or container
    - Navigate to the file explorer page
    - Click the upload button and attempt to add a new file
    - Click the upload button and attempt to add multiple files
    - Observe that the files have been added both in the file explorer and in the terminal

## Screenshots

image

### DIFF
--- a/src/pages/instances/UploadFileExplorerBtn.tsx
+++ b/src/pages/instances/UploadFileExplorerBtn.tsx
@@ -1,0 +1,134 @@
+import { type ChangeEvent, useRef, useState, type FC } from "react";
+import {
+  Button,
+  ConfirmationModal,
+  Icon,
+  usePortal,
+} from "@canonical/react-components";
+import type { LxdInstance } from "types/instance";
+import { useInstanceEntitlements } from "util/entitlements/instances";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import FileExplorerUploadModal from "./forms/FileExplorerUploadModal";
+import { pluralize } from "util/helpers";
+
+interface Props {
+  instance: LxdInstance;
+  currentPath: string;
+  refreshDirectoryList: () => void;
+  directoryContent: string[];
+}
+
+const UploadFileExplorerBtn: FC<Props> = ({
+  instance,
+  currentPath,
+  refreshDirectoryList,
+  directoryContent,
+}) => {
+  const [files, setFiles] = useState<File[]>([]);
+  const [duplicateFiles, setDuplicateFiles] = useState<string[]>([]);
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const isSmallScreen = useIsScreenBelow();
+  const { canAccessInstanceFiles } = useInstanceEntitlements();
+  const hasPermission = canAccessInstanceFiles(instance);
+
+  const handleClick = () => {
+    fileInputRef.current?.click();
+  };
+
+  const handleFileChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const selected = event.target.files;
+    if (!selected || selected.length === 0) return;
+
+    const selectedFiles = Array.from(selected);
+    setFiles(selectedFiles);
+
+    const duplicates = selectedFiles
+      .map((file) => file.name)
+      .filter((name) => directoryContent.includes(name));
+
+    if (duplicates.length > 0) {
+      setDuplicateFiles(duplicates);
+    } else {
+      openPortal(event);
+    }
+
+    // Reset the input so the same file(s) can be re-selected
+    if (fileInputRef.current) {
+      fileInputRef.current.value = "";
+    }
+  };
+
+  const handleConfirmDuplicate = (
+    event: React.MouseEvent<HTMLElement, MouseEvent>,
+  ) => {
+    setDuplicateFiles([]);
+    openPortal(event);
+  };
+
+  const handleCancelDuplicate = () => {
+    setDuplicateFiles([]);
+    setFiles([]);
+  };
+
+  const handleClose = () => {
+    setFiles([]);
+    closePortal();
+  };
+
+  const filePluralized = pluralize("file", files.length);
+
+  return (
+    <>
+      <input
+        type="file"
+        multiple
+        ref={fileInputRef}
+        onChange={handleFileChange}
+        className="u-hide"
+      />
+      <Button
+        onClick={handleClick}
+        hasIcon={!isSmallScreen}
+        disabled={!hasPermission || isOpen}
+        className="u-no-margin--bottom"
+        title={
+          hasPermission
+            ? "Upload files to current directory"
+            : "You do not have permission to manage files on this instance"
+        }
+      >
+        {!isSmallScreen && <Icon name="upload" />}
+        <span>Upload</span>
+      </Button>
+      {isOpen && (
+        <Portal>
+          <FileExplorerUploadModal
+            close={handleClose}
+            files={files}
+            instance={instance}
+            currentPath={currentPath}
+            refreshDirectoryList={refreshDirectoryList}
+          />
+        </Portal>
+      )}
+      {duplicateFiles.length > 0 && (
+        <ConfirmationModal
+          title={`Overwrite existing ${filePluralized}?`}
+          confirmButtonLabel={`Overwrite ${filePluralized}`}
+          onConfirm={handleConfirmDuplicate}
+          close={handleCancelDuplicate}
+        >
+          <p>Replace existing {filePluralized} with the same name:</p>
+          <ul>
+            {duplicateFiles.map((name) => (
+              <li key={name}>{name}</li>
+            ))}
+          </ul>
+        </ConfirmationModal>
+      )}
+    </>
+  );
+};
+
+export default UploadFileExplorerBtn;
--- a/src/pages/instances/forms/FileExplorerUploadModal.tsx
+++ b/src/pages/instances/forms/FileExplorerUploadModal.tsx
@@ -1,0 +1,142 @@
+import { useEffect, useRef, useState, type FC } from "react";
+import {
+  Button,
+  Modal,
+  useToastNotification,
+} from "@canonical/react-components";
+import type { UploadState } from "types/upload";
+import ProgressBar from "components/ProgressBar";
+import { humanFileSize } from "util/helpers";
+import type { LxdInstance } from "types/instance";
+import { deleteInstanceFile, uploadInstanceFile } from "api/instances";
+
+interface Props {
+  close: () => void;
+  files: File[];
+  instance: LxdInstance;
+  currentPath: string;
+  refreshDirectoryList: () => void;
+}
+
+const FileExplorerUploadModal: FC<Props> = ({
+  close,
+  files,
+  instance,
+  currentPath,
+  refreshDirectoryList,
+}) => {
+  const [uploadState, setUploadState] = useState<UploadState | null>(null);
+  const abortControllerRef = useRef<AbortController | null>(null);
+  const toastNotify = useToastNotification();
+  const milliseconds = 1000;
+  const [hasTimePassed, setHasTimePassed] = useState(false);
+  const [hasUploadFinished, setHasUploadFinished] = useState(false);
+  const [currentFileUploading, setCurrentFileUploading] = useState<string>();
+
+  const handleCancel = () => {
+    abortControllerRef.current?.abort();
+    close();
+  };
+
+  useEffect(() => {
+    void uploadFiles();
+  }, []);
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setHasTimePassed(true);
+      if (hasUploadFinished) {
+        close();
+      }
+    }, milliseconds);
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [hasUploadFinished]);
+
+  const uploadFiles = async () => {
+    const abortController = new AbortController();
+    abortControllerRef.current = abortController;
+
+    let successCount = 0;
+    let failError = null;
+    const uploadedPaths: string[] = [];
+
+    for (const file of files) {
+      const targetPath =
+        currentPath === "/" ? `/${file.name}` : `${currentPath}/${file.name}`;
+      setCurrentFileUploading(file.name);
+
+      try {
+        await uploadInstanceFile(
+          instance,
+          targetPath,
+          file,
+          setUploadState,
+          abortController,
+        );
+        successCount++;
+        uploadedPaths.push(targetPath);
+      } catch (e) {
+        failError = e;
+        const pathsToDelete: string[] = [];
+
+        if (abortController.signal.aborted) {
+          pathsToDelete.push(...uploadedPaths, targetPath);
+        } else {
+          pathsToDelete.push(targetPath);
+        }
+
+        await Promise.allSettled(
+          pathsToDelete.map(async (path) => deleteInstanceFile(instance, path)),
+        );
+
+        break;
+      }
+    }
+
+    const multipleFiles = successCount > 1;
+    const fileLabel = multipleFiles ? "files" : "File";
+    const message = multipleFiles
+      ? `${successCount} ${fileLabel} uploaded successfully.`
+      : `${files[0].name} uploaded successfully.`;
+
+    if (successCount === files.length) {
+      toastNotify.success(message);
+    } else {
+      toastNotify.failure(`Upload failed`, failError);
+    }
+
+    setHasUploadFinished(true);
+    refreshDirectoryList();
+    if (hasTimePassed) {
+      close();
+    }
+  };
+
+  return (
+    <Modal
+      close={handleCancel}
+      className="upload-instance-modal u-no-margin--bottom"
+      title="Uploading files"
+      closeOnOutsideClick={false}
+      buttonRow={
+        <Button className="u-no-margin--bottom" onClick={handleCancel}>
+          Cancel upload
+        </Button>
+      }
+    >
+      {uploadState && (
+        <>
+          <ProgressBar percentage={Math.floor(uploadState.percentage)} />
+          <p>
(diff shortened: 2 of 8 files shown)


src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,139 @@
...
+      </Button>
+      {uploadState && (
+        <div className="upload-file-explorer__progress">
+          <ProgressBar percentage={Math.floor(uploadState.percentage)} />
+          <span className="u-text--muted p-text--small">
+            {humanFileSize(uploadState.loaded)} of{" "}
+            {humanFileSize(uploadState.total ?? 0)}
+          </span>
+        </div>
+      )}
```
edlerd: This shouldn't be next to the button. I think after picking the file, we should open a modal and show the upload progress bar and an option to cancel inside the modal. Closing the modal should cancel the upload as well.

Currently there is no way to cancel the upload after it was started, and when navigating to another page, the progress vanishes while the upload completes in the background -- invisible to the user.

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,139 @@
...
+    <div className="upload-file-explorer">
+      <input
+        type="file"
+        multiple
+        ref={fileInputRef}
+        onChange={(e) => void handleFileChange(e)}
+        className="u-hide"
+      />
+      <Button
+        appearance="positive"
```
edlerd: This can be a standard button, no need for the primary styling.

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,139 @@
...
+          `Failed to upload ${file.name}`,
+          e instanceof Error ? e : new Error(String(e)),
+        );
+      }
+    }
+
+    if (successCount > 0) {
+      const fileLabel = successCount == 1 ? "file" : "files";
+      toastNotify.success(
+        `${successCount} ${fileLabel} ${successCount == 1 ? fileList[0].name + " " : ""}uploaded successfully.`,
```
edlerd: This should mention the instance as a rich chip. Maybe also the folder.

edlerd (commented):
UX looks promising and code mostly good, few things left for it to complete.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  };
+
+  return (
+    <Modal
+      close={handleCancel}
+      className="upload-instance-modal"
+      title="Uploading files"
+      closeOnOutsideClick={false}
+      buttonRow={
+        <Button appearance="negative" onClick={handleCancel}>
```
edlerd: The negative appearance is a bit too prominent, better use the standard button with just the border.

Image

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const handleCancel = () => {
+    abortControllerRef.current?.abort();
+    close();
+  };
+
+  useEffect(() => {
+    void uploadFiles();
+  }, []);
+
```
edlerd, line 44:
```suggestion
```
This shouldn't be a use effect. Better architecture is to call the function directly from the file selection submit event. The `uploadFiles` logic should be moved back into the button, and we keep the `uploadState` in the button as well. Then the modal receives the `uploadState`. And we expose a `cancelUpload` callback to this modal that triggers the abort controller in the button.

src/pages/instances/forms/FileExplorerUploadModal.tsx
edlerd: The button also should have no bottom margin.

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,87 @@
...
+        title={
+          hasPermission
+            ? "Upload files to current directory"
+            : "You do not have permission to manage files on this instance"
+        }
+      >
+        {!isSmallScreen && <Icon name="upload" light />}
+        <span>Upload</span>
+      </Button>
+      {isOpen && (
```
edlerd, line 104: Thanks for moving this into the modal, this is much better. There is a bit of a problem with small files, that let the modal flicker as it opens and closes within one second. Maybe we should define a minimum display time such as 3 or 5 seconds to avoid that glitch. wdyt?

edlerd (changes requested):
Thanks for the iteration, looks better already. Few new things and two previous comments are unresolved.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+          e instanceof Error ? e : new Error(String(e)),
+        );
+      }
+    }
+
+    if (successCount > 0) {
+      const fileLabel = successCount === 1 ? "file" : "files";
+      toastNotify.success(
+        `${successCount} ${fileLabel} ${successCount === 1 ? files[0].name + " " : ""}uploaded successfully.`,
+      );
```
edlerd: For the single count we should not mention the number, it should say "File foo uploaded to instance bar". and for multiple count "x files uploaded to instance bar". We can avoid "successfully" to shorten the phrase.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+      onUploadComplete();
+    }
+
+    if (failCount > 0 && successCount > 0) {
+      toastNotify.failure(
+        `${failCount} file${failCount > 1 ? "s" : ""} failed to upload.`,
```
edlerd: Same here with the singular/plural case as in the success case. Here especially we are not providing any information about which failed and which succeeded. Seems like a problem, I think a user would be left confused and we should give a bit more detail. See the bulk instance start/stop logic, where we offer a detailed breakdown at the end. Though that might be overkill here. But improving slightly would be good.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    let successCount = 0;
+    let failCount = 0;
+
+    for (const file of files) {
+      const targetPath =
+        currentPath === "/" ? `/${file.name}` : `${currentPath}/${file.name}`;
+
+      try {
+        await uploadInstanceFile(
```
omarelkashef, line 71: Is it possible to restructure the function to upload all files at the same time and not await one by one, then we check for how many succeeded?

omarelkashef (commented):
Really nice, one comment below.

src/pages/instances/forms/FileExplorerUploadModal.tsx
edlerd: I think the current approach is better. Otherwise, multiple files compete for the upload speed. Ideally, we would allow parallel uploads but cap it at 3 files max. More files would queue for the 3 parallel slots.

Though I think it's better to keep it simple as is with one file after another.

src/pages/instances/forms/FileExplorerUploadModal.tsx
Kxiru: I coped the structure from UploadInstanceFileBtn and the respctive modal. Should I still change the structure as it is currently consistent with the codebase.

src/pages/instances/UploadFileExplorerBtn.tsx
Kxiru: This doesn't happen to me locally, I had to physically find a large file so I could try and see the modal :sob: , but how would this work? And if it takes 3-5 seconds on average to upload something, is that too long? Maybe one second?

src/pages/instances/forms/FileExplorerUploadModal.tsx
Kxiru: Actioned the first part, but surely "successfully" needs to stay?

src/pages/instances/UploadFileExplorerBtn.tsx
edlerd: Two states, one for canCloseTimeout and one canCloseEvent. 
A set timeout that starts on open of the modal. It enables one of the two states after 3 seconds. We do flip the other state based on the upload completed event.

In both handlers we check if the other state is true and close if yes.

src/pages/instances/forms/FileExplorerUploadModal.tsx
edlerd: Yes, please change it here to simplify.

edlerd (changes requested):

src/pages/instances/forms/FileExplorerUploadModal.tsx
edlerd: It is a bit too much "success" in this message:

image

Let's use the proposed copy above.

Kxiru: Haha, sure

Kxiru: The propsed copy is the same as before. Please confirm:
`${multipleFiles ? successCount : ""} ${fileLabel} ${multipleFiles ? " " : files[0].name + " "}uploaded successfully.`,

omarelkashef (commented):
There is a situation where duplicate file show progress and notification but visually it is confusing because I am assuming they secretly overwrite the old one. If that is the intended behavior, maybe we can clarify it before the user uploads. If not, we can add -copy at the end of the file name before uploading. Or more easily, we tell the user that any uploads will overwrite files with the same name. Another thing is that when I select the file twice in a row, I do not get any feedback (progress bar or notification), but when I select file A then B then A again, it is fine (see the video for better explanation) 

https://github.com/user-attachments/assets/cc7e7b93-1812-4197-9a76-01a8d314fa32

edlerd:
> There is a situation where duplicate file show progress and notification but visually it is confusing because I am assuming they secretly overwrite the old one.

Yes, overwriting silently is what is happening indeed. Let's add a confirmation in that case. So before uploading, check if the file already exists and show a modal explaining "...file already exists, do you want to overwrite? >> Cancel / Upload and overwrite". If the file is new, upload it right away and skip the confirmation.

The current behavior of silently overwriting seems unexpected. Changing the name silently to append "-copy" also seems surprising and not intended.

> Another thing is that when I select the file twice in a row, I do not get any feedback (progress bar or notification), but when I select file A then B then A again, it is fine (see the video for better explanation)

This seems like a bug and should be fixed.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const timer = setTimeout(() => {
+      setHasTimePassed(true);
+      if (hasUploadFinished) {
+        close();
+      }
+    }, milliseconds);
+    return () => {
+      clearTimeout(timer);
+    };
+  }, []);
```
edlerd:
```suggestion
  }, [hasUploadFinished]);
```
This is currently broken and the upload modal does not close with small files.

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,139 @@
...
+          </p>
+          <ul>
+            {duplicateFiles.map((name) => (
+              <li key={name}>{name}</li>
+            ))}
+          </ul>
+          <p className="u-text--muted">
+            {duplicateFiles.length === 1 ? "This file" : "These files"} will be
+            overwritten.
+          </p>
```
edlerd:
```suggestion
```
Better use simple single line of copy (see other comment)

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,139 @@
...
+          confirmButtonLabel="Upload"
+          onConfirm={handleConfirmDuplicate}
+          close={handleCancelDuplicate}
+        >
+          <p>
+            The following{" "}
+            {duplicateFiles.length === 1
+              ? "file already exists"
+              : "files already exist"}{" "}
+            in this directory:
```
edlerd:
```suggestion
            Uploading these files will replace existing files with the same name.
```
Slightly simplifying the copy here. We need a pluralize as well.

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,139 @@
...
+            instance={instance}
+            currentPath={currentPath}
+            onUploadComplete={onUploadComplete}
+          />
+        </Portal>
+      )}
+      {duplicateFiles.length > 0 && (
+        <ConfirmationModal
+          title="Confirm file upload"
+          confirmButtonLabel="Upload"
```
edlerd:
```suggestion
          confirmButtonLabel="Overwrite file"
```
We should use pluralize on file here based on duplicate length.

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,139 @@
...
+            files={files}
+            instance={instance}
+            currentPath={currentPath}
+            onUploadComplete={onUploadComplete}
+          />
+        </Portal>
+      )}
+      {duplicateFiles.length > 0 && (
+        <ConfirmationModal
+          title="Confirm file upload"
```
edlerd:
```suggestion
          title="Overwrite existing files?"
```
This needs pluralize as well.

edlerd (changes requested):
The concern above raised by @omarelkashef of uploading the same file twice in a row is not addressed. When chosing the same file right after uploading it a second time, nothing happens.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+        `${multipleFiles ? successCount : ""} ${fileLabel} ${multipleFiles ? " " : files[0].name + " "}uploaded successfully.`,
+      );
+      onUploadComplete();
+    }
+    if (failCount > 0 && successCount > 0) {
+      toastNotify.failure(
+        `${multipleFiles ? failCount : ""} ${fileLabel} failed to upload.`,
+        new Error("Some files could not be uploaded"),
+      );
+    }
```
edlerd: The logic here is not sound. If both failcount and successcount are bigger than zero, we will trigger the success and then a failure notification, which seems wrong. If only the fail count is >0 and the success count is 0, then no notification will be shown.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+        );
+      }
+    }
+
+    const multipleFiles = successCount > 1;
+    const fileLabel = multipleFiles ? "files" : "File";
+
+    if (successCount > 0) {
+      toastNotify.success(
+        `${multipleFiles ? successCount : ""} ${fileLabel} ${multipleFiles ? " " : files[0].name + " "}uploaded successfully.`,
```
edlerd: Those are too many ternaries in one line, lets create a `const message = multipleFiles ? "... message 1" : ".. message 2"` to make this simpler.

src/pages/instances/UploadFileExplorerBtn.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {duplicateFiles.length > 0 && (
+        <ConfirmationModal
+          title={`Overwrite existing ${filePluralized}?`}
+          confirmButtonLabel={`Overwrite ${filePluralized}`}
+          onConfirm={handleConfirmDuplicate}
+          close={handleCancelDuplicate}
+        >
+          <p>
+            Uploading the following {filePluralized} will replace existing files
+            with the same name:
```
edlerd:
```suggestion
            Replace existing files with the same name:
```
A bit simpler.

We also need a pluralize on file/files here.

edlerd (changes requested):
Good progress, few issues left to resolve.

src/pages/instances/forms/FileExplorerUploadModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    if (successCount === files.length) {
+      toastNotify.success(message);
+      onUploadComplete();
+    } else if (successCount > 0) {
+      toastNotify.failure(
+        `${failCount} ${fileLabel} failed to upload.`,
+        new Error("Some files could not be uploaded"),
+      );
+      onUploadComplete();
```
edlerd:
```suggestion
    } else if (successCount > 0) {
      toastNotify.failure(
        `${failCount} ${fileLabel} failed to upload.`,
        new Error("Some files could not be uploaded"),
      );
```
When I upload a big file and cancel it mid way, this leaves behind the half uploaded file. So we must clear the cache on failure as well. Let's move this call out from the conditionals. I would also recommend renaming it from `onUploadComplete` to `refreshDirectoryList` or similar. Also rename it in the upstream component.

I wonder if we should do more, we could trigger a delete request to not leave behind the half empty file. But I am not 100% sure what best matches the user expectation here. Maybe discuss with others.

edlerd (commented):
Code and QA looks good to me, good work.

There is one edge case left that is not ideally handled. See details below.

src/pages/instances/forms/FileExplorerUploadModal.tsx
edlerd: Or we can change the error message to say "file was half uploaded", though that seems odd. I want to avoid a user uploading a file, the upload fails half way, then they see the file in the directory and wrongly assume the upload succeeded and ignore the error.

Kxiru: That's such an interesting edge case... Will make the initial amendments.

edlerd (approved):
LGTM, great result 👍

Kxiru:
> LGTM, great result 👍

Thanks a bunch!

PR #2075, opened 2026-06-18
chore(test) fix cluster test to rely on 3 nodes

edlerd:
## Done

- chore(test) fix cluster test to rely on 3 nodes


kimanhou:
Is there something wrong with local peering setup ?

image

edlerd:
> Is there something wrong with local peering setup ?

It seems unrelated, as it is only in one of the two browsers.

kimanhou (approved):

PR #2079, opened 2026-06-19
feat(load-balancer) guard load balancer health checks with an api extension

edlerd:
## Done

- feat(load-balancer) guard load balancer health checks with an api extension

## QA

1. Run the LXD-UI:
    - check an ovn network load balancer pool creation is not exposing health check configuration:
    - use a lxd latest edge backend with ovn configured:
      - sudo snap install microovn --channel=latest/edge
      -  microovn init (create a new cluster)
      -  lxd init (let it listen on the network)
      -  set lxd config to network.ovn.northbound_connection: ssl:127.0.0.1:6641
      -  connect the ui from this branch to the container as a backend
      -  In the ui configure the lxdbr0 created by lxd init above to have:
      -  IPv4 DHCP ranges: With about 20 ips, such as 10.115.100.170-10.115.100.190, needs to match the network mask
      -  IPv4 OVN ranges: With about 50 ips, such as 10.115.100.100-10.115.100.150, needs to match the network mask
      -  IPV4 routes: Set to a valid CIDR, such as 10.115.100.208/28
      -  in the ui create an ovn network
      -  use lxdbr0 as an uplink, other settings default
      -  in the ui visit ovn network detail page
      -  in the ui go to load balancers
      - in the ui create a load balancer pool

## Screenshots

image

### DIFF
--- a/src/context/useSupportedFeatures.tsx
+++ b/src/context/useSupportedFeatures.tsx
@@ -62,5 +62,8 @@ export const useSupportedFeatures = () => {
     hasReplicators: apiExtensions.has("replicators"),
     hasLoadBalancerPools: apiExtensions.has("network_load_balancer_pool"),
     hasStorageNvmeTcp: apiExtensions.has("storage_nvme_tcp"),
+    hasLoadBalancerHealthChecks: apiExtensions.has(
+      "network_load_balancer_pool_health_checks",
+    ),
   };
 };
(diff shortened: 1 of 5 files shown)


omarelkashef (commented):
Is the health check disabled or default when the api extension does not exist? Currently we show default. Maybe we should not even show that column in the load balancer pool table.

edlerd:
> Is the health check disabled or default when the api extension does not exist? Currently we show default. Maybe we should not even show that column in the load balancer pool table.

Ah yes, I missed hiding the column in the load balancer pools table, will add it shortly.

omarelkashef (dismissed):
LGTM, thank you! One non-blocking suggestion: should empty load balancer pool instance status be "-" instead of an empty string?

omarelkashef (approved):

PR #2081, opened 2026-06-19
test(file explorer): e2e tests

omarelkashef:
## Done

- Add instance file explorer e2e tests WD-36644

## QA

1. Run the LXD-UI:
    - Ensure new tests pass

### DIFF
--- a/tests/instances.spec.ts
+++ b/tests/instances.spec.ts
@@ -15,6 +15,18 @@ import {
   visitInstance,
 } from "./helpers/instances";
 import {
+  visitFileExplorer,
+  randomFileName,
+  openDirectory,
+  assertFileExists,
+  assertFileNotExists,
+  assertDirectoryExists,
+  deleteFile,
+  downloadFile,
+  uploadFile,
+  createFile,
+} from "./helpers/instance-file-explorer";
+import {
   assertCode,
   assertReadMode,
   setCodeInput,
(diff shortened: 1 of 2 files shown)


tests/helpers/instances.ts
```diff
@@ -238,3 +239,123 @@ export const migrateInstanceRootStorage = async (
...
+  project = "default",
+) => {
+  await visitInstance(page, instance, project);
+  await page.getByTestId("tab-link-File Explorer").click();
+  await expect(page.getByText("File explorer")).toBeVisible();
+};
+
+export const navigateToDirectory = async (
+  page: Page,
+  directoryPath: string,
```
Kxiru: As long as the type/image of the instane stays the same throughout every iteration of the test, passing along a hardcoded path should be fine. But some OS/images may have different file structures, so let's not pass through something deep.

tests/helpers/instances.ts
```diff
@@ -238,3 +239,123 @@ export const migrateInstanceRootStorage = async (
...
+    page,
+    new RegExp("^File\\s+.+\\s+uploaded successfully$"),
+  );
+};
+
+export const createFile = async (page: Page, fileName: string) => {
+  // Use terminal to create files for now. Later we might add a file creation feature in the file explorer.
+  await page.getByTestId("tab-link-Terminal").click();
+  await assertTextVisible(page, "~#");
+  await page.waitForTimeout(1000); // ensure the terminal is ready
```
Kxiru: I know this tests terminal creation of files, but let's assert visibility of the new file in the explorer, not the terminal.

tests/helpers/instances.ts
omarelkashef: Can you elaborate further? I do not follow.

tests/helpers/instances.ts
omarelkashef: I use the file explorer to assert the visibility of the file indeed. Check `assertFileExists`. I only use the terminal to create a new file until the create file is supported in the UI.

tests/helpers/instances.ts
```diff
@@ -238,3 +239,131 @@ export const migrateInstanceRootStorage = async (
...
+  return download;
+};
+
+export const uploadFile = async (page: Page, filePath: string) => {
+  const uploadButton = page.getByTitle("Upload files to current directory");
+  await uploadButton.click();
+
+  // Bypass the OS-native file picker dialog by directly setting the file input
+  await page.locator('input[type="file"]').setInputFiles(filePath);
+  await dismissNotification(page, new RegExp(".* uploaded successfully"));
```
edlerd:
```suggestion
  await dismissNotification(page, (" uploaded successfully"));
```
Let's use the full correctly expected message here, not some regexp matching.

tests/helpers/instances.ts
```diff
@@ -238,3 +239,131 @@ export const migrateInstanceRootStorage = async (
...
+  page: Page,
+  instance: string,
+  project = "default",
+) => {
+  await visitInstance(page, instance, project);
+  await page.getByTestId("tab-link-File Explorer").click();
+  await expect(page.getByText("Directory: root")).toBeVisible();
+};
+
+export const navigateToDirectory = async (
```
edlerd: I don't think we need this helper, it is a bit complex. I think a simpler way is to have this accept one directory and then call it multiple times:
```
openDirectory("home");
openDirectory("ubuntu");
```

tests/helpers/notification.ts
```diff
@@ -10,7 +10,10 @@ export const checkNotificationHidden = async (page: Page) => {
   await expect(notification).toBeHidden();
 };
 
-export const dismissNotification = async (page: Page, caption: string) => {
+export const dismissNotification = async (
+  page: Page,
+  caption: string | RegExp,
```
edlerd:
```suggestion
  caption: string,
```
We should be strict on the assertions to include everything from the message.

tests/instances.spec.ts
```diff
@@ -452,3 +461,75 @@ test("instance deletion (with and without force delete support)", async ({
...
+test("upload, download, and delete file from file explorer", async ({
+  page,
+}, testInfo) => {
+  await visitAndStartInstance(page, instance);
+  await visitFileExplorer(page, instance);
+
+  await navigateToDirectory(page, "/tmp");
+
+  const createdFileName = "test-file.txt";
+  const uploadedFileName = "uploaded-test-file.txt";
```
edlerd: Do we need to randomize those names?

edlerd (commented):
good start, few thoughts below.

tests/helpers/instances.ts
Kxiru: No action item here! Just saying that the instance type should stay consistent to avoid issues with inputting custom directory paths - which it already is.

tests/helpers/instances.ts
```diff
@@ -238,3 +243,136 @@ export const migrateInstanceRootStorage = async (
     `Instance ${instance} root storage successfully moved to pool ${pool}.`,
   );
 };
+
+// File Explorer Helpers
```
edlerd:
```suggestion
```
Better move those into a dedicated file.

tests/helpers/instances.ts
```diff
@@ -238,3 +243,136 @@ export const migrateInstanceRootStorage = async (
...
+    .filter({ hasText: type });
+};
+
+export const visitFileExplorer = async (
+  page: Page,
+  instance: string,
+  project = "default",
+) => {
+  await visitInstance(page, instance, project);
+  await page.getByTestId("tab-link-File Explorer").click();
```
edlerd: Let's avoid using test id and use the actual label to test also the experience of screen readers and accessibility.

tests/helpers/instances.ts
```diff
@@ -238,3 +243,136 @@ export const migrateInstanceRootStorage = async (
...
+  await expect(dirLink).toHaveCount(1);
+  await dirLink.click();
+
+  const breadcrumb = page.getByRole("navigation", {
+    name: "File Explorer Path",
+  });
+  await expect(
+    breadcrumb.getByText(directoryName, { exact: true }),
+  ).toBeVisible();
+  await expect(page.getByRole("grid").first()).toBeVisible();
```
edlerd:
```suggestion
```
This is unclear to me. What does this do? Can it be removed?

tests/helpers/instances.ts
```diff
@@ -238,3 +243,136 @@ export const migrateInstanceRootStorage = async (
...
+  await expect(dirLink).toHaveCount(1);
+};
+
+export const deleteFile = async (
+  page: Page,
+  fileName: string,
+  instnace: string,
+) => {
+  const fileRow = getFileExplorerRow(page, fileName, "file");
+  await expect(fileRow).toBeVisible();
```
edlerd:
```suggestion
```
We don't need this, clicking the delete button in the row ensures this implicitly. Playwright will fail if you try to make it click invisible elements.

tests/helpers/instances.ts
```diff
@@ -238,3 +243,136 @@ export const migrateInstanceRootStorage = async (
...
+    name: directoryName,
+    exact: true,
+  });
+  await expect(dirLink).toHaveCount(1);
+};
+
+export const deleteFile = async (
+  page: Page,
+  fileName: string,
+  instnace: string,
```
edlerd:
```suggestion
  instance: string,
```
typo

tests/helpers/instances.ts
```diff
@@ -238,3 +243,136 @@ export const migrateInstanceRootStorage = async (
...
+  const uploadButton = page.getByTitle("Upload files to current directory");
+  await uploadButton.click();
+
+  // Bypass the OS-native file picker dialog by directly setting the file input
+  await page.locator('input[type="file"]').setInputFiles(filePath);
+  await dismissNotification(page, " uploaded successfully");
+};
+
+export const createFile = async (page: Page, fileName: string) => {
+  // Use terminal to create files for now. Later we might add a file creation feature in the file explorer.
```
edlerd: The create file feature is already in an open PR. Should we keep this PR in draft until that lands so we can have a more straightforward way of creating files?

edlerd (commented):
Mostly lgtm, few comments on the proposed changes.

tests/helpers/instances.ts
omarelkashef: The opened PR #2101  seems to only allow adding new directories but not files.

edlerd (commented):
Code looks good, but CI is failing so the test is not sound yet.

edlerd (approved):
LGTM

PR #2090, opened 2026-06-23
fix(selectable-table) inline multiselect checkbox with dropdown toggle

edlerd:
## Done

- fix(selectable-table) inline multiselect checkbox with dropdown toggle

## QA

1. Run the LXD-UI:
    - open instance and identities list and ensure the select all checkbox and dropdown toggle are in one line

## Screenshots

image

### DIFF
--- a/src/components/SelectableMainTable.tsx
+++ b/src/components/SelectableMainTable.tsx
@@ -93,7 +93,7 @@ const SelectableMainTable: FC<Props> = ({
         <>
           <CheckboxInput
             label={<div className="u-off-screen">Select all</div>}
-            labelClassName="multiselect-checkbox"
+            className="multiselect-checkbox"
             checked={isAllSelected}
             indeterminate={isSomeSelected && !isAllSelected}
             onChange={isSomeSelected ? selectNone : selectPage}
--- a/src/pages/storage/StorageVolumes.tsx
+++ b/src/pages/storage/StorageVolumes.tsx
@@ -134,15 +134,7 @@ const StorageVolumes: FC = () => {
         ]
       : []),
     {
-      content: isSmallScreen ? (
-        <>
-          {TYPE_COL}
-          <br />
-          <div className="header-second-row">{CONTENT_TYPE_COL}</div>
-        </>
-      ) : (
-        TYPE_COL
-      ),
+      content: TYPE_COL,
       sortKey: isSmallScreen ? "contentType" : "type",
       style: {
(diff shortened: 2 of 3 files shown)


omarelkashef (commented):
Instances and identities are fixed, but volumes still look off.

Image

src/components/SelectableMainTable.tsx
```diff
@@ -94,6 +94,7 @@ const SelectableMainTable: FC<Props> = ({
           <CheckboxInput
             label={<div className="u-off-screen">Select all</div>}
             labelClassName="multiselect-checkbox"
+            className="multiselect-checkbox"
```
kimanhou, line 96: Do we need both labelClassName and className ?

edlerd: We can indeed remove the `labelClassName`.

edlerd:
> Instances and identities are fixed, but volumes still look off.
> 
> Image

Not sure what do you mean @omarelkashef ? The multiselect seems all right to me.

omarelkashef:
> > Instances and identities are fixed, but volumes still look off.
> > Image
> 
> Not sure what do you mean @omarelkashef ? The multiselect seems all right to me.



> > Instances and identities are fixed, but volumes still look off.
> > Image
> 
> Not sure what do you mean @omarelkashef ? The multiselect seems all right to me.

Notice the instances table for example how the select all input is aligned with the rest of the row, but in volumes it is not

image

image

edlerd:
Adressed alignment of the multiselect header on volumes page as well. good catch, @omarelkashef

omarelkashef (commented):
The multiselect looks perfect now. But I found another bug in the storage volumes table that we can fix here. On small screens, the "content type" column name changes to just "type" but it "content type" is still there.  

Image

edlerd:
> The multiselect looks perfect now. But I found another bug in the storage volumes table that we can fix here. On small screens, the "content type" column name changes to just "type" but it "content type" is still there.

The idea was that the content-type column merges into the type column on small screens and the heading should reflect that. I agree though this is more confusing than helping and simplified it to only mention "Type" in small screens.

omarelkashef (approved):
Thank you for fixing the tables! LGTM.

omarelkashef:
> > The multiselect looks perfect now. But I found another bug in the storage volumes table that we can fix here. On small screens, the "content type" column name changes to just "type" but it "content type" is still there.
> 
> The idea was that the content-type column merges into the type column on small screens and the heading should reflect that. I agree though this is more confusing than helping and simplified it to only mention "Type" in small screens.

It looks good now, previously it was confusing.

PR #2084, opened 2026-06-22
fix(replicator): force flag for promote and demote project

kimanhou:
## Done

- feat(replicators): add `force` flag for promoting and demoting project
Note: no `force` flag for clearing the replica mode.
- fix(ConfirmationButton): align confirmExtra checkbox with buttons in footer

## QA

1. Run the LXD-UI:
    - On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @Kxiru or @edlerd for access.
    - With a local copy of this branch, [build and run as described in the docs](https://github.com/canonical/lxd-ui/blob/main/CONTRIBUTING.md#setting-up-for-development).
2. Perform the following QA steps:
    - Go to a project's configuration. Open Replication tab.
    - Click on Promote or Demote depending on your project's replica mode.
    - Make sure a confirmation modal opens.
    - Make sure the force checkbox works.

## Screenshots

image

image

PR #2091, opened 2026-06-23
test(replicator): add project replication config e2e tests and remove unnecessary replicator validation and snapshot config

omarelkashef:
## Done

- add project replication config e2e tests WD-36615
- remove unnecessary replicator edit validation WD-37322
- remove replicator snapshot config WD-37607

## QA

1. Run the LXD-UI:
    - Ensure CI passes.


tests/helpers/projects.ts
```diff
@@ -90,3 +90,46 @@ export const deleteProject = async (page: Page, project: string) => {
 
   await confirmDelete(page, project);
 };
+
+export const openProjectReplicationTab = async (page: Page) => {
+  await page.getByText("Replication").click();
+};
+
+export const promoteProjectToLeader = async (page: Page) => {
```
kimanhou: Should this function and the following one include the project name in their arguments ? So that we can dismiss the notification `Project ${project} demoted to standby.`
It allowed me to realize I wasn't demoting/promoting the right project while developing.

tests/helpers/projects.ts
```diff
@@ -90,3 +90,46 @@ export const deleteProject = async (page: Page, project: string) => {
...
+};
+
+export const assertProjectReplicaMode = async (
+  page: Page,
+  expectedMode: string,
+) => {
+  await expect(page.locator('output[id="replica_mode"]')).toContainText(
+    expectedMode,
+  );
+};
```
kimanhou:
```suggestion
  await expect(page.getByRole("status", { name: "Replica mode" })).toHaveText(
    expectedMode,
  );
```

We should avoid locator if we can. Wdyt ?

tests/helpers/projects.ts
```diff
@@ -90,3 +90,46 @@ export const deleteProject = async (page: Page, project: string) => {
...
+  expectedCluster: string,
+) => {
+  await expect(page.getByLabel("Replica cluster")).toContainText(
+    expectedCluster,
+  );
+};
+
+export const saveProjectConfiguration = async (page: Page) => {
+  await page.getByRole("button").filter({ hasText: "Save" }).click();
+  await dismissNotification(page, "Project updated.");
```
kimanhou: Same remarks as above. It would be better to target the notification with `Project ${project} updated.`

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
+  test.skip(
+    lxdVersion === "5.0-edge" || lxdVersion === "5.21-edge",
+    "Replicators tests not supported for lxd 5.0 and 5.21",
+  );
+};
+
+export const randomReplicatorName = (): string => {
+  return `playwright-replicator-${Math.random().toString(36).substring(7)}`;
```
kimanhou: Do you want to use `randomNameSuffix` ?

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +63,108 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  // Demote to standby
+  await demoteProjectToStandby(page);
+  await assertProjectReplicaMode(page, "standby");
+  await expect(
+    page.getByText("This project is a read-only failover target"),
+  ).toBeVisible();
+
+  // Promote back to leader
+  await promoteProjectToLeader(page);
+  await assertProjectReplicaMode(page, "leader");
```
kimanhou: Do you want to also test clear replica mode ?

tests/helpers/projects.ts
```diff
@@ -90,3 +90,46 @@ export const deleteProject = async (page: Page, project: string) => {
...
+
+export const promoteProjectToLeader = async (page: Page, project: string) => {
+  await page.getByRole("button", { name: "Promote to leader" }).click();
+  await dismissNotification(page, `Project ${project} promoted to leader`);
+};
+
+export const demoteProjectToStandby = async (page: Page, project: string) => {
+  await page.getByRole("button", { name: "Demote to standby" }).click();
+  await dismissNotification(page, `Project ${project} demoted to standby`);
+};
```
edlerd, line 120: Those two need adjustments from https://github.com/canonical/lxd-ui/pull/2084 introducing a confirmation stage.

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,136 @@
+import { type LxdVersions, test, expect } from "../fixtures/lxd-test";
+import type { Page } from "@playwright/test";
+import { dismissNotification } from "./notification";
+import { randomNameSuffix } from "./name";
+
+export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
+  test.skip(
+    lxdVersion === "5.0-edge" || lxdVersion === "5.21-edge",
+    "Replicators tests not supported for lxd 5.0 and 5.21",
```
edlerd:
```suggestion
    "Replicators not supported for lxd 5.0 and 5.21",
```
nitpick

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,136 @@
...
+    const snapshotCheckbox = sidePanel.getByLabel("Snapshot");
+    await snapshotCheckbox.click();
+    await page.getByRole("option", { name: newSnapshot }).click();
+  }
+
+  await sidePanel.getByRole("button", { name: "Update" }).click();
+  await dismissNotification(page, `Replicator ${replicatorName} updated.`);
+};
+
+export const deleteReplicatorRow = async (
```
edlerd, line 62:
```suggestion
export const deleteReplicator = async (
```

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,136 @@
...
+  await sidePanel.getByRole("button", { name: "Create" }).click();
+  await dismissNotification(page, `Replicator ${replicatorName} created.`);
+};
+
+export const editReplicatorSidePanel = async (
+  page: Page,
+  replicatorName: string,
+  newDescription: string,
+  newSchedule?: string,
+  newSnapshot?: "true" | "false",
```
edlerd:
```suggestion
```
Can we remove the unused options?

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,136 @@
...
+) => {
+  const sidePanel = page.getByLabel("Side panel");
+  if (newDescription) {
+    await sidePanel.getByPlaceholder("Enter description").fill(newDescription);
+  }
```
edlerd:
```suggestion
    await sidePanel.getByPlaceholder("Enter description").fill(newDescription);
```
We don't need the guard here, the description is mandatory.

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  const dialog = page.getByRole("dialog", { name: "Confirm delete" });
+  const confirmButton = dialog.getByRole("button", {
+    name: "Delete replicator",
+  });
+  await confirmButton.click();
+  await dismissNotification(page, `Replicator ${replicatorName} deleted.`);
+};
+
+export const openReplicatorRowRunModal = async (
```
edlerd:
```suggestion
export const openReplicatorRunModal = async (
```

tests/helpers/projects.ts
```diff
@@ -90,3 +90,46 @@ export const deleteProject = async (page: Page, project: string) => {
 
   await confirmDelete(page, project);
 };
+
+export const openProjectReplicationTab = async (page: Page) => {
+  await page.getByText("Replication").click();
+};
```
edlerd:
```suggestion
```
Probably simpler without this indirection.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,114 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+
+  // Promote to leader
+  await promoteProjectToLeader(page, project);
+  await assertProjectReplicaMode(page, "leader");
+  await expect(
+    page.getByText("This project is the active source project for replication"),
+  ).toBeVisible();
+
+  const clusterLinkName = randomLinkName();
+  const token = createClusterLinkOnRemoteCluster(clusterLinkName);
```
edlerd: We need to delete this as well at the end.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,114 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  skipIfReplicatorsNotSupported(lxdVersion);
+  skipIfNotClustered(testInfo.project.name);
+
+  const project = randomProjectName();
+  await createProject(page, project);
+  await openProjectConfiguration(page);
+  await openProjectReplicationTab(page);
+
+  const clusterLinkName = randomLinkName();
+  const token = createClusterLinkOnRemoteCluster(clusterLinkName);
```
edlerd: this needs cleanup as well.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,114 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  await expect(
+    page.getByRole("row").filter({ hasText: "Updated description" }),
+  ).toBeVisible();
+
+  await openReplicatorRowRunModal(page, replicatorName);
+  const dialog = page.getByRole("dialog", { name: "Run replicator" });
+
+  // Leader pre-flight checks
+  await assertLeaderPreflightChecks(page, project, clusterLinkName);
+  await dialog.getByRole("button", { name: "Cancel" }).click();
```
edlerd: Shall we submit here and await the ".. started" and "...completed" notifications?

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,114 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
   await assertReadMode(page, "Direct cluster targeting", "Allow");
   await deleteProject(page, project);
 });
+
+// Project replication
+
+test("replica mode transitions: promotion and demotion", async ({
```
edlerd:
```suggestion
```
We probably don't need this as an extra test case. The one below is already promting / demoting and we can add the assertions there.

edlerd (commented):
Good start, few thoughts on the proposed tests below.

tests/helpers/replicators.ts
omarelkashef: This depends on how we set up the rest of the replicator tests. If we want to delete from inside the detail page, we would have a helper `deleteReplicator` so I wanted to make this different `deleteReplicatorRow` to indicate that it takes a replicator name and looks it up in a table (in project config page or replicators list) then deletes it. Similar thing for the run.

tests/helpers/replicators.ts
omarelkashef: Will they be needed for the other tests? @kimanhou

tests/projects-clustered.spec.ts
omarelkashef: The actual running will be in the other tests. i do not want to test it twice to avoid increasing the run time. Can you confirm that your PR will test the replicator run? @kimanhou

tests/projects-clustered.spec.ts
omarelkashef: I thought about that but when both tests are merged, it becomes quite long. wdyt?

tests/helpers/replicators.ts
kimanhou: Yes, they should but I can add them later.

tests/projects-clustered.spec.ts
kimanhou: Yes I confirm

tests/scripts/start_vm_cluster.sh
```diff
@@ -98,7 +98,8 @@ cat keys/lxd-ui.crt | sudo lxc exec vm1 -- lxc config trust add - || true
 
 cat <<EOF > .env.local
 LXD_UI_BACKEND_IP=$VM1_IP
-LXD_UI_CLUSTER_LINK_TARGET_VM=vm-standalone
+LXD_UI_CLUSTER_LINK_LOCAL_VM=vm1
```
kimanhou, line 101: Do we really need this ? We don't need to run the commands inside `lxd exec` if they are to be ran locally, do we ?

omarelkashef: In these tests we have a source and target clusters. When we clean up cluster links for example, we need to remove both of its ends so we have to `lxc exec` into both VMs.

omarelkashef: Maybe the name local is confusing. This variable is for the backend that the tests are ran again and is used as one end of cluster links, replicators, etc. Do you have suggestions for a clearer name?

kimanhou: This variable is only used in `lxd exec vm1 -- whatever cli`, right ? Can we just run `whatever cli` without `lxc exec` ?

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,241 @@
...
+    "Deleting all existing cluster links on the current cluster before test",
+  );
+  runCommand(`lxc exec ${localVm} -- sh -c '${deleteAllClusterLinksCommand}'`);
+  console.log(
+    "Deleting all existing cluster links on the remote cluster before test",
+  );
+  runCommand(`lxc exec ${remoteVm} -- sh -c '${deleteAllClusterLinksCommand}'`);
+};
+
+export const getDeleteAllReplicatorsCommand = (): string => {
```
kimanhou: export is not needed here as this function is not used outside of its file. Creating a function to return a constant is overkill, this can just be a constant. Finally this is declared after being used, it works but it's better to move it up the file, before it's called.

Suggestion ⬇️
```
const DELETE_ALL_REPLICATORS_COMMAND = 
  "for p in $(lxc project list --format csv | tail -n +1 | cut -d, -f1); do " +
  "lxc replicator list --project=$p --format csv | tail -n +1 | cut -d, -f1 | " +
  "xargs -r -n1 lxc replicator delete --project=$p; " +
  "done";
```

tests/helpers/cluster-links.ts
```diff
@@ -78,11 +77,15 @@ export const deleteClusterLink = async (page: Page, link: string) => {
   await dismissNotification(page, `Cluster link ${link} deleted.`);
 };
 
+export const getDeleteAllClusterLinksCommand = (): string => {
+  return "lxc cluster link list --format csv | tail -n +1 | cut -d, -f1 | xargs -r -n1 lxc cluster link delete";
+};
+
```
kimanhou:
```suggestion
export const DELETE_ALL_CLUSTER_LINKS_COMMAND = 
  "lxc cluster link list --format csv | tail -n +1 | cut -d, -f1 | xargs -r -n1 lxc cluster link delete";
```

tests/helpers/shell.ts
```diff
@@ -2,7 +2,7 @@ import { execSync } from "child_process";
 
 export const runCommand = (command: string) => {
   console.log("Running command: ", command);
-  const result = execSync(`sudo -E ${command}`).toString();
```
kimanhou, line 5: Was `sudo` not necessary ?

tests/scripts/start_vm_cluster.sh
kimanhou: As discussed, `lxd exec vm1` is needed.
In `tests/helpers/auth`, we have `getExecPrefix`. Can it be reused ?

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +67,88 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
   await assertReadMode(page, "Direct cluster targeting", "Allow");
   await deleteProject(page, project);
 });
+
+// Project replication
+
```
kimanhou: Nitpick: can the empty line be removed ?

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,241 @@
...
+  ).toBeVisible();
+  await expect(
+    page.getByText("Only outgoing replicators are shown here"),
+  ).toBeVisible();
+  await expect(
+    page.getByText("No replication configuration applied."),
+  ).toBeVisible();
+};
+
+export const setupReplicaClusterLink = async (
```
kimanhou: Naming is not ideal. I didn't expect the cluster link to be created in this function. Maybe `createAndConfigureReplicaClusterLink` even tho I admit it's longer.

tests/helpers/projects.ts
```diff
@@ -90,3 +90,52 @@ export const deleteProject = async (page: Page, project: string) => {
...
+export const assertProjectReplicaMode = async (
+  page: Page,
+  expectedMode: string,
+) => {
+  await expect(
+    page.getByRole("status", { name: "Replica mode" }),
+  ).toContainText(expectedMode);
+};
+
+export const assertProjectReplicaCluster = async (
```
kimanhou: Can this be included in `setupReplicaClusterLink` (or whatever it will be renamed) ? I am not sure this will be re-used elsewhere.

kimanhou (commented):
Partial review, will continue later

tests/helpers/shell.ts
omarelkashef: That was removed by mistake, good catch!

tests/scripts/start_vm_cluster.sh
omarelkashef: `getExecPrefix` hardcodes vm1. Do you want to change it to read LXD_UI_CLUSTER_LINK_LOCAL_VM?

tests/helpers/replicators.ts
omarelkashef: I think set up implies create and configure.

tests/scripts/start_vm_cluster.sh
kimanhou: Yes, it's probably best to replace hardcoded occurrences of `vm1` by `LXD_UI_CLUSTER_LINK_LOCAL_VM` 🙏

tests/helpers/cluster-links.ts
```diff
@@ -92,14 +94,22 @@ export const createClusterLinkOnRemoteCluster = (link: string) => {
 };
 
 export const deleteClusterLinkOnRemoteCluster = (link: string) => {
-  const targetVM = getRemoteClusterVm();
-  runCommand(`lxc exec ${targetVM} -- sh -c 'lxc cluster link delete ${link}'`);
+  const remoteVm = getRemoteClusterVm();
+  runCommand(`lxc exec ${remoteVm} -- sh -c 'lxc cluster link delete ${link}'`);
+};
+
+export const getRemoteClusterVm = () => {
```
kimanhou: Sorry, not related to your PR, but should `getRemoteClusterVm` and `getLocalClusterVm` live in `tests/helpers/cluster.ts` ?

tests/helpers/projects.ts
```diff
@@ -90,3 +90,43 @@ export const deleteProject = async (page: Page, project: string) => {
 
   await confirmDelete(page, project);
 };
+
+export const promoteProjectToLeader = async (page: Page, project: string) => {
```
kimanhou, line 94: Should we also test force promote and force demote ?

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const dialog = page.getByRole("dialog", { name: "Confirm delete" });
+  const confirmButton = dialog
+    .getByRole("button")
+    .filter({ hasText: "Delete" });
+  await confirmButton.click();
+  await dismissNotification(page, `Replicator ${replicatorName} deleted.`);
+};
+
+export const openReplicatorRowModal = async (
```
kimanhou: I think the name should include "run", even tho in some cases it's restore and not run. The component is called `RunReplicatorBtn`.

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,251 @@
...
+  );
+};
+
+export const deleteProjectOnRemoteCluster = (project: string) => {
+  const remoteVm = getRemoteClusterVm();
+  runCommand(`lxc exec ${remoteVm} -- sh -c 'lxc project delete ${project}'`);
+};
+
+export const cleanupReplicationState = (): void => {
+  const DELETE_ALL_REPLICATORS_COMMAND =
```
kimanhou: `DELETE_ALL_REPLICATORS_COMMAND` can be defined outside of the function `cleanupReplicationState`

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,251 @@
...
+  project: string,
+  clusterLinkName: string,
+) => {
+  await openReplicatorRowModal(page, replicatorName);
+  await assertLeaderPreflightChecks(page, project, clusterLinkName);
+  const dialog = page.getByRole("dialog", { name: "Run replicator" });
+  await dialog.getByRole("button", { name: "Cancel" }).click();
+};
+
+export const testStandbyPreflightChecksInModal = async (
```
kimanhou: Perhaps the "Overwrite local data" confirmation checkbox could be tested here too ? The "Restore" button should be disabled when the confirmation checkbox is unchecked and should become enabled once the confirmation checkbox is checked.

tests/replicators.spec.ts
```diff
@@ -0,0 +1,7 @@
+import { type LxdVersions, test } from "./fixtures/lxd-test";
```
kimanhou: I guess this file can be deleted for now ?

tests/helpers/cluster-links.ts
```diff
@@ -12,6 +12,9 @@ export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
   );
 };
 
+export const DELETE_ALL_CLUSTER_LINKS_COMMAND =
+  "lxc cluster link list --format csv | tail -n +1 | cut -d, -f1 | xargs -r -n1 lxc cluster link delete";
```
edlerd: I think due to the `tail` with argument `+1` this only deletes one cluster link, not all.

tests/helpers/replicators.ts
edlerd: I think we will always have a replicator table, also in the project configuration page. There is no single replicator detail page as far as I know.

tests/helpers/replicators.ts
edlerd: maybe `openReplicatorRunModal`?

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,251 @@
...
+  await expect(
+    dialog.getByText(`Project ${projectName} is in leader mode`),
+  ).toBeVisible();
+  await expect(
+    dialog.getByText(`Cluster link ${clusterLinkName} is reachable`),
+  ).toBeVisible();
+  await expect(
+    dialog.getByText(
+      new RegExp(
+        `This will sync \\d+ instances from the source project ${projectName} to the standby cluster ${clusterLinkName}`,
```
edlerd: Can we replace the \d with the actual number? We should know the number at this point.

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,251 @@
...
+  ).toBeVisible();
+  await expect(
+    dialog.getByText(`Cluster link ${clusterLinkName} is reachable`),
+  ).toBeVisible();
+  await expect(
+    dialog.getByText(
+      new RegExp(
+        `This will sync all instances from the ${clusterLinkName} cluster back to the ${projectName} project`,
+      ),
+    ),
```
edlerd, line 135:
```suggestion
    dialog.getByText(
        `This will sync all instances from the ${clusterLinkName} cluster back to the ${projectName} project`,
    ),
```
We don't need the regexp here.

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,251 @@
...
+    `lxc exec ${remoteVm} -- sh -c 'lxc project demote-replica ${project}'`,
+  );
+};
+
+export const deleteProjectOnRemoteCluster = (project: string) => {
+  const remoteVm = getRemoteClusterVm();
+  runCommand(`lxc exec ${remoteVm} -- sh -c 'lxc project delete ${project}'`);
+};
+
+export const cleanupReplicationState = (): void => {
```
edlerd: Let's split this into `deleteAllReplicators` and `deleteAllClusterLinks`, which each cleanup all respective entities on both local and remote LXD.

tests/helpers/replicators.ts
edlerd: This should be split into smaller units. I think the creation of the cluster link should not be a concern of this replicators helper. The test can call it first and then this function.

So the interface of this function also becomes clearer and more straightforward: `setReplicaCluster(page, project, cluster)`.

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,251 @@
...
+) => {
+  await createClusterLink(page, clusterLinkName, token);
+  await selectReplicaCluster(page, clusterLinkName);
+  await saveProjectConfiguration(page, project);
+  await expect(page.getByLabel("Replica cluster")).toContainText(
+    clusterLinkName,
+  );
+};
+
+export const testLeaderPreflightChecksInModal = async (
```
edlerd: No need for the indirection, we can make all of this part of `assertLeaderPreflightChecks`.

tests/helpers/replicators.ts
edlerd: This should be inlined to `assertStandbyPreflightChecks`.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  // Create standby project & promote to leader
+  createStandbyProjectOnRemoteCluster(project, clusterLinkName);
+  await promoteProjectToLeader(page, project);
+  await assertProjectReplicaMode(page, "leader");
+  await expect(
+    page.getByText("This project is the active source project for replication"),
+  ).toBeVisible();
+
+  // Create & edit replicator
+  const replicatorName = randomReplicatorName();
```
edlerd:
```suggestion
  const replicator = randomReplicatorName();
```
We can avoid the `Name` suffix in the variables of tests. We refer to the entity by name in all tests throughout anyhow.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+
+  // Project & replication config setup
+  await createProject(page, project);
+  await openProjectConfiguration(page);
+  await page.getByText("Replication").click();
+  await assertProjectReplicaMode(page, "None");
+  await assertReplicationConfigInitialState(page);
+
+  // Cluster link & replica cluster setup
+  const clusterLinkName = randomLinkName();
```
edlerd:
```suggestion
  const clusterLink = randomClusterLinkName();
```

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  ).toBeVisible();
+
+  // Create & edit replicator
+  const replicatorName = randomReplicatorName();
+  await createReplicator(page, replicatorName, clusterLinkName, project);
+  const replicatorRow = page
+    .getByRole("row")
+    .filter({ hasText: replicatorName });
+
+  await replicatorRow.getByRole("button", { name: "Edit replicator" }).click();
```
edlerd, line 111:
```suggestion
```
This should move into the `editReplicatorSidePanel` function.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  await editReplicatorSidePanel(page, replicatorName, "Updated description");
+
+  // Test leader pre-flight checks
```
edlerd:
```suggestion
```
The function name gives the same information as the comment.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  await testLeaderPreflightChecksInModal(
+    page,
+    replicatorName,
+    project,
+    clusterLinkName,
+  );
+
+  // Demote to standby & test standby pre-flight checks
+  await demoteProjectToStandby(page, project);
+  await assertProjectReplicaMode(page, "standby");
```
edlerd:
```suggestion
```
We can move the assertion into `demoteProjectToStandby`.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  await expect(
+    page.getByText("This project is a read-only failover target"),
+  ).toBeVisible();
```
edlerd:
```suggestion
```
This can move into `demoteProjectToStandby` as well.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+
+  // Test leader pre-flight checks
+  await testLeaderPreflightChecksInModal(
+    page,
+    replicatorName,
+    project,
+    clusterLinkName,
+  );
+
+  // Demote to standby & test standby pre-flight checks
```
edlerd:
```suggestion
```
I think the function names below will clarify this.

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  await assertProjectReplicaMode(page, "leader");
+
+  // Delete replicator & clear replica mode
+  await deleteReplicatorRow(page, replicatorName);
+  await expect(replicatorRow).toHaveCount(0);
+
+  await page.getByRole("button", { name: "Clear replica mode" }).click();
+  await assertProjectReplicaMode(page, "None");
+
+  // Cleanup
```
edlerd:
```suggestion
```

tests/projects-clustered.spec.ts
```diff
@@ -41,3 +66,86 @@ test("project edit configuration", async ({ page, lxdVersion }, testInfo) => {
...
+  deleteProjectOnRemoteCluster(project);
+  deleteClusterLinkOnRemoteCluster(clusterLinkName);
+  await visitClusterLinks(page);
+  await deleteClusterLink(page, clusterLinkName);
```
edlerd:
```suggestion
  await deleteClusterLink(page, clusterLinkName);
```
I think delete cluster link should also navigate there.

edlerd (commented):
A lot of small adjustment comments below.

tests/helpers/replicators.ts
omarelkashef: We do not know the number of instances in the tests. We will have to compute it somehow by visiting the instances list page for example, which will add unnecessary steps in my opinion.

edlerd: The test should create a new project, and then we can control the number of instances in it.

tests/helpers/replicators.ts
omarelkashef: We have a replicators detail page and we can delete from there. The detail page will be tested in the tests that @kimanhou adds.

tests/projects-clustered.spec.ts
omarelkashef: This will make `editReplicatorSidePanel` only useful in this test. I wanted to `editReplicatorSidePanel` to be used here and the replicators detail and list page tests. In those tests, the edit button will be in a row (for the list page) and in the header (for the detail page). All tests can open the side panel and then use this function.

tests/projects-clustered.spec.ts
omarelkashef: Most of the uses of `deleteClusterLink` is in cluster links tests where we are already in the correct page. If we make `deleteClusterLink` navifate to the links page, it will add unnecessary navigation to many tests.

tests/scripts/start_vm_cluster.sh
omarelkashef: If we do that, we should find a new name that is more general and does not have cluster link

tests/helpers/replicators.ts
```diff
@@ -0,0 +1,231 @@
...
+  ).toBeVisible();
+  await expect(
+    page.getByText("Only outgoing replicators are shown here"),
+  ).toBeVisible();
+  await expect(
+    page.getByText("No replication configuration applied."),
+  ).toBeVisible();
+};
+
+export const setupReplicaClusterLink = async (
```
edlerd:
```suggestion
export const setClusterForProject = async (
```
Maybe more clear?

edlerd (dismissed):
LGTM, tiny nitpick on naming left.

edlerd (approved):

PR #2097, opened 2026-06-25
fix(yaml): fix bug where YAML view does not preserve previous section

omarelkashef:
## Done

- fix bug where YAML view does not preserve previous section
- fix bug where YAML view does not open in create network tab if user is not on the GENERAL section

WD-37200

## QA

1. Run the LXD-UI:
    - Go to Instances/Profiles -> Create/Edit 
    - Go to any section and click on the YAML switch
    - Make edits and turn off the YAML switch (without saving is fine)
    - Make sure you are back on the same section before opening the YAML view
    - Go to Networking -> Networks -> Create
    - Make sure you can open the YAML view when in a section other than GENERAL (compare with previous behavior in the video attached below)

## Screenshots

https://github.com/user-attachments/assets/f7ff130e-3021-4e6e-9d5e-e5dd842d4fb3

### DIFF
--- a/src/components/forms/YamlSwitch.tsx
+++ b/src/components/forms/YamlSwitch.tsx
@@ -1,4 +1,4 @@
-import type { FC, FormEvent } from "react";
+import { useEffect, useRef, type FC, type FormEvent } from "react";
 import { Switch, usePortal } from "@canonical/react-components";
 import YamlConfirmation from "components/forms/YamlConfirmation";
 import { slugify } from "util/slugify";
@@ -26,8 +26,20 @@ const YamlSwitch: FC<Props> = ({
   const { openPortal, closePortal, isOpen, Portal } = usePortal();
   const yamlFormik = formik as FormikProps<{ yaml: string }>;
   const isSmallScreen = useIsScreenBelow();
+  const isYamlSection = slugify(section ?? "") === slugify(YAML_CONFIGURATION);
+  const previousSection = useRef<string>(
+    isYamlSection || !section ? MAIN_CONFIGURATION : section,
+  );
 
-  const isChecked = slugify(section ?? "") === slugify(YAML_CONFIGURATION);
+  useEffect(() => {
+    if (!section) {
+      return;
+    }
+
+    if (!isYamlSection) {
+      previousSection.current = section;
+    }
+  }, [section]);
 
   const handleSwitch = (e: FormEvent<HTMLInputElement>) => {
     if (yamlFormik.values.yaml) {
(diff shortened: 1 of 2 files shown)


src/components/forms/YamlSwitch.tsx
```diff
@@ -15,34 +15,57 @@ interface Props {
...
   disableReason?: string;
+  defaultSection?: string;
 }
 
 const YamlSwitch: FC<Props> = ({
   section,
   setSection,
   formik,
   disableReason,
+  defaultSection = MAIN_CONFIGURATION,
```
edlerd:
```suggestion
```
We don't need this here.

src/components/forms/YamlSwitch.tsx
```diff
@@ -15,34 +15,57 @@ interface Props {
...
+
+    if (slugify(section) !== slugify(YAML_CONFIGURATION)) {
+      previousSection.current = section;
+    }
+  }, [section]);
+
+  const getSectionToExitYaml = () =>
+    isYamlSection(previousSection.current)
+      ? defaultSection
+      : previousSection.current;
```
edlerd:
```suggestion
```
We don't need this, if we control properly the value in `previousSection` to be valid.

src/components/forms/YamlSwitch.tsx
```diff
@@ -15,34 +15,57 @@ interface Props {
...
   disableReason,
+  defaultSection = MAIN_CONFIGURATION,
 }) => {
   const { openPortal, closePortal, isOpen, Portal } = usePortal();
   const yamlFormik = formik as FormikProps<{ yaml: string }>;
   const isSmallScreen = useIsScreenBelow();
 
-  const isChecked = slugify(section ?? "") === slugify(YAML_CONFIGURATION);
+  const isYamlSection = (value?: string) =>
+    slugify(value ?? "") === slugify(YAML_CONFIGURATION);
```
edlerd: I don't think we need the extra function, and we can keep this as is -- execute immediately the check and store the result. Maybe rename `isChecked` to `isYamlSection` to be a bit clearer below.

edlerd (commented):
I think this can be simplified. See comments below.

src/components/forms/YamlSwitch.tsx
omarelkashef: It is added to not make the yaml switch aware of specific sections and in case we need to override for a future page, but we can add it then.

edlerd (approved):
LGTM, thanks for the fix!

PR #2098, opened 2026-06-25
chore(tests) add load balancer tests

edlerd:
## Done

- chore(tests) add load balancer tests

### DIFF
--- a/tests/network-load-balancer-clustered.spec.ts
+++ b/tests/network-load-balancer-clustered.spec.ts
@@ -1,0 +1,143 @@
+import { test } from "./fixtures/lxd-test";
+import {
+  createNetwork,
+  createOvnUplink,
+  deleteNetwork,
+  randomLoadBalancerPoolName,
+  randomNetworkName,
+  skipIfNotSupported,
+  supportsLoadBalancers,
+  visitNetwork,
+} from "./helpers/network";
+import { isClusteredTestProject, skipIfNotClustered } from "./helpers/cluster";
+import { dismissNotification } from "./helpers/notification";
+
+test.describe("Network Load Balancer", () => {
+  const UPLINK_NAME = randomNetworkName();
+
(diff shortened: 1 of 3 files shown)


tests/network-load-balancer-clustered.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+import { dismissNotification } from "./helpers/notification";
+
+test.describe("Network Load Balancer", () => {
+  const UPLINK_NAME = randomNetworkName();
+
+  test.beforeAll(async ({ browser, lxdVersion }, testInfo) => {
+    if (
+      lxdVersion === "5.0-edge" ||
+      lxdVersion === "5.21-edge" ||
+      !testInfo.project.name.includes(":clustered")
```
kimanhou: Not related to this PR, but is it time to unify these checks with `skipIfNotClustered` and `skipIfNotSupported` ? Maybe create functions `isNotClustered` and `isUnsupportedVersion`:

```
export const isNotClustered = (projectName: string) =>
  !projectName.includes(":clustered");

export const skipIfNotClustered = (projectName: string) => {
  test.skip(isNotClustered(projectName));
};

export const isUnsupportedVersion = (lxdVersion: LxdVersions) =>
  lxdVersion === "5.0-edge";

export const skipIfNotSupported = (lxdVersion: LxdVersions) => {
  test.skip(
    isUnsupportedVersion(lxdVersion),
    "Cluster tests not supported for lxd 5.0",
  );
};
```

The new functions could then be re-used in `beforeAll`:
```
if (
      isUnsupportedVersion(lxdVersion) ||
      isNotClustered(testInfo.project.name)
    ) {
      console.log("Skipping uplink creation");
      return;
    }
```

edlerd: Yes, we can simplify the checks here. Good idea. Adjusted it accordingly.

kimanhou (approved):

PR #2101, opened 2026-06-25
feat(File explorer): [WD-37160] Create new Directory

Kxiru:
## Done

- [x] Creates a new directory in a specified path.
- [x] Allow traversal of directories via input

## QA

1. Run the LXD-UI:
    - Using a running instance, or a container, navigate to the file explorer tab.
    - Observe the "Create" button
    - Attempt to create a new directory, both within the current directory and at a specified path.

## Screenshots

image
image
**Note: Modal closes after submitting to create within a dir that doesn't exist.**

Directory navigation
image

image

### DIFF
--- a/src/pages/instances/FileExplorerBreadcrumb.tsx
+++ b/src/pages/instances/FileExplorerBreadcrumb.tsx
@@ -1,7 +1,12 @@
-import type { FC } from "react";
-import { Link } from "react-router-dom";
+import { type FC, type KeyboardEvent, useEffect, useState } from "react";
+import { Link, useNavigate } from "react-router-dom";
+import { Button, Icon, Input } from "@canonical/react-components";
 import type { LxdInstance } from "types/instance";
-import { getFileExplorerDirectoryURL } from "util/instances";
+import {
+  getFileExplorerDirectoryURL,
+  validateDirectoryPathSyntax,
+} from "util/instances";
+import { fetchInstanceDirectory } from "api/instances";
 
 interface Props {
   currentPath: string;
@@ -9,42 +14,146 @@ interface Props {
 }
 
 const FileExplorerBreadcrumb: FC<Props> = ({ currentPath, instance }) => {
-  const segments = currentPath.split("/").filter(Boolean);
+  const [isEditing, setIsEditing] = useState(false);
+  const [path, setPath] = useState(currentPath);
+  const [isNavigating, setIsNavigating] = useState(false);
+  const [inputError, setInputError] = useState<string | undefined>(undefined);
+  const navigate = useNavigate();
 
+  useEffect(() => {
+    setPath(currentPath);
+    setInputError(undefined);
+    setIsEditing(false);
+  }, [currentPath]);
+
+  const handleNavigate = async () => {
+    const trimmedPath = path.trim();
+
+    if (trimmedPath === currentPath) {
+      setInputError(undefined);
+      setIsEditing(false);
+      return;
+    }
+
+    const syntaxError = validateDirectoryPathSyntax(trimmedPath);
+    setInputError(syntaxError);
+    if (syntaxError) {
+      return;
+    }
+
+    setIsNavigating(true);
+    try {
+      await fetchInstanceDirectory(
+        instance.name,
+        instance.project,
+        trimmedPath,
+      );
+      navigate(getFileExplorerDirectoryURL(trimmedPath, instance));
+    } catch (error) {
+      setInputError(
+        "Invalid path: " +
+          (error instanceof Error ? error.message : String(error)),
+      );
+    } finally {
+      setIsNavigating(false);
+    }
+  };
+
+  const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Enter") {
+      void handleNavigate();
+    }
+    if (e.key === "Escape") {
+      setPath(currentPath);
+      setInputError(undefined);
+      setIsEditing(false);
+    }
+  };
+
+  if (isEditing) {
+    return (
+      <div className="file-explorer-path-editor">
+        <Input
+          type="text"
+          className="file-explorer-path-input"
+          aria-label="Search path"
+          value={path}
+          onChange={(e) => {
+            setPath(e.target.value);
+          }}
+          onKeyDown={handleKeyDown}
+          disabled={isNavigating}
+          error={inputError}
+          autoFocus
+        />
+        <div>
+          <Button
+            appearance="base"
+            onClick={() => void handleNavigate()}
+            aria-label="Go to path"
+            disabled={isNavigating}
+            hasIcon
+          >
+            <Icon name="search" />
+          </Button>
+        </div>
+      </div>
+    );
+  }
+
+  const segments = currentPath.split("/").filter(Boolean);
   const breadcrumbs = [
-    {
-      label: "root",
-      path: "/",
-    },
+    { label: "root", path: "/" },
     ...segments.map((segment, index) => ({
       label: segment,
       path: "/" + segments.slice(0, index + 1).join("/"),
     })),
   ];
-  return (
-    <nav
-      className="p-breadcrumbs p-breadcrumbs--large"
-      aria-label="File Explorer Path"
-    >
-      <ol className="p-breadcrumbs__items breadcrumb-wrapper">
-        <li className="p-heading--4 breadcrumb-header">Directory:&nbsp;</li>
-        {breadcrumbs.map((crumb, index) => {
-          const isCurrentDirectory = index === breadcrumbs.length - 1;
 
-          return (
-            <li key={crumb.path} className="p-heading--4 continuous-breadcrumb">
-              {isCurrentDirectory ? (
-                <span>{crumb.label}</span>
-              ) : (
-                <Link to={getFileExplorerDirectoryURL(crumb.path, instance)}>
-                  {crumb.label}
-                </Link>
-              )}
-            </li>
-          );
-        })}
-      </ol>
-    </nav>
+  return (
+    <div className="file-explorer-breadcrumb-bar">
+      <nav
+        className="p-breadcrumbs p-breadcrumbs--large"
+        aria-label="File Explorer Path"
+      >
+        <ol className="p-breadcrumbs__items breadcrumb-wrapper">
+          <li className="p-heading--5 breadcrumb-header">Directory:&nbsp;</li>
+          {breadcrumbs.map((crumb, index) => {
+            const isCurrentDirectory = index === breadcrumbs.length - 1;
+
+            return (
+              <li
+                key={crumb.path}
+                className="p-heading--5 continuous-breadcrumb"
+              >
+                {isCurrentDirectory ? (
+                  <span>{crumb.label}</span>
+                ) : (
+                  <Link to={getFileExplorerDirectoryURL(crumb.path, instance)}>
+                    {crumb.label}
+                  </Link>
+                )}
+              </li>
+            );
+          })}
+          <li>
+            <Button
+              appearance="base"
+              className="u-no-margin--bottom"
+              onClick={() => {
+                setPath(currentPath);
+                setIsEditing(true);
+              }}
+              title="Search path"
+              aria-label="Search path"
+              hasIcon
+            >
+              <Icon name="search" />
+            </Button>
+          </li>
+        </ol>
+      </nav>
+    </div>
   );
 };
 
--- a/src/pages/instances/FileExplorerCreateBtn.tsx
+++ b/src/pages/instances/FileExplorerCreateBtn.tsx
@@ -1,0 +1,59 @@
+import { type FC } from "react";
+import { Button, Icon, usePortal } from "@canonical/react-components";
+import type { LxdInstance } from "types/instance";
+import { useInstanceEntitlements } from "util/entitlements/instances";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import FileExplorerCreateModal from "./forms/FileExplorerCreateModal";
+
+interface Props {
+  instance: LxdInstance;
+  currentPath: string;
+  refreshDirectoryList: () => void;
+}
+
+const FileExplorerCreateBtn: FC<Props> = ({
+  instance,
+  currentPath,
+  refreshDirectoryList,
+}) => {
+  const { openPortal, closePortal, isOpen, Portal } = usePortal();
+  const isSmallScreen = useIsScreenBelow();
+  const { canAccessInstanceFiles } = useInstanceEntitlements();
+  const hasPermission = canAccessInstanceFiles(instance);
+
+  const handleSuccess = () => {
+    refreshDirectoryList();
+    closePortal();
+  };
+
+  return (
+    <>
+      <Button
+        onClick={openPortal}
+        hasIcon={!isSmallScreen}
+        disabled={!hasPermission || isOpen}
+        className="u-no-margin--bottom"
+        title={
+          hasPermission
+            ? "Create a new directory"
+            : "You do not have permission to manage files on this instance"
+        }
+      >
+        {!isSmallScreen && <Icon name="plus" />}
+        <span>{isSmallScreen ? "Create" : "Create directory"}</span>
+      </Button>
+      {isOpen && (
+        <Portal>
+          <FileExplorerCreateModal
+            close={closePortal}
+            instance={instance}
+            currentPath={currentPath}
+            onSuccess={handleSuccess}
+          />
+        </Portal>
+      )}
+    </>
+  );
+};
+
+export default FileExplorerCreateBtn;
--- a/src/pages/instances/InstanceFileExplorer.tsx
+++ b/src/pages/instances/InstanceFileExplorer.tsx
@@ -12,6 +12,7 @@ import { isInstanceRunning } from "util/instanceStatus";
 import FileExplorerTable from "./FileExplorerTable";
 import FileExplorerBreadcrumb from "./FileExplorerBreadcrumb";
 import UploadFileExplorerBtn from "./UploadFileExplorerBtn";
+import FileExplorerCreateBtn from "./FileExplorerCreateBtn";
 import { fetchInstanceDirectory } from "api/instances";
 import { useSearchParams } from "react-router-dom";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
@@ -80,14 +81,26 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
 
   return (
     <Row className="general">
-      <div className="file-explorer-header">
-        <FileExplorerBreadcrumb currentPath={currentPath} instance={instance} />
-        <UploadFileExplorerBtn
-          instance={instance}
-          currentPath={currentPath}
-          refreshDirectoryList={invalidateCache}
-          directoryContent={directoryContent?.metadata ?? []}
-        />
+      <div className="file-explorer-controls-bar">
+        <div className="search-box-wrapper">
+          <FileExplorerBreadcrumb
+            currentPath={currentPath}
+            instance={instance}
+          />
+        </div>
+        <div className="file-explorer-actions">
+          <FileExplorerCreateBtn
+            instance={instance}
+            currentPath={currentPath}
+            refreshDirectoryList={invalidateCache}
+          />
+          <UploadFileExplorerBtn
+            instance={instance}
+            currentPath={currentPath}
+            refreshDirectoryList={invalidateCache}
+            directoryContent={directoryContent?.metadata ?? []}
+          />
+        </div>
       </div>
       {error ? (
         <>
--- a/src/pages/instances/forms/FileExplorerCreateModal.tsx
+++ b/src/pages/instances/forms/FileExplorerCreateModal.tsx
@@ -1,0 +1,117 @@
+import { useState, type FC, type FormEvent } from "react";
+import {
+  Button,
+  Input,
+  Modal,
+  useToastNotification,
+} from "@canonical/react-components";
+import type { LxdInstance } from "types/instance";
+import { createInstanceDirectory } from "api/instances";
+import {
+  validateDirectoryPathSyntax,
+  getFileExplorerDirectoryURL,
+} from "util/instances";
+import { useNavigate } from "react-router-dom";
+
+interface Props {
+  close: () => void;
+  instance: LxdInstance;
+  currentPath: string;
+  onSuccess: () => void;
+}
+
+const FileExplorerCreateModal: FC<Props> = ({
+  close,
+  instance,
+  currentPath,
+  onSuccess,
+}) => {
+  const [directoryName, setDirectoryName] = useState("");
+  const [parentPath, setParentPath] = useState(currentPath);
+  const [isCreating, setIsCreating] = useState(false);
+  const [inputError, setInputError] = useState<string | undefined>(undefined);
+
+  const toastNotify = useToastNotification();
+  const navigate = useNavigate();
+
+  const handleSubmit = async (e: FormEvent) => {
+    e.preventDefault();
+
+    const syntaxError = validateDirectoryPathSyntax(parentPath);
+    if (syntaxError) {
+      setInputError(syntaxError);
+      setIsCreating(false);
+      return;
+    }
+
+    const fullPath =
+      parentPath === "/"
+        ? `/${directoryName}`
+        : `${parentPath}/${directoryName}`;
+
+    setIsCreating(true);
+    try {
+      await createInstanceDirectory(instance, fullPath);
+      toastNotify.success(`Directory ${directoryName} created successfully.`);
+      navigate(getFileExplorerDirectoryURL(parentPath, instance));
+      onSuccess();
+    } catch (e) {
+      close();
+      toastNotify.failure("Directory creation failed", e);
+      setIsCreating(false);
+    }
+  };
+
+  return (
+    <Modal
+      close={close}
+      title="Create directory"
+      className="create-directory-modal"
+      buttonRow={
+        <>
+          <Button className="u-no-margin--bottom" onClick={close}>
+            Cancel
+          </Button>
+          <Button
+            appearance="positive"
+            className="u-no-margin--bottom"
+            disabled={!directoryName.trim() || isCreating}
+            onClick={(e) => void handleSubmit(e)}
+            loading={isCreating}
+          >
+            Create
+          </Button>
+        </>
+      }
+    >
+      <form onSubmit={(e) => void handleSubmit(e)}>
+        <Input type="submit" hidden value="Hidden input" />
+        <Input
+          id="directory-name"
+          label="Directory name"
+          type="text"
+          required
+          takeFocus
+          takeFocusDelay={100}
+          value={directoryName}
+          onChange={(e) => {
+            setDirectoryName(e.target.value);
+          }}
+          placeholder="Enter new directory name"
+        />
+        <Input
+          id="parent-path"
+          label="Parent path"
+          type="text"
+          value={parentPath}
+          onChange={(e) => {
+            setParentPath(e.target.value);
+          }}
+          error={inputError}
+        />
+      </form>
+    </Modal>
+  );
+};
+
+export default FileExplorerCreateModal;
--- a/src/pages/instances/forms/FileExplorerCreateModal.tsx
+++ b/src/pages/instances/forms/FileExplorerCreateModal.tsx
@@ -1,11 +1,13 @@
-import { useState, type FC, type FormEvent } from "react";
+import { useState, type FC } from "react";
 import {
   Button,
   Input,
   Modal,
+  useNotify,
   useToastNotification,
 } from "@canonical/react-components";
 import type { LxdInstance } from "types/instance";
+import NotificationRow from "components/NotificationRow";
 import { createInstanceDirectory } from "api/instances";
 import {
   validateDirectoryPathSyntax,
@@ -31,45 +33,59 @@ const FileExplorerCreateModal: FC<Props> = ({
   const [isCreating, setIsCreating] = useState(false);
   const [inputError, setInputError] = useState<string | undefined>(undefined);
 
+  const notify = useNotify();
   const toastNotify = useToastNotification();
   const navigate = useNavigate();
 
-  const handleSubmit = async (e: FormEvent) => {
+  const handleClose = () => {
+    notify.clear();
+    close();
+  };
+
+  const handleSubmit = async (e: { preventDefault: () => void }) => {
     e.preventDefault();
 
-    const syntaxError = validateDirectoryPathSyntax(parentPath);
+    setInputError(undefined);
+
+    const trimmedParentPath = parentPath.trim();
+    const trimmedDirectoryName = directoryName.trim();
+
+    const fullPath = trimmedParentPath.endsWith("/")
+      ? `${trimmedParentPath}${trimmedDirectoryName}`
+      : `${trimmedParentPath}/${trimmedDirectoryName}`;
+
+    const syntaxError =
+      validateDirectoryPathSyntax(trimmedParentPath) ||
+      validateDirectoryPathSyntax(fullPath);
+
     if (syntaxError) {
       setInputError(syntaxError);
       setIsCreating(false);
       return;
     }
 
-    const fullPath =
-      parentPath === "/"
-        ? `/${directoryName}`
-        : `${parentPath}/${directoryName}`;
-
     setIsCreating(true);
     try {
       await createInstanceDirectory(instance, fullPath);
-      toastNotify.success(`Directory ${directoryName} created successfully.`);
-      navigate(getFileExplorerDirectoryURL(parentPath, instance));
+      toastNotify.success(
+        `Directory ${trimmedDirectoryName} created successfully.`,
+      );
+      navigate(getFileExplorerDirectoryURL(trimmedParentPath, instance));
       onSuccess();
     } catch (e) {
-      close();
-      toastNotify.failure("Directory creation failed", e);
+      notify.failure("Directory creation failed", e);
       setIsCreating(false);
     }
   };
 
   return (
     <Modal
-      close={close}
+      close={handleClose}
       title="Create directory"
       className="create-directory-modal"
       buttonRow={
         <>
-          <Button className="u-no-margin--bottom" onClick={close}>
+          <Button className="u-no-margin--bottom" onClick={handleClose}>
             Cancel
           </Button>
           <Button
@@ -84,6 +100,7 @@ const FileExplorerCreateModal: FC<Props> = ({
         </>
       }
     >
+      <NotificationRow />
       <form onSubmit={(e) => void handleSubmit(e)}>
         <Input type="submit" hidden value="Hidden input" />
         <Input
--- a/src/sass/_file_explorer.scss
+++ b/src/sass/_file_explorer.scss
@@ -45,6 +45,20 @@
   .p-breadcrumbs {
     overflow: hidden;
   }
+
+  .p-breadcrumbs__items {
+    list-style: none;
+
+    > li {
+      display: inline;
+    }
+  }
+
+  .continuous-breadcrumb {
+    line-height: 1.8;
+    overflow-wrap: break-word;
+    word-break: break-all;
+  }
 }
 
 .file-explorer-path-editor {
--- a/src/sass/_file_explorer.scss
+++ b/src/sass/_file_explorer.scss
@@ -1,3 +1,19 @@
+.file-explorer-controls-bar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1rem;
+
+  .search-box-wrapper {
+    flex: 1 1 0;
+    max-width: 45rem;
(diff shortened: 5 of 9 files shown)


Kxiru:
Current design considerations for directory navigation as discussed in Mattermost:
Two options:
- Clicking on the current directory and using a component similar to the rename header (used in instances and other entities) to submit and automatically navigate to a new page and perhaps provide a toast notif if the directory does not exist.
  - The rename header component idea is more aligned with the design but @omarelkashef  suggests that it does not provide an explicit instruction tot he user that it can be used. We could circumvent that with an edit button but it would break existing patterns. Not impossible though.
image


- Even better, on pages such as Snapshots (attached 🫠) we can have a Search bar inline with the File explorer actions. Could that work? It kind of mimics how we can traverse directories natively in our file explorers.
  - It can be argued that semantically, a search bar implies we are looking for a file/directory, not whether we can "search" for a path. The browser URL bar and ubuntu file explorer bar are examples that do both.
image

image

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+        />
+        <Input
+          id="directory-path"
+          label="Full path"
+          type="text"
+          value={targetPath}
+          onChange={(e) => {
+            setTargetPath(e.target.value);
+          }}
+          placeholder={currentPath}
```
edlerd: this should be the value, not the placeholder

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+          id="directory-name"
+          label="Directory name"
+          type="text"
+          required
+          autoFocus
+          value={directoryName}
+          onChange={(e) => {
+            setDirectoryName(e.target.value);
+          }}
+          placeholder="my-directory"
```
edlerd: The placeholder seems a bit odd. How about "Enter new directory name"?

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    const basePath = targetPath.trim() || currentPath;
+    const fullPath =
+      basePath === "/" ? `/${directoryName}` : `${basePath}/${directoryName}`;
+
+    setIsCreating(true);
+    try {
+      await createInstanceDirectory(instance, fullPath);
+      toastNotify.success(`Directory ${directoryName} created successfully.`);
+      onSuccess(fullPath);
```
edlerd: I don't think we should move into the newly created directory after creation. It seems unexpected from other file explorer experiences.

src/pages/instances/FileExplorerCreateBtn.tsx
```diff
@@ -0,0 +1,63 @@
...
+        disabled={!hasPermission || isOpen}
+        className="u-no-margin--bottom u-no-margin--right"
+        title={
+          hasPermission
+            ? "Create a new directory"
+            : "You do not have permission to manage files on this instance"
+        }
+      >
+        {!isSmallScreen && <Icon name="plus" />}
+        <span>Create</span>
```
edlerd: Maybe this should say "Create directory", at least on desktop where we have enough space. Wdyt?

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+            loading={isCreating}
+          >
+            Create
+          </Button>
+        </>
+      }
+    >
+      <form onSubmit={(e) => void handleSubmit(e)}>
+        <Input
+          id="directory-name"
```
edlerd, line 107: Best autofocus on this field and allow to submit with enter. The enter submission needs a hidden field, you can find the pattern in other forms.

edlerd (commented):
Good start. A few thoughts on the draft below.

src/pages/instances/FileExplorerCreateBtn.tsx
Kxiru: It could but it seems inconsistent size wise with "Upload".

src/pages/instances/forms/FileExplorerCreateModal.tsx
Kxiru: I have amended this, now.

src/pages/instances/FileExplorerCreateBtn.tsx
edlerd: My concern is that currently it is unclear what the buttons do. A user will see "Create" and "Upload", I assume that can be perceived as confusing.

sophiecarroll:
Screenshot 2026-06-30 at 11 30 23 AM
Is this the correct version of file explorer (I looked in the demo on this PR)? Two quick things: the title of the tab at the top should be sentence case ("File explorer" not "File Explorer") to be consistent with the rest of the >1 word titles on the rest of the UI. Also can you change the pencil icon to a magnifying glass search icon? The pencil icon might lead people to believe they are editing the directory somehow instead of just searching. Also when I click it the original filepath doesn't stay, it starts with a cleared search box, and if the file path stays I think doubles as good example text for what type of content (a path) the user can enter.

src/pages/instances/FileExplorerCreateBtn.tsx
Kxiru: Surely clicking the button would clarify that? And/or it is implicit given we are in a File Explorer. I will rename it to "Create directory", let me know what you think.

Kxiru:
> the title of the tab at the top should be sentence case ("File explorer" not "File Explorer")...

Addressed!

> Also can you change the pencil icon to a magnifying glass search icon?

Also addressed.

src/pages/instances/forms/FileExplorerCreateModal.tsx
edlerd: Can we set the current path as the default value please?

src/pages/instances/forms/FileExplorerCreateModal.tsx
edlerd: Can we address the suggestion please?

src/pages/instances/FileExplorerCreateBtn.tsx
edlerd: Thanks! I think this is clearer now. On mobile we should shorted in to "Create" to avoid breaking lines though.

src/pages/instances/forms/FileExplorerCreateModal.tsx
edlerd: This seems unaddressed.

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,170 @@
...
+        appearance="base"
+        className="file-explorer-edit-btn u-no-margin--bottom"
+        onClick={() => {
+          setIsEditing(true);
+        }}
+        aria-label="Edit path"
+        hasIcon
+      >
+        <Icon name="search" />
+      </Button>
```
edlerd: I like this tiny toggle to enable edit mode.

Placement should be improved, there is a lot of whitespace between the path and the button in the screen below. There should be a fixed space, likely 1 or 2rem between the end of the path and the icon to make it clear what it refers to.

Image

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,170 @@
...
+        <Button
+          appearance="base"
+          className="file-explorer-go-btn u-no-margin--bottom"
+          onClick={() => void handleNavigate()}
+          aria-label="Go to path"
+          disabled={isNavigating}
+          hasIcon
+        >
+          <Icon name="chevron-right" />
+        </Button>
```
edlerd: When going into edit mode and then clicking this button (without changing the path), nothing happens.

edlerd (commented):
Good progress, two new concerns, and some of the older ones are still open.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+        />
+        <Input
+          id="directory-path"
+          label="Full path"
+          type="text"
+          value={targetPath}
+          onChange={(e) => {
+            setTargetPath(e.target.value);
+          }}
+          placeholder={"Directory path (optional)"}
```
edlerd:
```suggestion
          placeholder="Directory path"
```
We have an established pattern for required fields and should not say optional in placeholder text.

src/pages/instances/FileExplorerBreadcrumb.tsx
edlerd: Also the placement is not responsive for large nested structures. See how below the directory just ends and the nested directories are hidden

image

image

src/pages/instances/forms/FileExplorerCreateModal.tsx
Kxiru: I'm confused, all of these suggestions have already been addressed. Let me check my Repo...

Kxiru: Directory name input already has autofocus, and I have added `        <Input type="submit" hidden value="Hidden input" />` as requested. Please double check.

src/pages/instances/forms/FileExplorerCreateModal.tsx
edlerd: The placeholder was changed to "My directory" [in the source](https://github.com/canonical/lxd-ui/pull/2101/changes#diff-c074fd019699b5dedc443e7360657a309b374149022520571524a463faae2740R83). A pull will not change that ;)

src/pages/instances/FileExplorerBreadcrumb.tsx
Kxiru: Surely nothing should happen? Because you are not navigating away from the current directory. Do you want something to happen here?

src/pages/instances/forms/FileExplorerCreateModal.tsx
edlerd: I see the autofocus prop, but it doesn't work. On opening the modal the input does not receive the focus.

src/pages/instances/FileExplorerBreadcrumb.tsx
Kxiru: I understand that. This was a conscious decision so that when Edit is enabled, there is not any change in the structure of the screen. Both the directory path length and the search box length are the same and have a min-width. I can have the search box appear just after the end of the path, but it would mean that the UI / enter icon jumps, and is no longer in the same location as the search icon when in the Edit view. Can you confirm that this is what you would like to see?

src/pages/instances/FileExplorerBreadcrumb.tsx
edlerd: Expectation is for it to go back to read mode.

src/pages/instances/FileExplorerBreadcrumb.tsx
edlerd: I am not sure on the correct solution. Cutting long paths in the breadcrumb as we do now, is not right. Especially if the screen has enough space as in the screenshots above. The full path would fit if the breadcrumb had more width. It should be responsive, though, and we probably need a form of truncation for too long paths.

src/pages/instances/FileExplorerBreadcrumb.tsx
Kxiru: Even though it is a Chevron? Okay, I will add this.

src/pages/instances/FileExplorerBreadcrumb.tsx
Kxiru: Hmm... @sophiecarroll , could you weigh in on this, please?

src/pages/instances/forms/FileExplorerCreateModal.tsx
Kxiru: @edlerd I deleted my comment as I thought you were addressing a different component, but this has still been addressed as of the last push.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+          onChange={(e) => {
+            setDirectoryName(e.target.value);
+          }}
+          placeholder="Enter new directory name"
+        />
+        <Input
+          id="directory-path"
+          label="Full path"
+          type="text"
+          value={currentPath}
```
edlerd:
```suggestion
          value={targetPath}
```
currently we can't change this value. we need to use the state and initialize the state with current path.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -80,15 +82,26 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
...
-      <div className="file-explorer-header">
-        <FileExplorerBreadcrumb currentPath={currentPath} instance={instance} />
+      <div className="upper-controls-bar">
+        <div className="search-box-wrapper">
+          <FileExplorerBreadcrumb
+            currentPath={currentPath}
+            instance={instance}
+          />
+        </div>
+        <FileExplorerCreateBtn
```
edlerd: We need wrappers around the buttons, so they don't grow on overflow of the breadcrumb.

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -80,15 +82,26 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
...
     <Row className="general">
-      <div className="file-explorer-header">
-        <FileExplorerBreadcrumb currentPath={currentPath} instance={instance} />
+      <div className="upper-controls-bar">
+        <div className="search-box-wrapper">
+          <FileExplorerBreadcrumb
+            currentPath={currentPath}
+            instance={instance}
+          />
+        </div>
```
edlerd: Styling of this is messed up, see [this example](https://lxd-ui-2101.demos.haus/ui/project/default/instance/DemoInstance/file-explorer?project=default&path=%2Fdata%2Fsafd%2Ffff%2Flong-name%2Fnested-nested-nested-nested-nested-nested-nested).

We should give more space to the breadcrumb, so it can use the whitespace on the "Directory: root / ..." line and not keep it empty and wrap too early.

Image

edlerd (changes requested):

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+          onChange={(e) => {
+            setDirectoryName(e.target.value);
+          }}
+          placeholder="Enter new directory name"
+        />
+        <Input
+          id="directory-path"
+          label="Full path"
+          type="text"
+          value={targetPath}
```
edlerd: We have to initialize this with the current directory. When opening the modal this is displayed empty, which is confusing.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+          takeFocusDelay={100}
+          value={directoryName}
+          onChange={(e) => {
+            setDirectoryName(e.target.value);
+          }}
+          placeholder="Enter new directory name"
+        />
+        <Input
+          id="directory-path"
+          label="Full path"
```
edlerd:
```suggestion
          label="Parent path"
```
A bit more precise

edlerd: If this field is unchanged, the user stays on the current page and immediately sees the new folder.

However, on changing this field, the user does not see the new folder in another chosen directory. I think after submitting, the flow should show the parent path of the new folder. So after submitting the modal, we should take the user to the path entered here. Otherwise, they can't see the newly created directory.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+          type="text"
+          value={targetPath}
+          onChange={(e) => {
+            setTargetPath(e.target.value);
+          }}
+          placeholder={"Directory path"}
+          help="Leave blank to create within the current directory."
```
edlerd:
```suggestion
```
This can be simplified by defaulting this to the current directory. See other comment.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+        />
+        <Input
+          id="directory-path"
+          label="Full path"
+          type="text"
+          value={targetPath}
+          onChange={(e) => {
+            setTargetPath(e.target.value);
+          }}
+          placeholder={"Directory path"}
```
edlerd:
```suggestion
```
With a default, we don't need a placeholder here.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+const FileExplorerCreateModal: FC<Props> = ({
+  close,
+  instance,
+  currentPath,
+  onSuccess,
+}) => {
+  const [directoryName, setDirectoryName] = useState("");
+  const [targetPath, setTargetPath] = useState("");
```
edlerd:
```suggestion
  const [parentPath, setParentPath] = useState(currentPath);
```
A bit clearer naming and use the right default.

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,177 @@
...
+    if (trimmedPath === currentPath) {
+      setIsEditing(false);
+      return;
+    }
+
+    const syntaxError = validatePathSyntax(trimmedPath);
+    if (syntaxError) {
+      notify.failure("Invalid path", new Error(syntaxError));
+      return;
+    }
```
edlerd: This should not use notification, but use the [error prop](https://canonical.github.io/react-components/?path=/story/components-input--error) of the input instead.

Image

Image

src/pages/instances/InstanceFileExplorer.tsx
```diff
@@ -80,15 +82,28 @@ const InstanceFileExplorer: FC<Props> = ({ instance }) => {
...
+          />
+          <UploadFileExplorerBtn
+            instance={instance}
+            currentPath={currentPath}
+            refreshDirectoryList={invalidateCache}
+            directoryContent={directoryContent?.metadata ?? []}
+          />
+        </div>
       </div>
+      <NotificationRow />
```
edlerd:
```suggestion
```
I think we don't need this one if we keep the directory change errors on the input component.

If we decide to keep it, we have to fix the spacing:

Image

src/sass/_file_explorer.scss
```diff
@@ -15,12 +15,34 @@
...
-  display: inline-block;
-}
-
-.file-explorer-header {
+.file-explorer-breadcrumb-bar {
   align-items: center;
   display: flex;
-  justify-content: space-between;
+  gap: 0.5rem;
+  max-width: 25rem;
```
edlerd: This should take more of the available width and not break the line so early.

Image

Instead, it can go all the way until the line downwards from the "Logs" tab:

Image

But we must be careful, as this should still be responsive, and the buttons should stay in line with the directory.

src/sass/_file_explorer.scss
```diff
@@ -15,12 +15,34 @@
...
+  }
+
+  .p-breadcrumbs {
+    overflow: hidden;
+  }
+}
+
+.file-explorer-path-input {
+  margin-bottom: 0;
+  max-width: 25rem;
```
edlerd: This is not using the available space efficiently. In some extreme case light the one below the input could be wider to make editing easier:

Image

But we need to take care so it also looks good for short paths as those are the majority.

src/pages/instances/forms/FileExplorerCreateModal.tsx
edlerd: We already have validation of path input in the breadcrumb edit. Can we make that reusable to ensure the entered path here is valid?

We should not allow submitting with an invalid path and showing an error after submitting. The error should be on the field while editing with the error prop on the input instead.

Image

Image

edlerd (changes requested):
Good progress, few concerns on spacing, streamlining the flow and error handling

omarelkashef (commented):
This is really nice. I think the modal could be clearer. How about we use a prefixed input with the current directory as the prefix and only have one input? This way we won't need the help text and the user can add the full path of the directory in one input. Also, does LXD support creating recursive directories?

Image

sophiecarroll:
Screenshot 2026-06-30 at 3 39 30 PM
From mattermost: Just looked at the file explorer demo everything looks great! Is it too late to suggest leaving the icon to the right of the search bar to be a magnifying glass when you are typing in the search box? It changes into a clickable side arrow and I think keeping it as a magnifying glass might be a little clearer (similar to google etc)

Kxiru:
> Also, does LXD support creating recursive directories?

Unfortunately not.
Let me know what you think of the new version after the next push. We made some amendments in the WT session.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const [isCreating, setIsCreating] = useState(false);
+  const [inputError, setInputError] = useState<string | undefined>(undefined);
+
+  const toastNotify = useToastNotification();
+  const navigate = useNavigate();
+
+  const handleSubmit = async (e: FormEvent) => {
+    e.preventDefault();
+
+    const basePath = parentPath.trim() || currentPath;
```
edlerd:
```suggestion
```
Let's directly rely on `parentPath`. I don't think we need the additional variable with trim and fallback here with the latest updates.

edlerd: Also best move this after the `syntaxError` check, so we do validation first and handling of the valid input after.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+      toastNotify.failure("Directory creation failed", e);
+      setIsCreating(false);
+    }
+  };
+
+  return (
+    <Modal
+      close={close}
+      title="Create directory"
+      className="create-directory-modal u-no-margin--bottom"
```
edlerd:
```suggestion
      className="create-directory-modal"
```
We don't usually apply this class to modals. Is it needed here, or can we remove it?

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+            loading={isCreating}
+          >
+            Create
+          </Button>
+        </>
+      }
+    >
+      <form onSubmit={(e) => void handleSubmit(e)}>
+        <Input type="submit" hidden value="Hidden input" />
+
```
edlerd:
```suggestion
```

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+          takeFocus
+          takeFocusDelay={100}
+          value={directoryName}
+          onChange={(e) => {
+            setDirectoryName(e.target.value);
+          }}
+          placeholder="Enter new directory name"
+        />
+        <Input
+          id="directory-path"
```
edlerd:
```suggestion
          id="parent-path"
```

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,169 @@
...
+    setIsEditing(false);
+  }, [currentPath]);
+
+  useEffect(() => {
+    if (isEditing && inputRef.current) {
+      inputRef.current.focus();
+      inputRef.current.select();
+    }
+  }, [isEditing]);
+
```
edlerd:
```suggestion
```
We don't need this due to the auto focus prop on the input doing it already.

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,169 @@
...
   instance: LxdInstance;
 }
 
 const FileExplorerBreadcrumb: FC<Props> = ({ currentPath, instance }) => {
-  const segments = currentPath.split("/").filter(Boolean);
+  const [isEditing, setIsEditing] = useState(false);
+  const [path, setPath] = useState(currentPath);
+  const [isNavigating, setIsNavigating] = useState(false);
+  const [inputError, setInputError] = useState<string | undefined>(undefined);
+  const inputRef = useRef<HTMLInputElement>(null);
```
edlerd:
```suggestion
```

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,169 @@
...
+    const trimmedPath = path.trim();
+
+    if (trimmedPath === currentPath) {
+      setInputError(undefined);
+      setIsEditing(false);
+      return;
+    }
+
+    const syntaxError = validateDirectoryPathSyntax(trimmedPath);
+    setInputError(syntaxError);
```
edlerd, line 43:
```suggestion
    if (syntaxError) {
      return;
    }
```
We must not continue navigation if there is an error here.

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,169 @@
...
+    }
+  };
+
+  if (isEditing) {
+    return (
+      <div className="file-explorer-path-editor">
+        <Input
+          type="text"
+          className="p-form-control file-explorer-path-input"
+          placeholder="Search path..."
```
edlerd:
```suggestion
```
This will always have a value, probably no placeholder needed here.

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,169 @@
...
+      setIsEditing(false);
+    }
+  };
+
+  if (isEditing) {
+    return (
+      <div className="file-explorer-path-editor">
+        <Input
+          type="text"
+          className="p-form-control file-explorer-path-input"
```
edlerd:
```suggestion
          className="file-explorer-path-input"
```
This class seems unused.

src/util/instances.tsx
```diff
@@ -111,3 +111,20 @@ export const getFileExplorerFileURL = (
 
   return `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instance.name)}/files?${params.toString()}`;
 };
+
+export const validateDirectoryPathSyntax = (
+  path: string,
+): string | undefined => {
+  if (!path.startsWith("/")) {
+    return "Path must be absolute (start with /).";
```
```suggestion
    return "Path must start with /";
```
A bit simpler?

src/util/instances.tsx
```diff
@@ -111,3 +111,20 @@ export const getFileExplorerFileURL = (
...
+  }
+
+  const segments = path.split("/").filter(Boolean);
+  for (const segment of segments) {
+    if (segment === "." || segment === "..") {
+      return "Relative path segments (. and ..) are not allowed.";
```
```suggestion
      return "Relative path segments . and .. are not allowed.";
```
A bit simpler?

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,169 @@
...
+      setPath(currentPath);
+      setInputError(undefined);
+      setIsEditing(false);
+    }
+  };
+
+  if (isEditing) {
+    return (
+      <div className="file-explorer-path-editor">
+        <Input
```
edlerd, line 76: One styling issue was discovered on QA. In the error state, the search icon moves a little bit down. The icon should stay up, aligned with the input:

Image

edlerd (commented):
QA and responsiveness mostly LGTM. A few minimal issues or nitpicks that should be quick to resolve.

sophiecarroll:
1) After you click "create", I think the most expected behavior is to stay in the directory where you created the new one, so that on creation you see your new directory in the context of where you just created it, inside of automatically landing inside of it (eg. when you make a new folder in Google Drive, Finder, etc.). Example:
Screenshot 2026-07-08 at 10 51 33 AM

2) Going forward, we should make sure to include "* required field" on forms where we use asterisks (See example below)
Screenshot 2026-07-08 at 11 07 58 AM

3) the color of "Create network" button is green, and "Create directory" is clear. I'm going to think about this for a second and update with thoughts after my next meeting.

Questions:
- Can you create a symlink from inside LXD (UI or CLI)? Or is it only able to be uploaded from a symlink file on your computer?
- What's the history on calling this type of object a directory and not a folder? Not in the scope of this PR just curious
- After you create a new directory, can you rename it? Can we move files or folders to different locations after we create them? eg. I upload a pdf and then want to move it into a folder

edlerd:
> 1. After you click "create", I think the most expected behavior is to stay in the directory where you created the new one

Yes, I agree. We changed this yesterday because the user can change the parent directory in the "create directory" modal. I assumed we should navigate to the parent directory used. It could be the current one (if parent is left unchanged in the modal) or a different one (if the user changed the parent path in the modal). We should not move into the newly created directory but to the parent of it in my opinion.

src/pages/instances/forms/FileExplorerCreateModal.tsx
Kxiru: I think it can be removed! I don't remember at least why I added it, perhaps I was targeting something else.

src/pages/instances/FileExplorerBreadcrumb.tsx
Kxiru: How do you find these? :sob:

src/util/instances.tsx
Kxiru: I wonder if they need to be in single quotes?

Kxiru:
@sophiecarroll 
> Can you create a symlink from inside LXD (UI or CLI)? Or is it only able to be uploaded from a symlink file on your computer?

Within the UI you can only upload symlinks, though within the instance terminal, you can create symlinks inside the terminal, of course. Via the CLI I assume it's much the same.

> What's the history on calling this type of object a directory and not a folder? Not in the scope of this PR just curious

I believe this is the technical convention. Either works, but from the API layer we retrieve the 3 types, directory, file and symlink, so it is best to keep it the same as the backend/CLI for consistency.

> After you create a new directory, can you rename it? Can we move files or folders to different locations after we create them? eg. I upload a pdf and then want to move it into a folder

As of right now in the file explorer, you cannot rename or move the files/folders. You can however, do this within the terminal tab of the instance.

Kxiru:
> Going forward, we should make sure to include "* required field" on forms where we use asterisks (See example below)

It might be worth making a task for this, @sophiecarroll to apply this new convention across all instances of forms within the UI.

I'll have the UI navigate to the parent folder of where the new directory was created. Please review!

src/util/instances.tsx
edlerd: You mean `Relative path segments '.' and '..' are not allowed.`? I think that might look a bit too much special characters.

edlerd:
> > What's the history on calling this type of object a directory and not a folder? Not in the scope of this PR just curious
> 
> I believe this is the technical convention. Either works, but from the API layer we retrieve the 3 types, directory, file and symlink, so it is best to keep it the same as the backend/CLI for consistency.


Sophie and I had a chat on this yesterday as well. It seems "folder" is much more used terminology. Ubuntu and macOS use it. Also, it is shorter, so the button caption would be a bit lighter. Shall we make a rename over all occurrences in the UI for it? Though I think @sophiecarroll wanted to do a bit more research on it, and I think she should have the decision on this one.

sophiecarroll:
> > Going forward, we should make sure to include "* required field" on forms where we use asterisks (See example below)
> 
> It might be worth making a task for this, @sophiecarroll to apply this new convention across all instances of forms within the UI.
> 
> I'll have the UI navigate to the parent folder of where the new directory was created. Please review!

Sounds good. I'll create a ticket to add this to all the instances of *, and are you good to implement it in this create new directory instance?

Other than it looks good landing in the parent directory now!

sophiecarroll:
> > > What's the history on calling this type of object a directory and not a folder? Not in the scope of this PR just curious
> > 
> > 
> > I believe this is the technical convention. Either works, but from the API layer we retrieve the 3 types, directory, file and symlink, so it is best to keep it the same as the backend/CLI for consistency.
> 
> Sophie and I had a chat on this yesterday as well. It seems "folder" is much more used terminology. Ubuntu and macOS use it. Also, it is shorter, so the button caption would be a bit lighter. Shall we make a rename over all occurrences in the UI for it? Though I think @sophiecarroll wanted to do a bit more research on it, and I think she should have the decision on this one.

This is an interesting discussion because I think both terms have pros–I'm going to do a bit more research, and I asked Amy's opinion on whether it's best to prefer a term more industry standard vs. more aligned with CLI. This is also why CLI is so interesting for UX because it has a big impact on UI organization!

sophiecarroll:
> > > > What's the history on calling this type of object a directory and not a folder? Not in the scope of this PR just curious
> > > 
> > > 
> > > I believe this is the technical convention. Either works, but from the API layer we retrieve the 3 types, directory, file and symlink, so it is best to keep it the same as the backend/CLI for consistency.
> > 
> > 
> > Sophie and I had a chat on this yesterday as well. It seems "folder" is much more used terminology. Ubuntu and macOS use it. Also, it is shorter, so the button caption would be a bit lighter. Shall we make a rename over all occurrences in the UI for it? Though I think @sophiecarroll wanted to do a bit more research on it, and I think she should have the decision on this one.
> 
> This is an interesting discussion because I think both terms have pros–I'm going to do a bit more research, and I asked Amy's opinion on whether it's best to prefer a term more industry standard vs. more aligned with CLI. This is also why CLI is so interesting for UX because it has a big impact on UI organization!

I've done some more research on directory vs. folder and talked to Amy, and it seems like sticking to directory if that is the term used in CLI makes sense for consistency. I will also validate that this makes sense in testing

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    setInputError(undefined);
+
+    const trimmedParentPath = parentPath.trim();
+    const trimmedDirectoryName = directoryName.trim();
+
+    const fullPath =
+      trimmedParentPath === "/"
+        ? `/${trimmedDirectoryName}`
+        : `${trimmedParentPath}/${trimmedDirectoryName}`;
```
edlerd:
```suggestion
      trimmedParentPath.endsWith("/")
        ? `${trimmedParentPath}${trimmedDirectoryName}`
        : `${trimmedParentPath}/${trimmedDirectoryName}`;
```
This is a bit safer, the user might enter the parent path with a trailing slash.

src/pages/instances/forms/FileExplorerCreateModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+    setIsCreating(true);
+    try {
+      await createInstanceDirectory(instance, fullPath);
+      toastNotify.success(
+        `Directory ${trimmedDirectoryName} created successfully.`,
+      );
+      navigate(getFileExplorerDirectoryURL(trimmedParentPath, instance));
+      onSuccess();
+    } catch (e) {
+      toastNotify.failure("Directory creation failed", e);
```
edlerd: The toast error is not visible, the toast will be behind the modal.

Let's add a notification row within the modal instead of as a toast. We need the notify helper here as well, and we need to take care of clearing the notify error on close. We still need the toast helper for the success case.

Image

edlerd (commented):
Good work, some fine tuning on the error handling due to the changed handling of the modal not closing on errors (which is an improvement in itself).

src/pages/instances/forms/FileExplorerCreateModal.tsx
Kxiru: I think this line in general makes "//" a valid path upon directory creation. Should we have a special case when the user just wants to make a directory in root? ("/")

sophiecarroll (approved):

src/pages/instances/forms/FileExplorerCreateModal.tsx
edlerd: The root case is already handled as it also ends with a slash.

True your concern about double slash, users might also enter invalid characters. I wonder if we need more validation.

edlerd (commented):
Triggered a re-run of the failing tests. If they fail again, please check accordingly which ones broke.

edlerd (approved):
LGTM, thanks :)

Only tiny thing left regarding the extended validation of the directory name input. But we can also do that in the follow up PR already lined up.

Kxiru:
> Triggered a re-run of the failing tests. If they fail again, please check accordingly which ones broke.

Heya David!
I'm not sure which specific tests you reran, but at present, the Replicator, Storage Bucket and Network Peering tests are failing across two tests. I believe they are unrelated, however I will rebase on main and observe once more.

src/pages/instances/forms/FileExplorerCreateModal.tsx
Kxiru: We could certainly add a case to prevent consecutive slashes.

PR #2106, opened 2026-06-27
Feat: mobile & smaller screen image selector

vosdev:
## Done

- Make the image selector usable on medium, small and mobile screens by converting to a table view.
- Filters behind a toggle/button
- Search always available

Fixes usability on <1090 pixels width screens

## QA

1. Run the LXD-UI:

## Screenshots

Image selector on mobile (No black bars <500px)

image

little over 500px width:

image

Filters available behind a toggle/button:

image

792 pixels width:

image

Previous setup at 792 pixels, only showing 5 out of 8 columns, no select button/filters:

image

### DIFF
--- a/src/pages/images/ImageSelector.tsx
+++ b/src/pages/images/ImageSelector.tsx
@@ -3,6 +3,7 @@ import {
   Button,
   CheckboxInput,
   Col,
+  Icon,
   MainTable,
   Modal,
   Notification,
@@ -12,6 +13,7 @@ import {
   Select,
   Spinner,
 } from "@canonical/react-components";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
 import type { LxdImageType, RemoteImage } from "types/image";
 import { capitalizeFirstLetter } from "util/helpers";
 import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
@@ -53,7 +55,10 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
   const [hideRemote, setHideRemote] = useState(false);
   const [error, setError] = useState("");
   const [hideError, setHideError] = useState(false);
+  const [showFilters, setShowFilters] = useState(false);
   const { project } = useParams<{ project: string }>();
+
+  const isCardView = useIsScreenBelow(1090);
 
   const { data: settings, isLoading: isSettingsLoading } = useSettings();
   const { data: remoteImages, isLoading: isRemoteImagesLoading } =
@@ -168,6 +173,8 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
         onSelect(item, item.type ?? type);
       };
 
+      const onCellClick = isCardView ? undefined : selectImage;
+
       const displayRelease =
         item.os === "Ubuntu" &&
         item.release_title &&
@@ -226,54 +233,51 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
             content: item.os,
             role: "rowheader",
             "aria-label": "Distribution",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
           {
             content: displayRelease,
             role: "cell",
             "aria-label": "Release",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
           {
             content: displayVariant,
             role: "cell",
             "aria-label": "Variant",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
           {
             content: itemType,
             role: "cell",
             "aria-label": "Type",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
           {
-            className: "u-hide--small u-hide--medium",
             content: item.aliases.split(",").pop(),
             title: item.aliases.split(",").pop(),
             role: "cell",
             "aria-label": "Alias",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
           {
             content: getSource(),
             role: "cell",
             "aria-label": "Source",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
           {
-            className: "u-hide--small u-hide--medium",
             content: item.cached ? "Cached" : "Remote",
             role: "cell",
             "aria-label": "Cached",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
           {
-            className: "u-hide--small u-hide--medium",
             content: (
               <Button
                 onClick={selectImage}
                 type="button"
-                dense
+                dense={!isCardView}
                 className="u-no-margin--bottom"
                 appearance={
                   item.cached || item.server === LOCAL_IMAGE
@@ -286,7 +290,7 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
             ),
             role: "cell",
             "aria-label": "Action",
-            onClick: selectImage,
+            onClick: onCellClick,
           },
         ],
         sortData: {
@@ -307,21 +311,109 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
     {
       content: "Alias",
       sortKey: "alias",
-      className: "u-hide--small u-hide--medium",
     },
     {
       content: "Source",
     },
     {
-      className: "u-hide--small u-hide--medium",
       content: "Cached",
     },
     {
-      className: "u-hide--small u-hide--medium",
       content: "",
       "aria-label": "Actions",
     },
   ];
+
+  const filters = (
+    <div className="image-select-filters">
+      <Select
+        label="Distribution"
+        name="distribution"
+        onChange={(v) => {
+          setOs(v.target.value);
+          setRelease("");
+        }}
+        options={getOptionList((item: RemoteImage) => item.os)}
+        value={os}
+      />
+      <Select
+        label="Release"
+        name="release"
+        onChange={(v) => {
+          setRelease(v.target.value);
+        }}
+        options={getOptionList(
+          (item) => item.release,
+          (item) => item.os === os,
+        )}
+        value={release}
+        disabled={os === ""}
+      />
+      <Select
+        label="Variant"
+        name="variant"
+        onChange={(v) => {
+          setVariant(v.target.value);
+        }}
+        options={[
+          {
+            label: "Any",
+            value: ANY,
+          },
+        ].concat(
+          variantAll
+            .filter((item) => Boolean(item))
+            .map((item) => {
+              return {
+                label: item ?? "",
+                value: item ?? "",
+              };
+            }),
+        )}
+        value={variant}
+      />
+      <Select
+        label="Architecture"
+        name="architecture"
+        onChange={(v) => {
+          setArch(v.target.value);
+        }}
+        options={archAll.map((item) => {
+          return {
+            label: item,
+            value: item,
+          };
+        })}
+        value={arch}
+      />
+      <Select
+        label="Type"
+        name="type"
+        onChange={(v) => {
+          setType(
+            v.target.value === ANY
+              ? undefined
+              : (v.target.value as LxdImageType),
+          );
+        }}
+        options={[
+          {
+            label: "Any",
+            value: ANY,
+          },
+          ...instanceCreationTypes,
+        ]}
+        value={type ?? ""}
+      />
+      <CheckboxInput
+        checked={hideRemote}
+        label="Show only cached images"
+        onChange={() => {
+          setHideRemote((prev) => !prev);
+        }}
+      />
+    </div>
+  );
 
   return (
     <Modal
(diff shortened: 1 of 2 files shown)


src/pages/images/ImageSelector.tsx
```diff
@@ -53,8 +62,18 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
   const [hideRemote, setHideRemote] = useState(false);
   const [error, setError] = useState("");
   const [hideError, setHideError] = useState(false);
+  const [showFilters, setShowFilters] = useState(false);
   const { project } = useParams<{ project: string }>();
 
+  const isBelowLarge = useIsScreenBelow(largeScreenBreakpoint);
```
kimanhou: We have a beautiful hook for this !

```
```

Please have a look at `useIsScreenBelow`, it will be simpler 😉

src/sass/_forms.scss
```diff
@@ -414,18 +414,57 @@
     min-height: 100%;
   }
 
+  @media screen and (width < 500px) {
```
kimanhou: I'm not sure we need to handle such narrow screens. But if so, you can use Vanilla's breakpoints:
```
$breakpoint-x-small	460px
```

https://vanillaframework.io/docs/settings/breakpoint-settings

kimanhou (commented):
What about :
- removing the separating line above the `Select` button in each image (`display: none` on `::after`)
- making the `Select` button not dense (line 292) and full width (`width: 100%`) for mobile
- removing `onClick: selectImage` for each cell in mobile, as it's easy to unintentionnally click when scrolling

Image

src/sass/_forms.scss
vosdev: My Pixel 10 Pro is at 412px and my Samsung S23 is at 420 or something similar. A 6.3" and 6.1" phone, so I think most phones will be this narrow

I'll look into these breakpoint settings from the framework :)

src/pages/images/ImageSelector.tsx
```diff
@@ -473,6 +509,7 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
                 headers={headers}
                 rows={rows}
                 paginate={null}
+                responsive={screenSize === "small"}
```
edlerd:
```suggestion
                responsive
```
I don't think we need a conditional here. The responsive flag internally checks screen size, no?

src/sass/_forms.scss
edlerd: Breakpoint is slightly different, but it's probably okay to use `@include mobile {` here, which internally uses the small breakpoint.

src/pages/images/ImageSelector.tsx
```diff
@@ -307,126 +323,131 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
       content: "",
       "aria-label": "Actions",
     },
   ];
 
+  const filters = (
+    <div
+      className={classnames("image-select-filters", {
+        "image-select-filters--small": screenSize === "small",
+        "image-select-filters--medium": screenSize === "medium",
```
edlerd:
```suggestion
```
We should do this with CSS breakpoints instead. Avoid adding classes for styling specific dimensions. Such as `@include mobile {` and `@include medium {`.

src/sass/_forms.scss
```diff
@@ -414,18 +414,57 @@
...
+  .image-select-filters--small {
+    column-gap: $sph--large;
+    display: grid;
+    grid-template-columns: repeat(2, minmax(0, 1fr));
+    height: auto;
+    margin-bottom: $spv--large;
+    margin-right: 0;
+  }
+
+  .image-filter-toggle {
```
edlerd, line 459: Very nice idea to use the toggle here.

src/pages/images/ImageSelector.tsx
```diff
@@ -307,126 +323,131 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
-              checked={hideRemote}
-              label="Show only cached images"
-              onChange={() => {
-                setHideRemote((prev) => !prev);
-              }}
-            />
-          </div>
-        </Col>
-        <Col size={9}>
+        {screenSize === "large" && <Col size={3}>{filters}</Col>}
```
edlerd: Good idea to extract the filters into a dedicated const. I wanted to suggest using more composition pattern instead of the nested conditionals here and below. But I think that will make it not simpler. So probably fine as is.

edlerd (commented):
Thanks for your contribution! A few thoughts on the changeset can be found below.

src/pages/images/ImageSelector.tsx
vosdev: Got it!

I did have an issue with the `useIsScreenBelow()`. Large is defined as 1200, where in css it's `$breakpoint-large: 1090px;`. Some parts of lxd-ui use the useIsScreenBelow(largeScreenBreakpoint) and others make use of the css value `$breakpoint-large`. 

Working on small screen support i've been playing a lot with my screen size and this discrepancy has caused me to really scratch my head a few times. Is this difference intentional or by accident?

It would really make things a lot easier for my other branches if they were both 1090.

(And the default vanilla framework value is 1036px haha)

kimanhou: > Large is defined as 1200, where in css it's $breakpoint-large: 1090px;. Some parts of lxd-ui use the useIsScreenBelow(largeScreenBreakpoint) and others make use of the css value $breakpoint-large.

I am looking at the breakpoint inconsistency, it is indeed annoying.

vosdev: Oh what a timing, I just opened up my editor to continue. I will wait a bit :)

kimanhou: Actually it's gonna stay as it is.
If you want to use 1200px as a breakpoint, you can use `useIsScreenBelow(largeScreenBreakpoint)` in the code and in `@media screen and (min-width: 1200px)` or `@media screen and (max-width: 1199px)` CSS.
If you want to use 1090px as a breakpoint, you can use `useIsScreenBelow(1090)` in the code and in @media screen and `@include large` or `@include mobile-and-tablet ` CSS.

vosdev:
Sorry for the delay!

Your feedback has been put into the code. This results in slight changes in terms of breakpoints, but i've tested numerous resolutions between 400 and 1100 and everything seems to work as expected :)

src/pages/images/ImageSelector.tsx
```diff
@@ -307,126 +323,125 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+          {
+            label: "Any",
+            value: ANY,
+          },
+          ...instanceCreationTypes,
+        ]}
+        value={type ?? ""}
+      />
+      <CheckboxInput
+        aria-label="Only show cached images"
```
kimanhou: This is copy paste code so not your fault but could you please remove the `aria-label` ? 🙏 `CheckboxInput` internally handles it already

src/pages/images/ImageSelector.tsx
```diff
@@ -307,126 +323,125 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
-              checked={hideRemote}
-              label="Show only cached images"
-              onChange={() => {
-                setHideRemote((prev) => !prev);
-              }}
-            />
-          </div>
-        </Col>
-        <Col size={9}>
+        {screenSize === "large" && <Col size={3}>{filters}</Col>}
```
kimanhou: If you agree to use Vanilla's breakpoints, you can handle this with `className="u-hide--small u-hide--medium`
https://vanillaframework.io/docs/utilities/hide

src/pages/images/ImageSelector.tsx
```diff
@@ -307,126 +323,125 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+        <Col size={screenSize === "large" ? 9 : 12}>
```
kimanhou: If you agree to use Vanilla's breakpoints, you can handle this with `<Col size={9} small={12} medium={12}>`

src/pages/images/ImageSelector.tsx
```diff
@@ -53,8 +60,18 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
   const [hideRemote, setHideRemote] = useState(false);
   const [error, setError] = useState("");
   const [hideError, setHideError] = useState(false);
+  const [showFilters, setShowFilters] = useState(false);
   const { project } = useParams<{ project: string }>();
 
+  const isCardView = useIsScreenBelow(1090);
```
kimanhou, line 61: Any reason why 1090 specifically ?

src/pages/images/ImageSelector.tsx
```diff
@@ -53,8 +60,18 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+  const isMobile = useIsScreenBelow(smallScreenBreakpoint);
```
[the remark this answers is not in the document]
vosdev: I've tested at 1089, 1090 and 1091 pixels width and everything behaves as expected.
- Compact/table view at 1089px
- Traditional view at 1090px width. No card layout like Copilot claims
- Traditional view at 1091px width

src/pages/images/ImageSelector.tsx
```diff
@@ -53,8 +60,18 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+  let screenSize: ScreenSize = "large";
+  if (isMobile) {
+    screenSize = "small";
+  } else if (isCardView) {
+    screenSize = "medium";
+  }
```
kimanhou: I don't think we need to introduce a new type for handling different screen sizes.

```suggestion
  const isLargeScreen = !isCardView && !isMobile;
```

src/pages/images/ImageSelector.tsx
```diff
@@ -168,6 +185,8 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
         onSelect(item, item.type ?? type);
       };
 
+      const cellClick = isCardView ? undefined : selectImage;
```
kimanhou: Could you please rename this `onCellClick` ?

src/pages/images/ImageSelector.tsx
```diff
@@ -286,7 +302,7 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
             ),
```
kimanhou, line 286: If it's card view, should the "Select" button take the whole width of the card ?

Image

src/pages/images/ImageSelector.tsx
vosdev: Playing with resolutions on several screens i've noticed that 1090 is a breakpoint already used for for example the sidepanel on the instance list. 

It doesn't have to be 1090 specifically. I just felt around 1000-1100 is a fitting width to switch to cardView, based on the data in the table. I went with 1090 because it was already being used by the sidepanel.

With largeScreenBreakpoint it feels like we're jumping back to table view a bit too late. But if you want to stick to the existing ScreenBreakpoints then we can go 1200 with the largeScreenBreakpoint instead

src/pages/images/ImageSelector.tsx
vosdev: Yes :)

src/pages/images/ImageSelector.tsx
kimanhou: Understood ! `InstanceDetailPanel` uses `u-hide--medium u-hide--small` so we can do the same here.

vosdev: Using those classes for almost everything now. Some logic still uses the isCardView. It looks a lot cleaner now :)

Also FYI the 1090 comes from: https://github.com/canonical/lxd-ui/blob/main/src/sass/_settings.scss

kimanhou: > Also FYI the 1090 comes from: https://github.com/canonical/lxd-ui/blob/main/src/sass/_settings.scss

Yes, I saw.

src/sass/_forms.scss
```diff
@@ -428,18 +428,65 @@
...
+    }
+
+    @include mobile {
+      column-gap: $sph--large;
+      display: grid;
+      grid-template-columns: repeat(2, minmax(0, 1fr));
+      height: auto;
+      margin-bottom: $spv--large;
+      margin-right: 0;
+    }
```
kimanhou: We can avoid repetition here

```suggestion
  @include mobile-and-tablet {
    column-gap: $sph--large;
    display: grid;
    height: auto;
    margin-bottom: $spv--large;
    margin-right: 0;
  }

  @include mobile {
    grid-template-columns: repeat(2, minmax(0, 1fr));
  }

  @include medium {
    grid-template-columns: repeat(3, minmax(0, 1fr));
  }
```

Or use `@include mobile, @include medium` instead of `@include mobile-and-tablet` if you want more clarity.

kimanhou (commented):
Great work ! Just a tiny nitpicking on CSS, otherwise it looks very good.

src/sass/_forms.scss
vosdev: Done :)

src/pages/images/ImageSelector.tsx
```diff
@@ -307,126 +311,121 @@ const ImageSelector: FC<Props> = ({ onSelect, onClose }) => {
...
+          );
+        }}
+        options={[
+          {
+            label: "Any",
+            value: ANY,
+          },
+          ...instanceCreationTypes,
+        ]}
+        value={type ?? ""}
```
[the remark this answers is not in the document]
vosdev: Fixed!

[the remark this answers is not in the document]
kimanhou: Ah no I am not sure it should have changed.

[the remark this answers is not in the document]
kimanhou: Could you please put it back ? 🙏
That's why I resolved the conversation.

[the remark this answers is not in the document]
vosdev: And it's back!

kimanhou (approved):

PR #2113, opened 2026-06-30
test(Network peering): fix CI

kimanhou:
## Done

- Fix network local peering tests
```
  2 failed
    [chromium:lxd-5.21-edge:clustered] › tests/network-local-peering-clustered.spec.ts:43:3 › Network Local Peering › Create mutual peering between two OVN networks 
    [chromium:lxd-5.21-edge:clustered] › tests/network-local-peering-clustered.spec.ts:80:3 › Network Local Peering › Create non-mutual peering between two OVN networks 
```

## QA

1. Run the LXD-UI:
    - Rely on CI

## Screenshots

image

### DIFF
--- a/tests/helpers/network.ts
+++ b/tests/helpers/network.ts
@@ -62,9 +62,17 @@ export const prepareCreateNetwork = async (
   options: NetworkOptions = {},
 ) => {
   const { hasMemberSpecificParents, networkACL, uplink, ipv4, ipv6 } = options;
-  await gotoURL(page, "/ui/");
-  await page.getByRole("button", { name: "Networking" }).click();
-  await page.getByRole("link", { name: "Networks", exact: true }).click();
+  const networksLink = page.getByRole("link", {
+    name: "Networks",
+    exact: true,
+  });
+
+  // If the link isn't visible, the accordion is collapsed
+  if (!(await networksLink.isVisible())) {
+    await page.getByRole("button", { name: "Networking" }).click();
+  }
+
+  await networksLink.click();
   await page.getByRole("button", { name: "Create network" }).click();
   await page.getByRole("heading", { name: "Create a network" }).click();
   await page.getByRole("button", { name: "Type" }).click();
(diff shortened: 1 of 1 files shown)


tests/helpers/network.ts
```diff
@@ -213,7 +214,7 @@ export const getNetworkLink = async (page: Page, network: string) => {
 
 export const submitCreateNetwork = async (page: Page, network: string) => {
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForTimeout(5000);
+  await dismissNotification(page, `Network ${network} created.`);
```
edlerd: On 5.21 and before, we have the sync create network api call. That call often gets cancelled by the browser because the underlying network on the host (when running lxd on localhost) changes.

That is why we do not rely on the message here, but on a timeout instead. The message will more often not be displayed in CI. In newer versions such as on main or on the 6 track this would work though.

Then again, I think this is not addressing the underlying problem.

edlerd: The screen below taken from [this run](https://github.com/canonical/lxd-ui/actions/runs/28446245621/job/84296275930) shows the error on creating the network forward on the ovn network.

In the screencast we can see the uplink is created successfully, but also creating the ovn network fails or creates the ovn network in an error state.

image

omarelkashef: > On 5.21 and before, we have the sync create network api call. That call often gets cancelled by the browser because the underlying network on the host (when running lxd on localhost) changes.
> 
> That is why we do not rely on the message here, but on a timeout instead. The message will more often not be displayed in CI. In newer versions such as on main or on the 6 track this would work though.
> 
> Then again, I think this is not addressing the underlying problem.

Why does this happen even when we run the tests against a remote lxd inside a VM?

tests/helpers/network.ts
```diff
@@ -214,9 +224,10 @@ export const getNetworkLink = async (page: Page, network: string) => {
 
 export const submitCreateNetwork = async (page: Page, network: string) => {
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForTimeout(5000);
   const networkLink = await getNetworkLink(page, network);
-  await expect(networkLink).toBeVisible();
+  await expect(networkLink).toBeVisible({ timeout: 15000 });
```
omarelkashef: This defaults to the test timeout already. Do we need to override it?

tests/helpers/network.ts
edlerd: > > On 5.21 and before, we have the sync create network api call. That call often gets cancelled by the browser because the underlying network on the host (when running lxd on localhost) changes.
> > That is why we do not rely on the message here, but on a timeout instead. The message will more often not be displayed in CI. In newer versions such as on main or on the 6 track this would work though.
> > Then again, I think this is not addressing the underlying problem.
> 
> Why does this happen even when we run the tests against a remote lxd inside a VM?

Good question. For one, I am not 100% sure tests in LXD-CI run against a VM. For two, I assume the network could bubble up to the host when the VM is running on localhost.

PR #2117, opened 2026-07-01
fix(Navigation): highlight "Networks" for unmanaged networks

kimanhou:
## Done

- fix(Navigation): highlight "Networks" for unmanaged networks

## QA

1. Run the LXD-UI:
    - Go to the networks list page.
    - Click on an unmanaged network name (column Managed = No). Once on the detail page of an unmanaged network, "Networks" in the side navigation should be highlighted and the accordion open.
    - Make sure it's not broken for managed networks.

## Screenshots

Unmanaged
image

Managed
image

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -28,11 +28,12 @@ import { useLoggedInUser } from "context/useLoggedInUser";
 import { useSettings } from "context/useSettings";
 import { useIsScreenBelow } from "context/useIsScreenBelow";
 import { useIsClustered } from "context/useIsClustered";
-import { getReportBugURL } from "util/reportBug";
 import { AUTH_METHOD, authIcon } from "util/authentication";
+import { unmanagedNetworkDetailRoute } from "util/networks";
+import { ALL_PROJECTS } from "util/projects";
+import { getReportBugURL } from "util/reportBug";
 import DocLink from "components/DocLink";
 import AuthenticationTlsStepper from "./AuthenticationTlsStepper";
-import { ALL_PROJECTS } from "util/projects";
 
 const initialiseOpenNavMenus = (location: Location) => {
   const openPermissions = location.pathname.includes("/permissions/");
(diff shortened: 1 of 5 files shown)


src/components/Navigation.tsx
```diff
@@ -313,6 +317,9 @@ const Navigation: FC = () => {
                         <NavAccordion
                           baseUrls={[
                             `${ROOT_PATH}/ui/project/${encodeURIComponent(projectName)}/network`,
+                            ...(isUnmanagedNetworkDetail
+                              ? [location.pathname]
+                              : []),
```
edlerd: This approach is a bit backwards. We should rather have the template url here, and make the `NavAccordion` use the matcher.
```suggestion
                            `${ROOT_PATH}/ui/project/:project/member/:member/network/:name`,
```

edlerd (commented):
Thanks for picking this up. I'd slightly change to keep the matching contained in a single place. Then it should be good to go.

src/components/Navigation.tsx
kimanhou: I'm not following but my last changes seem to work, I don't know why 🤔

src/components/Navigation.tsx
```diff
@@ -313,6 +314,7 @@ const Navigation: FC = () => {
                         <NavAccordion
```
edlerd, line 314: We need to extend the `isActive` check in `NavAccordion` to use the matching function instead of the current `.includes` string comparison.

edlerd (commented):
Mostly LGTM, but the highlight fails in a cluster for unmanaged networks now.

src/components/Navigation.tsx
edlerd: I doesn't work for me, see comments below. Resolving this thread to continue it below.

src/components/Navigation.tsx
kimanhou: Okay I understood what you meant. I extended `NavAccordion` and `NavLink` to use `matchPath`. I'm not sure we should completely remove `.includes` tho.

edlerd (approved):
LGTM, thanks for spotting and fixing this!

PR #2118, opened 2026-07-01
WD-37548 Navigation - Unify experience between clustered and non-clustered environments

kimanhou:
## Done

- Navigation: replace "Server" by "Clustering" for non clustered environments

## QA

1. Run the LXD-UI:
    - Connect to a non clustered server. Make sure "Clustering" appears in the navigation.
    - Go thru each page under "Clustering" and make sure they work fine.
    - Connect to a clustered server and make sure nothing is broken.

## Screenshots

Server
image

Groups
image

Cluster links
image

Placement
image

Replicators
image


src/pages/cluster/ClusterMemberList.tsx
```diff
@@ -10,27 +12,74 @@ import {
...
+    return (
+      <BaseLayout
+        mainClassName="cluster-list"
+        title={
+          <HelpLink
+            docPath="/explanation/clustering/"
+            title="Learn more about clustering"
+          >
+            Server
+          </HelpLink>
```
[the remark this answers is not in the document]
kimanhou: Yeah but we haven't found a better way for now.

kimanhou:
Need to fix tests so converting this back to draft

omarelkashef (commented):
Thank you for the PR. A few comments below

1- These tabs don't look right.
Image

2- The text "To form a multi-node cluster and manage multiple members, you first need to enable clustering." does not need to wrap since the Notification takes the whole width. 
3-  I feel like the Notification component takes too much height. 
4- The page title is still Server which is different from the nav section name (Members). I don't think we do this anywhere else.
5- I am not sure this page can be found easy under Clustering -> Members. If a server is not clustered, I do not think a user will think to look into the Clustering section. 

Image

edlerd:
> Thank you for the PR. A few comments below
> 
> 1- These tabs don't look right. 

+1 I think there is some css missing

> 2- The text "To form a multi-node cluster and manage multiple members, you first need to enable clustering." does not need to wrap since the Notification takes the whole width. 

I think this is correct. The text shouldn't be too wide, as then it gets hard to read. But maybe the notification itself should be limited in width as well. Currently the "Learn more about clustering" in the footer is very far from the rest of the notification content.

> 3- I feel like the Notification component takes too much height.

Yes, the enable clustering button could use a different appearance and be moved to the footer, maybe?

> 4- The page title is still Server which is different from the nav section name (Members). I don't think we do this anywhere else.

I agree, though I think the notification immediately clarifies this, so maybe ok as is.

> 5- I am not sure this page can be found easy under Clustering -> Members. If a server is not clustered, I do not think a user will think to look into the Clustering section.

This is a valid concern on discoverability of the server hardware. If a user is moving between microcloud and unclustered LXD, they would expect it here, so no problem. But new users or users that only ever work with unclustered LXD might have trouble finding it. Maybe something to test with users? Or do you have ideas that could resolve this?

kimanhou:
@omarelkashef 

> 1- These tabs don't look right.

Fixed

> 2- The text "To form a multi-node cluster and manage multiple members, you first need to enable clustering." does not need to wrap since the Notification takes the whole width.

As David said, it's the default setting for paragraphs.

> 3- I feel like the Notification component takes too much height.
> Yes, the enable clustering button could use a different appearance and be moved to the footer, maybe?

It's a bit weird for the footer to have both "Learn more about clustering" and "Enabled clustering" because one is an external link, while the other one is a button. But I can try.

> 4- The page title is still Server which is different from the nav section name (Members). I don't think we do this anywhere else.

Yeah, we didn't find a better solution than this.

> 5- I am not sure this page can be found easy under Clustering -> Members. If a server is not clustered, I do not think a user will think to look into the Clustering section.

Same as 4

kimanhou:
@omarelkashef @edlerd I put "Enable clustering" in the notification's actions. What do we think ? (It does complicate the code a bit)

image

src/components/Navigation.tsx
```diff
@@ -509,105 +507,87 @@ const Navigation: FC = () => {
...
+                          {[
+                            <SideNavigationItem key="members">
+                              <NavLink
+                                to={`${ROOT_PATH}/ui/cluster/members`}
+                                title="Members"
+                                onClick={softToggleMenu}
+                                className="accordion-nav-secondary"
+                              >
+                                Members
+                              </NavLink>
```
edlerd: Shall we have a distinction between Members / Server here when unclustered?

src/pages/cluster/ClusterMemberList.tsx
```diff
@@ -10,27 +12,81 @@ import {
...
+        }
+      >
+        <Row>
+          <Notification
```
edlerd: I think this could be shorter, and not so wide like in the screen below.

Image

src/components/Navigation.tsx
kimanhou: Yes, done

src/pages/placement-groups/PlacementGroupList.tsx
```diff
@@ -16,33 +16,56 @@ import {
...
+        title={
+          <HelpLink
+            docPath="/howto/cluster_placement_groups/"
+            title="Learn how to use placement groups"
+          >
+            Placement groups
+          </HelpLink>
+        }
+      >
+        <NotClusteredEmptyState text="To manage placement groups, you first need to enable clustering." />
```
edlerd: I think we should already show the "Create placement group" button here, but disable it. And make the "Enable clustering" a non-primary button. That way a users knows they can create the placement groups here later on. Otherwise the primary button changes functionality with a potential for confusion.  Wdyt?

src/pages/cluster/ClusterMemberList.tsx
```diff
@@ -11,26 +12,51 @@ import {
...
+            Cluster members
+          </HelpLink>
+        }
+      >
+        <NotClusteredEmptyState
+          text="To manage multiple members, you first
+            need to enable clustering."
+        />
+      </BaseLayout>
+    );
```
edlerd:
```suggestion
```
Can this be removed now?

src/pages/cluster/ClusterServer.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <HelpLink
+          docPath="/explanation/clustering/"
+          title="Learn more about clustering"
+        >
+          Server
+        </HelpLink>
+      }
+    >
+      <Row>
+        <Notification
```
edlerd: The notification is sticky in the header. Can we make it closeable? I think it can reappear on reload, and we could close it with a local state to allow getting more space for the hardware section below. Wdyt?

src/sass/_cluster_list.scss
```diff
@@ -100,18 +104,34 @@
...
+    top: calc(
+      0.25rem + 1.5rem + 0.75rem /* header */ + 1.5rem
+        /* panel content padding top */ + 0.5rem - 1px
+        /* notification padding-top */ + 1.5rem
+        /* notification title line height */ + 0.5rem + 1rem + 0.5rem
+        /* notification meta */ + 0.5rem - 1px
+        /* notification padding-bottom */ + 1px + 1px
+        /* notification border-top and border-bottom */ + 1.5rem
+        /* notification margin-bottom */ + $spv--x-large
+    );
```
edlerd: I think we should just have a magic value here and do the calculation now to commit a single value. Wdyt?

edlerd (commented):
Good stuff, few ideas to refine.

src/pages/placement-groups/PlacementGroupList.tsx
kimanhou: Like this ?
image

src/pages/cluster/ClusterMemberList.tsx
kimanhou: Yes it can be removed but that means that if a user hits the URL manually, without using the Navigation, they will see a "broken" screen.

image


If we don't remove it, it gives a much nicer experience:
image

src/pages/cluster/ClusterServer.tsx
kimanhou: Yes, but once closed, the hardware menu stays in the same position. I could fix it but it's not straightforward.

src/sass/_cluster_list.scss
kimanhou: Yes, but I keep this until we settle on the design. It helps me iterate quickly. I can update the one on line 55, I think it won't change now.

src/pages/cluster/ClusterMemberList.tsx
omarelkashef: Maybe we should not register the members url if unclustered.

omarelkashef (commented):
I noticed the links page flashes an empty table before it shows the empty state

https://github.com/user-attachments/assets/9ef2a9a9-6241-418b-9f0c-a626ca7ab3ec

src/sass/_cluster_list.scss
kimanhou: Done

src/pages/cluster/ClusterServer.tsx
kimanhou: Done

src/pages/placement-groups/PlacementGroupList.tsx
edlerd: No, more similar to the replicators empty state. The two buttons can be next to each other in the empty state. The green button to the left of the "enable clustering" one. The enable clustering can be removed once clustering has happened. That way the big green button in the empty state doesn't change meaning. I think it helps to unconfuse.

kimanhou:
@omarelkashef It should be fixed now, can you please retry ?

src/pages/cluster/ClusterMemberList.tsx
kimanhou: Oh yeah that's a way better idea. Done

src/pages/placement-groups/PlacementGroupList.tsx
kimanhou: Ah sorry, like this:
image

edlerd: Not quite. The proposed has a problem as the Create group will become primary after enabling clustering, which will confuse users as they see again a green button in the same location, but it does a different thing.

Let's make "Create group" primary and on the left. Enable clustering to the right of it in the default style. After enabling the cluster, the green button enables and keeps meaning. The Enable clustering will be hidden then. Everything stays in its place and keeps the same meaning that way.

kimanhou: After enabling clustering, there is a `default` group. There is no empty state for Cluster groups.

image

However, there is an empty state for placement groups:
image

edlerd: > After enabling clustering, there is a `default` group. There is no empty state for Cluster groups.

Yes, I thought we were discussing placement groups the whole time, that is what the original comment refers to :)

edlerd: But now that you mention it -- maybe we should have it work the same on all those cluster empty states. Show the "Create :entity" in green and disabled. The "enable clustering" button in the default style next to it. Seems more consistent and signals to the user -- when you come back here after clustering is enabled, this is the place to create such an entity. Wdyt?

kimanhou: There are 2 screens that don't work for unclustered environments: Groups and Placement. I keep them consistent with each other for each iteration. 

For placement, the "not clustered" state and the empty state are different from each other (not the same text), so I don't think keeping the positive button for "+ Create placement group" really matters. 

I think it's more intuitive to have the positive button for the "next" step, so that the user knows where to click.

kimanhou: But I guess replicators is different so I'll match that.

omarelkashef:
> @omarelkashef It should be fixed now, can you please retry ?

Yes fixed, thank you!

src/pages/cluster/actions/EnableClusteringBtn.tsx
```diff
@@ -17,13 +22,14 @@ const EnableClusteringBtn: FC = () => {
...
-        hasIcon
+        appearance="default"
+        className={className}
+        hasIcon={hasIcon}
         onClick={openPortal}
         disabled={!canEdit}
         title={title}
       >
-        <Icon name="plus" light />
+        {hasIcon && <Icon name="plus" light />}
```
edlerd: This icon is wrongly rendered under the light theme

Image

src/pages/cluster/ClusterLinkList.tsx
```diff
@@ -148,11 +143,27 @@ const ClusterLinkList: FC<Props> = ({ variant = "main" }) => {
...
+      <BaseLayout
+        title={
+          <HelpLink
+            docPath="/explanation/clustering/"
+            title="Learn more about clustering"
+          >
+            Cluster links
+          </HelpLink>
+        }
+      >
```
edlerd: We shouldn't redefine this here. This way it tends to get out of sync with the main heading. We can simplify this to return only the spinner and not the heading.

`    return <Spinner className="u-loader" text="Loading..." isMainComponent />;`

src/pages/cluster/ClusterMemberHardware.tsx
```diff
@@ -130,7 +134,7 @@ const ClusterMemberHardware: FC<Props> = ({ member }) => {
               </Col>
             </Row>
           ))}
-          <div className="aside">
+          <div className="aside" style={{ top: hardwareNavigationTop }}>
```
edlerd:
```suggestion
          <div className="aside">
```
This should be done in CSS. We can use a `.cluster-server:has(.cluster-server-notification) .aside` filter on the parent to target this.

src/pages/placement-groups/PlacementGroupList.tsx
```diff
@@ -16,33 +15,58 @@ import {
...
+        mainClassName="placement-groups-list"
+        title={
+          <HelpLink
+            docPath="/howto/cluster_placement_groups/"
+            title="Learn how to use placement groups"
+          >
+            Placement groups
+          </HelpLink>
+        }
+      >
```
edlerd: We should not repeat this here from below. There should be one definition of the main layout.

src/pages/cluster/ClusterGroupList.tsx
```diff
@@ -20,15 +20,41 @@ import { useClusterGroups } from "context/useClusterGroups";
...
+        mainClassName="cluster-list"
+        title={
+          <HelpLink
+            docPath="/explanation/clustering/#cluster-groups"
+            title="Learn more about cluster groups"
+          >
+            Cluster groups
+          </HelpLink>
+        }
+      >
```
edlerd: We should not redefine the layout here.

edlerd (commented):
Mostly LGTM, few concerns left from detailed QA and code review.

src/pages/cluster/actions/EnableClusteringBtn.tsx
kimanhou: Should the icon be a "+" or the icon we use for Clustering ?

src/pages/cluster/ClusterGroupList.tsx
```diff
@@ -20,15 +20,41 @@ import { useClusterGroups } from "context/useClusterGroups";
...
+        <NotClusteredEmptyState
+          text="To organize your servers into groups, you first need to enable clustering."
+          extraButton={<CreateClusterGroupBtn disabled />}
```
edlerd: There should be a bottom margin in this usage here for mobile, where the two buttons wrap:

Image

This also applies to replicators. It is fine on the placement groups.

src/pages/cluster/actions/EnableClusteringBtn.tsx
edlerd: The + icon seems correct, but the color is not matching the theme.

tests/enable-clustering.spec.ts
```diff
@@ -7,9 +7,56 @@ test("check enabling clustering", async ({ page, lxdVersion }, testInfo) => {
...
+    ),
+  ).toBeVisible();
+  await page.getByRole("button", { name: "Enable clustering" }).click();
+  await page.getByRole("button", { name: "Close" }).click();
+
+  await page.getByRole("link", { name: "Links" }).click();
+  await expect(page.getByText("Cluster links", { exact: true })).toBeVisible();
+  await expect(
+    page.getByText("This server is not clustered", { exact: true }),
+  ).not.toBeVisible();
```
edlerd:
```suggestion
```
Why do we assert this text is missing here?

tests/enable-clustering.spec.ts
```diff
@@ -7,9 +7,56 @@ test("check enabling clustering", async ({ page, lxdVersion }, testInfo) => {
...
+    ),
+  ).toBeVisible();
+  await page.getByRole("button", { name: "Enable clustering" }).click();
+  await page.getByRole("button", { name: "Close" }).click();
+
+  await page.getByRole("link", { name: "Replicators" }).click();
+  await expect(page.getByText("Replicators", { exact: true })).toBeVisible();
+  await expect(
+    page.getByText("This server is not clustered", { exact: true }),
+  ).not.toBeVisible();
```
edlerd:
```suggestion
```
Same as above comment.

edlerd (dismissed):
Good work, thanks for aligning the clustering pages! Two small non-blocking questions on the tests.

tests/enable-clustering.spec.ts
kimanhou: To be sure that unclustered environments have access to Cluster links.

tests/enable-clustering.spec.ts
kimanhou: To be sure that unclustered environments have access to Replicators.

tests/enable-clustering.spec.ts
edlerd: This is a bit backwards assertion. Maybe check the create cluster link button is visible and enabled instead?

tests/enable-clustering.spec.ts
edlerd: Should we check for the "Create link / create replicator" button to be enabled instead?

edlerd (approved):
LGTM, good work 👍

PR #2122, opened 2026-07-06
fix(project config): responsiveness for replication tab

kimanhou:
## Done

- Fix responsiveness for project configuration's replication tab

## QA

1. Run the LXD-UI:
    - Go to a project's configuration. Open ""Replication" tab.
    - Make sure the actions buttons (Promote, Demote and Clear) are correctly displayed for all screen widths.

## Screenshots

Screen > 1200px
image

Screen = 990px
image

Screen = 887px (buttons wrap)
image

Screen < 620px (mobile)
image

### DIFF
--- a/src/pages/projects/forms/ProjectReplicaForm.tsx
+++ b/src/pages/projects/forms/ProjectReplicaForm.tsx
@@ -1,18 +1,23 @@
 import { type FC } from "react";
-import { Col, Field, Notification, Row } from "@canonical/react-components";
+import {
+  Col,
+  Notification,
+  OutputField,
+  Row,
+} from "@canonical/react-components";
 import ScrollableForm from "components/ScrollableForm";
 import type { FormikProps } from "formik/dist/types";
 import ClusterLinkSelector from "pages/cluster/ClusterLinkSelector";
 import ReplicatorList from "pages/cluster/ReplicatorList";
-import PromoteProjectBtn from "pages/projects/actions/PromoteProjectBtn";
+import ClearProjectReplicaModeBtn from "pages/projects/actions/ClearProjectReplicaModeBtn";
 import DemoteProjectBtn from "pages/projects/actions/DemoteProjectBtn";
+import PromoteProjectBtn from "pages/projects/actions/PromoteProjectBtn";
 import type { LxdConfigPair } from "types/config";
 import type {
   ProjectFormValues,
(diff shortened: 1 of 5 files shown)


src/context/useIsScreenBelow.tsx
```diff
@@ -4,7 +4,8 @@ import { useListener } from "@canonical/react-components";
 
 export const smallScreenBreakpoint = 620;
 export const mediumScreenBreakpoint = 820;
-export const largeScreenBreakpoint = 1200;
+export const largeScreenBreakpoint = 1090;
+export const largeScreenBreakpoint2 = 1200;
```
edlerd: Those changes should be reverted.

src/sass/_project_configuration.scss
```diff
@@ -64,6 +64,22 @@
...
   }
 
+  .replica-mode-action-buttons {
+    margin-bottom: $spv--small;
+
+    :first-child {
+      margin-bottom: $spv--small;
+    }
+
+    @include large-2 {
```
edlerd: We don't have to use a mixin here, we can use the 1200 responsive selector as in other places, like in `@media screen and (width < 1200px) {`

src/pages/projects/forms/ProjectReplicaForm.tsx
```diff
@@ -118,11 +124,11 @@ const ProjectReplicaForm: FC<Props> = ({ formik, isEdit }) => {
...
               </output>
 
-              <div className="replica-mode-action-buttons">
-                {getModeActionButtons(formik.values.replica_mode)}
-              </div>
+              {!isSmallScreen &&
+                getModeActionButtons(formik.values.replica_mode)}
             </div>
           </Field>
+          {isSmallScreen && getModeActionButtons(formik.values.replica_mode)}
```
edlerd: I think semantically those buttons should always be rendered next to the field and not inside of it when on large screens. Can we fix the layout so it renders next to it on big screens, but below on smaller screens and is always outside?

Also there is a layout bug with the loading state, where the buttons now jump around:

Image

kimanhou: I changed the layout, it's a bit risky but it seems to work.

I also fixed the loading state.

edlerd (approved):
LGTM, good work 👍

Issue #745, opened 2024-04-17
Error report


EdwinNoorlander:
```
import{u,j as t,L as a,N as o}from"./index-b3f3eaed.js";const n=({outlet:e})=>{const{isAuthenticated:r,isAuthLoading:s}=u();return s?t.jsx(a,{}):r?e:t.jsx(o,{to:"/ui/login",replace:!0})};export{n as default};

```

edlerd:
Thanks for your report. Could you please provide some context as to which version of LXD you are using and what was the context that you were in when this happened?

PR #2125, opened 2026-07-06
feat(design system): install new design system and consume KeyboardKey

kimanhou:
## Done

- Install new design syste,
- Consume one component: `KeyboardKey` in `MountedConfirmationButton`
- Increase text font of helper text in modal to match `KeyboardKey`'s font size

## QA

1. Run the LXD-UI:
    - Go to the Instance list page.
    - If no instance is running, start one.
    - Click on the "Stop" button of a running instance row, without pressing `Shift`.
    - Make sure the confirmation modal nicely displays the new KeyboardKey for `Shift`.

## Screenshots

Before
Screenshot from 2026-07-06 11-49-40

After
image

### DIFF
--- a/src/components/MountedConfirmationButton.tsx
+++ b/src/components/MountedConfirmationButton.tsx
@@ -6,6 +6,7 @@ import {
   type PropsWithSpread,
   type SubComponentProps,
 } from "@canonical/react-components";
+import { KeyboardKey } from "@canonical/react-ds-global";
 import type { MouseEvent, ReactNode } from "react";
 import { useModal } from "context/useModal";
 
@@ -72,9 +73,9 @@ export const MountedConfirmationButton = ({
       >
         {confirmationModalProps.children}
         {showShiftClickHint && (
-          <p className="p-text--small u-text--muted u-hide--small">
+          <p className="u-text--muted u-hide--small">
             Next time, you can skip this confirmation by holding{" "}
-            <code>SHIFT</code> and clicking the action.
+            <KeyboardKey keyValue="shift" /> and clicking the action.
           </p>
         )}
       </ConfirmationModal>,
--- a/yarn.lock
+++ b/yarn.lock
@@ -27,7 +27,7 @@
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.29.7.tgz#6f0237f0f36d2e51c0570a636faed9d2d0efe629"
   integrity sha512-locTkQyKvwIEgBzVrn8693ebc97F2U8ZHjbXwDXJ5Fn2TCpNwTlKcaKLkdHop5c/icOFE7qt7Q9JC5hnKNa6Gg==
 
-"@babel/core@^7.23.9", "@babel/core@^7.28.5":
+"@babel/core@^7.23.9", "@babel/core@^7.28.0", "@babel/core@^7.28.5":
   version "7.29.7"
   resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.29.7.tgz#80c10b17248082968b57a857b91640971f2070f7"
   integrity sha512-RgHBCvtjbOK2gXSNBNIkNoEc9qoVEtau3hj8gEqKQuL3HZAibKarWFEI3Lfm6EYKkLalOh8eSrj9b+ch9H/VBA==
@@ -291,7 +291,7 @@
     "@babel/plugin-transform-modules-commonjs" "^7.27.1"
     "@babel/plugin-transform-typescript" "^7.28.5"
 
-"@babel/runtime@^7.18.6":
+"@babel/runtime@^7.18.6", "@babel/runtime@^7.28.6":
   version "7.29.7"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.29.7.tgz#12022450c45a4da6d8d8287b18a4ff2ddb23f768"
   integrity sha512-Nq8OhGWiZIZGV6hLHoyAKLLcJihP/xFeBMGJoUrxTX2psI8dCifzLhZISFb+VWS3wFMRDmCGw5R+dOySCqPLhw==
@@ -305,7 +305,7 @@
     "@babel/parser" "^7.29.7"
     "@babel/types" "^7.29.7"
 
-"@babel/traverse@^7.29.7":
+"@babel/traverse@^7.28.0", "@babel/traverse@^7.29.7":
   version "7.29.7"
   resolved "https://registry.yarnpkg.com/@babel/traverse/-/traverse-7.29.7.tgz#c47b07a41b95da0907d026b5dd894d98de7d2f2d"
   integrity sha512-EhlfNQtZ+NK22w5BM61ciuiq1m58ed33Wr1Xan//ZRTy6hgjnwyCffRYwzsGXdASJSUJ1guZILsErh1eQcl+zw==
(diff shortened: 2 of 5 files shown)


src/components/MountedConfirmationButton.tsx
```diff
@@ -74,7 +75,7 @@ export const MountedConfirmationButton = ({
         {showShiftClickHint && (
           <p className="p-text--small u-text--muted u-hide--small">
             Next time, you can skip this confirmation by holding{" "}
-            <code>SHIFT</code> and clicking the action.
+            <KeyboardKey keyValue="shift" /> and clicking the action.
```
edlerd, line 78: I think the keyboard key needs to be a small font as well. Now we are mixing 16px and 14px font sizes in one sentence. I think it would be better to align it. Is there a small font version of the keyboard key component? If not, please request it upstream or suggest that we create an override for it.

Image

vite.config.ts
```diff
@@ -57,4 +58,21 @@ export default defineConfig({
...
+        __dirname,
+        "node_modules/@canonical/react-ds-global/dist/esm/lib",
+      ),
+    },
+    server: {
+      deps: {
+        inline: ["@canonical/react-ds-global"],
+      },
+    },
+  },
```
edlerd: Where do these build changes come from? Is it suggested by pragma?

edlerd (commented):
One question on the vite config change and a concern on the mixed font sizes.

src/components/MountedConfirmationButton.tsx
edlerd: Another option is that we drop the small font in the modal and only keep the muted color.

vite.config.ts
advl: This is a [subpath import feature ](https://johnnyreilly.com/only-node-subpaths-with-no-restricted-imports-and-perfectionist-sort-imports)we use internally but you shouldn't have to worry about or shouldn't impact you. It is fairly new, I will take a look tomorrow at why this ended up in your PR (it shouldn't)

src/components/MountedConfirmationButton.tsx
advl: Could we try dropping the small font and see how it goes ? We could perhaps simply make the size of the keys proportional to the text surrounding them ?

cc @diana-stanciulescu

vite.config.ts
edlerd: Thanks for the link @advl I think the link content only refers to the `alias` block. What about the `resolve` and `css: true`?

I think the lower block relates to tests exclusively. Then this is the wrong file, and we should move it to `vitest.config.ts`.

kimanhou: These changes were not suggested by pragma.

`dedupe` is to fix the error `Cannot read properties of null (reading 'useContext')`, which means that there are two different versions of React running at the same time.

Then the unit tests were failing with the error `TypeError: Unknown file extension ".css" for /project/node_modules/@canonical/react-ds-global/dist/esm/lib/_work_in_progress/CategoriesSection/styles.css`, hence `css: true`.
After fixing that, I got a different error when running the unit tests: `Error: Failed to resolve import "#lib/component/Icon/index.js" from "node_modules/@canonical/react-ds-global/dist/esm/lib/_work_in_progress/IconSection/IconSection.js". Does the file exist?`, hence the alias.

kimanhou: > we should move it to vitest.config.ts.

Ah yes sorry, didn't notice I was in `vite.config.ts`

edlerd: > These changes were not suggested by pragma.

If this is mandatory configuration to get it running, I think this should be suggested in the pragma docs. Especially as the setup of react + vite is a very common one. FYI @advl 

> `dedupe` is to fix the error `Cannot read properties of null (reading 'useContext')`, which means that there are two different versions of React running at the same time.

Where does the conflicting React version come from? Maybe a better approach is to streamline the versions.

If we can't streamline, I am not a fan of having package resolutions in the `vite.config`. Better keep it inside `package.json`, if possible. Then again, first choice should be streamlining the versions.

> Then the unit tests were failing with the error `TypeError: Unknown file extension ".css" for /project/node_modules/@canonical/react-ds-global/dist/esm/lib/_work_in_progress/CategoriesSection/styles.css`, hence `css: true`. After fixing that, I got a different error when running the unit tests: `Error: Failed to resolve import "#lib/component/Icon/index.js" from "node_modules/@canonical/react-ds-global/dist/esm/lib/_work_in_progress/IconSection/IconSection.js". Does the file exist?`, hence the alias.

Right, thanks for finding and implementing the fix. Seems good to keep in `vitest.config.ts`.

src/components/MountedConfirmationButton.tsx
kimanhou: Without small font

image

It looks good but it's a bit of a shame to have to adapt to the component. 
@advl Maybe KeyboardKey could take a className and that way we can pass `p-text--small` ?

kimanhou: In the meantime, I override the style of `KeyboardKey`

image

```
// Temporary fix for KeyboardKey in MountedConfirmationButton's ConfirmationModal
.p-modal__dialog {
  .ds.keyboard-key.code {
    @extend %small-text;
  }
}
```

edlerd: Looking at the focused screens now, I lean towards saying the unified font size actually looks better on the modal. It seems a bit more pleasant on the eye and easier to grasp. Personally, I'd be fine with that way forward.

vite.config.ts
advl: This is very helpful. I will do a repro in the afternoon prior to updating the docs.

kimanhou: > Where does the conflicting React version come from?

I'm not entirely sure but it looks like LXD UI and `react-components` were using React 18 under the hood (even tho both `package.json` have React 19) because of `react-table`. `react-table` is used in react-components' `ModularTable`, which is not used in LXD UI.

yarn.lock
edlerd, line 1: It seems like the whole lockfile changed. Can we keep the changes minimal?

edlerd (commented):
Thanks for the iteration, mostly LGTM, just one question regarding the lockfile left.

yarn.lock
kimanhou: Is it better now ? Sorry, I keep switching branches, and adding/removing dependencies so my `yarn.lock` is all over the place.

src/components/MountedConfirmationButton.tsx
```diff
@@ -72,9 +73,9 @@ export const MountedConfirmationButton = ({
       >
         {confirmationModalProps.children}
         {showShiftClickHint && (
-          <p className="p-text--small u-text--muted u-hide--small">
+          <p className="u-text--muted u-hide--small">
             Next time, you can skip this confirmation by holding{" "}
-            <code>SHIFT</code> and clicking the action.
+            <KeyboardKey keyValue="shift" /> and clicking the action.
```
edlerd, line 78: In light theme it looks all right

Image

but under the dark theme the keyboard shortcut does not adjust and is drawing too much attention

Image

edlerd: This would be a step back, especially considering the previous dark theme:

Image

edlerd (commented):
Code LGTM, but there is one more QA issue with the dark theme.

src/components/MountedConfirmationButton.tsx
kimanhou: This is weird, I don't have the same as you in dark theme

image

kimanhou: I could reproduce it by:
- being on the modal in dark theme
- switch to light theme
- not refreshing the page

LXD UI doesn't detect the theme change, I always have to refresh. But it looks like pragma does it automatically. It behaves the same when changing from light to dark:
image

kimanhou: We could implement a hook in LXD UI to detect changes of preference but it's unrelated to the current PR

```
import { useEffect, useState } from "react";
import { isDarkTheme, loadTheme } from "@canonical/react-components";

const QUERY = "(prefers-color-scheme: dark)";

export const useIsDarkTheme = (): boolean => {
  const [isDark, setIsDark] = useState(() => isDarkTheme(loadTheme()));

  useEffect(() => {
    const mediaQuery = window.matchMedia(QUERY);
    const listener = () => {
      setIsDark(isDarkTheme(loadTheme()));
    };
    mediaQuery.addEventListener("change", listener);
    return () => {
      mediaQuery.removeEventListener("change", listener);
    };
  }, []);

  return isDark;
};
```

kimanhou: https://github.com/user-attachments/assets/e87c9493-ae73-448d-ab2f-5b8c5be09e5d

edlerd: The "Shift" always has a white background in the dark theme to me, also after reload. I am using the demo server and a local build of this PR.

diana-stanciulescu:
hi all, this component does not yet have design documentation, hence the bare bones look to it. I am not sure when you need to go in production with this. We did have some explorations done, looking like this: 

Screenshot 2026-07-08 095207

I will discuss this with designers and get back to you with the tokens needed. I do not anticipate major design changes. One dependency is an update to  text/secondary/code - in this screenshot the text is text primary - 14, and the code text is text/secondary/code - 12 - which is currently using 14px and needs an update. Thanks!

src/components/MountedConfirmationButton.tsx
kimanhou: The bug happens because pragma uses the native CSS `light-dark()` function, which ignores our legacy `.is-dark` body class and relies strictly on the browser's system `color-scheme` property.

Since LXD UI uses `applyTheme` from react-components, I made a fix there: https://github.com/canonical/react-components/pull/1394

kimanhou: https://github.com/user-attachments/assets/3ecbc446-c1ac-4094-9a81-3a20ca5b443a

edlerd: Thanks for the fix, I think we should notify a wider web audience about this change. I think some teams have a copied that theme switcher and might miss the fix. Resolving it here, as it will be fixed as soon as the upstream fix lands and gets pulled in here.

edlerd (dismissed):
LGTM, thanks for the many follow-ups, testing the integration, and finding suitable solutions for all problems surfaced!

kimanhou:
@edlerd Sorry I updated react-components in this PR. Could you please re-approve ? 🙏

edlerd (approved):

advl:
A quick update on the hash aliases (`#lib`) - this is indeed an undocumented blocker atm, I was investigating and looking at options. I will need to make a small change on the way the apps are packaged to avoid the vite config change you have made. I was apparently too ambitious in getting an ideal build system to work without a bundler

Here is the recommended approach for now :
- Use the vite fix you have already done in the PR
- I will fix this upstream later in the week, and report
- follow-up with a fix pr, updating the deps and removing the unneeded config seam

How does that sound ?

advl:
Released pragma components v 0.30.0 which fixes the `#` imports, and then the need for a custom vite config. See pragma [#745](https://github.com/canonical/pragma/pull/745)

cc @kimanhou @edlerd

kimanhou:
@advl Thank you
I removed the `#` alias in `vitest.config`.
See LXD UI https://github.com/canonical/lxd-ui/pull/2143

PR #2131, opened 2026-07-08
fix(symlinks): [WD-37466] Fix Symlinks behaviour

### LINKED ISSUE #2109: File Explorer directory symlinks cannot be traversed
Symlinks to directories in the new file explorer are directly downloaded. I am not able to browse them in the UI.

/lib (symlink) links to: `https://example.com/1.0/instances/proxy/files?project=default&path=%2Flib`
/root (dir) links to: `https://example.com/ui/project/default/instance/proxy/file-explorer?project=default&path=%2Froot`

manually building the url: `https://example.com/ui/project/default/instance/proxy/file-explorer?project=default&path=%2Flib` doesn't work:

Image

`/lib` is a symlink to another directory `/usr/lib`, so it should be treated just like a directory

  # Metadata
    
  UI Version: 0.22
  Path: https://<ANONYMOUS_HOST>/ui/project/default/instance/proxy/file-explorer

Kxiru:
## Done

- [x] Introduces new Symlinks component
- [x] Symlinks should now be recursive and clicking on them should redirect to the folder. Note that some symlinks redirect to files.
  - For me, root/dev/fd is a symlink that redirects
  - Root/lib/libc.musl-x86_64.so.1 is a symlink that downloads a file 
- [x] Max depth of symlinks included
- [x] Error cases and validation included
- [x] Consistent styling (links, icons)

**PR Still in progress but early insights would be appreciated**

Fixes: https://github.com/canonical/lxd-ui/issues/2109

## QA

1. Run the LXD-UI:
    - Observe new Symlink functionality, including icons and redirection for directories and downloads for symlinks referring to files.

## Screenshots

image

### DIFF
--- a/src/api/instances.tsx
+++ b/src/api/instances.tsx
@@ -562,7 +562,8 @@ export const fetchInstanceDirectory = async (
 ): Promise<LxdFileExplorerItem> => {
   const params = new URLSearchParams();
   params.set("project", project);
-  params.set("path", path);
+  const normalizedPath = path.endsWith("/") ? path : path + "/";
+  params.set("path", normalizedPath);
 
   return fetch(
     `${ROOT_PATH}/1.0/instances/${encodeURIComponent(name)}/files?${params.toString()}`,
@@ -601,6 +602,23 @@ export const fetchInstanceFileHeader = async (
       modified: response.headers.get("x-lxd-modified") ?? "-",
     };
   });
+};
+
+export const fetchSymlinkContent = async (
+  instanceName: string,
+  project: string,
+  path: string,
+): Promise<string> => {
+  const params = new URLSearchParams();
+  params.set("project", project);
+  params.set("path", path);
+
+  return fetch(
+    `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instanceName)}/files?${params.toString()}`,
+    {
+      method: "GET",
+    },
+  ).then(handleTextResponse);
 };
 
 export const deleteInstanceFile = async (
--- a/src/pages/instances/FileExplorerSymlink.tsx
+++ b/src/pages/instances/FileExplorerSymlink.tsx
@@ -1,0 +1,117 @@
+import { type FC, useState } from "react";
+import {
+  Button,
+  Icon,
+  Spinner,
+  useToastNotification,
+} from "@canonical/react-components";
+import { useNavigate } from "react-router-dom";
+import type { LxdInstance } from "types/instance";
+import { fetchSymlinkContent, fetchInstanceFileHeader } from "api/instances";
+import {
+  getFileExplorerDirectoryURL,
+  getFileExplorerURL,
+  getFullPath,
+  resolveSymlinkTarget,
+} from "util/instances";
+
+const MAX_SYMLINK_DEPTH = 10;
+
+interface Props {
+  fileName: string;
+  parentPath: string;
+  instance: LxdInstance;
+}
+
+const isValidSymlinkTarget = (target: string): boolean => {
+  return target.length > 0 && target.length <= 4096 && !target.includes("\0");
+};
+
+const FileExplorerSymlink: FC<Props> = ({ fileName, parentPath, instance }) => {
+  const navigate = useNavigate();
+  const toastNotify = useToastNotification();
+  const [isLoading, setIsLoading] = useState(false);
+
+  const fullPath = getFullPath(parentPath, fileName);
+
+  const handleSymlink = async (
+    path: string,
+    depth = 0,
+    originalPath?: string,
+  ) => {
+    const navigatePath = originalPath ?? path;
+    if (depth >= MAX_SYMLINK_DEPTH) {
+      toastNotify.failure(
+        "Symlink resolution failed",
+        new Error("Too many levels of symbolic link nesting"),
+      );
+      return;
+    }
+
+    setIsLoading(true);
+    try {
+      const target = await fetchSymlinkContent(
+        instance.name,
+        instance.project,
+        path,
+      );
+
+      const trimmedTarget = target.trim();
+
+      if (!isValidSymlinkTarget(trimmedTarget)) {
+        toastNotify.failure(
+          "Symlink resolution failed",
+          new Error("Invalid symlink target"),
+        );
+        return;
+      }
+
+      const resolvedTarget = resolveSymlinkTarget(trimmedTarget, path);
+      const metadata = await fetchInstanceFileHeader(
+        instance.name,
+        instance.project,
+        resolvedTarget,
+      );
+
+      if (metadata.type === "directory") {
+        navigate(getFileExplorerDirectoryURL(navigatePath, instance));
+      } else if (metadata.type === "symlink") {
+        await handleSymlink(resolvedTarget, depth + 1, navigatePath);
+      } else {
+        const url = getFileExplorerURL(resolvedTarget, undefined, instance);
+
+        const link = document.createElement("a");
+        link.href = url;
+        link.download = fileName;
+        document.body.appendChild(link);
+        link.click();
+        link.remove();
+      }
+    } catch (error) {
+      toastNotify.failure(
+        "Symlink resolution failed",
+        error instanceof Error ? error : new Error("Failed to resolve symlink"),
+      );
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  return (
+    <Button
+      className="file-explorer-item u-no-margin--bottom"
+      appearance="link"
+      onClick={() => {
+        void handleSymlink(fullPath);
+      }}
+      disabled={isLoading}
+    >
+      {isLoading ? <Spinner /> : <Icon name="connected" />}
+      <span className="file-explorer-item__name" title={fileName}>
+        {fileName}
+      </span>
+    </Button>
+  );
+};
+
+export default FileExplorerSymlink;
--- a/src/util/instances.tsx
+++ b/src/util/instances.tsx
@@ -97,18 +97,21 @@ export const getFileExplorerDirectoryURL = (
   return `${ROOT_PATH}/ui/project/${encodeURIComponent(instance.project)}/instance/${encodeURIComponent(instance.name)}/file-explorer?${params.toString()}`;
 };
 
-export const getFileExplorerFileURL = (
+export const getFileExplorerURL = (
   parentPath: string,
-  fileName: string,
+  fileName: string | undefined,
   instance: LxdInstance,
 ) => {
   const params = new URLSearchParams();
   params.set("project", instance.project);
-  params.set(
-    "path",
-    parentPath === "/" ? `/${fileName}` : `${parentPath}/${fileName}`,
-  );
 
+  if (!fileName) {
+    params.set("path", parentPath);
+    return `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instance.name)}/files?${params.toString()}`;
+  }
+
+  const path = getFullPath(parentPath, fileName);
+  params.set("path", path);
   return `${ROOT_PATH}/1.0/instances/${encodeURIComponent(instance.name)}/files?${params.toString()}`;
 };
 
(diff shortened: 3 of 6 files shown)


src/pages/instances/FileExplorerSymlink.tsx
```diff
@@ -0,0 +1,129 @@
...
+    if (part === "..") {
+      resolved.pop();
+    } else if (part !== "." && part !== "") {
+      resolved.push(part);
+    }
+  }
+  return "/" + resolved.join("/");
+};
+
+const FileExplorerSymlink: FC<Props> = ({ fileName, parentPath, instance }) => {
```
edlerd, line 30: The symlinks should also be blue in the file list. Just like files and directories:

Image

src/pages/instances/FileExplorerSymlink.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+      const resolvedTarget = resolveSymlinkTarget(trimmedTarget, path);
+      const metadata = await fetchInstanceFileHeader(
+        instance.name,
+        instance.project,
+        resolvedTarget,
+      );
+
+      if (metadata.type === "directory") {
+        navigate(getFileExplorerDirectoryURL(resolvedTarget, instance));
```
edlerd: This should navigate treating the symlink as a directory, not the resolved target. So when navigating to `var` and `run`, the breadcrumb on the top should say `root/var/run`, not `root/run`.

edlerd (commented):
Did a quick QA, looking good already, but two issues surfaced. Will do code review once the other create directory PR lands and this is rebased. So we have the exclusive changes in this one instead of the combined ones we have now.

sophiecarroll:
This might just be the demo, but is this supposed to happen when I click a symlink?
Screenshot 2026-07-08 at 10 42 37 AM

src/pages/instances/FileExplorerSymlink.tsx
sophiecarroll: Seems good to me on the UX side! I would second David about symlinks being blue unless there's a reason to be black I don't know about, if it was discussed elsewhere.

Kxiru:
> This might just be the demo, but is this supposed to happen when I click a symlink?

I got this for some symlinks too but I wasn't sure if this was expected behaviour or not as I am not too familiar with the concept. My understanding is that symlinks should point to files, directories or maybe even other symlinks. @edlerd , do you have any idea why this might occur?

edlerd:
> > This might just be the demo, but is this supposed to happen when I click a symlink?
> 
> I got this for some symlinks too but I wasn't sure if this was expected behaviour or not as I am not too familiar with the concept. My understanding is that symlinks should point to files, directories or maybe even other symlinks. @edlerd , do you have any idea why this might occur?

This is a special system link that the LXD api doesn't handle right, or it simply returns a 404 instead of resolving it correctly.

Kxiru:
@edlerd 
> So we have the exclusive changes in this one instead of the combined ones we have now.

I'll rebase onto the parent branch anyway so that this is always pushed with the latest changes from there.

> This is a special system link that the LXD api doesn't handle right, or it simply returns a 404 instead of resolving it correctly.

Do we need to address this in any way?

src/pages/instances/FileExplorerSymlink.tsx
Kxiru: @sophiecarroll , nope! I'll make them blue. I can't find a standardised Vanilla for it though as the ones I have tried previously either do not resolve, or are coloured incorrectly relative to our other links, and it is bad practice to hard code the colour. I will scour the codebase for an example.

src/pages/instances/FileExplorerSymlink.tsx
Kxiru: What about if the breadcrumb said root/symlink? and the file explorer resolved to root/var/run?
I'm a bit confused as to what the standard output of the symlink should be at this point. Please see the latest iteration.

edlerd:
> > This is a special system link that the LXD api doesn't handle right, or it simply returns a 404 instead of resolving it correctly.
> Do we need to address this in any way?

I think there is not much we can do. You can report it, maybe it is a bug treating this special case wrongly in the LXD API layer.

src/pages/instances/FileExplorerSymlink.tsx
edlerd: Are you using a button @Kxiru ? There is a link appearance for the button.

src/pages/instances/FileExplorerSymlink.tsx
edlerd: New behavior after your change seems correct. :+1:

src/pages/instances/FileExplorerSymlink.tsx
Kxiru: I am using the primitive <button> element.

src/pages/instances/FileExplorerSymlink.tsx
```diff
@@ -0,0 +1,136 @@
...
+        "Symlink resolution failed",
+        error instanceof Error ? error : new Error("Failed to resolve symlink"),
+      );
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  return (
+    <button
```
edlerd: Let's use the `Button` with `appearance` set to `link` in favor of custom styling and the primitive.

src/pages/instances/FileExplorerBreadcrumb.tsx
```diff
@@ -1,50 +1,159 @@
...
+    }
+
+    setIsNavigating(true);
+    try {
+      await fetchInstanceDirectory(
+        instance.name,
+        instance.project,
+        trimmedPath,
+      );
+      navigate(getFileExplorerDirectoryURL(trimmedPath, instance));
```
edlerd, line 51: When in a nested symlinked path and going directly on the symlink (such as in `/var/run/cloud-init` and going for the `/var/run`), then the breadcrumb links to `/var/run` instead of `/var/run/` and fails because of the missing trailing slash.

Image

edlerd (commented):
QA mostly good, tiny issue with the breadcrumb and let's use the standard `Button`. Delaying full code review until the base is merged and this is rebased to show exclusive changes.

sophiecarroll (dismissed):

src/api/instances.tsx
```diff
@@ -562,7 +562,8 @@ export const fetchInstanceDirectory = async (
 ): Promise<LxdFileExplorerItem> => {
   const params = new URLSearchParams();
   params.set("project", project);
-  params.set("path", path);
+  const normalizedPath = path === "/" || path.endsWith("/") ? path : path + "/";
```
```suggestion
  const normalizedPath = path.endsWith("/") ? path : path + "/";
```
this is equivalent.

edlerd (commented):
Tiny issue. Postponing full review until the other PR is landed and this is rebased on main.

sophiecarroll (dismissed):

src/util/instances.tsx
```diff
@@ -97,17 +97,21 @@ export const getFileExplorerDirectoryURL = (
   return `${ROOT_PATH}/ui/project/${encodeURIComponent(instance.project)}/instance/${encodeURIComponent(instance.name)}/file-explorer?${params.toString()}`;
 };
 
-export const getFileExplorerFileURL = (
+export const getFileExplorerURLByType = (
```
```suggestion
export const getFileExplorerURL = (
```
I think this can be simpler.

src/util/instances.tsx
```diff
@@ -97,17 +97,21 @@ export const getFileExplorerDirectoryURL = (
...
-    "path",
-    parentPath === "/" ? `/${fileName}` : `${parentPath}/${fileName}`,
-  );
+
+  // If fileName is provided, construct the full path; otherwise, use the parentPath
+  const path = fileName
+    ? parentPath === "/"
+      ? `/${fileName}`
+      : `${parentPath}/${fileName}`
+    : parentPath;
```
edlerd: We shouldn't nest the ternaries. Better extract this in a function with early exits.

src/util/instances.tsx
```diff
@@ -97,17 +97,21 @@ export const getFileExplorerDirectoryURL = (
...
-  params.set(
-    "path",
-    parentPath === "/" ? `/${fileName}` : `${parentPath}/${fileName}`,
-  );
+
+  // If fileName is provided, construct the full path; otherwise, use the parentPath
+  const path = fileName
+    ? parentPath === "/"
+      ? `/${fileName}`
+      : `${parentPath}/${fileName}`
```
edlerd: Can we use the same `.endsWidth("/")` approach we established elsewhere? Actually, I think this should be a utility function, as it leaks into so many places. Something like `getFullPath(parentPath, fileName)`. It could handle all the cases. Then we can use it in all the places.

src/pages/instances/FileExplorerSymlink.tsx
```diff
@@ -0,0 +1,136 @@
...
+  return target.length > 0 && target.length <= 4096 && !target.includes("\0");
+};
+
+const resolveSymlinkTarget = (target: string, currentPath: string): string => {
+  if (target.startsWith("/")) {
+    return target;
+  }
+  const parentDir =
+    currentPath.substring(0, currentPath.lastIndexOf("/")) || "/";
+  const combined = parentDir === "/" ? `/${target}` : `${parentDir}/${target}`;
```
edlerd: Here we can also use the new `getFullPath` function.

src/pages/instances/FileExplorerSymlink.tsx
```diff
@@ -0,0 +1,136 @@
...
+  return "/" + resolved.join("/");
+};
+
+const FileExplorerSymlink: FC<Props> = ({ fileName, parentPath, instance }) => {
+  const navigate = useNavigate();
+  const toastNotify = useToastNotification();
+  const [isLoading, setIsLoading] = useState(false);
+
+  const fullPath =
+    parentPath === "/" ? `/${fileName}` : `${parentPath}/${fileName}`;
```
edlerd: Here we can also use `getFullPath`.

src/pages/instances/FileExplorerSymlink.tsx
```diff
@@ -0,0 +1,136 @@
...
+    <Button
+      className="file-explorer-item u-no-margin--bottom"
+      appearance="link"
+      onClick={() => {
+        handleSymlink(fullPath);
+      }}
+      disabled={isLoading}
+    >
+      {isLoading ? (
+        <Spinner className="file-explorer-symlink-spinner" />
```
edlerd:
```suggestion
        <Spinner />
```
This is unused.

src/pages/instances/FileExplorerSymlink.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const handleSymlink = async (
+    path: string,
+    depth = 0,
+    originalPath?: string,
+  ) => {
+    const navigatePath = originalPath ?? path;
+    if (depth >= MAX_SYMLINK_DEPTH) {
+      toastNotify.failure(
+        "Symlink resolution failed",
+        new Error("Too many levels of symbolic links"),
```
edlerd:
```suggestion
        new Error("Too many levels of symbolic link nesting"),
```
A bit clearer?

edlerd (commented):
QA looks great! Some thoughts to simplify the code below.

src/util/instances.tsx
Kxiru: I like the getFullPath idea, but I'm not sure how endsWith() can be used here. Could you review the latest version and let me know your thoughts in context?

src/util/instances.tsx
```diff
@@ -128,3 +131,29 @@ export const validateDirectoryPathSyntax = (
 
   return undefined;
 };
+
+export const getFullPath = (parentPath: string, fileName: string): string => {
+  return parentPath === "/" ? `/${fileName}` : `${parentPath}/${fileName}`;
+};
+
+export const resolveSymlinkTarget = (
```
edlerd, line 139: The logic in here is rather complex. Let's add a follow up to create a spec test for this util function.

edlerd (approved):
LGTM, good work :)

PR #2132, opened 2026-07-08
feat(Identities): group selection for token bearer identities

kimanhou:
## Done

- feat(Identities): group selection for token bearer identities
- Update success notification after updating the groups to display the name of the identity, the type and the ID.
- Update modal and success notification when deleting an identity to display the name of the identity, the type and the ID.
- 
## QA

1. Run the LXD-UI:
    - Create two identities with the CLI
    ```
    lxc auth identity create devlxd/csi-driver --group admins
    lxc auth identity show bearer/terraform-bot
    ```
    - Go to Permissions > Identities. Make sure the two identities are correctly shown in the table, with the correct group count.
    - Click on the group icon to open the side panel. Make sure "admins" is selected.
    - Remove and/or add groups and save. Make sure the success notification displays the name of the identity, the type and the ID.
    - Make sure the identity is saved correctly. You can double check with the CLI `lxc auth identity show bearer/terraform-bot`

## Screenshots

image

After adding a group to csi-driver
image
image

Success notification
image

Delete modal
image

Success notification of deletion
image

### DIFF
--- a/src/components/IdentityResource.tsx
+++ b/src/components/IdentityResource.tsx
@@ -1,26 +1,37 @@
 import type { FC } from "react";
+import ResourceLabel from "components/ResourceLabel";
+import ResourceLink from "components/ResourceLink";
 import type { LxdIdentity } from "types/permissions";
-import ResourceLabel from "./ResourceLabel";
-import { AUTH_METHOD } from "util/authentication";
+import { getIdentityIconType } from "util/permissionIdentities";
+import { ROOT_PATH } from "util/rootPath";
 
 interface Props {
   identity: LxdIdentity;
-  truncate?: boolean;
-  bold?: boolean;
+  variant?: "link" | "label";
 }
 
-const IdentityResource: FC<Props> = ({ identity, truncate, bold }) => {
-  const identityIconType =
-    identity.authentication_method == AUTH_METHOD.TLS
-      ? "certificate"
-      : "oidc-identity";
+const IdentityResource: FC<Props> = ({ identity, variant = "link" }) => {
+  const name = identity.name?.trim();
+  const label = name
+    ? `${name} (${identity.type}, ${identity.id})`
+    : `${identity.id} (${identity.type})`;
+
+  if (variant === "label") {
+    return (
+      <ResourceLabel
+        type={getIdentityIconType(identity.type)}
+        value={label}
+        bold
+        truncate
+      />
+    );
+  }
 
   return (
-    <ResourceLabel
-      type={identityIconType}
-      value={identity.type}
-      truncate={truncate}
-      bold={bold}
+    <ResourceLink
+      type={getIdentityIconType(identity.type)}
+      value={label}
+      to={`${ROOT_PATH}/ui/permissions/identities`}
     />
   );
 };
--- a/src/components/ResourceIcon.tsx
+++ b/src/components/ResourceIcon.tsx
@@ -32,7 +32,8 @@ export type ResourceIconType =
   | "device"
   | "setting"
   | "peering"
-  | "replicator";
+  | "replicator"
+  | "token-bearer";
 
 const resourceIcons: Record<ResourceIconType, string> = {
   container: "pods",
@@ -66,6 +67,7 @@ const resourceIcons: Record<ResourceIconType, string> = {
   metric: "statistics",
   "placement-group": "repository",
   replicator: "change-version",
+  "token-bearer": "private-key",
 };
 
 interface Props {
--- a/src/pages/permissions/panels/IdentityGroupsPanelConfirmModal.tsx
+++ b/src/pages/permissions/panels/IdentityGroupsPanelConfirmModal.tsx
@@ -14,9 +14,8 @@ import GroupsOrIdentityChangesTable from "./GroupOrIdentityChangesTable";
 import { updateIdentities } from "api/auth-identities";
 import { useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
+import IdentityResource from "components/IdentityResource";
 import usePanelParams from "util/usePanelParams";
-import ResourceLink from "components/ResourceLink";
-import { ROOT_PATH } from "util/rootPath";
 
 interface Props {
   onConfirm: () => void;
@@ -74,21 +73,27 @@ const IdentityGroupsPanelConfirmModal: FC<Props> = ({
           },
         });
 
-        const modifiedGroupNames = Object.keys(identityGroupsChangeSummary);
-        const successMessage =
-          modifiedGroupNames.length > 1 ? (
-            `Updated groups for ${modifiedGroupNames.length} identities.`
-          ) : (
-            <>
-              Updated groups for{" "}
-              <ResourceLink
-                type="oidc-identity"
-                value={modifiedGroupNames[0]}
-                to={`${ROOT_PATH}/ui/permissions/identities`}
-              />
-              .
-            </>
-          );
+        const modifiedIdentityIds = Object.keys(identityGroupsChangeSummary);
+        const isMultiple = modifiedIdentityIds.length > 1;
+        const [firstId, firstData] =
+          Object.entries(identityGroupsChangeSummary)[0] || [];
+        const firstIdentityDetails = selectedIdentities.find(
+          (id) => id.id === firstId,
+        );
+
+        const successMessage = isMultiple ? (
+          `Updated groups for ${modifiedIdentityIds.length} identities.`
+        ) : (
+          <>
+            Updated groups for{" "}
+            {firstIdentityDetails ? (
+              <IdentityResource identity={firstIdentityDetails} />
+            ) : (
+              firstData?.name?.trim() || firstId
+            )}
+            .
+          </>
+        );
 
         toastNotify.success(successMessage);
         panelParams.clear();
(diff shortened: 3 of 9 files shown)


edlerd (dismissed):
LGTM

edlerd (commented):
On updating an identity, can we show the name and type of the identity and then the id? Currently the success notification looks a bit cryptic with just the ID in the chip.

kimanhou:
@edlerd I want to change the way we update identities. See [JIRA task](https://warthogs.atlassian.net/browse/WD-37701).

> Currently the success notification looks a bit cryptic with just the ID in the chip.

Done
image

src/pages/permissions/panels/IdentityGroupsPanelConfirmModal.tsx
```diff
@@ -83,7 +84,7 @@ const IdentityGroupsPanelConfirmModal: FC<Props> = ({
               Updated groups for{" "}
               <ResourceLink
                 type="oidc-identity"
-                value={modifiedGroupNames[0]}
+                value={modifiedIdentities[0]?.name ?? modifiedGroupNames[0]}
```
edlerd: The name is not unique, so I think best is a combination of name, type and id to be specific.

edlerd (commented):
Thanks, a small adjustment is needed to the update message.

src/pages/permissions/panels/IdentityGroupsPanelConfirmModal.tsx
kimanhou: Like this ?
image

edlerd: I think this ok, yes. It will overflow in any case. Wdyt?

edlerd: Also, we should take care that the name could be empty, and in that case, we need a fallback.

kimanhou: What do you think of middle truncation for the ID ?
image

kimanhou: It only works with short identity name tho. Also "client token bearer" is longer than "client certificate".
image

edlerd: > What do you think of middle truncation for the ID ?

Probably too much complexity. The name could already satisfy the full available length if it is long enough.

src/pages/permissions/panels/IdentityGroupsPanelConfirmModal.tsx
```diff
@@ -74,21 +74,30 @@ const IdentityGroupsPanelConfirmModal: FC<Props> = ({
...
+        const identityType = firstIdentityDetails?.type ?? "Unknown type";
+        const modifiedIdentityLabel = `${displayName} (${identityType}, ${firstId})`;
+
+        const successMessage = isMultiple ? (
+          `Updated groups for ${modifiedIdentityIds.length} identities.`
+        ) : (
+          <>
+            Updated groups for{" "}
+            <ResourceLink
+              type="oidc-identity"
```
omarelkashef: This resource type depends on the identity type and not always oidc-identity, right? We will also probably need a new icon for bearer or add to certificate or oidc-identity and make the name more general.

omarelkashef (commented):
LGTM, only a small comment.

src/pages/permissions/panels/IdentityGroupsPanelConfirmModal.tsx
edlerd: Good point, there is already a `certificate` type and we should define a new `bearer-token` one. The underlying icon can be adjusted later on. We should set the right type here depending on the entity being updated.

src/components/IdentityResource.tsx
```diff
@@ -1,26 +1,54 @@
...
 }
 
-const IdentityResource: FC<Props> = ({ identity, truncate, bold }) => {
-  const identityIconType =
-    identity.authentication_method == AUTH_METHOD.TLS
-      ? "certificate"
-      : "oidc-identity";
+const IdentityResource: FC<Props> = ({ identity, variant = "link" }) => {
+  const getIcon = (identityType: string) => {
+    switch (identityType) {
```
omarelkashef: would it be easier to read this way? I am not sure.

```
if(identityType.startWith("Client certificate") || identityType.startWith("Cluster link certificate") ||  identityType.startWith("Metrics certificate") || identityType.startWith("Server certificate")) {
  return "certificate" 
}

if(identityType.includes("token bearer") {
  return "bearer-token"
}

return "oidc-identity"
```

src/components/IdentityResource.tsx
```diff
@@ -1,26 +1,54 @@
...
+      case "Client token bearer":
+      case "DevLXD token bearer":
+        return "bearer-token";
+      default:
+        return "oidc-identity";
+    }
+  };
+
+  const displayName = identity.name?.trim() || identity.id;
+  const label = `${displayName} (${identity.type}, ${identity.id})`;
```
omarelkashef: In the case of an empty identity.name, the label would be "id (type, id)". Would that be ok?

kimanhou: @edlerd Wdyt ?

edlerd: I think in the fallback case it should by "id (type)".

src/components/IdentityResource.tsx
kimanhou: Maybe. But less precise. I'm still undecided.

The question is also: what do we display in the chip and the type is "XXX (Pending)" ? Should we only keep the type XXX, ie displaying "Client certificate" instead of "Client certificate (Pending)" ?
Right now, the whole thing is displayed.

edlerd: or maybe 
```
if (identityType.includes("certificate")) {
  return "certificate" 
}
...
```

omarelkashef: That is even simpler yes

kimanhou: Done

src/components/IdentityResource.tsx
kimanhou: Done

omarelkashef (dismissed):
LGTM, thank you!

src/components/IdentityResource.tsx
```diff
@@ -1,26 +1,46 @@
...
+    if (identityType.toLowerCase().includes("token bearer")) {
+      return "token-bearer";
+    }
+    return "oidc-identity";
+  };
+
+  const displayName = identity.name?.trim() || identity.id;
+  const label = identity.name?.trim()
+    ? `${displayName} (${identity.type}, ${identity.id})`
+    : `${displayName} (${identity.type})`;
```
edlerd:
```suggestion
  const name = identity.name?.trim()
  const label = name
    ? `${name} (${identity.type}, ${identity.id})`
    : `${identity.id} (${identity.type})`;
```
A bit simpler.

src/components/ResourceIcon.tsx
```diff
@@ -66,6 +67,7 @@ const resourceIcons: Record<ResourceIconType, string> = {
   metric: "statistics",
   "placement-group": "repository",
   replicator: "change-version",
+  "token-bearer": "user",
```
edlerd:
```suggestion
  "token-bearer": "private-key",
```
I think this one is better fitting, wdyt?

Image

tests/permission-identities.spec.ts
```diff
@@ -47,7 +47,10 @@ test("manage groups for single identity", async ({ page, lxdVersion }) => {
     "added",
   );
   await page.getByRole("button", { name: "Confirm changes" }).click();
-  await dismissNotification(page, `Updated groups for ${identityBar}.`);
+  await dismissNotification(
+    page,
+    `Updated groups for  bar (OIDC client, ${identityBar}).`,
```
edlerd:
```suggestion
    `Updated groups for bar (OIDC client, ${identityBar}).`,
```
One space should be enough.

edlerd (commented):
Few tiny nitpicks below, otherwise LGTM.

src/components/ResourceIcon.tsx
kimanhou: Okay but please note it's already used for `"bucket-key": "private-key"`

edlerd: I think that is okay, for the buckets those are also api keys.

src/components/IdentityResource.tsx
```diff
@@ -1,26 +1,46 @@
...
+const IdentityResource: FC<Props> = ({ identity, variant = "link" }) => {
+  const getIcon = (identityType: string) => {
+    if (identityType.toLowerCase().includes("certificate")) {
+      return "certificate";
+    }
+    if (identityType.toLowerCase().includes("token bearer")) {
+      return "token-bearer";
+    }
+    return "oidc-identity";
+  };
```
edlerd: This duplicates a `getType` function in `PermissionIdentities`. Let's make it a util to reuse.

edlerd (commented):
one concern left, otherwise LGTM

edlerd (approved):
LGTM, good work 👍

Issue #2108, opened 2026-06-28
Additional Disk Device options


vosdev:
Helloooo,

At the moment only the basic disk device options exist in the UI. For some options you have to resort to the yaml.

https://canonical.com/lxd/docs/latest/reference/devices_disk/#device-options

Options like:
- readonly
- recursive
- required
- shift

Can I request these be made available via the UI? :)

Kxiru:
I think you will find that these will soon be made available to you. This was merged in a [recent PR](https://github.com/canonical/lxd-ui/pull/2133), Stay tuned!

<img width="615" height="668" alt="Image" src="https://github.com/user-attachments/assets/23be55ba-e428-47dc-9791-809bcb713e25" />

vosdev:
Lovely! Thank you :). UI is coming closer each day to full cli compatibility!

vosdev:
Oh I noticed the shift option is not showing on your screen. That is a very important one for LXCs, as without the shift option a host mount is pretty useless. Was this left out intentionally? If this is a VM then that makes sense :)

edlerd:
The shift option is also supported in the UI.

PR #2133, opened 2026-07-08
feat(devices): additional disk device options

omarelkashef:
## Done

- additional disk device options WD-37536

## QA

1. Run the LXD-UI:
   1. Add a read-only host bind-mount disk to a container
      - Navigate to Instances → select or create a **container** → Configuration → Disk devices tab.
      - Add a disk device using a host path (not a storage volume).
      - Set **Read-only** to `yes`.
      - Verify the **Recursive** field is disabled and its tooltip explains the conflict.
      - Save and confirm the instance config reflects `readonly: "true"` and no `recursive` key.
   2. Add a recursive host bind-mount disk to a container
      - On the same or a new container, add a host path disk device.
      - Set **Recursive** to `yes`.
      - Verify the **Read-only** field is disabled with a matching tooltip.
      - Save and confirm `recursive: "true"` is present and `readonly` is absent in the config.
   3. Volume-backed disk does not show Recursive or Shift
      - Add a disk device backed by a **custom storage volume** to a container.
      - Verify the **Recursive** and **Shift** fields are **not** present in the form.
      - Verify **Read-only** and **Required** are still visible.
      - Save successfully.
   6. VM disk device does not show Shift
      - Navigate to a **virtual machine** → Configuration → Disk devices tab.
      - Add a disk device (host path or volume).
      - Verify the **Shift** field is **not** present.
      - Verify **Read-only** and **Required** are visible.
   7. Shift field visible for container and profile disk devices
      - On a **container**, add a host bind-mount disk device.
      - Verify the **Shift** field is present; set it to `yes` and save.
      - Navigate to **Profiles** → Configuration→ Disk devices → add a host path disk.
      - Verify **Shift** is also present on the profile form.
   8. Inherited disk device displays the new properties
      - Apply a profile to a container where the profile's disk device has `readonly`, `recursive`, `required`, or `shift` set.
      - Edit the container → Disk devices tab.
      - Verify the inherited disk device row shows the correct labels and values for each property that is set.

## Screenshots

image

image

### DIFF
--- a/src/components/forms/DiskDeviceFormCustom.tsx
+++ b/src/components/forms/DiskDeviceFormCustom.tsx
@@ -1,7 +1,16 @@
 import type { FC } from "react";
-import { Button, Icon, Input, Label } from "@canonical/react-components";
+import {
+  Button,
+  Icon,
+  Input,
+  Label,
+  Select,
+} from "@canonical/react-components";
 import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
-import type { EditInstanceFormValues } from "types/forms/instanceAndProfile";
+import type {
+  CreateInstanceFormValues,
+  EditInstanceFormValues,
+} from "types/forms/instanceAndProfile";
 import CustomVolumeSelectBtn from "pages/storage/CustomVolumeSelectBtn";
 import type {
   CustomDiskDevice,
@@ -33,11 +42,19 @@ import {
 } from "util/devices";
 import { isInstanceCreation } from "util/instanceEdit";
 import { ensureEditMode } from "util/editMode";
-import { focusField } from "util/formFields";
+import { focusField, optionRenderer } from "util/formFields";
 import AttachDiskDeviceBtn from "pages/storage/AttachDiskDeviceBtn";
 import type { LxdProfile } from "types/profile";
 import type { LxdStorageVolume } from "types/storage";
 import StoragePoolRichChip from "pages/storage/StoragePoolRichChip";
+import { optionYesNo } from "util/options";
+import {
+  getConfigFieldDefault,
+  getConfigFieldDescription,
+  toConfigFields,
+} from "util/config";
+import ConfigFieldDescription from "pages/settings/ConfigFieldDescription";
+import { useConfigOptions } from "context/useConfigOptions";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
@@ -49,6 +66,33 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
   const readOnly = (formik.values as EditInstanceFormValues).readOnly;
   const existingDeviceNames = getExistingDeviceNames(formik.values, profiles);
   const isProfile = formik.values.entityType === "profile";
+  const { data: configOptions } = useConfigOptions();
+
+  const diskOptions = configOptions?.configs?.["device-disk"];
+  const diskConfigFields = diskOptions ? toConfigFields(diskOptions) : [];
+
+  const diskFieldLabel = ({
+    id,
+    label,
+    key,
+    required = false,
+  }: {
+    id: string;
+    label: string;
+    key: string;
+    required?: boolean;
+  }) => (
+    <div className="configuration-label-with-help">
+      <Label forId={id} required={required}>
+        {label}
+      </Label>
+      <span className="configuration-help">
+        <ConfigFieldDescription
+          description={getConfigFieldDescription(diskConfigFields, key)}
+        />
+      </span>
+    </div>
+  );
 
   const getInitialDeviceName = (
     deviceType: string,
@@ -89,7 +133,7 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
       formik.setFieldValue(`devices.${index}.path`, "");
     }
 
-    // If path must not exist for the device, remote it even if it's set for the existing device
+    // If path must not exist for the device, remove it even if it's set for the existing device
     if (volume.content_type === "block") {
       formik.setFieldValue(`devices.${index}.path`, undefined);
     }
@@ -111,6 +155,41 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
       <Icon name="edit" />
     </Button>
   );
+
+  const readOnlyYesNoValue = (
+    fieldName: string,
+    value: string | undefined,
+    option: string,
+  ) => {
+    const displayValue =
+      value ?? getConfigFieldDefault(diskConfigFields, option);
+
+    return (
+      <div className="custom-disk-read-mode">
+        <div className="custom-disk-value">
+          <b>{optionRenderer(displayValue, optionYesNo) || "-"}</b>
+        </div>
+        {editButton(fieldName)}
+      </div>
+    );
+  };
+
+  const getYesNoOptionsWithLxdDefault = (fieldKey: string) => {
+    const defaultValue = getConfigFieldDefault(diskConfigFields, fieldKey);
+    const defaultLabel = optionRenderer(defaultValue, optionYesNo);
+    const placeholderLabel = defaultLabel
+      ? `LXD default (${defaultLabel})`
+      : "LXD default";
+
+    return optionYesNo.map((option, index) =>
+      index === 0
+        ? {
+            ...option,
+            label: placeholderLabel,
+          }
+        : option,
+    );
+  };
 
   const rows: MainTableRow[] = [];
   let customDiskDeviceCount = 0;
@@ -151,7 +230,7 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
       rows.push(
         getConfigurationRowBase({
           className: "no-border-top has-margin-left",
-          configuration: <div className="u-text--muted">Source</div>,
+          configuration: <div>Source</div>,
           inherited: (
             <b className="mono-font">{(item as IsoVolumeDevice).source}</b>
           ),
@@ -161,7 +240,7 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
       rows.push(
         getConfigurationRowBase({
           className: "no-border-top has-margin-left",
-          configuration: <div className="u-text--muted">Pool</div>,
+          configuration: <div>Pool</div>,
           inherited: (
             <StoragePoolRichChip
               poolName={(item as IsoVolumeDevice).pool}
@@ -209,9 +288,7 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
 
       const volumeDeviceSource = () =>
         getConfigurationRowBase({
-          className: classnames("no-border-top inherited-with-form", {
-            "device-last-row": isVolumeDevice(item) && item.path === undefined,
-          }),
+          className: classnames("no-border-top inherited-with-form"),
           configuration: (
             <Label forId={`devices.${index}.pool`} className="u-text--muted">
               Pool / volume
@@ -253,9 +330,7 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
 
       const hostDeviceSource = () =>
         getConfigurationRowBase({
-          className: classnames("no-border-top inherited-with-form", {
-            "device-last-row": !isVolumeDevice(item) && item.path === undefined,
-          }),
+          className: classnames("no-border-top inherited-with-form"),
           configuration: (
             <Label forId={`devices.${index}.source`} className="u-text--muted">
               Host path
@@ -286,20 +361,33 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
           override: "",
         });
 
+      const isContainerOrProfile =
+        isProfile ||
+        (formik.values as CreateInstanceFormValues).instanceType ===
+          "container";
+      const isVolumeBackedDisk = isVolumeDevice(item);
+      const isReadonlyDisk = item.readonly === "true";
+      const isRecursiveDisk = item.recursive === "true";
+      const showMountPointField =
+        !isVolumeBackedDisk || item.path !== undefined;
+      const showRecursiveField = !isVolumeBackedDisk;
+      const showShiftField = isContainerOrProfile && !isVolumeBackedDisk;
+      const isRequiredLastRow = !showShiftField;
+
       rows.push(
         isVolumeDevice(item) ? volumeDeviceSource() : hostDeviceSource(),
       );
 
-      if (!isVolumeDevice(item) || item.path !== undefined) {
+      if (showMountPointField) {
         const hasError = isDiskDeviceMountPointMissing(formik, index);
         rows.push(
           getConfigurationRowBase({
-            className: "no-border-top inherited-with-form device-last-row",
+            className: "no-border-top inherited-with-form",
             configuration: (
               <Label
                 forId={`devices.${index}.path`}
-                required
                 className="u-text--muted"
+                required
               >
                 Mount point
               </Label>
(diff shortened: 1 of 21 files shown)


src/util/instanceAndProfilePayloads.tsx
```diff
@@ -87,6 +89,17 @@ export const formDeviceToPayload = (devices: FormDevice[]) => {
...
+        // LXD only supports recursive for additional bind-mounted host paths.
+        if (isVolumeDevice({ ...item, name } as FormDiskDevice)) {
+          delete item.recursive;
+          delete item.shift;
+        }
+
+        // The kernel does not support recursive read-only bind mounts.
+        if (item.readonly === "true") {
+          delete item.recursive;
+        }
```
edlerd: This should probably be surfaced to the user as an error and not silently ignored.

src/util/instanceAndProfilePayloads.tsx
```diff
@@ -87,6 +89,17 @@ export const formDeviceToPayload = (devices: FormDevice[]) => {
       if (item.type === "disk") {
         const { bare, ...rest } = item;
         item = { ...bare, ...rest };
+
+        // LXD only supports recursive for additional bind-mounted host paths.
+        if (isVolumeDevice({ ...item, name } as FormDiskDevice)) {
+          delete item.recursive;
+          delete item.shift;
+        }
+
```
edlerd: Do we need this? I think the form value builder should not create this combination.
```suggestion
```

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -330,6 +349,165 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+                  (isReadonlyDisk
+                    ? "Recursive read-only bind mounts are not supported"
+                    : undefined)
+                }
+              />
+            ),
+            override: "",
+          }),
+        );
+      }
```
edlerd, line 515: If recursive is available, it is mutually exclusive with read-only, right? Probably better to combine those into a single select with both as an option. It slightly diverts from the CLI and data model, though. But that also makes it clearer that only one of the two can be enabled at a time.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -330,6 +349,165 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+        getConfigurationRowBase({
+          className: classnames("no-border-top inherited-with-form", {
+            "device-last-row": isRequiredLastRow,
+          }),
+          configuration: (
+            <Label
+              forId={`devices.${index}.required`}
+              className="u-text--muted"
+            >
+              Required
```
edlerd: This needs a better label or some way of explaining what it means. The single word "Required" doesn't let the user know this will block the boot when the mount is unavailable.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -330,6 +349,165 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+      );
+
+      // The "shift" property cannot be used with custom storage volumes or VMs
+      if (showShiftField) {
+        rows.push(
+          getConfigurationRowBase({
+            className: "no-border-top inherited-with-form device-last-row",
+            configuration: (
+              <Label forId={`devices.${index}.shift`} className="u-text--muted">
+                Shift
```
edlerd: Here as well we need a bit more explanation as to what this is about. The metadata API should provide a good help text that we could surface in some way.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -330,6 +349,165 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
           }),
         );
       }
+
+      rows.push(
+        getConfigurationRowBase({
```
edlerd, line 423: In read mode we shouldn't show the dropdowns right away, but the value with an edit icon.

Image

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -152,16 +155,85 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
 
+      const isContainerOrProfile =
+        formik.values.entityType === "profile" ||
+        (formik.values as CreateInstanceFormValues).instanceType ===
+          "container";
+
+      const hasReadonly = item.disk.readonly !== undefined;
+      const hasRecursive = item.disk.recursive !== undefined;
+      const hasRequired = item.disk.required !== undefined;
+      const hasShift = isContainerOrProfile && item.disk.shift !== undefined;
```
edlerd: I think for all of those we need the notion of the LXD provided default value, that is available via the metadata api.

edlerd (commented):
Good start on this one. Some thoughts on making it a bit clearer to our users.

src/components/forms/DiskDeviceFormCustom.tsx
omarelkashef: I think that would be confusing and will add complexity. We will have to find a name that describes this combined field, and the fields will look different for different types of disks. Disabling with a message seems clearer to me. Wdyt?

edlerd: > Disabling with a message seems clearer to me. Wdyt?

Yes, that is more in line with the data model and CLI as well, agreed. I wonder how to clarify with the message considering the available space.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -112,6 +159,15 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
     </Button>
   );
 
+  const readOnlyYesNoValue = (fieldName: string, value: string | undefined) => (
+    <div className="custom-disk-read-mode">
+      <div className="custom-disk-value">
+        <b>{optionRenderer(value ?? "", optionYesNo) || "-"}</b>
```
edlerd: The fallback should better show the default from the metadata api.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -49,6 +63,39 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+    key: string;
+    required?: boolean;
+  }) => (
+    <div className="configuration-label-with-help">
+      <Label forId={id} required={required} className="u-text--muted">
+        {label}
+      </Label>
+      <span className="configuration-help">
+        <ConfigFieldDescription description={getDiskFieldDescription(key)} />
+      </span>
```
edlerd, line 93: We should have the description text also for inherited fields.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -49,6 +63,39 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+    label: string;
+    key: string;
+    required?: boolean;
+  }) => (
+    <div className="configuration-label-with-help">
+      <Label forId={id} required={required} className="u-text--muted">
+        {label}
+      </Label>
+      <span className="configuration-help">
+        <ConfigFieldDescription description={getDiskFieldDescription(key)} />
```
edlerd: I am not sure about the layout. In the table views (such as in "Resource limits") the labels are in bold and easy to distinguish from the help text. 

Maybe we should make the label here not muted and that makes it easy to distinguish description vs labels. Making the labels bold might complicate identifying individual devices, when many are connected of the same type (such as many disks on one instance/profile).

src/components/forms/DiskDeviceFormCustom.tsx
edlerd: Agreed, let's keep it as is.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -152,16 +172,93 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
+      const recursiveValue =
+        item.disk.recursive ?? getDiskOptionDefault("recursive");
+      const requiredValue =
+        item.disk.required ?? getDiskOptionDefault("required");
+      const shiftValue = item.disk.shift ?? getDiskOptionDefault("shift");
+
+      const hasReadonly = readonlyValue !== undefined;
+      const hasRecursive = recursiveValue !== undefined;
+      const hasRequired = requiredValue !== undefined;
+      const hasShift = isContainerOrProfile && shiftValue !== undefined;
```
edlerd: This needs to consider the disk type. Currently it always shows defaults for all four, even if only two of them are applicable.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -152,16 +172,93 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
+            !hasReadonly && !hasRecursive && !hasRequired && !hasShift
+              ? "device-last-row"
+              : undefined,
         }),
       );
+
+      if (hasReadonly) {
+        rows.push(
+          getInheritedDeviceRow({
+            label: "Read-only",
```
edlerd: Here on the inherited fields, we should also show the description next to the label.

src/context/useConfigOptions.tsx
```diff
@@ -0,0 +1,14 @@
+import { useQuery, type UseQueryResult } from "@tanstack/react-query";
+import { fetchConfigOptions } from "api/server";
+import { useSupportedFeatures } from "context/useSupportedFeatures";
+import { queryKeys } from "util/queryKeys";
+import type { LxdMetadata } from "types/config";
+
+export const useConfigOptions = (): UseQueryResult<LxdMetadata | null> => {
```
edlerd, line 7: Good idea to extract this 👍

edlerd (commented):
Good iteration, a few more thoughts on the changes below.

edlerd:
The help text from the API "Whether to fail if the source doesn’t exist" should probably be adjusted to talk about failing the boot. Could you mention this to the core team or make a PR directly with the suggestion?

src/components/forms/DiskDeviceFormCustom.tsx
omarelkashef: Should we also unmute the rows in the root disk section, or it looks fine different?

image

edlerd: Let's do it consistently in all device inputs, not only disk but also the other types.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -49,6 +63,43 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
   const existingDeviceNames = getExistingDeviceNames(formik.values, profiles);
   const isProfile = formik.values.entityType === "profile";
+  const { data: configOptions } = useConfigOptions();
+
+  const diskConfigOptions = (
+    configOptions?.configs as Record<string, LxcConfigOptionCategories>
+  )?.["device-disk"];
+  const diskConfigFields = diskConfigOptions
+    ? toConfigFields(diskConfigOptions)
+    : [];
```
edlerd:
```suggestion
  const diskOptions = configOptions?.configs?.["device-disk"];
  const configFields = diskOptions ? toConfigFields(diskOptions) : [];
```
Better extend the `LxdMetadata` type with `"device-disk": LxcConfigOptionCategories;` to avoid the cast here.

Also this can be slightly nicer with more concise namings.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -49,6 +63,43 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+
+  const getDiskFieldDescription = (key: string) => {
+    return diskConfigFields.find((field) => field.key === key)?.shortdesc;
+  };
+
+  const getDiskOptionDefault = (option: string): string | undefined => {
+    return diskConfigFields.find((field) => field.key === option)?.default;
+  };
```
edlerd:
```suggestion
  const getDescription = (key: string) => {
    return configFields.find((field) => field.key === key)?.shortdesc;
  };

  const getDefault = (option: string): string | undefined => {
    return configFields.find((field) => field.key === option)?.default;
  };
```
We can use more generic names on these as they are not specific.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -112,6 +163,23 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+    fieldName: string,
+    value: string | undefined,
+    option: string,
+  ) => {
+    const displayValue = value ?? getDiskOptionDefault(option);
+
+    return (
+      <div className="custom-disk-read-mode">
+        <div className="custom-disk-value">
+          <b>{optionRenderer(displayValue ?? "", optionYesNo) || "-"}</b>
```
edlerd:
```suggestion
          <b>{optionRenderer(displayValue, optionYesNo) || "-"}</b>
```
Can we remove the extra fallback here?

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -40,6 +48,33 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
+
+  const getDiskOptionDefault = (option: string): string | undefined => {
+    const configField = diskConfigFields.find((item) => item.key === option);
+
+    return configField?.default || undefined;
+  };
+
+  const getDiskFieldDescription = (key: string) => {
+    return diskConfigFields.find((field) => field.key === key)?.shortdesc;
+  };
```
edlerd: Can make these generic to avoid the duplication and to establish a single source of truth?

edlerd (commented):
QA LGTM, good work. A few thoughts on the code to slightly simplify it.

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -209,14 +271,12 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
 
       const volumeDeviceSource = () =>
         getConfigurationRowBase({
-          className: classnames("no-border-top inherited-with-form", {
-            "device-last-row": isVolumeDevice(item) && item.path === undefined,
+          className: classnames("no-border-top inherited-with-form"),
+          configuration: diskFieldLabel({
```
edlerd: I don't think we need the help text here, also the help text is wrong, as this is the pool and the volume.

Image

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -286,24 +344,34 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
-                Mount point
-              </Label>
-            ),
+            className: "no-border-top inherited-with-form",
+            configuration: diskFieldLabel({
+              id: `devices.${index}.path`,
+              label: "Mount point",
+              key: "path",
+              required: true,
+            }),
```
edlerd: I think we don't need the help text here either.

Image

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -330,6 +398,176 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+            key: "readonly",
+          }),
+          inherited: readOnly ? (
+            readOnlyYesNoValue(
+              `devices.${index}.readonly`,
+              item.readonly,
+              "readonly",
+            )
+          ) : (
+            <Select
```
edlerd, line 437: There is a surprise here, when the LXD default is shown and we go into edit mode, the value is unset in the edit form. The behaviour is correct, but we should not surprise the user. 


https://github.com/user-attachments/assets/b6d728e0-bb38-4ff5-981f-fadfc5ddaa53

edlerd (commented):
Thanks for the adjustments, a few concerns on the latest changes.

src/components/forms/DiskDeviceFormCustom.tsx
omarelkashef: I see you what you mean. Do you have suggestions not to surprise the user? I think displaying the default will complicate the component.

edlerd: I see two options. Either in read mode add an indicator that clarifies this is the LXD default. Though that complicates the read mode. Or change the caption of the select dropdown for the unselected from "Select an option" to "LXD default (Yes)" or something similar. So the edit mode is not a surprise.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -129,7 +161,7 @@ const DiskDeviceFormInherited: FC<Props> = ({
       if (isHostDiskDevice(item.disk)) {
         rows.push(
           getInheritedDeviceRow({
-            label: "Host path",
+            label: inheritedFieldLabel("Host path", "source", false),
```
edlerd: Let's undo this change, this can stay as it was.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -40,6 +51,27 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
   const readOnly = (formik.values as EditInstanceFormValues).readOnly;
+  const { data: configOptions } = useConfigOptions();
+
+  const diskOptions = configOptions?.configs?.["device-disk"];
+  const diskConfigFields = diskOptions ? toConfigFields(diskOptions) : [];
+
+  const inheritedFieldLabel = (
+    label: string,
+    key: string,
+    showDescription = true,
```
edlerd:
```suggestion
```
Instead of this extra argument, we can simply not use the function. No need for the extra wrapper in that case.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -139,7 +171,7 @@ const DiskDeviceFormInherited: FC<Props> = ({
       } else {
         rows.push(
           getInheritedDeviceRow({
-            label: "Pool / volume",
+            label: inheritedFieldLabel("Pool / volume", "pool", false),
```
edlerd: Here as well, this can stay plain text as it was.

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -152,16 +184,106 @@ const DiskDeviceFormInherited: FC<Props> = ({
...
+
+      const hasReadonly = readonlyValue !== undefined;
+      const hasRecursive = recursiveValue !== undefined;
+      const hasRequired = requiredValue !== undefined;
+      const hasShift = shiftValue !== undefined;
+
+      if (showMountPointField) {
+        rows.push(
+          getInheritedDeviceRow({
+            label: inheritedFieldLabel("Mount point", "path", false),
```
edlerd:
```suggestion
            label: "Mount point",
```

src/components/forms/DiskDeviceFormInherited.tsx
```diff
@@ -40,6 +51,27 @@ const DiskDeviceFormInherited: FC<Props> = ({
   project,
 }) => {
   const readOnly = (formik.values as EditInstanceFormValues).readOnly;
+  const { data: configOptions } = useConfigOptions();
+
+  const diskOptions = configOptions?.configs?.["device-disk"];
+  const diskConfigFields = diskOptions ? toConfigFields(diskOptions) : [];
+
+  const inheritedFieldLabel = (
```
edlerd:
```suggestion
  const labelWithHelpText = (
```
A bit clearer?

src/components/forms/DiskDeviceFormCustom.tsx
```diff
@@ -253,14 +335,13 @@ const DiskDeviceFormCustom: FC<Props> = ({ formik, project, profiles }) => {
...
+            label: "Host path",
+            key: "source",
+            showDescription: false,
           }),
-          configuration: (
-            <Label forId={`devices.${index}.source`} className="u-text--muted">
-              Host path
-            </Label>
-          ),
           inherited: readOnly ? (
```
edlerd, line 339: In this file we can also restore the original labels without the helper function and then remove the extra `showDescription` flag from the function.

edlerd (commented):
QA looks good. Some tiny ideas to simplify a bit, then should be good to merge.

One other thing I realized on QA is this modal should have a positive appearance for `Attach`. And the modal is missing the hidden field to allow submission with the "enter" key. But both can be addressed independently as well.

Image

edlerd (approved):
LGTM, good work 👍

Kxiru:
Linking to issue: https://github.com/canonical/lxd-ui/issues/2108

Issue #2082, opened 2026-06-19
API User management in the UI


vosdev:
Hello hello,

I wanted to create a user with limited (read only) permissions in the UI. The UI then provides instructions for UI and CLI usage. But my goal for this limited user was direct API usage with the python LXD library. 

Can you add instructions to obtain certificates directly, so without trust tokens meant for cli or web ui?

edlerd:
Hey and thanks for opening the issue :)

We do plan to add bearer token management to the UI in the near future until October. I think that will allow you to manage access for python LXD tooling from the UI.

We were discussing generating certificates right from the UI for other people. But found it not ideal, as private keys would then be generated by an admin and shared with the intended individual. I see your use case with APIs is slightly different and might be a good reason to add. Then again I wonder if baerer tokens are not the better approach for APIs.

vosdev:
Hmm but a bearer tokens often come with limited lifetime. My pipelines would still have to come up with a way to generate a bearer token, meaning authenticate in some way, right?

My current use case is a dynamic ansible inventory script, talking to LXD to obtain the running instances to build an inventory from at each ansible playbook run. Using certificates for authentication.

edlerd:
Bearer tokens can be used for long term api access. Their usage is fairly straight forward with adding an `Authentication: Bearer <token>` header to the API request.

The [docs](https://canonical.com/lxd/docs/latest/authentication/#authentication-bearer) are lightly off with mentioning 24h lifetime. That is only true for UI temporary access tokens meant for onboarding and will be fixed shortly in the docs.

A bearer identiy created `lxc auth identity create bearer/foo`, and token issued for it `lxc auth identity token issue bearer/foo` will have a 10 year lifetime by default.

I think that resolves your use case and our intended update of the UI to support creating and issuing bearer tokens will be the right way forward.

vosdev:
That sounds like a much better solution :)

vosdev:
@Kxiru https://github.com/canonical/lxd/pull/18791 is for another feature request of mine, and it is related to this issue

kimanhou:
Hi @vosdev 

Development is still on-going on both back end and front end.

edlerd:
This has been completed and is available in `latest/edge` today.

PR #2134, opened 2026-07-09
WD-36589 token bearer identities create

kimanhou:
## Done

- feat(Identites): create token bearer identities, with type `Client token bearer` and `DevLXD token bearer`.
- Identity creation is now a 2-step flow:
    1. Choose an identity type
    2. Input fields: name, expiry if bearer, groups

## QA

1. Run the LXD-UI:
3. Perform the following QA steps:
    - Go to Permissions > Identities
    - Click "+ Create identity".
    - Select "Bearer token (Main API)" and click "Continue".
    - Input a valid name, an expiry (optional) and select groups. Click "Create identity".
    - Make sure the identity is correctly create.
    - Do the same with selecting "Bearer token (DevLXD)" in step 1 of the flow.

## Screenshots

**Step 1**
image

**Step 2**
identity type: TLS
image

identity type: Client token bearer
image

identity type: DevLXD token bearer
image

**Success modal**
identity type: TLS (unchanged)
image

identity type: Client token bearer
image

identity type: DevLXD token bearer
image


src/pages/login/CreateTlsIdentityWithBearerToken.tsx
```diff
@@ -10,19 +10,20 @@ import {
 } from "@canonical/react-components";
```
omarelkashef, line 10: I am assuming this page lets users with bearer tokens create new identities, right? Should it be renamed  `CreateIdentityWithBearerToken`

src/pages/permissions/forms/NameWithGroupForm.tsx
```diff
@@ -1,14 +1,13 @@
...
 import { Form, Input } from "@canonical/react-components";
 import type { FormikProps } from "formik/dist/types";
 import type { IdpGroupFormValues } from "types/forms/idpGroup";
-import type { TLSIdentityFormValues } from "types/forms/tlsIdentity";
+import type { IdentityFormValues } from "types/forms/identity";
 
-interface Props {
-  formik: FormikProps<IdpGroupFormValues | TLSIdentityFormValues>;
-}
+type Props =
```
omarelkashef: Does this change has to be done this way? It deviated from the normal way we define props.

src/types/forms/identity.d.ts
```diff
@@ -0,0 +1,15 @@
+import type { IDENTITY_TYPE } from "util/permissionIdentities";
```
omarelkashef: I do not think we should be importing from other directories into types. The dependency should always point in the other direction.

omarelkashef (commented):
I did QA and a quick run through the code. Very nice work. I like the new flow alot.

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
kimanhou: No it actually creates a TLS identity. It's the screen we see when you authenticated with a magic link.

src/types/forms/identity.d.ts
kimanhou: IDENTITY_TYPE had to be defined in permissionIdentities because it is used at runtime for lookups and object keys in the same file. A `.d.ts` file does not emit runtime JavaScript.
And I want only one source of truth.

src/util/permissionIdentities.tsx
```diff
@@ -180,3 +202,174 @@ export const getIdentityName = (identity?: LxdIdentity): string => {
...
+> = {
+  [IDENTITY_TYPE.TLS]: {
+    codeSnippetTitle: "Your identity trust token",
+    notificationTitle: "Copy the identity trust token",
+    notificationBody: "to authenticate your client",
+    howToUseCli: (token: string) => (
+      <>
+        For use with the LXC command-line tool, run:
+        <CodeSnippetWithCopyButton
+          code={`lxc remote add ${location.hostname} ${token}`}
```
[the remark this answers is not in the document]
kimanhou: I don't want to change existing code.

src/pages/login/CreateTlsIdentityWithBearerToken.tsx
omarelkashef: Oh I see, thank you for clarifying!

src/pages/permissions/forms/NameWithGroupForm.tsx
```diff
@@ -1,14 +1,13 @@
...
-import type { TLSIdentityFormValues } from "types/forms/tlsIdentity";
+import type { IdentityFormValues } from "types/forms/identity";
 
 interface Props {
-  formik: FormikProps<IdpGroupFormValues | TLSIdentityFormValues>;
+  formik: FormikProps<IdpGroupFormValues> | FormikProps<IdentityFormValues>;
 }
 
-const NameWithGroupForm: FC<Props> = ({ formik }) => {
+const NameWithGroupForm = ({ formik }: Props) => {
```
edlerd: Why change the header here? The former is in line with the project style.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,352 @@
...
+                />
+              </div>
+              <Notification
+                severity="information"
+                title={IDENTITY_TYPE_HELP_TEXT[selectedIdentityType].title}
+              >
+                {IDENTITY_TYPE_HELP_TEXT[selectedIdentityType].description}
+              </Notification>
+            </div>
+          </ScrollableContainer>
```
edlerd, line 226: Let's adjust the layout here away from the form to something more like the instance migration:

Image

We can use muted text below the options instead of the information notification.

edlerd: The title can be "Choose an identity type" and the options can use their respective icons. I think we defined the secret key icon (similar to the one used for bucket keys) to be used for both bearer tokens.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -59,24 +59,33 @@ const CreateIdentityModal: FC<Props> = ({ onClose, token, identityName }) => {
         <>
           <CodeSnippetWithCopyButton
             code={token}
-            title="Your identity trust token"
+            title={
+              CREATE_IDENTITY_MODAL_TEXT[identity.identityType].codeSnippetTitle
```
edlerd: I am not sure about this pattern of defining a constant and reading it here. Maybe a better approach is to expose a `title`, `notification`, `howToUse` and `howToUseUi` prop and do the selection in the caller. 

That seems simpler than passing in a type and having the lookup from here. Wdyt?

src/util/permissionIdentities.tsx
```diff
@@ -180,3 +202,174 @@ export const getIdentityName = (identity?: LxdIdentity): string => {
...
+            blocks={[
+              {
+                appearance: CodeSnippetBlockAppearance.LINUX_PROMPT,
+                code: `curl -k -H "Authorization: Bearer ${token}" https://<lxd_address>/1.0
+{
+  ...
+  "metadata": {
+    "auth":"trusted"
+  }
+}`,
```
edlerd: Let's use the `location.hostname` just like in the other existing cases. Also I think we should avoid the answer section so this is a valid command to copy and run.

src/util/permissionIdentities.tsx
```diff
@@ -180,3 +202,174 @@ export const getIdentityName = (identity?: LxdIdentity): string => {
...
+            className="u-no-margin--bottom"
+            key="code-snippet"
+          />,
+          <span
+            className="u-text--muted p-text--small u-sv3"
+            key="code-snippet-help-text"
+          >
+            You can replace <code>&lt;lxd_address&gt;</code> with the hostname
+            or IP address of your LXD server.
+          </span>,
```
edlerd:
```suggestion
```
Probably not needed.

src/util/permissionIdentities.tsx
```diff
@@ -180,3 +202,174 @@ export const getIdentityName = (identity?: LxdIdentity): string => {
...
+          <DocLink docPath="/howto/auth_bearer/" key="learn-more-link">
+            Learn more about how to authenticate to the LXD API using bearer
+            tokens
+          </DocLink>,
```
edlerd: I think this is missing the external link icon.

src/util/permissionIdentities.tsx
```diff
@@ -180,3 +202,174 @@ export const getIdentityName = (identity?: LxdIdentity): string => {
...
+            You can verify trust by checking the <code>auth</code> field in the
+            response of <code>GET /1.0</code>:
+          </>,
+          <CodeSnippet
+            blocks={[
+              {
+                appearance: CodeSnippetBlockAppearance.LINUX_PROMPT,
+                code: `lxc exec c1 --env TOKEN=${token} -- bash
+root@c1# curl -H "Authorization: Bearer \${TOKEN}" -s --unix-socket /dev/lxd/sock http://custom.socket/1.0
+{"state":"Started","api_version":"1.0","instance_type":"container","location":"my-host","auth":"trusted"}`,
```
edlerd:
```suggestion
                code: `curl -H "Authorization: Bearer \${TOKEN}" -s --unix-socket /dev/lxd/sock http://custom.socket/1.0`,
```
I think this should only be the command with the token to be used. The surrounding text should explain that you use it inside of an instance.

src/util/permissionIdentities.tsx
```diff
@@ -180,3 +202,174 @@ export const getIdentityName = (identity?: LxdIdentity): string => {
...
+            key="code-snippet"
+          />,
+          <span
+            className="u-text--muted p-text--small u-sv3"
+            key="code-snippet-help-text"
+          >
+            Replace <code>c1</code> with your instance name.
+          </span>,
+          <DocLink docPath="/howto/devlxd_authenticate/" key="learn-more-link">
+            Learn more about how to authenticate to the DevLXD API
```
edlerd:
```suggestion
            How to authenticate to the DevLXD API
```
a bit more concise and it should have an icon

src/util/permissionIdentities.tsx
```diff
@@ -180,3 +202,174 @@ export const getIdentityName = (identity?: LxdIdentity): string => {
...
+          <span
+            className="u-text--muted p-text--small u-sv3"
+            key="code-snippet-help-text"
+          >
+            You can replace <code>&lt;lxd_address&gt;</code> with the hostname
+            or IP address of your LXD server.
+          </span>,
+          <DocLink docPath="/howto/auth_bearer/" key="learn-more-link">
+            Learn more about how to authenticate to the LXD API using bearer
+            tokens
```
edlerd:
```suggestion
            How to authenticate to the LXD API using bearer
            tokens
```
A bit more concise.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,352 @@
...
+              >
+                Continue
+              </ActionButton>
+            </>
+          ) : (
+            <>
+              <Button
+                appearance="base"
+                onClick={() => {
+                  setCurrentStep(1);
```
edlerd: I think the form should reset the fields once I choose a different type after going back. Wdyt?

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,352 @@
...
+                placeholder="e.g. 1d 3H 5M"
+                help={
+                  <>
+                    Space-separated durations: {"<number><unit>"}. Units are
+                    case-sensitive: y, m, w, d, H, M, S<br />
+                    If left empty, bearer tokens expire after 10 years, unless
+                    they are manually revoked.
+                  </>
+                }
+              />
```
edlerd: A better approach might be a selector with two options: "default (10 years)" and "custom". Once custom is selected an input appears next to it that allow to enter.

Kind of like this:

Image

and after selecting the other option:

Image

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,352 @@
...
+    queryClient.invalidateQueries({
+      queryKey: [queryKeys.identities],
+    });
+  };
+
+  const onSuccessInternal = (identity: IdentityFormValues, token: string) => {
+    onSuccess(identity, token);
+    invalidateCache();
+    closePanel();
+  };
```
edlerd: Can we move those things into the `onSuccess` handler in the caller and remove the `onSuccessInternal`?

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,352 @@
...
+              </Notification>
+            </div>
+          </ScrollableContainer>
+        ) : (
+          <>
+            <OutputField
+              id="identity-type"
+              label="Identity type"
+              value={IDENTITY_TYPE_HELP_TEXT[selectedIdentityType].title}
+            />
```
edlerd: This field looks a bit improper. Can we make the choice part of the title instead?

Image

edlerd (commented):
Good start, works technically like a charm. A few thoughts, mostly on the layout and representation of the flow.

omarelkashef:
When I select an identity type then go back without changing anything and choose another type, I get a name validation error. Is that expected from a UX perspective? Can it be solved by resetting errors in `handleIdentityTypeChange`?

image

kimanhou:
@omarelkashef it should be fixed with the new layout

src/api/auth-identities.tsx
```diff
@@ -110,3 +111,55 @@ export const createFineGrainedTlsIdentity = async (
...
+export const createAndIssueBearerToken = async (
+  name: string,
+  type: BearerIdentityType,
+  groups: string[],
+  expiry?: string,
+): Promise<{
+  token: string;
+}> => {
+  await createBearerIdentity(name, type, groups);
+  return await issueBearerToken(name, expiry);
```
omarelkashef: This can be left like this for simplicity, but a network error can cause a frustrating experience for the user if the identity creation succeeds but issuing the token fails because the user will retry with the same identity name and get an error that the name already exists.

src/pages/permissions/CreateIdentity.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const notify = useNotify();
+  const queryClient = useQueryClient();
+  const { openPortal, closePortal, isOpen, Portal } = usePortal({
+    programmaticallyOpen: true,
+  });
+  const [token, setToken] = useState("");
+  const [identity, setIdentity] = useState<IdentityFormValues>({
+    name: "",
+    groups: [],
+    identityType: "tls-certificate",
```
omarelkashef: Should this be IDENTITY_TYPE.TLS in case the name changes?

omarelkashef (commented):
Did a code review but not QA because the demo server is down. I found two small issues.

src/api/auth-identities.tsx
kimanhou: In the Edit, the user will be able to issue a new token for an identity.

kimanhou:
@omarelkashef I don't know why the demo server is down, as the deployment in Jenkins was successful.

src/api/auth-identities.tsx
omarelkashef: It is not easy to a notice a new identity is created in the table, especially that they will get a notification "Identity creation failed"

src/pages/permissions/CreateIdentity.tsx
```diff
@@ -0,0 +1,65 @@
+import { useState, type FC } from "react";
+import { useNotify, usePortal } from "@canonical/react-components";
+import { useQueryClient } from "@tanstack/react-query";
+import CreateIdentityModal from "pages/permissions/CreateIdentityModal";
+import CreateIdentityPanel from "pages/permissions/panels/CreateIdentityPanel";
+import type { IdentityFormValues } from "types/forms/identity";
+import usePanelParams, { panels } from "util/usePanelParams";
+import {
+  CREATE_IDENTITY_MODAL_TEXT,
+  type IDENTITY_TYPE,
```
edlerd:
```suggestion
  IDENTITY_TYPE,
```
This breaks the page and is not a type.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,359 @@
...
+  const goToTypeSelectionStep = () => {
+    formik.setTouched({}, false);
+    setCurrentStep(1);
+  };
+
+  const headerTitle = isTypeSelectionStep ? (
+    "Choose an identity type"
+  ) : (
+    <BackLink
+      linkText="Choose an identity type"
```
edlerd:
```suggestion
      linkText="Choose type"
```
Let's shorten the back link so the title always fits in a single line

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,359 @@
...
+              >
+                Back
+              </Button>
+              <Button
+                appearance="base"
+                onClick={closePanel}
+                className="u-no-margin--bottom"
+              >
+                Cancel
+              </Button>
```
edlerd, line 325: We should have just one cancel button and it should change behaviour. When on step 2, it should go back to step 1. When on step 1 it should close the panel. We should not show two buttons.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,359 @@
...
+                  Token expiry
+                </label>
+                <RadioInput
+                  label="Default (10 years)"
+                  checked={!isCustomExpiry}
+                  onChange={() => {
+                    setIsCustomExpiry(false);
+                    formik.setFieldValue("expiry", "");
+                  }}
+                />
```
edlerd, line 243: There is an annoyance, when I enter step 2 and want to click on "custom", the error for name pops in and shifts the layout, so my click goes somewhere else. Shall we disable other inputs and fields, until a name is entered?

https://github.com/user-attachments/assets/2126bee5-ae7f-45b1-9bbf-6ea8becf450f

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,359 @@
...
+                    }
+                    placeholder="e.g. 1d 3H 5M"
+                    disabled={!isCustomExpiry}
+                    help={
+                      <>
+                        Space-separated durations: {"<number><unit>"} <br />
+                        Units are case-sensitive: y, m, w, d, H, M, S
+                      </>
+                    }
+                  />
```
edlerd, line 276: I am not sure if radios with the always-visible input are the best option. I was thinking a select that has the two options "Default (10 years)" and "Custom". The input is initially hidden as long as "Default..." is selected. When switching the selection to "custom", the input is displayed on the right of the select with help text below.

That makes the whole side panel a bit lighter when seeing it for the first time. And it is not taking so much vertical space, leaving more room for the group selection below. Wdyt?

src/util/permissionIdentities.tsx
```diff
@@ -208,3 +226,158 @@ export const getIdentityIconType = (
...
+    codeSnippetTitle: "Your API bearer token",
+    notificationTitle: "Copy the API bearer token",
+    notificationBody: "to authenticate your clients and automated tools.",
+    howToUse: (token: string) => (
+      <List
+        className="u-no-margin--bottom"
+        items={[
+          <>
+            The token must be set as a bearer token in the{" "}
+            <code>Authorization</code> header.
```
edlerd:
```suggestion
            Set the bearer token in the{" "}
            <code>Authorization</code> header.
```
A bit more concise.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -59,24 +72,21 @@ const CreateIdentityModal: FC<Props> = ({ onClose, token, identityName }) => {
...
               setConfirmed(true);
             }}
           />
-          <Notification
-            severity="caution"
-            title="Copy the identity trust token"
-          >
-            You will need the token to authenticate your client. <br />
+          <Notification severity="caution" title={notification}>
+            You will need this token. <br />
```
edlerd:
```suggestion
```
I think we can leave this out.

src/util/permissionIdentities.tsx
```diff
@@ -208,3 +226,158 @@ export const getIdentityIconType = (
...
+            code={`curl -H "Authorization: Bearer \${TOKEN}" -s --unix-socket /dev/lxd/sock http://custom.socket/1.0`}
+            tooltipMessage="Copy command"
+            className="u-no-margin--bottom"
+            key="code-snippet"
+          />,
+          <span
+            className="u-text--muted p-text--small u-sv3"
+            key="code-snippet-help-text"
+          >
+            Run this command inside your instance shell.
```
edlerd:
```suggestion
            Run this command inside your instance.
```
Maybe a bit more concise?

src/util/permissionIdentities.tsx
```diff
@@ -208,3 +226,158 @@ export const getIdentityIconType = (
...
+          <>
+            The token must be set as a bearer token in the{" "}
+            <code>Authorization</code> header.
+          </>,
+          <>
+            You can verify trust by checking the <code>auth</code> field in the
+            response metadata of <code>GET /1.0</code>:
+          </>,
+          <CodeSnippetWithCopyButton
+            code={`curl -k -H "Authorization: Bearer ${token}" https://${location.hostname}/1.0`}
```
edlerd: We are missing the port here
```suggestion
            code={`curl -k -H "Authorization: Bearer ${token}" ${location.origin}/1.0`}
```

src/util/permissionIdentities.tsx
```diff
@@ -208,3 +226,158 @@ export const getIdentityIconType = (
...
+          <>
+            It must be set as a bearer token in the <code>Authorization</code>{" "}
+            header.
+          </>,
+          <>
+            You can verify trust by checking the <code>auth</code> field in the
+            response of <code>GET /1.0</code>:
+          </>,
+          <CodeSnippetWithCopyButton
+            code={`curl -H "Authorization: Bearer \${TOKEN}" -s --unix-socket /dev/lxd/sock http://custom.socket/1.0`}
```
edlerd:
```suggestion
            code={`curl -H "Authorization: Bearer ${token}" -s --unix-socket /dev/lxd/sock http://custom.socket/1.0`}
```
We should give the actual token here and pass it in as well.

src/pages/permissions/CreateIdentityModal.tsx
```diff
@@ -30,7 +43,7 @@ const CreateIdentityModal: FC<Props> = ({ onClose, token, identityName }) => {
           Identity{" "}
           <ResourceLabel
             type="certificate"
```
edlerd: This should be dynamic. For the new types it should be the secret key icon.

src/util/permissionIdentities.tsx
```diff
@@ -208,3 +226,158 @@ export const getIdentityIconType = (
...
+
+export const CREATE_IDENTITY_MODAL_TEXT: Record<
+  IdentityType,
+  {
+    codeSnippetTitle: string;
+    notificationTitle: string;
+    notificationBody: string;
+    howToUseCli?: (token: string) => ReactNode;
+    howToUseUi?: ReactNode;
+    howToUse?: (token: string) => ReactNode;
```
edlerd: This is a bit of a wired interface. Maybe more straight forward is to have only those two. And in the modal we evaluate all and if more than one is set, we show the tabs.
```suggestion
    howToUseCli?: (token: string) => ReactNode;
    howToUseUi?: ReactNode;
```

src/pages/permissions/CreateIdentity.tsx
```diff
@@ -0,0 +1,65 @@
...
+      {isOpen && (
+        <Portal>
+          <CreateIdentityModal
+            onClose={closePortal}
+            token={token}
+            identity={identity}
+            title={identityModalText.codeSnippetTitle}
+            notification={`${identityModalText.notificationTitle} ${identityModalText.notificationBody}`}
+            howToUse={identityModalText.howToUse}
+            howToUseCli={identityModalText.howToUseCli}
```
edlerd:
```suggestion
            howToUseCli={identityModalText.howToUseCli(token)}
```
Shall we evaluate it at this point and not pass the callback around?

src/pages/permissions/CreateIdentity.tsx
```diff
@@ -0,0 +1,65 @@
...
+    programmaticallyOpen: true,
+  });
+  const [token, setToken] = useState("");
+  const [identity, setIdentity] = useState<IdentityFormValues>({
+    name: "",
+    groups: [],
+    identityType: IDENTITY_TYPE.TLS,
+    expiry: "",
+  });
+  const identityModalText = CREATE_IDENTITY_MODAL_TEXT[identity.identityType];
```
edlerd:
```suggestion
  const identityModalCaptions = CREATE_IDENTITY_MODAL_CAPTIONS[identity.identityType];
```
A bit more precise?

src/api/auth-identities.tsx
edlerd: In other cases on storage pool creation, we issue a delete on the newly created pool if there is an error. I think we might want to do the same here as well. If there is an error, catch the error and delete, then rethrow the error to be displayed upstream.

edlerd (commented):
Thanks for the iterations, this is a great direction! Some more ideas to improve code and the flow/layout.

src/pages/permissions/panels/CreateIdentityPanel.tsx
kimanhou: I copied the design from IP addresses in when creating a network

image

kimanhou:
> In other cases on storage pool creation, we issue a delete on the newly created pool if there is an error. 

I can't find it. Is it for clustered pool ?

```
export const createClusteredPool = async (
  pool: LxdStoragePool,
  hasStorageAndNetworkOperations: boolean,
  sourcePerClusterMember?: ClusterSpecificValues,
  zfsPoolNamePerClusterMember?: ClusterSpecificValues,
  sizePerClusterMember?: ClusterSpecificValues,
): Promise<LxdOperationResponse> => {
  const { memberPoolPayload, clusterPoolPayload } =
    getClusterAndMemberPoolPayload(pool);

  const operations = await Promise.allSettled(
    clusterMembers.map(async (item) => {
      const clusteredMemberPool = {
        ...memberPoolPayload,
        config: {
          ...memberPoolPayload.config,
          source: sourcePerClusterMember?.[item.server_name],
          size: sizePerClusterMember?.[item.server_name],
          "zfs.pool_name": zfsPoolNamePerClusterMember?.[item.server_name],
        },
      };
      const operation = await createPool(clusteredMemberPool, item.server_name);
      return { operation, member: item.server_name };
    }),
  );

  const pendingOperations = operations.map((res) => {
    if (res.status === "rejected") {
      throw res?.reason as Error;
    }
    return res.value;
  });

  if (hasStorageAndNetworkOperations) {
    await Promise.all(
      pendingOperations.map(async ({ operation, member }) => {
        await waitForOperation(operation.metadata.id, member);
      }),
    );
  }

  return createPool(clusterPoolPayload);
};
```

src/pages/permissions/panels/CreateIdentityPanel.tsx
kimanhou: I prefer the radio buttons. It's a draft, the width of the select should be fixed but still I'm not convinced.

image


https://github.com/user-attachments/assets/3ae1a06d-4e35-4bc2-8f26-700551b74d3a

edlerd:
> > In other cases on storage pool creation, we issue a delete on the newly created pool if there is an error.
> I can't find it. Is it for clustered pool ?

I mixed this up with networks. For networks we do clean them up on failure.

src/api/auth-identities.tsx
```diff
@@ -110,3 +111,65 @@ export const createFineGrainedTlsIdentity = async (
...
+    return await issueBearerToken(name, expiry);
+  } catch (e) {
+    // Token issuance failed — delete the identity we just created and
+    // rethrow the original error so the caller can display it.
+    await fetch(
+      `${ROOT_PATH}/1.0/auth/identities/bearer/${encodeURIComponent(name)}`,
+      { method: "DELETE" },
+    ).catch(() => undefined);
+    throw e;
+  }
```
edlerd: Better use `deleteIdentity` instead of re-implementing it here.

src/pages/permissions/panels/CreateIdentityPanel.tsx
edlerd: okay, lets stay with the radios 👍

src/api/auth-identities.tsx
kimanhou: I originally tried but it looks ugly because of type mismatch

```
    deleteIdentity({
      authentication_method: "bearer",
      id: name,
    } as LxdIdentity).catch(() => undefined);
    throw e;
```

kimanhou:
> I mixed this up with networks. For networks we do clean them up on failure.

I modified the API layer. For network, we handle that in the component. I think it's cleaner in the API layer, no ?

Kxiru:
Linking this PR to the request for API access & management issue: https://github.com/canonical/lxd-ui/issues/2082

src/api/auth-identities.tsx
edlerd: It is a bit more verbose, but maybe cleaner as we don't reimplement the deletion:
```
    const identity = {
      id: name,
      name,
      type,
      authentication_method: AUTH_METHOD.BEARER,
      fine_grained: true,
    };
    deleteIdentity(identity).catch(() => undefined);
 ```
I think the `tls_certificate` should be optional on the `LxdIdentity` type.

**Comment by edlerd** [comment]
> I modified the API layer. For network, we handle that in the component. I think it's cleaner in the API layer, no ?

Yeah and no. You can argue both ways. It could be considered business logic and shouldn't be in the API. Also, the function name with `and` in it is considered an antipattern, as it is doing too much at once.
Personally, I think both options are okay here. No strong opinions from my side.

File: src/pages/permissions/panels/CreateIdentityPanel.tsx [hunk 35]
```diff
@@ -0,0 +1,356 @@
...
+          name: "",
+          groups: [],
+          identityType,
+          expiry: "",
+        },
+      });
+      return;
+    }
+
+    formik.setFieldValue("identityType", identityType);
```
edlerd:
```suggestion
```
This seems not needed considering the lines above.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+  const groupsAdded = new Set(formik.values.groups ?? []);
+  const isNameInvalid = !formik.values.name || !!formik.errors.name;
+  const isTypeSelectionStep = currentStep === 1;
+  const goToTypeSelectionStep = () => {
+    formik.setTouched({}, false);
+    setCurrentStep(1);
+  };
+
+  const headerTitle = isTypeSelectionStep ? (
```
edlerd:
```suggestion
  const headerTitle = currentStep == "typeSelection" ? (
```
Let's not use numbers but static strings, then we can also drop this const and keep things simpler.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+import { type IdentityFormValues } from "types/forms/identity";
+
+interface Props {
+  onSuccess: (identity: IdentityFormValues, token: string) => void;
+}
+
+const CreateIdentityPanel: FC<Props> = ({ onSuccess }) => {
+  const panelParams = usePanelParams();
+  const notify = useNotify();
+  const [currentStep, setCurrentStep] = useState<1 | 2>(1);
```
edlerd:
```suggestion
  const [currentStep, setCurrentStep] = useState<"typeSelection" | "detailsForm">("typeSelection");
```
Makes using this a bit clearer.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+  const [selectedIdentityType, setSelectedIdentityType] =
+    useState<IdentityType>(IDENTITY_TYPE.TLS);
```
edlerd:
```suggestion
```
This is already in formik, we shouldn't duplicate the state here.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+  const [isCustomExpiry, setIsCustomExpiry] = useState(false);
```
edlerd:
```suggestion
```
Let's keep this in formik as well.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+    formik.setTouched({}, false);
+    setCurrentStep(1);
+  };
+
+  const headerTitle = isTypeSelectionStep ? (
+    "Choose an identity type"
+  ) : (
+    <BackLink
+      linkText="Choose type"
+      title={IDENTITY_TYPE_HELP_TEXT[selectedIdentityType].title}
```
edlerd:
```suggestion
      title={IDENTITY_TYPE_HELP_TEXT[formik.values.identityType].title}
```

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+        e.preventDefault();
+      }}
+      onClick={goToTypeSelectionStep}
+    />
+  );
+
+  return (
+    <SidePanel loading={isLoading} hasError={Boolean(error)}>
+      <SidePanel.Header>
+        <SidePanel.HeaderTitle key={isTypeSelectionStep ? "start" : "details"}>
```
edlerd:
```suggestion
        <SidePanel.HeaderTitle key={currentStep}>
```
A bit simpler.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+          {headerTitle}
+        </SidePanel.HeaderTitle>
+      </SidePanel.Header>
+      <NotificationRow className="u-no-padding" />
+      <SidePanel.Content className="u-no-padding">
+        {isTypeSelectionStep ? (
```
edlerd:
```suggestion
        {currentStpe === "typeSelection" ? (
```

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+          >
+            <div className="choose-identity-type">
+              <FormLink
+                icon="certificate"
+                title={IDENTITY_TYPE_HELP_TEXT[IDENTITY_TYPE.TLS].title}
+                subText={IDENTITY_TYPE_HELP_TEXT[IDENTITY_TYPE.TLS].description}
+                subTextBelowTitle
+                onClick={() => {
+                  handleIdentityTypeChange(IDENTITY_TYPE.TLS);
+                  setCurrentStep(2);
```
edlerd:
```suggestion
```
This should happen inside `handleIdentityTypeChange`.

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+    panelParams.clear();
+    notify.clear();
+  };
+
+  const onError = (e: unknown) => {
+    formik.setSubmitting(false);
+    notify.failure("Identity creation failed", e);
+  };
+
+  const handleIdentityTypeChange = (identityType: IdentityType) => {
```
edlerd:
```suggestion
  const selectType = (identityType: IdentityType) => {
```

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,356 @@
...
+                Cancel
+              </Button>
+            </>
+          ) : (
+            <>
+              <Button
+                appearance="base"
+                onMouseDown={(e: MouseEvent<HTMLButtonElement>) => {
+                  e.preventDefault();
+                }}
```
edlerd:
```suggestion
type="button"
```
This shouldn't be needed if we mark this as not a submit button.

src/util/permissionIdentities.tsx
```diff
@@ -208,3 +226,156 @@ export const getIdentityIconType = (
...
+> = {
+  [IDENTITY_TYPE.TLS]: {
+    title: "TLS Certificate",
+    description:
+      "A certificate-based identity for long-lived client authentication to the LXD API.",
+  },
+  [IDENTITY_TYPE.BEARER_CLIENT]: {
+    title: "Bearer token (Main API)",
+    description:
+      "An API key alternative to TLS certificates. Best for external automation tools, integrations, or remote scripts interacting with LXD over the network.",
```
edlerd:
```suggestion
      "An API key for external automation tools, integrations, or remote scripts interacting with LXD over the network.",
```
A bit shorter.

src/sass/_form-link.scss
```diff
@@ -1,28 +1,46 @@
 @import "vanilla-framework/scss/vanilla";
 
 button.form-link {
```
edlerd, line 3: Nice improvements as a drive by to the migration modal responsiveness!

edlerd (commented):
QA LGTM; few nitpicks on the code to simplify.

src/pages/permissions/panels/CreateIdentityPanel.tsx
kimanhou: Mmh how do we differentiate if the user has selected Custom but hasn't input anything yet and Default ?
The expiry would be empty in both cases but the expiry input should be enabled if the user has selected Custom.

edlerd: We can add the `isCustomExpiry` as a new field to the formik state. I wasn't suggesting to remove the state and merge it with the `expiry` state, just moving this react state to formik as this seems like form state.

src/util/permissionIdentities.tsx
```diff
@@ -208,3 +226,156 @@ export const getIdentityIconType = (
...
+export const BEARER_EXPIRY_PATTERN = /^(\d+[ymwdHMS])(?:\s+\d+[ymwdHMS])*$/;
+
+export const IDENTITY_TYPE_HELP_TEXT: Record<
+  IdentityType,
+  { title: string; description: ReactNode }
+> = {
+  [IDENTITY_TYPE.TLS]: {
+    title: "TLS Certificate",
+    description:
+      "A certificate-based identity for long-lived client authentication to the LXD API.",
```
edlerd:
```suggestion
      "A certificate-based identity for long-lived client authentication.",
```
I was wondering how to include the UI in the previous wording. Maybe shortening is a nice way of making this simpler and including both UI and API. Because the former sounded like it is not for the UI.

src/api/auth-identities.tsx
```diff
@@ -110,3 +112,69 @@ export const createFineGrainedTlsIdentity = async (
...
+      name,
+      type,
+      groups,
+    }),
+  }).then(handleResponse);
+};
+
+const issueBearerToken = async (
+  name: string,
+  expiry?: string,
```
edlerd:
```suggestion
  expiry: string,
```
This is initialized to empty string, it can never really be null.

src/api/auth-identities.tsx
```diff
@@ -110,3 +112,69 @@ export const createFineGrainedTlsIdentity = async (
...
+    .then((data: LxdApiResponse<{ token: string }>) => {
+      return data.metadata;
+    });
+};
+
+export const createAndIssueBearerToken = async (
+  name: string,
+  type: BearerIdentityType,
+  groups: string[],
+  expiry?: string,
```
edlerd:
```suggestion
  expiry: string,
```

src/types/forms/identity.d.ts
```diff
@@ -0,0 +1,9 @@
+import type { IdentityType } from "util/permissionIdentities";
+
+export interface IdentityFormValues {
+  name: string;
+  groups?: string[];
+  identityType: IdentityType;
+  expiry?: string;
```
edlerd:
```suggestion
  expiry: string;
```

src/types/forms/identity.d.ts
```diff
@@ -0,0 +1,9 @@
...
+  isCustomExpiry?: boolean;
```
edlerd:
```suggestion
  isCustomExpiry: boolean;
```

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const notify = useNotify();
+  const [currentStep, setCurrentStep] = useState<
+    "typeSelection" | "detailsForm"
+  >("typeSelection");
+
+  const { data: groups = [], error, isLoading } = useAuthGroups();
+  const { data: identities = [] } = useIdentities();
+
+  if (error) {
+    notify.failure("Loading panel details failed", error);
```
edlerd:
```suggestion
    notify.failure("Loading details failed", error);
```
Users probably don't think in terms such as `panel`.

src/pages/permissions/panels/CreateIdentityPanel.tsx
edlerd: Somehow this is changed again, we close the panel on step 2. Let's go back to step 1 instead of closing the panel right away.

edlerd (commented):
QA and Code mostly LGTM, one stricter type and one behavior change of the cancel button should be the last requests.

src/pages/permissions/panels/CreateIdentityPanel.tsx
kimanhou: I also changed the label of the button from `Cancel` to `Back`

edlerd (dismissed):
LGTM, thanks!

edlerd (approved):

vosdev:
I just wanted to report two bugs on 6.9 regarding bearer identities and then I decided to check dev first.

Not only are they fixed, but 👏  This is lovely ! Really easy to use and also very nice to see expiry information.

At first I used my dev lxd-ui environment against a 6.9 backend and that had some issues. The bearer identity would say "No active token"  when there was a token, the API just didn't expose the expiry information yet. I upgraded LXD to 6/edge and after I re-issued tokens it stated there was an active token with an expiry date.

That made me think about how this would work with a 5.21 LTS backend (or older), those older api versions do not expose the expiry information and then the UI would simply say "No active token" which is confusing I think?

I'm not 100% sure but I just wanted to inform you about this potential case.

For example the `bla` bearer identity in this overview has a token issued that works for the coming 10 years, it just doesnt have an expiry date logged in the database. And that is causing it to show "No active token"

image

image


Bearer/token expiry was only added recently because of my feature request: https://github.com/canonical/lxd/issues/18542 https://github.com/canonical/lxd/pull/18791

@kimanhou

omarelkashef:
> hose older api versions do not expose the expiry information and then the UI would simply say "No active token" which is confusing I think?

Hello @vosdev , I am glad you like the new experience.  And yes, these older lxd versions do not expose token expiry information. But this is only an "issue" for development since that is the only place where you can use a UI that is not designed for that specific backend. If you use the UI that is shipped with the lxd snap, it is guaranteed to work fine. I hope that answers your concern. Let me know if you have further questions.

vosdev:
> > hose older api versions do not expose the expiry information and then the UI would simply say "No active token" which is confusing I think?
> 
> Hello @vosdev , I am glad you like the new experience.  And yes, these older lxd versions do not expose token expiry information. But this is only an "issue" for development since that is the only place where you can use a UI that is not designed for that specific backend. If you use the UI that is shipped with the lxd snap, it is guaranteed to work fine. I hope that answers your concern. Let me know if you have further questions.

Aah that is good to know that you thought of this! 😁😁 

No further questions!

PR #2135, opened 2026-07-10
feat(contextual-help): [WD-36580] Create reusable docslink component

Kxiru:
## Done

- Introduced Explanation Tooltip + Styling
- Created InstanceExplanationTooltip
- Replaced the HelpLink in the InstanceList with InstanceExplanationTooltip

## QA

1. Run the LXD-UI:
    - Hover over the InstanceExplanationTooltip in the instance list page and ensure it behaves normally and responsively.
    - Double check the link it goes to, and the copy. Taken from the [contextual help copy doc](https://docs.google.com/document/d/1vDmemn4eQm0Zr3szBRzGtWdZqGNayqHkUmtC8oZ0jNA/edit?tab=t.0) that was previously created.

## Screenshots

image

### DIFF
--- a/src/components/ExplanationTooltip.tsx
+++ b/src/components/ExplanationTooltip.tsx
@@ -1,0 +1,37 @@
+import { type FC } from "react";
+import { Icon, Tooltip } from "@canonical/react-components";
+import DocLink from "components/DocLink";
+
+interface Props {
+  explanation: string;
+  docPath: string;
+  docLabel?: string;
+}
+
+const ExplanationTooltip: FC<Props> = ({
+  explanation,
+  docPath,
+  docLabel = "Learn more",
+}) => {
+  const tooltip = (
+    <Tooltip
+      zIndex={1000}
+      tooltipClassName="explanation-tooltip-portal"
+      position="btm-right"
+      message={
+        <span className="explanation-tooltip">
+          <span>{explanation}</span>
+          <DocLink docPath={docPath} hasExternalIcon>
+            {docLabel}
+          </DocLink>
+        </span>
+      }
+    >
+      <Icon name="information" className="explanation-tooltip-icon" />
+    </Tooltip>
+  );
+
+  return tooltip;
+};
+
+export default ExplanationTooltip;
--- a/src/pages/instances/InstanceExplanationTooltip.tsx
+++ b/src/pages/instances/InstanceExplanationTooltip.tsx
@@ -1,0 +1,18 @@
+import type { FC, ReactNode } from "react";
+import ExplanationTooltip from "components/ExplanationTooltip";
+
+const InstanceExplanationTooltip: FC<{ children?: ReactNode }> = ({
+  children,
+}) => {
+  return (
+    <span className="explanation-tooltip-wrapper">
+      {children}
+      <ExplanationTooltip
+        explanation="Manage and monitor virtual machines and containers running on the LXD server."
+        docPath="/explanation/instances/"
+      />
+    </span>
+  );
+};
+
+export default InstanceExplanationTooltip;
--- a/src/sass/_explanation_tooltip.scss
+++ b/src/sass/_explanation_tooltip.scss
@@ -1,0 +1,23 @@
+.explanation-tooltip-wrapper {
+  align-items: center;
+  display: flex;
+  gap: $spv--small;
+}
+
+.explanation-tooltip-icon {
+  height: 1rem !important;
+  width: 1rem !important;
+}
+
+.explanation-tooltip {
+  display: flex;
+  flex-direction: column;
+  gap: $spv--small;
+}
+
+.explanation-tooltip-portal .p-tooltip__message {
+  margin: 0 1rem 0 0;
+  max-width: calc(60vw - 2rem);
+  white-space: normal;
+  width: max-content;
+}
--- a/src/pages/instances/InstanceList.tsx
+++ b/src/pages/instances/InstanceList.tsx
@@ -60,7 +60,7 @@ import {
 } from "util/operations";
 import NotificationRow from "components/NotificationRow";
 import SelectedTableNotification from "components/SelectedTableNotification";
-import HelpLink from "components/HelpLink";
+import InstanceExplanationTooltip from "pages/instances/InstanceExplanationTooltip";
 import type { LxdInstanceStatus } from "types/instance";
 import useSortTableData from "util/useSortTableData";
 import PageHeader from "components/PageHeader";
@@ -661,12 +661,9 @@ const InstanceList: FC = () => {
           <PageHeader>
             <PageHeader.Left>
               <PageHeader.Title>
-                <HelpLink
-                  docPath="/explanation/instances/#expl-instances"
-                  title="Learn more about instances"
-                >
+                <InstanceExplanationTooltip>
                   Instances
-                </HelpLink>
+                </InstanceExplanationTooltip>
               </PageHeader.Title>
               {hasInstances && selectedNames.length === 0 && (
                 <PageHeader.Search>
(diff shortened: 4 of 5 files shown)


sophiecarroll (dismissed):

src/pages/instances/InstanceExplanationTooltip.tsx
```diff
@@ -0,0 +1,18 @@
+import type { FC, ReactNode } from "react";
+import ExplanationTooltip from "components/ExplanationTooltip";
+
+const InstanceExplanationTooltip: FC<{ children?: ReactNode }> = ({
+  children,
```
omarelkashef, line 5: Do we need the children props? Since this component is only for instances we can have the "Instances" text here and get rid of the props.

src/components/ExplanationTooltip.tsx
```diff
@@ -0,0 +1,39 @@
...
+import DocLink from "components/DocLink";
+
+interface Props {
+  explanation: string;
+  docPath: string;
+  docLabel?: string;
+  position?: TooltipProps["position"];
+}
+
+const ExplanationTooltip: FC<Props> = ({
```
omarelkashef, line 11: Can the tooltip adjust for smaller screens?

Image

omarelkashef (commented):
Very nice work! One question and two comments. Are the rest of the pages planned for another PR? Should we just implement them here?

src/sass/_explanation_tooltip.scss
```diff
@@ -0,0 +1,16 @@
+.explanation-tooltip-wrapper {
+  align-items: center;
+  display: inline-flex;
+}
+
+.explanation-tooltip-icon {
+  height: 1rem !important;
+  margin-left: $spv--small;
```
edlerd:
```suggestion
```
This is causing the tooltip arrow to be wrongly positioned

Image

src/sass/_explanation_tooltip.scss
```diff
@@ -0,0 +1,16 @@
+.explanation-tooltip-wrapper {
+  align-items: center;
+  display: inline-flex;
+}
```
edlerd, line 5:
```suggestion
```
I don't think we need an extra wrapper, can this be a fragment with no class instead?

Kxiru:
> Are the rest of the pages planned for another PR? Should we just implement them here?
Yes, the other pages will all be in another PR. This PR will just address the infrastructure of the component and then it will be applied to the other entitites.

src/pages/instances/InstanceExplanationTooltip.tsx
Kxiru: True, but for code transparency from the InstanceList page, I want to make it clear that the "Instances" title is not an explicit part of this component, otherwise it couples the InstanceList header/title with the ExplanationTooltip, which theoretically is not quite what we want to achieve. For most usages, we won't need a child prop, but for headers, we can pass through the title of the page.

src/components/ExplanationTooltip.tsx
edlerd: This is still broken

Image

src/sass/_explanation_tooltip.scss
```diff
@@ -0,0 +1,16 @@
...
+.explanation-tooltip-icon {
+  height: 1rem !important;
+  width: 1rem !important;
+}
+
+.explanation-tooltip {
+  display: flex;
+  flex-direction: column;
+  gap: $spv--small;
+}
```
edlerd:
```suggestion
}

```
Code style

src/sass/_explanation_tooltip.scss
edlerd: The wrapper is still present?

edlerd (changes requested):

src/sass/_explanation_tooltip.scss
Kxiru: It has been repurposed as it now uses flex for the gap.

src/components/ExplanationTooltip.tsx
Kxiru: I wasn't finished, @edlerd :joy: . I'm having difficulty with this, I think due to the vanilla styling and the tooltip being in a portal? I will re-request reviews when I am done.

sophiecarroll:
Hi there, looks great, one question–do you think it should go to the top of the "containers and VMs" context page it links to? Right now the anchor link is at the bottom of the doc, and I'm worried that when people are routed to the bottom they'll think the info at the top of the page isn't relevant to them (it is because it's talking about containers and VMs)

Kxiru:
> Hi there, looks great, one question–do you think it should go to the top of the "containers and VMs" context page it links to?

Heya! Sorry what is "it"? You mentioned an anchor link but I have not used a href/anchor tag in this PR. Are you suggesting a drive by fix for something else or am I confused? @sophiecarroll

sophiecarroll:
> > Hi there, looks great, one question–do you think it should go to the top of the "containers and VMs" context page it links to?
> 
> Heya! Sorry what is "it"? You mentioned an anchor link but I have not used a href/anchor tag in this PR. Are you suggesting a drive by fix for something else or am I confused? @sophiecarroll

"it" means the tooltip, and when I click the link it doesn't just take me to a page but scrolls all the way down to the bottom. I'll attach a video

https://github.com/user-attachments/assets/ce494062-3eec-4bbf-af94-069dfdc3788a

Kxiru:
> "it" means the tooltip, and when I click the link it doesn't just take me to a page but scrolls all the way down to the bottom.

Ohh! Yes, it scrolls to the bottom because it is redirecting to the most relevant part of the doc, to the entity. But I'm not too fussed about this. I have changed this to redirect to the Instances page in general, as seems to be the structure across other general entity docspaths.

src/components/ExplanationTooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+  explanation,
+  docPath,
+  docLabel = "Learn more",
+  position = "right",
```
edlerd:
```suggestion
```
We probably don't need to expose the position.

src/components/ExplanationTooltip.tsx
edlerd: The issue is solved on some screens, but on narrow screens it still overflows outside the screen.

Also on desktop it now wraps too early, where it previously fit into a single line.

edlerd (commented):
Mostly LGTM, thanks for the iteration. Some concerns on responsiveness are left open.

Kxiru:
> The issue is solved on some screens, but on narrow screens it still overflows outside the screen. Also on desktop it now wraps too early, where it previously fit into a single line.

I worked on this for a few hours but could not find a solution due to the limitations of the fact that it is in a tooltip portal itself. I'll try again but please let me know if you have any suggestions. I might have to override some of the default/standard styling from ReactComponents.

Kxiru:
I was able to make a worthwhile iteration with the help of Copilot and javascript :melting_face:

src/components/ExplanationTooltip.tsx
```diff
@@ -0,0 +1,71 @@
...
+      }
+    });
+
+    observer.observe(document.body, { childList: true, subtree: true });
+
+    return () => {
+      observer.disconnect();
+    };
+  }, []);
+
```
edlerd:
```suggestion
```
This is far too complex for solving a layout issue.

src/sass/_explanation_tooltip.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.explanation-tooltip-portal .p-tooltip__message {
+  margin: 0 1rem 0 0;
+  max-width: calc(100vw - 2rem);
```
edlerd:
```suggestion
  max-width: calc(60vw - 2rem);
```
Instead, we can constrain the width of the tooltip with this css change.

edlerd (changes requested):

edlerd (approved):
LGTM, thanks for adding the reusable component!

PR #2137, opened 2026-07-10
feat(projects): Create projects list page

kimanhou:
## Done

- Create projects list page
- Update NavigationProjectSelector: 
    - change existing "All projects" link into "All instances"
    - add "All projects" link to project list page
    - add `hr` and make them visible
- Add "Projects" link on instance list page for all projects

## QA

1. Run the LXD-UI:
    - Go to `/ui/projects`
    - Make sure "All projects" is selected in the Navigation project selector.
    - Make sure the full list of projects is displayed.
    - Make sure you can create a project from this page.
    - Make sure you can search for a project.
    - Make sure the documentation link in the header is correct.
    - Click on a project name, it should take you to the project's configuration.
    - Click on a project instance count, it should take you to the instance list of this project.
    - Click on "Instances", it should take you to the instance list of all projects. You should be able to go back to the project list page by clicking "Projects".

## Screenshots

image

Screenshot from 2026-07-15 13-21-23

### DIFF
--- a/src/pages/projects/ProjectList.tsx
+++ b/src/pages/projects/ProjectList.tsx
@@ -1,0 +1,229 @@
+import { useState, type FC } from "react";
+import { Link } from "react-router-dom";
+import {
+  Col,
+  CustomLayout,
+  Icon,
+  MainTable,
+  Row,
+  ScrollableTable,
+  SearchBox,
+  Spinner,
+  TablePagination,
+  useNotify,
+} from "@canonical/react-components";
+import classnames from "classnames";
+import NotificationRow from "components/NotificationRow";
+import HelpLink from "components/HelpLink";
+import PageHeader from "components/PageHeader";
+import { useIsScreenBelow } from "context/useIsScreenBelow";
+import { useProjects } from "context/useProjects";
+import { useServerEntitlements } from "util/entitlements/server";
+import { defaultFirst } from "util/helpers";
+import { getInstancesUsedByProject } from "util/projects";
+import { ROOT_PATH } from "util/rootPath";
+import useSortTableData from "util/useSortTableData";
+
+const ProjectList: FC = () => {
+  const notify = useNotify();
+  const [query, setQuery] = useState<string>("");
+  const isSmallScreen = useIsScreenBelow();
+  const { canCreateProjects } = useServerEntitlements();
+  const { data: projects = [], error, isLoading } = useProjects();
+
+  if (error) {
+    notify.failure("Loading projects failed", error);
+  }
+
+  const filteredProjects = [...projects]
+    .sort(defaultFirst)
+    .filter((project) => {
+      if (!query) {
+        return true;
+      }
+      const q = query.toLowerCase();
+      return (
+        project.name.toLowerCase().includes(q) ||
+        project.description.toLowerCase().includes(q)
+      );
+    });
+
+  const headers = [
+    { content: "Name", sortKey: "name" },
+    {
+      content: "Description",
+      sortKey: "description",
+      className: "u-hide--small",
+    },
+    { content: "CPU limit", sortKey: "cpuLimit", className: "u-hide--small" },
+    {
+      content: "Memory limit",
+      sortKey: "memoryLimit",
+      className: "u-hide--small",
+    },
+    { content: "Disk limit", sortKey: "diskLimit", className: "u-hide--small" },
+    { content: "Instances", sortKey: "instances" },
+  ];
+
+  const rows = filteredProjects.map((project) => {
+    const instanceCount = getInstancesUsedByProject(project).length;
+    const projectInstancesPath = `${ROOT_PATH}/ui/project/${encodeURIComponent(project.name)}/instances`;
+    const projectConfigPath = `${ROOT_PATH}/ui/project/${encodeURIComponent(project.name)}/configuration`;
+    const cpuLimit = project.config["limits.cpu"] || "-";
+    const memoryLimit = project.config["limits.memory"] || "-";
+    const diskLimit = project.config["limits.disk"] || "-";
+
+    return {
+      key: project.name,
+      className: "u-row",
+      columns: [
+        {
+          content: (
+            <Link
+              to={projectConfigPath}
+              className="u-truncate"
+              title={project.name}
+            >
+              {project.name}
+            </Link>
+          ),
+          role: "rowheader",
+          "aria-label": "Name",
+        },
+        {
+          content: (
+            <div className="table-description" title={project.description}>
+              {project.description || "-"}
+            </div>
+          ),
+          role: "cell",
+          "aria-label": "Description",
+          className: "u-hide--small",
+        },
+        {
+          content: cpuLimit,
+          role: "cell",
+          "aria-label": "CPU limit",
+          className: "u-hide--small",
+        },
+        {
+          content: memoryLimit,
+          role: "cell",
+          "aria-label": "Memory limit",
+          className: "u-hide--small",
+        },
+        {
+          content: diskLimit,
+          role: "cell",
+          "aria-label": "Disk limit",
+          className: "u-hide--small",
+        },
+        {
+          content: (
+            <Link
+              to={projectInstancesPath}
+              className="u-truncate"
+              title={project.name}
+            >
+              {instanceCount}
+            </Link>
+          ),
+          role: "cell",
+          "aria-label": "Instances",
+        },
+      ],
+      sortData: {
+        name: project.name.toLowerCase(),
+        description: project.description.toLowerCase(),
+        instances: instanceCount,
+        cpuLimit,
+        memoryLimit,
+        diskLimit,
+      },
+    };
+  });
+
+  const { rows: sortedRows, updateSort } = useSortTableData({ rows });
+
+  if (isLoading) {
+    return <Spinner className="u-loader" text="Loading..." isMainComponent />;
+  }
+
+  return (
+    <CustomLayout
+      header={
+        <PageHeader>
+          <PageHeader.Left>
+            <PageHeader.Title>
+              <HelpLink docPath="/projects/" title="Learn more about projects">
+                Projects
+              </HelpLink>
+            </PageHeader.Title>
+            <PageHeader.Search>
+              <SearchBox
+                className="search-box margin-right--large u-no-margin--bottom"
+                id="search-project"
+                name="search-project"
+                type="text"
+                onChange={(value: string) => {
+                  setQuery(value);
+                }}
+                placeholder="Search"
+                value={query}
+                aria-label="Search"
+              />
+            </PageHeader.Search>
+          </PageHeader.Left>
+          <PageHeader.BaseActions>
+            <Link
+              className={classnames("p-button--positive u-no-margin--bottom", {
+                "is-disabled": !canCreateProjects(),
+                "has-icon": !isSmallScreen,
+              })}
+              to={`${ROOT_PATH}/ui/projects/create`}
+              title={
+                canCreateProjects()
+                  ? ""
+                  : "You do not have permission to create projects"
+              }
+              key="create-project"
+            >
+              <Icon name="plus" light className="u-hide--small" />
+              <span>Create project</span>
+            </Link>
+          </PageHeader.BaseActions>
+        </PageHeader>
+      }
+    >
+      <NotificationRow />
+      <Row>
+        <Col size={12}>
+          <ScrollableTable
+            dependencies={[filteredProjects, notify.notification]}
+            tableId="project-table"
+            belowIds={["status-bar"]}
+          >
+            <TablePagination
+              id="pagination"
+              data={sortedRows}
+              itemName="project"
+              className="u-no-margin--top"
+              aria-label="Table pagination control"
+            >
+              <MainTable
+                id="project-table"
+                className="project-table"
+                headers={headers}
+                sortable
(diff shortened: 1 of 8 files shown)


sophiecarroll:
This is going to be so useful, glad we're making it! No edits from design so far

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,246 @@
...
+              title={
+                canCreateProjects()
+                  ? ""
+                  : "You do not have permission to create projects"
+              }
+              key="create-project"
+            >
+              <Icon name="plus" light className="u-hide--small" />
+              <span>Create project</span>
+            </Link>
```
edlerd, line 193: Something odd happens, when I use this, the whole page refreshes. When I open create project through the side navigation, only the content page refreshes.

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,246 @@
...
+        },
+        {
+          content: (
+            <Link
+              href={projectInstancesPath}
+              className="u-truncate"
+              title={project.name}
+            >
+              {instanceCount}
+            </Link>
```
edlerd: Should we link only the project title to the instance view of that project and not have a link on this column? Or move this column to the end, or remove the cpu/memory/disk columns? Because we try to keep interactive parts of tables to the first or last columns.

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,246 @@
...
+          {projects.length === 0 && (
+            <EmptyState
+              className="empty-state"
+              image={<Icon name="repository" className="empty-state-icon" />}
+              title="No projects found"
+            >
+              <p>There are no projects yet.</p>
+            </EmptyState>
+          )}
+          {projects.length > 0 && (
```
edlerd:
```suggestion
```
There is always the default project, so we don't need an empty state.

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,246 @@
...
+          </PageHeader.Left>
+          <PageHeader.BaseActions>
+            <Link
+              className={classnames("p-button", {
+                "u-no-margin--bottom": !isSmallScreen,
+              })}
+              href={`${ROOT_PATH}/ui/all-projects/instances`}
+              key="see-instances"
+            >
+              <span>See instances for all projects</span>
```
edlerd: The caption is a bit long for a button, can we find a shorter one?

And I think we should also link to the project list from the all instances page.

edlerd (commented):
Great stuff, few thoughts on the new page.

src/pages/projects/ProjectList.tsx
kimanhou: I think it's because `NavigationProjectSelector` uses a `Button` with `navigate` in the `onClick` while I use `Links`. Semantically it makes more sense to use `Link` but in terms of UX it's better not to reload the whole page.

edlerd: Yes, the link is the better option, can you use it in the navigation as well? 

The problem is the `import { Link } from "react-router-dom";` instead of the one from react components in this component.

src/pages/instances/InstanceList.tsx
```diff
@@ -700,6 +701,17 @@ const InstanceList: FC = () => {
...
               <PageHeader.BaseActions>
+                {isAllProjects && (
+                  <Link
+                    className={classnames("p-button", {
+                      "u-no-margin--bottom": !isSmallScreen,
+                    })}
+                    href={`${ROOT_PATH}/ui/projects`}
+                    key="view-all-projects"
+                  >
+                    View all projects
```
```suggestion
                    Projects
```
A bit better? Maybe add the project icon as well.

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,232 @@
...
+          </PageHeader.Left>
+          <PageHeader.BaseActions>
+            <Link
+              className={classnames("p-button", {
+                "u-no-margin--bottom": !isSmallScreen,
+              })}
+              href={`${ROOT_PATH}/ui/all-projects/instances`}
+              key="view-all-instances"
+            >
+              View all instances
```
edlerd:
```suggestion
              Instances
```
Maybe this can be as short as this? Also, shall we add the instance icon?

src/pages/projects/ProjectList.tsx
edlerd: Can you also fix the import in `InstancesOverviewStatus` for the link component?

src/pages/instances/InstanceList.tsx
```diff
@@ -5,6 +5,7 @@ import {
   ColumnSelector,
   CustomLayout,
   Icon,
+  Link,
```
edlerd: Should also be imported from react router instead.

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,232 @@
...
+          ),
+          role: "cell",
+          "aria-label": "Instances",
+        },
+      ],
+      sortData: {
+        name: project.name.toLowerCase(),
+        description: project.description.toLowerCase(),
+        instances: instanceCount,
+      },
```
edlerd, line 142: We are missing sort keys for the memory, disk and cpu limits.

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,232 @@
...
+      header={
+        <PageHeader>
+          <PageHeader.Left>
+            <PageHeader.Title>
+              <HelpLink docPath="/projects/" title="Learn more about projects">
+                Projects
+              </HelpLink>
+            </PageHeader.Title>
+            <PageHeader.Search>
+              <SearchBox
```
edlerd, line 163: This component is lacking an `id` prop, which I think is mandatory for accessibility.

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,232 @@
...
+            >
+              <Icon name="plus" light className="u-hide--small" />
+              <span>Create project</span>
+            </Link>
+          </PageHeader.BaseActions>
+        </PageHeader>
+      }
+    >
+      <NotificationRow />
+      <Row className="no-grid-gap">
```
edlerd:
```suggestion
      <Row>
```
Seems not necessary.

edlerd (commented):
Good stuff, few small things left, then this should be good to merge.

kimanhou:
Should we change the `NavigationProjectSelector` ? Currently, clicking on "All projects" takes you to the instance list for all projects. This page is also accessible by the link "Instances" from the project list.
Shouldn't "All projects" link to the same page as "Projects" from the instance list ? Or should we rename "All projects" into "All instances" in the `NavigationProjectSelector` ?

sophiecarroll:
> Should we change the `NavigationProjectSelector` ? Currently, clicking on "All projects" takes you to the instance list for all projects. This page is also accessible by the link "Instances" from the project list. Shouldn't "All projects" link to the same page as "Projects" from the instance list ? Or should we rename "All projects" into "All instances" in the `NavigationProjectSelector` ?
Screenshot 2026-07-14 at 3 27 30 PM

Are you saying to link "all projects" on projects dropdown to the project list page in this PR? If so, I definitely think yes.

edlerd:
> Are you saying to link "all projects" on projects dropdown to the project list page in this PR? If so, I definitely think yes.

I agree, we should have a way to get to the project list from the navigation, and "All projects" sounds like a link to the project list page, which would be confusing. 

Some additional concerns 
- Keeping the current caption and changing the link would confuse existing users that expect the instance list from that link.
- Removing the link to all instances from the nav makes the global instance list harder to reach. Currently it is one click, then it would be two clicks: Going to all projects, the to all instances.

So I think the best way is to change the project dropdown to:
- Keep the link to all instance, but change the caption to something clearly stating the target page.
- Add a link to the project list

Maybe as simple as changing caption from "All projects" to "All instances" and in the footer right next to the "Create project" we can add a new "Projects" link to the list. Wdyt @sophiecarroll @kimanhou ?

sophiecarroll:
> > Are you saying to link "all projects" on projects dropdown to the project list page in this PR? If so, I definitely think yes.
> 
> I agree, we should have a way to get to the project list from the navigation, and "All projects" sounds like a link to the project list page, which would be confusing.
> 
> Some additional concerns
> 
> * Keeping the current caption and changing the link would confuse existing users that expect the instance list from that link.
> * Removing the link to all instances from the nav makes the global instance list harder to reach. Currently it is one click, then it would be two clicks: Going to all projects, the to all instances.
> 
> So I think the best way is to change the project dropdown to:
> 
> * Keep the link to all instance, but change the caption to something clearly stating the target page.
> * Add a link to the project list
> 
> Maybe as simple as changing caption from "All projects" to "All instances" and in the footer right next to the "Create project" we can add a new "Projects" link to the list. Wdyt @sophiecarroll @kimanhou ?

Agreed to all. 

My only thing is that I don't know if we even need to get to all instances from the project dropdown: we can still navigate to all instances by clicking the instances nav item while on all projects. It seemed unintuitive to me to get to instances through projects, which would now become the more intuitive version of getting to all projects through projects. I guess we need to know if a lot of people rely on the projects --> all instances method over the instances nav item–I can get some user testing validation on that.

kimanhou:
It's weird in the footer

image

I prefer them both at the top
image


By the way, is it normal that `hr` is never visible on dark background ?

edlerd:
> It's weird in the footer

I thought they can be side by side, create and list projects so the footer is a single line "Create project | Projects" with two buttons next to each other. Does it work better?

> By the way, is it normal that `hr` is never visible on dark background ?

This might be a bug, not sure, though we might apply custom styling.

sophiecarroll:
> > It's weird in the footer
> 
> I thought they can be side by side, create and list projects so the footer is a single line "Create project | Projects" with two buttons next to each other. Does it work better?
> 
> > By the way, is it normal that `hr` is never visible on dark background ?
> 
> This might be a bug, not sure, though we might apply custom styling.

I think I like the second way Kim Anh did it with all instances and all projects at the top of the dropdown and create project at the bottom. I think it makes sense to put the two "all"s together. The dropdown still doesn't feel super intuitive though–I put this on UX list for future research.

kimanhou:
"Create project" and "Projects" dont' fit side by side with the current width
Screenshot from 2026-07-15 13-16-15

We would have to increase the width:
Screenshot from 2026-07-15 13-17-23

I prefer the previous version with "All instances" and "All projects" at the top.

kimanhou:
There is custom CSS to make hr not visible in _project_select.scss. 
```
    hr {
      background-color: $colors--theme--background-default;
      margin-top: $spv--large;
    }
```

If I remove the override of the background color:

image

edlerd:
> If I remove the override of the background color:

I think this is best, with both options on top and the `hr`s visible.

src/pages/instances/InstanceList.tsx
```diff
@@ -700,6 +700,18 @@ const InstanceList: FC = () => {
...
+                    className={classnames("p-button", "has-icon", {
+                      "u-no-margin--bottom": !isSmallScreen,
+                    })}
+                    to={`${ROOT_PATH}/ui/projects`}
+                    key="view-all-projects"
+                  >
+                    <Icon name="folder" />
+                    <span>Projects</span>
+                  </Link>
+                )}
```
```suggestion
```
Please excuse the back and forth, but now I think we shouldn't have any button linking to projects here. Because it seems confusing that sometimes (on the all instances page) we do have the "projects" button and at other times (when in a specific project) we do not have the "projects" button.

Also, with the recent evolution of the project side panel, there is a more natural way to reach the project list page.

So maybe removing the link here, and also the link to "instances" from the project list. Wdyt?

src/pages/projects/ProjectList.tsx
```diff
@@ -0,0 +1,236 @@
...
+            <div className="table-description" title={project.description}>
+              {project.description || "-"}
+            </div>
+          ),
+          role: "cell",
+          "aria-label": "Description",
+          className: "u-hide--small",
+        },
+        {
+          content: project.config["limits.cpu"] || "-",
```
edlerd: Let's introduce a const for the cpu and other limits, to use for display here and for the sortkey below.

src/sass/_project_select.scss
```diff
@@ -71,7 +72,6 @@
     }
 
     hr {
-      background-color: $colors--theme--background-default;
```
edlerd, line 74: Again, sorry for the confusion. Seeing it now in action, it seems there are just too many lines. Maybe there was a reason for this override, and we should revert there. Wdyt?

edlerd (commented):
Thanks for the iterations! Mostly LGTM, tiny nitpicks, and questions on the design and code, then this should be good to merge.

edlerd (approved):
LGTM, thanks :)

sophiecarroll:
Interesting thought for the future: are people going to expect this to match the CLI "lxc project list" output more?

This is what it looks like: alternatively, should we add more info to the CLI list if we think it's useful on the UI like CPU limit, etc.?

Screenshot 2026-07-24 at 12 43 42 PM

kimanhou:
@sophiecarroll 

> are people going to expect this to match the CLI "lxc project list" output more

I don't know what people would expect or not to see on the dashboard. Please log your thoughts in the relevant locations (spec, Mattermost, JIRA) instead of a merged PR.

> should we add more info to the CLI list

You can talk to BE team.

sophiecarroll:
After meeting: add default network and storage pool

PR #2138, opened 2026-07-13
ci: use tiobe tics action

omarelkashef:
## Done

- update tics report workflow to upload test results and use tiobe action WD-37567

## QA

1. Run the LXD-UI:

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -109,6 +109,14 @@ jobs:
           DOTRUN_CONTAINER_ID=$(docker ps | awk '{print $1}' | tail -n1)
           docker stop $DOTRUN_CONTAINER_ID
 
+      - name: Upload Playwright test artifacts
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: playwright-test-artifacts
+          path: test-results
+          retention-days: 7
+
       - name: Upload coverage report
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yaml
```diff
@@ -118,62 +126,66 @@ jobs:
...
+            git push
+            if [ $? -eq 0 ]; then
+              echo "Successfully pushed HTML report to repo."
+              exit 0
+            fi
+          done
+      - name: Output Report URL as Worfklow Annotation
+        run: |
+          FULL_HTML_REPORT_URL=https://canonical.github.io/lxd-ui/coverage
+          echo "::notice title=Published Playwright Test Report::$FULL_HTML_REPORT_URL"
```
edlerd: Why change the publish coverage step? I think this can stay as it was.

.github/workflows/coverage.yaml
```diff
@@ -118,62 +126,66 @@ jobs:
...
 
   tics-report:
-    runs-on: [self-hosted, linux, amd64, tiobe, jammy]
+    runs-on: [self-hosted, linux, amd64, resolute]
     needs: publish-coverage-report
+    # Self-hosted runners share machine-level resources. Serialize TICS per branch
+    # to avoid qserver collisions for the same project database.
+    concurrency:
+      group: tics-report-${{ github.repository }}-${{ github.ref_name }}
```
edlerd:
```suggestion
    concurrency:
      group: tics-report-${{ github.repository }}
```
We only run TICS on main, and never on other branches.

.github/workflows/coverage.yaml
```diff
@@ -191,12 +203,13 @@ jobs:
...
-          # need to source .profile, or TICSQServer command won't be found
-          source ~/.profile
-          # Update path so eslint is used through yarn dependencies
-          export PATH="$PWD/node_modules/.bin:$PATH"
-          TICSQServer -project lxd-ui -tmpdir /tmp/tics -branchdir .
+        uses: tiobe/tics-github-action@a53037a6bb9f71b6a97cdeb0ac41632266804b9e # v3.8.0
+        with:
+          mode: qserver
+          project: lxd-ui
+          viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
```
edlerd, line 210:
```suggestion
          viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/TqiDashboard.html#axes=CodeType(production),ProjectGroup(PublicRepos),Project(lxd-ui)
```
I think this should be different?

edlerd (commented):
Few thoughts on the changes.

.github/workflows/coverage.yaml
omarelkashef: Only the indentation and a typo were fixed in this step.

.github/workflows/coverage.yaml
omarelkashef: Yes, but we might push to main before a previous run is finished and this can cause a TICS error "Project database already is connected to a running TICSQServer process"

.github/workflows/coverage.yaml
omarelkashef: The viewUrl should point to the configuration endpoint https://github.com/tiobe/tics-github-action#basic-parameters as in https://github.com/canonical/microcloud-cluster-manager/blob/main/.github/workflows/tics.yaml#L62-L69

.github/workflows/coverage.yaml
edlerd: Yes, I think a concurrency group based on the repository handles that. We should not add the `ref_name` / branch name to it.

.github/workflows/coverage.yaml
edlerd: Ah right, then all fine as is.

.github/workflows/coverage.yaml
omarelkashef: Oh I see what you mean. I'll fix it.

.github/workflows/coverage.yaml
```diff
@@ -118,62 +126,66 @@ jobs:
...
+      - name: Output Report URL as Workflow Annotation
+        run: |
+          FULL_HTML_REPORT_URL=https://canonical.github.io/lxd-ui/coverage
+          echo "::notice title=Published Coverage Report::$FULL_HTML_REPORT_URL"
 
   tics-report:
-    runs-on: [self-hosted, linux, amd64, tiobe, jammy]
+    runs-on: [self-hosted, linux, amd64, resolute]
     needs: publish-coverage-report
+    # Self-hosted runners share machine-level resources. Serialize TICS per branch
```
edlerd:
```suggestion
    # Self-hosted runners share machine-level resources. Serialize TICS
```
We removed the branch specifier from the concurrency group :)

edlerd (dismissed):
LGTM, tiny non-blocking nitpick on the comment.

.github/workflows/coverage.yaml
omarelkashef: Oh yes, good catch!

edlerd (approved):

PR #2140, opened 2026-07-14
feat(a11y-accessibility-audit) - [WD-36582] Create A11y accessibility check

Kxiru:
## Done

- [x] Added accessibility-audit.spec.ts test
- [x] Added audit tests for main entity pages and subpages
- [x] Refactored the Pr.yaml workflow to reduce repetition
- [x] Add upload artefact capability to CI
- [x] Aligns tests on features that are/n't in older LXD versions. (Eg; Cluster links are not supported in LXD 5.21)

Out of scope:
- Adding accessibility tests for side-panels
- ~Adding accessibility tests to CI to be automated (next PR)~
- Addressing any of the accessibility notices from "failed" tests.

## QA

1. Run the LXD-UI:
    - Within a local environment, run any of the tests on accessibility-audit.spec.ts and observe the output.

## Screenshots

Expected output for instances page:
image

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -87,88 +87,28 @@ jobs:
   a11y-check:
     name: a11y-check
     runs-on: ubuntu-latest
-    env:
-      LXD_CHANNEL: "latest/edge"
-      DEPLOYMENT: "unclustered"
-      LXD_OIDC_CLIENT_ID: "RYDnMpkygLAMfeo17lU7LYwWGxisRuRR"
-      LXD_OIDC_CLIENT_SECRET: "CNKX4UmrZKZJq5rJy5VM_JfcNPqkws1rwWWQk_q0oyZ8gABARr19ic7xrhPssGA1"
-      LXD_OIDC_ISSUER: "https://dev-h6c02msuggpi6ijh.eu.auth0.com/"
-      LXD_OIDC_AUDIENCE: "https://dev-h6c02msuggpi6ijh.eu.auth0.com/api/v2/"
-      LXD_OIDC_USER: "lxd-ui-e2e-tests@example.org"
-      LXD_OIDC_PASSWORD: "lxd-ui-e2e-password"
-      LXD_OIDC_GROUPS_CLAIM: "lxd-idp-groups"
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: false
 
-      - name: Generate SSL keys for HAProxy
-        run: |
-          mkdir -p keys
-          openssl req -x509 -newkey rsa:4096 -keyout keys/lxd-ui.key -out keys/lxd-ui.crt -days 365 -nodes -subj "/CN=localhost"
-          cat keys/lxd-ui.key keys/lxd-ui.crt > keys/lxd-ui.pem
-
-      - name: Install LXD
-        uses: canonical/setup-lxd@8c6a87bfb56aa48f3fb9b830baa18562d8bfd4ee # v1
+      - name: Setup LXD
+        uses: ./.github/actions/setup-lxd
         with:
-          bridges: "lxdbr0,workshopbr0"
-          channel: "latest/stable"
-
-      - name: Install Workshop
-        run: |
-          set -x
-          sudo lxd init --auto
-          sudo snap install workshop --classic
-          workshop launch --verbose
-
-      - name: Setup LXD backend (Unclustered in VM)
-        shell: bash
-        run: |
-          set -x
-          tests/scripts/start_vm_single_unclustered.sh ${{ env.LXD_CHANNEL }}
-
-
-      - name: Install LXD-UI dependencies
-        run: |
-          sudo chmod 0777 ../lxd-ui
-          workshop run dev install
-
-      - name: Setup for tests
-        shell: bash
-        env:
-          DEPLOYMENT: ${{ env.DEPLOYMENT }}
-          LXD_OIDC_ISSUER: ${{ env.LXD_OIDC_ISSUER }}
-          LXD_OIDC_CLIENT_ID: ${{ env.LXD_OIDC_CLIENT_ID }}
-          LXD_OIDC_CLIENT_SECRET: ${{ env.LXD_OIDC_CLIENT_SECRET }}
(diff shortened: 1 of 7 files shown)


.github/workflows/pr.yaml
```diff
@@ -84,6 +84,17 @@ jobs:
             echo "No disallowed Unicode characters found."
           fi
 
+  a11y-check:
+    name: a11y-check
+    runs-on: ubuntu-latest
+    steps:
+      # - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
omarelkashef: This step is necessary I believe.

tests/helpers/a11y.ts
```diff
@@ -0,0 +1,155 @@
...
+  await page.getByRole("link", { name: slug, exact: true }).first().click();
+};
+
+export const runA11yAudit = async (
+  slug: string,
+  page: Page,
+  testInfo: TestInfo,
+  parentSlug?: string,
+) => {
+  await clickSideNavItem(page, slug, parentSlug);
```
omarelkashef, line 71: Does our regular `goTo`  pattern not work here? This function could receive the page or url we want to audit. wdyt?

tests/helpers/ally.ts
```diff
@@ -0,0 +1,155 @@
...
+  parentSlug?: string,
+) => {
+  await clickSideNavItem(page, slug, parentSlug);
+
+  const results = await new AxeBuilder({ page })
+    .withTags(["wcag2a", "wcag2aa", "wcag21a", "wcag21aa", "best-practice"])
+    .analyze();
+
+  const passed = results.passes?.length ?? 0;
+  const failed = results.violations?.length ?? 0;
```
omarelkashef: I think violations is a better term here and in the report like the Axe framework uses.

tests/helpers/ally.ts
```diff
@@ -0,0 +1,155 @@
...
+      id: instance.id,
+      slug,
+      type,
+      tags: instance.tags,
+      description: instance.description,
+      nodes: [],
+    };
+    for (const node of instance.nodes) {
+      if (Array.isArray(node.target))
+        for (const target of node.target) resultObject["nodes"].push(target);
```
omarelkashef: I think the loop body is more readable inside curly braces.

tests/helpers/a11y.ts
```diff
@@ -0,0 +1,155 @@
...
+  const incomplete = results.incomplete?.length ?? 0;
+  const total = passed + failed + incomplete;
+  const percent = total === 0 ? 100 : (passed / total) * 100;
+
+  const severities = countSeverities(results);
+  const testName = testInfo.title;
+
+  printSummary(percent, passed, failed, incomplete, slug, testName, severities);
+  if (results.violations.length > 0) {
+    const result = generateResult(results, "violation", slug);
```
omarelkashef: should `generateResult` be combined with `printSummary` into `generateReport`?

omarelkashef (commented):
This is really cool! A few comments below.

tests/helpers/a11y.ts
Kxiru: Hmm. Technically it could work here, but I think it would be extra lines, such as needed to construct the URL. I think this approach is also fine, would there be any added benefit to sitching out to goToURL?

tests/helpers/a11y.ts
Kxiru: I think it's good to separate concerns and decouple functions.

omarelkashef: Yeah, I agree. They should stay as separate functions, but their calls can both be in a new function called generateReport

tests/helpers/a11y.ts
omarelkashef: The only advantage is that we use an already existing pattern.

Kxiru: Hmm... I mean, the clickSideNavItem function itself uses goToURL, so the pattern _has_ been reapplied. @edlerd , @kimanhou , any comments from yourselves, here?

tests/helpers/a11y.ts
Kxiru: I think your equivalent of generateReport is `runA11yAudit`, which exists already. Do you think it should be renamed?

.github/workflows/pr.yaml
```diff
@@ -84,6 +84,43 @@ jobs:
...
+        with:
+          persist-credentials: false
+
+      - name: Setup LXD
+        uses: ./.github/actions/setup-lxd
+        with:
+          lxd_channel: "latest/edge"
+          deployment: "unclustered"
+          browser: "chromium"
+          lxd_oidc_client_id: "RYDnMpkygLAMfeo17lU7LYwWGxisRuRR"
```
tarek-hanna, line 101: shouldn't those secrets and oidc config be moved to actions secrets?

.github/workflows/pr.yaml
```diff
@@ -84,6 +84,43 @@ jobs:
...
+          lxd_oidc_user: "lxd-ui-e2e-tests@example.org"
+          lxd_oidc_password: "lxd-ui-e2e-password"
+          lxd_oidc_groups_claim: "lxd-idp-groups"
+
+      - name: Install dependencies
+        shell: bash
+        run: yarn install --immutable
+
+      - name: Run a11y check
+        run: npx playwright test --project a11y-audit
```
tarek-hanna, line 114: Can you verify if npx is installed, i don't think it is ,, it is used to install playwright browser and run tests, but i dont see where  npx is installed. if it is installed in workshop.. these steps run `yarn install` and then `npx playwright test`  on the runner itself not workshop

.github/workflows/pr.yaml
Kxiru: They have always been in this PR.yaml file. I'm not sure making such a large move as changing them now would be appropriate without more context.

.github/workflows/pr.yaml
Kxiru: NPX isn't explicitly installed, it ships with node.js/npm which is in turn provded by the runner by default. Actions/setup-node pins it to version 24.

.github/workflows/pr.yaml
tarek-hanna: alright maybe you can keep a note  about that. but shouldn't block this PR at least :)

tarek-hanna (approved):
LGTM great work Nkeiru 🥇

PR #2141, opened 2026-07-14
test(replicator): fix e2e test

kimanhou:
## Done

- Fix replicator e2e test: add `pointer-events: none` to the description
- Skip file explorer test for 5.0 and 5.21
- Increase timeout for network

Fixes https://warthogs.atlassian.net/browse/WD-37781

## QA

1. Run the LXD-UI:
    - Rely on CI

## Screenshots

image

### DIFF
--- a/tests/helpers/instances.ts
+++ b/tests/helpers/instances.ts
@@ -63,7 +63,7 @@ export const createInstance = async (
 
   await page.getByRole("button", { name: "Create" }).first().click();
 
-  await page.getByText(`Created instance ${instance}.`).waitFor();
+  await dismissNotification(page, `Created instance ${instance}.`);
 };
 
 export const visitInstance = async (
--- a/tests/helpers/network.ts
+++ b/tests/helpers/network.ts
@@ -240,7 +240,7 @@ export const getNetworkLink = async (page: Page, network: string) => {
 
 export const submitCreateNetwork = async (page: Page, network: string) => {
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForTimeout(5000);
+  await page.waitForTimeout(10000);
   const networkLink = await getNetworkLink(page, network);
   await expect(networkLink).toBeVisible();
 };
--- a/tests/helpers/notification.ts
+++ b/tests/helpers/notification.ts
@@ -11,8 +11,10 @@ export const checkNotificationHidden = async (page: Page) => {
 };
 
 export const dismissNotification = async (page: Page, caption: string) => {
-  await page.getByText(caption).waitFor();
-  const notification = page.locator(".toast-notification");
+  const notification = page.locator(".toast-notification", {
+    hasText: caption,
+  });
+  await notification.waitFor({ state: "visible" });
   await notification
     .getByRole("button", { name: "Close notification" })
     .click();
--- a/tests/replicators-clustered.spec.ts
+++ b/tests/replicators-clustered.spec.ts
@@ -29,7 +29,13 @@ test("Replicator", async ({ page, lxdVersion }, testInfo) => {
   await createReplicator(page, replicator, clusterLink, project);
 
   const replicatorRow = page.getByRole("row").filter({ hasText: replicator });
-  await replicatorRow.getByRole("link", { name: replicator }).click();
+  const replicatorLink = replicatorRow.getByRole("link", { name: replicator });
+  await expect(replicatorLink).toBeVisible();
+  const replicatorHref = await replicatorLink.getAttribute("href");
+  if (!replicatorHref) {
+    throw new Error("Replicator detail link href is missing");
+  }
+  await page.goto(replicatorHref);
   await expect(page.getByRole("heading", { name: "General" })).toBeVisible();
   await expect(page.getByRole("heading", { name: "Target" })).toBeVisible();
   await expect(page.getByRole("heading", { name: "Sync" })).toBeVisible();
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -40,7 +40,7 @@ const config: PlaywrightTestConfig<TestOptions> = {
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
-    actionTimeout: 0,
+    actionTimeout: 120_000,
     baseURL: "https://localhost:8407/",
     ignoreHTTPSErrors: true,
     video: "retain-on-failure",
--- a/tests/helpers/instance-file-explorer.ts
+++ b/tests/helpers/instance-file-explorer.ts
@@ -1,9 +1,16 @@
 import type { Page } from "@playwright/test";
-import { expect } from "../fixtures/lxd-test";
+import { expect, test, type LxdVersions } from "../fixtures/lxd-test";
 import { dismissNotification } from "./notification";
 import { assertTextVisible } from "./permissions";
 import { visitInstance } from "./instances";
 import { randomNameSuffix } from "./name";
+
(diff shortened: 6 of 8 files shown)


tests/helpers/notification.ts
```diff
@@ -10,9 +10,15 @@ export const checkNotificationHidden = async (page: Page) => {
...
-  const notification = page.locator(".toast-notification");
+export const dismissNotification = async (
+  page: Page,
+  caption: string,
+  timeout = 30000,
+) => {
+  const notification = page.locator(".toast-notification", {
+    hasText: caption,
+  });
+  await notification.waitFor({ state: "visible", timeout });
```
omarelkashef: If I understand correctly https://playwright.dev/docs/api/class-locator#locator-wait-for-option-timeout, adding a timeout here will only reduce the time we wait for because it can never be more than the timeout of the test itself define in `playwright.cofig.ts`.

tests/helpers/network.ts
```diff
@@ -240,7 +240,7 @@ export const getNetworkLink = async (page: Page, network: string) => {
 
 export const submitCreateNetwork = async (page: Page, network: string) => {
   await page.getByRole("button", { name: "Create", exact: true }).click();
-  await page.waitForTimeout(5000);
+  await page.waitForTimeout(10000);
```
omarelkashef, line 243: Is there a reason we do not dismiss the notification here instead?

src/sass/_replicators.scss
```diff
@@ -1,4 +1,8 @@
 .replicator-table {
+  td .description.u-truncate {
```
omarelkashef: Why is this needed?

omarelkashef (commented):
Thank for the PR.  A few questions below.

src/sass/_replicators.scss
kimanhou: Playwright claims it couldn't click on the replicator name, even tho it was visible, enabled and stable, because the description was covering it.

tests/helpers/network.ts
kimanhou: From @edlerd:

On 5.21 and before, we have the sync create network api call. That call often gets cancelled by the browser because the underlying network on the host (when running lxd on localhost) changes.

That is why we do not rely on the message here, but on a timeout instead. The message will more often not be displayed in CI. In newer versions such as on main or on the 6 track this would work though.

src/sass/_replicators.scss
omarelkashef: That is weird that the description is interfering with the name.

tests/helpers/notification.ts
kimanhou: From Gemini:

Playwright handles hook timeouts and action timeouts completely differently.

120000 (2 minutes) is the global limit for the entire `beforeAll` block. However, individual actions inside that block have a much shorter fuse.

Here is what is actually happening under the hood:

By default, Playwright limits individual actions—like `locator.waitFor()`—to 30,000ms (30 seconds).

Even though your `beforeAll` hook has a maximum budget of 120,000ms, your `dismissNotification` helper is hitting Playwright's default 30-second action timeout inside the hook. When that 30-second timer expires, the step crashes.

Because it crashes inside beforeAll, Playwright halts everything, but the log shows "beforeAll" hook timeout... exceeded because the hook failed to complete its setup.

omarelkashef: the actionTimeout in `playwright.config.ts` is 0 https://github.com/canonical/lxd-ui/blob/main/playwright.config.ts#L43  which means there is no limit and it is only bounded by the timeout of the test.

src/pages/cluster/ReplicatorList.tsx
```diff
@@ -107,7 +107,10 @@ const ReplicatorList: FC<Props> = ({ variant = "main", project, cluster }) => {
         },
         {
           content: (
-            <div className="u-truncate" title={replicator.description}>
+            <div
+              className="description u-truncate"
+              title={replicator.description}
+            >
```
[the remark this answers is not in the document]
kimanhou: `title` still works

tests/helpers/notification.ts
kimanhou: In Playwright, a value of 0 disables this specific configuration tier. Because `actionTimeout` is disabled, locator actions fallback to Playwright's hardcoded default value, which is 30 seconds.

omarelkashef: You can see here that waitFor does not have a default and is controlled by actionTimeout https://playwright.dev/docs/api/class-locator#locator-wait-for  and you can see here that actionTimeout does not have a default  (no limit) https://playwright.dev/docs/test-timeouts#advanced-low-level-timeouts

src/sass/_replicators.scss
edlerd: This seems like a very hacky way to satisfy the test. Also it prevents the title from displaying in case of overflow. Can we find a simpler solution? We should fix the test maybe not to click, but to ensure the text is visible.

tests/helpers/instances.ts
```diff
@@ -63,7 +63,7 @@ export const createInstance = async (
 
   await page.getByRole("button", { name: "Create" }).first().click();
 
-  await page.getByText(`Created instance ${instance}.`).waitFor();
+  await dismissNotification(page, `Created instance ${instance}.`, 90000);
```
edlerd: We should not tweak individual timeouts. Better rely on changing the `actionTimeout` globally in the config file.

tests/helpers/notification.ts
edlerd: We should not customize this timeout. Better change the global `actionTimeout` in the config file to the highest value needed.

edlerd (commented):
Thank you for picking up the CI stability issues and finding the root cause and solutions to the problems. Two requests to alter the solutions below.

src/sass/_replicators.scss
edlerd: We could set `actionTimeout: 120_000` to not time out any actions before the whole test times out.

tests/replicators-clustered.spec.ts
```diff
@@ -29,7 +29,13 @@ test("Replicator", async ({ page, lxdVersion }, testInfo) => {
...
 
   const replicatorRow = page.getByRole("row").filter({ hasText: replicator });
-  await replicatorRow.getByRole("link", { name: replicator }).click();
+  const replicatorLink = replicatorRow.getByRole("link", { name: replicator });
+  await expect(replicatorLink).toBeVisible();
+  const replicatorHref = await replicatorLink.getAttribute("href");
+  if (!replicatorHref) {
+    throw new Error("Replicator detail link href is missing");
+  }
+  await page.goto(replicatorHref);
```
edlerd, line 38: Is this really needed? It looks like a complicated way for the previous check that did the same.

edlerd (approved):
LGTM, one question on the changes in the replicator test.

tests/replicators-clustered.spec.ts
kimanhou: It's just to make sure the name is a link and that it is visible. The only  absolutely necessary line is `await page.goto(replicatorHref);`

edlerd: Why don't we click the link but go through the hoops of reading the href and using `page.goto`?

kimanhou: It's what failed before. Playwright couldn't click on the link because it claimed the description was overlapping and preventing it to reach the link

edlerd: That sounds like there is a more profound problem here then. We shouldn't have to jump through such hoops, but have clean tests. Do you have examples of such cases and can we revert to the "simple" way of clicking here please in a follow up PR?

PR #2150, opened 2026-07-17
feat(identtiy): edit bearer identities WD-37701

omarelkashef:
## Done

- add identity edit panel WD-37701
## QA

1. Run the LXD-UI:
   1. Edit an identity - view details and manage groups
      - Navigate to Permissions -> Identities.
      - Click the "Edit identity" (pencil) icon on a row, or click the groups count link for an identity.
      - Verify the Edit identity panel opens showing read-only Name, ID and Auth method.
      - Add/remove auth groups and save; verify a "Confirm modification" dialog lists the group changes, confirm it, and verify a success toast and updated group count in the list.
   2. Issue a new token for a bearer identity
      - Open the Edit identity panel for an existing Client or DevLXD bearer identity.
      - Verify "Issue new token" button are shown (not present for TLS identities).
      - Click "Issue new token", confirm the warning that the previous token will be invalidated, and confirm.
## Screenshots

image
image

### DIFF
--- a/src/api/auth-identities.tsx
+++ b/src/api/auth-identities.tsx
@@ -6,8 +6,8 @@ import {
 import type { LxdApiResponse } from "types/apiResponse";
 import type { LxdIdentity, TlsIdentityTokenDetail } from "types/permissions";
 import { addEntitlements } from "util/entitlements/api";
-import type { BearerIdentityType } from "util/permissionIdentities";
 import { AUTH_METHOD } from "util/authentication";
+import type { BearerIdentityType } from "util/permissionIdentities";
 import { ROOT_PATH } from "util/rootPath";
 
 export const identitiesEntitlements = ["can_delete", "can_edit"];
@@ -131,7 +131,7 @@ const createBearerIdentity = async (
   }).then(handleResponse);
 };
 
-const issueBearerToken = async (
+export const issueBearerToken = async (
   name: string,
   expiry: string,
 ): Promise<{
--- a/src/context/useAuthGroups.tsx
+++ b/src/context/useAuthGroups.tsx
@@ -4,11 +4,13 @@ import { useAuth } from "./auth";
 import type { LxdAuthGroup } from "types/permissions";
 import { fetchGroups } from "api/auth-groups";
 
-export const useAuthGroups = (): UseQueryResult<LxdAuthGroup[]> => {
+export const useAuthGroups = (
+  enabled = true,
+): UseQueryResult<LxdAuthGroup[]> => {
   const { isFineGrained } = useAuth();
   return useQuery({
     queryKey: [queryKeys.authGroups],
     queryFn: async () => fetchGroups(isFineGrained),
-    enabled: isFineGrained !== null,
+    enabled: enabled && isFineGrained !== null,
   });
 };
--- a/src/pages/permissions/IdentityTokenModal.tsx
+++ b/src/pages/permissions/IdentityTokenModal.tsx
@@ -21,9 +21,10 @@ interface Props {
   notification: string;
   howToUseCli?: ReactNode;
   howToUseUi?: ReactNode;
+  titleSuffix?: string;
 }
 
-const CreateIdentityModal: FC<Props> = ({
+const IdentityTokenModal: FC<Props> = ({
   onClose,
   token,
   identity,
@@ -31,13 +32,14 @@ const CreateIdentityModal: FC<Props> = ({
   notification,
   howToUseCli,
   howToUseUi,
+  titleSuffix = "created successfully",
 }) => {
   const [isConfirmed, setConfirmed] = useState(false);
   const [howToUseActiveTab, setHowToUseActiveTab] = useState("ui-tab");
 
   return (
     <Modal
-      className="create-tls-identity"
+      className="identity-token-modal"
       title={
         <>
           Identity{" "}
@@ -49,7 +51,7 @@ const CreateIdentityModal: FC<Props> = ({
             bold
             truncate
           />{" "}
-          created successfully
+          {titleSuffix}
         </>
       }
       buttonRow={[
@@ -124,4 +126,4 @@ const CreateIdentityModal: FC<Props> = ({
   );
 };
 
-export default CreateIdentityModal;
+export default IdentityTokenModal;
--- a/src/pages/permissions/PermissionIdentities.tsx
+++ b/src/pages/permissions/PermissionIdentities.tsx
@@ -9,6 +9,7 @@ import {
   CustomLayout,
   NotificationConsumer,
 } from "@canonical/react-components";
+import EditIdentityGroupsBtn from "./actions/EditIdentityGroupsBtn";
 import SelectableMainTable from "components/SelectableMainTable";
 import SelectedTableNotification from "components/SelectedTableNotification";
 import { useEffect, useState, type FC } from "react";
@@ -22,11 +23,10 @@ import PermissionIdentitiesFilter, {
   type PermissionIdentitiesFilterType,
 } from "./PermissionIdentitiesFilter";
 import { useSettings } from "context/useSettings";
-import EditIdentityGroupsBtn from "./actions/EditIdentityGroupsBtn";
 import usePanelParams, { panels } from "util/usePanelParams";
 import PageHeader from "components/PageHeader";
 import HelpLink from "components/HelpLink";
-import EditIdentityGroupsPanel from "./panels/EditIdentityGroupsPanel";
+import EditIdentityPanel from "./panels/EditIdentityPanel";
 import Tag from "components/Tag";
 import BulkDeleteIdentitiesBtn from "./actions/BulkDeleteIdentitiesBtn";
 import DeleteIdentityBtn from "./actions/DeleteIdentityBtn";
@@ -41,6 +41,7 @@ import {
 import CreateIdentity from "pages/permissions/CreateIdentity";
 import PermissionIdentitiesActions from "pages/permissions/PermissionIdentitiesActions";
 import ResourceLabel from "components/ResourceLabel";
+import EditIdentityGroupsPanel from "./panels/EditIdentityGroupsPanel";
 
 const PermissionIdentities: FC = () => {
   const notify = useNotify();
@@ -118,15 +119,15 @@ const PermissionIdentities: FC = () => {
 
   const rows = filteredIdentities.map((identity) => {
     const isLoggedInIdentity = settings?.auth_user_name === identity.id;
-    const openGroupPanelForIdentity = () => {
-      panelParams.openIdentityGroups(identity.id);
+    const openEditIdentityPanel = () => {
+      panelParams.openEditIdentity(identity.id);
       setSelectedIdentityIds([identity.id]);
     };
 
     const getGroupLink = () => {
       if (canEditIdentity(identity)) {
         return (
-          <Button appearance="link" dense onClick={openGroupPanelForIdentity}>
+          <Button appearance="link" dense onClick={openEditIdentityPanel}>
             {identity.groups?.length || 0}
           </Button>
         );
@@ -196,17 +197,17 @@ const PermissionIdentities: FC = () => {
                 className="u-no-margin--bottom"
                 hasIcon
                 dense
-                onClick={openGroupPanelForIdentity}
+                onClick={openEditIdentityPanel}
                 type="button"
-                aria-label="Manage groups"
+                aria-label="Edit identity"
                 title={
-                  canEditIdentity()
-                    ? "Manage groups"
+                  canEditIdentity(identity)
+                    ? "Edit identity"
                     : "You do not have permission to modify this identity"
                 }
                 disabled={!canEditIdentity(identity)}
               >
-                <Icon name="user-group" />
+                <Icon name="edit" />
               </Button>
               {hasAccessManagementTLS && (
                 <DeleteIdentityBtn identity={identity} />
@@ -240,6 +241,10 @@ const PermissionIdentities: FC = () => {
   if (isLoading) {
     return <Spinner className="u-loader" text="Loading..." isMainComponent />;
   }
+
+  const editedIdentity = identities.find(
+    (identity) => identity.id === panelParams.identity,
+  );
 
   const getTablePaginationDescription = () => {
     // This is needed because TablePagination does not cater for plural identity
@@ -351,6 +356,14 @@ const PermissionIdentities: FC = () => {
       {panelParams.panel === panels.identityGroups && (
         <EditIdentityGroupsPanel
           identities={selectedIdentities}
+          onClose={() => {
+            setSelectedIdentityIds([]);
+          }}
+        />
+      )}
+      {panelParams.panel === panels.editIdentity && editedIdentity && (
+        <EditIdentityPanel
+          identity={editedIdentity}
           onClose={() => {
             setSelectedIdentityIds([]);
           }}
--- a/src/pages/permissions/panels/CreateIdentityPanel.tsx
+++ b/src/pages/permissions/panels/CreateIdentityPanel.tsx
@@ -1,8 +1,6 @@
 import {
   ActionButton,
   Button,
-  Input,
-  RadioInput,
   ScrollableContainer,
   SidePanel,
   useNotify,
@@ -21,10 +19,12 @@ import {
   createFineGrainedTlsIdentity,
 } from "api/auth-identities";
 import NameWithGroupForm from "pages/permissions/forms/NameWithGroupForm";
+import TokenExpirySelector from "pages/permissions/panels/TokenExpirySelector";
 import GroupSelection from "pages/permissions/panels/GroupSelection";
 import { base64EncodeObject } from "util/helpers";
 import {
   BEARER_EXPIRY_PATTERN,
+  BEARER_EXPIRY_VALIDATION_TEXT,
   IDENTITY_TYPE,
   IDENTITY_TYPE_HELP_TEXT,
   IDENTITY_TYPES_WITH_EXPIRY,
@@ -153,7 +153,7 @@ const CreateIdentityPanel: FC<Props> = ({ onSuccess }) => {
       then: (schema) =>
         schema.test(
           "valid-expiry-format",
-          "Use format like 1d 3H 5M with units y, m, w, d, H, M, or S",
+          BEARER_EXPIRY_VALIDATION_TEXT,
           (value) => !value || BEARER_EXPIRY_PATTERN.test(value.trim()),
         ),
       otherwise: (schema) => schema.notRequired(),
@@ -264,54 +264,30 @@ const CreateIdentityPanel: FC<Props> = ({ onSuccess }) => {
             {IDENTITY_TYPES_WITH_EXPIRY.includes(
               formik.values.identityType,
             ) && (
-              <div className="create-identity-panel-token-expiry u-sv1">
-                <label id="token-expiry-label" htmlFor="expiry">
-                  Token expiry
-                </label>
-                <RadioInput
-                  label="Default (10 years)"
-                  checked={!formik.values.isCustomExpiry}
-                  disabled={isNameInvalid}
-                  onChange={() => {
-                    formik.setFieldValue("isCustomExpiry", false);
+              <TokenExpirySelector
+                isCustomExpiry={!!formik.values.isCustomExpiry}
+                expiry={formik.values.expiry}
+                disabled={isNameInvalid}
+                error={
+                  formik.values.isCustomExpiry &&
+                  formik.touched.expiry &&
+                  formik.errors.expiry
+                    ? formik.errors.expiry
+                    : undefined
+                }
+                onIsCustomExpiryChange={(isCustom) => {
+                  formik.setFieldValue("isCustomExpiry", isCustom);
+                  if (!isCustom) {
                     formik.setFieldValue("expiry", "");
-                  }}
-                />
-                <div className="create-identity-panel-token-expiry-custom-container">
-                  <RadioInput
-                    label="Custom"
-                    checked={formik.values.isCustomExpiry}
-                    disabled={isNameInvalid}
-                    onChange={() => {
-                      formik.setFieldValue("isCustomExpiry", true);
-                    }}
-                  />
-                  <Input
-                    type="text"
-                    id="expiry"
-                    name="expiry"
-                    onChange={formik.handleChange}
-                    value={
-                      formik.values.isCustomExpiry ? formik.values.expiry : ""
-                    }
-                    error={
-                      formik.values.isCustomExpiry &&
-                      formik.touched.expiry &&
-                      formik.errors.expiry
-                        ? formik.errors.expiry
-                        : undefined
-                    }
-                    placeholder="e.g. 1d 3H 5M"
-                    disabled={isNameInvalid || !formik.values.isCustomExpiry}
-                    help={
-                      <>
-                        Space-separated durations: {"<number><unit>"} <br />
-                        Units are case-sensitive: y, m, w, d, H, M, S
-                      </>
-                    }
-                  />
-                </div>
-              </div>
+                  }
+                }}
+                onExpiryChange={(value) => {
+                  formik.setFieldValue("expiry", value);
+                }}
+                onExpiryBlur={() => {
+                  formik.setFieldTouched("expiry", true);
+                }}
+              />
             )}
             <label htmlFor="group-selection-table">Auth groups</label>
             <GroupSelection
@@ -364,9 +340,17 @@ const CreateIdentityPanel: FC<Props> = ({ onSuccess }) => {
                 onClick={() => void formik.submitForm()}
                 className="u-no-margin--bottom"
                 disabled={
-                  !formik.isValid || formik.isSubmitting || !formik.values.name
+                  !formik.isValid ||
+                  formik.isSubmitting ||
+                  !formik.values.name ||
+                  (formik.values.isCustomExpiry && !formik.values.expiry)
                 }
                 loading={formik.isSubmitting}
+                title={
+                  formik.values.isCustomExpiry && !formik.values.expiry
+                    ? "Please provide a valid expiry date"
+                    : undefined
+                }
               >
                 Create identity
               </ActionButton>
--- a/src/pages/permissions/panels/EditIdentityGroupsSection.tsx
+++ b/src/pages/permissions/panels/EditIdentityGroupsSection.tsx
@@ -1,0 +1,139 @@
+import { useEffect, useState, type FC } from "react";
+import type { LxdAuthGroup, LxdIdentity } from "types/permissions";
+import GroupSelection from "./GroupSelection";
+
+export interface IdentityGroupChanges {
+  currentGroups: Set<string>;
+  addedGroups: Set<string>;
+  removedGroups: Set<string>;
+}
+
+interface Props {
+  identity?: LxdIdentity;
+  groups: LxdAuthGroup[];
+  canEdit: boolean;
+  onChange: (
+    changes: IdentityGroupChanges,
+    modifiedGroups: Set<string>,
+  ) => void;
+}
+
+const EditIdentityGroupsSection: FC<Props> = ({
+  identity,
+  groups,
+  canEdit,
+  onChange,
+}) => {
+  const identityId = identity?.id ?? "";
+
+  const [selectedGroups, setSelectedGroupsState] = useState<Set<string> | null>(
+    null,
+  );
+
+  useEffect(() => {
+    if (!identity) {
+      setSelectedGroupsState(null);
+      return;
+    }
+
+    const initialGroups = new Set(identity.groups ?? []);
+    setSelectedGroupsState(initialGroups);
+    onChange(
+      {
+        currentGroups: initialGroups,
+        addedGroups: new Set<string>(),
+        removedGroups: new Set<string>(),
+      },
+      new Set<string>(),
+    );
+  }, [identityId]);
+
+  const originalGroups = new Set(identity?.groups ?? []);
+  const groupSelection = selectedGroups ?? originalGroups;
+
+  const getGroupChanges = (currentGroups: Set<string>) => {
+    const modifiedGroups = new Set<string>();
+    for (const group of groups) {
+      const isSelected = currentGroups.has(group.name);
+      const wasSelected = originalGroups.has(group.name);
+      if (isSelected !== wasSelected) {
+        modifiedGroups.add(group.name);
+      }
+    }
+
+    const addedGroups = new Set(
+      Array.from(modifiedGroups).filter((group) => currentGroups.has(group)),
+    );
+    const removedGroups = new Set(
+      Array.from(modifiedGroups).filter((group) => !currentGroups.has(group)),
+    );
+
+    return {
+      modifiedGroups,
+      changes: {
+        currentGroups,
+        addedGroups,
+        removedGroups,
+      },
+    };
+  };
+
+  const { modifiedGroups } = getGroupChanges(groupSelection);
+
+  const updateSelection = (currentGroups: Set<string>) => {
+    setSelectedGroupsState(currentGroups);
+    const { changes, modifiedGroups } = getGroupChanges(currentGroups);
+    onChange(changes, modifiedGroups);
+  };
+
+  const toggleGroup = (groupName: string) => {
+    if (!canEdit) {
+      return;
+    }
+    const newSelection = new Set(groupSelection);
+    if (newSelection.has(groupName)) {
+      newSelection.delete(groupName);
+    } else {
+      newSelection.add(groupName);
+    }
+    updateSelection(newSelection);
+  };
+
+  const setSelectedGroups = (
+    newSelectedGroups: string[],
+    isUnselectAll?: boolean,
+  ) => {
+    if (!canEdit) {
+      return;
+    }
+    if (isUnselectAll) {
+      updateSelection(new Set());
+    } else {
+      updateSelection(new Set(newSelectedGroups));
+    }
+  };
+
+  return (
+    <>
+      {identity && (
+        <>
+          <label htmlFor="group-selection-table">Auth groups</label>
+          <GroupSelection
+            groups={groups}
+            modifiedGroups={modifiedGroups}
+            parentItemName="identity"
+            parentItems={[identity]}
+            selectedGroups={groupSelection}
+            setSelectedGroups={setSelectedGroups}
+            toggleGroup={toggleGroup}
+            scrollDependencies={[groups, modifiedGroups.size, identity]}
+            disabled={!canEdit}
+            hasScrollableTable={false}
+          />
+        </>
+      )}
+    </>
+  );
+};
+
+export default EditIdentityGroupsSection;
--- a/src/pages/permissions/panels/EditIdentityPanel.tsx
+++ b/src/pages/permissions/panels/EditIdentityPanel.tsx
@@ -1,0 +1,200 @@
+import { useEffect, useState, type FC } from "react";
+import {
+  ConfirmationModal,
+  ScrollableContainer,
+  SidePanel,
+  useNotify,
+  useToastNotification,
+} from "@canonical/react-components";
+import { useQueryClient } from "@tanstack/react-query";
+import usePanelParams from "util/usePanelParams";
+import NotificationRow from "components/NotificationRow";
+import { useAuthGroups } from "context/useAuthGroups";
+import { updateIdentity } from "api/auth-identities";
+import { queryKeys } from "util/queryKeys";
+import { useIdentityEntitlements } from "util/entitlements/identities";
+import IdentityResource from "components/IdentityResource";
+import GroupsOrIdentityChangesTable from "./GroupOrIdentityChangesTable";
+import {
+  getChangesInGroupsForIdentities,
+  pivotIdentityGroupsChangeSummary,
+} from "util/permissionIdentities";
+import { type IdentityGroupChanges } from "./EditIdentityGroupsSection";
+import EditIdentityPanelContent from "./EditIdentityPanelContent";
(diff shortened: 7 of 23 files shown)


src/api/auth-identities.tsx
```diff
@@ -110,3 +111,74 @@ export const createFineGrainedTlsIdentity = async (
...
+): Promise<{
+  token: string;
+}> => {
+  await createBearerIdentity(name, type, groups);
+  try {
+    return await issueBearerToken(name, expiry);
+  } catch (e) {
+    // Token issuance failed — delete the identity we just created and
+    // rethrow the original error so the caller can display it.
+    await fetch(
```
kimanhou: Sorry, this has changed in my last push.

```
const identity = {
      id: name,
      name,
      type,
      authentication_method: AUTH_METHOD.BEARER,
      fine_grained: true,
    };
    deleteIdentity(identity).catch(() => undefined);
```

And you need to make `tls_certificate` optional in src/types/permissions.d.ts

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,289 @@
...
+        <OutputField
+          id="identity-type"
+          label="Type"
+          value={identity?.type ?? ""}
+        />
+        {isBearer && (
+          <div className="u-sv2">
+            <label htmlFor="identity-token-actions">Bearer token</label>
+            <div id="identity-token-actions" className="u-flex u-gap--small">
+              <Button
```
kimanhou: Any reason not to use ConfirmationButton for "Re-issue token" and "Revoke token" ? It would be simpler as the modal is included in the button.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,289 @@
...
+        className="edit-identity-panel"
+      >
+        <SidePanel.Header>
+          <SidePanel.HeaderTitle>Edit identity</SidePanel.HeaderTitle>
+        </SidePanel.Header>
+        <NotificationRow className="u-no-padding" />
+        <OutputField
+          id="identity-name"
+          label="Name"
+          value={identity ? getIdentityName(identity) : ""}
```
kimanhou: It might be worth having an early return if the identity doesn't exist. Then assume the identity always exist afterwards.

src/pages/permissions/panels/EditIdentityPanel.tsx
omarelkashef: Yes, good catch!

src/pages/permissions/panels/EditIdentityPanel.tsx
omarelkashef: That makes things simpler, yes!

src/api/auth-identities.tsx
```diff
@@ -131,9 +131,9 @@ const createBearerIdentity = async (
   }).then(handleResponse);
 };
 
-const issueBearerToken = async (
+export const issueBearerToken = async (
   name: string,
-  expiry: string,
+  expiry?: string,
```
edlerd: Expiry should be mandatory.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+                      }
+                      className="u-no-margin--bottom"
+                      loading={reissuingToken}
+                      shiftClickEnabled
+                      showShiftClickHint
+                      hasIcon
+                    >
+                      <Icon name="restart" />
+                      <span>Issue new token</span>
+                    </ConfirmationButton>
```
edlerd: Let's isolate this in a dedicated component.

We need an option to define an expiry here, similar as on creation of the identity.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+                      }
+                      className="u-no-margin--bottom"
+                      loading={revokingToken}
+                      shiftClickEnabled
+                      showShiftClickHint
+                      hasIcon
+                    >
+                      <Icon name="delete" />
+                      <span>Revoke token</span>
+                    </ConfirmationButton>
```
edlerd: Please isolate in a component.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+                          ? undefined
+                          : "You do not have permission to modify this identity"
+                      }
+                      className="u-no-margin--bottom"
+                      loading={revokingToken}
+                      shiftClickEnabled
+                      showShiftClickHint
+                      hasIcon
+                    >
+                      <Icon name="delete" />
```
edlerd: Theming is broken in light mode

Image

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+                      <span>Revoke token</span>
+                    </ConfirmationButton>
+                  </div>
+                </div>
+              </>
+            )}
+            <EditIdentityGroupsSection
```
edlerd: The layout should be responsive. When the available height reduces, the groups are unaccessible.

https://github.com/user-attachments/assets/3e4412d2-a748-43ae-b4d2-cfb011f880ce

src/pages/permissions/IdentityTokenModal.tsx
```diff
@@ -30,6 +31,7 @@ const CreateIdentityModal: FC<Props> = ({
   notification,
   howToUseCli,
   howToUseUi,
+  titleSuffix = "created successfully",
```
edlerd, line 35: Good idea to reuse this one.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+              id="identity-type"
+              label="Type"
+              value={identity.type}
+            />
+            {isBearer && (
+              <>
+                <OutputField
+                  id="identity-expires-at"
+                  label="Expires at"
+                  value={identity.expires_at || "Never"}
```
edlerd: Let's format this, we have sufficient helper functions for that.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+                          : "You do not have permission to modify this identity"
+                      }
+                      className="u-no-margin--bottom"
+                      loading={reissuingToken}
+                      shiftClickEnabled
+                      showShiftClickHint
+                      hasIcon
+                    >
+                      <Icon name="restart" />
+                      <span>Issue new token</span>
```
edlerd: The icons look slightly too high next to the text. But my eye might be wrong here:

Image

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+            <OutputField
+              id="identity-auth-method"
+              label="Auth method"
+              value={identity.authentication_method.toUpperCase()}
+            />
+            <OutputField
+              id="identity-type"
+              label="Type"
+              value={identity.type}
+            />
```
edlerd: This is a big wall of informational text. I wonder if the ID is relevant to show here. Maybe we can remove it. Also, we can consider condensing the two types into a single field or even hiding them completely here.

Image

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,372 @@
...
+                      hasIcon
+                    >
+                      <Icon name="restart" />
+                      <span>Issue new token</span>
+                    </ConfirmationButton>
+                    <ConfirmationButton
+                      type="button"
+                      appearance="negative"
+                      confirmationModalProps={{
+                        title: "Revoke token",
```
edlerd: After revocation, the UI should reflect that there is no token for this identity. It currently seems like there would be a token that never expires as a result of the revocation.

Image

edlerd (commented):
Thanks for rebasing this. I played around with it and have a few thoughts on the proposed layout and flow. We can also discuss in person.

src/api/auth-identities.tsx
```diff
@@ -153,6 +153,15 @@ const issueBearerToken = async (
...
 
+export const revokeBearerToken = async (name: string): Promise<void> => {
+  await fetch(
+    `${ROOT_PATH}/1.0/auth/identities/bearer/${encodeURIComponent(name)}/token`,
+    {
+      method: "DELETE",
+    },
+  ).then(handleResponse);
+};
+
```
edlerd: What is the user story, where you would want to revoke a token and not delete the identity or issue a new one? I wonder if this functionality should be removed completely to keep things simple. Yes it is a feature, but I don't see the use case yet. Best discuss with the backend team as well.

omarelkashef: I can not think of a reason why someone would use the revoke feature either. I am reaching out to the core team.

src/context/useAuthGroups.tsx
```diff
@@ -4,11 +4,13 @@ import { useAuth } from "./auth";
 import type { LxdAuthGroup } from "types/permissions";
 import { fetchGroups } from "api/auth-groups";
 
-export const useAuthGroups = (): UseQueryResult<LxdAuthGroup[]> => {
+export const useAuthGroups = (
+  enabled = true,
```
edlerd, line 8:
```suggestion
```
We probably can simplify here.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,335 @@
...
+  const isQueryEnabled = Boolean(identityId && authMethod);
+
+  const {
+    data: identity,
+    error,
+    isLoading,
+  } = useIdentity(identityId, authMethod, isQueryEnabled);
+
+  const isGroupQueryEnabled = !!identity;
+  const { data: groups = [] } = useAuthGroups(isGroupQueryEnabled);
```
edlerd:
```suggestion
  const { data: groups = [] } = useAuthGroups();
```
We can always send this query, why make it dependent on the identity being loaded?

src/pages/permissions/panels/CreateIdentityPanel.tsx
```diff
@@ -228,54 +228,31 @@ const CreateIdentityPanel: FC<Props> = ({ onSuccess }) => {
...
-                </label>
-                <RadioInput
-                  label="Default (10 years)"
-                  checked={!formik.values.isCustomExpiry}
-                  disabled={isNameInvalid}
-                  onChange={() => {
-                    formik.setFieldValue("isCustomExpiry", false);
+              <TokenExpirySelector
+                inputId="expiry"
+                isCustomExpiry={Boolean(formik.values.isCustomExpiry)}
```
edlerd:
```suggestion
                isCustomExpiry={formik.values.isCustomExpiry}
```
A bit simpler

src/types/forms/identity.d.ts
```diff
@@ -5,5 +5,5 @@ export interface IdentityFormValues {
   groups?: string[];
   identityType: IdentityType;
   expiry: string;
-  isCustomExpiry: boolean;
+  isCustomExpiry?: boolean;
```
edlerd, line 8: I think this should not be changed.
```suggestion
  isCustomExpiry: boolean;
```

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,335 @@
...
+                  modifiedGroups.size,
+                ]}
+                belowIds={["panel-footer"]}
+              >
+                <OutputField
+                  id="identity-name"
+                  label="Name"
+                  value={getIdentityName(identity)}
+                />
+                <OutputField id="identity-id" label="ID" value={identity.id} />
```
edlerd:
```suggestion
```
Shall we remove the id from this form?

edlerd (commented):
Thank for the iterations, few comments from QA below.

src/pages/permissions/panels/EditIdentityPanel.tsx
omarelkashef: The id is truncated in the table so this is the only place it shows. If the users need to copy or view it, we should keep it. Otherwise, we should remove it.

src/types/forms/identity.d.ts
omarelkashef: For the edit panel, we do not know whether the expiry is custom or not. So, we either make it optional or we create a new type for the edit case and change the interface of `CreateIdentityModal` and I thought the former is simpler.

src/pages/permissions/panels/EditIdentityPanel.tsx
omarelkashef: If there is no identity, we will not have the panel and therefore there is no reason for an unnecessary group query.

src/pages/permissions/panels/IssueNewTokenButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setExpiryError(undefined);
+    }
+  };
+
+  const handleConfirmReissue = () => {
+    onReissueToken(isCustomExpiry ? trimmedExpiry : "");
+  };
+
+  return (
+    <ConfirmationButton
```
edlerd: We should disable the `Next time, you can skip this confirmation by holding SHIFT and clicking the action.` feature, as we need confirmation on the default expiry.

src/util/usePanelParams.tsx
```diff
@@ -6,6 +6,8 @@ export interface PanelHelper {
   bucket: string | null;
   group: string | null;
   identity: string | null;
+  editIdentityId: string | null;
```
edlerd:
```suggestion
```
Let's reuse `identity` parameter and not introduce a custom one here. Wdyt?

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+const EditIdentityPanel: FC<Props> = ({ onClose }) => {
+  const panelParams = usePanelParams();
+  const notify = useNotify();
+  const toastNotify = useToastNotification();
+  const queryClient = useQueryClient();
+  const { canEditIdentity } = useIdentityEntitlements();
+
+  const identityId = panelParams.editIdentityId ?? "";
+  const authMethod = panelParams.editIdentityAuthMethod ?? "";
```
edlerd:
```suggestion
```
Let's remove the auth method from the params. The id is unique. We can pass in the `identities` array from the `PermissionIdentities` component in this `EditIdentityPanel`, then select the identity by id in here, dropping the auth method param and the additional `useIdentity` query.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const isBearer = identity ? isBearerIdentityType(identity.type) : false;
+
+  const invalidateIdentityQueries = () => {
+    queryClient.invalidateQueries({ queryKey: [queryKeys.identities] });
+  };
+
+  const handleReissueToken = (expiry: string) => {
+    if (!identity) {
+      return;
+    }
```
edlerd: We shouldn't silently abort. This should ideally be refactored so we don't need the early exit.

I think we can best isolate this into an `ReissueIdentityTokenBtn` and it makes the identity parameter mandatory itself.

That way we can also isolate the token state variables and the modal into a dedicated place, cleaning up this component.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+    setReissuingToken(true);
+    issueBearerToken(identity.name, expiry)
+      .then((response) => {
+        invalidateIdentityQueries();
+        toastNotify.success(
+          <>
+            New token issued for{" "}
+            <IdentityResource identity={identity} variant="label" />.
+          </>,
+        );
```
edlerd:
```suggestion
```
With the success dialog, I think we should not have an additional notification. It will be hidden behind the modal otherwise.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+  const identityFormValues: IdentityFormValues = {
+    name: identity?.name || identity?.id || "",
+    groups: identity?.groups ?? [],
+    identityType: identity?.type as IdentityType,
+    expiry: identity?.expires_at || "",
+  };
+
+  const identityModalCaptions =
+    CREATE_IDENTITY_MODAL_TEXT[identityFormValues.identityType];
```
edlerd: We are now picking from an object named create in the edit flow. Let's rename it so there are no surprises in the future on maintaining the code.

src/pages/permissions/panels/IssueNewTokenButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+  reissuingToken: boolean;
+  onReissueToken: (expiry: string) => void;
+}
+
+const IssueNewTokenButton: FC<Props> = ({
+  identity,
+  canEdit,
+  hasActiveToken,
+  reissuingToken,
+  onReissueToken,
```
edlerd:
```suggestion
```
I think the interface on this can be minimal, and all the state can become local, and the trigger for updates can be isolated in here as well.

src/pages/permissions/panels/IdentityTokenActions.tsx
```diff
@@ -0,0 +1,36 @@
...
+  onReissueToken: (expiry: string) => void;
+}
+
+const IdentityTokenActions: FC<Props> = ({
+  identity,
+  canEdit,
+  reissuingToken,
+  onReissueToken,
+}) => {
+  const hasActiveToken = Boolean(identity.expires_at);
```
edlerd:
```suggestion
```
This can be calculated internally in `IssueNewTokenButton`

src/pages/permissions/panels/IdentityTokenActions.tsx
```diff
@@ -0,0 +1,36 @@
...
+import IssueNewTokenButton from "./IssueNewTokenButton";
+
+interface Props {
+  identity: LxdIdentity;
+  canEdit: boolean;
+  reissuingToken: boolean;
+  onReissueToken: (expiry: string) => void;
+}
+
+const IdentityTokenActions: FC<Props> = ({
```
edlerd: Can we remove this indirection component?

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+                        isoTimeToString(identity.expires_at ?? "") ||
+                        "No active token"
+                      }
+                    />
+                    <IdentityTokenActions
+                      identity={identity}
+                      canEdit={canEdit}
+                      reissuingToken={reissuingToken}
+                      onReissueToken={handleReissueToken}
+                    />
```
edlerd: Let's make this inline and with base appearance to make the layout a bit slimer.

Image

src/pages/permissions/panels/GroupSelection.tsx
```diff
@@ -27,6 +27,7 @@ interface Props {
   preselectedGroups?: Set<string>;
   belowIds?: string[];
   disabled?: boolean;
+  useScrollableTable?: boolean;
```
edlerd:
```suggestion
  hasScrollableTable?: boolean;
```
Nice idea, small nitpick on the naming.

src/pages/permissions/panels/TokenExpirySelector.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </label>
+      <RadioInput
+        label="Default (10 years)"
+        checked={!isCustomExpiry}
+        disabled={disabled}
+        onChange={() => {
+          onIsCustomExpiryChange(false);
+        }}
+      />
+      <div className="create-identity-panel-token-expiry-custom-container">
```
edlerd:
```suggestion
      <div className="token-expiry-custom">
```
We should generalize the name for the generic component

src/pages/permissions/panels/TokenExpirySelector.tsx
```diff
@@ -0,0 +1,70 @@
...
+  isCustomExpiry,
+  expiry,
+  disabled = false,
+  error,
+  onIsCustomExpiryChange,
+  onExpiryChange,
+  onExpiryBlur,
+}) => {
+  return (
+    <div className="create-identity-panel-token-expiry u-sv1">
```
edlerd:
```suggestion
    <div className="token-expiry-selector u-sv1">
```

tests/permission-identities.spec.ts
```diff
@@ -115,3 +183,43 @@ test("manage groups for many identities", async ({ page, lxdVersion }) => {
...
+    expect(reissuedToken).not.toEqual(initialToken);
+    await dismissNotification(page, `New token issued for ${identityLabel}.`);
+    await closeTokenDisplayModal(page);
+  } finally {
+    // Clean up identity regardless of assertion outcomes.
+    const cleanup = await page.request.delete(
+      `/1.0/auth/identities/bearer/${encodeURIComponent(identityName)}`,
+    );
+    expect([200, 202, 204, 404]).toContain(cleanup.status());
+  }
```
edlerd: Let's avoid the try/catch here.

tests/permission-identities.spec.ts
```diff
@@ -115,3 +183,43 @@ test("manage groups for many identities", async ({ page, lxdVersion }) => {
...
 });
+
+test("reissue a new token for bearer identity", async ({
+  page,
+  lxdVersion,
+}) => {
+  skipIfFineGrainedAuthorisationNotSupported(lxdVersion);
+  skipIfBearerIdentitiesNotSupported(lxdVersion);
+
+  const identityName = randomIdentityName();
```
edlerd:
```suggestion
  const identity = randomIdentityName();
```
We use short names in tests.

tests/permission-identities.spec.ts
```diff
@@ -115,3 +183,43 @@ test("manage groups for many identities", async ({ page, lxdVersion }) => {
...
+  const identityLabel = `${identityName} (Client token bearer, ${identityName})`;
```
edlerd:
```suggestion
```
This can be inlined in the assertion.

tests/permission-identities.spec.ts
```diff
@@ -115,3 +183,43 @@ test("manage groups for many identities", async ({ page, lxdVersion }) => {
...
+    await closeTokenDisplayModal(page);
+
+    const identityRow = page.getByRole("row").filter({ hasText: identityName });
+    await expect(identityRow).toBeVisible();
+    await identityRow.getByLabel("Edit identity").click();
+
+    // Reissue and verify token rotation.
+    const reissuedToken = await issueTokenFromEditPanel(page, identityName);
+    expect(reissuedToken).not.toEqual(initialToken);
+    await dismissNotification(page, `New token issued for ${identityLabel}.`);
```
edlerd:
```suggestion
```
We should not have notifications and the modal.

tests/permission-identities.spec.ts
```diff
@@ -17,9 +19,72 @@ import {
...
+  expiry?: string,
+): Promise<string> => {
+  await page.getByRole("button", { name: "Issue new token" }).click();
+  const tokenModal = page
+    .locator(".p-modal")
+    .filter({ hasText: "Issue new token" });
+
+  if (expiry) {
+    await tokenModal.getByLabel("Custom").click();
+    await tokenModal.getByPlaceholder("e.g. 1d 3H 5M").fill(expiry);
```
edlerd: Can we use `getByLabel` to ensure the input has a proper label?

tests/permission-identities.spec.ts
```diff
@@ -17,9 +19,72 @@ import {
...
+    if (expiry === undefined) {
+      return true;
+    }
+
+    const payload = res.request().postDataJSON() as { expiry?: string };
+    return payload.expiry === expiry;
+  });
+
+  expect(response.ok()).toBeTruthy();
+};
```
edlerd:
```suggestion
```
We shouldn't do a busy wait like this, instead wait for the title of the success modal.

edlerd (commented):
Thanks for the big changeset. Several thoughts on it below.

src/pages/permissions/panels/EditIdentityPanel.tsx
edlerd: Or maybe better, pick the "editIdentity" in the `PermissionIdentities` based on its `identities` array and the id from the panel params. Then pass in the identity to the edit panel component. That is a cleaner interface.

src/pages/permissions/panels/EditIdentityGroupsSection.tsx
```diff
@@ -0,0 +1,140 @@
...
+    setSelectedGroupsState(initialGroups);
+    onChange(
+      {
+        currentGroups: initialGroups,
+        addedGroups: new Set<string>(),
+        removedGroups: new Set<string>(),
+      },
+      new Set<string>(),
+    );
+  }, [identityId, authMethod]);
```
edlerd:
```suggestion
  }, [identityId]);
```
`authMethod` seems unused inside and not needed as a trigger.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+}
+
+const EditIdentityPanel: FC<Props> = ({ identity, onClose }) => {
+  const panelParams = usePanelParams();
+  const notify = useNotify();
+  const toastNotify = useToastNotification();
+  const queryClient = useQueryClient();
+  const { canEditIdentity } = useIdentityEntitlements();
+
+  const { data: groups = [] } = useAuthGroups(!!identity);
```
edlerd:
```suggestion
  const { data: groups = [] } = useAuthGroups();
```
As this is the edit identity panel, it is expected that the identity is present. Only in an error case is the identity missing. Let's simplify here to always load the groups and not optimize for an error case.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+  getChangesInGroupsForIdentities,
+  pivotIdentityGroupsChangeSummary,
+} from "util/permissionIdentities";
+import { type IdentityGroupChanges } from "./EditIdentityGroupsSection";
+import EditIdentityPanelContent from "./EditIdentityPanelContent";
+import GroupSelectionActions from "../actions/GroupSelectionActions";
+import type { LxdIdentity } from "types/permissions";
+
+interface Props {
+  identity: LxdIdentity | undefined;
```
edlerd:
```suggestion
  identity: LxdIdentity;
```
We should not allow using this component without identity.

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    onClose();
+  };
+
+  const canEdit = canEditIdentity(identity);
+  const addedGroups = pendingGroupChanges?.addedGroups ?? new Set<string>();
+  const removedGroups = pendingGroupChanges?.removedGroups ?? new Set<string>();
+
+  const identityGroupsChangeSummary = identity
+    ? getChangesInGroupsForIdentities([identity], addedGroups, removedGroups)
+    : {};
```
edlerd:
```suggestion
  const identityGroupsChangeSummary = getChangesInGroupsForIdentities([identity], addedGroups, removedGroups);
```
A bit simpler

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    identityGroupsChangeSummary,
+  );
+
+  const closeGroupsConfirmModal = () => {
+    setConfirmingGroups(false);
+    setPendingGroupChanges(null);
+  };
+
+  const handleSaveGroups = () => {
+    if (!identity || !pendingGroupChanges) {
```
edlerd:
```suggestion
    if (!pendingGroupChanges) {
```

src/pages/permissions/panels/EditIdentityPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+      notify.failure(
+        "Update groups failed",
+        new Error("Identity or pending group changes are missing"),
```
edlerd:
```suggestion
        new Error("No pending group changes"),
```

src/pages/permissions/panels/EditIdentityPanelContent.tsx
```diff
@@ -0,0 +1,75 @@
...
+import EditIdentityGroupsSection, {
+  type IdentityGroupChanges,
+} from "./EditIdentityGroupsSection";
+import IssueNewTokenSection from "./IssueNewTokenSection";
+import { isoTimeToString } from "util/helpers";
+
+interface Props {
+  canEdit: boolean;
+  groups: LxdAuthGroup[];
+  identity?: LxdIdentity;
```
edlerd:
```suggestion
  identity: LxdIdentity;
```
This should simplify the component.

src/pages/permissions/panels/EditIdentityPanelContent.tsx
```diff
@@ -0,0 +1,75 @@
...
+  ) => void;
+}
+
+const EditIdentityPanelContent: FC<Props> = ({
+  canEdit,
+  groups,
+  identity,
+  onGroupChanges,
+}) => {
+  const isBearer = identity ? isBearerIdentityType(identity.type) : false;
```
edlerd:
```suggestion
  const isBearer = isBearerIdentityType(identity.type);
```

src/pages/permissions/panels/EditIdentityPanelContent.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <>
+      {identity && (
+        <>
```
edlerd:
```suggestion
```

src/pages/permissions/panels/IssueNewTokenButton.tsx
```diff
@@ -0,0 +1,31 @@
+import { type FC } from "react";
+import { Button, Icon } from "@canonical/react-components";
+
+interface Props {
+  canEdit: boolean;
+  isLoading: boolean;
```
edlerd:
```suggestion
```
A modal is visible while loading, we don't need a loading indicator on this button.

src/pages/permissions/panels/IssueNewTokenButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+        canEdit
+          ? undefined
+          : "You do not have permission to issue a new token for this identity"
+      }
+      className="u-no-margin--bottom"
+      onClick={onClick}
+      hasIcon
+    >
+      <Icon name="restart" />
+      <span>{isLoading ? "Issuing token..." : "Issue new token"}</span>
```
edlerd:
```suggestion
      <span>Issue new token</span>
```

src/pages/permissions/panels/IssueNewTokenSection.tsx
```diff
@@ -0,0 +1,163 @@
...
+      setExpiryError(BEARER_EXPIRY_VALIDATION_TEXT);
+    } else {
+      setExpiryError(undefined);
+    }
+  };
+
+  const closeReissueConfirmModal = () => {
+    if (reissuingToken) {
+      return;
+    }
```
edlerd: This can become an issue if the loading state gets stuck as we block the user from leaving. Maybe we should allow this.
```suggestion
  const closeReissueConfirmModal = () => {
```

src/pages/permissions/panels/IssueNewTokenSection.tsx
```diff
@@ -0,0 +1,163 @@
...
+                setExpiryError(undefined);
+              }
+            }}
+            onExpiryBlur={handleExpiryBlur}
+          />
+        </ConfirmationModal>
+      )}
+
+      {reissuedToken && identityModalCaptions && (
+        <CreateIdentityModal
```
edlerd: Shall we rename this component, to match both this new and existing use cases?

src/pages/permissions/panels/TokenExpirySelector.tsx
```diff
@@ -0,0 +1,70 @@
+import { Input, RadioInput } from "@canonical/react-components";
+import type { FC } from "react";
+
+interface Props {
+  inputId: string;
```
edlerd:
```suggestion
```
Do we have to expose this? I assume we can keep this an implementation detail of this component and don't have to expose it in the interface.

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -356,6 +357,16 @@ const PermissionIdentities: FC = () => {
           }}
         />
       )}
+      {panelParams.panel === panels.editIdentity && (
+        <EditIdentityPanel
+          identity={identities.find(
+            (identity) => identity.id === panelParams.identity,
+          )}
```
edlerd: We might need to introduce a const for the identity selection and guard the rendering for a stricter type. But I think the simpler internal implementation outweighs this small indirection.

tests/permission-identities.spec.ts
```diff
@@ -17,9 +19,47 @@ import {
...
+  expiry?: string,
+): Promise<string> => {
+  await page.getByRole("button", { name: "Issue new token" }).click();
+  const tokenModal = page
+    .locator(".p-modal")
+    .filter({ hasText: "Issue new token" });
+
+  if (expiry) {
+    await tokenModal.getByLabel("Custom").click();
+    await tokenModal.getByLabel("reissue-token-expiry").fill(expiry);
```
edlerd:
```suggestion
    await tokenModal.getByLabel("Token expiry").fill(expiry);
```
This should probably be the input label value here, no?

We might need to add a label just for the input with the `u-off-screen` class. That way screen readers announce the input properly and the layout can stay slim. From the test here, we can reference that off screen label.

edlerd (commented):
Very nice, thanks for the iteration! A few ideas to simplify and streamline from very detailed QA and code review.

src/pages/permissions/panels/IssueNewTokenSection.tsx
edlerd: * meaning rename the `CreateIdentityModal` above.

src/pages/permissions/panels/IssueNewTokenButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+  canEdit: boolean;
+  isLoading: boolean;
+  onClick: () => void;
+}
+
+const IssueNewTokenButton: FC<Props> = ({ canEdit, isLoading, onClick }) => {
+  return (
+    <Button
+      type="button"
+      appearance="base"
```
kimanhou, line 13: Are we set on the appearance of this button ? What about "default" when the expiry is in the future and "negative" when no active token is attached to the identity ?

Image

Image

omarelkashef: I think it makes sense for it to be default if it is inlined like this. Also, why make it negative if there is no token? I think the user should be able to issue a token if none exists. They might have revoked it in the cli and they should be able to issue a new one in the UI.

kimanhou: To catch the user's attention. We can't set it to "positive" because the "Edit" button in the footer is already positive.

edlerd: I prefer the slightly slimmer variant without any border. We also use it in other places, such as instance configuration. But no strong opinion.

Maybe we should make the button green/positive under some conditions. It seems like the primary goal to "issue new token", when a user explores an identity with an expired or non-existing token.

src/pages/permissions/IdentityTokenModal.tsx
```diff
@@ -21,23 +21,25 @@ interface Props {
...
 }) => {
   const [isConfirmed, setConfirmed] = useState(false);
   const [howToUseActiveTab, setHowToUseActiveTab] = useState("ui-tab");
 
   return (
     <Modal
-      className="create-tls-identity"
+      className="create-identity-modal"
```
edlerd:
```suggestion
      className="identity-token-modal"
```
Let's align the class to the new name as well.

tests/permission-identities.spec.ts
```diff
@@ -17,9 +19,47 @@ import {
...
+    await tokenModal.getByLabel("Custom").click();
+    await tokenModal.getByLabel("Token expiry value").fill(expiry);
+  }
+
+  await tokenModal.getByRole("button", { name: "Issue new token" }).click();
+  await expect(page.getByText("token issued successfully")).toBeVisible();
+
+  return getDisplayedToken(page);
+};
+
```
edlerd: Let's move these to the helpers file and keep only tests in this file.

edlerd (dismissed):
QA and code LGTM, thanks for the adjustments!

Two tiny non-blocking nitpicks on structure/naming.

src/pages/permissions/panels/IssueNewTokenSection.tsx
```diff
@@ -0,0 +1,160 @@
...
+        canEdit={canEdit}
+        onClick={() => {
+          setConfirmingReissue(true);
+        }}
+      />
+
+      {confirmingReissue && (
+        <ConfirmationModal
+          title="Issue new token"
+          confirmButtonLabel="Issue new token"
```
kimanhou, line 115: Should the button be disabled if Custom is selected but nothing is inputted yet ?

Image

omarelkashef: Fixed in a pair programming session.

edlerd (dismissed):
LGTM 👍

edlerd (dismissed):

edlerd (approved):

PR #2153, opened 2026-07-21
feat(test) - [WD-37811] Follow-up test for symlink resolution

Kxiru:
## Done

Creates a spec test in the utils folder to call the resolveSymlnkTarget function with different values and verify it’s output.

Tests:
- Relative paths
- Root paths
- Paths to directories
- Paths to files
- Paths to other symlinks

## QA

1. Run the LXD-UI:

## Screenshots

N/A

### DIFF
--- a/src/util/instances.spec.ts
+++ b/src/util/instances.spec.ts
@@ -1,0 +1,115 @@
+import { resolveSymlinkTarget } from "./instances";
+
+describe("resolveSymlinkTarget", () => {
+  describe("relative paths", () => {
+    it("resolves a sibling target relative to the current path", () => {
+      expect(resolveSymlinkTarget("bar", "/home/foo")).toBe("/home/bar");
+    });
+
+    it("resolves a target with ..", () => {
+      expect(resolveSymlinkTarget("../other", "/home/user/file.txt")).toBe(
+        "/home/other",
+      );
+    });
+
+    it("resolves a target with .", () => {
+      expect(resolveSymlinkTarget("./same", "/dir/current")).toBe("/dir/same");
+    });
+
+    it("resolves a target with multiple ..", () => {
+      expect(resolveSymlinkTarget("../../top", "/a/b/c/d")).toBe("/a/top");
+    });
+  });
+
+  describe("root (absolute) paths", () => {
+    it("returns an absolute target as-is", () => {
+      expect(resolveSymlinkTarget("/usr/bin/bash", "/home/user/link")).toBe(
+        "/usr/bin/bash",
+      );
+    });
+
+    it("returns root target as-is", () => {
+      expect(resolveSymlinkTarget("/", "/some/deep/path")).toBe("/");
+    });
+
+    it("returns an absolute target regardless of currentPath", () => {
+      expect(resolveSymlinkTarget("/etc/config", "/")).toBe("/etc/config");
+    });
+  });
+
+  describe("paths to directories", () => {
+    it("resolves a relative directory target", () => {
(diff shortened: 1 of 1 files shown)


tarek-hanna (dismissed):
LGTM .. and Copilot didn't complain either :D

 I wonder if it would make sense to add cases for empty string and null input, to validate error handling or possible edge cases, and also  2 relative paths (as input), since all cases here have at least 1 abs path.

Kxiru:
> LGTM .. and Copilot didn't complain either :D
> 
> I wonder if it would make sense to add cases for empty string and null input, to validate error handling or possible edge cases, and also 2 relative paths (as input), since all cases here have at least 1 abs path.

I think that is a valid concern, I will add these now.

src/util/instances.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  describe("two relative paths", () => {
+    it("resolves a sibling target with a relative currentPath", () => {
+      expect(resolveSymlinkTarget("bar", "home/user/file")).toBe(
+        "/home/user/bar",
+      );
+    });
+
+    it("resolves a target with .. and a relative currentPath", () => {
+      expect(resolveSymlinkTarget("../other", "home/user/file")).toBe(
+        "/home/other",
```
tarek-hanna: not sure if `home` being **relative** and suddenly becoming **absolute** is correct. i would expect it to resolve to something like `/root/.../home/user/other`

src/util/instances.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    it("throws when currentPath is null", () => {
+      expect(() =>
+        resolveSymlinkTarget("bar", null as unknown as string),
+      ).toThrow();
+    });
+  });
+
+  describe("two relative paths", () => {
+    it("resolves a sibling target with a relative currentPath", () => {
+      expect(resolveSymlinkTarget("bar", "home/user/file")).toBe(
```
tarek-hanna: same here. it should first get the absolute path for home wherever it  lives and then handles the relative path  in respect to bar, but again how does that work if i flip the input `expect(resolveSymlinkTarget( "home/user/file", "bar")).toBe( ??)`  >>> `/bar/home/user/file` does the input order matter here ?

tarek-hanna: i see `target` goes first and then `current path` so it might be easier to use if you enforce  the current path to always be absolute?  and check for `starts.with('/')` WDYT? this way the relative  path in the second input will say it is invalid input and should be absolute... again i dont know the exact use cases you need this symlink method for, so feel free to correct me or discard this :D

Kxiru:
@tarek-hanna Fair enough, we're testing undefined behaviour in this relativePaths block as currentPath is never relative. I will remove this test block entirely,  but I believe the Edge cases block is still relevant.

tarek-hanna:
> @tarek-hanna Fair enough, we're testing undefined behaviour in this relativePaths block as currentPath is never relative. I will remove this test block entirely, but I believe the Edge cases block is still relevant.

so basically what you are saying here is that we should enforce validation on the targetPath being absolute since "it can never be relative"?

tarek-hanna (approved):
Just to keep things in scope and unblock you. PR is approved. feel free to close/merge this one and have a follow up later with the extra validations (if needed)

Kxiru:
> so basically what you are saying here is that we should enforce validation on the targetPath being absolute since "it can never be relative"?

I'm saying that the currentPath is never relative, so it was unnecessary to test current paths that _are_. I'm not sure what you meant when you brought up the target paths though.

Thank you for unblocking!

PR #2164, opened 2026-08-03
fix(instance) avoid error reporting endless loop

### LINKED ISSUE #2159: Memory leak / cpu 100% on network disconnect
Hey!

Sometimes, not always (the most fun issues to troubleshoot) when I lose my network connection for some time (I am traveling today) I come back to a frozen LXD tab.

Image

Currently this tab is using 31GB memory and 160% cpu. I had to force close it. 

I was unable to open the developer console to try and debug what function is running in a loop. All I have is the error message on screen.

It doesn't always happen, just sometimes..

edlerd:
## Done

- fix(instance) avoid error reporting endless loop

Fixes #2159 

## QA

1. Run the LXD-UI:
    - open instance details > overview tab
    - stop lxd server
    - switch back to browser tab. see browser reporting single error, not crashing.

### DIFF
--- a/src/pages/instances/InstanceOverview.tsx
+++ b/src/pages/instances/InstanceOverview.tsx
@@ -1,7 +1,13 @@
 import { useEffect, type FC } from "react";
 import { isoTimeToString } from "util/helpers";
 import { ROOT_PATH } from "util/rootPath";
-import { Col, Row, useListener, useNotify } from "@canonical/react-components";
+import {
+  Col,
+  NotificationSeverity,
+  Row,
+  useListener,
+  useNotify,
+} from "@canonical/react-components";
 import type { LxdInstance } from "types/instance";
 import { updateMaxHeight } from "util/updateMaxHeight";
 import InstanceOverviewProfiles from "./InstanceOverviewProfiles";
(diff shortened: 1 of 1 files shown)


src/pages/instances/InstanceOverview.tsx
```diff
@@ -27,7 +33,9 @@ const InstanceOverview: FC<Props> = ({ instance }) => {
   const isClustered = useIsClustered();
 
   const onFailure = (title: string, e: unknown) => {
-    notify.failure(title, e);
+    if (notify.notification?.type !== NotificationSeverity.NEGATIVE) {
```
omarelkashef, line 36: This mitigates the issue reported, but also hides unrelated errors in other cases. How about we move the `onFailure` call of the child components into a useEffect with the error in the deps array so it is only invoked when the error changes?

edlerd: The problem here is that we have three components that can each raise a different error in case of a network failure. The error order is random, and we need to stop the endless "update error" > refresh > "update error" loop. This proposal here was the best I could think of to avoid the problem at hand.

omarelkashef: It solves the problem indeed. But in case the errors are genuine are not just network failures, only one error will be displayed. Is that ok?

edlerd: Good question. We are changing semantics here from "display last" to "display first". Probably okay. Wdyt?

Another option would be to not use the inline errors on pages, but use the toast errors. Though I think when page components fail to load, the inline error is the better choice.

kimanhou: > the inline error is the better choice

Yes

omarelkashef: No, inline errors are better.

omarelkashef (approved):
LGTM, thanks!

PR #2165, opened 2026-08-03
fix(up- or downgrade) refresh ui on server version change

### LINKED ISSUE #2123: Error report
# Description

A brief description of the problem. Should include what you were
attempting to do, what you did, what happened and what you expected to
see happen.

# Metadata

UI Version: 0.15.3
Path: /ui/project/default/storage/volumes

# Stacktrace

```
TypeError: Failed to fetch dynamically imported module: https://127.0.0.1:8443/ui/assets/StorageVolumes-D5ygOpRA.js?cacheBuster=1783328575512
```

edlerd:
## Done

- fix(up- or downgrad) refresh ui on server version change

Fixes #2123

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { Suspense, useEffect, type FC } from "react";
+import { Suspense, useEffect, type FC, useRef } from "react";
 import { Navigate, Route, Routes } from "react-router-dom";
 import ProjectRedirect from "pages/projects/ProjectRedirect";
 import ProjectLoader from "pages/projects/ProjectLoader";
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -159,6 +159,23 @@ const App: FC = () => {
...
+  useEffect(() => {
+    if (!serverVersion) {
+      return;
+    }
+    if (!initialVersion) {
+      setInitialVersion(serverVersion);
+      return;
+    }
+
+    if (serverVersion != initialVersion) {
```
kimanhou:
```suggestion
    if (serverVersion !== initialVersion) {
```

src/App.tsx
```diff
@@ -159,6 +159,23 @@ const App: FC = () => {
   setFavicon();
   setTitle();
 
+  const serverVersion = settings?.environment?.server_version;
+  const [initialVersion, setInitialVersion] = useState<string | null>(null);
```
kimanhou:
```suggestion
  const initialVersionRef = useRef<string | null>(null);
```

Could we use a `useRef` here instead of `useState` ?
Then in the `useEffect`, we could just do
```
  if (initialVersionRef.current === null) {
    initialVersionRef.current = serverVersion;
    return;
  }

  if (serverVersion !== initialVersionRef.current) {
    window.location.reload();
  }
```

src/App.tsx
```diff
@@ -159,6 +159,24 @@ const App: FC = () => {
...
+    if (!initialVersion.current) {
+      initialVersion.current = serverVersion;
+      return;
+    }
+
+    if (serverVersion !== initialVersion.current) {
+      // server was updated or downgraded and the UI is out of sync
+      window.location.reload();
+    }
+  }, [initialVersion, serverVersion]);
```
kimanhou:
```suggestion
  }, [serverVersion]);
```

We don't need `initialVersion` in the dependency array anymore

kimanhou (approved):

PR #2174, opened 2026-08-04
feat: migrate 3 icons to pragma WD-37614

omarelkashef:
## Done

- Migrate 3 icons to pragma

## QA

1. Run the LXD-UI:
   1. Verify the new icons: project, network, and image registry icons render correctly in all locations forms, panels, pages, table rows, etc and in both dark and light themes.

## Screenshots

New image registry icon

image


vite.config.ts
```diff
@@ -1,10 +1,54 @@
...
 
+const DS_ICONS_DIR = path.resolve(
+  __dirname,
+  "node_modules/@canonical/ds-assets/icons",
+);
+
+// Serves @canonical/ds-assets icons at /icons/ without copying files to public/.
+// - dev: intercepts requests to /icons/* and streams from node_modules
+// - build: emits each SVG as a build asset so they land in the output directory
+const dsAssetsIconsPlugin = (): Plugin => ({
```
edlerd, line 17: This is rather complicated, can we find an easier way? Maybe it can be enabled upstream?

src/components/ResourceIcon.tsx
```diff
@@ -1,4 +1,6 @@
 import { Icon } from "@canonical/react-components";
```
edlerd, line 1: Can we also alias this one as `VanillaIcon` to avoid any confusion?

src/components/ResourceIcon.tsx
```diff
@@ -35,27 +37,33 @@ export type ResourceIconType =
...
+// Icons available in @canonical/ds-assets — rendered via DsIcon (<svg><use>).
+// Move entries here from legacyResourceIcons as ds-assets gains coverage.
+const dsResourceIcons: Partial<Record<ResourceIconType, IconName>> = {
+  "image-registry": "image-registries",
+  network: "exposed",
+  project: "folder",
+};
+
+// Icons not yet in @canonical/ds-assets — rendered via vanilla-framework CSS.
+const legacyResourceIcons: Record<string, string> = {
```
edlerd, line 49:
```suggestion
const vanillaResourceIcons: Record<string, string> = {
```

vite.config.ts
```diff
@@ -1,10 +1,54 @@
...
+        continue;
+      }
+
+      this.emitFile({
+        type: "asset",
+        fileName: `icons/${name}`,
+        source: fs.readFileSync(path.join(DS_ICONS_DIR, name), "utf-8"),
+      });
+    }
+  },
```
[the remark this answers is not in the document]
edlerd: Yes, I think we should select the icons we bundle. Wonder if there is an easier way and if this actually works properly for the snap built.

edlerd (commented):
Few thoughts on the changes below.

vite.config.ts
omarelkashef: Maybe we can copy the icons from pragma to /public/icons in a script that we run as part of the build. wdyt?

edlerd: I quite liked the old way of importing the icons via css, and they would bundle automatically. Can you figure out in collaboration with @advl and team if pragma can work in the same way?

edlerd: A custom script or the solution proposed here would work for us, but I fear it is a lot of overhead that every project consuming pragma icons would have to copy. Hence the drive to find a more simple out of the box solution that is enabled upstream in pragma.

omarelkashef: I reached out to the design team, ineed. Let's see what they say.

omarelkashef:
## Trade-off: scan-for-used-icons vs. copy-all-icons

### Measured facts (lxd-ui, current state)
- Icons available in @canonical/ds-assets: 165 files, 664 KB total
- Icons detected as used by scripts/copy-ds-icons.js: 78 files, 316 KB
- Source files scanned: 1007 (.ts/.tsx), ~6.7 MB
- Script runtime: ~0.4 s (real), single pass, no watch mode
- Delta if we copy everything: +87 icons, +348 KB on disk

### Key context
Icons in public/icons/ are static assets served on demand — they are NOT bundled into the JS. So unused icons cost nothing at runtime for the user: the browser only fetches the icons a page actually renders. The 348 KB is purely distribution/image size, not page weight.

### Option A — Keep the scan (current)
Pros
- Ships ~50% fewer icon files (348 KB smaller build output / container image)
- public/icons/ stays a readable inventory of what the app actually uses

Cons
- 0.4 s added to build and to dev start (negligible, but non-zero; runs on  every start, not incremental)
- Correctness risk: the detection is a regex over string literals (/["']([a-z][a-z0-9-]*)["']/). It fails for any icon name that is not a literal in src/ — e.g. names built at runtime (`${type}-icon`), names coming from the API, config, or from @canonical/react-components internals. Result = a missing icon in production, discovered late.
- False positives: any unrelated string that happens to equal an icon name (e.g. "search", "close", "user") is copied anyway, so the savings are already softer than they look.
- Extra script to maintain and to reason about when icons "disappear".

### Option B — Copy all icons
Pros
- Zero correctness risk: any icon name, however it is computed, resolves
- Simpler: a single recursive copy, no scanning, no heuristics
- Build/start time drops to ~0 for this step (a 664 KB copy is faster than reading 6.7 MB of source)
- Removes a class of "works locally, broken in prod" bugs

Cons
- +348 KB in the build output / Docker image / snap
- public/icons/ no longer reflects actual usage



My recommendation: 
Option B. The performance argument in favor of option A does not hold: the scan costs 0.4 s of build time to save 348 KB of static files that are never all downloaded by any user. The saving is not worth it once you weigh in the real cost — the search can silently miss dynamically-constructed icon names and break the UI at runtime. We can revisit the approach if the icon set grows by an order of magnitude (thousands of icons / tens of MB), image size becomes a hard constraint

@edlerd @kimanhou @Kxiru

kimanhou:
@omarelkashef I agree with you on option B. That's also what Adrian recommended yesterday. It's not optimized but it's not worth the energy, time and risk. I hope the icon set won't grow too much.

edlerd:
> * +348 KB in the build output / snap

This will directly translate to a bigger snap size and the snap is downloaded by the users on each install or version update. So I think that we should maintain a list of used icons to avoid sending unused icons around.

omarelkashef:
> > * +348 KB in the build output / snap
> 
> This will directly translate to a bigger snap size and the snap is downloaded by the users on each install or version update. So I think that we should maintain a list of used icons to avoid sending unused icons around.

I do not think it is easy to maintain a list of all used icons. We currently have a list of resource icons but a lot of components use the Icon component directly. Also, how big is the snap, does 348 KB make a real difference?

edlerd:
> > > * +348 KB in the build output / snap
> > 
> > 
> > This will directly translate to a bigger snap size and the snap is downloaded by the users on each install or version update. So I think that we should maintain a list of used icons to avoid sending unused icons around.
> 
> I do not think it is easy to maintain a list of all used icons. We currently have a list of resource icons but a lot of components use the Icon component directly. Also, how big is the snap, does 348 KB make a real difference?

We currently do maintain such a list in [styles.scss](https://github.com/canonical/lxd-ui/blob/main/src/sass/styles.scss#L8-L77). Arguably now it is safer, as any new icon will not work in development until included in the list. With the new approach, things would work in development but fail in production, so it's a bit more brittle.

On the size of the snap, maybe @simondeziel has thoughts on it, as he maintained that previously.

Edit: FYI @advl

simondeziel:
> > > > * +348 KB in the build output / snap
> > > 
> > > 
> > > This will directly translate to a bigger snap size and the snap is downloaded by the users on each install or version update. So I think that we should maintain a list of used icons to avoid sending unused icons around.
> > 
> > 
> > I do not think it is easy to maintain a list of all used icons. We currently have a list of resource icons but a lot of components use the Icon component directly. Also, how big is the snap, does 348 KB make a real difference?
> 
> We currently do maintain such a list in [styles.scss](https://github.com/canonical/lxd-ui/blob/main/src/sass/styles.scss#L8-L77). Arguably now it is safer, as any new icon will not work in development until included in the list. With the new approach, things would work in development but fail in production, so it's a bit more brittle.

If only used icons can be bundled in the snap that's ideal, otherwise, 348 KB isn't that big of a deal but I really appreciate the effort to not inflate the snap needlessly!

tests/scripts/copy-ds-icons.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+ICONS_DIR="$ROOT/node_modules/@canonical/ds-assets/icons"
+DEST_DIR="$ROOT/public/icons"
```
edlerd: I think this will fail to copy them for the snap build. The snap takes the contents from the `build/` directory after running `yarn build`.

See [LXD build steps for the snap](https://github.com/canonical/lxd/blob/eae9c8d094bb542e4be01893f76e42b5132bc702/snap/snapcraft.yaml#L1714-L1717).

For the dev server, I think we don't need to copy them. We could proxy directly to the folder in `node_modules` from the vite config.

src/components/ResourceIcon.tsx
```diff
@@ -75,8 +87,22 @@ interface Props {
...
+  type in dsResourceIcons;
+
 const ResourceIcon: FC<Props> = ({ type, className }) => {
-  return <Icon name={resourceIcons[type]} className={className} />;
+  if (isDsResourceIcon(type)) {
+    return (
+      <DsIcon
+        icon={dsResourceIcons[type]}
+        className={className}
+        rootPath={`${ROOT_PATH}/ui/icons`}
```
edlerd:
```suggestion
        rootPath={`${ROOT_PATH}/ui/assets/icons`}
```
Let's load icons from the existing `assets` folder.

package.json
```diff
@@ -8,8 +8,9 @@
     "check-circular-deps": "npx depcruise src --max-depth 7",
     "clean": "rm -rf node_modules yarn-error.log *.log build/ .jekyll-metadata .bundle playwright-report test-results haproxy-local.cfg coverage",
     "build-html": "cp build/ui/index.html build/index.html",
+    "build-icons": "rm -rf build/ui/icons && cp -r node_modules/@canonical/ds-assets/icons build/ui/icons",
```
edlerd:
```suggestion
    "build-icons": "rm -rf build/ui/assets/icons && cp -r node_modules/@canonical/ds-assets/icons build/ui/assets/icons",
```

vite.config.ts
```diff
@@ -42,6 +42,16 @@ export default defineConfig({
     strictPort: true,
     hmr: process.env.CI ? false : undefined,
     proxy: {
+      "^/ui/icons/": {
```
edlerd:
```suggestion
      "^/ui/assets/icons/": {
```

edlerd (commented):
Thanks for the iteration, the approach is now sound and great. Small suggestion on serving the icons from the assets folder below.

vite.config.ts
edlerd: This clashes with the proxy rule below. We need to check if the priority of the rules is correctly applied. Or make the other rule not match `/ui/assets` globally, but ignore the nested icons one.

omarelkashef: I do not see how the priority is resolved in the vite official docs, but AI responses say that it is in the order of declaration, so we should be good.

src/components/ResourceIcon.tsx
```diff
@@ -75,8 +87,22 @@ interface Props {
...
+
 const ResourceIcon: FC<Props> = ({ type, className }) => {
-  return <Icon name={resourceIcons[type]} className={className} />;
+  if (isDsResourceIcon(type)) {
+    return (
+      <DsIcon
+        icon={dsResourceIcons[type]}
+        className={className}
+        rootPath={`${ROOT_PATH}/ui/assets/icons`}
+      />
```
edlerd, line 100: Sorry to drag this PR forever; just realizing now: This should probably be a component so we don't have to define the root path every time we use the icon component. But this can be a follow up as soon as we have a 2nd use case for the `DsIcon`.

edlerd (approved):
LGTM, thanks for driving this and the surrounding discussions. One thought for a follow-up below.

src/components/ResourceIcon.tsx
omarelkashef: Yes, I will create a ticket for it.

PR #2176, opened 2026-08-05
feat(OperationList): handle child operations

kimanhou:
## Done

- In OperationList, display child operations

## QA

1. Run the LXD-UI:
    - Go to Operations
    - To generate a bulk operation, you can stop all instances of the current project by running the command line `lxc stop --all` (or if you want to specify a project, run `lxc stop --all --project <project_name>`)
    - A bulk operation should have a chip with the count of child operations and a count for any statuses other than success
    - Click on the chevron or the chip to expand: extra rows for child operations should be displayed

## Screenshots

image

### DIFF
--- a/src/context/operationsProvider.tsx
+++ b/src/context/operationsProvider.tsx
@@ -143,3 +143,30 @@ export default OperationsProvider;
 export const useOperations = () => {
   return useContext(OperationsContext);
 };
+
+export const useOperationsWithChildren = () => {
+  const { isAuthenticated } = useAuth();
+
+  const {
+    data: operationList,
+    error,
+    isLoading,
+    isFetching,
+  } = useQuery({
+    queryKey: [queryKeys.operations, "with-children"],
+    queryFn: async () => fetchOperations(null, "2"),
+    enabled: isAuthenticated,
+  });
+
+  const failure = operationList?.failure ?? [];
+  const running = operationList?.running ?? [];
+  const success = operationList?.success ?? [];
+  const operations = [...failure, ...running, ...success];
+
+  return {
+    operations,
+    error,
+    isLoading,
+    isFetching,
+  };
+};
--- a/src/pages/operations/ChildOperationTrigger.tsx
+++ b/src/pages/operations/ChildOperationTrigger.tsx
@@ -1,0 +1,63 @@
+import type { FC } from "react";
+import { Button, Icon, List } from "@canonical/react-components";
+import type { LxdOperation, LxdOperationStatus } from "types/operation";
+import { pluralize } from "util/helpers";
+import { countByStatus } from "util/operations";
+
+interface Props {
+  operation: LxdOperation;
+  isExpanded: boolean;
+  onToggleExpansion: (id: string) => void;
+}
+
+const ChildOperationTrigger: FC<Props> = ({
+  operation,
+  isExpanded,
+  onToggleExpansion,
+}) => {
+  const childOperations = operation.children ?? [];
+  const childCount = childOperations.length || 0;
+
+  if (childCount === 0) {
+    return null;
+  }
+
+  const statusCount = countByStatus(childOperations);
+  const childStatusItems = (
+    ["Running", "Failure", "Cancelled"] as LxdOperationStatus[]
+  )
+    .map((status) => {
+      const count = statusCount[status] ?? 0;
+      return count > 0 ? `${count} ${status.toLowerCase()}` : "";
+    })
+    .filter(Boolean);
+
+  return (
+    <div className="child-operations-trigger">
+      <Button
+        hasIcon
+        dense
+        appearance="base"
+        onClick={() => {
+          onToggleExpansion(operation.id);
+        }}
+        aria-expanded={isExpanded}
+        className="bulk-toggle-btn"
+      >
+        <span className="p-chip is-dense u-no-margin">
+          <Icon name={isExpanded ? "chevron-up" : "chevron-down"} />
+          <List
+            middot
+            items={[
+              `${childCount} ${pluralize("child operation", childCount)}`,
+              ...childStatusItems,
+            ]}
+            className="u-no-margin--bottom"
+          />
+        </span>
+      </Button>
+    </div>
+  );
+};
+
+export default ChildOperationTrigger;
--- a/src/pages/operations/OperationList.tsx
+++ b/src/pages/operations/OperationList.tsx
@@ -1,4 +1,4 @@
-import { useState, type FC } from "react";
+import { useState, type FC, type ReactNode } from "react";
 import {
   EmptyState,
   Icon,
@@ -11,42 +11,161 @@ import {
   Spinner,
   CustomLayout,
 } from "@canonical/react-components";
-import CancelOperationBtn from "pages/operations/actions/CancelOperationBtn";
-import { isoTimeToString, nonBreakingSpaces } from "util/helpers";
-import type { LxdOperationStatus } from "types/operation";
-import OperationInstanceName from "pages/operations/OperationInstanceName";
+import { useOperationsWithChildren } from "context/operationsProvider";
 import NotificationRow from "components/NotificationRow";
-import { getInstanceName, getProjectName } from "util/operations";
-import { useOperations } from "context/operationsProvider";
-import RefreshOperationsBtn from "pages/operations/actions/RefreshOperationsBtn";
-import useSortTableData from "util/useSortTableData";
 import PageHeader from "components/PageHeader";
+import ChildOperationTrigger from "pages/operations/ChildOperationTrigger";
+import OperationDescription from "pages/operations/OperationDescription";
+import OperationTimeDates from "pages/operations/OperationTimeDates";
+import CancelOperationBtn from "pages/operations/actions/CancelOperationBtn";
+import RefreshOperationsBtn from "pages/operations/actions/RefreshOperationsBtn";
+import type { LxdOperation } from "types/operation";
+import {
+  getIconNameForStatus,
+  getInstanceName,
+  getProjectName,
+} from "util/operations";
+import useSortTableData from "util/useSortTableData";
+
+const renderParentFirstColumnContent = ({
+  operation,
+  isExpanded,
+  onToggleExpansion,
+}: {
+  operation: LxdOperation;
+  isExpanded: boolean;
+  onToggleExpansion: (id: string) => void;
+}) => {
+  return (
+    <div className="time-cell-content">
+      <OperationTimeDates operation={operation} />
+      <ChildOperationTrigger
+        operation={operation}
+        isExpanded={isExpanded}
+        onToggleExpansion={onToggleExpansion}
+      />
+    </div>
+  );
+};
+
+const renderChildFirstColumnContent = (operation: LxdOperation) => {
+  return (
+    <>
+      <div className="child-tree-rail">
+        <span className="tree-connector" />
+      </div>
+      <div className="time-cell-content">
+        <OperationTimeDates operation={operation} />
+      </div>
+    </>
+  );
+};
+
+const buildOperationRow = ({
+  operation,
+  className,
+  timeContent,
+  sortData,
+}: {
+  operation: LxdOperation;
+  className: string;
+  timeContent: ReactNode;
+  sortData: Record<string, unknown>;
+}) => {
+  return {
+    key: operation.id,
+    className,
+    columns: [
+      {
+        content: timeContent,
+        role: "cell",
+        "aria-label": "Time",
+        className: "time",
+      },
+      {
+        content: <OperationDescription operation={operation} />,
+        role: "rowheader",
+        "aria-label": "Description",
+        className: "description",
+      },
+      {
+        content: (
+          <>
+            {operation.err && <div>{operation.err}</div>}
+            {Object.entries(operation.metadata ?? {}).map(
+              ([metaKey, value]) => (
+                <div key={metaKey}>
+                  <span title={JSON.stringify(value)}>
+                    {metaKey}: {JSON.stringify(value)}
+                  </span>
+                </div>
+              ),
+            )}
+          </>
+        ),
+        role: "cell",
+        "aria-label": "Details",
+        className: "details",
+      },
+      {
+        content: (
+          <>
+            <Icon
+              name={getIconNameForStatus(operation.status)}
+              className="status-icon"
+            />
+            {operation.status}
+          </>
+        ),
+        role: "cell",
+        "aria-label": "Status",
+        className: "status",
+      },
+      {
+        content: <CancelOperationBtn operation={operation} />,
+        role: "cell",
+        className: "u-align--right cancel",
+        "aria-label": "Actions",
+      },
+    ],
+    sortData,
+  };
+};
 
 const OperationList: FC = () => {
   const notify = useNotify();
-  const { operations, isLoading, error } = useOperations();
+  const { operations, isLoading, error } = useOperationsWithChildren();
+
   const [query, setQuery] = useState<string>("");
+  const [expandedRowIds, setExpandedRowIds] = useState<Set<string>>(new Set());
 
   if (error) {
     notify.failure("Loading operations failed", error);
   }
 
+  const toggleRowExpansion = (id: string) => {
+    setExpandedRowIds((prev) => {
+      const next = new Set(prev);
+      if (next.has(id)) {
+        next.delete(id);
+      } else {
+        next.add(id);
+      }
+      return next;
+    });
+  };
+
   const headers = [
     { content: "Time", className: "time", sortKey: "created_at" },
-    { content: "Action", className: "action", sortKey: "action" },
-    { content: "Info", className: "info" },
+    {
+      content: "Description",
+      className: "description",
+      sortKey: "description",
+    },
+    { content: "Details", className: "details" },
     { content: "Status", className: "status status-header", sortKey: "status" },
     { "aria-label": "Actions", className: "cancel u-align--right" },
   ];
-
-  const getIconNameForStatus = (status: LxdOperationStatus) => {
-    return {
-      Cancelled: "status-failed-small",
-      Failure: "status-failed-small",
-      Running: "status-in-progress-small",
-      Success: "status-succeeded-small",
-    }[status];
-  };
 
   const filteredOperations = operations.filter((operation) => {
     const lowerCaseQuery = query.toLowerCase();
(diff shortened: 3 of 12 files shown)


src/pages/operations/OperationList.tsx
```diff
@@ -10,31 +11,65 @@ import {
   useNotify,
   Spinner,
   CustomLayout,
+  Chip,
 } from "@canonical/react-components";
+import { useQueries } from "@tanstack/react-query";
+import { fetchOperationFull } from "api/instances";
 import CancelOperationBtn from "pages/operations/actions/CancelOperationBtn";
```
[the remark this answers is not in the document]
kimanhou: Fixed

src/pages/operations/OperationList.tsx
```diff
@@ -76,7 +149,29 @@ const OperationList: FC = () => {
                 Last update:{" "}
                 {nonBreakingSpaces(isoTimeToString(operation.updated_at))}
               </div>
-            </>
+
+              {hasChildOperations && (
+                <div className="child-operations-trigger">
+                  <Button
+                    hasIcon
+                    dense
```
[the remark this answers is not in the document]
kimanhou: I'll wait until design is approved then I'll add e2e tests.

src/pages/operations/OperationList.tsx
```diff
@@ -59,15 +93,57 @@ const OperationList: FC = () => {
     );
   });
 
-  const rows = filteredOperations.map((operation) => {
+  const operationQueries = useQueries({
```
omarelkashef: Should we load the full children operation on page load or if the user expands the row?

src/pages/operations/OperationList.tsx
```diff
@@ -59,15 +93,57 @@ const OperationList: FC = () => {
...
+    queries: filteredOperations.map((operation) => ({
+      queryKey: [queryKeys.operations, operation.id, "children"],
+      queryFn: async () => fetchOperationFull(operation.id),
+      staleTime: 5_000,
+      enabled: (operation.child_count ?? 0) > 0,
```
omarelkashef: What is the difference between operation.child_count and operation.metadata.child_count used in other places?

src/pages/operations/OperationList.tsx
```diff
@@ -59,15 +93,57 @@ const OperationList: FC = () => {
...
+    bulkOperationChildrenById.set(operation.id, query.data?.children ?? []);
+    bulkOperationIsLoadingById.set(operation.id, query.isLoading);
+  });
+
+  const rows = filteredOperations.flatMap((operation) => {
     const projectName = getProjectName(operation);
-    return {
+    const childOperations = bulkOperationChildrenById.get(operation.id) ?? [];
+    const metadataChildCount = Number(operation.metadata?.child_count ?? 0);
+    const resourceChildCount = operation.resources?.instance?.length ?? 0;
```
omarelkashef: Are child operations only restricted to instances? If so, should `operation.resources.instance` be used or `operation.resources.instances`?

src/types/operation.d.ts
```diff
@@ -37,3 +38,7 @@ export interface LxdOperationList {
   running?: LxdOperation[];
   success?: LxdOperation[];
 }
+
+export interface LxdOperationFull extends LxdOperation {
```
omarelkashef: Is it necessary to have a new type? What happens if we add the `children` field to the existing `LxdOperation` type?

omarelkashef (commented):
Great feature, thank you! A few comments below

src/pages/operations/OperationList.tsx
kimanhou: I do it on payload to have the count by status "2 operations - 1 failed"

src/pages/operations/OperationList.tsx
kimanhou: Where is `operation.metadat.child_count` used ?

src/pages/operations/OperationList.tsx
omarelkashef: Yeah, I see. That is useful. But I am concerned that in case there are many operations and child operations, loading will slow down considerably.

src/pages/operations/OperationList.tsx
kimanhou: No, child operations are not restricted to instances, they can relate to any resource type, like images or storage volume snapshots.

Both `operation.resources.instance` and `operation.resources.instances` are used in getInstanceName:
```
export const getInstanceName = (operation?: LxdOperation): string => {
  // the url can be one of below formats
  // /1.0/instances/<instance_name>
  // /1.0/instances/<instance_name>?project=<project_name>

  const resources = [
    ...(operation?.resources?.instances || []),
    ...(operation?.resources?.instance || []),
  ];
  const candidates = getOperationEntityUrls(operation, resources);

  return (
    candidates
      ?.filter((item) => item.startsWith("/1.0/instances/"))
      .map((item) => item.split("/")[3])
      .pop()
      ?.split("?")[0] ?? ""
  );
};
```

src/types/operation.d.ts
kimanhou: No, I removed it. I am refactoring OperationList at the moment so haven't pushed yet.

src/pages/operations/OperationList.tsx
kimanhou: Not many operations have children and operations should be short lived so there should never a huge count of operations in the list.

src/pages/operations/OperationList.tsx
omarelkashef: I think it was removed in the latest push. All good now.

src/pages/operations/OperationList.tsx
omarelkashef: But `getInstanceName` only renders for operations related to instances, right?

src/pages/operations/OperationList.tsx
edlerd: There could be many with plans to make operations live for 24 hours to have some form of a log. Can we fetch the operations with a higher recursion factor to get them right away? If not possible, we should request an extension of the endpoint to support fetching all operations, including child details, in a single request.

src/pages/operations/OperationList.tsx
```diff
@@ -136,10 +238,81 @@ const OperationList: FC = () => {
...
+            className: "status",
+          },
+          {
+            content: <CancelOperationBtn operation={child} />,
+            role: "cell",
+            className: "u-align--right cancel",
+          },
+        ],
+        sortData: parentRow.sortData,
+      };
```
edlerd: Can we make this use the same renderer as the parent? Let's extract a row builder function to reuse here and for the parent. I have concerns on duplicating the building of the rows. Especially the logic for creating the description is already drifitng between the two, where it should be the same.

src/pages/operations/OperationList.tsx
```diff
@@ -136,10 +238,81 @@ const OperationList: FC = () => {
...
       },
     };
+
+    if (!isExpanded || !hasChildOperations) {
+      return [parentRow];
+    }
+
+    const childRows = childOperations.map((child, index) => {
+      const childName = getInstanceName(child) || child.id;
+      const isLast = index === childOperations.length - 1;
```
edlerd: there should be a way to identify this is the last of type with css, not introducing any logic here in JS. Something like this:
```
tr.child-row:has(+ tr:not(.child-row)),
tr.child-row:last-child {
```
```suggestion
```

edlerd (commented):
Good start works like a charm. A few thoughts on the structure below.

src/sass/_operation_list.scss
```diff
@@ -17,11 +17,68 @@
...
+      padding: $spv--x-small $spv--x-small;
+
+      &[aria-expanded="true"] {
+        background-color: unset;
+      }
+    }
+  }
+
+  .child-operation-row {
+    background-color: var(--color-background-layer3);
```
edlerd: This highlight doesn't work in the light theme

Image

src/sass/_operation_list.scss
```diff
@@ -17,11 +17,68 @@
...
+    }
+
+    .child-tree-rail {
+      height: 100%;
+      min-height: 2rem;
+      position: relative;
+
+      .tree-connector {
+        border-bottom: 2px solid var(--color-border);
+        border-bottom-left-radius: 4px;
```
edlerd:
```suggestion
```
Let's not use round corners.

src/sass/_operation_list.scss
```diff
@@ -17,11 +17,68 @@
...
+        padding-left: 0;
+      }
+    }
+
+    .child-tree-rail {
+      height: 100%;
+      min-height: 2rem;
+      position: relative;
+
+      .tree-connector {
```
edlerd: The height of the tree should be dynamic on the contents. Like in the example below, we should have a continuous line on the left hand side. With a small horizontal nook at the first line of the new row.

Image

edlerd (commented):
Few more thoughts on the layout

src/pages/operations/OperationList.tsx
kimanhou: Yes, if not an instance then `operation.id` is shown.

src/pages/operations/OperationList.tsx
kimanhou: Indeed recursion=2 returns the child operations alongside the parent operation.

src/context/operationsProvider.tsx
```diff
@@ -140,6 +144,41 @@ const OperationsProvider: FC<Props> = ({ children }) => {
...
+    operations,
+    runningOperations: running,
+    error,
+    isLoading,
+    isFetching,
+    refetchOperations: (options?: RefetchOptions) => {
+      void refetch(options);
+    },
+  };
 };
```
edlerd: I think this is too many things in one. Probably cleaner to define two entry points `useOperations` and `useOperationsWithChildren` to separate the two concerns.

src/pages/operations/actions/RefreshOperationsBtn.tsx
```diff
@@ -1,13 +1,12 @@
...
+import { useIsFetching, useQueryClient } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { ActionButton, Icon } from "@canonical/react-components";
-import { useOperations } from "context/operationsProvider";
 import classnames from "classnames";
 import { useIsScreenBelow } from "context/useIsScreenBelow";
 
 const RefreshOperationsBtn: FC = () => {
-  const { isFetching } = useOperations();
+  const isFetching = useIsFetching({ queryKey: [queryKeys.operations] }) > 0;
```
edlerd: If this consumes `useOperationsWithChildren` it would be simpler.

src/pages/operations/actions/RefreshOperationsBtn.tsx
edlerd: This is now the wrong key to invalidate. Because this button needs to refresh the operations with children for the list page. But maybe it should do a prefix match and refresh all in the status bar and the ones with children for the operation list.

src/pages/operations/OperationList.tsx
```diff
@@ -11,43 +12,202 @@ import {
...
+  return (
+    <div className="child-tree-rail">
+      <span className="tree-connector" />
+    </div>
+  );
+};
+
+const buildOperationRow = ({
+  operation,
+  key,
```
edlerd:
```suggestion
```
We can always use `operation.id` internally to simplify the interface.

src/pages/operations/OperationList.tsx
```diff
@@ -11,43 +12,202 @@ import {
...
+            }}
+            aria-expanded={isExpanded}
+            className="bulk-toggle-btn"
+          >
+            <Icon name={isExpanded ? "chevron-up" : "chevron-down"} />
+            <span className="p-chip is-dense u-no-margin">
+              {childOperationSummary}
+            </span>
+          </Button>
+        </div>
```
edlerd: Let's isolate this into a dedicated `ChildOperationTrigger` component and do all the calculations for the `        childOperationSummary` inside of it.

Also the `hasChildOperations` check and calculation can move inside. If not it can render null.

src/pages/operations/OperationList.tsx
```diff
@@ -11,43 +12,202 @@ import {
...
+  operation,
+  hasChildOperations,
+  isExpanded,
+  childOperationSummary,
+  onToggleRowExpansion,
+}: {
+  operation: LxdOperation;
+  hasChildOperations: boolean;
+  isExpanded: boolean;
+  childOperationSummary: string;
```
edlerd:
```suggestion
  isExpanded: boolean;
```
Those can move into the new component, reducing this interface.

src/pages/operations/OperationList.tsx
```diff
@@ -59,87 +219,65 @@ const OperationList: FC = () => {
...
+      ["Failure", "failed"],
+      ["Cancelled", "cancelled"],
+    ]
+      .map(([status, label]) => {
+        const count = statusCount[status as LxdOperationStatus] ?? 0;
+        return count > 0 ? `${count} ${label}` : "";
+      })
+      .filter(Boolean)
+      .join("\u00A0\u00A0\u2022\u00A0\u00A0");
+    const childOperationSummary = `${childCount} ${pluralize("child operation", childCount)}${childStatusSummary ? `\u00A0\u00A0\u2022\u00A0\u00A0${childStatusSummary}` : ""}`;
```
edlerd: This seems like a hacky way to re-create the inline `<List` component. Can we use it directly instead? Maybe we shouldn't use the chip component here at all. See https://canonical.github.io/react-components/?path=/story/components-list--middot

src/sass/_operation_list.scss
```diff
@@ -17,11 +19,90 @@
...
+      padding: $spv--x-small $spv--x-small;
+
+      &[aria-expanded="true"] {
+        background-color: unset;
+      }
+    }
+  }
+
+  .child-operation-row {
+    background-color: var(--color-background-layer2);
```
edlerd:
```suggestion
    background-color: $colors--theme--background-active;
```
This seems slightly better in terms of contrast for the light theme

src/pages/operations/OperationList.tsx
```diff
@@ -11,43 +12,202 @@ import {
...
+  );
+};
+
+const renderChildFirstColumnContent = () => {
+  return (
+    <div className="child-tree-rail">
+      <span className="tree-connector" />
+    </div>
+  );
+};
```
edlerd: The time is relevant on the children as well, I think we should show it too.

src/pages/operations/OperationList.tsx
```diff
@@ -59,87 +219,65 @@ const OperationList: FC = () => {
...
+        timeContent: renderChildFirstColumnContent(),
+        sortData: parentRowSortData,
+        descriptionRole: "cell",
+      });
+    });
+
+    return [parentRow, ...childRows];
   });
 
   const { rows: sortedRows, updateSort } = useSortTableData({ rows });
```
edlerd: I think we should by default sort by the most recent operations on top, as that is the expected behavior.

edlerd (commented):
Great progress, a few more thoughts on the recent changes.

src/pages/operations/actions/RefreshOperationsBtn.tsx
kimanhou: Can we use `refetchOperations` from `useOperationsWithChildren` ?

edlerd: Let's invalidate with prefix matching. So that also the operations potentially running live in the status bar get updated. As that would be the expectation to have all of them sync with this button.

There should be examples in the code base for the prefix matching using `invalidateQueries` with a `predicate`.

src/pages/operations/OperationList.tsx
kimanhou: I like the UI of the chip. I put the List inside the chip.

src/context/operationsProvider.tsx
```diff
@@ -143,3 +143,37 @@ export default OperationsProvider;
 export const useOperations = () => {
   return useContext(OperationsContext);
 };
+
+export const useOperationsWithChildren = () => {
+  const { isAuthenticated } = useAuth();
+  const contextOperations = useContext(OperationsContext);
```
edlerd:
```suggestion
```
We don't need those here, no?

src/context/operationsProvider.tsx
```diff
@@ -143,3 +143,37 @@ export default OperationsProvider;
...
+    enabled: isAuthenticated,
+  });
+
+  const failure = operationList?.failure ?? [];
+  const running = operationList?.running ?? [];
+  const success = operationList?.success ?? [];
+  const operations = [...failure, ...running, ...success];
+
+  return {
+    ...contextOperations,
```
edlerd:
```suggestion
```

src/context/operationsProvider.tsx
```diff
@@ -143,3 +143,37 @@ export default OperationsProvider;
...
+    operations,
+    runningOperations: running,
+    error,
+    isLoading,
+    isFetching,
+    refetchOperations: (options?: RefetchOptions) => {
+      void refetch(options);
+    },
```
edlerd:
```suggestion
    error,
    isLoading,
```
This can be simpler.

src/pages/operations/ChildOperationTrigger.tsx
```diff
@@ -0,0 +1,63 @@
...
+interface Props {
+  operation: LxdOperation;
+  isExpanded: boolean;
+  onToggleRowExpansion: (id: string) => void;
+}
+
+const ChildOperationTrigger: FC<Props> = ({
+  operation,
+  isExpanded,
+  onToggleRowExpansion,
```
edlerd:
```suggestion
  onToggleExpansion,
```
A bit clearer in this context.

src/pages/operations/OperationList.tsx
```diff
@@ -11,43 +11,189 @@ import {
...
+import { isoTimeToString, nonBreakingSpaces } from "util/helpers";
+import {
+  getIconNameForStatus,
+  getInstanceName,
+  getProjectName,
+} from "util/operations";
 import useSortTableData from "util/useSortTableData";
-import PageHeader from "components/PageHeader";
+
+const renderOperationDescription = (operation: LxdOperation) => {
```
edlerd:
```suggestion
const OperationDescription = (operation: LxdOperation) => {
```
This can be a component in a dedicated file.

src/pages/operations/OperationList.tsx
```diff
@@ -11,43 +11,189 @@ import {
...
+      <div>{operation.description}</div>
+      <OperationInstanceName operation={operation} />
+      <div className="u-text--muted u-truncate" title={projectName}>
+        Project: {projectName}
+      </div>
+    </>
+  );
+};
+
+const renderTimeDates = (operation: LxdOperation) => (
```
edlerd:
```suggestion
const OperationTimeDates = (operation: LxdOperation) => (
```
Should be a component too.

src/pages/operations/OperationList.tsx
```diff
@@ -11,43 +11,189 @@ import {
...
+    columns: [
+      {
+        content: timeContent,
+        role: "cell",
+        "aria-label": "Time",
+        className: "time",
+      },
+      {
+        content: renderOperationDescription(operation),
+        role: descriptionRole,
```
edlerd:
```suggestion
        role: "rowheader",
```
This should always be the row header, no?

src/pages/operations/OperationList.tsx
```diff
@@ -59,90 +205,56 @@ const OperationList: FC = () => {
...
+
+    const parentRowSortData = {
+      created_at: operation.created_at,
+      description: operation.description,
+      status: operation.status,
     };
+
+    const parentRow = buildOperationRow({
+      operation,
+      key: operation.id,
```
edlerd:
```suggestion
```

src/pages/operations/OperationList.tsx
```diff
@@ -59,90 +205,56 @@ const OperationList: FC = () => {
...
+      className: classnames("u-row", {
+        "has-child-operations": hasChildOperations,
+      }),
```
edlerd:
```suggestion
      className: "u-row",
```
seems unused.

src/sass/_operation_list.scss
```diff
@@ -1,4 +1,6 @@
 .operation-list {
```
edlerd, line 1: We need to tweak the list for smaller screens. Currently the table is responsive, and rows become cards, which bamboozles the tree layout.

Image

edlerd (commented):
A few ideas to simplify, and found one problem with responsiveness

src/context/operationsProvider.tsx
kimanhou: But I was using isFetching in `RefreshOperationsBtn`

edlerd: I missed that, we can keep it, sure.

src/sass/_operation_list.scss
kimanhou: Is this acceptable ?

image

tests/operations.spec.ts
```diff
@@ -27,3 +29,45 @@ test("instance operations are recognised on the Operations page", async ({
   await deleteInstance(page, instance);
   await validateOperation(page, `Deleting instance`);
 });
+
+test("bulk stop operation renders and expands child operations", async ({
+  page,
+}) => {
+  const firstInstance = randomInstanceName();
+  const secondInstance = randomInstanceName();
```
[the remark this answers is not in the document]
kimanhou: It is imported on line 5.

src/sass/_operation_list.scss
edlerd: Yeah, I think this is good :+1:

src/pages/operations/ChildOperationTrigger.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onToggleExpansion: (id: string) => void;
+}
+
+const ChildOperationTrigger: FC<Props> = ({
+  operation,
+  isExpanded,
+  onToggleExpansion,
+}) => {
+  const childOperations = operation.children ?? [];
+  const childCount = childOperations.length || operation.child_count || 0;
```
edlerd:
```suggestion
  const childCount = childOperations.length || 0;
```
I don't think we should fall back to the field here. It is an edge case anyhow, and we'd announce child operations that are not available to view. Wdyt?

src/pages/operations/ChildOperationTrigger.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return null;
+  }
+
+  const statusCount = countByStatus(childOperations);
+  const childStatusItems = (
+    ["Running", "Failure", "Cancelled"] as LxdOperationStatus[]
+  )
+    .map((status) => {
+      const count = statusCount[status] ?? 0;
+      return count > 0 ? `${count} ${status.toLowerCase()}` : "";
```
edlerd, line 31: We should use a non breaking space between number and status.

Image

src/pages/operations/ChildOperationTrigger.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <span className="p-chip is-dense u-no-margin">
+          <List
+            middot
+            items={[
+              `${childCount} ${pluralize("child operation", childCount)}`,
+              ...childStatusItems,
+            ]}
+            className="u-no-margin--bottom"
+          />
+        </span>
```
edlerd:
```suggestion
        <span className="p-chip is-dense u-no-margin">
          <Icon
            name={isExpanded ? "chevron-up" : "chevron-down"}
            className="u-no-margin--top"
          />
          <List
            middot
            items={[
              `${childCount} ${pluralize("child operation", childCount)}`,
              ...childStatusItems,
            ]}
            className="u-no-margin--bottom"
          />
        </span>
```
Slightly improving the spacing. Also, I think it is better to put the icon into the chip.

edlerd (commented):
Few ideas on styling the trigger. QA and Code otherwise LGTM.

edlerd (approved):
LGTM, great stuff :)

PR #2178, opened 2026-08-06
test(file explorer): extend file explorer tests to symlinks resolving…

omarelkashef:
… and directory creation

## Done

- extend file explorer tests to symlinks resolving and directory creation WD-37689

## QA

1. Run the LXD-UI:
    - Ensure CI passes

### DIFF
--- a/tests/helpers/instance-file-explorer.ts
+++ b/tests/helpers/instance-file-explorer.ts
@@ -1,9 +1,10 @@
 import type { Page } from "@playwright/test";
 import { expect, test, type LxdVersions } from "../fixtures/lxd-test";
+import { getLxcCmd } from "./auth";
 import { dismissNotification } from "./notification";
-import { assertTextVisible } from "./permissions";
 import { visitInstance } from "./instances";
 import { randomNameSuffix } from "./name";
+import { runCommand } from "./shell";
 
 export const skipIfFileExplorerNotSupported = (lxdVersion: LxdVersions) => {
   test.skip(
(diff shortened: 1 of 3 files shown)


tests/helpers/instance-file-explorer.ts
```diff
@@ -79,6 +90,53 @@ export const assertDirectoryExists = async (
...
+  page.once("dialog", async (dialog) => {
+    await dialog.accept();
+  });
+  await page.getByRole("link", { name: "File Explorer" }).click();
+
+  // Switching tabs drops the `path` search param; navigate back through the original path.
+  for (const segment of pathSegments) {
+    await openDirectory(page, segment);
+  }
+};
```
edlerd: This function is doing a bit too much. I think it should only run the command, and we shouldn't use the terminal for it. Instead we can do:
```suggestion
const runCommandForInstance = async (instance: string, command: string) => {
  runCommand(`${getLxcCmd()} exec ${instance} -- ${command}´);
};
```
The side effects of this function of reading the path and navigating back to it can be removed.

Though I haven't tested the suggestion, but assume it should work fine.

tests/instances.spec.ts
```diff
@@ -553,6 +558,67 @@ test("upload, download, and delete file from file explorer", async ({
...
 
+test("create directory from file explorer", async ({ page, lxdVersion }) => {
+  skipIfFileExplorerNotSupported(lxdVersion);
+
+  await visitAndStartInstance(page, instance);
+  await visitFileExplorer(page, instance);
+  await openDirectory(page, "tmp");
+
+  const directoryName = randomDirectoryName();
+  const directoryName2 = randomDirectoryName();
```
edlerd:
```suggestion
  const directory = randomDirectoryName();
  const directory2 = randomDirectoryName();
```
A bit shorter?

edlerd (commented):
Two small ideas to simplify, otherwise LGTM.

tests/helpers/instance-file-explorer.ts
omarelkashef: True, t makes it simpler. I will test it.

edlerd (approved):
LGTM, thanks for adding the test and simplifying the approach.

PR #2186, opened 2026-08-07
feat(identities) show expiration in identity list

edlerd:
## Done

- feat(identities) show expiration in identity list

## QA

1. Run the LXD-UI:
    - open permissions > identities
    - issue a new token for expiry in 1 second
    - check expiry warning within list

## Screenshots

### DIFF
--- a/src/pages/permissions/IdentityExpiration.tsx
+++ b/src/pages/permissions/IdentityExpiration.tsx
@@ -1,0 +1,26 @@
+import { type FC } from "react";
+import { isBeforeNow, isoTimeToString, UNDEFINED_DATE } from "util/helpers";
+import type { LxdIdentity } from "types/permissions";
+import { Icon, Tooltip } from "@canonical/react-components";
+
+interface Props {
+  identity: LxdIdentity;
+}
+
+const IdentityExpiration: FC<Props> = ({ identity }) => {
+  const expiry = identity.expires_at ?? UNDEFINED_DATE;
+  const isValid = expiry !== UNDEFINED_DATE;
+  const isExpired = isBeforeNow(expiry);
+  const caption = isValid ? isoTimeToString(expiry) : "-";
+
+  return isExpired ? (
+    <Tooltip message="Has expired" position="right">
+      <span>{caption}</span>
+      <Icon name="warning" className="u-margin-left--small" />
+    </Tooltip>
+  ) : (
+    caption
+  );
+};
+
+export default IdentityExpiration;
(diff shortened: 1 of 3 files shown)


src/util/helpers.tsx
```diff
@@ -34,6 +34,10 @@ export const isoTimeToString = (isoTime: string): string => {
   });
 };
 
+export const isBeforeNow = (time: string): boolean => {
+  return new Date(time) < new Date();
```
omarelkashef: Maybe it is worth to add a guard here.

src/pages/permissions/IdentityExpiration.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const IdentityExpiration: FC<Props> = ({ identity }) => {
+  const isExpired = identity.expires_at
+    ? isBeforeNow(identity.expires_at)
+    : false;
+  const caption = identity.expires_at
+    ? isoTimeToString(identity.expires_at)
+    : "-";
+
+  return isExpired ? (
```
omarelkashef, line 16: It looks a little off that the tooltip is on the oppsite side of the warning icon. Maybe the icon should come after the text and then the tooltip would be next to the icon. Also, I think we need a little more space between the icon and the expiration.
Image

omarelkashef (commented):
Thank you for the PR! A couple of comments below.

src/pages/permissions/IdentityExpiration.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface Props {
+  identity: LxdIdentity;
+}
+
+const IdentityExpiration: FC<Props> = ({ identity }) => {
+  const isExpired = identity.expires_at
+    ? isBeforeNow(identity.expires_at)
+    : false;
+  const caption = identity.expires_at
```
omarelkashef: I think the caption should be more granular. For TLS, it should say "Never" as opposed to "-" which is for bearer identities that do not have a token.

edlerd: A display value of "never" could be confusing. Because all TLS certificates have an expiration, just the LXD identity does not know which date is set on the cert.

kimanhou: How do we handle identities created before `expires_at` was added to the Identity object ? They don't have an expiry on the object but we can't display "Never" and we can't really display "we don't know, sorry" 😅

At the moment, the expiry is shown as "-", which is the least worst imo.

omarelkashef (dismissed):
LGTM, but needs a rebase!

src/pages/permissions/IdentityExpiration.tsx
omarelkashef: Yeah I agree "-" is the best solution for that edge case.

omarelkashef (approved):
Looks good, thank you

PR #2180, opened 2026-08-06
test(Identities): Create e2e test to create and delete identities

kimanhou:
## Done

- test(Identities): Create e2e test to create and delete identities
- Edit will come after #2150 is merged

## QA

1. Run the LXD-UI:
    - On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @Kxiru or @edlerd for access.
    - With a local copy of this branch, [build and run as described in the docs](https://github.com/canonical/lxd-ui/blob/main/CONTRIBUTING.md#setting-up-for-development).
2. Perform the following QA steps:
    - Trust CI

PR #2188, opened 2026-08-10
fix(bulk operations): improve e2e test to make it more robust

kimanhou:
## Done

- bulk operations: make e2e test more robust

## QA

1. Run the LXD-UI:
    - Trust CI.

## Screenshots

image

### DIFF
--- a/tests/operations.spec.ts
+++ b/tests/operations.spec.ts
@@ -3,6 +3,7 @@ import { getLxcCmd } from "./helpers/auth";
 import {
   createInstance,
   deleteInstance,
+  getInstanceCount,
   randomInstanceName,
   visitAndStartInstance,
   visitAndStopInstance,
@@ -50,28 +51,30 @@ test("bulk stop operation renders and expands child operations", async ({
   await visitAndStartInstance(page, secondInstance);
 
   const lxc = getLxcCmd();
+  const instanceCount = getInstanceCount("default");
   runCommand(`${lxc} stop --all --project default`);
 
   await visitOperations(page);
 
-  const childOperationsChip = page
-    .getByRole("button", { name: /2 child operations/i })
-    .first();
+  const childOperationsChip = page.getByRole("button", {
+    name: `${instanceCount} child operations`,
+  });
   await expect(childOperationsChip).toBeVisible();
-
   await childOperationsChip.click();
 
   const childRows = page.locator(
     "#operation-table tbody tr.child-operation-row",
   );
-  await expect(childRows).toHaveCount(2);
+  await expect(childRows).toHaveCount(instanceCount);
 
   const firstChildRow = childRows.filter({ hasText: firstInstance });
+  await expect(firstChildRow).toBeVisible();
   await expect(firstChildRow).toContainText("Stopping instance");
   await expect(firstChildRow).toContainText("Project: default");
   await expect(firstChildRow).toContainText("Success");
 
   const secondChildRow = childRows.filter({ hasText: secondInstance });
+  await expect(secondChildRow).toBeVisible();
   await expect(secondChildRow).toContainText("Stopping instance");
   await expect(secondChildRow).toContainText("Project: default");
   await expect(secondChildRow).toContainText("Success");
--- a/src/pages/operations/ChildOperationTrigger.tsx
+++ b/src/pages/operations/ChildOperationTrigger.tsx
@@ -49,7 +49,7 @@ const ChildOperationTrigger: FC<Props> = ({
           <List
             middot
             items={[
-              `${childCount} ${pluralize("child operation", childCount)}`,
+              `${childCount} child ${pluralize("operation", childCount)}`,
               ...childStatusItems,
             ]}
             className="u-no-margin--bottom"
--- a/tests/helpers/instances.ts
+++ b/tests/helpers/instances.ts
@@ -5,6 +5,8 @@ import { expect } from "../fixtures/lxd-test";
 import { visitLocalImages } from "./images";
 import { searchEntityListPage } from "./search";
 import { dismissNotification } from "./notification";
+import { runCommand } from "./shell";
+import { getLxcCmd } from "./auth";
 
 const DEFAULT_IMAGE = "alpine/3.23/cloud";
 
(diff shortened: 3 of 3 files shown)


src/util/permissionIdentities.tsx
```diff
@@ -251,17 +251,17 @@ export const IDENTITY_TYPE_HELP_TEXT: Record<
...
 > = {
   [IDENTITY_TYPE.TLS]: {
-    title: "TLS Certificate",
+    title: "Client certificate",
     description:
       "A certificate-based identity for long-lived client authentication.",
   },
   [IDENTITY_TYPE.BEARER_CLIENT]: {
-    title: "Bearer token (Main API)",
+    title: "Client token bearer",
```
edlerd, line 259: Should this be the other way around?
```suggestion
    title: "Client bearer token",
```

src/util/operations.tsx
```diff
@@ -165,3 +166,7 @@ export const getIconNameForStatus = (status: LxdOperationStatus) => {
     Success: "status-succeeded-small",
   }[status];
 };
+
+export const getChildOperationsLabel = (count: number): string => {
+  return `${count} child ${pluralize("operation", count)}`;
+};
```
edlerd, line 172: There is just one use case in production code. The test should not use the same code as production. Otherwise it doesn't really test. Shall we remove the function and inline this again?

tests/operations.spec.ts
```diff
@@ -50,28 +60,31 @@ test("bulk stop operation renders and expands child operations", async ({
...
   runCommand(`${lxc} stop --all --project default`);
 
   await visitOperations(page);
 
-  const childOperationsChip = page
-    .getByRole("button", { name: /2 child operations/i })
-    .first();
-  await expect(childOperationsChip).toBeVisible();
+  const childOperationsChip = page.getByRole("button", {
+    name: new RegExp(`^${getChildOperationsLabel(runningInstanceCount)}$`, "i"),
```
edlerd, line 69: We don't need to call `getChildOperationsLabel` here, because there are always at least two instances here as per the code above preparing the environment.

tests/operations.spec.ts
edlerd, line 1: Can we extract the fix to the operation test into a dedicated PR to not mix it into the type changes?

edlerd (commented):
There is an additional type confusion from `src/pages/permissions/panels/CreateIdentityPanel.tsx:209` and `src/pages/permissions/IdentityTokenModal.tsx:47` where different types are in use where they should be the same. It is confusing that the former is hardcoding the icons, where the latter is first transforming the identity type to a resource type, which is then transformed into an icon name.

src/util/permissionIdentities.tsx
edlerd: I see now this is exactly the type as reported by the API, so it seems correct as is.

tests/operations.spec.ts
kimanhou: I put the change on permissions identities type names in https://github.com/canonical/lxd-ui/pull/2180
This PR is only related to the e2e test of bulk operation, despite the branch name.

tests/operations.spec.ts
```diff
@@ -50,28 +60,34 @@ test("bulk stop operation renders and expands child operations", async ({
...
 
   await visitOperations(page);
 
-  const childOperationsChip = page
-    .getByRole("button", { name: /2 child operations/i })
-    .first();
-  await expect(childOperationsChip).toBeVisible();
+  const childOperationsChip = page.getByRole("button", {
+    name: new RegExp(
+      `^${runningInstanceCount} child ${pluralize("operation", runningInstanceCount)}$`,
```
edlerd:
```suggestion
      `^${runningInstanceCount} child operations`,
```
There are always at least two, no?

tests/operations.spec.ts
```diff
@@ -50,28 +60,34 @@ test("bulk stop operation renders and expands child operations", async ({
...
+  await childOperationsChip.first().click();
 
   const childRows = page.locator(
     "#operation-table tbody tr.child-operation-row",
   );
-  await expect(childRows).toHaveCount(2);
+  await expect(childRows).toHaveCount(runningInstanceCount);
 
-  const firstChildRow = childRows.filter({ hasText: firstInstance });
+  const firstChildRow = childRows.filter({ hasText: firstInstance }).first();
```
edlerd:
```suggestion
  const firstChildRow = childRows.filter({ hasText: firstInstance });
```
There should be only one occurrence with the instance name on it anyhow, no?

edlerd (dismissed):
LGTM, two non-blocking nitpicks below.

tests/operations.spec.ts
```diff
@@ -14,6 +14,15 @@ import {
 } from "./helpers/operations";
 import { runCommand } from "./helpers/shell";
 
+const getRunningInstanceCount = (project: string): number => {
```
edlerd: Can we move this to a helper file for instances? Then this file is for tests only.

tests/operations.spec.ts
```diff
@@ -50,28 +59,31 @@ test("bulk stop operation renders and expands child operations", async ({
...
   runCommand(`${lxc} stop --all --project default`);
 
   await visitOperations(page);
 
-  const childOperationsChip = page
-    .getByRole("button", { name: /2 child operations/i })
-    .first();
-  await expect(childOperationsChip).toBeVisible();
+  const childOperationsChip = page.getByRole("button", {
+    name: new RegExp(`^${runningInstanceCount} child operations$`, "i"),
```
edlerd:
```suggestion
    name: `${runningInstanceCount} child operations`),
```
I think we don't need the regexp here.

tests/operations.spec.ts
```diff
@@ -50,28 +59,31 @@ test("bulk stop operation renders and expands child operations", async ({
...
+  });
+  await expect(childOperationsChip.first()).toBeVisible();
```
edlerd:
```suggestion
  await expect(childOperationsChip).toBeVisible();
```
This is a bit imprecise. I think we need to find the correct wrapping operation row of the chip. Or actually, first might be fine as we sort by recency, and it will pick the right one.

edlerd (dismissed):
LGTM, few non-blocking thoughts on the changes.

edlerd (dismissed):
LGTM

edlerd (approved):

PR #2189, opened 2026-08-10
unidirectional cluster link create, display and edit

edlerd:
## Done

- unidirectional cluster link create, display and edit

Fixes WD-38124

## QA

1. Run the LXD-UI:
    - go to clustering > links and create cluster link
      - create a unidirectional link
      - create a bidirectional link
      - edit a unidirectional link
      - edit a bidirectional link
      - To create the token for a unidirectional link, create another VM in the same network as your LXD backend and run this command to generate the token: `lxc auth identity create cluster-link/foo`

## Screenshots

Screenshot from 2026-08-10 12-30-24

Screenshot from 2026-08-10 12-30-41

Screenshot from 2026-08-10 12-30-54

image

image

image

### DIFF
--- a/src/pages/cluster/ClusterLinkForm.tsx
+++ b/src/pages/cluster/ClusterLinkForm.tsx
@@ -1,25 +1,19 @@
-import { Form, Input, Label, RadioInput } from "@canonical/react-components";
+import { Form, Input, Label, OutputField } from "@canonical/react-components";
 import type { FC } from "react";
 import GroupSelection from "pages/permissions/panels/GroupSelection";
 import { useAuthGroups } from "context/useAuthGroups";
 import type { FormikProps } from "formik/dist/types";
-
-export interface ClusterLinkFormValues {
-  name: string;
-  description?: string;
-  token?: string;
-  tokenType?: "generate" | "consume";
-  authGroups: string[];
-  isCreating: boolean;
-  initialAuthGroups?: string[];
-}
+import ClusterLinkTokenInput from "pages/cluster/ClusterLinkTokenInput";
+import type { ClusterLinkFormValues } from "types/forms/clusterLink";
 
 interface Props {
   formik: FormikProps<ClusterLinkFormValues>;
 }
 
 const ClusterLinkForm: FC<Props> = ({ formik }) => {
-  const { data: authGroups = [] } = useAuthGroups();
+  const { data: authGroups = [] } = useAuthGroups(
+    formik.values.type === "bidirectional",
+  );
 
   const selectedGroups = new Set(formik.values.authGroups);
   const initial = new Set(formik.values.initialAuthGroups);
@@ -29,19 +23,33 @@ const ClusterLinkForm: FC<Props> = ({ formik }) => {
   const hasName = formik.values.name !== "";
 
   return (
-    <Form onSubmit={formik.handleSubmit}>
+    <Form onSubmit={formik.handleSubmit} className="cluster-link-form">
       {/* hidden submit to enable enter key in inputs */}
       <Input type="submit" hidden value="Hidden input" />
       {formik.values.isCreating && (
-        <Input
-          {...formik.getFieldProps("name")}
-          type="text"
-          label="Name"
-          placeholder="Enter name"
-          required
-          autoFocus
-          error={formik.touched.name ? formik.errors.name : null}
-        />
+        <>
+          <OutputField
+            id="type"
+            label="Cluster link type"
+            value={formik.values.type}
+          />
+          <ClusterLinkTokenInput formik={formik} />
+          <Input
+            {...formik.getFieldProps("name")}
+            type="text"
+            label="Cluster link name"
+            placeholder="Enter name"
+            required
+            disabled={
+              (formik.values.token === "" &&
+                (formik.values.type === "unidirectional" ||
+                  formik.values.tokenType === "consume")) ||
+              (formik.values.type === "bidirectional" &&
+                !formik.values.tokenType)
+            }
+            error={formik.touched.name ? formik.errors.name : null}
+          />
+        </>
       )}
       <Input
         {...formik.getFieldProps("description")}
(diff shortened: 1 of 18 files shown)


src/pages/cluster/ClusterLinkForm.tsx
```diff
@@ -12,36 +13,57 @@ export interface ClusterLinkFormValues {
...
+            <div title={formik.values.token} className="token-output">
+              <OutputField
+                id="token"
+                label="Token"
+                value={formik.values.token}
+              />
+            </div>
+          )}
+          {!showToken && (
+            <OutputField id="token" label="Token" value="<generate>" />
```
omarelkashef, line 55: "\<generate\>" looks more like an error or something that was left by mistake more than a placeholder? What about not displaying the field at all?

src/util/usePanelParams.tsx
```diff
@@ -203,8 +203,10 @@ const usePanelParams = (): PanelHelper => {
       setPanelParams(panels.editClusterGroups, { group });
     },
 
-    openEditClusterLink: (identity) => {
-      setPanelParams(panels.editClusterLink, { identity });
+    openEditClusterLink: (clusterLink) => {
```
omarelkashef, line 206: Is there a reason this looks different from other ways to set panel params?

omarelkashef (commented):
Great PR, the new panel looks really good! A few comments about the interactions below.

The radio button for token generate/consume is cut off.
Image


It took me some time to understand the purpose of the Notification. Maybe a muted help text will look more in place. wdyt? 
Image



I wonder if the instructions to create a token are easily maintainable. The page name can change. 

Image


How about decreasing the width of the modified status cell since it does not all of this and giving description more width?

Image


The rest of the edit panels show the name as its own output field not part of the title

Image

The interaction when there is an error creating a unidirectional link makes the panel hard to use.

https://github.com/user-attachments/assets/b2d09720-fa47-40d3-ae36-9c5b795e2951

src/pages/cluster/ClusterLinkTokenInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+    />
+  );
+
+  if (formik.values.type === "unidirectional") {
+    return <>{tokenInput}</>;
+  } else {
+    return (
+      <>
+        <div className="u-sv1">
+          <Select
```
omarelkashef, line 71: What if we make this a radio button? I think it looks better. At least, we should fix the fact that when it first renders, it has no default but the rest of the form shows fields as if the consume option was chosen

Image

src/pages/cluster/ClusterLinkTokenInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const ClusterLinkTokenInput: FC<Props> = ({ formik }) => {
+  const getTokenHelp = () => {
+    if (formik.values.type === "unidirectional") {
+      return (
+        <>
+          Paste the token you have generated on the target cluster.
+          <br />
+          To create one, open the LXD UI on the <b>target cluster</b>, then go
```
omarelkashef, line 17: I think when the UI cluster links docs are added to lxd docs, we should reference because this might go out of sync.

omarelkashef (commented):
Thank you for the iteration. Two comments below.

src/pages/cluster/modals/CreateClusterLinkModal.tsx
```diff
@@ -28,7 +29,8 @@ const CreateClusterLinkModal: FC<Props> = ({ onClose, token, linkName }) => {
       className="create-cluster-link"
       title={
         <>
-          Cluster link <ClusterLinkRichChip clusterLink={linkName} /> created
+          Cluster link <ResourceLabel type="cluster-link" value={linkName} />{" "}
```
omarelkashef, line 32: What is the reason for this change?

edlerd: The rich chip doesn't scale well in context of the heading, it looks too small.

src/pages/cluster/ClusterLinkTokenInput.tsx
edlerd: I want to give clear instructions right away to have fewer hops. Linking to the docs here would need a user to change focus and it seems a bit more complicated.

src/pages/cluster/ClusterLinkTokenInput.tsx
edlerd: These used to be radio buttons. The problem with radio input is that we should have a default and I want to force the user to make a conscious choice. Hence the switch away from radio to select with the prominent "select an option" default and all other elements are disabled.

As per the help text for token, I made it empty initially, that is better, indeed.

omarelkashef (commented):
One last thing and this will be good to merge after the tests are fixed. When switching from consume to generate, the token input should be cleared.

Image

src/pages/cluster/panels/EditClusterLinkPanel.tsx
```diff
@@ -89,8 +102,8 @@ const EditClusterLinkPanel: FC<Props> = ({ identity }) => {
     <SidePanel>
       <SidePanel.Header>
         <SidePanel.HeaderTitle className="u-truncate">
-          <span title={`Edit cluster link ${clusterLink?.name}`}>
-            Edit cluster link {clusterLink?.name}
+          <span title={`Edit cluster link ${clusterLink.name}`}>
+            Edit cluster link {clusterLink.name}
```
omarelkashef, line 106: Should the name be its own outputfield?

omarelkashef (approved):
Thank you for the many iterations. QA and code looks good. Only one non-blocking suggestion.

src/pages/cluster/panels/EditClusterLinkPanel.tsx
edlerd: I think having it in the header is the preferred way. The extra output field for the name is more design clutter and harder to digest, in my opinion.

PR #2193, opened 2026-08-10
feat: [WD-37337] Adding NVME to Powerstore mode options

Kxiru:
## Done

- Added nvme/tcp to Powerstore configuration list.
- Conditionally adds the above if the `storage_driver_powerstore_nvme` extension exists.

## QA

1. Run the LXD-UI:
    - In a Powerstore compatible backend (or backend with an added line to test 
## Screenshots

N/A

### DIFF
--- a/src/context/useSupportedFeatures.tsx
+++ b/src/context/useSupportedFeatures.tsx
@@ -65,5 +65,8 @@ export const useSupportedFeatures = () => {
     hasLoadBalancerHealthChecks: apiExtensions.has(
       "network_load_balancer_pool_health_checks",
     ),
+    hasStorageDriverPowerstoreNvme: apiExtensions.has(
+      "storage_driver_powerstore_nvme",
+    ),
   };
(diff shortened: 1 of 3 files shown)


omarelkashef (commented):
Code looks good. Only one comment: the inherited value from the backend is `nvme/tcp` but the the option in the dropdown is `NVMe over TCP`. Should we make the option match what the backend calls it to avoid confusion?
image

edlerd (approved):
LGTM, thanks :)

Kxiru:
> Code looks good. Only one comment: the inherited value from the backend is `nvme/tcp` but the the option in the dropdown is `NVMe over TCP`. Should we make the option match what the backend calls it to avoid confusion? 

Sorry, @omarelkashef , I didn't see this. The copy has been confirmed between myself and @musicdin so its all good to go. It's also more human readable. Thanks for being thorough though!

PR #2194, opened 2026-08-11
feat: [WD-37832] ExplanationTooltip for entity form headings

Kxiru:
## Done
V1: Each ET has a description
V2: Introduces 2nd round of ETs, ETs on creation & edit forms, and copy amendments.
V3. Introduces isConfigVariant. Introduces titles for each ET link. Some ETs have creation form variants and alternative links.
V4: Unified ETs comprised of a prefix and suffix. Some ETs have alternative links. Logic change.

Non-exhaustive list of all the ET header placements, as well as their config variant counterparts. (Exhaustive list in [copydoc](https://docs.google.com/document/d/1vDmemn4eQm0Zr3szBRzGtWdZqGNayqHkUmtC8oZ0jNA/edit?tab=t.0))
- [x] Profile creation & Edit form
- [x] Network creation & Edit form (+Forwards, +Load Balancers, +ACLs, +Local Peering)
- [x] Storage Pool creation & Edit form
- [x] Storage Volume creation & Edit form
- [x] Storage Buckets creation & Edit form
- [x] Custom ISO creation form
- [x] Groups / permissions creation & Edit forms
- [x] Cluster Links Edit form
- [x] Add titles of docs in place of the “Learn More” link title.
- [x] Reimplement new creation flow explanations.
- [x] ET tooltip arrow defaulting to the left of the ET icon on 9 pages.  ProjectList, ProfileList, NetworkList, StoragePools, StorageBuckets, Server page, Replicator list, Warnings page, Settings page

## QA

1. Run the LXD-UI:
    - Observe the changes above through the demo server, noting changes to creation form titles and fields.

## Screenshots

imageimage


src/pages/instances/CreateInstance.tsx
```diff
@@ -448,7 +449,14 @@ const CreateInstance: FC = () => {
     !formik.isValid || !formik.values.image || diskError || networkError;
 
   return (
-    <BaseLayout title="Create an instance" contentClassName="create-instance">
+    <BaseLayout
+      title={
+        <InstanceExplanationTooltip>
+          Create an instance
+        </InstanceExplanationTooltip>
```
edlerd: Reviewing it now in the context, I see the copy of the tooltip not being up for the forms as they are. The copy works for the list page, as it explains what the entity is, but I am not sure it works on the create/edit form in exactly the same form.

Such as in this example, you are creating an instance, so "Manage and monitor virtual machines and containers." does not work well. Here we might just need the link to the docs or evolve the copy to be helpful in the context of the form.

This applies to all the other forms as well. Some entities might need an explanation.

Image

src/pages/profiles/CreateProfile.tsx
```diff
@@ -176,7 +177,12 @@ const CreateProfile: FC = () => {
   }
 
   return (
-    <BaseLayout title="Create a profile" contentClassName="create-profile">
+    <BaseLayout
+      title={
+        <ProfileExplanationTooltip>Create a profile</ProfileExplanationTooltip>
```
edlerd, line 182: Here we might way "A profile is a configuration template for instances." instead of "Create and manage configuration templates for instances.".

Image

edlerd (commented):
I fear the copy is not 1:1 up for the forms as it is tailored to the list pages.

src/pages/instances/CreateInstance.tsx
Kxiru: Hmm, okay. We can definitely do custom lines in some places... But what exactly needs to be explained in the create sentence? It's not "use this form to create an instance" with a link to more instance info, it's supposed to provide additional context for what an instance actually is.

src/pages/profiles/CreateProfile.tsx
Kxiru: Oh I see, so rather than explaining what an entity is **for** (or what it does), we explain what it **is.**

Kxiru: So custom ETs again for all creation/edit forms? :grimacing:

src/pages/instances/CreateInstance.tsx
edlerd: Good question, I was wondering myself and didn't find a good suggestion. It could be something like "configure your instance". I think a better link than the current one for this exact case of instance form is https://lxd-ui-2194.demos.haus/documentation/explanation/instance_config/

src/pages/profiles/CreateProfile.tsx
edlerd: I think it depends on each individual case. Some copies and links might be fit for forms and list page.

For this one there is a better link target as well: https://lxd-ui-2194.demos.haus/documentation/reference/projects/

Maybe start with `ProfileFormExplanationTooltip` and `InstanceFormExplanationTooltip` and check the other cases.

src/sass/_explanation_tooltip.scss
```diff
@@ -4,6 +4,14 @@
   gap: $spv--small;
 }
 
+.explanation-tooltip-wrapper--overview-card {
```
edlerd: There is a visual bug that I missed in the previous PR. The arrow of the tooltip should be directly aligned with the icon it relates to.
Image

It correctly aligns when I remove the flex-gap on the container. But there should be a gap. Best find a way to get both alignment and the spacing before the icon.

Image

src/pages/instances/CreateInstance.tsx
Kxiru: I have created new create/edit ET variants in the [CopyDoc](https://docs.google.com/document/d/1vDmemn4eQm0Zr3szBRzGtWdZqGNayqHkUmtC8oZ0jNA/edit?tab=t.0). Please review.

src/pages/images/ImageRegistryExplanationTooltip.tsx
```diff
@@ -8,6 +8,7 @@ const ImageRegistryExplanationTooltip: FC<{ children?: ReactNode }> = ({
     <ExplanationTooltip
       explanation="Configure external image sources for importing images."
```
edlerd:
```suggestion
      explanation="Image registries connect to another LXD cluster to use its custom images."
```
Wdyt?

src/pages/networks/CreateNetworkAcl.tsx
```diff
@@ -159,9 +160,11 @@ const CreateNetworkAcl: FC = () => {
   return (
     <BaseLayout
       title={
-        <>
-          Create a network <abbr title="Access Control List">ACL</abbr>
-        </>
+        <NetworkAclExplanationTooltip>
+          <span>
+            Create a network <abbr title="Access Control List">ACL</abbr>
```
edlerd:
```suggestion
            Create a network ACL
```
I think we can remove this as the tooltip clarifies the abbreviation. It seems the existence of the abbreviation closes the tooltip when I hover with the mouse over it just for a split second. Hence, suggesting to remove it to make the doc link in the tooltip more easily accessible, as the tooltip won't close so quickly.

src/pages/networks/LocalPeeringExplanationTooltip.tsx
```diff
@@ -0,0 +1,18 @@
+import type { FC, ReactNode } from "react";
+import ExplanationTooltip from "components/ExplanationTooltip";
+
+const LocalPeeringExplanationTooltip: FC<{ children?: ReactNode }> = ({
+  children,
+}) => {
+  return (
+    <ExplanationTooltip
+      explanation="Route traffic directly between two networks on the same server without external gateways."
```
edlerd:
```suggestion
      explanation="Route traffic directly between two networks on the same cluster without external gateways."
```
I think "server" can be misunderstood as "single cluster member". It seems better to talk about the cluster here. Though in an unclustered case this can be slightly confusing as well. But maybe less so than the other way around.

src/pages/profiles/ProfileExplanationTooltip.tsx
```diff
@@ -1,13 +1,19 @@
...
+  children?: ReactNode;
+  isConfigVariant?: boolean;
+}> = ({ children, isConfigVariant }) => {
   return (
     <ExplanationTooltip
-      explanation="Create and manage configuration templates for instances."
+      explanation={
+        isConfigVariant
+          ? "Profiles are configuration templates for instances."
+          : "Create and manage configuration templates for instances."
```
edlerd:
```suggestion
         "Profiles are configuration templates for instances."
```
Having second thoughts on the list variant. I think the "Profiles are..." copy is easier to read. Should we unify this one and remove the config variant?

src/pages/projects/ProjectExplanationTooltip.tsx
```diff
@@ -6,8 +6,9 @@ const ProjectExplanationTooltip: FC<{ children?: ReactNode }> = ({
 }) => {
   return (
     <ExplanationTooltip
-      explanation="Organise and isolate entities and tenants into separate groups."
+      explanation="Projects organise and isolate entities and tenants into separate groups within your deployment."
```
edlerd:
```suggestion
      explanation="Projects organise and isolate entities and tenants into separate groups."
```
Deployment is a bit weird terminology here. We could say "server" but introduce the same ambiguity with "LXD cluster" as mentioned in the other comment. I think the explanation works equally well by simply removing that part.

src/pages/projects/ProjectConfigurationExplanationTooltip.tsx
edlerd: Why do we have a `ProjectConfigurationExplanationTooltip` and `ProjectExplanationTooltip` but for other entities such as instances, storage pools, and others we only have one component with the `isConfigVariant` flag? We should have one way to do this. I think splitting components as we do here with a dedicated config component might be better. Because it keeps the component itself simpler.

src/sass/_page_header.scss
```diff
@@ -6,11 +6,7 @@
     display: flex;
     flex: 2;
     flex-wrap: wrap;
-    row-gap: 1rem;
-
-    h1 {
-      margin-right: 2rem;
-    }
+    gap: 1rem 2rem;
```
edlerd, line 9: This change breaks the spacing for the auth groups page.

Before:

Image

After:

Image

edlerd (commented):
Great progress, I like the unified wording, it also seems simpler to gasp now. There are some smaller things and a layout issue as mentioned below.

src/pages/projects/ProjectConfigurationExplanationTooltip.tsx
Kxiru: The first is for the Configuration page and the second is for the project list page. I never actually paired them in my mental model but they absolutely can be joined. That is the only instance of this.

src/pages/images/ImageRegistryExplanationTooltip.tsx
Kxiru: "Image registries connect to other LXD clusters to use custom images."*?

src/pages/profiles/ProfileExplanationTooltip.tsx
Kxiru: Sure.

src/pages/profiles/ProfileExplanationTooltip.tsx
```diff
@@ -1,13 +1,14 @@
...
-const ProfileExplanationTooltip: FC<{ children?: ReactNode }> = ({
-  children,
-}) => {
+const ProfileExplanationTooltip: FC<{
+  children?: ReactNode;
+}> = ({ children }) => {
   return (
     <ExplanationTooltip
-      explanation="Create and manage configuration templates for instances."
+      explanation={"Profiles are configuration templates for instances."}
```
edlerd:
```suggestion
      explanation="Profiles are configuration templates for instances."
```

src/pages/projects/ProjectExplanationTooltip.tsx
```diff
@@ -1,13 +1,20 @@
...
-  children,
-}) => {
+const ProjectExplanationTooltip: FC<{
+  children?: ReactNode;
+  isConfigVariant?: boolean;
+}> = ({ children, isConfigVariant }) => {
   return (
     <ExplanationTooltip
-      explanation="Organise and isolate entities and tenants into separate groups."
+      className={"explanation-tooltip-wrapper--breadcrumb"}
```
edlerd:
```suggestion
      className="explanation-tooltip-wrapper--breadcrumb"
```

edlerd (dismissed):
QA and Code LGTM, with two non-blocking nitpicks on the formatting below.

src/pages/projects/ProjectExplanationTooltip.tsx
```diff
@@ -1,13 +1,20 @@
...
   return (
     <ExplanationTooltip
-      explanation="Organise and isolate entities and tenants into separate groups."
+      className="explanation-tooltip-wrapper--breadcrumb"
+      explanation={
+        isConfigVariant
+          ? "View and manage project-level configuration settings and resources."
+          : "Projects organise and isolate entities and tenants into separate groups."
+      }
       docPath="/projects/"
```
edlerd: From copilot:

> When isConfigVariant is true, this tooltip’s doc link still points to "/projects/". Previously the project-configuration breadcrumb tooltip linked to "/reference/projects/"; keeping the same docPath for both variants looks like an accidental regression in the docs destination.

edlerd (commented):
QA and code LGTM.

As discussed, let's remove the custom copy as a result of the `isConfigVariant` prop and always return the same copy that explains what an entity is with the same words. Ensuring that the copy is compatible with the list, create and edit modal use cases as well.

src/pages/cluster/ClusterGroupExplanationTooltip.tsx
```diff
@@ -1,13 +1,19 @@
...
-const ClusterGroupExplanationTooltip: FC<{ children?: ReactNode }> = ({
-  children,
-}) => {
+const ClusterGroupExplanationTooltip: FC<{
+  children?: ReactNode;
+  isConfigVariant?: boolean;
+}> = ({ children, isConfigVariant }) => {
+  const prefix =
+    "Cluster groups organise cluster members for instance placement.";
+  const suffix = " Create and manage them here.";
```
edlerd:
```suggestion
```
We decided against the suffix. This is obvious from the big green "Create group" button next to the tooltip and we don't need to bother the user with this extra copy.

edlerd (commented):
I am a bit confused, we decided not to use any suffixes yesterday, why are we adding them now?

src/pages/cluster/ClusterGroupExplanationTooltip.tsx
Kxiru: Sorry? We had a whole conversation confirming it should go forward in Standup? I'll confirm with the team because I was also suggesting we stick to V3.

Kxiru: You also said you appreciated having a unified new copy in Stand up. As you can see, all the copies have also been changed to be more descriptive of what the entity is rather than what the page does, so in actuality, do you just want me to get rid of the suffix clauses?

Kxiru: Please see the Copydoc where all 4 versions of the copies of these ETs have been logged.

kimanhou: > We decided against the suffix.

The final decision of yesterday's conversation wasn't clear to me. We agreed that the first sentence at least should describe what the entity is and should be the same everywhere. Then we discussed the possibility of adapating the explanation to its context. And we also said that, when on the Instances list page, it's obvious that the user can create and manage instances there. I believe "Create and manage them here" was a suggestion from @sophiecarroll from 2 days ago when you (@edlerd) had to leave standup.

I think we should keep the suffix concept but only use it when necessary, to add information or guidance. Most list pages don't need a suffix, but ACLs for example could benefit from having an explanation on how to use them in networks.

edlerd: I think generally adding the "Configure and manage them here" does not provide useful information and we should remove it everywhere.

I also agree some places might need additional explanation, but it should be decided on a case by case basis.

edlerd (changes requested):
See comment below.

src/pages/cluster/ClusterGroupExplanationTooltip.tsx
sophiecarroll: Agreed!

src/pages/images/ImageRegistryExplanationTooltip.tsx
```diff
@@ -1,13 +1,14 @@
...
-const ImageRegistryExplanationTooltip: FC<{ children?: ReactNode }> = ({
-  children,
-}) => {
+const ImageRegistryExplanationTooltip: FC<{
+  children?: ReactNode;
+}> = ({ children }) => {
   return (
     <ExplanationTooltip
-      explanation="Configure external image sources for importing images."
+      explanation="Image registries provide images from other LXD image servers."
```
edlerd:
```suggestion
      explanation="Image registries connect to other LXD clusters to use custom images."
```

src/pages/permissions/PermissionGroupExplanationTooltip.tsx
```diff
@@ -1,13 +1,14 @@
...
-const PermissionGroupExplanationTooltip: FC<{ children?: ReactNode }> = ({
-  children,
-}) => {
+const PermissionGroupExplanationTooltip: FC<{
+  children?: ReactNode;
+}> = ({ children }) => {
   return (
     <ExplanationTooltip
-      explanation="Organise identities into groups to manage permissions."
+      explanation="Groups organise identities for permission management."
```
edlerd:
```suggestion
      explanation="Auth groups organise identities for permission management."
```
We need to specialize this to distinguish it from cluster groups.

edlerd (dismissed):
QA and code LGTM, thanks!

Two small non-blocking suggestions below.

src/components/ExplanationTooltip.tsx
edlerd:
```suggestion
      <Icon name="help" className="explanation-tooltip-icon" />
```
Was this changed to information on purpose or accidentally as a drive-by when creating the new component? I assume the question mark/help one is more clear.

Help:

Image

Current information:

Image

Kxiru: I had a discussion on it with the design team in March: [Mattermost link](https://chat.canonical.com/canonical/pl/e7a97sh4c7f53ytsey69mdkrew).

It adds supplementary context without promising an extensive help system, and was also a point of contention in our user interviews where user expectations were in consistent with that they encountered.

edlerd: The thread seems to conclude that the question mark icon is preferred. So we had a regression when the ExplanationTooltip component landed using the information icon -- I guess?

edlerd (dismissed):
LGTM, thanks. One last question remains below.

src/pages/images/ImageRegistryExplanationTooltip.tsx
```diff
@@ -6,7 +6,7 @@ const ImageRegistryExplanationTooltip: FC<{ children?: ReactNode }> = ({
 }) => {
   return (
     <ExplanationTooltip
-      explanation="Configure external image sources for importing images."
+      explanation="Image registries connect to other LXD clusters to use custom images."
       docPath="/image-handling/"
     >
```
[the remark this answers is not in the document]
edlerd: Shall we add the "Learn more about image registries" copy here for the link?

src/components/ExplanationTooltip.tsx
Kxiru: My response in the thread wasn't necessarily my agreement, just acknowledgement of the two use cases. I do not think that a help text is entirely appropriate here as it implies that aid can be found at that tooltip, which was not the case and was also surfaced in the user interviews where users expected help but were given information (and an inconspicuous link to the docs).
Happy to revert if you so wish.

edlerd (approved):
LGTM, thanks!

sophiecarroll:
Are the tooltips up to date in this demo? I have a few short fixes @Kxiru 

1. Typo: "configuration" should be "configurations"
Screenshot 2026-08-24 at 10 42 19 AM

2. Overflow bug, tooltip hidden behind nav bar
Screenshot 2026-08-24 at 10 42 00 AM

3. definition of SSH key doesn't appear at the beginning of this tooltip
Screenshot 2026-08-24 at 10 43 10 AM

4. The tooltip here should be about "project configuration" and not "projects"–I think this is one of those times to add additional information after the definition of projects, like "Projects organise and isolate LXD resources into separate groups. Manage project configuration settings here." 
Screenshot 2026-08-24 at 10 46 52 AM

PR #2200, opened 2026-08-12
fix(instance) allow clearing of cpu and memory limit inputs

### LINKED ISSUE #2190: unable to remove 0 input on cpu count
Image

just a small annoyance

i typed 2, decided I wanted 4 so i removed the 2 and it turned into a 0. 

I am unable to remove the 0.

I had to type 04 and then remove the 0 again

edlerd:
## Done

- fix(instance) allow clearing of cpu and memory limit inputs

Fixes #2190

## QA

1. Run the LXD-UI:
    - instance detail page > configuration > resource limits
    - edit exposed cpu limit
    - type number, clear number to type another number, ensure no 0 is introduced
    - repeat test for memory limit below.

## Screenshots

image

### DIFF
--- a/src/components/forms/MemoryLimitSelector.tsx
+++ b/src/components/forms/MemoryLimitSelector.tsx
@@ -3,6 +3,7 @@ import { Input, RadioInput, Select } from "@canonical/react-components";
 import { BYTES_UNITS, MEM_LIMIT_TYPE, type MemoryLimit } from "types/limits";
 import MemoryLimitAvailable from "components/forms/MemoryLimitAvailable";
 import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
+import { valueAsNumber } from "util/valueAsNumber";
 
 interface Props {
   formik: InstanceAndProfileFormikProps;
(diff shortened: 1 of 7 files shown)


src/components/forms/MemoryLimitSelector.tsx
```diff
@@ -57,7 +58,10 @@ const MemoryLimitSelector: FC<Props> = ({
           step="Any"
           placeholder="Enter percentage"
           onChange={(e) => {
-            setMemoryLimit({ ...memoryLimit, value: +e.target.value });
+            setMemoryLimit({
+              ...memoryLimit,
+              value: valueAsNumber(e.target.value),
+            });
           }}
           value={`${memoryLimit.value ? memoryLimit.value : ""}`}
```
omarelkashef: I can't type 0 in the memory input  because of this line and  line 85.

omarelkashef (commented):
Thank you for the PR. One comment below.

omarelkashef (commented):
When I input 0 in the memory limit and hit save, nothing is sent in the payload and it is not reflected in the UI

Image

edlerd:
> When I input 0 in the memory limit and hit save, nothing is sent in the payload and it is not reflected in the UI

I think `0` is not a valid value, so discarding seems not wrong. Was it a surprise or what do you expect in this case to happen. Like why do you enter 0 and what is the intent?

omarelkashef:
> > When I input 0 in the memory limit and hit save, nothing is sent in the payload and it is not reflected in the UI
> 
> I think `0` is not a valid value, so discarding seems not wrong. Was it a surprise or what do you expect in this case to happen. Like why do you enter 0 and what is the intent?

Oh I see, maybe a validation error would be nice. It seems to the user that the value was accepted since there were no error notifications anywhere.

edlerd:
> Oh I see, maybe a validation error would be nice. It seems to the user that the value was accepted since there were no error notifications anywhere.

Changed this to surface an error on "0" input for memory.

omarelkashef (approved):
LGTM, thank you for the iterations!

PR #2206, opened 2026-08-14
feat: [WD-36583] A11y cases for side panels and modals

Kxiru:
## Done

- [x] Have tests only run in the a11y-audit suite
- [x] Add side panel tests for create panels
  - [x] Instance summary
  - [x] Storage bucket create panel
  - [x] Cluster link create panel
  - [x] Cluster replicator create panel
  - [x] Permissions identity create panel
  - [x] permissions group + IDP group create panels
 - [x] Add a11y tests for create pages
   - [x] Networks
   - [x] Instances
   - [x] Storage pools
   - [x] Storage volumes
   - [x] Network ACLs
   - [x] Profiles
   - [x] Project

 - [ ] ~~Add side panel tests for edit panels~~ This has been relegated to a [new task](https://warthogs.atlassian.net/browse/WD-38263?atlOrigin=eyJpIjoiMWY5ZmUyMzliMGU4NDA2Y2JmNzNmNjQ2ZmQ3MTBhMmIiLCJwIjoiaiJ9).
   - Storage Bucket edit panels need SBs to be created (Ceph env)
   - Only partial a11y testing possible for Clustered entities (Links, groups, placements, replicators & image registries)

- [x] Add tests for modals
  - [x] Migrate Instance (Recursive)
  - [x] Migrate Volume (Recursive)
  - [x] Custom ISO uplod
  - [x] Explort Instance
  - [x] Configure snapshot
  - [x] Delete project

Open questions:
- ~~Do we run the a11y test for clustered environments? To test cluster group creation & edit?~~ [New task](https://warthogs.atlassian.net/browse/WD-38263?atlOrigin=eyJpIjoiMWY5ZmUyMzliMGU4NDA2Y2JmNzNmNjQ2ZmQ3MTBhMmIiLCJwIjoiaiJ9)
- Do we test storage bucket edit?
- ~~Network Peering & Load Balancer create flows also cannot be captured without creating an OVN network.~~ [New task](https://warthogs.atlassian.net/browse/WD-38263?atlOrigin=eyJpIjoiMWY5ZmUyMzliMGU4NDA2Y2JmNzNmNjQ2ZmQ3MTBhMmIiLCJwIjoiaiJ9)

## QA

1. Run the LXD-UI:
    - Verify tests work locally

## Screenshots

N/A


tests/helpers/a11y.ts
```diff
@@ -130,3 +131,171 @@ export const runA11yAudit = async (
...
+  testInfo: TestInfo,
+): Promise<number> => {
+  await page.locator(PANEL_SELECTOR).waitFor({ state: "visible" });
+
+  const results = await new AxeBuilder({ page })
+    .include(PANEL_SELECTOR)
+    .withTags(["wcag2a", "wcag2aa", "wcag21a", "wcag21aa", "best-practice"])
+    .analyze();
+
+  const passed = results.passes?.length ?? 0;
```
omarelkashef: If I understand correctly, all the lines in this function except the one selecting the panelare repeated between in `runA11yAuditForPage`. Should we refactor what we can into a reusable function?

tests/helpers/a11y.ts
```diff
@@ -130,3 +131,171 @@ export const runA11yAudit = async (
...
+    await cancelButton.click();
+  } else {
+    await closeButton.click();
+  }
+
+  await panel.waitFor({ state: "hidden" });
+};
+
+export const hasTableRows = async (page: Page): Promise<boolean> => {
+  const rows = page.getByRole("row");
```
omarelkashef: I think this should be more strict as in it should look for rows in a specific table/container because the page might have rows in places we are not interested in or testing.

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,409 @@
...
+}, testInfo) => {
+  skipIfNotA11yProject(testInfo.project.name);
+
+  await clickSideNavItem(page, "Buckets", "Storage");
+
+  await page.getByRole("button", { name: "Create bucket" }).click();
+  await runA11yAuditForPanel("create-storage-bucket-panel", page, test.info());
+  await closePanel(page);
+
+  // const hasData = await hasTableRows(page);
```
omarelkashef: Is the committed code WIP?

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,409 @@
...
+  deleteInstance,
+  randomInstanceName,
+} from "./helpers/instances";
+import { openInstancePanel } from "./helpers/instancePanel";
 
-test("accessibility audit for instance page", async ({ page }) => {
+const instance = randomInstanceName();
+
+test("accessibility audit for instance page", async ({ page }, testInfo) => {
+  skipIfNotA11yProject(testInfo.project.name);
```
omarelkashef: Does it work if we have the `skipIfNotA11yProject` in a `test.beforeAll` or `test.beforeEach` hook instead of repeating it in every test?

omarelkashef (commented):
Thank you for extending the a11y checks. A couple of comments below.

tests/accessibility-audit.spec.ts
kimanhou: @omarelkashef The PR is in draft

omarelkashef: ah yes, my bad.

Kxiru: Yep, this is a draft PR with some outstanding questions.

tests/accessibility-audit.spec.ts
Kxiru: Test.skip() must be called inside a test() body and would not work as expected if placed in a beforeAll. The current approach is a standard Playwright pattern and is also seen across other test suites in the code base.

That being said, I have realised I can use describe blocks to structure the tests into different sections, and within those, they support beforeAll hooks, so I'll experiment with that and see if we can cut some lines.

tests/helpers/a11y.ts
Kxiru: Yes, there is significant overlap here, I can refactor this.

tests/helpers/a11y.ts
Kxiru: Because of the nature of the test (just opening side panels, not necessarily doing anything context specific), any row will actually do, and this check is simply to see whether an entity already exists or not. However there is an open question for some of the entities as to whether we should create resources in order to test deeper (Eg; create an Ovn network to test load balancers and network peering, or whether to use a clustered environment to test cluster links, groups and placement). In which case, more specificity may be worthwhile, but also not entirely necessary if we know there will only ever be one entity.

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,502 @@
...
+    await processA11yResults("create-project", results, test.info(), "a11y");
+  });
 });
 
-test("accessibility audit for settings page", async ({ page }) => {
-  const slug = "Settings";
-  await runA11yAudit(slug, page, test.info());
+// A11y tests for modal components
+test.describe("a11y: modals", () => {
+  const modalInstance = randomInstanceName();
```
edlerd: Let's structure the tests by category (like instance, profile, etc) instead of by type (page, modal, panel). That makes it much easier to understand which sections are covered and which are missing.

tests/helpers/a11y.ts
```diff
@@ -116,7 +100,7 @@ export const runA11yAudit = async (
       results,
     };
 
-    const filename = `a11y-${slug.toLowerCase()}-${timestamp}.json`;
+    const filename = `${filenamePrefix}-${slug.toLowerCase()}-${timestamp}.json`;
```
edlerd: Let's not use `a11y-modal` as a prefix. I would want the reports to be sorted by category. To see all the instance results together, not the modals grouped together. Instead of `filenamePrefix` better keep `a11y-` statically and introduce a `category` with values `model`, `side-panel` and `page`, to go after the `slug`.

tests/helpers/a11y.ts
```diff
@@ -130,3 +114,95 @@ export const runA11yAudit = async (
...
+  page: Page,
+  slug: string,
+  parentSlug?: string,
+): Promise<void> => {
+  await gotoURL(page, `/ui/project/default`);
+  await page.waitForLoadState("networkidle");
+
+  if (parentSlug) {
+    await page.getByRole("button", { name: parentSlug }).click();
+  }
```
edlerd: I think we need a check here if the parent is already open and only click it if closed.

tests/helpers/a11y.ts
```diff
@@ -130,3 +114,95 @@ export const runA11yAudit = async (
 
   return percent;
 };
+
+export const clickSideNavItem = async (
```
edlerd: This can be in `navigate.ts`.

tests/helpers/a11y.ts
```diff
@@ -130,3 +114,95 @@ export const runA11yAudit = async (
...
+  parentSlug?: string,
+): Promise<void> => {
+  await gotoURL(page, `/ui/project/default`);
+  await page.waitForLoadState("networkidle");
+
+  if (parentSlug) {
+    await page.getByRole("button", { name: parentSlug }).click();
+  }
+  await page.getByRole("link", { name: slug, exact: true }).first().click();
+  await page.waitForLoadState("networkidle");
```
edlerd: `waitForLoadState("networkidle")` is a bit brittle, a more stable approach is to wait for a concrete item on the page that is visible after loading.

tests/helpers/a11y.ts
```diff
@@ -130,3 +114,95 @@ export const runA11yAudit = async (
...
+  if (await cancelButton.isVisible()) {
+    await cancelButton.click();
+  } else {
+    await closeButton.click();
+  }
+
+  await panel.waitFor({ state: "hidden" });
+};
+
+export const hasTableRows = async (page: Page): Promise<boolean> => {
```
edlerd:
```suggestion
```
We don't need this helper. We should ensure that data exists, and wherever we call this function now, we should assume the data exists. If it does not, we should fail the test.

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,502 @@
...
+const instance = randomInstanceName();
 
-test("accessibility audit for network pages", async ({ page }) => {
-  const slug = "Networking";
-  await runA11yAudit("Networks", page, test.info(), slug);
-  await runA11yAudit("ACLs", page, test.info(), slug);
-  await runA11yAudit("IPAM", page, test.info(), slug);
-});
+test.describe("a11y: main pages", () => {
+  // eslint-disable-next-line no-empty-pattern
```
edlerd: Let's remove this grouping. We can have one group for all a11y tests.

tests/helpers/a11y.ts
```diff
@@ -130,3 +114,95 @@ export const runA11yAudit = async (
...
+  testInfo: TestInfo,
+  parentSlug?: string,
+  actionButtonName?: string,
+): Promise<number> => {
+  await clickSideNavItem(page, slug, parentSlug);
+
+  if (actionButtonName) {
+    await page.getByRole("button", { name: actionButtonName }).click();
+    await page.waitForLoadState("networkidle");
+  }
```
edlerd:
```suggestion
```
I think it is cleaner to not make this helper navigate. That should be a dedicated call from the test itself.

tests/helpers/a11y.ts
```diff
@@ -130,3 +114,95 @@ export const runA11yAudit = async (
...
+
+  return processA11yResults(slug, results, testInfo, "a11y-panel");
+};
+
+export const runA11yAuditForModal = async (
+  slug: string,
+  page: Page,
+  testInfo: TestInfo,
+): Promise<number> => {
+  const MODAL_SELECTOR = ".p-modal";
```
edlerd: This should be global, like the panel selector above.

edlerd (commented):
Great progress. The test is currently failing; please look at the logs. A few thoughts on the code below.

tests/helpers/a11y.ts
Kxiru: Sure thing. It was actually global first but then I changed it to be scoped as it is only used within this function. Why should it be global if it is only relevant to one place?

edlerd: Reading through the changes, it seemed odd, that the panel selector is global and the modal selector local.

tests/helpers/a11y.ts
Kxiru: It adds some repetition but I can undo that.

edlerd: It seems like an unexpected side effect, that `runA11yAudit` currently also does the navigation.

I was thinking to split this out into a dedicated `visit` function. So on the callers, we have `visit("identities")` and `runA11yAudit("identities",...)`. So we should not simply copy this to the callers but restructure the functions. That way the repetition is minimal, and the calling code gets more readable.

tests/helpers/a11y.ts
```diff
@@ -130,3 +115,68 @@ export const runA11yAudit = async (
...
+
+  const results = await new AxeBuilder({ page })
+    .include(MODAL_SELECTOR)
+    .withTags(WCAG_TAGS)
+    .analyze();
+
+  return processA11yResults(slug, results, testInfo, filenamePrefix);
+};
+
+export const closePanel = async (page: Page): Promise<void> => {
```
edlerd: This seems not related to a11y, should it be in the `navigate.ts` helper file?

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,631 @@
...
+  deleteIdpGroup,
+  randomIdpGroupName,
+} from "./helpers/permission-idp-groups";
 
-test("accessibility audit for instance page", async ({ page }) => {
-  const slug = "Instances";
-  await runA11yAudit(slug, page, test.info());
+const WCAG_TAGS = ["wcag2a", "wcag2aa", "wcag21a", "wcag21aa", "best-practice"];
+
+// --- Instances ---
```
edlerd: Instead of those comments, group the tests by type. Similarly to what you had previously with grouping by "page" and "modal," group it by entity type. See also `networks.spec.ts` for an example of grouping. That way we can satisfy two goals
1. We get rid of the brittle comments and get a more natural grouping by type as per actual code.
2. We can create and delete the instance in the `beforeAll` and `afterAll` block for the whole group. So we create one instance for all tests and not one for each instance test. That should be much faster and  stable for the CI run.
```suggestion
```

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,631 @@
...
+
+test("instance list page", async ({ page }, testInfo) => {
+  skipIfNotA11yProject(testInfo.project.name);
+  await clickSideNavItem(page, "Instances");
+  await runA11yAudit("Instances", page, test.info(), "instance");
```
edlerd:
```suggestion
  await runA11yAudit(page, test.info());
```
I think we can simplify this and don't need a prefix or the slug for each call here. We can structure the tests to have exactly one call to `runA11yAudit` per test. Some tests currently run multiple, but it should be straightforward to split those into individual tests with dedicated names. Then we don't need an additional prefix prop and neither the slug prop. We can use the `testInfo.title` instead to name the report.

So in the file name generation, we can do
```
const filename = `${testInfo.title}-${timestamp}.json`;
```

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,631 @@
...
+  await visitInstance(page, instance);
+  await page.getByRole("button", { name: "Migrate" }).click();
+  await runA11yAuditForModal(
+    "migrate-instance-method",
+    page,
+    test.info(),
+    "instance",
+  );
+
+  // Path A: root storage pool migration
```
edlerd: This should not be a comment but reflected in the test's name.
```suggestion
```

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,631 @@
...
+  skipIfNotA11yProject(testInfo.project.name);
+  await gotoURL(page, "/ui/project/default");
+  await page.getByRole("heading", { name: "Project configuration" }).waitFor();
+  await page.getByRole("button", { name: "default" }).click();
+  await page.getByRole("button", { name: "Create project" }).click();
+  await page.getByRole("heading", { name: "Create a project" }).waitFor();
+
+  const results = await new AxeBuilder({ page }).withTags(WCAG_TAGS).analyze();
+
+  await processA11yResults("create-project", results, test.info(), "project");
```
edlerd: We should use the `runA11yAudit` helper here. I think `processA11yResults` should not be exported in the helper and stay an implementation detail.

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,631 @@
...
+  skipIfNotA11yProject(testInfo.project.name);
+  const project = randomProjectName();
+  await createProject(page, project);
+
+  await gotoURL(page, "/ui/");
+  await page.getByRole("heading", { name: "Project configuration" }).waitFor();
+  await page.getByRole("button", { name: "default" }).click();
+  await page.getByRole("link", { name: project }).click();
+  await page.getByRole("heading", { name: "Project configuration" }).waitFor();
+  await page.getByRole("link", { name: "Configuration" }).click();
```
edlerd: I think we can use the `openProjectConfiguration` helper here instead.

edlerd (commented):
Good progress, thanks. Some thoughts to further simplify and better structure the tests.

tests/helpers/a11y.ts
Kxiru: What exactly are you referring to? the closePanel function?

tests/accessibility-audit.spec.ts
Kxiru: You told me not to use describe to group anything. I can put it back in.

tests/helpers/a11y.ts
Kxiru: I can move it there.

tests/accessibility-audit.spec.ts
edlerd: @Kxiru my point was we should not create an arbitrary grouping by page, modal and panel. Suggested back then to group by entity instead.

edlerd: @Kxiru see 
image

tests/helpers/a11y.ts
```diff
@@ -47,33 +51,11 @@ const countSeverities = (results: AxeResults): Record<string, number> => {
...
-};
-
-export const runA11yAudit = async (
-  slug: string,
-  page: Page,
+const processA11yResults = async (
+  results: AxeResults,
   testInfo: TestInfo,
-  parentSlug?: string,
+  slug?: string,
```
edlerd:
```suggestion
```
This seems unused now

tests/helpers/navigate.ts
```diff
@@ -7,3 +7,38 @@ export const gotoURL = async (page: Page, url: string) => {
...
+    const isExpanded = await parentButton.getAttribute("aria-expanded");
+    if (isExpanded !== "true") {
+      await parentButton.click();
+    }
+  }
+  await page.getByRole("link", { name: slug, exact: true }).first().click();
+  await page.waitForLoadState("networkidle");
+};
+
+const PANEL_SELECTOR = '[aria-label="Side panel"]';
```
edlerd: There should be just one occasion of this, there is a duplicate in the a11y helper.

tests/a11y-audit.spec.ts
```diff
@@ -1,100 +1,584 @@
...
+
+  test("summary panel", async ({ page }) => {
+    await openInstancePanel(page, instance);
+    await runA11yAuditForPanel("summary", page, test.info());
+  });
+
+  test("create page", async ({ page }) => {
+    await clickSideNavItem(page, "Instances");
+    await page.getByRole("button", { name: "Create instance" }).click();
+    await runA11yAudit(page, test.info());
```
edlerd: We should ensure we are on the create page before starting the a11y audit. Ideally we can refactor a navigation helper for most cases below like a small function as part of existing test helpers to reuse here. So we have as few logic in this test as possible to maintain.

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,584 @@
...
-  await runA11yAudit("Networks", page, test.info(), slug);
-  await runA11yAudit("ACLs", page, test.info(), slug);
-  await runA11yAudit("IPAM", page, test.info(), slug);
+    await page
+      .getByRole("row")
+      .getByRole("button", { name: "Select" })
+      .and(page.locator(":not([aria-disabled='true'])"))
+      .first()
+      .click();
+    await runA11yAuditForModal("confirm", page, test.info());
```
edlerd:
```suggestion
    await runA11yAuditForModal(page, test.info());
```
Let's remove the slug completely.

tests/accessibility-audit.spec.ts
```diff
@@ -1,100 +1,584 @@
...
+    await runA11yAuditForModal("select", page, test.info());
+
+    await page
+      .getByRole("row")
+      .getByRole("button", { name: "Select" })
+      .and(page.locator(":not([aria-disabled='true'])"))
+      .first()
+      .click();
+    await runA11yAuditForModal("confirm", page, test.info());
+  });
```
edlerd: Let's split this into two tests so we can remove the slug.

tests/a11y-audit.spec.ts
```diff
@@ -1,100 +1,584 @@
...
+
+  test.beforeAll(async ({ browser }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    const page = await browser.newPage();
+    await createVolume(page, volume);
+    await page.close();
+  });
+
+  test.afterAll(async ({ browser }) => {
+    const page = await browser.newPage();
```
edlerd, line 348: this is missing the skipIfNotA11yProject

tests/a11y-audit.spec.ts
```diff
@@ -1,100 +1,584 @@
...
-});
+  test.beforeAll(async ({ browser }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    const page = await browser.newPage();
+    await createInstance(page, instance);
+    await page.close();
+  });
+
+  test.afterAll(async ({ browser }) => {
+    const page = await browser.newPage();
```
edlerd, line 91: this is missing the skipIfNotA11yProject

edlerd (commented):
Please check the CI output, several tests are not excluded properly, such as [this one](https://github.com/canonical/lxd-ui/actions/runs/32312901703/job/96259503356?pr=2206) is executing a11y tests in the wrong context. Also the a11y test itself is currently failing.

edlerd:
I realized we are not testing detail pages at all. Such as the various sections of instance detail and other entities such as network detail pages are missing. Is there a tracking ticket to add them too? Or should we do it in this PR?

tests/helpers/a11y.ts
Kxiru: The slug is necessary for clickSideNavItem navigation and the print helper.

edlerd: We have no dependency on the `clickSideNavItem` in here, do we?

Kxiru: Sorry, I was on the wrong branch and forgot to delete this comment.

tests/a11y-audit.spec.ts
Kxiru: Please review the current iteration and advise.

edlerd (changes requested):
The newly added test suite is failing. Please fix @Kxiru .

tests/accessibility-audit.spec.ts
```diff
@@ -1,78 +1,729 @@
...
+    const page = await browser.newPage();
+    await deleteInstance(page, instance);
+    await page.close();
+  });
+
+  // eslint-disable-next-line no-empty-pattern
+  test.beforeAll(({}, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+  });
+
```
edlerd:
```suggestion
```
I am never sure if this actually works as intended. The custom eslint disabler is needed on every test case, making this a bit ugly. A better option is to put the `skipIfNotA11yProject` at the top of each test. It will be aligned with what we do in other tests, and it will be clear what is actually happening.

We still need to return early from the before* and after* handlers so they don't get run in other test suites even when the actual tests get skipped.

Probably best to make this aligned with the existing e2e test suites.

tests/accessibility-audit.spec.ts
```diff
@@ -1,78 +1,729 @@
...
+    await page.getByRole("button", { name: "Migrate" }).click();
+
+    const clusterButton = page.getByRole("button", {
+      name: "Migrate instance to a different cluster member",
+    });
+    test.skip(
+      !(await clusterButton.isVisible()),
+      "Not a clustered environment",
+    );
+
```
edlerd:
```suggestion
```
I think the suite should be opinionated whether it is run in a clustered environment or not. We don't plan to run it in both environments, so we should not silently hide the error here. Maybe we should remove all the clustered tests for now?

Either way, the right way to skip would be the `skipIfNotClustered` helper, not checking a button state.

tests/helpers/a11y.ts
edlerd, line 1: Can we align the naming, please? In the test we use `accessibility` and in the helper `a11y`. Let's settle on one of the two.

tests/helpers/a11y.ts
```diff
@@ -116,7 +88,7 @@ export const runA11yAudit = async (
       results,
     };
 
-    const filename = `a11y-${slug.toLowerCase()}-${timestamp}.json`;
+    const filename = `${testInfo.title}-${timestamp}.json`;
```
edlerd:
```suggestion
const testSuiteName = test.info().titlePath.at(-2);
    const filename = `${testSuiteName}-${testName}-${timestamp}.json`;
```
We can reuse the `testName` variable here instead.

We should prefix with the suite name. Otherwise, for the settings test the name will be "list page". We want it to be "settings - list page" instead.

edlerd (changes requested):
The new test suite is failing.

tests/accessibility-audit.spec.ts
Kxiru: I will switch to the skipIfNotClustered helper. We have a follow up task aleady to make the a11y ci clustered so we can test clustered items such as replicators, and address network forwards through the env that is - hopefully - also ovn compatible.
[Task](https://warthogs.atlassian.net/browse/WD-38263?atlOrigin=eyJpIjoiYjU2Yjk2MmNhNDVkNDI3MGE2ZTUxYzUyYTc3N2FlODAiLCJwIjoiaiJ9).

tests/helpers/a11y.ts
Kxiru: Will rename accessibility-audit.spec.ts to a11y-audit.spec.ts

tests/helpers/a11y.ts
Kxiru: Sounds good!

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,771 @@
...
+    await page.getByRole("button", { name: "Migrate" }).click();
+
+    await page
+      .getByRole("button", { name: "Move instance root storage" })
+      .click();
+
+    await page
+      .getByRole("row")
+      .getByRole("button", { name: "Select" })
+      .and(page.locator(":not([aria-disabled='true'])"))
```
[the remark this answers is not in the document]
Kxiru: Not applicable as the reactcomponents button uses aria-disabled, not HTML-native disabled. Also enablement is class baased and is not set on the element so the existing selector is correct for this library.

tests/helpers/a11y.ts
```diff
@@ -116,7 +88,8 @@ export const runA11yAudit = async (
       results,
     };
 
-    const filename = `a11y-${slug.toLowerCase()}-${timestamp}.json`;
+    const testSuiteName = test.info().titlePath.at(-2);
```
edlerd:
```suggestion
    const testSuiteName = testInfo.titlePath.at(-2);
```
I think this doesn't work as we need to use the correct reference to the calling test from the input to this method.

tests/helpers/a11y.ts
```diff
@@ -116,7 +88,8 @@ export const runA11yAudit = async (
...
+    const filename = `${testSuiteName}-${testName}-${timestamp}.json`;
     const outPath = testInfo.outputPath(filename);
```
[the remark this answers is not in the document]
edlerd: Can we do some sanitization as suggested?

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+    skipIfNotA11yProject(testInfo.project.name);
+    await visitInstance(page, instance);
+    await runA11yAudit(page, test.info());
+  });
+
+  test("summary panel", async ({ page }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    await openInstancePanel(page, instance);
+    await runA11yAuditForPanel(page, test.info());
+  });
```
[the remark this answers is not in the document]
edlerd: I think we need to use `testInfo` here and on all other occasions in this file. Otherwise, the reference to the test object is wrong.

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,764 @@
...
+
+  test("pools page", async ({ page }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    await clickSideNavItem(page, "Pools", "Storage");
+    await runA11yAudit(page, test.info());
+  });
+
+  test("pool detail page", async ({ page }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    await visitPool(page, "default");
```
edlerd: We have to create a pool similar to how we create the volume. In fact the volume should be created in the pool we create.

It is not guaranteed to have a pool with name default in the test environment.

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,764 @@
...
+      .getByRole("button", { name: "Select" })
+      .and(page.locator(":not([aria-disabled='true'])"))
+      .first()
+      .click();
+    await runA11yAuditForModal(page, test.info());
+  });
+});
+
+test.describe("images", () => {
+  const imageRegistry = randomImageRegistryName();
```
edlerd:
```suggestion
```
As this is not shared, let's keep it in the `image registry detail page` test.

edlerd (commented):
Nice how clean the test is becoming 👍 

Few thoughts and concerns below.

tests/helpers/a11y.ts
```diff
@@ -1,21 +1,35 @@
...
+const WCAG_TAGS = ["wcag2a", "wcag2aa", "wcag21a", "wcag21aa", "best-practice"];
+
+const MODAL_SELECTOR = ".p-modal";
+
+const sanitizeForFilename = (
+  value: string | undefined,
+  fallback = "unknown",
+): string => {
+  const raw = value?.trim() || fallback;
+  const sanitized = raw.replace(/[/\\:*?"<>|\n\r\0]/g, "_");
```
edlerd: Better have an allowed list of symbols than listing unallowed symbols. I.e. this is missing whitespace. Allowed should be only alphanumeric characters, minus and dot.

tests/helpers/a11y.ts
```diff
@@ -116,7 +98,9 @@ export const runA11yAudit = async (
       results,
     };
 
-    const filename = `a11y-${slug.toLowerCase()}-${timestamp}.json`;
+    const testSuiteName = sanitizeForFilename(testInfo.titlePath.at(-2));
+    const safeName = sanitizeForFilename(testName);
+    const filename = `${testSuiteName}-${safeName}-${timestamp}.json`;
```
edlerd:
```suggestion
    const testSuite = testInfo.titlePath.at(-2);
    const filename = sanitizeForFilename(`${testSuite}-${testName}-${timestamp}.json`);
```
A bit cleaner.

tests/helpers/selectors.ts
```diff
@@ -0,0 +1 @@
+export const PANEL_SELECTOR = '[aria-label="Side panel"]';
```
edlerd: Can we keep this in a11y.ts helper next to the modal selector?

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,774 @@
...
+
+  test("create identity type selection panel", async ({ page }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    await clickSideNavItem(page, "Identities", "Permissions");
+    await page.getByRole("button", { name: "Create identity" }).click();
+    await runA11yAuditForPanel(page, testInfo);
+    await closePanel(page);
+  });
+
+  test("create identity details panel", async ({ page }, testInfo) => {
```
edlerd:
```suggestion
  test("create identity certificate type panel", async ({ page }, testInfo) => {
```
We might want to add the other types here later on.

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,774 @@
...
+
+  test("create cluster link direction panel", async ({ page }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    await clickSideNavItem(page, "Links", "Clustering");
+    await page.getByRole("button", { name: "Create cluster link" }).click();
+    await runA11yAuditForPanel(page, testInfo);
+    await closePanel(page);
+  });
+
+  test("create cluster link details panel", async ({ page }, testInfo) => {
```
edlerd:
```suggestion
  test("create cluster link bidirectional panel", async ({ page }, testInfo) => {
```
We might want to test the other type later on.

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,774 @@
...
+    await createImageRegistry(page, imageRegistry, "SimpleStreams", {
+      url: "https://images.linuxcontainers.org",
+    });
+    await visitImageRegistry(page, imageRegistry);
+    await runA11yAudit(page, testInfo);
+    await deleteImageRegistry(page, imageRegistry);
+  });
+
+  test("image registries list page", async ({ page }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
```
edlerd, line 543:
```suggestion
    skipIfNotA11yProject(testInfo.project.name);
    skipIfImageRegistriesNotSupported(lxdVersion);
```
This will fail atm.

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,774 @@
...
+      .click();
+
+    await page
+      .getByRole("row")
+      .getByRole("button", { name: "Select" })
+      .and(page.locator(":not([aria-disabled='true'])"))
+      .first()
+      .click();
+    await runA11yAuditForModal(page, testInfo);
+  });
```
edlerd, line 161: This test needs a secondary storage pool, currently it fails as there is no option available to select.

tests/a11y-audit.spec.ts
```diff
@@ -0,0 +1,774 @@
...
+test.describe("storage", () => {
+  const volume = randomVolumeName();
+  const pool = randomPoolName();
+
+  test.beforeAll(async ({ browser }, testInfo) => {
+    if (!isA11yProject(testInfo.project.name)) {
+      return;
+    }
+    const page = await browser.newPage();
+    await createPool(page, pool);
```
edlerd, line 338: This is missing an import

edlerd (changes requested):
Good progress, few more comments and we should get the suite to pass.

tests/a11y-audit.spec.ts
Kxiru: Yes, still tidying. I;ll put the PR in draft.

Kxiru: Are you notified when there is a new push? Because usually I resend the re-review notification manually or in mattermost when the PR is stable.

edlerd: > Are you notified when there is a new push? 

yes, when in draft I usually skip the notification and when it is green/open then I start a review as I have time :)

tests/helpers/cluster.ts
```diff
@@ -26,6 +26,19 @@ export const isClusteredTestProject = (projectName: string) => {
...
+    metadata?: { environment?: { server_clustered?: boolean } };
+  };
+  return data?.metadata?.environment?.server_clustered ?? false;
+};
+
+export const skipIfNotClusteredEnvironment = async (page: Page) => {
+  const clustered = await isClusteredEnvironment(page);
+  test.skip(!clustered, "Skipping: LXD server is not clustered");
+};
+
```
edlerd, line 41: Can we use the existing `skipIfNotClustered` and `isClusteredTestProject` helpers instead?

edlerd (commented):
Didn't review the test in detail, but structurally it seems good.  Helpers LGTM, just one idea to reuse.

We should get CI to run on it, not sure why the github action is not picking this one up. Maybe trigger with another push?

tests/helpers/cluster.ts
Kxiru: No, because those helpers filter according to whether the test ends in ":clustered", not whether the environment is actually clustered or not, which is what I am trying to achieve. skipIfNotClustered and isClusteredTestProject are irrelevant because the test projects that these actually apply to are never even supposed to touch this area,

edlerd (changes requested):
Mostly LGTM, though the new test suite is still failing

Image

tests/a11y-audit.spec.ts
Kxiru: Lool, all good then :) I'll usually send notifications when I'm confident in the output. In the interim, please expect that things could be a little shaky!

playwright.config.ts
```diff
@@ -71,6 +71,7 @@ const config: PlaywrightTestConfig<TestOptions> = {
         lxdVersion: "latest-edge",
       },
       dependencies: ["login-chromium"],
+      testIgnore: "a11y-audit.spec.ts",
     },
```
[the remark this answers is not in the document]
Kxiru: All projects other than coverage have testmatch or testignore.

edlerd (approved):
Very nice, LGTM.

As a follow-up, we can add the instance detail page tabs and also ensure the list pages have content. I think often we currently check only the empty state.

Issue #1465, opened 2025-09-02
Feature Request: Support for user keys


vosdev:
In LXD the only form of tagging we have are via the user.* keys. 

https://documentation.ubuntu.com/lxd/latest/reference/instance_options/#instance-miscellaneous:user.*

These are key=value options that can be configured on an instance or profile.

Currently these are nowhere to be found in the UI. I've requested support for the search function for user keys here: https://github.com/canonical/lxd-ui/issues/1453

But I think it would also be fitting to display the user keys in the overview of an instance. For example they can be displayed in a table like Networks, Devices and Profiles.

Also the ability to edit the values in the configuration section is missing. Currently it's only possible via the YAML editor. I understand this is a design challenge as both the key and value are a free form field.

Played around a bit with the html of my lxd ui:

<img width="1829" height="741" alt="Image" src="https://github.com/user-attachments/assets/12d64633-0f70-4048-9693-06f3a2f86732" />

I placed them at the bottom in my example. I don't know if a better place would be higher up or somewhere else.

Another option would be this:, but then it would perhaps be a bit harder to optionally link to the profile it's coming from: 

<img width="2093" height="857" alt="Image" src="https://github.com/user-attachments/assets/cd146ac2-7cd3-45b4-aeba-107196d67940" />

With the profile chip added:

<img width="1219" height="401" alt="Image" src="https://github.com/user-attachments/assets/adf09e2a-f07f-42ab-a68b-805750468c6a" />

The whole table could be hidden if there are no userkeys/tags to keep it clean for users that do not make use of this feature :)

vosdev:
With the help of Claude I/we added the ability to edit keys on instances/profiles and show them on the instance detail/summary and instances list page.

I tried to keep the feedback from my previous PRs in mind.

At first I had a profile icon next to an inherited user key, but it looked out of place because its a chip for a user key and not a profile. So I kept the chips without an icon. If you have any suggestions for a chip icon I'm all ears!

What do you think about this style?

Edit keys on instance & profile:

<img width="1030" height="806" alt="Image" src="https://github.com/user-attachments/assets/e3f05252-c743-4611-a817-b56d42fcb977" />

profile key override:

<img width="910" height="353" alt="Image" src="https://github.com/user-attachments/assets/8862c17a-a297-40ad-ba55-bc810ca6dc54" />

Editting happens in similar fashion to the user key on the server settings page:

[...]

PR #2208, opened 2026-08-14
feat(Overview): Storage card

kimanhou:
## Done

- Implement Storage card for Overview page.

## QA

1. Run the LXD-UI:
    - Go to Overview page. Admire Storage card.

## Screenshots

image

### DIFF
--- a/src/pages/overview/StorageCard.tsx
+++ b/src/pages/overview/StorageCard.tsx
@@ -1,8 +1,85 @@
 import type { FC } from "react";
-import { Card } from "@canonical/react-components";
+import { Link } from "react-router-dom";
+import { Card, Icon, List, Spinner } from "@canonical/react-components";
+import ExplanationTooltip from "components/ExplanationTooltip";
+import { useAuth } from "context/auth";
+import { useStoragePools } from "context/useStoragePools";
+import StoragePoolDetails from "pages/overview/StoragePoolDetails";
+import type { LxdStoragePool } from "types/storage";
+import { ROOT_PATH } from "util/rootPath";
+import { getVolumesUsedByPool } from "util/storagePool";
+import { pluralize } from "util/helpers";
 
 const StorageCard: FC = () => {
-  return <Card className="overview-card storage"></Card>;
+  const { data: pools = [], error, isLoading } = useStoragePools();
+  const { defaultProject } = useAuth();
+
+  const totalVolumeCount = pools.reduce(
+    (count, pool: LxdStoragePool) => count + getVolumesUsedByPool(pool).length,
+    0,
+  );
+
+  const cardClassName = "overview-card storage";
+  const cardTitle = (
+    <>
+      <span className="overview-card-title">
+        <Icon name="storage-pool" /> Storage
+      </span>
+      <ExplanationTooltip
+        explanation={
+          <>
+            Storage pools host instance and image data. <br />
+            Storage volumes provide storage for instances.
+          </>
+        }
+        docPath="/storage"
+      />
+    </>
+  );
+
+  if (isLoading) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <Spinner className="u-loader" text="Loading storage pools..." />
+      </Card>
+    );
+  }
+
+  if (error) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <div className="error-message">
+          <Icon name="error" className="margin-right--large" /> Error while
+          loading storage pools: {error.message}
+        </div>
+      </Card>
+    );
+  }
+
+  return (
+    <Card className={cardClassName} title={cardTitle}>
+      <List
+        inline
+        middot
+        items={[
+          `${pools.length} ${pluralize("pool", pools.length)}`,
+          `${totalVolumeCount} ${pluralize("volume", totalVolumeCount)}`,
+        ]}
+      />
+      <div className="storage-pools-container">
+        {pools.map((pool) => (
+          <StoragePoolDetails pool={pool} key={pool.name} />
+        ))}
+      </div>
+      <div className="card-footer">
+        <Link
+          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(defaultProject)}/storage/pools`}
+        >
+          See all storage pools
+        </Link>
+      </div>
+    </Card>
+  );
 };
 
 export default StorageCard;
--- a/src/pages/overview/StorageCard.tsx
+++ b/src/pages/overview/StorageCard.tsx
@@ -1,8 +1,94 @@
 import type { FC } from "react";
-import { Card } from "@canonical/react-components";
+import { Link } from "react-router-dom";
+import { Card, Icon, Spinner } from "@canonical/react-components";
+import ExplanationTooltip from "components/ExplanationTooltip";
+import { useStoragePools } from "context/useStoragePools";
+import StoragePoolDetails from "pages/overview/StoragePoolDetails";
+import type { LxdStoragePool } from "types/storage";
+import { ROOT_PATH } from "util/rootPath";
+import { getVolumesUsedByPool } from "util/storagePool";
+import { pluralize } from "util/helpers";
 
 const StorageCard: FC = () => {
-  return <Card className="overview-card storage"></Card>;
+  const { data: pools = [], error, isLoading } = useStoragePools();
+
+  const poolsWithVolumeCount = pools.map((pool: LxdStoragePool) => {
+    return {
+      name: pool.name,
+      volumeCount: getVolumesUsedByPool(pool).length,
+    };
+  });
+  const totalVolumeCount = poolsWithVolumeCount.reduce(
+    (acc, pool) => acc + pool.volumeCount,
+    0,
+  );
+
+  const cardClassName = "overview-card storage";
+  const cardTitle = (
+    <>
+      <span className="overview-card-title">
+        <Icon name="storage-pool" /> Storage
+      </span>
+      <ExplanationTooltip
+        explanation={
+          <>
+            Storage pools host instance and image data. <br />
+            Storage volumes provide storage for instances.
+          </>
+        }
+        docPath="/howto/troubleshoot/"
+      />
+    </>
+  );
+
+  if (isLoading) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <Spinner className="u-loader" text="Loading..." />
+      </Card>
+    );
+  }
+
+  if (error) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <div className="error-message">
+          <Icon name="error" className="margin-right--large" /> Error while
+          loading storage pools: {error.message}
+        </div>
+      </Card>
+    );
+  }
+
+  return (
+    <Card className={cardClassName} title={cardTitle}>
+      {!isLoading && !error && (
+        <>
+          <p>
+            {pools.length} {pluralize("pool", pools.length)}&nbsp;|&nbsp;
+            {totalVolumeCount} {pluralize("volume", totalVolumeCount)}
+          </p>
+          <div className="storage-pools-container">
+            {pools.map((pool) => (
+              <StoragePoolDetails
+                pool={pool}
+                volumeCount={
+                  poolsWithVolumeCount.find((p) => p.name === pool.name)
+                    ?.volumeCount || 0
+                }
+                key={pool.name}
+              />
+            ))}
+          </div>
+          <div className="card-footer">
+            <Link to={`${ROOT_PATH}/ui/project/default/storage/pools`}>
+              See more
+            </Link>
+          </div>
+        </>
+      )}
+    </Card>
+  );
 };
 
 export default StorageCard;
--- a/src/pages/overview/StoragePoolDetails.tsx
+++ b/src/pages/overview/StoragePoolDetails.tsx
@@ -1,0 +1,39 @@
+import { type FC } from "react";
+import { Chip } from "@canonical/react-components";
+import StoragePoolSize from "pages/storage/StoragePoolSize";
+import type { LxdStoragePool } from "types/storage";
+import { pluralize } from "util/helpers";
+
+interface Props {
+  pool: LxdStoragePool;
+  volumeCount: number;
+}
+
+const StoragePoolDetails: FC<Props> = ({ pool, volumeCount }) => {
+  return (
+    <div className="storage-pool-details">
+      <p className="u-truncate" title={pool.name}>
+        <b>{pool.name}</b>
+        {pool.status === "Errored" && (
+          <>
+            <br />
+            <Chip
+              className="u-no-margin"
+              appearance="negative"
+              value="Errored"
+              lead="status"
+              isReadOnly
+            />
+          </>
+        )}
+      </p>
+
+      <p>
+        {volumeCount} {pluralize("volume", volumeCount)}
+      </p>
+      <StoragePoolSize pool={pool} hasMeterBar displayPercentage />
+    </div>
+  );
+};
+
+export default StoragePoolDetails;
--- a/src/pages/overview/StoragePoolDetails.tsx
+++ b/src/pages/overview/StoragePoolDetails.tsx
@@ -1,0 +1,69 @@
+import type { FC } from "react";
+import { Link } from "react-router-dom";
+import { Chip } from "@canonical/react-components";
+import { useAuth } from "context/auth";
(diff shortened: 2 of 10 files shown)


src/pages/storage/StoragePoolSize.tsx
```diff
@@ -47,6 +52,17 @@ const StoragePoolSize: FC<Props> = ({
...
 
+        const usedPercentage = total > 0 ? (used * 100) / total : 0;
+        const getMeterColor = (usedPercentage: number) => {
+          if (usedPercentage < 80) {
+            return "green";
+          } else if (usedPercentage < 95) {
+            return "orange";
+          } else {
+            return "red";
+          }
```
edlerd: We should keep this consistent. We have meters on the instance overview, the cluster list, and the storage pool list. Additionally, there are progress bars, such as on ISO upload or instance import.

I think there should be the same color on all of them, including the progress bars.

vosdev:
Related: https://github.com/canonical/lxd/issues/16689

The used storage metric is wrong for Ceph

sophiecarroll:
My UX responses to open discussion points @kimanhou :

> I'm not fan of the border around each pool detail

This is the current of the MAAS cluster dashboard, and it keeps the information visually separate between pools, so we should keep it for now. We can investigate during the next update if it still makes sense.

> I think displaying the percentage is useless

My reasoning for doing this is that engineers and users frequently cited 80% as being the usage point at which they start to be worried, so I wanted to make sure they could see that important number if they have 135/150 GiB for example, at which point it becomes hard to know off the top of your head what percent that is

> I added colors to the Meter depending on how full the pool is:
Green (< 80%) — Storage pool is healthy with plenty of free space
Orange (80-95%) — Storage pool is getting full, approaching capacity
Red (≥ 95%) — Storage pool is critically full, immediate attention needed

We decided not to use colors because of the reason I elaborated in the dashboard decisions section of my mid year break standup document:
Screenshot 2026-08-20 at 2 48 04 PM

So please remove colors

> Why aren't we using rich chips for storage pool 
names 

I'm open to hearing any supporting reasons for using the rich chip, but I didn't use rich chips because they made the design a little messy, and we display most of the important information from the rich chips in the description of the pool anyways. 

> "See more" currently points to /ui/project/default/storage/pools so clicking on it makes the project selector jump from "All projects" to "default".

This is unavoidable right now with the nav setup we have, but I would like to investigate this for the future because I agree it's confusing

> I added a read-only chip if the storage pool has status=Errored. Do we like it ? (cf conversation with vosdev in Feature Request: Support for user keys #1465 )

I'll have to investigate more about chips and if there's another way to display the error–I would say keep the chip for now, but it's true that it breaks our chip pattern so I'll keep thinking about alternatives

kimanhou:
I removed the Meter colors and added the explanation tooltip.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -58,8 +71,11 @@ const StoragePoolSize: FC<Props> = ({
         return (
           <Meter
             key={resourceKey}
-            percentage={(100 / total) * used || 0}
-            text={`${humanFileSize(used)} of ${humanFileSize(total)} used`}
+            percentage={usedPercentage}
+            text={
+              `${humanFileSize(used)} of ${humanFileSize(total)} used` +
```
edlerd: Let's round to whole numbers here, wdyt?

Before:

Image

After: 

Image

src/sass/_overview.scss
```diff
@@ -112,6 +112,29 @@
...
+          @include mobile-and-tablet {
+            grid-template-columns: repeat(2, minmax(0, 1fr));
+          }
+
+          @include mobile {
+            grid-template-columns: 1fr;
+          }
+
+          .storage-pool-details {
+            border: solid 1px var(--color-border);
```
edlerd: The border seems very dark compared to the storage card itself. In the dark theme it seems good, but in light theme the contrast of the border seems too high. Shall we use one with less contrast in the light theme?

Image

src/pages/overview/StorageCard.tsx
```diff
@@ -1,8 +1,95 @@
...
+  const totalVolumeCount = [...volumeCountByPoolName.values()].reduce(
+    (acc, count) => acc + count,
+    0,
+  );
+
+  const cardClassName = "overview-card storage";
+  const cardTitle = (
+    <>
+      <span className="overview-card-title">
+        <Icon name="storage-pool" /> Storage
```
edlerd, line 26: With this heading, is it immediately clear that the cards are representing storage pools? Maybe it is.

src/pages/overview/StorageCard.tsx
```diff
@@ -1,8 +1,95 @@
...
+      </span>
+      <ExplanationTooltip
+        explanation={
+          <>
+            Storage pools host instance and image data. <br />
+            Storage volumes provide storage for instances.
+          </>
+        }
+        docPath="/howto/troubleshoot/"
```
edlerd: the link seems wrong.

src/pages/overview/StorageCard.tsx
```diff
@@ -1,8 +1,95 @@
...
+          <Icon name="error" className="margin-right--large" /> Error while
+          loading storage pools: {error.message}
+        </div>
+      </Card>
+    );
+  }
+
+  return (
+    <Card className={cardClassName} title={cardTitle}>
+      {!isLoading && !error && (
```
edlerd:
```suggestion
```
If we return early above, we don't need these guards here, do we?

src/pages/overview/StorageCard.tsx
```diff
@@ -1,8 +1,95 @@
...
+        <>
+          <p>
+            {pools.length} {pluralize("pool", pools.length)}&nbsp;|&nbsp;
+            {totalVolumeCount} {pluralize("volume", totalVolumeCount)}
+          </p>
```
edlerd: Instead of the custom pipe symbol, should we use a [list with a middot](https://canonical.github.io/react-components/?path=/story/components-list--middot) here?

src/pages/overview/StoragePoolDetails.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const StoragePoolDetails: FC<Props> = ({ pool, volumeCount }) => {
+  const isClustered = useIsClustered();
+  const { defaultProject } = useAuth();
+  const { data: resources } = useStoragePoolResources(pool);
+  const isMultiMemberLocalCluster =
+    isClustered &&
+    isClusterLocalDriver(pool.driver) &&
+    ensureArray(resources).length > 1;
+  const poolDetailUrl = `${ROOT_PATH}/ui/project/${encodeURIComponent(defaultProject)}/storage/pool/${encodeURIComponent(pool.name)}`;
```
edlerd: If we are on the view for a specific project, this should not go to the default project.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -11,22 +11,33 @@ interface Props {
...
+  resources?: LxdStoragePoolResources[];
   forceSingleLine?: boolean;
+  displayPercentage?: boolean;
 }
 
 const StoragePoolSize: FC<Props> = ({
   pool,
   hasMeterBar,
   member,
+  resources,
```
```suggestion
```
I think we can fetch this internally, this should hit the same query cache and not result in duplicate queries. It seems way more straight forward to always fetch inside here.

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -11,22 +11,33 @@ interface Props {
   pool: LxdStoragePool;
   hasMeterBar?: boolean;
   member?: LxdClusterMember;
+  resources?: LxdStoragePoolResources[];
   forceSingleLine?: boolean;
+  displayPercentage?: boolean;
```
```suggestion
  hasPercentage?: boolean;
```
More aligned with existing `hasMeterBar`?

src/pages/overview/StorageCard.tsx
```diff
@@ -1,8 +1,95 @@
...
+  const volumeCountByPoolName = new Map<string, number>(
+    pools.map((pool: LxdStoragePool) => [
+      pool.name,
+      getVolumesUsedByPool(pool).length,
+    ]),
+  );
+  const totalVolumeCount = [...volumeCountByPoolName.values()].reduce(
+    (acc, count) => acc + count,
+    0,
+  );
```
edlerd: It is a bit odd to create the `volumeCountByPoolName` map just to deconstruct it right away and then do a single lookup in it.

Maybe simpler to do the counting of all volumes here and do the lookup inside `StoragePoolDetails` that already receives all information as part of the `pool` object.

edlerd (commented):
This works great on clustered and non clustered backends! Few thoughts and ideas on the proposed changes from a detailed QA and code review.

src/pages/overview/StorageCard.tsx
kimanhou: @sophiecarroll

src/pages/overview/StorageCard.tsx
kimanhou: Oops sorry !

src/pages/overview/StorageCard.tsx
kimanhou: @sophiecarroll

src/pages/storage/StoragePoolSize.tsx
kimanhou: Okay but for that I have to modify `humanFileSize`, which is used in other places. Is that okay ?

src/sass/_overview.scss
kimanhou: Let's use the same variable for the border color as `p-card`

image

src/pages/storage/StoragePoolSize.tsx
edlerd: Maybe we can make the precision an optional parameter to the function.

src/pages/overview/StoragePoolDetails.tsx
kimanhou: Right, is it okay to use `useCurrentProject` ?
The thing is, the overview page currently forces the project selector in the navigation to be on "All projects", so we don't have this use case yet. I was planning on changing that when re-working the projects card (wip).

src/pages/storage/StoragePoolSize.tsx
kimanhou: This stems from a Copilot review
image

I'll revert the changes.

src/sass/_overview.scss
edlerd: Fine by me, but not sure if that is the intended design. @sophiecarroll wdyt?

src/pages/overview/StoragePoolDetails.tsx
edlerd: Fine to keep it as is and address it as one in a follow up from my side.

edlerd: I assume `useCurrentProject` might be the right helper here, yes.

src/pages/storage/StoragePoolSize.tsx
edlerd: Yeah, I think the copilot suggestion is not good, as it leads to more complexity with a bigger interface and closer coupling.

src/sass/_overview.scss
sophiecarroll: Looks good to me

src/pages/overview/StorageCard.tsx
sophiecarroll: I think yes, I had this conversation with Amy a while ago when she was working on pools and she said people remembered the names of their pools, and we also define volumes as something separate within the card, as well as using the pools icon–we can confirm this in testing as well

src/pages/overview/StorageCard.tsx
sophiecarroll: Can someone give me a screenshot of which part this is talking about? I'm not clear on what a custom pope symbol is and where we're using it

kimanhou: Pipe
image

Middot
image

sophiecarroll: We used the pipe in quite a few cards: what is your reasoning for changing to middot? I'm fine with changing to middot but we would need to change permissions, instances, storage, and clustering.

edlerd: @sophiecarroll The middot is an established pattern and we have an existing component for it. For the pipe, we'd need to build and maintain a custom component. Hence, the idea to use the middot here to keep it simple and consistent.

edlerd: My suggestion was to change it here and in all the other cards at the same time, yes.

kimanhou: I'll change it to middot everywhere then

src/pages/overview/StoragePoolDetails.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { data: resources } = useStoragePoolResources(pool);
+  const volumeCount = getVolumesUsedByPool(pool).length;
+  const isMultiMemberLocalCluster =
+    isClustered &&
+    isClusterLocalDriver(pool.driver) &&
+    ensureArray(resources).length > 1;
+  const poolProject =
+    projectName !== "default" && projectName !== ALL_PROJECTS
+      ? projectName
+      : defaultProject;
```
edlerd:
```suggestion
  const project =
    currentProject !== ALL_PROJECTS
      ? currentProject
      : defaultProject;
```
Can this be a bit simpler?

src/pages/overview/StoragePoolDetails.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { getVolumesUsedByPool, isClusterLocalDriver } from "util/storagePool";
+
+interface Props {
+  pool: LxdStoragePool;
+}
+
+const StoragePoolDetails: FC<Props> = ({ pool }) => {
+  const isClustered = useIsClustered();
+  const { defaultProject } = useAuth();
+  const { projectName } = useCurrentProject();
```
edlerd:
```suggestion
  const { projectName: currentProject } = useCurrentProject();
```
Maybe making the below code a bit clearer?

src/pages/overview/StoragePoolDetails.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { data: resources } = useStoragePoolResources(pool);
+  const volumeCount = getVolumesUsedByPool(pool).length;
+  const isMultiMemberLocalCluster =
```
edlerd:
```suggestion
  const hasClusterMemberSpecificSize =
```
maybe clearer?

src/pages/storage/StoragePoolSize.tsx
```diff
@@ -47,19 +52,24 @@ const StoragePoolSize: FC<Props> = ({
         const used = poolResource.space.used || 0;
         const resourceKey = poolResource.memberName || `resource-${index}`;
 
+        const usedPercentage = total > 0 ? (used * 100) / total : 0;
+
         if (!hasMeterBar) {
           return (
             <div key={resourceKey}>
-              {`${humanFileSize(used)} of ${humanFileSize(total)} used`}
+              {`${humanFileSize(used, precision)} of ${humanFileSize(total, precision)} used`}
```
```suggestion
              {`${humanFileSize(used, 0)} of ${humanFileSize(total, 0)} used`}
```
I think we can hardcode it for the pool size, no need to make this customizable in the other usages of this component.

edlerd (dismissed):
QA and Code LGTM, few nitpicks on the code

src/pages/overview/StoragePoolDetails.tsx
edlerd: Actually this should be positive: 
```
  const project =
    currentProject === ALL_PROJECTS
      ? defaultProject
      : currentProject;
```

src/pages/overview/StorageCard.tsx
```diff
@@ -1,8 +1,85 @@
...
+      <div className="storage-pools-container">
+        {pools.map((pool) => (
+          <StoragePoolDetails pool={pool} key={pool.name} />
+        ))}
+      </div>
+      <div className="card-footer">
+        <Link
+          to={`${ROOT_PATH}/ui/project/${encodeURIComponent(defaultProject)}/storage/pools`}
+        >
+          See all storage pools
```
kimanhou, line 78: @sophiecarroll Please let me know once you've settled on a copy to replace "See more"

edlerd (approved):

sophiecarroll:
Updated 
Screenshot 2026-08-28 at 11 08 51 AM
design based on discussions: @kimanhou 

(I think the only thing to change is bottom right text)

PR #2211, opened 2026-08-19
feat(Overview): Cluster card

kimanhou:
## Done

- Implement Clustering card

## QA

1. Run the LXD-UI:
    - Go to the Overview page and admire new Clustering card.

## Screenshots

Clustered
Screenshot from 2026-08-28 14-49-59

With 1 offline member
image

Unclustered
image

### DIFF
--- a/src/pages/overview/ClusteringCard.tsx
+++ b/src/pages/overview/ClusteringCard.tsx
@@ -1,0 +1,163 @@
+import type { FC } from "react";
+import { Link } from "react-router-dom";
+import { Card, Icon, MainTable, Spinner } from "@canonical/react-components";
+import ListPipe from "components/ListPipe";
+import { useClusterMembers } from "context/useClusterMembers";
+import { useIsClustered } from "context/useIsClustered";
+import { useSettings } from "context/useSettings";
+import { pluralize } from "util/helpers";
+import {
+  getClusterMemberStatusIconName,
+  getClusterLeader,
+  getClusterMemberStatusCounts,
+} from "util/clusterMember";
+import type { LxdClusterMember, LxdClusterMemberStatus } from "types/cluster";
+import { ROOT_PATH } from "util/rootPath";
+import ClusterMemberStatus from "pages/cluster/ClusterMemberStatus";
+import ClusterMemberMemoryUsage from "pages/cluster/ClusterMemberMemoryUsage";
+import ClusterMemberCpuUsage from "pages/cluster/ClusterMemberCpuUsage";
+import ClusterMemberStorageUsage from "pages/cluster/ClusterMemberStorageUsage";
+import ClusteringTotalResources from "pages/overview/ClusteringTotalResources";
+
+const ClusteringCard: FC = () => {
+  const isClustered = useIsClustered();
+  const { data: settings } = useSettings();
+  const serverName = settings?.environment?.server_name;
+  const { data: members = [], error, isLoading } = useClusterMembers();
+
+  const getStatusSummary = (members: LxdClusterMember[]) => {
+    if (!members.length) {
+      return null;
+    }
+
+    const countsPerStatus = getClusterMemberStatusCounts(members);
+    const listItems = Object.entries(countsPerStatus).map(([status, count]) => (
+      <>
+        <Icon
+          name={getClusterMemberStatusIconName(
+            status as LxdClusterMemberStatus,
+          )}
+        />
+        {count} {status}
+      </>
+    ));
+
+    return <ListPipe items={listItems} />;
+  };
+
+  const cardClassName = "overview-card clustering";
+  const cardTitle = (
+    <>
+      <Icon name="cluster-host" /> {isClustered ? "Clustering" : "Server"}
+    </>
+  );
+
+  if (isLoading) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <Spinner
+          className="u-loader"
+          text={`Loading ${isClustered ? "cluster" : "server"} details...`}
+        />
+      </Card>
+    );
+  }
+
+  if (error) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <Icon name="error" className="margin-right--large" /> Error while
+        loading cluster details: {error.message}
+      </Card>
+    );
+  }
+
+  const headers = [
+    { content: "Members" },
+    { content: "Status" },
+    { content: "Memory" },
+    { content: "CPU" },
+    { content: "Storage" },
+  ];
+
+  const rows = members.map((member) => {
+    const rowKey = member.server_name;
+
+    return {
+      key: rowKey,
+      name: rowKey,
+      className: "u-row",
+      columns: [
+        {
+          content: (
+            <Link
+              to={`${ROOT_PATH}/ui/cluster/member/${encodeURIComponent(
+                member.server_name,
+              )}`}
+              className="u-truncate"
+            >
+              {member.server_name}
+            </Link>
+          ),
+          role: "rowheader",
+          "aria-label": "Member",
+          title: `Cluster member ${member.server_name}`,
+        },
+        {
+          content: <ClusterMemberStatus member={member} />,
+          "aria-label": "Status",
+        },
+        {
+          content: <ClusterMemberMemoryUsage member={member} />,
+          "aria-label": "Memory",
+        },
+        {
+          content: <ClusterMemberCpuUsage member={member} />,
+          "aria-label": "CPU",
+        },
+        {
+          content: <ClusterMemberStorageUsage member={member} />,
+          "aria-label": "Storage",
+        },
+      ],
+    };
+  });
+
+  const subtitleItems = isClustered
+    ? [
+        `Current: ${serverName}`,
+        `${members.length} ${pluralize("member", members.length)}`,
+        `Leader: ${getClusterLeader(members)?.server_name ?? "-"}`,
+        getStatusSummary(members),
+      ]
+    : [`Server: ${serverName}`, "Standalone"];
+
+  return (
+    <Card className={cardClassName} title={cardTitle}>
+      <p className="u-text--muted u-text--small">
+        <ListPipe items={subtitleItems} />
+      </p>
+
+      <ClusteringTotalResources />
+
+      {isClustered && (
+        <MainTable
+          className="cluster-members-table"
+          aria-label="Cluster members"
+          headers={headers}
+          rows={rows}
+        />
+      )}
+
+      <div className="card-footer">
+        <Link
+          to={`${ROOT_PATH}/ui/cluster/${isClustered ? "members" : "server"}`}
+        >
+          See more
+        </Link>
+      </div>
+    </Card>
+  );
+};
+
+export default ClusteringCard;
--- a/src/pages/overview/ClusteringCard.tsx
+++ b/src/pages/overview/ClusteringCard.tsx
@@ -1,0 +1,187 @@
+import { Fragment, type FC } from "react";
+import { Link } from "react-router-dom";
+import {
+  Card,
+  Icon,
+  List,
+  MainTable,
+  Spinner,
+  TablePagination,
+} from "@canonical/react-components";
+import { useClusterMembers } from "context/useClusterMembers";
+import { useIsClustered } from "context/useIsClustered";
+import { pluralize } from "util/helpers";
+import {
+  getClusterMemberStatusIconName,
+  getClusterLeader,
+  getClusterMemberStatusCounts,
+} from "util/clusterMember";
+import type { LxdClusterMember, LxdClusterMemberStatus } from "types/cluster";
+import { ROOT_PATH } from "util/rootPath";
+import ClusterMemberStatus from "pages/cluster/ClusterMemberStatus";
+import ClusterMemberMemoryUsage from "pages/cluster/ClusterMemberMemoryUsage";
+import ClusterMemberCpuUsage from "pages/cluster/ClusterMemberCpuUsage";
+import ClusteringTotalResources from "pages/overview/ClusteringTotalResources";
+import ClusterMemberExplanationTooltip from "pages/cluster/ClusterMemberExplanationTooltip";
+import ServerExplanationTooltip from "pages/cluster/ServerExplanationTooltip";
+import { ITEMS_PER_PAGE } from "pages/overview/overviewConstants";
+
+const ClusteringCard: FC = () => {
+  const isClustered = useIsClustered();
+  const { data: members = [], error, isLoading } = useClusterMembers();
+
+  const getStatusSummary = (members: LxdClusterMember[]) => {
+    if (!members.length) {
+      return [];
+    }
+
+    const countsPerStatus = getClusterMemberStatusCounts(members);
+    const listItems = Object.entries(countsPerStatus).map(([status, count]) => (
+      <Fragment key={status}>
+        <Icon
+          name={getClusterMemberStatusIconName(
+            status as LxdClusterMemberStatus,
+          )}
+        />
+        {count} {status}
+      </Fragment>
+    ));
+
+    return listItems;
+  };
+
+  const cardClassName = "overview-card clustering";
+  const cardTitle = (
+    <>
+      <span className="overview-card-title">
+        <Icon name="cluster-host" /> {isClustered ? "Clustering" : "Server"}
+      </span>
+      {isClustered ? (
+        <ClusterMemberExplanationTooltip />
+      ) : (
+        <ServerExplanationTooltip />
+      )}
+    </>
+  );
+
+  if (isLoading) {
+    return (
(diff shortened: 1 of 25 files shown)


vosdev:
That looks awesome !! I really dig it!

Here are some Q&A notes:



Am I right to say that the cpu usage will show 0 on the next reading when I stop or reboot an instance?

What happens to these graphs when a cluster member goes offline?

Interesting way of showing CPU usage. Why not a %?

My lab cluster at home has 4 nodes with 14 cores each, with a total of 56. I have had production clusters with 200+ cores. How do these larger numbers render?

This gives more priority to https://github.com/canonical/lxd/issues/18120 because the memory data you're showing is currently incorrect.

`getHostCpuUsage` isn't host cpu, but a sum of all instance cpu usage. It does not include processes running on the host. If you have additional services running on your LXD host that may utilize a lot of CPU this can show a very wrong graph. 
I have a feature request for host cpu metrics: https://github.com/canonical/lxd/issues/15366. That will fix this issue and also make your calculations a lot easier.

kimanhou:
Hi @vosdev, how are you ?

Thank you for your feedback !

> Am I right to say that the cpu usage will show 0 on the next reading when I stop or reboot an instance?

Yes

> What happens to these graphs when a cluster member goes offline?

Here are screenshots before and after I manually stopped micro2. Storage takes way longer to be calculated than when all members are online, and it may need some rework because it currently shows 0 as storage (maybe it should show "-" like CPU and memory).
Screenshot from 2026-08-19 14-59-41
Screenshot from 2026-08-19 15-03-07

> Interesting way of showing CPU usage. Why not a %?

We’re actually actively discussing this! Backend team suggests that load average could be a better metric, as it bakes in time, to smooth out load spikes. 

> How do these larger numbers render?

The Card should have a fixed height and you would have to scroll down inside it. That's the plan, I haven't implemented it yet for the Clustering card but the design is still open to discussion. We will for sure need to accomodate for large numbers of cluster members.

Thanks again for the detailed feedback! We are tracking the issues you mentioned, as they should indeed make these calculations cleaner and easier.

vosdev:
> Hi @vosdev, how are you ?

Pretty good! My girlfriend is moving in with me later this year! Work has been slow this summer but I have an interview for a potential new project tomorrow so that may soon change. In the meantime I am doing some Q&A for my favorite open source project. 


> The Card should have a fixed height and you would have to scroll down inside it. That's the plan, I haven't implemented it yet for the Clustering card but the design is still open to discussion. We will for sure need to accomodate for large numbers of cluster members.

Oh I was referring to cluster members with a very large CPU count. Your example nodes have 4 cores, I have had some servers with 96 cores. Will they show a [] for each core?


> Thanks again for the detailed feedback! We are tracking the issues you mentioned, as they should indeed make these calculations cleaner and easier.

Awesome!

kimanhou:
Awesome news ! We love having you onboard.

> Your example nodes have 4 cores, I have had some servers with 96 cores. Will they show a [] for each core?

Ah, yes. It's a new "segmented" design I just implemented for the Meter. It's experimental and we need to discuss what to do when there are many segments. Probably, over X segments, we don't consider it segmented and it will be a continuous bar, X being around 20-30. I just updated the code with this last suggestion.

vosdev:
I see a slight css difference. The bars for memory and storage have a 1px border, the new cpu sectioned bar does not.

My personal preference is without border, I think it looks more modern

Kxiru:
I love the look of this so far and can't wait to see it land.
Would you consider adding one of the "Enable clustering" buttons in the non-clustered view? Similar to how we have them in non-clustered views of the clustering pages?
image

Wdyt, @sophiecarroll @kimanhou ? Or is it redundant as See More links to the members/server page

I also agree that while storage is being calculated, it should show "-" rather than 0.

Kxiru:
Is there yet a view for people who do not have permission to see cluster member information?

sophiecarroll:
@Kxiru 
> Is there yet a view for people who do not have permission to see cluster member information?

Yup we display a permissions error and surface permissions card if not total admin

> Would you consider adding one of the "Enable clustering" buttons in the non-clustered view? Similar to how we have them in non-clustered views of the clustering pages?

We're not putting any action buttons on the overview because it is read only data for orientation, and then see more takes you to the pages where you can take actions and manage

sophiecarroll:
Some comments: Looks great so far! 

The reason I put the CPU bar into whole number sections is because I was told cores can't be partially used, so either the section is used or unused in whole numbers. How is it possible that a single core in the example screenshot is partially used?

Make sure text aligns with given designs: under Total CPU it should say "# of # cores used (#%)" and under Total memory it should say "#GiB of #GiB used (#%)". Right now CPU and Memory are missing percentages. Also make sure you include "allocated" as per the design in the memory and CPU status of each member.

kimanhou:
@sophiecarroll 

> How is it possible that a single core in the example screenshot is partially used?

See David's comment on the spec: https://docs.google.com/document/d/1Myu3NOPXoLd_80-X_6tZnNOPNMS9gwX_ljMm71aCgAs/edit?disco=AAACFwgFV6Y

sophiecarroll:
Updated card after all discussed changes:

Screenshot 2026-08-28 at 4 00 24 PM

@kimanhou As of august 31: make sure all text matches the design and show number of members offline (lmk if this is already done, I'm going off the screenshot because the demo is giving me 404 error)

sophiecarroll:
@kimanhou 

Looking great! Pagination needs an update: the one in the screenshots uses vanilla, pragma pagination looks different (used in my visual prototype, includes total number as well as pages)

Screenshot 2026-08-28 at 3 55 02 PM

Also memory needs to have % to match text on design (x of x used (%))

[EDIT] the design pragma version is ready but the engineering is not so we'll use vanilla for now

src/pages/overview/ClusteringTotalResources.tsx
```diff
@@ -0,0 +1,135 @@
...
+        queryKeys.cluster,
+        queryKeys.members,
+        member.server_name,
+        queryKeys.state,
+      ],
+      queryFn: async () => fetchClusterMemberState(member.server_name),
+      enabled: Boolean(member.server_name),
+      refetchInterval: 5000,
+    })),
+  });
```
edlerd: This does not scale on large clusters. Though I have no direct suggestion for improvement.

Also, do we consider that members can be down, and is the data fetching resilient to some of the members being unavailable? We should not block the data showing if some members are unresponsive. Especially if we already know their status is "wired" from the cluster member list query.

src/types/server.d.ts
```diff
@@ -23,6 +23,7 @@ export interface LxdSettings {
     server_clustered: boolean;
     storage_supported_drivers: SupportedStorageDriver[];
     backup_metadata_version_range?: number[];
+    server_name?: string;
```
edlerd, line 26:
```suggestion
```
I don't think this exists yet.

src/util/resourceDetails.tsx
```diff
@@ -94,3 +94,18 @@ export const extractResourceDetailsFromUrl = (
...
+  if (cores % 1 === 0) return cores.toString();
+  // Otherwise format to 1 decimal place
+  return cores.toFixed(1);
+};
+
+export const getCpuText = (percentage: number, isShort?: boolean): string => {
+  if (isShort) {
+    return `Load average (${percentage.toFixed(1)}%)`;
+  }
+  return `Load average in the last minute (${percentage.toFixed(1)}%)`;
```
edlerd:
```suggestion
    return `Load average (${percentage.toFixed(0)}%)`;
  }
  return `Load average in the last minute (${percentage.toFixed(0)}%)`;
```
I think less precision might be easier to read, wdyt?

src/util/resourceDetails.tsx
```diff
@@ -94,3 +94,18 @@ export const extractResourceDetailsFromUrl = (
...
   return resourceDetail;
 };
+
+export const formatCoreCount = (cores: number): string => {
+  if (cores === 0) return "0";
+  // If it's a whole number, don't show decimals (e.g. 2 instead of 2.0)
+  if (cores % 1 === 0) return cores.toString();
+  // Otherwise format to 1 decimal place
+  return cores.toFixed(1);
+};
```
edlerd: Do we really need this? Cores should always be a whole number, no?

edlerd (commented):
This is great, I really like it!

Few thoughts on the changes from a quick QA and code review.

src/types/server.d.ts
kimanhou: It does and it works ! But @sophiecarroll removed it from the designs so it's useless now. I will remove it.

```
  const server_name = settings?.environment?.server_name ?? "";
```

image

src/pages/overview/ClusteringTotalResources.tsx
kimanhou: Mmh indeed...
image

But then what do we show if a member is offline ? The total from the other members, excluding the offline one ?

src/util/resourceDetails.tsx
kimanhou: It was for a previous implementation, it's useless now. I will remove it.

src/types/server.d.ts
edlerd: Oh I see, this is the server that is handling the request. So it is the cluster member name, but I think we should not surface it to the user. So let's keep it in the type even though we are not displaying it. I agree with removing it from the design.

src/pages/overview/ClusteringTotalResources.tsx
edlerd: Good question. I think partial information would be ok, but we might need a hint that this is missing information from some members.

kimanhou: I did something

kimanhou: Is it too much ?

image

edlerd: I don't know, it could be okay from my side. It is a design question, though.

@sophiecarroll what do you think about the case with partial data and the suggested notification box in the card?

sophiecarroll: @edlerd @kimanhou Thanks for tagging me David! I think the "-" is a good solution for data unavailable.

Is it common to have members offline? If it's common to have members offline, I can think of a smaller solution to this message in order to not always have a large warnings in the block, but if it is infrequent this solution looks good. Brief copy change: Can the bottom line say "Resource usage includes data from online members only." ? We have the online members vs total members info at the top, and I think it's helpful to give what type member is included.

Also, I wanted to double check that when you click on memory, CPU, or status, it can sort by those categories. Sorry I can't see the demo but is that the case?

kimanhou: > when you click on memory, CPU, or status, it can sort by those categories.

Nope, that's a new requirement to me. Could you please provide more details ?

- When you click on Memory, what do you expect: sort by percentage or the number of GiB used or the total number of GiB ?
- Same question for CPU.
- Status would be sorted alphabetically, correct ?
- You don't want sorting by member's name ?

edlerd: > Is it common to have members offline? 

In most cases, no, not common. But at the same time likely to happen during maintenance or in outages. And in the scenarios it is *very* likely an operator will open the UI to do diagnostics or do the maintenance. And I was having the larger clusters in mind, where the more members there are, the more likely it is that at least one of them is unresponsive.

sophiecarroll: > when you click on memory, CPU, or status, it can sort by those categories.

> Nope, that's a new requirement to me. Could you please provide more details ?

Alright, I had represented that in my design with the downward chevron next to memory, but I can understand if it wasn't clear enough–memory: sort by percentage, CPU: sort by percentage, status: offline/online. 

> In most cases, no, not common. But at the same time likely to happen during maintenance or in outages. And in the scenarios it is very likely an operator will open the UI to do diagnostics or do the maintenance. And I was having the larger clusters in mind, where the more members there are, the more likely it is that at least one of them is unresponsive.

Alright, it sounds like having members offline is fairly common then. I'm going to think about this for a second and get back to you within a few hours, I'm having some conversations about it.

[EDIT] @kimanhou @edlerd I think I've found a smooth solution with some help from Sophie (content designer), we are refining the wording now. Having members offline could be fairly common and might not severely impede the user, so instead of a large possibly persistent warning, we can add a parenthetical after the subheading "# offline" to say something like "# online • # offline (not included in resource usage)". I will let you know as soon as we decide on the final wording

[EDIT] @kimanhou @edlerd Sophie and I decided on "# online • # offline (not included in resource usage)" in the subheading (so swapping the order of online and offline, and adding the parenthetical).
Screenshot 2026-09-02 at 9 41 46 AM

kimanhou: @sophiecarroll 

> You don't want sorting by member's name ?

sophiecarroll: This didn't come up as useful info so I didn't put it, but now that you bring it up, it would be consistent with our other table behavior through LXD so yes, let's be able to sort by name.

kimanhou: Understood.

https://warthogs.atlassian.net/browse/WD-38691

src/pages/overview/ClusteringCard.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { ROOT_PATH } from "util/rootPath";
+import ClusterMemberStatus from "pages/cluster/ClusterMemberStatus";
+import ClusterMemberMemoryUsage from "pages/cluster/ClusterMemberMemoryUsage";
+import ClusterMemberCpuUsage from "pages/cluster/ClusterMemberCpuUsage";
+import ClusteringTotalResources from "pages/overview/ClusteringTotalResources";
+import ClusterMemberExplanationTooltip from "pages/cluster/ClusterMemberExplanationTooltip";
+import ServerExplanationTooltip from "pages/cluster/ServerExplanationTooltip";
+import { ITEMS_PER_PAGE } from "pages/overview/overviewConstants";
+
+const ClusteringCard: FC = () => {
```
edlerd, line 29: The card looks pretty sad on an unclustered environment. Should we hide it there instead?

Image

src/pages/overview/ClusteringTotalResources.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { fetchClusterMemberState } from "api/cluster-members";
+import classnames from "classnames";
+import Meter from "components/Meter";
+import { useClusterMembers } from "context/useClusterMembers";
+import { useIsClustered } from "context/useIsClustered";
+import { useResources } from "context/useResources";
+import { queryKeys } from "util/queryKeys";
+import { getCpuText, getMemoryText } from "util/resourceDetails";
+
+const ClusteringTotalResources: FC = () => {
```
edlerd, line 13: Depending on the other question, we might be able to reduce complexity in this component. Because we could assume the environment is always clustered.

edlerd (commented):
Mostly LGTM, one question on the non clustered backend and potential simplification.

src/pages/overview/ClusteringCard.tsx
kimanhou: Total CPU should not be empty, I'm waiting for backend to tell us what endpoint we can use. 
But I'm okay if we don't have a Clustering card if the environment is unclustered.

@sophiecarroll Wdyt ?

edlerd: I think with the total CPU present, it would be interesting for an unclustered environment to have only those two metrics. 
Though the card looks very empty right now, maybe there is other information to display or can we make the card smaller?

sophiecarroll: I think we should keep the server card, as several users mentioned wanting to see cpu and memory information, and there was one user who cited her computer crashing because the cpu usage of LXD was too high and she didn't know. There was also a demand for memory and cpu per instances, which could be an interesting thing to explore next cycle, maybe having a list like we do with members.

kimanhou: > can we make the card smaller?

The card will take what it needs in terms of space once https://github.com/canonical/lxd-ui/pull/2224 is merged. It's part of the page layout ticket, I know it's a mess.

src/pages/overview/ClusteringTotalResources.tsx
kimanhou: I hesitated of having two components: one for clustered and one for unclustered. Do you think it would be simpler to read ?

edlerd: No strong preference in either direction of split or combined component. I was slightly confused by the naming, though. I think it can stay as is, no blocker from my side either.

edlerd (approved):
LGTM, open questions above resolved.

src/pages/overview/ClusteringTotalResources.tsx
kimanhou: Let's keep it as it then.

PR #2105, opened 2026-06-27
Feat: mobile config navigation

vosdev:
## Done

- Display instance configuration side-navigation as a dropdown on mobile

Fixes instance configuration navigation on mobile

## QA

1. Run the LXD-UI:
    - On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @Kxiru or @edlerd for access.
    - With a local copy of this branch, [build and run as described in the docs](https://github.com/canonical/lxd-ui/blob/main/CONTRIBUTING.md#setting-up-for-development).
2. Perform the following QA steps:
    - Navigate the configuration page for both create and edit instances on **mobile**, **small**, **medium** and **large** views
    - The devices collapse should not work when in dropdown view.

## Screenshots
image

image

PR #2212, opened 2026-08-19
feat(Configuration): Add initial user.key support

### LINKED ISSUE #1465: Feature Request: Support for user keys

vosdev:
## Done

- Allow configuration of user.* keys on instances and profiles
- Allow overriding inherited user keys
- Support large and smaller screens for viewing and editing

Partly fixes https://github.com/canonical/lxd-ui/issues/1465

(Claude Opus helped me with this feature)

## QA

1. Run the LXD-UI:

## Screenshots

# Large screen
### Create / Edit on the configuration page

image

image

### View on the instance overview

image

# small screen:

### Create / Edit on the configuration page

image

image

### View on the instance overview

image

### DIFF
--- a/src/components/UserKeyChips.tsx
+++ b/src/components/UserKeyChips.tsx
@@ -1,0 +1,28 @@
+import type { FC } from "react";
+import ExpandableList from "components/ExpandableList";
+import UserKeyChip from "components/UserKeyChip";
+import type { UserKey } from "types/forms/instanceAndProfile";
+
+interface Props {
+  userKeys: UserKey[];
+  displayCount?: number;
+}
+
+const UserKeyChips: FC<Props> = ({ userKeys, displayCount = 5 }) => {
+  if (userKeys.length === 0) {
+    return <>-</>;
+  }
+
+  return (
+    <div className="user-key-chips">
+      <ExpandableList
+        displayCount={displayCount}
+        items={userKeys.map((userKey) => (
+          <UserKeyChip key={userKey.key} userKey={userKey} />
+        ))}
+      />
+    </div>
+  );
+};
+
+export default UserKeyChips;
--- a/src/components/UserKeyListTable.tsx
+++ b/src/components/UserKeyListTable.tsx
@@ -1,0 +1,54 @@
+import type { FC } from "react";
+import { MainTable } from "@canonical/react-components";
+import type { UserKey } from "types/forms/instanceAndProfile";
+
+interface Props {
+  userKeys: UserKey[];
+}
+
+const UserKeyListTable: FC<Props> = ({ userKeys }) => {
+  if (userKeys.length === 0) {
+    return <>-</>;
+  }
+
+  const headers = [
+    { content: "Key", sortKey: "key", className: "u-text--muted key" },
+    { content: "Value", sortKey: "value", className: "u-text--muted value" },
+  ];
+
+  const rows = userKeys.map(({ key, value }) => {
+    return {
+      key,
+      className: "u-row",
+      columns: [
+        {
+          content: key,
+          role: "rowheader",
+          "aria-label": "Key",
+          className: "key",
+        },
+        {
+          content: value,
+          role: "cell",
+          "aria-label": "Value",
+          className: "value",
+        },
+      ],
+      sortData: {
+        key: key.toLowerCase(),
+        value: value.toLowerCase(),
+      },
+    };
+  });
+
+  return (
+    <MainTable
+      className="user-key-list-table"
+      headers={headers}
+      rows={rows}
+      sortable
+    />
+  );
+};
+
+export default UserKeyListTable;
--- a/src/components/forms/UserKeysForm.tsx
+++ b/src/components/forms/UserKeysForm.tsx
@@ -1,0 +1,428 @@
+import { useState, type FC, type KeyboardEvent, type ReactNode } from "react";
+import {
+  Button,
+  Icon,
+  Input,
+  MainTable,
+  PrefixedInput,
+  Tooltip,
+} from "@canonical/react-components";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import { useParams } from "react-router-dom";
+import type { InstanceAndProfileFormikProps } from "types/forms/instanceAndProfileFormProps";
+import type { UserKey } from "types/forms/instanceAndProfile";
+import { ensureEditMode } from "util/editMode";
+import { generateUUID } from "util/helpers";
+import {
+  getInheritedUserKeys,
+  getUserKeyNameError,
+  visibleUserKeys,
+} from "util/userKeys";
+import { useProfiles } from "context/useProfiles";
+import ProfileRichChip from "pages/profiles/ProfileRichChip";
+
+interface UserKeyDraft {
+  id: string;
+  originalKey?: string;
+  overrides?: string;
+  key: string;
+  value: string;
+}
+
+const draftAnchor = (draft: UserKeyDraft): string | undefined => {
+  return draft.originalKey ?? draft.overrides;
+};
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  disabledReason?: string;
+}
+
+const UserKeysForm: FC<Props> = ({ formik, disabledReason }) => {
+  const { project } = useParams<{ project: string }>();
+  const { data: profiles = [] } = useProfiles(project ?? "");
+  const [drafts, setDrafts] = useState<UserKeyDraft[]>([]);
+
+  const allKeys = formik.values.user_keys ?? [];
+  const localKeys = visibleUserKeys(allKeys);
+  const inheritedKeys = getInheritedUserKeys(formik.values, profiles);
+  const hasInherited = inheritedKeys.length > 0;
+
+  const setUserKeys = (userKeys: UserKey[]) => {
+    ensureEditMode(formik);
+    const internalKeys = allKeys.filter(
+      (item) => !localKeys.some((visible) => visible.key === item.key),
+    );
+    formik.setFieldValue("user_keys", internalKeys.concat(userKeys));
+  };
+
+  const addDraft = (draft: Omit<UserKeyDraft, "id">) => {
+    setDrafts(drafts.concat({ ...draft, id: generateUUID() }));
+  };
+
+  const updateDraft = (id: string, changes: Partial<UserKeyDraft>) => {
+    setDrafts(
+      drafts.map((draft) =>
+        draft.id === id ? { ...draft, ...changes } : draft,
+      ),
+    );
+  };
+
+  const discardDraft = (id: string) => {
+    setDrafts(drafts.filter((draft) => draft.id !== id));
+  };
+
+  const commitDraft = (draft: UserKeyDraft) => {
+    const userKey = { key: draft.key.trim(), value: draft.value };
+    setUserKeys(
+      draft.originalKey
+        ? localKeys.map((item) =>
+            item.key === draft.originalKey ? userKey : item,
+          )
+        : localKeys.concat(userKey),
+    );
+    discardDraft(draft.id);
+  };
+
+  const removeKey = (key: string) => {
+    setUserKeys(localKeys.filter((item) => item.key !== key));
+  };
+
+  const profileHeader = (
+    <>
+      <span className="u-hide--large">Profile</span>
+      <span className="u-hide--small u-hide--medium">
+        Inherited from profile
+      </span>
+    </>
+  );
+
+  const headers = [
+    { content: "Key", className: "key" },
+    { content: "Value", className: "value" },
+    ...(hasInherited ? [{ content: profileHeader, className: "profile" }] : []),
+    { "aria-label": "Actions", className: "actions u-align--right" },
+  ];
+
+  const profileColumn = (content: ReactNode = "") => {
+    return {
+      content,
+      role: "cell",
+      "aria-label": "Profile",
+      className: "profile",
+    };
+  };
+
+  const actionColumn = (content: ReactNode) => {
+    return {
+      content,
+      role: "cell",
+      "aria-label": "Actions",
+      className: "actions u-align--right",
+    };
+  };
+
+  const iconButton = (
+    label: string,
+    icon: string,
+    onClick: () => void,
+    key: string,
+  ) => {
+    return (
+      <Button
+        onClick={onClick}
+        dense
+        type="button"
+        hasIcon
+        appearance="base"
+        disabled={!!disabledReason}
+        title={disabledReason ?? label}
+        aria-label={`${label} ${key}`}
+      >
+        <Icon name={icon} />
+      </Button>
+    );
+  };
+
+  const draftRow = (draft: UserKeyDraft): MainTableRow => {
+    const takenKeys = localKeys
+      .filter((item) => item.key !== draft.originalKey)
+      .map((item) => item.key);
+    const keyError = getUserKeyNameError(draft.key.trim(), takenKeys);
+    const isValid = keyError === undefined && draft.value.length > 0;
+
+    const onKeyDown = (e: KeyboardEvent) => {
+      if (e.key === "Enter") {
+        e.preventDefault();
+        if (isValid) {
+          commitDraft(draft);
+        }
+      }
+      if (e.key === "Escape") {
+        e.preventDefault();
+        discardDraft(draft.id);
+      }
+    };
+
+    const isEdit = draft.originalKey !== undefined;
+
+    return {
+      key: draft.id,
+      columns: [
+        {
+          content: (
+            <div className="user-key-draft">
+              <div className="user-key-draft-fields">
+                <PrefixedInput
+                  id={`user-key-name-${draft.id}`}
+                  aria-label="User key"
+                  placeholder="key"
+                  immutableText="user."
+                  value={draft.key}
+                  error={draft.key.length > 0 ? keyError : undefined}
+                  onChange={(e) => {
+                    updateDraft(draft.id, { key: e.target.value });
+                  }}
+                  onKeyDown={onKeyDown}
+                  wrapperClassName="user-key-draft-field"
+                  autoFocus={!isEdit}
+                />
+                <Input
+                  id={`user-key-value-${draft.id}`}
+                  aria-label="User key value"
+                  placeholder="Value"
+                  type="text"
+                  value={draft.value}
+                  onChange={(e) => {
+                    updateDraft(draft.id, { value: e.target.value });
+                  }}
+                  onKeyDown={onKeyDown}
+                  wrapperClassName="user-key-draft-field"
+                  autoFocus={isEdit}
+                />
+              </div>
+              <div className="user-key-draft-actions">
+                <Button
+                  onClick={() => {
+                    discardDraft(draft.id);
+                  }}
+                  dense
+                  type="button"
+                  appearance="base"
+                >
+                  Cancel
+                </Button>
+                <Button
+                  onClick={() => {
+                    commitDraft(draft);
+                  }}
+                  dense
+                  type="button"
+                  appearance="positive"
+                  disabled={!isValid}
+                >
+                  Save
+                </Button>
+                {isEdit && (
+                  <Button
+                    onClick={() => {
+                      discardDraft(draft.id);
+                      removeKey(draft.originalKey as string);
+                    }}
+                    dense
+                    type="button"
+                    appearance="base"
+                    hasIcon
+                    aria-label={`Delete ${draft.originalKey}`}
+                  >
+                    <Icon name="delete" />
+                    <span>Delete</span>
+                  </Button>
+                )}
+              </div>
+            </div>
+          ),
+          role: "rowheader",
+          "aria-label": "Key",
+          className: "user-key-draft-cell",
+          colSpan: hasInherited ? 4 : 3,
+        },
+      ],
+    };
+  };
+
+  const inheritedRow = (
+    key: string,
+    value: string,
+    source: string,
+  ): MainTableRow => {
+    const isOverridden = localKeys.some((item) => item.key === key);
+    const isBeingOverridden = drafts.some((draft) => draft.overrides === key);
+
+    return {
+      key: `inherited-${key}`,
+      className: isOverridden ? "u-text--muted u-text--line-through" : "",
+      columns: [
+        {
+          content: key,
+          role: "rowheader",
+          "aria-label": "Key",
+          className: "key u-truncate",
+          title: key,
+        },
+        {
+          content: value,
+          title: value,
+          role: "cell",
+          "aria-label": "Value",
+          className: "value u-truncate",
+        },
+        profileColumn(
+          <ProfileRichChip
+            profileName={source}
+            projectName={project ?? ""}
+            className="force-truncate"
+            disabled={isOverridden}
+          />,
+        ),
+        actionColumn(
+          isOverridden
+            ? iconButton(
+                "Revert to the profile value of",
+                "restart",
+                () => {
+                  removeKey(key);
+                },
+                key,
+              )
+            : !isBeingOverridden &&
+                iconButton(
+                  "Override",
+                  "edit",
+                  () => {
+                    ensureEditMode(formik);
+                    addDraft({ overrides: key, key, value });
+                  },
+                  key,
+                ),
+        ),
+      ],
+    };
+  };
+
+  const localRow = (userKey: UserKey): MainTableRow => {
+    return {
+      key: `local-${userKey.key}`,
+      columns: [
+        {
+          content: userKey.key,
+          role: "rowheader",
+          "aria-label": "Key",
+          className: "key u-truncate",
+          title: userKey.key,
+        },
+        {
+          content: userKey.value,
+          title: userKey.value,
+          role: "cell",
+          "aria-label": "Value",
+          className: "value u-truncate",
+        },
+        ...(hasInherited ? [profileColumn()] : []),
+        actionColumn(
+          iconButton(
+            "Edit",
+            "edit",
+            () => {
+              ensureEditMode(formik);
+              addDraft({
+                originalKey: userKey.key,
+                key: userKey.key,
+                value: userKey.value,
+              });
+            },
+            userKey.key,
+          ),
+        ),
+      ],
+    };
+  };
+
+  const sortedKeys = [
+    ...new Set(
+      inheritedKeys
+        .map((item) => item.userKey.key)
+        .concat(localKeys.map((item) => item.key)),
+    ),
+  ].sort();
+
+  const rows = sortedKeys
+    .flatMap((key) => {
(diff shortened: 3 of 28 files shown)


src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,428 @@
...
+const draftAnchor = (draft: UserKeyDraft): string | undefined => {
+  return draft.originalKey ?? draft.overrides;
+};
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  disabledReason?: string;
+}
+
+const UserKeysForm: FC<Props> = ({ formik, disabledReason }) => {
```
edlerd, line 30: I think the edit flow is slightly confusing and can be simplified. In the edit mode, we should make all user keys editable, rendering the inputs for all existing user keys. Then we can remove the inline "Cancel" and "Save" buttons.

The problem is the "Save" button doesn't persist yet. It only updates the form state. We still have to submit the instance form. Likewise, if we have started editing a user key and not hit "save" on the row, then update other fields such as the description. On saving the instance, the user key update is not persisted as that row is still in an edit state itself.

Instead we can have the read mode as it is today. But enabling edit mode will render all entries in the user keys table as inputs with only the "remove" button next to the row. Changing any input will immediately update the form state and it will be persisted with the whole instance form. So we have:

Read mode:

key | value | pencil-icon

On clicking the pencil icon, all user key rows turn into edit mode with input fields for key and value. The value input for the row that we clicked the edit-button for should receive focus.

Edit mode:

key-input | value-input | trash-icon + "remove" copy

Lastly, the headings in the table should also align in edit mode with the columns.

Image

src/util/userKeys.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  return inherited;
+};
+
+export const getEffectiveUserKeys = (instance: LxdInstance): UserKey[] => {
+  const config = instance.expanded_config ?? instance.config;
+
+  return Object.keys(config)
+    .filter((key) => key.startsWith(USER_KEY_PREFIX) && !isInternalUserKey(key))
```
edlerd:
```suggestion
    .filter((key) => key.startsWith(USER_KEY_PREFIX))
```
Let's include the internal keys in all places. No need to hide them in my opinion.

src/components/UserKeyChips.tsx
```diff
@@ -0,0 +1,28 @@
...
+import ExpandableList from "components/ExpandableList";
+import UserKeyChip from "components/UserKeyChip";
+import type { UserKey } from "types/forms/instanceAndProfile";
+
+interface Props {
+  userKeys: UserKey[];
+  displayCount?: number;
+}
+
+const UserKeyChips: FC<Props> = ({ userKeys, displayCount = 5 }) => {
```
edlerd: I am not so sure if we should have the user keys in the list and side panel.

For one, there is the concern about chips being perceived as interactive in the rest of the UI, and here they are purely informative.

For two, the truncation behavior is not so clean in those places.

I don't have a clear conclusion on the best way forward. This is a design problem after all.

Truncated:

Image

Expanded:

Image

src/pages/profiles/ProfileDetailOverview.tsx
```diff
@@ -65,6 +67,14 @@ const ProfileDetailOverview: FC<Props> = ({ profile }) => {
                 <th className="u-text--muted">Description</th>
                 <td>{profile.description ? profile.description : "-"}</td>
               </tr>
+              <tr>
+                <th className="u-text--muted">User keys</th>
+                <td>
+                  <UserKeyChips
+                    userKeys={visibleUserKeys(parseUserKeys(profile))}
+                  />
```
edlerd: I think the user keys should be on the overview page for profiles and instances. But probably not so prominent, maybe below profiles and with a table. Similar to how it is on the configuration tab in read mode. That way parsing the "Which keys exist" and "What are their values" becomes way quicker. As the condensed chip format makes the mental load on parsing it a bit cumbersome.

Maybe something like this

Image

edlerd (commented):
Thanks for raising the PR! This makes it much easier to discuss and explore the solution space. Continuing on the [discussion from the issue](https://github.com/canonical/lxd-ui/issues/1465). A few detailed thoughts on the proposed changes.

src/components/forms/UserKeysForm.tsx
vosdev: That is actually a great suggestion, and it makes sense with the double "save" button. 

Sometimes all you need is a second set of eyes.

src/components/UserKeyChips.tsx
vosdev: Oh i tested combinations of long values but nothing showed up like that. I will have to put in some more hours there.

I have already disabled user keys in the instance list by default because not everyone uses them. It is optional to turn it on. 

Two friends of mine run large LXD clusters in production, one at a university and the other at a flower business. And both rely heavily on user keys so they were happy with my PR/suggestion. I will ask them for some example values that they use in prod, to get some more variety in my testing.

If Chips end up being the correct solution for user keys I can put some time into fixing these edge cases.

src/pages/profiles/ProfileDetailOverview.tsx
vosdev: Ahhh that is an exact copy of my initial suggestion in https://github.com/canonical/lxd-ui/issues/1465

That can be a good solution for the instance overview.

Downside is that it takes up a LOT more space, especially for smaller keys. But that downside is an upside for the larger values.

For example I was just told about the university using user.reverse-proxy=sub.domain.com/path/to/app to dynamically configure a central reverse proxy for students' VMs.

That is the kind of stuff that you want displayed properly.

vosdev: image

src/components/forms/UserKeysForm.tsx
vosdev: image

image


And buttons without text on small screens:

image

src/components/UserKeyChips.tsx
vosdev: I removed it from the instance list for now.

For the instance summary, I was thinking a table similar to the snapshot table. We have a snapshot/date table at the bottom, user keys will probably have enough space if we do show them with a similar table

edlerd: > For the instance summary, I was thinking a table similar to the snapshot table. We have a snapshot/date table at the bottom, user keys will probably have enough space if we do show them with a similar table

Thanks for the update. Yes I think that can work nicely :+1:

vosdev:
When editting a key, I currently show `user.environment` where `user.` is static and cannot be changed. I'm thinking it looks less crowded (especially on smaller screens) if we do not show the `user.` part.

src/pages/profiles/ProfileDetailOverview.tsx
edlerd: Nice :+1: 
It think it would be even nicer when the column width for "Key" was aligned with the "Name" from the profiles section above. 

We can probably add the user key table to this block in css:
```
  .profiles,
  .user-keys {
    th:nth-child(1) {
      width: 30%;
    }

    th:first-of-type,
    td:first-of-type {
      padding-left: 0;
    }
  }
```

vosdev: Yes, very clean!

image

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,365 @@
...
+        Inherited from profile
+      </span>
+    </>
+  );
+
+  const headers = [
+    { content: "Key", className: "key" },
+    { content: "Value", className: "value" },
+    ...(hasInherited ? [{ content: profileHeader, className: "profile" }] : []),
+    { "aria-label": "Actions", className: "actions u-align--right" },
```
edlerd:
```suggestion
    { "aria-label": "Actions", className: "actions" },
```
Let's left align the icons especially in case there are inherited and local rows.

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,365 @@
...
+      className: "profile",
+    };
+  };
+
+  const actionColumn = (content: ReactNode) => {
+    return {
+      content,
+      role: "cell",
+      "aria-label": "Actions",
+      className: "actions u-align--right",
```
edlerd:
```suggestion
      className: "actions",
```

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,365 @@
...
+        disabled={!!disabledReason}
+        title={disabledReason ?? label}
+        aria-label={`${label} ${target}`}
+      >
+        <Icon name={icon} />
+        {!isReadOnly && (
+          <span className="u-hide--small u-hide--medium">{label}</span>
+        )}
+      </Button>
+    );
```
edlerd, line 105:
```suggestion
    return (
      <Button
        onClick={onClick}
        type="button"
        hasIcon
        appearance="base"
        disabled={!!disabledReason}
        title={disabledReason ?? label}
        aria-label={`${label} ${target}`}
        className="u-no-margin--bottom"
      >
        <Icon name={icon} />
        <span className="u-hide--small u-hide--medium">{label}</span>
      </Button>
    );
```
We can use a more standard button here. Also, on second look, I think the pencil button can be with a label as well.

src/sass/_user_keys.scss
```diff
@@ -0,0 +1,71 @@
...
+      @include large {
+        width: 8rem;
+
+        .p-button--base {
+          padding-left: $sph--small;
+          padding-right: $sph--small;
+          width: auto;
+        }
+      }
+    }
```
edlerd, line 42:
```suggestion
    .actions {
      width: 10rem;
    }
```
Let's give this more width and keep the spacing closer to the standard.

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,365 @@
...
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  disabledReason?: string;
+}
+
+const UserKeysForm: FC<Props> = ({ formik, disabledReason }) => {
+  const { project } = useParams<{ project: string }>();
+  const { data: profiles = [] } = useProfiles(project ?? "");
+  const [focusIndex, setFocusIndex] = useState(-1);
+  const [pinnedCount, setPinnedCount] = useState(0);
```
edlerd:
```suggestion
```
I don't understand why we need this. Isn't the form state with user keys enough to keep track of the edits?

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  disabledReason?: string;
+}
+
+const UserKeysForm: FC<Props> = ({ formik, disabledReason }) => {
+  const { project } = useParams<{ project: string }>();
+  const { data: profiles = [] } = useProfiles(project ?? "");
+  const [focusIndex, setFocusIndex] = useState(-1);
```
edlerd:
```suggestion
```
The focus currently doesn't work. I think we can use the `focusField` helper directly and don't need the extra state for it.

src/pages/profiles/forms/ProfileDetailsForm.tsx
```diff
@@ -21,6 +22,9 @@ const ProfileDetailsForm: FC<Props> = ({ formik, isEdit, project }) => {
   const helpText = !isDefaultProfile
     ? "Click the name in the header to rename the profile."
     : "Default profile cannot be renamed.";
+  const missingNameReason = formik.values.name
+    ? undefined
+    : "Please enter a name before adding custom configuration";
```
edlerd, line 27:
```suggestion
```
The name is optional. LXD will generate a name if left empty.

src/util/userKeys.tsx
```diff
@@ -0,0 +1,119 @@
...
+    return "Key must not contain whitespace";
+  }
+  if (name.includes("=")) {
+    return "Key must not contain an equals sign";
+  }
+  if (existingKeys.includes(name)) {
+    return "This key already exists";
+  }
+  return undefined;
+};
```
edlerd, line 156: There is one flow that might trap a user in an invalid state:
1. Add user key
2. Do not fill the key name at all.
3. Scroll or navigate elsewhere in the configuration

At this stage the form cannot be submitted, and it is not exactly clear why the submit button is disabled. I wonder if we should allow submission and filter out empty keys similar to how keys with empty values are omitted on payload creation.

edlerd (commented):
Great progress on this one, thanks for the iteration!

Did a detailed QA and code review. See comments below.

src/pages/profiles/forms/ProfileDetailsForm.tsx
vosdev: That is for the profile creation page. All the other configuration (side nav) is also disabled with the same reason, until a name is filled in.

image
image


This disable was forgotten for the SSH key when that was added, so I added the same feature for SSH and User key button. It was a single line so I did not make a separate issue for it.

src/util/userKeys.tsx
vosdev: Once you start typing a key value (and leave the key empty), an error underneath the key pops up immediately.

image

The submit button is indeed disabled in this state. I am able to navigate away to a Devices or Snapshot page and go back to Main Configuration, the error is still there.

src/pages/profiles/forms/ProfileDetailsForm.tsx
vosdev: The instance creation page will enable the button after an image is selected, just like all other configuration options

src/components/forms/UserKeysForm.tsx
vosdev: The table renders local and inherited keys **merged** into one alphabetical sorted list. Form state keeps the keys in a stable order, but a row's position among inherited rows comes from the key name.

When adding a new key, typing the first letter will make it jump to its "correct" position, but I don't think it is very user friendly to have your inputs jump around the screen. So this pins them to the bottom until the state is saved.

You will only see this work on an instance with at least one inherited key.

This was a problem that popped up because I wanted to merge the inherited and local keys.

Before they were merged, they would order like this:
```
user.a | value1 | (profile X)
user.x | profile-value | (profile Y) (overridden)
user.ansible | true
user.owner | David
user.x | local-value
```
I wanted the local user keys directly below the profile key that it was overriding, hence the merged alphabetical sort :)

A merged & alphabetically sorted list looks like this:
```
user.a | value1 | (profile X)
user.ansible | true
user.owner | David
user.x | profile-value | (profile Y) (overridden)
user.x | local-value
```

or a real screenshot:
image

src/components/UserKeyChips.tsx
vosdev: I will put it in a separate PR once this is merged

src/components/forms/UserKeysForm.tsx
edlerd: I see, thanks for the detailed explanation. I think we can get the same behavior of sorting but without the extra pinned count variable:

On edit, we take all the existing user keys and sort them when initializing the user keys.
- When a key gets removed, the sorting stays stable.
- When a key is overriding an inherited value, it gets into the right place naturally.
- When a new key is added, we append it to the state array. Keeping the correct sorting order.

So we push the sorting information into the form state variable for the user keys. The problem of sorting is pushed to the initialization of the edit form, so we don't need to keep track of the pinned keys.

What do you think?

src/pages/profiles/forms/ProfileDetailsForm.tsx
edlerd: Ah, you are right. I didn't realize this is for profiles. Assuming wrongly this was for instances. Then all good as is :+1:

src/util/userKeys.tsx
edlerd: In my test I clicked "Add user key" and never entered a key name nor value. This kept the form submit button disabled, without a visible error, even on leaving the key input. Should we allow submission in this case where a user key and value are both empty?

vosdev: Did you modify anything else? When I tried to reproduce your steps, the only reason the submit button was still disabled was because nothing was added (besides the empty user key row/inputs). There are no pending changes.

If I click `Add user key` and leave it empty, then modify the description of the instance, I am able to save the instance config.

edlerd: Ah yes, you are right, my mistake. I was expecting to save with no pending changes. All good as is :)

vosdev: When clicking Edit elsewhere in the configuration, for example here:

image


The Save Changes button turns green, stating there is 1 `change`. But that is arguably a bug, because there is no change yet. "Saving" this will not change anything on the instance because no value has been selected in the dropdown yet.

edlerd: > The Save Changes button turns green, stating there is 1 `change`. But that is arguably a bug, because there is no change yet. "Saving" this will not change anything on the instance because no value has been selected in the dropdown yet.

I think the behavior matches the user expectations. I don't think the user keys proposed in this PR necessarily need to function in the same way. Furthermore, I think they are fine as proposed.

src/components/forms/UserKeysForm.tsx
vosdev: In my tests sorting at init didn't fix it because of the data coming from several sources (local and inherited), new rows didn't show up at the bottom and moved while I  was typing.

So I put the sorting info into the form state like you suggested, with a simple anchor in each row.

- I can modify existing keys without them jumping around
- Local, inherited and overridden keys merge alphabetically
- New items (without anchor) show up at the end.

And pinnedCount is gone

src/components/forms/UserKeysForm.tsx
vosdev: Am now using focusField, I missed that helper :)

vosdev:
Updated the details in the PR with new data and screenshots.

Some things I would like to change:

1. Change the "Inherited from profile" column to "Inherited". All columns on all the configuration pages name this column "Inherited".
This will open up a few `rem`'s of space on window sizes between 800 and 1000 pixels.

2. And also I'm thinking about removing the `user.` static text when editting user keys. That is also valuable room on a 820-1000 pixels width window + mobile screens. It's already clear that this area is for configuring user keys (Title + Info icon) so I don't think we need this added static text.


We have two input fields, a profile column and a button. When between 820 and 1000 pixels the whole configuration page is still very tight when you have the main sidebar open.

For example here is 900 pixels width with the main sidebar open:
image

824px window width is just completely unusable:

image

At 820 pixels the sidebar automatically collapses, and so does the configuration sidenav. So everything under 820 pixels is completely fine.


Someone with a window between 820 and ~1000 pixels would have to either increase their window size, or manually collapse the main sidebar to make enough room to edit user keys. I **could introduce a horizontal scrollbar** just like we have on other configuration pages? (See screenshot)

image


This issue is not related to this new addition but the configuration panel as a whole. 

The collapsable configuration sidenav I added with https://github.com/canonical/lxd-ui/pull/2105 solves this for <820 pixels, anything 820<->1000pixels is just in a difficult position. Adding a horizontal scrollbar is more of a bandaid than a real fix.

src/util/profiles.tsx
```diff
@@ -48,14 +48,22 @@ export const hasCloudInit = (profile: LxdProfile): boolean => {
...
+      (a, b) => profileNames.indexOf(b.name) - profileNames.indexOf(a.name),
     );
 };
+
+export const getAppliedProfiles = (
+  values: CreateInstanceFormValues | EditInstanceFormValues,
+  profiles: LxdProfile[],
+): LxdProfile[] => {
+  return getProfilesByName(values.profiles, profiles);
+};
```
edlerd: Can we undo the changes here? This seems like a no-op, just refactoring the code slightly without any interface or functional changes. The new `getProfilesByName` is never used, except internally. So probably best to keep this as it was.

src/components/UserKeyListTable.tsx
```diff
@@ -0,0 +1,54 @@
...
+    { content: "Value", sortKey: "value", className: "u-text--muted value" },
+  ];
+
+  const rows = userKeys.map(({ key, value }) => {
+    return {
+      key,
+      className: "u-row",
+      columns: [
+        {
+          content: key,
```
edlerd, line 25: I wonder if we should display the key with the `user.` prefix here. But maybe it's okay to leave it out as well.

tests/instances.spec.ts
```diff
@@ -229,6 +230,73 @@ test("instance edit cloud init configuration", async ({ page }) => {
   await assertCode(page, "Vendor data", "baz:");
 });
 
+const userKeyRow = (page: Page, name: string) => {
+  return page.locator(".user-key-list-table").getByRole("row", { name });
+};
```
edlerd: Let's move this into `tests/helpers/instances.ts` so we only have the tests in this file.

tests/instances.spec.ts
```diff
@@ -229,6 +230,73 @@ test("instance edit cloud init configuration", async ({ page }) => {
...
+  const table = page.locator("#user-keys-table");
+  const addUserKey = async (index: number, key: string, value: string) => {
+    await page.getByRole("button", { name: "Add user key" }).click();
+    // adding a row focuses its key input
+    await expect(
+      page.getByLabel(`User key ${index}`, { exact: true }),
+    ).toBeFocused();
+    await page.getByLabel(`User key ${index}`, { exact: true }).fill(key);
+    await page.getByLabel(`User key value ${index}`).fill(value);
+  };
```
edlerd: Let's move this also to `tests/helpers/instances.ts`.

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+interface Props {
+  formik: InstanceAndProfileFormikProps;
+  disabledReason?: string;
+}
+
+const UserKeysForm: FC<Props> = ({ formik, disabledReason }) => {
+  const { project } = useParams<{ project: string }>();
+  const { data: profiles = [] } = useProfiles(project ?? "");
+  const [isTableEditing, setIsTableEditing] = useState(false);
```
edlerd:
```suggestion
```
We should not shadow the formik state. Better rely directly on `formik.values.readOnly` and `ensureEditMode(formik)` to manage this.

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+  const userKeys = formik.values.user_keys ?? [];
+  const keyErrors = getUserKeyErrors(userKeys);
+  const inheritedKeys = getInheritedUserKeys(formik.values, profiles);
+  const hasInherited = inheritedKeys.length > 0;
+  const isReadOnly = formik.values.readOnly;
+  const showInputs = !isReadOnly && isTableEditing;
```
edlerd:
```suggestion
```
I think this is redundant and we can directly rely on `isReadOnly`.

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+  const removeUserKey = (index: number) => {
+    setUserKeys(userKeys.filter((_item, i) => i !== index));
+  };
+
+  useEffect(() => {
+    if (isReadOnly) {
+      setIsTableEditing(false);
+    }
+  }, [isReadOnly]);
+
```
edlerd:
```suggestion
```
With the suggestions above, we don't need this side effect.

src/components/forms/UserKeysForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+  const actionColumn = (content: ReactNode) => {
+    return {
+      content,
+      role: "cell",
+      "aria-label": "Actions",
+      className: "actions",
+    };
+  };
+
```
edlerd: We usually don't have helpers for building columns. I am not a huge fan of the idea, but also okay to keep as is.

edlerd (commented):
This looks great, thanks for improving it! Did a detailed QA and code review and have some more comments. Some are nonblocking, and I think the other ones are tiny and simple adjustments. Likely last ones before merging.

src/components/forms/UserKeysForm.tsx
vosdev: This snippet was introduced because the user key table jumped to edit mode as soon as you modified something on the `Main Configuration` page. Like typing in the description or adding/removing a profile.

I thought that was a bit ugly behavior, hence the solution.

Do you still want to get rid of isTableEditing ?

edlerd: The edit mode for all is what we do in other places. Once you start editing, everything turns into an edit mode. I think we should do it here as well.

src/components/UserKeyListTable.tsx
vosdev: The table header already implies it is for user keys, and it is valuable space for smaller screens. I would rather keep it hidden.

vosdev: We do show the `user.` prefix while editting on the `Main Configuration` page in the input fields. I am still considering removing that also to keep every view the same. What do you think?

image

edlerd: This is fundamentally a UX question, and I struggle to find the "right" answer. I think there is only a "best option" answer considering the tradeoffs. We should consider understanding/learnability of the feature and consistency as well as responsiveness and style.

Maybe @sophiecarroll can help find a great solution.

sophiecarroll: Hey! I would agree with removing "user." prefix, unless there are cases where you need to edit one of those specific characters, or adding and removing "user." is a useful functionality. If it's something that should always be there or the system will break, I would remove it both for visual clarity and error prevention.

vosdev: @sophiecarroll "Under the hood" these key/values are saved as `user.owner: sophiecarroll`. Hence the term User Keys. They are LXD's version of a "tag".

I will go ahead and remove them from the input field :) thanks for the quick response

vosdev:
All done!

edlerd:
> All done!

Thanks! I'll have a look at it later today.

edlerd (approved):
LGTM, thanks for the contribution!

PR #2214, opened 2026-08-21
Close any panel on ESC key press

edlerd:
## Done

- Close any panel on ESC key press
- Cleanup custom modal ESC handlers and rely on the one upstream in react-components

Fixes WD-38408

## QA

1. Run the LXD-UI:
- Check ESC key press on any side panel and modal you can find. Ensure panels go back one step on nested panels and close on the first step. Modals should always close on ESC key press.

### DIFF
--- a/src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
+++ b/src/components/forms/NetworkDevicesForm/edit/NetworkDevicePanel.tsx
@@ -45,6 +45,7 @@ import { useNetworkAcls } from "context/useNetworkAcls";
 import type { LxdNetwork } from "types/network";
 import { Link } from "react-router";
 import { ROOT_PATH } from "util/rootPath";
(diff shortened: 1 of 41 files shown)


omarelkashef (commented):
Thank you for standardizing this behavior! Only one observation below:

When I click on "See project configuration" in replicator preflight check modal, the modal does not close and nothing happens

https://github.com/user-attachments/assets/7547caf0-9830-4392-8199-be56fd88e4f3

edlerd:
> Thank you for standardizing this behavior! Only one observation below:
> 
> When I click on "See project configuration" in replicator preflight check modal, the modal does not close and nothing happens

Good catch! This happens because we are already on the page we link to. Let's hide the link in that case and only show it if we are in the replicator list. Adjusted the behaviour accordingly.

omarelkashef (approved):
LGTM, thank you!

PR #2226, opened 2026-08-25
feat(identities): explain identities types in the list page

omarelkashef:
## Done

- explain identities types in the list page WD-37735

## QA

1. Run the LXD-UI:
    - Go to Permissions -> Identities.
    - Hover over an identity type to see the explanations

## Screenshots
image

### DIFF
--- a/src/pages/permissions/IdentityTypeChip.tsx
+++ b/src/pages/permissions/IdentityTypeChip.tsx
@@ -1,0 +1,67 @@
+import type { FC } from "react";
+import classnames from "classnames";
+import { Tooltip } from "@canonical/react-components";
+import type { LxdIdentity } from "types/permissions";
+import { IDENTITY_TYPE_HELP_TEXT } from "util/permissionIdentities";
+
+interface Props {
+  identity: LxdIdentity;
+  className?: string;
+}
+
+const getIdentityType = (identity: LxdIdentity): string => {
+  return `${identity.authentication_method.toUpperCase()} - ${identity.type}`;
+};
+
+// Identity types can carry a suffix (e.g. "(pending)"), match against the known base types
+const getHelpText = (type: LxdIdentity["type"]) => {
+  const baseType = (
+    Object.keys(IDENTITY_TYPE_HELP_TEXT) as Array<
+      keyof typeof IDENTITY_TYPE_HELP_TEXT
+    >
+  ).find((value) => type.startsWith(value));
+  return baseType ? IDENTITY_TYPE_HELP_TEXT[baseType] : undefined;
+};
+
+const IdentityTypeChip: FC<Props> = ({ identity, className }) => {
+  const helpText = getHelpText(identity.type);
+  const identityType = getIdentityType(identity);
+
+  const chip = (
+    <span
+      className={classnames("p-chip is-dense is-inline", className)}
+      tabIndex={helpText ? 0 : undefined}
+    >
+      <span className="p-chip__value u-truncate">{identity.type}</span>
+    </span>
+  );
+
+  if (!helpText) {
+    return chip;
+  }
+
+  return (
+    <Tooltip
+      zIndex={1000}
+      position="right"
+      positionElementClassName="identity-type-chip-tooltip-wrapper"
+      tooltipClassName="identity-type-chip-tooltip"
+      message={
+        <>
+          <div className="p-text--small-caps u-text--muted u-no-margin--bottom">
+            Type
+          </div>
+          <div className="identity-type-chip-tooltip-title u-no-margin--top">
+            {identityType}
+          </div>
+          <div className="p-text--small-caps u-text--muted">Description</div>
+          <div>{helpText.description}</div>
+        </>
+      }
+    >
+      {chip}
+    </Tooltip>
+  );
+};
+
+export default IdentityTypeChip;
--- a/src/pages/permissions/PermissionIdentities.tsx
+++ b/src/pages/permissions/PermissionIdentities.tsx
@@ -43,6 +43,7 @@ import PermissionIdentitiesActions from "pages/permissions/PermissionIdentitiesA
 import ResourceLabel from "components/ResourceLabel";
 import IdentityExpiration from "pages/permissions/IdentityExpiration";
 import IdentityAdditionalIdpGroups from "pages/permissions/IdentityAdditionalIdpGroups";
+import IdentityTypeChip from "pages/permissions/IdentityTypeChip";
 
 const PermissionIdentities: FC = () => {
   const notify = useNotify();
@@ -72,9 +73,12 @@ const PermissionIdentities: FC = () => {
   const headers = [
     { content: "Name", className: "name", sortKey: "name" },
     { content: "ID", sortKey: "id", className: "identity-id" },
-    { content: "Auth method", sortKey: "authmethod", className: "auth-method" },
-    { content: "Type", sortKey: "type", className: "identity-type" },
     { content: "Expires", sortKey: "expires", className: "expires" },
+    {
+      content: "Type",
+      sortKey: "type",
+      className: "type",
+    },
     {
       content: "Groups",
       sortKey: "groups",
@@ -177,21 +181,16 @@ const PermissionIdentities: FC = () => {
           title: identity.id,
         },
         {
-          content: identity.authentication_method.toUpperCase(),
-          role: "cell",
-          "aria-label": "Auth method",
-          className: "auth-method",
-        },
-        {
-          content: identity.type,
-          role: "cell",
-          "aria-label": "Type",
-          className: "u-truncate identity-type",
-        },
-        {
           content: <IdentityExpiration identity={identity} />,
           role: "cell",
           "aria-label": "Expires",
+          className: "expires",
+        },
+        {
+          content: <IdentityTypeChip identity={identity} />,
+          role: "cell",
+          "aria-label": "Type",
+          className: "type",
         },
         {
           content: getGroupLink(),
@@ -222,7 +221,7 @@ const PermissionIdentities: FC = () => {
               <DeleteIdentityBtn identity={identity} />
             </>
           ),
-          className: "actions u-align--right",
+          className: "u-align--right actions",
           role: "cell",
           "aria-label": "Actions",
         },
@@ -230,8 +229,7 @@ const PermissionIdentities: FC = () => {
       sortData: {
         id: identity.id,
         name: name.toLowerCase(),
-        authentication_method: identity.authentication_method,
-        type: identity.type,
+        type: identity.type.toLowerCase(),
         expires: identity.expires_at ?? "",
         groups: identity.groups?.length || 0,
       },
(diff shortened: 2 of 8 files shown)


sophiecarroll:
After discussing with another designer we landed on a good solution for explaining identities: making the identity type name a rich chip, that when hovered over, displays the definition. The tooltip with all definitions was a good idea but too long for a tooltip, and I think it makes even more sense to attach the definition directly to the name.
Screenshot 2026-08-26 at 12 13 40 PM

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -73,7 +74,11 @@ const PermissionIdentities: FC = () => {
     { content: "Name", className: "name", sortKey: "name" },
     { content: "ID", sortKey: "id", className: "identity-id" },
     { content: "Auth method", sortKey: "authmethod", className: "auth-method" },
-    { content: "Type", sortKey: "type", className: "identity-type" },
+    {
+      content: "Type",
+      sortKey: "type",
+      className: "identity-type",
+    },
```
edlerd: Should we combine the Auth method and Type columns into a single "Type" column? I think they both relate to the same property, and we might not need the dedicated column for each.

src/sass/_permission_identities.scss
```diff
@@ -4,6 +4,28 @@
   margin-left: 0.8rem;
 }
 
+.identity-type-chip-tooltip .p-tooltip__message {
+  font-size: 0.75rem;
```
edlerd:
```suggestion
```
I think we should never customize the font size. Either use a small text helper or standard size.

src/sass/_permission_identities.scss
```diff
@@ -4,6 +4,28 @@
...
+  white-space: normal;
+  width: 18rem;
+}
+
+.identity-type-chip-tooltip-content {
+  .identity-type-chip-tooltip-label {
+    font-size: 0.65rem;
+    font-weight: 600;
+    letter-spacing: 0.02rem;
+    text-transform: uppercase;
```
edlerd:
```suggestion
```
Can we use an existing heading class instead of custom styles?

src/sass/_permission_identities.scss
```diff
@@ -4,6 +4,28 @@
...
+  }
+
+  .identity-type-chip-tooltip-title {
+    border-bottom: $border-thin;
+    font-weight: 600;
```
edlerd:
```suggestion
```
We should use existing helper classes to emphasize.

edlerd (commented):
I quite like the chips with tooltips. Though it breaks with the "interactive elements are in the outer columns" pattern. Maybe we should switch the column order?

A few ideas and remarks, also on the styling below.

omarelkashef:
> I quite like the chips with tooltips. Though it breaks with the "interactive elements are in the outer columns" pattern. Maybe we should switch the column order?

The last column is an actions column, so it has to stay there, and I do not think it is good option to put the type first, either.

edlerd:
> The last column is an actions column, so it has to stay there, and I do not think it is good option to put the type first, either.

I didn't mean it has to be the last column. But maybe swap it with the expires one to show it next to the groups. So the last 3 columns are interactive and in the middle the columns are text only.

sophiecarroll:
Overflow behavior bug: chip cuts off. We can't have the text wrap onto multiple lines with the chip, can we have the chip shorten width and then text becomes cut off with ...?
Screenshot 2026-08-26 at 5 55 43 PM

sophiecarroll:
Apologies if this is just an experiment, but why did you put Bearer/TLS in front of the type name? Also, the text should be normal not italicized for the hover over message.

omarelkashef:
> Apologies if this is just an experiment, but why did you put Bearer/TLS in front of the type name? Also, the text should be normal not italicized for the hover over message.

We decided to merge two related columns: auth method and type. That is why the type is now prefixed with TLS/Bearer. The italic should be removed now. It did not look good.

sophiecarroll:
> > Apologies if this is just an experiment, but why did you put Bearer/TLS in front of the type name? Also, the text should be normal not italicized for the hover over message.
> 
> We decided to merge two related columns: auth method and type. That is why the type is now prefixed with TLS/Bearer. The italic should be removed now. It did not look good.

@omarelkashef Why did you decide to merge the two columns? From a UX perspective, I think it makes it harder to orient yourself as a user because the name of the type is cut off easily with bearer or tls at the beginning, but I am curious about the reasoning

src/pages/permissions/IdentityTypeChip.tsx
```diff
@@ -0,0 +1,64 @@
...
+  type: LxdIdentity["type"];
+  authMethod: LxdIdentity["authentication_method"];
+  className?: string;
+}
+
+// Identity types can carry a suffix (e.g. "(pending)"), match against the known base types
+const getHelpText = (type: LxdIdentity["type"]) => {
+  const baseType = Object.values(IDENTITY_TYPE).find((value) =>
+    type.startsWith(value),
+  );
```
edlerd: Several types are missing here. I spotted: `Initial UI token bearer`, `OIDC client` and `Server certificate` I think there is also a metrics certificate and maybe more. It would be good to check the LXD source for an exhaustive list.

src/pages/permissions/IdentityTypeChip.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Tooltip } from "@canonical/react-components";
+import type { LxdIdentity } from "types/permissions";
+import {
+  IDENTITY_TYPE,
+  IDENTITY_TYPE_HELP_TEXT,
+} from "util/permissionIdentities";
+
+interface Props {
+  type: LxdIdentity["type"];
+  authMethod: LxdIdentity["authentication_method"];
```
edlerd:
```suggestion
  identity: LxdIdentity;
```
Let's simplify the interface, wdyt?

src/pages/permissions/IdentityTypeChip.tsx
```diff
@@ -0,0 +1,64 @@
...
+const getHelpText = (type: LxdIdentity["type"]) => {
+  const baseType = Object.values(IDENTITY_TYPE).find((value) =>
+    type.startsWith(value),
+  );
+  return baseType ? IDENTITY_TYPE_HELP_TEXT[baseType] : undefined;
+};
+
+const IdentityTypeChip: FC<Props> = ({ type, authMethod, className }) => {
+  const helpText = getHelpText(type);
+  const identityType = `${authMethod.toUpperCase()} - ${type}`;
```
edlerd: I think this should be in a helper, so we can reuse it for the sort data and in the edit side panel. Extracting guarantees this stays aligned in the future making it simpler to maintain.

edlerd (commented):
Thanks for the iteration. I think the combined column is much clearer and cleaner as well. Also, the refined position of the column has improved the impression of the table.

src/pages/permissions/IdentityTypeChip.tsx
edlerd: The comment relates to the construction of the `identityType`

omarelkashef:
> > > Apologies if this is just an experiment, but why did you put Bearer/TLS in front of the type name? Also, the text should be normal not italicized for the hover over message.
> > 
> > 
> > We decided to merge two related columns: auth method and type. That is why the type is now prefixed with TLS/Bearer. The italic should be removed now. It did not look good.
> 
> @omarelkashef Why did you decide to merge the two columns? From a UX perspective, I think it makes it harder to orient yourself as a user because the name of the type is cut off easily with bearer or tls at the beginning, but I am curious about the reasoning

We made that decision because they are very related concepts, and it did not make sense to have two separate columns for this info. However, I see your concern that it is cut off on small screens and it might not be very easy for users to look at the types. @edlerd

sophiecarroll:
> > > > Apologies if this is just an experiment, but why did you put Bearer/TLS in front of the type name? Also, the text should be normal not italicized for the hover over message.
> > > 
> > > 
> > > We decided to merge two related columns: auth method and type. That is why the type is now prefixed with TLS/Bearer. The italic should be removed now. It did not look good.
> > 
> > 
> > @omarelkashef Why did you decide to merge the two columns? From a UX perspective, I think it makes it harder to orient yourself as a user because the name of the type is cut off easily with bearer or tls at the beginning, but I am curious about the reasoning
> 
> We made that decision because they are very related concepts, and it did not make sense to have two separate columns for this info. However, I see your concern that it is cut off on small screens and it might not be very easy for users to look at the types. @edlerd

Can we talk about it for a few minutes before standup so I can get a better understanding?

omarelkashef:
> > > > > Apologies if this is just an experiment, but why did you put Bearer/TLS in front of the type name? Also, the text should be normal not italicized for the hover over message.
> > > > 
> > > > 
> > > > We decided to merge two related columns: auth method and type. That is why the type is now prefixed with TLS/Bearer. The italic should be removed now. It did not look good.
> > > 
> > > 
> > > @omarelkashef Why did you decide to merge the two columns? From a UX perspective, I think it makes it harder to orient yourself as a user because the name of the type is cut off easily with bearer or tls at the beginning, but I am curious about the reasoning
> > 
> > 
> > We made that decision because they are very related concepts, and it did not make sense to have two separate columns for this info. However, I see your concern that it is cut off on small screens and it might not be very easy for users to look at the types. @edlerd
> 
> Can we talk about it for a few minutes before standup so I can get a better understanding?

Yes, sure!

src/pages/permissions/IdentityTypeChip.tsx
```diff
@@ -0,0 +1,65 @@
...
+    </span>
+  );
+
+  if (!helpText) {
+    return chip;
+  }
+
+  return (
+    <Tooltip
+      position="right"
```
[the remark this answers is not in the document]
edlerd: This seems to be a problem indeed, can we add this as suggested?

edlerd (dismissed):
LGTM, tiny nitpick from the copilot suggestion

edlerd (dismissed):
LGTM, thanks :)

As a follow up can we also look into responsiveness for the table? Let's use the `responsive` prop on this and other permission tables and can remove some of the custom tailored styling / hiding of information for small screens.

src/pages/permissions/PermissionIdentities.tsx
```diff
@@ -222,16 +222,15 @@ const PermissionIdentities: FC = () => {
...
         },
       ],
       sortData: {
         id: identity.id,
         name: name.toLowerCase(),
-        authentication_method: identity.authentication_method,
-        type: identity.type,
+        type: getIdentityType(identity).toLowerCase(),
         expires: identity.expires_at ?? "",
         groups: identity.groups?.length || 0,
```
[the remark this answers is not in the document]
edlerd: @omarelkashef seems like an oversight when we switched the displayed content, not updating the sort data.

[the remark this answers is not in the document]
omarelkashef: Oh yes, I will fix it.

omarelkashef:
> LGTM, thanks :)
> 
> As a follow up can we also look into responsiveness for the table? Let's use the `responsive` prop on this and other permission tables and can remove some of the custom tailored styling / hiding of information for small screens.

I will create a ticket.

edlerd (approved):

PR #2231, opened 2026-08-26
CI: use workshop in coverage workflow

omarelkashef:
## Done

- use workshop in coverage workflow WD-37747

## QA

1. Run the LXD-UI:
    - Ensure coverage workflow passes

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -9,9 +9,42 @@ permissions:
   contents: read
 
 jobs:
+  js-coverage:
+    name: js-test-coverage
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 24
+
+      - name: Install LXD-UI dependencies
+        run: yarn install
+
+      - name: Run JS tests with coverage
+        run: yarn test-js-coverage
+
+      - name: Upload JS coverage report
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: js-coverage
+          path: coverage/unit
+          retention-days: 7
+          overwrite: true
+
   e2e-coverage:
-    name: e2e-test-coverage
+    name: e2e-test-coverage (${{ matrix.deployment }})
     runs-on: self-hosted-linux-amd64-noble-large-tiobe
+    strategy:
+      # Run the unclustered and clustered suites as independent parallel jobs
+      # so that a failure in one does not cancel or block the other.
+      fail-fast: false
+      matrix:
+        deployment: [unclustered, clustered]
     env:
       LXD_OIDC_CLIENT_ID: "RYDnMpkygLAMfeo17lU7LYwWGxisRuRR"
       LXD_OIDC_CLIENT_SECRET: "CNKX4UmrZKZJq5rJy5VM_JfcNPqkws1rwWWQk_q0oyZ8gABARr19ic7xrhPssGA1"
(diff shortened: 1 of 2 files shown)


omarelkashef:
Workflow runs:  https://github.com/omarelkashef/lxd-ui/actions/runs/33479212234/job/99777726265 and https://github.com/omarelkashef/lxd-ui/actions/runs/33479212234/job/99796007317

Notice that the two runs above have different artifacts, showing that rerunning failed jobs will produce fresh artifacts.

Copilot report:

## Coverage Workflow Equivalence Report

I compared the coverage artifact produced by the updated workflow with the artifact from the old successful workflow. The downstream consumers are `publish-coverage-report` and `tics-report`.

### Conclusion

The updated workflow preserves the same downstream artifact contract as the old workflow. The differences are expected coverage-content differences from splitting and merging the same raw inputs, not structural or naming differences that should break publishing or TICS.

### Artifact Structure

Both artifacts contain the same required top-level report files:

- `index.html`
- `cobertura-coverage.xml`
- `cobertura-coverage.zip`
- `base.css`
- `block-navigation.js`
- `prettify.css`
- `prettify.js`
- `sorter.js`
- `favicon.png`
- `sort-arrow-sprite.png`

The static report assets are byte-for-byte identical between the old and new artifacts. This confirms that the report format generated by `nyc report` has not changed.

### Cobertura XML Compatibility

Both reports use the same Cobertura structure:

- Root element: `coverage`
- Class filenames are repo-relative, for example `src/App.tsx` and `src/api/auth-groups.tsx`
- No absolute filenames appear in `<class filename="...">`
- `src/lib/**` and `src/types/**` remain excluded
- The same report types are generated: HTML and Cobertura XML

This is the key compatibility point for TICS-style consumers: the coverage file paths inside the XML remain stable and repo-relative.

### Zip Compatibility

Both artifacts include `cobertura-coverage.zip`, and both zip files contain the same internal path:

`coverage/playwright-report/cobertura-coverage.xml`

That matches the updated workflow command and preserves the same zipped Cobertura payload shape as the old workflow.

### Publish Step Compatibility

The updated workflow uploads the merged report as artifact `lxd-coverage` from:

`coverage/playwright-report`

The publish step downloads the same artifact name into:

`coverage`

That is equivalent to the old workflow’s published report shape: the contents of `coverage/playwright-report` become the root of the published `coverage` directory on `gh-pages`.

### TICS Step Compatibility

The TICS job does not directly download the coverage artifact in this workflow. Its inputs remain:

- Repository checkout
- Node dependencies
- TICS qserver configuration
- TICS auth token
- Project name `lxd-ui`

The dependency change only gates TICS on successful e2e coverage and successful publishing:

`needs: [e2e-coverage, publish-coverage-report]`

That is stricter than before, but it does not change the TICS action inputs.

### Temporary Runner Difference

The only notable structural difference observed was caused by the temporary runner change:

- Old artifact HTML source pages lived under `src/...`
- New artifact HTML source pages linked under `home/runner/work/lxd-ui/lxd-ui/src/...`

Since the runner change will be reverted, this difference should disappear. Even with that temporary difference, the new HTML report was internally linked correctly and the Cobertura XML still used repo-relative filenames, so it was not a blocking compatibility issue for the workflow handoff.

### Merge Confidence

The updated workflow keeps the downstream contract intact:

- Same final artifact name: `lxd-coverage`
- Same published report entry point: `index.html`
- Same Cobertura XML filename: `cobertura-coverage.xml`
- Same Cobertura zip filename: `cobertura-coverage.zip`
- Same zip internal XML path
- Same repo-relative coverage filenames
- Same exclusions
- Same static report asset names and contents

Based on the artifact comparison, the updated split-and-merge workflow should be equivalent for `publish-coverage-report` and `tics-report` once tests pass. It is reasonable to merge the PR from a workflow-setup perspective.

.github/workflows/coverage.yaml
```diff
@@ -20,112 +27,132 @@ jobs:
...
+          ENVIRONMENT: devel
+          PORT: 8407
+          VITE_PORT: 3000
         run: |
           set -x
-          sudo chmod 0777 ../lxd-ui
-          dotrun install
+          sudo chmod -R 0777 ../lxd-ui
+          yarn test-js-coverage
+          yarn test-e2e-cleanup
```
edlerd, line 86:
```suggestion
```
This shouldn't be necessary here, no?

.github/workflows/coverage.yaml
```diff
@@ -20,112 +27,132 @@ jobs:
...
+        env:
+          ENVIRONMENT: devel
+          PORT: 8407
+          VITE_PORT: 3000
         run: |
           set -x
-          sudo chmod 0777 ../lxd-ui
-          dotrun install
+          sudo chmod -R 0777 ../lxd-ui
+          yarn test-js-coverage
```
edlerd: I just realized now: Should we extract this into a dedicated step? I think the JS tests need no setup whatsoever, just a checkout and installing dependencies.

.github/workflows/coverage.yaml
```diff
@@ -20,112 +27,132 @@ jobs:
...
           path: coverage/playwright-report
           retention-days: 1
+          overwrite: true
 
   publish-coverage-report:
     name: publish-coverage-report
     runs-on: ubuntu-latest
-    needs: e2e-coverage
-    continue-on-error: true
+    # runs-on: self-hosted-linux-amd64-noble-large-tiobe
```
edlerd:
```suggestion
```

.github/workflows/coverage.yaml
```diff
@@ -20,112 +27,132 @@ jobs:
...
-        shell: bash
-        run: |
-          set -x
-          sudo lxc config set core.https_address "[::]:8443"
-          sudo lxc config trust add keys/lxd-ui.crt
-          sudo lxc config set cluster.https_address "127.0.0.1"
+  merge-coverage:
+    name: merge-coverage
+    runs-on: ubuntu-latest
+    # runs-on: self-hosted-linux-amd64-noble-large-tiobe
```
edlerd:
```suggestion
```

edlerd (commented):
Mostly LGTM, few comments below.

.github/workflows/coverage.yaml
omarelkashef: This will be reverted. It was only changed to run on ubuntu-latest to be able to trigger it from my fork.

.github/workflows/coverage.yaml
omarelkashef: This will be reverted. It was only changed to run on ubuntu-latest to be able to trigger it from my fork.

.github/workflows/coverage.yaml
omarelkashef: Yes it seems unnecessary

.github/workflows/coverage.yaml
edlerd: Publish coverage is a simple task and it should be okay to run on ubuntu-latest. No need for self hosted large runner on this task.

.github/workflows/coverage.yaml
edlerd: Merge as well is simple, this should run on the GitHub provided ubuntu latest runner.

.github/workflows/coverage.yaml
omarelkashef: Actually, it might be necessary to prevent stale coverage from contaminating reports since we run on a self-hosted runner.

edlerd: The self hosted does not mean we re-run with previous state. If we don't explicitly cache the directories, we should assume they are empty.

.github/workflows/coverage.yaml
```diff
@@ -20,112 +53,135 @@ jobs:
...
+          overwrite: true
 
   publish-coverage-report:
     name: publish-coverage-report
     runs-on: ubuntu-latest
-    needs: e2e-coverage
-    continue-on-error: true
+    # Only update the live site when the e2e tests actually passed; partial
+    # coverage from a failed run would misrepresent the real numbers.
+    needs: [js-coverage, e2e-coverage, merge-coverage]
```
edlerd:
```suggestion
    needs: merge-coverage
```
I think we can keep this simple, as the merge step already depends on the other two.

.github/workflows/coverage.yaml
omarelkashef: I assumed that since these are not ephemeral, data can be persisted between runs and it is theoretically possible that a job is run on a runner with stale data from a previous job run. Is that not true?

.github/workflows/coverage.yaml
```diff
@@ -20,112 +53,135 @@ jobs:
...
-          set -x
-          sudo lxc config set core.https_address "[::]:8443"
-          sudo lxc config trust add keys/lxd-ui.crt
-          sudo lxc config set cluster.https_address "127.0.0.1"
+  merge-coverage:
+    name: merge-coverage
+    runs-on: ubuntu-latest
+    needs: [js-coverage, e2e-coverage]
+    # Test failures still upload coverage artifacts for this job to merge.
+    if: ${{ !cancelled() }}
```
edlerd:
```suggestion
```
Let's simplify and not run the merge if we failed in a previous step. Wdyt?

.github/workflows/coverage.yaml
```diff
@@ -20,112 +53,135 @@ jobs:
...
-        if: always()
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-        with:
-          name: playwright-test-artifacts
-          path: test-results
-          retention-days: 7
 
       - name: Upload coverage report
-        if: always()
+        if: ${{ !cancelled() }}
```
edlerd:
```suggestion
```
I think we shouldn't upload on failure.

edlerd (commented):
I think we can slightly simplify the dependencies in the workflow. See below.

.github/workflows/coverage.yaml
edlerd: I assume they are not stateful, but not 100% sure. We can also keep this as is, no problem.

edlerd (approved):

PR #2236, opened 2026-08-26
fix(auth) surface auth errors to the user

edlerd:
## Done

- fix(auth) surface auth errors to the user

## QA

1. Run the LXD-UI:
    - run the UI against a backend that is from the 5.21 branch, such as `snap install lxd --channel=5.21/edge`
    - Ensure the UI loads and surfaces the error as below and we do not end up in an endless retry cycle for long times, ending in a confusing "project not found" page as before.

## Screenshots

image

### DIFF
--- a/src/context/appProviders.tsx
+++ b/src/context/appProviders.tsx
@@ -23,8 +23,13 @@ const queryClient = new QueryClient({
     queries: {
       retry: (failureCount, error: unknown) => {
         if (error instanceof LxdApiError) {
-          // Do not retry for 404 errors
-          if (error.status === 404) {
+          // Do not retry for most 4xx client errors (auth/validation), but allow retries for timeouts/rate limiting.
+          if (
+            error.status >= 400 &&
+            error.status < 500 &&
+            error.status !== 408 &&
+            error.status !== 429
+          ) {
             return false;
           }
         }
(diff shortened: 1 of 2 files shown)


src/context/appProviders.tsx
```diff
@@ -23,8 +23,8 @@ const queryClient = new QueryClient({
     queries: {
       retry: (failureCount, error: unknown) => {
         if (error instanceof LxdApiError) {
-          // Do not retry for 404 errors
-          if (error.status === 404) {
+          // Do not retry for 404 or 400 errors
+          if (error.status === 404 || error.status === 400) {
```
omarelkashef: Should this be generalized for all client errors 400-499?

src/context/auth.tsx
```diff
@@ -110,7 +114,8 @@ export const AuthProvider: FC<ProviderProps> = ({ children }) => {
         isAuthenticated: (settings && settings.auth !== "untrusted") ?? false,
         isAuthLoading:
           isSettingsLoading || isIdentityLoading || isProjectsLoading,
-        authError: settingsError ?? identityError,
+        authError:
```
omarelkashef, line 124: Is it accurate to return authentication errors for cases like sending the wrong entitlements?

omarelkashef (commented):
Thank you for PR, a couple of comments below!

src/context/auth.tsx
edlerd: Yes, because we need all these queries to settle and have the information to make the auth decisions.

src/context/appProviders.tsx
edlerd: Yeah, we can do that, will change it.

omarelkashef (approved):
LGTM, thank you!

PR #2239, opened 2026-08-27
feat(Networking): Prioritize main interfaces on IP address sorting

### LINKED ISSUE #2213: Prioritize guest interface IPv4/6 addresses in the instances list
Hey!

Many of my instances have several additional IPv4/IPv6 addresses. This can be for several reasons, such as nestedcontainers, virtual networks, CNIs, loadbalancer IPs (keepalived or bgp)

Image

All IPs are shown in alphabetical order. In my example, my authentik container's primary IP address is 172.22.0.19. It is hidden behind the `+ 1 more` text.

The IPs that are shown belong to the docker0 bridge and a virtual docker network's bridge.

For the kubernetes nodes in the example, the cilium bridge's IP addresses show first. 
For kube1, a loadbalancer IP shows as second and the node's real IP is also hidden.

I'd like to request the instance's main/configured interfaces to be shown first.

The `hwaddr` information is available in the same API call, which could be used to prioritize the LXD guest interfaces' IP addresses on top of secondary interfaces like docker bridges, cilium bridges and similar things.

A lesser form of address sorting already happens, with IPv6 local addresses always sorting at the end of the list. (`sortIpv6Addresses` in src/util/networks.tsx).

I can try to make a PR for this

vosdev:
This PR prioritizes instances' main interfaces over others (for example interfaces created by docker, kubernetes cni's or loadbalancer software)

It has annoyed me for quite some time, so here is a solution :)

## Done

- Added prioritization of IP addresses with a matching hwaddr configured on the instance to IP sorting.
- Put IPv4 local addresses (127.* and 169.254.*) at the bottom of the list, just like already happens with IPv6 fe80::

Fixes #2213

## QA

1. Run the LXD-UI:
    - Create instances with one or more network adapters
    - Add new network interfaces inside the guest, i.e. docker or kubernetes (or dummy interfaces)
    - Look at the listed IP addresses.

## Screenshots

Note the IPv4 and IPv6 of `verylonginstancename-1234567890`. `10.203.31.138` and `fd42:f409:293:999e:216:3eff:feb0:abe5` are its'  main IP addresses.
The main IPv4 address is 3rd in the list and not visible on the Instances list page.

A localhost addr and the docker0 interface (172.17.0.1) also takes the first two entries in the list for the `resolute` instance, causing a "real" IP address to not be visible.

side by side before & after:

image

`resolute` instance:
image

`verylonginstancename-1234567890` instance:
image

"Real" IPs now take preference and are visible for the `resolute` instance and `verylonginstancename-1234567890` instance.

### DIFF
--- a/src/util/networks.tsx
+++ b/src/util/networks.tsx
@@ -15,6 +15,7 @@ import {
   constructMemberError,
   type AbortControllerState,
 } from "util/helpers";
+import { getInstanceMacAddresses } from "util/instances";
 import { ROOT_PATH } from "util/rootPath";
 import type { AnyObject, TestFunction } from "yup";
 
@@ -45,19 +46,8 @@ export const typesWithNicDeviceAcls = [ovnType];
 export const typesWithNicStaticIPSupport = [bridgeType, ovnType];
 export const typesWithLocalPeerings = [ovnType];
 
-export const getIpAddresses = (
-  instance: LxdInstance,
-  family: IpFamily,
-): IpAddress[] => {
-  if (!instance.state?.network) return [];
-  return Object.entries(instance.state.network)
-    .filter(([key, _value]) => key !== "lo")
-    .flatMap(([key, value]) =>
-      value.addresses.map((item) => {
-        return { ...item, iface: key };
-      }),
-    )
-    .filter((item) => item.family === family);
+export const isLocalIPv4 = (ipv4Address: string): boolean => {
+  return ipv4Address.startsWith("127.") || ipv4Address.startsWith("169.254.");
 };
 
 export const isLocalIPv6 = (ipv6Address: string): boolean => {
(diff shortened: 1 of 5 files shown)


src/util/networks.tsx
```diff
@@ -75,15 +65,62 @@ export const isLocalIPv6 = (ipv6Address: string): boolean => {
...
+  addresses: IpAddress[],
+  mainInterfaces: Set<string>,
+): IpAddress[] => {
+  return [...addresses].sort(
+    (a, b) => getIpRank(a, mainInterfaces) - getIpRank(b, mainInterfaces),
+  );
+};
+
+// Match reported hwaddr to instance's volatile.<device>.hwaddr
+export const getMainInterfaces = (instance: LxdInstance): Set<string> => {
```
edlerd: I am not a fan of the name `mainInterface`. I think it doesn't clearly express what we are prioritizing on sorting. What do you think of `managedInterfaces`? I think what we are prioritizing are interfaces managed by LXD and not inherited from elsewhere. So maybe that wording is clearer.

We could use it here and in the argument names for the other functions.

edlerd (dismissed):
This is great and works seamlessly. Thanks for the contribution!

Tiny non-blocking nitpick on naming below. Please let me know what you think, can also merge as is.

src/util/networks.tsx
vosdev: That makes sense. Done!

edlerd (approved):
LGTM, thanks

PR #2241, opened 2026-08-27
feat: [WD-38192] Add ETs to Entity Forms

Kxiru:
## Done
Creates new ETs:
- SSH Keys
- Adds Config variant to ProjectExplanationTooltip (Feedback from @sophiecarroll in https://github.com/canonical/lxd-ui/pull/2194)

Adds ETs to the following pages:
- Instance creation form (Profile, SSH ETs)
- Profile creation form (SSH ETs)
- Storage Volume creation form (Storage pool ET)
- Storage Buckets creation form (Storage pool ET)
- Custom ISO creation form (Storage pool ET)
- Project configuration form (default storage pool & instance network)
- Project -> Replication -> Replicators for this project (Replicators ET)
- Storage Driver ET for the Storage Pool creation form

Other:
- Minor copy change in profile ET in instance creation.

## QA

1. Run the LXD-UI:

## Screenshots
image

image

### DIFF
--- a/src/components/forms/SshKeyExplanationTooltip.tsx
+++ b/src/components/forms/SshKeyExplanationTooltip.tsx
@@ -1,0 +1,19 @@
+import type { FC, ReactNode } from "react";
+import ExplanationTooltip from "components/ExplanationTooltip";
+
+const SshKeyExplanationTooltip: FC<{
+  children?: ReactNode;
+}> = ({ children }) => {
+  return (
+    <ExplanationTooltip
+      className="explanation-tooltip-wrapper--inline"
+      explanation="Cloud init must be enabled on the instance to apply the keys. Additional keys get applied on instance creation or restart. SSH keys are not removed automatically."
+      docPath="/reference/instance_options/#instance-cloud-init:cloud-init.ssh-keys.KEYNAME"
+      docLabel="Learn more about SSH keys"
+    >
+      {children}
+    </ExplanationTooltip>
+  );
+};
+
+export default SshKeyExplanationTooltip;
--- a/src/pages/instances/InstanceExplanationTooltip.tsx
+++ b/src/pages/instances/InstanceExplanationTooltip.tsx
@@ -7,6 +7,7 @@ const InstanceExplanationTooltip: FC<{
 }> = ({ children, isConfigVariant }) => {
   return (
     <ExplanationTooltip
+      className="explanation-tooltip-wrapper--inline"
       explanation="Instances are VMs or containers."
       docPath={
         isConfigVariant
--- a/src/pages/instances/forms/InstanceTargetSelect.tsx
+++ b/src/pages/instances/forms/InstanceTargetSelect.tsx
@@ -1,11 +1,9 @@
 import { useRef, type FC } from "react";
 import {
   CustomSelect,
-  Icon,
   Input,
   Select,
   Spinner,
-  Tooltip,
 } from "@canonical/react-components";
 import type { FormikProps } from "formik/dist/types";
 import type { CreateInstanceFormValues } from "types/forms/instanceAndProfile";
@@ -20,6 +18,8 @@ import type { SelectRef } from "@canonical/react-components/dist/components/Cust
 import PlacementGroupSelect from "pages/instances/forms/PlacementGroupSelect";
 import { PLACEMENT_GROUP_POLICY_COMPACT } from "pages/placement-groups/PlacementGroupForm";
 import { CLUSTER_GROUP_PREFIX } from "util/instances";
+import ClusterMemberExplanationTooltip from "pages/cluster/ClusterMemberExplanationTooltip";
+import ClusterGroupExplanationTooltip from "pages/cluster/ClusterGroupExplanationTooltip";
 
 export const TARGET = {
   AUTO: "auto",
@@ -162,7 +162,11 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
     return (
       <Input
         id="clusterMember"
-        label="Cluster member"
+        label={
+          <ClusterMemberExplanationTooltip>
+            Cluster member
+          </ClusterMemberExplanationTooltip>
+        }
         value={formik.values.target}
         disabled
         help="Member is determined by the selected ISO volume and can't be changed."
@@ -176,134 +180,141 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
     : "Please select an image before adding a placement group";
 
   return (
-    <div className="instance-target-selector">
-      <Select
-        id="target"
-        label={
-          <>
-            Target{" "}
-            <Tooltip
-              message={`LXD automatically selects a cluster member for the instance.\nPlacement groups allow you to control how instances are distributed across cluster members.`}
-            >
-              <Icon name="information" />
-            </Tooltip>
-          </>
-        }
-        wrapperClassName="select-input"
-        value={type}
-        options={[
-          {
-            label: "Auto",
-            value: TARGET.AUTO,
-          },
-          {
-            label: "Cluster group",
-            value: TARGET.CLUSTER_GROUP,
-            disabled: clusterGroupOptions.length === 0,
-          },
-          {
-            label: "Cluster member",
-            value: TARGET.CLUSTER_MEMBER,
-            disabled:
-              clusterMemberOptions.length === 0 ||
-              isProjectBlockingClusterMemberTargeting,
-          },
-          {
-            label: "Placement group",
-            value: TARGET.PLACEMENT_GROUP,
-          },
-        ]}
-        onChange={(e) => {
-          const type = e.target.value;
-          if (type === TARGET.AUTO) {
-            setAuto();
-          }
-          if (type === TARGET.CLUSTER_GROUP) {
-            setClusterGroup(clusterGroupOptions[0].value);
-            setTimeout(() => clusterGroupRef.current?.open(), 100);
-          }
-          if (type === TARGET.CLUSTER_MEMBER) {
-            setClusterMember(clusterMemberOptions[0].value);
-            setTimeout(() => clusterMemberRef.current?.open(), 100);
-          }
-          if (type === TARGET.PLACEMENT_GROUP) {
-            setPlacementGroup(placementGroups[0]?.name ?? "-");
-            if (placementGroups.length > 0) {
-              setTimeout(() => placementGroupRef.current?.open(), 100);
+    <>
+      <div className="instance-target-selector">
+        <Select
+          id="target"
+          label="Target"
+          wrapperClassName="select-input"
+          value={type}
+          options={[
+            {
+              label: "Auto",
+              value: TARGET.AUTO,
+            },
+            {
+              label: "Cluster group",
+              value: TARGET.CLUSTER_GROUP,
+              disabled: clusterGroupOptions.length === 0,
+            },
+            {
+              label: "Cluster member",
+              value: TARGET.CLUSTER_MEMBER,
+              disabled:
+                clusterMemberOptions.length === 0 ||
+                isProjectBlockingClusterMemberTargeting,
+            },
+            {
+              label: "Placement group",
+              value: TARGET.PLACEMENT_GROUP,
+            },
+          ]}
+          onChange={(e) => {
+            const type = e.target.value;
+            if (type === TARGET.AUTO) {
+              setAuto();
             }
-          }
-        }}
-        disabled={!formik.values.image}
-        error={
-          type === TARGET.PLACEMENT_GROUP && placementGroups.length === 0
-            ? "No placement groups found for this project."
-            : ""
-        }
-        title={missingImageTitle}
-      />
-      {type === TARGET.CLUSTER_GROUP && (
-        <CustomSelect
-          id="clusterGroup"
-          label="Cluster group"
-          wrapperClassName="select-input"
-          dropdownClassName="instance-target-dropdown"
-          selectRef={clusterGroupRef as SelectRef}
-          onChange={setClusterGroup}
-          value={selectedClusterGroup(formik.values.target)}
-          options={clusterGroupOptions}
-          header={
-            <div className="header">
-              <span className="name">Name</span>
-              <span className="members">Members</span>
-            </div>
-          }
-          disabled={!formik.values.image}
-        />
-      )}
-      {type === TARGET.CLUSTER_MEMBER && (
-        <CustomSelect
-          id="clusterMember"
-          label="Cluster member"
-          wrapperClassName="select-input"
-          dropdownClassName="instance-target-dropdown"
-          selectRef={clusterMemberRef as SelectRef}
-          onChange={setClusterMember}
-          value={selectedClusterMember(formik.values.target)}
-          options={clusterMemberOptions}
-          disabled={!formik.values.image}
-        />
-      )}
-      {type === TARGET.PLACEMENT_GROUP && (
-        <PlacementGroupSelect
-          value={formik.values.placementGroup}
-          setValue={setPlacementGroup}
-          project={project?.name ?? "default"}
-          ref={placementGroupRef as SelectRef}
-          disabled={!formik.values.image}
-          profileNames={formik.values.profiles}
-          isCreateInstance
-        />
-      )}
-      {canAnchorPlacementGroup && (
-        <CustomSelect
-          id="clusterMember"
-          label="Anchor"
-          wrapperClassName="select-input"
-          dropdownClassName="instance-target-dropdown"
-          selectRef={clusterMemberRef as SelectRef}
-          onChange={(value) => {
-            const target = value ?? undefined;
-            formik.setFieldValue("target", target);
+            if (type === TARGET.CLUSTER_GROUP) {
+              setClusterGroup(clusterGroupOptions[0].value);
+              setTimeout(() => clusterGroupRef.current?.open(), 100);
+            }
+            if (type === TARGET.CLUSTER_MEMBER) {
+              setClusterMember(clusterMemberOptions[0].value);
+              setTimeout(() => clusterMemberRef.current?.open(), 100);
+            }
+            if (type === TARGET.PLACEMENT_GROUP) {
+              setPlacementGroup(placementGroups[0]?.name ?? "-");
+              if (placementGroups.length > 0) {
+                setTimeout(() => placementGroupRef.current?.open(), 100);
+              }
+            }
           }}
-          value={selectedClusterMember(formik.values.target)}
-          options={[{ label: "Auto", value: "" }].concat(
-            ...clusterMemberOptions,
-          )}
           disabled={!formik.values.image}
-          help="Cluster member to anchor the placement group"
+          error={
+            type === TARGET.PLACEMENT_GROUP && placementGroups.length === 0
+              ? "No placement groups found for this project."
+              : ""
+          }
+          title={missingImageTitle}
         />
+        {type === TARGET.CLUSTER_GROUP && (
+          <CustomSelect
+            id="clusterGroup"
+            label={
+              <ClusterGroupExplanationTooltip>
+                Cluster group
+              </ClusterGroupExplanationTooltip>
+            }
+            wrapperClassName="select-input"
+            dropdownClassName="instance-target-dropdown"
+            selectRef={clusterGroupRef as SelectRef}
+            onChange={setClusterGroup}
+            value={selectedClusterGroup(formik.values.target)}
+            options={clusterGroupOptions}
+            header={
+              <div className="header">
+                <span className="name">Name</span>
+                <span className="members">Members</span>
+              </div>
+            }
+            disabled={!formik.values.image}
+          />
+        )}
+        {type === TARGET.CLUSTER_MEMBER && (
+          <CustomSelect
+            id="clusterMember"
+            label={
+              <ClusterMemberExplanationTooltip>
+                Cluster member
+              </ClusterMemberExplanationTooltip>
+            }
+            wrapperClassName="select-input"
+            dropdownClassName="instance-target-dropdown"
+            selectRef={clusterMemberRef as SelectRef}
+            onChange={setClusterMember}
+            value={selectedClusterMember(formik.values.target)}
+            options={clusterMemberOptions}
+            disabled={!formik.values.image}
+          />
+        )}
+        {type === TARGET.PLACEMENT_GROUP && (
+          <PlacementGroupSelect
+            value={formik.values.placementGroup}
+            setValue={setPlacementGroup}
+            project={project?.name ?? "default"}
+            ref={placementGroupRef as SelectRef}
+            disabled={!formik.values.image}
+            profileNames={formik.values.profiles}
+            isCreateInstance
+          />
+        )}
+        {canAnchorPlacementGroup && (
+          <CustomSelect
+            id="clusterMember"
+            label="Anchor"
+            wrapperClassName="select-input"
+            dropdownClassName="instance-target-dropdown"
+            selectRef={clusterMemberRef as SelectRef}
+            onChange={(value) => {
+              const target = value ?? undefined;
+              formik.setFieldValue("target", target);
+            }}
+            value={selectedClusterMember(formik.values.target)}
+            options={[{ label: "Auto", value: "" }].concat(
+              ...clusterMemberOptions,
+            )}
+            disabled={!formik.values.image}
+            help="Cluster member to anchor the placement group"
+          />
+        )}
+      </div>
+      {type === TARGET.AUTO && (
+        <p className="p-form-help-text">
+          LXD automatically selects a cluster member for the instance, unless a
+          target is specified.
+        </p>
       )}
-    </div>
+    </>
   );
 };
 
--- a/src/pages/profiles/ProfileExplanationTooltip.tsx
+++ b/src/pages/profiles/ProfileExplanationTooltip.tsx
@@ -3,10 +3,16 @@ import ExplanationTooltip from "components/ExplanationTooltip";
 
 const ProfileExplanationTooltip: FC<{
   children?: ReactNode;
-}> = ({ children }) => {
+  additionalInformation?: string;
+}> = ({ children, additionalInformation }) => {
   return (
     <ExplanationTooltip
-      explanation="Profiles are configuration templates for instances."
+      className="explanation-tooltip-wrapper--inline"
+      explanation={
+        additionalInformation
+          ? additionalInformation
+          : "Profiles are configuration templates for instances."
+      }
       docPath="/profiles/"
       docLabel="Learn more about profiles"
     >
--- a/src/pages/profiles/ProfileSelector.tsx
+++ b/src/pages/profiles/ProfileSelector.tsx
@@ -6,11 +6,11 @@ import {
   Select,
   useNotify,
   Spinner,
-  Tooltip,
 } from "@canonical/react-components";
 import { defaultFirst } from "util/helpers";
 import { focusField } from "util/formFields";
 import { useProfiles } from "context/useProfiles";
+import ProfileExplanationTooltip from "pages/profiles/ProfileExplanationTooltip";
 
 interface Props {
   project: string;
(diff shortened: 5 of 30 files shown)


sophiecarroll:
Pasting here for easier visibility (well, sorry the screenshots didn't paste)


Are the tooltips up to date in this demo? I have a few short fixes @Kxiru

Typo: "configuration" should be "configurations"
Screenshot 2026-08-24 at 10 42 19 AM
Overflow bug, tooltip hidden behind nav bar
Screenshot 2026-08-24 at 10 42 00 AM
definition of SSH key doesn't appear at the beginning of this tooltip
Screenshot 2026-08-24 at 10 43 10 AM
The tooltip here should be about "project configuration" and not "projects"–I think this is one of those times to add additional information after the definition of projects, like "Projects organise and isolate LXD resources into separate groups. Manage project configuration settings here."
Screenshot 2026-08-24 at 10 46 52 AM

Kxiru:
> Pasting here for easier visibility (well, sorry the screenshots didn't paste)

All of these have been addressed :)

src/sass/_explanation_tooltip.scss
```diff
@@ -9,7 +9,7 @@
   width: 1rem !important;
 }
 
-.explanation-tooltip-wrapper--breadcrumb {
+.explanation-tooltip-wrapper--inline {
```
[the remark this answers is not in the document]
Kxiru: Only one component uses this component (Projct header) and it has already been upadated.

tests/a11y-audit.spec.ts
```diff
@@ -699,7 +699,7 @@ test.describe("permissions", () => {
     const page = await browser.newPage();
     await visitIdpGroups(page);
     await deleteIdpGroup(page, idpGroup);
-    await deleteIdentity(page, identity);
+    await deleteIdentity(page, identity, "Client certificate");
```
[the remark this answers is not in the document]
Kxiru: Drive by fix so that the broken lint-js in main works.

src/components/forms/SshKeyExplanationTooltip.tsx
edlerd, line 1: Not related to this file, but the one in the screenshot was forgotten? I think we should use the standard text for placement groups.

Image

src/pages/profiles/ProfileSelector.tsx
```diff
@@ -67,20 +67,14 @@ const ProfileSelector: FC<Props> = ({
...
-              <br />
-              Profiles lower in this list take precedence and override values
-              from profiles above.
-            </>
-          }
+        <ExplanationTooltip
+          className="explanation-tooltip-wrapper--inline"
+          explanation="A profile stores a set of configuration, such as instance and device options. Profiles lower in this list take precedence and override values from profiles above."
+          docPath="/profiles/"
+          docLabel="Learn more about profiles"
```
edlerd: 1. We shouldn't duplicate the text already in `ProfileExplanationTooltip`. Otherwise, they will diverge, whereas they should align with future changes, creating a maintenance burden. Let's extend the component to allow passing an optional extra sentence instead.

2. The layout is producing a way too long tooltip. There should be line breaks after each sentence:

Image


3. The placement below hides the options. I think in this case we should use the placement like in the old behavior. It seems better suited:

Image

src/pages/projects/forms/ProjectDetailsForm.tsx
```diff
@@ -148,7 +150,11 @@ const ProjectDetailsForm: FC<Props> = ({ formik, project, isEdit }) => {
                 )
               }
               selectProps={{
-                label: "Default instance storage pool",
+                label: (
+                  <StoragePoolExplanationTooltip className="explanation-tooltip-wrapper--inline">
```
edlerd: This one and the default network one are missing in read mode

Image

src/pages/projects/ProjectConfigurationHeader.tsx
```diff
@@ -124,7 +124,7 @@ const ProjectConfigurationHeader: FC<Props> = ({ project }) => {
     <RenameHeader
       name={project.name}
       parentItems={[
-        <ProjectExplanationTooltip key="project-configuration">
+        <ProjectExplanationTooltip key="project-configuration" isConfigVariant>
```
edlerd: We should not produce such a long tooltip

Image

src/pages/storage/forms/StorageBucketForm.tsx
```diff
@@ -56,7 +57,11 @@ const StorageBucketForm: FC<Props> = ({ formik, bucket }) => {
           })}
           selectProps={{
             id: "bucket-create-pool",
-            label: "Storage pool",
+            label: (
+              <StoragePoolExplanationTooltip className="explanation-tooltip-wrapper--inline">
+                Storage pool
+              </StoragePoolExplanationTooltip>
```
edlerd: Placement above would be nicer here, not to hide the options or error message on the field below.

Image

src/pages/storage/forms/StorageDriverSelect.tsx
```diff
@@ -146,7 +147,11 @@ const StorageDriverSelect: FC<Props> = ({ formik }) => {
         <CustomSelect
           id="driver"
           name="driver"
-          label="Driver"
+          label={
+            <StorageDriverExplanationTooltip>
+              Driver
+            </StorageDriverExplanationTooltip>
```
edlerd, line 160: Position seems not ideal. Also I wonder if the message gives any extra information?

Image

tests/a11y-audit.spec.ts
edlerd: Please rebase on main, then this change not necessary and can be reverted.

edlerd (changes requested):

src/components/forms/SshKeyExplanationTooltip.tsx
Kxiru: It wasn't necessarily forgotten. I didn't include it as "Target" isn't an LXD Entity. This helptext is the original and was not touched or affected by any of the ET PRs.

So how would you like to proceed here? Create a Target ET with the same text (replacing the PG copy with what we have for PG right now?)

src/pages/storage/forms/StorageDriverSelect.tsx
Kxiru: I think the help-link is the most useful part of the ET.

src/components/forms/SshKeyExplanationTooltip.tsx
edlerd: Let's use the placement group ET with some extra text of "LXD automatically selects a cluster member for the instance.".

src/pages/storage/forms/StorageDriverSelect.tsx
edlerd: True, the link is useful :+1: 
Then let's maybe reconsider the default placement of the ET to be above. Wdyt?

sophiecarroll: The tooltip should explain the definition of driver at the beginning right to be consistent?

>Then let's maybe reconsider the default placement of the ET to be above. Wdyt?

Agreed for forms–user is focusing on selecting driver more than description at this point.

src/pages/profiles/ProfileSelector.tsx
Kxiru: > The layout is producing a way too long tooltip. There should be line breaks after each sentence:

There are certainly some long sentences in the explanations, so I would propose a max-width for the tooltip rather than line breaks at different sentences.

src/pages/instances/TargetExplanationTooltip.tsx
```diff
@@ -0,0 +1,24 @@
...
+}> = ({ children, hasAdditionalInformation }) => {
+  return (
+    <ExplanationTooltip
+      className="explanation-tooltip-wrapper--inline"
+      explanation={
+        hasAdditionalInformation
+          ? "Placement groups define rules for placing instances across cluster members. LXD automatically selects a cluster member for the instance."
+          : "Placement groups define rules for placing instances across cluster members."
+      }
+      docPath="/explanation/clusters/#automatic-placement-of-instances/"
```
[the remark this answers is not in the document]
edlerd: This need sto be fixed.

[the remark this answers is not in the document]
Kxiru: Sure, though I am waiting on the outcome of[ this mattermost thread ](https://chat.canonical.com/canonical/pl/q34zd1zdypnjtcjr6u1c54huwe) to continue this PR.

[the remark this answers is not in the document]
Kxiru: I'll push this fix now just to get rid of the error.

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -176,134 +174,133 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
...
+            disabled={!formik.values.image}
+            help="Cluster member to anchor the placement group"
+          />
+        )}
+      </div>
+      {type === TARGET.AUTO && (
+        <p className="p-form-help-text">
+          LXD automatically selects a cluster member for the instance, unless
+          target is specified.
+        </p>
```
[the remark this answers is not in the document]
Kxiru: @sophiecarroll Adding an "a" here to clarify meaning -- "--unless a target is specified".

src/pages/projects/ProjectExplanationTooltip.tsx
```diff
@@ -3,11 +3,16 @@ import ExplanationTooltip from "components/ExplanationTooltip";
 
 const ProjectExplanationTooltip: FC<{
   children?: ReactNode;
-}> = ({ children }) => {
+  hasAdditionalInformation?: boolean;
+}> = ({ children, hasAdditionalInformation }) => {
   return (
```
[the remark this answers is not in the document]
Kxiru: Fix addressed in followup push. Note to @edlerd , the Project configuration page ET has been removed as Design has said that we cannot have ETs in breadcrumb headers.

src/pages/storage/StoragePoolExplanationTooltip.tsx
```diff
@@ -1,20 +1,23 @@
...
+}> = ({ children, className, hasAdditionalInformation }) => {
   return (
     <ExplanationTooltip
+      className={classNames("explanation-tooltip-wrapper--inline", className)}
       explanation="Storage pools host instance and image data."
       docPath={
-        isConfigVariant
+        hasAdditionalInformation
           ? "/reference/storage_drivers/"
           : "/explanation/storage/"
```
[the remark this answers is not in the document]
Kxiru: This is intentional.

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -176,134 +174,133 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
...
+            ref={placementGroupRef as SelectRef}
+            disabled={!formik.values.image}
+            profileNames={formik.values.profiles}
+            isCreateInstance
+          />
+        )}
+        {canAnchorPlacementGroup && (
+          <CustomSelect
+            id="clusterMember"
+            label="Anchor"
```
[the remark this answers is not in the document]
Kxiru: This logic was not changed by this PR.

src/components/forms/SshKeyExplanationTooltip.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const SshKeyExplanationTooltip: FC<{
+  children?: ReactNode;
+}> = ({ children }) => {
+  return (
+    <ExplanationTooltip
+      className="explanation-tooltip-wrapper--inline"
+      explanation="Cloud init must be enabled on the instance to apply the keys. Additional keys get applied on instance creation or restart. SSH keys are not removed automatically."
+      docPath="/reference/instance_options/"
+      docLabel="Learn more about instance options"
```
edlerd:
```suggestion
      docPath="/reference/instance_options/#instance-cloud-init:cloud-init.ssh-keys.KEYNAME"
      docLabel="Learn more about instance options"
```

Shall we link to the detail section for ssh keys? The page is rather long otherwise.

We might also adjust the doc label.

src/pages/instances/InstanceExplanationTooltip.tsx
```diff
@@ -3,18 +3,19 @@ import ExplanationTooltip from "components/ExplanationTooltip";
 
 const InstanceExplanationTooltip: FC<{
   children?: ReactNode;
-  isConfigVariant?: boolean;
-}> = ({ children, isConfigVariant }) => {
+  hasAdditionalInformation?: boolean;
+}> = ({ children, hasAdditionalInformation }) => {
```
edlerd: The rename here is confusing. Wasn't the old name better suited?

src/pages/profiles/ProfileExplanationTooltip.tsx
```diff
@@ -3,10 +3,16 @@ import ExplanationTooltip from "components/ExplanationTooltip";
...
-}> = ({ children }) => {
+  hasAdditionalInformation?: boolean;
+}> = ({ children, hasAdditionalInformation }) => {
   return (
     <ExplanationTooltip
-      explanation="Profiles are configuration templates for instances."
+      className="explanation-tooltip-wrapper--inline"
+      explanation={
+        hasAdditionalInformation
+          ? "Profiles are configuration templates for instances. Profiles lower in this list take precedence and override values from profiles above."
```
edlerd: This `hasAdditionalInformation` seems very specific for a single use case. Shall we pass in an `additionalInformation` string instead to keep the component not too opinionated about that specific use case?

src/pages/projects/ProjectExplanationTooltip.tsx
```diff
@@ -6,8 +6,8 @@ const ProjectExplanationTooltip: FC<{
 }> = ({ children }) => {
   return (
     <ExplanationTooltip
-      className="explanation-tooltip-wrapper--breadcrumb"
-      explanation="Projects organize and isolate LXD resources into separate groups."
+      className="explanation-tooltip-wrapper--inline"
+      explanation="Projects organise and isolate LXD resources into separate groups."
```
edlerd:
```suggestion
      explanation="Projects organize and isolate LXD resources into separate groups."
```
I think we want the american english as standard, no?

src/pages/storage/forms/StorageBucketForm.tsx
```diff
@@ -41,12 +42,19 @@ const StorageBucketForm: FC<Props> = ({ formik, bucket }) => {
...
       {isEditing ? (
-        <OutputField
-          id="storage_bucket_pool"
-          label="Storage pool"
-          value={formik.values.pool}
-          help="Storage bucket pool can't be changed"
-        />
+        <>
+          <Label forId="storage_bucket_pool">
+            <StoragePoolExplanationTooltip>
```
edlerd, line 47: In this example it is a bit confusing, because the tooltip does not mention storage buckets. Not sure what is best to do here or maybe okay as is?

Image

src/pages/storage/forms/StorageDriverSelect.tsx
```diff
@@ -136,17 +137,28 @@ const StorageDriverSelect: FC<Props> = ({ formik }) => {
...
       {!formik.values.isCreating ? (
-        <OutputField
-          id={"driver"}
-          label={"Driver"}
-          value={storageDriverLabels[formik.values.driver]}
-          help={getHelpText()}
-        />
+        <>
+          <Label forId="driver">
+            <StorageDriverExplanationTooltip>
```
edlerd, line 142: This one is super nice to have a quick link to the docs!

src/pages/storage/forms/StorageBucketForm.tsx
edlerd: This also surfaces on volume creation/editing where we use the same tooltip. Maybe we should change it to "Storage pools host instance and other data." Wdyt @sophiecarroll ?

src/pages/storage/StoragePoolExplanationTooltip.tsx
```diff
@@ -1,20 +1,23 @@
...
 import type { FC, ReactNode } from "react";
 import ExplanationTooltip from "components/ExplanationTooltip";
 
 const StoragePoolExplanationTooltip: FC<{
   children?: ReactNode;
-  isConfigVariant?: boolean;
-}> = ({ children, isConfigVariant }) => {
+  className?: string;
+  hasAdditionalInformation?: boolean;
+}> = ({ children, className, hasAdditionalInformation }) => {
```
edlerd: Why the rename to `hasAdditionalInformation`? The previous name seems better suited.

src/pages/instances/forms/InstanceTargetSelect.tsx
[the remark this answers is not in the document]
sophiecarroll: Alright sounds good

src/pages/storage/forms/StorageBucketForm.tsx
sophiecarroll: Does it mean that with the added context of those two places, Storage pools host instance, image data, volume data, and storage bucket data? If storage pools host more than just instance and image data it does seem better to expand the definition. The wording of "instance and other data" is sounding a bit unnatural to me–maybe something like "Storage pools host data from instances, images, and more." ?

edlerd: > Does it mean that with the added context of those two places, Storage pools host instance, image data, volume data, and storage bucket data? I

Yes, exactly.

> Storage pools host data from instances, images, and more.

Sounds much better, indeed.

src/pages/storage/StoragePoolExplanationTooltip.tsx
Kxiru: Because not every ET with supplementary information is a config variant of an entity.

edlerd: I think the interfaces don't have to be the same. In this case here it is about changing to the config page in the docs, so I think the rename makes it a bit more confusing.

src/pages/instances/InstanceExplanationTooltip.tsx
Kxiru: I renamed isConfigVariant to hasAdditionalInformation across all entity tooltips as not all ETs with additional information are configuration variants.

src/pages/profiles/ProfileExplanationTooltip.tsx
Kxiru: I'm not sure I understand. Are you suggesting passing `additionalInformation` instead of `hasAdditionalInformation`, or in addition to the legacy `isConfigVariant`?

src/pages/storage/forms/StorageBucketForm.tsx
Kxiru: Alrighty, thnaks both! I'll change this.

src/pages/storage/forms/StorageDriverSelect.tsx
Kxiru: I agree! It's very helpful :)

src/pages/storage/StoragePoolExplanationTooltip.tsx
Kxiru: Okay, I will rename hasAdditionalInformation back is isConfigVariant, and only use hasAdditionalInformation for Profiles.

Kxiru: I wouldn't say that a storage driver is a config variant of a storage pool.

src/components/forms/SshKeyExplanationTooltip.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const SshKeyExplanationTooltip: FC<{
+  children?: ReactNode;
+}> = ({ children }) => {
+  return (
+    <ExplanationTooltip
+      className="explanation-tooltip-wrapper--inline"
+      explanation="Cloud init must be enabled on the instance to apply the keys. Additional keys get applied on instance creation or restart. SSH keys are not removed automatically."
+      docPath="/reference/instance_options/#instance-cloud-init:cloud-init.ssh-keys.KEYNAME"
+      docLabel="Learn more about SSH keys"
```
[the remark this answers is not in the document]
Kxiru: The provided link is valid to a section of the page.

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -176,134 +174,133 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
...
+                <span className="members">Members</span>
+              </div>
+            }
+            disabled={!formik.values.image}
+          />
+        )}
+        {type === TARGET.CLUSTER_MEMBER && (
+          <CustomSelect
+            id="clusterMember"
+            label="Cluster member"
```
edlerd: Did we consider using the ET for Cluster members here?

Image

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -176,134 +174,133 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
...
+            dropdownClassName="instance-target-dropdown"
+            selectRef={clusterMemberRef as SelectRef}
+            onChange={setClusterMember}
+            value={selectedClusterMember(formik.values.target)}
+            options={clusterMemberOptions}
+            disabled={!formik.values.image}
+          />
+        )}
+        {type === TARGET.PLACEMENT_GROUP && (
+          <PlacementGroupSelect
```
edlerd, line 281: Did we consider using the ET for Placement groups here?

Image

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -176,134 +174,133 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
...
+            type === TARGET.PLACEMENT_GROUP && placementGroups.length === 0
+              ? "No placement groups found for this project."
+              : ""
+          }
+          title={missingImageTitle}
         />
+        {type === TARGET.CLUSTER_GROUP && (
+          <CustomSelect
+            id="clusterGroup"
+            label="Cluster group"
```
edlerd: Did we consider using the ET for cluster groups here?

Image

src/pages/profiles/ProfileExplanationTooltip.tsx
edlerd: > Are you suggesting passing additionalInformation instead of hasAdditionalInformation


Yes, exactly. Instead of defining what additional information is, we allow to pass it in dynamically. Making the component easier to extend later on. Currently this component is very tailored towards a single use case and only supports additional information for that very case.

edlerd (commented):
Mostly LGTM, few questions on the target selector and one on the earlier discussion around structure for profiles.

src/pages/instances/forms/InstanceTargetSelect.tsx
Kxiru: I did with @sophiecarroll , but the pushback was that users in a clustered environment would already know what these are and that we may end up "overusing" ETs by applying them everywhere. I see both sides, and personally vote in favour of having them there for the 3 non-auto targets. Sophie, do you have anything to add?

src/pages/instances/forms/InstanceTargetSelect.tsx
Kxiru: As above.

src/pages/instances/forms/InstanceTargetSelect.tsx
Kxiru: As above.

src/pages/instances/forms/InstanceTargetSelect.tsx
sophiecarroll: My perspective was that we should put the ETs where the user would first discover the entity–on the main instances page, for example, but not in sub-forms that mention instances–because ETs are for learning and onboarding, and the user is probably beyond the level of first contact with the entity if they're doing more detailed work. It also avoids us overusing tooltips and being inconsistent, because if we put ETs everywhere the entity is mentioned in some pages and not in all it could be confusing, and if we put them in every subform every time they're mentioned, it could look cluttered. However, if it feels very helpful in this case or if you think people might not have learned what these are a previous time, I'm open to the possibility. @edlerd what do you think?

sophiecarroll: If the ETs were there, would it be for all three non-auto options that you can select? Or just cluster member?

Kxiru: > If the ETs were there, would it be for all three non-auto options that you can select? Or just cluster member?

The relevant ET would be there for the Cluster member, Placement group and cluster group, within their respective sections.

edlerd: Yes, I was suggesting it for all three. I suspect this could be the first time a user encounters the placement options, so I'm not sure if users are familiar with cluster members, placement groups, and cluster groups at that time.

Kxiru: I'll put them in :)

src/pages/instances/forms/InstanceTargetSelect.tsx
```diff
@@ -176,134 +174,133 @@ const InstanceTargetSelect: FC<Props> = ({ formik }) => {
...
-          profileNames={formik.values.profiles}
-          isCreateInstance
-        />
-      )}
-      {canAnchorPlacementGroup && (
-        <CustomSelect
-          id="clusterMember"
-          label="Anchor"
+    <>
+      <div className="instance-target-selector">
```
edlerd, line 184: On edit there is also a dedicated placement group selector, which is also appearing on profile edit. We might want to add the ET there as well:

Image

edlerd: Also I think the help text under the placement group selector reads not so ideal to me, maybe we can refine it?

src/pages/instances/forms/InstanceTargetSelect.tsx
sophiecarroll: Alright, sounds good

src/pages/instances/forms/InstanceTargetSelect.tsx
sophiecarroll: Yeah hmm I'm not sure 100% what it's trying to say, maybe we I can quickly in standup and we can rewrite it

src/pages/profiles/ProfileExplanationTooltip.tsx
edlerd: This seems unresolved.

edlerd (dismissed):
One small thing got overlooked in the last iteration.

src/pages/instances/forms/InstanceTargetSelect.tsx
sophiecarroll: @edlerd @Kxiru  Possible new copy for under the placement group selector: what do you guys think?

New: Assigning a placement group will not immediately move this instance to the member that satisfies the rules of the placement group. Any changes will apply when LXD scheduling events such as evacuation occur.

Old: Update does not move the instance. Changed placement policy applies only to future LXD scheduling events such as evacuation. Manage placement groups for this project.

edlerd: sounds great :+1: 

@Kxiru can we adjust the copy in the scope of this PR?

src/pages/profiles/ProfileExplanationTooltip.tsx
Kxiru: Ah, I misunderstood. I thought you were asking to only use hasAdditionalInformation on Profiles, hence it being the only use case. Then I thought you just wanted hasAdditionalInformation renamed to additionalInformation, but what you're actually asking for is hasAdditionalInformation being a string prop, not a boolean (and being renamed in the process). Okay, I'll address that now.

Kxiru:
> One small thing got overlooked in the last iteration.

This has now been addressed, I misunderstood the instruction.

src/pages/profiles/ProfileExplanationTooltip.tsx
```diff
@@ -3,10 +3,16 @@ import ExplanationTooltip from "components/ExplanationTooltip";
 
 const ProfileExplanationTooltip: FC<{
   children?: ReactNode;
-}> = ({ children }) => {
+  hasAdditionalInformation?: string;
```
edlerd:
```suggestion
  additionalInformation?: string;
```
A better suiting name?

edlerd (dismissed):

src/pages/profiles/ProfileExplanationTooltip.tsx
Kxiru: Yes, thank you. Pushed.

edlerd (approved):

PR #2216, opened 2026-08-21
feat: [WD-38208] Set a11y test threshold

Kxiru:
**##** Done

- Sets max violation (1) and minimum passing rate (90%) constants (Temporarily set to 3 and 80% respectively so that CI does not fail).
- Formats violations and outputs them. A throw happens after the report file is written so the artifat is always available for inspection even when the test fails.

## QA

1. Run the LXD-UI:
    - On the demo server via the link posted by @webteam-app below. This is only available for PRs created by collaborators of the repo. Ask @Kxiru or @edlerd for access.
    - With a local copy of this branch, [build and run as described in the docs](https://github.com/canonical/lxd-ui/blob/main/CONTRIBUTING.md#setting-up-for-development).
2. Perform the following QA steps:
- N/A
## Screenshots

N/A

PR #2251, opened 2026-09-01
feat(a11y): [WD-38581] add instance detail tab a11y and ensure list page content

Kxiru:
## Done

~~**https://github.com/canonical/lxd-ui/pull/2216 must be merged before this PR**~~ Can now be merged

- Adds a11y-audit coverage for all 6 instance detail page tabs: Configuration, Snapshots, File explorer, Terminal, Console, and Logs.
- Uses createAndStartInstance to support interactive tabs.
- Ensures ACL & ISO list pages are audited with actual content rather than empty states by creating a network ACL and uploading a custom ISO in test setup.

## QA

1. Run the LXD-UI:
    - Observe test suite or run tests locally

## Screenshots

N/A

### DIFF
--- a/tests/a11y-audit.spec.ts
+++ b/tests/a11y-audit.spec.ts
@@ -19,7 +19,7 @@ import {
   visitImageRegistry,
 } from "./helpers/image-registries";
 import {
-  createInstance,
+  createAndStartInstance,
   deleteInstance,
   randomInstanceName,
   visitInstance,
@@ -33,8 +33,14 @@ import {
   visitNetwork,
 } from "./helpers/network";
 import {
+  createNetworkAcl,
+  deleteNetworkAcl,
+  randomNetworkAclName,
+} from "./helpers/network-acls";
+import {
   createVolume,
   deleteVolume,
+  randomIsoName,
   randomVolumeName,
   visitVolume,
 } from "./helpers/storageVolume";
@@ -67,6 +73,7 @@ import {
   randomIdpGroupName,
   visitIdpGroups,
 } from "./helpers/permission-idp-groups";
+import { createCustomISO, deleteCustomISO } from "./helpers/custom-isos";
 
 test.describe("instances", () => {
   const instance = randomInstanceName();
@@ -78,7 +85,7 @@ test.describe("instances", () => {
       return;
     }
     const page = await browser.newPage();
-    await createInstance(page, instance);
+    await createAndStartInstance(page, instance);
     await createPool(page, pool);
     await createProject(page, project);
     await page.close();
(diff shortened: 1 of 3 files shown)


edlerd (changes requested):
There are unrelated commits in this PR. Also, the changes about custom ISO create/cleanup and network ACL create/cleanup seem unrelated and should be removed from this PR.

Kxiru:
> There are unrelated commits in this PR. Also, the changes about custom ISO create/cleanup and network ACL create/cleanup seem unrelated and should be removed from this PR.

This PR serves to add a11y for the detail tab, as well as ensure that there is content in some of the lists that we run a11y on. Namely the instance list, network list and custom ISO list, as these are easy to create entities for.

As I am restructuring the PR. I will put the list additions in a separate commit so that if you no longer want them, they can be removed easily.

The PR description depicts all of the changes in the PR and reasoning.

tests/a11y-audit.spec.ts
```diff
@@ -338,6 +394,18 @@ test.describe("storage", () => {
...
+    await gotoURL(page, "/ui/");
+    await page.getByRole("button", { name: "Storage", exact: true }).click();
+    await page.getByRole("link", { name: "Custom ISOs" }).click();
+    await page.getByRole("button", { name: "Upload custom ISO" }).click();
+    await page.getByLabel("Local file").setInputFiles(ISO_FILE);
+    await page.getByLabel("Alias").fill(isoName);
+    await page.getByRole("button", { name: "Upload", exact: true }).click();
+    await page
+      .getByText(`Custom ISO ${isoName} uploaded successfully`)
+      .waitFor();
```
edlerd: Let's not copy this here, but extract this into a `createCustomIso` helper to be used here and in the place you took it from.

tests/a11y-audit.spec.ts
```diff
@@ -346,6 +414,15 @@ test.describe("storage", () => {
...
     const page = await browser.newPage();
+
+    await gotoURL(page, "/ui/");
+    await page.getByRole("button", { name: "Storage", exact: true }).click();
+    await page.getByRole("link", { name: "Custom ISOs" }).click();
+    await page.getByPlaceholder("Search for custom ISOs").fill(isoName);
+    await page.getByRole("button", { name: "Delete" }).click();
+    await page.getByText("Delete", { exact: true }).click();
+    await page.getByText(`Custom iso ${isoName} deleted.`).waitFor();
+
```
edlerd: This should use a helper as well.

tests/a11y-audit.spec.ts
```diff
@@ -242,6 +248,51 @@ test.describe("instances", () => {
...
+    skipIfNotA11yProject(testInfo.project.name);
+    await visitInstance(page, instance);
+    await page.getByTestId("tab-link-Snapshots").click();
+    await runA11yAudit(page, testInfo);
+  });
+
+  test("file explorer tab", async ({ page }, testInfo) => {
+    skipIfNotA11yProject(testInfo.project.name);
+    await visitInstance(page, instance);
+    await page.getByTestId("tab-link-File explorer").click();
```
edlerd: We should generally avoid using `getBytestId` and use a selector that uses something like `getByRole("tab", { name: "File Explorer" })`. Also, we should ensure we are landing on the tab before starting the a11y audit. For this occasion here, we can use the `visitFileExplorer` helper. For the others below, we might need to introduce `visit...` helpers to use here and can fix the existing use cases where we navigate to the same tabs to use the new helpers as well.

edlerd (changes requested):

tests/a11y-audit.spec.ts
edlerd: The advantage of `getByRole("tab", { name: "File Explorer" })` is that it ensures an accessibility-compliant presentation of the tabs in the test instead of building a test that is tied to the implementation.

edlerd (approved):
LGTM, thanks.

PR #2274, opened 2026-09-04
feat(Overview): Clustering card if 1 or more member is offline

kimanhou:
## Done

- feat(Overview): Clustering card if 1 or more member is offline
- Display status count for all statuses in a fixed order (online, evacuated, offline, blocked)
- No changes to the Server card (if unclustered)

## QA

1. Run the LXD-UI:
    - Go to Overview page in a clustered environment with at least one offline member

## Screenshots

image

If all members are online
image

Unclustered (unchanged)
Screenshot from 2026-09-04 17-02-33


sophiecarroll:
Looks great! Make sure memory status bar says "x GiB of x GiB used (X%)"–right now it's missing "used". Added my +1 for the members offline text

kimanhou:
@sophiecarroll 
> Make sure memory status bar says "x GiB of x GiB used (X%)"–right now it's missing "used"

I don't think this is a good idea

image

sophiecarroll:
> @sophiecarroll
> 
> > Make sure memory status bar says "x GiB of x GiB used (X%)"–right now it's missing "used"
> 
> I don't think this is a good idea
> 
> image

Can you expand on why? It's pretty standard to clarify if X of Y in a status bar means X is used or X is free, otherwise it can cause some confusion, and it doesn't make any big changes here. It also is consistent with our other status bars in LXD

Screenshot 2026-09-04 at 11 22 29 AM

kimanhou:
@sophiecarroll Breaking into two lines generates other issues: it makes the row bigger, which makes the table bigger, which messes the fixed height I set for pagination, meaning the whole screen will jump when we change page.

When the table switches the card layout, it gets truncated.
image

The Clustering card is already far from responsive, let's not add difficulties.

sophiecarroll:
> @sophiecarroll Breaking into two lines generates other issues: it makes the row bigger, which makes the table bigger, which messes the fixed height I set for pagination, meaning the whole screen will jump when we change page.
> 
> When the table switches the card layout, it gets truncated. image
> 
> The Clustering card is already far from responsive, let's not add difficulties.

Ok, I see your point. Can you use it for the overall cluster memory bar, and then not use it for memory within cluster members? That way we can at least set a precedent that the bar signifies used and stay somewhat consistent with the rest of the UI, while avoiding the line break problems.

src/pages/overview/ClusteringCard.tsx
```diff
@@ -36,17 +36,21 @@ const ClusteringCard: FC = () => {
...
+      )
+      .map(([status, count]) => (
+        <Fragment key={status}>
+          <Icon
+            name={getClusterMemberStatusIconName(
+              status as LxdClusterMemberStatus,
+            )}
+          />
+          {count} {status}
+          {status !== "Online" && " (not included in resource usage)"}
```
edlerd: Status can be `Evacuated` and in that case it would be included. Also, if we have a mix of online, offline, evacuated, and blocked, then this suffix would be repeated multiple times.

sophiecarroll: @edlerd @kimanhou (Kim Anh can we continue this conversation here from the conversation we were having in Jira)

I've just learned about the Evacuated and Blocked statuses. To clarify, Evacuated and Online are included in resource usage, but Offline and Blocked are not? I may have to rethink the "not included in resources totals message"

edlerd: > Evacuated and Online are included in resource usage, but Offline and Blocked are not?

Yes, I think that is correct. But haven't tested it.

sophiecarroll: Will we have the same tooltip as we use in instance details for these status bars? I know at one point we talked about it but I don't know if we implemented it
Screenshot 2026-09-04 at 1 05 01 PM

kimanhou: > Evacuated and Online are included in resource usage, but Offline and Blocked are not?

Euh no, that's not what I implemented

```
  const onlineMemberNames = members
    .filter((member) => member.status === "Online")
    .map((member) => member.server_name);
```

> Will we have the same tooltip as we use in instance details for these status bars?

What Meter are you talking about ? Total memory or cluster member memor ?

kimanhou: > this suffix would be repeated multiple times

Yeah I know, that's why I asked @sophiecarroll if it was okay.

> Status can be Evacuated and in that case it would be included

Then we need to change the calculation of total memory.

kimanhou:
@sophiecarroll 

> Can you use it for the overall cluster memory bar, and then not use it for memory within cluster members? 

https://warthogs.atlassian.net/browse/WD-38762

src/pages/overview/ClusteringCard.tsx
sophiecarroll: Ok let's clarify which statuses are included, and I was talking about cluster total memory. Once we know for sure that "Evacuated and Online are included in resource usage, but Offline and Blocked are not" is or isn't true I will give an informed design update to how we should display this because it would change based on how many statuses are included

kimanhou: Right now, only Online is included. But it can change. It's really up to us.

kimanhou: > Will we have the same tooltip as we use in instance details for these status bars? on total memory

It can be implemented. It's already implemented for the cluster member's memory.

edlerd: > Right now, only Online is included. But it can change. It's really up to us.

I see. I assumed the totals include any members who do reply on the state API call. In that case we can also only include online members. It might be better actually, as those are the ones that can run workloads.

kimanhou: It is clearer in the Notification 😜

image

sophiecarroll: Ok, if we decided to only include data from online members, I will get back to you shortly about the design solution to clarify that resource totals use only online members. I definitely understand that the notification banner is very visible, but the reason I am hesitant to use it is because notification banners should generally be used for updates or changes, and not be a permanent feature. Can I let you know as soon as I can so I can think about the best way to design it?
Screenshot 2026-09-04 at 2 30 39 PM

kimanhou: Yeah sure. I'll put this PR in draft in the meantime.

Indeed we use Notification wrong in LXD UI. Here we should be using Announcement, but it's not ready yet.

sophiecarroll: After discussing a bit with some other designers I came up with this! Information is visibly surfaced but doesn't indicate change or error state, with a clearer hierarchy (clustering=first level, resource usage=second level, total memory/total cpu and column headers of the members table=third level

Screenshot 2026-09-04 at 4 21 08 PM

sophiecarroll:
Everything looks great, but we should include all the statuses even if the value 0 because of the following reasons: this is a discussion I had with David in Mattermost

I think it's better to show all states for clarity (eg. if user sees "evacuated" sometimes but not others they might wonder if they're not seeing that some members are evacuated, or that it's hidden somehow), and also to be consistent with instances card (shows all statuses permanently), and also if the subheading grows shorter or longer that might complicate the design solution for showing which types of statuses are included in the resource totals

sophiecarroll:
Looks great! I'll give my design +1 now, my one question is about the tooltip–on the server version the link still links to clusters which I don't think is right–we also have this in the server side nav, I might take a look at this but I think it falls outside of the scope of the PR.

Screenshot 2026-09-14 at 2 45 30 PM
Screenshot 2026-09-14 at 2 48 59 PM

kimanhou:
@sophiecarroll It was fixed by @Kxiru last week.

image

sophiecarroll:
> @sophiecarroll It was fixed by @Kxiru last week.
> 
> image

Yes we decided on that solution together!

src/pages/overview/ClusteringCard.tsx
```diff
@@ -130,6 +130,7 @@ const ClusteringCard: FC = () => {
           name={getClusterMemberStatusIconName(
             status as LxdClusterMemberStatus,
           )}
+          className="status-icon"
```
edlerd, line 133: The "status" header is not aligned with the text of the status below in the table. Though not sure if in scope for this PR.

Image

edlerd (dismissed):
LGTM, tiny nitpick on the cluster member table.

src/pages/overview/ClusteringCard.tsx
kimanhou: Fixed

image

src/pages/overview/ClusteringCard.tsx
```diff
@@ -52,7 +52,7 @@ const ClusteringCard: FC = () => {
 
   const headers = [
     { content: "Members", sortKey: "name" },
-    { content: "Status", sortKey: "status" },
+    { content: "Status", sortKey: "status", className: "status-header" },
```
edlerd, line 55: There is no line between the headers and the table contents in the design doc, but here we still have it. I think the removal of that line was intentional, as the card already has the box.

current:

Image

design:

Image

edlerd (commented):
Found a design discrepancy and there is some formatting issues with CSS

edlerd (approved):
LGTM, thanks.

PR #2280, opened 2026-09-04
Feat: [WD-37402] Instances overview card

Kxiru:
## Done

- [Instances Card Spec](https://docs.google.com/document/d/1Myu3NOPXoLd_80-X_6tZnNOPNMS9gwX_ljMm71aCgAs/edit?tab=t.6eii77atqs9t)
- ~Instance Empty State~

## QA

1. Run the LXD-UI:

## Screenshots
image


sophiecarroll:
A few responsivity bugs with text/icons: icon shape is changing and text is being cut off


Screenshot 2026-09-11 at 5 32 34 PM
Screenshot 2026-09-11 at 5 32 24 PM

Also pie chart should be the same height as the table (see reference below)

Screenshot 2026-09-11 at 5 37 00 PM

Once that's figured out everything looks good and I will give the +1!

Kxiru:
Thank you, @sophiecarroll ! Do you mind taking another look now that I'm we're out of draft? :)

sophiecarroll:
> Thank you, @sophiecarroll ! Do you mind taking another look now that I'm we're out of draft? :)

That looks great! One more small thing: can you make sure the subheading is closer to the heading than the body of the card? eg. move up "x VMs • x Containers" towards the Instances heading to match the design a bit closer. Other than that everything looks good. I'll go ahead and give my +1

src/pages/overview/InstancesOverviewStatus.tsx
```diff
@@ -1,39 +1,26 @@
...
-      <Link
-        className="status-link p-link--soft"
-        to={getStatusFilterHref(
-          capitalizeFirstLetter(status) as LxdInstanceStatus,
-        )}
-      >
-        <strong className="status-count">{count}</strong>
-      </Link>
-    </div>
+    <ChartLegend
```
edlerd, line 15: Let's remove the `ChartLegend` and `InstancesOverviewStatus` components. Currently we create a whole legend as an unordered list with a single item in it for every single status entry in the table. That seems overly verbose, and we have all the blocks for a much simpler and consistent approach.

Instead, we should refactor this bit from `InstanceStatusIcon` into a dedicated component:

```
    <>
      <Icon
        name={getIconNameForStatus(instance.status)}
        className={classnames("status-icon", {
          "u-animation--spin": instance.status === "Freezing",
        })}
      />
      {instance.status}
    </>
```

That new component should receive the status as the only prop. So we can reuse it instead of `InstancesOverviewStatus` and inside `InstanceStatusIcon`.

src/pages/overview/InstancesCard.tsx
```diff
@@ -81,67 +77,70 @@ const InstancesCard: FC = () => {
...
+
+        <MainTable
+          className="overview-table group-by-status-table"
+          aria-label="Instances by status"
+          headers={[
+            { content: "Status" },
+            { content: "Instances", className: "u-align--right" },
+          ]}
+          rows={rows}
+          responsive
```
edlerd, line 142: This table gets very hard to read on smaller screens. Maybe without responsive it is better? We might move the table below the donut or hide the donut chart for responsiveness.

Image

src/pages/overview/InstancesCard.tsx
```diff
@@ -81,67 +77,70 @@ const InstancesCard: FC = () => {
...
+          `${containerCount} ${pluralize("container", containerCount)}`,
+        ]}
+        className="u-no-margin--bottom"
+      />
 
-        <div className="group-by-type-container">
+      <div className="card-content">
+        <div className="group-by-status-chart">
+          <p className="chart-title">Instances by status</p>
           <DoughnutChart
```
edlerd, line 124: I think the donut chart is supposed to stay aligned with the table height, but in mid-sized screens it is bigger:

Image

On large screens the heights match:

Image

src/util/overviewInstances.ts
```diff
@@ -51,3 +82,35 @@ export const getInstanceDistribution = (
...
+export const getInstanceStatusCounts = (
+  distribution: InstanceDistribution,
+): Record<OverviewInstanceStatus, number> => {
+  return {
+    running: distribution.runningCount,
+    stopped: distribution.stoppedCount,
+    frozen: distribution.frozenCount,
+    error: distribution.errorCount,
+  };
+};
```
edlerd, line 95:
```suggestion
```
Let's not have this translation layer. As we generate the data above, we can generate it in with the correct naming immediately.

edlerd (commented):
Due to the iteration on the initial implementation, some things are more complex than they have to be. A few suggestions below. There are also few things open in terms of layout / responsiveness.

src/pages/overview/InstancesCard.tsx
sophiecarroll: I like this idea–for most sizes we have the donut, but we can hide it if it gets small enough, and we still have the essential information. @Kxiru

src/pages/overview/InstancesCard.tsx
sophiecarroll: ^ Correct

src/pages/overview/InstancesCard.tsx
Kxiru: @sophiecarroll How would you like to approach this exactly? To remove the entire card on certain screens? I could equally just move the entire table to below the graph on some screens which I think would be better.

sophiecarroll: I am putting some responsivity notes in the spec now, will tag you with conclusions

#1: @Kxiru see comment below, but the bottom of the donut chart should be aligned with the bottom of the status table, which it is not as you make the screen smaller. This should fix at least some responsivity for the mid-sized card, as it breaks quite fast right now when the donut stays large.

sophiecarroll: @Kxiru I just made some comments and explained my reasoning in the spec so I'm not leaving a super long comment here! Action items are in bold.

This is the exact place in the spec:
https://docs.google.com/document/d/1Myu3NOPXoLd_80-X_6tZnNOPNMS9gwX_ljMm71aCgAs/edit?tab=t.6eii77atqs9t

src/pages/overview/InstancesCard.tsx
edlerd: This issue seems to still be open and unaddressed.

edlerd (commented):
Code looks much better, though there is one [issue with responsiveness](https://github.com/canonical/lxd-ui/pull/2280#discussion_r4007273269) that seems not addressed yet.

src/pages/overview/InstancesCard.tsx
```diff
@@ -81,67 +79,73 @@ const InstancesCard: FC = () => {
...
-                color: containerColor,
-                label: `${containerCount} ${pluralize("container", containerCount)}`,
-              },
-            ]}
+            chartID="dashboard-instances-by-status-doughnut-chart"
+            className="group-by-status-doughnut-chart"
           />
         </div>
+
+        <MainTable
```
edlerd: There is a lot of space between the table and the "instances list" link in the footer. Can we make that more compact?

Image

sophiecarroll:
> Code looks much better, though there is one [issue with responsiveness](https://github.com/canonical/lxd-ui/pull/2280#discussion_r4007273269) that seems not addressed yet.

Hey, yes looks much better, but the donut doesn't resize with the height of the table, the status column header isn't aligned when it's just the table, and the subheading is still to far from the heading. But definitely looks much better.

src/pages/overview/InstancesCard.tsx
Kxiru: Addressed!

kimanhou:
@edlerd @sophiecarroll 
Should the "Status" header be aligned with the text below and not the icon, similarly to the members table in the Clustering card ?

At 1090px (width of the window), the layout of the Overview page switches from 2 columns to a single column. Just before this switch, the donut is still here:

image

And right after the switch, the Instances card takes the whole width, so does it look weird without the donut ?

image

edlerd:
> @edlerd @sophiecarroll
> Should the "Status" header be aligned with the text below and not the icon, similarly to the members table in the Clustering card ?

Yes, we should align the status headers always with the text and not with the icon.

sophiecarroll:
> @edlerd @sophiecarroll Should the "Status" header be aligned with the text below and not the icon, similarly to the members table in the Clustering card ?
> 

I agree, status header should be aligned to the left like the other text


> At 1090px (width of the window), the layout of the Overview page switches from 2 columns to a single column. Just before this switch, the donut is still here:
> 
> image
> And right after the switch, the Instances card takes the whole width, so does it look weird without the donut ?
> 
> image

I agree it's a big jump. The status labels used to line break instead of truncate, and my original advice was the remove the donut once they start to line break, so in the current card what if we removed the donut once the labels start to truncate like the screenshot below? Would having just one change at a time (donut + chart half width--> chart halfwidth --> chart whole width) make the jump seem less dramatic?

Screenshot 2026-09-23 at 10 19 32 AM

sophiecarroll:
> > And right after the switch, the Instances card takes the whole width, so does it look weird without the donut ?
> > image
> 
> I agree it's a big jump. The status labels used to line break instead of truncate, and my original advice was the remove the donut once they start to line break, so in the current card what if we removed the donut once the labels start to truncate like the screenshot below? Would having just one change at a time (donut + chart half width--> chart halfwidth --> chart whole width) make the jump seem less dramatic?
> 
> Screenshot 2026-09-23 at 10 19 32 AM

Edit from standup: when the card jumps to full width, can we keep the same distance between columns as no donut, half width card?

Kxiru:
- [ ] Align status headers

> ...what if we removed the donut once the labels start to truncate like the screenshot below? Would having just one change at a time (donut + chart half width--> chart halfwidth --> chart whole width) make the jump seem less dramatic?

So you would like to remove the donut earlier?

> Edit from standup: when the card jumps to full width, can we keep the same distance between columns as no donut, half width card?

I do not understand.

@sophiecarroll

Issue #2244, opened 2026-08-28
View/Manage oidc sessions


vosdev:
Hey!

Through my journey of OIDC I was just fighting a broken oidc session.

I discovered that I can list/delete/show these through the CLI. 

```
lxc auth oidc-session list
lxc auth oidc-session delete 01a042e2-d0d6-76da-840e-13dcc2f1821f
Deleted OIDC session "01a042e2-d0d6-76da-840e-13dcc2f1821f"
```

I think this would be a nice addition to the UI. 

It can be an optional nav item under Permissions, only showing when oidc is configured on the server?

You can force a re-login of users that have obtained new permissions, are having issues, users that need their permissions revoked ASAP etc.

related: 
https://github.com/canonical/lxd/issues/18835
https://github.com/canonical/lxd/issues/18873

edlerd:
Thanks for raising the issue. We will consider it for the next cycle, now is a great time to propose as we just start the six month planning.

The API is present at `/1.0/auth/oidc-sessions?recursion=1` and also for single items, including basic delete. This might be an extension to the identity list as well, so we can see the sessions in relation to the users. Or we could have a "your profile" page that includes active sessions to be deleted. There are many open design questions and usage patterns to be figured out as a first step.

vosdev:
> Or we could have a "your profile" page that includes active sessions to be deleted

I've actually been working on a "your profile" page for some time, and I also thought about adding a "My sessions" area

<img width="1499" height="930" alt="Image" src="https://github.com/user-attachments/assets/28fd8cae-5349-4522-b522-0e3a46f33194" />

PR #2285, opened 2026-09-07
feat(Account): identity and preferences page

vosdev:
## Done

- Makes the `logged-in-identity` menu entry clickable
- Create dedicated identity information page
- Move localStorage user preferences from server settings to dedicated user preference page
- Display effective permissions for logged in identity
- Move oidc/bearer Log Out button to the new identity accordion menu
- Edit button next to `assigned` deeplinks to editting your own user on the existing identities page

Fixes something I've wanted for some time :) and @edlerd brought it up in https://github.com/canonical/lxd-ui/issues/2244

(Claude Code has helped me with this PR 🤖 )

Been working with OIDC/RBAC lately and seeing the effective permissions has helped. Perhaps displaying effective permissions can also be done on the identities page for other identities, as a new feature.

I've spoken to several (2) people who had no idea they could change the default login project, they did not expect it to be mixed in with server settings, so this preferences page is now in a logical place. It also leaves plenty of room for expansion of additional user preferences.

Design wise the identity page is identical to the instance overview and storage pool page. That does mean a lot of the CSS had to be duplicated.

This will probably be the last feature for now as the slow summer is coming to an end :(

Would love to hear your thoughts about this small feature! Hope you like it :)

## QA

1. Run the LXD-UI:
    - View the new Identity page
    - Attempt to modify your assigned groups via the new identity page
    - Modify your local browser preferences via the dedicated preferences page

## Screenshots

image

image

image

image

image

image


Kxiru:
Hiya Vosdev!
I personally love love love this idea! I'd like to tie in our designer @sophiecarroll who may be able to use this as a base plate/layout should we decide to take this further. I agree that features such as theme and default project can be surfaced in more accessible areas.

Some additional considerations can be made around aligning the design with OIDC vs TLS authentication, as well as other settings that could be highlighted here.

Ps. Love your use of the help links!

edlerd:
Thanks for raising the PR. The identity and preferences seem like very nice additions to the functionality.

A few thoughts on the proposed changes and on scope:

* The navigation should probably not change. I think the only nav change needed is to turn the username entry in the nav into a link to the page that lists user information and preferences. I also think we should merge identity/preferences into a single profile page. It could be with tabs, or maybe listing with consecutive headings is the simplest and best to use here. Though a design question, so as always, there is no right answer :)

* The theme selector might go into the header inline with the page title. Similar to the instance actions on the instance detail page.

* The login project should likely be merged with the idea of project favorites. So we can allow users to create a project priority. Favorites can have an order, and the first one in the list is the login project. I think there are too many open questions on this topic. We should best isolate this into a dedicated PR. Keep this PR focused on introducing the identity/preferences page. That will make it simpler to find consensus.

vosdev:
> The theme selector might go into the header inline with the page title. Similar to the instance actions on the instance detail page.

That is also a good solution. A simple moon icon aligned right of the LXD/MicroCloud logo, toggling between light dark and auto (system). Or maybe at the bottom underneath the Report a bug/Log out link.

> Favorites can have an order, and the first one in the list is the login project.

I thought about orders when I suggested project favorites, combining it with the login project is a nice touch. 

It does make for an interesting design problem. With "just" favorites, a simple star icon in the project list would do. If you add ordering, that becomes a bit more complicated.

> The navigation should probably not change. I think the only nav change needed is to turn the username entry in the nav into a link to the page that lists user information and preferences.

That is also a fine solution. Reason for making a separate preference page was to allow for possible future additions to preferences. 

A project favorite order preference probably takes up a bit more space than just a dropdown menu. But with the theme toggle disappearing we don't need a separate page for just one preference.

I dumbed down the identity page suggestion for now. I will update the screenshots in a bit.

Unrelated sidenote:

It would mean a lot to me if you could leave a small testimony on my Ubuntu Community Member application:
https://discourse.ubuntu.com/t/daniel-vos-membership-application/87497

sophiecarroll:
Hey @vosdev ! I am the UX designer for LXD, and I was taking a look at your proposed solution and it looks like a great addition and very useful. I had a few quick suggestions from the design side that with some of the factors we're thinking about–let me know if these changes are feasible for you to make.

26 10 LXD Profile page merge request design feedback
(Here is a brief mockup of what I'm describing, sorry for the bluriness! I've pasted it a few times but it keeps being blurry. Let me know if you can't see something)

1. Thanks for changing it to be one page with a non-expanded nav item. Because Identity and preferences are two sections, can you change the heading of the page to match the name of the logged in user from the nav item, that way we maintain the hierarchy of having two subsections, instead of preferring one? Eg. in this screenshot the heading of the page would say “workshop-cert” instead of Identity, or in your screenshots it would say Daniel Vos. 

Screenshot From 2026-09-15 09-58-18

2. ^ Can you change Groups to say Auth groups? We recently changed the label of the side nav item to clarify which type of group. We want the option to edit authentication settings like groups/permissions to stay only in the Permissions nav items, in order to not have people split between two places where they can edit, so they automatically always know where to go. Can you remove the option to edit the Groups? I like your thinking that people might want to edit once they see their Groups, so can you put an explanation tooltip next to the Auth groups label? It can match our current label which says “Auth groups organize identities for permission management. [this is an addition] Edit them in the Permissions → Auth groups page. [line break] Learn more about authorization (link=https://lxd-ui-2280.demos.haus/documentation/explanation/authorization/)” 

Screenshot 2026-09-15 at 10 05 23 AM

3. Can you move Preferences to the top, to avoid it being hidden in case the user has a lot of effective permissions?

vosdev:
Hi @sophiecarroll !

I'm a user of the product 😄 

> can you change the heading of the page to match the name of the logged in user from the nav item

Sure!

> Can you change Groups to say Auth groups?

Will do!

> can you put an explanation tooltip next to the Auth groups label? It can match our current label which says “Auth groups organize identities for permission management. 

Will do!

> in order to not have people split between two places where they can edit, so they automatically always know where to go. Can you remove the option to edit the Groups

There is still only one place to edit groups. This edit button simply brings you to the edit screen on the `/permissions/identities` page with the edit panel open, like so:

https://localhost:8407/ui/permissions/identities?panel=edit-identity&identity=daniel%40vos.dev

image

Do you still want it removed? It is only to help you point to the right place. I am not adding a second place to be able to edit your permissions.

> Can you move Preferences to the top, to avoid it being hidden in case the user has a lot of effective permissions?

Yes, I think having it below General would be the best place then? So:
- General
- Preferences
- Auth Groups
- Effective Permissions

sophiecarroll:
> > in order to not have people split between two places where they can edit, so they automatically always know where to go. Can you remove the option to edit the Groups
> 
> There is still only one place to edit groups. This edit button simply brings you to the edit screen on the `/permissions/identities` page with the edit panel open, like so:
> 
> https://localhost:8407/ui/permissions/identities?panel=edit-identity&identity=daniel%40vos.dev
> 
> image
> Do you still want it removed? It is only to help you point to the right place. I am not adding a second place to be able to edit your permissions.

Ok thanks I didn't catch that it took you to the other page! It's true that people might want to edit their groups once the groups are in front of them, and it would be nice to surface where they should go. We're trying to not use actions that usually keep you in the same page (eg. pencil icon) to jump you to another page without specifying that it will be a different page, so it would still be best to remove the pencil icon. To keep the same functionality idea though, we could add a muted line of text under the group that explains where you can edit (this is a pattern we use to surface help text in other places like the examples below), with the "Identities" link being your current link https://localhost:8407/ui/permissions/identities?panel=edit-identity&identity=daniel%40vos.dev. I changed the text slightly to say "edit the groups assigned to your identity" because you're not actually editing the group itself, just which ones apply to you. 

Examples:
Screenshot 2026-09-15 at 5 26 02 PM
Screenshot 2026-09-15 at 5 15 09 PM

My quick mockup:
Screenshot 2026-09-15 at 5 39 36 PM


That means that you could also take that last sentence out of the tooltip and just use the Auth groups tooltip as is in the screenshot.

> > Can you move Preferences to the top, to avoid it being hidden in case the user has a lot of effective permissions?
> 
> Yes, I think having it below General would be the best place then? So:
> 
> * General
> * Preferences
> * Auth Groups
> * Effective Permissions

Sounds good to me.

Thanks for the back and forth!

vosdev:
image

I think this looks very nice, what do you think @sophiecarroll ?

image

sophiecarroll:
> image
> 
> I think this looks very nice, what do you think @sophiecarroll ?
> image

This looks amazing! Only one small things from my side--changing "Auth Groups" section heading to "Auth groups" because we use sentence case for multi word headings. Other than that, it looks great from my design perspective.

src/util/settings.tsx
```diff
@@ -76,23 +70,6 @@ export const getUserSettings = (
...
-    },
-    {
-      key: "user.ui_theme",
-      category: "user",
-      default: "",
-      shortdesc:
-        "Set UI to dark theme, light theme, or to match the system theme.",
-      type: "string",
-      isSaved: true,
-    },
```
edlerd, line 95: Removing those here can confuse users. I would suggest keeping the entries, but turn them into muted text with a quick note that those can now be found on the identities page. Similarly to what we did with the instance `secureboot` and `CSM` settings when they were moved:

Image

tests/helpers/identity.ts
```diff
@@ -0,0 +1,8 @@
+import { expect, type Page } from "@playwright/test";
+import { gotoURL } from "./navigate";
+
+export const visitIdentityPage = async (page: Page) => {
+  await gotoURL(page, "/ui/");
+  await page.locator('.sidenav-bottom-ul a[href$="/ui/identity"]').click();
```
edlerd, line 6:
```suggestion
  await expect(page.getByText("Preferences")).toBeVisible();
```
Better tie the e2e tests to visible aspects of the page than to implementation details such as css classes.

tests/identity.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+test("identity nav entry links to the identity page", async ({ page }) => {
+  await visitIdentityPage(page);
+
+  await expect(page.getByRole("heading", { name: "Identity" })).toBeVisible();
+  await expect(page.locator("#identity-name")).toBeVisible();
+  await expect(page.locator("#identity-id")).toBeVisible();
+  await expect(
+    page.getByRole("heading", { name: "Preferences" }),
+  ).toBeVisible();
```
edlerd, line 13: Better find other more visible aspects to assert here. Maybe sub-headings?

tests/login-project.spec.ts
```diff
@@ -5,38 +5,33 @@ import {
...
 
-  await visitServerSettings(page);
-  const settingRow = page.getByRole("row").filter({
-    has: page.getByText("user.ui_login_project"),
-  });
-  await settingRow.getByRole("button").click();
-  await settingRow
-    .getByRole("combobox", { name: SETTING_NAME })
+  await visitIdentityPage(page);
+  await page.locator(".readmode-button").click();
```
edlerd, line 18: There should be a title or aria-label on the button that we use to locate it here instead of a CSS class. That needs a small fix to the implementation to also apply the title on the button for accessibility.

tests/login-project.spec.ts
```diff
@@ -5,38 +5,33 @@ import {
...
+  await page
+    .getByRole("combobox", { name: "Login project" })
     .selectOption({ label: project });
-  await settingRow.getByRole("button", { name: "Save", exact: true }).click();
-  await dismissNotification(page, `Setting ${SETTING_NAME} updated.`);
+  await page.getByRole("button", { name: "Save", exact: true }).click();
+  await dismissNotification(page, "Login project updated.");
 
   await gotoURL(page, "/ui/");
   await expect(page).toHaveURL(new RegExp(`/ui/project/${project}/instances`));
```
edlerd, line 26: Better expect the project to be in the URL. The start page might shortly change to the dashboard, which would otherwise break this test.

tests/login-project.spec.ts
```diff
@@ -5,38 +5,33 @@ import {
...
 
-  await visitServerSettings(page);
-  await settingRow.getByRole("button").click();
-  await settingRow
+  await visitIdentityPage(page);
+  await page.locator(".readmode-button").click();
```
edlerd, line 29: Better use accessibility safe locators (same as other comment above).

src/sass/_identity.scss
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    .p-segmented-control__button {
+      margin-bottom: 0;
+    }
+  }
+
+  .identity-preference-value {
+    flex-grow: 1;
+    max-width: 20rem;
```
edlerd, line 111:
```suggestion
```
Let's rely on the available space so the reset button moves in line with the other actions.

Image

Image

src/sass/_identity.scss
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    .edit-icon {
+      min-width: 1rem;
+    }
+  }
+
+  .reset-button {
+    padding: 0 $sph--small;
+  }
```
edlerd, line 133:
```suggestion
```
It seems slightly more symmetric without this.

src/sass/_identity.scss
```diff
@@ -0,0 +1,134 @@
...
+    flex-grow: 1;
+    max-width: 20rem;
+  }
+
+  .readmode-button {
+    align-items: center;
+    display: flex;
+    padding: 0 1rem 0 0;
+    text-align: start;
+    width: 100%;
```
edlerd, line 119:
```suggestion
    max-width: 100%
```
The size can be dynamic depending on the project name. We only need to limit max in case of truncation.

src/sass/_identity.scss
```diff
@@ -0,0 +1,134 @@
...
+  .identity-preference-value {
+    flex-grow: 1;
+    max-width: 20rem;
+  }
+
+  .readmode-button {
+    align-items: center;
+    display: flex;
+    padding: 0 1rem 0 0;
+    text-align: start;
```
edlerd, line 118:
```suggestion
```
With dynamic sizing we probably can drop this.

src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,318 @@
...
+    if (!groups.length) {
+      return <span className="u-text--muted">-</span>;
+    }
+
+    return (
+      <List
+        inline
+        middot
+        items={groups}
+        className="effective-groups-list u-no-margin--bottom"
```
edlerd, line 49:
```suggestion
        className="u-no-margin--bottom"
```
This class has no effect here.

src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,318 @@
...
+        resource: permission.url,
+        entitlement: permission.entitlement,
+        grantedBy: grantingGroups.join(", "),
+      },
+    };
+  });
+
+  const editGroupsHint = currentIdentity ? (
+    <p className="identity-groups-hint u-text--muted u-no-margin--bottom">
+      Edit the groups assigned to your identity on the{" "}
```
edlerd, line 124:
```suggestion
      Edit the group assignments on the{" "}
```
Can we remove this bit to avoid saying "identity" twice in the sentence?

src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,318 @@
...
+              Identity details are not available for the current authentication
+              method or LXD version.
+            </Notification>
+          )}
+        </Col>
+      </Row>
+      <Row className="section">
+        <Col size={3}>
+          <h2 className="p-heading--5">
+            <ExplanationTooltip explanation="Preferences are stored in this browser only. They are not bound to your identity and do not apply to other browsers or devices.">
```
edlerd, line 206:
```suggestion
            <ExplanationTooltip explanation="Preferences are stored in this browser. They are not bound to your identity and do not apply to other browsers or devices.">
```
Tiny nitpick.

src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,318 @@
...
+                )}
+                {editGroupsHint}
+              </div>
+              {inheritedGroups.length > 0 && showIdpGroupsNotification && (
+                <Notification
+                  severity="information"
+                  className="idp-groups-notification"
+                  onDismiss={() => {
+                    setShowIdpGroupsNotification(false);
+                  }}
```
edlerd, line 276:
```suggestion
```
I think this can be non-closeable.

src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,318 @@
...
+                >
+                  Inherited groups are granted through identity provider group
+                  mappings, not on the identity itself. To change them, join or
+                  leave the matching group in your identity provider, or edit
+                  which LXD groups the mapping grants under{" "}
+                  <Link to={`${ROOT_PATH}/ui/permissions/idp-groups`}>
+                    IDP groups
+                  </Link>
+                  .
```
edlerd, line 285: I am not super happy with the copy here. It reads fairly complicated. Though I have no direct suggestion on how to clarify/simplify. Maybe @sophiecarroll has a good idea here?

src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,318 @@
...
+        },
+      ],
+      sortData: {
+        entityType: permission.entity_type,
+        resource: permission.url,
+        entitlement: permission.entitlement,
+        grantedBy: grantingGroups.join(", "),
+      },
+    };
+  });
```
edlerd, line 120: This seems like a suitable candidate for an `EffectivePermissionsTable` component to extract the logic in a reusable component. What do you think?

edlerd (commented):
The identity page is very nice. I really like it!

Did a detailed QA, layout, and code review. A few ideas and suggestions below to make it perfect.

src/pages/identity/Identity.tsx
edlerd: I mean those helpers together with the use of the main table below.

src/util/settings.tsx
vosdev: image

src/pages/identity/Identity.tsx
vosdev: Yes! Additionally, the UI could re-use this component on the Permissions page when viewing or editting an identity there.

src/util/settings.tsx
edlerd: Nice, looks great!

src/pages/identity/Identity.tsx
edlerd: The section with this notification looks a bit busy. There is the group info, then the help text and then a notification. Maybe there is a cleaner way?

image

src/pages/identity/Identity.tsx
```diff
@@ -0,0 +1,220 @@
...
+    <p className="identity-groups-hint u-text--muted u-no-margin--bottom">
+      Edit the group assignments on the{" "}
+      <Link
+        to={`${ROOT_PATH}/ui/permissions/identities?panel=edit-identity&identity=${encodeURIComponent(currentIdentity.id)}`}
+      >
+        Identities
+      </Link>{" "}
+      page.
+    </p>
+  ) : null;
```
edlerd: We should hide this hint if using has an unrestricted cert: 

Image

src/pages/permissions/AuthGroupList.tsx
```diff
@@ -0,0 +1,16 @@
...
+interface Props {
+  groups: string[];
+}
+
+const AuthGroupList: FC<Props> = ({ groups }) => {
+  if (!groups.length) {
+    return <span className="u-text--muted">-</span>;
+  }
+
+  return <List inline middot items={groups} className="u-no-margin--bottom" />;
```
edlerd: I am not 100% sure the inline list is best to display auth groups in the scale case:

Image

Maybe it's better to use a standard list:

```suggestion
  return <List items={groups} className="u-no-margin--bottom" />;
```

Image

edlerd (commented):
Thanks for the iteration. The code LGTM as is. Two edge cases on the scale case and combinations such as unrestricted certs can become better with small adjustments. Otherwise QA also seems all good to go!

src/pages/identity/Identity.tsx
edlerd: I think the "Edit the group assignments on the Identities page." help text is confusing when both assigned and inherited groups are visible. It works well only if no inherited groups exist.

src/pages/permissions/AuthGroupList.tsx
edlerd: Can we use the resource label for all auth groups?

```<ResourceLink
              type="auth-group"
              value={values.name}
              to={`${ROOT_PATH}/ui/permissions/groups`}
            />
```

Like in the notification:

image

src/pages/identity/Identity.tsx
vosdev: UI design can be a pain. Glad we have @sophiecarroll to help us out!

I agree that this is a bit much. There is currently more helper text than there is data.

We could move the contents inherited groups infobox to the ? next to Auth groups. That will clean up a lot.

The "Edit the group assignments on the Identities page." text was because initially I had a pencil edit button next to the Assigned groups. We could just get rid of that line and expect users to know they can modify it on the permissions/identities page.

edlerd: I think the "Edit the group assignments on the Identities page." works nicely if the group membership is only explicit without inheritance. In this case I would definitely keep it:

image

Just the case above with both assigned and inherited plus the notification we should probably hide it.

edlerd: > Just the case above with both assigned and inherited plus the notification we should probably hide it.

Also considering the notification already links to the IDP groups in the auth section.

src/pages/permissions/AuthGroupList.tsx
vosdev: If i knew we had one, it would have been there already! I totally missed it. Will look at it tomorrow.

kimanhou:
Hi @vosdev, could you please squash all your commits and sign off ? 🙏

src/pages/permissions/AuthGroupList.tsx
vosdev: image

vosdev: image

edlerd: Yeah, this looks much cleaner now with the chips and rules around the help text/notifications.

vosdev:
> Hi @vosdev, could you please squash all your commits and sign off ? 🙏

Yes, will do after everything is approved :)

src/pages/permissions/EffectivePermissionsTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+      },
+    };
+  });
+
+  return (
+    <MainTable
+      className="effective-permissions-table"
+      headers={headers}
+      rows={rows}
+      sortable
```
edlerd, line 90:
```suggestion
      sortable
      defaultSort="resource"
      defaultSortDirection="ascending"
```
Let's apply a default sort column and direction. So on initial load, the admin permission is always on top, and individual entities are grouped by type:

Before:

Image

After:

Image

edlerd (dismissed):
QA and Code LGTM

Tiny idea to improve the effective permission list.

As mentioned by @kimanhou please squash to a single signed commit so we can merge it.

edlerd (dismissed):
LGTM, thanks so much for adding the profile page and making it shine!

kimanhou:
Hi @vosdev 
The initial access test consistently fails on your PR because, after deleting the identity, we expect to be on the login page. Instead it shows an error.
@edlerd Do you have an idea why ?

image

I tested locally and got the same behavior as in the CI.


https://github.com/user-attachments/assets/17f65b3d-b08b-4dc1-b124-eb74a56fd99f


I also found an issue with navigation. When I click on Permissions, the whole section of the navigation disappears.


https://github.com/user-attachments/assets/8bcfbe70-5ad7-4762-90dc-83cd5541f00c

src/pages/permissions/actions/DeleteIdentityBtn.tsx
```diff
@@ -48,6 +48,7 @@ const DeleteIdentityBtn: FC<Props> = ({ identity }) => {
               queryKeys.identities,
               queryKeys.authGroups,
               queryKeys.settings,
+              queryKeys.currentIdentity,
```
kimanhou, line 51: I think this line introduces a bug. Deleting your own TLS identity now does not log you out. It only asks React Query to refetch /1.0/auth/identities/current. Maybe we should the same as for OIDC and force log out ?
@edlerd

kimanhou (commented):
I think I found a solution for the disappearing of section when clicking on Permissions or Clustering in the Navigation. In _pattern_navigation.scss, moving lines 119-120 to 124-125 (inside .sidenav-top-container) moves scroll ownership from each `.sidenav-top-ul` to the shared `.sidenav-top-container` which makes both top lists scroll as one area, and fixes the clipping.

```
.sidenav-top-ul {
  margin-bottom: 0;
}

.sidenav-top-container {
  height: 100%;
  overflow: auto;
  overflow-x: hidden;
}
```

edlerd:
>> I also found an issue with navigation. When I click on Permissions, the whole section of the navigation disappears.
> I think I found a solution for the disappearing of section when clicking on Permissions or Clustering in the Navigation. In _pattern_navigation.scss, moving lines 119-120 to 124-125 (inside .sidenav-top-container) moves scroll ownership from each `.sidenav-top-ul` to the shared `.sidenav-top-container` which makes both top lists scroll as one area, and fixes the clipping.

@kimanhou This was a bug introduced and fixed last week.
@vosdev can you rebase on main? That should resolve the problem with navigation described above.

src/pages/permissions/actions/DeleteIdentityBtn.tsx
edlerd: Good catch @kimanhou !

I wonder if we should just remove the invalidation of the current identity here. Restoring what was here before.  Then we can rely on the existing logic to log the person out, and it won't break the existing flow. In my local tests, with removing this change it resolved the issue.

kimanhou: @vosdev Why did you add this in the first place ?

edlerd: > @vosdev Why did you add this in the first place ?

I think there are several other cases where we failed to invalidate the current identity cache. Previously that would not surface, but with the profile page relying on it, this becomes relevant. Various current identity query invalidations were fixed in this PR. This occasion of the delete button is one other similar case. But actually now we see it is a bit different, and we should rely on existing logic instead of invalidating this one.

vosdev: > @vosdev Why did you add this in the first place ?

When modifying my identity group membership, i noticed that my groups did not update on the identity page. I had to F5.

By invalidating the query that worked immediately. 

I did not expect to surface a different bug by doing so. Will look at it later today :)

Will also rebase on Main. I noticed the nav bug also, but when i was typing the issue i saw it was solved a few hours earlier.

vosdev: I was a bit eager and added the invalidate query to too many refresh lists. I have removed them from the (bulk) delete identities files.

src/sass/_identity.scss
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  .identity-group-value {
+    align-items: center;
+    display: flex;
+    gap: $sph--small;
+  }
+
+  .identity-groups-hint {
+    font-size: 0.875rem;
```
kimanhou: Do you want to use the classname `p-text--small` instead of this CSS line ?

src/sass/_identity.scss
```diff
@@ -0,0 +1,129 @@
...
+  .identity-preferences {
+    display: flex;
+    flex-direction: column;
+    gap: $spv--large;
+    margin-bottom: $spv--large;
+    padding-top: $spv--small - 0.375rem;
+
+    .identity-group-row {
+      align-items: flex-start;
+      padding: 0;
```
kimanhou: `u-no-padding` ?

src/sass/_identity.scss
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    .identity-group-label {
+      padding-top: 0.375rem;
+    }
+
+    .p-segmented-control__button {
+      margin-bottom: 0;
```
kimanhou: `u-no-margin--bottom` ?

kimanhou (dismissed):
Minor comments on CSS

edlerd (dismissed):
LGTM, thanks.

vosdev:
Updated for the 3 css nitpicks from @kimanhou. Squashed and rebased. Now should be all good :)

src/sass/_identity.scss
vosdev: The theme buttons are inside ThemeSwitcher, which takes no class name, so I can't place it anywhere unless we affect all the other rows as well.

PR #2298, opened 2026-09-14
feat: [WD-38687] Maintain required form formatting

Kxiru:
## Done

- Create instance form (Base image)
- Create netowrk forward (Listen Address)
- Enable Clustering Form validation.

## QA

1. Run the LXD-UI:
    - Review the 3 areas above and ensure they are according to the "[spec](https://warthogs.atlassian.net/browse/WD-37731)".

## Screenshots

N/A


sophiecarroll:
Looks good to me

src/pages/instances/forms/InstanceCreateDetailsForm.tsx
```diff
@@ -77,7 +78,7 @@ const InstanceCreateDetailsForm: FC<Props> = ({
       </Row>
       <Row>
         <Col size={12}>
-          <p className="p-form__label">Base Image*</p>
+          <Label required>Base Image</Label>
```
edlerd:
```suggestion
          <Label required forId="select-image">Base Image</Label>
```
We should not create an orphan label. Let's pin this to the first option below.

src/pages/cluster/EnableClusteringModal.tsx
```diff
@@ -117,14 +115,16 @@ const EnableClusteringModal: FC<Props> = ({ onClose }) => {
       <Input
         label="Server name"
         type="text"
-        required
+        error={formik.touched.serverName ? formik.errors.serverName : null}
         {...formik.getFieldProps("serverName")}
```
edlerd, line 119: This and the field below are both required. Can we restore it?

Kxiru: @sophiecarroll redesigned the flow of this form. See the last item of the [spec](https://warthogs.atlassian.net/browse/WD-37731). She required that as both items are required, it is implicit and therefore validation should occur on button press rather than inline. At least, that's how I understood it.

edlerd: I agree with the change to show error on submit and enable submit right away. I do think we should show the asterisk on both inputs though -- they are both required to have a value for a successful submission.

Kxiru: I see. @sophiecarroll , you explicitly asked these be removed in the spec. Is it an issue to add them back in?

sophiecarroll: @edlerd This is the guidance in the design system that says do not indicate required fields for small forms with 1-3 fields. So I would say from that perspective we shouldn't have them, but I'm open to reasons why we could make an exception here or why it's particularly important in this case

Screenshot 2026-09-15 at 12 17 29 PM

edlerd: > This is the guidance in the design system that says do not indicate required fields for small forms with 1-3 fields

I wonder about the reasons for this choice. To me it seems better to always indicate required fields. Without it, the user runs into the error and gets frustrated. I see no advantage of not showing the field as required, only disadvantages.

edlerd: I don't see why a form being big or small makes a difference here.

Kxiru: @sophiecarroll

sophiecarroll: This is the reasoning in the spec:
Screenshot 2026-09-15 at 2 14 47 PM

To me, it makes sense because server name and cluster address are the only two fields, and I would automatically assume they're required, and if I didn't (unusual), I would be alerted to that if I clicked submit and couldn't because of errors. However, if we feel strongly that people might not know that information is necessary, we can leave the asterisks.

Screenshot 2026-09-15 at 2 17 01 PM

edlerd: Thanks, I see for login or sign-up how the requirement of filling the fields is obvious and doesn't need to be communicated. For this one I was not so sure. Personally I assumed people might not consider it mandatory. But then again I am heavily biased and would trust your decision here. So let's remove them as suggested; it is cleaner after all.

edlerd (dismissed):
LGTM, thanks!

edlerd (approved):
LGTM, thanks

PR #2299, opened 2026-09-15
chore(test) fix operation tests to allow for operations such as delete instance to stick around for longer

edlerd:
## Done

- chore(test) fix operation tests to allow for operations such as delete instance to stick around for longer

Making tests more reliable. Previous failures:
- https://github.com/canonical/lxd/actions/runs/34871510395/job/104071869550
- https://github.com/canonical/lxd/actions/runs/34597907785/job/103260143031
- https://github.com/canonical/lxd/actions/runs/34393342137/job/102609818672
- https://github.com/canonical/lxd/actions/runs/34494508045/job/103029967732


Kxiru (approved):
Code review looks good. Though how does this address the highlighted issue of operations taking longer when all it is doing is moving forward when the (first) title selector is visible.

edlerd:
> Though how does this address the highlighted issue of operations taking longer when all it is doing is moving forward when the (first) title selector is visible.

The tests linked in the description failed because previous tests left behind duplicate matching operations. The change makes the test resilient to such leftover operations.

Kxiru:
> The change makes the test resilient to such leftover operations.

Okay, so the test would make multiple operation instances, and in trying to get the operation in question, it would error as the selector matched multiple rows? So now we only target the first row? Is this correct?

PR #2304, opened 2026-09-16
feat: [WD-38704] Overview page empty states

Kxiru:
## Done

- [x] Instance empty state
- [x] Warnings
- [ ] Storage pools
- [ ] Permissions (?)

## QA

1. Run the LXD-UI:
    - Review the Overview page with a user that only has view-only-project permissions.

## Screenshots
image

image


sophiecarroll:
Everything looks good except can you make sure the padding on the left and right sides of the empty state message are equal?

Kxiru:
> Everything looks good except can you make sure the padding on the left and right sides of the empty state message are equal?

The padding **is** equal, it's just there is a new line in the link text. For some reason the box will not wrap between "list" and "page" on my screen at least. Did you test using the demo on your screen?.

sophiecarroll:
> > Everything looks good except can you make sure the padding on the left and right sides of the empty state message are equal?
> 
> The padding **is** equal, it's just there is a new line in the link text. For some reason the box will not wrap between "list" and "page" on my screen at least. Did you test using the demo on your screen?.

Both warnings and Instances look good now, I see what you mean. I had to go to a project with 0 instances to test.

src/pages/overview/InstancesCard.tsx
```diff
@@ -41,10 +42,16 @@ const InstancesCard: FC = () => {
...
-      {!isLoading && !error && instances.length > 0 && ` (${instances.length})`}
-    </span>
+    <>
+      <span className="overview-card-title">
+        <Icon name="pods" /> Instances
+        {!isLoading &&
+          !error &&
+          instances.length > 0 &&
+          ` (${instances.length})`}
+      </span>{" "}
```
edlerd, line 52:
```suggestion
      <span className="overview-card-title">
        <Icon name="pods" /> Instances
        {!isLoading &&
          !error &&
          instances.length > 0 &&
          ` (${instances.length})`}
      </span>
```
Can we remove the space here?

src/pages/overview/InstancesCardEmptyState.tsx
```diff
@@ -0,0 +1,34 @@
+import { useCurrentProject } from "context/useCurrentProject";
+import { useProject } from "context/useProjects";
+import { Link } from "react-router-dom";
+import { useProjectEntitlements } from "util/entitlements/projects";
+import { getInstancesUrl } from "util/projects";
+
+const InstancesCardEmptyState = () => {
+  const { canCreateInstances } = useProjectEntitlements();
+  const { project, isAllProjects } = useCurrentProject();
+  const { data: defaultProject } = useProject("default", isAllProjects);
```
edlerd, line 10: Better use `getDefaultProject` helper here, as the user might not have access to the `default` project.

src/pages/overview/WarningsCardEmptyState.tsx
```diff
@@ -0,0 +1,17 @@
...
+const WarningsCardEmptyState = () => {
+  return (
+    <>
+      <div className="overview-empty-state u-no-margin--bottom">
+        <div className="overview-card-subtitle">No warnings found</div>
+      </div>
+      <div className="card-footer">
+        <Link to={`${ROOT_PATH}/ui/warnings?status=new`}>Warnings list</Link>
+      </div>
+    </>
```
edlerd, line 13:
```suggestion
    <OverviewCardEmptyState title="No warnings found" footer={(<Link to={`${ROOT_PATH}/ui/warnings?status=new`}>Warnings list</Link>
```
Can we introduce a `OverviewCardEmptyState` component to reuse?

edlerd (commented):
Code mostly LGTM, few comments on the draft.

sophiecarroll:
Looks good to me, just wanted to check that the warnings card in the second screenshot will use the same empty state spacing as instances and storage

Kxiru:
> Looks good to me, just wanted to check that the warnings card in the second screenshot will use the same empty state spacing as instances and storage

Hmm, I'm not sure what you mean. The warnings card appears as it does in the [spec](https://docs.google.com/document/d/1Myu3NOPXoLd_80-X_6tZnNOPNMS9gwX_ljMm71aCgAs/edit?tab=t.prg1thxmcrex#heading=h.6nxmimybncel) and in the first screenshot. There is no additional information to display as a subtitle for the warnings card so it maintains the single, title emboldened and centered styling.

src/pages/overview/CardEmptyState.tsx
```diff
@@ -0,0 +1,31 @@
+import type { FC, ReactNode } from "react";
+
+interface Props {
+  subtitle: string;
+  message?: ReactNode;
+  footerLink?: ReactNode;
```
kimanhou: Isn't the `footerLink` always here ?

```suggestion
  footerLink: ReactNode;
```

src/pages/overview/CardEmptyState.tsx
```diff
@@ -0,0 +1,31 @@
...
+  centered?: boolean;
+}
+
+const CardEmptyState: FC<Props> = ({
+  subtitle,
+  message,
+  footerLink,
+  centered,
```
kimanhou:
```suggestion
  centered = !message
```

src/pages/overview/CardEmptyState.tsx
```diff
@@ -0,0 +1,31 @@
...
+}) => {
+  const isCentered = centered ?? !message;
+
+  return (
+    <>
+      <div
+        className={`overview-empty-state u-no-margin--bottom${isCentered ? " is-centered" : ""}`}
```
kimanhou:
```suggestion
        className={classnames("overview-empty-state u-no-margin--bottom", {
          "is-centered": isCentered,
        })}
```

It's easier to read with `classnames`

src/pages/overview/CardEmptyState.tsx
```diff
@@ -0,0 +1,31 @@
...
+      >
+        <div className="overview-card-subtitle">{subtitle}</div>
```
kimanhou:
```suggestion
        <p className="overview-card-subtitle">{subtitle}</p>
```

This should be a `p`, not a `div`. You can then adjust margin if needed.

src/pages/overview/CardEmptyState.tsx
kimanhou: You can even remove the "centered" props and handle the exception of PermissionsCard in CSS.

kimanhou: Or not use the CardEmptyState component for PermissionsCard. That would greatly simplify the code for PermissionsCard, as you just need to hide "Your groups:".

src/pages/overview/ProjectsCard.tsx
```diff
@@ -55,6 +58,24 @@ const ProjectsCard: FC = () => {
     );
   }
 
+  if (projects.length === 0) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
```
kimanhou, line 63: According to the spec, the title should be "Projects (0)".

Spec:

Image

Current branch:

Image

src/pages/overview/InstancesCard.tsx
```diff
@@ -41,10 +42,16 @@ const InstancesCard: FC = () => {
...
-    <span className="overview-card-title">
-      <Icon name="pods" /> Instances
-      {!isLoading && !error && instances.length > 0 && ` (${instances.length})`}
-    </span>
+    <>
+      <span className="overview-card-title">
+        <Icon name="pods" /> Instances
+        {!isLoading &&
+          !error &&
+          instances.length > 0 &&
```
kimanhou, line 50: According to the spec, the title should be "Instances (0)" when no instances is found.

Spec:

Image

Current branch:

Image

src/pages/overview/WarningsCardEmptyState.tsx
```diff
@@ -0,0 +1,16 @@
+import { Link } from "react-router-dom";
```
kimanhou: I don't think this component is necessary, this can be inline in WarningsCard directly.

src/sass/_overview.scss
```diff
@@ -26,6 +26,10 @@
     margin-bottom: 0;
     min-width: 0;
 
+    .overview-card-subtitle {
+      font-weight: 500;
```
kimanhou, line 30: Is there really a visible different between font-weight 400 and 500 ? 
@sophiecarroll 

Image

sophiecarroll:
> > Looks good to me, just wanted to check that the warnings card in the second screenshot will use the same empty state spacing as instances and storage
> 
> Hmm, I'm not sure what you mean. The warnings card appears as it does in the [spec](https://docs.google.com/document/d/1Myu3NOPXoLd_80-X_6tZnNOPNMS9gwX_ljMm71aCgAs/edit?tab=t.prg1thxmcrex#heading=h.6nxmimybncel) and in the first screenshot. There is no additional information to display as a subtitle for the warnings card so it maintains the single, title emboldened and centered styling.

The first screenshot shows the message with only the first line of the text centered, and the second one shows the message centered but based on a wider text box including the second line of text. I just want to make sure the first line of text on warnings would have the same padding on the left as instances/storage, even though it continues to have one line in an empty state.

src/pages/overview/InstancesCard.tsx
sophiecarroll: @Kxiru @kimanhou Good catch on the instances and projects empty state headings Kim Anh–actually, after we've fleshed out the empty state messages, I think we can leaving the headings without a (0) for consistency, now that I'm seeing them all the empty states come together. I will update in the spec.

src/sass/_overview.scss
sophiecarroll: I think font usage looks good as is–there is a small visible difference, and it enforces the hierarchy

src/pages/overview/InstancesCard.tsx
Kxiru: So no entity counts when no entities.

Kxiru:
There are only 2 warnings screenshots in the spec as of today, and only two warnings screenshots in this PR. Neither of which match this description. Can you clarify?

src/pages/overview/ProjectsCard.tsx
Kxiru: @kimanhou , what kind of environment / user permissions gave you this view? Even with a user with only can create project permissions, I cannot view No projects AND the subtitle listed here.

@sophiecarroll , some cards use "All instances list" or "Instances list" as the footer link. Is it worth aligning this messaging?

kimanhou: As mentioned in the spec, this view should never happen but we handle it in case someone manages to break LXD. To get this view, you need to tweak the code locally.

kimanhou:
@sophiecarroll 

>  the second one shows the message centered but based on a wider text box including the second line of text. 

Which second line of text are you talking about ? To be sure, are we talking about the Warnings card ? Because the Warnings card never has a second line of text, as you can't create a warning (this is mentioned in your comment in the spec).

src/pages/overview/ProjectsCard.tsx
```diff
@@ -69,7 +90,7 @@ const ProjectsCard: FC = () => {
             )}
             <Link to={getInstancesUrl(projectName)}>
               Project instances list
-            </Link>
+            </Link>{" "}
```
kimanhou, line 93: Do we need this extra space ?

src/pages/overview/PermissionsCard.tsx
```diff
@@ -2,43 +2,60 @@ import type { FC } from "react";
...
+        <Icon name="user" /> Permissions
+      </span>
+      <PermissionGroupExplanationTooltip />
+    </>
   );
 
+  if (!effectiveGroups || effectiveGroups.length === 0) {
+    return (
+      <Card className={cardClassName} title={cardTitle}>
+        <CardEmptyState
```
kimanhou, line 33: I wouldn't use the CardEmptyState component here. But if you absolutely want to, put title and footerLink in variables and re-use them here and one line 49 and lines 61-63.

Also I still think that there should not be a "centered" props in CardEmptyState. It can be handled by CSS.

sophiecarroll:
> There are only 2 warnings screenshots in the spec as of today, and only two warnings screenshots in this PR. Neither of which match this description. Can you clarify?

Hey Nkeiru, this is something I only noticed once I saw the empty states with the subheadings combined with the ones without, which results in different padding because of different widths being centered, but after playing around with the warnings card, I realized it would make it more complicated than it needs to be to align this. Let's just keep warnings empty state as is, sorry for the false alarm

src/pages/overview/ProjectsCard.tsx
sophiecarroll: Seconding Kim Anh that this is a super edge case but we wanted to account for it just incase. 

> some cards use "All instances list" or "Instances list" as the footer link. Is it worth aligning this messaging?

I'll paste my response from Mattermost for ease of finding it:

The "Instances list" link stays the same always on the instances card, and leads always to level of the project (eg. all instances when all projects is selected). The projects card link changes based on the level: if you are in all projects on the sidenav dropdown, it should say "All instances list" to reinforce that it is across all projects, and if you are inside of a project in the nav dropdown it should say "Project instances list" to reinforce that it is only the list inside one project. The reason I include All/Project in the projects card is because it can change so I want to be extra clear. Does that make sense or do you think we should reevaluate?

Kxiru: Alrighty, @sophiecarroll

src/pages/overview/PermissionsCard.tsx
Kxiru: I think there are enough examples of both centered and uncentered Empty state structures to warrant "centered"'s use as a prop. Otherwise more function signatures will just be populated to add the class than if we were just focusing on the non-centered instances. I think the default state of !subtitle is an appropriate middleground, and I have already implemented this.

src/pages/overview/ProjectsCard.tsx
Kxiru: @sophiecarroll , Kim-Anh posted the spec vs current implementation of the Projects card above. You have already approved the design but I just wanted to bring it to your attention one more time to confirm. As the projects card is significantly wider than the instance / storage pages, the margins that worked for the last two cards don't work for the projects card (same would go for any longer card). In short, a box that is 68% the sizeof it's parent box (as is the case currently) appears differently for short vs long cards.

If you would like to match the spec exactly design wise, I can look into custom scss for the projects card. Alternatively, if you are happy with it as is (see the third screenshot in the description), then that is fine too.

src/pages/overview/PermissionsCard.tsx
kimanhou: All the cards are centered except Permissions and you should not use CardEmptyState for Permissions, so "centered" props is useless.

src/sass/_overview.scss
```diff
@@ -262,3 +257,32 @@
...
+
+  @include large {
+    width: 75%;
+  }
+
+  @include extra-large {
+    width: 68%;
+  }
+
+  &__full-width {
```
kimanhou, line 272: Is this used ?

src/pages/overview/ProjectsCard.tsx
kimanhou: FYI this is how it currently looks when all cards are in their empty state:
large screen
image

small screen

https://github.com/user-attachments/assets/73000578-ee61-46e7-8f38-1d62fad12853



The discrepancy between @sophiecarroll 's design and @Kxiru 's implementation is:
- in the design, text is centered in the card, ie same margin left and right.
- in the implementation, the width of the text and the margins depend on the window size:
    - on medium and small screens (< 1090px), the layout of the Overview page is a one-column layout, the width of the text is 100% of the parent's width (minus padding)  and the text is aligned left, except for Warnings where it's centered.
    - on large screens (> 1090px), the width of the text is 75% of the parent's width. It means that the left margin is equal to 12.5% of the parent's width. Same for the right margin but it doesn't always appear as such, because the text doesn't always take the full availabe width.
    - on extra large screens (> 1681px), the width of the text is 68% of the parent's width -> margin left is 16%

src/sass/_overview.scss
Kxiru: Yes

src/pages/overview/PermissionsCard.tsx
Kxiru: If anything, that supports the fact that centered should **not** be explicitly required in "all the cards", and instead only passed as an exception for the Permissions card to avoid duplication of code that can just be included in the component.

PR #2308, opened 2026-09-17
feat(profile) allow copy profile to other project and project refresh

### LINKED ISSUE #2283: profile copy --refresh --target-project
Hey!

Can I request this feature from the cli for the UI?

https://canonical.com/lxd/docs/latest/reference/manpages/lxc/profile/copy/

The refresh option and the target-project option.

The name should be able to stay the same if going to another project.

edlerd:
## Done

- feat(profile) allow copy profile to other project and project refresh

Fixes #2283

## QA

1. Run the LXD-UI:
    - go to an existing profile
    - copy it
    - copy it to another project
    - copy it over an existing profile and check the new "refresh existing" checkbox

## Screenshots

Screenshot from 2026-09-17 12-47-28


src/pages/profiles/forms/CopyProfileForm.tsx
```diff
@@ -51,35 +63,53 @@ const CopyProfileForm: FC<Props> = ({ profile, close }) => {
     toastNotify.success(message, actions);
   };
 
-  const getCopiedProfileName = (oldProfileName: string): string => {
-    const newProfileName = truncateEntityName(oldProfileName, "-copy");
-    return getUniqueResourceName(newProfileName, profiles);
+  const copyProfile = async (values: LxdProfileCopy) => {
```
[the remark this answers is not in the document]
omarelkashef: Good idea to add test coverage here or in a follow-up ticket

src/pages/profiles/forms/CopyProfileForm.tsx
```diff
@@ -1,44 +1,56 @@
...
+    project: string,
+    copyProfileRefresh: boolean,
+  ) => {
     const profileURL = `${ROOT_PATH}/ui/project/${encodeURIComponent(project)}/profile/${encodeURIComponent(name)}`;
+    const verb = copyProfileRefresh ? "Refreshed" : "Created";
+
     const message = (
       <>
-        Created profile{" "}
+        {verb} profile{" "}
```
omarelkashef, line 61: Should this be Creation/refresh started instead of created/refreshed since it is  operation?

src/pages/profiles/forms/CopyProfileForm.tsx
```diff
@@ -118,11 +161,38 @@ const CopyProfileForm: FC<Props> = ({ profile, close }) => {
...
+          }
+        />
+        <CheckboxInput
+          {...formik.getFieldProps("isRefresh")}
+          checked={formik.values.isRefresh}
+          onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+            formik.setFieldValue("isRefresh", e.target.checked);
+          }}
+          label="Refresh existing profile"
+          disabled={!targetExists}
```
omarelkashef: We should probably add a title why it is disabled. 

Image

src/pages/profiles/forms/CopyProfileForm.tsx
```diff
@@ -51,35 +63,53 @@ const CopyProfileForm: FC<Props> = ({ profile, close }) => {
...
         .then(() => {
-          notifySuccess(values.profileName, profile.project ?? "");
+          notifySuccess(
+            values.profileName,
+            values.targetProject,
+            values.isRefresh,
+          );
         })
         .catch((e) => {
           toastNotify.failure("Profile copy failed.", e, profileLink);
```
omarelkashef: The message here should be generic as well.

src/pages/profiles/forms/CopyProfileForm.tsx
```diff
@@ -118,11 +161,38 @@ const CopyProfileForm: FC<Props> = ({ profile, close }) => {
...
+            };
+          })}
+        />
         <Input
           {...formik.getFieldProps("profileName")}
           type="text"
           label="New profile name"
-          error={formik.touched.profileName ? formik.errors.profileName : null}
+          error={
+            targetExists && !formik.values.isRefresh
```
omarelkashef: This gives an error after I type an existing profile name but before I have touched the isRefresh check. 

Image

src/pages/profiles/forms/CopyProfileForm.tsx
```diff
@@ -118,11 +161,38 @@ const CopyProfileForm: FC<Props> = ({ profile, close }) => {
...
+          id="targetProject"
+          label="Target project"
+          options={projects.filter(canCreateProfiles).map((project) => {
+            return {
+              label: project.name,
+              value: project.name,
+            };
+          })}
+        />
         <Input
```
omarelkashef: The current interaction forces the user to type in the profile name to refresh from memory. How about we move the isRefresh checkbox above and if it is selected, the profile name becomes a dropdown?

src/pages/profiles/forms/CopyProfileForm.tsx
```diff
@@ -90,6 +120,14 @@ const CopyProfileForm: FC<Props> = ({ profile, close }) => {
     },
   });
 
+  const { data: targetProfiles = [] } = useProfiles(
+    formik.values.targetProject,
+  );
```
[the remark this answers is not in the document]
omarelkashef: We should disable submitting while this is loading.

omarelkashef (commented):
Thank you for this PR.  A few comments below.

We should probably add a max width to the target project dropdown as projects names can get very long 

Image

sophiecarroll:
Looks good to me, is there a logic to the automatic order of profiles in the profiles list page? When the profile was copied over to another project, and I went to the profiles page of the destination project, I expected the profile-copy to be at the top (or bottom), then I tried to look for it alphabetically, but I couldn't immediately find the profile I copied, which I would probably want to use right after copying/creating it.

omarelkashef (commented):
The interaction looks much better now. A few comments remain:

Should the notification indicate that it is an operation? Maybe say refresh/copy started?

Image

We probably should not show the current profile in the dropdown if the current project is selected.

Image


The project dropdown should truncate long names 

Image

sophiecarroll:
> The interaction looks much better now. A few comments remain:
> 
> Should the notification indicate that it is an operation? Maybe say refresh/copy started?
> 
> Image
> We probably should not show the current profile in the dropdown if the current project is selected.
> 
> Image
> The project dropdown should truncate long names
> 
> Image

I agree with all of this

edlerd:
> Should the notification indicate that it is an operation? Maybe say refresh/copy started?

The success notification is triggered after the operation succeeded.

> We probably should not show the current profile in the dropdown if the current project is selected.

Ah yes, we should filter it out. Will add that.

> The project dropdown should truncate long names

I think relying on the browser here is the right thing to do. We are using the standard select component.

omarelkashef:
> 
> I think relying on the browser here is the right thing to do. We are using the standard select component.

But the long names stretch the dropdown indefinitely.

omarelkashef:
> > Should the notification indicate that it is an operation? Maybe say refresh/copy started?
> 
> The success notification is triggered after the operation succeeded.

Oh I assumed the operation had not finished because of the "1 operation progress" note at the bottom

edlerd:
> > I think relying on the browser here is the right thing to do. We are using the standard select component.
> 
> But the long names stretch the dropdown indefinitely.

I think that is correct and depends on the browser. If we truncate, we have no way to surface the full length with the standard select as it only supports one text label.

kimanhou:
I tried copying an existing profile but I could not select a target project. Is this expected ?

https://github.com/user-attachments/assets/20207f3f-eb94-4e61-9c66-911d8eac2e23

After creating a profile, should we jump to the profile list ? We currently stay on the same page, which is the original profile overview or configuration. When creating a profile, we do so from the profile list, which allows us to see the newly created profile appear. The user might be used to that visual feedback, and would probably want to interact with the newly creating profile. This is especially true when copying a profile to another project. Wdyt ?

edlerd:
> I tried copying an existing profile but I could not select a target project. Is this expected ?

I can't reproduce the error and this should work as expected with a standard select.

> After creating a profile, should we jump to the profile list ? We currently stay on the same page, which is the original profile overview or configuration. When creating a profile, we do so from the profile list, which allows us to see the newly created profile appear. The user might be used to that visual feedback, and would probably want to interact with the newly creating profile. This is especially true when copying a profile to another project. Wdyt ?

Yes, we should land on the target profile. I switched the logic on success to land in the target profile page.

kimanhou:
@edlerd 

> I can't reproduce the error and this should work as expected with a standard select.

It now works.

